PR #1091, opened 2017-09-26
[RFR] Function as children without permissions

### LINKED ISSUE #1009: Render <Resource /> from application's props
I would like to render <Resource /> which depends on application's props

For example, 
```
constructor(props) {
  super(props);
  this.state = {
    resources: []
  };
}

componentDidMount() {
  let self = this;
  setTimeout(() => {
      self.setState({ resources: ['posts'] });
  },3000);
}
```

and render <Admin /> by using this code

```
render() {
     return (
                     <Admin title='Post Admin' restClient={jsonServerRestClient('https://jsonplaceholder.typicode.com')}>
                {
                    this.state.resources.map(resource => {
                        return <Resource name={resource} list={PostList} />;
                    })
                }
            </Admin>
     )

}
```

`PostList` is the same one as tutorial

but it turns out with an error
screen shot 2560-09-07 at 6 23 42 pm

I did some investigation in List.js from your lib. I found `state.admin` doesn't have key which name `posts`

I would like to make a flexible admin page which is can adapt to various endpoint.

djhi:
Allows to use the function as child pattern in components that support it, without an authClient.
This function might return a promise.

Fixes #1009

- [x] Implementation
- [x] Documentation

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -428,6 +428,7 @@ const App = () => (
 ```
 
 ## `initialState`
+
 The `initialState` prop lets you pass preloaded state to Redux. See the [Redux Documentation](http://redux.js.org/docs/api/createStore.html#createstorereducer-preloadedstate-enhancer) for more details.
 
 ## `history`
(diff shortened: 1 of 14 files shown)


ThieryMichel:
Ok pour moi

PR #1097, opened 2017-09-27
[RFR] [Feature] can set autoHideDuration for Notification's

### LINKED ISSUE #911: Allow changing autoHideDuration for <Notification>
Right now the autoHideDuration is hardcoded to 4s https://github.com/marmelab/admin-on-rest/blob/master/src/mui/layout/Notification.js#L41

It would be nice to change it though.

ie. 4s is too little for ppl to read the notification sometimes so i want it longer

natrim:
resolves #911

the autoHideDuration can be set globally: `<Notification autoHideDuration={5000} />`
or even different for every message: `showNotification('you_fail', 'warning', 30000)`

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -334,3 +334,14 @@ export default connect(mapStateToProps, {
     setSidebarVisibility: setSidebarVisibilityAction,
 })(MyLayout);
 ```
+
+
+## Notification's
+
+If you use your own layout (or custom login page), then you have access to the `Notification` component.
+
+Here you can set the `autoHideDuration` prop (defaults to 4000, aka 4 seconds).
+
+Using the `autoHideDuration` you can set the delay after which the notification closes.
+
+*TIP*: if you use the `redux` `showNotification` action, then you can define `autoHideDuration` per message as the third parameter to `showNotification` action.
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -336,12 +336,14 @@ export default connect(mapStateToProps, {
 ```
 
 
-## Notification's
+## Notifications
 
-If you use your own layout (or custom login page), then you have access to the `Notification` component.
+If you use your own layout (or custom login page), then you probably use the `<Notification>` component.
 
-Here you can set the `autoHideDuration` prop (defaults to 4000, aka 4 seconds).
+You can override the notification duration by setting the `autoHideDuration` prop. It defaults to 4000, i.e. 4 seconds.
 
-Using the `autoHideDuration` you can set the delay after which the notification closes.
+```jsx
+<Notification autoHideDuration={5000} />
+```
 
-*TIP*: if you use the `redux` `showNotification` action, then you can define `autoHideDuration` per message as the third parameter to `showNotification` action.
+**Tip**: if you use the `showNotification` action, then you can define `autoHideDuration` per message as the third parameter of the `showNotification` action creator.
--- a/src/actions/notificationActions.js
+++ b/src/actions/notificationActions.js
@@ -1,10 +1,6 @@
 export const SHOW_NOTIFICATION = 'AOR/SHOW_NOTIFICATION';
 
-export const showNotification = (
-    text,
-    type = 'info',
-    autoHideDuration = undefined
-) => ({
+export const showNotification = (text, type = 'info', autoHideDuration) => ({
     type: SHOW_NOTIFICATION,
     payload: { text, type, autoHideDuration },
 });
--- a/src/actions/notificationActions.js
+++ b/src/actions/notificationActions.js
@@ -1,8 +1,12 @@
 export const SHOW_NOTIFICATION = 'AOR/SHOW_NOTIFICATION';
 
-export const showNotification = (text, type = 'info') => ({
+export const showNotification = (
+    text,
+    type = 'info',
+    autoHideDuration = undefined
+) => ({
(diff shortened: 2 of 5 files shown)


src/actions/notificationActions.js
```diff
@@ -1,8 +1,12 @@
 export const SHOW_NOTIFICATION = 'AOR/SHOW_NOTIFICATION';
 
-export const showNotification = (text, type = 'info') => ({
+export const showNotification = (
+    text,
+    type = 'info',
+    autoHideDuration = undefined
```
fzaninotto: a default value of `undefined` makes no sense. Please remove the default value.

src/reducer/admin/notification.js
```diff
@@ -6,12 +6,17 @@ import {
 const defaultState = {
     text: '',
     type: 'info', // one of 'info', 'confirm', 'warning'
+    autoHideDuration: undefined, // undefined means use global duration
```
fzaninotto: This is where you should put the default duration of 4000

src/actions/notificationActions.js
natrim: it's to explicitly show that is should be undefined

src/reducer/admin/notification.js
natrim: if you put it here it will override the global one set in prop though

natrim:
theres no documentation where to add it though
i did not manage to find any docs for `Notification`

fzaninotto:
You're right, I suggest adding a section in the Theming chapter.

natrim:
ok, tried adding some simple one

src/actions/notificationActions.js
natrim: removed then

docs/Theming.md
```diff
@@ -334,3 +334,14 @@ export default connect(mapStateToProps, {
     setSidebarVisibility: setSidebarVisibilityAction,
 })(MyLayout);
 ```
+
+
+## Notification's
+
+If you use own layout (or custom login page), then you have access to the `Notification` component.
```
djhi: s/you use own/you use your own/

natrim: replaced

src/reducer/admin/notification.js
fzaninotto: gotcha

src/reducer/admin/notification.spec.js
```diff
@@ -7,22 +7,29 @@ import reducer from './notification';
...
         assert.deepEqual(
-            { text: 'foo', type: 'warning' },
+            { text: 'foo', type: 'warning', autoHideDuration: undefined },
             reducer(undefined, {
                 type: SHOW_NOTIFICATION,
-                payload: { text: 'foo', type: 'warning' },
+                payload: {
+                    text: 'foo',
+                    type: 'warning',
+                    autoHideDuration: undefined,
```
fzaninotto: you don't need to add the payload here

src/reducer/admin/notification.spec.js
```diff
@@ -7,22 +7,29 @@ import reducer from './notification';
...
             })
         );
     });
     it('should set text to empty string upon HIDE_NOTIFICATION', () => {
         assert.deepEqual(
-            { text: '', type: 'warning' },
+            { text: '', type: 'warning', autoHideDuration: undefined },
             reducer(
-                { text: 'foo', type: 'warning' },
+                { text: 'foo', type: 'warning', autoHideDuration: undefined },
```
fzaninotto: no need to add the payload here either

src/reducer/admin/notification.spec.js
```diff
@@ -7,22 +7,29 @@ import reducer from './notification';
 
 describe('notification reducer', () => {
     it('should return empty notification by default', () => {
```
fzaninotto: you should add one test showing that a custom `autoHideDuration` in the payload updates the state accordingly

src/reducer/admin/notification.js
```diff
@@ -6,12 +6,17 @@ import {
...
     switch (type) {
         case SHOW_NOTIFICATION:
-            return { text: payload.text, type: payload.type };
+            return {
+                text: payload.text,
+                type: payload.type,
+                autoHideDuration: payload.autoHideDuration,
+            };
         case HIDE_NOTIFICATION:
             return { ...previousState, text: '' };
```
fzaninotto: I think this misses a `autoHideDuration: undefined`, otherwise a single notification with a specific duration will override the duration of all following notifications.

src/reducer/admin/notification.js
fzaninotto: Replace "undefined  means use global duration" by "use default duration from Notification"

src/reducer/admin/notification.js
natrim: nope, because every call of showNotification action set's new value or undefined
same as type

src/reducer/admin/notification.spec.js
natrim: ok, added

src/reducer/admin/notification.spec.js
natrim: removed

src/reducer/admin/notification.spec.js
natrim: removed

fzaninotto:
Thanks!

PR #1098, opened 2017-09-27
[RFR] [Feature] SimpleList redirect prop

### LINKED ISSUE #926: Allow changing what action happens on click in <SimpleList>
The `SimpleList ` right now redirects to Edit on click.

It would be nice to be able change the action (or even disable clicking).

(ie. i wanna it go to Show not Edit)

edit: probably something like:

```
import React from 'react';
import { List, ListItem } from 'material-ui/List';

const tertiaryStyle = { float: 'right', opacity: 0.541176 };

const SimpleList = ({
    ids,
    data,
    basePath,
    primaryText,
    secondaryText,
    secondaryTextLines,
    tertiaryText,
    leftAvatar,
    leftIcon,
    rightAvatar,
    rightIcon,
    action,
}) => (
    <List>
        {ids.map(id => (
            <ListItem
                key={id}
                primaryText={
                    <div>
                        {primaryText(data[id], id)}
                        {tertiaryText &&
                            <span style={tertiaryStyle}>
                                {tertiaryText(data[id], id)}
                            </span>
                        }
                    </div>
                }
                secondaryText={secondaryText && secondaryText(data[id], id)}
                secondaryTextLines={secondaryTextLines}
                leftAvatar={leftAvatar && leftAvatar(data[id], id)}
                leftIcon={leftIcon && leftIcon(data[id], id)}
                rightAvatar={rightAvatar && rightAvatar(data[id], id)}
                rightIcon={rightIcon && rightIcon(data[id], id)}
                containerElement={action === 'none' ? 'span' : <Link to={`${basePath}/${id}${action ? `/${action}` : ''}`} />}
            />
        ))}
    </List>
);

SimpleList.propTypes = {
    ids: PropTypes.array,
    data: PropTypes.object,
    basePath: PropTypes.string,
    primaryText: PropTypes.func,
    secondaryText: PropTypes.func,
    secondaryTextLines: PropTypes.number,
    tertiaryText: PropTypes.func,
    leftAvatar: PropTypes.func,
    leftIcon: PropTypes.func,
    rightAvatar: PropTypes.func,
    rightIcon: PropTypes.func,
    action: PropTypes.string,
};

SimpleList.defaultProps = {
    action: '', // or default to 'none'?
};

export default SimpleList;
```

natrim:
to choose where to redirect on click
resolves #926

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -430,7 +430,7 @@ export const PostList = (props) => (
 );
 ```
 
-**Tip**: You can set the `linkType` prop to `"show"` to link to the `<Show>` page instead.
+**Tip**: The `<SimpleList>` items link to the edition page by default. You can set the `linkType` prop to `show` to link to the `<Show>` page instead.
 
 ```jsx
 // in src/posts.js
@@ -449,24 +449,7 @@ export const PostList = (props) => (
 );
 ```
 
-If you set the `linkType` prop to `false` (boolean, not string), it will remove the link.
-
-```jsx
-// in src/posts.js
-import React from 'react';
-import { List, SimpleList } from 'admin-on-rest';
-
-export const PostList = (props) => (
-    <List {...props}>
-        <SimpleList
-            primaryText={record => record.title}
-            secondaryText={record => `${record.views} views`}
-            tertiaryText={record => new Date(record.published_at).toLocaleDateString()}
-            linkType={false}
-        />
-    </List>
-);
-```
+Setting the `linkType` prop to `false` (boolean, not string) removes the link in all list items.
 
 ## The `<SingleFieldList>` component
 
--- a/docs/List.md
+++ b/docs/List.md
@@ -430,6 +430,44 @@ export const PostList = (props) => (
 );
 ```
 
+**Tip**: You can set the `linkType` prop to `"show"` to link to the `<Show>` page instead.
+
+```jsx
+// in src/posts.js
+import React from 'react';
+import { List, SimpleList } from 'admin-on-rest';
+
+export const PostList = (props) => (
+    <List {...props}>
+        <SimpleList
+            primaryText={record => record.title}
+            secondaryText={record => `${record.views} views`}
+            tertiaryText={record => new Date(record.published_at).toLocaleDateString()}
+            linkType="show"
+        />
+    </List>
+);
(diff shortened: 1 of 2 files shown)


docs/List.md
```diff
@@ -430,6 +430,44 @@ export const PostList = (props) => (
 );
 ```
 
+**Tip**: You can set what route the click on list item redirects to by setting the `redirect` prop.
+
```
djhi: Please specify all the possible values and their effects

natrim: it accept's the same as redirect props on edit and create

added the enums to doc

fzaninotto:
The `redirect` name is incorrect, there is no redirection in that case. 

Besides, I'd prefer if you let the user override the `containerElement` altogether. So I suggest introducing an `itemContainerElement` prop instead of `redirect`.

natrim:
@fzaninotto well i wanted to make it simple and similar to other components

https://github.com/marmelab/admin-on-rest/blob/master/src/mui/field/ReferenceField.js
is even simpler, with limited `linkType `

should i remake this pull with the `itemContainerElement ` ?

natrim:
hm, i could probably just rename it to some other name and allow Element type

so the user could pass string for link
false for disabling
and element for replacing the container

fzaninotto:
You're right, `linkType` already exists and it makes sense to reuse the same convention.

natrim:
@fzaninotto should i the rename the `redirect` to `linkType` then ?
or some other change?

fzaninotto:
renaming is fine. Just check that the semantics are the same as with the `ReferenceField` component.

natrim:
what do you mean with `semantics`?

`ReferenceField ` accepts only: show, edit, true, false
    and defaults to edit

`SimpleList ` defaults to show
   - with the pull it accepts all redirectsTo supports
     - not 'true' though

fzaninotto:
I mean that `SimpleList` should also accept `show`, `edit`, `true`, and `false`. It can default to `show`, though.

natrim:
@fzaninotto renamed and changed to work same as in `ReferenceField `

fzaninotto:
Thanks!

natrim:
edition page?
i think it would be better to say: editing page

Issue #896, opened 2017-07-26
[RFR] Remove `addField` prop in Input components


fzaninotto:
## How `addField` currently works

Currently, all the aor Input components create their own redux-form `<Field>` component. Or rather, they express that they want to be decorated by a redux-form `<Field>`, thanks to the `addField: true` default Prop. For instance:

```js
// in LongTextInput.js
// simplified for this explanation
const LongTextInput = ({ input, isRequired, label, meta: { touched, error }, options, source, elStyle, resource }) => (
    <TextField {...input} multiLine fullWidth errorText={touched && error} style={elStyle}
        floatingLabelText={<FieldTitle label={label} source={source} resource={resource} isRequired={isRequired} />}
        {...options}
    />
);

LongTextInput.propTypes = {
    addField: PropTypes.bool.isRequired,
    // ...
};

LongTextInput.defaultProps = {
    addField: true,
    // ...
};

export default LongTextInput;
```

What happens is that `<SimpleForm>` and `<TabbedForm>` iterate over all their children, and pass them to a `<FormField>` component as the `input` prop:

```js
// in SimpleForm.js
// simplified for this explanation
export class SimpleForm extends Component {
    render() {
        const { children, invalid, record, resource, basePath } = this.props;
        return (
            <form>
                <div>
                    {Children.map(children, input => input && (
                        <div key={input.props.source}>
                            <FormField input={input} resource={resource} record={record} basePath={basePath} />
                        </div>
                    ))}
                </div>
            </form>
        );
    }
}
```

Then `<FormField>` inspects its `input` prop props. If the `addField` prop is detected, FormField decorates the input with a redux-form field.

```js
// in FormField.js
// simplified for this explanation
const FormField = ({ input, ...rest }) => {
    if (input.props.addField) {
        return (
            <Field
                {...rest}
                {...input.props}
                name={input.props.source}
                component={input.type}
                isRequired={isRequired(input.props.validate)}
            />
        );
    }
    return (typeof input.type === 'string') ? input : React.cloneElement(input, rest);
};
```

The great advantage of this is that we don't have to repeat the `<Field>` decoration for every input component. 

## The problem

Every time a developer decorates one of aor components, e.g. to customize the style, or set a prop always the same, like so:

```js
const RedLongTextInput = (props) => <div style={{ backgroundColor: 'red' }}>
    <LongTextInput {...props} />
</div>
```

They get this horrible error:

> Cannot read property 'touched' of undefined

[...]

PR #1129, opened 2017-10-04
[RFR] [BC Break] Remove addField prop

fzaninotto:
- [x] Create an `addField` HOC
- [x] Unit test the `addField` HOC
- [x] Make core input components work without `addField` and `addLabel` props
- [x] Update documentation for custom inputs
- [x] Find a solution for fields used as inputs (who all have `addLabel` turned on)
- [x] Remove the `<FormField>` component
- [x] Write the Upgrade guide

Note: this PR temporarily disables the `<RichTextInput>` call, because that external package needs to be updated.

Refs #896


djhi:
If we're going to force input implementations to explicitly state they want a label, why don't we document they have to wrap their input with the `FormField` directly ?

example/posts.js
```diff
@@ -44,7 +44,7 @@ import {
     required,
     translate,
 } from 'admin-on-rest'; // eslint-disable-line import/no-unresolved
-import RichTextInput from 'aor-rich-text-input';
+// import RichTextInput from 'aor-rich-text-input';
```
djhi: Did you forget to uncomment ?

src/mui/form/addField.js
```diff
@@ -0,0 +1,56 @@
...
+                [source]:
+                    typeof defaultValue === 'function'
+                        ? defaultValue()
+                        : defaultValue,
+            });
+        }
+
+        render() {
+            const { input, validate, ...props } = this.props;
+            return input ? ( // An ancestor is already decorated by Field
```
djhi: :+1:

src/mui/input/FileInput.js
```diff
@@ -39,8 +43,6 @@ export class FileInput extends Component {
     };
 
     static defaultProps = {
-        addLabel: true,
```
djhi: Shouldn't it have a default `addLabel: false` ?

example/posts.js
fzaninotto: No, see the PR description: If I leave it uncommented, the example breaks because the `RichRextInput` hasn't been updated to use the `addField` HOC.

src/mui/input/FileInput.js
fzaninotto: `addLabel` is no longer used in Input components, that's why I added a `<Labeled>` component in the `render()` method.

example/posts.js
djhi: Oh sorry about that

PR #1140, opened 2017-10-09
withWidth, use constants instead of magic numbers

oliviertassinari:
The width constants seem not to be used in the aor codebase. Is this on purpose?

### DIFF
--- a/src/auth/withPermissionsFilteredChildren.js
+++ b/src/auth/withPermissionsFilteredChildren.js
@@ -42,21 +42,19 @@ export default BaseComponent => {
             const { authClient, ...props } = this.props;
             return (
                 <BaseComponent {...props}>
-                    {children && Array.isArray(children) ? (
-                        children.map(
-                            child =>
-                                child
-                                    ? cloneElement(child, {
-                                          key:
-                                              child.props.name ||
-                                              child.props.source ||
-                                              child.props.label,
-                                      })
-                                    : null
-                        )
-                    ) : (
-                        children
-                    )}
+                    {children && Array.isArray(children)
+                        ? children.map(
+                              child =>
+                                  child
+                                      ? cloneElement(child, {
+                                            key:
+                                                child.props.name ||
+                                                child.props.source ||
+                                                child.props.label,
+                                        })
+                                      : null
+                          )
+                        : children}
                 </BaseComponent>
             );
         }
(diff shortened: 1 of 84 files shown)


fzaninotto:
You're absolutely right. But you need to confirm to *prettier* code formatting. May I suggest you turn eslint auto-formatting on save in your IDE?

oliviertassinari:
@fzaninotto This change was done with the code editor of GitHub. I will replace all the occurrences.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
fzaninotto: Please revert the prettier-related changes, these should not be in this PR.

src/auth/withPermissionsFilteredChildren.js
```diff
@@ -42,21 +42,19 @@ export default BaseComponent => {
             const { authClient, ...props } = this.props;
             return (
                 <BaseComponent {...props}>
-                    {children && Array.isArray(children) ? (
```
fzaninotto: please revert this change, as well as all the changes due to prettier configuration change

src/mui/input/CheckboxGroupInput.js
```diff
@@ -128,21 +128,17 @@ export class CheckboxGroupInputComponent extends Component {
             <Checkbox
                 key={get(choice, optionValue)}
                 checked={
-                    value ? (
```
fzaninotto: please revert changes in this file

.prettierrc
oliviertassinari: There is not package lock of the dependencies. Feel free to amand my PR. As having the write access on the repository, you have write access to my fork's branch.

oliviertassinari: Also I expect changing any part of the package.json invalidating travis cache and making all consecutives build red.

fzaninotto: Not sure I understand what you mean. Do you say you had to change the prettier configuration because the prettier version wasn't fixed in the package, and prettier updated their defaults in a new version?

If that's the case, you should just remove the `~` in the prettier version in the `package.json`.

oliviertassinari:
I'm giving up. I have tried with `"prettier": "1.5.2",` the diff is even worse. I have no clue about the prettier version used to format the source code.

fzaninotto:
It should be easier to do in the `next` branch now.

PR #1147, opened 2017-10-12
Blog example with dot.notation reference

Phocea:
The blog sample is lacking an example where the dot notation is used to reference another resources.
This PR modify the data.json to include such an example

### DIFF
--- a/example/comments.js
+++ b/example/comments.js
@@ -36,7 +36,7 @@ export CommentIcon from 'material-ui/svg-icons/communication/chat-bubble';
 
 const CommentFilter = ({ ...props }) => (
     <Filter {...props}>
-        <ReferenceInput source="post_id" reference="posts">
+        <ReferenceInput source="post.id" reference="posts">
             <SelectInput optionText="title" />
         </ReferenceInput>
     </Filter>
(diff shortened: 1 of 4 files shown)


fzaninotto:
I like the idea, but not the field you chose to illustrate the dot notation, because I never saw such cases in real life (keeping a reference in an object). Can't you use `post.pictures.first` instead?

Phocea:
Funny enough all of our API are like this for reference ID's, because the reference id can be through one id or through 2 single id's.. (not my design). This explains why we have the reference in an object. ReferenceField works fine with it though :)

However, `post.pictures.first` is using the upload feature and not a  `ReferenceField`, and pictures are not an independent resource.
Also, Changing comment -> post relationship removed the nominal `ReferenceField `case from the example.

Could we add a reviewer to the posts and link it to the declared user as an object `reviewer :  { id: 1}` ?

fzaninotto:
So you want to illustrate dot notation specifically on a `ReferenceField`? What's so special about it?

Phocea:
Just to show it's possible really. On field, input and also filters.
Its actually not working in our codebase, so I had to replicate it on the example blog to make sure it was working as expected. Problem is actully with our custom rest client, but thought it might be useful to keep the use case in the blog example.

If not, thats fine. I have a fork of it :)

fzaninotto:
Gotcha. But I want to keep the example small and easy to work on, so it can't showcase all features (or it would be doubling down on the documentation). Maybe try on admin-on-rest-demo.

PR #1149, opened 2017-10-12
[RFR] Fix Refresh in Edit and Show views

### LINKED ISSUE #1131: Refresh button does not refresh data in <ReferenceManyField>

**What you were expecting:**

Refresh button should refresh data in ReferenceManyField

**What happened instead:**

Data in ReferenceManyField is not refreshed (no API call) when Refresh button click

**Steps to reproduce:**

Add ReferenceManyField with reference="xyz" target="zyx", Datagrid, and TextField. Click on Refresh button.

fzaninotto:
The Refresh button didn't force refetch of related records (e.g. in `<ReferenceManyField>`).

Closes #1131

### DIFF
--- a/src/mui/detail/TabbedShowLayout.js
+++ b/src/mui/detail/TabbedShowLayout.js
@@ -1,12 +1,9 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import compose from 'recompose/compose';
 import { Tabs, Tab } from 'material-ui/Tabs';
-import getDefaultValues from '../form/getDefaultValues';
 
 const divStyle = { padding: '0 1em 1em 1em' };
-export class TabbedShowLayout extends Component {
+class TabbedShowLayout extends Component {
     constructor(props) {
         super(props);
         this.state = {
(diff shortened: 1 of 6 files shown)


src/mui/detail/TabbedShowLayout.js
```diff
@@ -67,16 +65,11 @@ TabbedShowLayout.propTypes = {
...
     basePath: PropTypes.string,
     translate: PropTypes.func,
+    version: PropTypes.number,
 };
 
 TabbedShowLayout.defaultProps = {
     contentContainerStyle: { borderTop: 'solid 1px #e0e0e0' },
 };
 
-const enhance = compose(
```
fzaninotto: This probably came from a bad copy/paste of the `TabbedFormLayout`. There is no `initialValues` used in the component.

qstyler:
This refresh doesn't really work in the following case.

I have a <ReferenceManyField /> that obviously displays some related records. I have an action that removes one of the related models. I see that request to the API is fired. But the number of the records remain the same.

fzaninotto:
@qstyler Your sentence is missing some words, please rephrase.

> I have a that obviously displays some related records.

qstyler:
@fzaninotto  Sorry. I must have been very sleepy when I wrote this. I have created another issue for that. #1648

PR #835, opened 2017-07-06
[RFR] Introduce selector functions to query the state

lutangar:
This is a follow up of #624.

### Selectors
Selectors are *getter for the state*:
1. Decouple data selection code from component itself, thus decoupling `state` structure from components: the only layer between the state tree and the connected components being the selectors.
2. Ease state refactoring, as the choice here was to *colocate* selectors with there reducers. The reducer slices and their selectors are entirely independent from other bites.

Most of the selectors are simple functions. Some selectors are querying 2 different parts of the state and are using [reselect](https://github.com/reactjs/reselect).

#### Synthax
Selectors are using the following syntax:
```
const getResource = (state, { resource }) => getResources(state)[resource];
```

1. Always put state argument in the last position.
* Selector can be used as a `createSelector` argument when building selectors with reselect:
```
 createSelector(
  getResource,
  posts => ...
)
```

* Selector can be used as an argument of `select` in `redux-saga`:
```
yield(select(getResource, { resource: 'posts' }));
```

2. Put selectors function that perform some logic at the smaller bite possible to reduce complexity and avoid unnecessary coupling:

Before:
https://github.com/larriereguichet/admin-on-rest/blob/next/src/reducer/admin/references/oneToMany.js#L32-L41

After:
https://github.com/larriereguichet/admin-on-rest/blob/533460447c6688c2e6cb87b82cd6c291720dc232/src/reducer/admin/resource/data.js#L104-L111

3. If a selector depends on 2 different state slices, again, find the smallest slice that contains both of the needed slice.
ex: https://github.com/larriereguichet/admin-on-rest/blob/533460447c6688c2e6cb87b82cd6c291720dc232/src/reducer/admin/index.js#L54-L69

#### Tests
Tests were added but I'm not convinced of the relevance of some, any suggestions on the way to go are more than welcome.

PR #1151, opened 2017-10-12
[RFR] Introduce selector functions to query the state

fzaninotto:
Supersedes #835


fzaninotto:
Not good enough, reopening #835 to discuss it with PR author.

PR #1159, opened 2017-10-17
[RFR] Upgrade to React 16

fzaninotto:
- [x] upgrade to React 16
- [x] upgrade to enzyme 3
- [x] Fix tests that break
- [x] Fix warnings in tests


packages/ra-input-rich-text/package.json
```diff
@@ -30,8 +30,8 @@
     },
     "homepage": "https://github.com/marmelab/react-admin#readme",
     "peerDependencies": {
-        "react": "^15.4.0 || ^16.0.0-beta.5",
-        "react-dom": "^15.4.0 || ^16.0.0-beta.5"
+        "react": "~16.0.0",
```
djhi: Are you sure about this ? Shouldn't we use `^` and ranges like before ?

djhi:
Looks like you missed a warning though:
> Warning: Failed prop type: The prop `refreshView` is marked as required in `List`, but its value is `undefined`.

in packages/react-admin/src/mui/list/List.spec.js

fzaninotto:
No, it will be fixed in another PR - removing a deprecated feature

PR #1160, opened 2017-10-17
[RFR] [BC Break] Remove deprecated list refresh method

fzaninotto:


UPGRADE.md
```diff
@@ -177,3 +177,7 @@ const choices = [
 const SexInput = props => <SelectInput {...props} choices={choices}/>;
 export default SexInput;
 ```
+
+## No More `refresh` method On `<List>`
+
+The Refresh button now uses Redux to force a refetch of the data. As a consequence, the various views (List, Show, Edit) don't need to pass an action creator down to the Refresh button. We've removed the `refresh` method of the `<List>` component - but the refresh button is still there and works fine. It's just that, if you relied on the presence of that method, you should now dispatch a `refreshView` action creator by hand.
```
djhi: An example might be nice here

UPGRADE.md
```diff
@@ -178,6 +178,40 @@ const SexInput = props => <SelectInput {...props} choices={choices}/>;
 export default SexInput;
 ```
 
-## No More `refresh` method On `<List>`
+## No More `refresh` Prop Passed To `<List>` Actions
 
-The Refresh button now uses Redux to force a refetch of the data. As a consequence, the various views (List, Show, Edit) don't need to pass an action creator down to the Refresh button. We've removed the `refresh` method of the `<List>` component - but the refresh button is still there and works fine. It's just that, if you relied on the presence of that method, you should now dispatch a `refreshView` action creator by hand.
+The Refresh button now uses Redux to force a refetch of the data. As a consequence, the List view no longer passes the `refresh` prop to the `<Actions>` component. If you relied on that prop to refresh the list, you must now use the new `<RefreshButton>` component.
+
+```jsx
```
djhi: I was thinking about a custom button using the new `refreshView` action. Maybe it should appear somewhere in the documentation but not in the upgrade guide

fzaninotto: I prefer to emphasize the use of a component rather than trying to explain the intricacies of Redux connect and dispatch. Here, it's much simpler.

djhi: Understood but we'll have to document it anyway because, imo, a lot of custom actions will require to refresh the data once they succeeded

PR #1170, opened 2017-10-20
ImageField: allow to style the img tag

alanpoulain:
Useful for example if you want to change the size of the image:
```javascript
<ImageField key="imageLink" source="imageLink" elStyle={{image: {height: 40}}} />
``` 
If you are okay with this, I could add tests and/or docs.

### DIFF
--- a/packages/react-admin/src/mui/field/ImageField.js
+++ b/packages/react-admin/src/mui/field/ImageField.js
@@ -19,6 +19,11 @@ export const ImageField = ({ elStyle = {}, record, source, src, title }) => {
         return <div />;
     }
 
+    const imageStyle = {
+        ...styles.image,
+        ...elStyle.image,
+    };
+
     if (Array.isArray(sourceValue)) {
         const style = {
             ...styles.list,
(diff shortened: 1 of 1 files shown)


Simperfit:
ping @fzaninotto, this is blocking for us, could you please look at it if you have some time ?

fzaninotto:
This is a new feature, so it should be PRed agains the `next` branch.

alanpoulain:
@fzaninotto it's done.

djhi:
Now that we use MaterialUI v1 and JSS on `next`, the `ImageField` can customized using:

```jsx
const MyImageField = withStyles({ image: { height: 40 })(ImageField);
MyImageField.defaultProps = ImageField.defaultProps;
```

ricardovf:
This is how im doing:

```js
const styles = {
  image: { width: "50px" }
};

const AvatarField = withStyles(styles)(({ classes, ...props }) => (
  <ImageField classes={classes} source="frameUrl" label="Frame" {...props} />
));

<AvatarField />
```

PR #1177, opened 2017-10-24
AutocompleteInput does not render correctly first time

BenjaminFaal:
When wrapped in a ReferenceInput it does not render the value the first time


fzaninotto:
Please provide a failing test case, I can't reproduce the bug.

BenjaminFaal:
Im sorry, it happens when i use it within a EmbeddedArrayInput from https://github.com/MhdSyrwan/aor-embedded-array, i guess i will open a issue there.

PR #1178, opened 2017-10-24
[RFR] Minor copywriting updates

netsgnut:
This PR proposes two small updates regarding `src/i18n/messages.js`.

1. Removed space before `?` for `aor.message.are_you_sure`.
2. Moved the trailing `.` from `src/mui/layout/NotFound.js` to `aor.message.not_found` instead to allow proper i18n for languages that don't typically use `.` for sentence ends, e.g. Japanese and Chinese.

### DIFF
--- a/src/i18n/messages.js
+++ b/src/i18n/messages.js
@@ -42,7 +42,7 @@ export default {
         message: {
             yes: 'Yes',
             no: 'No',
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

netsgnut:
No problem. Thanks for the awesome library ;)

PR #1187, opened 2017-10-29
Update index.md

faviouz:
The documentation's readme suggests installing via npm as a dev dependency.

```
npm install --save-dev admin-on-rest
```

It should not be a dev dependency, `--save` should have been used instead. As of npm5, that can actually even be omitted. [`npm install` will `--save` by default](http://blog.npmjs.org/post/161081169345/v500).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -46,7 +46,7 @@ Admin-on-rest is available from npm. You can install it (and its required depend
 using:
 
 ```sh
-npm install --save-dev admin-on-rest
+npm install admin-on-rest
 ```
 
 ## Usage


faviouz:
On the other hand, the tutorial page suggests using yarn for installation. It may be in your interest to make it consistent.

fzaninotto:
Thanks

christiaanwesterbeek:
So, what's the preferred way? I'm new to aor. Not sure to start with this

```javascript
npm install -g create-react-app
create-react-app test-admin
cd test-admin/
yarn add admin-on-rest
yarn start
```

or 
```javascript
npm install admin-on-rest
```

Oh well, I'll probably learn soon enough...

djhi:
There's no preferred way.

If you want to get starting quickly, use the first method with `create-react-app`.

If you don't want to use `yarn` and have your project without `create-react-app`, use npm

PR #1191, opened 2017-11-01
Default List page param

Toosick:
When you have a query string that doesn't have a page param the Pagination component has `NaN-NaN of 1`. This defaults the page to 1 when you don't set it in your Link react router dom component.

See https://github.com/marmelab/admin-on-rest/issues/982


djhi:
Superseded by #1288

PR #1195, opened 2017-11-01
Small removal of unneeded template tag

surbas:
The tags don't seem needed based on rest of document. It was unrendered by github.


fzaninotto:
No, this is necessary when rendering the pages by Jekyll, as in https://marmelab.com/admin-on-rest/Authentication.html

PR #1202, opened 2017-11-03
pass props in DashBoardMenuItem component

AlexandrKnyazev1395:
I have just pass props to the next component.
(/issues/1201)


fzaninotto:
This is not a bug fix, so it has to be done against the `next` branch. And since we're deeply focused on the material-ui 1.0 migration, this likely won't be merged until the migration is complete.

djhi:
No news for some time and we're going to do it in next

PR #1204, opened 2017-11-05
add component factory and java rest building blocks to Ecosystem.md

zifnab87:
if you feel react-admin-java-rest should be under Miscellaneous that's fine too

### DIFF
--- a/docs/Ecosystem.md
+++ b/docs/Ecosystem.md
@@ -32,10 +32,6 @@ See the [translation](./Translation.md#available-locales) page.
 - [moonlight-labs/aor-jsonapi-client](https://github.com/moonlight-labs/aor-jsonapi-client): a REST client for [JSON API](http://jsonapi.org/)
 - [mhdsyrwan/aor-rest-client-router](https://github.com/MhdSyrwan/aor-rest-client-router): a REST client that enables routing to many REST clients based on resource name.
 
-## Rest servers
-
-- [zifnab87/react-admin-java-rest](https://github.com/zifnab87/react-admin-java-rest): building blocks for Java REST API Server bootstrapping customized for admin-on-rest using Java Spring Boot, Hibernate and JPA. Fully working example: [zifnab87/react-admin-demo-java-rest](https://github.com/zifnab87/react-admin-demo-java-rest)
-
 ## Miscellaneous
 
 - [marmelab/aor-realtime](https://github.com/marmelab/aor-realtime): enable realtime updates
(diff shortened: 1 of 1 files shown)


docs/Ecosystem.md
```diff
@@ -32,6 +32,11 @@ See the [translation](./Translation.md#available-locales) page.
 - [moonlight-labs/aor-jsonapi-client](https://github.com/moonlight-labs/aor-jsonapi-client): a REST client for [JSON API](http://jsonapi.org/)
 - [mhdsyrwan/aor-rest-client-router](https://github.com/MhdSyrwan/aor-rest-client-router): a REST client that enables routing to many REST clients based on resource name.
 
+## Rest servers
+
+- [zifnab87/react-admin-java-rest](https://github.com/zifnab87/react-admin-java-rest): building blocks for Java REST API Server bootstrapping customized for admin-on-rest using Java Spring Boot, Hibernate and JPA. Fully working example: [zifnab87/react-admin-demo-java-rest](https://github.com/zifnab87/react-admin-demo-java-rest)
```
fzaninotto: Nice, but how is that specific to aor? It's just a REST API provider, right?

zifnab87: Thanks! react-admin-java-rest uses Spring Boot's Criteria API to build automatically all the filtering/sorting queries that are needed from admin-on-rest. It helps developers expose ?filter=[...] for all the resources easily and wraps all the resources with paging information and does configures the serialization/deserialization. With this boilerplate code removed I was able to build the Java REST API back-end with MySQL for admin-on-rest-demo in about 4-5 hours. I am planning to extract an npm package that will include the minimally invasive restClient that is needed to work with this backend. it is still part of react-admin-demo-java-rest

fzaninotto: aor doesn't require any API signature. You'd better write a `restClient` for an existing API provider, rather than design an API for one of the (mostly educational) `restClient` implementations.

zifnab87: I wanted to stick to the defaults - the way that admin-on-rest-demo works. Of course I am going to make a new package that exposes a restClient to use react-admin-java-rest properly. Since initially we migrated from Drupal I had to find a quick way to bootstrap filtering and paging using Java similar to what that Drupal Views had (and they were painful after a certain extend). Do you feel it is not valuable to anyone?

fzaninotto: Yes, I feel it's not in the right direction.

zifnab87: ok I guess we can leave that for later. Please merge only ra-component-factory links then. Thank you

fzaninotto:
Thanks!

PR #1213, opened 2017-11-09
make auto complete possible to store inital values

vedmalex:
this make possible to setup initial value from list of choises in case when chioces are loaded in async way.
otherwise we see empty values, because it is chosen from empty initial load.


fzaninotto:
Any new feature must be PRed against the `next` branch. Also, please add a unit test.

djhi:
@vedmalex Any news on this one ?

vedmalex:
@djhi not yet... sorry... i've make local copy of files... and have relax :) i have more files to see but now too busy with current tasks

djhi:
Superseded by #1340

PR #1214, opened 2017-11-10
[RFR] Make the Layout component themeable again

fzaninotto:
To support theming, react-admin must accept a `theme` component in the `<Layout>` component. Due to the order in which mui 1.0 registers theme and width providers, this implies separating the width-dependent elements from the `Layout`.

Necessary to bring theming functionality to the demo.


packages/react-admin/src/mui/list/List.js
```diff
@@ -268,7 +269,9 @@ export class List extends Component {
                         </div>
                     ) : (
                         <CardContent style={styles.noResults}>
-                            {translate('ra.navigation.no_results')}
+                            <Typography type="body1">
```
fzaninotto: this is a "cavalier" change (it has nothing to see with the rest of the PR) but it's not worth a PR on its own.

PR #905, opened 2017-07-30
[RFR] Layoutless routes

marcw:
At this time, when using the `customRoutes` prop, the routes are rendered in the default Layout.
While this is the expected behavior, a user might need to add routes which are not rendered inside the default layout (i.e: password forgot dance, or user registration). The only way of doing that is to follow the [custom app tutorial](https://marmelab.com/admin-on-rest/CustomApp.html) which comes at other expenses (maintenance, loss of dev speed). 

At a quick glance on the code, it was pretty easy to add a `layoutlessRoutes` props which are rendered outside of the default Layout. This is the purpose of this PR. Feedbacks are welcome :)

PR #1218, opened 2017-11-12
Added customPages prop on Admin

cipriancaba:
This will allow the creation of completely custom pages within the Admin component

I often feel the need to add completely custom routes to the admin (eg: `/register`).. I cannot do this in the context of unauthenticated page, which means that the default layout would be visible

Therefore I am suggesting the addition of a simple change on the `Admin` component which will allow the creation of completely independent routes without the fuss of creating your own admin component

If this is something you would consider, I am open to feedback and willing to update the docs to reflect this change

### DIFF
--- a/src/Admin.js
+++ b/src/Admin.js
@@ -26,6 +26,7 @@ const Admin = ({
     customReducers = {},
     customSagas = [],
     customRoutes = [],
+    customPages = [],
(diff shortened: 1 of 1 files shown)


cipriancaba:
Of course this was already added in #905 
Sorry about the confusion

PR #1226, opened 2017-11-18
AutocompleteInput: translate is a required property

arlair:
AutocompleteInput: translate is a required property, but it is not defaulted?

I think this is meant to be defaulted to the imported translate? I saw property warnings in the console and this change removed the warnings.

### DIFF
--- a/src/mui/input/AutocompleteInput.js
+++ b/src/mui/input/AutocompleteInput.js
@@ -195,6 +195,7 @@ AutocompleteInput.defaultProps = {
     options: {},
     optionText: 'name',
     optionValue: 'id',
+    translate,
     translateChoice: true,
 };
 


djhi:
No, the `translate` function in props is supplied by the `translate` Higher Order Function which is used at last line of the file. Please open an issue with an codesandbox showcasing the warning.

arlair:
Ahh, In that case I think the issue is there is a named export and the default export. If you use a named import it is not calling translate. Some IDEs will auto pull the imports and you do not know unless you look at the source code that you need the default export and not the named one, i.e.,

`export class AutocompleteInput extends Component {`

So the only export should be the default one?

djhi:
No, we use the named one for unit tests

PR #1228, opened 2017-11-20
Documentation update for 'Writing Actions'

danyalaytekin:
In 'Writing Actions', the suggested styling for `CommentEditActions` (`float: right`) leads to the custom actions being non-clickable.  I've updated the styles to mirror those suggested in the documentation for [CreateEdit](https://github.com/marmelab/admin-on-rest/blob/master/docs/CreateEdit.md#actions).

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -78,7 +78,6 @@ export const CommentList = (props) =>
 
 Or, in the `<Edit>` page, as a [custom action](./CreateEdit.md#actions):
 
-{% raw %}
 ```jsx
 // in src/comments/CommentEditActions.js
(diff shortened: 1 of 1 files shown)


fzaninotto:
Tests fail, for a mysterious reason. Perhaps a regression in the latest material-ui update?

danyalaytekin:
The example website's 'Published at' input seems to behave differently when the 'create post' page is visited by hand, compared to when it is visited during the test.  During the test run, it takes a long time to be set to the default value.  Just an observation, not sure if relevant.

djhi:
Thanks!

PR #1229, opened 2017-11-20
Added jsnext:main and module

tkvw:
This allows usage of es6 modules instead of the transpiled es5 module.
My IDE (Intellij IDEA) doesn't recognize the transpiled exports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
         "src"
     ],
     "main": "lib/index",
+    "module": "src/index",
+    "jsnext:main": "src/index",
     "scripts": {
         "test": "make test"
     },


tkvw:
Too soon.

fzaninotto:
Can you elaborate on the "too soon" comment?

tkvw:
Yeah, I created the issue "too soon". I had to do some more investigation, because it didn't work as expected. In the end it has to do with `react-admin` using certain `stage-x` functionality so the code needs to be transpiled to es6 before using `module`. For some reason, 'jsnext:main' works (I have it included in my branch,so my IDE (Intellij IDEA) recognizes imported modules [@Issue](https://youtrack.jetbrains.com/issue/WEB-29832)). I think this has to do with the fact that babel/webpack ignoring `jsnext:main` and using `module` if available.

PR #1238, opened 2017-11-24
Allow field to supply a custom header component. 

tkvw:
This PR allows a field to supply a component to render the header. 
@See the test added: 

        it('should allow custom header renderer in field', () => {
            const wrapper = shallow(
                <DatagridHeaderCell
                    currentSort={{}}
                    field={<Field source="title" header={<MyHeader />} />}
                    updateSort={() => true}
                />
            );
            assert.equal(wrapper.find('MyHeader').length, 1);
        });

It splits the `DatagridHeaderCell` component in two components: `DatagridHeaderCell` and `DatagridHeaderCellContent`. The latter is the default renderer if none is supplied on the field.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ lib
 es6
 docs/_site/
 packages/ra-example/static
+# IDE 
+.idea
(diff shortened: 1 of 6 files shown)


tkvw:
This is not sufficient for my usecase.. I'll refactor and PR later

PR #1241, opened 2017-11-27
Added selectable / bulk action feature

tkvw:
Allow a list to be selectable

TODO: 
- [x]  Code
- [x]  Example app
- [ ]  Test
- [ ]  Document
- [x]  Add feature to simplelist, currently only DataGrid is supported 
- [ ] Add confirmation option for actions 
Usage: 
```
    <List
        {...props}
        selectable
        selectMode="bulk" // One of single, page or bulk
    >
```
prop | description 
--- | ----
selectable | Activates a select checkbox
selectMode | `single`: allows per row selection<br> `page`: allows selection on the current page only, a page change looses the selected items.<br> `bulk`: keeps selection between page changes.<br> Defaults to `bulk`

Actions can be added by supplying a `selectActionButton` prop to the Actions component of Admin. 
This should be a component of type `<SelectActionButton>`. The default value is: 
```
const DefaultSelectActionButton = props => (
    <SelectActionButton {...props}>
        <SelectActionMenuItem action="delete" label="ra.action.delete" />
    </SelectActionButton>
);

```

The action is dispatched using a redux action `AOR/CRUD_EXECUTE_LIST_ACTION` a default saga is available to execute batch deletes, it listens for the action with identifier 'delete'. 
You can implement you own logic by writing customSaga's for Admin. 
`<Admin customSaga={[myActionSaga]} ....> `

Example:


.gitignore
```diff
@@ -4,3 +4,5 @@ lib
 es6
 docs/_site/
 packages/ra-example/static
+# IDE 
+.idea
```
djhi: Please keep those developer specific ignore to your global gitignore: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

packages/ra-example/posts.js
```diff
@@ -82,6 +82,14 @@ export const PostList = ({ ...props }) => (
         {...props}
         filters={<PostFilter />}
         sort={{ field: 'published_at', order: 'DESC' }}
+        selectable
+        selectMode="bulk"
+        selectActions={[
```
djhi: Not fan of the actions as objects here. It would be better to pass a component like we do for `actions` already.
Maybe we can somehow just use the current `actions` but passing it a selection of items as props ?

packages/react-admin/package.json
```diff
@@ -9,6 +9,7 @@
         "src"
     ],
     "main": "lib/index",
+    "jsnext:main": "src/index",
```
djhi: Nice addition, however not related to this PR

packages/react-admin/src/sideEffect/saga/crudResponse.js
```diff
@@ -17,6 +17,12 @@ import {
 import { showNotification } from '../../actions/notificationActions';
 import resolveRedirectTo from '../../util/resolveRedirectTo';
 
+function* handleNotification(requestPayload, ...notificationArgs) {
+    yield requestPayload.skipNotification
```
djhi: why not yielding the `put` only if condition matches and do nothing otherwise ?

djhi:
Thanks for giving it a go :)

.gitignore
tkvw: Nice one! Didn't know of the existence of this option.

packages/react-admin/src/sideEffect/saga/crudResponse.js
tkvw: Yeah, I was asking myself that question, mainly kept it because of copy paste from original.

packages/react-admin/package.json
tkvw: Will revert and PR in other request

packages/ra-example/posts.js
tkvw: So I thought about this a bit. Using a selectActions attribute allows the usage of actions without implementing a custom Actions component. It's just for ease of use. The SelectActionButton requires to know what actions to render. Maybe this can be defaulted to a specific list (delete) and if the user wants to overwrite a custom actions component can be used.

djhi: I think we can handle that without a new `selectActions` prop just by passing the currently selected items to the existing component specified by the `actions` prop [here](https://github.com/marmelab/admin-on-rest/blob/master/src/mui/list/List.js#L246). We might provide a default `Delete` button handling this case in the default [`Actions` component](https://github.com/marmelab/admin-on-rest/blob/master/src/mui/list/Actions.js).

tkvw: I think I know what you mean. I'll refactor a bit and come back

tkvw:
Made actions responsive and added selectable to SimpleList.

packages/react-admin/src/mui/button/RefreshButton.js
```diff
@@ -3,8 +3,10 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import compose from 'recompose/compose';
 import Button from 'material-ui/Button';
+import IconButton from 'material-ui/IconButton';
```
djhi: I do like the work you have done on the responsiveness of actions but it should be done in a separate PR as it makes this one harder to review. The bulk operations handling is going to be complex so I'd rather keep it separated

packages/react-admin/src/mui/list/Actions.js
```diff
@@ -30,6 +33,11 @@ const Actions = ({
                     filterValues,
                     context: 'button',
                 })}
+            {selectable &&
+                React.cloneElement(selectActions, {
```
djhi: I'd rather have the default Actions here directly

packages/react-admin/src/mui/list/List.js
```diff
@@ -18,17 +18,26 @@ import Header from '../layout/Header';
...
 import withChildrenAsFunction from '../withChildrenAsFunction';
+import SelectActionButton from './SelectActionButton';
+import SelectActionMenuItem from './SelectActionMenuItem';
 
 const styles = {
     noResults: { padding: 20 },
 };
 
+const DefaultSelectActionButton = props => (
+    <SelectActionButton {...props}>
```
djhi: This should be in the Actions component directly. If you want to customize actions, it should be done as usual. Bulk actions should not be handled differently

packages/react-admin/src/mui/button/RefreshButton.js
tkvw: I understand, do you want it based from next of can I branch from this branch? Saves me some work 👍

tkvw:
Updated as requested and created a separate PR for the responsive part.

wadjeroudi:
Finally! good job. This PR and the multiple notifications PR were really needed. Thx.

Toosick:
Dang we finished this feature at the same time too. We are using aor-graphql. We kind of took a shortcut for the time being and passed around the Apollo client instead of going through the aor-graphql client as we wanted to send ~100 - 150 actions in 1 request instead of spamming the API with 1 request per action like you have here I think. Ideally we should probably make everything go through the admin-on-rest specified client but this was a faster implementation for our use case.

admin-on-rest changes (some places probably need more generalization): https://github.com/iolap/admin-on-rest/commit/4eedf0593d9d92f11c15bd54af699d56e3d25f7f

Features:
- This change includes a Popover menu in the top left of the table w/ dynamically created actions
- Customizable actions
- Default delete action
- Confirmation screen w/ and w/o input
- Combined w/ our main project repo we were able to send 1 request for all the updates/deletes.

Could possibly merge the 2 implementations?

tkvw:
> Dang we finished this feature at the same time too. We are using aor-graphql. We kind of took a shortcut for the time being and passed around the Apollo client instead of going through the aor-graphql client as we wanted to send ~100 - 150 actions in 1 request instead of spamming the API with 1 request per action like you have here I think. Ideally we should probably make everything go through the admin-on-rest specified client but this was a faster implementation for our use case.

Good job! 👍  The changes should be against next, so some implementation differences there, for example a Table is not `multiSelectable` anymore in material-ui next. This branch uses a delete per request to be compatible with the dataProvider/Restclient used in Admin. But it's only a default. Simply add a custom actionbutton (possible floating) and a custom saga to implement this, something like: 

```
<SelectActionButton>
    <SelectActionMenuItem action="deleteFast" label="ra.action.delete"/>
</SelectActionButton>
.. saga:
    function* watchSelectActions() {
        yield all([
            takeEvery(
                action =>
                    action.type === CRUD_EXECUTE_LIST_ACTION &&
                    action.meta.action === 'deleteFast',
                handleSelectDeleteFastAction
            )
        ]);
    };   
    function* handleSelectDeleteFastAction(({payload:{reload=true,ids}})){
       // Do Graphql delete fast 
    } 
```

> This change includes a Popover menu in the top left of the table w/ dynamically created actions

This can be done by rendering a custom SelectActionButton 

> Customizable actions 

This can be done by using SelectActionItem

> Default delete action

This is implemented, see `sideEffect/saga/defaultListActionSaga`

> Confirmation screen w/ and w/o input

That's a good one, where is this implemented? I'll a a task for this. 

> Combined w/ our main project repo we were able to send 1 request for all the updates/deletes.
   Could possibly merge the 2 implementations?

I don't think that's a good idea. Only one implementation should be merged.

djhi:
> Combined w/ our main project repo we were able to send 1 request for all the updates/deletes

Yep, this is a must have. We should probably introduce one or more AOR actions for bulk operations here. That would then be the `dataProvider` job (old `restClient`) to send one or more requests.

I think only one new AOR action would be needed, maybe with a `type` property (`delete`, `myCustomUpdate`). Thoughts ?

tkvw:
😕 I am missing something. 

> Yep, this is a must have. We should probably introduce one or more AOR actions for bulk operations here. That would then be the dataProvider job (old restClient) to send one or more requests.

What are `AOR actions` in this context? `<SelectActionItem>`?

> I think only one new AOR action would be needed, maybe with a type property (delete, myCustomUpdate). Thoughts ?

Again, what is a `AOR action` in this context?

djhi:
*Redux Actions*: https://github.com/marmelab/admin-on-rest/blob/master/src/actions/dataActions.js
*DataProvider Actions*: https://github.com/marmelab/admin-on-rest/blob/master/src/rest/types.js

tkvw:
Ah, now I understand what you mean. Yeah, something like action/verb `BULK` and have it typed should do the trick. For the action, `ids` should be required and optionally have a `data` {Object} arg (for bulk update) and `previousData` array.

tkvw:
I thought about the modals part a bit more and came up with an applicaton wide pluggable solution.    https://github.com/marmelab/admin-on-rest/pull/1251

kfern:
Hi! Any idea when will this RP be merged?

djhi:
We are currently focused on bug fixes and small devices compatibility. Besides this PR will take some time to review. Finally, it needs rebase

tkvw:
@djhi : do you want to include it in 2.0.0? If so, I can spend some time on rebase.

djhi:
That would be very nice indeed :+1:

fzaninotto:
Any updates on the rebase?

tkvw:
Yes, working on it. Expect to finish today.

tkvw:
Finished rebase and major refactoring. 
Outstanding changes: 
### Introduced dataProvider verb 'BULK_ACTION'

### Request format: 
```
{ action: string, actionMeta:{Object}, data:{Object},ids:{mixed[]}, previousData: {Object} }
```
Field | Description
---|---
action | Builtin support for UPDATE and DELETE, but any action value can be used
actionMeta | Additional metadata send from the action executed
data | Values to update in case of action=UPDATE
ids | Array of ids to perform the action on 
previousData | Object keyed by id with the known values in the store

### Response format: 
```
{ data: { {resolved:{data:Record},rejected:error}[] }
```
The dataprovider should always `resolve` the request and respond with all individual `resolves` and `rejections` in data entry array. 

### addBulkActionFeature dataProvider middleware
The example app includes a dataProvider middleware to use the `UPDATE` and `DELETE` action with all existing dataproviders! Please note that this is not very efficient, because a request is made for each UPDATE/DELETE. Best would be to support batch actions on the server and write support in the dataprovider. 


### Added `crudExecuteBulkAction` actionCreator, it expects: 

Argument | Description
---|----
resource | The resource to execute the bulk action on 
action | The bulk action to execute 
ids | Array of ids to execute the bulk action on 
data | Optional object with values to use with execution of bulk action. In case of action UPDATE this should contain the values to update for each record 
previousData | Object keyed by id with the previousdata
actionMeta | See actionMeta below

ActionMeta: 

Key | Description 
---|---
cacheAction | The cacheAction to use for the BULK_ACTION, can be one of DELETE,UPDATE or NONE, defaults to UPDATE. 
selection | `{keepSuccess:bool,keepFailed:bool}` allows to set what to do with selection when the action has finished
notification | `{success:string,failed:string, failedArgs:func}` message labels to use then action finishes. Success is called with the number of successful handled records, and failed is called for each rejection. `failedArgs` allows to supply a function to generate message arguments. It will be called with `requestPayload` and the `rejection`.
refreshList | bool, instruct to reload the list after execution
\* | All other values are passed to `actionMeta` of dataProvider

## Added bulkActions component for ListActions
```
import { BulkActions } from 'react-admin';
<List actions={<ListActions bulkActions={<BulkActions>{...bulkactions}</BulkActions>}/>}/>
```
`bulkActions` should be `BulkAction` items, there is build in support for: 

### BulkActionDelete
This button removes all items in the selection. 
```
import { BulkActions,BulkActionDelete } from 'react-admin';
<List actions={<ListActions bulkActions={<BulkActions><BulkActionDelete/></BulkActions>}/>}/>
```
### BulkActionUpdate
This action will trigger a Dialog with configurable Form to update multiple values which are applied to all selected items. 
```
import { BulkActions,BulkActionUpdate,SimpleForm,TextInput } from 'react-admin';
<List actions={<ListActions bulkActions={
            <BulkActions>
                <BulkActionUpdate>
                    <SimpleForm>
                        <TextInput source="title" />
                    </SimpleForm>
                </BulkActionUpdate>
            </BulkActions>
}/>}/>
```
### BulkAction
This is a generic action item which can be used for all other actions or is a startpoint for different actions. 
```
import { BulkActions,BulkAction } from 'react-admin';
<List actions={<ListActions bulkActions={
    <BulkActions>
        <BulkActionCustom 
          label="app.action.email" 
          action="EMAIL" 
          actionMeta={{template:'subscribe'}}
          cacheAction={{NONE}}
          notificationSuccess="app.notification.email.success"
          notificationFailed="app.notification.email.failed"
          keepSelectionFailed // Keep selection of failed items
        />
    </BulkActions>
}/>}/>
```
Please note that this action should be implemented by the dataprovider!

fzaninotto:
I haven't read the code yet, but i can already say: great job! 

Here are a few remarks:

## Why are there several `selectMode` possible values?

I don't understand this setting, and probably the users won't understand it either. There should be only one mode, not configurable. From what I understand, it's the `bulk` mode.

## Simplify `List` API for bulk actions

As for the API to customize the bulk actions, I think a 

```<List bulkActions={<BulkActions><BulkActionDelete/></BulkActions>} />```

Would be preferable over your

```<List actions={<ListActions bulkActions={<BulkActions><BulkActionDelete/></BulkActions>}/>}/>```

It would have the same effect under the hood (the `List` component passing the `bulkActions` prop to the `ListActions` component if defined). If the `bulkActions` isn't empty, the `<ListActions>` would display the bulk action button automatically.

## No generic bulk action or data provider verb

I'm worried by the idea of a generic bulk action and data provider verb. Why can't a bulk delete emit several `DELETE` actions instead of one bulk action? Also, why should every bulk action issue a request on the data provider? I think we should leave the implementation to the user. I know it's a lot of work for the user (adding a reducer, a saga, and updating the data provider), but I'm really afraid that the strong formalism will become a burden in the future

## Move BulkActionUpdate to another PR

This is something that I'm not sure should be in the core. It would help keep that PR slimmer if you removed the feature.

## Is BulkActionCustom really a good idea?

The amount of controller logic that must be passed to a (presentational) component makes me think. I'm not sure this is a good API. Again, let's leave the implementation to the end users, and see which patterns emerge, before deciding on an API that will be hard to change in the future.

**Comment by tkvw** [comment]
> I haven't read the code yet, but i can already say: great job!

👍  thanks, it was some work, but it really fits well with `react-admin` now

> Why are there several selectMode possible values?

Mode | Description
---|----
`single` | Allow only one row to be selected at any time (disables selectAll)
`page` | Only allow selection of items on the visible page, so if multiple pages are available a navigation to the second page loses the selection of the previous page 
`bulk` | Allow selection across pages 

I think these modes are pretty much what a user could ask for need and the most sensible, `bulk` is the default, so no problems there. I'll vote to keep it in. 

> Simplify List API for bulk actions

Yes, much cleaner 👍 

> No generic bulk action or data provider verb

Consider a user wants to remove 10 records. In the most ideal situation there is only one single request from RA to the server. This would require a server that supports batch removeall (there are plenty) and a dataprovider that somehow knows how to batch the requests. 
Using the `BULK_ACTION` verb the dataProvider has this opportunity, this is an edge case, but for efficient usage of bulk actions this is a must I think (I am actually using it). On the server side this would allow single transaction update's and a rollback on the entire operation if it fails (this is not possible with individual requests). 
Btw this idea comes from @djhi, see a couple of posts above 😄 

**Please note:** that the supplied dataProvider middleware (`addBulkActionFeature.js`) in the example app does **exactly** what you mention. It add support of the `BULK_ACTION` verb to *all* dataproviders.  It triggers 10 different delete requests and supports the `DELETE` and `UPDATE` actions out of the box with zero configuration. 

> I'm really afraid that the strong formalism will become a burden in the future 

The format heavily use the same formalism as the current format. Each entry in the `data` response is an object with either `{resolved:response}` or `{rejected:error}`, where response and error are exactly the same a the responses from the `DELETE` and `UPDATE` verb. So it only extends on the current format. 

> Move BulkActionUpdate to another PR

I know it's a bit much to check, but if it helps I can move it to another PR. This component is also missing a detail I think. When presented it should check if there are matching values in the selected records and present these values as `initialValues` for the batch update form. 

> Is BulkActionCustom really a good idea?

Yes, it is 😄 The really nice thing about the `BULK_ACTION` verb, that it allows the user to define whatever action they want to use. 
Consider a resource displaying a list of virtual machines. A virtual machine can be powered down by executing a POST to `/machine/id/powerdown`. This is does not fit in the current implementation of `react-admin` dataprovider. 
Now using the `BULK_ACTION` verb: 

Powerdown middleware: 
```
(fetch,next) => (type, resource, params) => 
   if(type === BULK_ACTION && params.action === 'powerdown'){
       return Promise.all(params.ids.map(id=>
          fetch(`/machine/${id}/powerdown?forced=${params.actionMeta.forced?'true':'false'}`,{method:'POST'})
             .then(response=> ({resolved:response})
             .catch(error=> ({rejected: error})             
   }
   return next(type,resource,params); 
```
and in the ui: 
```
   const PowerDownAction = props => (
    <BulkActionCustom
        label="ra.action.powerdown"
        action="powerdown"
        cacheAction="NONE" // Or UPDATE if the server returns interesting values 
        notificationSuccess="ra.notification.bulk_action.powerdown.success"
        notificationFailed="ra.notification.bulk_action.powerdown.failed"
        refreshList // Does the list needs to be refreshed? 
        keepSelectionFailed // Keep the failed selection so the user can retry to powerdown the failed servers
        {...props}
    />
);
   const PowerDownForcedAction = props => (
    <BulkActionCustom
        label="ra.action.powerdown_force"
        action="powerdown"
        actionMeta={{
            forced:true
        }}
        cacheAction="NONE" // Or UPDATE if the server returns interesting values 
        notificationSuccess="ra.notification.bulk_action.powerdown.success"
        notificationFailed="ra.notification.bulk_action.powerdown.failed"
        refreshList // Does the list needs to be refreshed? 
        keepSelectionFailed // Keep the failed selection so the user can retry to powerdown the failed servers
        {...props}
    />
);

```
\o/ I think this really elegant

tkvw:
I refactored my commits, so this PR is more readable. I added comments where applicable in each commit message. Please start with first commit.

tkvw:
Closed in favor of https://github.com/marmelab/admin-on-rest/pull/1521

PR #1243, opened 2017-11-27
Added jsnext:main for es6 packagers/loaders

tkvw:
Added jsnext:main
@See: https://github.com/marmelab/admin-on-rest/pull/1241#discussion_r153205547

### DIFF
--- a/packages/react-admin/package.json
+++ b/packages/react-admin/package.json
@@ -10,6 +10,7 @@
     ],
     "main": "lib/index",
     "jsnext:main": "src/index",
+    "module": "src/index",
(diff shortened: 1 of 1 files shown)


tkvw:
This does not work. There should be an ES6 build where `module` and `jsnext:main` points to, the problem now is that this project requires to be transpiled because of stage-x features being used. [Export default](https://github.com/babel/proposals/issues/17) for example.  
What do you think?

tkvw:
Closing for now, only helpful for my IDE (Intellij IDEA).

PR #1244, opened 2017-11-27
[RFR] Add missing import in Menu Documentation

jdemangeon:
Since I've updated my project from AOR 1.2.3 to 1.3.2, my own "CustomMenu" does not work anymore because we need to "connect" it manually. There's no BC in CHANGELOG about that, I don't know when change append exactly.

Anyway, I've added a missing import on the new "CustomMenu" example in documentation.


fzaninotto:
Superseded by #1197 and #1298

PR #1245, opened 2017-11-27
Allow multiple notifications

tkvw:
This PR allows multiple notifications. It does not display multiple notifications/snackbars but enqueues them. 
@see https://storage.googleapis.com/material-design/publish/material_v_12/assets/0B6Okdz75tqQsX3VhUHNoaWVEbDQ/components-snackbars-usage-consecutivesbdifferenttype-005.webm

It also allows to pass `messageArgs` to `showNotification`. The `messageArgs` will be passed to translate.


packages/react-admin/src/actions/notificationActions.js
```diff
@@ -1,8 +1,31 @@
 export const SHOW_NOTIFICATION = 'AOR/SHOW_NOTIFICATION';
 
-export const showNotification = (text, type = 'info', autoHideDuration) => ({
+/**
+ * @typedef {Object} notificationOptions
```
djhi: :heart_eyes:

fzaninotto:
Excellent, thanks!

tkvw:
Just a heads up, this currently doesn't work as expected, @see: https://github.com/mui-org/material-ui/pull/9670, I've fixed it by a PR to material-ui.

PR #1251, opened 2017-11-29
[RFR]: Modals/Dialog proposal

tkvw:
This PR allows usage of dialogs/modals from **anywhere** in the application. I've included an [implementation](https://github.com/marmelab/admin-on-rest/pull/1251/files#diff-319b8abdb62799f1b7a4a145a644ea73) which checks if a form is dirty and asks to continue. 

Todo: 
- [X] Write initial code 
- [ ] Discuss feature for acceptance and usage in the application (are there more use cases besides dirtycheck in form)
- [ ] Add example to example app  
- [ ] Document feature 
- [X] Tests 

Although this is a very simple use case, it shows a modal. Modals can be added application wide (like `confirm_yes_no` or you can add them to whatever page you want to use it on). This allows use cases like a tutorial to render on first visit of a specific page or a quick add user page.  What's really nice is that a modal can be triggered from a sideeffect (saga). 

It uses [redux-promising-modals](https://github.com/AKolodeev/redux-promising-modals) which is a cool utility library for modals and redux. 

To call a modal from a component ([using render props](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce)): 
```
<Modals render={({pushModal})=> 
    <Button onClick={()=> pushModal('confirm_yes_no_cancel',{
   title: 'ra.message.something', // Overwrite default title
   message: 'message.label' // Overwrite default message 
}).then(answer=> {
 // Handle answer accordingly. 
})}
}}/>
```
Note: A promise which resolves with the answer is returned for convenience. You probably want to use a saga in more complex dialogs.

Call a modal from saga: 
```
// Gui input from a saga! 
const answer = yield put.resolve(
    yield call(pushModal, 'confirm_yes_no')
);
```

The redux actionTypes are: 
```
PUSH_MODAL_WINDOW,
INSERT_MODAL_WINDOW,
POP_MODAL_WINDOW,
CLEAR_MODAL_WINDOWS
```

Creating a custom modal couldn't be easier: 
```
const myModal = ()=> <Modal type="tutorial-step-1" render={({
open, // Indicates the model should be shown on screen 
translate, // For translating the messages in the dialog 
popModal, // Closes the currently displaying modal and shows the next one if available
nextModal, // Opens the next modal
prevModal, // Opens the previous modal,
insertModal, // Insert a modal 
shiftModal, // Remove the first modal 
})=>{
    return <Dialog open={open} ... >

    </Dialog>
}}
```
They can be registered by supplying an array of modals to customModals prop of `Admin` or simply add them to the page where you want to use them! 

Modals can be chained, etc..to much for now 😄


Toosick:
Well that was quick. 💯

packages/react-admin/package.json
```diff
@@ -58,6 +58,7 @@
         "recompose": "~0.25.0",
         "redux": "~3.7.2",
         "redux-form": "~7.0.3",
+        "redux-promising-modals": "^2.1.2",
```
djhi: Please use `~`

tkvw:
Yeah, I screwed up this commit a bit. I started using git effectively for only a couple months ago, so a lot to learn. Fixed now. The only unrelated thing in there now should be the `DirtyBlocker` of form.

packages/react-admin/src/mui/modals/Modals.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import compose from 'recompose/compose';
+import * as actions from './actions';
```
djhi: Why import every possible actions ? That should be the responsibility of the user inside its render callback

packages/react-admin/src/mui/modals/Modal.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import compose from 'recompose/compose';
+import shouldUpdate from 'recompose/shouldUpdate';
+
+import * as actions from './actions';
```
djhi: Why import every possible actions ? That should be the responsibility of the user inside its render callback

tkvw: The render prop supplies these actions because they are redux connected actions. https://github.com/tkvw/admin-on-rest/blob/ce55ef916971e0bff19205cd9f3cfd1376742d00/packages/react-admin/src/mui/modals/Modal.js#L42
So Modal is nothing more than a bridge to these actions.

packages/react-admin/src/mui/modals/Modals.js
tkvw: Same as the Modal comment. They are exported as redux actions to the render prop.

packages/react-admin/src/mui/modals/Modal.js
djhi: Yes but why do we need this bridge. Shouldn't the consumer of the `Modal` just import what it needs by itself ?

tkvw: The entire purpose of `Modal` is: 
* to register a modal type (just a string identifier), which allows the Modal to be called by
* to abstract redux-promising-modals usage 

For a user to create a new modal is as simple as: 
```
<Modal type='myConfirm' render={(popModal)=>{
   <Dialog ...> 
}} /> 
```
This is much clearer than to create a connected Dialog and importing the appropriate `redux-promising-modals` actions (which requires more knowledge of redux-promising-modals). 

So it leads to cleaner code. 
But the user is free to implement it's own modal it's pretty straight forward. Basically just use mapstate and connect the appropriate actions.

tkvw:
Is there interest in this PR? If not I'll add it to an addons package I'm creating

fzaninotto:
I love the idea of a dirty modal, but I think it can be implemented with mui Dialogs without introducing a whole modals subframework in react-admin. What I don't want is to force users to use react-admin to display a modal - they should be able to do it with mui alone.

tkvw:
So, the `DirtyBlocker` was only an example implementation of the goal of this PR; adding modals to `react-admin`.  You can copy the `DirtyBlocker` component without much problem with a custom Dialog implementation. I'll close this PR and move it to userland.

fzaninotto:
As seen in #1589, we're moving away from modals to optimistic rendering for a better User Experience.

PR #1252, opened 2017-11-30
Logo in drawer proposal

tkvw:
I think it would be nice to allow a logo in the drawer.


fzaninotto:
Nice idea, thanks.

But I'm not interested to put it in the core, plus it's possible to do it in userland using a custom Layout.

PR #1258, opened 2017-12-04
Fix fields not updated.

### LINKED ISSUE #1257: Edit View not being updated when fields get changed.- Again

**What you were expecting:**
Edit Form updates when its children changes

**What happened instead:**
Edit Form not updated when its children changes
Due to the fact that `WithPermissionsFilteredChildren` is implemented in a way only `initializeChildren` at `componentDidMount`.
**Steps to reproduce:**
https://github.com/marmelab/admin-on-rest/issues/509

**Related code:**

```
```

**Other information:**

**Environment**

* Admin-on-rest version:
* React version:
* Browser:

yshing:
This should fix
#1257

### DIFF
--- a/src/auth/withPermissionsFilteredChildren.js
+++ b/src/auth/withPermissionsFilteredChildren.js
@@ -18,6 +18,10 @@ export default BaseComponent => {
             this.initializeChildren(this.props.children);
         }
 
+        componentWillReceiveProps(nextProps){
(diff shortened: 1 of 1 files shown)


fzaninotto:
@djhi OK for you?

djhi:
Nope, this results in a infinite render loop inside the Edit component. It can be seen on the example app for the posts.

PR #1260, opened 2017-12-05
[RFR] Review permissions

djhi:
Follow #1263 

- [x] Review permissions at resource level
- [x] Review permissions at admin level
- [x] Update documentation
- [x] Upgrade guide


packages/react-admin/src/reducer/admin/auth.js
```diff
@@ -0,0 +1,67 @@
+import {
+    USER_CHECK_SUCCESS,
+    USER_LOGIN_SUCCESS,
+} from '../../actions/authActions';
+
+const initialState = {
+    user: null,
```
fzaninotto: I suggest we initialize it using localStorage, and use the saga to store the credentials there. What do you think ?

packages/react-admin/src/reducer/admin/auth.js
```diff
@@ -0,0 +1,67 @@
...
+    permissions: {},
+};
+
+const getPermissionsKey = (route, resource, routeParams) => {
```
fzaninotto: Could you add js doc and an example ?

packages/react-admin/src/AdminRoutes.js
```diff
@@ -1,28 +1,39 @@
...
-            }
+    async componentWillReceiveProps(nextProps) {
+        if (nextProps.permissions !== this.props.permissions) {
+            await this.initializeResources(nextProps);
+        }
+    }
 
+    getPermissions = () => {
+        const { userCheck, location, match } = this.props;
+        userCheck({}, location && location.pathname, match && match.params);
```
fzaninotto: Am I understanding correctly that each visit to a page triggers an AUTH_CHECK (the same check that at login) instead of AUTH_GET_PERMISSIONS? If so, I think it’s not a good idea. We should allow a separate implementation of authentication and authorization in the auth client

djhi: Yes, each location change will trigger an AUTH_CHECK. I'm not sure about the benefits of such separation. Can you explain ?

packages/react-admin/src/reducer/admin/auth.js
djhi: Added jsDoc on the top level export (inside `reducer/index.js`)

packages/react-admin/src/reducer/admin/auth.js
djhi: We did not enforce how to store the authentication related until now. I'm still not sure we should.

packages/react-admin/src/auth/WithPermissions.js
```diff
@@ -0,0 +1,91 @@
...
+import compose from 'recompose/compose';
+import getContext from 'recompose/getContext';
+
+import { userCheck } from '../actions/authActions';
+import { AUTH_GET_PERMISSIONS } from '../auth/types';
+
+/**
+ * Retrieve permissions for a specific route.
+ *
+ * Useful for Route components ; used internally by CrudRoute.
```
fzaninotto: you should tell (and ullistrate in the example) that it injects the `permissions` prop to the child.

djhi: Done

packages/react-admin/src/AdminRoutes.js
```diff
@@ -7,22 +7,44 @@ import getContext from 'recompose/getContext';
...
+    componentDidUpdate(prevProps, prevState) {
+        if (
+            prevState.permissions !== this.state.permissions &&
+            this.state.permissions !== initialPermissions
+        ) {
+            this.initializeResources();
+        }
+    }
 
+    getPermissions = async () => {
```
fzaninotto: As this is the job of `WithPermissions`, why don't we decorate `<AdminRoutes>` with `<WithPermissions>` instead of duplicating the code here?

djhi: Because `AdminRoutes` is also responsible for rendering custom routes which may not be protected and `WithPermissions` will check for authentication

packages/react-admin/src/auth/WithPermissions.js
```diff
@@ -0,0 +1,107 @@
...
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import compose from 'recompose/compose';
+import getContext from 'recompose/getContext';
+
+import { userCheck } from '../actions/authActions';
+import { AUTH_GET_PERMISSIONS } from '../auth/types';
+
+/**
+ * Retrieve permissions for a specific route.
```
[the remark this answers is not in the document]
djhi: Done

fzaninotto:
And needs rebase...

Issue #1256, opened 2017-12-03
Is the documentation on custom rest client understandable?


te-online:
**What you were expecting:**

A short, yet comprehensible documentation on how to create a custom rest client for admin on rest.

**What happened instead:**

The documentation states that it's very easy, but I didn't find it very easy. Maybe this is my fault, but maybe also the documentation could be improved.

**Steps to reproduce:**

Imagine you're new to admin-on-rest. Try to understand how to create a custom rest client from this guide https://github.com/marmelab/admin-on-rest/blob/master/docs/RestClients.md#writing-your-own-rest-client.

----

It feels like the documentation was written to be friendly and helpful. But after using it extensively, I have to say that it sometimes says something is simple, when it's instead rather difficult and this makes me feel stupid (see also http://bradfrost.com/blog/post/just/).

Take for example this statement in the docs for creating a custom rest client:
> But don't be afraid, it's easy!

Personally, I didn't find it easy at all.

My proposal is therefore to add the following to the documentation of custom rest clients:
* Add that you need to adjust the imports when copying from the original simple rest client as stated here https://stackoverflow.com/questions/44745113/creating-a-custom-rest-client-admin-on-rest.
* Add that you need to update this file every time admin-on-rest is updated and the simple rest client might get any optimizations.
* Add context to the snippets under the respective “Examples:” paragraphs.
* Rephrase “But don't be afraid, it's easy!” to “Follow the guide below to create a custom rest client”.

I'm sorry in case I can't articulate well enough what I mean, but there are a few more places where I had this feeling when reading the docs (e.g. https://github.com/marmelab/admin-on-rest/blob/master/docs/Fields.md#writing-your-own-field-component). Writing this, I hope you understand it as constructive feedback. :-)

Phocea:
Myself and my team followed the documentation and also looked at the several existing examples and plugins. We had zero problems. 

You should raise a PR with the documentation update you suggested, it would be faster than listing it here :)

djhi:
I agree with Thomas, we should not assume anything is easy in the documentation. I'll work on it asap

djhi:
But as @Phocea said, a PR will be welcome :)

te-online:
> I agree with Thomas, we should not assume anything is easy in the documentation. I'll work on it asap

Thank you so much, for considering this :-)

> We had zero problems.

Okay, maybe I **am** indeed stupid 😅 (Or you guys are just very good at what you do.) 😉

[...]

PR #1264, opened 2017-12-05
Update custom rest client docs in RestClient.md

te-online:
* Add that you need to adjust the imports when copying from the original simple rest client as stated here https://stackoverflow.com/questions/44745113/creating-a-custom-rest-client-admin-on-rest.
* Add that you need to update this file when admin-on-rest is updated and the simple rest client might get any optimizations and you want to benefit from them. 
* Add a short intro to the format paragraphs, to get to know the context.
* Rephrase “But don't be afraid, it's easy!” to “Follow the guide below to create a custom rest client”.

Addresses #1256

### DIFF
--- a/docs/RestClients.md
+++ b/docs/RestClients.md
@@ -255,7 +255,31 @@ render(
 
 ## Writing Your Own REST Client
 
-Quite often, none of the the core REST clients match your API exactly. In such cases, you'll have to write your own REST client. But don't be afraid, it's easy!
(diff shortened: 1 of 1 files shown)


fzaninotto:
I'd prefer that you backport my changes from #1311, and adapt them to the current naming scheme. It's better to keep the documentation pretty much in sync between 1.x and 2.x.

te-online:
Yes, I'd vote for that too. Let me see if I can find the time to update my merge request next week with your changes from #1311. However I don't really see how this would affect the sync between 1.x and 2.x?

te-online:
Continued in #1441.

PR #1265, opened 2017-12-06
Bug: Validate not being passed on redux-form field.

tkvw:
This bug is also visible in the example app, edit a post title (which is required) and remove all text. The validation error will not appear and the post is saved.


djhi:
Thanks!

afilp:
@tkvw Is the master OK with this or we need to create a PR for master too? (I am using master, not next)

PR #1268, opened 2017-12-06
[RFR] Add a className prop to all components 

djhi:
Follow #1266 

- [x] Fields
- [x] Inputs
- [x] Core components (`Edit`, `Show`, etc.)
- [x] Datagrid
- [x] Theming Documentation
- [x] Component CSS documentation
- [x] Upgrade guide


fzaninotto:
Awesome work. Can you update the theming / styling guide to mention the preferred way to override styles, using `className` with JSS?

djhi:
Yes, but we'll have to talk about the Datagrid styles before

packages/react-admin/src/mui/list/Datagrid.js
```diff
@@ -79,40 +78,42 @@ class Datagrid extends Component {
...
             headerOptions,
             bodyOptions,
             rowOptions,
         } = this.props;
         return (
             <Table
-                style={options && options.fixedHeader ? null : styles.table}
+                className={classnames(
+                    {
+                        [classes.table]: !options || !options.fixedHeader,
```
fzaninotto: material-ui 1.0 doesn't support fixed headers anymore

packages/react-admin/src/mui/list/DatagridBody.js
```diff
@@ -27,7 +28,8 @@ const DatagridBody = ({
     >
         {ids.map((id, rowIndex) => (
             <TableRow
-                style={rowStyle ? rowStyle(data[id], rowIndex) : styles.tr}
+                className={classes.row}
```
fzaninotto: I think you should add a `rowEven` and `rowOdd` classes, that the user could override in `classes`

packages/react-admin/src/mui/list/DatagridCell.js
```diff
@@ -1,42 +1,30 @@
...
-    );
-    return (
-        <TableCell className={className} style={computedStyle} {...rest}>
-            {React.cloneElement(field, { record, basePath, resource })}
-        </TableCell>
-    );
-};
+}) => (
+    <TableCell
+        className={classnames(className, field.props.className)}
```
fzaninotto: missing what we talked about on inspecting the alignment required by the field

packages/react-admin/src/mui/list/DatagridHeaderCell.js
```diff
@@ -38,67 +37,52 @@ export const DatagridHeaderCell = ({
...
-    return (
-        <TableCell className={className} style={style}>
-            {field.props.sortable !== false && field.props.source ? (
-                <Button
-                    onClick={updateSort}
-                    data-sort={field.props.source}
-                    className={classes.sortButton}
-                >
+}) => (
+    <TableCell className={classNames(className, field.props.headerClassName)}>
```
fzaninotto: missing what we talked about on inspecting the alignment required by the field

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "private": true,
-    "name": "react-admin",
+    "name": "react-admin-lerna",
```
fzaninotto: ??

docs/Theming.md
```diff
@@ -9,71 +9,206 @@ Whether you need to adjust a CSS rule for a single component, or change the colo
 
 ## Overriding A Component Style
 
-Most react-admin components support two style props to set inline styles:
+Every react-admin component provides a className property which is always applied to the root element.
```
fzaninotto: s/className/`className`/

docs/Theming.md
```diff
@@ -9,71 +9,206 @@ Whether you need to adjust a CSS rule for a single component, or change the colo
...
-                headerStyle={{ textAlign: 'right' }}
+            <TextField source="price" />
+            <MyEditButton />
+        </Datagrid>
+    </List>
+);
+```
+{% endraw %}
+
+For some components, when the `className` property isn't enough and you need to access deeper elements, you can take advantage of the `classes` property to customize them. Take a look at a component documentation to know what classes are available. For instance, you can have a look at the [Datagrid CSS documentation](./List.md#the-datagrid-component).
```
fzaninotto: s/to know what classes/to know which classes/

package.json
djhi: Because jest was complaining about two packages having the same name

djhi: Besides, this package.json will never be published on npm so it doesn't really matter right ?

fzaninotto: ok

PR #1277, opened 2017-12-11
Bugfix: Fieldtitle does not update on locale change

tkvw:
PR fixed issue that field titles aren't updated on locale change, as shown below:


tkvw:
Changed mui buttons as well because connect implements SCU and blocks if translate changes.

PR #1278, opened 2017-12-11
[RFR]: Async i18n messages proposal

tkvw:
This PR allows i18n messages to be loaded async and allows to change the locale through a query param. 
Usage: 
```
<Admin
messages={{
fr: import('./i18n/fr.js').then(m=>m.default)
}}
```


djhi:
IMO, it would be better to add/change the current locale actions so that they may include messages. This way, it would be possible to load them on demand and from any source (file, db, API, whatever).

tkvw:
@djhi : I changed the implementation to what you've suggested. I had some issues with my initial solution and this implementation is cleaner.

tkvw:
Moved list query issue to separate PR https://github.com/marmelab/admin-on-rest/pull/1326 and rebased.

Phocea:
A way to force reload of the locale without reloading the app would also be great. Then we could administer and reload locale via react-admin :)

packages/ra-example/app.js
```diff
@@ -21,13 +19,9 @@ import {
...
 
 import data from './data';
-import * as customMessages from './i18n';
 import authClient from './authClient';
 
-const messages = {
-    fr: { ...frenchMessages, ...customMessages.fr },
-    en: { ...englishMessages, ...customMessages.en },
-};
+import englishMessages from './i18n/en';
```
fzaninotto: you probably forgot to update the English messages to use the custom messages, like in French.

packages/react-admin/src/Admin.js
```diff
@@ -44,7 +45,11 @@ const Admin = ({
         appReducer(action.type !== USER_LOGOUT ? state : undefined, action);
     const saga = function* rootSaga() {
         yield all(
-            [crudSaga(dataProvider, authClient), ...customSagas].map(fork)
+            [
+                crudSaga(dataProvider, authClient),
+                i18nSaga({ locale, messages, enableLocaleQuerySwitcher }),
```
fzaninotto: I'd prefer that you put the `i18nSaga` in the `crudSaga`, to avoid composing sagas here. Remember, the code of the `Admin` component becomes the code to duplicate for people who want to build a custom app. The less work they have, the better.

packages/react-admin/src/sideEffect/saga/i18nSaga.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    function* handleUserLogoutAction() {
+        yield* loadLocaleMessages(initialLocale);
+    }
+
+    return function* watchLocaleChanges() {
+        yield all([
+            takeLatest(CHANGE_LOCALE, handleChangeLocaleAction),
+            put(changeLocale(initialLocale)),
```
fzaninotto: won't this create a rerender each time the app renders for the first time?

packages/react-admin/src/sideEffect/saga/i18nSaga.js
```diff
@@ -0,0 +1,68 @@
...
+import { all, call, put, select, takeLatest } from 'redux-saga/effects';
+import { LOCATION_CHANGE } from 'react-router-redux';
+
+const resolveMessages = messages =>
+    Promise.resolve(typeof messages === 'function' ? messages() : messages);
+const localeSelector = state => state.i18n.locale;
+
+export default ({
+    locale: initialLocale,
+    messages,
```
fzaninotto: Storing the messages in the saga feels weird to me. I can't put the finger on why - probably because usually, sagas don't persist anything. storing stuff is the store's job.

packages/react-admin/src/actions/localeActions.js
```diff
@@ -4,3 +4,12 @@ export const changeLocale = locale => ({
     type: CHANGE_LOCALE,
     payload: locale,
 });
+export const CHANGE_MESSAGES = 'RA/CHANGE_MESSAGES';
+
+export const changeMessages = (locale, messages) => ({
+    type: CHANGE_MESSAGES,
+    payload: {
+        locale,
```
fzaninotto: why does this action take the `locale` as parameter?

tkvw:
Changed as requested, introduced an `i18nDataprovider` and moved locale change using query param to user domain. I've included it in the example-app as a custom saga.

packages/react-admin/src/Admin.js
```diff
@@ -29,10 +29,10 @@ const Admin = ({
     dashboard,
     history,
     locale,
-    messages = {},
     menu = Menu,
     catchAll,
     dataProvider,
+    i18nProvider,
```
djhi: Can we set a default `i18nProvider` which would load messages as before ?

tkvw: ```locale => this.props.messages[locale]``` ?

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
 ```
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
+
+## Using asynchronous messages loading 
+
+By default React Admin loads all configured lanuages and languagebundles on startup. If you have very large language files, you can load messagebundles asynchronous.
```
djhi: s/languagebundles/their messages/

djhi: s/messagebundles/messages/

djhi: s/asynchronous/asynchronously/

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
 ```
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
+
+## Using asynchronous messages loading 
+
+By default React Admin loads all configured lanuages and languagebundles on startup. If you have very large language files, you can load messagebundles asynchronous. 
+React-admin uses an `i18nProvider` to load messages. A custom dataprovider can be supplied to the `Admin` component by using the `i18nProvider` prop.
```
djhi: s/dataprovider/i18nProvider/

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
 ```
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
+
+## Using asynchronous messages loading 
+
+By default React Admin loads all configured lanuages and languagebundles on startup. If you have very large language files, you can load messagebundles asynchronous. 
+React-admin uses an `i18nProvider` to load messages. A custom dataprovider can be supplied to the `Admin` component by using the `i18nProvider` prop. 
+The signature of the `i18nProvider` is: 
+```jsx
```
djhi: Please add a new empty line above. Otherwise, it will not render correctly.

Screenshot:

screen

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
...
+locale => {Promise}
+```
+The promise should resolve to a messages object. 
+
+An example asynchronous `i18nProvider` is: 
+```jsx
```
djhi: Same here

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
...
+
+By default React Admin loads all configured lanuages and languagebundles on startup. If you have very large language files, you can load messagebundles asynchronous. 
+React-admin uses an `i18nProvider` to load messages. A custom dataprovider can be supplied to the `Admin` component by using the `i18nProvider` prop. 
+The signature of the `i18nProvider` is: 
+```jsx
+locale => {Promise}
+```
+The promise should resolve to a messages object. 
+
+An example asynchronous `i18nProvider` is:
```
djhi: s/is/would be/

docs/Translation.md
```diff
@@ -343,3 +379,31 @@ translate('not_yet_translated', { _: 'Default translation' })
...
+      en: () => englishMessages,
+      fr: () => import('./i18n/fr').then(m => m.default),
+      nl: () => import('./i18n/nl').then(m => m.default)
+   }
+   return locale => messages[locale]();
+} 
+export default asyncI18nProvider;
+```
+
+Like in this provider, it makes sense to include the primary language of your app synchronous.
```
djhi: s/synchronous/synchronously/

packages/react-admin/src/reducer/i18n/index.js
```diff
@@ -0,0 +1,16 @@
...
+import messages from './messages';
+import loading from './loading';
+import requested from './requested';
+
+export default initialLocale => {
+    return combineReducers({
+        locale: locale(initialLocale),
+        messages,
+        loading,
+        requested,
```
djhi: I don't understand this `requested` reducer. Where is it used ?

tkvw: It's not used in `react-admin`, so maybe it shouldn't be there. It's there so an user can query the locale being loaded. `{locale:'en',requested:null}` -> `{type:CHANGE_LOCALE,payload:'nl'}` -> `{locale:'en',requested:'nl'}` ->possibility to show ui loading `nl` -> `{type:CHANGE_LOCALE_SUCCESS}` -> `{locale:'nl',requested:null}`

djhi: I think this should stay in userland then

docs/Translation.md
```diff
@@ -343,3 +379,34 @@ translate('not_yet_translated', { _: 'Default translation' })
 ```
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
+
+## Using asynchronous messages loading 
+
+By default React Admin loads all configured lanuages and their messages on startup. If you have very large language files, you can load messages asynchronously.
```
djhi: s/lanuages/languages/

djhi:
Well, it seems we have a problem with our e2e tests on travis only. I can confirm the tests pass locally for your PR. I'm investigating

djhi:
We finally fixed the e2e tests issue. Can you please rebase your PR ?

fzaninotto:
I don't understand why the `yarn.lock` was modified, since there is no modification of the `package.json`.... Can you remove the change from that file (that would remove the need for a rebase)?

tkvw:
My bad..

packages/react-admin/src/Admin.js
```diff
@@ -29,10 +29,10 @@ const Admin = ({
     dashboard,
     history,
     locale,
-    messages = {},
     menu = Menu,
     catchAll,
     dataProvider,
+    i18nProvider = locale => this.props.messages[locale],
```
fzaninotto: please extract this function to a `defaultI18nProvider` to avoid redefining a new function (and therefore failing the prop identity check) whenever the `<Admin>` component is rerendered.

packages/react-admin/src/Admin.js
```diff
@@ -63,7 +66,7 @@ const Admin = ({
 
     return (
         <Provider store={store}>
-            <TranslationProvider messages={messages}>
+            <TranslationProvider>
```
fzaninotto: you need to update the `CustomApp` documentation, too.

packages/react-admin/src/i18n/TranslationProvider.js
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import Polyglot from 'node-polyglot';
 import { connect } from 'react-redux';
 import { compose, withContext } from 'recompose';
-
 import defaultMessages from 'ra-language-english';
+import defaultsDeep from 'lodash/defaultsDeep';
```
fzaninotto: Use `lodash.defaultsdeep` instead (already a dependency)

packages/react-admin/src/sideEffect/saga/i18n.js
```diff
@@ -0,0 +1,33 @@
...
+    function* loadInitialMessages() {
+        const initialLocale = yield select(getLocale);
+        yield put(changeLocale(initialLocale));
+    }
+
+    return function*() {
+        yield all([
+            takeLatest(CHANGE_LOCALE, loadMessages),
+            takeLatest(USER_LOGOUT, loadInitialMessages),
+            call(loadInitialMessages),
```
fzaninotto: I'm afraid that this one will cause an initial double render of the `<Admin>` children. Can you check that it doesn't?

packages/react-admin/src/sideEffect/saga/i18n.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+    function* loadInitialMessages() {
+        const initialLocale = yield select(getLocale);
+        yield put(changeLocale(initialLocale));
+    }
+
+    return function*() {
+        yield all([
+            takeLatest(CHANGE_LOCALE, loadMessages),
+            takeLatest(USER_LOGOUT, loadInitialMessages),
```
fzaninotto: What's the rationale of doing it on user logout?

tkvw: the messages are removed from the store using the `resettableAppReducer`, so they should be reloaded on user logout. Also the user locale can be different from the site default locale.

fzaninotto: That means, IMO, that the *initial state* of the i18n reducer is wrong.

packages/react-admin/src/sideEffect/saga/i18n.js
tkvw: Yes I had the same concern, my observations are that it does when the messages are loaded async, but synchronous I don't see this effect. In my custom app I have a loader which `races` the `CHANGE_LOCALE_SUCCESS` and `CHANGE_LOCALE_FAILURE` before displaying `react-admin`, maybe this should be added?

packages/react-admin/src/sideEffect/saga/i18n.js
tkvw: I see what you're saying, but my problem is with the how. The messages aren't known on initial render of `Admin`, not until the `i18n` saga finishes loading them. I can add loading state to `Admin`, but that complicates `Admin` more, any suggestions?

packages/react-admin/src/sideEffect/saga/i18n.js
fzaninotto: Let's summarize the problems: 

1. The loading of the messages can be async to avoid loading all messages at startup
2. We don't want to block the rendering of the app while it waits for an async message fetch
3. But the app requires a default locale, loaded synchronously, for the first render
4. The locale can depend on the user (a custom `authSaga` can dispatch a `CHANGE_LOCALE` upon login).

The solution, IMO, is to implement two actions in the `i18nProvider`:

- `GET_DEFAULT_MESSAGES`: must be synchronous, is called by the `<Admin>` component, then passed to the i18n *reducer* to initialize the default i18n state
- `CHANGE_LOCALE`: can be asynchronous, and is called by the i18n *sagas*

That way, no need to reset messages upon logout, because the state has a default set of i18n messages.

The way I see it, the i18nProvider could still be a simple function taking the locale as only argument, and returning a promise for messages (if the locale argument is defined), or directly messages (if no locale is passed). Or, it could be a more structured function, similar to the `dataProvider`, taking type and params, as follows:

```js
const i18nProvider = (messages, defaultMessages) => (type, params) => {
    if (type === 'GET_DEFAULT_MESSAGES') {
        return defaultMessages;
    }
    if (type === 'CHANGE_LOCALE') {
        return messages[params.locale];
    }
    throw new Error('Undefined action type', type); 
}
```

Having to pass both a `messages` and a `defaultMessages` props to the `<Admin>` component will be cumbersome, so I guess we can take advantage of the BC break in 2.0 and ask only for one parameter: the `i18nProvider`. The default `i18nProvider` would simply be:

```js
import englishMessages from 'ra-language-english'
const i18nProvider = (type, params) => englishMessages;
```

What do you think?

tkvw:
Thanks for your feedback! Very nice way of solving this. I've refactored the code to reflect the changes, I've changed `i18nProvider` to `messagesProvider` which is a better name in my opinion. Let me know what you think.

docs/Translation.md
```diff
@@ -343,3 +389,52 @@ translate('not_yet_translated', { _: 'Default translation' })
...
+            return messages[locale]
+                ? typeof messages[locale] === 'function'
+                  ? messages[locale]()
+                  : messages[locale]
+                : Promise.reject(`Locale ${locale} is not supported`);
+        default:
+            throw new Error(`Undefined messagesProvider action type ${type}`);
+    }
+};
+export default createMessagesProvider(undefined, null);
```
fzaninotto: bad export

examples/simple/sagas.js
```diff
@@ -0,0 +1,28 @@
+import { parse } from 'query-string';
+import { getLocale, changeLocale } from 'react-admin';
+import { call, put, select, takeLatest } from 'redux-saga/effects';
+import { LOCATION_CHANGE } from 'react-router-redux';
+
+const localeParam = 'locale';
+
+function* handleChangeLocaleAction({ payload: { search } }) {
```
fzaninotto: I'm not fan of that use case - we shouldn't encourage messing up with the url. Changing the locale should be a function exposed in the UI. Besides, to illustrate how a custom saga can change the messages, why don't you illustrate the use of cookies for the locale?

docs/Translation.md
```diff
@@ -12,27 +12,37 @@ The react-admin interface uses English as the default language. But it also supp
...
 - a `locale` prop expecting a string ('en', 'fr', etc), and
-- a `messages` prop, expecting a dictionary object.
+- a `messagesProvider` prop, expecting an `messagesProvider` as explained below.
+
+## `messagesProvider`
+
+Provides messages for a specific locale. React-admin ships with a `createMessagesProvider` factory to easily create your own `messagesProvider`. 
+`createMessagesProvider` requires a single argument `messageBundles` and allows an optional `defaultMessageBundle`. 
+* `messageBundles`: required, an object with a language pack per locale. These language packs can be resolved async by returning a Promise.
+* `defaultMessageBundle`: optional, a default language pack where the application starts with. The `defaultMessageBundle` must always be resolved synchronous and defaults to the messages from `ra-language-english`.
```
fzaninotto: no, it's not optional, on the contrary: it's THE required parameter.

docs/Translation.md
```diff
@@ -12,27 +12,37 @@ The react-admin interface uses English as the default language. But it also supp
 To handle translations, the `<Admin>` component supports:
 
 - a `locale` prop expecting a string ('en', 'fr', etc), and
-- a `messages` prop, expecting a dictionary object.
+- a `messagesProvider` prop, expecting an `messagesProvider` as explained below.
+
+## `messagesProvider`
+
+Provides messages for a specific locale. React-admin ships with a `createMessagesProvider` factory to easily create your own `messagesProvider`. 
+`createMessagesProvider` requires a single argument `messageBundles` and allows an optional `defaultMessageBundle`.
```
fzaninotto: I'm not fan of this `createMessagesProvider`. It obfuscates the logic behind the provider, and it introduces a second way of doing i18n - which is hard to explain. 

I think you can document how to pass custom messages simply using the following example:

```js
const messages = {
  en: englishMessages,
  fr: frenchMessages,
};
const i18nProvider = (type, params) => {
    if (type === 'GET_DEFAULT_MESSAGES') {
        return englishMessages;
    }
    if (type === 'CHANGE_LOCALE') {
        return messages[params.locale];
    }
    throw new Error('Undefined action type', type); 
}
```

docs/Translation.md
```diff
@@ -125,6 +135,42 @@ class LocaleSwitcher extends Component {
 export default connect(undefined, { changeLocale: changeLocaleAction })(LocaleSwitcher);
 ```
 
+## Changing the locale using the browser URL
```
fzaninotto: Not the best illustration for a custom saga IMO (see other comment on that code snippet)

docs/Translation.md
```diff
@@ -12,27 +12,37 @@ The react-admin interface uses English as the default language. But it also supp
 To handle translations, the `<Admin>` component supports:
 
 - a `locale` prop expecting a string ('en', 'fr', etc), and
-- a `messages` prop, expecting a dictionary object.
+- a `messagesProvider` prop, expecting an `messagesProvider` as explained below.
```
fzaninotto: I liked the `i18nProvider` name. The term *messages* isn't specific enough IMO (it can relate to notifications in many people's mind).

packages/react-admin/src/reducer/i18n/index.js
```diff
@@ -0,0 +1,14 @@
+import { combineReducers } from 'redux';
+import locale from './locale';
+import messages from './messages';
+import loading from './loading';
+
+export default (initialLocale, defaultMessages) => {
+    return combineReducers({
```
fzaninotto: no need to `return` here, use the braceless function form.

packages/react-admin/src/reducer/index.js
```diff
@@ -5,16 +5,17 @@ import admin, {
...
 } from './admin';
-import localeReducer from './locale';
+import i18n, { getLocale as adminGetLocale } from './i18n';
 
-export default (customReducers, locale) =>
+export default (customReducers, locale, messages) =>
     combineReducers({
         admin,
-        locale: localeReducer(locale),
+        i18n: i18n(locale, messages),
```
fzaninotto: the function should be named `i18nReducer` for consistency.

packages/react-admin/src/sideEffect/saga/crudSaga.js
```diff
@@ -3,13 +3,15 @@ import auth from './auth';
 import crudFetch from './crudFetch';
 import crudResponse from './crudResponse';
 import referenceFetch from './referenceFetch';
+import i18n from './i18n';
 
 /**
  * @param {Object} dataProvider A Data Provider function
  */
-export default (dataProvider, authClient) =>
+export default (dataProvider, authClient, messagesProvider) =>
```
fzaninotto: i18nProvider

packages/react-admin/src/Admin.js
fzaninotto: Still required

tkvw:
Reverted back to `i18nProvider`, updated the docs and updated requested changes.

docs/Translation.md
```diff
@@ -12,27 +12,60 @@ The react-admin interface uses English as the default language. But it also supp
 To handle translations, the `<Admin>` component supports:
 
 - a `locale` prop expecting a string ('en', 'fr', etc), and
-- a `messages` prop, expecting a dictionary object.
+- a `i18nProvider` prop, expecting an `i18nProvider` as explained below.
 
-React-admin ships with the English locale by default. If you want to use another locale, you'll have to install a third-party package. For instance, to change the interface to French, install the `ra-language-french` npm package, then configure the `<Admin>` component as follows:
```
fzaninotto: React-admin can still work without provider and default to English, right?

docs/Translation.md
```diff
@@ -12,27 +12,60 @@ The react-admin interface uses English as the default language. But it also supp
...
+## `i18nProvider`
+
+Provides messages for a specific locale. React-admin ships with a default `i18nProvider` which supports the English language.  
+If you want to use another locale, you'll have to install a third-party package. For instance, to change the interface to French, install the `ra-language-french` npm package, then configure the `<Admin>` component as follows:
 
 ```jsx
 import React from 'react';
 import { Admin, Resource, resolveBrowserLocale } from 'react-admin';
```
fzaninotto: I'd import the `GET_DEFAULT_MESSAGES` and `GET_LOCALE_MESSAGES` constants, too.

docs/Translation.md
tkvw: Yes, the default provider loads the english messages. Not sure if this should be changed to an empty messages object, because the same bundle is loaded in the `TranslationProvider` as default messages.  Having no messages in the defaultProvider feels weird, let me know what you want.

tkvw: Yeah, so I rephrased it a bit, probably should have left the `React-admin ships with the English locale by default.` sentence in there, will change.

UPGRADE.md
```diff
@@ -63,6 +63,45 @@ import { jsonServerRestClient } from 'admin-on-rest';
...
+import React from 'react';
+import { Admin, Resource, resolveBrowserLocale, GET_DEFAULT_MESSAGES, GET_LOCALE_MESSAGES } from 'react-admin';
+import frenchMessages from 'ra-language-french';
+import englishMessages from 'ra-language-english';
+
+const messages = {
+  en: () => englishMessages,
+  fr: () => import('ra-language-french').then( bundle => bundle.default ),
+};
+const i18nProvider = (type, params) => {
```
djhi: This is already an advanced usage scenario. I'd rather keep this example the most simple possible using the default i18nProvider to have the same behavior as before. Then I'll introduce what this new provider allows one to do in a separate paragraph (something like *Lazily loading a locale messages*)

Phocea:
I have a question. 
Since the proper loading of the application will depend on an async call to fetch the locale, should we use something like [PersistGate](https://github.com/rt2zz/redux-persist/blob/6547840c3c1971b6af602c5400fa58a7efa2540b/docs/PersistGate.md), to ensure the App is ready ?

More over, the whole logic of [redux persis](https://github.com/rt2zz/redux-persist) could be used and the locale added to the store ?

Locale aside, having the possibility to at least use some kind of redux persistence mechanism without moving to a custom app would be a great addition to react-admin.

tkvw:
Hello @Phocea : there is no async call to fetch the locale. This PR focuses at loading the messages of a locale async. The provider requires the initial application messages to be loaded synchronous, so the loading of the application can continue. You are allowed to trigger a `changeLocale` from a startup saga to setup the correct locale of the user and load the messages sync/async (you're free to implement whatever you want in the `i18nProvider`)

UPGRADE.md
```diff
@@ -63,6 +63,45 @@ import { jsonServerRestClient } from 'admin-on-rest';
...
+import frenchMessages from 'ra-language-french';
+import englishMessages from 'ra-language-english';
+
+const messages = {
+  en: englishMessages,
+  fr: frenchMessages,
+};
+const i18nProvider = (type, params) => {
+    if (type === GET_DEFAULT_MESSAGES) {
+        return messages.en; // The default messages should always resolve synchronous!!
```
djhi: s/synchronous/synchronously/

UPGRADE.md
```diff
@@ -63,6 +63,45 @@ import { jsonServerRestClient } from 'admin-on-rest';
 import jsonServerRestClient from 'ra-data-json-server';
 ```
 
+## Introduced `i18nProvider` for loading of locale messages
+
+Previously you would supply `react-admin` with a `messages` prop object to allow alternative locales, this should be replaced with an `i18nProvider`. The new `i18nProvider` allows you to load the messages asynchronously. @See [`i18nProvider`](./Translation.md#i18nProvider)
```
djhi: What's the `@` before `See` ?

UPGRADE.md
```diff
@@ -63,6 +63,45 @@ import { jsonServerRestClient } from 'admin-on-rest';
...
 ```
 
+## Introduced `i18nProvider` for loading of locale messages
+
+Previously you would supply `react-admin` with a `messages` prop object to allow alternative locales, this should be replaced with an `i18nProvider`. The new `i18nProvider` allows you to load the messages asynchronously. @See [`i18nProvider`](./Translation.md#i18nProvider) 
+
+```jsx
+// before
+import { enMessages } from 'admin-on-rest';
+const messages = { 'en': enMessages };
```
fzaninotto: the two examples are not the same, since you're passing one language here and two in the second snippet. An upgrade guide should show how to achieve *the same thing* on the two versions.

docs/CustomApp.md
```diff
@@ -52,9 +52,17 @@ import { UserList, UserEdit, UserCreate } from './User';
 // your app labels
 import messages from './i18n';
 
+const i18nProvider = (type, params) => {
+    if (type === 'CHANGE_LOCALE') {
+        return messages[params.locale];
+    }
+    throw new Error('Undefined action type', type);
```
fzaninotto: this won't work, the initial message is `GET_DEFAULT_MESSAGES` and it will be rejected. Besides, isn't it `GET_LOCALE_MESSAGES`?

UPGRADE.md
tkvw: jsdoc notation, shouldn't be there

fzaninotto:
@tkvw Thanks a lot for your contribution, it is much appreciated!

As you may have seen on this PR, managing contributions requires a lot of time on our part. In the near future, we want to focus on releasing the 2.0 rather than adding new features. If you want to give us a hand, please triage new issues/PRs, fix bugs, and write documentation.

tkvw:
@fzaninotto : np! Thanks for your inputs. I'll try to be more secure on my side the next time. Can you add labels 2.0.0 milestone labels to the PR's you want to include, so I can focus on them?

fzaninotto:
If you can start by reading every open issue and seeing if you can help reproduce / localize the reported bugs, that would be a great help/

PR #1280, opened 2017-12-11
Add options prop in dashboard menu item

bodo22:
To customize the DashboardMenuItem component, e.g. the MenuItem's `style` prop, the `options` prop is spread at the end of DashboardMenuItem.


packages/react-admin/src/mui/layout/DashboardMenuItem.js
```diff
@@ -16,20 +16,22 @@ const styles = {
     },
 };
 
-const DashboardMenuItem = ({ classes, onClick, translate }) => (
+const DashboardMenuItem = ({ classes, onClick, translate, options }) => (
```
djhi: why not just `...props` ?

bodo22: I followed naming in most input components but now realized the most remaining prop name is props :)

packages/react-admin/src/mui/layout/DashboardMenuItem.js
```diff
@@ -16,20 +16,22 @@ const styles = {
     },
 };
 
-const DashboardMenuItem = ({ classes, onClick, translate }) => (
+const DashboardMenuItem = ({ classes, onClick, translate, props }) => (
```
djhi: Not what I meant. Why don't you simply pass all extra props ?

```js
const DashboardMenuItem = ({ classes, onClick, translate, ...props }) => (
```

bodo22: I was following the pattern used in most input components, e.g. `BooleanInput`, `TextInput`. Should I still change it?

djhi: Yes please, we're going to apply this to all components anyway for `next`

djhi: Besides, sorry but you'll have to rebase as we updated `next` to add `className` to all components too

djhi:
The commits in this PR have been messed up. Can you either clean them or make a new PR ?

PR #1281, opened 2017-12-12
Connecting saga and redux-form through promises proposal

tkvw:
This PR allows direct communication from redux-saga to redux-form through promises. It uses a customized implementation of https://github.com/mhssmnn/redux-form-saga which is a smart solution to do this. 

What it basically does is to change the actionCreator to emit a promise instead of a pojo (plain old javascript object) which can be handed to handleSubmit of redux-form. The action creator implements methods `success` and `failure` which emit a pojo action object which is monitored by a saga and this saga resolves or rejects the promise. 

Because it implements static accessors for `SUCCESS`, `FAILURE` and `LOADING` the dataActions look a bit cleaner. 

I implemented this with a basic callback `onError` and `onSuccess` in `Create` and `Edit` (which actually don't make much sense now, because of the redirect). But it should give some insight in the power of the pattern used. This would easily allow implementation of multiple form identifiers in redux-form and server-side validation callback (simple transform .catch to redux's ValidationError). 

I had to change the signature of the action creators to accept an object instead of params (which looks cleaner imho as well), because the actionCreator can be called with or without dispatch (the latter simple emits a default pojo action). 

Not sure if you're interested in this for `ra-admin` , but I am going to use it in my project and thought to share.

### DIFF
--- a/packages/react-admin/src/mui/detail/Edit.js
+++ b/packages/react-admin/src/mui/detail/Edit.js
@@ -9,7 +9,7 @@ import Header from '../layout/Header';
 import Title from '../layout/Title';
 import {
     crudGetOne as crudGetOneAction,
-    crudUpdate as crudUpdateAction,
+    crudUpdate,
 } from '../../actions/dataActions';
 import DefaultActions from './EditActions';
 import translate from '../../i18n/translate';
@@ -100,17 +100,24 @@ export class Edit extends Component {
     }
 
     updateData(resource = this.props.resource, id = this.props.id) {
-        this.props.crudGetOne(resource, id, this.getBasePath());
+        this.props.crudGetOne({
+            resource,
+            id,
+            basePath: this.getBasePath(),
+        });
     }
 
-    save = (record, redirect) => {
-        this.props.crudUpdate(
-            this.props.resource,
-            this.props.id,
-            record,
-            this.props.data,
-            this.getBasePath(),
-            redirect
+    save = (record, redirect, dispatch) => {
+        return crudUpdate(
+            {
+                resource: this.props.resource,
+                id: this.props.id,
+                data: record,
+                previousData: this.props.data,
+                basePath: this.getBasePath(),
+                redirectTo: redirect,
+            },
+            dispatch
         );
     };
 
@@ -186,7 +193,6 @@ Edit.propTypes = {
     actions: PropTypes.element,
     children: PropTypes.node,
     crudGetOne: PropTypes.func.isRequired,
-    crudUpdate: PropTypes.func.isRequired,
     data: PropTypes.object,
     hasDelete: PropTypes.bool,
     hasShow: PropTypes.bool,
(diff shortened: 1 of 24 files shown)


packages/react-admin/src/mui/detail/Edit.js
```diff
@@ -100,18 +100,29 @@ export class Edit extends Component {
...
-        this.props.crudUpdate(
-            this.props.resource,
-            this.props.id,
-            record,
-            this.props.data,
-            this.getBasePath(),
-            redirect
-        );
+    save = (record, redirect, dispatch) => {
+        return crudUpdate(
```
djhi: why not connecting this action ? Connected action creators do not return anything ?

packages/react-admin/src/mui/form/SimpleForm.js
```diff
@@ -11,7 +11,9 @@ const formStyle = { padding: '0 1em 1em 1em' };
 
 export class SimpleForm extends Component {
     handleSubmitWithRedirect = (redirect = this.props.redirect) =>
-        this.props.handleSubmit(values => this.props.save(values, redirect));
+        this.props.handleSubmit((values, dispatch) =>
```
djhi: I can find any reference to `dispatch` in redux-form. Where does it come from ?

tkvw: https://github.com/erikras/redux-form/blob/6b69da594c369bcdc8acd7a14f0af605c0abbb8d/src/handleSubmit.js#L39

packages/react-admin/src/mui/detail/Edit.js
tkvw: The action createor emits a promise so can't be connected as a normal actions object (where each action creator should create a pojo action). That's why it receives dispatch as second parameter. It will dispatch the action internally and return a promise to listen to.

packages/react-admin/src/mui/form/SimpleForm.js
djhi: Even better: https://redux-form.com/7.2.0/docs/api/reduxform.md/#-code-onsubmit-function-code-optional-

I was afraid it was not documented

djhi:
Can you explain the value of this approach ? What does it solve ?

tkvw:
My most important use case (and initial need for redux-form-saga) is server-side validation, I want the field validation to return to the redux-form.

Another use case I am working on is so allow a custom `action` on the dataProvider. I think the `verbs` defined by you are great but limit me in executing arbitrary, none REST, actions, like `/dog/2/bark` . These actions would work great with the promises, because I can do something like: 
```
crudCustom(path='/dog/2/bark').then(didBark=> updateDidBarkComponentState())
```
These actions belong to the component, require to use the dataprovider for setup headers etc and are to simple to be handled by a saga. (My editor is currently opened on a `<Custom>` component to handle custom form submission using `crudCustom` like `<Create>` and `<Update>`. 

Interesting sideeffects: 
* this approach sets the form `submitting` property to true which would allow the gui to show the form is submitting (in case of long during submits). 
* this decouples the need of one single form identifier `record-form`, for example a `reset` instruction result can be returned to the promise when using Create instead of explicitly resetting `record-form`

tkvw:
I think this should be in userland as well, closing for now.

PR #1283, opened 2017-12-12
Xmysql REST client

soaserele:
[Xmysql](https://github.com/o1lab/xmysql) - one command to generate REST APIs for any MySql database.


djhi:
As explained in #1282:

> Nice! But this is not something we're going to maintain. 
> I suggest you create a standalone package for it, and open a PR in the restClients documentation, in the section dedicated to third-party clients.

PR #1294, opened 2017-12-14
WithPermission: Component can unmount before setting state.

tkvw:
Minor issue with switching path before permissions are resolved.  This happens if an unauthorized user browses to a restricted path directly and is redirected to login.


djhi:
Cannot reproduce the problem on next anymore, probably fixed with #1301. Can you confirm and close if needed ?

tkvw:
Yes, this looks fixed.

PR #1271, opened 2017-12-07
Allow logout redirect

tkvw:
Currently logout always redirects to `/login`, this PR allows to redirect to another page. 

Usage: 
```
<Admin 
   logout={()=><Logout redirectTo="/"/>}..
```

PR #1295, opened 2017-12-14
Reinitialize resources after logout

tkvw:
Resources are removed from the redux store after logout (which is fine if the user is redirected to `/login`), but if the user is redirected to another page the resources aren't reinitialized. This PR simply calls `getPermissions()` to  trigger this if needed.

### DIFF
--- a/packages/react-admin/src/AdminRoutes.js
+++ b/packages/react-admin/src/AdminRoutes.js
@@ -23,7 +23,9 @@ export class AdminRoutes extends Component {
 
     componentDidUpdate(prevProps, prevState) {
         if (
+            this.props.resources &&
             this.props.resources.length === 0 &&
+            prevProps.resources &&
             prevProps.resources.length !== 0
         ) {
             this.getPermissions();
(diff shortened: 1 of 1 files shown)


fzaninotto:
If I understand correctly, this is a problem only once #1271 is merged?

tkvw:
Correct. What's going on: Resources are only declared on initial mount of `<AdminRoutes>`, this is not an issue when redirecting to `/login`, because `<Switch>`  takes care of unmounting the `<AdminRoutes>` component. But if you don't redirect to `/login` `<AdminRoutes>` will stay in previous state, and because `resources` is cleared in the redux store because of `USER_LOGOUT` there aren't any active routes anymore. This PR simply takes care of the resources being declared after logout by checking if the resources count changed to 0.

djhi:
Is this still needed ?

tkvw:
Yes, I think this is still the case.  `getPermissions` is only called in `componentWillMount` of AdminRoutes.js. If you redirect to a `customRoute (with layout)` after logout (instead of `/login`), the getPermissions will still reflect the previous permissions, because the router didn't move away from `AdminRoutes`'s  `<Switch>`.

djhi:
Unfortunately, the solution here is not an option anymore. I think we should introduce an `isLoggedIn` entry in the redux store.

PR #1299, opened 2017-12-16
[RFR]: Show loading component while resources are being declared.

tkvw:
When resources are being loaded async, the `NotFound` component is shown while they are not declared. This behaviour can be shown by: 
```
    <Admin
        authClient={authClient}
        dataProvider={delayedDataProvider}
        title="Example Admin"
        locale="en"
        messages={messages}
    >
        {permissions =>
            new Promise(resolve => {
                setTimeout(
                    () =>
                        resolve([
                            <Resource
                                name="posts"
                                list={PostList}
                                create={PostCreate}
                                edit={PostEdit}
                                show={PostShow}
                                remove={Delete}
                                icon={PostIcon}
                            />,
                        ]),
                    1000
                );
            })}
    </Admin>,
```
This is the current behaviour: 

After applying this PR:

PR #1301, opened 2017-12-18
[RFR] Review Resource component

djhi:
- [x] Review Resource component
- [x] Updated tests
- [x] Updated documentation

Also included the Loading component from #1299 by @tkvw (Thanks :))


packages/ra-example/posts.js
```diff
@@ -153,27 +153,16 @@ export const PostCreate = ({ ...props }) => (
         <SimpleForm
             toolbar={<PostCreateToolbar />}
             defaultValue={{ average_note: 0 }}
-            validate={values => {
```
fzaninotto: not good, please use another view for debugging field-level validation, we need this one to test form-level validation

packages/react-admin/src/AdminRoutes.js
```diff
@@ -17,16 +15,11 @@ export class AdminRoutes extends Component {
     // Can't use null or undefined here as authClient may return any those values
     state = { permissions: initialPermissions };
```
fzaninotto: I'd set the `childrenToRender` key to `[]` to make it explicit.

packages/react-admin/src/AdminRoutes.js
```diff
@@ -40,98 +33,115 @@ export class AdminRoutes extends Component {
...
+        if (
+            typeof nextProps.children === 'function' &&
+            nextState.permissions !== this.state.permissions
+        ) {
+            const { children } = nextProps;
+
+            const childrenFuncResult = children(nextState.permissions);
+            if (childrenFuncResult.then) {
+                childrenFuncResult.then(resolvedChildren => {
+                    this.setState({
```
fzaninotto: Is it a good practice to update the state (`childrenToRender`) while it was just updated (`premissions`)? Won't it cause a double render?

djhi: Yes but it only renders a Loading screen if there are no children to render anyway. Are you worried about updated permissions ? In this case, yes too but I don't see any other way to do it

packages/ra-example/posts.js
```diff
@@ -172,7 +183,6 @@ const emptyKeycode = [];
 // Store the validation functions array in a variable to avoid
```
fzaninotto: please remove the comment, too

PR #1315, opened 2017-12-20
[RFR] Apply recommandations for running Jest on Travis

djhi:
See Jest [documentation](https://facebook.github.io/jest/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server)


fzaninotto:
Are you sure this improves test duration?

djhi:
Improve no, but at least it prevent it from timing out

PR #1316, opened 2017-12-20
[Fix]: Set notification message to empty if no notification

tkvw:
If `hideNotification` is called from a sideEffect the notification can become undefined. Message is a required prop, so simply display empty message.

### DIFF
--- a/packages/react-admin/src/mui/layout/Notification.js
+++ b/packages/react-admin/src/mui/layout/Notification.js
@@ -60,9 +60,13 @@ class Notification extends React.Component {
             <Snackbar
                 open={this.state.open}
                 message={
-                    notification &&
-                    notification.message &&
-                    translate(notification.message, notification.messageArgs)
+                    (notification &&
+                        notification.message &&
+                        translate(
+                            notification.message,
+                            notification.messageArgs
+                        )) ||
+                    ''
                 }
                 autoHideDuration={
                     (notification && notification.autoHideDuration) ||


djhi:
Wouldn't it better to render `null` instead of the `Snackbar` ?

tkvw:
Nope, you would lose the `fade-out` `fade-in` transformation effect on state open.

djhi:
As we fixed the timeout issue in tests on `next` (due to the jest update), can you rebase ?

fzaninotto:
I don't understand the use case, can you provide a more detailed scenario to reproduce the bug?

tkvw:
If a request fails for some reason (401) `crudResponse` saga will show a Snackbar error with the error message. `auth` saga will push to login and hide the notification. The `hideNotication` will pop the notification from the notification queue, but the Snackbar is still rendering (property `open` is changed though). As it expects an error message and none is available, it will throw an error in the console, and this is ugly. A lot of words for a PR with only `|| ''` 😄

fzaninotto:
Well, that means that the proper fix is to avoid rendering the snackbar in that case...

tkvw:
Imo not, because the Snackbar has a `disappearing` transition when `open` is changed to `false`. So if you simply remove it from the DOM, it will not disappear with a transition but suddenly disappear. 
If you want to remove the snackbar, first a `hide` action should be triggered; this should set the `open` state to false and after a while (how long the transition takes) a close action should be triggered which actually removes the notification from the queue and removes the snackbar from the DOM.

fzaninotto:
I can't reproduce any of the problems you try to fix:

- an empty snackbar showing briefly upon 401 response
- an unexpected error in the console (apart from the one logged by the `crudResponse` saga when catching a `FAILURE`

Can you provide a minimal CodeSandbox to reproduce the problem?

tkvw:
As I can't reproduce the behaviour myself for now, I am closing this issue. If it reappears, I'll reopen it.

PR #1317, opened 2017-12-20
[RFR] Upgrade to MUI beta 24

djhi:
- [x] `Sidebar`, `Notification` and `FilterButton` : `OnRequestClose` to `onClose`
- [x] `Toolbar`: `noGutter` to `disableGutters`
- [x] Different handling of breakpoints between MUI `Hidden` component and our `Responsive` component => `AppBarMobile` and the main `Header` are both displayed on small devices
- [x] Typography misused (warning in both console and tests)


packages/react-admin/src/mui/layout/AppBarMobile.js
```diff
@@ -33,39 +33,29 @@ const styles = {
     },
 };
 
-class AppBarMobile extends Component {
-    handleLeftIconButtonTouchTap = event => {
```
fzaninotto: weird that this wasn't used anymore...

packages/react-admin/src/mui/layout/Layout.js
```diff
@@ -83,7 +83,7 @@ const Layout = ({
     <MuiThemeProvider theme={createMuiTheme(theme)}>
         <div className={classnames(classes.root, className)} {...rest}>
             <div className={classes.appFrame}>
-                <Hidden xsDown>
+                <Hidden smDown>
```
fzaninotto: needs more thorough tests ; unless something changed on the mui side, this looks odd.

djhi: Something changed. It's not inclusive anymore: https://github.com/mui-org/material-ui/releases/tag/v1.0.0-beta.24

packages/react-admin/src/mui/layout/AppBarMobile.js
djhi: Agreed, but it's not and the button is working

PR #1322, opened 2017-12-21
[RFR] Make all buttons responsive

djhi:
- [x] Display Icon only on small devices
- [x] Fix CreateButton display on small devices

screen

screen

### DIFF
--- a/packages/react-admin/src/mui/button/DeleteButton.js
+++ b/packages/react-admin/src/mui/button/DeleteButton.js
@@ -16,8 +16,8 @@ const styles = {
         display: 'inline-flex',
         alignItems: 'center',
     },
-    iconPaddingStyle: {
-        paddingRight: '0.5em',
+    label: {
+        paddingLeft: '0.5em',
     },
 };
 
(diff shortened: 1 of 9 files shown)


tkvw:
I think this makes https://github.com/marmelab/admin-on-rest/pull/1250 obsolete?

djhi:
Yes, sorry. I commented on your PR but forgot to actually submit the comment :stuck_out_tongue_closed_eyes:

packages/react-admin/src/mui/button/DeleteButton.js
```diff
@@ -29,16 +30,31 @@ const DeleteButton = ({
...
-        className={classnames(classes.link, className)}
-        component={Link}
-        color="accent"
-        to={`${linkToRecord(basePath, record.id)}/delete`}
-        {...rest}
-    >
-        <ActionDelete className={classes.iconPaddingStyle} />
-        {label && translate(label)}
-    </Button>
+    <Responsive
```
fzaninotto: I'm wondering, why do you use `<Responsive`> at the top level, forcing you to repeat the `<Button>` code, instead of putting it inside the `<Button>`? Is it to avoid introducing a `<span>` to have several children when there is both text and icon?

djhi: What do you think of:

```jsx
    <Button
        className={classnames(classes.link, className)}
        component={Link}
        color="accent"
        to={`${linkToRecord(basePath, record.id)}/delete`}
        {...rest}
    >
        <ActionDelete />
        <Responsive
            small={<span />}
            medium={<span>{label && translate(label)}</span>}
        />
    </Button>
``` ?

fzaninotto: much better, but the icon needs a special padding on Mobile

djhi: No we need a margin between the icon and the label. Solution is to put it on the label

fzaninotto:
How about refactoring our buttons to use a commun react-admin `<Button>` component?

PR #1324, opened 2017-12-21
[RFR] Ensure whole form is accessible on small devices

### LINKED ISSUE #1274: Toolbar hovers over content on mobile
I think this shouldn't be `fixed` but `static`.  The toolbar's width is 100%, position fixed and background white so it overflows the content on small screens. Or maybe there's some css trickery I am not aware of to fix this. 
https://github.com/marmelab/admin-on-rest/blob/9f4138a75ce23ffe17075abb50d5db4cb025d4e4/src/mui/form/Toolbar.js#L9

djhi:
fixes #1274

Before
screen

After
screen

### DIFF
--- a/packages/react-admin/src/mui/form/SimpleForm.js
+++ b/packages/react-admin/src/mui/form/SimpleForm.js
@@ -4,21 +4,22 @@ import { reduxForm } from 'redux-form';
 import { connect } from 'react-redux';
 import compose from 'recompose/compose';
 import { withStyles } from 'material-ui/styles';
-import withWidth from 'material-ui/utils/withWidth';
 import classnames from 'classnames';
 
 import getDefaultValues from './getDefaultValues';
 import FormInput from './FormInput';
 import Toolbar from './Toolbar';
 
-const styles = {
+const styles = theme => ({
     form: {
-        padding: '0 1em 1em 1em',
+        [theme.breakpoints.up('sm')]: {
+            padding: '0 1em 1em 1em',
+        },
+        [theme.breakpoints.down('sm')]: {
+            padding: '0 1em 5em 1em',
+        },
     },
-    formSmall: {
-        padding: '0 1em 5em 1em',
-    },
-};
+});
 
 const sanitizeRestProps = ({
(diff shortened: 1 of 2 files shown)


packages/react-admin/src/mui/form/SimpleForm.js
```diff
@@ -4,14 +4,20 @@ import { reduxForm } from 'redux-form';
...
+import withWidth from 'material-ui/utils/withWidth';
 import classnames from 'classnames';
 
 import getDefaultValues from './getDefaultValues';
 import FormInput from './FormInput';
 import Toolbar from './Toolbar';
 
 const styles = {
-    form: { padding: '0 1em 1em 1em' },
+    form: {
```
fzaninotto: no, the right way is to use `theme.breakpoints`, as in `mui/layout/Layouts.js`. No need to decorate with `withWidth`

packages/react-admin/src/mui/form/TabbedForm.js
```diff
@@ -12,12 +12,18 @@ import compose from 'recompose/compose';
 import Divider from 'material-ui/Divider';
 import Tabs, { Tab } from 'material-ui/Tabs';
 import { withStyles } from 'material-ui/styles';
+import withWidth from 'material-ui/utils/withWidth';
```
fzaninotto: same, not needed

packages/react-admin/src/mui/form/SimpleForm.js
```diff
@@ -67,6 +74,7 @@ export class SimpleForm extends Component {
             submitOnEnter,
             toolbar,
             version,
+            width,
```
fzaninotto: not needed

packages/react-admin/src/mui/form/TabbedForm.js
```diff
@@ -87,6 +94,7 @@ export class TabbedForm extends Component {
             toolbar,
             translate,
             version,
+            width,
```
fzaninotto: not needed

PR #1326, opened 2017-12-21
[Fix/RFR]: Limit valid query params and make sure filter is initialized.

tkvw:
When you want to browse to a list using the browser url with query params, you get an error as shown below, this PR fixes this.


djhi:
Thanks

PR #1329, opened 2017-12-22
Allow disabled option on Create / Save / Edit / Delete buttons

lgiraudel:
Hey,

I have this case where I allow user to see the Edit view based on READ permission and then disable Save button if he doesn't have EDIT permission. But unfortunately, the Save button can't be disabled, I have to create the whole RaisedButton when user doesn't have permission:

```
const CustomToolbar = ({ permissions, ...props }) => (
  <Toolbar {...props}>
    {canEdit(permissions, true) ? (
      <SaveButton redirect="list" submitOnEnter />
    ) : (
      <RaisedButton
        type="button"
        label="Save"
        icon={<ContentSave />}
        primary={true}
        disabled={true}
        style={{ margin: "10px 24px", position: "relative" }}
      />
    )}
  </Toolbar>
);

export const MyObjectEdit = props => (
  <Edit {...props}>
    {permissions => (
      <SimpleForm toolbar={<CustomToolbar permissions={permissions} />}>
         // ...
      </SimpleForm>
    )}
  </Edit>
);
```

I wanted something easier:

```
const CustomToolbar = ({ permissions, ...props }) => (
  <Toolbar {...props}>
      <SaveButton redirect="list" submitOnEnter disabled={canEdit(permissions, true)} />
  </Toolbar>
);
```

### DIFF
--- a/src/mui/button/CreateButton.js
+++ b/src/mui/button/CreateButton.js
@@ -23,16 +23,20 @@ const styles = {
     },
 };
 
-const CreateButton = ({
+export const CreateButton = ({
     basePath = '',
     translate,
     label = 'aor.action.create',
     width,
+    disabled,
 }) =>
     width === 1 ? (
         <FloatingActionButton
             style={styles.floating}
-            containerElement={<Link to={`${basePath}/create`} />}
+            containerElement={
+                disabled ? 'div' : <Link to={`${basePath}/create`} />
+            }
+            disabled={disabled}
         >
             <ContentAdd />
         </FloatingActionButton>
@@ -41,8 +45,11 @@ const CreateButton = ({
             primary
             label={label && translate(label)}
             icon={<ContentAdd />}
-            containerElement={<Link to={`${basePath}/create`} />}
+            containerElement={
+                disabled ? 'div' : <Link to={`${basePath}/create`} />
+            }
             style={styles.flat}
+            disabled={disabled}
         />
     );
 
@@ -51,6 +58,7 @@ CreateButton.propTypes = {
     label: PropTypes.string,
     translate: PropTypes.func.isRequired,
     width: PropTypes.number,
+    disabled: PropTypes.bool,
 };
 
 const enhance = compose(
--- a/src/mui/button/CreateButton.spec.js
+++ b/src/mui/button/CreateButton.spec.js
@@ -1,0 +1,31 @@
+import assert from 'assert';
+import { shallow } from 'enzyme';
+import React from 'react';
+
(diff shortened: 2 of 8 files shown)


src/mui/button/CreateButton.js
```diff
@@ -23,16 +23,20 @@ const styles = {
...
     label = 'aor.action.create',
     width,
+    disabled,
 }) =>
     width === 1 ? (
         <FloatingActionButton
             style={styles.floating}
-            containerElement={<Link to={`${basePath}/create`} />}
+            containerElement={
+                disabled ? 'div' : <Link to={`${basePath}/create`} />
```
fzaninotto: I think it doesn't make sense to show a disabled floating action button. Hide it in that case.

src/mui/button/CreateButton.spec.js
```diff
@@ -0,0 +1,31 @@
...
+        assert.equal(wrapper.type().muiName, 'FlatButton');
+    });
+
+    it('should be displayed as a <Link />', () => {
+        const wrapper = shallow(<CreateButton translate={translate} />);
+
+        assert.equal(wrapper.prop('containerElement').type, Link);
+    });
+
+    it('should not be displayed as a <Link /> when disable', () => {
```
fzaninotto: s/disable/disabled/

fzaninotto:
Oh, and I forgot: the `master` branch is feature-freezed. To add a new feature, you must use the `next` branch.

src/mui/button/CreateButton.js
lgiraudel: Really? My goal is to display a disabled button, maybe we can add another property to hide button when disabled?

fzaninotto: Again, from a usability perspective, a "disabled" floating action button is an aberration IMO. A [floating action button](https://material.io/guidelines/components/buttons-floating-action-button.html) is used for a promoted action. As it floats over the content, if it's not active, it's just obscuring the content behind it.

fzaninotto:
Bump. Any updates on this one?

lgiraudel:
Sorry, wasn't working on React those days so I didn't spend much time on it.

I'm ok with the Floating Action button, I was not aware about what it represent in Material UI.

I worked on the documentation, unfortunately there is no reference page for buttons so I had to create the whole documentation for them.
And I had some troubles to create a new page: links from the reference page are not working on my local environment (even on existing pages).

lgiraudel:
Don't have time to finish this and since I'm no longer working with React those days, it doesn't help.

Sorry :(

PR #1332, opened 2017-12-23
Update index.js

mattaylor:
GET request were broken because flattenObject is now exported from fetchUtils.

### DIFF
--- a/packages/ra-data-json-server/src/index.js
+++ b/packages/ra-data-json-server/src/index.js
@@ -1,7 +1,6 @@
 import { stringify } from 'query-string';
 import {
     fetchUtils,
-    flattenObject,
     GET_LIST,
     GET_ONE,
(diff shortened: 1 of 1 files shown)


mattaylor:
Please merge this as ra-data-server is currently broken

fzaninotto:
Thanks!

lakano:
Hi @fzaninotto !
Could you confirm me this is patched inside the beta1 please?
I have the same error « _reactAdmin.flattenObject is not a function » and I'm runnning on the 2.0.0-beta1 for all packages.
Merci :)

fzaninotto:
No, we've not redeployed the beta with the fix yet.

lakano:
Thank you for your reply. We've fixed that on the beta1 by moving it outside of the node_modules package and applying the fix like manually.

PR #1334, opened 2017-12-28
update CommentList example in docs

BATMAH69:
Deleted the assignment of the reference type in defaultProps.
It is a bad practice to use the reference types in defaultProps.


fzaninotto:
Where is this documented as a bad practice?

PR #1335, opened 2017-12-29
[RFR]: Add extendable adminsettings 

tkvw:
Add extendable adminSettings to allow basic configuration be passed arround.
This allows to `loginUrl` to be configurable, but can be extended for different usage as well.

### DIFF
--- a/packages/react-admin/src/Admin.js
+++ b/packages/react-admin/src/Admin.js
@@ -18,6 +18,7 @@ import Menu from './mui/layout/Menu';
 import Login from './mui/auth/Login';
 import Logout from './mui/auth/Logout';
 import TranslationProvider from './i18n/TranslationProvider';
+import DefaultSettings from './AdminSettings';
 
 const Admin = ({
     appLayout,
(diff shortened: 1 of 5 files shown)


fzaninotto:
I'm not fan of a general purpose settings object passed around like that. It tends to dilute the responsibility of each component.

Also, from what I see, you implement 3 different features in this PR:

- customize the url root. Why do you need that?
- disable console logging of HTTP errors. Why do you need that?
- redirection after login: can already be done by using an alternate `Login` component, and dispatch the `userLogin` action with a custom default path name.

tkvw:
> customize the url root. Why do you need that?

I think you mean login url. There's no imminent need, but I just dislike the static '/login' all over the code. 

> disable console logging of HTTP errors. Why do you need that?

I dislike messages in the console that I don't have influence over. `console.error` in crudResponse can't be disabled. 

> redirection after login: can already be done by using an alternate Login component, and dispatch the 
userLogin action with a custom default path name.

👍 

But I don't feel strong about this PR, so I will close it.

PR #1336, opened 2017-12-29
[RFR]: Login page bugfix

tkvw:
Because of passing down the additional props to the outer div of the LoginPage there were alot of console errors, because all redux-form props weren't sanitized. Normally redux-form should be wrapped around a Form, so I externalized the Form.

### DIFF
--- a/packages/react-admin/src/mui/auth/Login.js
+++ b/packages/react-admin/src/mui/auth/Login.js
@@ -34,15 +34,6 @@ const styles = theme => ({
     icon: {
         backgroundColor: theme.palette.secondary[500],
     },
-    form: {
-        padding: '0 1em 1em 1em',
-    },
-    input: {
-        marginTop: '1em',
-    },
-    button: {
-        width: '100%',
-    },
 });
 
 const sanitizeRestProps = ({
--- a/packages/react-admin/src/mui/auth/Login.js
+++ b/packages/react-admin/src/mui/auth/Login.js
@@ -1,21 +1,15 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { propTypes, reduxForm, Field } from 'redux-form';
-import { connect } from 'react-redux';
 import compose from 'recompose/compose';
-import Card, { CardActions } from 'material-ui/Card';
+import Card from 'material-ui/Card';
 import Avatar from 'material-ui/Avatar';
-import Button from 'material-ui/Button';
-import TextField from 'material-ui/TextField';
-import { CircularProgress } from 'material-ui/Progress';
 import LockIcon from 'material-ui-icons/LockOutline';
 import { withStyles } from 'material-ui/styles';
 
 import defaultTheme from '../defaultTheme';
-import { userLogin } from '../../actions/authActions';
-import translate from '../../i18n/translate';
 import Notification from '../layout/Notification';
+import LoginForm from './LoginForm';
 
 const styles = theme => ({
     main: {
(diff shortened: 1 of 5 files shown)


fzaninotto:
Excellent. Would you mind extracting the `LoginForm` to another file, and exporting it so that end user can reuse it?

tkvw:
Changed as requested.

packages/react-admin/src/sideEffect/saga/auth.js
```diff
@@ -33,7 +36,8 @@ export default authClient => {
                         type: USER_LOGIN_SUCCESS,
                         payload: authPayload,
                     });
-                    yield put(push(meta.pathName || '/'));
+                    const redirectTo = yield select(nextPathnameSelector);
```
fzaninotto: why that change? You're breaking a feature that was added to redirect to a specific page after login (`meta.pathName`).

fzaninotto: Also, I'm afraid than when the `nextPathnameSelector` function is called, the routing state is already in `/login`, so it doesn't make sense to redirect there?

tkvw: > why that change? You're breaking a feature that was added to redirect to a specific page after login (meta.pathName).

I wanted to remove the dependency of https://github.com/marmelab/admin-on-rest/blob/d82fb2dc9df9d75c5922197a07fd49dfe977128d/packages/react-admin/src/mui/auth/Login.js#L102 in the `loginForm` and I decided that it was best to put the logic to resolve the previous location in the saga instead of the component. The previous location state is reduced so available to select. But ignoring `meta.pathName` is an oversight, fixed it. 

> Also, I'm afraid than when the nextPathnameSelector function is called, the routing state is already in /login, so it doesn't make sense to redirect there?

Yes, but the previous location state is stored separately, @see: https://github.com/marmelab/admin-on-rest/blob/d82fb2dc9df9d75c5922197a07fd49dfe977128d/packages/react-admin/src/sideEffect/saga/auth.js#L59 so it is up for grabs.

packages/react-admin/src/mui/auth/Login.js
```diff
@@ -171,39 +88,18 @@ class Login extends Component {
...
-                errors.username = translate('ra.validation.required');
-            if (!values.password)
-                errors.password = translate('ra.validation.required');
-            return errors;
-        },
-    }),
-    connect(mapStateToProps, { userLogin }),
-    withStyles(styles)
-);
+const enhance = compose(withStyles(styles));
```
fzaninotto: no need to `compose` now that there is only one HOC.

packages/react-admin/src/mui/auth/Login.js
```diff
@@ -99,23 +65,8 @@ const sanitizeRestProps = ({
  *     );
  */
 class Login extends Component {
```
fzaninotto: You should turn it into a functional component since it has no other method than `render()`

packages/react-admin/src/sideEffect/saga/auth.js
```diff
@@ -33,7 +36,11 @@ export default authClient => {
                         type: USER_LOGIN_SUCCESS,
                         payload: authPayload,
                     });
-                    yield put(push(meta.pathName || '/'));
+                    let redirectTo = meta.pathName;
```
fzaninotto: Please use a ternary expression instead of a `let`

packages/react-admin/src/sideEffect/saga/auth.js
fzaninotto: Then you need to change the `userLogin` action signature, too.

packages/react-admin/src/sideEffect/saga/auth.js
tkvw: Not sure what you mean here, a ternary expression using `yield` in the expression?  Can you show what you mean?

djhi:
```js
const redirectTo = meta.pathName || yield select(nextPathnameSelector);
```
?

tkvw: Parser says no. 
```
Unexpected token (39:54)
> 39 |                     const redirectTo = meta.pathName || yield select(nextPathnameSelector);
     |                                                       ^
  40 |                     yield put(push(redirectTo || '/'));
```

djhi: Sorry:

```js
const redirectTo = yield meta.pathName || select(nextPathnameSelector);
```

tkvw: 👍 **stupid** of me, I didn't realize you can yield any value.

djhi:
Thanks!

Issue #785, opened 2017-06-22
Adding custom middleware


anuragagarwal561994:
Is there a way to add custom store middleware?
Right now I making custom Admin component to do so.

djhi:
No, there isn't. Can you explain your use case ?

anuragagarwal561994:
So when I am building my application I require logger or it becomes very confusing while managing states. For adding a logger right now I made an exact copy of the Admin from the source code and then just added `redux-logger` in it. There can be other middlewares which can be used in similar way.

djhi:
For the logger part, isn't the redux devtools enough ?
![screen](http://i.imgur.com/YMGF85J.png)

fzaninotto:
If you need a custom store, you'll have to swap the `<Admin>` component for a component of your own. Since `<Admin>` only has like 5 components, it shouldn't be too hard.

llioor:
@fzaninotto do you have a good example for implementing the whole admin component with in a custom app? On docs I can see just AppBar and Routes example but when I try to implement the "MyLayout" things became to be complicated.

PR #1337, opened 2017-12-29
[RFR]: Allow additional Redux middleware

tkvw:
I like to use additional redux middleware. It would be useful to allow this.


fzaninotto:
This is outside the 80% use cases IMO, and falls in the zone where developers must use a custom app.

I want to avoid making `<Admin>` a super component merging all the possible options of `redux`, `saga` and all.

fzaninotto:
also see: #785

PR #1338, opened 2017-12-29
[RFR]: Add context to AUTH_CHECK in authClient

tkvw:
When evaluating AUTH_CHECK in the `authClient` additional information is required. This PR supplies `resource`, `type`, `location` and routeParams to the authParams and allows additional `authParams` to be supplied on the `Resource` component. 

It also adds a `isLoggedIn` check on `WithPermissions`, because it should reevaluate when the `isLoggedIn` state changes.


fzaninotto:
> additional information is required

Who requires it? To do what?

You're adding *authorization* logic to the *authentication* layer, which is a bad thing IMO. Asking for *authentication* should not depend on the resource required - this is what authorizations are for.

So fine grained access to routes and resources should be done in `AUTH_GET_PERMISSIONS`, not in `AUTH_CHECK`

tkvw:
You're 100 % right, I misunderstood it, thanks for the feedback and clearing it up. I still think `WithPermissions` should be reevaluated on logout. I'll submit a separate PR for this.

PR #1339, opened 2017-12-29
[RFR]: Only redirect if error status indicates it's required

tkvw:
Changed how `CRUD_GET_ONE_FAILURE` is handled. Consider that the fetch to `users/1` fails with a 401.  This will trigger a `USER_LOGOUT`, but `CRUD_GET_ONE_FAILURE` will also be handled. The latter moves the user to `/users`, so effectively the following behaviour is happening: 
`users/1` -> `/login` -> `/users` -> `/login`. This is not desired, so I changed CRUD_GET_ONE_FAILURE only to redirect to basepath if status 404 (not found) or 403 (forbidden) is returned and show a notification accordingly.

### DIFF
--- a/packages/ra-language-english/index.js
+++ b/packages/ra-language-english/index.js
@@ -71,6 +71,7 @@ module.exports = {
             deleted: 'Element deleted',
             bad_item: 'Incorrect element',
             item_doesnt_exist: 'Element does not exist',
+            item_forbidden: "You're not authorized for this item",
             http_error: 'Server communication error',
         },
         validation: {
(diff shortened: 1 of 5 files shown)


fzaninotto:
Thanks for the PR.

I can reproduce the double redirection, but its not a big deal IMO. The screen doesn't redraw after the first redirection. So even if the URL changes twice, the content of the screen only changes once. 

Besides, I'm not fan of your solution, which ties the `crudResponse` saga too much with HTTP. If you look carefully, there is no HTTP code in the admin-on-rest code (only in providers). This is because a DataProvider can use HTTP, or WebSockets, or even a local data store - in which case, 404 and 401 don't mean anything. 

So my opinion on this is:

- we can live with the double redirection
- if it bothers you, please try and submit a patch that doesn't rely on HTTP status codes.

So I'm closing this PR because I won't merge the patch as is. Feel free to open a new PR if you have an alternative solution.

tkvw:
Thanks for your extensive answer, I understand your motivation and gave another whack at it @see: https://github.com/marmelab/admin-on-rest/pull/1349. 

The current behaviour is undesirable for me, because the history changes. So the redirect after the login is to `/list` instead of the initial item requested. I want to be able to send a link to a user, authorize and show the item.

PR #1341, opened 2018-01-02
check that window exists before checking for devToolsExtension

twDuke:


twDuke:
build timeout on yarn install can't seem to retry.

PR #1344, opened 2018-01-03
Check that window exists before checking for devToolsExtension

twDuke:
Using a-o-r as part of a gatsbyjs app, this change makes it possible to build.

### DIFF
--- a/src/Admin.js
+++ b/src/Admin.js
@@ -52,7 +52,9 @@ const Admin = ({
         initialState,
         compose(
             applyMiddleware(sagaMiddleware, routerMiddleware(routerHistory)),
-            window.devToolsExtension ? window.devToolsExtension() : f => f
+            typeof window !== 'undefined' && window.devToolsExtension
+                ? window.devToolsExtension()
+                : f => f
         )
     );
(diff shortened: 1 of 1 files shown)


twDuke:
I need some help with the linter, i'm having trouble running it locally using the make test command, tests run fine but then i get 
Running linter...
'.' is not recognized as an internal or external command,
operable program or batch file.
make: *** [lint] Error 1

I can run it using 
yarn run eslint

but i don't get the same error as the CI produces, i run it in the admin on rest repo folder so it should pick up the .eslintrc config. 

Also if someone could help me understand what the feedback from eslint means. I would be very thankfull.  
 Replace `·?·window.devToolsExtension()` with `⏎················?·window.devToolsExtension()⏎···············`

tkvw:
You can run make prettier to prettify the sources. Check the makefile for the command. You can download a make executable for gitbash for windows. This is what I am using. Run make test to test your changes before pushing to git and PR.

twDuke:
I have make running, make test works except that the lint fails. Thanks for the input.

fzaninotto:
Thanks!

PR #1345, opened 2018-01-03
[RFR]: Export component Loading

tkvw:
I think this component should be exported.

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -486,3 +486,22 @@ You can override the notification duration by setting the `autoHideDuration` pro
 ```
 
 **Tip**: if you use the `showNotification` action, then you can define `autoHideDuration` per message as the third parameter of the `showNotification` action creator.
+
+
+## Loading
+
+For consistency the `Loading` component of `react-admin` is exported. 
+
+Supported props: 
+
+Prop | Type | Default | Descriptions
+---|---|---|---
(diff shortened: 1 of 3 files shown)


fzaninotto:
Can you elaborate? In which case should end users use it?

I worry about it because every public component should be documented, maintained, and kept BC. We try to keep the public API surface small.

tkvw:
For consistency in the loading component, it allows the user to use it if needed. I use it for startup of the app (reauthentication and locale loading).

fzaninotto:
Then you should add it to the list of public components (in `docs/Referecne.md`) and document it (probably in `docs/Theming.md`).

tkvw:
Done

docs/Theming.md
```diff
@@ -486,3 +486,22 @@ You can override the notification duration by setting the `autoHideDuration` pro
 ```
 
 **Tip**: if you use the `showNotification` action, then you can define `autoHideDuration` per message as the third parameter of the `showNotification` action creator.
+
+
+## Loading
+
+For consistency the `Loading` component of `react-admin` is exported.
```
djhi: A description of the component's behavior (display a circular progress with optional messages, maybe add ing a link the MUI component too) and utility (display the same loading component as `react-admin` on custom pages for consistency)  would be better.

djhi:
Thanks!

PR #1350, opened 2018-01-05
[RFR]: Reevaluate permissions on user logout.

tkvw:
When an user logs out, the permissions of `WithPermissions` should be reevaluated.


djhi:
Thanks!

PR #1354, opened 2018-01-08
[RFR]: Support server validation

tkvw:
This PR introduces a validation provider to transform validation errors returned from the server to `redux-form` [SubmissionErrors](https://redux-form.com/7.2.0/docs/api/submissionerror.md/). This allows to inform the user of validation errors returned from the server.

The signature is simple: 
```
validationProvider = error => {Promise}
```
It should reject when it cannot convert the error (the default implementation rejects all errors) and resolve with the errors when it succeeds. 

Example: 
```
httpValidationProvider = error => 422 === error.status? Promise.resolve(error): Promise.reject(error);
```

An interesting sideeffect of this PR is, that the `redux-form` actions are triggered correctly, because a Promise is returned from `handleSubmit` @see: https://redux-form.com/7.2.0/docs/api/reduxform.md/#-code-onsubmit-function-code-optional-. 

All errors are translated before displaying. This allows to work with message labels returned from the server. 

For clarity the `redux-form` errors object looks like: 
```
{
    _error: <genericerror>,
    field_a:<error_label>,
    field_b:<error_label>,
}
```


tkvw:
This PR was getting unreadable, so I put some effort into refactoring my commits. Now each commit will tell the story of the entire PR, so it can be easier understood.

packages/react-admin/src/mui/form/GenericFormError.js
```diff
@@ -0,0 +1,27 @@
...
+import { withStyles } from 'material-ui/styles';
+
+const styles = theme => ({
+    root: {
+        backgroundColor: theme.palette.background.default,
+        padding: 2 * theme.spacing.unit,
+    },
+});
+
+const GenericFormError = ({ translate, error, classes, ...rest }) =>
```
djhi: Can you add a `className` prop as well and merge it with `classes.root` using `classnames` as we do in other components ?

fzaninotto:
I have an alternative to propose: what if developers could throw `SubmissionError` directly in their `dataProvider`? It's just a matter of filtering errors by type in the `crudFetch` sage, and letting redux-form errors bubble up. That way, errors from the data provider would be displayed in the form. You can keep the translation part of the PR.

I suggest that because I feel uncomfortable adding a new level of complexity : you add a new validation phase before submission, but with most data providers, it's done by the same controller. I prefer to leave the implementation details in the hands of the user.

What do you think?

tkvw:
Changed the implementation so the `dataProvider` can throw a `SubmissionError`. Added changes requested by @djhi and superseed https://github.com/marmelab/admin-on-rest/pull/1349. Added a `requestIdentifier` to `resolve/reject` the correct form promise.

packages/react-admin/src/mui/form/SanitizedForm.js
```diff
@@ -0,0 +1,58 @@
...
+    autofill,
+    submit,
+    resource,
+    redirect,
+    array,
+    ...rest
+}) => rest;
+
+const SanitizedForm = ({ renderLayout, ...props }) =>
+    renderLayout ? (
```
djhi: Can you rename it `render` ?

packages/react-admin/src/mui/form/SanitizedForm.js
```diff
@@ -0,0 +1,58 @@
...
+    translate,
+    autofill,
+    submit,
+    resource,
+    redirect,
+    array,
+    ...rest
+}) => rest;
+
+const SanitizedForm = ({ renderLayout, ...props }) =>
```
djhi: Can you rename `renderLayout` to `render` ?

packages/react-admin/src/mui/form/SanitizedForm.js
```diff
@@ -0,0 +1,58 @@
...
+    renderLayout ? (
+        renderLayout({
+            ...props,
+            getFormProps: sanitizeFormProps,
```
djhi: The name implies it is a function. Can you either make it a function or rename it `formProps` ? I think the second is better as a function does not bring anything here

packages/react-admin/src/mui/form/SanitizedForm.js
```diff
@@ -0,0 +1,58 @@
...
+        })
+    ) : (
+        <form {...sanitizeFormProps(props)} />
```
djhi: Wouldn't it be better to only have the `render` prop and provide a default for it returning this ?

tkvw:
Moved form modifications to https://github.com/marmelab/admin-on-rest/pull/1389

fzaninotto:

PR #1362, opened 2018-01-09
Example fix

alauper:
this doesn't look like valid code and doesn't use the values parameter sent in, which I think was the point of the example.

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -317,11 +317,11 @@ const number = value => value && isNaN(Number(value)) ? 'Must be a number' : und
 const minValue = min => value =>
   value && value < min ? `Must be at least ${min}` : undefined;
 
-const ageValidation = (value, values) => {
+const ageValidation = (value, allValues) => {
     if (!value) {
         return 'The age is required';
     }
-    if (values.age < 18) {
+    if (allValues.age < 18) {
         return 'Must be over 18';
     }
     return [];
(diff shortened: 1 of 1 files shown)


alauper:
There's actually a few ambiguities with that example.

It's not clear where ageValidation is being called from.  Are the passed in parameters "value, values" or "value, allValues"?  And do you return undefined or []?

docs/CreateEdit.md
```diff
@@ -321,7 +321,7 @@ const ageValidation = (value, values) => {
     if (!value) {
         return 'The age is required';
     }
-    if (age < 18) {
+    if (values.age < 18) {
```
djhi: Agreed, this example is wrong. About your question, the parameters are `(value, allValues)`. So here, `age` should be `value`.

docs/CreateEdit.md
```diff
@@ -317,11 +317,11 @@ const number = value => value && isNaN(Number(value)) ? 'Must be a number' : und
 const minValue = min => value =>
   value && value < min ? `Must be at least ${min}` : undefined;
 
-const ageValidation = (value, values) => {
+const ageValidation = (value, allValues) => {
     if (!value) {
         return 'The age is required';
     }
-    if (age < 18) {
+    if (allValues.age < 18) {
```
djhi: I meant `allValues.age` should simply be `value`.

alauper: I actually thought allValues.age would demonstrate how to use the allValues parameter.  But I agree with your opinion too.  I updated the commit.  Thanks!

PR #1366, opened 2018-01-11
Prop name is validate not validation

alauper:
The docs seem to use "validation" and "validate" as the prop name.  I believe the prop is always "validate", is that correct?  I think the help docs section title "Validation" is fine, but the prop definition list and examples should use validate.


fzaninotto:
Absolutely, thanks for fixing it!

PR #1368, opened 2018-01-12
[RFR] Bootstrap new examples

djhi:
We want to borrow the examples concept from [next.js](https://github.com/zeit/next.js).

Inside the `examples` folder, we will add very focused examples apps to demonstrate various recipes.


fzaninotto:
please remove the `static` folder from the commited files, and rebase to avoid a too large commit.

PR #1369, opened 2018-01-12
[RFR] Change dependencies to peerDependencies

tkvw:
The plugins shouldn't dependent on react-admin but rather have a peerDependency on it. For testing it should be added to devDependencies.

### DIFF
--- a/packages/ra-data-fakerest/package.json
+++ b/packages/ra-data-fakerest/package.json
@@ -31,7 +31,9 @@
     },
     "homepage": "https://github.com/marmelab/react-admin#readme",
     "dependencies": {
-        "fakerest": "~2.1.0",
+        "fakerest": "~2.1.0"
+    },
+    "peerDependencies": {
         "react-admin": "^2.0.0-beta1"
     }
 }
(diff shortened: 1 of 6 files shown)


fzaninotto:
Why? They can't work without `react-admin` anyway.

tkvw:
If my project includes both `react-admin` and `ra-input-richt-text`, npm will pull in two copies of `react-admin`. 
```
npm_modules
  -> react-admin
  -> ra-input-rich-text
     -> npm_modules
          -> react-admin
```
I saw this when I analyzed my constructed production build. It included two copies of react-admin. I did some reading on the npm dependency model, @see: https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/ (I come from maven, so a big difference in dependency handling), and my take on it is that the plugins have a public contract/interface with react-admin, so no direct dependency is required. 

This actually makes sense, because when I want to upgrade `react-admin`, the plugins don't care, because they have there own dependency, but npm should actually say something like: `the peerdepencies of plugin xx is incompatible with this react-admin version.`

fzaninotto:
The blog post you linked to doesn't explain why peer dependencies should be used when the package has a public contract with a dependency, so I looked for another source of truth. I found this:

https://nodejs.org/en/blog/npm/peer-dependencies/

The npm people designed peer dependencies specifically for plugins - and what our `ra-` packages are. 

So you're right: `ra-` packages should declare `react-admin` as `peerDependency`, and if they have tests depending on the package, add it in `devDependencies, too.

PR #1372, opened 2018-01-15
[RFR] Fix Makefile for running the simple example

djhi:


djhi:
Only for prettier (not lint). I fixed it

PR #1374, opened 2018-01-15
[RFR]: Add validation to filterform

tkvw:
FilterForm should respect field validation, this PR fixes that (test included). 
Fixes https://github.com/marmelab/admin-on-rest/issues/1207

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -255,10 +255,15 @@ It does so by inspecting its `context` prop.
 
 **Tip**: Don't mix up this `filters` prop, expecting a React element, with the `filter` props, which expects an object to define permanent filters (see below).
 
+#### Styling
 The `Filter` component accepts the usual `className` prop but you can override many class names injected to the inner components by React-admin thanks to the `classes` property (as most Material UI components, see their [documentation about it](https://material-ui-next.com/customization/overrides/#overriding-with-classes)). This property accepts the following keys:
 
 * `form`: applied to the root element when rendering as a form.
 * `button`: applied to the root element when rendering as a button.
+
+#### Validation
+The `Filter` component accepts a `validate` prop to validate the filter values before a list is filtered. Field level validations are supported as well. 
+For more information on validation, see: [`<SimpleForm>`](./CreateEdit.html#validation) 
 
 ### Records Per Page
 
--- a/e2e/pages/ListPage.js
+++ b/e2e/pages/ListPage.js
@@ -1,4 +1,4 @@
-import { By, until } from 'selenium-webdriver';
+import { By, until, Key } from 'selenium-webdriver';
 
 export default url => driver => ({
     elements: {
@@ -9,6 +9,8 @@ export default url => driver => ({
         filter: name => By.css(`.filter-field[data-source='${name}'] input`),
         filterMenuItems: By.css(`.new-filter-item`),
         menuItems: By.css(`[role=menuitem`),
+        filterError: source =>
+            By.css(`.filter-field[data-source="${source}"] p`),
         filterMenuItem: source =>
             By.css(`.new-filter-item[data-key="${source}"]`),
         hideFilterButton: source =>
@@ -161,7 +163,12 @@ export default url => driver => ({
         if (clearPreviousValue) {
             filterField.clear();
         }
-        filterField.sendKeys(value);
+        if (clearPreviousValue && '' === value) {
+            // Ensure onChange is called
+            filterField.sendKeys(value, Key.TAB);
+        } else {
+            filterField.sendKeys(value);
+        }
         driver.sleep(500);
         return this.waitUntilDataLoaded();
     },
@@ -183,6 +190,10 @@ export default url => driver => ({
         );
         hideFilterButton.click();
         return this.waitUntilDataLoaded(); // wait for debounce and reload
+    },
+
+    getFilterError(name) {
+        return driver.findElement(this.elements.filterError(name)).getText();
     },
 
     logout() {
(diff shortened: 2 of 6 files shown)


packages/react-admin/src/mui/list/FilterForm.js
```diff
@@ -47,11 +47,16 @@ const sanitizeRestProps = ({
     displayedFilters,
     setFilter,
     setFilters,
+    debounce,
```
fzaninotto: not used

packages/react-admin/src/mui/list/Filter.js
```diff
@@ -69,7 +71,6 @@ export class Filter extends Component {
         const {
             classes = {},
             context,
-            debounce,
```
fzaninotto: why would you want it to be in the `rest` parameters and pass it to the `<FilterForm>` if it doesn't use it?

packages/react-admin/src/mui/list/Filter.js
```diff
@@ -23,20 +24,21 @@ export class Filter extends Component {
...
-            this.filters = filtersWithoutEmpty;
+            const filtersWithoutEmpty = removeEmpty(values);
+            if (!isEqual(values, this.filters)) {
+                this.props.setFilters(filtersWithoutEmpty);
+                this.filters = filtersWithoutEmpty;
+            }
         }
-    }, this.props.debounce);
+    };
+    handleFilterFormOnChange = debounce(
```
fzaninotto: this is the responsibility of the `FilterForm` component IMO

packages/react-admin/src/mui/list/Filter.js
```diff
@@ -23,20 +24,21 @@ export class Filter extends Component {
...
 
-    componentWillUnmount() {
-        if (this.props.setFilters) {
-            this.setFilters.cancel();
-        }
-    }
-
-    setFilters = debounce(filters => {
-        if (!isEqual(filters, this.filters)) {
+    handleFilterFormSubmit = (values, dispatch, props) => {
```
fzaninotto: no, the `FilterButton` should pass a `setFilters` function to the form. It's the form's responsibility to call this function both on change and on submit.

packages/react-admin/src/mui/list/Filter.js
```diff
@@ -23,20 +24,21 @@ export class Filter extends Component {
         this.filters = nextProps.filterValues;
     }
 
-    componentWillUnmount() {
-        if (this.props.setFilters) {
-            this.setFilters.cancel();
```
fzaninotto: this shouldn't be removed

packages/react-admin/src/mui/list/FilterForm.js
```diff
@@ -88,60 +93,69 @@ export class FilterForm extends Component {
...
-                                <div>
-                                    <Field
-                                        allowEmpty
-                                        {...filterElement.props}
-                                        name={filterElement.props.source}
-                                        component={filterElement.type}
-                                        resource={resource}
-                                        record={emptyRecord}
-                                    />
+            <form onSubmit={handleSubmit}>
```
fzaninotto: what's the interest of handling submission? Validation?

tkvw: Yes, I've tried to use: 
```
reduxForm({
  onChange: (values,dispatch,props)=>{
     props && props.valid && props.setFilters(values)
 } 
})
```
but `valid` is not updated/evaluated ontime @see: https://github.com/erikras/redux-form/issues/3754 

And either way, only `submit` allows the gui to nicely show the validation errors.

packages/react-admin/src/mui/list/FilterForm.js
```diff
@@ -172,15 +192,40 @@ export const mergeInitialValuesWithDefaultValues = ({
...
+const combineWithProps = props => ({
+    ...mergeInitialValuesWithDefaultValues(props),
+    ...filterFormProps(props),
+});
+
 const enhance = compose(
     withStyles(styles),
     translate,
-    withProps(mergeInitialValuesWithDefaultValues),
+    withProps(combineWithProps),
```
fzaninotto: I don't understand why you add the `onChange` and `cancelOnChangeDebounce` methods as props rather than just methods of the component class. It obfuscates the code IMO.

tkvw: onChange is a prop of the `reduxForm` hoc, so either the form should be moved in it's own component (something like `FilterInnerForm`) decorated by `reduxForm` and pass the `onChange` prop to it using component method class. Or this solution needs to be taken. Simply putting the debounce on the `reduxForm` decorator, like:
```
reduxForm({
   onChange: debounce(...
)}
```
does not work, because this would hide the `cancel` function of debounce. 
So I opted for this solution, but have no problem with externalizing the component. Let me know.

afilp:
@tkvw Thank you for your contribution! As we need this soon, is it possible that you make the changes so that the merge is done? Thanks a lot!

packages/react-admin/src/mui/list/Filter.js
tkvw: it should, because setFilters is not debounced anymore, it's moved down

packages/react-admin/src/mui/list/Filter.js
tkvw: not sure what you're asking, but debounce is the number of ms to debounce the filter function. This is only used by form and not by button.

tkvw:
@afilp I don't think there are any requested changes left to be handled

afilp:
@tkvw I still see this PR open, with requested changes from @fzaninotto. Am I missing something here? (I refer to your mention that no requested changes are left to be handled). Thanks again!

fzaninotto:
Needs rebase

fzaninotto:

PR #1376, opened 2018-01-15
[RFR] Upgrade material-ui-chip-input

### LINKED ISSUE #1139: react16 SelectArrayInput autocomplete not working

**What you were expecting:**

I hope SelectArrayInput can select autocomplete value.
**What happened instead:**

in react 16.0.0 ,SelectArrayInput autocomplete not working
**Steps to reproduce:**

**Related code:**

```
            <SelectArrayInput choices={[
                {
                    id:1,
                    name:'xxx'
                },
                {
                    id:2,
                    name:'yyy'
                }
            ]} optionText="name"/>
```

**Other information:**

In react15.4,SelectArrayInput autocomplete it's ok 

**Environment**

* Admin-on-rest version:^1.3.2
* Last version that did not exhibit the issue (if applicable):15.4
* React version:16.0
* Browser:chrome

### LINKED ISSUE #1255: SelectArrayInput does not work with react >= 16
Values does not appear inside the component after selecting them in the dropdown menu, they are expected to appear.

Can be reproduced by running admin-on-rest-demo with this dependencies:

        "react": "^16.2.0",
        "react-dom": "^16.2.0",
        "react-router-dom": "^4.2.2",
        "react-tap-event-plugin": "^3.0.2",

May be it is because admin-on-rest depends on material-ui-chip-input@~0.17.2, which does not support react 16, it's latest version is 0.18.3.

It seems that the onBlur event handler receives an empty array as the 'event' argument when called by ChipInput.
Redux state log shows that the selected value is added to the array after CHANGE event but removed after BLUR event.

**Environment**
Admin-on-rest version 1.3.2
React 16 or 16.2
Google Chrome 58.0.
Ubuntu 16.04

djhi:
Fixes #1139
Fixes #1255

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,20 @@
 {
     "name": "admin-on-rest",
     "version": "1.3.3",
-    "description":
-        "A frontend Framework for building admin applications on top of REST services, using ES6, React and Material UI",
-    "files": ["*.md", "docs/*.md", "lib", "src"],
+    "description": "A frontend Framework for building admin applications on top of REST services, using ES6, React and Material UI",
+    "files": [
+        "*.md",
+        "docs/*.md",
+        "lib",
+        "src"
+    ],
     "main": "lib/index",
     "scripts": {
         "test": "make test"
     },
-    "authors": ["François Zaninotto"],
+    "authors": [
+        "François Zaninotto"
+    ],
     "repository": "marmelab/admin-on-rest",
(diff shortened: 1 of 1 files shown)


ronadi:
Hi @djhi  Is it possible to add **onBeforeRequest** function on SelectArrayInput to handle before add chip?

PR #1382, opened 2018-01-16
[RFR] Passes all React Router props to custom routes

### LINKED ISSUE #1273: Customroutes router props
Why is only the `location` router prop passed here? I think this leads to unexpected behaviour. I expected to have access to all three router props (`history` and `match` as well)
https://github.com/marmelab/admin-on-rest/blob/9f4138a75ce23ffe17075abb50d5db4cb025d4e4/src/Admin.js#L88

djhi:
Fixes #1273

### DIFF
--- a/packages/react-admin/src/Admin.js
+++ b/packages/react-admin/src/Admin.js
@@ -72,9 +72,9 @@ const Admin = ({
                             <Route
                                 exact
                                 path="/login"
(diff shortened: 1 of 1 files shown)


djhi:
@fzaninotto Should we backport this to `master` ?

fzaninotto:
No, I don't think so.

PR #1384, opened 2018-01-16
dynamic redirects

fu2re:
redirect prop in forms or save button can be a function now.
I have some problems during tests before I have started my work:
 FAIL  packages/react-admin/src/mui/input/ImageInput.spec.js
  ● Test suite failed to run

    Cannot find module './Mime' from 'index.js'

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:194:17)
      at Object.<anonymous> (node_modules/mime/index.js:3:12)

Also I have figure out what sanitizeRestProps is cut off redirect property from form. Any passed value is became "list". SaveButton property work as expected. Is there any hidden new feature? Please fix it before 2.0 is released.

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -444,6 +444,22 @@ export const PostEdit = (props) => (
 );
 ```
 
+In case when this is not enough you can also use a function. For example if you want set the recently added object as value of reference field in the another object:
+
+```jsx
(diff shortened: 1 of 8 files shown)


fzaninotto:
Needs rebase

fzaninotto:
Also, this might be achieved otherwise using a custom saga

PR #1390, opened 2018-01-17
[RFR] Update documentation

djhi:
I also updated some of the code examples to be in sync with the screenshots and removed some deprecated ones.


docs/CreateEdit.md
```diff
@@ -482,52 +482,6 @@ export const PostEdit = (props) => (
 
 **Tip**: Don't forget to also set the `redirect` prop of the Form component to handle submission by the `ENTER` key.
 
-## Declaring Inputs At Runtime
-
-You might want to dynamically define the inputs when the `<Create>` or `<Edit>` components are rendered. They both accepts a function as their child and this function can return a Promise. If you also defined an `authClient` on the `<Admin>` component, the function will receive the result of a call to `authClient` with the `AUTH_GET_PERMISSIONS` type (you can read more about this in the [Authorization](./Authorization.md) chapter).
```
fzaninotto: why is it removed?

docs/Fields.md
```diff
@@ -28,9 +28,11 @@ All field components accept the following attributes:
 * `source`: Property name of your entity to view/edit. This attribute is required.
 * `label`: Used as a table header of an input label. Defaults to the `source` when omitted.
 * `sortable`: Should the list be sortable using `source` attribute? Defaults to `true`.
-* `elStyle`: A style object to customize the look and feel of the field element itself
-* `style`: A style object to customize the look and feel of the field container (e.g. the `<td>` in a datagrid).
+* `className`: A class's name (usually generated by JSS) to customize the look and feel of the field element itself
```
fzaninotto: s/A class's name/A class name/

docs/Fields.md
```diff
@@ -28,9 +28,11 @@ All field components accept the following attributes:
...
+* `cellClassName`: A class's name (usually generated by JSS) to customize the look and feel of the field container (e.g. the `<td>` in a datagrid).
```
fzaninotto: s/A class's name/A class name/

docs/Fields.md
```diff
@@ -28,9 +28,11 @@ All field components accept the following attributes:
...
+* `headerClassName`: A class's name (usually generated by JSS) to customize the look and feel of the field header (e.g. the `<th>` in a datagrid).
```
fzaninotto: s/A class's name/A class name/

docs/Fields.md
```diff
@@ -669,54 +677,106 @@ import { UrlField } from 'react-admin';
 
 ## Styling Fields
 
-All field components accept the `style` prop, which overrides the default style of the field *container*:
+All field components accept a `className` prop, allowing you to customize their style to your liking. We advise you to use Material UI styling solution, JSS to generate those classes. See their [documentation](https://material-ui-next.com/customization/css-in-js/#api) about that.
```
fzaninotto: s/to use Material UI styling solution, JSS to generate those classes/to use the Material UI styling solution, JSS, to generate those classes/

docs/Fields.md
```diff
@@ -669,54 +677,106 @@ import { UrlField } from 'react-admin';
...
 // renders in the datagrid as
-<td style="color: purple;"><span>2</span></td>
+<td><span class="[class name generated by JSS]">2</span></td>
 ```
 {% endraw %}
 
-If you want to override the styles of the field *element*, use the `elStyle` prop instead:
+React-admin usually delegates the rendering of fields components to material-ui components. Refer to the material-ui documentation to see the default styles for elements.
+
+You may want to customize the cell's style inside a `DataGrid`. You can use the `cellClassName` for that:
```
fzaninotto: s/the cell's style/the cell style/

docs/Fields.md
```diff
@@ -669,54 +677,106 @@ import { UrlField } from 'react-admin';
...
 );
-// renders in the datagrid as
-<td style="text-align:right"><span>2</span></td>
 // renders in the table header as
-<th style="text-align:right;font-weight:bold"><button>Price</button></td>
+<th class="[class name generated by JSS]"><button>Price</button></td>
 ```
 {% endraw %}
 
+Finally, sometimes, you just want to right align the text of a cell. Use the `textAlign` prop which accepts either `left` or `right`:
```
fzaninotto: s/the `textAlign` prop which accepts/the `textAlign` prop, which accepts/

docs/Inputs.md
```diff
@@ -168,22 +168,14 @@ This input does not handle `null` values. You would need the `<NullableBooleanIn
 
 You can use the `options` prop to pass any option supported by the Material UI `Toggle` components.
 
-{% raw %}
```
fzaninotto: Why did you remove this one? If it's not good, you should replace it with another one.

docs/List.md
```diff
@@ -519,46 +523,3 @@ export const CommentList = (props) => (
 
 As you can see, nothing prevents you from using `<Field>` components inside your own components... provided you inject the current `record`. Also, notice that components building links require the `basePath` component, which is also injected.
 
-## Declaring Fields At Runtime
-
-You might want to dynamically define the fields when the `<List>` component is rendered. It accepts a function as its child and this function can return a Promise. If you also defined an `authClient` on the `<Admin>` component, the function will receive the result of a call to `authClient` with the `AUTH_GET_PERMISSIONS` type (you can read more about this in the [Authorization](./Authorization.md) chapter).
```
fzaninotto: again, I don't understand why you removed this part.

docs/Show.md
```diff
@@ -157,46 +157,3 @@ export const PostShow = (props) => (
     </Show>
 );
 ```
-
-## Declaring Fields at Runtime
-
-You might want to dynamically define the fields when the `<Show>` component is rendered. It accepts a function as its child, and this function can return a Promise. If you defined an `authClient` on the `<Admin>` component, the function child of the `<Show>` component will receive the result of a call to `authClient` with the `AUTH_GET_PERMISSIONS` type (you can read more about this in the [Authorization](./Authorization.md) chapter).
```
fzaninotto: Why did you remove it?

docs/Tutorial.md
```diff
@@ -545,6 +545,8 @@ This works exactly the way you expect. The lesson here is that react-admin takes
 
 ![Responsive List](./img/responsive-list.gif)
 
+**Note:** We switched to a custom API for those screenshots in order to demonstrate how to use some of the `SimpleList` component props.
```
fzaninotto: Why this comment? is it because the screenshot doesn't reflect the text? If so, is it worth mentioning?

docs/CreateEdit.md
djhi: Because we don't support child as a function anymore

docs/Inputs.md
djhi: Because the `labelPosition` prop does not exist anymore and we have yet to decide how to handle it, if possible

docs/Show.md
djhi: Same reason, child as a function is not supported anymore

docs/List.md
djhi: Same reason, child as a function is not supported anymore

docs/Tutorial.md
djhi: The screenshot should reflect the text but that's not the reason. The `jsonplaceholder` does not provide the fields we want to display on mobile to demonstrate the SimpleList props. As the user may be trying to reproduce our example, it will be frustrating as he/she will see undefined date and views. I thought it was nice to at least explain why he/she can't achieve the same exact result.

docs/Inputs.md
djhi: Oh and by default, it is display at the right of the Switch

PR #1392, opened 2018-01-17
[RFR] AdminConfig component

tkvw:
I think the `AdminConfig` component better separates concerns in the react-admin app, it also makes it easier to implement `custom app` applications. Simply including `AdminConfig` and create the layout.

### DIFF
--- a/packages/react-admin/src/Admin.js
+++ b/packages/react-admin/src/Admin.js
@@ -40,59 +40,61 @@ const Admin = ({
             dataProvider={dataProvider}
             initialState={initialState}
         >
-            <Switch>
-                <Route
-                    exact
-                    path="/login"
-                    render={({ location }) =>
-                        createElement(loginPage || Login, {
-                            location,
-                            title,
-                        })}
-                />
-                {customRoutes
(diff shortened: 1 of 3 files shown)


fzaninotto:
I'm -1 for this change, for several reasons:

1. When developers write a custom app, they want entire control over the reducers, the sagas, the layout, etc. Otherwise they have the `<Admin>` tag. This new component obfuscates the things they can override. 
2. If they just want to customize the layout, they already have an option for that, without implementing a custom app
3. We'd have to document the new component *in addition to* the current way to build the current app (because of 1.) That would be 3 ways of defining an admin... too much choice for the developer.

What do other people think about it?

tkvw:
Thanks for your response. I understand your arguments, but I still think there's alot going on in the `Admin` component. I stubbed this out in my own project and thought it was nice enough to PR. It's mostly cosmetic, the only big change is that it allows a `customMiddleware` to be defined (which is the only thing I'm missing on the `Admin` component. Anyway, I'll keep it open for a couple of more days and if there's no interest, I'll close the issue.

tkvw:
Closing due to lack of interest.

PR #1399, opened 2018-01-18
[RFR] Add demo and tutorial example apps

djhi:
It required to downgrade our version of jest to the version used by `react-scripts` so that the example apps, created with CRA, do not mess up with our tests


djhi:
Ok, but adding a make target for every example app we or our contributors will add in the future is going to be a PITA and we might end up with a very long makefile.

PR #1405, opened 2018-01-22
[RFR] Examples - demo with graphql

djhi:
- [x] Backported updates from `aor-graphql` repository
- [x] Import the graphql demo
- [x] Introduce `ra-data-graphql-simple`
- [x] Convert the demo to use `json-graphql-server`
- [x] Use the client fake server


djhi:
Note about the `ra-data-graphql-simple` package. I started it by simply copying `ra-data-graphcool`. It'll need more work to be really usable but I this PR is getting really large already. I'll work on it on another PR

fzaninotto:
Why did you need to override the `getOneCommandQuery` in `aorApolloClient`?

PR #1406, opened 2018-01-22
[RFR] Fix crud-fetch cancelPrevious

tkvw:
I have a custom component (LocaleSelector) which uses `crudFetch` to retrieve data. But when the component is rendered on the same page as a List, the `crudFetch` of List will cancel the previous fetch of the LocaleSelector component. I think the cancellation should only happen if the resources are the same and `cancelPrevious` is requested. 
This PR add's this behaviour.


packages/react-admin/src/sideEffect/saga/crudFetch.js
```diff
@@ -18,13 +19,12 @@ const crudFetch = dataProvider => {
...
-            put({ type: FETCH_START }),
-        ]);
-        let response;
         try {
-            response = yield call(
+            yield all([
+                put({ type: `${type}_LOADING`, payload, meta }),
+                put({ type: FETCH_START }),
+            ]);
+            let response = yield call(
```
djhi: Why those changes ? They don't seem related to this PR

tkvw: They are, because the generator can be `yielding`  on the puts, while `cancel` is called from the `watchCrudFetch` generator, if this happens the `cancel` finalizer will never run, and the loading state isn't reduced correctly.

afilp:
Can this be merged? Thanks!

fzaninotto:
Can you please add unit tests?

tkvw:
Can this be merged? Or do you require additional tests?

fzaninotto:
needs rebase

tkvw:
Rebased

PR #1409, opened 2018-01-23
[RFR] change extraction of status for AUTH_ERROR

### LINKED ISSUE #1408: Authentication Documentation AUTH_ERROR status code correction
**What you were expecting:**
as also noted in https://github.com/marmelab/admin-on-rest/issues/1095
I had to use the Error internals to extract the status as it is not directly on params (as stated in the docs: https://github.com/marmelab/admin-on-rest/blob/master/docs/Authentication.md#catching-authentication-errors-on-the-api)
```js
if (type === AUTH_ERROR) {
        const status  = params.message.status;
        if (status === 401 || status === 403) {
             localStorage.removeItem('token');
            return Promise.reject({message: 'Authentication Error: Token Expired or Forbidden'});
        }
        return Promise.resolve();
		
    }
```
**What happened instead:**
In documentation it states: 
```js
if (type === AUTH_ERROR) {
        const { status } = params;
        if (status === 401 || status === 403) {
            localStorage.removeItem('token');
            return Promise.reject();
        }
        return Promise.resolve();
    }
```

zifnab87:
Fixes #1408

### DIFF
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -138,7 +138,7 @@ export default (type, params) => {
         // ...
     }
     if (type === AUTH_ERROR) {
-        const { status } = params;
+        const status  = params.message.status;
         if (status === 401 || status === 403) {
             localStorage.removeItem('token');
             return Promise.reject();


djhi:
Thanks!

PR #1410, opened 2018-01-23
[RFR] AutocompleteInput issues

tkvw:
There are a couple of issues with `AutocompleteInput`: 

- [X] When the suggestion is undefined it errors https://github.com/moroshko/react-autosuggest/pull/481
- [X] After the selectionlist appears and Escape is pressed, the selections do not reappear. 
- [X] `getSuggestionValue` does not check if `optionText` could be a function. 
- [X] Added a bunch of tests

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -136,6 +136,8 @@ By default the component matches choices with the current input searchText, if i
 }/>
 ```
 
+If you want to limit the initial choices shown to the current value only, you can set the `limitChoicesToValue` prop.  
+
 Lastly, use the `options` attribute if you want to override any of Material UI's `<AutoComplete>` attributes:
 
 {% raw %}
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -68,7 +68,7 @@ Then you can display a text input to edit the author first name as follows:
 
 ## `<AutocompleteInput>`
 
-To let users choose a value in a list using a dropdown with autocompletion, use `<AutocompleteInput>`. It renders using [Material ui's `<AutoComplete>` component](http://www.material-ui.com/#/components/auto-complete) and a `fuzzySearch` filter. Set the `choices` attribute to determine the options list (with `id`, `name` tuples).
+To let users choose a value in a list using a dropdown with autocompletion, use `<AutocompleteInput>`. It renders using [react-autosuggest](http://react-autosuggest.js.org/) and a `fuzzySearch` filter. Set the `choices` attribute to determine the options list (with `id`, `name` tuples).
 
 ```jsx
 import { AutocompleteInput } from 'react-admin';
@@ -125,6 +125,16 @@ However, in some cases (e.g. inside a `<ReferenceInput>`), you may not want the 
 <AutocompleteInput source="gender" choices={choices} translateChoice={false}/>
 ```
 
+
+By default the component matches choices with the current input searchText, if it finds a match this choice will be selected. For example, using the `choices`: `[{id:'M',name:'Male',id:'F',name:'Female'}]` and the user enters the text `male` then the component will set the input value to `M`. Using the `inputValueMatcher` prop the component allows you to change how choices are matched. For example, given the choices: `[{id:1,iso2:'NL',name:'Dutch'},{id:2,iso2:'EN',name:'English'},{id:3,iso2:'FR',name:'French'}]` you can create the following `inputValueMatcher` to match choices on the iso2 code: 
+```javascript
+<AutocompleteInput inputValueMatcher={
+    (input,suggestion,getOptionText) => 
+        input.toUpperCase().trim() === suggestion.iso2 || 
+        input.toLowerCase().trim() === getOptionText(suggestion).toLowerCase().trim()
+}/>
+```
+
 Lastly, use the `options` attribute if you want to override any of Material UI's `<AutoComplete>` attributes:
 
 {% raw %}
@@ -135,8 +145,6 @@ Lastly, use the `options` attribute if you want to override any of Material UI's
 }} />
 ```
 {% endraw %}
-
-Refer to [Material UI Autocomplete documentation](http://www.material-ui.com/#/components/auto-complete) for more details.
 
 **Tip**: If you want to populate the `choices` attribute with a list of related records, you should decorate `<AutocompleteInput>` with [`<ReferenceInput>`](#referenceinput), and leave the `choices` empty:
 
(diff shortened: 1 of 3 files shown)


djhi:
I was wondering whether it would be simpler to use [downshift](https://github.com/paypal/downshift) here. Thoughts ?

tkvw:
Cool, didn't know that one. I think this is a good idea, the `react-autosuggest` library seems to be inactive at the moment. I'll take a whack at it.

djhi:
Can we keep the downshift update for another PR ?

tkvw:
Yes I will, what about the current PR, do I need to separate the `ReferenceInput` and `ReferenceArrayInput` from the `Autosuggest` component?

djhi:
If those changes aren't directly related to the `Autosuggest` bugs, please do so

tkvw:
Refactored and separated `ReferenceInput` and `ReferenceArrayInput` to separate issue.

fzaninotto:
Could you add tests to demonstrate the fixes?

tkvw:
I feel like closing this issue in favor of the Downshift solution, what do you think?

djhi:
No, let's keep `react-suggest` for now. We need some time to decide whether downshift actually make the code more readable and less prone to errors

djhi:
I guess this PR now depends on #1420 ?

tkvw:
> I guess this PR now depends on #1420 ?

Nope, but it's wise to merge them together. `selectedItem` isn't a required prop. Just it's primary source to reduce the `selectedItem`.

docs/Inputs.md
```diff
@@ -152,6 +160,20 @@ import { AutocompleteInput, ReferenceInput } from 'react-admin'
...
+`inputValueMatcher` | Optional | `Function` | `(input, suggestion, getOptionText) => input.toLowerCase().trim() === getOptionText(suggestion).toLowerCase().trim()` | Allows to define how choices are matched with the searchText while typing.   
+`optionValue` | Optional | `string` | `id` | Fieldname of record containing the value to use as input value 
+`optionText` | Optional | <code>string &#124; Function</code> | `name` | Fieldname of record to display in the suggestion item or function which accepts the currect record as argument (`(record)=> {string}`)
+`setFilter` | Optional | `Function` | null | A callback to inform the `searchText` has changed and new `choices` can be retrieved based on this `searchText`. Signature `searchText => void`. This function is automatically setup when using `ReferenceInput`. 
+`selectedItem` | Optional | `Object` | null | The current selectedItem. If this value is not supplied, it will be reduced from the `choices` list.
```
djhi: > If this value is not supplied, it will be reduced from the `choices` list.

I don't get this

tkvw: The component expects a selectedItem object to render the `optionText` from. If the selectedItem is not supplied it will look it from the `choices` array (matching the `optionValue` of each record with the `input.value`). The latter was the default behaviour on the previous `AutocompleteInput`. Now it simply allows you supply the `selectedItem` directly (which is what the updated `ReferenceInput` does now).

djhi: I suggest you put this explanation in the documentation :)

docs/Inputs.md
```diff
@@ -125,6 +125,16 @@ However, in some cases (e.g. inside a `<ReferenceInput>`), you may not want the
 <AutocompleteInput source="gender" choices={choices} translateChoice={false}/>
 ```
 
+
+By default the component matches choices with the current input searchText, if it finds a match this choice will be selected. For example, using the `choices`: `[{id:'M',name:'Male',id:'F',name:'Female'}]` and the user enters the text `male` then the component will set the input value to `M`. Using the `inputValueMatcher` prop the component allows you to change how choices are matched. For example, given the choices: `[{id:1,iso2:'NL',name:'Dutch'},{id:2,iso2:'EN',name:'English'},{id:3,iso2:'FR',name:'French'}]` you can create the following `inputValueMatcher` to match choices on the iso2 code: 
+```javascript
```
djhi: Can you add an empty line before this one ? Otherwise, the markup for code won't be displayed

docs/Inputs.md
djhi: Can you remove this prop ?

packages/react-admin/src/mui/input/AutocompleteInput.js
```diff
@@ -91,65 +92,70 @@ const styles = theme => ({
  */
 export class AutocompleteInput extends React.Component {
     state = {
-        firstLoad: true,
+        dirty: false,
+        inputValue: null,
         searchText: '',
+        selectedItem: null,
```
djhi: Please remove

tkvw: Euh no? This is a state variable, not a prop. It caches the selectedItem.

docs/Inputs.md
tkvw: done

djhi:
I just tried those changes on the simple example locally and it seems that when editing a resource, the component now just show the selected one in its choices. You now have to explicitly erase the content and type a new search to see other choices. We added this feature deliberately to at least show some choices along side the selected one (optimistic UI). Can you add this behavior back ?

tkvw:
> I just tried those changes on the simple example locally and it seems that when editing a resource, the component now just show the selected one in its choices. You now have to explicitly erase the content and type a new search to see other choices. We added this feature deliberately to at least show some choices along side the selected one (optimistic UI). Can you add this behavior back ?

Wow, I actually spend time to get rid of this behaviour. Can we agree to make it configurable (I can default it's behaviour to what you want). I think this is counter intuitive, and in the case the optionText renders an item which has additonalinfo (like a persons badge or something), it should only show the selecteditem item imo.

docs/Inputs.md
```diff
@@ -152,6 +163,19 @@ import { AutocompleteInput, ReferenceInput } from 'react-admin'
 
 **Tip**: React-admin's `<AutocompleteInput>` has only a capital A, while material-ui's `<AutoComplete>` has a capital A and a capital C. Don't mix up the components!
 
+### Properties
+Prop | Required/Optional | Type | Default | Description
+---|---|---|---|---
+`choices` | Required | `Object[]` | - | List of items to autosuggest
+`resource` | Required | `string` | - | The resource working on. This field is passed down by wrapped components like `Create` and `Edit`.  
+`source` | Required |  `string` | - | Name of field to edit, it's type should correspond to the type retrieved from `optionValue` 
+`allowEmpty` | Optional | `boolean` | `false` | If `false` and the searchText typed did not match any suggestion, the searchText will revert to the current value when the field is blurred. If `true` and the `searchText` is set to `''` then the field will set the input value to `null`.
```
djhi: Can you either remove this line or specify that it will be passed from the `ReferenceInput`. Users should not specify `allowEmpty` directly on the `AutocompleteInput`

djhi: Can you either remove this line or specify that it will be passed from the `ReferenceInput`. Users should not specify `allowEmpty` directly on the `AutocompleteInput`

djhi:
Thanks ! We'll fix the doc ourselves :)

PR #1412, opened 2018-01-24
[RFR] Refactor i18nProvider

djhi:


UPGRADE.md
```diff
@@ -117,20 +117,18 @@ const messages = {
...
-        return messages.en; // The default messages should always resolve synchronously!! 
+const i18nProvider = locale => {
+    if (locale === 'fr') {
+        return frMessages;
     }
-    return messages[params.locale];
+
+    // If locale wasn't specified it's a request for default messages
+    // If locale is not supported, fallback to english
+    return enMessages;
```
fzaninotto: why don't you simply return `messages[locale]` all the time?

docs/Translation.md
```diff
@@ -9,63 +9,61 @@ The react-admin interface uses English as the default language. But it also supp
...
-    }
-    if (type === GET_LOCALE_MESSAGES) {
-        return messages[params.locale];
+const i18nProvider = locale => {
+    if (locale === 'fr') {
+        return frenchMessages;
     }
-    throw new Error('Undefined action type', type); 
+
+    // Always fallback on english
```
fzaninotto: fallback *to* english

docs/Translation.md
```diff
@@ -9,63 +9,61 @@ The react-admin interface uses English as the default language. But it also supp
...
-  fr: frenchMessages,
-};
-const i18nProvider = (type, params) => {
-    if (type === GET_DEFAULT_MESSAGES) {
-        return englishMessages;
-    }
-    if (type === GET_LOCALE_MESSAGES) {
-        return messages[params.locale];
+const i18nProvider = locale => {
+    if (locale === 'fr') {
```
fzaninotto: use a simple example first with object lookup in every case, then speak about the difference between first and subsequent locale changes in another example. Otherwise, this first example looks intimidating.

docs/Translation.md
```diff
@@ -9,63 +9,61 @@ The react-admin interface uses English as the default language. But it also supp
...
+The i18nProvider may return a promise for any subsequent calls, which can be useful to only load the needed locale. For example:
 
 ```javascript
+import englishMessages from '../en.js';
+
 const asyncMessages = {
-  en: ()=> import('../en.js').then(messages => messages.default),
-  fr: ()=> import('../fr.js').then(messages => messages.default),
+    fr: () => import('../i18n/fr.js').then(messages => messages.default),
+    it: () => import('../i18n/it.js').then(messages => messages.default),
```
fzaninotto: fr / it ? why not en / fr ?

UPGRADE.md
djhi: For the fallback ?

docs/Translation.md
djhi: Just wanted to show several locales and we don't need to asynchronously import `en` as it is already there

djhi: But the fallback won't work here anyway...

djhi: Fixed the fallback

PR #1422, opened 2018-01-25
[WIP]: use downshift as container for AutocompleteInput

tkvw:
Arguments: 
* `react-autosuggest` is inactive the last months 
* downshift uses no JSF, only render props, cleaner implementation 
* suggested by @djhi 

- [x] Rewrite component 
- [X] Update tests
- [X] Added behavioural tests 
- [X] Allow customized rendering of suggestion items
- [ ] Allow component to be externalized
- [ ] Update docs


tkvw:
Only the docs should be added, but it's ready for code review.

fzaninotto:
Let's first try to make the current implementation work fine. If we hit a wall, we'll reconsider this.

Thanks for your work though.

PR #1425, opened 2018-01-26
[RFR] Fix Edit view refresh does not cancel changes

### LINKED ISSUE #1417: Refresh button in the Edit view does not refresh data.
**What you were expecting:**
To have original values pulled from API and set in the form fields.

**What happened instead:**
The form field values do not get updated with the original values coming from the API.

**Steps to reproduce:**
Open [any item on the demo app](https://marmelab.com/admin-on-rest-demo/#/customers/121), change some data, hit Refresh button.

**Environment**

* Admin-on-rest version: latest

djhi:
Fixes #1417

### DIFF
--- a/src/mui/detail/Edit.js
+++ b/src/mui/detail/Edit.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { Card, CardText } from 'material-ui/Card';
 import compose from 'recompose/compose';
 import inflection from 'inflection';
+import { reset } from 'redux-form';
 import ViewTitle from '../layout/ViewTitle';
 import Title from '../layout/Title';
 import {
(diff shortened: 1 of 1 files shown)


src/mui/detail/Edit.js
```diff
@@ -30,7 +31,11 @@ export class Edit extends Component {
 
     componentWillReceiveProps(nextProps) {
         if (this.props.data !== nextProps.data) {
-            this.setState({ record: nextProps.data }); // FIXME: erases user entry when fetch response arrives late
+            this.setState({ record: nextProps.data });
+
+            // As a refresh of the current record won't actually change the record values,
+            // we need to explicitly ask redux-form to reset the form
+            this.props.resetForm('record-form');
             if (this.fullRefresh) {
```
fzaninotto: the `fullRefresh` thingie is never updated, I think it's a trace from the past (when we didn't use the store for refreshes). 

We need another way to force a redraw.

PR #1430, opened 2018-01-27
[RFR]: AutocompleteArrayInput and jest snapshot testing

tkvw:
Follow #1410 

Created a rudimentary implementation of `SelectArrayInput`, before going forward I like to know what direction you want to go. This implementation just adds a `startAdornment` to the (fixed) `AutocompleteInput` (that's why the AutocompleteInput changes are in the PR as well). It supports adding while typing, delete using backspace, and a click handler. It has no additional external dependencies. 

- [X] Create component
- [X] Add tests
- [X] Add docs


fzaninotto:
IMO, this is a great start, but in a direction different to what we want. 

We want to get a `SelectArrayInput` fast, and that means without autocomplete. Maybe we'll introduce an `AutocompleteArrayInput` in the future, but the only thing that we want to implement before releasing 2.0, is a simple `SelectArrayInput`. Your PR makes one component more complex, we're looking for many, simpler components.

tkvw:
Fine, by me. It's actually just a composition of `AutocompleteInput` and a startadornment, so not very complicated. Like I said, it's based on the PR https://github.com/marmelab/admin-on-rest/pull/1410 of `Autocomplete` which I think is still broken. So this implementation is actually much smaller than it looks. 

I just wanted to mention; thanks for all the work and effort you and your team put into the project.

fzaninotto:
The simple `<SelectArrayInput>` was implemented in #1495. This PR should therefore be repurposed to create an `<AutocompleteArrayInput>`.

tkvw:
Changed it to `<AutocompleteArrayInput>`

tkvw:
Rebased and added jest snapshot testing: 

Snapshot testing: 
* Gives better understanding of what's rendered
* Makes testing easier
* Snapshots should be versioned and verified
* Only snapshot `shallow`, not `mount`

See: 

* https://facebook.github.io/jest/docs/en/snapshot-testing.html
* https://hackernoon.com/testing-react-components-with-jest-and-enzyme-41d592c174f

fzaninotto:
We're very aware of snapshot testing, and we don't want it. It tends to test too much details, and breaks too often. Please use more atomic unit tests instead.

fzaninotto:

PR #1436, opened 2018-01-29
Update SimpleShowLayout.js

afilp:
Contrary to the DatagridBody, we could not have conditional rendering of fields. This was failing because of a missing check in this file.

### DIFF
--- a/src/mui/detail/SimpleShowLayout.js
+++ b/src/mui/detail/SimpleShowLayout.js
@@ -12,36 +12,38 @@ const SimpleShowLayout = ({
     version,
 }) => (
     <div style={style} key={version}>
-        {Children.map(children, field => 
-        field ? (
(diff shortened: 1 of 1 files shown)


djhi:
Hi @afilp, thanks for taking the time to make this fix! We use eslint and prettier to ensure a consistent code style. Can you make sure you apply prettier on your changes ?

afilp:
@djhi I am not sure if i did this correctly, I used WebStorm to reformat the file based on eslint suggestions. Is this what you are looking for now? (2nd commit)

afilp:
@djhi Will you release a new library on npm soon? We really need this change. Thanks!

djhi:
Unfortunately, we won't be able to release a new version before next week. Thank you for your patience :)

PR #1443, opened 2018-01-29
[RFR] Upgrade the graphql data provider to Apollo 2

djhi:


packages/ra-data-graphql/src/buildApolloClient.js
```diff
@@ -8,9 +9,13 @@ export default options => {
     const { uri, ...otherOptions } = options;
 
     if (uri) {
+        const link = new HttpLink({ uri });
+        const cache = new InMemoryCache().restore({});
+
         return new ApolloClient({
+            link,
+            cache,
```
djhi: They are put before on purpose. User may override them

PR #1450, opened 2018-01-31
[RFR] Fix WithPermissionsFilteredChildren should not put regular children in state

### LINKED ISSUE #1144: WithPermissionsFilteredChildren stores the children as state, making it impossible to control which children of a list are rendered
First I would like to say that I really appreciate the work you have done with this great framework. 

I have an issue with WithPermissionsFilteredChildren. Apparently this component stores the children as state and by that it controls which children are rendered. This makes it is impossible for a component wrapping a e.g List to control the children. 

Although there may be reasons of which I'm not aware, I don't feel that a framework should interfere with [regular expectations of how React children work](http://mxstbr.blog/2017/02/react-children-deepdive/#manipulating-children). Are there any particular reasons for storing the children as state? 

I have made an example of how I would expect it to work in this fiddle: https://jsfiddle.net/activist/L19pgec6/
If you try to do the same (with the respectively named components) in admin-on-rest the second child/paragraph will render even though the wrapping component removes it. 

Any thoughts on this? Would it be possible to rewrite this part of the code? Or if the children must be stored as state, could a callback possibly be provided to update the children state? 

Also see https://stackoverflow.com/questions/46606712/is-there-a-simple-way-of-implementing-a-column-picker-for-a-list for relevant example.

**Environment**

* Admin-on-rest version: 1.3.1
* React version: 15.4.2

### LINKED ISSUE #1257: Edit View not being updated when fields get changed.- Again

djhi:
Fixes #1144
Fixes #1257

### DIFF
--- a/docs/Authorization.md
+++ b/docs/Authorization.md
@@ -124,9 +124,9 @@ export const UserCreate = ({ ...props }) =>
                 toolbar={<UserCreateToolbar permissions={permissions} />}
                 defaultValue={{ role: 'user' }}
             >
-                <TextInput source="name" validate={[required]} />
+                <TextInput source="name" validate={required} />
                 {permissions === 'admin' &&
-                    <TextInput source="role" validate={[required]} />}
+                    <TextInput source="role" validate={required} />}
             </SimpleForm>}
     </Create>;
 ```
--- a/docs/Authorization.md
+++ b/docs/Authorization.md
@@ -167,7 +167,7 @@ const UserFilter = ({ ...props }) =>
                 alwaysOn
             />,
             <TextInput key="name" source="name" />,
-            permissions === 'admin' ? <TextInput source="role" /> : null,
+            permissions === 'admin' ? <TextInput key="role" source="role" /> : null,
         ]}
     </Filter>;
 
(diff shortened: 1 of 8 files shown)


docs/Authorization.md
```diff
@@ -167,7 +167,7 @@ const UserFilter = ({ ...props }) =>
                 alwaysOn
             />,
             <TextInput key="name" source="name" />,
-            permissions === 'admin' ? <TextInput source="role" /> : null,
+            permissions === 'admin' ? <TextInput key="role" source="role" /> : null,
```
alexisjanvier: As `key` param seems to be required, should this not be pointed in the documentation?

djhi: Its standard React :)

PR #1454, opened 2018-02-01
[RFR] Added FAQ entry about unique child key in datagrid

djhi:


docs/FAQ.md
```diff
@@ -35,6 +36,15 @@ const convertHTTPResponseToREST = (response, type, resource, params) => {
 };
 ```
 
+## I get warning about unique key for child in array
```
alexisjanvier: :+1:

Issue #854, opened 2017-07-11
Hide loader in Reference Fields when the REST client call fails


natrim:
Sometimes when i return to resource list, the ReferenceArrayField is not updated and is stuck in loading state.

There is no request seen in network panel in Chrome.
And the only message is console is many of: finalize has been cancelled

I have not managed to reproduce it reliably, so i guess its some timeout thingie.

It would probably be nice if there was some timeout on the reference request?
And displaying some error icon with retry on click?

fzaninotto:
If you can't give us a reliable way to reproduce it, there is no way we can fix it...

Feel free to reopen when you have a reliable failing test case.

natrim:
well the only way i can reproduce it is to simulate network failure during fetching of the references

ie. i set slow network in chrome, and then set it to offline after fetching the list, but before the referenced fields are fetched

fzaninotto:
Just to rephrase: when your server doesn't return the response to a ReferenceArrayField request,  it lets the field in loading state. It's probably true for every Reference field (not only ReferenceArrayField).

There is currently no way the ReferenceArrayField can detect a failure in the REST client call. All it does is update when the referenced data changes. But I can see why it's a problem.

I'm reopening and marking it as an enhancement request.

fzaninotto:
As a side note, this is probably VERY difficult. For instance, some REST clients make n requests to fetch references. If only one of those requests fails, how can the concerned Reference field know about it?

natrim:
well the simplest fix would be to add one more prop named timeout (and set timeout based on it) and if it elapses and the field is still in loading state, then mark it as failed

fzaninotto:
No, handling timeout is your rest client's business. For aor, a timeout isn't different from another REST failure. And aor components don't send REST requests directly anyway, so there is no way to determine if a REST failure is related to a given component.

llioor:
@fzaninotto I think that the issue is "Too Many Attempts" (error code 429).
For example:
I have Country model and I have State model. When I edit the Country "United States" I use `<ReferenceArrayField>`  to load the States (50). In the network console I can see many requests to load all chosen States (50 requests). This can be an issue if you are using "throttling" in your API. If I go back to the list or refresh the page I got "429" error code and the `<ReferenceArrayField>` is stuck in loading status.

djhi:
[...]

PR #1459, opened 2018-02-01
[RFR] Decoupling allowEmpty prop from reference selectors

### LINKED ISSUE #953: Update docs ReferenceField about option allowEmpty
There is no mention about option `allowEmpty` in docs about ReferenceField https://marmelab.com/admin-on-rest/Fields.html#referencefield
We can see progress bar without this option. 
So, if this fix already done or will be done in PR https://github.com/marmelab/admin-on-rest/pull/873 you can close it. Thank you

### LINKED ISSUE #1153: allowEmpty mandatory on Create form using a ReferenceInput
**What you were expecting:**
Using `ReferenceInput` inside a `Create` `Simpleform` to work without having to use the `allowEmpty` prop

**What happened instead:**
The component is disabled

**Related code:**
https://codesandbox.io/s/52o4m4pxo4

**Other information:**
To be able to enter a resource with a mandatory `ReferenceInput`, there is no other choice than to use the allowEmpty prop in the `Create `config. Which is very contradictory, specially together with `validation={{ required: true }}`

Since there is no reference id yet, and allowEmpty is false, this piece of code is executed in `ReferenceInput`

```
            if (!referenceRecord && !allowEmpty) {
                return _react2.default.createElement(_Labeled2.default, {
                    label: typeof label === 'undefined' ? 'resources.' + resource + '.fields.' + source : label,
                    source: source,
                    resource: resource
                });
            }
```
and the dropdown is deactivated

image

Could we change this condition and not do this if we are on a create form maybe ?

**Environment**

* Admin-on-rest version: 1.3.1
* React version: 15.6.1
* Browser: Chrome

alexisjanvier:
Fix #953 
Fix #1153 

`allowEmpty` should only be used to add or not an empty choice in the choices list, with a value of null.
It must not interfere with the fetch of references.

The reference selectors must indicate either that the reference fetch is in progress, that the reference fetch has failed, or display the *choice component* with choices based on the fetched references.

### todo
- [x] Display of a spinner during references fetch
fetchok

- [x] Display fetch error if it exists
fetchko

- [x] Render children component after fetch with allowEmpty prop binded
- [x] Update documentation about `allowEmpty`

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -581,7 +581,7 @@ import { ReferenceInput, SelectInput } from 'admin-on-rest'
 ```
 
 Set the `allowEmpty` prop when you want to add an empty choice with a value of null in the choices list.
-Be careful, disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
+Disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
 
 ```jsx
 import { ReferenceInput, SelectInput } from 'admin-on-rest'
@@ -697,7 +697,7 @@ import { ReferenceArrayInput, SelectArrayInput } from 'admin-on-rest'
 ```
 
 Set the `allowEmpty` prop when you want to add an empty choice with a value of null in the choices list.
-Be careful, disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
+Disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
 
 ```js
 import { ReferenceArrayInput, SelectArrayInput } from 'admin-on-rest'
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -30,7 +30,7 @@ All input components accept the following attributes:
 
 * `source`: Property name of your entity to view/edit. This attribute is required.
 * `defaultValue`: Value to be set when the property is `null` or `undefined`.
-* `validate`: Validation rules for the current property (see the [Validation Documentation](./CreateEdit.md#validation))
+* `validate`: Validation rules for the current property (see the [Validation Documentation](./CreateEdit.html#validation))
 * `label`: Used as a table header of an input label. Defaults to the `source` when omitted.
 * `style`: A style object to customize the look and feel of the field container (e.g. the `<div>` in a form).
 * `elStyle`: A style object to customize the look and feel of the field element itself
@@ -580,7 +580,8 @@ import { ReferenceInput, SelectInput } from 'admin-on-rest'
 </Admin>
 ```
 
-Set the `allowEmpty` prop when the empty value is allowed.
+Set the `allowEmpty` prop when you want to add an empty choice with a value of null in the choices list.
+Be careful, disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
 
 ```jsx
 import { ReferenceInput, SelectInput } from 'admin-on-rest'
@@ -695,7 +696,8 @@ import { ReferenceArrayInput, SelectArrayInput } from 'admin-on-rest'
 </Admin>
 ```
 
-Set the `allowEmpty` prop when the empty value is allowed.
+Set the `allowEmpty` prop when you want to add an empty choice with a value of null in the choices list.
+Be careful, disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
 
 ```js
 import { ReferenceArrayInput, SelectArrayInput } from 'admin-on-rest'
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -254,7 +254,7 @@ export const PostEdit = (props) => (
     <Edit title={<PostTitle />} {...props}>
         <SimpleForm>
             <DisabledInput source="id" />
-            <ReferenceInput label="User" source="userId" reference="users" validate={[required]}>
+            <ReferenceInput label="User" source="userId" reference="users" validate={required}>
                 <SelectInput optionText="name" />
             </ReferenceInput>
             <TextInput source="title" />
@@ -266,7 +266,7 @@ export const PostEdit = (props) => (
 export const PostCreate = (props) => (
     <Create {...props}>
         <SimpleForm>
-            <ReferenceInput label="User" source="userId" reference="users" validate={[required]}>
+            <ReferenceInput label="User" source="userId" reference="users" validate={required}>
                 <SelectInput optionText="name" />
             </ReferenceInput>
             <TextInput source="title" />
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -249,11 +249,12 @@ const PostTitle = ({ record }) => {
     return <span>Post {record ? `"${record.title}"` : ''}</span>;
 };
 
+const required = value => (value ? undefined : 'Required');
 export const PostEdit = (props) => (
     <Edit title={<PostTitle />} {...props}>
         <SimpleForm>
             <DisabledInput source="id" />
-            <ReferenceInput label="User" source="userId" reference="users">
+            <ReferenceInput label="User" source="userId" reference="users" validate={[required]}>
                 <SelectInput optionText="name" />
             </ReferenceInput>
             <TextInput source="title" />
@@ -265,7 +266,7 @@ export const PostEdit = (props) => (
 export const PostCreate = (props) => (
     <Create {...props}>
         <SimpleForm>
-            <ReferenceInput label="User" source="userId" reference="users" allowEmpty>
+            <ReferenceInput label="User" source="userId" reference="users" validate={[required]}>
                 <SelectInput optionText="name" />
             </ReferenceInput>
             <TextInput source="title" />
@@ -281,7 +282,6 @@ If you've understood the `<List>` component, the `<Edit>` and `<Create>` compone
 
 As for the `<ReferenceInput>`, it takes the same props as the `<ReferenceField>` (used earlier in the list page). `<ReferenceInput>` uses these props to fetch the API for possible references related to the current record (in this case, possible `users` for the current `post`). It then passes these possible references to the child component (`<SelectInput>`), which is responsible for displaying them (via their `name` in that case), and letting the user select one. `<SelectInput>` renders as a `<select>` tag in HTML.
 
-**Tip**: The `<Edit>` and the `<Create>` components use the same `<ReferenceInput>` configuration, except for the `allowEmpty` attribute, which is required in `<Create>`.
 
 To use the new `<PostEdit>` and `<PostCreate>` components in the posts resource, just add them as `edit` and `create` attributes in the `<Resource>` component:
 
@@ -341,7 +341,7 @@ import { Filter, ReferenceInput, SelectInput, TextInput } from 'admin-on-rest';
 const PostFilter = (props) => (
     <Filter {...props}>
         <TextInput label="Search" source="q" alwaysOn />
-        <ReferenceInput label="User" source="userId" reference="users" allowEmpty>
+        <ReferenceInput label="User" source="userId" reference="users">
             <SelectInput optionText="name" />
         </ReferenceInput>
     </Filter>
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -229,7 +229,7 @@ An admin interface is about displaying remote data, but also about editing and c
 ```jsx
 // in src/posts.js
 import React from 'react';
-import { List, Edit, Create, Datagrid, ReferenceField, TextField, EditButton, DisabledInput, LongTextInput, ReferenceInput, SelectInput, SimpleForm, TextInput } from 'admin-on-rest';
+import { List, Edit, Create, Datagrid, ReferenceField, TextField, EditButton, DisabledInput, LongTextInput, ReferenceInput, required, SelectInput, SimpleForm, TextInput } from 'admin-on-rest';
 
 export const PostList = (props) => (
     <List {...props}>
@@ -249,7 +249,6 @@ const PostTitle = ({ record }) => {
     return <span>Post {record ? `"${record.title}"` : ''}</span>;
 };
 
-const required = value => (value ? undefined : 'Required');
 export const PostEdit = (props) => (
     <Edit title={<PostTitle />} {...props}>
         <SimpleForm>
--- a/example/comments.js
+++ b/example/comments.js
@@ -162,6 +162,7 @@ export const CommentEdit = props => (
         <SimpleForm>
             <DisabledInput source="id" />
             <ReferenceInput
+                label="Post"
                 source="post_id"
                 reference="posts"
                 perPage={5}
@@ -177,19 +178,20 @@ export const CommentEdit = props => (
 );
 
 const defaultValue = { created_at: new Date() };
+const required = value => (value ? undefined : 'Required');
 export const CommentCreate = props => (
     <Create {...props}>
         <SimpleForm defaultValue={defaultValue}>
             <ReferenceInput
+                label="Post"
                 source="post_id"
                 reference="posts"
-                allowEmpty
-                validation={{ required: true }}
+                validate={[required]}
             >
                 <SelectInput optionText="title" />
             </ReferenceInput>
             <DateInput source="created_at" />
-            <LongTextInput source="body" />
+            <LongTextInput source="body" validate={[required]} />
         </SimpleForm>
     </Create>
 );
--- a/example/comments.js
+++ b/example/comments.js
@@ -19,6 +19,7 @@ import {
     TextField,
     TextInput,
     minLength,
+    required,
     translate,
     Show,
     ShowButton,
@@ -178,7 +179,6 @@ export const CommentEdit = props => (
 );
 
 const defaultValue = { created_at: new Date() };
-const required = value => (value ? undefined : 'Required');
 export const CommentCreate = props => (
     <Create {...props}>
         <SimpleForm defaultValue={defaultValue}>
--- a/example/comments.js
+++ b/example/comments.js
@@ -186,12 +186,12 @@ export const CommentCreate = props => (
                 label="Post"
                 source="post_id"
                 reference="posts"
-                validate={[required]}
+                validate={required}
             >
                 <SelectInput optionText="title" />
             </ReferenceInput>
             <DateInput source="created_at" />
-            <LongTextInput source="body" validate={[required]} />
+            <LongTextInput source="body" validate={required} />
         </SimpleForm>
     </Create>
 );
--- a/src/mui/input/ReferenceArrayInput.js
+++ b/src/mui/input/ReferenceArrayInput.js
@@ -2,16 +2,15 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import debounce from 'lodash.debounce';
-import LinearProgress from 'material-ui/LinearProgress';
-import TextField from 'material-ui/TextField';
 
-import Labeled from '../input/Labeled';
 import {
     crudGetMany as crudGetManyAction,
     crudGetMatching as crudGetMatchingAction,
 } from '../../actions/dataActions';
 import { getPossibleReferences } from '../../reducer/admin/references/possibleValues';
-import { progessStyle, progessContainerStyle } from './ReferenceInput';
+import ReferenceLoadingProgress from './ReferenceLoadingProgress';
+import ReferenceError from './ReferenceError';
+import translate from '../../i18n/translate';
 
 const referenceSource = (resource, source) => `${resource}@${source}`;
 
@@ -167,6 +166,7 @@ export class ReferenceArrayInput extends Component {
             onChange,
             children,
             meta,
+            translate,
         } = this.props;
 
         if (React.Children.count(children) !== 1) {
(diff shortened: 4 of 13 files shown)


src/mui/input/ReferenceInput.js
```diff
@@ -164,20 +175,24 @@ export class ReferenceInput extends Component {
             onChange,
             children,
             meta,
+            record,
         } = this.props;
-        if (!referenceRecord && !allowEmpty) {
+        if (!matchingReferences.length) {
```
fzaninotto: not good: for instance, for an autocomplete, if I type a radom string, there will be no matching reference, even once the fetch has succeeded

src/mui/input/ReferenceInput.js
```diff
@@ -218,6 +233,7 @@ ReferenceInput.propTypes = {
     meta: PropTypes.object,
     onChange: PropTypes.func,
     perPage: PropTypes.number,
+    record: PropTypes.object,
```
fzaninotto: unused

src/reducer/admin/references/possibleValues.js
```diff
@@ -9,6 +12,11 @@ export default (previousState = initialState, { type, payload, meta }) => {
                 ...previousState,
                 [meta.relatedTo]: payload.data.map(record => record.id),
             };
+        case CRUD_GET_MATCHING_FAILURE:
+            return {
+                ...previousState,
+                [meta.relatedTo]: { error: payload.error },
```
alexisjanvier: I'm not very fan of having either an object or an array for this state value... Does anyone have a better solution in mind?

src/mui/input/ReferenceArrayInput.js
```diff
@@ -172,6 +176,43 @@ export class ReferenceArrayInput extends Component {
...
+                        ) : (
+                            label
+                        )
+                    }
+                    errorText={matchingReferences.error}
+                />
+            );
+        }
+
         if (!(referenceRecords && referenceRecords.length > 0) && !allowEmpty) {
```
alexisjanvier: IMO, `refrenceRecords` should be deprecated. @fzaninotto @djhi are you agree with that ?

djhi: Can you elaborate ?

alexisjanvier: Here, we test if we have the referenceIds data in the state resources
```
const referenceIds = props.input.value || [];
const data = state.admin.resources[props.reference].data;
referenceRecords: referenceIds.reduce((references, referenceId) => {
            if (data[referenceId]) {
                references.push(data[referenceId]);
            }
            return references;
        }, []),
```
If not, and if `allowEmpty` is not `true`, we show a disabled input. 
I think we don't need to do this anymore (because we test if value is required with `validate`, and not with `allowEmpty` prop), so we should remove this `recordReference` from props.

src/mui/input/ReferenceArrayInput.js
```diff
@@ -182,8 +185,29 @@ export class ReferenceArrayInput extends Component {
                             label
                         )
                     }
-                    source={source}
-                    resource={resource}
+                >
+                    <span style={progessContainerStyle}>
```
djhi: Why not extracting completely this component as `ReferenceLoadingProgress` and put the styles in it ?

src/reducer/admin/references/possibleValues.js
djhi: Add another property ?

```js
[`${meta.relatedTo}`_error]
```

docs/Tutorial.md
```diff
@@ -249,11 +249,12 @@ const PostTitle = ({ record }) => {
     return <span>Post {record ? `"${record.title}"` : ''}</span>;
 };
 
+const required = value => (value ? undefined : 'Required');
 export const PostEdit = (props) => (
     <Edit title={<PostTitle />} {...props}>
         <SimpleForm>
             <DisabledInput source="id" />
-            <ReferenceInput label="User" source="userId" reference="users">
+            <ReferenceInput label="User" source="userId" reference="users" validate={[required]}>
```
fzaninotto: I think you must now extract the validation array, too, due to a redux-form BC break. Ping @djhi

src/mui/input/ReferenceArrayInput.js
fzaninotto: 👍

src/mui/input/ReferenceArrayInput.js
```diff
@@ -182,8 +185,29 @@ export class ReferenceArrayInput extends Component {
...
+                </Labeled>
+            );
+        }
+
+        if (matchingReferences.error) {
+            return (
+                <TextField
+                    disabled={true}
+                    hintText={
+                        typeof label === 'undefined' ? (
```
fzaninotto: missing call to translation helper

src/mui/input/ReferenceArrayInput.js
```diff
@@ -182,8 +185,29 @@ export class ReferenceArrayInput extends Component {
...
+                            `resources.${resource}.fields.${source}`
+                        ) : (
+                            label
+                        )
+                    }
+                    errorText={matchingReferences.error}
```
fzaninotto: missing call to translation helper

src/mui/input/ReferenceArrayInput.js
```diff
@@ -172,7 +175,7 @@ export class ReferenceArrayInput extends Component {
             );
         }
 
-        if (!(referenceRecords && referenceRecords.length > 0) && !allowEmpty) {
```
fzaninotto: You miss a case where the current record references non-existent records in the distant resource. E.g. an article with a list of author_ids, some of which don't exist in the Author resource. The previous code was there to avoid the app to break in that case (and we should keep this security).

src/mui/input/ReferenceInput.js
```diff
@@ -157,15 +169,21 @@ export class ReferenceInput extends Component {
...
             basePath,
             onChange,
             children,
             meta,
         } = this.props;
-        if (!referenceRecord && !allowEmpty) {
+
+        if (React.Children.count(children) !== 1) {
+            throw new Error(
+                '<ReferenceInput> only accepts a single child (like <SelectInput>)'
```
fzaninotto: s/ (like <SelectInput>)//

src/mui/input/ReferenceInput.js
```diff
@@ -175,8 +193,29 @@ export class ReferenceInput extends Component {
...
+                    </span>
+                </Labeled>
+            );
+        }
+
+        if (matchingReferences.error) {
+            return (
+                <TextField
+                    disabled={true}
+                    hintText={
```
fzaninotto: missing call to translation helper. Same for error.

src/mui/input/ReferenceInput.js
```diff
@@ -157,15 +169,21 @@ export class ReferenceInput extends Component {
...
             source,
-            referenceRecord,
             allowEmpty,
             matchingReferences,
             basePath,
             onChange,
             children,
             meta,
         } = this.props;
-        if (!referenceRecord && !allowEmpty) {
```
fzaninotto: Same remark as for the `ReferenceArrayInput`, the app musn't break if the `referenceRecord` doesn't exist.

src/reducer/admin/references/possibleValues.js
fzaninotto: We already use this pattern otherwise, not a big deal IMO

src/reducer/admin/references/possibleValues.js
```diff
@@ -20,17 +28,23 @@ export const getPossibleReferences = (
...
-        referenceSource
-    ]
-        ? Array.from(state.admin.references.possibleValues[referenceSource])
-        : [];
+    if (!state.admin.references.possibleValues[referenceSource]) {
+        return null;
+    }
+
+    if (state.admin.references.possibleValues[referenceSource].error) {
+        return state.admin.references.possibleValues[referenceSource];
```
fzaninotto: why skip the special map / filter processing of the possible values in that case?

docs/Tutorial.md
djhi: Probably but doesn't validate support a single function too ?

`validate={required}`

src/mui/input/ReferenceArrayInput.js
djhi: Don't forget to the `_` option

src/reducer/admin/references/possibleValues.js
alexisjanvier: Okay, I'll leave it like that.

src/reducer/admin/references/possibleValues.js
alexisjanvier: Because this case corresponds to a fetch error return

src/mui/input/ReferenceArrayInput.js
```diff
@@ -278,13 +267,12 @@ function mapStateToProps(state, props) {
...
+const enhance = compose(
+    connect(mapStateToProps, {
+        crudGetMany: crudGetManyAction,
+        crudGetMatching: crudGetMatchingAction,
+    }),
+    translate
+);
 
-export default ConnectedReferenceInput;
+export default enhance(ReferenceArrayInput);
```
djhi: The exported component needs to define its default props, with `addField: true` in it

src/mui/input/ReferenceInput.js
```diff
@@ -288,13 +263,12 @@ function mapStateToProps(state, props) {
...
+const enhance = compose(
+    connect(mapStateToProps, {
+        crudGetOne: crudGetOneAction,
+        crudGetMatching: crudGetMatchingAction,
+    }),
+    translate
+);
 
-export default ConnectedReferenceInput;
+export default enhance(ReferenceInput);
```
djhi: The exported component needs to define its default props, with addField: true in it

alexisjanvier:
@fzaninotto @djhi review apllied

src/mui/input/ReferenceInput.js
```diff
@@ -157,28 +160,37 @@ export class ReferenceInput extends Component {
...
-            referenceRecord,
             allowEmpty,
             matchingReferences,
             basePath,
             onChange,
             children,
             meta,
+            translate,
         } = this.props;
-        if (!referenceRecord && !allowEmpty) {
```
djhi: Aren't we still missing that case ?

djhi:
I think we are missing some cases here:

`ReferenceInput`: If the currently referenced resource cannot be fetched for any reason, we should display that information without preventing edition.

`ReferenceArrayInput`: If any of the currently referenced resources cannot be fetched for any reason, we should display that information for each of them (even all if needed) without preventing edition.

Ping @fzaninotto

src/mui/input/ReferenceInput.js
fzaninotto: What we should do: 

- Display an error message saying that the reference wasn't found in the API
- display the input with an empty value. It should be functional (i.e. allow to set a new value)

src/mui/input/ReferenceInput.js
```diff
@@ -157,28 +160,37 @@ export class ReferenceInput extends Component {
...
+
+        if (React.Children.count(children) !== 1) {
+            throw new Error('<ReferenceInput> only accepts a single child');
+        }
+
+        if (matchingReferences === null) {
+            return <ReferenceLoadingProgress label={translatedLabel} />;
+        }
+
+        if (error) {
```
fzaninotto: Same as above, please handle the case where the `matchingReferences` are loaded but not the `referenceRecord`.

src/mui/input/ReferenceArrayInput.js
```diff
@@ -172,20 +176,32 @@ export class ReferenceArrayInput extends Component {
...
+        if (
+            !error &&
+            input.value &&
+            input.value.length > 0 &&
+            !referenceRecords.length
+        ) {
+            error = translate('aor.input.references.missing');
+        }
+
+        if (matchingReferences === null) {
```
fzaninotto: why don't you do that *before* building the error message? It would provide an early return.

src/mui/input/ReferenceInput.js
```diff
@@ -157,28 +160,37 @@ export class ReferenceInput extends Component {
...
+                : label;
+        const translatedLabel = translate(finalLabel, { _: finalLabel });
+        const error =
+            matchingReferences && matchingReferences.error
+                ? translate(matchingReferences.error, {
+                      _: matchingReferences.error,
+                  })
+                : null;
+
+        if (React.Children.count(children) !== 1) {
```
fzaninotto: Same as above, please move up to return earlier.

src/mui/input/ReferenceInput.js
```diff
@@ -157,28 +160,37 @@ export class ReferenceInput extends Component {
...
+            throw new Error('<ReferenceInput> only accepts a single child');
+        }
+
+        if (matchingReferences === null) {
```
fzaninotto: I worry about another case:

1. The input displays and loads matching references => `matchingReferences` is null
2. The API returns an error => `matchingReferences` is `{ error: 'toto' }`
3. The user clicks "Refresh", which re-triggers the 2 fetch actions. But at that time, the value of `matchingReferences` is still an error, so the loader won't show...

src/mui/input/ReferenceInput.js
djhi: The error shouldn't be displayed by the `ReferenceInput` itself right ? It should just be passed to the child as `ReferenceInput` don't actually render anything.

fzaninotto: You're right

fzaninotto:
also, please consider #854

alexisjanvier:
@fzaninotto #854 is about `<Reference*Field>` not `<Reference*Input>`. This should be adressed in another PR.

alexisjanvier:
Here a summary of the new behavior of reference inputs

## `<ReferenceInput>`

#### A spinner is displayed as long as:
 * we search for matching references and there are no reference already linked to the ressource,
 * we search for matching references and also data for a reference already linked to the ressource.
capture d ecran de 2018-02-12 16-26-30

#### An error is displayed if:
 * we have a server error return on the search of matching references and that there are no reference already linked to the ressource,
 * we have a server error return on the search of matching references and the data of reference already associated with the object are not found.
capture d ecran de 2018-02-12 16-28-21

#### A warning is displayed with a children's error when : 
 * we have a server error return on the search of matching references but we found data for reference already associated with the object
 * we didn't find any data for a reference already associated with the object, but we have a valid return (even if empty) from the server on the search of matching references.
capture d ecran de 2018-02-12 16-29-25

## `<ReferenceArrayInput>`

#### A spinner is displayed as long as:
 * we search for matching references and there are no references already linked to the ressource,
 * we search for matching references and also data for any references already linked to the ressource.
capture d ecran de 2018-02-12 16-22-43

#### An error is displayed if:
 * we have a server error return on the search of matching references and that there are no references already linked to the ressource,
 * we have a server error return on the search of matching references and no data for any of the references already linked to the ressource were found.
capture d ecran de 2018-02-12 16-24-19

#### A warning is displayed with a children's error when : 
 * we have a server error return on the search of matching references but we found the data for at least one reference already linked with the ressource,
 * We did not find the data for at least one of the references already associated with the ressource, but we have a valid return (even if empty) from the server on the search of matching references.
capture d ecran de 2018-02-12 16-25-11

docs/Inputs.md
```diff
@@ -580,7 +580,8 @@ import { ReferenceInput, SelectInput } from 'admin-on-rest'
 </Admin>
 ```
 
-Set the `allowEmpty` prop when the empty value is allowed.
+Set the `allowEmpty` prop when you want to add an empty choice with a value of null in the choices list.
+Be careful, disabling `allowEmpty` does not mean that the input will be required. If you want to make the input required, you must add a validator as indicated in [Validation Documentation](./CreateEdit.html#validation). Enabling the `allowEmpty` props just adds an empty choice (with `null` value) on top of the options, and makes the value nullable.
```
djhi: s/Be careful, disabling/Disabling/

docs/Inputs.md
```diff
@@ -695,7 +696,8 @@ import { ReferenceArrayInput, SelectArrayInput } from 'admin-on-rest'
...
```
djhi: s/Be careful, disabling/Disabling/

example/comments.js
```diff
@@ -177,19 +178,20 @@ export const CommentEdit = props => (
 );
 
 const defaultValue = { created_at: new Date() };
+const required = value => (value ? undefined : 'Required');
```
djhi: You can simply import the `required` function from `admin-on-rest`

docs/Tutorial.md
```diff
@@ -249,11 +249,12 @@ const PostTitle = ({ record }) => {
     return <span>Post {record ? `"${record.title}"` : ''}</span>;
 };
 
+const required = value => (value ? undefined : 'Required');
```
djhi: You can simply import the `required` function from `admin-on-rest`

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,701 @@ import React from 'react';
...
+                    ...data,
+                    input: { value: [1, 2] },
+                    matchingReferences: { error: 'error' },
+                },
+                false,
+                'it is ready if linked reference data are not found, but the references fetch is finished with error.'
+            );
+        });
+
+        it('should claim an error if needed', () => {
```
djhi: claim ?

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,701 @@ import React from 'react';
...
+            assert.equal(ErrorElement.length, 0);
+            const MyComponentElement = wrapper.find('MyComponent');
+            assert.equal(MyComponentElement.length, 1);
+            assert.deepEqual(MyComponentElement.prop('meta'), {
+                error: '*fetch error*',
+                touched: true,
+            });
+        });
+
+        it('should send an error to the children if references were found and but selected references are not complete', () => {
```
djhi: s/and but/and/

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,701 @@ import React from 'react';
...
+            assert.equal(ErrorElement.length, 0);
+            const MyComponentElement = wrapper.find('MyComponent');
+            assert.equal(MyComponentElement.length, 1);
+            assert.deepEqual(MyComponentElement.prop('meta'), {
+                error: '*aor.input.references.many_missing*',
+                touched: true,
+            });
+        });
+
+        it('should send an error to the children if references were found and but selected references are empty', () => {
```
djhi: s/and but/and/

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,701 @@ import React from 'react';
...
+            );
+            const SpinnerElement = wrapper.find('ReferenceLoadingProgress');
+            assert.equal(SpinnerElement.length, 0);
+            const ErrorElement = wrapper.find('ReferenceError');
+            assert.equal(ErrorElement.length, 0);
+            const MyComponentElement = wrapper.find('MyComponent');
+            assert.equal(MyComponentElement.length, 1);
+        });
+
+        it('should not render enclosed component if allowEmpty is true', () => {
```
djhi: should not render ?

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,701 @@ import React from 'react';
...
+                    {...defaultProps}
+                    allowEmpty
+                    meta={{ touched: false }}
+                >
+                    <MyComponent />
+                </ReferenceArrayInput>
+            );
+
+            const myComponent = wrapper.find('MyComponent');
+            assert.notEqual(myComponent.prop('meta', undefined));
```
djhi: `assert.deepEqual(myComponent.prop('meta'), { touched: false })` ?

src/mui/input/ReferenceInput.js
```diff
@@ -2,15 +2,59 @@ import React, { Component } from 'react';
...
     crudGetMatching as crudGetMatchingAction,
 } from '../../actions/dataActions';
 import { getPossibleReferences } from '../../reducer/admin/references/possibleValues';
+import ReferenceLoadingProgress from './ReferenceLoadingProgress';
+import ReferenceError from './ReferenceError';
+import translate from '../../i18n/translate';
 
 const referenceSource = (resource, source) => `${resource}@${source}`;
 const noFilter = () => true;
+const getFinalLabel = (label, resource, source) =>
```
djhi: useless function imo, why not just put the following where it is called ?
```jsx
label || `resources.${resource}.fields.${source}`
```

src/mui/input/ReferenceArrayInput.js
```diff
@@ -2,14 +2,74 @@ import React, { Component } from 'react';
...
     crudGetMany as crudGetManyAction,
     crudGetMatching as crudGetMatchingAction,
 } from '../../actions/dataActions';
 import { getPossibleReferences } from '../../reducer/admin/references/possibleValues';
+import ReferenceLoadingProgress from './ReferenceLoadingProgress';
+import ReferenceError from './ReferenceError';
+import translate from '../../i18n/translate';
 
 const referenceSource = (resource, source) => `${resource}@${source}`;
+const getFinalLabel = (label, resource, source) =>
```
djhi: useless function imo, why not just put the following where it is called ?
```jsx
label || `resources.${resource}.fields.${source}`
```

src/mui/input/ReferenceArrayInput.spec.js
```diff
@@ -2,202 +2,704 @@ import React from 'react';
...
-        const wrapper = shallow(
-            <ReferenceArrayInput {...defaultProps}>
-                <MyComponent />
-            </ReferenceArrayInput>
-        );
-        const MyComponentElement = wrapper.find('MyComponent');
-        assert.equal(MyComponentElement.length, 0);
-    });
+import {
+    getSelectedReferencesStatus,
```
fzaninotto: It feels strange to test individual methods of a component. IMO, these are logic methods which should be placed out of the component. But we'll do that in another PR.

src/mui/input/ReferenceArrayInput.js
```diff
@@ -172,33 +230,44 @@ export class ReferenceArrayInput extends Component {
...
-            label:
-                typeof label === 'undefined'
-                    ? `resources.${resource}.fields.${source}`
-                    : label,
+            label: translatedLabel,
             resource,
-            meta,
+            meta: {
+                ...meta,
+                error: dataStatus.warning,
```
fzaninotto: I'm afraid this may override actual errors from the filed level validator. Did you test it?

src/mui/input/ReferenceArrayInput.js
```diff
@@ -232,44 +301,45 @@ ReferenceArrayInput.propTypes = {
...
     crudGetMatching: crudGetMatchingAction,
 })(ReferenceArrayInput);
 
-ConnectedReferenceInput.defaultProps = {
+ConnectedReferenceArrayInput.defaultProps = {
     addField: true,
 };
 
-export default ConnectedReferenceInput;
+export default translate(ConnectedReferenceArrayInput);
```
fzaninotto: Why do you decorate with translate here and not line 336?

src/mui/input/ReferenceArrayInput.js
```diff
@@ -2,15 +2,72 @@ import React, { Component } from 'react';
...
 import { getPossibleReferences } from '../../reducer/admin/references/possibleValues';
+import ReferenceLoadingProgress from './ReferenceLoadingProgress';
+import ReferenceError from './ReferenceError';
+import translate from '../../i18n/translate';
 
 const referenceSource = (resource, source) => `${resource}@${source}`;
 
+export const getSelectedReferencesStatus = (input, referenceRecords) =>
+    !input.value || input.value.length === referenceRecords.length
+        ? 'ready'
```
fzaninotto: this should be a constant, same for 'incomplete' and 'empty'

src/mui/input/ReferenceArrayInput.js
alexisjanvier: After review, it's working.

ezgif com-video-to-gif

ezgif com-video-to-gif 1

src/mui/input/ReferenceArrayInput.js
alexisjanvier: After review, it's working.

ezgif com-video-to-gif

ezgif com-video-to-gif 1

alexisjanvier:
@fzaninotto Review applied

src/mui/input/ReferenceArrayInput.js
```diff
@@ -336,10 +344,10 @@ function mapStateToProps(state, props) {
 const ConnectedReferenceArrayInput = connect(mapStateToProps, {
     crudGetMany: crudGetManyAction,
     crudGetMatching: crudGetMatchingAction,
-})(ReferenceArrayInput);
+})(translate(ReferenceArrayInput));
```
djhi: would be better to use `recompose/compose` here, as done everywhere

fzaninotto:
merged \o/

Now @djhi, you "just" have to port it to `next`...

PR #1460, opened 2018-02-01
[RFR] Backport master 1.3.5

djhi:


packages/react-admin/src/mui/field/ReferenceField.js
```diff
@@ -2,13 +2,22 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import get from 'lodash.get';
+import { withStyles } from 'material-ui/styles';
+import compose from 'recompose/compose';
+import classenames from 'classnames';
```
fzaninotto: typo

packages/react-admin/src/mui/field/ReferenceField.js
```diff
@@ -135,6 +145,7 @@ ReferenceField.propTypes = {
     allowEmpty: PropTypes.bool.isRequired,
     basePath: PropTypes.string.isRequired,
     children: PropTypes.element.isRequired,
+    classes: PropTypes.object,
```
fzaninotto: add to default props, too (`{}`) to ease tests.

packages/react-admin/src/mui/field/ReferenceField.spec.js
```diff
@@ -16,6 +20,7 @@ describe('<ReferenceField />', () => {
                 reference="bar"
                 basePath=""
                 crudGetManyAccumulate={crudGetManyAccumulate}
+                classes={classes}
```
fzaninotto: should not be necessary once you have default prop for `classes`.

packages/react-admin/src/mui/field/ReferenceField.js
djhi: Done

Issue #1120, opened 2017-10-03
[RFC] Filters get rerendered and change value on typing


kopax:
# issue type

- [x] bug

# description

We have implemented the CustomApp example so we are not using `<Admin />`.

When we try to type in any filter fields, the input text is replaced constantly making the typing inconvenient.

![out](https://user-images.githubusercontent.com/1866564/31112190-7c51547e-a83c-11e7-9e0e-5468723f7cce.gif)

I think this is due to the filter setting the URL that trigger the `@@router/LOCATION_CHANGE`.

Version 1.3.1

Thanks again for the awesome lib.

fzaninotto:
The reducer part seems to be outdated in the CustomApp tutorial. Can you try the following:

```js
// replace
const reducer = combineReducers({
    admin: adminReducer,
    locale: localeReducer(),
    form: formReducer,
    routing: routerReducer,
});

// with
import { createAppReducer } from 'admin-on-rest';
const reducer = createAppReducer({});
```

?

kopax:
I have replaced

```jsx
import { combineReducers } from 'redux';
import { routerReducer } from 'react-router-redux';
import { reducer as formReducer } from 'redux-form';
import { adminReducer, localeReducer } from 'admin-on-rest';
export default function createReducer(injectedReducers) {
  return combineReducers({
    route: routerReducer,
    admin: adminReducer,
    locale: localeReducer(),
    form: formReducer,
    ...injectedReducers,
  });
}
```

with

```jsx
import { routerReducer } from 'react-router-redux';
import { createAppReducer } from 'admin-on-rest';

export default function createReducer(injectedReducers) {
  return createAppReducer({
    route: routerReducer,
    ...injectedReducers,
  });
}
```

But this didn't solve anything.

fzaninotto:
I really don't know where it comes from. This works fine with the standard `<Admin>` app. I suggest you look at the code of `<Admin>` and `<Resource>` to see the differences with your implementation.

fzaninotto:
wait, `createAppReducer` already includes the router reducer. You're adding it twice...

kopax:
The `createAppReducer` indeed contain the `routing` store. But now it's exactly the same as before except I am using `routing` for `route`. 
Our only way around was to use the `<Admin />` and stick to `mui` for the frontend. :(
I have no clue either why it is going crazy like this, do you want me to create a GitHub reproduction ?

fzaninotto:
Any news about this one?

[...]

PR #1462, opened 2018-02-02
[RFR] Refactor resources handling

### LINKED ISSUE #1445: CustomRoutes refresh does not trigger `declareResources`.
**What you were expecting:**
Refreshing page from a  declared `customRoutes` using F5 and remaining in the same location and context.

**What happened instead:**
App crashes telling me: `Cannot read property 'list' of undefined`

**Steps to reproduce:**
I am not sure how to reproduce the error as I have other `customRoutes` that do not have the same issue

**Related code:**
First, I realized that in most other situations, or if I move the same routes from `customRoutes` to `Resource` just for the intention of testing the App would not crash. The bug seems to be inconsistent.

After logging in `initializeResources(children)` and within the `List` component, it seemed that `this.props.declareResources` was not called before the `List` component was mounted.

**Environment**

* React version: "16.0.0",
* Browser: Chrome

djhi:
Fixes #1445

- [x] Refactor resources handling
- [x] Added e2e tests to validate

### DIFF
--- a/e2e/pages/CustomPage.js
+++ b/e2e/pages/CustomPage.js
@@ -1,0 +1,45 @@
+import { By, until } from 'selenium-webdriver';
+
+export default url => driver => ({
+    elements: {
+        appLoader: By.css('.app-loader'),
+        total: By.css('.total'),
+        bodyFromLayout: By.css('.body'),
+    },
+
+    navigate() {
+        driver.navigate().to(url);
+        return this.waitUntilDataLoaded();
+    },
+
+    waitUntilDataLoaded() {
+        let continued = true;
+        return driver
+            .wait(until.elementLocated(this.elements.appLoader), 2000)
+            .catch(() => (continued = false)) // no loader - we're on the same page !
+            .then(
+                () =>
+                    continued
+                        ? driver.wait(
+                              until.stalenessOf(
+                                  driver.findElement(this.elements.appLoader)
+                              )
+                          )
+                        : true
+            )
+            .catch(() => {}) // The element might have disapeared before the wait on the previous line
+            .then(() => driver.sleep(100)); // let some time to redraw
+    },
+
+    hasBody() {
+        return driver
+            .findElements(this.elements.bodyFromLayout)
+            .then(elements => elements.length > 0);
+    },
+
+    getTotal() {
+        return driver
+            .findElement(this.elements.total)
+            .then(element => element.getText());
+    },
+});
--- a/e2e/tests/customPages.js
+++ b/e2e/tests/customPages.js
@@ -1,0 +1,36 @@
+import assert from 'assert';
+import driver from '../chromeDriver';
+import customPageFactory from '../pages/CustomPage';
+
+describe('Custom Pages', () => {
+    const CustomPageNoLayout = customPageFactory(
+        'http://localhost:8083/#/custom'
+    )(driver);
+
+    const CustomPageWithLayout = customPageFactory(
+        'http://localhost:8083/#/custom2'
+    )(driver);
+
+    describe('Without Layout', () => {
+        before(async () => await CustomPageNoLayout.navigate());
+
+        it('should not display the layout', async () => {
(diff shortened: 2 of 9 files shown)


santaclauze:
https://github.com/santaclauze/aor-customRoutes-bug

Still have the same problem. There are situations where we would like to pass a component that has been created with your own components List/Create/Edit/Show but in a customeRoutes.

In my example, PostList and UserList are passed to customeRoutes and still return an error:
`Cannot read property 'list' of undefined`.

As stated in my original post, it seems that the resources are set but not yet loaded.

djhi:
No, #1462 targets `master` while #1466 targets `next`. We haven't released this fix yet

santaclauze:
kk ty for the details. Looking forward.

kopax:
I have forked `master` and tagged a version [1.3.2-fixed](https://github.com/kopaxgroup/admin-on-rest/tree/1.3.2-fixed) with `lib` pushed so it can be installed from `npm` : 

I have used this version over the @santaclauze [reproduction](https://github.com/marmelab/admin-on-rest/issues/1445#issuecomment-361818465) with these additionals commands: 

    npm install --save git+https://git@github.com:kopaxgroup/admin-on-rest.git#1.3.2-fixed
    make install
    make run

`customRoutes` aren't rendering at all, Is this supposed to fix #1445 ?

fzaninotto:
1.4.0 has been released today. Can you check if it fixes your issue?

santaclauze:
@fzaninotto unfortunately not. 

The problem seems to be the same, `declareResources` is not called before `customRoutes` is rendered. 

https://github.com/santaclauze/aor-customRoutes-bug-test-2

```
make install
make run
```

then navigate between:

http://localhost:8081/#/posts
and
http://localhost:8081/#/comments

and refresh the pages manually.

I left some logs which seems to suggest that `declareResources` is now called, but only after the rendering of the component which creates the error related to `mapStateToProps` as the data is not yet loaded in the resource that we are trying to display (in this case comments).

test_aor

fzaninotto:
ping @djhi

djhi:
No, they are, see the custom routes pages included in the simple example. The issue here is that their declaration is asynchronous in the simple example but our default CRUD pages (List, Create, Edit, Show and Delete) are not meant to be used inside custom routes.

@fzaninotto: we can make them usable in custom routes quite easily by fixing the `mapStateToProps` for each of them, ensuring the resource is here before trying to read it.

fzaninotto:
Not in 1.0, this is a new feature. 

@santaclauze I suggest you open an enhancement request for using `List`, `Create`, `Edit`, `Show` and `Delete` components in custom routes. We may implement it in version 2.0.

afilp:
@fzaninotto This would be great for 2.0, I also need this functionality. Thanks!

kopax:
> @fzaninotto: we can make them usable in custom routes quite easily by fixing the mapStateToProps for each of them, ensuring the resource is here before trying to read it.

@djhi just as a comment: I believe it would not be the best place to fix it. As people may extend react-admin, it should be done at the application level instead of component level.

fzaninotto:
@kopax if you have an idea of how to fix it properly, please open a PR!

kopax:
@fzaninotto is there a chance to get in touch with you?

fzaninotto:
Sure, you can email me using my first name at my company name, followed by a dot and then by a com.

kopax:
@fzaninotto I've sent you an email

fzaninotto:
@kopax I didn't receive it. I suggest you go to the marmelab.com website and hit the Contact button.

kopax:
@fzaninotto I've forwarded the email to contact@.

fzaninotto:
Still not received. I believe your email wasn't sent.

kopax:
> Not in 1.0, this is a new feature.

@fzaninotto not sure if this is related but I've noticed that it also break the `customApp.md`. See [#1120](https://github.com/marmelab/admin-on-rest/issues/1120) for more details.

If this is also 1.0 issues, should it be fixed in 1.0 ?

PR #1466, opened 2018-02-02
[RFR] Refactor resources handling

djhi:


Makefile
```diff
@@ -25,36 +25,36 @@ run-graphcool-demo: build ## run the demo example
 
 build-react-admin:
 	@rm -rf ./packages/react-admin/lib
-	@NODE_ENV=production ./node_modules/.bin/babel ./packages/react-admin/src -d ./packages/react-admin/lib --ignore '*.spec.js' --ignore '*.test.js'
```
djhi: tests files were added to lib, leading to errors in tests

packages/react-admin/src/AdminRoutes.js
```diff
@@ -1,71 +1,103 @@
...
     componentWillMount() {
         this.initializeResources(this.props);
     }
 
-    initializeResources = nextProps => {
-        if (typeof nextProps.children === 'function') {
-            this.initializeResourcesAsync(nextProps);
+    initializeResources = props => {
+        if (typeof props.children === 'function') {
+            props.authClient(AUTH_GET_PERMISSIONS).then(permissions => {
```
alexisjanvier: There is no risk that resources won't be declared when custom routes will be mounted?

djhi: The e2e tests confirm that yes :)

packages/react-admin/src/AdminRoutes.js
```diff
@@ -81,69 +113,114 @@ export class AdminRoutes extends Component {
...
-                                        to={`/${childrenToRender[0].props
-                                            .name}`}
+                <Route
+                    path="/"
+                    render={() =>
+                        createElement(appLayout, {
+                            children: (
+                                <Switch>
+                                    {customRoutes &&
+                                        customRoutes.map((route, index) => (
```
fzaninotto: you should filter out the custom routes without layout, which were already rendered above.

packages/react-admin/src/AdminRoutes.js
```diff
@@ -81,69 +113,114 @@ export class AdminRoutes extends Component {
...
+                                                render={route.props.render}
+                                                children={route.props.children} // eslint-disable-line react/no-children-prop
+                                            />
+                                        ))}
+                                    {Children.map(childrenToRender, child => (
+                                        <Route
+                                            path={`/${child.props.name}`}
+                                            render={props =>
+                                                cloneElement(child, {
+                                                    context: 'route',
```
fzaninotto: So you did keep the context after all...

packages/react-admin/src/mui/layout/Layout.js
```diff
@@ -97,14 +97,7 @@ const Layout = ({
                         open && classes.contentShift
                     )}
                 >
-                    <AdminRoutes
```
fzaninotto: if the layout changes, this is a BC break, and it should be documented in the UPGRADE guide. The theming doc needs an update, too.

e2e/tests/server.js
```diff
@@ -23,20 +23,6 @@ before(
         })
 );
 
-before(() => {
```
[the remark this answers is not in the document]
djhi: Because I had weird issues when some tests were trying to click buttons hidden under notifications. You may have noticed that the chrome driver is not started with the `start-maximized` option to ensure the largest screen possible. 

Another option would have been to implement some custom webdriver code which would allow to scroll until an element is clickable but it required more time to implement

packages/react-admin/src/AdminRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        return (
+            <div>
+                {// Render every resources children outside the React Router Switch
+                // as we need all of them and not just the one rendered
+                Children.map(childrenToRender, child =>
+                    cloneElement(child, {
+                        key: child.props.name,
+                        // The context prop instruct the Resource component to not render anything
```
fzaninotto: s/instruct/instructs/

packages/react-admin/src/AdminRouter.js
```diff
@@ -0,0 +1,193 @@
...
+                        // but simply register itself as a known resource
```
fzaninotto: s/register/to register/

fzaninotto:
Needs rebase now that `authClient` has been renamed

packages/react-admin/src/mui/layout/Layout.js
djhi: Done

PR #1477, opened 2018-02-05
[RFR] Update documentation about new RefreshButton syntax

### LINKED ISSUE #1473: Is custom action of <List> workable?
I'm trying to add a button on the top of the <List>, then I copied the sample code like this:
```
import { showNotification, CardActions, CreateButton, FlatButton, NavigationRefresh} from 'admin-on-rest';

const cardActionStyle = {
    zIndex: 2,
    display: 'inline-block',
    float: 'right',
};

const PostActions = ({ resource, filters, displayedFilters, filterValues, basePath, showFilter, refresh }) => (
    <CardActions style={cardActionStyle}>
        {filters && React.cloneElement(filters, { resource, showFilter, displayedFilters, filterValues, context: 'button' }) }
        <CreateButton basePath={basePath} />
        <FlatButton primary label="refresh" onClick={refresh} icon={<NavigationRefresh />} />
        {/* Add your custom actions */}
        <FlatButton primary label="Custom Action" onClick={customAction} />
    </CardActions>
);

    <List {...props} actions={<PostActions />}>
        ...
    </List>
);
```

I found the problems as followed:
1. refresh is undefined since the import works fine
2. customAction should be a function, I guess, so I declare a function like this:
```
const customAction = function () {
    fetch('/crawler', { method: 'GET'})
        .then(() => {
            showNotification('Crawled All!');
        })
        .catch((e) => {
            console.error(e);
            showNotification('Error: crawle all', 'warning')
        });
}
```
But there's the same error like "refresh", here's error messages:
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `ListActions`.

I knew the custom button could do the same job but it's complicated to my requirement that's just to call a API and show a message of result.
So any hint will be appreciated!

alexisjanvier:
Fix #1473

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -115,8 +115,7 @@ You can replace the list of default actions by your own element using the `actio
 ```jsx
 import { CardActions } from 'material-ui/Card';
 import FlatButton from 'material-ui/FlatButton';
-import NavigationRefresh from 'material-ui/svg-icons/navigation/refresh';
-import { ListButton, ShowButton, DeleteButton } from 'admin-on-rest';
+import { DeleteButton, ListButton, RefreshButton, ShowButton } from 'admin-on-rest';
 
 const cardActionStyle = {
     zIndex: 2,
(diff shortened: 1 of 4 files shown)


alexisjanvier:
yes, of course

alexisjanvier:
@djhi done

fzaninotto:
To be backported to `next`

fzaninotto:
It looks to be already OK in `next`

PR #1479, opened 2018-02-05
[RFR] Fix warnings in SimpleShowLayout

djhi:


djhi:
We were passing all rest props to the container `div`. So we got warnings about unknown props for it

PR #1486, opened 2018-02-05
[RFR] Upgrade material-ui to v1.0.0-beta.32

djhi:
The latest 2 versions of material-ui introduce BC breaks:

* https://github.com/mui-org/material-ui/releases/tag/v1.0.0-beta.32
* https://github.com/mui-org/material-ui/releases/tag/v1.0.0-beta.31

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -646,7 +646,7 @@ const UserCreateToolbar = ({ permissions, ...props }) =>
                 label="user.action.save_and_add"
                 redirect={false}
                 submitOnEnter={false}
-                raised={false}
+                variant="flat"
             />}
     </Toolbar>;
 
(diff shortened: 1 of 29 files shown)


UPGRADE.md
```diff
@@ -646,7 +646,7 @@ const UserCreateToolbar = ({ permissions, ...props }) =>
                 label="user.action.save_and_add"
                 redirect={false}
                 submitOnEnter={false}
-                raised={false}
+                variant="flat"
```
fzaninotto: This is the default, so it's not necessary. See https://material-ui-next.com/demos/buttons/.

Same for all the other flat buttons.

djhi: That would be true if it was a simple Mui Button. However, our `SaveButton` set the variant to `raised` by default

fzaninotto: meh

djhi: See marmelab/admin-on-rest/pull/1486/files/a074016fcd20b1a9f2a2edd25ac58b3437a40c57#diff-3affbe736d22fedadd658b6ffd6cb70f

PR #1490, opened 2018-02-05
[RFR] Use React Router NavLink in menu

### LINKED ISSUE #1487: Use NavLink instead of Link in MenuItemLink component
To be able to highlight current resource

djhi:
screen

Fixes #1487

### DIFF
--- a/packages/react-admin/src/mui/layout/MenuItemLink.js
+++ b/packages/react-admin/src/mui/layout/MenuItemLink.js
@@ -5,7 +5,10 @@ import { MenuItem } from 'material-ui/Menu';
 import { withStyles } from 'material-ui/styles';
 
 const styles = theme => ({
-    active: { color: theme.palette.secondary.main },
+    active: {
+        backgroundColor: theme.palette.primary.main,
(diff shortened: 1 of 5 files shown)


djhi:
We don't have any documentation yet for the `MenuItemLink`. Moreover, we need to document the CSS API for our components which uses `classes`.

fzaninotto:
The use of secondary color is ugly IMHO. Material-ui uses an inverse background, I think it's better

image

packages/react-admin/src/mui/layout/MenuItemLink.js
```diff
@@ -20,6 +24,7 @@ export class MenuItemLink extends Component {
 
     render() {
         const {
+            activeClassName,
```
fzaninotto: why do you add that prop? Developers can already pass their own `classes`

djhi: because it's easier ? I'll remove it if you care

fzaninotto: yes please

PR #1491, opened 2018-02-05
[RFR] Allow to overwrite basePath on actions inside Datagrid

### LINKED ISSUE #800: Allow to override basePath on EditButton
```
<List {...props}>
    <Datagrid>
        <EditButton basePath="anything" />
    </Datagrid>
</List>
```

It seems basePath on any child of Datagrid will always be overwritten, that is: be the basePath of the record resource. More flexibility would be provided if the basePath set as prop would be retained and applied instead.

Adding the following after src/mui/list/DatagridCell.js:7 does the trick for me:
```
basePath = !!field.props.basePath && field.props.basePath !== basePath ? field.props.basePath : basePath
```
see [diff](https://github.com/pimschaaf/admin-on-rest/commit/b94a0b98a2d23293a87b0062bbb4df50a5c1062b)

I'm unsure if the code style is right and if more props should be retained similarly.

Kmaschta:
Fixes #800

It allows to overwrite the `basePath` of an `EditButton` inside a `Datagrid`:

```js
<List {...props}>
    <Datagrid>
        <EditButton basePath="anything" />
    </Datagrid>
</List>
```

I read the documentation but I'm not sure where to add this additional prop. Any help?

### DIFF
--- a/packages/react-admin/src/mui/list/DatagridCell.js
+++ b/packages/react-admin/src/mui/list/DatagridCell.js
@@ -40,7 +40,11 @@ export const DatagridCell = ({
         )}
         {...sanitizeRestProps(rest)}
     >
-        {React.cloneElement(field, { record, basePath, resource })}
+        {React.cloneElement(field, {
+            record,
+            basePath: field.props.basePath || basePath,
+            resource,
+        })}
     </TableCell>
 );
 
--- a/packages/react-admin/src/mui/list/DatagridCell.spec.js
+++ b/packages/react-admin/src/mui/list/DatagridCell.spec.js
@@ -1,17 +1,40 @@
 import assert from 'assert';
 import React from 'react';
+import PropTypes from 'prop-types';
 import { shallow } from 'enzyme';
 
 import { DatagridCell } from './DatagridCell';
 
 describe('<DatagridCell />', () => {
     const Field = () => <div />;
+    Field.propTypes = {
+        type: PropTypes.string,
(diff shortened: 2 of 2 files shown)


djhi:
> I read the documentation but I'm not sure where to add this additional prop. Any help?

In the `List` section ?

jmchza:
I thought this meant to change the basePath of the URL that hoes to the API not the internal one to the frontend site, hopefully that makes sense.

So essentially I want to override the basePath so I can call a different URL on the backend side. If this is not what was intended , how can I override a action of the EditButton or ShowButton to call a different URL?

Or the other option that I can think of is: if we have a way to override the attribute that gets picked at the moment the button gets clicked, that is for an instance for complex structures instead of sending the id to lookup you could use something like: user.id.

Thanks in advance.

djhi:
Hi, and thanks for your question. As explained in the [react-admin contributing guide](https://github.com/marmelab/react-admin/blob/master/README.md#contributing), the right place to ask a "How To" question, get usage advice, or troubleshoot your own code, is [StackOverFlow](http://stackoverflow.com/questions/tagged/react-admin).

This makes your question easy to find by the core team, and the developer community. Unlike Github, StackOverFlow has great SEO, gamification, voting, and reputation. That's why we chose it, and decided to keep GitHub issues only for bugs and feature requests.

And once you get a response, please continue to hang out on the react-admin channel in StackOverflow. That way, you can help newcomers and share your expertise!

PR #1492, opened 2018-02-05
[RFR] Refactor reference selectors

fzaninotto:
- [x] Change selectors signature to be `(state, store)` as much as possible
- [x] Moved selectors to the right reducer
- [x] Fix memoization not working when there are several ReferenceInputs on a page (cf https://github.com/reactjs/reselect#sharing-selectors-with-props-across-multiple-component-instances)
- [x] Added unit tests

Follows #1420

### DIFF
--- a/packages/react-admin/src/mui/input/ReferenceArrayInput.js
+++ b/packages/react-admin/src/mui/input/ReferenceArrayInput.js
@@ -34,6 +34,7 @@ const sanitizeRestProps = ({
     optionText,
     optionValue,
     record,
+    referenceSource,
     resource,
     allowEmpty,
     source,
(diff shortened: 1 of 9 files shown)


packages/react-admin/src/mui/input/ReferenceArrayInput.js
```diff
@@ -163,21 +164,29 @@ export class ReferenceArrayInput extends Component {
...
 
-    fetchOptions = ({ reference, source, resource } = this.props) => {
+    fetchOptions = (props = this.props) => {
+        const {
+            crudGetMatching,
+            reference,
+            source,
+            resource,
+            referenceSource,
+        } = props;
```
djhi: why not using destructuring directly in the method signature ?

packages/react-admin/src/mui/input/ReferenceArrayInput.js
```diff
@@ -289,33 +299,43 @@ ReferenceArrayInput.defaultProps = {
...
-        matchingReferences: getPossibleReferences(
-            referenceState,
-            possibleValues,
-            inputIds
-        ),
-    })
-);
 
-export default compose(
+const makeMapStateToProps = () =>
```
djhi: Why is it a function ?

packages/react-admin/src/mui/input/ReferenceArrayInput.js
fzaninotto: because it's less legible when there is a default value and many props IMO

PR #1493, opened 2018-02-06
[RFR] Rename `onMenuTap` to `onMenuClick` 

djhi:
Follow #1490

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -159,12 +159,12 @@ import { connect } from 'react-redux';
 import { MenuItemLink, getResources } from 'react-admin';
 import { withRouter } from 'react-router-dom';
 
-const Menu = ({ resources, onMenuTap, logout }) => (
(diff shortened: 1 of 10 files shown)


docs/Admin.md
```diff
@@ -190,7 +193,42 @@ const App = () => (
 );
 ```
 
-**Tip**: If you use authentication, don't forget to render the `logout` prop in your custom menu component. Also, the `onMenuTap` function passed as prop is used to close the sidebar on mobile.
+**Tip**: If you use authentication, don't forget to render the `logout` prop in your custom menu component. Also, the `onMenuClick` function passed as prop is used to close the sidebar on mobile.
+
+The `MenuItemLink` component make use of React Router [`NavLink`](https://reacttraining.com/react-router/web/api/NavLink) component, hence allowing to customize its style when it targets the current page.
```
fzaninotto: s/make use of React Router/makes use of the React Router/

PR #1304, opened 2017-12-18
[RFR] Deprecate SelectArrayInput component

djhi:

PR #1495, opened 2018-02-06
[RFR] SelectArrayInput

### LINKED ISSUE #1483: Re-implement SelectArrayInput in next
material-ui 1.0 supports `Select` with the `multiple` tag, see https://material-ui-next.com/demos/selects/#multiple-select. So we no longer need to depend on `material-ui-chip-input`.

It's now time to reintroduce the `SelectArrayInput`, knowing that it doesn't handle autocomplete (which is a BC break), but works for cases with a limited number of choices nonetheless.

djhi:
Fix #1483 

- [x] Implementation
- [x] Tests
- [x] Documentation

screen


docs/Inputs.md
```diff
@@ -728,6 +728,69 @@ import { SelectInput, ReferenceInput } from 'react-admin'
...
+```js
+const choices = [
+   { id: 'books', name: 'myroot.category.books' },
+   { id: 'sport', name: 'myroot.category.sport' },
+];
+```
+
+Lastly, use the `options` attribute if you want to override any of the `<Select>` attributes:
+
+{% raw %}
```
alexisjanvier: What's this `{% raw %}` ?

djhi: To avoid Jekyll messing with code samples

djhi: I must say I'm not sure when we should use it ^^

fzaninotto: when there is a double mustache in the code snippet

docs/Inputs.md
```diff
@@ -728,6 +728,69 @@ import { SelectInput, ReferenceInput } from 'react-admin'
...
+```js
+<SelectArrayInput source="category" options={{ fullWidth: true }} />
+```
+{% endraw %}
+
+Refer to [the Select documentation](http://www.material-ui.com/#/components/select) for more details.
+
+**Tip**: The `SelectArrayInput` component **cannot** be used inside a `ReferenceInput`.
```
fzaninotto: add: but it can be used inside a `ReferenceManyInput`

packages/react-admin/src/mui/input/SelectArrayInput.js
```diff
@@ -0,0 +1,260 @@
...
+            throw new Error(
+                "The SelectInput component wasn't called within a redux-form <Field>. Did you decorate it and forget to add the addField prop to your component? See https://marmelab.com/react-admin/Inputs.html#writing-your-own-input-component for details."
+            );
+        }
+        const { touched, error } = meta;
+
+        return (
+            <FormControl
+                className={classnames(classes.root, className)}
+                {...sanitizeRestProps(rest)}
```
fzaninotto: Am I right to think that is the user adds a prop to `SelectArrayInput` that she wants to pass to  to the `Select`, the `FormControl` will pass it automatically?

djhi: No, as with any other components, you'll have to use the `options` props

docs/Inputs.md
djhi: Not sure about that. It might but if you have many items for the referenced resource, it will be unusable. Haven't tested it though

fzaninotto:
It seems that #1304 removed the documentation for `<ReferenceArrayInput>`, could you add it back? Also, you need to remove the BC break mentioned in the UPGRADE guide.

PR #1503, opened 2018-02-07
[RFR] SingleFieldList now accepts a linkType prop

### LINKED ISSUE #740: Allow linkType on Reference*Field
It would be nice if I can specify a linkType on ReferenceManyField or ReferenceArrayField so that when using a SingleFieldList the containing Field would be wrapped with Link.

I could then do something like:

```
<ReferenceManyField label="Comments by" reference="comments" target="post_id" linkType="show">
  <SingleFieldList>
    <ChipField source="author.name" />
  </SingleFieldList>
</ReferenceManyField>
```

Which would make the Chip link to the respective comment.

### LINKED ISSUE #1291: Support resource names containing a slash
**What you were expecting:**
I have a resource `user/comments` and `posts`. In the comment list I have a ReferenceField with refers to which post a comment belongs. So I have it something like:
```jsx
<ReferenceField
  label="Post"
  source="post_id"
  reference="posts"
>
  <TextField source="name" />
</ReferenceField>
```
So the URL of the link I expect it to be `/posts/:id`.

**What happened instead:**
The actual url generated is: `user/posts/:id` which is not expected.

**Environment**

* React version: 15.6.2
* Browser: Any

djhi:
- [x] `SingleFieldList` accepts a `linkType` prop (which defaults to `edit`)
- [x] Ensure we correctly handle resource names with slashes in both `SingleFieldList` and `ReferenceField`
- [x] Documentation

Fixes #740 
Fixes #1291

### DIFF
--- a/packages/react-admin/src/mui/field/ReferenceField.js
+++ b/packages/react-admin/src/mui/field/ReferenceField.js
@@ -67,16 +67,17 @@ export class ReferenceField extends Component {
 
     render() {
         const {
+            allowEmpty,
             basePath,
-            classes,
+            children,
             className,
+            classes,
+            linkType,
             record,
-            source,
             reference,
             referenceRecord,
-            allowEmpty,
-            children,
-            linkType,
+            resource,
+            source,
             translateChoice,
             ...rest
         } = this.props;
@@ -86,36 +87,23 @@ export class ReferenceField extends Component {
         if (!referenceRecord && !allowEmpty) {
             return <LinearProgress />;
         }
-        const rootPath = basePath
-            .split('/')
-            .slice(0, -1)
-            .join('/');
+        const rootPath = basePath.replace(`/${resource}`, '');
+
         const href = linkToRecord(
             `${rootPath}/${reference}`,
             get(record, source)
         );
-        if (linkType === 'edit' || linkType === true) {
+
+        let resourceLinkPath =
+            linkType === 'edit' || linkType === true
+                ? href
+                : linkType === 'show' ? `${href}/show` : false;
+
+        if (linkType === 'edit' || linkType === true || linkType === 'show') {
             return (
                 <Link
                     className={classnames(classes.link, className)}
-                    to={href}
-                    {...sanitizeRestProps(rest)}
-                >
-                    {React.cloneElement(children, {
-                        record: referenceRecord,
-                        resource: reference,
-                        allowEmpty,
-                        basePath,
-                        translateChoice: false,
-                    })}
-                </Link>
-            );
-        }
-        if (linkType === 'show') {
-            return (
-                <Link
-                    className={classnames(classes.link, className)}
-                    to={`${href}/show`}
+                    to={resourceLinkPath}
                     {...sanitizeRestProps(rest)}
                 >
                     {React.cloneElement(children, {
@@ -154,6 +142,7 @@ ReferenceField.propTypes = {
     record: PropTypes.object,
     reference: PropTypes.string.isRequired,
     referenceRecord: PropTypes.object,
+    resource: PropTypes.string,
     source: PropTypes.string.isRequired,
     translateChoice: PropTypes.func,
     linkType: PropTypes.oneOfType([PropTypes.string, PropTypes.bool])
--- a/packages/react-admin/src/mui/field/ReferenceField.js
+++ b/packages/react-admin/src/mui/field/ReferenceField.js
@@ -87,19 +87,12 @@ export class ReferenceField extends Component {
         if (!referenceRecord && !allowEmpty) {
             return <LinearProgress />;
         }
-        const rootPath = basePath.replace(`/${resource}`, '');
+        const rootPath = basePath.replace(resource, reference);
+        const resourceLinkPath = !linkType
+            ? false
+            : linkToRecord(rootPath, get(record, source), linkType);
 
-        const href = linkToRecord(
-            `${rootPath}/${reference}`,
-            get(record, source)
-        );
-
-        let resourceLinkPath =
-            linkType === 'edit' || linkType === true
-                ? href
-                : linkType === 'show' ? `${href}/show` : false;
-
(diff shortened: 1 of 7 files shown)


packages/react-admin/src/mui/field/ReferenceField.js
```diff
@@ -86,36 +87,23 @@ export class ReferenceField extends Component {
         if (!referenceRecord && !allowEmpty) {
             return <LinearProgress />;
         }
-        const rootPath = basePath
-            .split('/')
-            .slice(0, -1)
-            .join('/');
+        const rootPath = basePath.replace(`/${resource}`, '');
```
fzaninotto: Smart! But what if the resource name is `edit` (as in video edit) or `show` (as in trade show)? You should handle this special case

packages/react-admin/src/mui/field/ReferenceField.js
```diff
@@ -86,36 +87,23 @@ export class ReferenceField extends Component {
...
-            );
-        }
-        if (linkType === 'show') {
+
+        let resourceLinkPath =
+            linkType === 'edit' || linkType === true
+                ? href
+                : linkType === 'show' ? `${href}/show` : false;
+
+        if (linkType === 'edit' || linkType === true || linkType === 'show') {
```
fzaninotto: simplify with a `if(resourceLinkPath)`?

packages/react-admin/src/mui/field/ReferenceField.spec.js
```diff
@@ -54,6 +54,23 @@ describe('<ReferenceField />', () => {
         const linkElement = wrapper.find('WithStyles(Link)');
         assert.equal(linkElement.prop('to'), '/bar/123');
     });
+    it('should render a link to the Edit page of the related record even if the resouce contains slashes', () => {
```
fzaninotto: s/resouce/resource/

packages/react-admin/src/mui/list/SingleFieldList.js
```diff
@@ -16,35 +21,106 @@ const styles = {
...
+            ...rest
+        } = this.props;
+
+        return (
+            <div
+                className={classnames(classes.root, className)}
+                {...sanitizeRestProps(rest)}
+            >
+                {ids.map(id => {
+                    const href = linkToRecord(basePath, id);
```
fzaninotto: As it's repeated twice, couldn't you put the edit/show logic inside this `linkToRecord` helper function?

packages/react-admin/src/mui/list/SingleFieldList.js
```diff
@@ -16,35 +21,106 @@ const styles = {
  *         <ChipField source="title" />
  *     </SingleFieldList>
  * </ReferenceManyField>
+ * 
+ * By default, it includes a link to the <Edit> page of the related record
+ * (`/books/:id` in the previous example).
+ *
+ * Set the linkType prop to "show" to link to the <Show> page instead.
```
fzaninotto: You should add that to the `List.md` doc, too.

packages/react-admin/src/mui/field/ReferenceField.js
djhi: I checked that and when in an edit view, basepath is `/${resource}`, not `/${resource}/edit`

djhi: Added tests for those cases

packages/react-admin/src/mui/list/SingleFieldList.js
djhi: Done

packages/react-admin/src/mui/field/ReferenceField.spec.js
```diff
@@ -54,14 +55,65 @@ describe('<ReferenceField />', () => {
...
+    });
+    it('should render a link to the Edit page of the related record when the resource is named edit or show', () => {
+        let wrapper = shallow(
+            <ReferenceField
+                record={{ fooId: 123 }}
+                source="fooId"
+                referenceRecord={{ id: 123, title: 'foo' }}
+                reference="edit"
+                resource="show"
+                basePath="/edit"
```
fzaninotto: So we're in a list of `show` resources but the basepath is `/edit`? I'm not sure that makes sense, it should be `/show`

packages/react-admin/src/mui/field/ReferenceField.spec.js
```diff
@@ -71,14 +123,49 @@ describe('<ReferenceField />', () => {
...
     });
+    it('should render a link to the Show page of the related record when the resource is named edit or show and linkType is show', () => {
+        let wrapper = shallow(
+            <ReferenceField
+                record={{ fooId: 123 }}
+                source="fooId"
+                referenceRecord={{ id: 123, title: 'foo' }}
+                reference="edit"
+                resource="show"
+                basePath="/edit"
```
fzaninotto: same as above, the basepath should be `/show` if the resource is `show`

packages/react-admin/src/mui/list/SingleFieldList.spec.js
```diff
@@ -0,0 +1,163 @@
...
+import ChipField from '../field/ChipField';
+
+describe('<SingleFieldList />', () => {
+    it('should render a link to the Edit page of the related record by default', () => {
+        const wrapper = shallow(
+            <SingleFieldList
+                ids={[1, 2]}
+                data={[{ id: 1, title: 'foo' }, { id: 2, title: 'bar' }]}
+                resource="bar"
+                basePath="/bar"
```
fzaninotto: I don't understand this test case. Is it about a reference many of `bar` to `bar`? I suggest you use `post` and `author` or `comment` to make more sense of it.

djhi: The `SingleFieldList` receive the referenced resource as `resource` and the related `basePath`. It does not have knowledge about the original resource which reference its own resource.

PR #1514, opened 2018-02-09
[RFR] Display error message when the form doesn't validate

### LINKED ISSUE #1513: Display error message when the form doesn't validate
When I click the save button and the form doesn't validate, the errors may not be on screen, so I sometimes wonder why the action didn't succeed. 

We should display a generic message, like "Save failed: validation error", maybe in a Snackbar?

djhi:
Fix #1513

screen

### DIFF
--- a/packages/ra-language-french/index.js
+++ b/packages/ra-language-french/index.js
@@ -50,7 +50,7 @@ module.exports = {
                 "L'URL saisie est incorrecte, ou vous avez suivi un mauvais lien.",
             loading:
                 'La page est en cours de chargement, merci de bien vouloir patienter.',
-            invalid_form: `Le formulaire n'est pas valide. Veuillez vérifier les champs en erreur.`,
+            invalid_form: `Le formulaire n'est pas valide.`,
         },
         navigation: {
             no_results: 'Aucun résultat',
--- a/packages/ra-language-french/index.js
+++ b/packages/ra-language-french/index.js
@@ -50,6 +50,7 @@ module.exports = {
                 "L'URL saisie est incorrecte, ou vous avez suivi un mauvais lien.",
             loading:
                 'La page est en cours de chargement, merci de bien vouloir patienter.',
+            invalid_form: `Le formulaire n'est pas valide. Veuillez vérifier les champs en erreur.`,
         },
         navigation: {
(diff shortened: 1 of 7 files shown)


fzaninotto:
You should use a `warning` type in the notification IMO.

packages/ra-language-french/index.js
```diff
@@ -50,6 +50,7 @@ module.exports = {
                 "L'URL saisie est incorrecte, ou vous avez suivi un mauvais lien.",
             loading:
                 'La page est en cours de chargement, merci de bien vouloir patienter.',
+            invalid_form: `Le formulaire n'est pas valide. Veuillez vérifier les champs en erreur.`,
```
fzaninotto: Isn't it too long?

packages/react-admin/src/mui/button/SaveButton.spec.js
```diff
@@ -5,7 +5,7 @@ import { SaveButton } from './SaveButton';
 
 const translate = label => label;
 
-describe('<SaveButton />', () => {
+describe.only('<SaveButton />', () => {
```
fzaninotto: nooooooooo...

Don't you use Jest?

packages/react-admin/src/mui/button/SaveButton.js
```diff
@@ -37,12 +38,22 @@ const sanitizeRestProps = ({
...
+            redirect,
+            saving,
+            showNotification,
+        } = this.props;
+
+        if (saving) {
             // prevent double submission
             e.preventDefault();
         } else {
+            if (invalid) {
```
fzaninotto: I'm not sure what `invalid` means for redux-form, but can you confirm this notification shows only *after* the user has submitted the form?

packages/react-admin/src/mui/button/SaveButton.spec.js
djhi: oups ^^

packages/react-admin/src/mui/button/SaveButton.js
djhi: I confirm because we only check for an invalid form inside the click handler :)

PR #1521, opened 2018-02-12
[RFR] Bulk actions 

djhi:
- [x] Allow to select items in Datagrid
- [x] Show a bulk actions button when items are selected
- [x] Add a default DELETE_MANY bulk action
- [x] Allow to specify custom bulk actions
- [x] Tests
- [x] Documentation
- [x] Update all dataProviders


tkvw:
Does this mean I can close my PR?

docs/DataProviders.md
```diff
@@ -360,10 +371,25 @@ export default (type, resource, params) => {
             options.method = 'PUT';
             options.body = JSON.stringify(params.data);
             break;
+        case UPDATE_MANY:
+            const query = {
+                filter: JSON.stringify({ id: params.ids }),
+            };
+            url = `${apiUrl}/${resource}?${stringify(query)}`;
+            options.method = 'PATCH';
```
alexisjanvier: Is `PATCH` a better method than `PUT` for an **update many**?

djhi: https://stackoverflow.com/questions/24241893/rest-api-patch-or-put

packages/react-admin/src/mui/list/Datagrid.js
```diff
@@ -76,25 +78,43 @@ class Datagrid extends Component {
...
+            onSelect,
+            onToggleItem,
             ...rest
         } = this.props;
+
         return (
             <Table className={classnames(classes.table, className)} {...rest}>
                 <TableHead>
                     <TableRow className={classes.row}>
+                        <TableCell padding="checkbox">
```
alexisjanvier: Couldn't the display of `<Checkbox>` (and `<DatagridRow>`) be conditioned by the addition or not of the prop `bulkActions` on `<List>` ?

packages/react-admin/src/reducer/admin/resource/list/selectedIds.js
```diff
@@ -0,0 +1,26 @@
+import {
+    SET_LIST_SELECTED_IDS,
+    TOGGLE_LIST_ITEM,
+} from '../../../../actions/listActions';
+
+const initialState = [];
+
+export default (previousState = initialState, action) => {
+    if (action.type === SET_LIST_SELECTED_IDS) {
```
alexisjanvier: Is there a single state for all list selections ?
```
// from listActions.js
export const setListSelectedIds = (resource, ids) => ({
    type: SET_LIST_SELECTED_IDS,
    payload: ids,
    meta: { resource },
});
```

docs/DataProviders.md
alexisjanvier: ok, thanks ;)

packages/react-admin/src/mui/list/Datagrid.js
djhi: No, because we always provide a DefaultBulkActions, see the `List` component

packages/react-admin/src/reducer/admin/resource/list/selectedIds.js
djhi: I don't get the question

djhi: Oh you mean a single one for all resources ?

alexisjanvier: yes :)

djhi: Ok, so no there is a `selectedIds` sub reducer for each resource

docs/List.md
```diff
@@ -109,6 +110,106 @@ export const PostList = (props) => (
...
+);
+
+export const PostList = (props) => (
+    <List {...props} bulkActions={<PostBulkActions />}>
+        ...
+    </List>
+);
+```
+
+A custom bulk action, is a material-ui [`MenuItem`](http://www.material-ui.com/#/components/menu), for which you are entirely responsible. It will receive several props allowing it to perform its work:
```
djhi: I wonder whether we should provide a `BulkAction` component hiding the need to call `onCloseMenu` by the user. Thoughts ?

packages/ra-data-fakerest/src/index.js
```diff
@@ -92,12 +94,22 @@ export default (data, loggingEnabled = false) => {
                         ...params.data,
                     }),
                 };
+            case UPDATE_MANY:
+                params.ids.forEach(id =>
```
fzaninotto: Are you sure that Fakerest supports that?

packages/react-admin/src/mui/list/BulkActions.js
```diff
@@ -0,0 +1,109 @@
...
+    ...rest
+}) => rest;
+
+class BulkActions extends Component {
+    state = {
+        anchorElement: null,
+    };
+
+    handleClick = event => {
+        this.setState({ anchorElement: event.currentTarget });
```
fzaninotto: Storing an element in the state doesn't smell good. It feels like you should use a `ref` here, and store `isOpen` in the state instead.

packages/react-admin/src/mui/list/Datagrid.js
fzaninotto: I agree with @alexisjanvier, we need a way to disable checkboxes on the datagrid altogether.

packages/react-admin/src/mui/list/DatagridRow.js
```diff
@@ -0,0 +1,98 @@
...
+            ...rest
+        } = this.props;
+        return (
+            <TableRow
+                className={className}
+                key={id}
+                style={style}
+                {...sanitizeRestProps(rest)}
+            >
+                <TableCell padding="checkbox">
```
fzaninotto: Same as above, it should be conditional in case there is no `bulkActions`.

packages/react-admin/src/sideEffect/saga/crudResponse.js
```diff
@@ -39,6 +45,17 @@ function* handleResponse({ type, requestPayload, error, payload }) {
                       ),
                   ])
                 : yield [put(showNotification('ra.notification.updated'))];
+        case CRUD_UPDATE_MANY_SUCCESS: {
+            const actions = [put(showNotification('ra.notification.updated'))];
```
fzaninotto: I think we need a custom message here, something like `3 elements deleted`, instead of `'Element deleted`. Same for update down there.

fzaninotto:
@tkvw We discussed over your PR first, and decided that we wanted to implement something simpler, leaving out some of the features you added, so they can be made in userland. We took inspiration from your code for actions and reducers of course. Thanks for submitting it in the first place.

The rest of the discussion will take place in this very PR, so yes, you can close your own.

Feel free to voice your opinion concerning @djhi's proposal.

docs/List.md
fzaninotto: -1 for that, connecting an action creator isn't the end of the world.

packages/react-admin/src/mui/list/Datagrid.js
djhi: Passing `false` to `bulkActions` ?

docs/List.md
djhi: It's not an action creator, just a kind of callback prop to ensure the bulk action menu is closed after triggering the bulk action

packages/ra-data-fakerest/src/index.js
djhi: The e2e tests confirm it

tkvw:
@fzaninotto : Thanks for the feedback.

packages/react-admin/src/reducer/admin/resource/list/selectedIds.js
tkvw: 👍 for `selectedIds` per resource.

packages/react-admin/src/actions/dataActions.js
```diff
@@ -82,6 +102,23 @@ export const crudDelete = (
...
+export const CRUD_DELETE_MANY_LOADING = 'RA/CRUD_DELETE_MANY_LOADING';
+export const CRUD_DELETE_MANY_FAILURE = 'RA/CRUD_DELETE_MANY_FAILURE';
+export const CRUD_DELETE_MANY_SUCCESS = 'RA/CRUD_DELETE_MANY_SUCCESS';
+
+export const crudDeleteMany = (
+    resource,
+    ids,
+    basePath,
+    refresh = true,
+    unselectAll = true
```
tkvw: This shouldn't be configurable, but always true for `Delete`.

packages/react-admin/src/sideEffect/saga/crudResponse.js
tkvw: Also, because this is a batch request, the response can contain success and failed responses.So this should be separated (like I did in my PR). 
I think the failed response should keep selected state (so a retry is easy) and a individual notification should be shown for each failed entity. 
A general message like `3 elements deleted, 2 failed`

packages/react-admin/src/sideEffect/saga/crudResponse.js
```diff
@@ -72,6 +89,17 @@ function* handleResponse({ type, requestPayload, error, payload }) {
                       ),
                   ])
                 : yield [put(showNotification('ra.notification.deleted'))];
+        case CRUD_DELETE_MANY_SUCCESS: {
+            const actions = [put(showNotification('ra.notification.deleted'))];
```
tkvw: Same as above

fzaninotto: I agree, although this is an advanced feature that we can delay for now.

djhi: The response should return the ids of the updated/deleted items. So it should be easy to determine which have failed in a custom saga

packages/react-admin/src/reducer/admin/resource/list/ids.js
```diff
@@ -70,6 +71,19 @@ export default resource => (
 
             return newState;
         }
+        case CRUD_DELETE_MANY_SUCCESS: {
+            const newState = previousState.filter(
```
tkvw: Same issue here, if a individual `DELETE` fails the state would become out of sync. Using the `requestPayload` ids is a bad idea imo. Also, the data state isn't cleared (this can be delayed for later implementation). 
And what about `CRUD_UPDATE_MANY_SUCCESS`? should the state be updated? or are you relying on `refresh`?

djhi: Yes, we rely on refresh. Also there is no individual `DELETE`, we introduced `DELETE_MANY` and `UPDATE_MANY`. Error handling of individual requests if that's how the dataProvider handle those types, should be handled by the dataProvider

tkvw: Ok, so it's considered an atomic operation. It would be nice to add a parameter to the `dataActions` to disable the default `showNotification` on failure. This allows me to implement a custom saga on failure and not showing the notification.

djhi: Agreed but out of this PR scope IMO

Issue #1524, opened 2018-02-14
Layout duplicate when using a Custom Layout


valentinchevalier:
**What you were expecting:**
Use a Custom Layout

**What happened instead:**
The admin layout is duplicate

![admin-on-rest](https://user-images.githubusercontent.com/11924144/36207738-3ac91842-1197-11e8-904e-6848561c1ce9.JPG)

**Steps to reproduce:**
Install admin on rest ([Admin-on-REST Tutorial](https://marmelab.com/admin-on-rest/Tutorial.html))
Follow the [custom layout documentation ](https://marmelab.com/admin-on-rest/Theming.html#using-a-custom-layout)

**Related code:**
The code of the documentation
* CodeSandbox (https://codesandbox.io/s/nklv6w52kj)

**Environment**

* Admin-on-rest version: 1.4.0
* Last version that did not exhibit the issue (if applicable): 1.3.4
* React version: 16.2.0
* Browser: every browsers

djhi:
Sorry about that, we did not update this part of the documentation on `1.4.0`

PR #1526, opened 2018-02-14
[RFR] Fix custom layout documentation

djhi:


fzaninotto:
Fix #1524

ramwaykar:
I have upgraded the admin-on-rest from 1.3.4 to 1.4.0 and then modified the code as you suggested.
AdminRoutes
-                                    customRoutes={customRoutes}
-                                    dashboard={dashboard}
-                                >
-                                    {children}
-                                </AdminRoutes>
But i am getting some notification like this
"<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered<Resource> elements are for configuration only and should not be rendered
["](/home/user/Desktop/area.png)

ramwaykar:
/home/user/Desktop/area.png

PR #1531, opened 2018-02-15
[RFR] Backport decoupling allowEmpty prop from reference selectors

alexisjanvier:
on next


packages/react-admin/src/mui/input/ReferenceDataStatus.js
```diff
@@ -0,0 +1,97 @@
+export const getStatusForInput = ({
```
fzaninotto: I wouldn't name this file with a capital R since it's not a component

PR #1533, opened 2018-02-15
Fix npm warnings

oliviertassinari:
~16.0.0 is a range (tilde) constraint. It means that it will match several versions.
In fact, the current constraint will be satisfied by any version matching >=16.0.0 <16.1.0.
It won't work with the latest version of React 16.2.0. Hence the following warnings in my console:

> warning " > react-admin@2.0.0-beta2" has incorrect peer dependency "react@^15.4.0 || ~16.0.0".
warning " > react-admin@2.0.0-beta2" has incorrect peer dependency "react-dom@^15.4.0 || ~16.0.0".

^16.0.0 is a range (caret) constraint. It means that it will match several versions.
In fact, the current constraint will be satisfied by any version matching >=16.0.0 <17.0.0.

http://jubianchi.github.io/semver-check/

### DIFF
--- a/packages/react-admin/package.json
+++ b/packages/react-admin/package.json
@@ -27,8 +27,8 @@
         "react-test-renderer": "~16.0.0"
     },
     "peerDependencies": {
-        "react": "^15.4.0 || ~16.0.0",
-        "react-dom": "^15.4.0 || ~16.0.0"
+        "react": "^15.4.0 || ^16.0.0",
+        "react-dom": "^15.4.0 || ^16.0.0"
     },
     "dependencies": {
         "autosuggest-highlight": "^3.1.1",


fzaninotto:
Oh, but we *want* to use tildes, because of semver. If a package breaks BC in a major version (like React sometimes does, like material-ui always does), we don't want to have to patch aor in emmergency. 

So in order to support React 16.2 (for which we haven't tested the lib), we should set the dependency on `"react@~16.0.0 || ~16.1.0 || ~16.2.0"` instead.

We prefer to be restrictive.

oliviertassinari:
> We prefer to be restrictive.

@fzaninotto I have never seen them breaking semver. I see that you have been carefully thinking of the problem. My pull-request isn't relevant. Thanks for the feedback.

PR #1538, opened 2018-02-15
[RFR] Refactor resource reducers

tkvw:
There is alot going on in the resource reducers. Function reducers were created while reducing; a bad thing imo. This PR simplifies the resource reducer by using `combineReducer`.


fzaninotto:
You're changing several things. Would you mind explaining the differences before / after?

tkvw:
The resource state reducer consists of three reducers: `data`, `props` and `list`. Currently, for each reducer a new reducer is created while reducing; for example there's a dataReducer with signature `resource=> (state,action)`, and then called like: 
```
    const newState = resources.reduce(		 +        ? { ...previousState, [action.meta.resource]: nextResourceState }
        (acc, resource) => ({		 +        : previousState;
           ...acc,		
            [resource]:		
                action.meta.resource === resource		
                    ? {		
                          props: previousState[resource].props,		
                          data: dataReducer(resource)(		
                              previousState[resource].data,		
                              action		
                          ),		
                          list: listReducer(resource)(		
                              previousState[resource].list,		
                              action		
                          ),		
                      }		
                    : {		
                          props: previousState[resource].props,		
                          data: previousState[resource].data,		
                          list: previousState[resource].list,		
                      },		
        }),		
        {}		
    );
```
So a new function is created each time, this shouldn't happen (or should be avoided in a reducer)

fzaninotto:
Sorry, I still don't understand. 

> So a new function is created each time

Which function are you referring to?

> this shouldn't happen 

Why? Did you notice performance / memory problems?

tkvw:
> Which function are you referring to?

```
dataReducer(resource)(		
     previousState[resource].data,		
     action		
)
```
and 
```
listReducer(resource)(		
    previousState[resource].list,		
    action		
)
```
> Why? Did you notice performance / memory problems?

No, I just prefer the reducers to be as clean and simple as possible.

fzaninotto:
Your PR seems to do much more than that. I fail to see it as a simple refactoring. For me, it adds or removes existing features. Would you mind extracting the refactoring from the other changes, to let us see if it's relevant?

tkvw:
? I fail to see your conclusion. The only thing that's added, is the `props` reducer, this one was obfuscated in the `REGISTER_RESOURCE` reduction, so this is just more understandable. The rest is just refactoring and removing unnecessary code.

fzaninotto:
The following change:

```diff
-export default resource =>
-    combineReducers({
-        ids: ids(resource),
-        params: params(resource),
-        selectedIds,
-        total: total(resource),
-    });
+export default combineReducers({
+    ids,
+    params,
+    selectedIds,
+    total,
+});
```

Seems to change the feature, not be a refactoring.

Same for:

```diff
     if (action.type === UNREGISTER_RESOURCE) {
-        const newState = Object.keys(previousState).reduce((acc, key) => {
-            if (key === action.payload) {
-                return acc;
-            }
-
-            return { ...acc, [key]: previousState[key] };
+        return Object.keys(previousState).reduce((nextState, key) => {
+            if (key !== action.payload) nextState[key] = previousState[key];
+            return nextState;
         }, {});
-        return newState;
     }
```

I must be too dumb or too tired to understand your PR, so I'll look again sometime in the future.

tkvw:
> I must be too dumb or too tired to understand your PR, so I'll look again sometime in the future.

Let's keep it at the latter, because the first is true, that would make my qualification unbearable. It probably has something to do with me describing/ explaining the changes. I'll give it try once I have more time to do so.

tkvw:
First of all let me say that the current implementation works, but imo it's not as clean as it should be.
So what's going on? Whenever a resource action is received in the resource reducer,  a new reducer is created to reduce the state of a specific resource using the `dataReducer` factory and the `listReducer` factory. The `listReducer` factory creates 3 more reducers (ids, params and total). I am arguing that these factories are not needed. A single predefined reducer can be used to reduce the resourcestate. So the outcome is the same but done differently; it's better readable and less code, my definition of code refactoring. 

So;
``` 
-export default resource =>
-    combineReducers({
-        ids: ids(resource),
-        params: params(resource),
-        selectedIds,
-        total: total(resource),
-    });
+export default combineReducers({
+    ids,
+    params,
+    selectedIds,
+    total,
+});
```
does not change the feature, it does the same differently; the PR implementation can now simply call `reducer(state)=>nextState` instead of `reducer(resource)(state)=>nextState`. I can't see the need for the latter, a reducer should do calculation only.

My consideration for the next change is: if you `reduce` an array and spread each iteration to the accumulator, what will happen is that a new Object is created each iteration and shallow copied (babel uses Object.assign). This will impact performance when having alot of resources. I see this being used in the rest of the reducer as well. Object spreading should be done with care. 

```
     if (action.type === UNREGISTER_RESOURCE) {
-        const newState = Object.keys(previousState).reduce((acc, key) => {
-            if (key === action.payload) {
-                return acc;
-            }
-
-            return { ...acc, [key]: previousState[key] };
+        return Object.keys(previousState).reduce((nextState, key) => {
+            if (key !== action.payload) nextState[key] = previousState[key];
+            return nextState;
         }, {});
-        return newState;
     }
```

I hope this explains it a bit more.

tkvw:
Rebased

tkvw:
This PR also ensures that fetches to unregistered resources are still reduced (simply remove the `tags` resource from the `simple` example and the application still runs). This is much safer I think. A user can easily make a mistake by excluding a resource based on permissions although it's required (and authorized) by a view of another resource.

djhi:
> A user can easily make a mistake by excluding a resource based on permissions although it's required (and authorized) by a view of another resource.

IMO we should not allow it. Instead we should indeed throw an error and I believe the error has been made more explicit recently

tkvw:
> IMO we should not allow it

Any specific reason against it? If the user isn't authorized the backend should disallow the request. Imo the only reason (I can think off) to register a resource, it to allow a selector to query available menu options.

fzaninotto:

PR #1542, opened 2018-02-15
[RFR] Customize mui Palette to be less ugly

fzaninotto:
- [x] use secondary for sidebar and app bar
- [x] use primary for active elements (that's the mui convention)
- [x] Find the right palette
- [x] Use icon-only sidebar when minimized
- [x] Make the app bar move with the window (except on mobile, where is is still fixed)
- [x] Fix scrolling and background color glitches

Before:

kapture 2018-02-16 at 14 37 56

After:

kapture 2018-02-16 at 14 32 51


fzaninotto:
OK, the palette is good enough, and I don't want to have a too big PR because the rebase after #1543 will be painful. Switching to RFR.

fzaninotto:
Superseded by #1557

Issue #1393, opened 2018-01-17
Unexpected behavior of Resource


PigoDotRb:
Hi, sorry for my lack of experience: this is my first issue.

**What you were expecting:**
I was trying to create a resource named `posttypes`.
I expect that AOR manages it like a normal resource.

**What happened instead:**
The item in the menu renders correctly but when I click on it, AOR throws an Error:
```
Uncaught TypeError: Cannot read property 'search' of undefined
    at getLocationSearch (List.js:425)
    at index.js:86
    at index.js:36
...
```

**Steps to reproduce:**
Just implement code below.

**Related code:**
```jsx
// App.js
import React from 'react';
import { Admin, Resource } from 'admin-on-rest';

import restClient from './restClient';

import { PostTypeList } from './posttypes';

const App = () => (
  <Admin restClient={restClient}>
    <Resource name="posttypes" list={PostTypeList} />
  </Admin>
);

export default App;


// posttypes.js
import React from 'react';
import { List, Datagrid, TextField } from 'admin-on-rest';

export const PostTypeList = (props) => (
  <List>
    <Datagrid>
      <TextField source="id" />
      <TextField source="slug" />
    </Datagrid>
  </List>
);
```

**Other information:**
`restClient` is just a custom REST client made up from [simple.js](https://github.com/marmelab/admin-on-rest/blob/master/src/rest/simple.js), just changed a few things but the behavior is the same when I pass it to `Admin` or not.

**Environment**

* Admin-on-rest version: `1.3.3`
* React version: `16.2.0`
* Browser: Chromium `63.0.3239.84 Built on Ubuntu , running on Ubuntu 16.04 (64 bit)`
* Stack trace (in case of a JS error):
```
Uncaught TypeError: Cannot read property 'search' of undefined
    at getLocationSearch (List.js:425)
    at index.js:86
    at index.js:36
    at Function.mapStateToProps [as mapToProps] (List.js:438)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at Function.detectFactoryAndVerify (wrapMapToProps.js:52)
    at mapToPropsProxy (wrapMapToProps.js:43)
    at handleFirstCall (selectorFactory.js:26)
    at pureFinalPropsSelector (selectorFactory.js:74)
    at Object.runComponentSelector [as run] (connectAdvanced.js:26)
    at Connect.initSelector (connectAdvanced.js:178)
    at new Connect (connectAdvanced.js:119)
[...]

PR #1545, opened 2018-02-15
[RFR] Better error messages for references of undeclared Resources

Kmaschta:
References to a non-declared Resource is a common error in ng-admin and admin-on-rest (see #1393).
The current error message is pretty misleading.

I improved it a bit with a description of what is happening and how to resolve it.
For example, now if you remove the resource `tags` of the simple example:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ install: package.json ## install dependencies
 run: run-simple
 
 run-simple: ## run the simple example
-	@cd examples/simple && ./node_modules/.bin/webpack-dev-server --hot --inline --config ./webpack.config.js
+	@cd examples/simple && ./node_modules/.bin/webpack-dev-server -d --hot --inline --config ./webpack.config.js
 
 run-tutorial: build ## run the tutorial example
 	@cd examples/tutorial && yarn start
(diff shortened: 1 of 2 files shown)


djhi:
Needs rebase

Makefile
```diff
@@ -9,7 +9,7 @@ install: package.json ## install dependencies
 run: run-simple
 
 run-simple: ## run the simple example
-	@cd examples/simple && ./node_modules/.bin/webpack-dev-server --hot --inline --config ./webpack.config.js
+	@cd examples/simple && ./node_modules/.bin/webpack-dev-server -d --hot --inline --config ./webpack.config.js
```
djhi: Isn't it the default ?

Kmaschta: Nope!

packages/ra-core/src/reducer/admin/references/oneToMany.js
```diff
@@ -20,8 +20,28 @@ export const getIds = (state, relatedTo) =>
 export const getReferences = (state, reference, relatedTo) => {
     const ids = getIds(state, relatedTo);
     if (typeof ids === 'undefined') return undefined;
+
+    if (!state.admin.resources[reference]) {
+        // prettier-ignore
+        console.error(
```
djhi: eslint issue: `no-console`

djhi: > error  Unexpected console statement  no-console

PR #1548, opened 2018-02-15
[RFR] Sanitize AutocompleteInput

tkvw:
I tried to set the `fullWidth` attribute on `AutoCompleteInput` and expected it to work like all other Inputs does, but it didn't work. So I sanitized the component and passed through the props to bring it inline with the other input components.


fzaninotto:
Can you rebase?

tkvw:
Done.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -281,22 +299,13 @@ export class AutocompleteInput extends React.Component {
...
-                margin="normal"
-                className={classnames(classes.root, className)}
-                inputRef={ref}
                 error={!!(touched && error)}
                 helperText={(touched && error) || helperText}
+                className={classnames(classes.root, className)}
                 {...options}
-                InputProps={{
-                    classes: {
-                        input: classes.input,
```
fzaninotto: You're breaking existing code by removing that

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -281,22 +299,13 @@ export class AutocompleteInput extends React.Component {
...
-                    },
-                    ...InputProps,
-                    ...other,
-                }}
+                {...input}
+                {...sanitizeRestProps(rest)}
```
fzaninotto: In the previous code, the rest props are passed to the Input, now they are passed to the TextField.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -281,22 +299,13 @@ export class AutocompleteInput extends React.Component {
...
+                className={classnames(classes.root, className)}
                 {...options}
-                InputProps={{
-                    classes: {
-                        input: classes.input,
-                    },
-                    ...InputProps,
-                    ...other,
-                }}
+                {...input}
```
fzaninotto: why did you add these props?

packages/ra-ui-materialui/src/input/ReferenceInput.js
```diff
@@ -54,6 +54,7 @@ export const ReferenceInputView = ({
     classes,
     className,
     error,
+    fullWidth,
```
fzaninotto: This is exactly the type of props that we want to pass with `rest` instead of identifying them individually here.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
tkvw: I brought the component inline with `TextInput`, which is better designed. Same applies for this component.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
tkvw: Correct, imo previous was wrong.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
tkvw: The `classes` prop is still passed through, but `classes.input` doesn't make sense (there's no `classes.input` in the styles of `AutocompleteInput`). Not sure why it was there in the first place.

packages/ra-ui-materialui/src/input/ReferenceInput.js
tkvw: I did this, because of the `rest` props were still being passed to `Labeled`, this is no long the case so this can be changed.

tkvw:
So what I did was to make the `TextField` generation inline with `TextInput` and added a `propsSanitizer`.

fzaninotto:

PR #1549, opened 2018-02-15
[RFR]: SelectArrayInput render child instead of Chip

tkvw:
The simple example shows : 
```
<ReferenceArrayInput reference="tags" source="tags">
    <SelectArrayInput>
        <ChipField source="name" />
    </SelectArrayInput>
</ReferenceArrayInput>
```
This make me think that the `SelectArrayInput` uses `ChipField` to render, but it doesn't. Currently it renders a `<Chip >` field. This PR changes this to clone the child element instead.

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.js
@@ -6,7 +6,6 @@ import { MenuItem } from 'material-ui/Menu';
 import Input, { InputLabel } from 'material-ui/Input';
 import { FormControl, FormHelperText } from 'material-ui/Form';
 import { withStyles } from 'material-ui/styles';
-import Chip from 'material-ui/Chip';
 import compose from 'recompose/compose';
 import classnames from 'classnames';
 import { addField, translate, FieldTitle } from 'ra-core';
@@ -156,6 +155,7 @@ export class SelectArrayInput extends Component {
 
     render() {
         const {
+            children,
             choices,
             classes,
             className,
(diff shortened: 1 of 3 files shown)


fzaninotto:
Can you rebase?

tkvw:
Done.

packages/ra-ui-materialui/src/input/SelectArrayInput.js
```diff
@@ -201,13 +201,13 @@ export class SelectArrayInput extends Component {
...
-                                        key={get(choice, optionValue)}
-                                        label={get(choice, optionText)}
-                                        className={classes.chip}
-                                    />
-                                ))}
+                                .map(choice =>
+                                    React.cloneElement(children, {
+                                        key: get(choice, optionValue),
+                                        record: choice,
+                                        className: classes.chip,
```
fzaninotto: I think we need to pass more props to the child. The `ChipField`, for instance, expects a `source` in addition to the `record`.

Please add a unit test to prove that your change works.

tkvw: No, the `source` prop is setup in the component, so already available. For example: 
```
<SelectArrayInput>
    <ChipField source="name"/>
</SelectArrayInput>
```

tkvw: Test added

fzaninotto: you're right, my bad

fzaninotto:
Thanks!

PR #1557, opened 2018-02-16
[RFR] Improve colors, sidebar, and layout

fzaninotto:
- [x] use secondary for sidebar and app bar
- [x] use primary for active elements (that's the mui convention)
- [x] Find the right palette
- [x] Use icon-only sidebar when minimized
- [x] Make the app bar move with the window (except on mobile, where is is still fixed)
- [x] Fix scrolling and background color glitches

Before:

kapture 2018-02-16 at 14 37 56

After:

kapture 2018-02-16 at 14 32 51

Supersedes #1542

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Logout.js
+++ b/packages/ra-ui-materialui/src/auth/Logout.js
@@ -12,7 +12,7 @@ import Responsive from '../layout/Responsive';
 import { translate, userLogout as userLogoutAction } from 'ra-core';
 
 const styles = {
-    iconPaddingStyle: { paddingRight: '0.5em' },
+    iconPaddingStyle: { paddingRight: '1.2em' },
 };
 
 const sanitizeRestProps = ({
--- a/packages/ra-ui-materialui/src/button/DeleteButton.js
+++ b/packages/ra-ui-materialui/src/button/DeleteButton.js
@@ -2,21 +2,39 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ActionDelete from 'material-ui-icons/Delete';
 import { linkToRecord } from 'ra-core';
+import { withStyles } from 'material-ui/styles';
+import { fade } from 'material-ui/styles/colorManipulator';
+import classnames from 'classnames';
 
-import Link from '../Link';
+import { Link } from 'react-router-dom';
 import Button from './Button';
+
+const styles = theme => ({
+    deleteButton: {
+        color: theme.palette.error.main,
+        '&:hover': {
+            backgroundColor: fade(theme.palette.error.main, 0.12),
+            // Reset on mouse devices
+            '@media (hover: none)': {
+                backgroundColor: 'transparent',
+            },
+        },
+    },
+});
 
 const DeleteButton = ({
     basePath = '',
     label = 'ra.action.delete',
     record = {},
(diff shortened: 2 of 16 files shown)


afilp:
Thanks for these changes!

Can you release a new BETA version, so that I we can immediately incorporate to our upgraded apps? 

(P.S. Could you release BETA minor versions very often, for example every 3 days, so that we are able to incorporate all changes to our apps and provide valuable feedback to you? I am on Windows and have difficulties to "make" the latest lib from the latest "next" branch, I would like to have these in 'npm', of possible)

fzaninotto:
We've changed so much of the codebase lately that we need time to do more tests and check for regressions in features and documentation. That is not compatible with a regular release schedule, unless we want to annoy (I'm being polite) our users - which we don't.

If you want to be on the cutting edge, and if you are ready to live with the consequences, please learn to do the build on your own.

afilp:
True... 

What if you had an **ALPHA** version line? That could be updated almost daily? 

This way we could use and test the latest changes on-the-fly, without expecting proper documentation or tests. We would know that it is an "AS-IS" solution (still we could find bugs and report them).

And if we would like to "lock" our app in a specific Alpha version we could do this in `package.json`.

fzaninotto:
The `next` branch is the alpha version.

afilp:
@fzaninotto I tried to create a ``lib`` with the latest ``src`` on the ``next`` branch (based on your suggestion to use the `make` logic).

What do I do wrong? There is no `lib` created like the one that we can see in the `npm` package. In other words, do you have somewhere documented the exact steps to get the same package like the one that you host in the `npm` (but with the latest commits included)?

afilp:
Thanks! 

Can you please document somewhere the process to `make` the package ourselves and use it with our app? This is very important as we expect to be using the latest possible fixes on the `next` branch. We are using `next` branch in production, so this is very important.

Thanks again!

fzaninotto:
> We are using `next` branch in production

Don't. This is risky, we don't encourage this.

As for building the packages yourself, I guess it implies a combination of `make build` and `npm link`, but you'll have to figure that out by yourself.

afilp:
1. We thought to be using it because it says "BETA", not ALPHA, and we would not like to invest time in an old version (current stable one) which has so much differences with the one that is coming 'next'. But the main reason is that the old version is "feature-fixes" and has several limitations that when we report them we get (reasonably) an answer similar to this: "This is a new feature so it goes to 'next'". And then it is implemented in 'next'. So, you understand, it is a  "catch 22" problem if we we want to have a slightly better functionality (not just bug fixes, but slightly better/proper functionality).

2. About the build, I assume that you have a script that already does this? Can't this be documented?

Thanks again, I really appreciate your work. You seem to have a great team there in France!

fzaninotto:
You don't need to `make` your packages since I just release the beta 3 to npm. Just update your `package.json`.

About the build, we use `make build`, then `lerna publish`. End of it. If you don't feel comfortable using `npm link` (read the docs!), you should fork our repo, change the package name to something only you will use, and publish when you want using `lerna publish`.

I'm closing this conversation as you keep on asking for our help on building an intermediate version, and I keep on telling you we don't support this scenario. This isn't going anywhere.

PR #1563, opened 2018-02-16
[RFR] Example: create button for referenced resource

### LINKED ISSUE #1402: Example: How to add a create button with a ReferenceManyField
**What you were expecting:**

https://marmelab.com/admin-on-rest/Fields.html#referencemanyfield

 On the documentation I have the option to show a list of related fields using a ```<ReferenceManyField>``` which is great, and it works. But following the same documentation, wanting to hide the resource from the sidebar menu

> Note: You must add a ```<Resource>``` for the reference resource - admin-on-rest needs it to fetch the reference data. You can omit the list prop in this reference if you want to hide it in the sidebar menu.)

I'm also blocking my way from creating new rows on that resource, because there is no "Create" button when showing a list of related rows nested inside another resource (and hiding it from the sidebar menu)

**What happened instead:**
There is no create button, so there is no way to create rows on a nested resource.
reference-many-field-datagrid

I saw the pull request #744 but this is different since this is related to ```<ReferenceManyField>``` not ```<ReferenceArrayInput>```

**Steps to reproduce:**
Same snippet of code on the documentation:
```js
import React from 'react';
import { Edit, Datagrid, SimpleForm, DisabledInput, DateField, EditButton, ReferenceManyField, TextField, TextInput } from 'admin-on-rest';

    <Edit {...props}>
        <SimpleForm>
            <DisabledInput label="Id" source="id" />
            <TextInput source="title" />
            <ReferenceManyField label="Comments" reference="comments" target="post_id">
                <Datagrid>
                    <TextField source="body" />
                    <DateField source="created_at" />
                    <EditButton />
                </Datagrid>
            </ReferenceManyField>
        </SimpleForm>
    </Edit>
);
```

**Environment**

* Admin-on-rest version: 1.3.3
* Last version that did not exhibit the issue (if applicable): n/a
* React version: latest
* Browser: any

djhi:
Fix #1402


fzaninotto:
-1 for adding a custom example for that, because:

- we can add it to the simple demo
- every new example we add, we'll need to maintain. As there are many similarities between demos, this adds to our maintenance work
- the place for a solution to a custom problem like that would be StackOverflow / codesandbox.

Can you add it to the simple demo instead?

djhi:
Actually, I was wondering if we should add an `Examples` or `Recipes` section in the documentation with links to codesandboxes instead of having multiple example apps directly in our repository like next.js

fzaninotto:
I think you're reinventing StackOverflow ;)

djhi:
That was the whole point of the `examples` section actually as explained when we introduced it: https://github.com/marmelab/admin-on-rest/pull/1368

I do agree that keeping many examples in our repository would make it too heavy though.

fzaninotto:
I doubt that a person with a specific question about prepopulated create forms will look in the Examples section of the documentation. My bet is that, most of the time, they'll google their question. In that case, they'll probably end up in Stack Overflow.

So for discoverability, StakOverflow is already better than us. Additionally, we could drop a few grains here and there in the documentation, where users may find them. 

Sharing snippets and examples is a good idea. Implementing an entire blog admin clone for that is too much work. So IMO, a codesandbox link or a stack overflow snippet is the right format.

PR #1570, opened 2018-02-19
[RFR] Documentation - Examples section

djhi:
Add an `Examples` section to the documentation with a first example showing how to add a Create button linked to a ReferenceManyField

### DIFF
--- a/docs/Examples.md
+++ b/docs/Examples.md
@@ -1,0 +1,19 @@
+---
+layout: default
+title: "Examples"
+---
+
+# Examples
+
+## Adding a Create button linked to a ReferenceManyField
+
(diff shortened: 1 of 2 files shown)


fzaninotto:
IMO, There are enough sections in the documentation already. Use either the FAQ, or add a section in the `CreateEdit` chapter.

djhi:
I disagree. IMI, FAQ should be for troubleshooting. Besides, I really think we should have a dedicated section for examples and, eventually, add links to it inside pure documentation sections such as `CreateEdit`.

fzaninotto:
Better : implement this in one of the existing examples (e.g. create comment for post), and add a link to the code in CreateEdit.

djhi:
My idea for the examples was to always start from the posts/comment base and show a specific recipe in addition.

But ok. I fear that we might end up with a very complicated simple example though, where someone won't be able to find answer to a simple question.

PR #1577, opened 2018-02-20
[RFR] Document how to pass variables to notifications

### LINKED ISSUE #1351: Add interpolation support on showNotification()
Currently the Notification component just passes the translated message to the Snackbar.
```js
class Notification extends React.Component {
    handleRequestClose = () => {
        this.props.hideNotification();
    };

    render() {
        const style = {};
        const { primary1Color, accent1Color } = getStyles(this.context);
        const { type, translate, message } = this.props;
        if (type === 'warning') {
            style.backgroundColor = accent1Color;
        }
        if (type === 'confirm') {
            style.backgroundColor = primary1Color;
        }
        return (
            <Snackbar
                open={!!message}
                message={!!message && translate(message)}
                autoHideDuration={4000}
                onRequestClose={this.handleRequestClose}
                bodyStyle={style}
            />
        );
    }
}
```
However, if someone wants to interpolate a string into the Notification like so:
```js
showNotification('myroot.hello.world', {name: 'Planet Earth'});
```
It won't work, because the second argument of the `showNotificationAction` should be the type of notification. 

So I propose an additional optional parameter like so: 
```js
showNotification('myroot.hello.world', 'info', {name: 'Planet Earth'})
```

djhi:
Fixes #1351

### DIFF
--- a/docs/Translation.md
+++ b/docs/Translation.md
@@ -365,3 +365,25 @@ translate('not_yet_translated', { _: 'Default translation' })
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
 
+## Variables with notifications
+
+It is possible to pass variables for polyglot interpolation with custom notifications. For example:
+
+```js
+showNotification('myroot.hello.world', 'info', { messageArgs: { name: 'Planet Earth' } });
+```
+
+Assuming you have the following in your custom messages:
+
+```js
+// in src/App.js
+const messages = {
+    en: {
+        myroot: {
+            hello: {
+                world: 'Hello, %{name}!',
+            },
+        },
+    },
+};
+```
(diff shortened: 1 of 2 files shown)


docs/Translation.md
```diff
@@ -365,3 +365,25 @@ translate('not_yet_translated', { _: 'Default translation' })
 
 To find more detailed examples, please refer to [http://airbnb.io/polyglot.js/](http://airbnb.io/polyglot.js/)
 
+## Variables with notifications
```
fzaninotto: s/Variables with notifications/Notifications With Variables/

docs/_layouts/default.html
```diff
@@ -670,6 +670,9 @@
                             <li class="chapter">
                                 <a href="#using-specific-polyglot-features">Using Specific Polyglot Features</a>
                             </li>
+                            <li class="chapter">
+                                <a href="#variables-with-notifications">Variables with notifications</a>
```
fzaninotto: s/variables-with-notifications/notifications-with-variables/
s/Variables with notifications/Notifications With Variables/

PR #1580, opened 2018-02-21
[RFR] Fix first warnings about missing proptypes

fzaninotto:
`make lint` currently outputs 172 warnings, mostly about missing prop types. They pollute the build STDOUT and obscure more interesting outputs. This PR fixes all these warnings.

- [x] Fix warnings in `make lint`
- [x] Fix warnings in `make test-unit`


fzaninotto:
Back to WIP, I'll fix the warnings in unit tests, too

PR #1583, opened 2018-02-21
[RFR] Use material ui features

fzaninotto:
Since we implemented the components with early versions of material-ui 1.0, their have implemented features that we also implemented on our side.

Let's catch up by using their features instead of ours. Benefits: less code to maintain, better alignment of header and content in datagrid.

Before:

image

After:

image

### DIFF
--- a/examples/simple/users.js
+++ b/examples/simple/users.js
@@ -24,12 +24,25 @@ import {
     required,
     translate,
 } from 'react-admin'; // eslint-disable-line import/no-unresolved
+import { InputAdornment } from 'material-ui/Input';
 import PeopleIcon from 'material-ui-icons/People';
+import SearchIcon from 'material-ui-icons/Search';
 export const UserIcon = PeopleIcon;
 
 const UserFilter = ({ permissions, ...props }) => (
     <Filter {...props}>
-        <TextInput label="user.list.search" source="q" alwaysOn />
+        <TextInput
+            label="user.list.search"
+            source="q"
+            alwaysOn
+            InputProps={{
+                endAdornment: (
+                    <InputAdornment position="end">
+                        <SearchIcon color="disabled" />
+                    </InputAdornment>
+                ),
+            }}
+        />
         <TextInput source="name" />
         {permissions === 'admin' ? <TextInput source="role" /> : null}
     </Filter>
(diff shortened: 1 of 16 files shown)


packages/ra-ui-materialui/src/list/BulkActions.js
```diff
@@ -101,26 +102,32 @@ class BulkActions extends Component {
                 >
                     <MoreVertIcon />
                 </Button>
-                <Menu
+                <Popover
```
fzaninotto: The popover used to appear on top of the bulk actions button

packages/ra-ui-materialui/src/list/DatagridRow.js
```diff
@@ -47,6 +48,7 @@ class DatagridRow extends Component {
                 className={className}
                 key={id}
                 style={style}
+                hover={hover}
```
fzaninotto: they implemented row hover!

packages/ra-ui-materialui/src/list/BulkActions.js
djhi: As Menu passes its props down to Popover, isn't this bug fixed just with the `anchorOrigin` prop ?

fzaninotto: probably. I just used the same markup as for the Filters dropdown, to make them consistent.

But now I see that the `Poppovers` demo was removed from the mui doc, so I'm wondering if your way isn't better, and we should change the Filters dropdown instead... What do you think?

djhi: I really don't know. I think the Popover is a low level component though. Maybe we should ask on the mui repository or ping Olivier ?

fzaninotto: I discovered 2 things:

- `Menu` uses `Popover` and `MenuItem` internally. To avoid repetition, it's better that we use `Menu`
- Material design explains that menus *should* open above the button, not under. The rationale is that the finger should not have to move to select the default menu item. So I removed the position setting in both menus (bulk actions, filters) to have them show up as material design says.

examples/simple/users.js
```diff
@@ -50,7 +63,7 @@ export const UserList = ({ permissions, ...props }) => (
                 />
             }
             medium={
-                <Datagrid>
+                <Datagrid hover={false}>
```
djhi: Why disabling it ?

PR #1584, opened 2018-02-21
Enable override ImageField <img> styles

manelpb:


fzaninotto:
We won't add new features to master. Please open a PR on `next` instead (and by the way, I think that in `next`, this is already possible out of the box).

brownieboy:
Was this ever implemented?  If so, any documentation on how to do it?

I'd like to set the borderRadius style on the image, for example.  At present, this style gets set on the parent Div, which has no affect on the actual image.

PR #1587, opened 2018-02-22
Remove unused code in Admin.js

afilp:


djhi:
It's not unused. It provides the defaults

djhi:
Think about the `react-admin` package as a distribution. It provides a default admin (`ra-core`) using material UI (`ra-ui-materialui`)

afilp:
@djhi Sorry about that, I was struggling to find the reason why my custom menu does not respect the 'WithPermissions' logic, and looking everywhere in the code. I just reported this issue:
https://github.com/marmelab/admin-on-rest/issues/1586

(sorry for no sandbox, not sure how to do this with login/logout logic)

PR #1589, opened 2018-02-23
[RFR] Add "Undo" Feature on Save and Delete

fzaninotto:
Follows #1588

kapture 2018-02-23 at 15 04 38

kapture 2018-02-23 at 17 06 59

## Problem

Optimistic updates introduce a usability problem: when the user deletes a record, she is redirected to the datagrid where she sees her record deleted (good), then the deletion is sent to the server, and only when the server returns, she sees the notification about the deletion. In practice, that means a delay of 500ms to several seconds between the redirection and the notification. It's bad.

Another issue is the confirmation dialog: When the user clicks on the Delete button, 80% of the time, she *wants* to delete the record, so the dialog is a burden. The UX should optimize the 80% and make the workflow more complex for the other 20%; not the other way around.

## Solution: Undo

Since we are doing optimistic updates, we could go all in, and give the notification right away, even before the server has been notified of the change.

In that notificaiton, we could display an "undo" link that would replace the Dialog.

## Implementation

The actual server request is sent *after* the notification times out. If the user clicks on "undo", the current view should be refreshed to get updated data from the server.

## To Do

- [x] Add Undo actions
- [x] Translate "Undo"
- [x] Remove opacity change on body when loading. We don't want to emphasize the "loading" anymore, since we consider that the client data is the source of truth in general
- [x] Implement Undo Delete
- [x] Implement Undo Delete Many
- [x] Implement Undo Edit
- [x] Implement Undo Edit Many
- [x] Make most side effects (redirection, notification, refresh) standalone saga watching specific metas, to allow users to add actions with these types of side effects, and let undoable actions be a simple wrapper around normal actions
- [x] rename sagas for consistency
- [x] Make all these actions work either in undo of confirm mode (!)
- [ ] ~~document how to enable / disable undo actions~~ (I'm not sure we want to expose the non-undoable versions. These are internal APIs for test purpose)


packages/ra-ui-materialui/src/button/DeleteButton.js
```diff
@@ -26,44 +30,38 @@ const styles = theme => ({
 });
 
 class DeleteButton extends Component {
-    state = { dialogOpen: false };
```
fzaninotto: We need two buttons actually: one with the Confirm dialog, and one with the undo snackbar. Scratching my head about how to achieve that.

djhi:
This is very promising !
> Make all these actions work either in undo of confirm mode (!)
document how to enable / disable undo actions

Does this mean we'll support both mode: confirmable or cancellable actions ?

packages/ra-core/src/sideEffect/saga/cancelSaga.js
```diff
@@ -0,0 +1,119 @@
...
+            if (action.payload.unselectAll) {
+                actions.push(put(setListSelectedIds(action.meta.resource, [])));
+            }
+            yield all(actions);
+            break;
+        }
+    }
+    yield put(startOptimisticMode());
+    // dispatch action in optimistic mode (no fetch)
+    yield put(action);
```
fzaninotto: I realize that instead of dispatching the same action twice (which has the effect to update the store twice), I'd better dispatch the action with an `_OPTIMISTIC` suffix. That's the type of action that the `ids` and `data` reducers should watch (not the ones dispatching the fetch call). 

That would probably ease the implementation of the Confirm alternatives

packages/ra-core/src/sideEffect/saga/cancelSaga.js
```diff
@@ -0,0 +1,119 @@
...
+                    showNotification('ra.notification.updated', 'info', {
+                        messageArgs: {
+                            smart_count: action.payload.ids.length,
+                        },
+                        cancellable: true,
+                    })
+                ),
+            ];
+            if (action.payload.unselectAll) {
+                actions.push(put(setListSelectedIds(action.meta.resource, [])));
```
fzaninotto: since these are synchronous and not side effects ad all, they should be move to a reducer

packages/ra-core/src/sideEffect/saga/cancelSaga.js
```diff
@@ -0,0 +1,119 @@
...
+    stopOptimisticMode,
+} from '../../actions/cancelActions';
+import { refreshView } from '../../actions/uiActions';
+import resolveRedirectTo from '../../util/resolveRedirectTo';
+import { setListSelectedIds } from '../../actions/listActions';
+
+function* handleCancelRace(cancellableAction) {
+    const { payload: { action, delay: cancelDelay } } = cancellableAction;
+    // display cancellable notification
+    switch (action.type) {
```
fzaninotto: instead of doing a check on the action type, the notification should be in the action. That would make the saga very generic (and reusable for user actions).

fzaninotto:
> Does this mean we'll support both mode: confirmable or cancellable actions ?

We will if it's not too much work. Otherwise, I'll remove confirmable actions without regret.

packages/ra-core/src/actions/cancelActions.js
```diff
@@ -0,0 +1,21 @@
+export const CANCELLABLE = 'RA/CANCELLABLE';
```
fzaninotto: I wonder if we should use the term "Undoable" instead of "Cancelable"?

docs/List.md
```diff
@@ -211,20 +210,37 @@ class ResetViewsAction extends Component {
...
+
+    render() {
+        return null;
+    }
+}
+
+export default connect(undefined, { startUndoable })(ResetViewsAction);
+```
+
+Note that the `crudUpdateMany` action creator is *not* present in the `mapDispatchToProps` arcgument of `connect()` in that case. Only `startUndoable` needs to be dispatched in this case, using the result of the `crudUpdateMany()` call as parameter.
```
djhi: s/arcgument/argument/

packages/ra-core/src/sideEffect/notification.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Notification Side Effects
+ */
+function* handleNotification({ type, meta: { notification } }) {
+    const { body, level = 'info', messageArgs = {} } = notification;
+    yield put(
+        showNotification(body, level, {
+            messageArgs,
+            undoable: type.indexOf('_OPTIMISTIC') !== -1,
```
djhi: Not fan of the optimistic case detection mechanism. Can't we use a `optimistic` property in the metas instead ?

fzaninotto:
Review applied

packages/ra-core/src/reducer/admin/resource/list/ids.js
```diff
@@ -61,25 +60,19 @@ export default resource => (previousState = [], { type, payload, meta }) => {
                 ...previousState.slice(index + 1),
             ];
 
-            Object.defineProperty(
-                newState,
-                'fetchedAt',
-                previousState.fetchedAt
-            );
+            Object.defineProperty(newState, 'fetchedAt', {
```
jpetitcolas: Why using `defineProperty` instead of `newState.fetchedAt`?

fzaninotto: to make a non-enumerable propery

tkvw:
This is so cool, the only thing I don't like is that you require to set `undoable={false}` on `Edit` if you want to opt-out. I rather see `truethy` props. So if the prop is `undoable` it should be `false` by default. So imo `disableUndoable` prop is better for `Edit` .

fzaninotto:
I'm glad you like it!

As for the ability to opt-out, I'm still wondering if we should leave the choice to the user. I designed the `UNDOABLE` action and saga to work regardless of the enclosed action, and I needed to test the same action in undoable and non-undoable for that. But I don't want to maintain 2 UX scenarios in the long run. So the `undoable` prop, which is not documented, is only here for debugging purposes for now. Consider it a private method!

If we keep it, I agree with you that it should be renamed as you suggest. Would you mind opening a PR for that?

tkvw:
My use case is to allow server-side validation on a form. So I like to opt-out if I have a form which requires this. I'll create a PR for renaming the prop.

normanzb:
It is a bit off topic, but calling this feature "undoable" is misleading. 
This feature really has nothing todo with "undo", the data hasn't been sent to server at the time user click the undo button. So if you haven't done anything yet, how can you undo it?
IMO the true "undoable" should commit the changes immediately to server and keep a record of every changes, when "undo" is clicked, the record in client memory is used to recover the original version of data.
I think this feature should be called something like "regrettable action"

PR #1597, opened 2018-02-28
Add onBeforeAdd to SelectArrayInput

ronfogel:
Added new function - onBeforeAdd to SelectArrayInput. this was needed to be able to transfrom the new value to a different value based on some conditions and even to stop adding a new  value based on some conditions

Real usecase: choices are provided by server side, if user add new value with same letters but different letter case ("FrEe" against "free") it added a new value which we didnt want. to prevent that I added the onAddBefore so we could transform the "FrEe" to "free". also I was able to prevent duplication of same tag more than once.

### DIFF
--- a/src/mui/input/SelectArrayInput.js
+++ b/src/mui/input/SelectArrayInput.js
@@ -87,6 +87,8 @@ export class SelectArrayInput extends Component {
     };
 
     handleAdd = newValue => {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Why can't you use [redux-form's `parse`](https://redux-form.com/7.3.0/docs/api/field.md/#-code-parse-value-name-gt-parsedvalue-code-optional-), which you can specify as a custom prop on every input in admin-on-rest?

PR #1598, opened 2018-02-28
[RFR] Datagrid style

tkvw:
Allow stylesheets to override column style per resource and add a static width to select column in datagrid.

### DIFF
--- a/e2e/pages/ListPage.js
+++ b/e2e/pages/ListPage.js
@@ -28,9 +28,9 @@ export default url => driver => ({
         deleteBulkActionsButtonMenuItem: By.css(
             '.bulk-actions-menu-item:last-child'
         ),
-        selectAll: By.css('.select-all'),
-        selectedItem: By.css('.select-item input:checked'),
-        selectItem: By.css('.select-item input'),
+        selectAll: By.css('.select-all input'),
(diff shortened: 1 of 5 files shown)


fzaninotto:
Lacks documentation.

Also, I had forgotten about the ability to override styles using plain CSS. In other terms, it's not tested, and it may break in various other places.

tkvw:
Not sure what you're expecting from me? Should I add docs, or don't you want to support external css anymore?

fzaninotto:
What does your change allow that a user can't do with `withStyles`?

If you're adding a way to override the styles of a cell with CSS instead of JSS, you must explain it in the `Theming` and `List` documentation.

fzaninotto:
bump

tkvw:
I can add docs for this, but only if you want to support custom styling through external css. I've thought about it a bit more and think your comment about `withStyles` makes sense. This is just a dirty quick way of applying style. So let me know what to do or you can close this topic.

fzaninotto:
Somehow related to #1694, I think we should use only one style customization path, and that should be `withStyles`. But, to allow developers to override inner components, we should pass specific fields of the `classes` prop to children component. 

Anyway, that invalidates the plain CSS approach, and I'm going to close this issue. Thanks for suggesting the merge, though.

PR #1599, opened 2018-02-28
[RFR]: Allow title to be an element

tkvw:
Changed some proptypes and use `cloneElement` when title is an element. This allows the element to use the correct `theme` (themeprovider is initialized in the layout).

### DIFF
--- a/packages/ra-core/src/CoreAdminRouter.js
+++ b/packages/ra-core/src/CoreAdminRouter.js
@@ -171,7 +171,7 @@ CoreAdminRouter.propTypes = {
     logout: PropTypes.node,
     menu: componentPropType,
     theme: PropTypes.object,
-    title: PropTypes.string,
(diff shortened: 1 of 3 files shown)


fzaninotto:
Can you explain the use case a bit more?

tkvw:
To allow the title to be a element (which was allowed before?), for example I want the title to render a Link to '/'. But because title is supplied to `<Admin>` it's created outside scope of the `ThemeProvider` so a `cloneElement` should be used to get access to the `theme` object in `withStyles`. 

Title: 
```
const styles = theme => ({
  root: theme.palette.primary.contrastText
})
const Title = ...
export default withStyles(styles)(Title)
```

jpetitcolas:
@fzaninotto: we also encountered the same issue: we want the title to be a logo, hence an `` tag.

fzaninotto:
Fine, thanks!

PR #1601, opened 2018-03-01
[RFR]: Layout reuse enhanced layout (another proposal)

JacquesBonet:
Another proposal of change theme and style composition (Can't change style with tkvw solution)

* Usage

```
import { defaultLayoutWithThemeAndStyle } from 'ra-ui-materialui/lib/layout/Layout'
import theme from './theme'

//copy past style from ra-ui-materialui/src/layout/Layout.js and modify it
const styles = theme => ({
    root: {
        ....
    },
    content: {
       ...
    },
});

export default defaultLayoutWithThemeAndStyle(theme, styles)
```

* Its also possible to specify a new layout

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Layout.js
+++ b/packages/ra-ui-materialui/src/layout/Layout.js
@@ -116,10 +116,10 @@ export const layoutWithTheme = PlainLayout => {
         </MuiThemeProvider>
     );
     LayoutWithTheme.propTypes = {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Please explain the problem, what users have to do before the change, and how you changes makes thinks easier.

fzaninotto:
Please contribute to #1600 instead of creating a new Pr on the same subject

PR #1606, opened 2018-03-03
Update UPGRADE.md: Use "function as child"

afilp:
Type "function as child" expression, instead of "child as a function".


fzaninotto:
Thanks!

PR #1614, opened 2018-03-06
[RFR] Move `setFilters` debounce to core

jpetitcolas:
Currently, `setFilters` is debounced only in Material UI package. As a consequence, when implementing a custom theme without the `ra-ui-materialui` package, we also need to add manually a debounce. Otherwise, editing a text input filter would spam API (see tests for a real world scenario).

As I can't see any scenario where we would need to spam API (causing network slowness), let's move it to the core package.

### DIFF
--- a/packages/ra-core/src/controller/ListController.js
+++ b/packages/ra-core/src/controller/ListController.js
@@ -88,6 +88,10 @@ export class ListController extends Component {
         }
     }
 
+    componentWillUnmount() {
+        this.setFilters.cancel();
+    }
+
     componentWillReceiveProps(nextProps) {
         if (
             nextProps.resource !== this.props.resource ||
--- a/packages/ra-core/src/controller/ListController.js
+++ b/packages/ra-core/src/controller/ListController.js
@@ -7,8 +7,11 @@ import { push as pushAction } from 'react-router-redux';
 import compose from 'recompose/compose';
 import { createSelector } from 'reselect';
 import inflection from 'inflection';
+import debounce from 'lodash.debounce';
+import isEqual from 'lodash.isequal';
 import pickBy from 'lodash.pickby';
 
+import removeEmpty from '../util/removeEmpty';
 import queryReducer, {
     SET_SORT,
     SET_PAGE,
@@ -71,7 +74,7 @@ export class ListController extends Component {
     componentDidMount() {
         if (
             !this.props.query.page &&
-            !this.props.ids.length &&
+            !(this.props.ids || []).length &&
             this.props.params.page > 1 &&
             this.props.total > 0
         ) {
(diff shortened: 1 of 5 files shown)


djhi:
You just need to fix the eslint errors :)

packages/ra-core/src/controller/ListController.js
```diff
@@ -166,8 +169,15 @@ export class ListController extends Component {
 
     setPage = page => this.changeParams({ type: SET_PAGE, payload: page });
 
-    setFilters = filters =>
-        this.changeParams({ type: SET_FILTER, payload: filters });
+    setFilters = debounce(filters => {
+        if (isEqual(filters, this.props.filterValues)) {
```
fzaninotto: I'm not sure about this check. In the previous version, we kept a local version of the `filters`, and now you use `filterValues`

Are you sure this is correct?

jpetitcolas: It works when running demo. Yet, E2E tests fail. I'm checking out why.

fzaninotto:
Can you please rebase?

packages/ra-core/src/controller/ListController.js
```diff
@@ -71,7 +74,7 @@ export class ListController extends Component {
     componentDidMount() {
         if (
             !this.props.query.page &&
-            !this.props.ids.length &&
+            !(this.props.ids || []).length &&
```
fzaninotto: in which case isn't `ids` an array?

packages/ra-core/src/controller/ListController.js
fzaninotto: let me ask this in a different way: why did you remove the local cache with `this.filters`?

jpetitcolas:
> let me ask this in a different way: why did you remove the local cache with this.filters?

The local cache has been introduced in [Fix push to same location bug due to redux onChange commit](https://github.com/marmelab/admin-on-rest/commit/c88f52685a4787e7ec28d348d237c8a4d45a2efb#diff-a72191de31abba2b5717726bba2abec1) to (from what I understand) fix an issue with ReduxForm `onChange` property, causing multiple redirects. Checking on current version, there is no more such issue:

1. Updating form value updates correctly URL and list,
2. Same when clearing a filter value,
3. Same when adding a new filter,
4. Same when hiding a filter.

I am not sure why the class property has been introduced, as IMO, the `debounce` should be enough. If we want to optimize to prevent some extra `setFilters`, we may check directly on the `filters` prop. 

`this.filters` mimics the `this.props.filters`as we update it on both constructor and `componentWillReceiveProps` methods. That's why I removed it as there is no extra logic on it.

PR #1616, opened 2018-03-07
[RFR] Fix next E2E tests

jpetitcolas:
* [x] Fix E2E tests
* [ ] Reset filters when navigating away on ListController

### DIFF
--- a/e2e/pages/ShowPage.js
+++ b/e2e/pages/ShowPage.js
@@ -76,18 +76,17 @@ export default (url, initialField = 'title') => driver => ({
         return driver.sleep(200);
     },
 
-    delete() {
-        return driver
-            .findElement(this.elements.deleteButton)
-            .then(button => button.click())
-            .then(() =>
-                driver.findElement(this.elements.deleteConfirmButton).click()
-            )
-            .then(() => driver.sleep(200))
-            .then(() =>
-                driver.wait(until.elementLocated(this.elements.snackbar), 3000)
-            )
-            .then(() => driver.findElement(this.elements.body).click()) // dismiss notification
-            .then(() => driver.sleep(200)); // let the notification disappear (could block further submits)
+    async delete() {
+        const deleteButton = await driver.findElement(
+            this.elements.deleteButton
+        );
+        await deleteButton.click();
+
+        const deleteConfirmationButton = await driver.findElement(
+            this.elements.deleteConfirmButton
+        );
+        await deleteConfirmationButton.click();
+
+        await this.waitUntilDataLoaded();
     },
 });
--- a/e2e/tests/list.js
+++ b/e2e/tests/list.js
@@ -11,7 +11,9 @@ describe('List Page', () => {
         'http://localhost:8083/#/comments'
     )(driver);
 
-    beforeEach(async () => await ListPagePosts.navigate());
+    beforeEach(async () => {
+        await ListPagePosts.navigate();
+    });
 
     describe('Pagination', () => {
         it('should display paginated list of available posts', async () => {
@@ -30,7 +32,7 @@ describe('List Page', () => {
         });
     });
 
-    describe.only('Filtering', () => {
+    describe('Filtering', () => {
         it('should display `alwaysOn` filters by default', async () => {
             await driver.wait(
                 until.elementLocated(ListPagePosts.elements.filter('q'))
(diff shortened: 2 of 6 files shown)


e2e/pages/ShowPage.js
```diff
@@ -76,18 +76,17 @@ export default (url, initialField = 'title') => driver => ({
...
+            this.elements.deleteButton
+        );
+        await deleteButton.click();
+
+        const deleteConfirmationButton = await driver.findElement(
+            this.elements.deleteConfirmButton
+        );
+        await deleteConfirmationButton.click();
+
+        await this.waitUntilDataLoaded();
```
jpetitcolas: Just wait the list page has loaded instead of relying on a MaterialUI component. I guess their API changed.

e2e/tests/list.js
```diff
@@ -11,7 +11,7 @@ describe('List Page', () => {
         'http://localhost:8083/#/comments'
     )(driver);
 
-    before(async () => await ListPagePosts.navigate());
+    beforeEach(async () => await ListPagePosts.navigate());
```
jpetitcolas: All tests were dependent, causing some troubles when adding an `.only` and revealing some issues.

djhi: Yes and that's expected actually. Some tests are relying on the previous

jpetitcolas:
Switching to RFR.

e2e/pages/ShowPage.js
```diff
@@ -76,18 +76,17 @@ export default (url, initialField = 'title') => driver => ({
...
 
-    delete() {
-        return driver
-            .findElement(this.elements.deleteButton)
-            .then(button => button.click())
-            .then(() =>
-                driver.findElement(this.elements.deleteConfirmButton).click()
-            )
-            .then(() => driver.sleep(200))
-            .then(() =>
```
fzaninotto: I prefer that you don't remove this part, as it's used to avoid that the snackbar is still over the create button in further tests

e2e/tests/list.js
```diff
@@ -76,10 +80,15 @@ describe('List Page', () => {
...
+            await ListPagePosts.setFilterValue('q', 'quis culpa impedit');
+
             await ListPageComments.navigate();
             await ListPagePosts.navigate();
-            assert.equal(await ListPagePosts.getNbPagesText(), '1-10 of 13');
+
+            const filterValue = await ListPagePosts.getFilterValue('q');
+            assert.equal(filterValue, 'quis culpa impedit');
+            assert.equal(await ListPagePosts.getNbPagesText(), '1-1 of 1');
         });
```
fzaninotto: please reset the filter before leaving this test case

e2e/tests/list.js
```diff
@@ -121,8 +131,13 @@ describe('List Page', () => {
...
 
         it('should allow to trigger the delete bulk action on selected items', async () => {
+            await ListPagePosts.toggleSelectSomeItems(3);
             await ListPagePosts.applyDeleteBulkAction();
             assert.equal(await ListPagePosts.getNbPagesText(), '1-10 of 10');
         });
```
fzaninotto: please untoggle all at the end

e2e/tests/list.js
```diff
@@ -121,8 +131,13 @@ describe('List Page', () => {
...
     });
+
+    afterEach(async () => {
+        await ListPagePosts.setFilterValue('q', '');
```
fzaninotto: please remove

packages/ra-ui-materialui/src/list/FilterForm.js
```diff
@@ -180,6 +181,7 @@ const enhance = compose(
     reduxForm({
         form: 'filterForm',
         enableReinitialize: true,
+        destroyOnUnmount: false, // do not destroy to preserve state across navigation
```
fzaninotto: this is not a good solution

e2e/pages/ShowPage.js
jpetitcolas: The notification for a successful delete has disappeared. WIll be addressed in another PR.

packages/ra-ui-materialui/src/list/FilterForm.js
fzaninotto: After discussion, it is the right solution, because React 16 mounts before unmounting, so the redux-form is initialized then destroyed if we don't do that.

tkvw: Another solution would be to use a `form` identifier based on the resource, something like: 
```
withProps(({resource})=> ({
    form: `${resource}FilterForm`
})),
reduxForm({
   enableReinitialize: true,
})
```

fzaninotto: Yes, we thought about it. But it won’t work for the edit and create forms (which suffer from the same problem), because in that case, making the form name resource specific has nasty side effects (like breaking the dépendent input add-on).

PR #1618, opened 2018-03-08
[RFR] Fix form unmount issues

### LINKED ISSUE #1613: Linking from page with form to another page with form destroys the redux-form
**update**

All the debugging lead me to this:
https://github.com/erikras/redux-form/issues/3435

It is a problem with React 16 and the way redux-form handles the initializing and destroying of the forms in their components lifecycle 

**What you were expecting:**
I am linking from an edit page with a form to a create page with a form passing some values by the query parameters. I am trying to fill these parameters as default values to the form. That's how I found this.

**What happened instead:**
It looks like the destroyOnUnmount function of redux-form destroys the form **after**  it has been initalized.

Maybe this comes somewhere between redux-form 6.6.3 which is used in aor 1.2.1 and redux-form 7.0.4

**Related code:**
See an example application here:
https://github.com/nik-lampe/admin-on-rest-form-destruction

I had problems recreating this bug in the older version of admin-on-rest which is included in the CodeSandbox you're refering to: https://codesandbox.io/s/ElxGNxBY0

It seems to work with admin-on-rest 1.2.1. I have made a branch in the repo for this.

But updating the dependencies in the CodeSandbox seem to break the code, that's why I just created the sample application.

**Steps to reproduce:**
1. Click on comments list.
2. Click an create.
3. See the initial value for the body.

Then:
1. Click on posts list. 
2. Edit any Post.
3. Click on "Add Comment" in the action bar.
4. The body value should have a default value 'Test Body' but it doesnt. (I am not actually using any of the query parameters, I made this link just to show me use case for why I built it)
5. Debugging the redux store shows the destruction of the record-form

1.4.0:
admin-on-rest 1.4.0

1.2.1:
admin-on-rest 1.2.1

**Environment**

* Admin-on-rest version: 1.5.0

jpetitcolas:
Fixes #1613.

A bug is present from the upgrade to React16:

1. Edit a post on /#/posts/5,
2. Change URL and go on /#/comments/5 instead

Comment form is empty. This is due to the way React16 handle component lifecycle: it mounts the new component, then unmount the previous one. Hence, our form values are first initialized during the mounting of the new component, and then destroyed by the unmounting. As we are using the same name, it raises the issue.

We may either change form names, or never unmount our forms. As the first solution would break a lot of ecosystem plug-in (for instance [aor-dependant-input](https://github.com/marmelab/aor-dependent-input)), we choose the latter solution. Memory footprint should not be too much increased, as we only have two different forms: `filtersForm` and `recordForm`.

### DIFF
--- a/e2e/pages/EditPage.js
+++ b/e2e/pages/EditPage.js
@@ -75,6 +75,11 @@ export default url => driver => ({
         return input.sendKeys(value);
     },
 
+    async getInputValue(name) {
+        const input = await driver.findElement(this.elements.input(name));
+        return await input.getAttribute('value');
+    },
+
     clickInput(name) {
         const input = driver.findElement(this.elements.input(name));
         input.click();
--- a/e2e/tests/edit.js
+++ b/e2e/tests/edit.js
@@ -3,51 +3,69 @@ import driver from '../chromeDriver';
 import editPageFactory from '../pages/EditPage';
 
 describe('Edit Page', () => {
-    const EditPage = editPageFactory('http://localhost:8083/#/posts/5')(driver);
+    const EditPostPage = editPageFactory('http://localhost:8083/#/posts/5')(
+        driver
+    );
+    const EditCommentPage = editPageFactory(
+        'http://localhost:8083/#/comments/5'
+    )(driver);
 
-    beforeEach(async () => await EditPage.navigate());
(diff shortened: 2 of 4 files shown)


natrim:
would be nice to have the `destroyOnUnmount: false,` on master branch too

PR #1621, opened 2018-03-08
[RFR]: Revert fetch saga to takeEvery

tkvw:
Because of the new `select` in the fetch saga there is potential danger to miss a fetch request. This PR ensures all fetches are catched by using a Channel to allow requests to be queued if the saga is yielding on another action.


fzaninotto:
Now I'm thinking: why did I use `takeLatest` for fetch actions in the first place? It alls goes back to my first commit, but I can't remember the reason why I thought it was better to do that than a `takeEvery`. In which case is it relevant to cancel the previous fetch? In my opinion as of today, it's premature optimization. 

Your previous fix came from that `takeLatest`, and it broke #1625. Shouldn't we roll back to `takeEvery` instead?

tkvw:
My main reason was to allow multiple requests to different resources at the same time. The `cancelPrevious` would end any previous running `fetch` request to other resources as well which is wrong imo (this would never happen in the current `react-admin` implementation, but I'm using the fetch actions in my custom app) . I'm not sure https://github.com/marmelab/admin-on-rest/issues/1625 is caused by this,there are only two `CRUD_GET_MANY` actions in the log, so I think there's something else at play. This saga listens to the `CRUD_GET_MANY` action but does not remove one. So @afilp should come up with a reproducible example I think.

afilp:
Would this comment help to understand where to look in the code?
https://github.com/marmelab/admin-on-rest/issues/1625#issuecomment-371992272

I would gladly investigate which Filter prop may cause the initiation of the problem, if you could give me some hints/thoughts.

tkvw:
@afilp : the reference to the same resource is causing the problem. `ReferenceInputController` uses `crudGetOne` to get the reference, this uses `cancelPrevious` on the same resource. You should see something like that in the redux log (`FETCH_CANCEL`). 
But I seriously doubt this behaviour diffs from the `takeEvery`, the `cancelPrevious` was in there as well. 

@fzaninotto : any thoughts? Make `CRUD_GET_ONE` `cancelPrevious=false`? or have `CRUD_GET_LIST` and `CRUD_GET_ONE` to run simultaneous?

afilp:
@tkvw Indeed, I see a CANCEL, with this order of actions:

image

fzaninotto:
Let me reiterate: this problem, and the previous fix by @tkvw, come from the fact that the `fetch` saga only takes the *latest* fetch action for a certain resource and type, i.e. it cancels the previous one. Instead of circumventing this, I propose to challenge it. 

Why do we use `takeLatest` instead of `takeEvery`? The only use case I can come up with is to avoid race conditions if a user browses to a post list, then immediately to a comments list, but the response for the comment list comes *before* the response to the post list.

It's nice to have a safeguard for that case, but seing how it has nasty side effects, and considering it's rare, my suggestion is to remove the `cancelPrevious` logic altogether, and to pass *every* action with the `fetch` meta to the `fetch` saga.

Thoughts?

tkvw:
👍 if it is a important use case for the end user it can still be implemented in the `dataProvider`.

tkvw:
Reverted and added an option to `opt-out` of  `isOptimistic` skipping. Not sure about the property name though, `ignoreOptimistic`.

fzaninotto:
Why do you need this `isOptimistic` in the first place? 

Also, IMO the data actions should be optimistic-agnostic. It's the job of the `undo` saga to transform underlying actions to make them optimistic.

tkvw:
You brought it in on commit https://github.com/marmelab/admin-on-rest/commit/7b7a317ad574ee2dcbe3242b234fd4598376f254 so I figured you wanted to keep it. 

It probably has to do with the fact that you don't want the list to refresh when there's an optimistic action underway (which is good imo).

fzaninotto:
Yes, we should keep the `isOptimistic` from the state (toggle by the undo sage), but not the `isOptimistic` in the action creators.

tkvw:
Ah you mean the `ignoreOptimistic` in the dataActions creators, I understand you don't want them there.

My main issue is: when the ui is in `optimistic` mode, not any fetch will run. What if I have an unrelated `TickerComponent` which uses `fetch` and it decides to update? 

A solution could be to add the `resource` prop to `START_OPTIMISTIC_MODE` so we can check in the fetch saga that only the fetches to this resource are ignored. What do you think?

fzaninotto:
It won’t work if there are ReferenceField in the data grid for instance.

tkvw:
Not sure if I see a solution without using the `dataActions` for it. It can be made optimistic agnostic by using a different `meta` flag, `force` for example. But this still requires an additional arg on the dataActions. 

Or you can opt not to support this at all ofcourse.

tkvw:
Reverted to `takeEvery` and removed `fetch` tests ( imo there isn't much use in a saga unit test if there's no behavioural test to do).

afilp:
I think that another reason to use the `takeEvery` instead of the `takeLatest` is this:

I have a list of "assets" (with "primary" key being the "asset_id").
One of the fields is "parentAsset_id". Which means that the field references its own collection in order to get another "asset_id"as "parentAsset_id".

Unfortunately, this `SelectInput` again shows a "loading" indicator, and in "Edit" which is not a parallel process (I may wait 10 seconds to press "Edit", so no other loading exists by that time).

Is it the same problem? Or `takeLatest` does not even solve this?

fzaninotto:
Your PR has failing tests.

I don't understand your use case: you mention a list, then a `SelectInput`... shouldn't it be `SelectField`? Can you rephrase?

tkvw:
> Your PR has failing tests.

There was a problem with the CI server, the e2e tests didn't run.

fzaninotto:
Thanks!

afilp:
@fzaninotto I meant to say `I have "assets"`, I changed the wording, I refer to the 'Edit' part if we do not use <DependentInput /> but straight <SelectInput />. But do not bother, I will test with the new changes and will let you know.

afilp:
Yes, that works too, so not problems now. Thanks!

PR #1622, opened 2018-03-08
[RFR]: Pass through action in generic Fetch actions

tkvw:
I need to monitor each individual fetch request to the initiator. Currently there's no way of telling which `FETCH_*` action belongs to which action. I like to tie them down to an action by simply passing the action in the generic fetch actions (`FETCH_START,FETCH_END,FETCH_ERROR and FETCH_CANCEL`). This is not in anyway helpful for `ra` at the moment, but would help me.


fzaninotto:
Sorry, I don't see a use case for react-admin.

PR #1623, opened 2018-03-08
[RFR]: Remove passing through rest props to Labelled component

tkvw:
For `ReferenceInput` and `ReferenceArrayInput` remove passing through rest props to Labelled component. Is gives a lot of errors in the console, because `Loading` does support there props.


fzaninotto:
Thanks!

PR #1624, opened 2018-03-08
Fix 'bulkActions' boolean functionality

afilp:
The page breaks (with error message), when using `bulkActions` prop as a boolean `false` (for hiding the checkboxes):
```
<List
    {...props}
    title={<TitleList />}
    bulkActions={false}
  >
```

This PR fixes this problem.

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListActions.js
+++ b/packages/ra-ui-materialui/src/list/ListActions.js
@@ -21,13 +21,14 @@ const Actions = ({
 }) => {
     return (
         <CardActions className={className} {...rest}>
-            {cloneElement(bulkActions, {
-                basePath,
-                filterValues,
-                resource,
-                selectedIds,
-                onUnselectItems,
-            })}
+            {bulkActions &&
+                cloneElement(bulkActions, {
+                    basePath,
+                    filterValues,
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

afilp:
You're welcome! 

BTW, I see you are on fire!! 
(a burst of commits today, which I love, you know it by now... :-) ) Thanks again for your efforts!

tkvw:
@afilp : did you consider using a `thruthy` value instead of `false`? i.e.: 
```
<List disableBulkActions />
```
looks nicer than
```
<List bulkActions={false} />
```

afilp:
@tkvw Initially I did, but then I realized that this has been deliberate feature. 

See the code here: https://github.com/marmelab/admin-on-rest/blob/8ef62c18259e3b2270e1ad2b7643776b583ec31c/packages/ra-ui-materialui/src/list/List.js#L277

And here:
https://github.com/marmelab/admin-on-rest/blob/next/docs/List.md
**(search for `disable bulk actions`)**

fzaninotto:
> you are on fire!!

No, this is GitHub rebase showing all my commits from the past 2 weeks in one day.

PR #1627, opened 2018-03-09
[RFC] Add a maximum page limit to the List view

Kmaschta:
The goal of this PR is to introduce a new feature to the List view: limit the page number if needed by providing a new optional prop `maxPage` to the `List` and `Pagination` component.

I don't know if it will be widely used.. for one of our clients, it's a "security issue".
So, I made a PR in case it might be useful.

## TODO
- [x] Update the documentation
- [x] Use `maxPage` in the Pagination
- [x] Pass the `maxPage` prop through the List & List controller
- [ ] Add a E2E test (?)
- [ ] Add this prop in at least one example

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -22,6 +22,7 @@ Here are all the props accepted by the `<List>` component:
 * [`bulkActions`](#bulk-actions)
 * [`filters`](#filters) (a React element used to display the filter form)
 * [`perPage`](#records-per-page)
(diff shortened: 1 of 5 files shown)


Kmaschta:
Finally, it is totally possible to write a custom Pagination component (or wrapper) with a new prop to limit the pages.

It's not worth it to introduce and maintain this prop in the core API.

PR #1630, opened 2018-03-09
[WIP]: Load list data after undoable edit

tkvw:
Fixes https://github.com/marmelab/admin-on-rest/issues/1626

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/list/initiated.js
+++ b/packages/ra-core/src/reducer/admin/resource/list/initiated.js
@@ -1,0 +1,12 @@
+import { CRUD_GET_LIST_SUCCESS } from '../../../../actions/dataActions';
+export default resource => (previousState = false, { type, meta }) => {
+    if (!meta || meta.resource !== resource) {
+        return previousState;
+    }
+    switch (type) {
+        case CRUD_GET_LIST_SUCCESS:
+            return true;
+        default:
+            return previousState;
+    }
+};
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/reducer/admin/resource/list/initiated.js
```diff
@@ -0,0 +1,12 @@
+import { CRUD_GET_LIST_SUCCESS } from '../../../../actions/dataActions';
+export default resource => (previousState = false, { type, meta }) => {
```
fzaninotto: how about using the `total` reducer instead (modifying it to start with a `null` state instead of zero)?

tkvw: I was actually looking for one to use, and opted out from `ids` because a list can be empty, but didn't notice total, which is a good place to do this 👍

tkvw:
What do you mean with `placeholder`? Something like: `List data will be loaded soon` or `Click here to initialize data`? 

I think the main problem is the `stale` record issue. If you issue an undoable action => request list => list data received => you'll see the current data as before the undoable action data was applies => undoable action is applied => the updated data is displayed. 

So there'll be a short period of time you'll see the stale record in the list (if it is in view).  Not sure, but I think this issue would remain by issuing a refresh from the ra refresh button, because it all depends on the fact if the undoable actions was applied already or still delayed while the new list data was received.

fzaninotto:
I mean a table with a few rows populated by grey rectangles, cf https://www.sitepoint.com/how-to-speed-up-your-ux-with-skeleton-screens/

As for your explanation of the `stale` issue, I didn't understand you. Could you develop a bit?

fzaninotto:
Bump.

fzaninotto:
Superseded by #1728

PR #1631, opened 2018-03-09
[RFR]: Use ui to trigger undoable complete action

tkvw:
So a user be like: let me check the notification in the snackbar by hovering my mouse over it. After a couple of seconds he/she understands the message and decides to undo the action, but it does not work, the record is already removed. see screenshot: 

This PR removes the `delay` from the saga and allows the UI to call the complete: 
As you can see nothing happens while the mouse is over the snackbar.


fzaninotto:
I didn't know that hovering on the snackbar paused the notification. Your modification makes sense.

fzaninotto:
Thanks!

PR #1635, opened 2018-03-11
[RFR] fix GET_MANY_REFERENCE example in RestClients.md

Kiailandi:
The GET_MANY_REFERENCE example code was broken because of a `;` before the `.then`, probably carried over from the previous example.


djhi:
Thanks

PR #1636, opened 2018-03-11
[RFR]: Show notifications if available at mount

tkvw:
Notifications didn't show if there are any available before the `Notification` component is mounted.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Notification.js
+++ b/packages/ra-ui-materialui/src/layout/Notification.js
@@ -26,13 +26,13 @@ class Notification extends React.Component {
         open: false,
     };
     componentWillMount = () => {
-        this.setNextState();
+        this.setOpenState(this.props);
     };
     componentWillReceiveProps = nextProps => {
-        this.setNextState(nextProps);
+        this.setOpenState(nextProps);
     };
 
-    setNextState = ({ notification } = this.props) => {
+    setOpenState = ({ notification }) => {
         this.setState({
             open: !!notification,
         });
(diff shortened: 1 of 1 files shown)


fzaninotto:
Can you explain in which case this can happen, i.e. which bug is it supposed to fix?

tkvw:
Sure, user navigates to a restricted page without authorization -> with a custom saga I call `showNotification` -> user is redirected to `/login` -> `/login` has it's own `Notification` component so when it mounts, the notification in the redux-store will not render (because the initial state `open` is `false`). The `open` state should be recalculated on mount of the component (just like `componentWIllReceiveProps`)

packages/ra-ui-materialui/src/layout/Notification.js
```diff
@@ -25,9 +25,16 @@ class Notification extends React.Component {
     state = {
         open: false,
     };
+    componentWillMount = () => {
+        this.setNextState();
```
fzaninotto: shouldn't you use `this.props` here?

tkvw: The args of `setNextState` defaults to `this.props`

fzaninotto: True, I hadn't paid attention. Or rather, the code wasn't clear enough. Could you please pass `this.props` here and have no default value in `setNextState`?

packages/ra-ui-materialui/src/layout/Notification.js
```diff
@@ -25,9 +25,16 @@ class Notification extends React.Component {
...
+    };
     componentWillReceiveProps = nextProps => {
+        this.setNextState(nextProps);
+    };
+
+    setNextState = ({ notification } = this.props) => {
```
fzaninotto: While we're at it, please rename this method to `setOpenState`

fzaninotto:
Thanks!

PR #1641, opened 2018-03-12
[RFR]: Notification can be undefined when hidden through saga

tkvw:
When a saga calls `hideNotification` and the `Notification` component is visible, the `notification.undoable` check fails. This PR ensures notification exists.


fzaninotto:
Thanks!

PR #1646, opened 2018-03-14
[RFR]: Allow custom validation message and allow message to be a function

tkvw:
Allow to customize each validation message and allow message to be a function. 

The later is useful if the user wants to allow lazy initialization of the error message (useful for nested lookups). 

It's a breaking change, because `required` validator should now be used as `required()`, `number` as `number()` and `email` as `email()`. This is inline with all other validators.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -25,7 +25,7 @@
 - [Logout is now displayed in the AppBar on desktop](#logout-is-now-displayed-in-the-appbar-on-desktop)
 - [Data providers should support two more types for bulk actions](#data-providers-should-support-two-more-types-for-bulk-actions)
 - [react-admin addon packages renamed with ra prefix and moved into root repository](#react-admin-addon-packages-renamed-with-ra-prefix-and-moved-into-root-repository)
-
+- [require,number and email validation should be renamed to require(),number() and validation()](#validations-should-be-initialized)
 
 ## Admin-on-rest Renamed to React-Admin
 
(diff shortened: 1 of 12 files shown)


docs/CreateEdit.md
```diff
@@ -380,12 +380,12 @@ export const ProductEdit = ({ ...props }) => (
...
 
-* `required` if the field is mandatory,
+* `required(message)` if the field is mandatory,
 * `minValue(min, message)` to specify a minimum value for integers,
 * `maxValue(max, message)` to specify a maximum value for integers,
 * `minLength(min, message)` to specify a minimum length for strings,
 * `maxLength(max, message)` to specify a maximum length for strings,
-* `number` to check that the input is a valid number,
+* `number(message)` to check that the input is a valid number,
 * `email` to check that the input is a valid email address,
```
danejorgensen: * `email(message)` to check that the input is a valid email address,

UPGRADE.md
```diff
@@ -1325,3 +1325,27 @@ import buildGraphcoolProvider from 'aor-graphql-client-graphcool';
...
 import buildGraphcoolProvider from 'ra-data-graphcool';
 ```
+
+## Validations should be initialized
+
+Previously the message of `required`,`number` and `email` weren't customizable.
+
+Update your `require`,`number` and `email` validations. 
+ 
+```jsx
```
fzaninotto: I prefer the `diff` syntax (see in the first half of the UPGRADE document)

UPGRADE.md
```diff
@@ -1325,3 +1325,27 @@ import buildGraphcoolProvider from 'aor-graphql-client-graphcool';
 // after
 import buildGraphcoolProvider from 'ra-data-graphcool';
 ```
+
+## Validations should be initialized
+
+Previously the message of `required`,`number` and `email` weren't customizable.
```
fzaninotto: That's not the gist of the change. You should be more explicit, something like:

> The `required`,`number` and `email` validators must now be executed just like the other validators, not passed as function arguments
>
>```diff
> -<TextInput source="foo" validate={[required, maxSize(2)]} />
> +<TextInput source="foo" validate={[required(), maxSize(2)]} />
>```

fzaninotto:
Thanks!

PR #1649, opened 2018-03-14
[RFR]: Default redirect disables passing redirect to SimpleForm

tkvw:
The `redirect` is defaulted in `EditController` and should be passed down to `SaveButton`. The `defaultProp` in `SaveButton` overrides this.
```
<Edit>
<SimpleForm redirect={false}>
</SimpleForm>
</Edit>
```
This currently does not work, the redirect still is `list`. (I know I can pass custom `toolbar`).

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.js
+++ b/packages/ra-ui-materialui/src/button/SaveButton.js
@@ -117,7 +117,6 @@ SaveButton.propTypes = {
 };
 
 SaveButton.defaultProps = {
-    redirect: 'list',
     handleSubmitWithRedirect: () => () => {},
 };
 


fzaninotto:
The `redirect` prop of the form dictates what happens to the form when submitted with Enter. The `redirect` prop of the Save button dictates what happens to the form when submitted with the button.

If you override the `redirect` prop of the `SaveButton`, it's no longer possible to have two save buttons with different redirects (save=>list and save=>add).

This is properly documented, I don't see the bug.

tkvw:
> The redirect prop of the form dictates what happens to the form when submitted with Enter. 

Yeah, so this does not work, this PR fixes this behaviour. 

> If you override the redirect prop of the SaveButton, it's no longer possible to have two save buttons with different redirects (save=>list and save=>add).

This PR only removes the default value of `redirect` on the `SaveButton` this should only be set explicit.
SaveButton passes `redirect` to `handleSubmitWithRedirect`, but because it's defaulted to `list` in SaveButton this will always be `list` unless specified explicitly. 

`handleSubmitWithRedirect` in `SimpleForm` uses `this.props.redirect` as default value (which is passed from `EditController`). As SaveButton always passes a value (because of the default `redirect` prop) this value will never be used.

Pinpickle:
I'm also being hit by the issue fixed by this PR (on `2.0.0-beta4`). The [instructions given in the docs](https://marmelab.com/admin-on-rest/CreateEdit.html#redirection-after-submission) for custom redirects do not change where the "Save" button redirects to (on edit or update views).

fzaninotto:
Thanks!

PR #1653, opened 2018-03-14
[RFR]: Allow translate fallback keys

tkvw:
Adds fallback keys for translatable fields. This allows an array of messages to be supplied to `translate`. 
A fallback is added for the `FieldTitle` lookup: `resources.${resource}.fields.${source}` and `common.fields.${source}` is checked. 

Todo: 

- [X] Add docs

### DIFF
--- a/docs/Translation.md
+++ b/docs/Translation.md
@@ -220,6 +220,7 @@ However, before humanizing names, react-admin checks the `messages` dictionary f
 
 - `${locale}.resources.${resourceName}.name` for resource names (used for the menu and page titles)
 - `${locale}.resources.${resourceName}.fields.${fieldName}` for field names (used for datagrid header and form input labels)
+- `${locale}.common.fields.${fieldName}` for common field names (used for datagrid header and form input labels)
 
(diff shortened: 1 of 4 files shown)


fzaninotto:
I don't see the interest, please explain why the current way is a problem.

tkvw:
I have a lot of resources using the same field sources: `id`,`dateCreated`,`lastUpdated`,`createdBy`,etc. 
Instead of writing the same translation for for each resource, this allows a user to write a generic translation for these fields once. 
```
users.fields.dateCreated:'dateCreated'
accounts.fields.dateCreated:'dateCreated'
```
vs
```
common.fields.dateCreated: 'dateCreated'
```

fzaninotto:
This is advanced usage scenario, I guess this may be the job of a custom `i18nProvider`.

tkvw:
Yeah, that's an alternate way of doing it. I think it's nice to have as default, but I can live without.

fzaninotto:
After consideration, we won't add this ability to the core. Thanks for your contribution, though.

PR #1657, opened 2018-03-15
Update redux-form to 7.1.X

mauchede:
Hello,

We have [some warnings on api-platform/admin](https://travis-ci.org/api-platform/admin/jobs/353729496#L501) due to [redux-form](https://github.com/erikras/redux-form/issues/3385).

Update `redux-form` to (at least) `7.1.0` resolves the problem (I used `7.1.1` because `7.1.0` had [a problem with immutablejs](https://github.com/erikras/redux-form/issues/3493)).

According to [your version policy](https://github.com/marmelab/admin-on-rest/pull/1533), I kept `tilde` for version range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
         "react-router-redux": "~5.0.0-alpha.6",
         "recompose": "~0.25.0",
         "redux": "~3.7.2",
-        "redux-form": "~7.0.3",
+        "redux-form": "~7.1.1",
         "redux-saga": "~0.15.0",
         "reselect": "~3.0.0"
     }


fzaninotto:
I'm not sure we'll merge this, as we're limiting updates on the `master` branch.

mauchede:
I understand you want to be focus on `next` branch. But the warnings are very noisy: on Travis we have more "warning" lines than "legit" lines. Your project also has the problem but [it is printed only once](https://travis-ci.org/marmelab/admin-on-rest/builds/352190356#L1220)

We have only two possibilities:
* run Jest with `--silent` flag.
* force `npm` / `yarn` to retrieve a fixed version.

Unfortunately, these solutions are not acceptable.

This fix will give some time for all project that depending of admin-on-rest (like our project) to prepare the update to next version.

By the way, the problem is also present on `next`. If you wish, I can make another PR for `next` branch.

fzaninotto:
Thanks

mauchede:
Thank you for merging!

PR #1665, opened 2018-03-16
[RFR] Use the new onSuccess meta for reviews actions

djhi:


examples/demo/src/reviews/reviewSaga.js
```diff
@@ -1,23 +1,9 @@
 import { all, put, takeEvery } from 'redux-saga/effects';
-import { push } from 'react-router-redux';
 import { showNotification } from 'react-admin';
-import {
-    REVIEW_APPROVE_SUCCESS,
-    REVIEW_APPROVE_FAILURE,
-    REVIEW_REJECT_SUCCESS,
-    REVIEW_REJECT_FAILURE,
-} from './reviewActions';
+import { REVIEW_APPROVE_FAILURE, REVIEW_REJECT_FAILURE } from './reviewActions';
```
djhi: Should we keep those ?

fzaninotto: I think we should try to remove this saga altogether. What prevents you from it?

examples/graphcool-demo/src/reviews/reviewActions.js
```diff
@@ -8,7 +8,19 @@ export const REVIEW_APPROVE_SUCCESS = 'REVIEW_APPROVE_SUCCESS';
 export const reviewApprove = (id, data, basePath) => ({
     type: REVIEW_APPROVE,
     payload: { id, data: { ...data, status: 'accepted' }, basePath },
-    meta: { resource: 'Review', fetch: UPDATE, cancelPrevious: false },
+    meta: {
+        resource: 'reviews',
+        fetch: UPDATE,
+        cancelPrevious: false,
```
fzaninotto: no longer useful

examples/demo/src/reviews/reviewActions.js
```diff
@@ -19,5 +31,17 @@ export const REVIEW_REJECT_SUCCESS = 'REVIEW_REJECT_SUCCESS';
 export const reviewReject = (id, data, basePath) => ({
     type: REVIEW_REJECT,
     payload: { id, data: { ...data, status: 'rejected' }, basePath },
-    meta: { resource: 'reviews', fetch: UPDATE, cancelPrevious: false },
+    meta: {
+        resource: 'reviews',
+        fetch: UPDATE,
+        cancelPrevious: false,
```
fzaninotto: no longer useful

examples/demo/src/reviews/reviewActions.js
```diff
@@ -8,7 +8,19 @@ export const REVIEW_APPROVE_SUCCESS = 'REVIEW_APPROVE_SUCCESS';
 export const reviewApprove = (id, data, basePath) => ({
     type: REVIEW_APPROVE,
     payload: { id, data: { ...data, status: 'accepted' }, basePath },
-    meta: { resource: 'reviews', fetch: UPDATE, cancelPrevious: false },
+    meta: {
+        resource: 'reviews',
+        fetch: UPDATE,
+        cancelPrevious: false,
```
fzaninotto: no longer useful

examples/demo/src/reviews/reviewSaga.js
```diff
@@ -1,23 +1,9 @@
...
     yield all([
-        takeEvery(REVIEW_APPROVE_SUCCESS, function*() {
-            yield put(
-                showNotification(
-                    'resources.reviews.notification.approved_success'
-                )
-            );
-            yield put(push('/reviews'));
-        }),
         takeEvery(REVIEW_APPROVE_FAILURE, function*({ error }) {
```
fzaninotto: Now I understand why we still need the saga. That means something is missing in the core sagas (`fetch` and `undo`); we should be able to handle `onFailure` side effects from the action meta, too.

djhi: Wouldn't it be redundant with the default dataProvider error handling ?

fzaninotto: No, the `dataProvider` has no particular handling. It can return exceptions, that the `fetch` saga redispatches as `_FAILURE` actions. But the side effects of the fetch error (redirection, notification, refresh) could be described as data, in the initial action, just like the `onSuccess` meta.

PR #1671, opened 2018-03-19
[RFR]: SelectArrayInput don't translate when optionText is a React Component

tkvw:
The translate option did not check if the optionText is a component. Component clones should not be translated. Added `translate` prop to clone properties.


fzaninotto:
Why would you add the `translate` prop? It's the component(s responsibility to grab the `translate` prop using the `translate` HOC. Otherwise, you risk that the component doesn't need `translate` and passes it to underlying components, possibly DOM elements, and trigger a warning.

tkvw:
I thought about that, but because it's an `OptionText` with potentially a lot of choices, I thought this would be more efficient. But I'll remove it

tkvw:
Translate prop removed

fzaninotto:
thanks!

PR #1672, opened 2018-03-19
[RFR] ReferenceArrayInput label not rendered, because it's initialized with empty string

tkvw:
Fixes topic.


fzaninotto:
Sorry, I don't understand the problem. How do you use the component, what do you expect, what happens that's not correct?

tkvw:
When you add a field enclosed with `<ReferenceArrayInput>` no default label (`resources.<resource>.fields.<source>`) will be rendered, because the label prop is defaulted to an empty string.

fzaninotto:
Thanks!

PR #1676, opened 2018-03-20
Create ISSUE_TEMPLATE.md

santhosh1190:
I am facing the checkbox issue in List page.

I have selected the some rows from list. After that If I click any place on the page which means outside of Grid, the selected checkbox became unselect.
I do not understand why it is happening like this.

Due to this problem, After selected rows I am not able to click the button to perform action on selected rows.
Someone help me please!

test

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,24 @@
+### Expected behavior
+
+1. Added the Checkbox in Grid view.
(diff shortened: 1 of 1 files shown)


fzaninotto:
Dude, you haven't read the issue template at all, you create a PR to get help... Your bad behavior creates noise and distracts us from actually making admin-on-rest better. Please read before you post.

PR #1677, opened 2018-03-20
[RFR] Introduce onFailure meta to handle failure side effects

fzaninotto:
- [x] Unroll failure side effects in `fetch` saga
- [x] Move failure side effects from the `error` saga to data action creators
- [x] Use and apply on the `demo` example
- [x] Document failure side effects

Action creators can now declare failure side effects alongside success side effects, in the `meta` property:

```js
export const crudDeleteMany = (resource, ids, basePath, refresh = true) => ({
    type: CRUD_DELETE_MANY,
    payload: { ids },
    meta: {
        resource,
        fetch: DELETE_MANY,
        onSuccess: {
            notification: {
                body: 'ra.notification.deleted',
                level: 'info',
                messageArgs: {
                    smart_count: ids.length,
                },
            },
            basePath,
            refresh,
            unselectAll: true,
        },
        onFailure: {
            notification: {
                level: 'warning',
            },
            refresh: true,
        },
    },
});
```

This removes the need for custom sagas in many cases.

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -279,6 +279,12 @@ export const commentApprove = (id, data, basePath) => ({
 +            redirectTo: '/comments',
 +            basePath,
 +        },
++        onFailure: {
++            notification: {
++                body: 'resources.comments.notification.approved_failure',
++                level: 'warning',
++            },
++        },
     },
 });
 ```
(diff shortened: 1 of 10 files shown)


docs/Actions.md
```diff
@@ -229,30 +231,87 @@ export const commentApprove = (id, data, basePath) => ({
...
+-        redirectTo: '/comments',
+-        basePath,
++        onSuccess: {
++            notification: {
++                body: 'resources.comments.notification.approved_success',
++                level: 'info',
++            },
++            redirectTo: '/comments',
++            basePath,
++        },
```
djhi: Shouldn't we also add an `onFailure` example ?

fzaninotto:
Review applied

PR #1680, opened 2018-03-20
[RFR] Change upgrade syntax to use diff for snippets

fzaninotto:
Same content, but more readable, and smaller (255 additions and 597 deletions).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -352,8 +352,6 @@ All react-admin components now accept a `className` prop instead of the `elStyle
 + export default withStyles(styles)(UserList);
 ```
 
-Yes, it's way longer to write and it's a pain, but that's material-ui's choice of CSS-in-JS implementation - and there is nothing the react-admin team can do about it. If this is a consolation, we suffered to upgrade our own apps to JSS, too.
-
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -352,59 +352,42 @@ All react-admin components now accept a `className` prop instead of the `elStyle
 + export default withStyles(styles)(UserList);
 ```
 
-In addition to `elStyle`, Field and Input components used to support a `style` prop to override the styles of the *container element* (the `<td>` in a datagrid). This prop is no longer supported in react-admin. Instead, the `Datagrid` component will check if its children have a `headerClassName` and `cellClassName` props. If they do, it will apply those classes to the table header and cells respectively.
-
-```jsx
-// before
-import { EmailField, List, Datagrid } from 'react-admin';
+Yes, it's way longer to write and it's a pain, but that's material-ui's choice of CSS-in-JS implementation - and there is nothing the react-admin team can do about it. If this is a consolation, we suffered to upgrade our own apps to JSS, too.
```
djhi: Not sure we should add that

fzaninotto: You're right, we can mention it in a blog post, not in an upgrade guide.

Issue #1681, opened 2018-03-21
feature: ReferenceManyField to support referencing record with other than id field


vedmalex:
Need feature to support referencing fields other than id|

fzaninotto:
You must map this reference in the rest client. We won’t add anything to the ReferenceManyField to support special identifiers.

vedmalex:
you mean that in next version?
but in current stable it is not working.

PR #1682, opened 2018-03-21
should support referencing record with other than id field #1681

vedmalex:
implement freature #1681 
very simple


fzaninotto:
No, the right place to do that is in the rest client.

PR #1687, opened 2018-03-22
[RFR] Upgrade ChromeDriver

fzaninotto:
Because e2e tests tend to fail often


tkvw:
I think the previous failure had to do with downgraded state of `unsplash` so pages didn't fully load, because the image wasn't available.

PR #1688, opened 2018-03-22
[RFR] DependsOn

djhi:
- [x] Merge `DependentInput` and `DependentField` into `DependsOn`
- [x] Move it the `ra-core` package
- [x] Update the upgrade guide
- [x] Update the documentation


packages/ra-core/src/util/DependsOn.js
```diff
@@ -47,7 +48,7 @@ export const DependentInputComponent = ({
     );
 };
 
-DependentInputComponent.propTypes = {
+DependsOnView.propTypes = {
```
fzaninotto: Why `DependsOnView` instead of `DependsOn`?

djhi: Because of the default export which is `DependsOn`. That will avoid eslint warning about importing a named export as default

docs/Fields.md
```diff
@@ -773,6 +773,129 @@ PriceField.defaultProps = {
 ```
 {% endraw %}
 
+## Customize views depending on its record values
+
+When you want to display fields only when some other fields are present or have a specific value, you can use the `DependsOn` component.
+
+The `DependsOn` component accepts the following props:
+
+* `dependsOn`: Either a string indicating the name of the field to check (eg: `hasEmail`) or an array of fields to check (eg: `['firstName', 'lastName']`). You can specify deep paths such as `author.firstName`.
```
fzaninotto: I think this should be renamed to `source`

PR #1694, opened 2018-03-23
Customization

JacquesBonet:
Little changes for better flexibility

Header: define its own color attribute, permitting Header component to have specific color

Layout: Appbar call, add classes attribute

DataGrid: TableHead call: add className attribute

List: Header call, add classes attributes


JacquesBonet:
Jest pull request must be accepted to permit to this request not to have error.

JacquesBonet:
With modification:

image

fzaninotto:
I don't like passing the entire `classes` to a child. It's better to define a key. 

Your PR mixes app bar and datagrid header, please split it into two. 

Also, please document your changes to that developers understand that they can override the styles by reading the documentation.

PR #1700, opened 2018-03-25
[RFR] Update material-ui and material-ui-icons dependencies

jeromemacias:
It seems that there is no breaking changes that impact react-admin


djhi:
Thanks @jeromemacias :)

PR #1703, opened 2018-03-27
[RFR] ReferenceInputController this.props missing in componentWillMount

tkvw:
ReferenceInputController this.props missing in componentWillMount so it will not fetch the options on mount.


djhi:
Thanks!

PR #1704, opened 2018-03-27
Update FAQ.md

grantliu:
Update FAQ.md to add parentheses around mapped objects, otherwise it would return compile error


djhi:
Thanks!

fzaninotto:
Bad target branch... This was merged into #1686

PR #1706, opened 2018-03-27
Fix login layout in IE 11

pavyarov:
Fix issue described in https://github.com/marmelab/admin-on-rest/issues/1705


fzaninotto:
Thanks!

fzaninotto:
I'm not porting it to the `next` branch as it is currently not compatible with IE11 (we're using `Array.includes()` and other things that IE11 doesn't support)

PR #1707, opened 2018-03-28
[RFR]: Render props

tkvw:
Allows to use render props to generate children on key components: SimpleForm, TabbedForm, Create, Edit and Show.


djhi:
Can you explain the problem this PR solves ?

fzaninotto:
If this is about using a render function as child of `Edit` instead of using a `<Form>` component for instance, I'm -1. You can already use the `EditController` with a render function, so this only brings confusion.

tkvw:
Yeah, so I kinda agree. I am using it and maybe somebody will search the issues and find it a reasonable solution for his/here problem.

PR #1709, opened 2018-03-29
NotFound with permission

qooplmao:
At the moment the `NotFound` page/component doesn't check for authentication. This means that when an app is using authentication and a user hits a 404, the `NotFound` page is shown whether they are authenticated or not. Clicking any resource links would kick the user our to `/login` but that doesn't seem the most user friendly.

This PR just wraps the regular `NotFound` page with a `Restricted` then set that as the default `catchAll` if provided with an `authClient`. If an unauthenticated user hits the `NotFound` page they are then automatically redirected to the `/login` page.

### DIFF
--- a/src/Admin.js
+++ b/src/Admin.js
@@ -60,9 +60,7 @@ const Admin = ({
     sagaMiddleware.run(saga);
 
     const logout = authClient ? createElement(logoutButton || Logout) : null;
-    const catchAll = authClient
-        ? createElement(customCatchAll || NotFoundWithPermission)
-        : customCatchAll;
+    const catchAll = customCatchAll;
 
     return (
         <Provider store={store}>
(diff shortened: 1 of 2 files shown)


fzaninotto:
For new features, please PR against `next`, not `master`. I'll close this one and let you reopen a new one.

stefsava:
@fzaninotto 
Seem fixes a bug and don't introduce BC break, master should be a valid target.
@qooplmao 
Anyway your PR is useful, reopen in next, please.
IMHO

fzaninotto:
In fact, I don't think there is a bug, and I don't think we should implement your recommendation. 

The reason is that users can add routes without authentication to the `<Admin>` component using the `customRoutes` prop. That means that parts of the admin can be accessible without authentication (and not only the Login page). So I don't see the point of hiding the 404 page behind authentication itself.

qooplmao:
What about just adding it to the `Authentication` section of the documentation?

PR #1716, opened 2018-04-01
Update version of components, especially material-ui from beta32 to b…

JacquesBonet:
Update versions of some components especially material-ui, from beta32 to beta40


fzaninotto:
Please contribute to #1700 instead.

PR #1717, opened 2018-04-03
[WIP] Support embedded arrays: Add ArrayField and ArrayInput

### LINKED ISSUE #1662: CollectionField iterator and ReferenceCollectionInput
**What you were expecting:**

I want to populate a list iterator from a source of a collection. Not an array. And furthermore, I don't 
want to use a reference field. All of the data I need is contained in the collection. 

```
            <ReferenceArrayField label="Groups" source="roles">
                <SingleFieldList>
                    <ChipField source="name" />
                </SingleFieldList>
            </ReferenceArrayField>
```

Network Response:

```
[{
	"email": "bo.hatchet@mail.com",
	"id": 6,
	"roles": [{
		"id": 3,
		"name": "reg role"
	}]
}, {
	"email": "bo.ratchet@mail.com",
	"id": 7,
	"roles": [{
		"id": 1,
		"name": "admin"
	},{
		"id": 2,
		"name": "sales"
	}]
}]
```

**What happened instead:**

Currently there seems to be no iterator field to work on a source that is a collection. Only ReferenceArrayField and ReferenceManyField which both require and utilize a reference endpoint.

fzaninotto:
- [x] Add `<ArrayField>` component
- [x] Document `<ArrayField>` component
- [ ] Unit Test `<ArrayField>` component
- [x] Add `<ArrayInput>` component
- [ ] Document `<ArrayInput>` component
- [ ] Unit Test `<ArrayInput>` component

Closes #1662 

ArrayField preview:

image

ArrayInput preview:

kapture 2018-04-03 at 9 09 12

### DIFF
--- a/packages/ra-ui-materialui/src/field/ArrayField.js
+++ b/packages/ra-ui-materialui/src/field/ArrayField.js
@@ -1,0 +1,136 @@
+import { Component, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import get from 'lodash/get';
+import pure from 'recompose/pure';
+
+/**
+ * Display a collection
+ * 
+ * Ideal for embedded arrays of objects, e.g.
+ * {
+ *   id: 123
+ *   tags: [
+ *     { name: 'foo' },
+ *     { name: 'bar' }
+ *   ]
+ * }
+ * 
+ * The child must be an iterator component
+ * (like <Datagrid> or <SingleFieldList>).
+ * 
+ * @example Display all the backlinks of the current post as a <Datagrid>
+ * // post = {
+ * //   id: 123
+ * //   backlinks: [
+ * //       {
+ * //           date: '2012-08-10T00:00:00.000Z',
+ * //           url: 'http://example.com/foo/bar.html',
+ * //       },
+ * //       {
+ * //           date: '2012-08-14T00:00:00.000Z',
+ * //           url: 'https://blog.johndoe.com/2012/08/12/foobar.html',
+ * //       }
+ * //    ]
+ * // }
+ *     <ArrayField source="backlinks">
+ *         <Datagrid>
+ *             <DateField source="date" />
+ *             <UrlField source="url" />
+ *         </Datagrid>
+ *     </ArrayField>
+ *
+ * @example Display all the tags of the current post as <Chip> components
+ * // post = {
+ * //   id: 123
+ * //   tags: [
+ * //     { name: 'foo' },
+ * //     { name: 'bar' }
+ * //   ]
+ * // }
+ *     <ArrayField source="tags">
+ *         <SingleFieldList>
+ *             <ChipField source="name" />
+ *         </SingleFieldList>
+ *     </ArrayField>
+ *
+ * If you need to render a collection in a custom way, it's often simpler
+ * to write your own component:
+ * 
+ * @example
+ *     const TagsField = ({ record }) => (
+ *          <ul>
+ *              {record.tags.map(item => (
+ *                  <li key={item.name}>{item.name}</li>
+ *              ))}
+ *          </ul>
+ *     )
+ *     TagsField.defaultProps = { addLabel: true };
+ */
+class ArrayField extends Component {
+    constructor(props) {
+        super(props);
+        this.state = props.record
+            ? this.getDataAndIds(props.record, props.source)
+            : {
+                  data: {},
+                  ids: [],
+              };
+    }
+
+    componentWillReceiveProps(nextProps, prevProps) {
+        if (nextProps.record !== prevProps.record) {
+            this.setState(
+                this.getDataAndIds(nextProps.record, nextProps.source)
+            );
+        }
+    }
+
+    getDataAndIds(record, source) {
+        const list = get(record, source);
+        return {
+            data: list.reduce((prev, item) => {
+                prev[JSON.stringify(item)] = item;
+                return prev;
+            }, {}),
+            ids: list.map(JSON.stringify),
+        };
+    }
+
+    render() {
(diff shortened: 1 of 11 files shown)


djhi:
I advise against using icon only buttons in general. I'd rather have buttons with explicit texts. Here, maybe just the plus button should be a button with text and we may keep the minus one. I would make it a trash or cross icon though

docs/Inputs.md
```diff
@@ -160,17 +197,18 @@ import { AutocompleteInput, ReferenceInput } from 'react-admin'
...
-`optionValue` | Optional | `string` | `id` | Fieldname of record containing the value to use as input value 
-`optionText` | Optional | <code>string &#124; Function</code> | `name` | Fieldname of record to display in the suggestion item or function which accepts the currect record as argument (`(record)=> {string}`)
-`setFilter` | Optional | `Function` | null | A callback to inform the `searchText` has changed and new `choices` can be retrieved based on this `searchText`. Signature `searchText => void`. This function is automatically setup when using `ReferenceInput`. 
-`suggestionComponent` | Optional | Function | `({ suggestion, query, isHighlighted, props }) => <div {...props} />` | Allows to override how the item is rendered. 
+
+| Prop | Required | Type | Default | Description |
+| ---|---|---|---|--- |
+| `choices` | Required | `Object[]` | - | List of items to autosuggest |
+| `resource` | Required | `string` | - | The resource working on. This field is passed down by wrapped components like `Create` and `Edit`.   |
+| `source` | Required |  `string` | - | Name of field to edit, it's type should correspond to the type retrieved from `optionValue`  |
```
djhi: s/correspond to/match/

fzaninotto:
Switching to Ready For Review

packages/ra-ui-materialui/package.json
```diff
@@ -42,6 +35,7 @@
         "react-dropzone": "~4.0.1",
         "react-redux": "~5.0.4",
         "react-router-dom": "~4.2.2",
+        "react-transition-group": "^2.2.1",
```
fzaninotto: Here I'm using `^` to be sure I get the same version as material-ui

djhi:
How does it look on mobile ?

packages/ra-ui-materialui/src/field/ArrayField.js
```diff
@@ -0,0 +1,140 @@
...
+ *     const TagsField = ({ record }) => (
+ *          <ul>
+ *              {record.tags.map(item => (
+ *                  <li key={item.name}>{item.name}</li>
+ *              ))}
+ *          </ul>
+ *     )
+ *     TagsField.defaultProps = { addLabel: true };
+ */
+export class ArrayField extends Component {
```
djhi: As this component does not include any UI, shouldn't it be in `ra-core` ?

packages/ra-ui-materialui/src/input/ArrayInput.js
```diff
@@ -0,0 +1,108 @@
...
+ * A form iterator is a component accepting a fields object
+ * as passed by redux-form's <FieldArray> component, and defining a layout for
+ * an array of fields. For instance, the <SimpleFormIterator> component
+ * displays an array of fields in an unordered list (<ul>), one sub-form by
+ * list item (<li>). It also provides controls for adding and removing
+ * a sub-record (a backlink in this example).
+ * 
+ * @see https://redux-form.com/7.3.0/examples/fieldarrays/
+ */
+export class ArrayInput extends Component {
```
djhi: Same

fzaninotto: This one does

```js
import { FormControl } from 'material-ui/Form';
import { InputLabel } from 'material-ui/Input';
```

packages/ra-ui-materialui/src/field/ArrayField.js
fzaninotto: I though about that, too. It's just that we don't have the folder structure in `ra-core` to accommodate plain fields (not just controllers). I was being lazy, I'll do it.

fzaninotto:
It looks OK on Mobile

image

djhi:
Kinda hard to distinguish items on mobile: too many line separator imo. Maybe adding a bit more space around fields and eventually items inside the array input would make it more readable

packages/ra-ui-materialui/src/field/ArrayField.js
fzaninotto: But the `ArrayField.spec.js` file does use many components from `ra-ui-material-ui`. In fact, it tests the whole integration with an iterator. So it's not easy to move the component and its tests to `ra-core`.

We need to think more about what goes into core and what goes into ui. But in a future PR.

fzaninotto:
I think the whole idea of ArrayField isn't good for mobile UIs. Whatever the look and feel, it will be clunky IMO. This type of component is really suited for larger screens.

packages/ra-ui-materialui/src/field/ArrayField.js
djhi: ok

nitoygo:
ReferenceInput not working with ArrayInput.
How to do something like the following?
```
        <ArrayInput source="post">
          <SimpleFormIterator>
            <ReferenceInput source="commentedOn" reference="user">
              <AutocompleteInput optionText="username" />
            </ReferenceInput>
          </SimpleFormIterator>
        </ArrayInput>
```

zavadpe:
@fzaninotto I can confirm that use case @nitoygo described is not working. I guess it's related to #1783.

fzaninotto:
Yes indeed, and #1792 should fix it.

PR #1723, opened 2018-04-04
Added slovak translations for react-admin

zavadpe:
Added ra-language-slovak translation package for react-admin
and removed aor-language-slovak from previous version list.


fzaninotto:
Thanks!

PR #1724, opened 2018-04-05
Impossible to call a TextInput with an input attribut

JacquesBonet:
Impossible to call TextInput with input attribute (to put a value to the TextInput).

My call:

```
<TextInput
                        label="Tifiz Number"
                        source="name"
                        input={{
                            value: value.name,
                        }}
                        translate={translate}
                        validate={validateRequired}
                    />
```

I have the following error message:
```

Error: The TextInput component wasn't called within a redux-form <Field>. Did you decorate it and forget to add the addField prop to your component? See https://marmelab.com/react-admin/Inputs.html#writing-your-own-input-component for details.
TextInput.render
/home/oem/Documents/software/ticatag/ticatag-front/packages/ra-ui-materialui/lib/input/TextInput.js:109
  106 |     rest = (0, _objectWithoutProperties3.default)(_props, ['className', 'input', 'isRequired', 'label', 'meta', 'options', 'resource', 'source', 'type', 'value']);
  107 | 
  108 | if (typeof meta === 'undefined') {
> 109 |     throw new Error("The TextInput component wasn't called within a redux-form <Field>. Did you decorate it and forget to add the addField prop to your component? See https://marmelab.com/react-admin/Inputs.html#writing-your-own-input-component for details.");
  110 | }  
```

My TextInput is call inside a redux-form, because call inside a SimpleForm (SimpleForm call redux-form)

image

The link given on the error message (https://marmelab.com/react-admin/Inputs.html#writing-your-own-input-component), redirect to PAGE NOT FOUND

With my modiication, I can set a value to my TextInput.

### DIFF
--- a/packages/ra-ui-materialui/src/input/TextInput.js
+++ b/packages/ra-ui-materialui/src/input/TextInput.js
@@ -48,12 +48,9 @@ export class TextInput extends Component {
             type,
             ...rest
         } = this.props;
-        if (typeof meta === 'undefined') {
-            throw new Error(
-                "The TextInput component wasn't called within a redux-form <Field>. Did you decorate it and forget to add the addField prop to your component? See https://marmelab.com/react-admin/Inputs.html#writing-your-own-input-component for details."
-            );
-        }
-        const { touched, error } = meta;
+
+        const touched = meta ? meta.touched : false
+        const error = meta ? meta.error : false
 
         return (
             <TextField


fzaninotto:
Why would you want 'to put a value to the TextInput'? I think you're looking for `defaultValue`. Otherwise, the value come from the record grabbed from the data provider.

`TextInput` won't allow to override the `value` (that would mess up with `redux-form` badly). But you can still write your own input components.

JacquesBonet:
Why I need that, because I'm developing a wizard: Not  all fields are visually on the same screen.

JacquesBonet:
In TextInput.spec.js, Its not me who write that:

   ```
 it('should use a mui TextField', () => {
        const wrapper = shallow(
            **<TextInput {...defaultProps} input={{ value: 'hello' }} />**
        );
        const TextFieldElement = wrapper.find('TextField');
        assert.equal(TextFieldElement.length, 1);
        assert.equal(TextFieldElement.prop('value'), 'hello');
        assert.equal(TextFieldElement.prop('type'), 'text');
    });
```

Have a good day

fzaninotto:
in `TextInput.spec.js`, notice that we do:

`import { TextInput } from './TextInput';`

Instead of

`import TextInput from './TextInput';`

This is a unit test, testing the non-decorated component. In a real app, you are using the component wrapped in a redux-form `<Field>`. In this case, it's not advised to override the `input` prop.

PR #1727, opened 2018-04-05
[RFR] Update Tutorial for better first contact

fzaninotto:
The tutorial was a bit hard to follow. I made shorter sections, added a bit more explanations, and removed non-related content.

Also, I added the ability to run an `<Admin>` with no child (just displays a confirmation screen). IMO, it provides a better developer experience for first time users, by allowing baby steps.

### DIFF
--- a/packages/ra-core/src/CoreAdminRouter.js
+++ b/packages/ra-core/src/CoreAdminRouter.js
@@ -10,6 +10,12 @@ import { isLoggedIn } from './reducer';
 import { userLogout } from './actions/authActions';
 import RoutesWithLayout from './RoutesWithLayout';
 
+const welcomeStyles = {
+    width: '50%',
+    margin: '40vh',
+    textAlign: 'center',
+};
+
 export class CoreAdminRouter extends Component {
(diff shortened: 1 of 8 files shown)


packages/ra-core/src/CoreAdminRouter.js
```diff
@@ -93,15 +99,27 @@ export class CoreAdminRouter extends Component {
...
 
-        let childrenToRender =
-            typeof children === 'function' ? childrenFromState : children;
+        if (typeof children !== 'function' && !children) {
+            return (
+                <div style={welcomeStyles}>
+                    React-admin is properly configured.<br />
+                    Now you can add a first &lt;Resource&gt; as child of
+                    &lt;Admin&gt;.
+                </div>
```
djhi: Why not also add a code snipper ? Great welcome message though :+1:

fzaninotto: Because the code for that message is bundled to production, even though it's only useful in development. I don't want to add too much weight to the final production bundle.

PR #1728, opened 2018-04-05
[RFR] Refresh after success of optimistic action

### LINKED ISSUE #1626: Undoable edit and refresh
When you refresh an edit resource you'll get unexpected behaviour using `undoable` edits, see the behaviour below: 

There should be a check if the resource is empty.

### LINKED ISSUE #1725: The Datagrid doesn't automatically refresh data after a save
That's a regression. It used to refresh automatically after redirection from a edit / create form.

* react-admin version: 2.0.0-beta4
* reproduced on simple example

image

fzaninotto:
Fixes #1725
Fixes #1626

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/list/total.js
+++ b/packages/ra-core/src/reducer/admin/resource/list/total.js
@@ -1,4 +1,5 @@
 import {
+    CRUD_GET_ONE_SUCCESS,
     CRUD_GET_LIST_SUCCESS,
     CRUD_DELETE_OPTIMISTIC,
     CRUD_DELETE_MANY_OPTIMISTIC,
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/reducer/admin/resource/list/total.js
```diff
@@ -8,6 +9,9 @@ export default resource => (previousState = 0, { type, payload, meta }) => {
     if (!meta || meta.resource !== resource) {
         return previousState;
     }
+    if (type === CRUD_GET_ONE_SUCCESS) {
+        return previousState == 0 ? 1 : previousState;
```
fzaninotto: This is to avoid the "no results found" after a redirection from the Edit view if the user didn't display the list before. The bug was an artefact of optimistic rendering.

packages/ra-core/src/sideEffect/undo.js
```diff
@@ -34,7 +34,10 @@ export function* handleUndoRace(undoableAction) {
         // if not cancelled, redispatch the action, this time immediate, and without success side effect
         yield put({
             ...action,
-            meta: metaWithoutSuccessSideEffects,
+            meta: {
+                ...metaWithoutSuccessSideEffects,
+                onSuccess: { refresh: true },
```
fzaninotto: The forced refresh allows to make sure the displayed data is up to date with the server data after an optimistic operation.

PR #1745, opened 2018-04-13
[RFR] Add support for defaultValues on alwaysOn inputs

### LINKED ISSUE #1740: Allow AlwaysOn filters with default value
First off, I love admin-on-rest! I created an awesome administrative portal for our REST API in no-time, so huge thanks to you guys!

I have one issue however I can't find a workaround for. I want to have this:

- A date filter that is always visiable (alwaysOn), so user don't have to click 'add filter' first
- The filter must have a default value (some date value, for example current date like this: new Date()
- The user must be able to change the filter value (that's why I use 'filters' and not the permanent 'filter' on the grid)

So, I created a a simple grid, showing a count for a certain subject with 2 date filters: 

`

    const GridFilters = (props) => (<Filter {...props}>

		<DateInput source="to_date" label="To date" alwaysOn options={{
				mode: "portrait",
				locale: "nl"
			}} />		
	
		<DateInput source="initialfilter_from_date" label="From date" defaultValue={props.date} alwaysOn options={{
				mode: "portrait",
				locale: 'nl',
				defaultDate: props.date,
			}} />					
    </Filter>);
    <List title="Dategrid" perPage={50} filters={<GridFilters date={new Date()} />} {...props}>
        <Datagrid >
           <TextField source="subject" />
           <NumberField source="count" />
	</Datagrid>
</List>`

My app has a custom menu with multiple pages. If I directly go to the url that this page is visible on (for example: http://localhost:3001/#/admin/dategrid , the default value is set in the input and applied in the filter. 

However, if I switch to another page in my menu and switch back to this page, the default filter is not set in the input nor in the filters sent to my API. In my custom restClient the params.filters are not filled. There are no errors in the console and if I press F5 the filter is set again.

Maybe needed info for reproduction info: the grid is added to my <Admin.. component as a resource ad passed to the 'list' property of a resources.

Ps. an additional bug I found with the DateInput is that setting a defaultValue like this: new Date() works, however, set a custom date like this: new Date(2018,1,1,0,0,0) doesn't work. Don't know why, but the filter does not get applied. I have worked around that by manually checking inputs fields with default values in my restClient and manually adding them to the querystring, but it doesn't work out of the box. This issue maybe better belongs in a seperate issue however since it seems not really related.

Thanks!

Ps. I use admin-on-rest 1.4.0 according to my package.json

fzaninotto:
Closes #1740 

For technical reasons, react-admin does not accept children of `<Filter>` having both a `defaultValue` and `alwaysOn`. To set default values for always on filters, use the `filterDefaultValues` prop of the `<List>` component instead.

```jsx
// in src/posts.js
const PostFilter = (props) => (
    <Filter {...props}>
        <TextInput label="Search" source="q" alwaysOn />
        <BooleanInput source="is_published" alwaysOn />
        <TextInput source="title" defaultValue="Hello, World!" />
    </Filter>
);

    <List {...props} filters={<PostFilter />} filterDefaultValues={{ is_published: true }}>
        ...
    </List>
);
```


djhi:
Great but I fear that having two ways for adding filter default values might be very confusing. I'd rather only keep this new prop

fzaninotto:
We need both, one that can be overridden by the user (that's `filterDefaultValues`), and one that cannot be overridden (that's `filter`). This is explained in the documentation:

```js
const filterSentToDataProvider = { ...filterDefaultValues, ...filterChosenByUser, ...filters };
```

Issue #1361, opened 2018-01-09
ReferenceField: allow specifying sort column


tomaszmadeyski:
Currently when you add `ReferenceField` like this

```
<ReferenceField label="Post" source="post_id" reference="posts">
      <TextField source="title" />
</ReferenceField>
```

and once you click on the column header to sort it will be sorted by `post_id` and not `title`. 

**Fix for this would be allowing to specify which column should be passed in `sort` specification in query string.**

It can be seen on [official demo](https://marmelab.com/admin-on-rest-demo) - go to orders and click on `Customer` column. 

![ezgif com-crop](https://user-images.githubusercontent.com/1555694/34724677-ffa08e3a-f54e-11e7-8fe1-501909211658.gif)

So in the end such sorting is very confusing from user's perspective since sorting by id is not relevant information.

There's already (not mine) [StackOverflow question](https://stackoverflow.com/q/47959129/2393077) about that (with no answers)

style2joy:
thanks for bringing it here.. hope it gets some attention.

afilp:
I would also need a fix for this, sorting by the irrelevant id shows wrong results on the Datagrid.

Chusynuk:
Hey I have the same problem, trying to sort this fields using the sort prop from admin-on-rest documentation.

should I use reference on each Texfield tag? thx

``` 
<List 
{...props}
      sort={{ field: "username", order: "DESC" }}
      title="Admin Users"
    >
      <Datagrid>
        <TextField source="username" label="username" />
        <TextField source="role" label="role" />
        <DateField source="created_at" label="created_at" type="date" />
      </Datagrid>
    </List>
```

ArneZsng:
Hi, just to be clear: Nobody is currently working on this, right? If not, I would probably work on a PR to get this functionality in. I assume it would make the most sense to add this to `next`?

fzaninotto:
I confirm that nobody from my team is currently working on it (we assign issues when we start working on them). New features should be PRed against `next`. 

I'm looking forward to reading your contribution!

ArneZsng:
@fzaninotto I made a proposal in #1746, please let me know what your thoughts are.

ArneZsng:
@zifnab87 My approach (now found in https://github.com/marmelab/react-admin/pull/1755) will just swap out the sorting parameter for the GET request if `sortBy` is present on a reference field.

Example: If the source of you reference field is ID, that source information is used for two things:
* Constructing the link of each row
[...]

PR #1746, opened 2018-04-13
Add optional sortBy to ReferenceField

ArneZsng:
#### Features
:rocket: Add optional sortBy to ReferenceField to sort by other than source of reference

I could think of two ways to solve issue #1361:
- Add an optional field to ReferenceField that is used to generate the link in case it is present and let it always sort by `source`.
- Add an optional field to ReferenceField that is used to sort by in case it is present and let the link always be generated by `source`.

I opted for the second option here but implementing the first is also possible. Please let me know what you think and if you have any preferences with regards to the API of this project.


fzaninotto:

PR #1755, opened 2018-04-17
Add optional sortBy to ReferenceField

ArneZsng:
Reopen of https://github.com/marmelab/react-admin/pull/1746

#### Features
:rocket: Add optional sortBy to ReferenceField to sort by other than source of reference

I could think of two ways to solve issue #1361:
- Add an optional field to ReferenceField that is used to generate the link in case it is present and let it always sort by `source`.
- Add an optional field to ReferenceField that is used to sort by in case it is present and let the link always be generated by `source`.

I opted for the second option here but implementing the first is also possible. Please let me know what you think and if you have any preferences with regards to the API of this project.

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -901,8 +901,7 @@ export const UserList = (props) => (
 );
 ```
 
-<<<<<<< HEAD
-**Tip**: In such custom fields, the `source` is optional. React-admin uses it to determine which column to use for sorting when the column header is clicked.
+**Tip**: In such custom fields, the `source` is optional. React-admin uses it to determine which column to use for sorting when the column header is clicked. In case you use the `source` property for additional purposes, the sorting can be overridden by the `sortBy` property on any `Field` component.
 
 ## Adding Label To Custom Field Components In The Show View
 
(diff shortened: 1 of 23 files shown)


ArneZsng:
@fzaninotto I'd be happy to get your feedback on this.

packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
```diff
@@ -35,9 +35,12 @@ export const DatagridHeaderCell = ({
                 enterDelay={300}
             >
                 <TableSortLabel
-                    active={field.props.source === currentSort.field}
+                    active={
+                        (field.props.sortBy || field.props.source) ===
```
fzaninotto: Hard to read in this shape, perhaps you can invert the test to have the `||` in the rightmost part?

ArneZsng:
@fzaninotto I added `sortBy` to the PropTypes of all `Field` components and added some docs. Can you take another look at it?

docs/Fields.md
```diff
@@ -491,6 +491,14 @@ To change the link from the `<Edit>` page to the `<Show>` page, set the `linkTyp
 </ReferenceField>
 ```
 
+By default, `<ReferenceField>` is sorted by its `source`. To specify another attribute to sort by, set the `sortBy` prop to the according attribute's name.
+
+```jsx
+<ReferenceField label="User" source="userId" reference="users" sortBy="name">
```
fzaninotto: Not a good example imo, the `sortBy` prop should be something like `user.name`. That is, if the API understands it...

ArneZsng:
@fzaninotto Changed the base and updated the docs. Let me know if this works now.

swrobel:
@fzaninotto this would be amazing! Is there anything left that's preventing you from approving?

fzaninotto:
Thanks!

PR #1763, opened 2018-04-20
Update Makefile - Fix building demo

afilp:


djhi:
That's actually why we started naming your files with kebab-case on a project with @alexisjanvier

PR #1791, opened 2018-04-27
fix NumberInput step prop

Klucherev:
as update to material-ui, prop step no longer exist

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.js
+++ b/packages/ra-ui-materialui/src/input/NumberInput.js
@@ -73,7 +73,7 @@ export class NumberInput extends Component {
                 margin="normal"
                 error={!!(touched && error)}
                 helperText={touched && error}
-                step={step}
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/input/NumberInput.js
```diff
@@ -73,7 +73,7 @@ export class NumberInput extends Component {
                 margin="normal"
                 error={!!(touched && error)}
                 helperText={touched && error}
-                step={step}
+                inputProps={{step}}
```
djhi: No, it should be in `inputProps`. See the material-ui doc: https://material-ui-next.com/api/text-field/

djhi: > inputProps => Properties applied to the native input element.

So here, we set the `step` attribute which is supported by native inputs

PR #1803, opened 2018-05-04
[RFR] Fix changing id type in accumulate

byymster:
## Issue that fixed:
When you use GraphQl, id as Int and ReferenceField in DataGrid you get error:
image

* Object.keys changes type of id to string 
* Object.values don't change type
* Uniqueness not affected

### DIFF
--- a/packages/ra-core/src/sideEffect/accumulate.js
+++ b/packages/ra-core/src/sideEffect/accumulate.js
@@ -15,11 +15,11 @@ const addIds = (resource, ids) => {
         debouncedIds[resource] = {};
     }
     ids.forEach(id => {
-        debouncedIds[resource][id] = true;
+        debouncedIds[resource][id] = id;
     }); // fast UNIQUE
 };
 const getIds = resource => {
-    const ids = Object.keys(debouncedIds[resource]);
+    const ids = Object.values(debouncedIds[resource]);
     delete debouncedIds[resource];
     return ids;
 };


byymster:
Hi @fzaninotto.
I'm using `ra-data-graphcool` but this issue can be reproduced in any graphQL data provider.
I have old project with int id's in database.
My graphQl schema looks like:
```graphql
type Query {
  allRedirects(first: Int = 20, skip: Int = 0, orderBy: String, filter: DefaultFilter): [Redirect]
}

type Redirect {
  id: Int!
  ...........
  shortDomain: ShortDomain
}

type ShortDomain {
  id: Int!
  ...........
}
```

I have list view for `Redirect`:
```jsx
<List
      {...props}
      sort={{ field: "updated", order: "DESC" }}
      title={"Redirects"}
    >
      <Datagrid>
        <TextField source={"id"} />
        <ReferenceField
          source="shortDomain.id"
          reference="ShortDomain"
          linkType="show"
        >
          <TextField source={"domain"} />
        </ReferenceField>
        //...........
         <ShowButton />
        <EditButton />
      </Datagrid>
    </List>
```

So when Datagrid renders it's accumulate id's of `ShortDomain` and then use them to get information for all references at once:
```graphql
query allShortDomains($filter: DefaultFilter) {
  items: allShortDomains(filter: $filter) {
    id
    ...........
  }
  total: _allShortDomainsMeta(filter: $filter) {
    count
    __typename
  }
}
```
with variables:
```js
{filter: {id_in: ["2", "3"]}}
```
`id_in: ["2", "3"]` - generated by accumulate and it's strings because Object.keys returns strings, I propose to use Object.values because it's don't affect types - that's critical for GraphQL data provider.

byymster:
@fzaninotto This information is enough for you?

fzaninotto:
Thanks!

PR #1808, opened 2018-05-08
Next

kopax:


djhi:
Can you please explain what you are trying to achieve in this pull request ? There are more than 250 commits, 539 files changed...

At the minimum, please add a description, otherwise it will be closed

kopax:
Sorry wrong merge target, I lost the track of it. This remote is using our react-intl requirement. I still think we should provide the TranslationProvider to the lib using a props. I could do it and maintain it, instead of doing ours.

PR #1812, opened 2018-05-08
[RFR] Split example components into standalone files

fzaninotto:
The previous example code structure encouraged the bad practice of packing several components into a single file.

This PR makes the example code much more readable and maintainable, and introduces the recommended pattern for importing resource components and passing them to the `<Admin>` app:

```js
import posts from './posts';

render(
    </Admin>,
    document.getElementById('root')
);
```


fzaninotto:
Note for reviewers: no code was added or deleted, it's just a reorganization.

PR #1822, opened 2018-05-11
[RFR] Nicer dashboard in demo example

fzaninotto:
Before:

After:


djhi:
Not sure it's nicer but both look nice :+1:

canhlinh:
cool

PR #1842, opened 2018-05-21
Allow override of link and cache parameter in ra-data-graphql

terkiterje:
I was having a hard time trying to set the Authorization header with ra-data-graphql-simple dataprovider. In buildApolloClient.js in ra-data-graphql the provided cache and link parameters in clientOptions are effectively ignored and setting link causes Apollo to complain about missing cache with "In order to initialize Apollo Client, you must specify link & cache properties on the config object.". This PR fixes the issue for me.

It seems to me like a simple mistake and not intended behavior, since "otherOptions" will never contain link or cache, so finalLink and finalCache will always be initialized to undefined here.


fzaninotto:
ping @djhi

PR #1849, opened 2018-05-23
Support hosting react-admin from a sub-directory

onpaws:
react-router supports hosting from a non-root directory of a web host using a prop called `basename`.
The idea is to expose this prop so users[1] can optionally configure per recommended guidelines[2], [3].

This proposal is intended to be a starting point for this discussion - not "final".
`<Admin>` seems like the most obvious place to put this but of course let's chat - what makes sense to you?

[1] https://stackoverflow.com/questions/50402692/mount-react-admin-under-admin
[2] https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string
[3] https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#building-for-relative-paths

### DIFF
--- a/packages/ra-core/src/CoreAdmin.js
+++ b/packages/ra-core/src/CoreAdmin.js
@@ -23,6 +23,7 @@ const CoreAdmin = ({
     customReducers = {},
     customSagas = [],
     customRoutes = [],
+    basename,
     dashboard,
     history,
     menu,
@@ -69,7 +70,7 @@ const CoreAdmin = ({
     return (
         <Provider store={store}>
             <TranslationProvider>
-                <ConnectedRouter history={routerHistory}>
+                <ConnectedRouter basename={basename} history={routerHistory}>
                     <Switch>
                         <Route
                             exact


fzaninotto:
👍 

Can you document this `<Admin>` prop in the relevant doc chapter?

onpaws:
No problem, Github is showing me this warning:
image
But i'll just go ahead and switch it anyway.

onpaws:
I'm interested to hear how this works for folks that aren't me.

@fzaninotto do you recommend any places where react-admin users might be willing to test out new features? e.g. Slack etc?

swrobel:
Is there any way to install this from github? If so, I can certainly test.

onpaws:
Yes, there is,
Because of the way GitHub handles project renames, you'll likely have to
clone a repo called admin-on-rest from my account.
I'll post back with instructions.

swrobel:
This actually isn't the correct solution. When you aren't using BrowserHistory (ie: providing your own history), the `basename` [needs to be provided](https://github.com/ReactTraining/react-router/issues/6083#issuecomment-380801810) to [`createHistory()`](https://github.com/ReactTraining/history#using-a-base-url).

react-admin could still be changed to allow basename to be provided, but it should be passed to [its createHistory call](https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/CoreAdmin.js#L54) rather than attempting to do it this way (which won't work based on my testing).

onpaws:
Thanks for checking it out @swrobel!
Last year when I first tried admin-on-rest I noticed the # in the URL which comes from [this](https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/CoreAdmin.js#L5) flavor of the history singleton.

I speculate admin-on-rest may have been originally targeting pre IE11 browsers.
I wonder if it's time to consider migrating to BrowserHistory these days? Obligatory [caniuse](https://caniuse.com/#feat=history)

onpaws:
@fzaninotto what do you think?

onpaws:
I went ahead and migrated this PR to BrowserHistory. 
@swrobel curious hwo this works for you.

swrobel:
@onpaws I doubt that will be accepted as [BrowserRouter prevents providing a custom history prop](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/BrowserRouter.js#L23)

onpaws:
OK. While I investigate I assume it'd be desirable to drop the `#` in URLs

swrobel:
@onpaws sorry, I should've looked at your code before you commented. This should be fine. I confused `BrowserHistory` with `BrowserRouter`.

onpaws:
Understood. I was a bit confused thanks for posting back :)

onpaws:
So according to
https://marmelab.com/react-admin/Admin.html#history
the hashRouter supports working without a server.
That's nice I guess, but the more I think about it, I wonder for how many `react-admin` users this is important and worth the (minor) inconvenience of hash URLs.

Don't want to sound presumptuous -- to me it just seems like more users may prefer to use the app without the # in the URL. 
What would be the objections to keeping the history prop on `<Admin>` but swapping the default to BrowserHistory?

docs/Admin.md
```diff
@@ -442,6 +443,17 @@ const Foo = () => (
 export default Foo;
 ```
 
+## `basename`
+
+If you need to host react-admin in a subdirectory on your web host, e.g.
+
+`http://example.com/admin/`
+
+please specify a [basename](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string) via the basename prop to teach react-router..
```
fzaninotto: Remove "please", start the line by a capital letter, put basename between backticks

By the way, this sentence doesn't make sense. "To teach react-router": to teach what? And readers probably don't know about react-router so far in the documentation.

docs/Admin.md
```diff
@@ -442,6 +443,17 @@ const Foo = () => (
...
+
```
fzaninotto: Missing example

packages/ra-core/src/CoreAdmin.js
```diff
@@ -2,7 +2,7 @@ import React, { createElement } from 'react';
 import PropTypes from 'prop-types';
 import { createStore, compose, applyMiddleware } from 'redux';
 import { Provider } from 'react-redux';
-import createHistory from 'history/createHashHistory';
+import createHistory from 'history/createBrowserHistory';
```
fzaninotto: no, this is a BC break, please revert

fzaninotto:
react-admin uses a hash router by default. If you want to use another history, pass it to the admin using the `history` prop.

As for knowing why, it's simply because we host our SPAs on static servers. Is it a general practice? Is the opposite a general practice? I don't know, but since we built react-admin for our usage first that's the default behavior.

docs/Admin.md
onpaws: Thanks for the feedback! 
I'm still reading through existing docs to grok the style and this isn't finished yet. I'll incorporate your instructions

onpaws:
OK, reverting the BrowserHistory change.

stefsava:
I do not think it is necessary to change the code, just extend the documentation.
after
```
import createHistory from 'history/createBrowserHistory';

const history = createHistory();

const App = () => (
    <Admin history={history}>
        ...
    </Admin>
);
```
add

If you need to host react-admin in a subdirectory on your web host, e.g.
`http://example.com/admin/`
add basename prop in createHistory
```
const history = createHistory({ basename: 'admin' });
```
as is documented at [basename](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string).

PR #1860, opened 2018-05-26
[RFR] Fix SelectArrayInput Chip label should render same as selected menu item option

jeromemacias:
**Steps to reproduce**

- use `SelectArrayInput` with props `optionText` containing a function or a React element
- select one element: `Chip` will be empty

I don't know how to test this fix.


fzaninotto:
Thanks!

PR #1863, opened 2018-05-28
[RFR] [ra-input-rich-text] Remove necessity to use css loader

fzaninotto:
CSS loaders aren't always working... they require a specific loader, which is included in create-react-app but not everywhere. For this reason, a basic codesandbox created based on the simple example fails to display the rich text input (cf https://codesandbox.io/s/pw74xww3p0).

But since we have JSS with material-ui, it's possible to use JSS for styling Quill.

This is backwards compatible, since it should not change the resulting style. It just simplifies the bundler configuration required.


fzaninotto:
Back to WIP, there seems to be a problem with the hyperlink popup

PR #1879, opened 2018-06-05
[RFR] Use CSSTransition instead of opacity to animate bulk actions button

### LINKED ISSUE #1795: One may click 'bulk Delete' button, even though options for deletion is not selected.
**What you were expecting:**

Its impossible to click on 'bulk delete' button, without selecting options to delete.

**What happened instead:**

Its possible to click on 'bulk delete' button, without selecting options to delete.

**Steps to reproduce:**

1. In a List view, don't select any options.
2. Click on area where 'bulk delete' action is placed.
[example](http://joxi.ru/Rmz1ERgSW8eZdr): 

**Other information:**

This bug can be fixed, adding 'visibility: hidden' style for button.
**Environment**

* React-admin version: 2.0.0-RC2
* React version: 16.2.0
* Browser: Chrome (last)

fzaninotto:
Closes #1795

### DIFF
--- a/packages/ra-ui-materialui/src/list/BulkActions.js
+++ b/packages/ra-ui-materialui/src/list/BulkActions.js
@@ -1,5 +1,6 @@
 import React, { cloneElement, Children, Component } from 'react';
 import PropTypes from 'prop-types';
+import { CSSTransition } from 'react-transition-group';
 import Menu from '@material-ui/core/Menu';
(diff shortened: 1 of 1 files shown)


fzaninotto:
Back to WIP: the bulk button isn't displayed when a user comes back to a List page where they already selected rows

PR #1881, opened 2018-06-06
[RFR] Add ability to customize Layout components (menu, appbar, notification)

fzaninotto:
Also, deprecating the ability to customize the menu from the `<Admin>` component, as it makes more sense to use a custom Layout instead.

Your custom layout can simply extend the default `<Layout>` component if you only want to override the appBar, the menu, or the notification component. For instance:

```jsx
// in src/MyLayout.js
import { Layout } from 'react-admin';
import MyAppBar from './MyAppBar';
import MyMenu from './MyMenu';
import MyNotification from './MyNotification';

const MyLayout = (props) => <Layout 
    {...props}
    appBar={MyAppBar}
    menu={MyMenu}
    notification={MyNotification}
/>;

export default MyLayout;
```

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -31,7 +31,7 @@ Here are all the props accepted by the component:
 * [`title`](#title)
 * [`dashboard`](#dashboard)
 * [`catchAll`](#catchall)
-* [`menu`](#menu)
+* [`menu`](#menu) (deprecated)
 * [`theme`](#theme)
 * [`appLayout`](#applayout)
 * [`customReducers`](#customreducers)
@@ -152,6 +152,8 @@ const App = () => (
 
 ## `menu`
 
+**Tip**: This prop is deprecated. To override the menu component, use a [custom layout](#appLayout) instead.
+
 React-admin uses the list of `<Resource>` components passed as children of `<Admin>` to build a menu to each resource with a `list` component.
 
 If you want to add or remove menu items, for instance to link to non-resources pages, you can create your own menu component:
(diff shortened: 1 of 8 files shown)


docs/Admin.md
```diff
@@ -31,7 +31,7 @@ Here are all the props accepted by the component:
 * [`title`](#title)
 * [`dashboard`](#dashboard)
 * [`catchAll`](#catchall)
-* [`menu`](#menu)
+* [`menu`](#menu) (deprecated)
```
djhi: Why leaving it at all in the documentation then ?

docs/Admin.md
```diff
@@ -152,6 +152,8 @@ const App = () => (
 
 ## `menu`
 
+**Tip**: This prop is deprecated. To override the menu component, use a [custom layout](#appLayout) instead.
+
 React-admin uses the list of `<Resource>` components passed as children of `<Admin>` to build a menu to each resource with a `list` component.
 
 If you want to add or remove menu items, for instance to link to non-resources pages, you can create your own menu component:
```
djhi: Shouldn't this part be removed too ?

docs/Admin.md
```diff
@@ -270,9 +234,7 @@ For more details on predefined themes and custom themes, refer to the [Material
 
 ## `appLayout`
 
-If you want to deeply customize the app header, the menu, or the notifications, the best way is to provide a custom layout component. It must contain a `{children}` placeholder, where react-admin will render the resources. If you use material UI fields and inputs, it should contain a `<MuiThemeProvider>` element. And finally, if you want to show the spinner in the app header when the app fetches data in the background, the Layout should connect to the redux store.
-
-Use the [default layout](https://github.com/marmelab/react-admin/blob/master/packages/react-admin/src/mui/layout/Layout.js) as a starting point, and check [the Theming documentation](./Theming.html#using-a-custom-layout) for examples.
+If you want to deeply customize the app header, the menu, or the notifications, the best way is to provide a custom layout component through the `appLauout` prop.
```
djhi: s/appLauout/appLayout/

docs/Admin.md
fzaninotto: because it's deprecated, not removed. People using this prop would be surprised not to find the related documentaiton. We'll remove it in 3.0.

docs/Admin.md
fzaninotto: No, same comment as above

christiaanwesterbeek:
Does this break anything? Would like to have this in 2.0.4

djhi:
It does not break anything but as it adds new features, it will be in `2.1.0`

fredcomp3111:
Hello. Is the current version of react admin no longer accept appBar as props for <Layout/> component?
I tried:
```
const appLayout = (props) => <Layout
    {...props}
    appBar={MyAppBar}
    menu={myMenu}
/>;
```
But get warning and not being able to change the appbar: **_React does not recognize the `appBar` prop on a DOM element._**

fzaninotto:
version 2.1.0 does. If you think there is a bug, I suggest you make create a codesandbox to reproduce it, and open a new issue here.

PR #1883, opened 2018-06-06
[RFR] Add media to dashboard

fzaninotto:
The demo is our showcase, so it better be good lookin'. And what's more good lookin' than a good ol' beard? (pardon my accent, I'm French)


djhi:
^^

PR #1902, opened 2018-06-13
fix withStyles usage samples

stefsava:
- In withStyles usage sample add destructuring parameters with ({ classes, ...props })
- In sample where override the field header use headerClassName property instead of cellClassName


djhi:
Thanks!

PR #1912, opened 2018-06-15
[RFR] Pass the sort, pagination, and filters to the child of ReferenceInput

### LINKED ISSUE #1585: Pass 'total' property to ReferenceInput children
I want to use pagination in my custom children of ReferenceInput, but I haven't 'total' property there. 
Now in children of ReferenceInput I have 'choices' property and method 'setPagination'. To realize paginationh, I need to know 'total' property, like in List component

fzaninotto:
Closes #1585

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -677,7 +677,7 @@ You can tweak how this component fetches the possible values using the `perPage`
 
 ## `<ReferenceInput>`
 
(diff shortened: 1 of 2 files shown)


djhi:
Btw, it might also be a good idea to do the same for the `List` component and its child. We currently don't pass the `filters`

PR #1916, opened 2018-06-15
[RFR] Add validation to `<RichTextInput>`

### LINKED ISSUE #207: <RichTextInput> doesn't display validation errors
With the following edition form:

```js
    <Edit title={<PostTitle />} {...props}>
        <DisabledInput label="Id" source="id" />
        <RichTextInput source="body" validation={{ required: true }} />
    </Edit>
);
```

It's impossible to edit a post with no `body` - and there are no explanations.

fzaninotto:
Closes #207

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -3,6 +3,8 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Quill from 'quill';
 import { addField } from 'react-admin';
+import FormHelperText from '@material-ui/core/FormHelperText';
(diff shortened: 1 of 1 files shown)


packages/ra-input-rich-text/src/index.js
```diff
@@ -45,18 +47,23 @@ export class RichTextInput extends Component {
     }
 
     onTextChange = () => {
-        this.props.input.onChange(this.editor.innerHTML);
+        const value =
+            this.editor.innerHTML == '<p><br></p>' ? '' : this.editor.innerHTML;
```
fzaninotto: Quill doesn't like void. Removing the content of a quill editor results in a non-empty content (!).

PR #1923, opened 2018-06-19
[WIP] Editable Datagrid

fzaninotto:
Excel-like datagrid that replaces the Edit view, because that's what some of our customers want.

Added as a standalone package because the dependency weights 212kB minified.

- [ ] ~~`<List>` injects pagination-related props to children (`page`, `perPage`, `setPage`, `setPerPage`, `total`)~~(moved to #2028)
- [x] Integrate `ReactDataGrid` external component with List
- [x] hook up sorting
- [x] hook up pagination (sort of)
- [x] hook up row selection
- [x] hook up edition
- [ ] ~~Fix missing resize on menu minification~~ (known react-aditable-datagrid bug, see https://github.com/adazzle/react-data-grid/issues/916)
- [x] Fix selection loss after saving cell change
- [ ] Fix warnings / missing proptypes
- [ ] Add row
- [x] How do we handle deletion? via BulkDelete
- [ ] Add documentation

kapture 2018-06-19 at 9 18 04


djhi:
@RotatingFans Well, actually in this PR: https://github.com/marmelab/react-admin/pull/1923/files#diff-0f00f51285768f6e2170200382644ebaR27

RotatingFans:
I see. Guess I missed it.

afilp:
This feature would be great! Is there something we could do to help? I see that most of the work has been done. Thanks!

fzaninotto:
I don't think we'll be working on this one in the near future. If anybody in the community wants to take over and publish it as a standalone package, I'm sure the feature would gain traction.

RotatingFans:
After looking through the react-admin code, I believe there are a few ways to implement the ability to add a row: 

1. Make the create button open a dialog with certain fields, like a mini createForm. Relatively simple, but would require the user to manually pass the controller as a prop and somehow notify the listController to use it correctly.
2. Make the Editable datagrid a stateless component as controlled by the list controller. Easiest from the User perspective, put pretty ugly on the backend.
3. Make a new Listcontroller that implements the editable datagrid and other list features in one React component. This would both be simple from the user perspective and simple on the backend(albeit a bit redundant in parts).

However, we would have to do something if a user has a non editable field such as a primary key in the table. 
Possible solutions:
1. For Option 1, save the data when closing the dialog and update the datatable with the server response.
2. For Option 2 & 3, monitor when a new row is updated, run a create, get the returned primary key and continue to use updates. Relatively expensive in backend calls
3. For Options 2 & 3, Add an empty row(User cannot edit the primary key values). Let user make edits to whole data grid. Add a save button at the bottom. Probably the best choice. Saving as updates happen is excessive in most instances and hampers performance.

Lastly, in order to make this a ready to publish feature, I recommend the following additional changes based on my experience:

- Load a few pages below at a time(probably about 5x the number of rows that are on a page)
- Improve Lazy loading
- Improve responsiveness. This is annoying and requires a good bit of manual css modifications
- Implement MUI components into the editors and such. Not too difficult to implement

FYI, I implemented reactDataGrid into a project of my own as a custom component within react-admin. I plan to work on some of these issues, but I wanted feedback about the best way to approach the solution to keep consistent with the intended design of the application. If you want to see my implementation, you can look [here(its ugly atm)](https://gitlab.com/ABOS-Software/ABOS/blob/c536f0f196dc8d7621c1be29167a517ced65de40/client/src/resources/ProductsGrid.js). It does not use lazy loading or the list controller in any way(It is a component on a simple form).

fzaninotto:
We won't add this feature in the core. Please contact marmelab if you're interested in this feature.

PR #1925, opened 2018-06-19
Fix multiple broken links in DataProviders doc

alireza-ahmadi:
`react-admin/` was missing in multiple links in the Data Providers doc page. This pull request fixes those broken links.


fzaninotto:
Thanks!

PR #834, opened 2017-07-06
Support nested filter object

mtakayuki:
There are two problems when a deep property is specified in `Filter` like:

```javascript
const MyFilter = props => (
  <Filter {...props}>
    <TextInput source='foo.bar' label='Foo' />
  </Filter>
);


const MyList = props => (
  <List {...props} filters={<MyFilter />}
    ...
  </List>
);
```

* Invalid query string (E.g. `"foo=%5Bobject%20Object%5D"`) is generated for json server. This should b flatten like `"foo.bar=val"`.
* When the text input value is cleared, the filter value is not removed from filter parameters.

PR #1936, opened 2018-06-20
[RFR] Fix error in jsonserver provider on Windows

### LINKED ISSUE #1935: ReferenceError: Buffer is not defined ... in ra-core util fetch.js (running ra on Windows)
First steps from Tutorial (posts list) returns error:

"ReferenceError: Buffer is not defined
    at isValidObject (webpack-internal:///./node_modules/ra-core/lib/util/fetch.js:77:20)
    at Object.flattenObject (webpack-internal:///./node_modules/ra-core/lib/util/fetch.js:87:9)
...
"

**Environment**
* mozilla-neutrino + react preset 
* React-admin version: 2.0.3
* Windows 10

**Solved by:**
`global.Buffer = global.Buffer || require('buffer').Buffer;`

fzaninotto:
Closes #1935

### DIFF
--- a/packages/ra-core/src/util/fetch.js
+++ b/packages/ra-core/src/util/fetch.js
@@ -54,7 +54,7 @@ const isValidObject = value => {
     }
 
     const isArray = Array.isArray(value);
-    const isBuffer = Buffer.isBuffer(value);
+    const isBuffer = Buffer && Buffer.isBuffer(value);
     const isObject =
         Object.prototype.toString.call(value) === '[object Object]';
     const hasKeys = !!Object.keys(value).length;


packages/ra-core/src/util/fetch.js
```diff
@@ -54,7 +54,7 @@ const isValidObject = value => {
     }
 
     const isArray = Array.isArray(value);
-    const isBuffer = Buffer.isBuffer(value);
+    const isBuffer = Buffer && Buffer.isBuffer(value);
```
djhi: So we won't know if it's a buffer on Windows ?

fzaninotto: `Buffer` is a Node object - it does not exist natively in browsers. It's emulated by the `buffer` npm module, which creates a global `Buffer` object. This `buffer` module is required by webpack. So if you use webpack, you don't have the bug. I think the bug appears not if you're on Windows, but if you use another bundler (like rollup). My fix patches that.

But the heart of the problem is that the value cannot be a Buffer since it doesn't make sense for browsers. This test was added in #834, and I validated the PR, but I don't understand this Buffer test anymore.

alexicum: Yes, you are right: it's not Windows problem. But neutrino using Webpack too. May be some CRA module setup global Buffer object unlike neutrino.

PR #1939, opened 2018-06-20
[WIP] Document components using Docz

fzaninotto:
Testing [Docz](https://www.docz.site/) for API doc generation.

- [x] Make docz work
- [ ] Document a few components
- [ ] Tweak publish script to generate components doc alongside the existing doc

I'm not convinced that docz is finished. Some of the basic requirements aren't yet met, so I'll put this issue on pause until docz matures a bit.

localhost_3000_packages-ra-ui-materialui-src-button-edit-button 2


djhi:
> Some of the basic requirements aren't yet met

Can you elaborate ?

fzaninotto:
Out the top of my head:

- automated Props table don't work on our components (either because we export more than one component per file, or because we use HOCs, see https://github.com/pedronauck/docz/issues/65) (react-docgen limitation)
- The `<Playground>` tag outputs its own code (https://github.com/pedronauck/docz/issues/11)
- The `<Playground>` tag outputs context (https://github.com/pedronauck/docz/issues/75)
- The docs aren't mobile-friendly (https://github.com/pedronauck/docz/issues/49)
- The default theme is too large
- I'm missing an interactive playground (à la [storybook knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs))

ashim-k-saha:
Definitely, we will get so much help with this kind of documentation. In the above example, I got some more clarity. +1

brikou:
TS support with Docz is quite good, and automated Props table works as expected (default export is not required anymore, IProps signature is well parsed, just as Cmpt.defaultProps). You should give another try with newly converted TS components.

fzaninotto:
The problem is we already have to migrate all ra-core to TypeScript to be able to migrate components. Wanna give us a hand?

brikou:
After playing with Docz, I don't thing this is the right tool to build a robust documentation.
Why? Because of MDX and lack of TS support. Documentation cannot be checked by TS.

fzaninotto:
This Pr is too old. If we ever generate API doc, we'll give Docz another try.

PR #1942, opened 2018-06-21
[RFR] Allows to customize the redux-form name

djhi:
Follow #1953 
- [x] Allow to customize the redux-form name
- [x] Simple Example: Add a quick create in modal from a `ReferenceInput`
- [x] Simple Example: Add a quick show preview in modal from a `ReferenceInput`
- [x] Tests
- [x] Documentation
- [x] Ensure the newly created resource keep showing up in the `ReferenceInput` possible values

## Important changes

Not sure those are breaking changes though as they should not impact any documented feature

- The Controller components for views are no longer responsible of the route parsing
- The `saving` is now passed to the `Toolbar` components by the form components (`SimpleForm` and `TabbedForm`). This is the `Toolbar` responsability to pass it down to its children. This might affect users who have custom toolbars.

Imgur


djhi:
ping @jeromemacias

examples/simple/src/comments/PostQuickCreate.js
```diff
@@ -0,0 +1,36 @@
...
+        >
+            <TextInput source="title" />
+            <LongTextInput source="teaser" />
+        </SimpleForm>
+    </Create>
+);
+
+PostQuickCreate.defaultProps = {
+    basePath: '/posts',
+    formName: 'post-create',
```
fzaninotto: why do you set them as default props instead of passing them as props to `Create` in JSX? It would make the intent clearer.

examples/simple/src/comments/PostReferenceInput.js
```diff
@@ -0,0 +1,65 @@
...
+                    <DialogContent>
+                        <PostQuickCreate />
+                    </DialogContent>
+                </Dialog>
+            </div>
+        );
+    }
+}
+
+const PostReferenceInput = connect(
```
fzaninotto: isn't it the type of very local state that we'd better handle in pure React?

examples/simple/src/index.js
```diff
@@ -29,6 +40,7 @@ render(
             />,
             <Route exact path="/custom2" component={CustomRouteLayout} />,
         ]}
+        customSagas={customSagas}
```
fzaninotto: seems overkill to use a saga just to display a form. Isn't there a more local solution?

examples/simple/src/comments/PostQuickCreate.js
djhi: Less to write where you use it. But both are ok I guess

packages/ra-core/src/controller/CreateController.js
```diff
@@ -48,14 +49,6 @@ import { crudCreate as crudCreateAction } from '../actions';
  *     export default App;
  */
 class CreateController extends Component {
-    getBasePath() {
-        const { location } = this.props;
-        return location.pathname
```
fzaninotto: I don't understand, removing this will probably cause regressions

packages/ra-core/src/controller/EditController.js
```diff
@@ -59,25 +60,17 @@ export class EditController extends Component {
...
         ) {
-            this.props.resetForm('record-form');
+            this.props.resetForm(this.props.formName);
             this.updateData(nextProps.resource, nextProps.id);
         }
     }
 
-    getBasePath() {
-        const { location } = this.props;
-        return location.pathname
```
fzaninotto: same

examples/simple/src/comments/PostReferenceInput.js
djhi: The problem is, how do you know when to close the modal ? I wanted to close it only when the create succeded

examples/simple/src/index.js
djhi: Same

packages/ra-core/src/controller/CreateController.js
djhi: Why ?

djhi: The e2e tests say it's ok ;

fzaninotto: because we take into account only the final part of the URL. An URL looking like `/admin/posts` won't work anymore.

examples/simple/src/comments/PostQuickCreate.js
fzaninotto: I prefer in props than in default props

djhi: Sure, no prb

examples/simple/src/comments/PostReferenceInput.js
fzaninotto: Then use optimistic updates ;)

packages/ra-core/src/controller/CreateController.js
djhi: I'll test that but at any rate, this should not be the responsability of those components. They don't need to be aware of routing. The logic you mention could be applied in the `Resource` component

fzaninotto: agreed

examples/simple/src/comments/PostReferenceInput.js
djhi: And how do we react to failures ?

fzaninotto: refresh and show an error notification (it's what we do for optimistic deletes)

djhi: I do agree on one thing though: the UI could be improved. This is a very rough and naive exemple

djhi: > refresh and show an error notification (it's what we do for optimistic deletes)

Even after the user has then submitted the Comment Create form ?

packages/ra-core/src/controller/CreateController.js
djhi: > because we take into account only the final part of the URL.

What made you think that ?

In `Resource` at [L75](https://github.com/marmelab/react-admin/pull/1942/files#diff-3332c15e881ebff85a83708a32c91a1aR75):

```js
const basePath = match.url;
```

`match.url` will be whatever the url is at this point: `/posts` or `/admin/posts` or am I missing something ?

examples/simple/src/comments/PostReferenceInput.js
```diff
@@ -0,0 +1,54 @@
...
+        event.preventDefault();
+        this.setState({ showDialog: true });
+    };
+
+    handleClose = () => {
+        this.setState({ showDialog: false });
+    };
+
+    handleSave = post => {
+        console.log({ post });
```
fzaninotto: `console.log()` spotted

examples/simple/src/comments/PostQuickCreate.js
```diff
@@ -0,0 +1,88 @@
...
+            </SimpleForm>
+        );
+    }
+}
+
+const mapStateToProps = state => ({
+    submitting: state.admin.loading > 0,
+});
+const mapDispatchToProps = dispatch => ({
+    setFetchStart: () => dispatch({ type: FETCH_START }),
```
fzaninotto: shouldn't we create an action creator for those?

examples/simple/src/comments/PostQuickCreate.js
```diff
@@ -0,0 +1,88 @@
...
+        onSave: PropTypes.func.isRequired,
+        setFetchEnd: PropTypes.func.isRequired,
+        setFetchStart: PropTypes.func.isRequired,
+        submitting: PropTypes.bool.isRequired,
+    };
+
+    handleSave = values => {
+        const { setFetchStart, setFetchEnd, onSave } = this.props;
+
+        setFetchStart();
```
fzaninotto: I'd prefer `fetchStart()`

examples/simple/src/comments/PostQuickCreate.js
```diff
@@ -0,0 +1,88 @@
...
+    SimpleForm,
+    TextInput,
+    Toolbar,
+    required,
+} from 'react-admin'; // eslint-disable-line import/no-unresolved
+import dataProvider from '../dataProvider';
+import CancelButton from './PostQuickCreateCancelButton';
+import SaveButton from './PostQuickCreateSaveButton';
+
+const PostQuickCreateToolbar = ({ submitting, onCancel, ...props }) => (
```
fzaninotto: Why do you need a custom toolbar? To close the window on cancel? Then add a comment to explain it, it's not immediate.

examples/simple/src/comments/PostQuickCreate.js
```diff
@@ -0,0 +1,88 @@
...
+    <Toolbar {...props}>
+        <SaveButton saving={submitting} />
+        <CancelButton onClick={onCancel} />
```
fzaninotto: Why would the user click on cancel rather than close the dialog using the cross?

examples/simple/src/comments/PostQuickCreateSaveButton.js
```diff
@@ -0,0 +1,83 @@
+import React, { Component } from 'react';
```
fzaninotto: why do we need a custom save button?

examples/simple/src/comments/PostReferenceInput.js
```diff
@@ -0,0 +1,134 @@
...
+                </Dialog>
+            </Fragment>
+        );
+    }
+}
+
+const formSelector = formValueSelector('record-form');
+
+const mapStateToProps = state => ({
+    post_id: formSelector(state, 'post_id'),
```
fzaninotto: why do you need it? react-admin injects the `record` to inputs, too.

packages/ra-core/src/form/FormField.js
```diff
@@ -68,4 +68,10 @@ export class FormField extends Component {
     }
 }
 
-export default connect(undefined, { initializeForm })(FormField);
+// HACK: For some reason, connect was not injecting the initializeForm action creator as
+// a prop. Using this function was the only way to make it work again :x-files:
+const mapDispatchToProps = dispatch => ({
```
fzaninotto: Can you test again without this hack? It's worrying.

examples/simple/src/comments/PostQuickCreateSaveButton.js
djhi: Because of [this](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/button/SaveButton.js#L123)

And [this](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/reducer/admin/saving.js#L12)

In a nutshell, the save button display a `CircularProgress` if it reads `saving` from the reducer which is only set when our CRUD actions are dispatched. 

In our case, we are using the `dataProvider` directly so that we can get its result. Hence, we don't dispatch those CRUD actions and cannot set the `saving` prop on the button

examples/simple/src/comments/PostReferenceInput.js
djhi: `record` !== form values

fzaninotto: then why don't you use redux-form's `<Field>` instead?

examples/simple/src/comments/PostQuickCreateSaveButton.js
fzaninotto: Can't we make the save button reusable by getting the `saving` state not in `<SaveButton>` but in `<SimpleForm>`?

examples/simple/src/comments/PostQuickCreate.js
fzaninotto: 👍

examples/simple/src/comments/PostQuickCreateSaveButton.js
djhi: Good idea :+1:

examples/simple/src/comments/PostReferenceInput.js
djhi: I don't get this

packages/ra-core/src/sideEffect/redirection.js
```diff
@@ -30,7 +31,7 @@ export function* handleRedirection({
                   )
               )
           )
-        : yield put(reset('record-form')); // explicit no redirection, reset the form
+        : yield put(reset(formName)); // explicit no redirection, reset the form
```
fzaninotto: If I understand correctly, this side effect is the reason why you pass the form name to all the data actions? Because I've been wondering why you updated all the action creators.

djhi:
Figured why the e2e tests fail. As selenium doesn't really reload the pages, the post created in the new test is added to the available posts for the other tests. I'll fi that this evening

packages/ra-core/src/controller/CreateController.js
djhi: Btw, would it be better to extract those changes in a dedicated PR ?

examples/simple/src/comments/PostReferenceInput.js
djhi: ah, got it! Much simpler indeed :+1:

djhi:
There is one remaining bug:

1. Create a new post from the `ReferenceInput`
2. It appears in the `ReferenceInput` choices
3. Select another post as the value of the `ReferenceInput`

=> The previously created post does not appear anymore in the choices

This is because creating the post does not add it inside the `possibleValues` reducer.

djhi:
Superseded by #1955

PR #1948, opened 2018-06-25
[RFR] Migrate e2e tests to cypress

djhi:
- [x] Migrate tests
- [x] Cleanup
- [x] Document test commands
- [x] Find a better way to start the server before running the tests (probably custom script using cypress API: https://docs.cy)press.io/guides/guides/continuous-integration.html#Module-API)


djhi:
> I find the cypress output too verbose

We might be able to finetune that by specifying a mocha reporter

fzaninotto:
You're RFR but the checklist is still not complete...

djhi:
Forgot to do it. Completed !

fzaninotto:
If I understand correctly, you've remove the getters from the e2e Page objects. This worries me a bit, since it leads to plumbing in the e2e tests themselves, and a bit of code repetition. Can't you put the getters in the page objects with Cypress?

djhi:
@fzaninotto 

> If I understand correctly, you've remove the getters from the e2e Page objects. This worries me a bit, since it leads to plumbing in the e2e tests themselves, and a bit of code repetition. Can't you put the getters in the page objects with Cypress?

Can you check the files inside the `cypress/integration` folder. It looks fine to me already.

djhi:
I'll just make use of the `beforeEach` hook to avoid some code repetition

djhi:
@fzaninotto Done!

fzaninotto:
\o/

PR #1954, opened 2018-06-25
[RFR] Manage form saving state inside Form components

djhi:


djhi:
In a nutshell, the SaveButton is no longer responsible for retrieving its saving state so it can be reused in forms not connected to our main admin state

fzaninotto:
Gotcha

PR #1955, opened 2018-06-25
[RFR] Allow to customize the redux form name on Form components 

djhi:
Follow #1954 

- [x] Implementation
- [x] Tests
- [x] Documentation

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -163,6 +163,9 @@ Here are all the props accepted by the `<SimpleForm>` component:
 * [`submitOnEnter`](#submit-on-enter)
 * [`redirect`](#redirection-after-submission)
 * [`toolbar`](#toolbar)
+* `save`: The function invoked when the form is submitted. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `saving`: A boolean indicating whether a save operation is ongoing. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `form`: The name of the [`redux-form`](https://redux-form.com/7.4.2/docs/api/reduxform.md/#-code-form-string-code-required-). It defaults to `record-form` and should only be modified when using the `SimpleForm` outside of a `Create` or `Edit` component.
 
 ```jsx
 export const PostCreate = (props) => (
@@ -192,6 +195,9 @@ Here are all the props accepted by the `<TabbedForm>` component:
 * [`submitOnEnter`](#submit-on-enter)
 * [`redirect`](#redirection-after-submission)
 * [`toolbar`](#toolbar)
+* `save`: The function invoked when the form is submitted. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `saving`: A boolean indicating whether a save operation is ongoing. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `form`: The name of the [`redux-form`](https://redux-form.com/7.4.2/docs/api/reduxform.md/#-code-form-string-code-required-). It defaults to `record-form` and should only be modified when using the `TabbedForm` outside of a `Create` or `Edit` component.
 
 {% raw %}
 ```jsx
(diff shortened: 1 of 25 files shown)


djhi:
@fzaninotto I decided to not allow the customization of the redux-form name for now. It isn't required to make the customization we want available to our users.

e2e/tests/customPages.js
```diff
@@ -19,7 +19,8 @@ describe('Custom Pages', () => {
         });
 
         it('should have retrieved the number of posts', async () => {
-            assert.equal(await CustomPageNoLayout.getTotal(), '13');
+            // 14 because we created one in custom-forms tests
```
fzaninotto: I'd prefer if you deleted the record you created. We should avoid side effects in e2e tests.

packages/ra-ui-materialui/src/form/SimpleForm.js
```diff
@@ -136,11 +141,12 @@ SimpleForm.defaultProps = {
 
 const enhance = compose(
     connect((state, props) => ({
+        form: props.form || REDUX_FORM_NAME,
```
fzaninotto: why is it in `connect` while it doesn't depend on `state`?

docs/CreateEdit.md
```diff
@@ -163,6 +163,9 @@ Here are all the props accepted by the `<SimpleForm>` component:
 * [`submitOnEnter`](#submit-on-enter)
 * [`redirect`](#redirection-after-submission)
 * [`toolbar`](#toolbar)
+* `save`: The function invoked when the form is submitted. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `saving`: A boolean indicating whether a save operation is ongoing. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
+* `form`: The name of the [`redux-form`](https://redux-form.com/7.4.2/docs/api/reduxform.md/#-code-form-string-code-required-). It defaults to `record-form` and should only be modified when using the `SimpleForm` outside of a `Create` or `Edit` component.
```
fzaninotto: shouldn't it be `formName` everywhere?

examples/simple/src/comments/CommentEdit.js
```diff
@@ -31,7 +31,15 @@ const editStyles = {
...
     <EditController {...props}>
-        {({ resource, record, redirect, save, basePath, version }) => (
+        {({
+            resource,
+            record,
+            redirect,
+            save,
+            basePath,
+            version,
+            formName,
```
fzaninotto: There, I don't understand. Since the form name isn't configurable in Edit and Create controllers, they should not define it, and let the `<SimpleForm>` use a default value when the prop isn't defined.

packages/ra-ui-materialui/src/form/SimpleForm.js
djhi: it should not depend on `state`, we're setting the redux-form name here. I tried to set `defaultProps` on enhance but it didn't work. This achieve the same results by setting the `form` prop **before** the `reduxForm` HOC is applied

docs/CreateEdit.md
djhi: No, this is in fact the `reduxForm` `form` prop

examples/simple/src/comments/CommentEdit.js
djhi: Oups, forgot to remove this one

packages/ra-ui-materialui/src/form/SimpleForm.js
djhi: I could have used the `withProps` HOC from recompose though but thought it was unnecessary to introduce yet another one just for that

e2e/pages/CustomFormPage.js
```diff
@@ -0,0 +1,54 @@
+import { By, until } from 'selenium-webdriver';
```
fzaninotto: shouldn't you rewrite e2e tests to use cypress?

e2e/tests/list.js
```diff
@@ -17,6 +17,7 @@ describe('List Page', () => {
 
     describe('Pagination', () => {
         it('should display paginated list of available posts', async () => {
+            // 14 because we created one in custom-forms tests
```
fzaninotto: please remove comment

e2e/pages/CustomFormPage.js
djhi: Well, the cypress PR was on `master` so in order to do that we should first backport `master` changes into `next`

fzaninotto: oh, I thought the cypress migration was in `next`. That will be a problem...

djhi: Why would it be a problem ? We'll have to backport master anyway and tests aren't introducing any changes to the lib

djhi:
Nope, the `PostQuickCreate` will call its `onSave` prop only when the dataProvider returns and that's when we know we can close the modal

fzaninotto:
My bad, you're right. That's because you're using the `dataProvider` directly. As discussed, it's possible to use `fetch` actions instead, provided we introduce a new side effect... and I this I'd prefer that version, as it would be a great illustration for the new side effect.

PR #1963, opened 2018-06-26
Fix datagrid error

alexicum:
Uncaught TypeError: Cannot read property 'field' of undefined

Check that currentSort is available, because it is not required.


alexicum:
I'm using Datagrid alone to show second list near resource List.
Here it is. Removing `currentSort={{}}` will cause this warning:

```
render() {
    return (
      <Datagrid className={styles.DbProducts} data={this.props.data} ids={this.props.ids} currentSort={{}}>
        <TextField source="id" />
        <TextField source="name" />
      </Datagrid>
    );
  }
```

fzaninotto:
`<Datagrid>` is supposed to be used inside `<List>` or `<ReferenceManyField>`. Both inject the `currentSort` prop. If you use it outside of the normal use cases, you should respect the same convention and inject a default sort, too.

As far as I understand, your use case isn't standard and we don't want to support it. So I'm closing this PR. Feel free to reopen if you disagree.

alexicum:
May be it would be better to make currentSort required in Datagrid and DatagridHeaderCell?

fzaninotto:
You're right. Would you like to open a PR for that?

alexicum:
https://github.com/marmelab/react-admin/pull/1970

PR #1966, opened 2018-06-27
[RFR] Remove full-icu package

fzaninotto:
If the unit tests pass, it means that the node 10.5 installed on Travis is built with full icu by default, and we don't need the full_icu package anymore.


fzaninotto:
Nope :(

Issue #1938, opened 2018-06-20
FormDataConsumer passes undefined as source inside ArrayInput


pedrohh:
**What you were expecting:**
`<FormDataConsumer>` passes the correct source path when inside an `<ArrayInput`> element.

**What happened instead:**
Instead of passing the source `attachments[0].title` it is passing `attachments[0].undefined`.

**Steps to reproduce:**
Put the `<FormDataConsumer>` example from [documentation](https://marmelab.com/react-admin/Inputs.html#linking-two-inputs) inside `<ArrayInput>` element.

**Related code:**
```
export const ArticleCreate = (props) => (
    <Create {...props}>
        <SimpleForm>
            <TextInput source="text"/>
            <FormDataConsumer>
                {({ formData, ...rest }) =>
                    <TextInput source="text" {...rest} />
                }
            </FormDataConsumer>
            <ArrayInput source="attachments">
                <SimpleFormIterator>
                    <TextInput source="title"/>
                    <FormDataConsumer>
                        {({ formData, ...rest }) =>
                            <TextInput source="title" {...rest} />
                        }
                    </FormDataConsumer>
                </SimpleFormIterator>
            </ArrayInput>
        </SimpleForm>
    </Create>
);
```

**Images:**

- https://image.ibb.co/jVjbxy/1.png
- https://image.ibb.co/b9GpHy/2.png

**Environment**

* React-admin version: 2.0.3
* React version: 16.4.0
* Browser: Chrome

fzaninotto:
Duplicate of #1871, fixed in v2.0.4.

pedrohh:
@fzaninotto Hello I'm trying this with the version 2.1.1 and is still not working. The problem is when using FormDataConsumer inside an ArrayInput. As you can see in the image bellow, the path is correct, but the field shows as undefined:

- https://image.ibb.co/dSFA6d/Untitled.png

pedrohh:
@fzaninotto @djhi can you check this please? I tried again with the latest v2.1.4 and this error still happens.  FormDataConsumer inside an ArrayInput returns the source as undefined.

djhi:
That's a corner case. Here's what happen:

`ArrayInput` passes the correct props to `FormDataConsumer` including the source which is indeed not defined on `FormDataConsumer`, hence the `undefined`.

`FormDataConsumer` passes all its props to the children function so, including the `source` prop which is `undefined`.

[...]

PR #1968, opened 2018-06-28
[RFR] Support routing in TabbedForm & TabbedShowLayout

djhi:
This PR adds support for routing driven tab selection both in `TabbedForm` and `TabbedShowLayout`.

- [x] Remove unecessary `exact` from `Resource` to allow urls such as `/posts/10/show/comments`
- [x] Support tab routing by default in `TabbedForm`. Use `path` prop if provided and fallback on tab index if not
- [x] Support tab routing by default in `TabbedShowLayout`. Use `path` prop if provided and fallback on tab index if not
- [x] Tests

Screencast

### DIFF
--- a/packages/ra-core/src/Resource.js
+++ b/packages/ra-core/src/Resource.js
@@ -78,7 +78,6 @@ export class Resource extends Component {
             <Switch>
                 {create && (
                     <Route
-                        exact
                         path={`${match.url}/create`}
                         render={routeProps => (
                             <WithPermissions
@@ -95,7 +94,6 @@ export class Resource extends Component {
                 )}
                 {show && (
                     <Route
-                        exact
                         path={`${match.url}/:id/show`}
                         render={routeProps => (
                             <WithPermissions
@@ -115,7 +113,6 @@ export class Resource extends Component {
                 )}
                 {edit && (
                     <Route
-                        exact
                         path={`${match.url}/:id`}
                         render={routeProps => (
                             <WithPermissions
@@ -135,7 +132,6 @@ export class Resource extends Component {
                 )}
                 {list && (
                     <Route
-                        exact
                         path={`${match.url}`}
                         render={routeProps => (
                             <WithPermissions
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -1,8 +1,6 @@
 import React, { Component, Children, cloneElement } from 'react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
 import Tabs from '@material-ui/core/Tabs';
-import MuiTab from '@material-ui/core/Tab';
 import Divider from '@material-ui/core/Divider';
 import { withStyles } from '@material-ui/core/styles';
 import { withRouter, Route } from 'react-router-dom';
@@ -136,6 +134,8 @@ TabbedShowLayout.propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     classes: PropTypes.object,
+    location: PropTypes.object,
+    match: PropTypes.object,
     record: PropTypes.object,
     resource: PropTypes.string,
     basePath: PropTypes.string,
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -5,7 +5,7 @@ import Tabs from '@material-ui/core/Tabs';
 import MuiTab from '@material-ui/core/Tab';
 import Divider from '@material-ui/core/Divider';
 import { withStyles } from '@material-ui/core/styles';
-import { withRouter, Link, Route } from 'react-router-dom';
+import { withRouter, Route } from 'react-router-dom';
 import compose from 'recompose/compose';
 import { translate } from 'ra-core';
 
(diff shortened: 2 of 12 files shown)


nicholasio:
👍

fzaninotto:
I'd prefer if the tabbed layouts supported routing by default. This should work out of the box, without any additional code in the `simple` example.

djhi:
Ok, should we try to infer the pathname for each tab from it's label and allow overriding it with a `pathname` prop ?

fzaninotto:
Why can't we use the tab index in the URL instead?

djhi:
> Why can't we use the tab index in the URL instead?

1. Because indexes in url are ugly
2. Because it's better from a usability perspective to have a text matching the active tab, IMO
3. Because people will ask to customize it

djhi:
I tried adding routing support by default on the TabbedForm. Let me know what you think. I'll tackle the TabbedShowLayout asap

christiaanwesterbeek:
If you use tab names in the URL in a multilingual app with, you'll mix languages. Since resource names probably remain English and the tab names get translated.

djhi:
@fzaninotto I also tried to have the same way to handle tabs in `TabbedForm` and `TabbedShowLayout`, removing the `context` hack in the `Tab` component

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -90,33 +90,50 @@ export class TabbedShowLayout extends Component {
...
-                        (tab, index) =>
-                            tab &&
-                            cloneElement(tab, {
-                                context: 'header',
-                                value: index,
-                            })
-                    )}
+                    {Children.map(children, (tab, index) => {
+                        if (!tab) return null;
+                        const tabPath = `${match.url}${tab.props.path
```
fzaninotto: I'd prefer that you leave that rendering part in `Tab`, in another rendering *context*, as it was done previously. *It's always bad decide too many things for your children* (you can quote me on this).

packages/ra-ui-materialui/src/form/TabbedForm.js
```diff
@@ -111,45 +105,55 @@ export class TabbedForm extends Component {
...
-                                        ) && this.state.value !== index
-                                            ? classes.errorTabButton
-                                            : null
-                                    )}
-                                />
-                            ) : null
-                    )}
+                    {Children.map(children, (tab, index) => {
+                        if (!tab) return null;
+                        const tabPath = `${match.url}${tab.props.path
```
fzaninotto: Same remark

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
djhi: I'm -1 on this. First it's kind of a hack. Second, our two tabbed components manage their tabs differently. I'd rather have them do the same thing as it is easier to reason about them this way.

Finally, we are still deciding things for our children with this hack imo

packages/ra-core/src/Resource.js
```diff
@@ -115,7 +113,6 @@ export class Resource extends Component {
                 )}
                 {edit && (
                     <Route
-                        exact
```
fzaninotto: doesn't it mean that the route `/posts/12/show` will also render this route?

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
fzaninotto: Except it allows to change the way a tab header or content is rendered just by providing a custom `<Tab>` component. With your approach, it's no longer possible. I prefer that we keep the rendering context.

packages/ra-core/src/Resource.js
djhi: No because we're in a switch and the show route will take precedence

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
djhi: Fair enough. Shouldn't we apply the recipe to the FormTab then ?

However, instead of supplying a custom Tab, shouldn't we suggest a custom Form component altogether when further customization is needed?

packages/ra-core/src/Resource.js
fzaninotto: gotcha

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
fzaninotto: Both the `Tab` and the `FormTab` should use this double render trick.

Writing a custom Form component requires way more knowledge (redux, redux-form, form name, initial values, etc) than writing a custom tab IMO.

fzaninotto: And when I think about it, it makes `SimpleForm` and `TabbedForm` good candidates for a refactoring based on a `FormController` with a render prop.

djhi: I was thinking about it too :) However we will probably have to come up with another solution for detecting tabs with errors though

djhi: Let's merge this and start a refactoring in another PR ?

djhi: Or should I first revert the Tab change and apply the context pattern to the FormTab?

fzaninotto: Yes, please readd context, and we’ll refactor the controller part in another PR. 

I’m waiting for this (#1938) PR merge to release 2.1.0.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -90,33 +88,41 @@ export class TabbedShowLayout extends Component {
             >
                 <Tabs
                     scrollable
-                    value={this.state.value}
+                    value={location.pathname}
```
fzaninotto: :+1: smart way of dealing with tab values. I just think you should add a comment leading to line 97 below because I was scratching my head until uI read the rest of the component source.

packages/ra-ui-materialui/src/form/TabbedForm.spec.js
```diff
@@ -9,28 +9,11 @@ const translate = label => label;
 const muiTheme = { textField: { errorColor: 'red' } };
 
 describe('<TabbedForm />', () => {
-    it('should render tabs', () => {
```
fzaninotto: why did you remove this one?

djhi: Because of React Router, it seems that it doesn't render in tests

PR #1981, opened 2018-06-30
[RFR] Fix LongTextInput when used inside a TabbedForm

### LINKED ISSUE #1956: LongTextInput is not auto expandable textarea
Please do not submit support requests or "How to" questions here. For that, go to [StackOverflow](http://stackoverflow.com/questions/tagged/react-admin).

**What you were expecting:**
`<LongTextInput>` display correct

**What happened instead:**
if `<LongTextInput>` in `<TabbedForm>` have data, it not auto expandable textarea 

screen shot 2561-06-26 at 17 01 38 copy

if `<LongTextInput>` in `<TabbedForm>` no data, 

screen shot 2561-06-26 at 17 06 36 copy

**Steps to reproduce:**
1. git clone https://github.com/marmelab/react-admin.git && cd react-admin
2. make install && make run-demo
3. Login demo/demo
4. Go to Customers menu and click edit some customer
5. Go to tab Address

**Related code:**

```
<FormTab label="resources.customers.tabs.address">
                <LongTextInput
                    source="address"
                    formClassName={classes.address}
                />
                <TextInput source="zipcode" formClassName={classes.zipcode} />
                <TextInput source="city" formClassName={classes.city} />
</FormTab>
```

**Other information:**

**Environment**

* React-admin version: 2.0.3
* Last version that did not exhibit the issue (if applicable): v2.0.0-RC4
* React version: 16.3.2
* Browser: Chrome Version 67.0.3396.87 (Official Build) (64-bit)

fzaninotto:
The problem occurs when using `<LontTextInput>` inside a `<TabbedForm>`, but not if it's used in the first tab... The bug is due to a conjunction of:

- material-ui, which computes the field height based on the `scrollHeight` value of a dummy textarea,
- the way we hide inactive tabs with `display=none`, which makes all `scrollHeight` computation return 0
- the fact that we decorate fields with redux-form's `<FormField>`, which decorates the field with `connect()` and therefore makes them pure, and prevents a rerender when the tab becomes active.

I must confess the bug wasn't easy to locate, but the fix is a one liner. 

Closes #1956

### DIFF
--- a/packages/ra-ui-materialui/src/form/TabbedForm.js
+++ b/packages/ra-ui-materialui/src/form/TabbedForm.js
@@ -153,6 +153,19 @@ export class TabbedForm extends Component {
                                 record,
                                 basePath,
                                 hidden: this.state.value !== index,
+                                /**
+                                 * Force redraw when the tab becomes active
+                                 * 
+                                 * This is because the fields, decorated by redux-form and connect,
+                                 * aren't redrawn by default when the tab becomes active.
+                                 * Unfortunately, some material-ui fields (like multiline TextField)
+                                 * compute their size based on the scrollHeight of a dummy DOM element,
+                                 * and scrollHeight is 0 in a hidden div. So they must be redrawn
+                                 * once the tab becomes active.
+                                 * 
+                                 * @ref https://github.com/marmelab/react-admin/issues/1956
+                                 */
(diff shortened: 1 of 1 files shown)


djhi:
:muscle:

Issue #1892, opened 2018-06-08
Clone Button


fzaninotto:
Common need: a button in an edit or show vjew, which redirects to a Create view prepopulated with the values from the record.

PR #1987, opened 2018-07-03
[RFR] Document how to prefill a create form based on another record

### LINKED ISSUE #1930: Document how to add a button to create a record related to the current one
e.g. "create comment" in post

fzaninotto:
I think the `<CreateController>` should support `location.state` by default... But that would be a new feature, to be added to the `next` branch, while this PR documents the `master` branch.

Edit: well, implementing `location.state` support is a must have to enable #1892, so I guess I'll implement it (in another PR).

Closes #1930


djhi:
Nice :) I think it should support the querystring by default too. I was thinking about it this very morning while writing the advanced tutorials

PR #1990, opened 2018-07-03
Fix validators memoization

### LINKED ISSUE #1989: validate regex is being shared (aliased?) across sibling TextInput
**code snippet**:
```ts
export const BrandCreate = (props: any) => (
  <Create title="Create a Brand" {...props}>
    <SimpleForm redirect={"edit"}>
      <TextInput source="bt_brand_id" label="Brand Id" validate={[required(), regex(/^[a-z\-_0-9]{2,20}$/)]} />
      <TextInput source="bt_brand_name" label="Brand Name" validate={[required(), minLength(2), maxLength(100)]} />
      <TextInput
        source="bt_brand_main_website"
        // https://www.regextester.com/94502 - modified. protocol is a must
        validate={[regex(/^(?:http(s)?:\/\/)[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/)]}
      />
    </SimpleForm>
  </Create>
);
```

**What you were expecting:**
bt_brand_main_website should be validated against the regex url pattren

**What happened instead:**
bt_brand_main_website is validated against the validate of `bt_brand_id` field

screen shot 2018-07-03 at 15 47 27

**Environment**

* React-admin version: 2.1.0/1
* Last version that did not exhibit the issue (if applicable): 2.0
* React version: 16.4.2
* Browser: chrome 67

Bnaya:
fixes #1989 
Add `.prettierrc` with the settings take from `.eslintrc` makes vscode (and all the other related tools?) more happy

### DIFF
--- a/packages/ra-core/src/form/validate.js
+++ b/packages/ra-core/src/form/validate.js
@@ -70,11 +70,14 @@ export const number = memoize(
             : undefined
 );
 
-export const regex = memoize(
+export const regex = lodashMemoize(
     (pattern, message = 'ra.validation.regex') => (value, values, props) =>
         !isEmpty(value) && typeof value === 'string' && !pattern.test(value)
             ? getMessage(message, { pattern }, value, values, props)
-            : undefined
+            : undefined,
+    (pattern, message) => {
+        return pattern.toString() + message;
+    }
 );
 
 export const email = memoize((message = 'ra.validation.email') =>
--- a/packages/ra-core/src/form/validate.spec.js
+++ b/packages/ra-core/src/form/validate.spec.js
@@ -201,6 +201,24 @@ describe('Validators', () => {
                 'not foo'
             );
         });
+
+        it('should memoize the validator when the regex pattren and message are the same', () => {
+            expect(regex(/foo/, 'placeholder')).toBe(
+                regex(/foo/, 'placeholder')
+            );
+        });
+
+        it('should create new validator when the regex pattren is different', () => {
+            expect(regex(/foo/, 'placeholder')).not.toBe(
+                regex(/notfoo/, 'placeholder')
+            );
+        });
+
+        it('should create new validator when message is different', () => {
+            expect(regex(/foo/, 'placeholder')).not.toBe(
+                regex(/foo/, 'another placeholder')
+            );
+        });
     });
     describe('email', () => {
         it('should return undefined if the value is empty', () => {


packages/ra-core/src/form/validate.js
```diff
@@ -70,11 +70,14 @@ export const number = memoize(
...
 );
 
-export const regex = memoize(
+export const regex = lodashMemoize(
     (pattern, message = 'ra.validation.regex') => (value, values, props) =>
         !isEmpty(value) && typeof value === 'string' && !pattern.test(value)
             ? getMessage(message, { pattern }, value, values, props)
-            : undefined
+            : undefined,
+    (pattern, message) => {
```
Bnaya: I can add `pattern instanceof RegExp` sanity check here or something similar.
But i don't think its a must

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
djhi: Please remove this file. What is vscode not happy about ?

fzaninotto:
Can you please add a unit test?

.prettierrc
Bnaya: this is the most up to date way to share prettier config across all related tooling 
https://prettier.io/docs/en/configuration.html
with this file, vscode prettier extension simply picking the correct configurations

djhi: We included prettier as part of eslint with `eslint-config-prettier` and `eslint-plugin-prettier`. Configure your editor to apply eslint on save instead of prettier and you'll be ok :)

djhi: In VScode, this is done through the eslint extension

Bnaya:
I'm not sure what to test, i can make sure that memoization works:
 `regexp(/a/) === regexp(/a/)`
And that its using correct key:
 `regexp(/a/) === regexp(/b/)`

Is it ok?

fzaninotto:
Yes, that’s OK

Bnaya:
Added tests

Bnaya:
Also removed the `.prettierrc` file

.prettierrc
djhi: If we can have this file instead of our rules inside `.eslintrc` I'll merge that separate PR :) I just don't want our rules duplicated in two different files

djhi:
Thanks!

djhi:
You can also just use your very own regex validation function in your project in the mean time

PR #1991, opened 2018-07-03
[RFR] Add ability to clone a record

### LINKED ISSUE #1892: Clone Button

fzaninotto:
- [x] Add support for location state and search in <Create> (refs #1987)
- [x] Add `<CloneButton>` (closes #1892)
- [x] Document it
- [x] Unit test it


docs/CreateEdit.md
```diff
@@ -147,51 +147,14 @@ Using a custom `EditActions` component also allow to remove the `<DeleteButton>`
...
-            <NumberInput source="nb_views" />
-        </SimpleForm>
-    </Create>
-);
-```
-
-While using the `record` to set default values works here, it doesn't work with `<Edit>`. So it's recommended to use [the `defaultValue` prop in the Form component](#default-values) instead.
-
-However, there is a valid use case for presetting the `record` prop: to prepopulate a record based on a related record. For instance, in a `PostShow` component, you may want to display a button to create a comment related to the current post, that would lead to a `CommentCreate` page where the `post_id` is preset.
+By default, the `<Create>` view starts with an empty `record`. However, if the `location` object (injected by [react-router](https://reacttraining.com/react-router/web/api/location)) contains a `record` in its `state`, the `<Create>` view uses that `record` instead of the empty object.
```
djhi: Ok so you're making my not yet published first tutorial on advanced react-admin obsolete already! :angry:

packages/ra-core/src/controller/CreateController.js
```diff
@@ -47,7 +49,21 @@ import { crudCreate as crudCreateAction } from '../actions';
  *     );
  *     export default App;
  */
-class CreateController extends Component {
+export class CreateController extends Component {
+    constructor(props) {
```
djhi: why not using class properties to initialize `record`?

packages/ra-ui-materialui/src/button/CloneButton.js
```diff
@@ -0,0 +1,48 @@
...
+    basePath = '',
+    label = 'ra.action.clone',
+    record = {},
+    ...rest
+}) => (
+    <Button
+        component={Link}
+        to={{
+            pathname: `${basePath}/create`,
+            state: { record: omitId(record) },
```
fzaninotto: I can't believe how easy this is.

packages/ra-core/src/i18n/translate.js
```diff
@@ -33,9 +34,10 @@ const translate = BaseComponent => {
         translate: PropTypes.func.isRequired,
         locale: PropTypes.string.isRequired,
     };
-    TranslatedComponent.displayName = `TranslatedComponent(${
-        BaseComponent.displayName
-    })`;
+    TranslatedComponent.displayName = wrapDisplayName(
```
fzaninotto: I know this commit has nothing to see with the PR, but while testing the CloneButton with enzyme, I was really bothered by the `WithStyles(TranslatedComponent(undefined))` that I saw everywhere. Now they are `WithStyles(translate(Button))`, much better.

djhi: It was on my to-do list too :+1:

djhi:
Well done ! This will be a much appreciated feature :)

vipzhicheng:
Great component, but I have a special feature request, I don't know how to implement it base on CloneButton.

For now, When I click Clone, the record in create page is same as origin, but what I need is , the title becomes 'Clone of ' + original title, and the status of record always set to 0 to prevent publishing accidentally.

Is this a feature request? or already exist feature but not document?

djhi:
Hi, and thanks for your question. As explained in the [react-admin contributing guide](https://github.com/marmelab/react-admin/blob/master/README.md#contributing), the right place to ask a "How To" question, get usage advice, or troubleshoot your own code, is [StackOverFlow](http://stackoverflow.com/questions/tagged/react-admin).

This makes your question easy to find by the core team, and the developer community. Unlike Github, StackOverFlow has great SEO, gamification, voting, and reputation. That's why we chose it, and decided to keep GitHub issues only for bugs and feature requests.

PR #1995, opened 2018-07-05
[RFR] Add Error Page

fzaninotto:
- [x] Create error boundary page
- [x] Find a way to make navigation work after an error
- [x] Document how to override it

kapture 2018-07-05 at 10 29 54

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Error.js
+++ b/packages/ra-ui-materialui/src/layout/Error.js
@@ -64,7 +64,7 @@ const Error = ({
             <AppBarMobile />
         </Hidden>
         <div className={classnames(classes.container, className)} {...rest}>
-            <h1 className={classes.title}>
+            <h1 className={classes.title} role="alert">
                 <ErrorIcon className={classes.icon} />
                 {translate('ra.page.error')}
             </h1>
--- a/packages/ra-ui-materialui/src/layout/Error.js
+++ b/packages/ra-ui-materialui/src/layout/Error.js
@@ -69,17 +69,19 @@ const Error = ({
                 {translate('ra.page.error')}
             </h1>
             <div>{translate('ra.message.error')}</div>
-            <ExpansionPanel className={classes.panel}>
-                <ExpansionPanelSummary expandIcon={<ExpandMoreIcon />}>
-                    {translate('ra.message.details')}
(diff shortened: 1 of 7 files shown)


packages/ra-ui-materialui/src/layout/Error.js
```diff
@@ -0,0 +1,105 @@
...
+        <Hidden mdUp>
+            <AppBarMobile />
+        </Hidden>
+        <div className={classnames(classes.container, className)} {...rest}>
+            <h1 className={classes.title}>
+                <ErrorIcon className={classes.icon} />
+                {translate('ra.page.error')}
+            </h1>
+            <div>{translate('ra.message.error')}</div>
+            <ExpansionPanel className={classes.panel}>
```
djhi: Might be a good idea to only display this panel when `process.env.NODE_ENV` is `development` ?

fzaninotto: you're right.

packages/ra-ui-materialui/src/layout/Layout.js
```diff
@@ -53,18 +54,35 @@ const styles = theme => ({
...
-            errorInfo,
+    constructor(props) {
+        super(props);
+        /**
+         * Reset the error state upon navigation
+         *
+         * @see https://stackoverflow.com/questions/48121750/browser-navigation-broken-by-use-of-react-error-boundaries
+         * */
+        props.history.listen(() => {
+            if (this.state.hasError) {
```
djhi: neat :+1:

packages/ra-ui-materialui/src/layout/Error.js
```diff
@@ -0,0 +1,107 @@
...
+    className,
+    translate,
+    ...rest
+}) => (
+    <Fragment>
+        <Hidden mdUp>
+            <AppBarMobile />
+        </Hidden>
+        <div className={classnames(classes.container, className)} {...rest}>
+            <h1 className={classes.title}>
```
djhi: Accessibility: add the `role` attribute with the `alert` value.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

PR #1996, opened 2018-07-05
[RFR] Use redux-form onSubmit allowing remote submit

### LINKED ISSUE #1797: submitOnEnter not working
```submitOnEnter={false}``` does not work, form submits anyway.

**What you were expecting:**
Form not to submit.

**What happened instead:**

**Steps to reproduce:**

**Related code:**
```
    <Edit {...props}>
        </SimpleForm>
    </Edit>
);
```

**Other information:**

**Environment**

* React version: 16.3.2
* Browser: Chrome 66

### PR DESCRIPTION
**Pull request by djhi**
Also fixes #1797

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.js
+++ b/packages/ra-ui-materialui/src/button/SaveButton.js
@@ -56,6 +56,7 @@ export class SaveButton extends Component {
             if (e) {
                 e.preventDefault();
             }
+
             handleSubmitWithRedirect(redirect)();
         }
     };
@@ -81,7 +82,7 @@ export class SaveButton extends Component {
                 className={classnames(classes.button, className)}
                 variant={variant}
                 type={type}
-                onClick={this.handleClick}
+                onClick={submitOnEnter ? undefined : this.handleClick}
                 color={saving ? 'default' : 'primary'}
                 {...sanitizeRestProps(rest)}
             >
--- a/packages/ra-ui-materialui/src/form/SimpleForm.spec.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.spec.js
@@ -8,7 +8,7 @@ import TextInput from '../input/TextInput';
 describe('<SimpleForm />', () => {
     it('should embed a form with given component children', () => {
         const wrapper = shallow(
-            <SimpleForm>
+            <SimpleForm handleSubmit={() => {}}>
                 <TextInput source="name" />
                 <TextInput source="city" />
             </SimpleForm>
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: packages/ra-ui-materialui/src/button/SaveButton.js [hunk 1]
```diff
@@ -81,7 +82,7 @@ export class SaveButton extends Component {
                 className={classnames(classes.button, className)}
                 variant={variant}
                 type={type}
-                onClick={this.handleClick}
+                onClick={submitOnEnter ? undefined : this.handleClick}
```
fzaninotto: Does that mean that if submitOnEnter is true (which is the default), the save button doesn't work anymore? That sounds wrong.

packages/ra-ui-materialui/src/form/SimpleForm.spec.js
```diff
@@ -8,7 +8,7 @@ import TextInput from '../input/TextInput';
 describe('<SimpleForm />', () => {
     it('should embed a form with given component children', () => {
         const wrapper = shallow(
-            <SimpleForm>
+            <SimpleForm handleSubmit={() => {}}>
```
fzaninotto: If this is required, then your PR is a BC break

packages/ra-ui-materialui/src/button/SaveButton.js
djhi: Nope, it fallback to the default html submit on enter, hence the necessity to have the `onSubmit` prop on the form

packages/ra-ui-materialui/src/form/SimpleForm.spec.js
djhi: The `SimpleForm` in the tests is the not decorated by redux-form one. `handleSubmit` is injected by redux-form

packages/ra-ui-materialui/src/button/SaveButton.js
fzaninotto: Gothcha

packages/ra-ui-materialui/src/form/SimpleForm.spec.js
fzaninotto: ok then

djhi:
> I'm lost in your PR between what's injected by redux-form and what's the developer's responsibility. A documentation on that feature would help.

Well actually nothing should have changed for our users. We just allow remote submit using the `submit` action for redux-form. I don't know what to document here

djhi:
Ok, one thing has changed: we don't display a notification when clicking the submit button if the form is invalid. However, this notification was redundant with the error messages imo

djhi:
Closed in favor of #2032

PR #1998, opened 2018-07-05
[RFR] Prevent submit on enter when submitOnEnter is false

### LINKED ISSUE #1797: submitOnEnter not working

djhi:
Fixes #1797

### DIFF
--- a/examples/simple/src/users/UserCreate.js
+++ b/examples/simple/src/users/UserCreate.js
@@ -14,7 +14,7 @@ const UserCreateToolbar = ({ permissions, ...props }) => (
         <SaveButton
             label="user.action.save_and_show"
             redirect="show"
-            submitOnEnter={true}
+            submitOnEnter={false}
         />
         {permissions === 'admin' && (
             <SaveButton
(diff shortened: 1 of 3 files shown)


fzaninotto:
I'm surprised that we need to do that, especially according to https://github.com/erikras/redux-form/issues/572#issuecomment-268905929...

Are you positive that it's the right way?

djhi:
No I'm not and I just saw a bad side effect: enter does not work in `LongTextInput` anymore... However, there something not quite right with the solution from the comment you mention: we want to set the `onSubmit` prop on our form to allow remote submit and because that's the way a form should be used. I'll try something else

djhi:
Ok, I found a better way. Unfortunately, it requires the changes from #1996

djhi:
Summary:

```jsx
// in SimpleForm

handleSubmit = event => {
    const { handleSubmit, redirect, save, submitOnEnter } = this.props;
    if (!submitOnEnter && event.key === 'Enter') {
        event.preventDefault();
    }
    handleSubmit(values => save(values, redirect))(event);
};

// in render...
<form
    className={classnames('simple-form', className)}
    onSubmit={this.handleSubmit}
    {...sanitizeRestProps(rest)}
>
```

djhi:
Included in #1996

PR #2019, opened 2018-07-16
[RFR] Fix list filterDefaultValues is shared accross resources

djhi:
- [x] Refactor/simplify resource reducers
- [x] Add a failing e2e test
- [x] Fix

### DIFF
--- a/packages/ra-core/src/controller/ListController.js
+++ b/packages/ra-core/src/controller/ListController.js
@@ -138,14 +138,14 @@ export class ListController extends Component {
                 ? this.props.query
                 : { ...this.props.params };
         const filterDefaultValues = this.props.filterDefaultValues || {};
-        Object.keys(filterDefaultValues).forEach(name => {
-            if (!query.filter) {
-                query.filter = {};
-            }
-            if (!query.filter[name]) {
-                query.filter[name] = filterDefaultValues[name];
-            }
-        });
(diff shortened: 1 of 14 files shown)


packages/ra-core/src/controller/ListController.js
```diff
@@ -138,14 +138,14 @@ export class ListController extends Component {
...
-        Object.keys(filterDefaultValues).forEach(name => {
-            if (!query.filter) {
-                query.filter = {};
-            }
-            if (!query.filter[name]) {
-                query.filter[name] = filterDefaultValues[name];
-            }
-        });
+
+        query.filter = Object.keys(filterDefaultValues).reduce(
```
djhi: This is the fix

packages/ra-core/src/controller/ListController.js
```diff
@@ -138,14 +138,8 @@ export class ListController extends Component {
...
-        Object.keys(filterDefaultValues).forEach(name => {
-            if (!query.filter) {
-                query.filter = {};
-            }
-            if (!query.filter[name]) {
-                query.filter[name] = filterDefaultValues[name];
-            }
-        });
+
+        query.filter = { ...filterDefaultValues, ...query.filter };
```
djhi: This is the fix

PR #2020, opened 2018-07-17
[RFR] Fix optimistic delete in List

djhi:


packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -69,6 +71,42 @@ export default (previousState = initialState, { type, payload, meta }) => {
             .map(record => ({ ...record, ...payload.data }));
         return addRecords(updatedRecords, previousState);
     }
+    if (type === CRUD_DELETE_OPTIMISTIC) {
+        const newState = Object.keys(previousState).reduce((acc, id) => {
```
fzaninotto: You're doing an `omit` here, which can be written in one line in ES6:

```js
const { id, ...nextState } = previousState;
```

packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -69,6 +71,42 @@ export default (previousState = initialState, { type, payload, meta }) => {
...
+        }, {});
+
+        Object.defineProperty(newState, 'fetchedAt', {
+            value: previousState.fetchedAt,
+        });
+
+        return newState;
+    }
+    if (type === CRUD_DELETE_MANY_OPTIMISTIC) {
+        const newState = Object.keys(previousState).reduce((acc, id) => {
```
fzaninotto: For clarity, I'd rewrite this one with a `filter` and a simpler `reduce`, as follows:

```js
const newState = Object.entries(previousState)
   .filter(([key]) => !payload.ids.includes(key))
   .reduce((obj, [key, val]) => ({ ...obj, [key]: val }), {});
```

packages/ra-core/src/reducer/admin/resource/data.js
djhi: mmm `id` is a variable, not a string key

djhi: `const { [payload.id]: removed, ...newState } = previousState;` :tada:

djhi: I :heart: es6

fzaninotto:
Would you mind adding a little unit test at the reducer level to prove that your fix works?

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: 💪 much better

PR #2028, opened 2018-07-18
[RFR] Simplify ListView to allow easier override

fzaninotto:
- [x] Expose `setPerPage()` in `List`
- [x] ListView should inject the same props to all its children (actions, list, and pagination)
- [x] Expose omit and pick for ListController props to allow easier override

Supersedes parts of #1923


djhi:
Prettier issues. Looks good otherwise :+1:

PR #2032, opened 2018-07-18
[RFR] Pass handleSubmit directly to buttons

djhi:
> No breaking changes, `handleSubmitWithRedirect` is still passed

## Use case

Having a custom save button which alter the values before actually saving them (see the `Save as Commentable` button in the `PostCreate` component of the `simple` example)

## Current issue

It's currently not possible to alter the form values before dispatching the save action:
- Dispatching a `redux-form` `change` action is asynchronous so we can't submit directly after changing the value
- Custom saga doesn't work either as it is not possible to dispatch a `redux-form` `submit` action because we don't have an `onSubmit` on our forms

## Solution

Passing the `redux-form` `handleSubmit` function to the buttons allows the user to do whatever he wants and still benefits from the default validation mechanisms

## TODO

- [x] Implementation
- [x] Tests
- [x] Documentation

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -10,9 +10,9 @@ describe('Create Page', () => {
     it('should put the current date in the field by default', () => {
         const currentDate = new Date();
         const currentDateString = currentDate.toISOString().slice(0, 10);
-        cy
-            .get(CreatePage.elements.input('published_at'))
-            .should(el => expect(el).to.have.value(currentDateString));
+        cy.get(CreatePage.elements.input('published_at')).should(el =>
+            expect(el).to.have.value(currentDateString)
+        );
     });
 
     it('should redirect to show page after create success', () => {
@@ -51,11 +51,38 @@ describe('Create Page', () => {
         CreatePage.setValues(values);
         CreatePage.submitAndAdd();
         cy.url().then(url => expect(url).to.contain('/#/posts/create'));
-        cy
-            .get(CreatePage.elements.input('title'))
-            .should(el => expect(el).to.have.value('')); // new empty form
+        cy.get(CreatePage.elements.input('title')).should(el =>
+            expect(el).to.have.value('')
+        ); // new empty form
 
         ShowPage.navigate();
+        ShowPage.ShowPage.delete();
+    });
+
+    it.only('should allow to call a custom action updating values before submit', () => {
+        const values = [
+            {
+                type: 'input',
+                name: 'title',
+                value: 'Test title',
+            },
+            {
+                type: 'textarea',
+                name: 'teaser',
+                value: 'Test teaser',
+            },
+            {
+                type: 'checkbox',
+                name: 'commentable',
+                value: false,
+            },
+        ];
+
+        CreatePage.setValues(values);
+        CreatePage.submitWithAverageNote();
+        ShowPage.waitUntilVisible();
+        ShowPage.gotoTab(3);
+        cy.contains('10');
         ShowPage.delete();
     });
 
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -59,7 +59,7 @@ describe('Create Page', () => {
         ShowPage.ShowPage.delete();
     });
 
-    it.only('should allow to call a custom action updating values before submit', () => {
+    it('should allow to call a custom action updating values before submit', () => {
         const values = [
             {
                 type: 'input',
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -462,7 +462,7 @@ Sometimes, you may want your custom action to alter the form values before actua
 - `handleSubmitWithRedirect` which calls the default form save methods
 - `handleSubmit` which is the same prop as in [`react-form`](https://redux-form.com/7.4.2/docs/api/props.md/#-code-handlesubmit-eventorsubmit-function-code-)
 
-Knowing this, you can dispatch a custom action with a button. For instance, in the `simple` example:
+Knowing this, you can dispatch a custom action with a button and still benefit from the default crud action side effects (notifications, optimistic ui, undo, etc.). For instance, in the `simple` example:
 
 ```jsx
 // A custom action creator which modifies the values before calling the default crudCreate action creator
@@ -515,8 +515,6 @@ const PostCreateToolbar = props => (
     </Toolbar>
 );
 ```
-
-The advantages of this is that we still benefit from the default `crudCreate` action side effects.
 
 ## Custom Sagas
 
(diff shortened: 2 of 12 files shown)


fzaninotto:
What about when the user submits the form by pressing `<Enter>`? The custom handler won't be called...

cypress/integration/create.js
```diff
@@ -51,11 +51,38 @@ describe('Create Page', () => {
...
-            .should(el => expect(el).to.have.value('')); // new empty form
+        cy.get(CreatePage.elements.input('title')).should(el =>
+            expect(el).to.have.value('')
+        ); // new empty form
 
         ShowPage.navigate();
+        ShowPage.ShowPage.delete();
+    });
+
+    it.only('should allow to call a custom action updating values before submit', () => {
```
fzaninotto: .only spotted

packages/ra-ui-materialui/src/input/BooleanInput.js
```diff
@@ -24,14 +24,17 @@ export class BooleanInput extends Component {
             ...rest
         } = this.props;
 
+        const { value, ...inputProps } = input;
```
fzaninotto: this change seems unrelated, what's the motivation?

examples/simple/src/posts/PostCreate.js
```diff
@@ -13,6 +16,44 @@ import {
     Toolbar,
 } from 'react-admin'; // eslint-disable-line import/no-unresolved
 
+const saveWithNoteButton = (values, basePath, redirectTo) =>
+    crudCreate('posts', { ...values, average_note: 10 }, basePath, redirectTo);
+
+const SaveWithNoteButton = connect(
```
fzaninotto: I don't like the direct HOC decoration, can you add an intermediary component?

examples/simple/src/posts/PostCreate.js
```diff
@@ -13,6 +16,44 @@ import {
     Toolbar,
 } from 'react-admin'; // eslint-disable-line import/no-unresolved
 
+const saveWithNoteButton = (values, basePath, redirectTo) =>
```
fzaninotto: would be better named `saveWithNote`, as it's independent of the button context

packages/ra-ui-materialui/src/input/BooleanInput.js
djhi: Can be done in another PR. I stumbled upon this one while testing this feature. `BooleanInput` do not pass the input props (`name`, etc.) to the inner input element

djhi: I found this while trying to set a `BooleanInput` value with Cypress.

djhi: Should I open a PR on master for this ?

docs/Actions.md
```diff
@@ -455,6 +455,69 @@ And that's all it takes to make a fetch action optimistic. Note that the `startU
...
 The fact that react-admin updates the internal store if you use custom actions with the `fetch` meta should be another motivation to avoid using raw `fetch`.
 
+## Altering the Form Values before Submitting
+
+Sometimes, you may want your custom action to alter the form values before actually sending them to the `dataProvider`. For those cases, you should know that every buttons inside a form [Toolbar](/CreateEdit.html#toolbar) receive two props:
+
+- `handleSubmitWithRedirect` which calls the default form save methods
+- `handleSubmit` which is the same prop as in [`react-form`](https://redux-form.com/7.4.2/docs/api/props.md/#-code-handlesubmit-eventorsubmit-function-code-)
+
+Knowing this, we can dispatch a custom action with a button. For instance, in the `simple` example:
```
fzaninotto: s/ we can dispatch/ you can dispatch/

docs/Actions.md
```diff
@@ -455,6 +455,69 @@ And that's all it takes to make a fetch action optimistic. Note that the `startU
...
+
+```jsx
+// A custom action creator which modifies the values before calling the default crudCreate action creator
+const saveWithNote = (values, basePath, redirectTo) =>
```
fzaninotto: please add the `import { crudCreate } from 'react-admin'` to make your example easier to copy and paste.

docs/Actions.md
```diff
@@ -455,6 +455,69 @@ And that's all it takes to make a fetch action optimistic. Note that the `startU
...
+            label="post.action.save_with_average_note"
+            redirect="show"
+            submitOnEnter={false}
+            variant="flat"
+        />
+    </Toolbar>
+);
+```
+
+The advantages of this is that we still benefit from the default `crudCreate` action side effects.
```
fzaninotto: The advantages over what? What are the default action side effects? Also, prefer 'you' over 'we' in the documentation.

docs/CreateEdit.md
```diff
@@ -630,12 +630,69 @@ Here are the props received by the `Toolbar` component when passed as the `toolb
...
 * `pristine`: A boolean indicating whether the form is pristine (eg: no inputs have been changed yet)
 * `redirect`: The default form's redirect
 * `saving`: A boolean indicating whether a save operation is ongoing.
 * `submitOnEnter`: A boolean indicating whether the form should be submitted when pressing `enter`
 
+**Tip**: To alter the form values before submitting, you should use the `handleSubmit` prop. For instance, in the `simple` example:
+
+```jsx
+// A custom action creator which modifies the values before calling the default crudCreate action creator
+const saveWithNote = (values, basePath, redirectTo) =>
```
fzaninotto: Please link to the other page rather than duplicating the example. You're doubling the cost of documentation maintenance.

djhi:
> I still don't know what happens when the user submits the form with <Enter>. Nothing changed on that side?

Yup nothing changed :)

djhi:
> What about when the user submits the form by pressing `<Enter>`? The custom handler won't be called...

Just tested it locally, if the button with the custom handler is the only one with a `submitOnEnter` set to `true` (you can only have one anyway), it works as expected

docs/Actions.md
```diff
@@ -455,6 +455,67 @@ And that's all it takes to make a fetch action optimistic. Note that the `startU
...
+
+Sometimes, you may want your custom action to alter the form values before actually sending them to the `dataProvider`. For those cases, you should know that every buttons inside a form [Toolbar](/CreateEdit.html#toolbar) receive two props:
+
+- `handleSubmitWithRedirect` which calls the default form save methods
+- `handleSubmit` which is the same prop as in [`react-form`](https://redux-form.com/7.4.2/docs/api/props.md/#-code-handlesubmit-eventorsubmit-function-code-)
+
+Knowing this, you can dispatch a custom action with a button and still benefit from the default crud action side effects (notifications, optimistic ui, undo, etc.). For instance, in the `simple` example:
+
+```jsx
+// A custom action creator which modifies the values before calling the default crudCreate action creator
```
fzaninotto: please add the `import { crudCreate } from 'react-admin'` line to make your example easier to copy and paste.

packages/ra-ui-materialui/src/input/BooleanInput.js
fzaninotto: nevermind, let's keep it here

PR #2033, opened 2018-07-18
[RFR] Fix Theme doesn't apply to Fields

### LINKED ISSUE #2031: Themes don't apply to Fields
Themes don't seem to be applied to Fields, only inputs.
When overriding the theme's `palette` no difference occurs in e.g. TextFields. (Their labels are changed)

djhi:
Fixes #2031 

Before:
Imgur

After:
Imgur

Note: There is still an issue with the tab buttons but it seems to be on material-ui side


fzaninotto:
tests fail

PR #2037, opened 2018-07-19
[RFR] Fix form keep values after navigation

### LINKED ISSUE #2021: When you sequentially navigate between Create forms of different resources, the set of Redux state values: form -> record-form -> values not cleared.
When you sequentially navigate between Create forms of different resources, the set of Redux state values: form -> record-form -> values, not cleared. As a result, the set contains fields from different resources.

* React-admin version: 2.1.2
* React version: 16.3.2
* Browser: Chrome (64-bit) 67.0.3396.99

djhi:
Fixes #2021

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.js
@@ -147,7 +147,6 @@ const enhance = compose(
     })),
     translate, // Must be before reduxForm so that it can be used in validation
     reduxForm({
-        destroyOnUnmount: false,
         enableReinitialize: true,
         keepDirtyOnReinitialize: true,
     }),
--- a/packages/ra-ui-materialui/src/form/TabbedForm.js
+++ b/packages/ra-ui-materialui/src/form/TabbedForm.js
@@ -274,7 +274,6 @@ const enhance = compose(
     }),
     translate, // Must be before reduxForm so that it can be used in validation
     reduxForm({
-        destroyOnUnmount: false,
         enableReinitialize: true,
         keepDirtyOnReinitialize: true,
     }),


packages/ra-ui-materialui/src/form/SimpleForm.js
```diff
@@ -147,7 +147,6 @@ const enhance = compose(
     })),
     translate, // Must be before reduxForm so that it can be used in validation
     reduxForm({
-        destroyOnUnmount: false,
```
djhi: Confirmed the bug fixed when we added this is still fixed with the two props below

packages/ra-ui-materialui/src/form/TabbedForm.js
```diff
@@ -274,7 +274,6 @@ const enhance = compose(
     }),
     translate, // Must be before reduxForm so that it can be used in validation
     reduxForm({
-        destroyOnUnmount: false,
```
djhi: same

packages/ra-ui-materialui/src/form/SimpleForm.js
fzaninotto: You mean #1613?

fzaninotto: Yes, #1997 also fixes #1613, so we no longer need that line.

mobiletradingpartners:
this code seems to be present still in ra-ui-materialui/src/form/SimpleForm.js and /TabbedForm.js, in version 2.8.5 of the package

...and the bugs referenced (eg caching of data from previous form invocations, and therefore saving of incorrect form data) still exist in that version

lazjramos:
Hi, we are using the version 2.9.2 and still we have the same problem. I looked in SimpleForm.js in the master branch and still it has the line: "destroyOnUnmount: false,".

Please, can you tell me what must we have to do?

PR #2056, opened 2018-07-20
[RFR] Fix record value in SimpleFormIterator children

### LINKED ISSUE #1835: ShowButton in ArrayField points to wrong route
**What you were expecting:**
`ShowButton` can be used in different scenarios when using references in Datagrid.

**What happened instead:**
`ShowButton`'s route is incorrect when used in specific scenario described below.

**Related code:**
Showing list of images inside places reference and wanted to add `ShowButton` to link to `images` reference.
```
<ArrayField source="images" >
  <Datagrid>
    <TextField label="ID" source="id" />
    ...
    <ShowButton basePath="images" />
  </Datagrid>
</ArrayField>
```
This produces original reference show route with id if used without basePath (/places/<IMAGE_ID>/show) and append basePath and id to original reference path otherwise (/places/<PLACE_ID>/images/<IMAGE_ID>/show).

tried also (with or without basePath):
```
<ArrayField source="images" >
  <Datagrid>
    <TextField label="ID" source="id" />
    ...
    <ReferenceField source="id" reference="media">
      <ShowButton />
    </ReferenceField>
    <ShowButton basePath="images" />
  </Datagrid>
</ArrayField>
```
This produces original reference show route with id (/places/<IMAGE_ID>/show) no matter basePath.

**Environment**
* React-admin version: 2.0.0-RC4
* React version: 16.3.2

fzaninotto:
Imagine that `resource` is `posts`, `record` is `{ id: 123, comments: [{ id: 456, body: 'foobar' }] }`.

`<ArrayInput source="comments">` properly injects the `record` and `source` to `<SimpleFormIterator>`. But `<SimpleFormIterator>` injects this record (`{ id: 123, comments: [{ id: 456, body: 'foobar' }] }`) directly to its children, while in fact they're only interested in the sub records (`{ id: 456, body: 'foobar' }`).

This PR fixes that.

Closes #1835

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -126,7 +126,7 @@ export class SimpleFormIterator extends Component {
                                                     input.props.label ||
                                                     input.props.source,
                                             })}
-                                            record={records[index]}
+                                            record={records && records[index]}
                                             resource={resource}
                                         />
                                     ))}
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -2,6 +2,7 @@ import React, { Children, cloneElement, Component } from 'react';
 import PropTypes from 'prop-types';
 import compose from 'recompose/compose';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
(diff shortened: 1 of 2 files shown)


fzaninotto:
Back to Work in Progress: the error message may be misinterpreted as to say that *all* responses should be arrays. The error message should indicate the verb, and hint that the solution lies in the `dataProvider`.

PR #2060, opened 2018-07-22
[RFR] Debounce calls to crudGetMatching

### LINKED ISSUE #2004: Aggregate crudGetMatching requests from ReferenceInput
Browser is stuck (page load forever and after a while the "kill page" prompt appears)

ArrayInput should aggregate all the needed requests and made only 1 request for each resource it uses with all the ids it uses and bind the data correctly.

Create a component that makes the above but not as generic as I would want it to be.

**Additional context**
Created a code sandbox to show my problem
https://r7jxw6748n.codesandbox.io/#/posts/1 (if you need to login, user: admin, password: password)
https://imgur.com/arni0wi

djhi:
Fixes #2004

- [x] Implementation
- [x] Tests


djhi:
I don't really know how to test it...

fzaninotto:
Why can't you use the existing `accumulate` saga?

djhi:
> Why can't you use the existing accumulate saga?
Because we don't need to accumulate those calls but to debounce them. Checkout the key I build in the `debounce` side effect

packages/ra-core/src/sideEffect/debounce.js
```diff
@@ -0,0 +1,58 @@
...
+ * `SUCCESS` or `FAILURE` counterparts and removing the action from our debounced tasks
+ * list
+ */
+export function* handleDebouncedAction(
+    key,
+    debouncedAction,
+    isRelatedAction = defaultIsRelatedAction(debouncedAction.type, key)
+) {
+    yield put(debouncedAction);
+    yield take(isRelatedAction, handleFinalize, key);
```
fzaninotto: I don't agree with the strategy. The debounce should be action agnostic, i.e. be usable even with non-`fetch` actions, and debounce not based on the response, but based on a delay.

packages/ra-core/src/controller/input/ReferenceInputController.js
```diff
@@ -172,7 +173,7 @@ export class ReferenceInputController extends Component {
         } = props;
         const { pagination, sort, filter } = this.state;
 
-        crudGetMatching(
+        crudGetMatchingDebounce(
```
fzaninotto: or we can simply used `lodash/debounce` here to avoid creating a new action creator and saga...

packages/ra-core/src/controller/input/ReferenceInputController.js
```diff
@@ -154,16 +155,16 @@ export class ReferenceInputController extends Component {
     };
 
     fetchReference = (props = this.props) => {
-        const { crudGetOne, input, reference } = props;
+        const { crudGetManyAccumulate, input, reference } = props;
         const id = input.value;
         if (id) {
-            crudGetOne(reference, id, null, false);
+            crudGetManyAccumulate(reference, [id], null, false);
```
fzaninotto: 👍 quick win

packages/ra-core/src/controller/input/ReferenceInputController.js
djhi: And this happens when used inside an `ArrayInput` for example

packages/ra-core/src/sideEffect/debounce.js
djhi: So you mean we should pass the key in the action itself (meta or payload) ?

packages/ra-core/src/sideEffect/accumulate.js
```diff
@@ -4,57 +4,61 @@ import { call, cancel, fork, put, takeEvery } from 'redux-saga/effects';
...
+    const accumulatedValue = new Set(accumulations) || new Set();
+    ids.forEach(id => accumulatedValue.add(id));
+    return Array.from(accumulatedValue);
 };
-const getIds = resource => {
-    const ids = Object.values(debouncedIds[resource]);
-    delete debouncedIds[resource];
-    return ids;
+
+const getAccumuledValue = key => {
```
fzaninotto: getAccumulatedValue

packages/ra-core/src/sideEffect/accumulate.js
```diff
@@ -4,57 +4,61 @@ import { call, cancel, fork, put, takeEvery } from 'redux-saga/effects';
...
+ * Cancel call to finalize, accumulate values, and call finalize
  *
  * @example
  * accumulate({ type: CRUD_GET_MANY_ACCUMULATE, payload: { ids: [1, 3, 5], resource: 'posts' } })
  */
-function* accumulate({ payload, meta }) {
-    const { ids, resource } = payload;
-    if (tasks[resource]) {
-        yield cancel(tasks[resource]);
+export function* accumulate(action) {
```
fzaninotto: why change the signature? It seems you only use `{ payload, meta }` anyway

packages/ra-core/src/sideEffect/accumulate.js
```diff
@@ -4,57 +4,61 @@ import { call, cancel, fork, put, takeEvery } from 'redux-saga/effects';
...
-    if (!debouncedIds[resource]) {
-        debouncedIds[resource] = {};
-    }
-    ids.forEach(id => {
-        debouncedIds[resource][id] = id;
-    }); // fast UNIQUE
+let accumulations = {};
+
+const addIds = (accumulations, { payload: { ids } }) => {
+    const accumulatedValue = new Set(accumulations) || new Set();
```
fzaninotto: missing the per ressource (per key now) discrimination

packages/ra-core/src/sideEffect/accumulate.js
```diff
@@ -4,57 +4,61 @@ import { call, cancel, fork, put, takeEvery } from 'redux-saga/effects';
...
  * @see https://redux-saga.js.org/docs/recipes/#debouncing
  */
-function* finalize(resource, actionCreator) {
+export function* finalize(key, actionCreator) {
     // combined with cancel(), this debounces the calls
     yield call(delay, 50);
-    yield put(actionCreator(resource, getIds(resource)));
-    delete tasks[resource];
+    const accumulatedValue = getAccumuledValue(key);
+    const action = actionCreator(key, accumulatedValue);
```
fzaninotto: why create a `const` here? It makes your diff harder to read for no good reason.

packages/ra-core/src/actions/accumulateActions.js
```diff
@@ -1,9 +1,39 @@
-import { crudGetMany } from './dataActions';
+import { crudGetMany, crudGetMatching } from './dataActions';
 
 export const CRUD_GET_MANY_ACCUMULATE = 'RA/CRUD_GET_MANY_ACCUMULATE';
 
 export const crudGetManyAccumulate = (resource, ids) => ({
     type: CRUD_GET_MANY_ACCUMULATE,
     payload: { resource, ids },
-    meta: { accumulate: crudGetMany },
+    meta: { accumulate: (resource, ids) => crudGetMany(resource, ids) },
```
fzaninotto: I don't understand that change

packages/ra-core/src/sideEffect/accumulate.js
```diff
@@ -4,57 +4,91 @@ import { call, cancel, fork, put, takeEvery } from 'redux-saga/effects';
...
-    delete tasks[resource];
+
+    // Get the latest accumulated value for the provided key
+    const accumulatedValue = getAccumulatedValue(key);
+
+    // For backward compatibility, we pass the key (which may be a resource name) as the first parameter
+    const action = actionCreator(key, accumulatedValue);
+
+    yield put(action);
+    delete tasks[key];
```
fzaninotto: you should delete accumulations, too

packages/ra-core/src/sideEffect/accumulate.spec.js
```diff
@@ -0,0 +1,142 @@
...
+            const saga = accumulateFactory(tasks, accumulations, finalize)({
+                payload: { resource: 'posts', ids: [1, 2] },
+                meta: { accumulate: crudGetMany },
+            });
+
+            expect(saga.next().value).toEqual(
+                fork(finalize, 'posts', crudGetMany)
+            );
+
+            expect(accumulations).toEqual({
```
fzaninotto: could you also test the `tasks` value?

packages/ra-core/src/sideEffect/accumulate.spec.js
```diff
@@ -0,0 +1,142 @@
...
+import { call, cancel, fork, put } from 'redux-saga/effects';
+import { createMockTask } from 'redux-saga/utils';
+
+import { accumulateFactory, finalizeFactory } from './accumulate';
+import { crudGetMany } from '../actions';
+
+describe('accumulate saga', () => {
+    describe('backward compatibility', () => {
+        it('proceeds with the accumulated action when no more actions are dispatched', () => {
+            const finalize = finalizeFactory({}, {});
```
fzaninotto: it feels weird not to use the same `tasks` and `accumulations` objects as for the other factory. I know you don't need it in this test, but that gives a wrong usage example.

packages/ra-core/src/sideEffect/accumulate.spec.js
djhi: Nope, because it will be `{ posts: undefined }` as `fork` returns that in tests. That's actually why I had to use `createMockTask` to test task cancellation

fzaninotto: gotcha

fzaninotto:
I'm super happy to have this merged onto `next`. Thanks!

PR #2062, opened 2018-07-22
[RFR] dataProvider response format validation: items without id

djhi:


packages/ra-core/src/sideEffect/fetch.js
```diff
@@ -38,6 +38,17 @@ function validateResponseFormat(
         );
         throw new Error('ra.notification.data_provider_error');
     }
+    if (
+        fetchActionsWithArrayOfRecordsResponse.includes(type) &&
```
fzaninotto: this will make all existing providers fail for the DELETE_MANY and UPDATE_MANY verbs, whose response format was left fuzzy (array of records or array of ids?), probably because it's not used.

djhi: Should I limit this message to the `GET_LIST`, `GET_MANY` and `GET_MATCHING` then ?

PR #2064, opened 2018-07-22
[RFR] Fix submitOnEnter

### LINKED ISSUE #1797: submitOnEnter not working

djhi:
Fixes #1797

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.js
@@ -84,6 +84,8 @@ export class SimpleForm extends Component {
         return (
             <form
                 className={classnames('simple-form', className)}
+                // Prevent automatic submission when hitting enter
+                action="javascript: void(0)"
                 {...sanitizeRestProps(rest)}
             >
                 <div className={classes.form} key={version}>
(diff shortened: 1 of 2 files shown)


fzaninotto:
HTML forms understands a click on `enter` as a submission if:

- the form has only one input
- the form has a submit button

See https://codepen.io/anon/pen/wxdEjJ for the proof and https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission for the spec.

fzaninotto:
any news on this one?

djhi:
I have a solution/hack.

The issue is that when all the form buttons have their `submitOnEnter` set to `false`, the users can still hit `Enter` and the form will be submitted using its action attribute (the HTML classic one) which only accept strings...

To avoid having to set `onKeyPress` handlers, the most simple solution is to use this `action` prop on our forms by default: `action="javascript: void(0)"`.

It does not prevent submitting the form with Enter when at least one button has its `submitOnEnter` prop set to `true` nor with click on a button.

@fzaninotto What do you think?

PR #2067, opened 2018-07-23
[RFR] Fix ArrayInput defaultValue

### LINKED ISSUE #2046: The default value of an ArrayInput doesn't work with a SimpleFormIterator
**What you were expecting:**
I want to have a default value for an `ArrayInput` with a `SimpleFormIterator` as a child.

**What happened instead:**
The input isn't populated as expected and stay empty.

**Steps to reproduce:**
1. Have an `ArrayInput` with a `SimpleFormIterator` as a child
2. Set the prop `defaultValue` with a correct format

**Related code:**

See the first input in `src/posts/PostCreate.js` file within the following Code Sandbox:
https://codesandbox.io/s/1wqxkv6pml

The code is the following:

```jsx
<ArrayInput
    source="backlinks"
    defautltValue={[
        { date: "2012-08-09T00:00:00.000Z", url: "http://example.com/bar/baz.html" }
    ]}
>
    <SimpleFormIterator>
        <DateInput source="date" />
        <TextInput source="url" />
    </SimpleFormIterator>
</ArrayInput>
```

**Other information:**
Related question on SO: https://stackoverflow.com/questions/51425225/arrayinput-how-to-set-default-value

**Environment**

* React-admin version: from **2.0.2** to **2.1.2** at least
* React version: 16.3.1
* Browser: Chrome

djhi:
Follow #2084 
Fixes #2046 

- [x] Introduce `withDefaultValue` to allow input components not using `addField` to dispatch the `initializeForm` action for their default values
- [x] Refactor FormField to use the new HOC
- [x] Use the new HOC in `ArrayInput`
- [x] Fix SimpleFormIterator so that it initializes correctly when it has default values
- [x] Tests

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -15,6 +15,17 @@ describe('Create Page', () => {
         );
     });
 
+    it('should put the ArrayInput default value', () => {
+        const currentDate = new Date();
+        const currentDateString = currentDate.toISOString().slice(0, 10);
+        cy.get(CreatePage.elements.input('backlinks[0].date')).should(el =>
+            expect(el).to.have.value(currentDateString)
+        );
+        cy.get(CreatePage.elements.input('backlinks[0].url')).should(el =>
+            expect(el).to.have.value('http://google.com')
+        );
+    });
+
     it('should redirect to show page after create success', () => {
         const values = [
             {
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -56,7 +56,7 @@ describe('Create Page', () => {
         ); // new empty form
 
         ShowPage.navigate();
-        ShowPage.ShowPage.delete();
+        ShowPage.delete();
     });
 
     it('should allow to call a custom action updating values before submit', () => {
--- a/packages/ra-core/src/form/withDefaultValue.js
+++ b/packages/ra-core/src/form/withDefaultValue.js
@@ -1,0 +1,60 @@
+import { Component, createElement } from 'react';
+import PropTypes from 'prop-types';
+
+import { connect } from 'react-redux';
+import { initializeForm as initializeFormAction } from '../actions/formActions';
+
+export class DefaultValue extends Component {
+    static propTypes = {
+        decoratedComponent: PropTypes.oneOfType([
+            PropTypes.element,
+            PropTypes.func,
+        ]),
+        defaultValue: PropTypes.any,
+        initializeForm: PropTypes.func.isRequired,
+        input: PropTypes.object,
+        source: PropTypes.string,
+        validate: PropTypes.oneOfType([PropTypes.func, PropTypes.array]),
+    };
+
+    componentDidMount() {
+        const { defaultValue, input, initializeForm, source } = this.props;
+        if (typeof defaultValue === 'undefined' || input) {
+            return;
+        }
+        initializeForm({
+            [source]:
+                typeof defaultValue === 'function'
+                    ? defaultValue()
+                    : defaultValue,
+        });
+    }
+
+    componentDidUpdate(prevProps) {
+        const { defaultValue, input, initializeForm, source } = this.props;
+        if (typeof defaultValue === 'undefined' || input) {
+            return;
+        }
+
+        if (defaultValue !== prevProps.defaultValue) {
+            initializeForm({
+                [source]:
+                    typeof defaultValue === 'function'
+                        ? defaultValue()
+                        : defaultValue,
+            });
+        }
+    }
+
+    render() {
+        const { initializeForm, decoratedComponent, ...props } = this.props;
+        return createElement(decoratedComponent, props);
+    }
+}
+
+export default DecoratedComponent => {
+    return connect(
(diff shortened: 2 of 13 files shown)


fzaninotto:
The refactoring makes the change difficult to understand... Can you either split the commit into two (one refactoring, one fix), or explain what you've changed?

djhi:
@fzaninotto Clearer ?

packages/ra-core/src/form/withDefaultValue.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+        initializeForm({
+            [source]:
+                typeof defaultValue === 'function'
+                    ? defaultValue()
+                    : defaultValue,
+        });
+    }
+
+    componentDidUpdate(prevProps) {
```
fzaninotto: why use `componentDidUpdate` instead of the previous `componentWillReceiveProps`?

packages/ra-core/src/form/withDefaultValue.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    render() {
+        const { initializeForm, component, ...props } = this.props;
+        return createElement(component, props);
+    }
+}
+
+export default DecoratedComponent => {
```
fzaninotto: use shorthand function (loose the brackets and `return`)

packages/ra-core/src/form/withDefaultValue.spec.js
```diff
@@ -0,0 +1,53 @@
...
+                    initializeForm={initializeForm}
+                    component={BaseComponent}
+                    source="bar"
+                    defaultValue="foo"
+                />
+            );
+            assert.equal(initializeForm.mock.calls.length, 1);
+            assert.deepEqual(initializeForm.mock.calls[0][0], { bar: 'foo' });
+
+            wrapper.setProps({ defaultValue: 'bar', initializeForm });
```
fzaninotto: why do you reset `initializeForm`? It's a function...

packages/ra-ui-materialui/src/form/SimpleFormIterator.js
```diff
@@ -65,7 +65,11 @@ export class SimpleFormIterator extends Component {
         // but redux-form doesn't provide one (cf https://github.com/erikras/redux-form/issues/2735)
         // so we keep an internal map between the field position and an autoincrement id
         this.nextId = 0;
-        this.ids = props.fields ? props.fields.map(() => this.nextId++) : [];
+        this.ids = props.defaultValue
+            ? props.defaultValue.map(() => this.nextId++)
```
fzaninotto: this one I don't understand. Can you explain the rationale?

cypress/integration/create.js
```diff
@@ -15,6 +15,17 @@ describe('Create Page', () => {
         );
     });
 
+    it('should put the ArrayInput default value', () => {
+        const currentDate = new Date();
+        const currentDateString = currentDate.toISOString().slice(0, 10);
+        cy.get(CreatePage.elements.input('backlinks[0].date')).should(el =>
+            expect(el).to.have.value(currentDateString)
```
fzaninotto: I'm afraid this is subject to race conditions, where the `Date` generated in the test and the one generated in the default value won't match.

packages/ra-core/src/form/withDefaultValue.js
djhi: because `componentWillReceiveProps` is deprecated. See https://reactjs.org/blog/2018/03/29/react-v-16-3.html

cypress/integration/create.js
djhi: No more than in the first create test at L10

packages/ra-ui-materialui/src/form/SimpleFormIterator.js
djhi: Explained in comments

packages/ra-ui-materialui/src/form/SimpleFormIterator.js
```diff
@@ -64,7 +64,16 @@ export class SimpleFormIterator extends Component {
         // but redux-form doesn't provide one (cf https://github.com/erikras/redux-form/issues/2735)
         // so we keep an internal map between the field position and an autoincrement id
         this.nextId = 0;
-        this.ids = props.fields ? props.fields.map(() => this.nextId++) : [];
+
+        // We check whether we have a defaultValue (which must be an array) before checking
+        // the fields prop which will always be empty for a new record.
+        // Without it, our ids wouldn't match the default value and we would get key warnings
+        // on the CssTransition element inside our render method
+        this.ids = props.defaultValue
```
fzaninotto: I still don't understand why you don't check `props.fields` *before* checking `props.defaultValue`. Here is my rationale:

1. if there is a record, use the fields
2. if there is no record, use the default value

The 1st case doesn't work like that with your code.

By the way, since `this.nextId` is initialized just above, I think a faster way to do `props.fields.map(() => this.nextId++)` is `this.nextId = props.fields.length`.

PR #2070, opened 2018-07-24
[RFR] Fix link color in ReferenceField

fzaninotto:
ReferenceField links have lost their color since #2033

### DIFF
--- a/examples/demo/src/products/GridList.js
+++ b/examples/demo/src/products/GridList.js
@@ -20,6 +20,10 @@ const styles = {
         background:
             'linear-gradient(to top, rgba(0,0,0,0.8) 0%,rgba(0,0,0,0.4) 70%,rgba(0,0,0,0) 100%)',
     },
+    price: {
+        display: 'inline',
+        fontSize: '1em',
+    },
     link: {
         color: '#fff',
     },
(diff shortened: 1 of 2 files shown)


examples/demo/src/products/GridList.js
```diff
@@ -37,16 +41,16 @@ const GridList = ({ classes, ids, data, basePath }) => (
...
-                                        source="price"
-                                        record={data[id]}
-                                        options={{
-                                            style: 'currency',
-                                            currency: 'USD',
-                                        }}
-                                    />
-                                </b>
+                                <NumberField
+                                    className={classes.price}
```
fzaninotto: another side effect of the addition of `Typography` was that this component also lost its color

packages/ra-ui-materialui/src/field/ReferenceField.js
```diff
@@ -36,11 +35,9 @@ export const ReferenceFieldView = ({
 
     if (resourceLinkPath) {
         return (
-            <Link
-                className={classnames(classes.link, className)}
-                to={resourceLinkPath}
-            >
+            <Link to={resourceLinkPath} className={className}>
                 {React.cloneElement(children, {
+                    className: classes.link, // force color override for Typography components
```
fzaninotto: or should I apply the `className` prop to the child, too? I'm afraid it's a BC break.

PR #2071, opened 2018-07-24
Fix the 'directory already exists' issue when re-making the project

afilp:
I received an error message that the 'docs' directory already exists in packages, when updating the project with a new make (i.e. the folders were already there).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ build-react-admin:
 	@echo "Transpiling react-admin files...";
 	@rm -rf ./packages/react-admin/docs
 	@cd ./packages/react-admin && yarn -s build
-	@mkdir packages/react-admin/docs
+	@mkdir -p packages/react-admin/docs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ build-react-admin:
 	@echo "Transpiling react-admin files...";
 	@rm -rf ./packages/react-admin/docs
 	@cd ./packages/react-admin && yarn -s build
-	@mkdir packages/react-admin/docs
+	@mkdir -p packages/react-admin/docs
```
djhi: It should have been removed by L39

fzaninotto:
Thanks!

fzaninotto:
oops, I merged too soon. @djhi, should I revert?

fzaninotto:
Reverted in #2073. The fix isn't good. The problem is that the command isn't run from the correct directory. @djhi will fix it.

PR #2081, opened 2018-07-25
[RFR] Fix cypress failing on chrome 68

djhi:
See https://github.com/cypress-io/cypress/issues/2037


fzaninotto:
Your fix seems to differ from [the recommended fix](https://github.com/cypress-io/cypress/issues/2037#issuecomment-407898194).

djhi:
I missed that one. Going to try it now

Issue #1906, opened 2018-06-13
FormDataConsumer is resetting the form when a contained component has a defaultValue


stopmachine:
**What you were expecting:**
For `FormDataConsumer` to be evaluated once and render the inner components if needed.

**What happened instead:**
The `FormDataConsumer` is rendered two times, once with the correct form values, then if an inner component has a `defaultValue` the form is reset and `FormDataConsumer` is evaluated again with the form values reset.

**Steps to reproduce:**
See [this sandbox](https://codesandbox.io/s/129r0lx3l), which was edited in `PostCreate.js` at line 60. If you type a character in the Create Post page in the Title field it immediately disappears, and the Average note field does not appear, until you type the character again.

**Related code:**
CodeSandbox:
https://codesandbox.io/s/129r0lx3l

**Other information:**

**Environment**

* React-admin version: 2.0.2
* Last version that did not exhibit the issue (if applicable):
* React version: 16.3.17
* Browser: Firefox 60.0.1 and Chrome 67.0.3396.62 on Ubuntu 16.04
* Stack trace (in case of a JS error):

ashim-k-saha:
+1

fzaninotto:
Bug reproduced and confirmed, thanks for taking the time to isolate and reproduce it with a sandbox.

Here are the Redux actions dispatched when there is no `defaultValue`:

![2018-07-04_1727](https://user-images.githubusercontent.com/99944/42285444-efbcf260-7faf-11e8-90a9-36a5e029e669.png)

And when there is one:

![2018-07-04_1728](https://user-images.githubusercontent.com/99944/42285450-f4d1af20-7faf-11e8-8452-9e1d314f99ee.png)

Our `RA/INITIALIZE_FORM` action seems to be creating the problem.

fzaninotto:
#1997 has been reverted in #2075 because it caused side effects, so this bug effectively reappears.

djhi:
Fixed by #2082

PR #2082, opened 2018-07-25
[RFR] Fix form reinitializations 

### LINKED ISSUE #1994: SelectInput with default value causes Create form reset when switching form tabs
**What you were expecting:**
`SelectInput` works as expected and doesn't cause Create form to reset its values when switching between form tabs.

**What happened instead:**
`SelectInput` with defaultValue set causes Create form reset its values when switching between form tabs.

**Steps to reproduce:**
1. Create form with couple of tabs containing `SelectInput` with defaultValue set
2. Fill up values in one tab and try to switch between tabs

**Related code:**
```
<Create {...props}>
    <TabbedForm>
      <FormTab label="Tab 1">
        <TextInput source="title" />
        <SelectInput
          source="type"
          choices={[
            { id: "type1", name: "Type 1" },
            { id: "type2", name: "Type 2" }
          ]}
          defaultValue={"type1"}
        />
      </FormTab>
      <FormTab label="Tab 2">
        <TextInput source="title" />
      </FormTab>
      <FormTab label="Tab 3">
        <SelectInput
          source="category"
          choices={[
            { id: "cat1", name: "Category 1" },
            { id: "cat2", name: "Category 2" }
          ]}
          defaultValue={"cat1"}
        />
      </FormTab>
   </TabbedForm>
</Create>
```
or check [this](https://codesandbox.io/s/6n0xlnjzkn) CodeSandbox.
The following happens in Product Create form:
1. Tab1.title is filled and switched to Tab2 -> form is completely reset
2. If Tab1 SelectInput doesn't have defaultValue set, switching to Tab2 reset form only once (?)
3. If Tab1 and Tab2 SelectInputs both don't have defaultValue, switching tabs works as expected.

**CodeSandbox**
https://codesandbox.io/s/6n0xlnjzkn

**Other information:**
I noticed this behaviour in earlier versions (I think it's there from at least 2.0.2).

**Environment**

* React-admin version: 2.1.0
* React version: 16.3.1
* Browser: any

djhi:
Follow #2084 

- [x] Update redux-form
- [x] Introduce the `resetForm` action and use it in the record side effect instead of the `LOCATION_CHANGE` (Fixes #1994 & #1906)
- [x] Ensure the form is destroyed when unmounted (Ensure #2021 is still fixed)
- [x] Tests

### DIFF
--- a/packages/ra-core/src/sideEffect/recordForm.js
+++ b/packages/ra-core/src/sideEffect/recordForm.js
@@ -14,6 +14,7 @@ export function* handleLocationChange({ payload: { pathname } }) {
         !lastLocationPathName.startsWith(pathname) &&
         !pathname.startsWith(lastLocationPathName)
     ) {
+        lastLocationPathName = pathname;
         yield put(resetForm());
     }
 }
--- a/packages/ra-core/src/sideEffect/recordForm.js
+++ b/packages/ra-core/src/sideEffect/recordForm.js
@@ -2,23 +2,35 @@ import { put, takeEvery } from 'redux-saga/effects';
 import { LOCATION_CHANGE } from 'react-router-redux';
 import { resetForm } from '../actions/formActions';
 
-let lastLocationPathName = undefined;
+export const handleLocationChangeFactory = (
+    initialLocationPathname = undefined
+) => {
+    let lastLocationPathname = initialLocationPathname;
 
-export function* handleLocationChange({ payload: { pathname } }) {
-    if (!lastLocationPathName) {
-        lastLocationPathName = pathname;
-        return;
+    function* handleLocationChange({ payload: { pathname } }) {
+        // We don't want to reset the form when switching tabs in TabbedForm
+        // so we store the current location. If it was the first location change,
+        // we have nothing left to do
+        if (!lastLocationPathname) {
+            lastLocationPathname = pathname;
+            return;
+        }
+
+        // We check wether we are changing from a tab path to the form root path
+        // or vice-versa
+        if (
+            !lastLocationPathname.startsWith(pathname) &&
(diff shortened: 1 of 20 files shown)


djhi:
It might be better to target the `next` branch as I introduce a new action and a new saga... @fzaninotto, your call :)

fzaninotto:
yep, please target the `next` branch

packages/ra-core/src/sideEffect/recordForm.js
```diff
@@ -0,0 +1,36 @@
...
+        // we have nothing left to do
+        if (!lastLocationPathname) {
+            lastLocationPathname = pathname;
+            return;
+        }
+
+        // We check wether we are changing from a tab path to the form root path
+        // or vice-versa
+        if (
+            !lastLocationPathname.startsWith(pathname) &&
```
fzaninotto: I'm afraid of side effects like switching from `/` to `/posts`, or from `/posts` to `/posts_comments`, or from `/posts` to `/posts/123`, which should trigger a form reset but won't...

Can't you be a bit more specific with your test?

djhi: Why reseting the `record-form` in those cases?

- from `/` to `/posts` (there are no `record-form` on any)
- from `/posts` to `/posts/123` (there are no `record-form` on the first and coming from any other page which may have contained one will triggered a reset)
- from `/posts` to `/posts/123` (same)

fzaninotto:
You never know if developers will add the ability to edit a record directly from the list. Be as conservative as possible, so as to avoid side effects. So if you don't want to dispatch the reset between tabs, use a regexp to validate that the url is, indeed, a tab URL.

PR #2089, opened 2018-07-27
[RFR] ra-treeview

djhi:
> An addon which add treeview related components to react-admin

- [x] Simple usage
```js
import React from 'react';
import {
    List,
    EditButton,
    DeleteButton,
    TextField,
} from 'react-admin';
import { Tree } from 'ra-materialui-treeview';

export const CategoriesList = (props) => (
    <List {...props}>
        <Tree>
            <TextField source="name" />
            <EditButton />
            <DeleteButton />
        </Tree>
    </List>
);
```

- [x] Edition usage
```js
import {
    List,
    EditButton,
    DeleteButton,
    SaveButton,
    TextInput,
} from 'react-admin';
import { EditableTree } from 'ra-materialui-treeview';

const CategoriesList = () => (
    <List {...this.props}>
        <EditableTree>
            <TextInput source="name" /> {/* Quick name edition */}
            <SaveButton />
            <EditButton />
            <DeleteButton />
        </EditableTree>
    </List>
);
```

- [x] Deletion
- [x] Keep the nodes opened state after refresh (after deletion for example)
- [x] Split `Tree` in two 2 components: `Tree` and `EditableTree`
- [x] submit on enter (it currently do the HTML form submission when set to `false`, same issue as #1797)
- [x] README ra-tree-core
- [x] README ra-tree-ui-materialui
- [x] Tests

Imgur


packages/ra-materialui-treeview/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```sh
+npm install --save react-beautiful-dnd
+# or
+yarn add react-beautiful-dnd
+```
+
+## Usage
+
+With a category ressource having this shape:
```
fzaninotto: "this shape": be more precise. We're talking about a tree structure, where the parent-child relationship is stored using a `parent_id` field.

packages/ra-materialui-treeview/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```json
```
fzaninotto: use js onstead of json, or make it valid JSON (missing quotes).

packages/ra-materialui-treeview/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+export const CategoriesList = (props) => (
+    <List {...props}>
+        <Treeview>
+            <TextField source="name" />
+        </Treeview>
+    </List>
+);
+```
+
```
fzaninotto: probably missing an explanation about what react-admin does (fetch the entire list, reconstitute the tree based on the `parent_id`, then display the root) and a screenshot about what's possible after that.

packages/ra-materialui-treeview/README.md
```diff
@@ -0,0 +1,96 @@
...
+    EditButton,
+    DeleteButton,
+    SaveButton,
+    TextInput,
+} from 'react-admin';
+import Treeview from 'ra-materialui-treeview';
+
+const CategoriesList = () => (
+    <List {...this.props}>
+        <Treeview>
```
fzaninotto: In my opinion, it feels weird that the editable and non-editable versions share the same component. Does that mean that you add a form even though I didn't ask for it? These should be two separate components.

packages/ra-materialui-treeview/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "homepage": "https://github.com/marmelab/react-admin#readme",
+    "scripts": {
+        "build": "rimraf ./lib && cross-env NODE_ENV=production babel --quiet ./src -d ./lib --ignore spec.js,test.js",
+        "watch": "rimraf ./lib && cross-env NODE_ENV=production babel --watch ./src -d ./lib --ignore spec.js,test.js"
+    },
+    "peerDependencies": {
+        "react": "~16.0.0 || ~16.1.0 || ~16.2.0 || ~16.3.0",
+        "react-admin": "^2.0.0",
+        "react-beautiful-dnd": "^8.0.5",
+        "react-dom": "~16.0.0 || ~16.1.0 || ~16.2.0 || ~16.3.0"
```
fzaninotto: you can make it `^16.4` right away

packages/ra-materialui-treeview/README.md
fzaninotto: probably missing an explanation about what react-admin does (fetch the entire list, reconstitute the tree based on the `parent_id`, then display the root) and a screenshot about what's possible after that.

packages/ra-materialui-treeview/src/Treeview.js
```diff
@@ -0,0 +1,145 @@
...
+            classes,
+            data: { fetchedAt, ...data },
+            getHierarchizedData,
+            parentSource,
+            resource,
+            theme,
+        } = this.props;
+
+        if (data) {
+            const hierarchizedData = getHierarchizedData(
```
fzaninotto: replace `getHierarchizedData` by `getTreeFromArray`?

fzaninotto: Also, a `const` inside an `if` block isn't a good sign.

packages/ra-materialui-treeview/src/Treeview.js
```diff
@@ -0,0 +1,145 @@
...
+            return (
+                <List
+                    classes={{
+                        root: classes.root,
+                    }}
+                    dense
+                    disablePadding
+                >
+                    {hierarchizedData.map(node => (
+                        <TreeviewNode
```
fzaninotto: I'd rename `TreeviewNode` to `TreeNode`. 90% of the React components we're creating concern the view, so we shouldn't prefix them all.

packages/ra-materialui-treeview/src/TreeviewNode.js
```diff
@@ -0,0 +1,116 @@
...
+    basePath,
+    classes,
+    children,
+    node,
+    onChange,
+    resource,
+    theme,
+}) =>
+    node.__children.length > 0 ? (
+        <ListItem
```
fzaninotto: Very long ternary expression, sign that you should probably create two components for the two cases.

packages/ra-materialui-treeview/src/Treeview.js
```diff
@@ -0,0 +1,145 @@
...
+                            key={node.id}
+                            basePath={basePath}
+                            classes={{
```
fzaninotto: double mustache, which will be bad for rendering speed. I'd say: let the nodes be responsible for their own style

packages/ra-materialui-treeview/src/TreeviewNode.js
```diff
@@ -0,0 +1,116 @@
...
+            button
+            classes={{
```
fzaninotto: why don't you pass the full `classes` object then?

packages/ra-materialui-treeview/src/TreeviewNode.js
```diff
@@ -0,0 +1,116 @@
...
+                root: classes.root,
+            }}
+            dense
+            disableGutters
+            style={{ paddingLeft: theme.spacing.unit * 4 }}
```
fzaninotto: is this `style` and injected `theme` temporary?

packages/ra-materialui-treeview/src/TreeviewNode.js
```diff
@@ -0,0 +1,116 @@
...
+                            >
+                                {children}
+                            </TreeviewNode>
+                        ))}
+                    </List>
+                </ExpansionPanelDetails>
+            </ExpansionPanel>
+        </ListItem>
+    ) : (
+        <ListItem
```
fzaninotto: this one is the same as in the previous case. Shouldn't the ternary expression be moved down one level?

examples/simple/src/tags/TagList.js
```diff
@@ -0,0 +1,26 @@
...
+    List,
+    SaveButton,
+    ShowButton,
+    TextInput,
+} from 'react-admin';
+import { Treeview, TreeviewNodeActions } from 'ra-materialui-treeview';
+
+const TagList = props => (
+    <List {...props} perPage={1000}>
+        <Treeview parentSource="parent_id">
```
fzaninotto: I'd rename `Treeview` to `Tree`

packages/ra-materialui-treeview/README.md
djhi: Done

packages/ra-materialui-treeview/README.md
djhi: Yes, that's what it means. Does it matter ?

fzaninotto:
You don’t have to split packages, but separating the logic and the presentation will make your component easier to override

packages/ra-materialui-treeview/README.md
fzaninotto: Yep, I think it matters when we use one component for two purposes. Perhaps it can be the developer’s choice (using an EditableTreeNode as child instead of a TreeNode)?

packages/ra-materialui-treeview/src/Treeview.js
fzaninotto: Because the variable is undefined if the condition isn’t met. Prefer a ternary expression, to give the variable a meaningful value in all cases.

packages/ra-materialui-treeview/README.md
djhi: done below

packages/ra-materialui-treeview/src/TreeviewNode.js
fzaninotto: But this one doesn’t depend on props...

packages/ra-materialui-treeview/src/TreeviewNode.js
djhi: To avoid unknown classes warnings from JSS

packages/ra-materialui-treeview/src/TreeviewNode.js
djhi: woops ^^ good catch! It was before and I forgot to update it

djhi:
> You don’t have to split packages, but separating the logic and the presentation will make your component easier to override

That's actually a good idea to split packages like we do with the core IMO

djhi:
@fzaninotto Should I revert the work on drag and drop from this PR and start a new one for this feature?

djhi:
Ok so I reverted the changes related to drag & drop from this PR and will start a new one later for that feature.

I tried to implement it with https://github.com/atlassian/react-beautiful-dnd but it is currently difficult (impossible?) to support trees (see https://github.com/atlassian/react-beautiful-dnd/issues/302 and https://github.com/atlassian/react-beautiful-dnd/issues/498#issuecomment-395589945).

In the next PR, I will work on a less sexy implementation using buttons (:scream: ) to move items around. Here are the options I'm considering:

- A **Move** button on each node which, when clicked, makes a **Select** button to appear on every node but itself, and change itself to **Cancel**

- A **Move** dropdown button on each node allowing to select a parent with autocomplete (downshift :heart:)

I think the first option is better UX wise.

dwaltrip:
@djhi Hey, popping in here from react-beautiful-dnd Issue 302.

I came up with a workaround for giving each Droppable it's own `onDragEnd` handler, which might be all you need for a sane implementation in a tree-based UI. I've had this solution deployed for a week now on my project, and it seems to be working pretty well. Perhaps it would make `react-beautiful-dnd` viable for your tree-view.

Here's the code: https://gist.github.com/dwaltrip/005a520aeaebe328a8a33998be2babc4

To use it, just change your import statement to `import { Draggable, Droppable, DragDropContext } from '../some/path/wrapped-react-beautiful-dnd';`. Then you can use isolated `onDragEnd` handlers for each `Droppable`.

Feel free to ignore this if it's not helpful. I thought it couldn't hurt to make the suggestion. I almost had to use `move up` / `move down` buttons in my own my project, which I was not terribly excited about =)

djhi:
Hey @dwaltrip, thanks for popping in and for the suggestion :)

So you implemented the workaround explained in the second issue? It looks great! I have been wondering whether drag & drop was actually a good thing though...

Anyway, I'll give a try with the code you posted but will probably end up implementing both options anyway :)

christiaanwesterbeek:
We've been using `react-beautiful-dnd` with react-admin too in a custom CardList (like Datagrid, but with cards). We moved to `react-sortable-hoc`. Both have feature pairity. The latter however has 2 added benefits.

1. `react-beautiful-dnd` only does lists (either horizontal or vertical), but not grids. A design guru that I work with has tried everything. react-sortable-hoc works with both grids and lists.
2. `react-sortable-hoc` has an easier api.

Compare 
```
// react-beautiful-dnd
<DragDropContext onDragEnd={this.onDragEnd}>
  <Droppable droppableId="droppable">
    {(provided, snapshot) => (
      <div
        ref={provided.innerRef}
        style={getListStyle(snapshot.isDraggingOver)}
      >
        {this.state.items.map((item, index) => (
          <Draggable key={item.id} draggableId={item.id} index={index}>
            {(provided, snapshot) => (
              <div
                ref={provided.innerRef}
                {...provided.draggableProps}
                {...provided.dragHandleProps}
                style={getItemStyle(
                  snapshot.isDragging,
                  provided.draggableProps.style
                )}
              >
                {item.content}
              </div>
            )}
          </Draggable>
        ))}
        {provided.placeholder}
      </div>
    )}
  </Droppable>
</DragDropContext>

// It gets worse if you work with seperate component for a DragHandle.
```

```
// react-sortable-hoc
const SortableList = SortableContainer(({items}) => {
  return (
    <ul>
      {items.map((value, index) => (
        <SortableItem
          key={`item-${index}`}
          index={index}
          sortIndex={index}
          value={value}
        />
      ))}
    </ul>
  );
});

// It's painless if you want a seperate component for a DragHandle.
```
_The comparison is not completely fair, because the first includes stuff about styling, but the latter really is almost painless._

You'll thank me if you've transitioned.

djhi:
Wow, I didn't know about this one, it looks great too. Thanks @christiaanwesterbeek ! I'll definitely investigate it too!

djhi:
@christiaanwesterbeek Thanks for the info but unfortunately, `react-sortable-hoc` is only for **sorting** items, while we want to also **move** them (dropping a node on another one should make the dropped node a child of the targeted node)

christiaanwesterbeek:
> `react-sortable-hoc` is only for sorting

I didn't realize that. I see now there's some good progress made on moving between containers, but that isn't merged yet. The community seems pretty active about it though
https://github.com/clauderic/react-sortable-hoc/pull/138

fzaninotto:
Would you mind rebasing on `next`?

djhi:
Closing this for now, I worked on a better version in the following repository:
https://github.com/djhi/ra-tree

We will probably include it in the core later

PR #2099, opened 2018-07-30
[RFR] Make List even easier to override

fzaninotto:
Developers wanting to customize the List layout are often intimidated by the amount of code they have to copy from the `ListView` component. Continuing on #2028, this PR simplifies the `ListView` code even more, by moving some of its logic to sub components (`Datagrid`, `SimpleList`, and `Pagination`). It means it's not strictly backwards compatible: for people having their own Grid or Pagination component, the display might be wrong when the filters return no result, or when the last page is exceeded. But I think the added benefit may qualify it for the 2.2 release.

As an example of easier override, this PR introduces the `TabbedDatagrid`, a way to turn a filter into Excel-like tabs. It's a custom component used by the Demo on the Orders list.

kapture 2018-07-30 at 15 46 52

The `ListView` source is now as simple as

```jsx
export const ListView = ({
    // component props
    actions = <DefaultActions />,
    filters,
    bulkActions = <DefaultBulkActions />,
    pagination = <DefaultPagination />,
    // overridable by user
    children,
    className,
    classes = {},
    exporter,
    title,
    ...rest
}) => {
    const { defaultTitle, isLoading, version, total } = rest;
    const controllerProps = getListControllerProps(rest);
    const titleElement = <Title title={title} defaultTitle={defaultTitle} />;
    return (
        <div
            className={classnames('list-page', classes.root, className)}
            {...sanitizeRestProps(rest)}
        >
            <Card>
                <Header
                    className={classes.header}
                    title={titleElement}
                    actions={React.cloneElement(actions, {
                        ...controllerProps,
                        className: classes.actions,
                        bulkActions,
                        exporter,
                        filters,
                    })}
                />
                {filters &&
                    React.cloneElement(filters, {
                        ...controllerProps,
                        context: 'form',
                    })}
                <div key={version}>
                    {children &&
                        React.cloneElement(children, {
                            ...controllerProps,
                            hasBulkActions: !!bulkActions,
                        })}
                    {pagination &&
                        React.cloneElement(pagination, controllerProps)}
                </div>
            </Card>
        </div>
    );
};
```


packages/ra-ui-materialui/src/list/List.js
```diff
@@ -113,48 +111,27 @@ export const ListView = ({
...
-                ) : (
-                    <CardContent className={classes.noResults}>
-                        <Typography variant="body1">
-                            {translate('ra.navigation.no_results')}
-                        </Typography>
-                    </CardContent>
-                )}
+                <div key={version}>
+                    {children &&
+                        React.cloneElement(children, {
```
djhi: Isn't it a breaking change to not handle the no results case anymore ?

fzaninotto: If you use a custom datagrid, it is a small breaking change because it will show the datagrid headers with empty content. The ra datagrid has been updated to show nothing in this case. And in the case of the demo, the custom image grid for posters still works as before... because it has no headers.

If you use a custom pager, it is a small breaking change because it will not show the out of bounds message when filtering too much or going to a non-existing page. The ra Pagination has been updated to show the messages in this case. My opinion is that custom pagination components are rare, so we should no worry too much about it.

So it's not a BC break if you use the ra components. But we'll need to mention it in the blog post of the release.

djhi:
Nice example on the demo!

PR #2100, opened 2018-07-30
[RFR] Drag & drop support for EditableTree

djhi:
Follow #2089 

Imgur

- [x] Make it work
- [x] Make it right
- [x] Make it fast (not as fast as I'd like yet but acceptable with a throttled CPU x6). Note there is a big difference between development and production.
- [x] Make it beautiful by default
- [x] Add a top level dropzone to transform a node into a top level node
- [x] Fix the damn bug when dropping over a `TextInput` (it clears the value)


djhi:
Closing this for now, I worked on a better version in the following repository:
https://github.com/djhi/ra-tree

We will probably include it in the core later

PR #2119, opened 2018-08-01
[RFR] Fix mobile layout

fzaninotto:
- [x] Fix tab bar height
- [x] Fix toolbar alignment

|Before|After|
|------|-----|
|localhost_8080_ iphone 6_7_8|localhost_8080_ iphone 6_7_8 1|

### DIFF
--- a/packages/ra-ui-materialui/src/layout/ViewTitle.js
+++ b/packages/ra-ui-materialui/src/layout/ViewTitle.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import CardContent from '@material-ui/core/CardContent';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/layout/ViewTitle.js
```diff
@@ -10,11 +10,14 @@ import AppBarMobile from './AppBarMobile';
...
-                title={title}
-                {...rest}
-            />
+            <Fragment>
+                <AppBarMobile
+                    className={classnames('title', className)}
+                    title={title}
+                    {...rest}
+                />
+                <span> </span>
```
djhi: why the span ?

fzaninotto: To enable the `justify-content: space-between` to work even though the first element of the list (the title) is removed. Therefore I add another element to be the first, otherwise the Actions become the first element and are aligned left.

PR #2120, opened 2018-08-01
[RFR] Allow the FormDataConsumer to be used inside an ArrayInput

### LINKED ISSUE #1938: FormDataConsumer passes undefined as source inside ArrayInput

djhi:
Fix #1938

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -33,11 +33,25 @@ describe('Create Page', () => {
         );
     });
 
-  it('should have a working array input with references', () => {
+    it('should have a working array input with references', () => {
         cy.get(CreatePage.elements.addAuthor).click();
         cy.get(CreatePage.elements.input('authors[0].user_id')).should(
             el => expect(el).to.exist
         );
+        cy.get(CreatePage.elements.input('authors[0].role')).should(
+            el => expect(el).to.not.exist
+        );
+    });
+
+    it.only('should have a working array input with a scoped FormDataConsumer', () => {
+        cy.get(CreatePage.elements.addAuthor).click();
+        CreatePage.setValues([
+            {
+                type: 'input',
+                name: 'authors[0].user_id',
+                value: 'Annamarie Mayer{enter}',
+            },
+        ]);
         cy.get(CreatePage.elements.input('authors[0].role')).should(
             el => expect(el).to.exist
         );
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -43,7 +43,7 @@ describe('Create Page', () => {
         );
     });
 
-    it.only('should have a working array input with a scoped FormDataConsumer', () => {
+    it('should have a working array input with a scoped FormDataConsumer', () => {
         cy.get(CreatePage.elements.addAuthor).click();
         CreatePage.setValues([
             {
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -1296,24 +1296,26 @@ const PostEdit = (props) => (
     <Edit {...props}>
         <SimpleForm>
             <ArrayInput source="authors">
-                <TextInput source="name" />
+                <SimpleFormIterator>
+                    <TextInput source="name" />
 
-                <FormDataConsumer>
-                    {({
-                        formData, // The whole form data
-                        scopedFormData, // The data for this item of the ArrayInput
-                        getSource, // A function to get the valid source inside an ArrayInput
-                        ...rest,
-                    }) =>
-                        scopedFormData.name ? (
-                            <SelectInput
-                                source={getSource('role')} // Will translate to "authors[0].role"
-                                choices={['main', 'coauthor']}
-                                {...rest}
-                            />
-                        ) : null
-                    }
-                </FormDataConsumer>
+                    <FormDataConsumer>
+                        {({
+                            formData, // The whole form data
+                            scopedFormData, // The data for this item of the ArrayInput
+                            getSource, // A function to get the valid source inside an ArrayInput
+                            ...rest,
+                        }) =>
+                            scopedFormData.name ? (
+                                <SelectInput
+                                    source={getSource('role')} // Will translate to "authors[0].role"
+                                    choices={['main', 'coauthor']}
+                                    {...rest}
+                                />
+                            ) : null
+                        }
+                    </FormDataConsumer>
+                </SimpleFormIterator>
             </ArrayInput>
         </SimpleForm>
     </Edit>
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -1284,6 +1284,42 @@ const OrderEdit = (props) => (
 ); 
 ```
 
+**Tip**: When using a `FormDataConsumer` inside an `ArrayInput`, the `FormDataConsumer` will provide two additional properties to its children function:
+
+- `scopedFormData`: an object containing the current values of the currently rendered item from the `ArrayInput`
+- `getSource`: a function which will translate the source into a valid one for the `ArrayInput`
+
+```jsx
+import { FormDataConsumer } from 'react-admin';
+
+const PostEdit = (props) => (
+    <Edit {...props}>
+        <SimpleForm>
+            <ArrayInput source="authors">
+                <TextInput source="name" />
+
+                <FormDataConsumer>
+                    {({
+                        formData, // The whole form data
+                        scopedFormData, // The data for this item of the ArrayInput
+                        getSource, // A function to get the valid source inside an ArrayInput
+                        ...rest,
+                    }) =>
+                        scopedFormData.name ? (
+                            <SelectInput
+                                source={getSource('role')} // Will translate to "authors[0].role"
+                                choices={['main', 'coauthor']}
+                                {...rest}
+                            />
+                        ) : null
+                    }
+                </FormDataConsumer>
+            </ArrayInput>
+        </SimpleForm>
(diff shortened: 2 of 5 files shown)


cypress/integration/create.js
```diff
@@ -33,11 +33,25 @@ describe('Create Page', () => {
...
         cy.get(CreatePage.elements.addAuthor).click();
         cy.get(CreatePage.elements.input('authors[0].user_id')).should(
             el => expect(el).to.exist
         );
+        cy.get(CreatePage.elements.input('authors[0].role')).should(
+            el => expect(el).to.not.exist
+        );
+    });
+
+    it.only('should have a working array input with a scoped FormDataConsumer', () => {
```
fzaninotto: `.only` spotted

docs/Inputs.md
```diff
@@ -1284,6 +1284,42 @@ const OrderEdit = (props) => (
...
+- `getSource`: a function which will translate the source into a valid one for the `ArrayInput`
+
+```jsx
+import { FormDataConsumer } from 'react-admin';
+
+const PostEdit = (props) => (
+    <Edit {...props}>
+        <SimpleForm>
+            <ArrayInput source="authors">
+                <TextInput source="name" />
```
fzaninotto: missing `SimpleFormIterator`

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -42,8 +43,58 @@ const REDUX_FORM_NAME = 'record-form';
...
  * );
  */
-const FormDataConsumer = ({ children, formData, ...rest }) => {
-    const ret = children({ formData, ...rest });
+const FormDataConsumer = ({ children, formData, source, ...rest }) => {
+    let scopedFormData = formData;
+    let getSource;
+    let getSourceHasBeenCalled = false;
+
+    // If we have a source, we are in an ArrayInput component
```
fzaninotto: I'm not very confident about this assumption. It seems an indirect way of detecting that we are in an `ArrayInput` component, based on the current `SimpleFormIterator` implementation. Isn't there a more robust way to detect it?

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -42,8 +43,58 @@ const REDUX_FORM_NAME = 'record-form';
...
+    if (source) {
+        const matches = ARRAY_SOURCE_REGEX.exec(source);
+
+        if (matches.length > 2) {
+            const [, arraySource, index] = matches;
```
fzaninotto: I'm surprised that it lints

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -42,8 +43,58 @@ const REDUX_FORM_NAME = 'record-form';
...
+            scopedFormData = formData[arraySource][parseInt(index)];
+            getSource = scopedSource => {
+                getSourceHasBeenCalled = true;
+                return `${arraySource}[${index}].${scopedSource}`;
```
fzaninotto: you should use `lodash.get()`, because there may be a nested object, e.g. `<TextInput source="author.name" />`

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { getFormValues } from 'redux-form';
 
 const REDUX_FORM_NAME = 'record-form';
+const ARRAY_SOURCE_REGEX = /^(.*)\[(\d*)\]/;
```
fzaninotto: I think the actual shape differs a bit. When I look at `SimpleFormIterator`'s way of recreating the source, it reads like this:

```
${member}.${input.props.source}
```

Where `member` is something like `backlinks[3]`.

So when used with a `SimpleFormIterator`, which has no `source`, the result has a closing dot (``backlinks[3].`). This `source` doesn't make sense, and I don't think you should use it inside `SimpleFormIterator`. Instead, you should detect in `SimpleFormIterator` when the child has no `source`, and inject the `index` instead.

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -42,8 +43,58 @@ const REDUX_FORM_NAME = 'record-form';
...
+    if (
+        ret &&
+        !getSourceHasBeenCalled &&
+        process.env.NODE_ENV !== 'production'
+    ) {
+        console.warn( // eslint-disable-line
+            `You have an input inside a FormDataConsumer inside an ArrayInput and you did not called the getSource function supplied by the FormDataConsumer component. This is required for your input to get the proper source and name.
+
+            <ArrayInput source="users">
+                <TextInput source="name" />
```
fzaninotto: Missing `SimpleFormIterator`

packages/ra-core/src/form/FormDataConsumer.js
```diff
@@ -42,8 +43,58 @@ const REDUX_FORM_NAME = 'record-form';
...
+
+    if (
+        ret &&
+        !getSourceHasBeenCalled &&
+        process.env.NODE_ENV !== 'production'
+    ) {
+        console.warn( // eslint-disable-line
+            `You have an input inside a FormDataConsumer inside an ArrayInput and you did not called the getSource function supplied by the FormDataConsumer component. This is required for your input to get the proper source and name.
+
+            <ArrayInput source="users">
```
fzaninotto: the auto formatting added *a lot* of leading spaces on each line, which I don't think will look good on the console.

djhi: It was fine but I extracted it anyway to be sure :)

screen

I indented the code on purpose but it may be not so nice to copy/paste it with this indentation. What do you think?

packages/ra-core/src/form/FormDataConsumer.js
djhi: Like passing an optional prop?

packages/ra-core/src/form/FormDataConsumer.js
djhi: We're generating a string here, from another string.

packages/ra-core/src/form/FormDataConsumer.js
djhi: In the case of the `FormDataConsumer`, which has no `source`, we end up with `backlinks[3].undefined`

djhi: Thanks to your inputs, I ended up not using any regex at all :+1:

packages/ra-core/src/form/FormDataConsumer.js
paulito415: Hi @djhi,

First, thanks for the awesome `FormDataConsumer` - really central and useful component for react-admin. I have a quick question but couldn't find you on the Discord at reactiflux. 

I have a nested `ArrayInput` both of which have `FormDataConsumer` within. In my inner `FormDataConsumer`, when I set `source={getSource('field')}` on the page it shows it's still getting the outer `getSource` element. Is there a way for me to really get the inside one?

Thanks,
Paul

djhi: Please use StackOverflow as explained in the issue template

paulito415: OK just did - https://stackoverflow.com/questions/60343419/how-to-call-getsource-within-nested-arrayinput-formdataconsumer-in-react-admin. Thanks!

paulito415: Hi @djhi sorry to contact you here again. But other than stack overflow, is there any other place I can ask this question? Are the #help-js and #help-react channels on Discord at reactiflux an appropriate place? I just actually wanted to know whether this is supported. If not, I can quickly change my way of doing this, instead of continuing to bang my head on the wall :) Thanks!

fzaninotto: No, StackOverflow is the only way to get free support. If you have an urgent need for support, you can consider paid support by Marmelab.

paulito415: OK thanks Francois! I was able to solve m problem by entirely restructuring my code / layout.

PR #2122, opened 2018-08-02
[RFR] Fix FileInput documentation

### LINKED ISSUE #2083: onDrop is not working for FileInput
**What you were expecting:**

> The `FileInput` component accepts all react-dropzone properties

According to documentation, `FileInput` should take `onDrop` props of react-dropzone.

**What happened instead:**
But `onDrop` is not working as expected. Maybe due to [this](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/FileInput.js#L175).

**Related code:**
```jsx
  <FileInput
    onDrop={this.onDrop}
    source="files"
    label="Files"
    accept="application/pdf, image/jpeg, image/png"
  >
    <FileField  source="src" title="title" />
  </FileInput>
```

**Environment**

* React-admin version: 2.1.2
* Last version that did not exhibit the issue (if applicable): NA
* React version: 16.3.2
* Browser: Google Chrome

djhi:
Another option would be to get the DropZone supported props from the props transparently (using DropZone.propTypes).

Fix #2083

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -445,7 +445,7 @@ Writing a custom field component for displaying the current value(s) is easy:  i
 
 When receiving **new** files, `FileInput` will add a `rawFile` property to the object passed as the `record` prop of children. This `rawFile` is the [File](https://developer.mozilla.org/en-US/docs/Web/API/File) instance of the newly added file. This can be useful to display informations about size or mimetype inside a custom field.
 
-The `FileInput` component accepts all [react-dropzone properties](https://github.com/okonet/react-dropzone#features), in addition to those of react-admin. For instance, if you need to upload several files at once, just add the `multiple` DropZone attribute to your `<FileInput />` field.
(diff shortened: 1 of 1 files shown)


fzaninotto:
I agree, this is not consistent with the other inputs, and our habit of passing all props down if we don't use them. could you use the second solution instead?

djhi:
> this is not consistent with the other inputs

I believe it actually is. We always pass the `rest` props to the **root** component and the `options` props to the inner component which may need them 

https://marmelab.com/react-admin/Inputs.html#autocompleteinput

> Lastly, `<AutocompleteInput>` renders a meterial-ui `<TextField>` component. Use the `options` attribute to override any of the `<TextField>` attributes

*PS: we have a typo in this paragraph btw :stuck_out_tongue:*

Also true for the `BooleanInput`, the `NullableBooleanInput`, the `CheckboxGroupInput`, the `RadioButtonGroupInput`, the `SelectInput` and the `SelectArrayInput` components

fzaninotto:
You've convinced me!

PR #2123, opened 2018-08-02
[RFR] Fix Autocomplete list is cut off by content area

### LINKED ISSUE #2097: Autocomplete list is cut off by content area
bug_autocomplete

**Environment**

* React-admin version: 2.1.4
* React version: 16.3.0
* Browser: Google Chrome 68.0.3440.75

djhi:
Fixes #2097

Before:
Before

After:
After

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.js
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.js
@@ -4,6 +4,7 @@ import get from 'lodash/get';
 import isEqual from 'lodash/isEqual';
 import Autosuggest from 'react-autosuggest';
 import Paper from '@material-ui/core/Paper';
+import Popper from '@material-ui/core/Popper';
 import TextField from '@material-ui/core/TextField';
 import MenuItem from '@material-ui/core/MenuItem';
 import { withStyles } from '@material-ui/core/styles';
@@ -96,6 +97,8 @@ export class AutocompleteInput extends React.Component {
         selectedItem: null,
         suggestions: [],
     };
+
+    inputEl = null;
 
     componentWillMount() {
         const selectedItem = this.getSelectedItem(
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -272,6 +275,13 @@ export class AutocompleteInput extends React.Component {
 
         const { touched, error, helperText = false } = meta;
 
+        // We need to store the input reference for our Popper element containg the suggestions
+        // but Autosuggest also need this reference (it provides the ref prop)
```
fzaninotto: s/also need/also needs/

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -307,9 +317,14 @@ export class AutocompleteInput extends React.Component {
         const { containerProps, children } = options;
 
         return (
-            <Paper {...containerProps} square>
-                {children}
-            </Paper>
+            <Popper
+                {...containerProps}
```
fzaninotto: I think te `containerProps` should be passed to the `Paper` element, otherwise it's a BC break.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -307,9 +317,14 @@ export class AutocompleteInput extends React.Component {
...
+                open
+                anchorEl={this.inputEl}
+                placement="bottom-start"
+            >
+                <Paper>{children}</Paper>
```
fzaninotto: what happened to the `square` prop?

fzaninotto:
I had to revert this PR as it made the suggestions disappear altogether. So I'm reopening #2097 as it's not fixed.

PR #2124, opened 2018-08-02
[RFR] Prepare for allowing forms inside List

djhi:
In the case of the incoming Tree component for example, we must allow all buttons inside the node edition forms. Most of them don't care about `handleSubmit`, `handleSubmitWithRedirect` and `submitOnEnter`.


fzaninotto:
This seems like a new feature to me. Can't you PR against `next` for this one?

djhi:
We just sanitize more props here. Are you sure you want me to rebase it ?

fzaninotto:
We'll implement that within ra-tree

PR #2128, opened 2018-08-03
[RFR] Decouples optimistic side effects from the action's type

djhi:
The idea is to allow custom actions to be handled by our reducers like the standard crud actions, regarding to their effect on data.

For example, we could have a custom action using a custom fetch type and yet have it update the resource(s) optimistically.

- [x] data reducer
- [x] list reducer
- [x] total reducer
- [x] Tests
- [x] Documentation

### DIFF
--- a/cypress/integration/customFetch.js
+++ b/cypress/integration/customFetch.js
@@ -1,0 +1,42 @@
+import listPageFactory from '../support/ListPage';
+
+describe('Edit Page', () => {
+    const ListCommentPage = listPageFactory('/#/comments');
+
+    it('should handle successful custom FETCH actions', () => {
+        ListCommentPage.navigate();
+
+        // Check the first comment status
+        cy.get(ListCommentPage.elements.commentStatus(1)).should(el =>
+            expect(el.text()).to.contains('rejected')
+        );
+        // Click on the first comment approve button
+        cy.get(ListCommentPage.elements.commentApproveButton(1)).click();
+
+        // Check the first comment status
+        cy.get(ListCommentPage.elements.commentStatus(1)).should(el =>
+            expect(el.text()).to.contains('approved')
+        );
+    });
+
+    it('should handle failed custom FETCH actions', () => {
+        ListCommentPage.navigate();
+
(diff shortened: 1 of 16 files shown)


fzaninotto:
IMO it doesn't feel right to use the metas in a reducer. All our metas are used by sagas. I'd say what you are looking to achieve is a side effect, so it should be handled by a saga, too. And you'll have to 

Also, I don't understand when you'd use that instead of `createOptimistic`. Can you be more specific about use cases?

djhi:
> IMO it doesn't feel right to use the metas in a reducer. All our metas are used by sagas.

We already used them in resource reducers, `fetch` at least as you can see in the diff.

> Also, I don't understand when you'd use that instead of `createOptimistic`. Can you be more specific about use cases?

I don't know what `createOptimistic` is, `startUndoable` ? The use cases are explained in the documentation. Ping me again if that's not clear enough

fzaninotto:
I think #2952 makes this PR obsolete, can you confirm?

djhi:
> I think #2952 makes this PR obsolete, can you confirm?

#2684 re-implemented part of it and I think #2952 only addresses part of it too.

This PR added the ability to have a custom fetch type (for using custom queries in the dataProvider) while having the resource reducers handling the resulting optimistic effects as specified in the action. Let's take the approve/reject comment case as an example:

Usually, we have have a classic crud patch route to which we send an updated comment. The resource reducers use the action payload to optimistically update their data.

But what if you have to call very specific `/comments/approve/:id` and `comments/reject/:id` routes on your API (or graphql mutations) ?
 
First, you need a custom action with a custom fetch type but then the reducers won't see the action at all.
Moreover, you don't need to put anything but the comment's id in the payload of the custom action so the reducers wouldn't know how to update their data anyway.
The thing is, you might know what your backend will return (the updated comment) and would like to optimistically update the resource. 

This PR allowed one to specify the data to use for the optimistic update.

fzaninotto:
Not necessary anymore in v3 as the data hooks allow custom side effects

PR #2131, opened 2018-08-05
Tutorial.md adding prop-types package external dependency

igneel64:
Prop types have become an external dependency so we should instruct a first-time user to install this externally using yarn. This change does just that adding two lines in order to complete this need.

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -200,7 +200,6 @@ UrlField.propTypes = {
 export default UrlField;
 ```
 
-
 ## Handling Relationships
 
 In JSONPlaceholder, each `post` record includes a `userId` field, which points to a `user`:
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -200,6 +200,12 @@ UrlField.propTypes = {
 export default UrlField;
 ```
 
+To take advantage of the prop-types library, you can go ahead and install it.
+
+```sh
+yarn add prop-types
+```
+
 ## Handling Relationships
 
 In JSONPlaceholder, each `post` record includes a `userId` field, which points to a `user`:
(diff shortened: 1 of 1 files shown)


docs/Tutorial.md
```diff
@@ -200,6 +200,12 @@ UrlField.propTypes = {
 export default UrlField;
 ```
 
+To take advantage of the prop-types library, you can go ahead and install it.
+
+```sh
+yarn add prop-types
```
djhi: Would be better to add it at line 23 IMO:

```
yarn add react-admin prop-types
```

igneel64: Can do and I agree with the setup 'before hand'. But since this is not 'required' to have the example done, the reader might as well choose not to use prop-types.  That is also the case because in the whole next part of the tutorial we do not use them again. It would be unwise in my opinion to add it from the start (as this also is a tutorial).
What do you think?

djhi: In that case, I would remove all references to PropTypes. @fzaninotto?

igneel64: If I may, removing them entirely will not be of benefit.
As a 'solution' library as many will see react-admin, it might be used a starting point for someone wanting to either learn React, or has to implement something for the first time on it on his job for example.
As an established library, it would be nice to give the initiative to the reader to see and search for a good practice as prop-types and the tutorial is the right place for it. In my opinion removing this is not the right choice in the 'community' side.

fzaninotto: Let's leave it, but added line 23 as @djhi recommended

docs/Tutorial.md
```diff
@@ -200,6 +200,7 @@ UrlField.propTypes = {
 export default UrlField;
 ```
 
+
```
fzaninotto: no need for this extra line, please remove

fzaninotto:
Thanks!

PR #2137, opened 2018-08-07
Fix Not Bounding correctly <label> and <input> in login form

### LINKED ISSUE #2136: Not Bounding correctly <label> and <input>  in login form
**What you were expecting:**
Bound correctly <label> and <input> tags in login form when using translation

**What happened instead:**
Username and password on login form elements do not have associated labels

**Steps to reproduce:**
Use Login Form, and test it with browser extension that checks accesibility like [Google Lighthouse](https://github.com/GoogleChrome/lighthouse)

**Related code:**

To resolve this issue i will do a PR based on following html code generated via material-ui Field 
```
<label for="username"> <lable>
<input id="username"></input>
<label for="password"> <lable>
<input id="password"></input>

```

**Other information:**
It is very important to validate Accesibility.

**Environment**

* React-admin version: 2.2.0
* Browser: Chrome

josx:
Closes #2136

### DIFF
--- a/packages/ra-ui-materialui/src/auth/LoginForm.js
+++ b/packages/ra-ui-materialui/src/auth/LoginForm.js
@@ -44,6 +44,7 @@ const LoginForm = ({ classes, isLoading, handleSubmit, translate }) => (
         <div className={classes.form}>
             <div className={classes.input}>
                 <Field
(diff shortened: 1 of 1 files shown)


djhi:
Thanks!

PR #2152, opened 2018-08-10
[RFR] Fix FileInput documentation

### LINKED ISSUE #2151: FileInput doesn't accept multiple files when setting 'multiple' property using options prop
**What you were expecting:**
A FileInput that accepts multiple files.

**What happened instead:**
The input only accepts a single file.

**Steps to reproduce:**
Set multiple using the `options` prop rather than the `multiple` prop.

From [the docs](https://github.com/marmelab/react-admin/blob/master/docs/Inputs.md#fileinput):
> The FileInput component accepts an options prop into which you can pass all the react-dropzone properties. For instance, if you need to upload several files at once, just add the options={{ multiple: true }} DropZone attribute to your <FileInput /> field.

**Related code:**
```
<FileInput source="files" options={{ multiple: true }}>
    <FileField source="src" title="title" />
</FileInput>
```

**Other information:**
I will create a pull request for FileInput to check the options prop in addition to the (legacy?) multiple prop. Otherwise, we should update the documentation to mention the multiple prop.

**Environment**
* React-admin version: 2.2.0

lihop:
Closes #2151

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -445,7 +445,8 @@ Writing a custom field component for displaying the current value(s) is easy:  i
 
 When receiving **new** files, `FileInput` will add a `rawFile` property to the object passed as the `record` prop of children. This `rawFile` is the [File](https://developer.mozilla.org/en-US/docs/Web/API/File) instance of the newly added file. This can be useful to display informations about size or mimetype inside a custom field.
 
-The `FileInput` component accepts an `options` prop into which you can pass all the [react-dropzone properties](https://github.com/okonet/react-dropzone#features). For instance, if you need to upload several files at once, just add the {% raw %}`options={{ multiple: true }}`{% endraw %} DropZone attribute to your `<FileInput />` field.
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/input/FileInput.js
```diff
@@ -75,13 +76,15 @@ export class FileInput extends Component {
     }
 
     onDrop = files => {
-        const updatedFiles = this.props.multiple
+        const multiple = this.props.options.multiple || this.props.multiple;
```
djhi: I think we should support only one of the two. As it is a very common usecase, I wonder if it should be `this.props.multiple` in which case, we should merge it in the options before passing them to the `DropZone` component. This should also be documented if chosen.

djhi: The code must be updated as well. We currently check for `this.props.multiple` but it should be `this.props.options.multiple`

djhi: Nvm, we actually pass `this.props.multiple` to the DropZone...

lihop:
Closed in favour of #2235.

PR #2155, opened 2018-08-10
[RFR] UI Diet

### LINKED ISSUE #1500: AppBarMobile is missing from the documentation - is there a way to override it?
We needed to override the default AppBarMobile (https://github.com/marmelab/admin-on-rest/search?utf8=%E2%9C%93&q=AppBarMobile&type=) since it seems that it appears on top of our custom AppBar when in mobile. I assume that this component is internally used .. how is it connected with the rest of the application so it can be overriden?

Thank you

fzaninotto:
## Problem

In List and Edit pages, the content appears very low on the page, mostly because:

1. We write the page title in the *content* (the white paper). In contrast, the App Bar title never changes, and this is wasted space.
2. In the list, the filters are rendered *below* the action buttons, moving the content even further down
3. Action buttons are small but they use large icons

## Solution

1. Display the page title in the AppBar (this requires a [`Portal`](https://reactjs.org/docs/portals.html)).
2. In the list, display the filters where the page title used to be.
3. In buttons, adapt the icon size to the button size

But then the details pages (Edit and show) have only one button, the Delete button, which seems to float there. So:

1. Remove the Delete button from the Show view
2. Move the Delete button in the Toolbar, at the bottom of the form, in the Edit view.

- [x] Rework the `<Title>` component to use a `Portal`
- [x] Display the page title in the `AppBar`
- [x] Display the filters on the top left of the content
- [x] Move the `<Delete>` button down to the `Toolbar` 
- [x] Update `<Button>` to adapt the icon size to the button size
- [x] Make it lint
- [x] Add integration tests for the title
- [x] Make test pass
- [x] Update the doc

Small BC break: custom pages with layout have an empty title. Developers will have to manually define the `<Title>`

|Before|After|
|--|--|
|||

|Before|After|
|--|--|
|||

In Edit and Show views, the Show and Edit buttons still push the content down... only if they exist. That means that if a resource has only a Show or an Edit (and not both), then the form appears right away. IMO, that's the general case.

One side effect is that the App "title" ('like "My Beautiful Admin") doesn't appear very much. By default, it's only on error pages. Of course, it's injected in the Dashboard, so it's easy to display there.

Another side effect is that it's harder to delete one resource. You can't go to the `Show` page anymore, and in the `Edit` page, it's under the form instead of below it (so sometimes below the fold).

Another side effect is that by default, the UI is pretty bare. It's good for the users, who can focus on their content and custom actions. It's not as good for de Waouh effect of a default admin 🙄 

Another side effect is to get rid of alternative mobile versions for the Title and AppBar (closes #1500), and to make the markup simpler.

Overall I fell that this makes the admins better. The best example is the demo:

kapture 2018-08-10 at 16 57 11

Comments are welcome.

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -69,7 +69,7 @@ The `dataProvider` is also the ideal place to add custom HTTP headers, authentic
 
 ## `title`
 
-By default, the header of an admin app uses 'React Admin' as the main app title. It's probably the first thing you'll want to customize. The `title` prop serves exactly that purpose.
+On error pages, the header of an admin app uses 'React Admin' as the main app title. Use the `title` to customize it.
 
 ```jsx
 const App = () => (
@@ -81,17 +81,17 @@ const App = () => (
 
 ## `dashboard`
 
-By default, the homepage of an an admin app is the `list` of the first child `<Resource>`. But you can also specify a custom component instead. To fit in the general design, use Material UI's `<Card>` component, and react-admin's `<ViewTitle>` component:
+By default, the homepage of an an admin app is the `list` of the first child `<Resource>`. But you can also specify a custom component instead. To fit in the general design, use Material UI's `<Card>` component, and react-admin's `<Title>` component to set the title in teh AppBar:
 
 ```jsx
 // in src/Dashboard.js
 import React from 'react';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
-import { ViewTitle } from 'react-admin';
+import { Title } from 'react-admin';
 export default () => (
     <Card>
-        <ViewTitle title="Welcome to the administration" />
+        <Title title="Welcome to the administration" />
         <CardContent>Lorem ipsum sic dolor amet...</CardContent>
     </Card>
 );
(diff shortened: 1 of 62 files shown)


fzaninotto:
Switching to RFR

docs/Admin.md
```diff
@@ -81,17 +81,17 @@ const App = () => (
 
 ## `dashboard`
 
-By default, the homepage of an an admin app is the `list` of the first child `<Resource>`. But you can also specify a custom component instead. To fit in the general design, use Material UI's `<Card>` component, and react-admin's `<ViewTitle>` component:
+By default, the homepage of an an admin app is the `list` of the first child `<Resource>`. But you can also specify a custom component instead. To fit in the general design, use Material UI's `<Card>` component, and react-admin's `<Title>` component to set the title in teh AppBar:
```
Kmaschta: s/an an/an/

Kmaschta: s/teh/the/

packages/ra-ui-materialui/src/button/CreateButton.js
```diff
@@ -40,12 +34,11 @@ const CreateButton = ({
     classes = {},
     translate,
     label = 'ra.action.create',
-    size = 'small',
     ...rest
 }) => (
     <Responsive
         small={
-            <Button
+            <MuiButton
```
djhi: I don't get this change.Why?

packages/ra-ui-materialui/src/layout/RecordTitle.js
```diff
@@ -1,10 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Title from './Title';
+import TitleDeprecated from './TitleDeprecated';
 
+/**
+ * @deprecated Use TitleForRecord instead
+ */
 const RecordTitle = ({ defaultTitle, record, title }) =>
```
djhi: Aren't we missing a console warning here?

djhi:
Awesome!

packages/ra-ui-materialui/src/button/CreateButton.js
fzaninotto: Because we're using our own `Button` for the `medium` size, and the names collide.

packages/ra-ui-materialui/src/layout/RecordTitle.js
fzaninotto: since all occurrences of `RecordTitle` were inside a `Header`, and we already warn about it, that would trigger two warnings in a row... I think one is enough.

packages/ra-ui-materialui/src/button/CreateButton.js
djhi: Oh I did not see that we were using the MUI button for both before... :+1:

PR #2160, opened 2018-08-13
[RFR] Change links from .md to .html

fzaninotto:
.md links only work when browsing the doc via the GitHub file explorer, .html links only work when browsing the Jekyll-generated documentation online. We choose the latter.

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -76,7 +76,7 @@ export const CommentList = (props) =>
     </List>;
 ```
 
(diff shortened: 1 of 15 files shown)


fzaninotto:
No, sorry, this is a bad idea. GitHub already does the change  (see https://blog.github.com/2016-12-05-relative-links-for-github-pages/). It just doesn't work when testing with Jekyll locally.

PR #2164, opened 2018-08-14
[WIP] scroll bar invisible in large list page + AppBar Headroom 

### LINKED ISSUE #1819: Two horizontal scroll in version ^2.0.0-RC3
Hello, I'm using the react-admin version ^2.0.0-RC3, but I get an error on a table with multiple columns with 2 horizontal scroll (in body and table), the following image is my error:

img

I want only horizontal scroll in table and columns is not stretch, so can you support with options in <Resource>

Thank you.

zyhou:
## Description

peek 2018-08-22 16-45
peek 2018-08-23 10-20

- [x] Test Chrome
- [x] Test Firefox
- [x] Test IE
- [x] Test Edge

fix #1819

### DIFF
--- a/packages/ra-ui-materialui/src/layout/AppBar.js
+++ b/packages/ra-ui-materialui/src/layout/AppBar.js
@@ -10,6 +10,7 @@ import { withStyles } from '@material-ui/core/styles';
 import MenuIcon from '@material-ui/icons/Menu';
 import compose from 'recompose/compose';
 import { toggleSidebar as toggleSidebarAction } from 'ra-core';
+import Headroom from 'react-headroom';
 
 import LoadingIndicator from './LoadingIndicator';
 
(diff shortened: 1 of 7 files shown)


fzaninotto:
Not good: this makes the toolbar fixed on desktop, which is not intended.

zyhou:
I like the material design guidelines. I will look into headroom.js.

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -61,6 +62,10 @@ const styles = theme => ({
     },
 });
 
+const styleHeadroom = {
```
djhi: This style is always applied on the `appBar`, why not setting it on the `appBar` class directly ?

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -70,41 +75,45 @@ const AppBar = ({
...
-                {typeof title === 'string' ? title : React.cloneElement(title)}
-            </Typography>
-            {logout &&
-                cloneElement(logout, {
-                    className: classes.logout,
-                })}
-        </Toolbar>
-        <LoadingIndicator className={classes.loadingIndicator} />
-    </MuiAppBar>
+    <Headroom className={classNames(classes.appBar)} style={styleHeadroom}>
```
djhi: Breaking change here, you don't merge the `className` anymore

packages/ra-ui-materialui/src/layout/AppBar.js
zyhou: Because `style` update inline style and I need to force the position, look [headroom/index.js](https://github.com/KyleAMathews/react-headroom/blob/master/src/index.js#L312)

packages/ra-ui-materialui/src/layout/AppBar.js
zyhou: I need advice on this one.

In fact, `className={className}` is on `<MuiAppBar />` components, maybe I have to put on `<Headroom` ?

djhi: Convention is to apply `className` on the root element. This might be a breaking change though

zyhou:
I close this PR, I have to use `next` branch. I do another PR.

zyhou:
Follow : https://github.com/marmelab/react-admin/pull/2206

PR #2165, opened 2018-08-14
[RFR] Inject the open prop to the Menu component

fzaninotto:
Because custom menus can look differently when open or closed. 

My use case: a menu with a sub-level. I want the submenus to have a left padding when the menu is open, but not when it is closed.


djhi:
For some reason, Travis did not update its status on github but the build failed

PR #2167, opened 2018-08-15
Minor Menu Item Style Fixes

ragboyjr:
- Fixes issue of scrunched elements when you there
  are a lot of resources
- Fixes cut off name of resource when it's too long to
  fit in the window

Before:

screen shot 2018-08-15 at 3 40 26 pm

screen shot 2018-08-15 at 3 41 27 pm

After:

screen shot 2018-08-15 at 3 41 43 pm

### DIFF
--- a/packages/ra-ui-materialui/src/layout/MenuItemLink.js
+++ b/packages/ra-ui-materialui/src/layout/MenuItemLink.js
@@ -10,11 +10,14 @@ const styles = theme => ({
         color: theme.palette.text.secondary,
         display: 'flex',
         alignItems: 'flex-start',
+        flexShrink: 0,
+        whiteSpace: 'normal',
+        height: "auto"
     },
     active: {
         color: theme.palette.text.primary,
     },
-    icon: { paddingRight: '1.2em' },
+    icon: { paddingRight: '1.2em', height: "24px" },
 });
 
 export class MenuItemLink extends Component {


ragboyjr:
By the way, the documentation for this project is insanely detailed and helpful. I was able to override the menu styles on in my apps installation within 30 minutes or so. Very very impressed by the level of detail!

djhi:
Hi @ragboyjr! Thanks for your contribution. Can you run the `make prettier` command to ensure our prettier rules are enforced?

ragboyjr:
@djhi updated!

packages/ra-ui-materialui/src/layout/MenuItemLink.js
```diff
@@ -10,6 +10,8 @@ const styles = theme => ({
         color: theme.palette.text.secondary,
         display: 'flex',
         alignItems: 'flex-start',
+        minHeight: 'min-content',
```
djhi: According to https://developer.mozilla.org/fr/docs/Web/CSS/width and https://caniuse.com/#search=min-content, this won't work neither on edge nor IE. Can you think of an alternative?

ragboyjr: Any minHeight would do. I think the height is set at 24px, so setting the min to that value would do the trick as well.

djhi: Btw, isn't `flexShrink: 0;` enough to resolve this ?

ragboyjr: @djhi I want to say no, because if it was, the layout bug wouldn't be there.

djhi: Have you tried to add `flexShrink: 0` to the `root` style ? I had a similar issue on another project and it solved it

djhi: Any news ?

ragboyjr:
@djhi updated to use flexShrink, had to make some minor changes to the height to get it to work perfectly.

djhi:
Just tried to reproduce the issue and was unable to do so. What version of react-admin are you using and on which browser are you seeing this ?

This is what I see on mobile:
mobile

And on desktop:
desktop

ragboyjr:
2.2.0

ragboyjr:
It happens on both chrome/safari.

ragboyjr:
And the issue doesn't show up on desktop, it only affects mobile.

djhi:
> And the issue doesn't show up on desktop, it only affects mobile.

I posted a mobile screenshot too where I can't see the issue either

fzaninotto:
Bump, can you please provide a CodeSandbox demonstrating the issue?

ragboyjr:
@djhi @fzaninotto I haven't had time to test this off of the latest master, it is possible that has been resolved by other changes.

fzaninotto:
OK then, I'm closing this one. Feel free to reopen if you happen to reproduce the issue on the latest master.

PR #2171, opened 2018-08-17
Typo fix for ArrayInput documentation

igneel64:


fzaninotto:
No, that's not a typo. The field allows to edit, not to add. I don't understand your change.

igneel64:
@fzaninotto 
Sorry it must have been an autocomplete mistake for the 'addition'. See new commit.

Out of curiosity I search for 'edition' and if it is can be actually used in our case but I could not find something. So it might be better to express it in this way as shown in the new commit.

fzaninotto:
Thanks!

PR #2174, opened 2018-08-17
Remove const client = new ApolloClient();

kfern:
Hi!

"client" is not used in example, so const client = new ApolloClient() can be removed


djhi:
Thanks!

gl4ssiest:
`const client = new ApolloClient();` still on readme

JulienMattiussi:
@gl4ssiest Better late than never ;)

PR #2175, opened 2018-08-17
[RFR] Bulk actions toolbar

fzaninotto:
Material Design encourages a specific way to present bulk actions, quite different from the one we've implemented (see https://material.io/design/components/data-tables.html#tables-within-cards). 

Bulk actions toolbar

This is better in terms of usability, because the bulk actions appear right after selecting a row (instead of after clicking on the "XXX selected" button, which is a bad button label anyway).

This PR implements this proposal, in a backwards-compatible way, by adding a new `bulkActionButtons` prop to the `<List>` component:

- [x] Create `BulkActionsToolbar` component
- [x] Make the change backwards compatible
- [x] Document it

```jsx
const PostListBulkActions = props => (
    <Fragment>
        <ResetViewsButton {...props} />
        <BulkDeleteButton {...props} />
    </Fragment>
);

const PostList = props => (
    <List {...props} bulkActionButtons={<PostListBulkActions />} >
      ...
   </List>
)
```

kapture 2018-08-17 at 19 52 21

This video shows that a List view that still uses the `bulkActions` prop continues to work.

bulkactionsbuttons


djhi:
This is a lot nicer indeed :+1:

PR #2179, opened 2018-08-19
[RFR] Display warnings only when not in production mode

djhi:


fzaninotto:
I think we have written at least twice `if(process.env.NODE_ENV === 'development'` in the code. Could you fix these ones to make the warnings consistent?

PR #2184, opened 2018-08-20
Added a Danish translation of React-Admin

nikri:
A translated version of the React-Admin language file


nikri:
Cool, I will do that instead of 👍

PR #2203, opened 2018-08-23
[RFR] Fix Select inputs validation

### LINKED ISSUE #2142: SelectInput and SelectArrayInput meta
**What you were expecting:**
When I select my input, if there is an error, the error would be displayed.

**What happened instead:**
The touched meta is always false. So the error is not displayed.
If I submit, the error is displayed but in black color (for the SelectArrayInput)=> When I see on MaterialUi, I suppose the FormControl need the "error" attribute.

```jsx
<FormControl className={classes.formControl} error>
          <InputLabel htmlFor="name-error">Name</InputLabel>
          <Select
            value={this.state.name}
            onChange={this.handleChange}
            name="name"
            renderValue={value => `⚠️  - ${value}`}
            input={<Input id="name-error" />}
          >
            <MenuItem value="">
              <em>None</em>
            </MenuItem>
            <MenuItem value="hai">Hai</MenuItem>
            <MenuItem value="olivier">Olivier</MenuItem>
            <MenuItem value="kevin">Kevin</MenuItem>
          </Select>
          <FormHelperText>Error</FormHelperText>
        </FormControl>
```

**Steps to reproduce:**
For SelectInput, add the required() validator, click and unclick.

For SelectArrayInput , create a valitor : 
```jsx
export const requiredSelectArrayInput = memoize((message = 'ra.validation.required') =>
    Object.assign(
        (value, values, props) =>
            !isEmpty(value) && value.length > 0 ? undefined : getMessage(message, undefined, value, values, props),
        { isRequired: true }
    )
);
```

**Environment**

* React-admin version: 2.2.0
* React version:16.3
* Browser: Chrome

djhi:
Fix #2142

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.js
@@ -137,7 +137,7 @@ export class SelectArrayInput extends Component {
     handleChange = event => {
         this.props.input.onChange(event.target.value);
         // HACK: For some reason, redux-form does not consider this input touched without calling onBlur manually
-        this.props.input.onChange(event.target.value);
+        this.props.input.onBlur(event.target.value);
         this.setState({ value: event.target.value });
     };
     renderMenuItemOption = choice => {
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.js
@@ -136,6 +136,8 @@ export class SelectArrayInput extends Component {
 
     handleChange = event => {
         this.props.input.onChange(event.target.value);
+        // HACK: For some reason, redux-form does not consider this input touched without calling onBlur manually
+        this.props.input.onChange(event.target.value);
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/SelectArrayInput.js
```diff
@@ -135,6 +135,8 @@ export class SelectArrayInput extends Component {
     }
 
     handleChange = event => {
+        this.props.input.onChange(event.target.value);
```
Kmaschta: This is `onChange` not `onBlur`. Hm?

packages/ra-ui-materialui/src/input/SelectInput.js
```diff
@@ -187,7 +190,9 @@ export class SelectInput extends Component {
             );
         }
         const { touched, error, helperText = false } = meta;
+        const helperTextError = (touched && error) || helperText;
 
+        console.log({ meta, helperTextError });
```
fzaninotto: `console.log()` spotted

PR #2206, opened 2018-08-24
[RFR] Implement Headroom in AppBar

### LINKED ISSUE #1819: Two horizontal scroll in version ^2.0.0-RC3

zyhou:
## Description

We have some issue on AppBar component like #1819. 

In fact, we don't follow [ the material design guidelines](https://material.io/design/components/app-bars-top.html#behavior), which specify that the scrollbar should hide on scroll down, and show up (fixed) on scroll up (see https://material-components.github.io/material-components-web-catalog/#/component/top-app-bar/standard for an example).

A discussion on the same subject in the [material-ui tracker: mui-org/material-ui#12337](https://github.com/mui-org/material-ui/issues/12337).

## Solution

I implement [react-headroom](https://github.com/KyleAMathews/react-headroom) container in AppBar component. I export `<Headroom />` component which is already configure for react-admin.

- [x] Clean example
- [x] Documentation
- [x] Add test E2E
- [x] Test Chrome
- [x] Test Firefox
- [x] Test Edge
- [x] Test IE

## Result

peek 2018-08-24 12-11
peek 2018-08-24 12-10

fix #1819

### DIFF
--- a/cypress/integration/list.js
+++ b/cypress/integration/list.js
@@ -15,6 +15,18 @@ describe('List Page', () => {
         });
     });
 
+    describe('Headroom', () => {
+        it('should hide/show the appBar when scroll action appears', () => {
+            cy.viewport(1280, 500);
+
+            cy.scrollTo(0, 200);
+            cy.get(ListPagePosts.elements.headroom).should('not.be.visible');
+
+            cy.scrollTo(0, -100);
+            cy.get(ListPagePosts.elements.headroom).should('be.visible');
+        });
+    });
+
     describe('Pagination', () => {
         it('should display paginated list of available posts', () => {
             cy.contains('1-10 of 13');
--- a/cypress/integration/list.js
+++ b/cypress/integration/list.js
@@ -15,7 +15,7 @@ describe('List Page', () => {
         });
     });
 
-    describe('Headroom', () => {
+    describe('Auto-hide AppBar', () => {
         it('should hide/show the appBar when scroll action appears', () => {
             cy.viewport(1280, 500);
 
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -635,6 +635,9 @@
                                 <a href="#using-a-custom-layout">Using a Custom Layout</a>
                             </li>
                             <li class="chapter">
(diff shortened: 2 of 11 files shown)


cypress/integration/list.js
```diff
@@ -15,6 +15,22 @@ describe('List Page', () => {
...
+            cy.get(ListPagePosts.elements.title).contains('Posts List');
+        });
+
+        it('should hide the appBar when scroll down action appears', () => {
+            cy.scrollTo(0, 500);
+            cy.get(ListPagePosts.elements.headroom).should('not.be.visible');
+        });
+
+        it('should show the appBar when scroll top action appears', () => {
+            cy.scrollTo(0, -500);
```
djhi: Wouldn't it be better to only scroll up a little (less than 500, maybe only 200) ?

djhi: My understanding is that it should appear when you start scrolling up, right ?

zyhou: - Hep, I can 200px or less (like 100px), I try in local

- Yes exactly

docs/_layouts/default.html
```diff
@@ -634,6 +634,9 @@
                             <li class="chapter">
                                 <a href="#using-a-custom-layout">Using a Custom Layout</a>
                             </li>
+                            <li class="chapter">
+                                <a href="#using-a-custom-appbar">Using a Custom AppBar</a>
```
djhi: I think this should be a section of the `Using a Custom Layout` chapter

djhi: Nevermind this one

cypress/integration/list.js
```diff
@@ -15,6 +15,18 @@ describe('List Page', () => {
         });
     });
 
+    describe('Headroom', () => {
```
fzaninotto: s/Headroom/Auto-hide AppBar/

PR #2212, opened 2018-08-24
[RFR] Allow changing the sidebar width using a Layout prop

fzaninotto:
Until now, changing the sidebar width (to accommodate shorter or longer resource names) used to require a lot of boilerplate. No more!

```jsx
const MyLayout = props => <Layout {...props} sidebarWidth={200} />;
```

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -365,19 +365,20 @@ const App = () => (
 );
 ```
 
-Your custom layout can extend the default `<Layout>` component if you only want to override the sidebar width, the appBar, the menu, the notification component, or the error page. For instance:
+Your custom layout can extend the default `<Layout>` component if you only want to override the sidebar, the appBar, the menu, the notification component, or the error page. For instance:
 
 ```jsx
 // in src/MyLayout.js
 import { Layout } from 'react-admin';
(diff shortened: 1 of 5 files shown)


fzaninotto:
Do you mean:

```jsx
const MySidebar = props => <Sidebar {...props} sidebarWidth={200} />;
const MyLayout = props => <Layout {...props} sidebar={<MySidebar />} />;
```

?

djhi:
Yes, exactly

djhi:
Shouldn't we allow to override the closed state width as well ?

fzaninotto:
wait, something's wrong

docs/Theming.md
```diff
@@ -405,6 +407,19 @@ const MyLayout = props => <Layout
 />;
 ```
 
+You can speciy the `Sidebar` size by setting the `size` property:
```
djhi: Shoulnd't we document the `closedSize` prop too ?

fzaninotto: The current documentation for this type of theming is already verbose and unclear. We'll probably need to rewrite it somehow.

Besides, it's a really corner case IMO. We'll document it when we have an API doc like docz. Until then, let's consider it an easter egg.

PR #2218, opened 2018-08-27
[RFR] Remove babel stage-0 and use babel-preset-env

### LINKED ISSUE #2210: Remove `stage-0` preset in `.babelrc`
Babel is [deprecating `stage-` presets](https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets) and it makes sense.

React-admin uses the `stage-0` preset. We should replace it with the features we actually use.

Kmaschta:
Fix #2210

I'm not sure about the `last 2 versions`, it includes a huge list of browsers.


fzaninotto:
`babel-preset-env` adds support for `babel-preset-es2015`, `babel-preset-es2016`, and `babel-preset-es2017`. It doesn't add support for all the `stage-0` features. We need to carefully check which ones we used in react-admin, and add the required preset. After a rapid check, that means at least:

- babel-plugin-transform-export-extensions (stage 1)
- babel-plugin-transform-class-properties (stage 2)
- babel-plugin-transform-object-rest-spread (stage 3)
- babel-plugin-syntax-trailing-function-commas (stage 3)
- babel-plugin-transform-async-to-generator (stage 3)

Kmaschta:
Review applied.

fzaninotto:
Can you only add the plugins for the syntax we actually use in the simple example?
Can you add the related dependencies in the simple example package.json?
Can you also make sure the `tutorial` example is correct?

Kmaschta:
Review applied. (waiting for GitHub to notice it)

`tutorial` is OK thanks to `react-scripts`

Issue #1958, opened 2018-06-26
Any plan on supporting Prisma ?


Weakky:
Hey there,

Is there any plan on crafting an adaptator for [Prisma](https://prisma.io)'s generated GraphQL API ?

`react-admin` already supports Graphcool's GQL API. Although Prisma should 
normally be used as a data-layer behind a handcrafted GraphQL server, I think it should be ok to consider reaching the endpoint directly when dealing with backoffice CRUD stuff.

Thanks!

djhi:
There are no plan for supporting Prisma by ourselves. However, with `ra-data-graphql` you should be able to roll your own. Read the `ra-data-graphcool` code for inspiration

YannickB:
I'd like to second this demand, it seems that Prisma is the future for graph.cool anyway https://www.prisma.io/forum/t/graphcool-framework-and-prisma/2237

The Prisma API https://www.prisma.io/docs/reference/prisma-api/overview-ohm2ouceuj seems the natural evolution of the GraphCool API, with some good improvements, and they are clearly trying to establish a standard for Graphql CRUD operations https://github.com/opencrud/opencrud.

IMO, it totally make sense that react-admin provide a driver for Prisma/OpenCRUD in its core, and follow the standard they are establishing.

YannickB:
@Weakky Great! I am not focused on it right now but I definitely will later.

Weakky:
@YannickB I completely forgot about *OpenCRUD* however. I totally agree that this provider should not be prisma focused, but more *OpenCRUD* focused. This will allow to reach a wider audience in the future, starting by Prisma *and* GraphCMS.

djhi:
I built `ra-data-graphcool` as a way to test the `ra-data-graphql` package and because I personally really liked graphcool (and Prisma) and used it on a side project. Unfortunately, we don't use Prisma on any projects for now, and neither do I, so we won't work on a Prisma dataProvider any time soon.

Please keep in mind that this is an open source project. We work on it when we can or when some features are required for our customers. You're welcome to become one of them though :wink:. Contact us at this address: [contact@marmelab.com](mailto:contact@marmelab.com)

@Weakky I love your README! This is a great explanation of `ra-data-graphql`! Would you mind opening a PR to improve its documentation? If you don't have time for it, would you mind if I shamelessly copy parts of yours? :heart: 

[...]

PR #2220, opened 2018-08-27
[RFR] Add a data provider for Prisma in documentation

Weakky:
As discussed in #1958, here's my PR to reference the data provider for Prisma


fzaninotto:
Thanks!

PR #2222, opened 2018-08-27
Add 'How does it work' section to ra-data-graphql README

Weakky:
As discussed in #1958, here's a modified version of the `ra-data-graphql` explanations, even though I said that I'd let you write it on your own.

If there are any updates or content that you'd like to add/remove, feel free to ask


Weakky:
Embarrassing mistakes, my bad. Typos should be removed now

PR #2226, opened 2018-08-27
[RFR] Fix TabbedLayout, TabbedForm and List display regarding scroll

### LINKED ISSUE #2094: TabbedForm overflows viewport when there are a lot of FormTabs
**What you were expecting:**
When you add a lot of FormTabs in a TabbedForm inside a Edit, I expect a behavior like this: [Material UI Scrollable Tabs](https://material-ui.com/demos/tabs/#scrollable-tabs) 

**What happened instead:**

It overflows the viewport in an annoying way

**Steps to reproduce:**
I reproduced it in CodeSandbox, I recommend to **open in a new window**

**Related code:**
CodeSandbox (https://codesandbox.io/s/0p6v43wnl)

* React-admin version: 2.1.4
* React version: 16.3.2
* Browser: Chrome

djhi:
Fixes #2094

As decided with the core team, we won't try to do anything special regarding horizontal scroll. We display everything and the horizontal scroll-bar of the window will be displayed normally.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -87,8 +87,6 @@ export class TabbedShowLayout extends Component {
                 {...sanitizeRestProps(rest)}
             >
                 <Tabs
-                    scrollable
-                    scrollButtons="off"
                     // The location pathname will contain the page path including the current tab path
                     // so we can use it as a way to determine the current tab
                     value={location.pathname}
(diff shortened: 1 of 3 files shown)


djhi:
Should I backport this on master ?

fzaninotto:
Bad for the list. Customers expect the horizontal scroll bar to show at the bottom of the window if there is an overflow, and in your example, it might be under the page limit.

So we need a separate solution for the edit and list views.

djhi:
@fzaninotto I updated the PR as discussed, see new comment. There's still a side effect on IE/Edge though. See the next comment

fzaninotto:
So basically we're giving up on scrollable tabs... It's a tradeoff that we must remember. And I'm pretty sure that it could work without some CSS (or JS) width calculation, we must be missing something. We'll have to revisit this problem in the future.

Issue #1782, opened 2018-04-25
Stylesheet class names conflict with production build (issue with JSS minification)


jpetitcolas:
We noticed a severe issue when passing our project in production. Our app seems to be empty:

![image](https://user-images.githubusercontent.com/688373/39239547-85f792fa-4881-11e8-97f3-efb5f177857b.png)

Seems, because everything is correctly set up in the DOM:

![image](https://user-images.githubusercontent.com/688373/39239610-b08d156c-4881-11e8-85e9-03d1f6a1879c.png)

The issue seems related to JSS, which has some troubles generating class names. Each time it see a `withStyles`, it increments a counter. Yet, it seems that MaterialUI and `react-admin` uses two distinct counters. Hence, when the classes are minified (removing the component display name), classes are renamed such as:

* MaterialUI: `jss1`, `jss2`, etc.
* React Admin: `jss1`, `jss2`, etc.

This duplication cause the issue mentioned above.

We temporarily fixed it on our project side, using a custom class name generator, using the same counter for every classes.

``` js
const escapeRegex = /([[\].#*$><+~=|^:(),"'`\s])/g;
let classCounter = 0;

// Heavily inspired of Material UI:
// @see https://github.com/mui-org/material-ui/blob/9cf73828e523451de456ba3dbf2ab15f87cf8504/src/styles/createGenerateClassName.js
// The issue with the MUI function is that is create a new index for each
// new `withStyles`, so we handle have to write our own counter
export const generateClassName = (rule, styleSheet) => {
    classCounter += 1;

    if (process.env.NODE_ENV === 'production') {
        return `c${classCounter}`;
    }

    if (styleSheet && styleSheet.options.classNamePrefix) {
        let prefix = styleSheet.options.classNamePrefix;
        // Sanitize the string as will be used to prefix the generated class name.
        prefix = prefix.replace(escapeRegex, '-');

        if (prefix.match(/^Mui/)) {
            return `${prefix}-${rule.key}`;
        }

        return `${prefix}-${rule.key}-${classCounter}`;
    }

    return `${rule.key}-${classCounter}`;
};
```
Then we wrapped our `Admin` component in:

```js
import JssProvider from 'react-jss/lib/JssProvider';

export default () => (
    <JssProvider generateClassName={generateClassName}>
        <Admin />
    </JssProvider>
);
```
[...]

PR #2228, opened 2018-08-27
[RFR] Warn About Outdated Material-UI version

Kmaschta:
Here is a small idea that I found to warn developers about the outdated Material-UI issue on production (#1782).

It can be added in addition to #2215. I've considered other alternatives like throwing an error at the build phase or read the `@material-ui/core/package.json` but it's the less intrusive method I found.

Here is the result on the console:

image

### DIFF
--- a/packages/ra-ui-materialui/src/index.js
+++ b/packages/ra-ui-materialui/src/index.js
@@ -8,3 +8,7 @@ export * from './layout';
 export * from './list';
 export Link from './Link';
 export defaultTheme from './defaultTheme';
+
+import { warnForUnsupportedMuiVersion } from './warnings';
+
+warnForUnsupportedMuiVersion();
--- a/packages/ra-ui-materialui/src/warnings.js
+++ b/packages/ra-ui-materialui/src/warnings.js
@@ -1,0 +1,17 @@
+/* eslint-disable no-console */
+import { ToggleButtons } from '@material-ui/core';
+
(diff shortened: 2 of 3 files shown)


packages/ra-ui-materialui/package.json
```diff
@@ -53,6 +53,7 @@
         "react-transition-group": "^2.2.1",
         "recompose": "~0.26.0",
         "redux": "~3.7.2",
-        "redux-form": "~7.4.0"
+        "redux-form": "~7.4.0",
+        "warning": "^4.0.2"
```
djhi: We didn't use the `warning` package for all other warnings. I think we should either drop it or use it everywhere

fzaninotto: we can drop it here

Kmaschta: It is used on ra-core.

Kmaschta:
Review applieed

fzaninotto:
In fact, I think it's the responsibility of material-ui to do that. The same way React warns the user if there are more than one version loaded, material-ui should do the same.

Kmaschta:
Nvm, I opened an issue on MUI about this. https://github.com/mui-org/material-ui/issues/12781

oliviertassinari:
> In fact, I think it's the responsibility of material-ui to do that. React warns developers if there are more than one version loaded, material-ui should do the same.

Done https://github.com/mui-org/material-ui/pull/15422

PR #2230, opened 2018-08-28
[RFR] Expose ES modules in order to enable tree shaking

### LINKED ISSUE #2204: Expose ES Modules (ESM) in order to allow Tree-Shaking
Currently, all the ESM `import` and `export` are transpiled to `require` and `module.exports`.

My suggestion is to not transpile these ESM imports/exports to their CommonJS counterparts.

# Enabling Tree-Shaking
Any application that uses `react-admin` has no choice but include its **whole** codebase in its JS bundle.

It's the case because of the extensive usage of `index.js` containing a lot of imports.
[This kind of files](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/index.js) are allowing a very nice API for React Admin, but each of these are including its children modules in the bundle.

Tree shaking to the rescue ? Sadly, the [Webpack Tree Shaking](https://webpack.js.org/guides/tree-shaking/) feature relies on these `import` and `export` to find and mark unused exports and delete them with production bundle

**React Admin is already publishing its sources in `src/`, so what's the issue?**

We indeed can import directly the ESM via thanks to:

```js
import { Admin } from 'react-admin/src'; 
```

The issue is that React Admin is split in a lot of packages, and all of them relies on the transpiled versions of each others.

For example, we can import the Admin ES module from `react-admin/src`, but this lib imports CoreAdmin from `ra-core`, which is transpiled in CommonJS.
Hence, the webpack can't shake the whole tree and let a lot of unused code in the bundle.

# Implementation Options & Consequences

## With Breaking Change (`react-admin@3` ?)
At first glance, all we need to do is change `.babelrc` a bit:

```diff
-    "presets": ["es2015", "stage-0", "react"],
+    "presets": [["es2015", { "modules": false }], "stage-0", "react"],
```

But this will introduce a major breaking change: the library users will then be forced to transpile ES modules to CommonJS on their own.

This solution would be the best in the long run, since the most common usage of React Admin is from a fresh `create-react-app`, [as explained in the tutorial](https://marmelab.com/react-admin/Tutorial.html#setting-up).

## With An Additionnal ES Submodule

Another solution is to do like [`date-fns` is doing in its next major version](https://date-fns.org/v2.0.0-alpha.9/docs/ECMAScript-Modules): expose the CommonJS modules by default and have a submodule where the library is transpiled with ES modules in `/esm`.

```js
import { Admin } from 'react-admin;' // CommonJS
import { Admin } from 'react-admin/esm'; // ESM - Tree shaking friendly
```

This solution isn't easy to implement since any package should exports an ES submodule, and each parent package should duplicate some code in order to not cutting the tree.

For example, the `react-admin` index should import `ra-core` but the `react-admin/esm` index must import from `ra-core/esm`, and so on.

## Maintaining A Parallel Package `react-admin-esm`
In a similar fashion, it is possible to fork this repository or having parallel packages.

```js
import { Admin } from 'react-admin;' // CommonJS
import { Admin } from 'react-admin-esm'; // ESM - Tree shaking friendly
```

The implementation would be easier: `react-admin` would be a simple shell for `react-admin-esm` with a bit of additional transpilation.

We'll then be able to use aliases like `'react-admin': 'react-admin-esm'` or something to find a way to a tree-shakable codebase.

This solution might have its drawbacks too.
In conclusion, I think there is no silver bullet to enable tree-shaking but it would be a great addition since this project is growing.

Kmaschta:
Fixes #2204 

## Rationale

As explained in #2204, the goal of this PR is to expose ES modules from the `react-admin` packages in order to enable the usage of Tree Shaking by webpack.

To do so, I take the following measures:

- The assets are now transpiled twice, once in `lib/` for **CommonJS** and once in `esm/` for **ES Modules** (thanks to the `BABEL_ENV` variable and the config in `.babelrc`)
- In all `package.json`, the `esm` folder is now included to the NPM package
- In all `package.json`, a new key `module` tells to consumers that ES modules are available at `esm/index.js`
- In all `package.json` again, the `sideEffects` is set to `false` telling that nothing is done at the import phase (useful to webpack for tree shaking)

## Benchmark

<details>
<summary>Tested codebase and configurations</summary>

- Default webpack configuration in `production` mode

- `index.js`

```jsx
// index.js
import React from 'react';
import ReactDOM from 'react-dom';
import { Admin, Resource, List, Datagrid, TextField } from 'react-admin';

export const testSomething = () => { console.log('not used'); };

    <List {...props}>
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <TextField source="body" />
        </Datagrid>
    </List>
);

const dataProvider = jsonServerProvider('http://jsonplaceholder.typicode.com');
const App = () => (
    </Admin>
);

ReactDOM.render(<App />, document.getElementById('root'))
```

- `.babelrc`

```json
{
    "presets": [
        ["env", { "loose": true, "modules": false }],
        "react"
    ],
    "plugins": [
        "transform-react-jsx",
        "add-module-exports",
        [
            "babel-plugin-transform-builtin-extend",
            {
                "globals": [
                    "Error"
                ]
            }
        ],
        [
            "transform-runtime",
            {
                "polyfill": false,
                "regenerator": true
            }
        ]
    ]
}

```

</details>

### React Admin 2.2.0

Total bundle size: 1.19 MB

image

### Current Branch

Total bundle size: 840 KB

image

## Improvement ideas

- Instead of transpiling twice, we can only have `index.js` files that aren't not transpiled but refers to the common JS files.

## TODO
- [x] Expose ES modules into a separate folders `/esm`
- [x] Link the ES modules with the `module` field of each `package.json`
- [x] Ensure the tree shaking works
- [x] Fix / Add some tests

### DIFF
--- a/examples/simple/webpack.config.js
+++ b/examples/simple/webpack.config.js
@@ -32,7 +32,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-core',
-                'src'
+                'lib'
             ),
             'ra-ui-materialui': path.join(
                 __dirname,
@@ -40,7 +40,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-ui-materialui',
-                'src'
+                'lib'
             ),
             'react-admin': path.join(
                 __dirname,
@@ -48,7 +48,7 @@ module.exports = {
                 '..',
                 'packages',
                 'react-admin',
-                'src'
+                'lib'
             ),
             'ra-data-fakerest': path.join(
                 __dirname,
@@ -56,7 +56,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-data-fakerest',
-                'src'
+                'lib'
             ),
             'ra-input-rich-text': path.join(
                 __dirname,
@@ -64,7 +64,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-input-rich-text',
-                'src'
+                'lib'
             ),
             'ra-tree-core': path.join(
                 __dirname,
@@ -72,7 +72,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-tree-core',
-                'src'
+                'lib'
             ),
             'ra-tree-ui-materialui': path.join(
                 __dirname,
@@ -80,7 +80,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-tree-ui-materialui',
-                'src'
+                'lib'
             ),
             'ra-tree-language-english': path.join(
                 __dirname,
--- a/examples/simple/webpack.config.js
+++ b/examples/simple/webpack.config.js
@@ -32,7 +32,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-core',
-                'lib'
+                'src'
             ),
             'ra-ui-materialui': path.join(
                 __dirname,
@@ -40,7 +40,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-ui-materialui',
-                'lib'
+                'src'
             ),
             'react-admin': path.join(
                 __dirname,
@@ -48,7 +48,7 @@ module.exports = {
                 '..',
                 'packages',
                 'react-admin',
-                'lib'
+                'src'
             ),
             'ra-data-fakerest': path.join(
                 __dirname,
@@ -56,7 +56,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-data-fakerest',
-                'lib'
+                'src'
             ),
             'ra-input-rich-text': path.join(
                 __dirname,
@@ -64,7 +64,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-input-rich-text',
-                'lib'
+                'src'
             ),
             'ra-tree-core': path.join(
                 __dirname,
@@ -72,7 +72,7 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-tree-core',
-                'lib'
+                'src'
             ),
             'ra-tree-ui-materialui': path.join(
                 __dirname,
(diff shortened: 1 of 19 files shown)


examples/simple/webpack.config.js
```diff
@@ -32,55 +32,55 @@ module.exports = {
                 '..',
                 'packages',
                 'ra-core',
-                'src'
+                'lib'
```
djhi: Are you sure about this ? It looks like it's going to take the `cjs` files, thus not building them itself when changed

Kmaschta: Not sure this is the best solution, but it works on my machine®!
I considered your solution but is it the role of the simple example to build the RA sources? Idk

On the CI, the build is done before the test, hence it avoid to double-build these sources.

djhi: Not its role but invaluable when working on one of those packages

Kmaschta: Then, just to be sure, you suggest that I should add all the babel plugins related to the all the packages in `examples/simple/.babelrc`?
Or I would just use the same and find somewhere to add `BABEL_ENV=cjs`.

djhi: I don't what would work best tbh. IMO, adding the `BABEL_ENV=cjs` looks easier

djhi: Don't forget our windows users though ;)

fzaninotto: Don't change the simple example webpack config. We use a direct link to src because it lets us develop on the packages and see the impact in the example *without rebuilding*. Your change would break that.

Kmaschta:
So, in order to not change the webpack config of the simple example, I was forced to add `BABEL_ENV=cjs` in the Makefile.
Because the simple example use the root `.babelrc` and totally ignores the babel config at `examples/config/.babelrc`.

Since it's not the goal of this PR to fix that, I opened an issue for that #2280

fzaninotto:
Awesome!

PR #2246, opened 2018-08-30
[RFR] fix ReferenceArrayInputController error on undefined record

natrim:
ReferenceArrayInputController will make error in componentWillReceiveProps if record does not exist's

happens sometimes on initial render

fixed same way as in ReferenceInputController


fzaninotto:
bad fix. `{} === {}` returns false in JS. Please don't use `||`, but check the existence before testing the equality.

fzaninotto:
My bad, the fix is correct, I read the code too fast. Sorry for the noise.

PR #2249, opened 2018-08-30
[RFR] Finally fix form reset with navigation

### LINKED ISSUE #2245: POST request params in create page is getting params from last edit page
**What you were expecting:**
For example, there are 2 tables named, "User" and "Address".
First I go to "User" edit page and do editing then save, or even do nothing in edit page.
Then I go to "Address" create page, when I save this new record, the POST request is expected to be using "Address" fields only.

**What happened instead:**
But the POST request is now using "Address" fields plus "User" fields, which makes the request failed to call.

**Steps to reproduce:**
For example like
Go to XXXXX/user/1/edit
Then go to XXXXX/address/create, then try to create a new record

**Environment**
* React-admin version: ^2.1.2
* React version: ^16.4.1
* Browser: chrome

djhi:
Fixes #2245

I removed the formMiddleware I introduced to allow inline editing with the upcoming Tree component. It's most probably isn't needed anymore but I'll fix it on the next branch at the Tree level if that's the case


fzaninotto:
We need a unit / e2e test to make sure the bug you fix doesn't reappear in the future.

djhi:
I tweaked the tests a little to cover the following cases:

- Navigating between two pages of the same resource (edit to create for example) should not kept the redux form values nor the record values

- Navigating between two pages of the different resource (comments/edit to posts/create for example) should not kept the redux form values nor the record values

PR #2252, opened 2018-08-30
[RFR] Fix wrong warning on FormDataConsumer

djhi:
It should only be displayed when the `FormDataConsumer` is used inside an `ArrayInput`.

Also do not pass `ArrayInput` related props if not needed


fzaninotto:
Can you add a unit test? If it's a fix, why not push it to master?

djhi:
Not on `master` because this feature is on `next`

PR #2247, opened 2018-08-30
[RFR] RichTextInput - allow fullWidth

natrim:
pass fullWidth in RichTextInput to FormControl be able to make the wysiwyg full width

PR #2253, opened 2018-08-30
[RFR] RichTextInput - allow fullWidth

natrim:
pass fullWidth in RichTextInput to FormControl be able to make the wysiwyg full width

port of #2247 from `master` to `next`


fzaninotto:
thanks!

PR #2255, opened 2018-08-31
Added bs-react-admin project to Ecosystem documentation.

ctbucha:


djhi:
Thanks! Would you mind explaining this comment in the README?

> Need to pass this to Resource as ReactJS instead of ReasonReact since react-admin uses it in a non-standard way

Isn't standard to accept functions as prop?

fzaninotto:
Thanks!

djhi:
Thank you for explaining this, I don't have any experience with Reason, yet :) Is it an issue with `cloneElement` too?

Issue #2243, opened 2018-08-29
GraphQL adaptators improvements


Weakky:
Hey there,

As of now, most GraphQL adaptators aren't taking full advantage of one of GraphQL's biggest benefit: **being able to query nested fields in one request**, as deep as the schema allow us to.
Instead, they're more or less imitating the way REST works, by making X request for X reference needed (although some great optimizations were already done in that regard https://marmelab.com/blog/2016/10/18/using-redux-saga-to-deduplicate-and-group-actions.html)

I think there are ways we could improve that, and I suggest we use this place as a central thread to discuss about what could be done.

## Suggestion n°1

**Include all scalar fields of linked types by default**

As of now, we're only fetching **ids** from linked types, which forces us to use `<ReferenceField />` on every linked type.

```js
const linkedResource = introspectionResults.resources.find(
  r => r.type.name === type.name
);

if (linkedResource) {
  return { ...acc, [field.name]: { fields: { id: {} } } };
}
```

One quick-fix that could be done, is to automatically fetch all scalar fields of linked types:

```js
const linkedResource = introspectionResults.resources.find(r => r.type.name === type.name);

if (linkedResource) {
  // Fetch all linked resource scalar fields
  const linkedScalarFields = linkedResource.type.fields
    .filter(field => getFinalType(field.type).kind !== TypeKind.OBJECT)
    .reduce((acc, field) => ({ ...acc, [field.name]: {} }), {});

  return { ...acc, [field.name]: { fields: linkedScalarFields } };
}
```

This way, we're able to use `<TextField source="linkedType.scalarField" />`, which already cover lots of cases and greatly improve the amount of request made.
I think the few more bytes that those additional fields will take on every requests are more than worth the amount of request that it will save.

## Suggestion n°2

**Make this overridable**

After thinking about it, @fzaninotto, I don't think there's a need for a function to transform the `introspectionResults`. The introspection query already fetches all the existing types (and their fields) anyway.

If I'm not mistaking, overriding the behavior described above is actually already possible and that's what you've done [here](https://github.com/marmelab/react-admin/blob/master/examples/graphcool-demo/src/dataProvider.js#L42-L47), by overriding the query that fetches the `Command`.

I think we could provide a simpler and more concise API to do that, for two reasons:

- The query name and the params are not needed for what we're trying to achieve. The params will always have to be declared anyway, and it can become heavy when overriding `GET_LIST` actions (having filters, sorting and pagination as input).

[...]

PR #2257, opened 2018-08-31
[RFR] GraphQL providers enhancements

djhi:
Following #2243

- [x] Deprecate the `override` option
- [x] Exports the `buildQuery` function in dialect packages
- [x] Document how to wrap `buildQuery` to override specific queries

### DIFF
--- a/packages/ra-data-graphcool/README.md
+++ b/packages/ra-data-graphcool/README.md
@@ -87,6 +87,41 @@ Or supply your client directly with:
 buildGraphcoolProvider({ client: myClient });
 ```
 
+### Overriding a specific query
+
+The default behavior might not be optimized especially when dealing with references. You can override a specific query by wrapping the `buildQuery` function:
+
+```js
+// in src/dataProvider.js
+import buildGraphcoolProvider, { buildQuery } from 'ra-data-graphcool';
+
+const overrideBuildQuery = buildQuery => (fetchType, resource, params) => {
+    const builtQuery = buildQuery(fetchType, resource, params);
+
+    if (resource === 'Command' && fetchType === 'GET_ONE') {
+        return {
+            // Use the default query variables and parseResponse
+            ...buildQuery,
+            // Override the query
+            query: gql`
+                query Command($id: ID!) {
+                    data: Command(id: $id) {
+                        id
+                        reference
+                        customer {
+                            id
+                            firstName
+                            lastName
+                        }
+                    }
+                }`,
+        };
+    }
+
+    return builtQuery;
+}
+```
+
 ### Customize the introspection
 
 These are the default options for introspection:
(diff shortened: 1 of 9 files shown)


packages/ra-data-graphcool/README.md
```diff
@@ -87,6 +87,41 @@ Or supply your client directly with:
...
 
+### Overriding a specific query
+
+The default behavior might not be optimized especially when dealing with references. You can override a specific query by wrapping the `buildQuery` function:
+
+```js
+// in src/dataProvider.js
+import buildGraphcoolProvider, { buildQuery } from 'ra-data-graphcool';
+
+const overrideBuildQuery = buildQuery => (fetchType, resource, params) => {
```
Weakky: I think the signature is wrong here: `(buildQuery) => (introspectionResults) => (fetchType, resource, params)`.

Same goes for `const builtQuery = buildQuery(introspectionResults)(fetchType, resource, params);`

packages/ra-data-graphcool/README.md
```diff
@@ -87,6 +87,41 @@ Or supply your client directly with:
...
+    const builtQuery = buildQuery(fetchType, resource, params);
+
+    if (resource === 'Command' && fetchType === 'GET_ONE') {
+        return {
+            // Use the default query variables and parseResponse
+            ...buildQuery,
```
Weakky: Shouldn't it be `builtQuery` instead?

djhi: Good catch!

packages/ra-data-graphcool/README.md
djhi: Right !

djhi: It seems cumbersome to use it that way, isn't it? Lot of boilerplate again...

What do you think of the following usage:

```js
// in src/dataProvider.js
import buildGraphcoolProvider from 'ra-data-graphcool';

const buildQuery = defaultBuildQuery => (fetchType, resource, params) => {
    const builtQuery = defaultBuildQuery(fetchType, resource, params);

    if (resource === 'Command' && fetchType === 'GET_ONE') {
        return {
            // Use the default query variables and parseResponse
            ...builtQuery,
            // Override the query
            query: gql`
                query Command($id: ID!) {
                    data: Command(id: $id) {
                        id
                        reference
                        customer {
                            id
                            firstName
                            lastName
                        }
                    }
                }`,
        };
    }

    return builtQuery;
}

export default buildGraphcoolProvider({ buildQuery })
```

djhi: Actually a bad idea as one may need the introspection results for advanced scenarios

packages/ra-data-graphql/src/index.js
```diff
@@ -58,6 +58,12 @@ export default async options => {
         ...otherOptions
     } = merge({}, defaultOptions, options);
 
+    if (override && process.env.NODE_ENV === 'production') {
```
fzaninotto: I think we still use the `override` option in the `dataProvider` of the `examples/graphcool-demo`

PR #2262, opened 2018-08-31
[RFR] ·Add documentation on the data returned from create requests 

### LINKED ISSUE #2087: CREATE type data request returns params instead of response
**What you were expecting:**
The dataProvider CREATE type data request to return created data from the server.

**What happened instead:**
It returns the params you pass to it instead. If your backend adds data to the record, you are unable to access that data unless you fetch the record again.

I'm not sure if this is intended.

**Related code:**
https://github.com/marmelab/react-admin/blob/master/packages/ra-data-json-server/src/index.js#L123
https://github.com/marmelab/react-admin/blob/master/packages/ra-data-simple-rest/src/index.js#L127

cuococarlos:
closes #2087

### DIFF
--- a/packages/ra-data-json-server/README.md
+++ b/packages/ra-data-json-server/README.md
@@ -99,6 +99,13 @@ const httpClient = (url, options = {}) => {
 
 Now all the requests to the REST API will contain the `Authorization: SRTRDFVESGNJYTUKTYTHRG` header.
 
+**Note**: In case of REST verb "CREATE" consider that the response body is the same as the request body but with the object ID injected .
+```
+case CREATE:
+return { data: { ...params.data, id: json.id } };
+```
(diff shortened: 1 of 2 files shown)


djhi:
Thank you for contributing! However, as @fzaninotto said in the linked issue (#2087), we don't want to make this change now as it will be a breaking one and it is easy to implement your own data provider.

I apologize as my last comment on this issue missed the word `documentation`. Indeed, we only want to make those data providers behavior regarding `CREATE` clearer. I updated this comment accordingly

cuococarlos:
Ok.Tomorrow, i will write documentation in another pr and close this.

cuococarlos:
I added the documentation on this topic. but I do not know if it will be necessary to add it in another document too

fzaninotto:
Thanks!

PR #2272, opened 2018-09-03
[RFR] AutocompleteArrayInput

### LINKED ISSUE #2138: Multi-select Autocomplete
The [example](https://github.com/marmelab/react-admin/blob/0e86299a3ad7fa685a457d8b7067e93ee98bf8fc/packages/ra-ui-materialui/src/input/ReferenceArrayInput.js#L196) does not provide filtering as it is decribed in the example text. Though being the filterToQuery prop is required, it should allow for filtering automatically.

The filterToQuery function never fires being there is not area to input text.

```
field.input = props => (
    <ReferenceArrayInput
        source={field.name}
        reference={field.reference.name}
        label="Destination"
        key={field.name}
        filter={{isLeaf: true}}
        filterToQuery={searchText => {console.log('Filter the Destination: ' + searchText); return ({ displayname: searchText })}}
        sort={{field: 'displayname', order: 'ASC'}}
        {...props}>
        <SelectArrayInput optionText="displayname" optionValue="id"  />
    </ReferenceArrayInput>
);
```

This should bring up the items that are already selected, and other possible items, then as you type, filter the items so they can be selected and added to the multiselect.

If this is not the way to do this, or if more custom code should be inserted, please describe, or direct to a tutorial instead of a faulty example.

**Environment**

* React-admin version: 2.2.0
* Last version that did not exhibit the issue (if applicable): false
* React version: 16.4.1
* Browser:Chrome 68.0.3440.84

djhi:
Screencast

- [x] Implementation
- [x] Polishing & Code cleaning
- [x] Tests
- [x] Documentation

Fixes #2138

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -174,7 +174,7 @@ By default the component matches choices with the current input searchText: if i
 
 If you want to limit the initial choices shown to the current value only, you can set the `limitChoicesToValue` prop.  
 
-Lastly, `<AutocompleteInput>` renders a meterial-ui `<TextField>` component. Use the `options` attribute to override any of the `<TextField>` attributes:
+Lastly, `<AutocompleteInput>` renders a material-ui `<TextField>` component. Use the `options` attribute to override any of the `<TextField>` attributes:
 
 {% raw %}
 ```jsx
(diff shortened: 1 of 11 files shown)


djhi:
IMO we should not auto-select items when we get a match from autocompletion on this component. It makes sense for the `AutocompleteInput` but it's disturbing here.

fzaninotto:
Awesome!

PR #2279, opened 2018-09-05
[RFR] Fix SaveButton misaligned CircularProgress

natrim:
Fix SaveButton misaligned CircularProgress on saving (because of the icon's paddingRight)

before:
before

after fix:
after


fzaninotto:
Thanks!

PR #2283, opened 2018-09-06
Added Hungarian translation

phelion:
Hi, I created the hungarian translation for react-admin, and published it to npm. Now I added the path of the git repo to the Translation doc


phelion:
Thank you too!:)

PR #2284, opened 2018-09-06
[WIP] Avoid to build twice in order to expose ESM

Kmaschta:
Exposing ESM is good for library consumers and performance.
But building twice the entire libary and multiplying the codebase in the node module is not the best solution to do so.

The goal of this PR is to manually importing the right modules in order to have only imports and exports in the `esm` folder for the same result.

## TODO
- [x] Remove the dedicated process for building ESM packages
- [ ] For each package, fill the `esm` folder with proper imports
- [ ] Find a way to test these ESM folder

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,56 +1,39 @@
 {
-    "env": {
-        "cjs": {
-            "presets": [
-                ["env", {
-                    "targets": {
-                        "browsers": ["last 2 versions"]
-                    }
-                }],
-                "react"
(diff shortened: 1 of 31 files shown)


Kmaschta:
All things considered, it's not a good solution ... The tree will not be as deep as a full transpilation, and it will force the developers to maintain two versions of the API.

If someone have a better idea, feel free to open an issue!

PR #2286, opened 2018-09-07
[RFR] Make UserMenu override more idiomatic

fzaninotto:
The previous way of adding items to the user menu didn't allow to replace the user menu (e.g. to customize the user icon). Now it's possible.

Developers can replace the default user menu by their own by setting the `userMenu` prop of the `<AppBar>` component. For instance, to add custom menu items, just decorate the default `<UserMenu>` by adding children to it:

```js
import { AppBar, UserMenu, MenuItemLink } from 'react-admin';
import SettingsIcon from '@material-ui/icons/Settings';

const MyUserMenu = props => (
    <UserMenu {...props}>
        <MenuItemLink
            to="/configuration"
            primaryText="Configuration"
            leftIcon={<SettingsIcon />}
        />
    </UserMenu>
);

const MyAppBar = props => <AppBar {...props} userMenu={<MyUserMenu />} />;

```

It's a bit more indirect, but a lot more powerful.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/UserMenu.js
+++ b/packages/ra-ui-materialui/src/layout/UserMenu.js
@@ -37,6 +37,7 @@ class UserMenu extends React.Component {
 
     render() {
         const { children, label, logout, translate } = this.props;
+        if (!logout && !children) return null;
         const { anchorEl } = this.state;
         const open = Boolean(anchorEl);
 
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/layout/UserMenu.js
```diff
@@ -37,6 +37,7 @@ class UserMenu extends React.Component {
 
     render() {
         const { children, label, logout, translate } = this.props;
+        if (!logout && !children) return null;
```
djhi: I  wonder if we should allow to not specify the `logout`

PR #2294, opened 2018-09-09
fixes rowsPerPageOptions

antonversal:
Somehow `rowsPerPageOptions` was lost and does not work anymore. 
The PR fixes the issue.


fzaninotto:
Thanks!

PR #2318, opened 2018-09-13
[RFR] Add warnings to translate HOC when used directly to translate a string

djhi:
screenshot

### DIFF
--- a/packages/ra-core/src/i18n/translate.js
+++ b/packages/ra-core/src/i18n/translate.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import wrapDisplayName from 'recompose/wrapDisplayName';
+import warning from '../util/warning';
 
 /**
  * Higher-Order Component for getting access to the `translate` function in props.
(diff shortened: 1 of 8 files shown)


packages/ra-core/src/i18n/translate.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import wrapDisplayName from 'recompose/wrapDisplayName';
+import warning from 'warning';
```
fzaninotto: this package is internal to facebook (see https://github.com/facebook/fbjs), use `console.warn()` instead, and please remove usage from the code.

packages/ra-core/src/i18n/translate.js
```diff
@@ -21,6 +22,15 @@ import wrapDisplayName from 'recompose/wrapDisplayName';
  * @param {*} BaseComponent The component to decorate
  */
 const translate = BaseComponent => {
+    warning(
+        typeof BaseComponent !== 'string',
```
fzaninotto: how about function components?

packages/ra-core/src/i18n/translate.js
djhi: > This is a mirror of Facebook's (FB) warning module used within React's source code (and other FB software). As such this module will mirror their code as much as possible.

packages/ra-core/src/i18n/translate.js
djhi: they are not string right ?

djhi: warning displays its warning if the condition is **not** met. Here, if `BaseComponent` **is** a string, then we'll have a warning

packages/ra-core/src/i18n/translate.js
djhi: I remembered that I introduced this package when working on permissions. I wonder whether we should use it everywhere now. It only display warnings when not in production so that one check we don't have to verify ourselves ;)

fzaninotto:
One last demand: please open against `next` rather than `master`. You're changing dependencies, and this is, technically a new feature.

PR #2322, opened 2018-09-14
[RFR] Fix resetform when navigating to same page

### LINKED ISSUE #2291: Double-click on a Icon from the Menu reset the edition form
If we have a `MenuItemLink` that points to an edition view, if you click on this menu item **icon** (not the menu label!) twice, it resets the edition form.

**Steps to reproduce:**
1. Have a custom menu with a menu item which points to an edition view

```jsx
const Menu = ({ onMenuTap, logout }) => {
  return (
    <div>
      <MenuItemLink
        to="/posts/13"
        primaryText="Post 13"
        exact
        leftIcon={createElement(PostIcon)}
        onClick={onMenuTap}
      />
    </div>
  );
};
```

2. Plug this menu to this admin

```jsx
render(
  <Admin menu={Menu}>
        {/* resources */}
  </Admin>,
)
```

3. Click on this menu item to go to `/posts/13`

4. When the data is loaded, click again but on the menu **icon**, the form is resetted. But not by clicking on the text.

**Related code:**
There is an example on this bug on the following sandbox: https://codesandbox.io/s/24ov7qyx1y

**Environment**

* React-admin version: At least 2.2.0

[Related question on Stack Overflow](https://stackoverflow.com/questions/52184610/react-admin-clicking-menuitemlink-for-2nd-time-clears-form-inputs)

djhi:
Fixes #2291

### DIFF
--- a/packages/ra-core/src/sideEffect/recordForm.js
+++ b/packages/ra-core/src/sideEffect/recordForm.js
@@ -1,13 +1,23 @@
 import { put, takeEvery } from 'redux-saga/effects';
 import { LOCATION_CHANGE } from 'react-router-redux';
 import { destroy } from 'redux-form';
+import isEqual from 'lodash/isEqual';
+
 import { resetForm } from '../actions/formActions';
 import { REDUX_FORM_NAME } from '../form/constants';
 
-export function* handleLocationChange({ payload: { state } }) {
-    if (state && state.skipFormReset) {
+let previousLocation;
+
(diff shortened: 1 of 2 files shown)


fzaninotto:
I'm not sure about this one. We reset the form on purpose when the redirect is "edit", for the "save and add" feature.

djhi:
We have end to end tests for the save and add feature and they still pass ;)

PR #2332, opened 2018-09-18
[RFR] Add <DateTimeInput>

fzaninotto:
~~Due to the way `<input type="datetime-local" /> works on Chrome, I had to store a local draft state and couldn't resort to redux-form's `parse` and `format` (see commit list).~~~

I managed to make it a stateless component thanks to redux-form's `parse` and `format`.

### DIFF
--- a/packages/ra-core/src/form/addField.js
+++ b/packages/ra-core/src/form/addField.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import FormField from './FormField';
 
-export default BaseComponent => {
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/form/addField.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import FormField from './FormField';
 
-export default BaseComponent => {
+export default (BaseComponent, fieldProps = {}) => {
```
djhi: This should be documented though

PR #2335, opened 2018-09-19
[RFR] Add error prop to FormHelperText, to mark it as error on error meta

natrim:
Add's error prop to FormHelperText, to mark it as error on error meta

Makes the error text red (well, the error color defined by theme) like other errors in TextInput's

changes RichTextInput, RadioButtonGroupInput, CheckboxGroupInput and SelectArrayInput

note: added as pull to master instead of next as i consider it a bug that these components have inconsistent way to display error with other input's

**edit: maybe i should add some borderBottom to make it even more consistent?
edit2: added images**

before:
old

old

after (in this PR):
new

new

with borderBottom (not in this PR, shpuld i add it?):
withborder

withborder

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -71,7 +71,7 @@ export class RichTextInput extends Component {
                 className="ra-rich-text-input"
             >
                 <div ref={this.updateDivRef} />
-                {error && <FormHelperText>{error}</FormHelperText>}
+                {error && <FormHelperText error>{error}</FormHelperText>}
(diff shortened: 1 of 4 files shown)


fzaninotto:
Nice! I don't think the border bottom is necessary in the Boolean input.

But this is a new feature, so as per our release policy, it must be PRed against the `next` branch rather than the `master` branch. Can you please rebase and repush?

natrim:
@fzaninotto rebased on next

fzaninotto:
Sorry, you're right, I meant RadioButtonGroupInput.

fzaninotto:
Thanks!

PR #2339, opened 2018-09-19
[RFR] Fix create to edit form values 

### LINKED ISSUE #2324: 2.3.0 - CreateForm with single TextInput doesn't show any value after form submission
**What you were expecting:**
CreateForm with single TextInput should show stored valued after form submssion

**What happened instead:**
CreateForm with single TextInput doesn't show any value after form submission

**Steps to reproduce:**
Post -> Create -> add title -> Save

**Related code:**
* https://codesandbox.io/s/6w0v218loz

* React-admin version: 2.3.0 / 2.3.1

djhi:
- [x] Added an e2e test
- [x] Added a form middleware intercepting location changes
- [x] Unit test the middleware

Fixes #2324

### DIFF
--- a/docs/CustomApp.md
+++ b/docs/CustomApp.md
@@ -66,8 +66,8 @@ export default ({
         compose(
             applyMiddleware(
                 sagaMiddleware,
-                routerMiddleware(history),
                 formMiddleware,
+                routerMiddleware(history),
                 // add your own middlewares here
             ),
             typeof window !== 'undefined' && window.devToolsExtension


packages/ra-core/src/form/formMiddleware.js
```diff
@@ -0,0 +1,16 @@
+import { LOCATION_CHANGE } from 'react-router-redux';
+import { beforeLocationChange } from '../actions/formActions';
+
+const formMiddleware = () => next => action => {
```
fzaninotto: I and other developers will be unable to understand why you're doing that. This file requires many, many comments to describe the issue and the solution.

djhi:
Oh no, now that it is RFR and green, that's the moment where you say "Can you target next?", right @fzaninotto? :sob:

djhi:
> I think you must update the CustomApp doc, too.

Actually the `CustomApp` doc was outdated and still referencing the `formMiddleware`

Issue #2319, opened 2018-09-13
Delete button cannot be hidden/conditionally displayed in Edit toolbar 


kctrlv:
**What you were expecting:**
Being able to override the DeleteButton showing in the toolbar based on permissions. 

**What happened instead:**
The DeleteButton is hardcoded into the toolbar after the span, there is no way to remove or conditionally display it.

**Steps to reproduce:**
Spin up any edit page that has a resource ID and note that the DeleteButton is on the lower right. Attempting to make your own custom toolbar with conditional logic to display the buttons will not override the hardcoded DeleteButton that comes with it.

**Related code:**
Toolbar.js Lines 94-100

kctrlv:
Potential fix: move the lines that deal with the DeleteButton logic (94-101) to the same section above that conditionally displays SaveButton (Children.count(children) === 0). This way it will still show on the default toolbar but will let a user override it with their custom buttons.  

P.S. Thank you for all your hard efforts on this project.

fzaninotto:
I don't understand why you're saying that you can't create a custom Toolbar without DeleteButton. I don't understand your potential fix either.

kctrlv:
@fzaninotto #2340 fixes the issue, thank you.

PR #2340, opened 2018-09-19
[RFR] [BC break] Easier custom toolbar

fzaninotto:
## Problem

Removing the delete button from the Edit toolbar used to require to create a Toolbar component from scratch. 

Refs #2319

## Solution

With this PR, it's as easy as passing just the `<SaveButton>` as only child of a custom `<Toolbar>`:

```jsx
const PostEditToolbar = props => (
    <Toolbar {...props} >
        <SaveButton />
    </Toolbar>
);

    <Edit {...props}>
        <SimpleForm toolbar={<PostEditToolbar />}>
            ...
        </SimpleForm>
    </Edit>
```

There is a slight BC break here: people with custom Toolbar for Edit views used to have the Delete button displayed even though it wasn't in their custom toolbar. With that change, the Edit button will disappear. I think, with proper warning in the release notes, we can assume this BC break (on a feature that is very young and probably not very widely used to date).

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -612,7 +612,7 @@ The most common use case is to display two submit buttons in the `<Create>` view
 For that use case, use the `<SaveButton>` component with a custom `redirect` prop:
 
 ```jsx
-import { Edit, SimpleForm, SaveButton, Toolbar } from 'react-admin';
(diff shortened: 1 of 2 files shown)


palyvodaBoi:
bad solution (as usual)
Save button loses its 'disabled' functionality with custom toolbar

I can believe it's hard to add 'showDeleteButton' property

PR #2351, opened 2018-09-24
[RFR] Automatically add id attribute on input if not specified and accessibility fixes

djhi:
- [x] Infer `id` from `source` prop and set the `id` and `htmlFor` props automatically
- [x] Adds `aria-label` to buttons

This allows material-ui to bind labels to their inputs using the `for` attribute, simplifying e2e tests with libraries such as `react-test-utils` and providing better accessibility.

For example, using [`cypress-testing-library`](https://github.com/kentcdodds/cypress-testing-library) in `cypress`, with an input written as:

```jsx
<TextInput source="first_name" />
```

We can now write tests like:

```js
cy.getByLabelText('First Name').type('Gildas');
```

Before, for the same result, we would have written the input as:

```jsx
<TextInput id="first_name" source="first_name" />
```

*NOTE*: it is still possible to provide override the default `id` inferred from the `source` prop

### DIFF
--- a/packages/ra-ui-materialui/src/button/Button.js
+++ b/packages/ra-ui-materialui/src/button/Button.js
@@ -61,6 +61,7 @@ const Button = ({
                 className={classnames(classes.button, className)}
                 color={color}
                 size={size}
+                aria-label={label && translate(label, { _: label })}
                 {...rest}
             >
                 {alignIcon === 'left' &&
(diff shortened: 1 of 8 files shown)


fzaninotto:
Awesome! Although, strictly speaking, this is an evolution, so it should land in `next` rather than `master`. Can you rebase to the `next` branch?

PR #2358, opened 2018-09-26
[RFR] Docs: renaming aor-feathers-client to ra-data-feathers

josx:
Signed-off-by: José Luis Di Biase <josx@interorganic.com.ar>


josx:
nice to hear someone whot is using it ;)

PR #2359, opened 2018-09-26
[RFR] Docs: Killing whitespaces

josx:
Signed-off-by: José Luis Di Biase <josx@interorganic.com.ar>


djhi:
Why should we do that?

Kmaschta:
I don't know, but I like this PR!

PR #2360, opened 2018-09-26
Adds Spanish language package

alfonsodev:
Hola,
I've made the translation for Spanish messages, I'm not sure if you would like to have the package merged in the main repo, or otherwise you prefer it to be in a separate repository.  

If you know Spanish maybe you can review and give your opinion, perhaps people from south America would use different words, I'm open to find synonyms  that work best for all of us.   
I found some expressions are hard to translate for example . 
```
            clear_input_value: 'Despejar valor',
```
I'm pretty sure there is better way to express that, perhaps "limpiar valor" ?, "restablecer valor" ?
and I'm ignoring "input" there because it sounds weird "limpiar valor de entrada" but you might think otherwise.   
Let me know your opinion.  
Thanks.

### DIFF
--- a/packages/ra-language-spanish/README.md
+++ b/packages/ra-language-spanish/README.md
@@ -1,0 +1,29 @@
+# Spanish Messages for React-Admin
+
+Spanish messages for [react-admin](https://github.com/marmelab/react-admin), the frontend framework for building admin applications on top of REST/GraphQL services.
+
+![react-admin demo](http://static.marmelab.com/react-admin.gif)
+
+## Installation
+
+```sh
+npm install --save ra-language-spanish
+```
+
+## Usage
+
+```js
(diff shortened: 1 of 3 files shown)


djhi:
Thanks! We won't be able to maintain this translation. Can you publish it as a stand-alone package ?

alfonsodev:
Thank for the quick reply, I'll publish it separately.

djhi:
Btw, are you aware of the existing spanish translation package https://github.com/BlackBoxVision/ra-language-spanish?

It is already on our official list https://marmelab.com/react-admin/Translation.html#available-locales

alfonsodev:
yes, I saw it after translating it, I've checked now, but it is not updated, I'll try to make a pull request there to update it or in case there is no answer I'll point here to mine too.

PR #2369, opened 2018-10-01
feat(form-tab): add className to tab content

despatates:
As it's already done on Tab, apply className to both FormTab tab and content elements.

### DIFF
--- a/packages/ra-ui-materialui/src/form/FormTab.js
+++ b/packages/ra-ui-materialui/src/form/FormTab.js
@@ -29,8 +29,8 @@ class FormTab extends Component {
         );
     };
 
-    renderContent = ({ children, hidden, ...rest }) => (
-        <span style={hidden ? hiddenStyle : null}>
+    renderContent = ({ children, className, hidden, ...rest }) => (
+        <span className={className} style={hidden ? hiddenStyle : null}>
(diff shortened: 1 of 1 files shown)


despatates:
My use case: I want to apply a CSS class to the content of one tab of a `TabbedForm`, and not to the others.

Currently, I can do it with `TabbedShowLayout`/`Tab`, the same `className` is rendered for the header and the content.

```jsx
<TabbedShowLayout>
  <Tab className="my-tab-class">
    ...
  </Tab>
</TabbedShowLayout>
```

The CSS selector are `a.my-tab-class` for the tab header and `span.my-tab-class` for its content.

I expected to have the same behavior with tabbed forms.

I'm interested to know your point of view about it.
Do you think it's a wrong behavior of `Tab`?

fzaninotto:
Indeed. Adding the same class to two very different components doesn't make sense. So `TabbedShowLayout` does it wrong. For your use case, you probably need a custom form layout. 

I'm closing this PR to avoid duplicating the same mistake twice. Thanks for your suggestion anyway.

despatates:
Thanks, that's what I did. 
Let me know if you remove the `className` from `Tab`, I'm using it.

PR #2376, opened 2018-10-01
[RFR] Component Inference

fzaninotto:
To improve the developer experience, react-admin can display content before developers choose which fields or input should be displayed inside a `List` or an `Edit`, based on inference on content.

This inference is then serialized to the console, to be copied/pasted directly into the code.

- [x] Add inference engine
- [x] Add support for `Datagrid` inference in `<ListGuesser>`
- [ ] ~~Add support for `Filter` inference in `<ListGuesser>`~~ (v2)
- [x] Add support for `SimpleForm` inference in `<EditGuesser>`
- [x] Add support for `SimpleShowLayout` inference in `<ShowGuesser>` 
- [x] Output inferred components as a string in the console
- [x] Make sure inferences work at least on the Demo example
- [x] Make sure inference does not force loading of all fields and inputs when cherry picking is enabled
- [x] Document inference in `List`, `Show`, and `Edit` chapters
- [x] Rewrite tutorial to show inference

Usage:

```js
import { Admin, Resource, ListGuesser, EditGuesser } from 'react-admin';

const App = () => (
    <Admin dataProvider={...}/
        <Resource name="posts" list={ListGuesser} edit={EditGuesser} />
        <Resource name="comments" list={ListGuesser} edit={EditGuesser} />
    </Admin>
);
```


fzaninotto:
Finally, after squashing most of the commits of this PR (to avoid drowning future blamers with my hesitations), it's Ready For Review!

packages/ra-core/package.json
```diff
@@ -42,6 +42,7 @@
     "dependencies": {
         "babel-runtime": "~6.26.0",
         "classnames": "~2.2.5",
+        "date-fns": "^1.29.0",
```
Kmaschta: By adding this version of `date-fns`, you'll import the **whole** library to the react-admin's bundle.

The pre-release `date-fns@2.0.0-alpha.22` supports tree shaking and expose ES modules. And it's stable enough to be imported in React Admin IMHO.

See comparison between:
- [date-fns@1.29.0](https://bundlephobia.com/result?p=date-fns@1.29.0)
- [date-fns@2.0.0-alpha.22](https://bundlephobia.com/result?p=date-fns@2.0.0-alpha.22)

packages/ra-core/src/inference/assertions.js
```diff
@@ -0,0 +1,31 @@
+import parseDate from 'date-fns/parse';
+
+export const isNumeric = value => !isNaN(parseFloat(value)) && isFinite(value);
+export const valuesAreNumeric = values => values.every(isNumeric);
+
+export const isInteger = value => Number.isInteger(value);
+export const valuesAreInteger = values => values.every(isInteger);
+
+export const isBoolean = value => typeof value === 'boolean';
```
Kmaschta: What abount nullable booleans? :p

fzaninotto: the inference currently falls back to string for all nullable values. That's a good idea of evolution for v2!

packages/ra-core/package.json
fzaninotto: I'm not comfortable adding an alpha dependency at this point, especially since the benefit is only in development...

djhi: ping @Kmaschta

Kmaschta: It's ok, if it's only used on dev!

PR #2381, opened 2018-10-02
[RFR] Feat/show button nested data

Oupsla:
Hello folks !

A little pull request to enhance the ShowButton and be able to query nested data to create the link.

Can be reproduced to other buttons 😄 

Tell me if I miss something since it is my first PR on this project (btw incredible project)

### DIFF
--- a/packages/ra-ui-materialui/src/button/ShowButton.js
+++ b/packages/ra-ui-materialui/src/button/ShowButton.js
@@ -4,6 +4,7 @@ import shouldUpdate from 'recompose/shouldUpdate';
 import ImageEye from '@material-ui/icons/RemoveRedEye';
 import { Link } from 'react-router-dom';
 import { linkToRecord } from 'ra-core';
+import { get } from 'lodash';
 
 import Button from './Button';
 
@@ -11,11 +12,12 @@ const ShowButton = ({
     basePath = '',
     label = 'ra.action.show',
     record = {},
+    source = 'id',
     ...rest
 }) => (
     <Button
         component={Link}
-        to={`${linkToRecord(basePath, record.id)}/show`}
+        to={`${linkToRecord(basePath, get(record, source))}/show`}
         label={label}
         {...rest}
     >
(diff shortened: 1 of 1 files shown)


fzaninotto:
Hi Nicolas, and thanks for your contribution. 

Could you explain a bit more about your use case? All resources must have an `id` field, why would you want to use something else?

Besides, your PR needs documentation, and if there is already a unit test for that component, a new test for the feature you added.

Oupsla:
Hello @fzaninotto ,

To explain my use case, here is an example.

I got User and Organisation. 
User and Organisation are linked with a many-to-many reliationship (with a liaison table, and this liaison table contains additional data about the relation).

I'm using graphql to query my data and I end up with a query like this
`{
  Organisation(id: 1) {
    id
    identification
    users {
      id
      type
      user {
        id
        email
        firstname
      }
    }
  }
}` 

And in my organisationShow : 

`<Show title={<OrganisationTitle />} {...props}>
    <TabbedShowLayout>
      <Tab label="summary">
        <TextField source="id" />
        <TextField source="identification" />
      </Tab>
      <Tab label="users" path="users">
        <ArrayField source="users">
          <Datagrid>
            <TextField label="User Id" source="user.id" />
            <EmailField label="Email" source="user.email" />
            <TextField label="Firstname" source="user.firstname" />
            <TextField label="Liaison Type" source="type" />
            <ShowButton source="user.id" basePath="/User" />
          </Datagrid>
        </ArrayField>
      </Tab>
    </TabbedShowLayout>
  </Show>`

As you can see my user (and his id) is nested in the users record (that represents the many-to-many relation).
Tell me if you think it is a good idea in this case to add the source property on Button.

I'll work on the documentation and unit test 👍

fzaninotto:
This is very specific, I don't think we'll support it in the core. You can easily roll on your own `<ShowButton>` with this custom logic.

So I'm closing this PR, thanks for submitting it anyway!

Issue #2372, opened 2018-10-01
ra-data-json-server UPDATE_MANY uses PATCH instead of PUT


jheiska:
**What you were expecting:**
UPDATE_MANY to send multiple PUT calls.

**What happened instead:**
Multiple PATCH calls are made instead.

**Steps to reproduce:**
Use ra-data-json-server UPDATE_MANY -action.

**Other information:**
https://github.com/marmelab/react-admin/blob/master/packages/ra-data-json-server/src/index.js

**Environment**

* React-admin version: 2.3.2

Mclovinn:
I would like to work on this

Kmaschta:
@Mclovinn Go for it! Just open a PR and we'll help you.
Be sure to read the [contribution guidelines](https://github.com/marmelab/react-admin/blob/master/.github/CONTRIBUTING.md).

jheiska:
It does, but I think it's confusing that UPDATE uses PUT but UPDATE_MANY uses PATCH. Especially when it's stated in the file:

https://github.com/marmelab/react-admin/blob/master/packages/ra-data-json-server/src/index.js

"json-server doesn't handle filters on UPDATE route, so we fallback to calling UPDATE n times instead"

Also, in my case, the backend doesn't support PATCH calls at all so it's a big problem for me. I don't know if this is a common problem or is PATCH usually supported? I'm quite new to web development.

fzaninotto:
The problem is that your fix is a breaking change. Anyone using the `ra-data-json-server provider` with a custom API modeled after the provider will have the update_many stop working.

So there are two ways to handle this:

1. Consider that the provider was only made for json-server in the first place. As it supports both PUT and PATCH for updates, it's not a BC break. Anyone who modeled an API based on that provider is in a non-supported use case, and we don't care if they break
2. Delay this change until we release a major version

@djhi any opinion?

jheiska:
Sorry, I don't completely understand what you're saying. There won't be a change until the next major release and I should just update my backend to support PATCH calls?

Maybe you should change the comments in ra-data-json-server? As it is, UPDATE_MANY doesn't call the same http method as UPDATE, so saying "we fallback to calling UPDATE n times" is quite misleading.

fzaninotto:
Fixed by #2382

PR #2382, opened 2018-10-02
[WIP] Change HTTP PATH method by PUT on UPDATE_MANY action

Mclovinn:
- [x] Fix HTTP method
- [x] Check regresion
- [x] Run prettier

See #2372

### DIFF
--- a/packages/ra-data-json-server/src/index.js
+++ b/packages/ra-data-json-server/src/index.js
@@ -138,7 +138,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
             return Promise.all(
                 params.ids.map(id =>
                     httpClient(`${apiUrl}/${resource}/${id}`, {
-                        method: 'PATCH',
+                        method: 'PUT',
                         body: JSON.stringify(params.data),
                     })
                 )


Mclovinn:
@Kmaschta maybe you can help me.
How can I test my little change? Any idea?

Kmaschta:
This package doesn't have unit tests. Yet!
We use [jest](https://jestjs.io/) for unit tests. It is already configured and you just have to run it with `make test-unit-watch`.
You'll just have to create a new file `index.spec.js`, there is a lot of these files, you can take them as examples.

[Data providers](https://marmelab.com/react-admin/DataProviders.html) like `ra-data-simple-rest` are basically a function which takes an `httpClient` and return a closure containing a big switch to build a request.

For your case, in order to test your change, you'll have to:
- mock the HTTP client
- call the data provider with the correct fixtures `simpleRestProvider('url', mockedHttpClient)(UPDATE_MANY, ...)`
- test that mocked HTTP client is correctly called with the correct arguments, including the method `PUT`

fzaninotto:
As explained in #2372, this is not just a fix, it's a potential BC break. That being said, I'm willing to accept it on a minor release since it's only a BC break for people who replicated (imperfectly) the API of JSONRestServer instead of using their lib.

However, this needs to be done on the `next` branch rather than on `master` - even a minor BC beak is a no-go for a bugfix release. Can you please rebase on `next`?

PR #2391, opened 2018-10-04
[RFR] Custom User Menu icon

Luwangel:
The idea is to let the user choosing the Icon but also the Component without rewriting the whole UserMenu.

## Todo

- [x] Make possible the change of the UserMenu icon
- [x] Make possible the change of the UserMenu icon component
- [x] Take Screenshots
- [x] Write docs

## Screenshots

**Changing UserMenu icon**

selection_107

**Changing UserMenu component**

selection_113

### DIFF
--- a/packages/ra-ui-materialui/src/layout/UserMenu.js
+++ b/packages/ra-ui-materialui/src/layout/UserMenu.js
@@ -53,7 +53,7 @@ class UserMenu extends React.Component {
                         color="inherit"
                         onClick={this.handleMenu}
                     >
-                        {icon && cloneElement(icon)}
+                        {icon}
                     </IconButton>
                 </Tooltip>
                 <Menu
--- a/packages/ra-ui-materialui/src/layout/UserMenu.js
+++ b/packages/ra-ui-materialui/src/layout/UserMenu.js
@@ -11,7 +11,6 @@ class UserMenu extends React.Component {
         children: PropTypes.node,
         label: PropTypes.string.isRequired,
         logout: PropTypes.node,
-        customIcon: PropTypes.node,
         icon: PropTypes.node,
         translate: PropTypes.func.isRequired,
     };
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/layout/UserMenu.js
```diff
@@ -36,23 +39,35 @@ class UserMenu extends React.Component {
...
+            icon,
+            logout,
+            translate,
+        } = this.props;
         if (!logout && !children) return null;
         const { anchorEl } = this.state;
         const open = Boolean(anchorEl);
 
+        const menuIconProps = {
+            'arial-label': label && translate(label, { _: label }),
```
djhi: We should provide a default label too imo. Something like `Profile` ?

Luwangel: Currently there is a default label used as default prop: `ra.auth.user_menu`.

djhi: Oh sorry missed it

Luwangel:
By custom icon you mean remove the possibility to change the node and keep only the overriding of the icon prop by another icon from material-ui?

packages/ra-ui-materialui/src/layout/UserMenu.js
```diff
@@ -36,23 +38,24 @@ class UserMenu extends React.Component {
     };
 
     render() {
-        const { children, label, logout, translate } = this.props;
+        const { children, label, icon, logout, translate } = this.props;
         if (!logout && !children) return null;
         const { anchorEl } = this.state;
         const open = Boolean(anchorEl);
 
+        const menuIconProps = {
```
fzaninotto: I don't need you need to extract the props now that only the `icon` prop is accepted.

PR #2401, opened 2018-10-08
Allow to override SnackBar style in Notification Component

### LINKED ISSUE #2400: Cannot override SnackBar style in Notification Component
Since Notification component is using Snackbar, that would be great to customize SnackBar style for example.

**What you were expecting:**
I want to customize `anchorOriginBottomCenter` css property from SnackBar : 
`<Notification classes={{ root: ..., anchorOriginBottomCenter: classes.anchorPosition }}`

**What happened instead:**
`<Notification />` ignore classes 

**Steps to reproduce:**
Create a notification component, try to modify SnackBar through the css api https://material-ui.com/api/snackbar/#css-api

rameshsyn:


packages/ra-ui-materialui/src/layout/Notification.js
```diff
@@ -104,6 +104,7 @@ class Notification extends React.Component {
                         </Button>
                     ) : null
                 }
+                classes
```
djhi: This is most probably not working. Have you tested it ? You're not passing any value to the `classes` prop here, so it's probably considered a boolean

djhi: Moreover, to avoid warnings from JSS, the `classes` object should probably be cleaned up by removing classes unknown to the Snackbar component (`confirm`, `warning` and `undo`)

rameshsyn:
@djhi Thanks for pointing out the issue in the code.

djhi:
Thanks for contributing! However, there's still one thing as I pointed out in my comments: to avoid warnings from JSS, the `classes` object should probably be cleaned up by removing classes unknown to the Snackbar component (`confirm`, `warning` and `undo`).

Can you check there are no warnings ?

packages/ra-ui-materialui/src/layout/Notification.js
```diff
@@ -71,7 +71,7 @@ class Notification extends React.Component {
             hideNotification,
             ...rest
         } = this.props;
-
+        const { warning, confirm, undo, ...snackbarClasses } = classses;
```
djhi: Typo here. It should be `classes` with only 2 `s`. You obviously didn't test your changes at all so I'm not comfortable with merging your PR.

fzaninotto:
This is a new feature, and as such the PR should be made against `next`, not `master`

packages/ra-ui-materialui/src/layout/Notification.js
rameshsyn: @djhi I just tested in example demo, there is also duplicate variable declaration "undo" in same render function scope. Is there any other solution to remove unknown classes? I could fix the scope issue by separating the logic to function. 
```
getSnackbarClasses = ({ warning, confirm, undo, ...snackbarClasses }) => snackbarClasses

// In render function
const snackbarClasses = this.getSnackbarClasses(classes);
```

djhi: Yup, you can use the ability to rename a property when destructuring.

```js
const {
    warning,
    confirm,
    undo: undoClass, // Rename classes.undo to undoClass in this scope to avoid name conflicts
    ...snackbarClasses,
} = classes;
```

rameshsyn: Thanks, that's a nice solution. :smile:

rameshsyn:
@djhi  @fzaninotto Pull request against `next` branch.

PR #2409, opened 2018-10-10
[RFR] Button - don't display empty label and center icon if no label

natrim:
If you provide no label to Button, then on mobile mode there is empty tooltip visible
snimek obrazovky 2018-10-10 v 9 40 52

and on desktop mode the icon is not center due to empty span with padding
snimek obrazovky 2018-10-10 v 9 41 01

this PR fixes that

btw. no idea if this should go to next or master, so made it for next first

### DIFF
--- a/packages/ra-ui-materialui/src/button/Button.js
+++ b/packages/ra-ui-materialui/src/button/Button.js
@@ -45,23 +45,29 @@ const Button = ({
 }) => (
     <Responsive
         small={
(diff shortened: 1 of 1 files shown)


afilp:
I have noticed that too, thanks for the fix! I believe it could go to master also, but @fzaninotto has the final saying.

fzaninotto:
Thanks!

PR #2410, opened 2018-10-10
Passing props into react-autosuggest

AskseL:
Adds ability for passing props into React-Autosuggest using AutocompleteInput (such as focusInputOnSuggestionClick)

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.js
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.js
@@ -453,6 +453,7 @@ export class AutocompleteInput extends React.Component {
             source,
             className,
             options,
+            ...rest
         } = this.props;
         const { suggestions, searchText } = this.state;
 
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -444,6 +444,7 @@ export class AutocompleteInput extends React.Component {
             source,
             className,
             options,
+            autosuggestProps = {}
```
djhi: We should use the same convention as for the other input/field components: pass all unknown props to the root component, here `Autosuggest`. 

I suggest replacing this prop by `...rest`

fzaninotto:
I've rebased the `next` branch on `master` (it was lagging behind the latest minor). Could you please rebase your PR on the new `next`?

AskseL:
changed custom props to {...rest}

packages/ra-ui-materialui/src/input/AutocompleteInput.js
AskseL: done

fzaninotto:
Thanks!

PR #2411, opened 2018-10-10
Use 'network-only' apollo option for graphql introspection.

### LINKED ISSUE #2397: graphql introspection returns null data
Using `ra-data-opencrud` (https://github.com/Weakky/ra-data-opencrud) which again uses `introspection.js` in `ra-data-graphql`:

**What you were expecting:**
I would expect introspection not to crash, because inspecting the network request shows that there is in fact a valid schema returned by the server.

**What happened instead:**
 the graphql introspection query throws an error:

```
Uncaught (in promise) TypeError: Cannot read property '__schema' of null
    at introspection.js:76
```

**Steps to reproduce:**
1. Run a Prisma server which returns a valid response to introspection at `localhost:4466`
2. Create a simple react app using OpenCrud provider

**Related code:**
```
import buildOpenCrudProvider from 'ra-data-opencrud';
class App extends Component {
  ...
  componentDidMount() {
    buildOpenCrudProvider({ clientOptions: { uri: 'http://localhost:4466/' } })
      .then(dataProvider => this.setState({ dataProvider }));
  }
  ...
}
```

**Other information:**
A bit of investigation from my end points in the direction of `fetchPolicy` in apollo client. React-admin already sets `fetchPolicy: 'network-only'` for ordinary graphql queries, (which I figure is a fix for this exact problem) but not for the introspection. A related apollo-client issue is https://github.com/apollographql/apollo-client/issues/3030, but it appears it has never been properly fixed. I suggest setting `fetchPolicy: 'network-only'` to the introspection query also to work around this bug in apollo-client.

**Environment**

* React-admin version: 2.3.3
* React version: 16.4.14
* Browser: any Chrome/Firefox

The crashing code in is in `packages/ra-data-graphql/src/introspection.js`:
```
export default async (client, options) => {
    const schema = options.schema
        ? options.schema
        : await client
              .query({
                  query: gql`
                      ${introspectionQuery}
                  `,
              })
              .then(({ data: { __schema } }) => __schema); // <-- crashes here: data is null
    ...
};
```

audunhalland:
Fixes #2397 

Not very trivial to test with jest, as it's a browser-only issue. The change mimics the behaviour already already in place for regular graphql queries (See issue).

### DIFF
--- a/packages/ra-data-graphql/src/introspection.js
+++ b/packages/ra-data-graphql/src/introspection.js
@@ -33,6 +33,7 @@ export default async (client, options) => {
         ? options.schema
         : await client
               .query({
+                  fetchPolicy: 'network-only',
                   query: gql`
                       ${introspectionQuery}
                   `,


djhi:
Thanks!

PR #2417, opened 2018-10-11
[RFR] Fix double asterix on required RadioButtonGroupInput

### LINKED ISSUE #2395: RadioButtonGroupInput: required() validation adds two asterisks instead of one
Self explanatory title.

fzaninotto:
Closes #2395

Incidentally, this bug teaught me that material-ui now handles the required label at the `FormControl` level. We'll have to use that in the future.

### DIFF
--- a/packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
+++ b/packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
@@ -132,17 +132,11 @@ export class RadioButtonGroupInput extends Component {
         return (
             <FormControl
                 component="fieldset"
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
```diff
@@ -132,17 +132,11 @@ export class RadioButtonGroupInput extends Component {
         return (
             <FormControl
                 component="fieldset"
-                required={isRequired}
```
fzaninotto: `FieldTitle` already handles that (see line 144)

PR #2421, opened 2018-10-11
Fix snippet to be correct if copy-pasted

daytonn:
I copy-pasted this example. It did not work. Implicit return is finicky and is probably an anti-pattern because of the inconsistency.

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -100,33 +100,35 @@ const PostActions = ({
     resource,
     selectedIds,
     showFilter
-}) => (
-    <CardActions>
-        {bulkActions && React.cloneElement(bulkActions, {
-            basePath,
-            filterValues,
-            resource,
-            selectedIds,
-            onUnselectItems,
-        })}
-        {filters && React.cloneElement(filters, {
-            resource,
-            showFilter,
-            displayedFilters,
-            filterValues,
-            context: 'button',
-        }) }
-        <CreateButton basePath={basePath} />
-        <ExportButton
-            resource={resource}
-            sort={currentSort}
-            filter={filterValues}
-            exporter={exporter}
-        />
-        <RefreshButton />
-        {/* Add your custom actions */}
-        <Button primary onClick={customAction}>Custom Action</Button>
-    </CardActions>
+}) => {
+    return (
+        <CardActions>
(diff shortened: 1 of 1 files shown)


djhi:
> Implicit return is finicky and is probably an anti-pattern because of the inconsistency

I use implicit return all the time. Can you elaborate?

daytonn:
This is my personal opinion which is why I prefaced it with "probably". I certainly don't begrudge the use of implicit return generally but since implicit return has specific syntactic preconditions, I find it to be an error-prone form. In documentation, there is a high likelihood of people copying examples to be modified in the implementation. Avoiding this form will ensure that developers with varying levels of experience will have a consistent experience when using the documentation. I certainly don't mean to tell you how to write your documentation, but I find consistency to be of higher value than syntactic sugar in exemplary code.

djhi:
Ok but then, why not fix all the implicit returns ?

djhi:
@fzaninotto Any thoughts ?

daytonn:
@djhi I didn't want to be presumptuous and change everything. I encountered this problem specifically so I thought I'd issue a pull request. If it is generally accepted that the documentation should refrain from using implicit returns I could certainly do so.

djhi:
I see, good point :+1: Thanks for raising the issue btw, it is a very valid one. I'll wait for @fzaninotto opinion on this though

fzaninotto:
The original snippet doesn't work because `customAction` is undefined. But it's still undefined in your version, so it doesn't solve the problem. Or am I missing something?

Also, I fail to understand the problems you mention about implicit returns, can you give examples of cases it won't work?

daytonn:
@fzaninotto implicit returns require the parenthesis on multiple lines which are easy to miss. It's not that they don't work, it's that they require specific syntax depending on the context. For this reason they are error-prone and can trip up less-experienced developers. This is my opinion so feel free to disregard if you disagree.

In regards to this fixing the error I don't recall if I changed that or not so it very well could be that I simply misdiagnosed the issue. I didn't notice the parenthesis so it looked like an error to me. This either illustrates my point or just proves I need glasses.

fzaninotto:
Well, I don't agree, so I won't merge this one. No hard feelings though your next contributions are welcome!

PR #2426, opened 2018-10-12
[RFR] Bootstrap TypeScript migration

fzaninotto:
We've decided to gradually migrate react-admin to TypeScript. Explaining why isn't the purpose of this description, but we expect more stability and a better developper experience.

Our strategy is:

- [x] Switch transpiler from Babel to TypeScript (both for CJS and ESM builds)
- [x] Change the minimum code in the packages to make it work (mostly `export *`)
- [x] Change one file from the core to TypeScript to make sure the whole chain works
- [x] Check that the simple example still works
- [x] Migrate the simple example to Babel 7
- [x] Replace ESLint by TSLint
- [x] Check that the demo example still works
- [x] Check that tests still pass
- [ ] ~~Document the new standard~~(it will only be a standard for packages we've migrated)

Then, we'll rename files from .js to .ts/.tsx and add types little by little (outside of the scope of this PR)

Comments are welcome!


tsconfig.json
```diff
@@ -0,0 +1,78 @@
+{
+    "compilerOptions": {
+        /* Basic Options */
+        "target":
+            "ES3" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
```
djhi: Es3 ?

djhi: Isn't it `tsc` which is used to build now according to the package.json files ?

fzaninotto:
@brikou we use babel to test react-admin with the simple example, but we'll use Typescript to compile the ES6/TS code to ES5 for packaging, so we need both the `jsx:react` and the emit.

tsconfig.json
fzaninotto: yes indeed

PR #2427, opened 2018-10-12
added suggestion limit prop to AutocompleteInput

mohinderps:


oksuz:
@djhi I need this improvement also.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -455,6 +462,7 @@ export class AutocompleteInput extends React.Component {
             options,
         } = this.props;
         const { suggestions, searchText } = this.state;
+        const slicedSuggestions = this.sliceSuggestions(suggestions);
```
fzaninotto: The result will be a new array for each `render()`, which will force a rerender of the autosuggest even if the suggestions didn't change. This is bad, so you should slice when setting the suggestions in the state, not when reading the state.

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -520,6 +528,7 @@ AutocompleteInput.propTypes = {
     translate: PropTypes.func.isRequired,
     translateChoice: PropTypes.bool.isRequired,
     inputValueMatcher: PropTypes.func,
+    suggestionLimit: PropTypes.number,
```
fzaninotto: please document and unit test this new prop

mohinderps:
Okay, Will do.

fzaninotto:
Hi! Could you find a few minutes tu update your PR?

fzaninotto:

PR #2454, opened 2018-10-20
Update introspection type check with queryType and mutationType

yishus:
Checking for queryType and mutationType instead opens up the possibility for data providers to make use of GraphQL endpoints where the type name for query and mutation types are not 'Query' and 'Mutation'.


BatuAksoy:
https://github.com/marmelab/react-admin/blob/56b1d4096a5e68e442d29b1b7b12699d3bfc3f4b/packages/ra-data-graphql/src/introspection.js#L46

I have an introspection which has mutationType = null, is it valid according to GraphQL Spec?
If it is, promise fails with a TypeError, because it cannot find 'name' in null.

PR #2386, opened 2018-10-03
Add ra-jsonapi-client to third-party Data Providers

henvo:
Hey!

First of all: Awesome project.

I needed a data provider for my JSONAPI backend and first tried this client:
[moonlight-labs/aor-jsonapi-client](https://github.com/moonlight-labs/aor-jsonapi-client)

Unfortunately this project is deprecated and does not seem to be maintained anymore.
It was already marked as deprecated with this pull request a couple of days ago:
https://github.com/marmelab/react-admin/pull/2300

So I've written my own:
[henvo/ra-jsonapi-client](https://github.com/henvo/ra-jsonapi-client)

The client is now in a reasonably stable state and I would like to submit this to the third-party data providers.

Let me know what you think and have a great day.

PR #2455, opened 2018-10-20
Switch jsonapi data provider in ecosystem chapter

henvo:
This is a follow up PR for https://github.com/marmelab/react-admin/pull/2386 where we switched the data provider for JSONAPI and forgot to change this also in the 'Ecosystem' chapter.

Also a small typo fix.


fzaninotto:
Thanks!

PR #2467, opened 2018-10-24
Fix Can't use SimpleList whitin edit form

### LINKED ISSUE #2462: Can't use SimpleList whitin edit form
i've use with Responsive component for xsmall size SimpleList
as for react-admin 2.1 it is works fine but for 2.4 it is broken
i've reproduced the sample scenario for that bug

the same applied for show and edit-form

just select any post with comments and go to the comment it wil not display any, while for 2.1 for the original sample it is working

[sample project](https://codesandbox.io/s/qk3rz6mj5w)

vedmalex:
Fixes #2462

it seems like it fixing issue. because the component not always recieves the total property, especially when it is used inside show or form

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList.js
+++ b/packages/ra-ui-materialui/src/list/SimpleList.js
@@ -58,7 +58,7 @@ const SimpleList = ({
     total,
     ...rest
 }) =>
-    (isLoading || total > 0) && (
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks for your contribution! It does solve the issue, however I think the right way it to pass the `total` to `ReferenceManyField`. I'll do that in another PR.

fzaninotto:
See #2487

PR #2499, opened 2018-11-05
add default item value on SimpleFormIterator

ronfroy:
we need to set default value when we create a new item. The value must be an object or a string (when  the array is an array of string)

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -85,9 +85,9 @@ export class SimpleFormIterator extends Component {
     };
 
     addField = () => {
-        const { fields } = this.props;
+        const { fields, defaultItemValue } = this.props;
         this.ids.push(this.nextId++);
-        fields.push({});
+        fields.push(defaultItemValue);
     };
 
     render() {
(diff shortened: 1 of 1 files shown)


ronfroy:
@fzaninotto hello. what's about this change?

fzaninotto:
I don't understand the use case because I see no example usage. Also, I thin kthis is a new feature, so it should be PRed against `next`.

ronfroy:
My first use case is simple:

I want to have default values when adding a new item on an SimpleFormIterator 

My second use case is tu initialize new item with string to lets the
  SimpleFormIterator to deal with array of string and not only array of object.

afilp:
@fzaninotto We also have this need, somwhow different maybe, we need the `defaultValue` per field to be respected for new items of the iterator, not just for the 1st one.

Our use case is this:

```
         <SimpleFormIterator>

            <BooleanInput
              label="Enabled"
              source="enabled"
              defaultValue={true}
              style={{display: 'inline-flex'}}
            />
...etc...
```

JackyW83:
I notice same behavior, the defaultValue prop of input don't work in iterator except first one

fzaninotto:
This bug doesn't appear on `master` (v3) anymore. I'm closing the PR, thanks for the patch anyway!

PR #2502, opened 2018-11-05
Should render suggestions

ginman86:
This PR allows for the creator of an `<AutocompleteInput />` to pass in an optional `shouldRenderSuggestions` function. This function is supported by `react-autosuggest` and is documented here: https://github.com/moroshko/react-autosuggest#should-render-suggestions-prop

ex: `<AutocompleteInput shouldRenderSuggestions={(val) => { return val.trim() > 2 }} />` would not render any suggestions until the 3rd character was entered.


fzaninotto:
Thanks!

PR #2505, opened 2018-11-06
fix(styles): remove incorrect quotes

moklick:


moklick:
From the spec: ["Generic font family names are keywords and must NOT be quoted."](https://www.w3.org/TR/CSS21/fonts.html#propdef-font-family)

fzaninotto:
Thanks!

PR #2521, opened 2018-11-08
[RFR] Migrate ra-core actions to Typescript

xavierhans:
Migrate the files in the `action` folder of the `ra-core` package to Typescript


packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -2,7 +2,19 @@ import { crudGetMany, crudGetMatching } from './dataActions';
 
 export const CRUD_GET_MANY_ACCUMULATE = 'RA/CRUD_GET_MANY_ACCUMULATE';
 
-export const crudGetManyAccumulate = (resource, ids) => ({
+export const crudGetManyAccumulate = (
+    resource: string,
+    ids: []
+): {
+    type: string;
```
fzaninotto: can't you be stricter with the type, given that we always know *which* string the function returns?

packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -11,12 +23,19 @@ export const crudGetManyAccumulate = (resource, ids) => ({
...
-) => {
+    reference: string,
+    relatedTo: string,
+    pagination: object,
+    sort: object,
+    filter: object
+): {
+    type: string;
+    meta: {
+        accumulate: any;
```
fzaninotto: no, it should be `() => object`

packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -2,7 +2,19 @@ import { crudGetMany, crudGetMatching } from './dataActions';
 
 export const CRUD_GET_MANY_ACCUMULATE = 'RA/CRUD_GET_MANY_ACCUMULATE';
 
-export const crudGetManyAccumulate = (resource, ids) => ({
+export const crudGetManyAccumulate = (
```
fzaninotto: You will need to export the return type for each action creator, because it will be used in the reducers.

packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -2,7 +2,19 @@ import { crudGetMany, crudGetMatching } from './dataActions';
...
+    resource: string,
+    ids: []
+): {
+    type: string;
+    payload: {
+        resource: string;
+        ids: [];
+    };
+    meta: {
+        accumulate: any;
```
fzaninotto: For the metas, we need to be a bit smarter. As each meta has a special signification (and triggers a particular side effect), we probably need a subtype for `accumulate`, to be reused everywhere the `accumulate` meta is present.

fzaninotto:
Switching to RFR. All actions and reducers aren't merged, but nothing is blocking the merge and we'll continue little by little.

fzaninotto:
There is no way to tell git that it's a rename when its internal heuristics don't detect it. So back to RFR.

xavierhans:
Looks good to me, there might be a lot of `any` used for the actions payload. I don't know if we can be more precise on these types.

Issue #2506, opened 2018-11-07
Datagrid's last column doesn't line up / has wrong padding


aliang:
**What you were expecting:**
Header of last column of Datagrid should line up with data in column

**What happened instead:**
Header of last column has no padding, but cells in the table have 12px padding, thus making the cells not line up

**Steps to reproduce:**
None in particular, but you need to resize your browser window and/or switch the view in the CodeSandbox to be wide enough to show the Tags column. It won't show up in the narrow Responsive view because the last column is hidden.

**Related code:**
https://codesandbox.io/s/9o0jo2myqo

**Other information:**
![image](https://user-images.githubusercontent.com/162961/48104079-c80fb380-e1e6-11e8-97e0-7777207cd84e.png)

**Environment**

* React-admin version: 2.4.1
* Last version that did not exhibit the issue (if applicable):
* React version: 16.3.1
* Browser: Chrome
* Stack trace (in case of a JS error):

djhi:
Hi and thanks for reporting. If you look at the dev console, you'll see that the last column has the same padding as the others, only its width changes and it's set automatically by material-ui. I'm afraid there is nothing we can do about that

aliang:
Thanks for the quick response.

I think we might be talking about a different comparison. It sounds like you're comparing the last column with the other columns, but I am comparing the last column's header with the last column's data rows.

- In the last column, the `th` padding is `0`, but the `tr` padding is `0 12px`.
- In the other columns, the `th` and `tr` padding are both `0 12px`.

I think the fix might be in https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/Datagrid.js. I haven't had a chance to test it yet, but I think that `styles.headerCell` needs the same padding style that is in `styles.rowCell`:
```
'&:last-child': {
    padding: '0 12px',
}
```

aliang:
Also here are a couple more screenshots that might clarify things.

Styles and padding for `th`:
<img width="899" alt="image" src="https://user-images.githubusercontent.com/162961/48116038-a845b300-e219-11e8-9e5e-c9e7494c6b98.png">

Styles and padding for `td`:
<img width="892" alt="image" src="https://user-images.githubusercontent.com/162961/48116079-cc08f900-e219-11e8-86b0-40e2858f5531.png">

fzaninotto:
Fixed by #2522

PR #2522, opened 2018-11-08
Fix Datagrid header cell padding

aliang:
This fixes the issue I described in #2506. It didn't look like there are any automated tests around this so I checked it in the app built by `make run-simple`. Let me know if there is anything else I can help with around this issue.

### DIFF
--- a/packages/ra-ui-materialui/src/list/Datagrid.js
+++ b/packages/ra-ui-materialui/src/list/Datagrid.js
@@ -22,6 +22,9 @@ const styles = {
     },
     headerCell: {
         padding: '0 12px',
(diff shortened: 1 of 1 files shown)


Haraldson:
Tried installing this pull request directly by doing `yarn add marmelab/react-admin#2522/head`, but no dice. Also tried using `patch-package` on `ra-ui-materialui` without any luck – I guess this change has to be baked into react-admin in a build step?

Crossing my fingers the PR will be accepted soon, it looks absolutely horrible at the moment. :)

fzaninotto:
Thanks!

PR #2528, opened 2018-11-13
Automatically focus on 1st input (username), on load of LoginForm

afilp:
A minor change. 

Reason?

When using `KeepPass`, it can add `username/password` automatically with CTRL-V (inside KeePass).

However, the 1st input (username) needs to be focused. So, I have to first focus there with the mouse, then go to KeePass, etc.

I also believe that it is nice to be able to start typing immediately, without having to focus somewhere (like GMAIL login does, for example).

Thanks!


fzaninotto:
Also, this is a new feature, so please PR against the `next` branch instead of `master`.

afilp:
@djhi You are right! I was unaware of this prop there. I will do it like you suggest.
@fzaninotto I will merge with next.

afilp:
Sorry, did not do that correctly, I will it again in a new PR

PR #2530, opened 2018-11-13
Automatically focus on 1st input (username), on load of LoginForm

afilp:


afilp:
(Sorry for closing the old one and creating a new one, I do not use GitHub a lot and I messed up something with the 2nd commit when trying to do it through here)

djhi:
Thanks !

PR #2532, opened 2018-11-14
Accept object as valid parameter for toolbar

pnthang01:
This input should accept object as valid parameter for toolbar for advance custom.

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -18,7 +18,7 @@ export class RichTextInput extends Component {
         meta: PropTypes.object,
         options: PropTypes.object,
         source: PropTypes.string,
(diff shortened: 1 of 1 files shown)


fzaninotto:
Can you explain the use case and document the new option? 

Also, this is technically a new feature, so it should be PRed against the `next` branch

pnthang01:
Hi, base on this Quill's documentation for toolbar, it can custom container and handlers(object). And I need to to use this custom handlers but your component only accepts bool or array as param.
https://quilljs.com/docs/modules/toolbar/
Thanks

fzaninotto:
OK, still waiting for the documentation and the rebase against `next`

fzaninotto:
Bump

fzaninotto:
No news for too long, closing

PR #2535, opened 2018-11-14
[RFR] Add types to side effects

fzaninotto:
Part of the ongoing effort to add TypeScript support to ra-core


packages/ra-core/src/reducer/admin/resource/list/selectedIds.ts
```diff
@@ -1,17 +1,31 @@
...
+    SetListLelectedIdsAction,
     TOGGLE_LIST_ITEM,
+    ToggleListItemAction,
 } from '../../../../actions/listActions';
 import { CRUD_DELETE_OPTIMISTIC } from '../../../../actions/dataActions';
 
 const initialState = [];
 
-export default (previousState = initialState, action) => {
+type State = any[];
```
djhi: I think every reducer should probably export its state definition and we should make a complete state definition by composing all of them in the root `types.ts` file

fzaninotto: I think that `combineReducer` does it. Let's wait until the main reducer is migrated to TS to see.

djhi: As discussed yes, but that won't allow your users to have a typed state in their `mapStateToDispatch` functions

PR #2539, opened 2018-11-15
Add disabled option in SelectInput

### LINKED ISSUE #2512: Disable option in SelectInput
I've added list of user statuses into our admin page, and transition to few of them should be not possible, the problem is that it's not possible to mark option as disabled. This is supported in material ui, since `MenuItem` has prop `disabled` but `SelectInput` is not passing that prop down

```
<SelectInput source="status" choices={[
  { id: "DISABLED", name: "disabled", disabled: true},
  { id: "ENABLED", name:"enabled"}
]}
```

mikaoelitiana:
Fixes #2512


fzaninotto:
Neat!

As it's a new feature, could you open the PR against the `next` branch rather than the `master` branch?

mikaoelitiana:
@fzaninotto  I will close this one and create another one

PR #2542, opened 2018-11-16
Add missing route parameter

trsteel88:


docs/Authorization.md
```diff
@@ -248,7 +248,7 @@ const MyPage = ({ permissions }) => (
         }
     </Card>
 )
-const MyPageWithPermissions = ({ location, match }) => (
+const MyPageWithPermissions = ({ location, match }, route) => (
```
fzaninotto: nope, functional components only accept one parameter.

trsteel88: The function inside is referencing a route parameter which does not exist though. Where does that come from?

Issue #2399, opened 2018-10-08
Upgrade Material UI to Latest Stable


ragboyjr:
I'm pretty new to material ui, but it looks like latest stable is at 3.2, but react-admin is still on ^1.4. When implementing customizations to the react-admin, it caused me to lose some time because I was trying to figure out why certain Material UI code wasn't working and then I found out react-admin just uses an older version.

Is there anything stopping this project from being able to bump up to material ui 3.2 other than just the work to port it over?

Thanks again for all the hard work into this library!

fzaninotto:
Material-ui 3 is a BC break over material-ui 1, so upgrading that dependency would be a BC break for react-admin, too.We'd have to release react-admin 3.0. Since the latest stable (2.0) is not that old (May 2018), we'll wait a few more weeks before breaking BC again.

oliviertassinari:
Material-UI v3 introduces the following breaking change over Material-UI v1:
1. Drop support of Firefox [45-51]

fzaninotto:
We haven't listed the browser versions we support. So in react-admin terms, dropping support for FF < 52 isn't a BC break per se ;)

However, that would still require a 3.0, and major versions frighten users because of the upgrade cost. I'd rather continue using an old material-ui version for another couple of months rather than rushing a major just to get the new features.

@oliviertassinari this is where our release strategies diverge.

phacks:
@fzaninotto Hi! Coming back to this issue since recent (post `v3.0.0`) versions fix a bug with Modals I encountered in my React Admin project. Looks like [`v3.3.1`](https://github.com/mui-org/material-ui/releases/tag/v3.3.1) contains the actual bug fix I need.

I totally understand the need to not jump majors too frequently and I appreciate the consideration you have regarding developer experience. However, would you have an estimation of a timeline for the release of major version updating MaterialUI ? Depending on that timeline, I can either try and help you release and/or monkey-patch MaterialUI on my end.

Should you need more info about the bug I can try and create a reproduction repo, but I’m not sure that it is valuable since the fix was already shipped in either https://github.com/mui-org/material-ui/pull/13351 or https://github.com/mui-org/material-ui/pull/13389.

fzaninotto:
The guys at material-ui are hard at work planning the next BC break with the hooks rewrite, so we'll definitely wait for that effort to finish before releasing a major on our side.

phacks:
Oh yes that makes sense, thanks for the update!

djhi:
[...]

PR #2551, opened 2018-11-21
[#2399] chore: hardcode TablePaginationAction proptypes

phacks:
_See #2399 for context._

The import

```
import TablePaginationActions from '@material-ui/core/TablePaginationActions';
```

does not work anymore in material-ui’s more recent versions, as per
https://github.com/mui-org/material-ui/pull/13524.

This commit makes transitioning from `material-ui@1.4.1` to `material-ui@3.5.1` easier.


fzaninotto:
weird, some tests are failing which are not related to your PR. But it seems you definitely broke one test.

PR #2556, opened 2018-11-22
Add 'icon' prop to buttons

afilp:
We had the need to customize the icon shown on the `ShowButton`.

This commit changes just the `ShowButton`, so that you comment whether you like this change or not.

If you agree with this, let me know so that we also update the rest of the buttons located inside the `button` folder.

Thanks.

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkDeleteButton.js
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteButton.js
@@ -36,6 +36,24 @@ const styles = theme => ({
 });
 
 class BulkDeleteButton extends Component {
+    static propTypes = {
+        basePath: PropTypes.string,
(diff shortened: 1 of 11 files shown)


fzaninotto:
Love it! Would you mind going a bit further and do the same for other buttons?

afilp:
Sure! That was the idea! First to see if you want it, then take the time to do the rest. I will do this most likely during the weekend.

Thanks!

afilp:
Added 2nd commit for the rest of the buttons.

Please note that in some "class"-using buttons I moved the `propTypes` and `defaultProps` as `static`. This was done for consistency (some classes used the `static` approach, others the named declaration at the end).

djhi:
Thanks!

PR #2562, opened 2018-11-24
functionality to access to global store

astiag:
General purpose of this PR is to have access global store in react-admin out side of a react component in our projects.
it will be helpful in cases that we are using some custom restClient provider.
this will help us to access important info inside store like locale to add it in request headers

### DIFF
--- a/docs/Translation.md
+++ b/docs/Translation.md
@@ -162,6 +162,34 @@ class LocaleSwitcher extends Component {
 export default connect(undefined, { changeLocale: changeLocaleAction })(LocaleSwitcher);
 ```
 
+## Sending locale info in requests header
+
+You can access global store. this can be achieved by using `getStore`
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks for the PR and the explanation. 

To get the current locale from inside the dataProvider, I don't think you should use the store. You see, the store purpose is to store data for the UI. Here, you're making it store data for something else. Also, the store is supposed to be used only by `connect` - exposing it is a leaky abstraction.

For your use case, update your i18nProvider to store the current locale somewhere (e.g localStorage or global variable), and update your dataProvider to access that locale. It's your job to make the two providers communicate - and the Redux store should not be involved in the process.

So I'm closing this PR as I don't see any use case where exposing the store would be the only possible solution, and exposing it is too dangerous.

Thanks for your PR anyway.

Issue #2534, opened 2018-11-14
AppBar Toolbar customization using children


alanpoulain:
**Is your feature request related to a problem? Please describe.**
If you want to add elements to the `Toolbar` of the `AppBar`, you need to copy all the `AppBar` into your custom one.

**Describe the solution you'd like**
Why not allowing children to be passed to the `Toolbar`? Maybe between `Typography` and `LoadingIndicator`.

**Describe alternatives you've considered**
It could also be a render prop for the `Toolbar`.

**Additional context**
Example:
```js
import React from 'react'
import { AppBar } from 'react-admin'

export default props =>
  <AppBar {...props}>
    <!-- Toolbar components here -->
  </AppBar>
```

fzaninotto:
Fixed by #2777

PR #2573, opened 2018-11-26
render children inside AppBar

astiag:
this PR will render children inside AppBar component

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -409,22 +409,6 @@ const MyAppBar = props => <AppBar {...props} userMenu={<MyUserMenu />} />;
 const MyLayout = props => <Layout {...props} appBar={MyAppBar} />;
 ```
 
-`AppBar` can accept children so you are able to add more custom components in it.
-
-```jsx
-import { AppBar, UserMenu, MenuItemLink } from 'react-admin';
-import SomeCustomComponent from './SomeCustomComponent';
-
-const MyAppBar = props => (
-    <AppBar {...props}>
-        <SomeCustomComponent />
-    </AppBar>
-);
-
-const MyLayout = props => <Layout {...props} appBar={MyAppBar} />;
-
-```
-
 You can also customize the default icon by setting the `icon` prop to the `<UserMenu />` component.
 
 {% raw %}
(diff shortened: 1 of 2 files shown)


fzaninotto:
please add documentation and example usage

alanpoulain:
Related to this: https://github.com/marmelab/react-admin/issues/2534

fzaninotto:
Let me rephrase my remark: please add documentation and example usage in the files of your PR (i.e. modify the react-admin documentation, not only the code).

alanpoulain:
I can but I am not the author of this PR :sweat_smile:

astiag:
I know your rules about having everything documented.
but the thing is you already have it [here](https://github.com/marmelab/react-admin/blob/master/docs/Theming.md#using-a-custom-appbar)

btw your docs regarding to adding children to `AppBar` will cause some problems

```
const MyAppBar = props => (
    <AppBar {...props}>
        <Toolbar>
            <Typography variant="title" id="react-admin-title" />
        </Toolbar>
    </AppBar>
);
```
in this case you will have `Toolbar` inside `Toolbar`. [see here](https://github.com/marmelab/react-admin/blob/1f47c28b87b3b63b1455894a05a2d818d102c283/packages/ra-ui-materialui/src/layout/AppBar.js#L68)

so I also will fix this one in a new commit in this PR

alanpoulain:
@astiag No I don't think it's the case. You are mixing up `AppBar` from `@material-ui` and `AppBar` from `ra-ui-materialui`. The first can take children (and a `Toolbar`), the second can't.

astiag:
@alanpoulain you are right. didnt notice that.
anyway I'll add something in documents about having children inside `AppBar`

astiag:
@fzaninotto please review again

docs/Theming.md
```diff
@@ -409,6 +409,22 @@ const MyAppBar = props => <AppBar {...props} userMenu={<MyUserMenu />} />;
 const MyLayout = props => <Layout {...props} appBar={MyAppBar} />;
 ```
 
+`AppBar` can accept children so you are able to add more custom components in it.
```
djhi: This should probably go inside the `Using a Custom AppBar` section.

fzaninotto: 👍

fzaninotto:
Any progress?

docs/Theming.md
astiag: will change it asap

BartoGabriel:
I think it would be good to combine this feature, with the possibility of disabling the LoadingIndicator or the UserMenu of the AppBar.
Since the user having the ability to customize AppBar, he could create their own components, without having to make a new AppBar.
With two properties it could be solved.

astiag:
> Any progress?
@fzaninotto @djhi I've moved the description to `Using a Custom AppBar` section

docs/Theming.md
astiag: done

fzaninotto:
Thanks for the PR, I've finally decided to do a slightly different implementation (replacing the content with children instead of appending the children to the default content) in #2777. So I'll close this PR.

PR #2577, opened 2018-11-26
Change node polyglot dependency tag

hamidfzm:
Node polyglot has some issues with Farsi locale in version 2.2.0 and it is fixed in 2.3.0. But as I looked tag version in ra-core is 2.2.0 and creates conflicts with 2.3.0. I suggest using ^2.2.0 to fix this issue.


fzaninotto:
Thanks!

Issue #528, opened 2017-03-30
Empty content on first load with RichTextInput


DjLeChuck:
First, sorry for the double post: https://github.com/marmelab/aor-rich-text-input/issues/3

I don't know if this issue is related to aor in general or is specific to RichTextInput...

Coming from the listing, the field is empty. After hitting refresh button, the field is filled.

If I log the value received in `componentDidMount`, I see it's empty the first time and not the second one.

![demo](https://media.giphy.com/media/xUPGcdJejn3pIleTJu/giphy.gif)

DjLeChuck:
The bug only appear if we come from the listing. Using a direct link to the edit page is OK (that's why a reload works...)

DjLeChuck:
After digging more, I see the component receive the value after his initialization. So, if I use `componentWillReceiveProps` I can fill Quill:

```
  componentWillReceiveProps(nextProps) {
    const {
      input: {
        value,
      },
    } = nextProps;

    this.quill.pasteHTML(value);
  }
```

fzaninotto:
Yes, this is a bug for marmelab/aor-rich-text-input, closing this one.

PR #2581, opened 2018-11-27
Add componentWillReceiveProps for RichTextInput

abstractart:
Today i found bug with ```RichTextInput``` component. On first load it will be empty.
Proposed solution: call ```quill.pasteHTML``` in ```componentWillReceiveProps```.

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -47,6 +47,12 @@ export class RichTextInput extends Component {
         this.quill.on('text-change', debounce(this.onTextChange, 500));
     }
 
+    componentWillReceiveProps(nextProps) {
+       if (nextProps.input.value !== this.props.input.value) {
+          this.quill.pasteHTML(nextProps.input.value);
+       }
+    }
+
     componentWillUnmount() {
         this.quill.off('text-change', this.onTextChange);
         this.quill = null;


fzaninotto:
Can you explain the bug in more details, and provide a CodeSandbox to reproduce it?

abstractart:
1. When i click to Edit button i navidated to Form. RichTextInput component is empty.
2. When i press ctrl+f5 RichTextInput is filled. 

First time i searched by issues and found [similar](https://github.com/marmelab/react-admin/issues/528).

Next i read code in [repo](https://github.com/marmelab/aor-rich-text-input) and found [this].(https://github.com/marmelab/aor-rich-text-input/blob/master/src/index.js#L41)

I Append this function to my local installation of ```ra-input-rich-text```  and rebuild. RichTextInput fills on first load.

userbq201:
+1

fzaninotto:
I'm sorry, I can't seem to reproduce the bug in the simple example. Could you please provide a link to a CodeSandbox proving the issue?

fzaninotto:
Any progress?

userbq201:
@fzaninotto i forgot. And Today i will reproduce

userbq201:
I don't to reproduce. Example crash with babel-runtime error.
So the error is that `this.quill.setContents` call onse in `componentDidMount` method.
If props come after mount - value not set. This is critical bug.
To solve this problem you need add `componentDidUpdate` or `willReceiveProps` and call `this.quill.setContents` if has value

fzaninotto:

afilp:
@fzaninotto 

I also believe there is a problem with `RichTextInput` not updating on changes of its REDUX form value, when this change comes from outside this particular input.

This is our case:
We use the same RichTextInput `source` (eg: `title.en`) twice in the same form (the 2nd one is in a "translations" dialog, which is part of the form). When we change the `title.en` rich text in the "translations" dialog and then close the dialog, unfortunately the `title.en` outside the dialog (but in the same form) still shows the old text.

It does seem that something *special* needs to happen when its input props change, like the proposed `this.quill.pasteHTML(nextProps.input.value);`

Does the above description help on defining the problem?

afilp:
Since my use case is different and therefore a different solution may be needed, I created a new issue to follow there:  https://github.com/marmelab/react-admin/issues/2832

Thanks!

PR #2586, opened 2018-11-29
Allow icon to be changed on SaveButton

trsteel88:
Allows the SaveButton icon to be customised.

Also solves https://stackoverflow.com/questions/53177554/how-to-change-savebutton-icon-in-react-admin


trsteel88:
Oh good to hear. Thanks @djhi.

Will keep that in mind for next time.

PR #2589, opened 2018-11-29
[RFR] Speed Up Default Login Page Time To Interactive

Kmaschta:
## Problem

CSS background image block the main thread when loading an image.
On the default Login page, we load a HUGE image and it makes the login form unresponsive with a slow internet connection.

## Solution

This PR makes the background image loading async so that it don't block the interactivity.


fzaninotto:
Thanks!

Issue #2483, opened 2018-10-29
Export button should not show up on empty List page


fzaninotto:
When new users discover a react-admin app, and they have no data yet, the list screen is intimidating.  It shows a "no results found" and an export button...

![image](https://user-images.githubusercontent.com/99944/47641685-019d4a80-db67-11e8-8b40-4834881c6522.png)

- If there is no result, the export button should not appear - or should appear greyed out

afilp:
@fzaninotto Regarding the default "EXPORT" functionality, in most cases we do not need this, as it would require us to do major work on custom "exporter" functions.

Instead of us having to create our own `actions` component, could we just use `exporter = {false}` in order to disable the "EXPORT" functionality? (not show it at all).

The above proposal is exactly similar to what we do with `bulkActions={false}`.

afilp:
Concerning my latest comment above, I have created a PR for your evaluation.

I already use this option in my projects (i.e. hiding the export button inn some Lists) and I have the feeling that others may need it too. It is up to you to decide. Thanks!

PR #2595, opened 2018-11-30
[RFR] Disable the Export button when there is nothing to export

Luwangel:
Related to this issue https://github.com/marmelab/react-admin/issues/2483

## Todo

- [x] Check the total and disable the export button when the total is 0
- [x] Write doc

selection_137

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -100,7 +100,8 @@ const PostActions = ({
     onUnselectItems,
     resource,
     selectedIds,
-    showFilter
+    showFilter,
+    total
 }) => (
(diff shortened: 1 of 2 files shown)


fzaninotto:
Nice! Don't forget to add the `total` propType defineition. Also, please PR against `next` as this is a new feature.

fzaninotto:
Thanks!

fzaninotto:
@Luwangel You change triggers a warning on Mobile in the simple example (because the Export button is minimized, and uses a tooltip). Would you mind taking a look?

PR #2596, opened 2018-11-30
[RFR] Login Background Image: handle empty ref

Kmaschta:
Even though the React doc ensure the ref creation is done before the
componentDidMount, it can happen that the ref is set to null until the
next render.

So, to handle this case the component will now try to load the image on
the componentDidMount, but if the ref doesn't exist, it will try again
on the following componentDidUpdate. The try will be done only once.

Refs:
- https://reactjs.org/docs/refs-and-the-dom.html

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -471,7 +471,9 @@ const App = () => (
 );
 ```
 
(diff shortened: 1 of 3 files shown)


Kmaschta:
Good point :+1:

PR #2600, opened 2018-12-01
[RFR] Fixes arrow in sorted column header when field use sortBy props

### LINKED ISSUE #2306: Arrow for sortorder not appearing on header with sortBy
**What you were expecting:**
To see an arrow next to the sorted column header

**What happened instead:**
No arrow indicating the sort direction

**Steps to reproduce:**
<List title={<ListTitle/>} {...props} filters={<LotenFilter/>}>
        <Datagrid>
            <ReferenceField label="Prop1" source="projectId" reference="projects" sortBy="prop1">
                <TextField source="prop1"/>
            </ReferenceField>
            <TextField label="Prop2" source="prop2"/>
        </Datagrid>
    </List>

**Environment**

* React-admin version: 2.3.0
* React version: 16.4.2
* Browser: chrome

louisbl:
Fixes #2306

### DIFF
--- a/packages/ra-ui-materialui/src/list/Datagrid.js
+++ b/packages/ra-ui-materialui/src/list/Datagrid.js
@@ -156,8 +156,8 @@ class Datagrid extends Component {
                                         currentSort={currentSort}
                                         field={field}
                                         isSorting={
-                                            field.props.source ===
-                                            currentSort.field
+                                            currentSort.field ===
+                                            (field.props.sortBy || field.props.source)
                                         }
                                         key={field.props.source || index}
                                         resource={resource}


packages/ra-ui-materialui/src/list/Datagrid.js
```diff
@@ -156,8 +156,8 @@ class Datagrid extends Component {
                                         currentSort={currentSort}
                                         field={field}
                                         isSorting={
-                                            field.props.source ===
-                                            currentSort.field
+                                            currentSort.field ===
```
fzaninotto: the || feels a bit weird here. Would you mind making the two conditions explicit?

```
isSorting={field.props.sortBy ? currentSort.field === field.props.sortBy : currentSort.field === field.props.source}
```

louisbl: I use the same style as seen in DatagridHeaderCell [here](https://github.com/marmelab/react-admin/blob/60fa0bdd43a11f4514cdd830d2d07811bde7f5dd/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js#L42) and [here](https://github.com/marmelab/react-admin/blob/60fa0bdd43a11f4514cdd830d2d07811bde7f5dd/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js#L56).
Should I change the || though?

louisbl:
Where do I add the test?
In the 
[list integration test]https://github.com/marmelab/react-admin/blob/master/cypress/integration/list.js?

packages/ra-ui-materialui/src/list/Datagrid.js
fzaninotto: Oh, you're right. So for consistency's sake, let's keep it like that.

fzaninotto:
yes, exactly

louisbl:
@fzaninotto I've added some integration tests for the sort arrow on column header.

The last commit is a bit problematic: in order to *really* test this bug fix I had to modify the simple demo example to add the `sortBy` props on the `tags` ReferenceField.

The problem is that the `ra-data-fakerest` does not seem to support sorting by a reference field so the rows are actually not sorted.

PR #2601, opened 2018-12-01
[RFR] Included 'position' prop to AppBar component.

JulioITS:
Following the official material-ui [docs](https://material-ui.com/api/app-bar/#props), added the position prop.

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -90,7 +90,7 @@ const VisitorFilter = withStyles(filterStyles)(({ classes, ...props }) => (
     </Filter>
 ));
 
-// The List component supports the `root`, `header`, `actions` and `noResults` CSS classes. Here we override the `header` and `actions` classes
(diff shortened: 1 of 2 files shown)


JulioITS:
Hello @djhi, I completely missed the `...rest` part.

Nonetheless, the precedence must be taken into account here.
Since the `AppBar.js` specifies those `props`, it would overwrite my own `props`, as you can see in this [Pen](https://codepen.io/juliobulgarelli/pen/bQZYvR).

I see 2 possibilities here:

1. Accept them as `props` (what I suggested with the PR)
2. Remove the specifications from the `AppBar.js` file

Looking forward for your considerations.

Thanks for taking the time to evaluate this PR.

djhi:
That's because you excluded your overriden props from the rest object you're passing to the underlying component:

```js
const {color, backgroundColor, ...rest} = this.props;
```

Just pass `this.props` and they will override the defaults:

```js
return(
  <AppBar color="white" backgroundColor="black" {...this.props} />
);
```

JulioITS:
Hello again.

Understood, thanks for the explanation.

PR #2614, opened 2018-12-03
Add enableReducers and renderProp option to TestContext to help with integration testing

### LINKED ISSUE #2612: enable reducers within `TestContext` component
While writing tests around our react-admin system, we found that we were not able to assert statements which rely on form actions being dispatched.

Rather than relying on the internals, and implementation details of our components, we'd prefer to test our forms as users would interact with them.  This means (simulating) clicking buttons and then asserting correct actions have been dispatched.  Currently, because reducers aren't integrated within `TestContext`, anything that happens after a dispatch (e.g. form submissions) won't be fired.

I would like to be able use real reducers when using `TestContext`.

To prevent this from being a breaking change (i.e. once setup, actions will be dispatched and peoples tests may start breaking if they didn't expect this to happen e.g. dispatch an action may update state or unmount components), I propose adding a `enableReducers` flag.  This flag would swap the static state currently returned buy the store within `TestContext` with working reducers.

Alternatively, we could use the real react-admin `Admin` component within our tests, but this feels like is expects, and would require, a lot more boilerplate to get working.

The other alternative would be to test the internals of our components.  but I'd prefer not to as this doesn't ensure all aspects are plumbed together properly.

**Additional context**
I expect this would feel, for some people, less like unit tests and more like integration tests.  While this is true I do not see a downside to enabling people to perform more integration-like tests earlier on. 

This is the proposed code:
```js

const TestContext = ({ store, enableReducers, children }) => {
    const storeWithDefaults = enableReducers
        ? createStore(combineReducers(defaultReducers), merge(defaultStore, store))
        : createStore(() => merge(defaultStore, store));
    
    return (
        <Provider store={storeWithDefaults}>
            <TranslationProvider>{children}</TranslationProvider>
        </Provider>
    );
};
```

### LINKED ISSUE #2613: return the store when using TestContext
When trying to write tests using TestContext we would like to assert on certain actions being dispatched.

I propose, using renderProps, to return the created `store` from `TestContext` to enable developers to spy on the store.dispatch method.

the alternative would be to unit test `mapDispatch` method directly, outside the context of the components.  This would be brittle and allow mistakes to creep in.  I feel it would be better to test the component completely plumbed in.

**Additional context**
This would be a non-breaking change and the renderProp option would be opt-in.

he is an example of the proposed code:

```js
const TestContext = ({ store, enableReducers, children }) => {
    const storeWithDefault = createStore(() => merge(store, defaultStore));

    const showChildren = () => (typeof children === 'function' ? children({ store: storeWithDefault }) : children);

    return (
        <Provider store={storeWithDefault}>
            <TranslationProvider>{showChildren()}</TranslationProvider>
        </Provider>
    );
};
```

peter-mouland:
* Add children as a renderProp option
    * return the store if the `children` is a function
 * Add `enableReducers` prop option
   * will allow dispatching of actions

> these are opt-in change, not breaking changes.

fixes #2613
fixes #2612

### DIFF
--- a/docs/UnitTesting.md
+++ b/docs/UnitTesting.md
@@ -46,3 +46,39 @@ You can then provide additional props, as needed, to your component (such as the
 
 At this point, your component should `mount` without errors and you can unit test your component.
 
+
+## Enabling reducers to ensure actions are dispatched
+
+If you component relies on a a reducer, e.g. redux-form submission, you can enable reducers using the `enableReducers` prop:
+
+```jsx harmony
+    myCustomEditView = mount(
+        <TestContext enableReducers>
+            <MyCustomEditView />
+        </TestContext>
+    );
+```
+
+This means that reducers will work as they will within the app.  For example, you can now submit a form and redux-form will cause a re-render of your component.
+
+
+## Spying on the store 'dispatch'
+
+If you are using `mapDispatch` within connected components, it is likely you will want to test that actions have been dispatched with the correct arguments.  You can return the `store` being used within the tests using a `renderProp`. 
+
+```jsx harmony
+    let dispatchSpy;
+    myCustomEditView = mount(
+        <TestContext>
+            {({ store }) => {
+                dispatchSpy = jest.spyOn(store, 'dispatch');
+                return <MyCustomEditView />
+            }}
+        </TestContext>,
+    );
+    
+    it('should send the user to another url', () => {
+        myCustomEditView.find('.next-button').simulate('click');
+        expect(dispatchSpy).toHaveBeenCalledWith(`/next-url`);
+    });
+```
--- a/packages/ra-core/src/util/TestContext.js
+++ b/packages/ra-core/src/util/TestContext.js
@@ -1,19 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { createStore } from 'redux';
+import { createStore, combineReducers } from 'redux';
 import { Provider } from 'react-redux';
 import { reducer as formReducer } from 'redux-form';
 import { TranslationProvider } from 'ra-core';
 import merge from 'lodash/merge';
 
-const defaultStore = {
-    admin: {
-        resources: {},
-        references: { possibleValues: {} },
-        ui: { viewVersion: 1 },
-    },
-    form: formReducer(),
-    i18n: { locale: 'en', messages: {} },
+import createAdminStore from '../createAdminStore';
+
+export const defaultStore = {
+  admin: {
+    resources: {},
+    references: { possibleValues: {} },
+    ui: { viewVersion: 1 },
+  },
+  form: formReducer(),
+  i18n: { locale: 'en', messages: {} },
 };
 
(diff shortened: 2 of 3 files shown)


packages/ra-core/src/util/TestContext.js
```diff
@@ -16,6 +16,12 @@ const defaultStore = {
     i18n: { locale: 'en', messages: {} },
 };
 
+const defaultReducers = {
```
fzaninotto: why don't you use react-admin's `createAdminStore` instead?

packages/ra-core/src/util/TestContext.js
```diff
@@ -28,23 +34,41 @@ const defaultStore = {
...
+ * );
  */
-const TestContext = ({ store, children }) => {
-    const storeWithDefault = createStore(() => merge(store, defaultStore));
+const TestContext = ({ store, enableReducers, children }) => {
+    const storeWithDefault = enableReducers
+        ? createStore(combineReducers(defaultReducers), merge(defaultStore, store))
+        : createStore(() => merge(defaultStore, store));
+
+    const showChildren = () => (typeof children === 'function' ? children({ store: storeWithDefault }) : children);
```
fzaninotto: better rename it to `renderChildren`

docs/UnitTesting.md
```diff
@@ -46,3 +46,39 @@ You can then provide additional props, as needed, to your component (such as the
...
+## Spying on the store 'dispatch'
+
+If you are using `mapDispatch` within connected components, it is likely you will want to test that actions have been dispatched with the correct arguments.  You can return the `store` being used within the tests using a `renderProp`. 
+
+```jsx harmony
+    let dispatchSpy;
+    myCustomEditView = mount(
+        <TestContext>
+            {({ store }) => {
+                dispatchSpy = jest.spyOn(store, 'dispatch');
```
fzaninotto: smart. But I'm not sure we're note testing implementation details. @djhi, what do you think?

packages/ra-core/src/util/TestContext.js
peter-mouland: ah nice - didn't realise that was there.  I'll check out `createAdminStore ` and update the PR in the morning.

docs/UnitTesting.md
peter-mouland: yeah, i've tussled with this too. I've settled with this solution because it covers most bases, i.e normally we could test actions are dispatched by testing the UI updates, but some side-effects don't affect UI, or sometimes multiple things happen and only update the UI once.  Because of this, i think spying on the dispatch function and directly asserting it is called with the correct actions is the better way off ensuring the component/containers are written and integrated properly.  I've seen others test the `mapDispatch` function directly or only test the action was called, but both of these miss out important steps.

Happy to hear your thoughts though or show you the real-life use case for more clarity.

peter-mouland:
I've now rebased from `next` and made the suggested changes.  Let me know if you have any other questions.

PR #2634, opened 2018-12-07
[RFR] Add ability to attach an expansion panel to a datagrid row

fzaninotto:
- [x] Add `expand` prop to `<Datagrid>`
- [x] Document it
- [x] Test it

peek 07-12-2018 15-02


cypress/integration/list.js
```diff
@@ -171,16 +171,50 @@ describe('List Page', () => {
...
+            cy.contains('Annamarie Mayer')
+                .parents('tr')
+                .click();
             cy.contains('Summary').should(el => expect(el).to.exist);
         });
     });
+
+    describe('expand panel', () => {
+        it('should show an expand button opening the expand element', () => {
+            cy.contains('1-10 of 13'); // wait for data
```
djhi: You should add an expectation here: `.should(el => expect(el).to.exist)`

cypress/integration/list.js
```diff
@@ -171,16 +171,50 @@ describe('List Page', () => {
...
+            cy.get('[role="expand"]')
+                .eq(0)
+                .click();
+            cy.get('[role="expand-content"]').should(
```
djhi: I avoid magic strings expectations as they prevent tools such as the IDE autocompletion and type checking to do their job. I suggest to use the following form: `.should(el => expect(el).to.contain(''))`

cypress/integration/list.js
```diff
@@ -171,16 +171,50 @@ describe('List Page', () => {
...
+                'contain',
+                'Curabitur eu odio ullamcorper, pretium sem at, blandit libero. Nulla sodales facilisis libero, eu gravida tellus ultrices nec. In ut gravida mi. Vivamus finibus tortor tempus egestas lacinia. Cras eu arcu nisl. Donec pretium dolor ipsum, eget feugiat urna iaculis ut.'
+            );
+            cy.get('.datagrid-body').should(
```
djhi: same

cypress/integration/list.js
```diff
@@ -171,16 +171,50 @@ describe('List Page', () => {
...
+
+        it('should accept multiple expands', () => {
+            cy.contains('1-10 of 13'); // wait for data
+            cy.get('[role="expand"]')
+                .eq(0)
+                .click();
+            cy.get('[role="expand"]')
+                .eq(1)
+                .click();
+            cy.get('[role="expand-content"]').should('have.length', 2);
```
djhi: same

cypress/integration/list.js
```diff
@@ -171,16 +171,50 @@ describe('List Page', () => {
...
+                'Curabitur eu odio ullamcorper, pretium sem at, blandit libero. Nulla sodales facilisis libero, eu gravida tellus ultrices nec. In ut gravida mi. Vivamus finibus tortor tempus egestas lacinia. Cras eu arcu nisl. Donec pretium dolor ipsum, eget feugiat urna iaculis ut.'
+            );
+            cy.get('.datagrid-body').should(
+                'not.contain',
+                '[role="expand-content"]'
+            );
+        });
+
+        it('should accept multiple expands', () => {
+            cy.contains('1-10 of 13'); // wait for data
```
djhi: Missing expectation

cypress/integration/list.js
fzaninotto: I don't want to do an assertion here, just wait. Also, the other tests in this file don't do that either.

cypress/integration/list.js
fzaninotto: Same as above, it's just a wait

cypress/integration/list.js
fzaninotto: I've done some reading. 

The Cypress docs says you're wrong:

https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#For-instance

>**Core Concept**
>
>All DOM based commands automatically wait for their elements to exist in the DOM.
>
>You never need to write .should('exist') after a DOM based command.

And it repeats it:

https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-wait-for-cy-get

And it repeats it again:

https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Asynchronous

I only found one person saying what you say:

* https://stackoverflow.com/questions/53428220/cypress-io-contains-not-waiting-for-element

I am very unsure that you're right. If you can reproduce the `contains` not waiting before executing the next cypress command behavior, I suggest you open an issue in the Cypress bug tracker, because that's a bug of theirs.

Until then, I'll assume Cypress does what their doc says.

djhi: I admit I haven't tested it

PR #2644, opened 2018-12-10
Fix DataGrid export

### LINKED ISSUE #2637: DatagridBody is not exported from react-admin
**What you were expecting:**
importing following statement:
```js
import {DatagridBody as RaDatagridBody} from 'react-admin';
```

**What happened instead:**
Got following error:
```bash
Failed to compile.

./src/resources/Transaction/list/DatagridBody.js
Attempted import error: 'DatagridBody' is not exported from 'react-admin' (imported as 'RaDatagridBody').
```

**Steps to reproduce:**
Just import it :)

**Other information:**
As documented in https://marmelab.com/react-admin/List.html#the-datagrid-component, we can import datagrid body from react admin but it's not possible.

**Environment**

* React-admin version: 2.5.1

hamidfzm:
Export DatagridBody, DatagridRow, DatagridHeaderCell, and DatagridCell according to documentation to customize data grid at any level
Fixes #2637

### DIFF
--- a/packages/ra-ui-materialui/src/list/index.js
+++ b/packages/ra-ui-materialui/src/list/index.js
@@ -2,6 +2,10 @@ import BulkActions from './BulkActions';
 import BulkActionsToolbar from './BulkActionsToolbar';
 import BulkDeleteAction from './BulkDeleteAction';
 import Datagrid from './Datagrid';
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

Issue #2571, opened 2018-11-26
ra-data-graphql-simple: problem GET_MANY_REFERENCE in buildVariables.js


wmwart:
Hi.

```
export const  UserEdit = props => {
    return (
        <Edit {...props}>
            <SimpleForm>
                <DisabledInput source="id" />
                <TextInput source="name" />
                <ReferenceManyField
                    reference="Customer"
                    target="userId"
                >
                    <Datagrid>
                        <TextField source="organizationName" />
                    </Datagrid>
                </ReferenceManyField>
            </SimpleForm>
        </Edit>
    );
}
```
It was expected that the record would be returned as in the documentation.

The query filter gets an object in the object, which prevents you from working further.
expected filter: 
variables: {filter: {userId:"cjoa0w1z5005y0b91ka78ug3r"}}

Here's what gets into Request payload:
```
{operationName: "allCustomers", variables: {filter: {id: {id: "cjoa0w1z5005y0b91ka78ug3r"}}},…}
operationName: "allCustomers"
query: "query allCustomers($filter: CustomerFilter) {↵  items: allCustomers(filter: $filter) {↵    id↵    type↵    organizationName↵    organizationInn↵    payment↵    __typename↵  }↵  total: _allCustomersMeta(filter: $filter) {↵    count↵    __typename↵  }↵}↵"
variables: {filter: {userId: {id: "cjoa0w1z5005y0b91ka78ug3r"}}}
filter: {userId: {id: "cjoa0w1z5005y0b91ka78ug3r"}}
userId: {id: "cjoa0w1z5005y0b91ka78ug3r"}
id: "cjoa0w1z5005y0b91ka78ug3r"

```

the response from the server:
```
operationName: null
link.js:48
variables:
link.js:49
{
link.js:50
  "_v0_where": {
    "AND": [
      {
        "userId_contains": "[object Object]"
      }
    ]
  },
  "_v1_orderBy": "id_ASC"
}

```
* React-admin version: 2.4.2
* Last version that did not exhibit the issue (if applicable):
* React version:16.5.2
* Browser: Chrome 70.0.3538.110
* Use: ra-data-graphql-simple

fzaninotto:
I'm sorry, I don't understand your issue.

> It was expected that the record would be returned as in the documentation.

Which record? According to which part of the documentation? 

> The query filter gets an object in the object, which prevents you from working further.

What?

> Here's what gets into Request payload

[...]

PR #2649, opened 2018-12-11
ra-data-graphql-simple: Fix filter id applied incorrectly to get many ref

redorb:
Fixes issue #2571


Redorb:
I'll check the failing tests when I get home.

Redorb:
@fzaninotto So I was looking at my actual fix I put in a week or so back in my own codebase and it's a bit different than the one outlined in the PR. Here's the code snippet, I'm doing it in the data provider as a custom build query:

```
const customBuildQuery = introspection => (fetchType, resource, params) => {
const builtQuery = buildQuery(introspection)(fetchType, resource, params);

if (fetchType === 'GET_MANY_REFERENCE') {
		// Targeting this weird bug: https://github.com/marmelab/react-admin/issues/2571
		// Hopefully this gets fixed soon.
		for (let filterOpt in builtQuery.variables.filter) {
			if (
				typeof builtQuery.variables.filter[filterOpt] === 'object' &&
				filterOpt.includes("Id") &&
				builtQuery.variables.filter[filterOpt].id
			) {
				builtQuery.variables.filter[filterOpt] = builtQuery.variables.filter[filterOpt].id
			}
		}
	}
	return builtQuery;
}
```

It's probably overly targeted, but we wanted to be safe without digging really deep into the codebase. Do you know where the equivalent would be? I think it should be just something as simple as:

```
case GET_MANY_REFERENCE: {
    const parts = params.target.split('.');
    return filter: { [parts[0]]: params.id };
}
```

I won't be near my main computer though for a few more days.

Redorb:
Alright after looking through the rest of the code in this area it makes sense. Not sure how the code in that issue worked for the OP though. I amended the commit with something similar to what I'm doing in my workaround, let me know if I need to update anything else.

christiaanwesterbeek:
I encountered the same problem and fixed it according to this PR. I vote to re-open and merge this PR.

djhi:
I don't know why @redorb closed this PR

redorb:
I can't remember why I ended up closing it. Feel free to look through the PR if it's still in need of a fix.

redorb:
Merged master into the branch, sorry if anything is weird. I'm doing these edits in the github UI since I'm not going to be near my laptop for a month or so.

Issue #2651, opened 2018-12-12
Allow passing `replace={true}` to FormTabs Link.


waynebloss:
Hello, I will make a pull request for this feature if you approve. It should be trivial.

**Is your feature request related to a problem? Please describe.**
I want react-router to replace the history entry instead of creating a new one every time you change tabs in TabbedForm. That way the user can quickly use the back button to go back to the list view.

**Describe the solution you'd like**
At the very least, I'd like to be able to have `FormTab` pass `replace={true}` to its `MuiTab/Link` element. It turns out that `<FormTab replace>` *does* do this, but that same `replace={true}` property gets passed down to every input element as well, resulting in:

> Warning: Received `true` for a non-boolean attribute `replace`.
> If you want to write it to the DOM, pass a string instead: replace="true" or replace={value.toString()}.

**Describe alternatives you've considered**
I tried passing `replace="true"` since `react-router-dom`s code for `Link` does a simple truthy test, but they actually guarded the type with `PropTypes`, so that produces:

> Warning: Failed prop type: Invalid prop `replace` of type `string` supplied to `Link`, expected `boolean`.

Despite the warnings, everything works the way I want it to regardless of how I pass the prop. So one other alternative is that I could just ignore this warning. It also builds just fine. It's a bit annoying to have the error during development though.

**Additional context**
This is easy to fix in `FormTab`, just sanitize the `replace` prop and pass it to `MuiTab`.

fzaninotto:
Fixed by #2654

PR #2654, opened 2018-12-12
Fix extra FormTab/Tab props are passed to two different components

waynebloss:
For issue #2651.

I looked at every component that used `Link` and decided that `Tab` and `FormTab` were the only components where this ability really made sense.

I briefly considered adding docs for this prop, but `Tab` and `FormTab` are not documented and neither are their properties, so I thought it best to leave that for a different pull request.

Thank you.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Tab.js
+++ b/packages/ra-ui-materialui/src/detail/Tab.js
@@ -65,7 +65,7 @@ class Tab extends Component {
         />
     );
 
-    renderContent = ({ className, children, ...rest }) => (
+    renderContent = ({ className, children, basePath, record, resource }) => (
         <span className={className}>
             {React.Children.map(
                 children,
@@ -89,7 +89,11 @@ class Tab extends Component {
                             ) : typeof field.type === 'string' ? (
                                 field
                             ) : (
-                                React.cloneElement(field)
+                                React.cloneElement(field, {
+                                    basePath,
+                                    record,
+                                    resource,
+                                })
                             )}
                         </div>
                     )
--- a/packages/ra-ui-materialui/src/detail/Tab.js
+++ b/packages/ra-ui-materialui/src/detail/Tab.js
@@ -83,6 +83,9 @@ class Tab extends Component {
                                 <Labeled
                                     label={field.props.label}
                                     source={field.props.source}
+                                    basePath={basePath}
+                                    record={record}
+                                    resource={resource}
                                 >
                                     {field}
                                 </Labeled>
--- a/packages/ra-ui-materialui/src/detail/Tab.js
+++ b/packages/ra-ui-materialui/src/detail/Tab.js
@@ -83,17 +83,13 @@ class Tab extends Component {
                                 <Labeled
                                     label={field.props.label}
                                     source={field.props.source}
-                                    {...sanitizeRestProps(rest)}
                                 >
                                     {field}
                                 </Labeled>
                             ) : typeof field.type === 'string' ? (
                                 field
                             ) : (
-                                React.cloneElement(
-                                    field,
-                                    sanitizeRestProps(rest)
-                                )
+                                React.cloneElement(field)
                             )}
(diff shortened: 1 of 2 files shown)


fzaninotto:
Wait, it's not the right fix. The problem is that FormTab passes extra props both to the tab header and to the tab inputs. That can never work, because there is no prop that can work on such different components. 

The right fix is to *remove* the `{...sanitizeRestProps(rest)}` in the second part to let you pass any extra prop (like `replace`) to the root element only.

waynebloss:
> The right fix is to remove the `{...sanitizeRestProps(rest)}` in the second part...

So I think what you're saying is to undo all of the other changes and simply remove `{...sanitizeRestProps(rest)}` from [the FormTab renderContent block on this line](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/form/FormTab.js#L38), allowing all of the previously sanitized props (`label, icon, value, translate`) to be passed down to each child of FormTab? (Same for [detail/Tab.js renderContent](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/detail/Tab.js#L86)?)

I don't think this will work because `sanitizeRestProps` is already letting the `replace` prop through to each child of FormTab/Tab and when that happens there is an error:

> Warning: Received true for a non-boolean attribute replace.
> If you want to write it to the DOM, pass a string instead: replace="true" or replace={value.toString()}.

waynebloss:
Oh wait, you're saying to remove it - so none of those props will be passed. I got it, thanks!

waynebloss:
OK, I removed the forwarding of props in the `renderContent` block of both `FormTab` and `Tab`.

In `Tab` there were 2 places that it forwarded props (to the `Labeled` element and the `cloneElement`).

fzaninotto:
You were too aggressive when removing the props you passed to FormInput. See SimpleForm to understand which props need to be passed explicitly.

waynebloss:
OK, sorry about that - I am still waiting over an hour to get this `cypress` package installed and I can't run any tests on my own machine. So, I just pushed to let Travis do the tests.

I looked at the code for `SimpleForm` and I saw what it was passing to `FormInput`. So, I passed those same props from `FormTab`. However, running my own project in developer mode, I don't see any of those props (`basePath`, `record` or `resource`) in devtools being passed to my `FormTab`, so I'm not sure how this is going to forward those props to its children.

fzaninotto:
you'll have the same problem with Tab, see SimpleShowLayout for an example of what props to pass to children (`record`, `resource`, `basePath`).

waynebloss:
OK, I just don't see those props anywhere in devtools on a `Tab` or a `FormTab`, so I wasn't sure at all; wanted to run a test. I'll push that change in a minute.

waynebloss:
I already knew what to do before I looked at `SimpleShowLayout` so I didn't put the props in the same order. Sorry about that.

waynebloss:
I'm guessing that the 1 test is failing because of another prop that is not being passed? I'm not sure where the error is coming from. The stack trace is useless. The `replace` property mentioned in the error has nothing to do with any `replace` property that I was passing, since I removed all reference to that property from `FormTab` and `Tab`. Now, the only place that I find the symbol `replace` being used in the entire package, is in [`RichTextField`](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/field/RichTextField.js#L9)

My original solution passed all tests since it only added the feature and changed nothing else. Now we're working on something else. I could start passing some of the other props that I see in devtools, but it takes too long to run these tests and I already spent over 3 hours on this.

This feature turned into something a lot less trivial and I don't think I can spend anymore time on it. Without this feature, things work in production without any errors. I just wanted to fix an error in my console. Sorry.

fzaninotto:
Thanks for your time. I'll take it from then on and try to finish it.

fzaninotto:
Thanks, you were very close to the target!

PR #2658, opened 2018-12-13
[RFR] Translate FileInput placeholder

djhi:


djhi:
Just noticed it actually accepts a component... :tired_face:

PR #2660, opened 2018-12-13
[RFR] FileInput - Display validation errors

djhi:


examples/simple/src/posts/PostEdit.js
```diff
@@ -70,7 +70,7 @@ const PostEdit = props => (
                         { id: 42, name: 'Sean Phonee' },
                     ]}
                 />
-                <ImageInput multiple source="pictures" accept="image/*" validate={required()}>
```
fzaninotto: I don't understand that change. How are you supposed to test it?

djhi: Well I added it to test manually in the first place but ok, I'll write tests for it :cry:

fzaninotto: no no no, it will get really tedious if the form doesn't accept posts without files. It's ok like that.

PR #2661, opened 2018-12-13
Fix demo application for IE 11

### LINKED ISSUE #2653: react-admin demo not working for IE11
What you were expecting:
That the react-admin demo should work with IE11

What happened instead:
Hangs at loader page. Console error is "Object does not support assign method.."

Steps to reproduce:
Go to https://marmelab.com/react-admin-demo/ using IE 11. Page won't open.

phacks:
Fixes #2653

CRA 2.0 needs polyfills to work with IE11, so I followed [this doc](https://github.com/facebook/create-react-app/blob/master/packages/react-app-polyfill/README.md) and everything seems to be back up for IE11.

### DIFF
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -17,6 +17,7 @@
         "ra-language-french": "^2.0.0",
         "react": "~16.3.1",
         "react-admin": "^2.0.0",
+        "react-app-polyfill": "^0.1.3",
         "react-dom": "~16.3.1",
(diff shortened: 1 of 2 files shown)


djhi:
Thanks!

fzaninotto:
I've redeployed the demo with your fix and checked that it's now OK with IE11. Thanks for your contribution!

PR #2662, opened 2018-12-13
Fix DeleteButton undoable={false} in List does not refresh List

### LINKED ISSUE #2425: <DeleteButton undoable={false} /> in List does not refresh List
Please do not submit support requests or "How to" questions here. For that, go to [StackOverflow](http://stackoverflow.com/questions/tagged/react-admin).

**What you were expecting:**

In List with `<DeleteButton undoable={false} />`, after user click delete, request sent to server, and item deleted from list.

**What happened instead:**

user click delete, request sent to server, but item still remains in the list.

record

**Steps to reproduce:**

add <DeleteButton undoable={false} /> in /examples/simple/src/posts/PostList -> PostListActionToolbar section.

```
                    <PostListActionToolbar>
                        <EditButton />
                        <ShowButton />
                        <DeleteButton undoable={false} />
                    </PostListActionToolbar>

```

**Related code:**

```
```

**Other information:**

**Environment**

* React-admin version: master branch
* React version:
* Browser: Chrome

BartoGabriel:
After making an delete, by default you must refresh.

Closes #2425

### DIFF
--- a/packages/ra-core/src/actions/dataActions.js
+++ b/packages/ra-core/src/actions/dataActions.js
@@ -197,7 +197,8 @@ export const crudDelete = (
     id,
     previousData,
     basePath,
-    redirectTo = 'list'
+    redirectTo = 'list',
+    refresh = true
 ) => ({
     type: CRUD_DELETE,
     payload: { id, previousData },
@@ -212,6 +213,7 @@ export const crudDelete = (
                     smart_count: 1,
                 },
             },
+            refresh,
             redirectTo,
             basePath,
         },


fzaninotto:
Can you describe the bug you’re trying to fix ?

BartoGabriel:
Sorry, I put it in the commit message. #2425

normanzb:
the same change need to be done in crudUpdate as well.

normanzb:
i think the complete fix for this issue should: 
1. add refresh in crudDelete and crudUpdate
2. remove the refresh here https://github.com/marmelab/react-admin/blob/29fc75e19b917bb37952dcbeb881b840f2a07b38/packages/ra-core/src/sideEffect/undo.ts#L41

The refresh of ui shouldn't be triggered by undo effect, it should be triggered by the event that the date has be updated.
Without removing bullet point 2 above, i am afraid the refresh will be triggered twice?

djhi:
> This directly executes the action, refreshes the list, but still displays the Undo button on the toast message...

It displays the `undo` but it won't have the expected effect as the underlying delete action will have been dispatched

te-online:
> It displays the undo but it won't have the expected effect as the underlying delete action will have been dispatched

Exactly :-) That's why it's a drawback – it suggests you can undo the action, but you can't. I wasn't clear on that, thanks for pointing it out.

normanzb:
> But with the `crudUpdate` I never noticed that the list is not refreshed, because I normally don't update any record while being on the list.

That happens to me on edit view. 

I have a custom made edit view, for my edit view, I prefer to stay on edit view once the editing is finished. if I enable undoable for that edit view, even when the update is sent the interface style unchanged. Anyway I think it would be nice to keep the code consistent.

normanzb:
> As to your fear of triggering refresh twice: I couldn't reproduce that. With or without undo, it seems to refresh just once.

I just did test on my edit view, it does trigger twice now when `undoable=true` is set on my edit view.
You can easily test it by insert a break point into the render() function of edit view, when the undo snackbar goes away, there will be 2 consecutive calls.

fzaninotto:
This will add a useless rerender even on optimistic actions, but it's the proper way to fix the bug.

fzaninotto:
Thanks!

PR #2676, opened 2018-12-18
Fix <RichTextInput> not working on IE11

### LINKED ISSUE #2675: <RichTextInput> not working on IE11
**What you were expecting:**

On IE11, `<RichTextInput />` should work as expected.

**What happened instead:**

`<RichTextInput />` causes a “Something went wrong” error on IE11.

image

**Steps to reproduce:**

Visit https://marmelab.com/react-admin-demo/#/products/3/description (demo app) on IE11, with the credentials `demo/demo`

**Other information:**

IE11 throws an `Unspecified Error`, which is not very helpful…

image

**Environment**

* React-admin version: `^2.0.0`
* Last version that did not exhibit the issue (if applicable): I don’t know yet
* React version: `~16.3.1`
* Browser: IE11
* Stack trace (in case of a JS error): see above

phacks:
Fixes #2675

The underlying problem is in the [css-vendor](https://github.com/cssinjs/css-vendor) library, specifically in the cssVendor.supportedValue method that breaks on IE11 with content and '' as parameters.

the `content: ''` attribute is currently not working on Chrome either.

### DIFF
--- a/packages/ra-input-rich-text/src/styles.js
+++ b/packages/ra-input-rich-text/src/styles.js
@@ -18,7 +18,7 @@ export default {
                     right: 0,
                     bottom: 0,
                     height: 1,
-                    content: '',
+                    content: '""',
                     position: 'absolute',
                     transition:
                         'background-color 200ms cubic-bezier(0.4, 0, 0.2, 1) 0ms',
@@ -30,7 +30,7 @@ export default {
                     right: 0,
                     bottom: 0,
                     height: 2,
-                    content: '',
+                    content: '""',
                     position: 'absolute',
                     transform: 'scaleX(0)',
                     transition:


Kmaschta:
Are you sure this change doesn't affect the display of the component?
I would use `content: ' '` with a space, but I'm not sure.

phacks:
@Kmaschta My guess is that since the property is considered as invalid in Chrome (see screenshot), it currently has no effect on the display. I can however try and screenshot multiple situations in Chrome and see if each of them works in IE11: 

- current;
- without the `content` (the suggested PR);
- with a proper `content: ' '`

I may have some time today, otherwise tomorrow

fzaninotto:
The empty content was added in aad754ae6670a491d11e85a3ed585ddec254ff83, probably for a good reason. Removing the `content` field in the CSS style probably removes the entire ::before and ::after styles (see https://stackoverflow.com/questions/17067918/why-do-the-before-and-after-pseudo-elements-require-a-content-property), so I don't think this patch is harmless.

phacks:
@fzaninotto Good catch! I’ll do a more thorough review of any impacts (and/or try to set the content to `' '` instead of `''` and will get back to you with what I found!

phacks:
@fzaninotto Hi, and happy new year 🥳

Found a way better approach to solving this: as per https://github.com/cssinjs/jss/issues/866 (JSS is a dependency of MaterialUI), the property `content: ''` is invalid (which explains why it wasn’t properly parsed by Chrome).

To translate the proper CSS `content: ''` in JSS, one must specify `content: '""'`. I updated the PR to reflect this.

Works like a charm now! However, there still is an issue on the layout (long text requires horizontal scrolling), which I’ll try to fix in a future PR (see screenshot):

image

Thanks!

phacks:
Note : this fix has the side effect of repairing the “selected” visual hints for the input.

_Before:_

jan-12-2019 16-15-07

_After:_

jan-12-2019 16-14-28

Is that OK for you?

djhi:
Thanks!

Issue #2370, opened 2018-10-01
Autocomplete : Popover (dropdown) does not stay underneath


AkselsLedins:
Hello !

**What you were expecting:**
That the dropdown stays underneath and won't cover the input.

**What happened instead:**
The dropdown is covering the input.

**Steps to reproduce:**
Works in demo app:
https://marmelab.com/react-admin-demo/#/reviews?filter=%7B%7D&order=DESC&page=1&perPage=25&sort=date

* Add filter -> Customer
* Click on customer reference input (filter)
* Scroll down a little

![popover-bug](https://user-images.githubusercontent.com/18367406/46276767-69a94280-c561-11e8-8583-6449819f5e93.gif)

you can click outside and click again the input, the dropdown will still be on top on the input, preventing you from typing anything.

It's really annoying :disappointed:

FACOLOMBANI:
I am having the same problem when using AutocompleteArrayInput and AutocompleteInput below a custom component that loads after both inputs.
![autocompleteerror](https://user-images.githubusercontent.com/11721821/46660570-98559780-cbb7-11e8-973b-729556e54a57.gif)

FACOLOMBANI:
I fixed my issue by adding a the prop 'disablePortal={true}' to the Popper component in the renderSuggenstionsContainer.

afilp:
@FACOLOMBANI Whas that change inside the `react-admin` code? Or in your project's code? Thanks.

FACOLOMBANI:
I did this inside react-admin code. I forked the component for customization because i needed to group choices, render groups and secondary information, so that was the  fix to the Popper in Both AutocompleteArrayInput and AutocometeInput. In this page you will find the description for disablePortal prop https://material-ui.com/api/popper/. Hope it helps.

FACOLOMBANI:
@aflip i looked at my code again and saw i also added a modifier in the Popper component, look at the modifiers prop below 

Here is the description https://popper.js.org/popper-documentation.html#modifiers..keepTogether

```
<Popper
                className={className}
                open
                anchorEl={this.inputEl}
                disablePortal={true}
                placement="bottom-start"
                modifiers={{
                    keepTogether: {
                        enabled: true
                    }
                }}
            >
                <Paper square className={classes.suggestionsPaper} {...containerProps}>
                    {children}
                </Paper>
            </Popper>
```

I could create a PR for this.

afilp:
Wait, I just created one, I am pusing now! :-)

[...]

PR #2677, opened 2018-12-18
Fix absence of 'modifiers' and 'disablePortal' for material-ui Popper

afilp:
Related issue: https://github.com/marmelab/react-admin/issues/2370


djhi:
Thanks but I think we shouldn't try to replicate props from the `Popper` component one by one. I'll have a PR for this soon, allowing to pass a `suggestionContainerProps` prop containing the props to pass to the `Popper` component. This is the favored pattern in MUI

afilp:
Ok!

fzaninotto:
superseded by #2678

PR #2678, opened 2018-12-18
[RFR] Allow to override the Popper props in AutocompleteInput

### LINKED ISSUE #2370: Autocomplete : Popover (dropdown) does not stay underneath

djhi:
I'm wondering whether we should adopt the same pattern for overriding the `TextInput` props: instead of passing them directly in options, putting them under a `textInputProps` key.

Fixes #2370


Kmaschta:
I like the idea, it is very explicit.

PR #2679, opened 2018-12-18
[RFR] Document Resource without list prop / Menu

Kmaschta:


docs/FAQ.md
```diff
@@ -106,3 +107,28 @@ You have probably specified a version requirement for `@material-ui/core` that i
...
+
+```jsx
+<Admin>
+    <Resource name="reference" create={PostReference} edit={EditReference} />
+</Admin>
+```
+
+But with the default menu, resources without `list` prop aren't shown.
+
+In order to have a specific resource without `list` prop listed on the menu, you have to have to [write your own custom menu](./Theming.html#using-a-custom-menu).
```
fzaninotto: have to have to

Makefile
```diff
@@ -88,7 +88,13 @@ build-data-generator:
 build: build-ra-core build-ra-ui-materialui build-react-admin build-ra-data-fakerest build-ra-data-json-server build-ra-data-simple-rest build-ra-data-graphql build-ra-data-graphcool build-ra-data-graphql-simple build-ra-input-rich-text build-ra-realtime build-ra-tree-core build-ra-tree-ui-materialui build-data-generator ## compile ES6 files to JS
 
 doc: ## compile doc as html and launch doc web server
-	@yarn -s doc
+	@docker run -it --rm \
```
fzaninotto: please remove and create a new target for docker-doc. Mac users can't use docker, so we must leave them a way out.

PR #2684, opened 2018-12-20
[RFR] Rely on meta to handle optimistic actions

djhi:
Context: https://marmelab.com/react-admin/Actions.html#using-a-custom-action-creator

We were relying on the action type to handle optimistic effects. This prevent custom actions to be handled optimistically too. 

Consider the following case: I have an `ACCEPT_CHANGE` and `REJECT_CHANGE` actions. Both will remove their targeted item from the `change` resource but will be caught by a custom saga to be applied on another resource. From the `change` resource perspective, they should be removed after being either accepted or rejected.

```js
export const acceptChange = (
    id: string,
    previousData: any,
    basePath: string
) => ({
    type: ACCEPT_CHANGE,
    payload: { id, previousData },
    meta: {
        resource: 'changes',
        fetch: DELETE,
        onSuccess: {
            notification: {
                body: 'myapp.notification.accepted',
                level: 'info',
                messageArgs: {
                    smart_count: 1,
                },
            },
            redirectTo: 'list',
            basePath,
        },
        onFailure: {
            notification: {
                level: 'warning',
            },
        },
    },
});
```

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/data.js
+++ b/packages/ra-core/src/reducer/admin/resource/data.js
@@ -1,18 +1,15 @@
 import { FETCH_END } from '../../../actions/fetchActions';
 import {
+    CREATE,
+    DELETE,
+    DELETE_MANY,
     GET_LIST,
-    GET_ONE,
     GET_MANY,
     GET_MANY_REFERENCE,
-    CREATE,
+    GET_ONE,
     UPDATE,
+    UPDATE_MANY,
 } from '../../../dataFetchActions';
-import {
-    CRUD_DELETE_OPTIMISTIC,
-    CRUD_DELETE_MANY_OPTIMISTIC,
-    CRUD_UPDATE_OPTIMISTIC,
-    CRUD_UPDATE_MANY_OPTIMISTIC,
-} from '../../../actions/dataActions';
 
 import getFetchedAt from '../../../util/getFetchedAt';
 
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
 const initialState = {};
 Object.defineProperty(initialState, 'fetchedAt', { value: {} }); // non enumerable by default
 
-export default (previousState = initialState, { type, payload, meta }) => {
-    if (type === CRUD_UPDATE_OPTIMISTIC) {
-        const updatedRecord = { ...previousState[payload.id], ...payload.data };
-        return addRecords([updatedRecord], previousState);
-    }
-    if (type === CRUD_UPDATE_MANY_OPTIMISTIC) {
-        const updatedRecords = payload.ids
-            .reduce((records, id) => records.concat(previousState[id]), [])
-            .map(record => ({ ...record, ...payload.data }));
-        return addRecords(updatedRecords, previousState);
-    }
-    if (type === CRUD_DELETE_OPTIMISTIC) {
-        const { [payload.id]: removed, ...newState } = previousState;
+export default (previousState = initialState, { payload, meta }) => {
+    if (meta && meta.optimistic) {
+        if (meta.fetch === UPDATE) {
+            const updatedRecord = { ...previousState[payload.id], ...payload.data };
+            return addRecords([updatedRecord], previousState);
+        }
+        if (meta.fetch === UPDATE_MANY) {
+            const updatedRecords = payload.ids
+                .reduce((records, id) => records.concat(previousState[id]), [])
+                .map(record => ({ ...record, ...payload.data }));
+            return addRecords(updatedRecords, previousState);
+        }
+        if (meta.fetch === DELETE) {
+            const { [payload.id]: removed, ...newState } = previousState;
 
-        Object.defineProperty(newState, 'fetchedAt', {
-            value: previousState.fetchedAt,
-        });
+            Object.defineProperty(newState, 'fetchedAt', {
+                value: previousState.fetchedAt,
+            });
 
-        return newState;
-    }
-    if (type === CRUD_DELETE_MANY_OPTIMISTIC) {
-        const newState = Object.entries(previousState)
-            .filter(([key]) => !payload.ids.includes(key))
-            .reduce((obj, [key, val]) => ({ ...obj, [key]: val }), {});
+            return newState;
+        }
+        if (meta.fetch === DELETE_MANY) {
+            const newState = Object.entries(previousState)
+                .filter(([key]) => !payload.ids.includes(key))
+                .reduce((obj, [key, val]) => ({ ...obj, [key]: val }), {});
 
-        Object.defineProperty(newState, 'fetchedAt', {
-            value: previousState.fetchedAt,
-        });
+            Object.defineProperty(newState, 'fetchedAt', {
+                value: previousState.fetchedAt,
+            });
 
-        return newState;
+            return newState;
+        }
     }
     if (!meta || !meta.fetchResponse || meta.fetchStatus !== FETCH_END) {
         return previousState;
     }
+
     switch (meta.fetchResponse) {
         case GET_LIST:
         case GET_MANY:
--- a/packages/ra-core/src/reducer/admin/resource/data.js
+++ b/packages/ra-core/src/reducer/admin/resource/data.js
@@ -54,8 +54,16 @@ export const addRecordsFactory = getFetchedAt => (
 
 const addRecords = addRecordsFactory(getFetchedAt);
 
+// We track the last time data was fetched by adding a property on the data which is an array
+// (Hence using defineProperty)
+const updateDataFetchedTime = (data, fetchedAt) => {
+    Object.defineProperty(data, 'fetchedAt', {
+        value: fetchedAt,
+    });
+}
+
 const initialState = {};
-Object.defineProperty(initialState, 'fetchedAt', { value: {} }); // non enumerable by default
+updateDataFetchedTime(initialState, {}); // non enumerable by default
 
 export default (previousState = initialState, { payload, meta }) => {
     if (meta && meta.optimistic) {
@@ -72,9 +80,7 @@ export default (previousState = initialState, { payload, meta }) => {
         if (meta.fetch === DELETE) {
             const { [payload.id]: removed, ...newState } = previousState;
 
-            Object.defineProperty(newState, 'fetchedAt', {
-                value: previousState.fetchedAt,
-            });
+            updateDataFetchedTime(newState, previousState.fetchedAt);
 
             return newState;
         }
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
...
-        const updatedRecords = payload.ids
-            .reduce((records, id) => records.concat(previousState[id]), [])
-            .map(record => ({ ...record, ...payload.data }));
-        return addRecords(updatedRecords, previousState);
-    }
-    if (type === CRUD_DELETE_OPTIMISTIC) {
-        const { [payload.id]: removed, ...newState } = previousState;
+export default (previousState = initialState, { payload, meta }) => {
+    if (meta && meta.optimistic) {
+        if (meta.fetch === UPDATE) {
```
ThieryMichel: A switch would be better IMO

packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
...
-    }
-    if (type === CRUD_DELETE_MANY_OPTIMISTIC) {
-        const newState = Object.entries(previousState)
-            .filter(([key]) => !payload.ids.includes(key))
-            .reduce((obj, [key, val]) => ({ ...obj, [key]: val }), {});
+            return newState;
+        }
+        if (meta.fetch === DELETE_MANY) {
+            const newState = Object.entries(previousState)
+                .filter(([key]) => !payload.ids.includes(key))
```
ThieryMichel: You can put the filter logic directly in the reduce to avoid traversing the array twice

packages/ra-core/src/reducer/admin/resource/data.js
ThieryMichel: Add a getUpdatedRecords function to get the updatedResults

packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
...
-        const { [payload.id]: removed, ...newState } = previousState;
+export default (previousState = initialState, { payload, meta }) => {
+    if (meta && meta.optimistic) {
+        if (meta.fetch === UPDATE) {
+            const updatedRecord = { ...previousState[payload.id], ...payload.data };
+            return addRecords([updatedRecord], previousState);
+        }
+        if (meta.fetch === UPDATE_MANY) {
+            const updatedRecords = payload.ids
+                .reduce((records, id) => records.concat(previousState[id]), [])
```
ThieryMichel: Unless previousState[id] returns an array you can use a map here. And in this case you can do it in a single map

```js
.map(id => ({
    ...previousState[id],
    ...payload.data,
}));
```

packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
...
+                .map(record => ({ ...record, ...payload.data }));
+            return addRecords(updatedRecords, previousState);
+        }
+        if (meta.fetch === DELETE) {
+            const { [payload.id]: removed, ...newState } = previousState;
 
-        Object.defineProperty(newState, 'fetchedAt', {
-            value: previousState.fetchedAt,
-        });
+            Object.defineProperty(newState, 'fetchedAt', {
```
ThieryMichel: why not using the spread operator?

packages/ra-core/src/reducer/admin/resource/data.js
```diff
@@ -60,40 +57,43 @@ const addRecords = addRecordsFactory(getFetchedAt);
...
+        }
+        if (meta.fetch === DELETE_MANY) {
+            const newState = Object.entries(previousState)
+                .filter(([key]) => !payload.ids.includes(key))
+                .reduce((obj, [key, val]) => ({ ...obj, [key]: val }), {});
 
-        Object.defineProperty(newState, 'fetchedAt', {
-            value: previousState.fetchedAt,
-        });
+            Object.defineProperty(newState, 'fetchedAt', {
```
ThieryMichel: spread ?

packages/ra-core/src/reducer/admin/resource/data.spec.js
```diff
@@ -99,19 +96,24 @@ describe('data addRecordsFactory', () => {
     });
 });
 
-describe('Resources data reducer', () => {
-    describe('CRUD_DELETE_OPTIMISTIC', () => {
+describe.only('Resources data reducer', () => {
```
ThieryMichel: only spotted

packages/ra-core/src/reducer/admin/resource/data.spec.js
```diff
@@ -99,19 +96,24 @@ describe('data addRecordsFactory', () => {
...
+    describe('optimistic DELETE', () => {
         it('removes the deleted record', () => {
             const state = {
                 record1: { id: 'record1', prop: 'value' },
                 record2: { id: 'record2', prop: 'value' },
                 record3: { id: 'record3', prop: 'value' },
             };
 
+            console.log('test');
```
ThieryMichel: debug spotted

packages/ra-core/src/reducer/admin/resource/data.js
djhi: I disagree, I won't have a default case

packages/ra-core/src/reducer/admin/resource/data.js
djhi: less readable and we won't deal with many items anyway.

packages/ra-core/src/reducer/admin/resource/data.js
ThieryMichel: default case represent the case where meta.fetch has an unexpected value, it should be handled and the user warned. IMO

fzaninotto:
It took me some time to understand the purpose of this PR so let me summarize my findings here.

The `undo` saga catches every `UNDOABLE` action and dispatches the inner action, modified as follows:
- the name is changed to '[original_name]_OPTIMISTIC'
- the `optimistic: true` meta is added.
- other modifications not relevant to this issue

For instance, for an undoable crudDelete action, the saga dispatches the following action:

```js
// Original action
{
    type: CRUD_DELETE,
    payload: { id, previousData },
    meta: {
        resource,
        fetch: DELETE,
        onSuccess: {
            notification: {
                body: 'ra.notification.deleted',
                level: 'info',
                messageArgs: {
                    smart_count: 1,
                },
            },
            redirectTo,
            basePath,
        },
        onFailure: {
            notification: {
                level: 'warning',
            },
        },
    },
}

// dispatched action
{
    type: CRUD_DELETE_OPTIMISTIC,
    payload: { id, previousData },
    meta: {
        optimistic: true,
        resource,
        fetch: DELETE,
        notification: {
            body: 'ra.notification.deleted',
            level: 'info',
            messageArgs: {
                smart_count: 1,
            },
        },
        redirectTo,
        basePath,
    },
}
```

This is fine.

The problem lies in the `data` reducer, which catches optimistic actions to modify the store immediately (without waiting for the server response). And it catches them by *name*, i.e. it will react to `CRUD_DELETE_OPTIMISTIC`.

This prevents usage of custom optimistic actions relying on the same fetch metas, for instance the `ACCEPT_CHANGE` action described by @djhi. Once redispatched by the `undo` saga, this action will become `ACCEPT_CHANGE_OPTIMISTIC`, and will be ignored by the `data` reducer. This is a pity, since this action is basically a `CRUD_UPDATE` with a special name.

So the proposal is to change the `data` reducer to react not on the optimistic action *names* anymore, but on the optimistic *meta* instead. that way, the `data` reducer would react to the `ACCEPT_CHANGE_OPTIMISTIC` action, not because of its name, but because it's optimistic.

I'm :+1: for that change.

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: `meta.fetch` can be anything the user wants, provided that they deal with it in their `dataProvider`. This is not REST, but really a dictionary of verbs. We shouldn't limit what the user can do, or send a warning when an unknown `meta.fetch` is used.

fzaninotto: I don't understand this one.

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: I think the `reduce` has another purpose: avoid updating non-existing ids.

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: because `newState` is an array. I know, it's counter-intuitive. The previous code should probably have included a comment for that.

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: 👍 with @djhi here

packages/ra-core/src/reducer/admin/resource/data.js
fzaninotto: same, `newState` is an array.

packages/ra-core/src/reducer/admin/resource/data.js
ThieryMichel: ok

packages/ra-core/src/reducer/admin/resource/data.js
djhi: I'll add a comment on this one

packages/ra-core/src/reducer/admin/resource/data.js
ThieryMichel: I mean extract the logic to get the updatedResults in a separate function.
This way you can call addRecords a single time at the end.

djhi: Not all cases require to call `addRecords`, I think it is more readable this way

packages/ra-core/src/reducer/admin/resource/data.js
ThieryMichel: Even so, it makes no difference. We create a new array any way from the ids in the payload.
And we will always merge the previousState with the data from the payloads, so we might as well do it in a single map.

djhi: agreed

PR #2686, opened 2018-12-20
Fix EditButton in Datagrids using rowClick

plattdl:
Fixes a similar issue to #2457 where clicking on an Edit Button inside of a Datagrid with rowClick set to "show" doesn't handle the button event properly.


fzaninotto:
Good catch! I see the `ShowButton` still has the same issue, would you mind opening another PR for that one?

plattdl:
@fzaninotto Done!

PR #2704, opened 2018-12-31
Add 'redirect' prop to 'DeleteButton' in 'Toolbar'

afilp:
I could not redirect to a custom list page after pressing Delete in the default toolbar.

This fixes it.

### DIFF
--- a/packages/ra-ui-materialui/src/form/Toolbar.js
+++ b/packages/ra-ui-materialui/src/form/Toolbar.js
@@ -79,6 +79,7 @@ const Toolbar = ({
                                 basePath={basePath}
                                 record={record}
                                 resource={resource}
+                                redirect={redirect}
                             />
                         )}
                 </div>


fzaninotto:
Hi @afilp,

And thanks for the PR.

The `redirect` prop is documented as affecting the *submission*, not the *deletion*. This addition could bring unexpected changes. Also, some developers may want a different redirection for Edit and for Delete. 

Besides, having a custom Delete redirection is already possible without changing the react-admin code, by dispatching a custom DELETE action with a `redirect` meta.

As this is not a bulletproof change, and it's not strictly required, we won't merge it to the core.

Thanks for your contribution anyway!

afilp:
@fzaninotto Thank you for your time devoted to explain. I wish you and your team all the best for 2019!

Just a short explanation of my `use case` here: 

I have a resource `movements` (that has a CRUD interface). Then, there is another resource called `movementAssets`. Each `movement` has many `movementAssets`, so a CRUD interface also exists for them BUT each `movementAsset` need to always have the `movementId` as a field, as a `foreign key` to the `movement` where it belongs.

The only way I could imagine doing this was to pass the `movementId` to the CRUD interface (List, etc.) of the `movementAssets`.  Example: `http://localhost:4000/#/movementAssets?movementId=5c26a16af82aea0798ed4ef4&order=DESC&page=1&perPage=25`

(I could not depend on the REDUX store for this, because REFRESH causes lots of problems, we do need a query param to keep state on REFRESH OR use localStorage everywhere)

So, from now on, every action on the `movementAssets` would need to keep the `movementId` parameter on the URL. This works well with the `redirect` for `CREATE` and `EDIT`, because I use the `redirect function` and the `movementId` is a hidden field of the `data`, so I take it from the `data` and use it for the redirect path.

Trouble is ONLY with the `DELETE`, because there is no option to `redirect` the page to the `List` with my parameter (`movementId`) added.

I hope you see the trouble we may have for using a "related CRUD, based on a particular foreign key ID". Maybe I miss something that can make this easier.

PR #2706, opened 2018-12-31
[RFR] Display a skeleton instead of empty datagrid when loading

fzaninotto:
This PR slightly improves the User Experience by showing a skeleton datagrid during the first load instead of nothing. On slow networks, the previous user experience could lead the users to thing there is a bug.

kapture 2018-12-31 at 16 02 54

Note that this only affects the *first* load. On subsequent renderings of the datagrid, react-admin still displays stale data while fetching up-to-date data (optimistic rendering).

The skeleton tries to minimize screen fickering by displaying checkboxes only when bulk actions are on, by showing as many columns as the datagrid has children, etc.

Incidentally, this also solves a UI problem that I used to think of "the lone record". When a users opens the app directly in a record detail page and goes back to the list, the list used to display only that record while loading the first page. Now it displays the skeleton instead.

kapture 2018-12-31 at 16 14 26

Tested on `Datagrid` directly descending a `List`, and descending a `ReferenceManyField`.

kapture 2018-12-31 at 16 17 35

Note that if the dataProvider fails fetching the list the first time, the skeleton will remain on screen. That's assumed. 

Adds a new entry in the resource state to avoid changing the initial value of the list data or ids.

image

When React Suspense is a thing, we'll add a time threshold to display the skeleton only after a short delay.


packages/ra-core/src/controller/field/ReferenceManyFieldController.js
```diff
@@ -135,7 +135,7 @@ export class ReferenceManyFieldController extends Component {
             currentSort: this.state.sort,
             data,
             ids,
-            isLoading: typeof ids === 'undefined',
+            loadedOnce: typeof ids !== 'undefined',
```
fzaninotto: this can be considered a BC break, but it was never documented, so I'm confident it should not break much.

PR #2711, opened 2019-01-01
GraphQL-simple: Update building metaArgs

tim-stasse:
Only the filter variable is required by the graphql server, to calculate the total count. Additional variables add no value, however if left out of the schema, an error is produced as the client attempts to pass variables for `page` and `perPage`.


djhi:
Is this still needed @tim-stasse? Can you first describe how to reproduce the issue?

fzaninotto:

Issue #2712, opened 2019-01-01
cypress: cy.get(...).get(...) doesn't work


JacquesBonet:
In List.js file of cypress project we have 5 instructions with multiple get() like this one:

    cy.get(ListPagePosts.elements.sortBy('"tags.name"')).get("svg").should('be.visible');

This instruction is equivalent to:

    cy.get("svg").should('be.visible');

So we check the visibility of all svg dom element in the page and not of the svg element with a parent using class "tags.name".

PR #2713, opened 2019-01-01
Correct consecutive get()  in cypress instruction #2712

JacquesBonet:
- Create a new svg access function permitting to acces the right element.
- correct another problem about headroom css class. This class doesn't exist with a production build using react-script, so use alternative css class.

### DIFF
--- a/cypress/package.json
+++ b/cypress/package.json
@@ -3,6 +3,7 @@
     "name": "e2e",
     "version": "0.1.0",
     "scripts": {
+        "serve": "node ./infinitStart.js",
         "start": "./node_modules/.bin/cypress open",
         "test": "node ./start.js"
     },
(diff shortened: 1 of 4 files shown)


cypress/package.json
```diff
@@ -3,6 +3,7 @@
     "name": "e2e",
     "version": "0.1.0",
     "scripts": {
+        "serve": "node ./infinitStart.js",
```
djhi: Why is this needed ? Looks irrelevant to this PR

djhi: Are you aware of the npm commands inside the root `package.json`. If you can't use make, you can:

launch a terminal and start `yarn run-simple`

launch another terminal and start `yarn test-e2e-local`

Can you remove the new ones?

JacquesBonet: sure

Issue #2693, opened 2018-12-23
graph cool demo error


sorgnardt:
when launching the graphcool-demo, it get the following error

instanceOf.mjs:33 Uncaught ReferenceError: require is not defined
    at Module.../../packages/ra-data-graphql/node_modules/graphql/jsutils/instanceOf.mjs (instanceOf.mjs:33)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/node_modules/graphql/type/definition.mjs (definition.mjs:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/node_modules/graphql/type/validate.mjs (validate.mjs:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/node_modules/graphql/graphql.mjs (graphql.mjs:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/node_modules/graphql/index.mjs (index.mjs:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/esm/introspection.js (introspection.js:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphql/esm/index.js (index.js:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module.../../packages/ra-data-graphcool/esm/index.js (index.js:1)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module../src/dataProvider.js (index.js:3)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module../src/App.js (App.css?8166:45)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Module../src/index.js (index.css?b42f:45)
    at __webpack_require__ (bootstrap:787)
    at fn (bootstrap:150)
    at Object.0 (index.js:190)
    at __webpack_require__ (bootstrap:787)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.chunk.js:1

jabolina:
[...]

PR #2715, opened 2019-01-02
Build new

JacquesBonet:
Build process change:
--> simplification of the package.json files:
   - use only devDependencies for all packages module
   - no more esm module
--> use typescript build process except for cypress and simple project
--> permits debug with original source file
--> build is faster

correct issue #2693
correct some duplicated attributes declaration
add .idea on .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 npm-debug.log
 yarn-error.log
 lerna-debug.log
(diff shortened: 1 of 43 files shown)


fzaninotto:
No, thanks. esm module is necessary to allow tree-shaking.

JacquesBonet:
Yes esm module is necessary to allow tree-shaking, but is tree-shaking really useful?

I made a test on building (yarn build) projects. I obtain the following size for the packed js files

Demo: without the pull request: 2522kb, with pull request: 2443kb

Tutorial: without the pull request: 1032kb, with pull request: 945kb

==> tree-shaking don't seem useful

PR #2720, opened 2019-01-03
Should render suggestions - AutocompleteArrayInput

davidpicarra:
Following #2502, this PR allows for the creator of an `<AutocompleteArrayInput />` to pass in an optional `shouldRenderSuggestions` function.

Was trying to set this function for the `<AutocompleteArrayInput />` and noticed that #2502 only adds this functionality to `<AutocompleteInput />`.


fzaninotto:
Could you please PR against `next` instead of `master`, as it's a new feature?

packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -449,7 +449,7 @@ export class AutocompleteInput extends React.Component {
         ) {
             return shouldRenderSuggestions(val);
         }
-
+
```
djhi: Can you revert this ?

davidpicarra: My bad, should be fixed now

fzaninotto:
Thanks!

Issue #2317, opened 2018-09-13
Scrollable tabs on TabbedShowLayout


MROFerreiro:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I want the possibility to scroll on `<Tabs>` of `<TabbedShowLayout>`, when using mobile devices and there are many tabs, some become out of view and is not possible to scroll horizontally to see the hidden ones.

**Additional context**
Material-Ui `<Tabs>` component receives  prop called scrollable that enables horizontal scrolling `<TabbedShowLayout>` could receive the same prop.

in19farkt:
I have this problem on the desktop.

alauper:
I have this problem on TabbedForm.   Will you be supporting the TabbedShowLayout option [here](https://github.com/marmelab/react-admin/pull/2740) with the TabbedForm component?

MROFerreiro:
Fixed in #2740 issue can be closed.

Kmaschta:
Fixed by https://github.com/marmelab/react-admin/pull/2740

PR #2740, opened 2019-01-08
[RFR] Add options prop to TabbedShowLayout

lucas2595:
After facing the same problems as #2317, this PR allows to pass props through `<TabbedShowLayout>` to material-ui's `<Tabs>`. This way the user can enable all the different Tabs modes (such as `scrollable`, `centered`), and pass other props that might be useful. This is done through the `options` prop in `<TabbedShowLayout>`.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -31,6 +31,7 @@ const getTabFullPath = (tab, index, baseUrl) =>
  *
  * Receives the current `record` from the parent `<Show>` component,
  * and passes it to its childen. Children should be Tab components.
+ * The object passed as `options` props is passed to the material-ui <Tabs> component
  *
  * @example
  *     // in src/posts.js
@@ -77,6 +78,7 @@ export class TabbedShowLayout extends Component {
             translate,
             version,
             value,
+            options,
             ...rest
         } = this.props;
 
@@ -91,6 +93,7 @@ export class TabbedShowLayout extends Component {
                     // so we can use it as a way to determine the current tab
                     value={location.pathname}
                     indicatorColor="primary"
+                    {...options}
                 >
                     {Children.map(children, (tab, index) => {
                         if (!tab) return null;
@@ -144,6 +147,7 @@ TabbedShowLayout.propTypes = {
     value: PropTypes.number,
     version: PropTypes.number,
     translate: PropTypes.func,
+    options: PropTypes.object,
 };
 
 const enhance = compose(
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -7,6 +7,7 @@ import compose from 'recompose/compose';
 import { translate } from 'ra-core';
 
 import CardContentInner from '../layout/CardContentInner';
+import TabbedShowLayoutTabs from './TabbedShowLayoutTabs';
 
 const sanitizeRestProps = ({
     children,
@@ -18,6 +19,7 @@ const sanitizeRestProps = ({
     initialValues,
     staticContext,
     translate,
+    tabs,
     ...rest
 }) => rest;
 
@@ -31,7 +33,7 @@ const getTabFullPath = (tab, index, baseUrl) =>
  *
  * Receives the current `record` from the parent `<Show>` component,
  * and passes it to its childen. Children should be Tab components.
- * The object passed as `options` props is passed to the material-ui <Tabs> component
+ * The component passed as `tabs` props replaces the default material-ui's <Tabs> component.
  *
  * @example
  *     // in src/posts.js
@@ -78,7 +80,7 @@ export class TabbedShowLayout extends Component {
             translate,
             version,
             value,
-            options,
+            tabs,
             ...rest
         } = this.props;
 
@@ -88,14 +90,15 @@ export class TabbedShowLayout extends Component {
                 key={version}
                 {...sanitizeRestProps(rest)}
             >
-                <Tabs
-                    // The location pathname will contain the page path including the current tab path
-                    // so we can use it as a way to determine the current tab
-                    value={location.pathname}
-                    indicatorColor="primary"
-                    {...options}
-                >
-                    {Children.map(children, (tab, index) => {
+                {React.cloneElement(
+                    tabs,
+                    {
+                      // The location pathname will contain the page path including the current tab path
+                      // so we can use it as a way to determine the current tab
+                      value: location.pathname,
+                      match
+                    },
+                    Children.map(children, (tab, index) => {
                         if (!tab) return null;
 
                         // Builds the full tab tab which is the concatenation of the last matched route in the
@@ -108,8 +111,8 @@ export class TabbedShowLayout extends Component {
                             context: 'header',
                             value: tabPath,
                         });
-                    })}
-                </Tabs>
+                    })
+                )}
                 <Divider />
                 <CardContentInner>
                     {Children.map(
@@ -147,7 +150,11 @@ TabbedShowLayout.propTypes = {
     value: PropTypes.number,
     version: PropTypes.number,
     translate: PropTypes.func,
-    options: PropTypes.object,
+  tabs: PropTypes.element.required
+};
+
+TabbedShowLayout.defaultProps = {
+    tabs: <TabbedShowLayoutTabs />
 };
 
 const enhance = compose(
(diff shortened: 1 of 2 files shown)


lucas2595:
That indeed sounds better. Just made another commit to it implementing that.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -86,13 +90,15 @@ export class TabbedShowLayout extends Component {
...
-                    {Children.map(children, (tab, index) => {
+                {React.cloneElement(
+                    tabs,
+                    {
+                      // The location pathname will contain the page path including the current tab path
+                      // so we can use it as a way to determine the current tab
+                      value: location.pathname,
+                      match
+                    },
+                    Children.map(children, (tab, index) => {
```
fzaninotto: I'd put that into the TabbedShowLayoutTabs component. In my opinion, TabbedShowLayout should not mess with its grand-children (the children of its child Tabs).

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component, Children, cloneElement } from "react";
+import PropTypes from "prop-types";
+import Tabs from "@material-ui/core/Tabs";
+import Divider from "@material-ui/core/Divider";
+import { withRouter, Route } from "react-router-dom";
```
fzaninotto: I suspect that some of these are unused, please check the linter.

fzaninotto:
Please make this PR against `next` instead of `master` as it's a new feature.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
lucas2595: Should I pass `getTabFullPath` as a prop to TabbedShowLayoutTabs well, then? Or do I replicate it inside TabbedShowLayouTabs?

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
lucas2595: When I run `make linter` it's only returning some `console.log` calls in other files (ListGuesser for example). When running `make prettier`, it changes most of the projects files with a different style. Did I set up something wrong?

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
fzaninotto: As `getTabFullPath` will only be used in the child, it should be defined there.

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
fzaninotto: We currently have a problem with the `make prettier` command, so you can ignore it. I'm surprised that `make lint` doesn't show the unused imports. You'll have to check those imports one by one.

lucas2595:
Sorry about the messed up history, first time forking and making a PR on github here. All those force-pushes were to fix a rebase gone wrong. Should I redo it from scratch to clean things up?

fzaninotto:
I can squash and merge, don't worry about that.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
lucas2595: Done in https://github.com/marmelab/react-admin/pull/2740/commits/c2dd91d80f392fe0480745a0b2f9218d42f0fb26

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
lucas2595: Done in https://github.com/marmelab/react-admin/pull/2740/commits/c2dd91d80f392fe0480745a0b2f9218d42f0fb26

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -77,36 +79,22 @@ export class TabbedShowLayout extends Component {
...
+            <div className={className} key={version} {...sanitizeRestProps(rest)}>
+                {cloneElement(
+                    tabs,
+                    {
+                        // The location pathname will contain the page path including the current tab path
+                        // so we can use it as a way to determine the current tab
+                        value: location.pathname,
+                        match,
+                    },
+                    [...children],
```
fzaninotto: Why not put `children` directly here, instead of `[...children]`?

lucas2595: Sorry, followed [this](https://reactjs.org/docs/react-api.html#cloneelement) blindly.

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
```diff
@@ -0,0 +1,34 @@
+import React, { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import Tabs from '@material-ui/core/Tabs';
+
+const getTabFullPath = (tab, index, baseUrl) =>
+    `${baseUrl}${
+      tab.props.path ? `/${tab.props.path}` : index > 0 ? `/${index}` : ''
+    }`;
+
+const TabbedShowLayoutTabs = ({ value, children, match, ...rest }) => (
```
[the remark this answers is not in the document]
fzaninotto: I think it'd be better to pass the `rest` props to `Tab`, that way it's possible to add `Tab` props just by decorating `TabbedShowLayoutTabs`

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.js
```diff
@@ -21,6 +21,7 @@ const TabbedShowLayoutTabs = ({ value, children, match }) => (
             return cloneElement(tab, {
                 context: 'header',
                 value: tabPath,
+                ...rest,
```
fzaninotto: Sorry, you misunderstood my comment, I wanted the rest props to be passed to `Tabs`, not to its children.

lucas2595: oh, yeah, you said "`Tab`" and that got me confused haha. Well... should I keep on explicitly  passing `value={value}` in `Tabs` as well, or do I just remove it since it'll be in `rest`?

fzaninotto:
Terrific, thanks a lot!

MROFerreiro:
Is this really working? I tried to use an options object with props defined in [material-ui tabs component](https://v1-0-0.material-ui.com/api/tabs/) in `<TabbedShowLayout>` like: `<TabbedShowLayout options={{scrollable=true}} />` but it doesn't work. How is this supposed to be used? 
For what I saw in source code, there is no options props treated by `<TabbedShowLayout>` neither in `<TabbedShowLayoutTabs>` the new tabs. Am I missing something?

lucas2595:
I'm sory @MROFerreiro, the PR title became misleading after it was decided to use a `tabs` prop in `<TabbedShowLayout>` like: `<TabbedShowLayout tabs={<TabbedShowLayoutTabs/>}>` (where `<TabbedShowLayoutTabs/>` is the old tabs component that was being used before this PR) instead of the `options` props suggestion.
For what you're trying to do, you can either use `<TabbedShowLayout tabs={<TabbedShowLayoutTabs scrollable={true}/>}>` (since the props passed to `TabbedShowLayoutTabs` are passed to the material-ui's `Tabs` inside it), or you can create your own `tabs` component, and then use it like: `<TabbedShowLayout tabs={<MyCustomTabs/>}>`.
@fzaninotto should I edit the PR's title to prevent this confusion?

MROFerreiro:
Currently is not possible to use your solution with `<TabbedShowLayoutTabs>` since it is not export in [index of details](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/detail/index.js) simply adding `export {default as TabbedShowLayoutTabs} from './TabbedShowLayoutTabs'` would fix it, currently I copied your implementation to a custom component and used it. Seems fine, but I need to do some corrections on my side in order for it to work properly.
That said, @fzaninotto, since is a simple hotfix I don't think is needed to do another merge request.

fzaninotto:
Please open a new PR to export and document `TabbedShowLayoutTabs`

MROFerreiro:
@lucas2595 can you do it? Since you were the one that build this feature you would produce better docs than I and faster. Don't forget the export of the new Tabs component. 😃

lucas2595:
Sure thing, @MROFerreiro. Sorry for the delay, got caught up with with some other stuff in the past days. I'll do it right now.

PR #2638, opened 2018-12-10
Add the option to not show the 'Export Button'

afilp:
Related comment: https://github.com/marmelab/react-admin/issues/2483#issuecomment-435852774

PR #2746, opened 2019-01-09
Fix propType warning when using exporter false in List

afilp:
Related to https://github.com/marmelab/react-admin/pull/2638#issuecomment-452768564


fzaninotto:
Thanks!

Issue #2750, opened 2019-01-10
CheckboxGroupInput is not aligned with other Inputs in List Filters


Luwangel:
**What you were expecting:**

I use a `CheckboxGroupInput` in `Filters`, but this one is not aligned with the others inputs. I want the label and the checkboxes to be aligned as you can see below.

![selection_001](https://user-images.githubusercontent.com/5584839/50965321-f0ab7680-14d1-11e9-9bc9-0acd8fafd4ac.png)

**What happened instead:**

According to this screenshot, we can see that the labels are not aligned.

![selection_002](https://user-images.githubusercontent.com/5584839/50965711-108f6a00-14d3-11e9-841b-1a52eb12e2ad.png)

**Steps to reproduce:**

- Create a `CheckboxGroupInput` filter
- Select the filter in the list and compare it to others

``` jsx
const PostFilter = props => (
    <Filter {...props}>
        <TextInput
            source="title"
            defaultValue="Qui tempore rerum et voluptates"
        />
        <CheckboxGroupInput
            label="Languages"
            choices={[
                { id: 'english', name: 'English' },
                { id: 'french', name: 'French' },
            ]}
            {...props}
        />
    </Filter>
);
```

**Related code:**

It's a bug related to how the checkboxes are designed in Material UI. In facts, they use a `IconButton` which has a default size which is bigger than the size expected to align our filters.

**Environment**

* React-admin version: 2.6

PR #2751, opened 2019-01-10
[RFR] Fix CheckboxGroupInput design

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/2750

## Todo

In `CheckboxGroupInput.js` file

- [x] Change the label style which is not the same as Material UI Input one
- [x] Pass a new style to the `Checkbox` component which forces the height to 32px

selection_003


fzaninotto:
Great!

PR #2753, opened 2019-01-10
Fix ReferenceManyField doesn't rerender when the filter props changes

kandebr:
Component ReferenceManyField takes filter prop but now it doesn't call fetching new data if filter changed. This PR can fix it.


fzaninotto:
For a mysterious reason, the CI decided not to run on your previous commit. Would you mind re-pushing to trigger a build?

fzaninotto:
Thanks!

Issue #1683, opened 2018-03-21
[v3] improve peerDependency


kopax:
singletons should be a peer; `react-redux` thus probably should be

singletons (something that shares mutable state, and needs to have only one copy in the dependency graph)

Also `redux` is a way to store state. it is *definitely* a singleton.

If your lib has `react-redux` v3.0 as a dep, and someone else has v3.1, you'll have two copies in memory and things will break.

`react-redux` should be a peer dep, so should `react-router`

Also, it should be using ^ not ~ throughout, because ~ means "just patch" and ^ means "patch or minor"

You want semver ranges to be as wide as possible to allow for as much deduping and upgrading as possible.

fzaninotto:
If someone installs another version of Redux, it’s they responsibility. If they want to share the same store, they must use the same as admin-on-rest. 

As for using ^ instead of ~, I disagree. Using ^ guarantees that the application will break easily, because lib authors don’t use delver properly. Believe me, we’ve followed this path before.

ljharb:
@fzaninotto fwiw you're highly wrong about `^` - lib authors who don't use semver properly could publish a breaking change in a patch, too, which ~ allows. The *only* way to protect yourself from this kind of thing is with a lockfile, which is an app-level concern - published packages should use as wide a range as possible. Using `~`, or even using `=`, will also not protect against transitive dependencies from updating.

kopax:
@fzaninotto I agree with ljharb. About peerDeps, I can only ask now why do you want such restriction ?

The customApp example let us think the opposite.

fzaninotto:
Peer dependencies make the install process harder and more error prone. We want the setup to be easy for newcomers - sometimes at the expense of expert users like the ones who want to integrate admin-on-rest in another app.

ljharb:
Not if you make it both a peer and a production dep - users don’t ave to do anything manually, but are prevented from having conflicts in their dep graph.

We can keep debating as long as you like, but you’re wrong here on every metric.

kopax:
That seems like a good and hybrid solution. Can we reopen this?

fzaninotto:
@ljharb I'm often wrong on many metrics, but it' the first time I happen to be wrong on every metric. Thanks for pointing out. By the way, which metric are you referring to? 

But since I've got an expert in the field of dependencies, let me ask a few questions, because many things remain unclear to me in npm land.

[...]

Issue #1999, opened 2018-07-05
Deprecation of react-router-redux


vascofg:
React-router-redux is deprecated as can be seen here:
https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux

Any plans on finding an alternative?

djhi:
Not yet but I guess we'll have to use https://github.com/supasate/connected-react-router

ignlg:
I've been using https://github.com/FormidableLabs/redux-little-router with some projects, just to point at another option

azjgard:
@ignlg Thanks a ton for passing that along!

fenghuizhang:
> com
Can we use react-router's {withRouter} replace react-router-redux's {routerReducer} ?

fenghuizhang:
> Not yet but I guess we'll have to use https://github.com/supasate/connected-react-router

Can we use react-router's {withRouter} replace react-router-redux's {routerReducer} ?

maicWorkGithub:
> > Not yet but I guess we'll have to use https://github.com/supasate/connected-react-router
> 
> Can we use react-router's {withRouter} replace react-router-redux's {routerReducer} ?

https://reacttraining.com/react-router/core/guides/redux-integration/deep-integration

They say not suggested use router in redux, I'm a noob, why you choose use router in redux? what's the benefit?

edy:
What about [React Navigation](https://reactnavigation.org)? They support [Web and SSR](https://reactnavigation.org/docs/en/web-support.html) too.

fzaninotto:
After re-reading https://reacttraining.com/react-router/core/guides/redux-integration/deep-integration, the proposed solution (do not integrate react-router with redux) seems cumbersome and a very big BC break, because each action that *may* lead to a redirection must have the `history` object in its payload.

So we'll go with [connected-react-router](https://github.com/supasate/connected-react-router), but they haven't integrated react-redux v7 yet (see https://github.com/supasate/connected-react-router/issues/287).

fzaninotto:
Fixed by #3170

PR #2760, opened 2019-01-11
fix(dependencies): moved redux, react-redux and react-router-redux to peerDependencies.

kopax:
HI @fzaninotto. Sorry for the delay, we have not received a notification when you reopened #1683.

I have seen that you already moved part of the `dependencies` into `peerDependencies`.

**redux** should also be in `peerDependencies`, so does **react-redux**. 

I am not sure for `react-router-redux`, does this package use a store reference or does it just export a reducer to be used somewhere? 

By doubt and because it would solve all problems, I also added it as a `peerDependencies`.


Kmaschta:
Isn't this change a breaking one?

kopax:
@Kmaschta you are absolutely right, I forgot to mention it, you should add a not for people.

Moving a `dependencies` into `peerDependences` **and** `dependencies` will change nothing in development (you must have it at some point). In production, it will require the user to have our list of `peerDependencies` installed in it's `dependencies` or `devDependencies`.

If the user does not have `redux` in it's `devDependencies` or `dependencies`, then it will break.

If the user have `redux` or at least two sub depencencies that depend on `redux`, then it will work.

fzaninotto:
Thanks! 

This really is a breaking change, so we won't merge it before 3.0. And we're waiting for react 16.8 and material-ui 4.0 to break BC for good, so this PR won't be merged before a few months at least.

kopax:
Sure. About `react-router-redux`, the version `5.0.0-alpha.9` was never released. Instead they renamed the package `connected-router`.

I can't find the source code of `5.0.0-alpha.9`, but https://github.com/supasate/connected-react-router/blob/master/src/reducer.js show they are doing the reducer in here. It should also be a `peerDependencies`. 

Could we, at the same time. replace `react-router-redux@5.0.0-alpha.9` with `connected-router` (work the same way, just the import and the dependency change). 

At the same time, we could also update `redux`, `react-redux`, `react-router-dom`, `react-router`.

I've already played with all these versions and none of theme are breaking changes.

fzaninotto:
see #1999 for react-router-redux

PR #2780, opened 2019-01-17
Add an optional emptyValue to SelectInput

edy:
#2234 introduced a breaking change for our application. This change adds an optional custom value which is used when allowEmpty is set.

Previously `null` was set as an empty value, now you can set null again.

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectInput.js
+++ b/packages/ra-ui-materialui/src/input/SelectInput.js
@@ -10,6 +10,7 @@ import ResettableTextField from './ResettableTextField';
 const sanitizeRestProps = ({
     addLabel,
     allowEmpty,
+    emptyValue,
     basePath,
     choices,
     className,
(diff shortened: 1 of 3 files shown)


sghribi:
Thanks @edy for this PR!

Is it ok to be merged for 2.7.1?

edy:
sure

packages/ra-ui-materialui/src/input/SelectInput.js
```diff
@@ -155,7 +156,10 @@ export class SelectInput extends Component {
 
     addAllowEmpty = choices => {
         if (this.props.allowEmpty) {
-            return [<MenuItem value="" key="null" />, ...choices];
+            const value = this.props.hasOwnProperty('emptyValue')
+                ? this.props.emptyValue
```
djhi: Using `defaultProps` might be better than this manual check

edy: i agree. will change that later this day

edy:
> Is it ok to be merged for 2.7.1?

would you merge this for 2.7.2?

sghribi:
Hi @djhi @fzaninotto 

Could you review/validate this PR?

Thanks!

djhi:
Sorry @edy, as it is a new prop, can you please target the `next` branch?

edy:
Done.

Why isn't it possible to merge into master? All tests pass and the change is backward compatible too.

djhi:
As I said in previous comment, this adds a new prop, hence a new feature. If we do merge it, it will be on the `next` branch for the `2.8` release (semver ?)

edy:
also done

fzaninotto:
Thanks!

PR #2782, opened 2019-01-17
[RFR] Check minimum required props are passed to controllers

Kmaschta:
When we forgot to pass the props from the parent to the controller, the app displays a huge error.

```js
// For example:
const PostShow = () => (
    <Show title={<PostTitle />} />
);

// Instead of:
const PostShow = props => (
    <Show title={<PostTitle />} {...props} />
);
```

## Current Error Message:

image

## New Error Message

image


Kmaschta:
Last minute change: `throw Error` instead of a `console.error(); return null;` since the default Layout catch rendering errors. (screenshots updated)

@djhi You're right, that's why I wrote a HOC :)

PR #2796, opened 2019-01-21
New data provider ra-strapi-rest added to DataProviders.md

nazirov91:


Kmaschta:
Hey! Thanks for the contribution, your data provider seems great!

I see that you have copy/pasted the code of `ra-data-simple-rest` in order to write your own provider.
May I suggest you to take `ra-data-simple-rest` and wrap it in order to adjust it for Strapi?

See the documentation about that: https://marmelab.com/react-admin/DataProviders.html#decorating-your-data-provider-example-of-file-upload

Kmaschta:
You can update it before to call the simple rest provider, with something like:

```js
const customProvider = (apiUrl) => (type, resource, params) => {
    const modifiedUrl = `${apiUrl}/prod`;

    return simpleRestDataProvider(modifiedUrl)(type, resource, params)
        .then((result) => /* update result */);
};
```

PR #2799, opened 2019-01-22
[RFR] Demo improvements

fzaninotto:
- [x] Add a two-level menu
- [x] Add a loader to the posters list
- [x] Generate invoices in the data generator
- [x] Add an Invoice list with expand
- [x] Display the Review detail on the side of the list
- [x] Use a normal Edit view on Mobile

localhost_3000_ 3

reviews


examples/demo/src/invoices/InvoiceList.js
```diff
@@ -0,0 +1,51 @@
...
+const ListFilters = props => (
+    <Filter {...props}>
+        <DateInput source="date_gte" alwaysOn />
+        <DateInput source="date_lte" alwaysOn />
+    </Filter>
+);
+
+const InvoiceList = props => (
+    <List {...props} filters={<ListFilters />} perPage={25}>
+        <Datagrid expand={<ShowGuesser />}>
```
djhi: Using the ShowGuesser does not allow you to override the `title` prop. It can be seen in the last screenshot: there are two titles concatenated (list and show) when expanding an line

fzaninotto: Fixed in the next commit. I only used the guesser temporarily.

PR #2806, opened 2019-01-24
[RFR] Display selected choices in their selection order in SelectArrayInput

Luwangel:
The behavior of the `SelectArrayInput` doesn't follow the way they are stored. I've changed the render method in the `SelectArrayInput` component in order to fix this.

In addition, this change respects the [Multiple Select demo](https://material-ui.com/demos/selects/#multiple-select) in Material-UI.

## Scenario

In the following screenshot, we can find *on the left the new version**, and **on the right the old version**.

Selection order:
- (1) Compulsive
- (2) Reviewer
- (3) Regular

Choices order:
- **(1) Compulsive**
- Collector
- Ordered once
- **(3) Regular**
- Returns
- **(2) Reviewer**

We notice that without this PR, the `SelectArrayInput` displays the items in the choices order instead of the selection order. However, in the storage, the items are ordered following the user selection order.

selection_001

## Todo

- [x] Change the `renderValue` method in `SelectArrayInput` to display selected choices in their selection order

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.js
@@ -141,6 +141,7 @@ export class SelectArrayInput extends Component {
         this.props.input.onBlur(event.target.value);
         this.setState({ value: event.target.value });
     };
+
     renderMenuItemOption = choice => {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Excellent, thanks!

PR #2816, opened 2019-01-28
[RFR] Use default colors for most buttons

djhi:
We currently set the color prop of all buttons to `primary` which defeat the purpose of having a primary button.

This PR sets the default color of all buttons to `default`. However, it keeps the `primary` color for some cases when it makes sense such as the `CreateButton` in fab mode (on mobile devices) or the `SaveButton` used in forms. It can still be overriden by supplying the `color` prop.

Same logic is applied to the `DeleteButton` which should not stand out by default (it is currently always red). However, a new `danger` value is supported by its `color` prop in addition of the material-ui values for when it is needed (such as in a confirmation modal)

Before: before
After: after

Before: before
After: after

### DIFF
--- a/examples/simple/src/tags/TagList.js
+++ b/examples/simple/src/tags/TagList.js
@@ -23,7 +23,7 @@ const TagDragPreview = props => (
 
 const CustomNodeActions = props => (
     <NodeActions {...props}>
-        <SaveButton variant="flat" />
+        <SaveButton color="default" variant="flat" />
         <IgnoreFormProps>
             <EditButton />
             <ShowButton />
--- a/packages/ra-ui-materialui/src/button/Button.js
+++ b/packages/ra-ui-materialui/src/button/Button.js
@@ -37,7 +37,7 @@ const Button = ({
     children,
     classes = {},
     className,
-    color = 'primary',
+    color = 'default',
     disabled,
     label,
     size = 'small',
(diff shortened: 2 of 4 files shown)


djhi:
> But now there is no more primary button in the List view... I'm not convinced.

Because there is no need to have one here ?

fzaninotto:
Material Design emphasizes using a different color for buttons in its new version. I see no example of button using the same color as the text.

https://material.io/design/components/buttons.html#

The difference in hierarchy (primary / not primary) is noted by variants of the button style (outlined / contained), not by color.

djhi:
Do we agree about the delete button? I can make 2 separate PR

fzaninotto:
hmm... I don't like it either. It's a dangerous button, it should have a danger color.

djhi:
Can we at least add the `danger` value for the `DeleteButton` `color` prop, which allows to easily change its styles. I'll make `danger` the default

djhi:
See the `DeleteButton` in this PR but I'll change the default value of the color prop to `danger`

fzaninotto:
I think that setting a color not supported by material-ui is confusing, because developers expect that prop to be passed as is to the mui button, and the mui button doesn't support the color danger. I'd prefer a `danger` prop set to `true` by default, which triggers the custom style, rather than a custom color.

PR #2817, opened 2019-01-28
[RFR] Fix error when using Edit component as expand in list

### LINKED ISSUE #2804: Edit component inside datagrid expansion explicitly requires location property
When placed an `Edit` component inside Datagrid's `expand` as per the example

```
const MyExpansion = props => (
    <Edit {...props} title=" " >
...
```

```
<Datagrid rowStyle={rowStyle} expand={<MyExpansion/>}>
...
```
I get the following error:

```
Uncaught Error: <Edit> component is not properly configured, some essential props are missing.
Be sure to pass the props from the parent. Example:

const MyEdit = props => (
    <Edit {...props}></Edit>
);

The missing props are: location
    at checkMinimumRequiredProps.js:28
```

This goes away if I add `location` prop as follows:
```
<Edit {...props} title=" "  location={props.location}>
```

The documentation does not mention this, so either this is a bug or docs should be updated (about having to pass location object explicitly).

**Environment**

* React-admin version: ^2.6.2
* Last version that did not exhibit the issue (if applicable): ^2.6.0
* React version: ^16.7.0
* Browser: Chrome

fzaninotto:
`location` isn't required at all.

Closes #2804

### DIFF
--- a/packages/ra-core/src/controller/EditController.js
+++ b/packages/ra-core/src/controller/EditController.js
@@ -176,11 +176,7 @@ function mapStateToProps(state, props) {
 }
 
 export default compose(
-    checkMinimumRequiredProps('Edit', [
-        'basePath',
-        'location',
-        'resource',
-    ]),
+    checkMinimumRequiredProps('Edit', ['basePath', 'resource']),
     connect(
         mapStateToProps,
         {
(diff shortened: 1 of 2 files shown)


fzaninotto:
good catch, I missed those.

PR #2823, opened 2019-01-28
trigger refresh after crudUpdate. this is to fix a similar issue to #2425…

normanzb:
I have a custom edit view allowing user to do in-place editing.
Once field editing is done, the user stay on the same edit view so they can work on the other fields.
When undoable={false} is set on the edit view, the edit view does not get refreshed after editing.
But when undoable={true} is set, the edit view does get refreshed.

### DIFF
--- a/packages/ra-core/src/actions/dataActions.js
+++ b/packages/ra-core/src/actions/dataActions.js
@@ -119,7 +119,8 @@ export const crudUpdate = (
     data,
     previousData,
     basePath,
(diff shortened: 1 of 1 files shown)


djhi:
Sorry about that, we are migrating to TypeScript. Would mind rebasing ?

natrim:
got suprised by this one too, created #3506 to close this, as it seems no response for half year

fzaninotto:
superseded by #3506

PR #2825, opened 2019-01-29
[RFR] Fix undefined issue in asyncValidation for SelectInput

erkimiilberg:
Property with undefined value was passed to asyncValidation when onBlur event occured


erkimiilberg:
**Steps to reproduce:**
* Have an asyncValidated form with `<SelectInput source="a" choices={[ id: 'b', name: 'Only choice' ]} />`
* Change value in the input a
* Debug data (first argument) passed to asyncValidate method

When validation is called on input change, then data equals `{ a: 'b' }`
When validation is called on input blur, then data equals `{ a: undefined }`

This bug is created by https://github.com/marmelab/react-admin/pull/2203
If you look closely, there is value passed for onBlur in `SelectArrayInput` but not for `SelectInput`

fzaninotto:
Thanks, then!

PR #2831, opened 2019-01-30
[RFR] Fix default filter values cannot be removed by user

### LINKED ISSUE #2472: Cleared filterDefaultValues field is reset to default when another filter is added
**What you were expecting:**
`filterDefaultValues` should only set the value of a `Filter` field on the first page load, not when another filter is populated

**What happened instead:**
If a `Filter` field with an associated `filterDefaultValues` is cleared, then another filter is populated, the first filter will be reset with it's default value.

**Steps to reproduce:**
Go the to posts page of the below codesandbox, clear the `title` filter, enter any text in the `search` box, notice that the title filter has been reset to it's `filterDefaultValues` value

**Related code:**
https://codesandbox.io/s/6vrvnn0kzr - react-admin v2.4
https://codesandbox.io/s/k5y489x333 - react-admin v2.3

**Environment**

* React-admin version: 2.4 & 2.3
* Last version that did not exhibit the issue (if applicable): Unknown
* React version: 16.3.0 and 16.3.2
* Browser: Chrome

fzaninotto:
- [x] Fix code
- [x] Add e2e test

After:

peek 30-01-2019 10-41

Closes #2472

### DIFF
--- a/cypress/integration/list.js
+++ b/cypress/integration/list.js
@@ -24,7 +24,9 @@ describe('List Page', () => {
             cy.viewport(1280, 500);
 
             cy.scrollTo(0, 200);
-            cy.get(ListPagePosts.elements.headroomUnpinned).should('not.be.visible');
+            cy.get(ListPagePosts.elements.headroomUnpinned).should(
+                'not.be.visible'
+            );
 
             cy.scrollTo(0, -100);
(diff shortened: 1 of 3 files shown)


fzaninotto:
Switching to RFR

Issue #2346, opened 2018-09-21
autocomplete is blocked (happens in demo app to)


TomJannes:
**What you were expecting:**
Being able to type in the autocompleteinput

**What happened instead:**
autocompleteinput is blocked, i can not change the value. this also happens in the demo application.

**Steps to reproduce:**
```
<ReferenceInput label="Company" source="companyId" reference="companies" filterToQuery={searchText => ({ name: searchText })}>
    <AutocompleteInput optionText="name" optionValue="id"/>
</ReferenceInput>
```

**Environment**

* React-admin version: 2.3.1
* React version: 16.5.2
* Browser: chrome
* Stack trace (in case of a JS error): no error

AkselsLedins:
2.3.2 aswell

We are experiencing this since a week approx. (even in production :crying_cat_face: )

Demo (try to select a customer):
https://marmelab.com/react-admin-demo/#/commands

![demo-bug](https://user-images.githubusercontent.com/18367406/45886073-67ecbb80-bdb8-11e8-81f9-5153fb402d0f.gif)

@valours @ArnaudD

AkselsLedins:
@kamil-borecki Thank you it helped

TomJannes:
Solved my problem, thanks!

afilp:
@fzaninotto @djhi I am also in need to change this behavior because I cannot type anything else if one item is already matched. So, I will have to Find/Replace all `<AutocompleteInput>` to add this custom function. 

It is easy to do, but since this will be a common request by others (I believe so), can we have a different behavior there? Like `inputValueMatcher = {false}`.

I would prefer to be the opposite and have this as a default but I understand that this would break previous implementations. If the `false` was the default instead, then the user would have to explicitly specify the `inputValueMatcher` either as `true` (i.e. using the default function that you already have) OR `() => {\\do different things here}`.

jheiska:
Still have this problem in 2.4.4 Is anyone working on this? It's a very annoying bug. Some other problems with the AutocompleteInput component, too. I'll list a few here. The problems are present when I have the input inside a ReferenceInput. I only use autocomplete inside reference, so not sure where the problem lies.

- When you have selected a value, basically the only way to get back to the "autocomplete mode" is to select the whole text and delete it - you cannot use backspace to remove some letters from the end. I think this is what the original issue referred to. 

- Sometimes when you are writing on the input, it randomly selects a value. 

[...]

PR #2833, opened 2019-01-30
[RFR] Fix AutocompleteInput by removing auto selection

### LINKED ISSUE #2688: AutocompleteInput auto select the option before i click from suggestion list
I'm using AutocompleteInput to pick value from categories if search the word "ag" it auto fills the  input with word "Agent" instead of showing suggestion list

**What should happen**

it should show the suggestions if i type "ag" or something other instead of pre populating the suggestion into input

djhi:
The main issue was that when used in a `ReferenceInput`, whenever the filter changed,  the `AutocompleteInput` first checked if one of the already available choices matched and if only one was found, it would not even try to fetch others from the `dataProvider`.

As we also had numerous complaints about the auto-selection feature, we decided to remove it completely.

Fixes #2688, #2346

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -55,9 +55,10 @@ describe('Create Page', () => {
             {
                 type: 'input',
                 name: 'authors[0].user_id',
-                value: 'Annamarie Mayer{enter}',
+                value: 'Annamarie Mayer',
             },
         ]);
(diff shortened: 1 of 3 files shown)


djhi:
I'm not sure whether this should be considered a breaking change or not

PR #2842, opened 2019-01-31
Fix `CreateButton` not being re-translated when language change at runtime

### LINKED ISSUE #2830: Buttons in custom Dashboard not re-rendered on language switch
**What you were expecting:**
I have a set a custom `Dashboard` component. In this dashboard component, I added a `CreateButton` to create a particular type of resource:

```
<CreateButton resource="pets" basePath="/pets" />
```

All is working fine with this. I just added automatic switching of language using a language switcher in the app bar (which is set in a custom `Layout`, if it has some importance).

Now, the language switcher works great for the tooltips in the app bar and for my custom translations. 

**What happened instead:**
The `<CreateButton>` label stays the same, it's not re-rendered with the updated `translate` method.

**Steps to reproduce:**
1. Create a custom dashboard and add a `CreateButton` component in it
2. Create a dynamic language switcher (simply dispatching the `changeLocale` action should be enough)
3. See the `CreateButton` not being re-rendered.

**Other information:**
If it's better to report/ask on StackOverflow, don't hesitate to tell me.

**Environment**

* React-admin version: 2.4.1
* React version: 16.7

maoueh:
The `CreateButton` is using an optimization to re-render only a certain set of keys change. However, the `translate` prop was not part of that. This was causing the component to not re-render when the translate function was updated due to language having changed.

This is now fixed by specifying the `translate` as being a prop change that causes a re-render.

Fixes #2830

### DIFF
--- a/packages/ra-ui-materialui/src/button/CreateButton.js
+++ b/packages/ra-ui-materialui/src/button/CreateButton.js
@@ -77,7 +77,7 @@ CreateButton.propTypes = {
 
 const enhance = compose(
     translate,
-    onlyUpdateForKeys(['basePath', 'label']),
+    onlyUpdateForKeys(['basePath', 'label', 'translate']),
     withStyles(styles)
 );
 


maoueh:
I was really lucky (or it was a real bad luck) to hit that as `CreateButton` was the only place where `onlyUpdateForKeys` was used in conjunction with `translate` HOC :)

PR #2851, opened 2019-02-05
Add additionalMenuItems prop to Menu

Minishlink:
Hello,

Sometimes you only want to add some menu items without overriding the whole `Menu` base component.
This adds an `additionalMenuItems` prop that shows custom menu items.

Another more flexible strategy would be add a `renderAdditionalMenuItems` prop that gives `onMenuClick` as parameter.

Let me know what you think :)

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Menu.js
+++ b/packages/ra-ui-materialui/src/layout/Menu.js
@@ -43,6 +43,7 @@ const Menu = ({
     resources,
     translate,
     logout,
+    additionalMenuItems,
     ...rest
 }) => (
     <div className={classnames(classes.main, className)} {...rest}>
@@ -56,6 +57,19 @@ const Menu = ({
                     primaryText={translatedResourceName(resource, translate)}
                     leftIcon={
                         resource.icon ? <resource.icon /> : <DefaultIcon />
+                    }
+                    onClick={onMenuClick}
+                    dense={dense}
+                />
+            ))}
+        {!!additionalMenuItems &&
+            additionalMenuItems.map(menuItem => (
+                <MenuItemLink
+                    key={menuItem.key || menuItem.primaryText}
+                    to={menuItem.to}
+                    primaryText={menuItem.primaryText}
+                    leftIcon={
+                        menuItem.icon ? <menuItem.icon /> : <DefaultIcon />
                     }
                     onClick={onMenuClick}
                     dense={dense}


Minishlink:
@djhi This looks good, but it doesn't seem possible since we need to pass onMenuClick.

What do you think of : 

```js
const renderAdditionnalMenuItems = ({ onMenuClick, dense }, position) => {
  if (position === 'ABOVE_DASHBOARD') return [<MenuItem {...blablabla} />];
  ...
};
const MyMenu = props => <MyMenu>{renderAdditionalMenuItems}</MyMenu>
```

This way you have access to `onMenuClick` and you can specify where these buttons will be.

djhi:
You're right, we need a render prop here. I suggest we support both classic children and children as a function.
Instead of the `position` argument, I would add `renderDashboard` and `renderResources` functions as arguments of the children function. Usage:

```jsx
const MyMenu = () => (
    <Menu>
        {({ renderDashboard, renderResources, onMenuClick }) => (
             <>
                <MenuItem onClick={onMenuClick}>...</MenuItem>
                {renderDashboard()}
                <MenuItem onClick={onMenuClick}>...</MenuItem>
                {renderResources()}
                <MenuItem onClick={onMenuClick}>...</MenuItem>
             </>
        )}
    </Menu>
)
```

fzaninotto:
This is harder to do than to write a complete new Menu component...

fzaninotto:
Again, I'm -1 with this change, because the net benefit for the developer is negative IMO. They need to learn a new syntax, there are cases it doesn't cover (custom dashboard menu style, custom logout menu style), and it's not easier to use than to write a custom menu. 

We could export more things from the menu to make custom menus easier (like the translatedResourceName() and the complicated decoration), but I don't think that `additionalMenuItems` is the solution.

Maybe cloning children and passing them `onMenuClick` is enough?

```jsx
const MyMenu = props => (
    <Menu {...props}>
               <DashboardMenuItem />
                <MenuItemLink to="/books" primaryText="Books" leftIcon={<BookIcon />} dense="dense" />
                <MenuItemLink to="/reviews" primaryText="Reviews" leftIcon={<ReviewIcon />} dense="dense" />
                <MenuItemLink to="/authors" primaryText="Authors" leftIcon={<AuthorIcon />} dense="dense" />
                <MenuItem to="/custom">...</MenuItem>
                 <Responsive xsmall={logout} medium={null} />
    </Menu>
)
```

fzaninotto:
Any updates?

fzaninotto:

Minishlink:
@fzaninotto Yes I'll open a new one once I have the time with your proposal.

Issue #2736, opened 2019-01-07
Choose the default order value for sorting


frankPairs:
**Is your feature request related to a problem? Please describe.**
I need to be able to sort descending by default. The problem is ra-core automatically assign the ASC value to sort.order when the sort.field changes.

**Describe the solution you'd like**
I would like to be able to to define this default value. For example, I want to assign DESC value to sort.order when sort.field changes.

fzaninotto:
The default value is set in the [queryReducer](https://github.com/marmelab/react-admin/blob/d147689d9b4a0e21aacc62114ebdabbb6f63eb80/packages/ra-core/src/reducer/admin/resource/list/queryReducer.js#L30). To override it, a user should extend `ListController` and override the `changeParams` method to use their own `queryReducer`... Or write an entirely new `ListController`. Not easy.

I agree that this could be made easier. But I'm not sure this is a common need.

Can you explain your use case a bit more? Which type of data should be ordered DESC by default ?

frankPairs:
@fzaninotto thanks for the quick response.

Yes, I saw that this value is setted in the queryReducer.

The use case is:

I have a list of users with a created at and updated at values. 

By default, I have the list ordered by **updated at** DESC via List sort property because we want to show the latests updated users.

When  the user clicks to order by created at, now the library is ordering by **created at** ASC so we are showing the first created users. 

We want to continue showing the latest users.

Yeah, it is not maybe a common need but I think that would be nice to keep the default order setted in the List sort property also when the sort.field changes.

fzaninotto:
I guess we could change the queryReducer to accept not only a field, but also a direction. That way, it's up to the List (and you can override the DatagridHeaderCell for that) to decide in which direction the sort should go.

I'm marking this as an enhancement. As we don't need it ourselves, we won't implement it in the near future. Feel free to open a PR to add this feature though.

fzaninotto:
Fixed by #4518

PR #2855, opened 2019-02-06
Can set order value for SET_SORT action

frankPairs:
Hi!

This PR is related to the issue https://github.com/marmelab/react-admin/issues/2736.

Now we can pass a order value. This order value will change the state only when sort value is different than the previousState.sort.

It allows us to define a default order value every time sort value is new.

Regards!


packages/ra-core/src/reducer/admin/resource/list/queryReducer.spec.js
```diff
@@ -66,4 +66,50 @@ describe('Query Reducer', () => {
...
+            const updatedState = queryReducer(
+                {},
+                {
+                    type: 'SET_SORT',
+                    payload: { sort: 'foo', order: SORT_DESC },
+                }
+            );
+            assert.deepEqual(updatedState, { sort: 'foo', order: SORT_DESC, page: 1  });
+        });
+        it('should set order with the opposite order sort value is the same llke the previous state', () => {
```
djhi:
```suggestion
        it("should set order as the opposite of the one in previous state when sort hasn't change", () => {
```

packages/ra-core/src/reducer/admin/resource/list/queryReducer.spec.js
```diff
@@ -66,4 +66,50 @@ describe('Query Reducer', () => {
...
+                    sort: 'foo', order: SORT_DESC, page: 1
+                },
+                {
+                    type: 'SET_SORT',
+                    payload: { sort: 'foo' },
+                }
+            );
+            assert.deepEqual(updatedState, { sort: 'foo', order: SORT_ASC, page: 1  });
+        });
+        it('should set order with the opposite order sort value is the same llke the previous state even when we pass a specific order', () => {
```
djhi:
```suggestion
        it("should set order as the opposite of the one in previous state even if order is specified in the payload when sort hasn't change", () => {
```

djhi:
By the way, as this is technically a new feature, would you mind targeting the `next` branch?

frankPairs:
@djhi of course. I am going to change the target and the text titles.

thanks for reviewing

fzaninotto:
we need to rebase `next` on `master`, as the `queryReducer` is now written in TypeScript

djhi:
Sorry about that @frankPairs, we are migrating to TS

frankPairs:
@djhi  @fzaninotto  it's ok, i will update the PR using TS

frankPairs:
@djhi @fzaninotto sorry I did a rebase.

I am going to create another PR from the next branch with this changes, I think that it is simpler.

thanks

PR #2868, opened 2019-02-11
Touch ability to ra-rich-text-input

ggarcia92:
This component shows the error message without user touch the element, with this update, the message error only appears when user click or select the input. It is a solution to simalate the way that other react-admin inputs work.

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -61,10 +61,6 @@ class RichTextInput extends Component {
         this.quill = null;
     }
 
-    componentWillReceiveProps(nextProps){
-        this.editor.innerHTML = nextProps.input.value
-    }
-
     onTextChange = () => {
         const value =
             this.editor.innerHTML === '<p><br></p>' ? '' : this.editor.innerHTML;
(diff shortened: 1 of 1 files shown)


fzaninotto:
Sorry, I don't understand the issue. Can you describe it in more details, perhaps with a screencast?

ggarcia92:
Hi @fzaninotto . All react-admin input components show an validation error message only if user touched the input. For example, if i have a text input with required validation, when the browser render the form no error message is show until user click the input field to enter some value. But RichTextInput show this message when form is rendered. Regards

fzaninotto:
Redux-form already tracks the `dirty` and `pristine` states of a field, I don't think you need to track your own state. Try using Redux-form instead.

packages/ra-input-rich-text/src/index.js
```diff
@@ -61,6 +61,10 @@ class RichTextInput extends Component {
         this.quill = null;
     }
 
+    componentWillReceiveProps(nextProps){
```
fzaninotto: I'm not convinced this solves a bug. Please remove this change and open a new PR for it, where we can choose to merge it or leave it independently of the other issue.

ggarcia92: hi @fzaninotto , I will remove this commit and open a new PR with it. Thanks Regards

fzaninotto:
Also, please consider my earlier comment:

> Redux-form already tracks the `dirty` and `pristine` states of a field, I don't think you need to track your own state. Try using Redux-form instead.

fzaninotto:
Any updates?

fzaninotto:

PR #2872, opened 2019-02-12
Update Unit Testing documentation

mexitalian:
The previous import location threw an error for me as it couldn't be found.


djhi:
The `react-admin` package just re-export what `ra-core` exports. So both imports should work

Kmaschta:
I somewhat agree with that PR: `ra-core` isn't documented (yet) and it might be confusing to the early adopters why we require that package in a simple unit test example.

djhi:
@fzaninotto What do you think about it? I agree with @Kmaschta

fzaninotto:
Thanks!

PR #2873, opened 2019-02-13
Fix GET_MANY in ra-data-json-server data provider returns too many results

paradoxxxzero:
I think using the [json-server](https://github.com/typicode/json-server) `like` feature is not what we want here, since an `id_like=2|33|444` will match 2, 33, 444 but also 23, 133 and so forth.
We could restrict the match with delimiters (`^` and `$`) but since it supports multiple filters query, it would be simpler to use just that.

### DIFF
--- a/packages/ra-data-json-server/src/index.js
+++ b/packages/ra-data-json-server/src/index.js
@@ -81,7 +81,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
                 break;
             case GET_MANY: {
                 const query = {
-                    [`id_like`]: params.ids.join('|'),
+                  id: params.ids
                 };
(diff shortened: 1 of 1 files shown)


fzaninotto:
Are you sure this works? I tested the [`/posts?id=[1,2]`](http://jsonplaceholder.typicode.com/posts?id%3D%5B1%2C2%5D) URL on JSONPlaceholder but it doesn't seem to filter the list.

paradoxxxzero:
This is not `/posts?id=[1,2]` but `/posts?id=1&id=2`

fzaninotto:
Gotcha, thanks for the explanation.

fzaninotto:
And thanks for the PR!

PR #2878, opened 2019-02-13
[RFR] Migrate ra-core form code to TypeScript

djhi:


packages/ra-core/src/form/validate.ts
```diff
@@ -23,7 +48,10 @@ const getMessage = (message, messageArgs, value, values, props) =>
 // If we define validation functions directly in JSX, it will
 // result in a new function at every render, and then trigger infinite re-render.
 // Hence, we memoize every built-in validator to prevent a "Maximum call stack" error.
-const memoize = fn => lodashMemoize(fn, (...args) => JSON.stringify(args));
+const memoize = (fn: any) =>
+    lodashMemoize(fn, (...args) => JSON.stringify(args));
+
+type Required = (message?: string | MessageFunc) => Validator;
```
fzaninotto: I don't understand why you needed to create one type per validator, can you please explain? Is it because `memoize` loses the types?

djhi: Exactly

fzaninotto: OK, thanks

djhi: I found a better way

PR #2881, opened 2019-02-13
[RFR] Migrate ra-core controllers to TypeScript

djhi:
- [x] Fields
- [x] Inputs
- [x] Views


packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -6,16 +6,19 @@ export interface CrudGetManyAccumulateAction {
...
-        ids: [];
+        ids: any[];
     };
     readonly meta: {
         accumulate: any;
     };
 }
 
+// Used to type the dispatcher function (the one injected by mapDispatchToProps)
+export type CrudGetManyAccumulate = (resource: string, ids: any[]) => void;
```
djhi: Maybe this should be named `DispatchCrudGetManyAccumulate`. @fzaninotto, what do you think?

fzaninotto: "Dispatch" should only be used when the action creator is decorated by redux. Otherwise, we may end up with a `dispatchDispatchCrudGetManyAccumulate` prop name.

djhi: Well, this is actually the type used for the function passed as prop by the `mapToDispatch` argument of connect

fzaninotto: understood, but it's not a dispatcher at that step

djhi: Hu ? It is

packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -1,12 +1,13 @@
...
+import { Pagination, Sort } from '../types';
 
 export const CRUD_GET_MANY_ACCUMULATE = 'RA/CRUD_GET_MANY_ACCUMULATE';
 
 export interface CrudGetManyAccumulateAction {
     readonly type: typeof CRUD_GET_MANY_ACCUMULATE;
     readonly payload: {
         resource: string;
-        ids: [];
+        ids: any[];
```
fzaninotto: why not `Identifier[]`?

packages/ra-core/src/actions/accumulateActions.ts
```diff
@@ -15,7 +16,7 @@ export interface CrudGetManyAccumulateAction {
 
 export const crudGetManyAccumulate = (
     resource: string,
-    ids: []
+    ids: any[]
```
fzaninotto: Same remark

packages/ra-core/src/actions/accumulateActions.ts
djhi: Good point, it should be `Identifier[]`

packages/ra-core/src/controller/CreateController.tsx
```diff
@@ -50,7 +81,13 @@ import checkMinimumRequiredProps from './checkMinimumRequiredProps';
  *     );
  *     export default App;
  */
-export class CreateController extends Component {
+export class CreateControllerView extends Component<Props> {
```
fzaninotto: I don't agree with the renaming. In that particular case, it is not a view, it is a controller. If you're looking for a better name, try `CreateControllerCore` or `UnconnectedCreateController`.

Same for other renamings in ra-core controllers.

packages/ra-core/src/controller/CreateController.tsx
```diff
@@ -50,7 +81,13 @@ import checkMinimumRequiredProps from './checkMinimumRequiredProps';
...
+    public static defaultProps = {
+        record: {},
```
fzaninotto: `Partial<Record>`?

packages/ra-core/src/controller/CreateController.tsx
```diff
@@ -1,13 +1,44 @@
...
+    hasCreate: boolean;
+    hasEdit: boolean;
+    hasList: boolean;
+    hasShow: boolean;
+    isLoading: boolean;
+    location: Location;
+    match: Match;
+    record?: Record;
+    resource: string;
+    title: string | ReactNode;
```
fzaninotto: You didn't transpose the propTypes correctly. `title`, for instance, was not required. Please double check for all components.

packages/ra-core/src/controller/EditController.tsx
```diff
@@ -1,13 +1,53 @@
...
+    crudGetOne,
+    crudUpdate,
+    startUndoable as startUndoableAction,
+} from '../actions';
 import { REDUX_FORM_NAME } from '../form';
 import checkMinimumRequiredProps from './checkMinimumRequiredProps';
+import { Translate, Record, Dispatch } from '../types';
+import { RedirectionSideEffect } from '../sideEffect';
+
+interface ChildrenFuncParams {
```
fzaninotto: :+1: This is making the code much clearer

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -70,9 +145,29 @@ import checkMinimumRequiredProps from './checkMinimumRequiredProps';
...
+        perPage: 10,
+        sort: {
+            field: 'id',
+            order: SORT_DESC,
+        },
+    };
+
     state = {};
 
+    setFilters = debounce(filters => {
```
fzaninotto: why did you move that function?

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -310,9 +395,11 @@ export class ListController extends Component {
...
             onToggleItem: this.handleToggleItem,
             onUnselectItems: this.handleUnselectItems,
-            page: parseInt(query.page || 1, 10),
-            perPage: parseInt(query.perPage, 10),
-            refresh: this.refresh,
+            page:
+                (typeof query.page === 'string'
+                    ? parseInt(query.page, 10)
+                    : query.page) || 1,
+            perPage: parseInt(query.perPage.toString(), 10),
```
fzaninotto: please use the same test as in the previous line. I prefer not casting to string to get a number in the end

packages/ra-core/src/controller/ShowController.tsx
```diff
@@ -1,11 +1,41 @@
...
+    children: (params: ChildrenFuncParams) => ReactNode;
+    crudGetOne: Dispatch<typeof crudGetOneAction>;
+    record?: Record;
+    hasCreate: boolean;
+    hasEdit: boolean;
+    hasShow: boolean;
+    hasList: boolean;
+    id: string;
+    isLoading: boolean;
+    location: object;
```
fzaninotto: I think we can use the React Router types here

packages/ra-core/src/controller/field/ReferenceArrayFieldController.tsx
```diff
@@ -1,10 +1,30 @@
...
 import { connect } from 'react-redux';
 import get from 'lodash/get';
 
 import { crudGetManyAccumulate as crudGetManyAccumulateAction } from '../../actions';
 import { getReferencesByIds } from '../../reducer/admin/references/oneToMany';
+import { ReduxState, Record, RecordMap, Dispatch } from '../../types';
+
+interface ChildrenFuncParams {
+    loadedOnce: boolean;
+    ids: any[];
```
fzaninotto: identifiers

packages/ra-core/src/controller/field/ReferenceArrayFieldController.tsx
```diff
@@ -1,10 +1,30 @@
...
+    data: RecordMap;
+    referenceBasePath: string;
+    currentSort: any;
```
fzaninotto: Sort

packages/ra-core/src/controller/field/ReferenceManyFieldController.tsx
```diff
@@ -61,11 +93,19 @@ import {
...
-    }
+export class ReferenceManyFieldControllerView extends Component<Props> {
+    public static defaultProps: Partial<Props> = {
+        filter: {},
+        perPage: 25,
+        sort: { field: 'id', order: 'DESC' },
+        source: 'id',
+    };
+
+    public state = {
```
fzaninotto: Could you type it?

packages/ra-core/src/controller/input/ReferenceInputController.tsx
```diff
@@ -16,9 +15,56 @@ import {
...
+
+interface Props {
+    allowEmpty?: boolean;
+    basePath: string;
+    children: (params: ChildrenFuncParams) => ReactNode;
+    crudGetMatchingAccumulate: Dispatch<typeof crudGetMatchingAccumulateAction>;
+    crudGetManyAccumulate: Dispatch<typeof crudGetManyAccumulateAction>;
+    filter?: object;
+    filterToQuery: (filter: {}) => any;
+    input?: {
```
fzaninotto: don't we have a type from redux-form for that?

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -75,7 +75,11 @@ export const getReferences = (state: ReduxState, reference, relatedTo) => {
         }, {});
 };
 
-export const getReferencesByIds = (state: ReduxState, reference, ids) => {
+export const getReferencesByIds = (
+    state: ReduxState,
+    reference: string,
+    ids: any[]
```
fzaninotto: identifier

packages/ra-core/src/reducer/admin/resource/list/params.ts
```diff
@@ -30,7 +30,7 @@ const paramsReducer: Reducer<State> = (
 ) => {
     switch (action.type) {
         case CRUD_CHANGE_LIST_PARAMS:
-            return action.payload;
+            return { ...previousState, ...action.payload };
```
fzaninotto: please revert

packages/ra-core/src/types.ts
```diff
@@ -4,6 +4,10 @@ export interface Record {
     [key: string]: any;
 }
 
+export interface RecordMap {
+    [id: string]: Record;
```
fzaninotto: unfortunately, there is a bug in TypeScript that doesn't let you use Identifer as a key type, but I'm afraid this will break if the id is a number

packages/ra-core/src/controller/CreateController.tsx
djhi: Yes and actually, that's a prop of the `ra-ui-materialui` `Create` component.

packages/ra-core/src/controller/ListController.tsx
djhi: That's not me, tslint did it. I suppose it's because it considers it a private field and not a private function

packages/ra-core/src/controller/input/ReferenceInputController.tsx
djhi: We would have to include the redux-form types in our dev deps and last time I tried that, they were not good

packages/ra-core/src/controller/CreateController.tsx
djhi: I type `defaultProps` instead

packages/ra-core/src/controller/input/ReferenceInputController.tsx
djhi: I added them and it seems ok

packages/ra-core/src/controller/CreateController.tsx
fzaninotto: I saw you remove the title prop here, but I suspect you didn't cross check the isRequired attribute on all props. hasEdit, for instance, is required in your version but was not require din propTypes

packages/ra-core/src/controller/ShowController.tsx
```diff
@@ -1,11 +1,41 @@
...
+    translate: Translate;
+    version: number;
+}
+
+interface Props {
+    basePath: string;
+    children: (params: ChildrenFuncParams) => ReactNode;
+    crudGetOne: Dispatch<typeof crudGetOneAction>;
+    record?: Record;
+    hasCreate: boolean;
```
fzaninotto: still missing optional here

fzaninotto:
\o/

djhi:
Wait before we test them. I'm sure we're going to be spanked hard by TS when using those definitions in the other packages :cold_sweat:

Issue #2885, opened 2019-02-14
Create button redirect with more data in url querystring


developerium:
Hi Thanks for this great tool. Helped me a lot!

Sometimes when I redirect my users to create page, I want to pass some parameters to that page, with url querystrings.

Using CreateButton, I think it would be ideal to pass an object of key value pairs, and have CreateButton add it to url. I know I can invoke an action to add the parameter to form, but I would very much like it in the url, since they can share it, and I don't want to lose the parameters.

What do you guys think?

developerium:
I can make a pull request if you think it's a cool feature

djhi:
Do you mean this: https://marmelab.com/react-admin/CreateEdit.html#prefilling-a-create-record ?

developerium:
Yes this is also a good idea, thanks

djhi:
Oh you meant to handle this automatically on the `CreateButton`? If so, yes, a PR would be nice :+1:

developerium:
I made a pull request [#2892](https://github.com/marmelab/react-admin/pull/2892)

red1:
Can we expect this branch to be merged ?

developerium:
@red1 I don't think it is going to be merged with the approach I took on my branch, but feel free to use the code from my branch and use it

PR #2892, opened 2019-02-16
Add additional data to CreateButton

developerium:
Hi guys, in following of [this](https://github.com/marmelab/react-admin/issues/2885) discussion, I made this pull request.
Let me how it looks like.

Thanks

### DIFF
--- a/packages/ra-ui-materialui/src/button/CreateButton.js
+++ b/packages/ra-ui-materialui/src/button/CreateButton.js
@@ -35,35 +35,47 @@ const CreateButton = ({
     translate,
     label = 'ra.action.create',
     icon = <ContentAdd />,
+    data,
     ...rest
-}) => (
-    <Responsive
-        small={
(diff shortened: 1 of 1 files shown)


djhi:
Thanks for this PR. Unfortunately, as explained in the [documentation](https://marmelab.com/react-admin/CreateEdit.html#prefilling-a-create-record) (end of this section), this would not work correctly for people using numbers as identifiers.

I don't think we can have a good default here so maybe documenting it is enough. I suggest you use that component in your project but we probably won't inlude that in the core.

developerium:
Could you direct me to how can I make it better?

djhi:
Unfortunately no. I really believe there is no good default suitable for everyone so we shouldn't include one. We made it easy to override though and I think it's enough. If you think your component could be useful for others, feel free to distribute it as an addon package but we won't include it in.

PR #2893, opened 2019-02-16
Add .idea folder to ignored files

developerium:
I use webstorm, and I know this is not a major change, but it would be nice to ignore jetbrains IDE generated files by default.
Let me know what you guys think.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ examples/**/static
 examples/**/dist
 cypress/videos
 cypress/screenshots
+.idea


fzaninotto:
Sorry, we won't track every file every editor may leave in project directories. Instead, I recommend you add the .idea name to the list of ignored file in your global git configuration.

See https://stackoverflow.com/a/5724484 for details.

developerium:
Thanks for the advice

PR #2898, opened 2019-02-18
Corrected prop types for ArrayInput

kujon:
It wasn't letting arrays through, even though they work absolutely fine.


kujon:
you're welcome!

PR #2899, opened 2019-02-18
[RFR] Add withDataProvider HOC and Query/Mutation components to ease custom queries

### LINKED ISSUE #2645: [RFC] side effect helper
## Problem

Adding side effects to actions from within a component is cumbersome. There is [the `callback` saga](https://marmelab.com/react-admin/Actions.html#custom-sagas), which allows to add a function to be executed on success or failure, but that requires a custom action. It can't be used e.g. after a regular crud action.

## Proposed solution

Provide a helper to add success or failure side effects to an existing action. The helper is an action creator that takes an action as parameter, and returns another action with side effects added.

Except I can't find a nice API with no drawbacks...

### Option 1 - jQuery style

* Pros: Explicit, familiar to frontend devs
* Cons: Old school

```jsx
import React from 'react';
import { withSideEffect, crudUpdate } from 'react-admin';

class ApproveButton extends React.Component {
    handleClick() {
        const { withSideEffect, id } = this.props;
        const action = crudUpdate('comment', id, { is_approved: true } );
        withSideEffect(action, {
            success: data => {
                // do something with the data
            },
            error: error => {
                // do something with the error
            }
        })
    }

    render() {
        // ...
    }
}

export default connect(null, { withSideEffect })(MyComponent);
```

### Option 2 - NodeJs style

Expect a callback as second argument.

* Pros: Familiar to NodeJs users
* Cons: This syntax is more uncommon on the frontend, and more and more deprecated on the backend. Also, it doesn't encourage catching errors. 

```jsx
import React from 'react';
import { withSideEffect, crudUpdate } from 'react-admin';

class ApproveButton extends React.Component {
    handleClick() {
        const { withSideEffect, id } = this.props;
        const action = crudUpdate('comment', id, { is_approved: true } );
        withSideEffect(action, (data, error) => {
            if (error) {
                // do something with the error
            }
            // do something with the data
        })
    }

    render() {
        // ...
    }
}

export default connect(null, { withSideEffect })(MyComponent);
```

### Option 3 - Positioned arguments style

* Pros: Concise
* Cons: Not explicit, not intuitive, requires reading the doc

```jsx
import React from 'react';
import { withSideEffect, crudUpdate } from 'react-admin';

class ApproveButton extends React.Component {
    handleClick() {
        const { withSideEffect, id } = this.props;
        const action = crudUpdate('comment', id, { is_approved: true } );
        withSideEffect(
            action,
            data => { // success side effect
                // do something with the data
            },
            error => { // error side effect
                // do something with the error
            })
    }

    render() {
        // ...
    }
}

export default connect(null, { withSideEffect })(MyComponent);
```

### Option 4 - fetch / Promise style

Requires treating actions as mutable objects, and modifying actions *after* being dispatched, but before being handled by sagas. Yes, it's possible.

* Pros: Familiar, Intuitive
* Cons: Misleading, error-prone (since this isn't a real Promise, so it can't be `await`ed nor chained further), Not robust (modifying actions after the reducer isn't a Redux documented feature)

```jsx
import React from 'react';
import { fetchWithRedux, crudUpdate } from 'react-admin';

class ApproveButton extends React.Component {
    handleClick() {
        const { fetchWithRedux, id } = this.props;
        const action = crudUpdate('comment', id, { is_approved: true } );
        fetchWithRedux(action)
            .then(data => {
                // do something with the data
            })
            .catch(error => {
                // do something with the error
            })
    }

    render() {
        // ...
    }
}

export default connect(null, { fetchWithRedux })(MyComponent);
```

### Option 5 - no style 

Same hack as option 4, except that, by renaming then and error to success and failure, it looks less like a Promise

* Pros: Somewhat familiar, although I don't remember which AJAX library used that syntax
* Cons: Still hackish and not robust

```jsx
import React from 'react';
import { withSideEffect, crudUpdate } from 'react-admin';

class ApproveButton extends React.Component {
    handleClick() {
        const { withSideEffect, id } = this.props;
        const action = crudUpdate('comment', id, { is_approved: true } );
        withSideEffect(action)
            .success(data => {
                // do something with the data
            })
            .failure(error => {
                // do something with the error
            })
    }

    render() {
        // ...
    }
}

export default connect(null, { withSideEffect })(MyComponent);
```

### Option 6 ? Feel free to suggest a better API!

ThieryMichel:
## Problem 

Calling the API directly is a bit cumbersome, especially when handling side effects or chaining calls. We need a better way to do that.

## Solution

- [x] Add `withDataProvider()` HOC
- [x] Document `withDataProvider()` HOC
- [x] Add `<Query>` component
- [x] Document `<Query>` component
- [x] Add `<Mutation>` component
- [x] Document `<Mutation>` component
- [x] Use those in the demo example
- [x] Test all that

Note: Of course we'll use a hook for that once we make the move to React 16.8.

## Sample Syntax

**withDataProvider**

```jsx
import { withDataProvider, showNotification } from 'react-admin';

class PostList extends Component {
    state = {
        posts: [],
    }

    componentDidMount() {
        const { dataProvider, dispatch } = this.props;
        dataProvider('GET_LIST', 'posts', { filter: { status: 'pending' }})
           .then(({ data: posts }) => this.setState({ posts }))
           .catch(error => dispatch(showNotification(error.message, 'error')))
    }

    render() {
        const { posts } = this.state;
        return (
           <Fragment>
               {posts.map((post, index) => <PostDetail post={post} key={key} />)}
           </Fragment>
        );
    }
}

PostList.propTypes = {
    dataProvider: PropTypes.func.isRequired,
};

export default withDataProvider(PostList);
```

**Query**

```jsx
const UserProfile = ({ record }) => (
    <Query type="GET_ONE" resource="users" payload={{ id: record.id }}>
        {({ data, loading, error }) => {
            if (loading) { return <Loading />; }
            if (error) { return <p>ERROR</p>; }
            return <div>User {data.username}</div>;
        }}
    </Query>
);
```

**Mutation**

```jsx
const ApproveButton = ({ record }) => (
    <Mutation
        type="UPDATE"
        resource="comments"
        payload={{ id: record.id, data: { isApproved: true } }}
    >
        {(approve) => (
            <FlatButton label="Approve" onClick={approve} />
        )}
    </Mutation>
);
```

Closes #2645


fzaninotto:
`withDataProvider` isn't tested directly, but across `<Query>` and `<Mutation>`. This should be enough. Switching to RFR.

docs/Actions.md
```diff
@@ -1,17 +1,25 @@
...
 
-## The Simple Way
+* [Using `fetch`](#the-simple-way-using-fetch)
+* [Using the `dataProvider`](#using-the-data-provider-instead-of-fetch)
+* [Using the `withDataProvider` Decorator](#using-the-withdataprovider-decorator)
+* [Using the `<Query>` and `<Mutation>` Components](#query-and-mutation-components)
+* [Using a Custom Action Creator](#using-a-custom-action-creator)
 
-Here is an implementation of the "Approve" button that works perfectly:
+**Tip**: Most of the time, the `<Query>` and `<Mutation>` components will do everything you need wit hthe simplest API.
```
djhi: s/wit hthe/with the/

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+-       dataProvider(UPDATE, 'comments', { id: record.id, data: updatedRecord })
+-           .then(() => {
+-               dispatch(showNotification('Comment approved'));
+-               dispatch(push('/comments'));
+-           })
+-           .catch((e) => {
+-               dispatch(showNotification('Error: comment not approved', 'warning'))
+-           });
+-   }
++       dataProvider(UPDATE, 'comments', { id: record.id, data: updatedRecord }, {
```
djhi: Neat :)

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+-   dispatch: PropTypes.func.isRequired,
+    record: PropTypes.object,
+};
+
+export default withDataProvider(ApproveButton);
+```
+
+React-admin can handle the following side effects:
+
+- `notification`: Display a notification. The property value should be an object describing the notification to display. The `body` can be a translation key. `level` can be either `info` or `warning`.
```
djhi: Probably worth mentionning the message args too

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+- `redirectTo`: Redirect the user to another page. The property value should be the path to redirect the user to.
+- `basePath`: This is not a side effect, but it's used internally to compute redirection paths. Set it when you have a redirection side effect.
+- `refresh`: Force a rerender of the current view (equivalent to pressing the Refresh button). Set to true to enable.
+- `unselectAll`: Unselect all lines in the current datagrid. Set to true to enable.
+- `callback`: Execute an arbitrary function. The value should be the function to execute. It receives the `requestPayload` and the response `payload`.
```
djhi: Probably worth mentionning it receives them in a single object

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+You can benefit from optimistic rendering when you call the `dataProvider` prop function, too. You just need to pass the `undoable: true` option in the options parameter:
+
+```diff
+// in src/comments/ApproveButton.js
+class ApproveButton extends Component {
+    handleClick = () => {
+        const { dataProvider, dispatch, record } = this.props;
+        const updatedRecord = { ...record, is_approved: true };
+        dataProvider(UPDATE, 'comments', { id: record.id, data: updatedRecord }, {
++           undoable: true,
```
djhi: :tada:

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+}
+```
+
+The fact that react-admin can handle side effects and undo a call to the API if you use `withDataProvider` should be a good motivation to prefer it to raw `fetch`.
+
+## `<Query>` and `<Mutation>` Components
+
+When using the `withDataProvider` decorator to fetch data from the API, you must create a stateful class component to handle the initial state, the loading state, the loaded state, and the error state. That's a lot of boilerplate for a simple query.
+
+For such cases, react-admin provides a `<Query>` component, which uses `withDataProvider` under the hood. It leverages the render props pattern to reduce the boilerplate.
```
djhi: render props are soooo 2018 ^^

docs/Actions.md
```diff
@@ -289,67 +432,22 @@ class ApproveButton extends Component {
...
-- `basePath`: This is not a side effect, but it's used internally to compute redirection paths. Set it when you have a redirection side effect.
+Just like for the `withDataProvider`, you can associate side effects to a fetch action declaratively by setting the appropriate keys in the action `meta`.
 
-## Success and Failure Side Effects
-
-React-admin triggers all side effects declared in the `meta` property of an action *simultaneously*. So in the previous example, the "notification approved" notification appears when the `COMMENT_APPROVE` action is dispatched, i.e. *before* the server is even called. That's a bit too early: what if the server returns an error?
-
-In practice, most side effects must be triggered only after the `fetch` side effect succeeds or fails. To support that, you can enclose side effects under the `onSuccess` and `onFailure` keys in the `meta` property of an action:
+So the side effects will be declared in the action creator rather than in the component. For instance, to display a notification when the `COMMENT_APPROVE` action is sucessfully dispatched, add the `notification` meta:
```
djhi: s/sucessfully/successfully/

docs/Actions.md
```diff
@@ -385,37 +477,18 @@ export const commentApprove = (id, data, basePath) => ({
...
-            level: 'info',
-        },
-        redirectTo: '/comments',
-        basePath,
-    },
-}
-```
-
-And then, the side effects will trigger. With this code, approving a review now displays the correct notification, and redirects to the comment list.
+The side effects accepted in the `meta` field of the aciton are the same as in the fourth parameter of the `dataProvider` function injected by `withDataProvider`:
```
djhi: s/aciton/action/

docs/Actions.md
```diff
@@ -385,37 +477,18 @@ export const commentApprove = (id, data, basePath) => ({
...
-
-For its own fetch actions, react-admin uses an approach called *optimistic rendering*. The idea is to handle the `fetch` actions on the client side first (i.e. updating entities in the Redux store), and re-render the screen immediately. The user sees the effect of their action with no delay. Then, react-admin applies the success side effects, and only after that, it triggers the call to the data provider. If the fetch ends with a success, react-admin does nothing more than a refresh to grab the latest data from the server. In most cases, the user sees no difference (the data in the Redux store and the data from the data provider are the same). If the fetch fails, react-admin shows an error notification, and forces a refresh, too.
+- `notification`: Display a notification. The property value should be an object describing the notification to display. The `body` can be a translation key. `level` can be either `info` or `warning`.
```
djhi: same remark about message args

docs/Actions.md
```diff
@@ -385,37 +477,18 @@ export const commentApprove = (id, data, basePath) => ({
...
+- `redirectTo`: Redirect the user to another page. The property value should be the path to redirect the user to.
+- `refresh`: Force a rerender of the current view (equivalent to pressing the Refresh button). Set to true to enable.
+- `unselectAll`: Unselect all lines in the current datagrid. Set to true to enable.
+- `callback`: Execute an arbitrary function. The meta value should be the function to execute. It receives the `requestPayload` and the response `payload`.
```
djhi: same about those being received inside a single object

packages/ra-core/src/sideEffect/callback.ts
```diff
@@ -21,9 +22,10 @@ interface ActionWithSideEffect {
 function* handleCallback({
     payload,
     requestPayload,
+    error,
     meta: { callback },
 }: ActionWithSideEffect) {
-    yield call(callback, { payload, requestPayload });
+    yield call(callback, { payload, requestPayload, error });
```
djhi: The `error` field should be mentionned in the documentation too

packages/ra-core/src/util/Mutation.tsx
```diff
@@ -0,0 +1,83 @@
+import { Component, ReactNode } from 'react';
+import withDataProvider from './withDataProvider';
+
+type DataProviderCallback = (
+    type: string,
```
djhi: Maybe we can explicitly allow only the mutation types here

packages/ra-core/src/util/Query.tsx
```diff
@@ -0,0 +1,80 @@
+import { Component, ReactNode } from 'react';
+import withDataProvider from './withDataProvider';
+
+type DataProviderCallback = (
+    type: string,
```
djhi: Maybe we can explicitly allow only the query types here

docs/Actions.md
fzaninotto: A bit too long to describe for here. We'll have to write that down in another section, and place a reference here. To be done in a future PR.

docs/Actions.md
fzaninotto: again, to be documented elsewhere

docs/Actions.md
```diff
@@ -1,17 +1,25 @@
 ---
 layout: default
-title: "Actions"
+title: "Queryng the API"
```
Kmaschta: s/Queryng/Querying/

packages/ra-core/src/util/Mutation.tsx
fzaninotto: No, because we want to allow custom dataProvider verbs in general.

docs/Actions.md
Kmaschta: Explain why it's the simplest API: it handle handle the data provider nicely, etc.
And add a link to the paragraph related to these components

packages/ra-core/src/util/Query.tsx
fzaninotto: Same answer as above

docs/Actions.md
```diff
@@ -1,17 +1,25 @@
...
+* [Using `fetch`](#the-simple-way-using-fetch)
+* [Using the `dataProvider`](#using-the-data-provider-instead-of-fetch)
+* [Using the `withDataProvider` Decorator](#using-the-withdataprovider-decorator)
+* [Using the `<Query>` and `<Mutation>` Components](#query-and-mutation-components)
+* [Using a Custom Action Creator](#using-a-custom-action-creator)
 
-Here is an implementation of the "Approve" button that works perfectly:
+**Tip**: Most of the time, the `<Query>` and `<Mutation>` components will do everything you need wit hthe simplest API.
+
+## The Simple Way: Using `fetch`
```
Kmaschta: Since we advice users to use Query and Mutation, we should not say the fetch is the simple way?

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+{% endraw %}
+
+Just like the `dataProvider` injected prop, the `<Query>` component expects three paramaters: `type`, `resource`, and `payload`. It fetches the data provider on mount, and passes the data to its child component once the response from the API arrives.
+
+The `<Query>` component is designed to read data from the API. When calling the API to update ("mutate") data, use the `<Mutation>` component instead. It passes a callback to trigger the API call to its child function. And the `<ApproveButton>` component from previous sections is a great use case for demonstrating `<Mutation>`:
+
+```jsx
+import { Mutation } from 'react-admin';
+
+const options =
```
Kmaschta: It miss something after the `options =`

docs/Actions.md
```diff
@@ -214,26 +217,184 @@ class ApproveButton extends Component {
...
+            if (loading) { return <Loading />; }
+            if (error) { return <p>ERROR</p>; }
+            return <div>User {data.username}</div>;
+        }}
+    </Query>
+);
+```
+{% endraw %}
+
+Just like the `dataProvider` injected prop, the `<Query>` component expects three paramaters: `type`, `resource`, and `payload`. It fetches the data provider on mount, and passes the data to its child component once the response from the API arrives.
```
Kmaschta: s/paramaters/parameters/

docs/Actions.md
```diff
@@ -692,8 +767,13 @@ You can find a complete example of a custom Bulk Action button in the `List` doc
...
 
-Which style should you choose for your own action buttons?
+Which style should you choose for your own action buttons? Here is a quick benchmark:
 
-The first version (with `fetch`) is perfectly fine, and if you're not into unit testing your components, or decoupling side effects from pure functions, then you can stick with it without problem.
 
-On the other hand, if you want to promote reusability, separation of concerns, adhere to react-admin's coding standards, and if you know enough Redux and Saga, use the final version.
+Solution | Advantages | Drawbacks 
+---------|------------|-----------
+`fetch`  | Nothing to learn | Requires duplication of authentication, does not handle the loading state, adds boilerplate
```
Kmaschta: Can you use lists in the advantages and drawbacks columns?

fzaninotto:
I'm not fan of writing a redirection page in Markdown - and I 'm not even sure it's possible. Let's keep the file name (and the URL) as it is, as the menu label has been properly changed.

PR #2904, opened 2019-02-20
add a note about CloneButton component to docs

### LINKED ISSUE #2781: CloneButton should sanitize its rest props
**What you were expecting:**
Using a `<CloneButton />` inside an `<Toolbar />` from an `Edit` view should not trigger any React warnings

**What happened instead:**
Lots of warnings about unkown props passed by the `<Toolbar />`

**Steps to reproduce:**
Using a `<CloneButton />` inside an `<Toolbar />` from an `Edit` view .

**Related code:**
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/button/CloneButton.js#L25

**Environment**

* React-admin version: 2.6.1

mnlbox:
Hi guys,
It's my first PR here. Really thanks about this awesome project. :heart: 

Closes #2781

**/CC:** @djhi


fzaninotto:
Thanks!

PR #2909, opened 2019-02-21
fix(List.js): add children style

alirni:
We do not have access to holders of `children` for personalization. And this holder has been blocking our layout. This `div` should be deleted or added to that `CSS class`.
That's why I added this `class` through which we can personalize.

### DIFF
--- a/packages/ra-ui-materialui/src/list/List.js
+++ b/packages/ra-ui-materialui/src/list/List.js
@@ -33,6 +33,7 @@ export const styles = {
         justifyContent: 'space-between',
         alignSelf: 'flex-start',
     },
+    children: { display: 'flex' },
     noResults: { padding: 20 },
 };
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/List.js
```diff
@@ -33,6 +33,7 @@ export const styles = {
         justifyContent: 'space-between',
         alignSelf: 'flex-start',
     },
+    children: { display: 'flex' },
```
fzaninotto: You're breaking BC by adding a new flex style. If you just want to make it configurable, you just need to have an empty object literal here.

fzaninotto:
Bump, please rebase on `next` and initialize the new style to empty properties.

fzaninotto:

PR #2912, opened 2019-02-21
Add allowDuplicates property to AutocompleteArrayInput and fix #2311

### LINKED ISSUE #2311: AutocompleteArrayInput allows to pick repeated values (probably a feature?)
```js
import { AutocompleteArrayInput } from 'react-admin';

<AutocompleteArrayInput source="category" choices={[
    { id: 'programming', name: 'Programming' },
    { id: 'lifestyle', name: 'Lifestyle' },
    { id: 'photography', name: 'Photography' },
]} />
```

Just use the example component. It allows to pick the same value twice. If you remove one of the repeated values, all of them are removed, so it looks like it probably doesn't make sense to allow repeated values (bug).

If it does make sense, there should be an attribute to control whether we want repeated values or not (feature)?

tiagoschenkel:
The method **`handleSuggestionSelected`** is adding the new value directly to the **`input`** through **`onChange`** method. Even when the property **`allowDuplicates`** is enabled in the **`AutocompleteArrayInputChip`** this element will be added.

This merge request adds a new property to **`AutocompleteArrayInput`** component, named **`allowDuplicates`**, that is used to allow or disallow insertion of duplicated elements.

Fixes #2311

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.js
@@ -159,12 +159,18 @@ export class AutocompleteArrayInput extends React.Component {
     };
 
     handleSuggestionSelected = (event, { suggestion, method }) => {
-        const { input } = this.props;
+        const { allowDuplicates, input } = this.props;
 
-        input.onChange([
-            ...(this.state.inputValue || []),
-            this.getSuggestionValue(suggestion),
-        ]);
+        const values = this.state.inputValue || [];
+        const newValue = this.getSuggestionValue(suggestion);
+
+        if (allowDuplicates || values.indexOf(newValue) < 0) {
+            // Add selected value to the list
+            input.onChange([...values, newValue]);
+
+            // Ensure to reset the filter
+            this.updateFilter('');
+        }
 
         if (method === 'enter') {
             event.preventDefault();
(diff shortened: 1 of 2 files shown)


tiagoschenkel:
Hi @djhi,

I could write some unit tests, but this would take more time.

Do you have in mind any specific test case?

For me is not 100% clear yet if this **`allowDuplicates`** property is a _feature_ or a _bug_.

Yes, I'm adding a new property thinking about keeping backward compatibility, but them this flag should be disabled by default since the current behavior of the **`AutocompleteArrayInput`** is to not filter out duplicates.

My question is, does it make sense to add a new flag as a feature? Or should be the case that this is not and intended behaviour, then we should handle this as a **`bug`** and simple get rid of duplicated elements without having to add a new _flag_.

Got it?

Best!

djhi:
> Do you have in mind any specific test case?

That it allows or not allows duplicates depending on the `allowDuplicates` prop ?

> My question is, does it make sense to add a new flag as a feature? Or should be the case that this is not intended behavior, then we should handle this as a **`bug`** and simply get rid of duplicated elements without having to add a new _flag_.

I'm sure there are cases where you want to allow duplicates so I think it's good to have a flag and it should be false by default

tiagoschenkel:
@djhi,

I'm trying to simulate an `onSuggestionSelected`event from `Autosuggest` component, to validate if the `onChange` method will be called or not based in the `allowDuplicates` property.

I'm not able to force a `select` event, so the `inputValue` state property is never changing.

My last two test cases are incomplete even when they are passing.

Could you help with this?

fzaninotto:
Superseded by #4026

PR #2914, opened 2019-02-22
Limit autocomplete height on AutocompleteArrayInput

tiagoschenkel:
The component **`AutocompleteArrayInput`** will render all the `choices` in a full container without scroll options. This is not a good idea when dealing with huge lists, as a list of possible countries in the world.

The property **`choicesMaxHeight`** will limit the size of the choices container, enabling the scrolling of the content.


fzaninotto:
Please reproduce what's been done on the `<AutocompleteInput>` component instead (limiting the choices paper to 50% of the screen).

tiagoschenkel:
Removed the parameter **`choicesMaxHeight`** and implemented same changes added to **`AutocompleteInput`** (#2415).

This pull request is now against **`master`** since would be considered as a bug fix instead of a new feature.

fzaninotto:
Thanks!

PR #2916, opened 2019-02-22
Richtext clean value after form submit

ggarcia92:
Allow reset value after submit just like other react-admin inputs.


fzaninotto:
Superseded by #2930

PR #2924, opened 2019-02-24
[RFR] Migrate Core Components to TypeScript

djhi:
- [x] Resource
- [x] RoutesWithLayout
- [x] CoreAdminRouter
- [x] CoreAdmin
- [x] helpers
- [x] Configure TS to output definitions


packages/ra-core/src/CoreAdmin.tsx
```diff
@@ -15,30 +10,41 @@ import withContext from 'recompose/withContext';
...
-    loading: ComponentType;
-    locale: string;
-    loginPage: ComponentType | boolean;
-    logoutButton: ComponentType;
-    menu: ComponentType;
-    theme: object;
-    title: ReactNode;
+    i18nProvider?: I18nProvider;
+    initialState?: object;
+    loading?: ComponentType;
```
fzaninotto: loading is required in CoreAdminRouter. We can provide an empty default tough.

packages/ra-core/src/CoreAdmin.tsx
```diff
@@ -15,30 +10,41 @@ import withContext from 'recompose/withContext';
...
+    locale?: string;
+    loginPage?: LoginComponent | boolean;
```
fzaninotto: the ? here is weird. It's either a component or false, but it can't be undefined. We should make it required and give it a false default value.

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+const welcomeStyles: CSSProperties = {
     width: '50%',
     margin: '40vh',
     textAlign: 'center',
 };
 
-export class CoreAdminRouter extends Component {
-    state = { children: [] };
+interface Props extends LayoutProps {
+    appLayout?: LayoutComponent;
```
fzaninotto: required

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+    catchAll?: CatchAllComponent;
```
fzaninotto: required

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+    children: AdminChildren;
+    customRoutes?: CustomRoutes;
+    loading?: ComponentType;
```
fzaninotto: required

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+}
+
+interface EnhancedProps {
+    authProvider?: AuthProvider;
+    isLoggedIn?: boolean;
```
fzaninotto: required

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+    userLogout: Dispatch<typeof userLogoutAction>;
+}
+
+interface State {
+    children: any[];
```
fzaninotto: why not use the AdminChildren type?

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
-    state = { children: [] };
+interface Props extends LayoutProps {
+    appLayout?: LayoutComponent;
+    catchAll?: CatchAllComponent;
+    children: AdminChildren;
+    customRoutes?: CustomRoutes;
+    loading?: ComponentType;
+}
+
+interface EnhancedProps {
```
fzaninotto: :+1: good practice

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -7,53 +15,82 @@ import getContext from 'recompose/getContext';
...
+    initializeResourcesAsync = async (props: Props & EnhancedProps) => {
         const { authProvider } = props;
         try {
             const permissions = await authProvider(AUTH_GET_PERMISSIONS);
             const { children } = props;
 
-            const childrenFuncResult = children(permissions);
-            if (childrenFuncResult.then) {
-                childrenFuncResult.then(resolvedChildren => {
+            if (typeof children === 'function') {
```
fzaninotto: no need to check here: it must be a function as initializeResourcesAsync is only called in initializeResources if it's a function.

packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -122,20 +159,22 @@ export class CoreAdminRouter extends Component {
...
-                Children.map(childrenToRender, child =>
-                    cloneElement(child, {
-                        key: child.props.name,
-                        // The context prop instructs the Resource component to not render anything
-                        // but simply to register itself as a known resource
-                        context: 'registration',
-                    })
+                Children.map(
+                    childrenToRender,
+                    (child: React.ReactElement<any>) =>
```
fzaninotto: We need a type here. We're passing a context props, we must check that the child accepts it

packages/ra-core/src/CoreAdmin.tsx
```diff
@@ -96,12 +102,12 @@ React-admin requires a valid dataProvider function to work.`);
             <TranslationProvider>
                 <ConnectedRouter history={this.history}>
                     <Switch>
-                        {loginPage !== false ? (
+                        {loginPage !== false && loginPage !== true ? (
```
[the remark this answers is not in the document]
djhi: It means it's not a boolean, as it does equals `true` nor `false` and that make TS happy for what follows. Agreed about the `undefined` case

packages/ra-core/src/CoreAdmin.tsx
djhi: Making it required means users **have** to provide it. I think it must stay optionnal and we may provide a default `false` value

packages/ra-core/src/CoreAdminRouter.tsx
djhi: Because the `children` property of the state is only used to store the result of the `children` **function** from the props. This function returns an array of `Resource` elements

packages/ra-core/src/CoreAdminRouter.tsx
djhi: This makes TS happy

packages/ra-core/src/CoreAdminRouter.tsx
djhi: You're aware that won't be used anywhere outside this component ?

packages/ra-core/src/types.ts
```diff
@@ -62,3 +66,32 @@ export interface ReduxState {
 export type Dispatch<T> = T extends (...args: infer A) => any
     ? (...args: A) => void
     : never;
+
+type RenderResourcesFunction = (permissions: any) => any[];
+export type AdminChildren = RenderResourcesFunction | ReactNode;
```
fzaninotto: ReactNode is a bit too broad here, we could use the Resource props and impose a `ReactElement<ResourceProps>`

packages/ra-core/src/CoreAdminRouter.tsx
djhi: Actually we even know it's an array of Resource elements

packages/ra-core/src/Resource.tsx
```diff
@@ -102,7 +135,8 @@ export class Resource extends Component {
                                     createElement(show, {
                                         basePath,
                                         id: decodeURIComponent(
-                                            props.match.params.id
+                                            (props.match as ResourceMatch)
```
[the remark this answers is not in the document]
djhi: How?

PR #2926, opened 2019-02-25
Fix for list not updating when filter prop is changed

### LINKED ISSUE #2923: List component does not update when filter property is updated
**What you were expecting:**
When dynamically updating the List `filter` property using a redux action the list data should immediately be updated.

**What happened instead:**
The list data is delayed in updating creating a confusing UX.

**Steps to reproduce:**
* [CodeSandbox](https://codesandbox.io/s/6x9ljr1q6r) and [demo](https://6x9ljr1q6r.codesandbox.io/#/users)
1. Sort by ID to generate query string params
2. Select "Group A" from the dropdown (no change to list)
3. Select "Group B" from the dropdown  (list is now filtered by "Group A")

If query string params are not present an error is thrown to the browser  console `Warning: Missing translation for key: "Cannot read property 'toLowerCase' of null"`.

**Other information:**
I think this is related to the `shouldUpdateComponent` function in the [ListController](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/ListController.tsx#L220) it seems like perhaps just adding the filter property to this list could fix the issue. However, I wasn't sure if it was intentionally excluded for performance reasons...

I think [this stack overflow issue](https://stackoverflow.com/questions/51378884/how-to-use-a-custom-reducers-state-as-a-permanent-filter-in-a-list) has the same root cause.

**Environment**

I have replicated this with two environments:

* React-admin version: 2.4.0, React version: 16.5.2
* React-admin version: 2.7.2, React version: 16.8.3

kara-todd:
fixes #2923

- ListController: update to use `componentDidUpdate` instead of `componentWillReceiveProps`
- ListController:  Add `filter` and `params` to list of props comparison values for `shouldComponentUpdate`

### DIFF
--- a/packages/ra-core/src/controller/ListController.tsx
+++ b/packages/ra-core/src/controller/ListController.tsx
@@ -198,7 +198,6 @@ export class UnconnectedListController extends Component<Props> {
         const {
             resource,
             query,
-            params,
             filter,
             sort,
             perPage,
@@ -212,12 +211,11 @@ export class UnconnectedListController extends Component<Props> {
             prevProps.query.page !== query.page ||
             prevProps.query.perPage !== query.perPage ||
             !isEqual(prevProps.query.filter, query.filter) ||
-            !isEqual(prevProps.params, params) ||
             !isEqual(prevProps.filter, filter) ||
             !isEqual(prevProps.sort, sort) ||
             !isEqual(prevProps.perPage, perPage)
         ) {
-            this.updateData(Object.keys(query).length > 0 ? query : params);
+            this.updateData(Object.keys(query).length > 0 && query);
         }
 
         if (prevProps.version !== version) {
--- a/packages/ra-core/src/controller/ListController.tsx
+++ b/packages/ra-core/src/controller/ListController.tsx
@@ -194,31 +194,34 @@ export class UnconnectedListController extends Component<Props> {
         this.setFilters.cancel();
     }
 
-    componentDidUpdate(prevProps: Props) {
-        const {
-            resource,
-            query,
-            filter,
-            sort,
-            perPage,
-            version,
-        } = this.props;
-
-        if (
-            prevProps.resource !== resource ||
-            prevProps.query.sort !== query.sort ||
-            prevProps.query.order !== query.order ||
-            prevProps.query.page !== query.page ||
-            prevProps.query.perPage !== query.perPage ||
-            !isEqual(prevProps.query.filter, query.filter) ||
-            !isEqual(prevProps.filter, filter) ||
-            !isEqual(prevProps.sort, sort) ||
-            !isEqual(prevProps.perPage, perPage)
-        ) {
-            this.updateData(Object.keys(query).length > 0 && query);
+    componentWillReceiveProps(nextProps: Props) {
+        if (!isEqual(nextProps.filter, this.props.filter)) {
+            const filtersWithoutEmpty = removeEmpty(this.getFilterValues());
+            this.changeParams({
+                type: SET_FILTER,
+                payload: filtersWithoutEmpty,
+            });
         }
 
-        if (prevProps.version !== version) {
+        if (
+            nextProps.resource !== this.props.resource ||
+            nextProps.query.sort !== this.props.query.sort ||
+            nextProps.query.order !== this.props.query.order ||
+            nextProps.query.page !== this.props.query.page ||
+            nextProps.query.perPage !== this.props.query.perPage ||
+            !isEqual(nextProps.query.filter, this.props.query.filter) ||
+            !isEqual(nextProps.filter, this.props.filter) ||
+            !isEqual(nextProps.sort, this.props.sort) ||
+            !isEqual(nextProps.perPage, this.props.perPage)
+        ) {
+            this.updateData(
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/controller/ListController.tsx
```diff
@@ -194,25 +194,34 @@ export class UnconnectedListController extends Component<Props> {
...
-            !isEqual(nextProps.filter, this.props.filter) ||
-            !isEqual(nextProps.sort, this.props.sort) ||
-            !isEqual(nextProps.perPage, this.props.perPage)
+            prevProps.resource !== resource ||
+            prevProps.query.sort !== query.sort ||
+            prevProps.query.order !== query.order ||
+            prevProps.query.page !== query.page ||
+            prevProps.query.perPage !== query.perPage ||
+            !isEqual(prevProps.query.filter, query.filter) ||
+            !isEqual(prevProps.query.filter, query.filter) ||
```
djhi: Duplicate ?

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -194,25 +194,34 @@ export class UnconnectedListController extends Component<Props> {
...
+            !isEqual(prevProps.params, params) ||
```
fzaninotto: you've added the params key here, it has nothing to do with the bug you mentioned. Please remove.

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -194,25 +194,34 @@ export class UnconnectedListController extends Component<Props> {
...
+            !isEqual(prevProps.filter, filter) ||
+            !isEqual(prevProps.sort, sort) ||
+            !isEqual(prevProps.perPage, perPage)
         ) {
-            this.updateData(
-                Object.keys(nextProps.query).length > 0
-                    ? nextProps.query
-                    : nextProps.params
-            );
+            this.updateData(Object.keys(query).length > 0 && query);
```
fzaninotto: no, you've changed the logic here. Please revert.

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -222,6 +231,12 @@ export class UnconnectedListController extends Component<Props> {
             nextProps.translate === this.props.translate &&
             nextProps.isLoading === this.props.isLoading &&
             nextProps.version === this.props.version &&
+            nextProps.filter === this.props.filter &&
+            nextProps.params.sort !== this.props.params.sort &&
```
fzaninotto: Same remark as above, don't add the test for params key as it has nothing to do with the bug you're solving.

packages/ra-core/src/controller/ListController.tsx
kara-todd: @fzaninotto this was due to the second bug I mentioned. If you apply the dynamic filter with no query params for example `/#/users` the params with defaults of `null` are called on the update data queries. This throws the `"Cannot read property 'toLowerCase' of null"` error.

Since the params data is merged with the defaults in the `this.getQuery()` I thought that just not passing the params in this case would still work and properly merge the defaults, but perhaps I misunderstood the logic here?

packages/ra-core/src/controller/ListController.tsx
kara-todd: @fzaninotto in my testing when I removed the params comparison on `shouldComponentUpdate` the list failed to immediately update with no query string present.  (`/#/users`)

This caused the sort filter to have to be clicked twice to be applied.

packages/ra-core/src/controller/ListController.tsx
kara-todd: whoops yes. sorry about that, thanks.

kara-todd:
@djhi @fzaninotto if the logic for the change is accepted I'm happy to add a test for it, but if it needs to be implemented differently I'd appreciate some direction on how you feel I should change it and still resolve the `'toLowerCase' of null` error.

packages/ra-core/src/controller/ListController.tsx
fzaninotto: with the previous logic, if there is no query string in the URL, we use the params from the Redux store, which are always defined. 

I don't know about the `Cannot read property 'toLowerCase' of null"` error, I suggest we deal with it in another PR, and you provide a failing test case to demonstrate it.

packages/ra-core/src/controller/ListController.tsx
fzaninotto: That means you have broken something in your PR that used to work before. Applications built with the current version have no problem displaying a resource list page without query string in the URL.

kara-todd:
@djhi @fzaninotto I've reverted the code as requested. This does expose the `
Cannot read property 'toLowerCase' of null"` error when changing filters, but that can be dealt with separately as requested.

packages/ra-core/src/controller/ListController.tsx
pashakopot: been fixing this 3 times in my projects, please add this to master someday )

pashakopot:
been fixing this 3 times in my projects, please add this to master someday )

djhi:
Superseded by #3308

Issue #2465, opened 2018-10-24
RichTextInput doesn't accept defaultValue prop


manolisan:
**What you were expecting:**
Using a component like that `<RichTextInput defaultValue="<p>Hello</p>" />` should result in a RichTextInput component having a prefiled value.

**What happened instead:**
The component renders without the default value.

**Related code:**
 `<RichTextInput defaultValue="Initial Value" />`

**Environment**

 - React-admin version: 2.3.3
 - React version: 16.3.2
 - Browser: Chrome 69.0

fzaninotto:
Confirmed and reproduced

fzaninotto:
in fact, it's a duplicate of #2464

manolisan:
I am not sure if accepting defaultValue prop will fix #2464 

#2464 is not about the default value of the RichTextInput  component. 

**Behavior of Input components on redux-change actions**
In general, on an Input component, when a redux-state change happens, the Input component should re-render it's value on React.

**Behavior of RichTextInput on redux-change actions caused by other components**
However if the componet is a RichTextInput and the redux-state change is caused by an action other than writing in that specific component _(e.g. writing on another RichTextInput component with the same source)_, it doesn't re-render it in React.

Let me know if it's clear or not, to provide you a code example in the next days.

PR #2930, opened 2019-02-26
Fix RichTextInput does not update when content is changed outside the editor

### LINKED ISSUE #2464: RichTextInput doesn't update on changes caused by redux-form actions
**What you were expecting:**
In one view with two **RichTextInput** components with the **same source**, changing the first one should change the value of the other and vise versa. 
However it is working as expected with two **TextInput** components with the **same source**, and they are in sync.

**What happened instead:**
Both RichTextInput component populate data on the same record in redux, but each component renders only what is typed within in it and not the value of the redux form.

**Steps to reproduce:**
Simply put two RichTextInput components with the same source in one SimpleForm.

**Other information:**

- Looking at the code, it appears that the Quill instance is populated with data only on mount. So, when new data comes asynchronously from REDUX, the RichTextInput keeps it's initial value.
 - Remounting a RichTextInput component leads to rendering the value of the redux form.

**Environment**

* React-admin version: 2.3.3
* React version: 16.3.2
* Browser: Chrome 69.0

fzaninotto:
Closes #2464, #2465

### DIFF
--- a/packages/ra-input-rich-text/src/index.js
+++ b/packages/ra-input-rich-text/src/index.js
@@ -49,6 +49,18 @@ export class RichTextInput extends Component {
         this.quill.on('text-change', debounce(this.onTextChange, 500));
     }
 
+    componentDidUpdate(prevProps) {
+        if (prevProps.input.value !== this.props.input.value) {
+            const selection = this.quill.getSelection();
+            this.quill.setContents(
+                this.quill.clipboard.convert(this.props.input.value)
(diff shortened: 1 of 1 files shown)


afilp:
@fzaninotto I am afraid that this fix caused another bug: If you type **spaces** inside the `RichTextInput` they are trimmed.

I created a new issue for this here: https://github.com/marmelab/react-admin/issues/3067

PR #2946, opened 2019-03-02
Fix button usage to show primary color

developerium:
Hi guys.

When I was going through documents, I noticed setting the primary color on button was outdated.
What do you think about this?


developerium:
Cool :+1:

Issue #2866, opened 2019-02-11
[RFC] Confirm Button decorator 


fzaninotto:
React-admin has an (undocumented') `<Confirm>` component, which lets developer open a confirmation dialog as follows:

```jsx
<Confirm
     isOpen={true}
     title="Delete Item"
     content="Are you sure you want to delete this item?"
     confirm="Yes"
     confirmColor="primary"
     cancel="Cancel"
     onConfirm={() => { // do something }}
     onClose={() => { // do something }}
/>
```

The problem is that developers still have to dispatch actions manually in their own `onConfim` handler. It's quite cumbersome when they just need to make a DeleteButton pessimistic.

I suggest transforming the `Confirm` component so that it accepts a button as child. When provided, this button *replaces* the confirmation button, and dispatches its own action - no need to pass an `onConfirm` handler. This `Confirm` button would handle its own open state, no need to pass it: 

```jsx
<Confirm
     title="Delete Item"
     content="Are you sure you want to delete this item?"
     confirm="Yes"
     confirmColor="primary"
     cancel="Cancel"
>
    <DeleteButton undoable={false} />
</Confirm>
```

When cloning the child, the `Confirm` button could even override the `undoable`, `label`, and `color` props... Also, as the API is a bit different than the (non-documented, but probably used) `Confirm` component, we should probably create another component to keep backwards compatibility.

Any opinion?

fzaninotto:
After playing with the idea, I think it's not a good one.

First, here is my implementation:

```jsx
import React, { Fragment, Component } from 'react';
import Confirm from './Confirm';

class WithConfirm extends Component {
    state = { isOpen: false };
    event = null;
    constructor(props) {
        super(props);
        this.button = React.createRef();
    }

    handleClick = e => {
        this.setState({ isOpen: true });
        e.stopPropagation();
    };

    handleDialogClose = () => {
        this.setState({ isOpen: false });
    };

    handleConfirm = () => {
        this.button.current.firstElementChild.click();
        this.setState({ isOpen: false });
    };

    render() {
        const { children, title, content, ...rest } = this.props;
        return (
            <Fragment>
                <span onMouseDown={this.handleClick}>
                    <span ref={this.button}>
                        {React.cloneElement(children, rest)}
                    </span>
                </span>
                <Confirm
                    isOpen={this.state.isOpen}
                    title={title}
                    content={content}
                    onConfirm={this.handleConfirm}
                    onClose={this.handleDialogClose}
                />
[...]

PR #2955, opened 2019-03-04
Non-undoable delete buttons show confirmation dialog

fzaninotto:
Refs #2866 

By default, the Save and Delete actions are undoable, i.e. react-admin only sends the related request to the data provider after a short delay, during which the user can cancel the action. This is part of the "optimistic rendering" strategy of react-admin ; it makes the user interactions more reactive.

You can disable this behavior by setting `undoable={false}`. With that setting, clicking on the Delete button displays a confirmation dialog. Both the Save and the Delete actions become blocking, and delay the refresh of the screen until the data provider responds.

```jsx
const PostEdit = props => (
    <Edit undoable={false} {...props}>
        ...
    </Edit>
```

**Tip**: If you want a confirmation dialog for the Delete button but don't mind undoable Edits, then pass a [custom toolbar](#toolbar) to the form, as follows:

```jsx
import {
    Toolbar,
    SaveButton,
    DeleteButton,
    Edit,
    SimpleForm,
} from 'react-admin';
import { withStyles } from '@material-ui/core';

const toolbarStyles = {
    toolbar: {
        display: 'flex',
        justifyContent: 'space-between',
    },
};

const CustomToolbar = withStyles(toolbarStyles)(props => (
    <Toolbar {...props}>
        <SaveButton />
        <DeleteButton undoable={false} />
    </Toolbar>
));

const PostEdit = props => (
    <Edit {...props}>
        <SimpleForm toolbar={<CustomToolbar />}>
            ...
        </SimpleForm>
    </Edit>
```

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -181,7 +181,7 @@ const Aside = ({ record }) => (
 
 By default, the Save and Delete actions are undoable, i.e. react-admin only sends the related request to the data provider after a short delay, during which the user can cancel the action. This is part of the "optimistic rendering" strategy of react-admin ; it makes the user interactions more reactive.
 
-You can disable this behavior by setting `undoable={false}`. Wit hthat setting, clicking on the Delete button displays a confirmation dialog. Both the Save and the Delete actions become blocking, and delay the refresh of the screen until the data provider responds.
+You can disable this behavior by setting `undoable={false}`. With that setting, clicking on the Delete button displays a confirmation dialog. Both the Save and the Delete actions become blocking, and delay the refresh of the screen until the data provider responds.
 
 ```jsx
 const PostEdit = props => (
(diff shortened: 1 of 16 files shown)


docs/CreateEdit.md
```diff
@@ -176,6 +177,53 @@ const Aside = ({ record }) => (
 
 **Tip**: Always test that the `record` is defined before using it, as react-admin starts rendering the UI before the API call is over.
 
+### Undoable
+
+By default, the Save and Delete actions are undoable, i.e. react-admin only sends the related request to the data provider after a short delay, during which the user can cancel the action. This is part of the "optimistic rendering" strategy of react-admin ; it makes the user interactions more reactive.
+
+You can disable this behavior by setting `undoable={false}`. Wit hthat setting, clicking on the Delete button displays a confirmation dialog. Both the Save and the Delete actions become blocking, and delay the refresh of the screen until the data provider responds.
```
djhi: s/Wit hthat/With that/

PR #2960, opened 2019-03-04
Fix multiple redirect events on auth error

### LINKED ISSUE #2939: Multiple redirect to login in case multiple requests to api presents on the page
**What you were expecting:**

When user is unauthorized and more than one request to api presents on the page, he should be redirected to login page only once

**What happened instead:**

When user is unauthorized and more than one request to API presents on the page, user will be redirected to login page multiple times (after first successful login, user will be again redirected to login form, and it will happens multiple times. Number of times exactly match the number of requests to API on the page)

**Related code:**

I believe that this happens because of this saga:
```
return function* watchAuthActions() {
        yield all([
            takeEvery(action => action.meta && action.meta.auth, handleAuth),
            takeEvery(FETCH_ERROR, handleAuth),
        ]);
    };
```

**Proposed solution:**

May be it is better to use `takeLatest` instead of `takeEvery` for authentication errors. If you are fine with this solution, I can create a PR

**Environment**

* React-admin version: 2.7.1
* React version: 16.6

fxzhukov:
Closes #2939 
**Proposed solution:**

Proposed solution: Use `takeLatest` instead of `takeEvery` saga effect to handle authentication errors.

### DIFF
--- a/packages/ra-core/src/sideEffect/auth.ts
+++ b/packages/ra-core/src/sideEffect/auth.ts
@@ -1,4 +1,11 @@
-import { all, put, call, select, takeEvery } from 'redux-saga/effects';
+import {
+    all,
+    put,
+    call,
+    select,
+    takeLatest,
+    takeEvery,
+} from 'redux-saga/effects';
 import { push, replace } from 'react-router-redux';
 
 import { AuthProvider } from '../types';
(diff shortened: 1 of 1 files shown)


fzaninotto:
I think you should use a `takeLeading` instead of a `takeLatest`. The latter will cancel previous calls while the former blocks future calls until the saga completes.

fxzhukov:
@fzaninotto there is one problem with `takeLeading` - we have redux-saga v0.16.0, but `takeLeading` was introduced only in v1.0.0:
https://github.com/redux-saga/redux-saga/releases
Not sure if i can update package to the major version without breaking anything

djhi:
I tried to upgrade `redux-saga` through yarn resolutions on a project and had no issues. It might impact other users though.

I think we should use `takeLatest` and note that we should migrate to `takeLeading` when releasing the next major version

fzaninotto:
Thanks!

PR #2964, opened 2019-03-05
Actually showing random image, instead of same image at /login

developerium:
Hi guys

At login, I always got the same image! I dug deeper and saw the problem was from [this address](https://source.unsplash.com/random/1600x900/daily). By removing `/daily`, you actually get random image.

What do you guys think?

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Login.js
+++ b/packages/ra-ui-materialui/src/auth/Login.js
@@ -154,7 +154,7 @@ Login.propTypes = {
 };
 
 Login.defaultProps = {
-    backgroundImage: 'https://source.unsplash.com/random/1600x900/daily',
+    backgroundImage: 'https://source.unsplash.com/random/1600x900',
     theme: defaultTheme,
     loginForm: <DefaultLoginForm />,
 };


Kmaschta:
Actually, it was totally random at first, but we changed this behavior since it was disturbing to some of our users.

NikitaVlaznev:
It would be great to provide a way to configure the default Login Page at least by passing background image url.

ashim-k-saha:
@NikitaVlaznev `backgroundImage` is the prop, you can use with login component.

JBaczuk:
@NikitaVlaznev You can pass it through admin component like this:

```
import { Admin, Login } from 'react-admin';

const MyLoginPage = () => <Login backgroundImage="/background.jpg" />;

const App = () => (
    <Admin loginPage={MyLoginPage}>
        // ...
    </Admin>
);
```

See [Using a Custom Login](https://marmelab.com/react-admin/Theming.html#using-a-custom-login-page)

NikitaVlaznev:
Thanks!

NikitaVlaznev:
Is it a supposed behavior that Custom login, Custom error and custom notification sections are not listed in the table of contents (Theming)?
https://marmelab.com/react-admin/Theming.html#using-a-custom-login-page

PR #2966, opened 2019-03-06
[RFR] Fix breaking change in ra-data-json-server

### LINKED ISSUE #2965: Breaking change in ra-data-json-server between 2.7.0 and 2.7.3
**What you were expecting:**
I am not sure if you're following Semver, but if you are - somewhere between version 2.7.0 and 2.7.3,  `ra-data-json-server` caused a breaking change, with [this commit](https://github.com/marmelab/react-admin/commit/429c4b606cacc2d3d85a8b1d61e1f8c5e99f31b1#diff-c3f1afe296439d9aaf3e711c8637b96b), because it no longer sends the `id_like` parameter, instead it sends the resources `id` field to get related resources.

Also, instead of a list of ids separated by a pipe (`|`) character, it now sends the same parameter twice when it wants to get more than one resource item, _e.g._ to get category `8` and category `203` it sends `http://localhost:9000/admin/categories?id=8&id=203` instead of `http://localhost:9000/admin/categories?id_like=8%7C203`

**What happened instead:**
Semver wasn't used.

**Steps to reproduce:**
Install the newest version 2.7.3 after being on 2.7.0.

**Related code:**
none

**Other information:**
none

**Environment**

* React-admin version: 2.7.3
* Last version that did not exhibit the issue (if applicable): 2.7.0
* React version: 16.8.4
* Browser: N/A

Kmaschta:
Fixes #2965

### DIFF
--- a/packages/ra-data-json-server/src/index.js
+++ b/packages/ra-data-json-server/src/index.js
@@ -82,6 +82,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
             case GET_MANY: {
                 const query = {
                     id: params.ids,
+                    [`id_like`]: params.ids.join('|'), // Legacy
                 };
                 url = `${apiUrl}/${resource}?${stringify(query)}`;
                 break;


packages/ra-data-json-server/src/index.js
```diff
@@ -82,6 +82,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
             case GET_MANY: {
                 const query = {
                     id: params.ids,
+                    [`id_like`]: params.ids.join('|'), // Legacy
```
djhi: I'm not sure this is a proper fix. APIs will receives a query with both filters. What are they supposed to do?

Kmaschta: Hadn't thought of that.
I'm not sure: should we revert the change and wait for the 3.0 to change the behavior, or adding a param and tell to users to ignore it?

Option 2: add a optional parameter to the data provider when to handle this change, like:
```js
jsonServerProvider('http://jsonplaceholder.typicode.com', { addIdParamToQuery: true })
```

fzaninotto:
As explained in the related issue, we don't think there is a bug to fix, so I'm closing this PR. Thanks for helping out anyway!

PR #2968, opened 2019-03-06
[RFR] Add contributors to the README

Kmaschta:
I just wanted to try the CLI, and I was thinking about adding it on react-admin, so here we are!

Didn't tried to categorize or anything, let me know if I forgot someone and so on.
Order is based on the contributions (see stats on GitHub)

## Preview

image


fzaninotto:
What's the purpose of adding contributors to the readme when GitHub already lists them elsewhere?

Kmaschta:
List contributors that helps the project without touching the code, like the reviewers, people that answer questions on SO, etc

djhi:
Why not if that can be updated automatically, otherwise I think we already have enough to maintain

Kmaschta:
It's helped by a CLI, but it requires a manual intervention :/

PR #2982, opened 2019-03-09
Prepare ra-ui-materialui Migration to TypeScript

djhi:
I locally added the `checkJs` option to the package `tsconfig.json` file (not commited). This gives us the opportunity to fix many issues before starting the migration to TS.

- [x] Use `createStyles`
- [x] Check `isValidElement` before cloning where necessary
- [x] Use `Children.only` before cloning when necessary
- [x] Update `material-ui` to `1.5.1` (which we already use on many projects anyway)
- [x] Fix all other possible errors


djhi:
175 errors left, from 235 when I started... :weary:

PR #2984, opened 2019-03-10
Migrate ra-ui-materialui to TypeScript

djhi:
- [x] auth


Kmaschta:
pray

fzaninotto:
The perimeter of this PR is ambitious - I'd prefer that you split the migration to avoid a large code review. As is, the perimeter of this PR is good enough - just get it working!

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -8,49 +8,46 @@ import {
...
-        margin: '1em',
-        display: 'flex',
-        justifyContent: 'center',
-    },
-    icon: {
-        backgroundColor: theme.palette.secondary[500],
-    },
-});
+interface Props {
+    backgroundImage: string;
```
fzaninotto: can be empty I believe

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -143,15 +139,13 @@ class Login extends Component {
     }
 }
 
+const Login = withStyles(styles)(LoginView) as ComponentType<Props>;
+
 Login.propTypes = {
```
fzaninotto: why do you keep the propTypes?

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -114,18 +104,24 @@ class Login extends Component {
...
+            loginForm,
+            ...rest
+        } = this.props;
 
         return (
             <MuiThemeProvider theme={this.theme}>
                 <div
                     className={classnames(classes.main, className)}
-                    {...sanitizeRestProps(rest)}
+                    {...rest}
```
fzaninotto: why did you remove the sanitization?

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -85,16 +79,12 @@ class Login extends Component {
     // the componentDidMount, but if the ref doesn't exist, it will try again
     // on the following componentDidUpdate. The try will be done only once.
     // @see https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element
-    updateBackgroundImage = (lastTry = false) => {
+    updateBackgroundImage = () => {
```
fzaninotto: why did you remove the lastTry? See the comment above the function.

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -143,15 +139,13 @@ class Login extends Component {
...
-    classes: PropTypes.object,
     className: PropTypes.string,
-    input: PropTypes.object,
     loginForm: PropTypes.element,
-    meta: PropTypes.object,
-    previousRoute: PropTypes.string,
+    theme: PropTypes.object,
 };
 
 Login.defaultProps = {
```
fzaninotto: migrate that to static in the component.

packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -39,7 +61,13 @@ const renderInput = ({
 const login = (auth, dispatch, { redirectTo }) =>
     dispatch(userLogin(auth, redirectTo));
 
-const LoginForm = ({ classes, isLoading, handleSubmit, translate }) => (
+const LoginFormView: SFC<Props & EnhancedProps> = ({
+    classes,
+    isLoading,
+    handleSubmit,
+    translate,
+    ...rest
```
fzaninotto: you don't do anything with the `rest` you added

packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -77,30 +105,35 @@ const LoginForm = ({ classes, isLoading, handleSubmit, translate }) => (
...
     withStyles(styles),
-    translate,
+    withTranslate,
     connect(mapStateToProps),
     reduxForm({
         form: 'signIn',
-        validate: (values, props) => {
-            const errors = {};
+        validate: (values: FormData, props: TranslationContextProps) => {
+            const errors = { username: '', password: '' };
```
fzaninotto: That means that both fields have an empty error, it's different from having no error.

packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -77,30 +105,35 @@ const LoginForm = ({ classes, isLoading, handleSubmit, translate }) => (
...
+            }
             return errors;
         },
     })
 );
 
-export default enhance(LoginForm);
+const LoginForm = enhance(LoginFormView);
+
+LoginForm.propTypes = {
```
fzaninotto: why is it still here?

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -1,44 +1,64 @@
...
+    redirectTo,
+    translate,
+    userLogout,
+    ...rest
+}) => (
     <MenuItem
         className={classnames('logout', classes.menuItem, className)}
         onClick={userLogout}
-        {...sanitizeRestProps(rest)}
+        {...rest}
```
fzaninotto: you should keep the rest sanitization

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -1,44 +1,64 @@
...
 /**
  * Logout button component, to be passed to the Admin component
  *
  * Used for the Logout Menu item in the sidebar
  */
-const Logout = ({ classes, className, translate, userLogout, ...rest }) => (
+const LogoutView: SFC<Props & EnhancedProps> = ({
+    classes,
+    className,
+    locale,
```
fzaninotto: not used

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -1,44 +1,64 @@
...
+    redirectTo,
```
fzaninotto: not used

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -1,44 +1,64 @@
...
+    translate,
```
fzaninotto: not used

packages/ra-ui-materialui/src/button/BulkDeleteButton.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { Component, MouseEvent } from 'react';
```
fzaninotto: please make sure that the last changes to that component were taken into account

packages/ra-ui-materialui/src/auth/Login.tsx
djhi: As discussed, Props interfaces are for TypeScript compilation, propTypes are for our users who are not using TypeScript.

packages/ra-ui-materialui/src/auth/Login.tsx
djhi: First, because this code does not work at all, read the original carefully and look where we were passing `lastTry`. Second because TypeScript agrees with me

packages/ra-ui-materialui/src/auth/Login.tsx
fzaninotto: That means that the type of the Props for that component is too restrictive, and will warn the user when they pass other props. You should extend the prop types of the child to whom you pass the rest props.

packages/ra-ui-materialui/src/auth/LoginForm.tsx
djhi: No, empty string is the same as no error

packages/ra-ui-materialui/src/auth/Logout.tsx
djhi: This is the rest sanitization

packages/ra-ui-materialui/src/auth/Logout.tsx
djhi: See previous answer

packages/ra-ui-materialui/src/auth/Logout.tsx
djhi: See previous answer

packages/ra-ui-materialui/src/auth/Logout.tsx
djhi: See next comment

packages/ra-ui-materialui/src/auth/LoginForm.tsx
djhi: As explained in a previous comment, TypeScript does not generate propTypes and not everyone use TypeScript. PropTypes are still valuable

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -8,49 +13,48 @@ import {
...
-    avatar: {
-        margin: '1em',
-        display: 'flex',
-        justifyContent: 'center',
-    },
-    icon: {
-        backgroundColor: theme.palette.secondary[500],
-    },
-});
+interface Props extends HtmlHTMLAttributes<HTMLDivElement> {
```
fzaninotto: we should not extend from `HtmlHTMLAttributes`, but use an union type in the component definition:

```
class LoginView extends Component<Props & WithStyles<typeof styles> & HtmlHTMLAttributes<HTMLDivElement>>
```

I think having the Props type extending nothing makes the intent of the component clearer.

packages/ra-ui-materialui/src/auth/Login.tsx
fzaninotto: then you should remove the comment above the method, too

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -114,18 +123,25 @@ class Login extends Component {
...
 
     render() {
-        const { classes, className, loginForm, ...rest } = this.props;
+        const {
+            backgroundImage,
+            classes,
+            className,
+            loginForm,
+            staticContext,
+            ...rest
```
fzaninotto: Some of the props that were in `sanitizeRestProps` are missing here

packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -39,7 +61,12 @@ const renderInput = ({
 const login = (auth, dispatch, { redirectTo }) =>
     dispatch(userLogin(auth, redirectTo));
 
-const LoginForm = ({ classes, isLoading, handleSubmit, translate }) => (
+const LoginFormView: SFC<Props & EnhancedProps> = ({
```
fzaninotto: I think SFC is deprecated in favor of FunctionalComponent

packages/ra-ui-materialui/src/auth/Login.tsx
djhi: Because they were unnecessary. I checked

packages/ra-ui-materialui/src/auth/Login.tsx
djhi: Why not

Issue #2761, opened 2019-01-11
Graphql version breaks webpack bundeling


dsinai14:
See the following issue:
When using graphql 0.13.2 as a dependency, it breaks the webpack bundeling and throw an error, for example when running with the new create-react-app.

An upgrasde to  graphql@^14.0.2 should solve this problem

[Runtime Error "require is not defined" when bundling with webpack](https://github.com/graphql/graphiql/issues/721)

djhi:
Fixed by #2993

PR #2993, opened 2019-03-13
Fix the graphql demo

### LINKED ISSUE #2990: graphql-demo does not work since v2.4.0
**What you were expecting:**

Try the demo with a GraphQL backend.

```sh
make run-graphql-demo
```

**What happened instead:**

Once you build & run the graphql-demo, at the first page load the graphql-js lib does not load properly.

At line : `/* WEBPACK VAR INJECTION */}.call(this, require("./../../process/browser.js")))`

**Steps to reproduce:**

```sh
git checkout v2.7.3
make install
make build
make run-graphql-demo
```

**Environment**

* React-admin version: 2.4.0 to 2.8.0
* Last version that did not exhibit the issue (if applicable): 2.3.4
* React version: 16.3.2
* NodeJS version : 10.15.3
* Browser: Firefox - Chrome
```js
ReferenceError: require is not defined 8.chunk.js:10937:1
```

djhi:
Fixes #2990

This was most probably caused by the CRA update which updated babel. Upgrading the graphql dependencies fixed it.

I also fixed the demo which had a new `invoices` resource added in the REST provider but not the graphql one.

### DIFF
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -8,7 +8,7 @@
         "data-generator": "^2.4.0",
         "fakerest": "~2.1.0",
         "fetch-mock": "~6.3.0",
-        "json-graphql-server": "~2.1.1",
+        "json-graphql-server": "~2.1.3",
         "ra-data-fakerest": "^2.0.0",
         "ra-data-graphql-simple": "^2.0.0",
         "ra-data-simple-rest": "^2.0.0",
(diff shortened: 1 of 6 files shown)


fzaninotto:
Does it also fix #2761?

djhi:
Yes, closed it

PR #2998, opened 2019-03-13
[RFR] Add German translations for ra-tree-ui-materialui

straurob:
This PR adds German translations for the `ra-tree-ui-materialui` package.

https://github.com/marmelab/react-admin/tree/master/packages/ra-tree-ui-materialui


fzaninotto:

We won't be able to maintain this package, as we don't speak German, so we won't add it to the react-admin sources.

I advise you to publish it as a standalone package under your name, and add a link to it in the react-admin Ecosystem documentation.

PR #3001, opened 2019-03-14
Fix ra-data-graphql-simple handling of UPDATE mutations

### LINKED ISSUE #2776: ra-data-graphql-simple UPDATE_MANY incorrect build variable {data}
**What you were expecting:**
When creating the BulkArchiveButton, I expected correct execution mutation

**What happened instead:**
UPDATE_MANY a few times causes the execution of the UPDATE
the required ID argument is not passed to the {data} parameter

**Steps to reproduce:**
try to update any resource through the CustomBulkActionButton

**Other information:**
Wrote the BulkArchiveButton as in the example in the documentation:

```
class BulkArchiveButton extends Component {
    handleArchive = () => {
        const {  basePath, resource, selectedIds, startUndoable } = this.props;
        startUndoable(
            crudUpdateMany(resource, selectedIds, { isArchved: true }, basePath)
        );
    };
    render() {
        const { className, classes, filterValues } = this.props;
        return (
            <Button 
                size="small"
                variant="outlined"
                label="raExt.action.archive"
                onClick={this.handleArchive}
                className={`${classes.button} ${classes.archiveButton} ${className}`}
            >
                <ArchiveIcon />
            </Button>
        );
    }
}
```

This is the request this button sends:
default

That's what gets a response:
default

To solve the problem fixed 

> C:\...\ui-react-admin\node_modules\ra-data-graphql-simple\esm\index.js

```
...
               if (fetchType === UPDATE_MANY) {
                var ids = params.ids, otherParams_2 = __rest(params, ["ids"]);
                return Promise.all(params.ids.map(function (id) {
                    
                 //return defaultDataProvider(UPDATE, resource, __assign({ id: id }, otherParams_2));
                    return defaultDataProvider(UPDATE, resource, {data:__assign({ id: id }, otherParams_2.data)});

                })).then(function (results) {
                    var data = results.reduce(function (acc, _a) {
                        var data = _a.data;
                        return acc.concat([data.id]);
                    }, []);
                    return { data: data };
                });
            }
...
```
Please, fix for next release.
**Environment**

* React-admin version: 2.6.0

djhi:
- [x] Fix `ra-data-graphql-simple` handling of `UPDATE` mutations
- [x] Fix error messages still referring to aor
- [x] Fix type in `ra-data-graphql` variable name
- [x] Fix graphql demo as it needs to override `DELETE` requests

Fixes #2776

### DIFF
--- a/examples/demo/src/dataProvider/graphql.js
+++ b/examples/demo/src/dataProvider/graphql.js
@@ -1,35 +1,67 @@
-import buildApolloClient from 'ra-data-graphql-simple';
+import buildApolloClient, { buildQuery as buildQueryFactory } from 'ra-data-graphql-simple';
+import { DELETE } from 'ra-core';
+import gql from 'graphql-tag';
+const getGqlResource = resource => {
+    switch (resource) {
+        case 'customers':
+            return 'Customer';
+
+        case 'categories':
+            return 'Category';
+
+        case 'commands':
+            return 'Command';
+
+        case 'products':
+            return 'Product';
+
+        case 'reviews':
+            return 'Review';
+
+        case 'invoices':
+            return 'Invoice';
+
+        default:
+            throw new Error(`Unknown resource ${resource}`);
+    }
+};
+
+const customBuildQuery = introspectionResults => {
+    const buildQuery = buildQueryFactory(introspectionResults);
+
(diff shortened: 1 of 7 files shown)


wmwart:
Thank you, it works!

PR #3004, opened 2019-03-14
Fix AutocompleteInput crashes when a choice doesn't have a string value

### LINKED ISSUE #2991: AutocompleteInput throws an error if resource[optionText] isn't a string
Hi guys

I noticed in [here](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/AutocompleteInput.js#L181), if the result field from api response is null, (usually when first input is put in page), then `suggestionLabel.toString()` will throw an error saying null doesn't have method `toString`.

What do you think about this? Should we put a default filter to avoid being null, or convert to string differently?

developerium:
Closes #2991


fzaninotto:
Thanks!

developerium:
No problem @fzaninotto

PR #3011, opened 2019-03-16
[RFR] Update documentation with German translation package

straurob:
This PR adds a link to the translation section. It refers to a new standalone package for the German translations of the ra-tree-ui-materialui package as propsed in PR #2998.

PR #3014, opened 2019-03-17
[RFR] Update documentation with German translation package

straurob:
This PR adds a link to the translation section. It refers to a new standalone package for the German translations of the ra-tree-ui-materialui package as propsed in PR #2998.

Due to problems with my local repo I had to recreate the fork. Thus, PR #3011 is obsolete and can be continued here.


fzaninotto:
hm, this one is even worse than the previous one...

straurob:
Let's see if the latest change looks better. I'm sorry... this is embarassing :-|

fzaninotto:
Yep, that one is all right. Thanks!

straurob:
@fzaninotto 
I just stumpled upon this. This PR was added to the [2.8.2 milestone](https://github.com/marmelab/react-admin/tree/v2.8.2) and is also listed in the corresponding changelog. However I don't see the commit in the actual file.

https://github.com/marmelab/react-admin/blob/v2.8.2/docs/Translation.md 

Could you check this?

fzaninotto:
ah, it's been pushed against `next` rather than `master`. I didn't pay attention before merging. So this will appear in the next minor release.

straurob:
Ah ok, I understand. Thanks for the clarification.

PR #3021, opened 2019-03-18
fix(Resource): Added exact to all <Route />

### LINKED ISSUE #2910: [Bug] resourcesRoutes list is sensitive to ordering
We have the following `resourcesRoutes`:

```jsx
    <Resource
      name="project-management/projects"
      list={ProjectList}
      edit={ProjectEdit}
      create={ProjectCreate}
      show={ProjectShow}
    />,
    <Resource
      name="project-management/projects/services"
      list={ProjectServiceList}
      edit={ProjectServiceEdit}
      create={ProjectServiceCreate}
      show={ProjectServiceShow}
    />,
```

This fail to generate a working `<Route />`, because `project-management/projects/services` will catch before `project-management/projects`.

Instead we must order it DESC :  

```jsx
    <Resource
      name="project-management/projects/services"
      list={ProjectServiceList}
      edit={ProjectServiceEdit}
      create={ProjectServiceCreate}
      show={ProjectServiceShow}
    />,
    <Resource
      name="project-management/projects"
      list={ProjectList}
      edit={ProjectEdit}
      create={ProjectCreate}
      show={ProjectShow}
    />,
```

This will make a valid `<Route />`.

It can also be solved using `exact` props, and `customRoutes`, but this is too sensitive for being an expected behaviour, and it is not specified anywhere in the documentation.

If you agree, this ticket consists of taking all the user props from `<Resource />` and pass them to the `<Route />` component.

https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/Resource.tsx#L80

kopax:
Fix #2910 

This will prevent any resource route to be selected not intentionally.

### DIFF
--- a/packages/ra-core/src/Resource.tsx
+++ b/packages/ra-core/src/Resource.tsx
@@ -72,6 +72,7 @@ export class Resource extends Component<ResourceProps & ConnectedProps> {
         }
 
         const resource = {
+            exact: true,
             resource: name,
             options,
             hasList: !!list,
--- a/packages/ra-core/src/Resource.tsx
+++ b/packages/ra-core/src/Resource.tsx
@@ -72,7 +72,6 @@ export class Resource extends Component<ResourceProps & ConnectedProps> {
         }
 
         const resource = {
-            exact: true,
             resource: name,
             options,
             hasList: !!list,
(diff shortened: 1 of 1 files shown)


kopax:
I see, is that because `<Route />` props within `<Resource />` will not be used as it is declared? (`cloneElement`).

I am surprised, I will close this one and prepare another PR with the recommended change.

PR #3026, opened 2019-03-20
fix(resourcesRoutes): Added prop exact to all resources routes

### LINKED ISSUE #2910: [Bug] resourcesRoutes list is sensitive to ordering

kopax:
Fix #3021  
Fix #2910

### DIFF
--- a/packages/ra-core/src/Resource.tsx
+++ b/packages/ra-core/src/Resource.tsx
@@ -86,6 +86,7 @@ export class Resource extends Component<ResourceProps & ConnectedProps> {
             <Switch>
                 {create && (
                     <Route
+                        exact
                         path={`${match.url}/create`}
                         render={routeProps => (
                             <WithPermissions
@@ -103,6 +104,7 @@ export class Resource extends Component<ResourceProps & ConnectedProps> {
                 )}
                 {show && (
                     <Route
+                        exact
                         path={`${match.url}/:id/show`}
                         render={routeProps => (
                             <WithPermissions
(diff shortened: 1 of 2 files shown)


Kmaschta:
Well, well! E2E tests are failing, especially on with custom pages.

kopax:
Ok, I have no clue yet, I'll have a look now

kopax:
@djhi I have investigated and I found out the fix your recommended was not actually the right one.
Adding `exact` here will create a `<Route path={`/${child.props.name}`} />`

image

What we want is to set the `exact` props to the 4 crud route, not just this one. I offer to restore the 1st fix.

fzaninotto:
As seen in #2910, we won't address that issue.

Issue #2428, opened 2018-10-13
AutocompleteInput pop-up after choice is selected


Ken-Lim:
**What you were expecting:**
After choice is selected, AutocompleteInput filled the textInput with selected choice

**What happened instead:**
After choice is selected, AutocompleteInput filled the textInput with selected choice then pop-up all the choices again

**Steps to reproduce:**
1. Goto demo site, to edit an order (e.g. https://marmelab.com/react-admin-demo/#/commands/280)
2. Enter a customer name in the Customer field
3. Select the choice with your mouse
4. All choices were pop-up again

Please see the attached GIF for the issue
![autocomplete](https://user-images.githubusercontent.com/26237537/46902325-230aff00-cef6-11e8-8c14-0c733d5ad249.gif)

fzaninotto:
Confirmed. Weird bug.

Ken-Lim:
> Hello @Ken-Lim and @fzaninotto,
> I'm still experiencing this issue with an Autocomplete inside a Reference Input, with react-admin 2.4.0 and react-autosuggest ^9.4.2 in package-lock.json.
> 
> Little note : The ReferenceInput is outside a Create/Edit page so using a SimpleForm and overriding the form prop
> 
> ```
>  render() {
>     const { deviceId } = this.state;
> 
>     return (
>         <SimpleForm form="replace-device-form" resource="Device" save={this.handleSubmit}>
>             <ReferenceInput
>                 label="Test issue 2428"
>                 source="deviceId"
>                 reference="Device"    
>                 onChange={(event, deviceId) => console.log(`New deviceId ${deviceId}`)}
>             >
>                 <AutocompleteInput optionText="identification" />
>             </ReferenceInput>
>         </SimpleForm>
>     )
> }
> ```
> 
> Any idea ? Upgrading to 9.4.2 has fixed the issue for you ?

I upgraded to react-admin 2.4.0, which uses the react-autosuggest 9.4.2.
Unfortunately, the issue is not resolve for me.

FACOLOMBANI:
i fixed this issue by adding focusInputOnSuggestionClick={false} props to the AutoSuggest component.

fzaninotto:
reopening as the issue seems not fixed

Oupsla:
Forwarding message for example of the issue

> Here : https://codesandbox.io/s/lrro0q4wnm
> 
> Go to the custom page, then write something (like omnis) in the autocomplete input and then select something (and the selection container will open again)

[...]

PR #3031, opened 2019-03-20
[RFR] New AutocompleteInput using Downshift

### LINKED ISSUE #3519: Reimplement AutocompleteInput and AutocompleteArrayInput using downshift
The V3 Roadmap (#3194) include a rewrite of `<AutocompleteInput>` and `<AutocompleteArrayInput>` using  `downshift`.

Here is the current PR implementation: #3031 

This issue purpose is to list all problems the new implementation should fix:

- #2961 The suggestion list is wrongly placed when a second search as few results
- #2428 The suggestion list appears again after a suggestion is selected
- #3098 The `<AutocompleteInput>` doesn't work in a `<ReferenceInput>`
- #3148 The selected value is always the first suggested item (even if it shouldn't)

ThieryMichel:
Fixes #3519, #2428

- [x] Implementation
- [x] Tests
- [x] Upgrade Guide
- [x] Documentation

### DIFF
--- a/cypress/integration/edit.js
+++ b/cypress/integration/edit.js
@@ -88,9 +88,6 @@ describe('Edit Page', () => {
             .clear()
             .type('Sed quo');
         cy.get('[role="tooltip"]').within(() => {
-            cy.contains(
-                'Accusantium qui nihil voluptatum quia voluptas maxime ab similique'
-            );
             cy.contains('Sed quo et et fugiat modi').click();
         });
         cy.get('[role="tooltip"]').should(el => expect(el).to.not.exist);
@@ -104,9 +101,7 @@ describe('Edit Page', () => {
             .clear()
             .type('architecto aut');
         cy.get('[role="tooltip"]').within(() => {
-            cy.contains('Sed quo et et fugiat modi');
             cy.contains('Sint dignissimos in architecto aut');
-            cy.contains('A voluptas eius eveniet ut commodi dolor');
         });
     });
 
--- a/examples/simple/src/posts/PostEdit.js
+++ b/examples/simple/src/posts/PostEdit.js
@@ -3,6 +3,7 @@ import React from 'react';
 import {
     TopToolbar,
     AutocompleteArrayInput,
+    AutocompleteInput,
     ArrayInput,
     BooleanInput,
     CheckboxGroupInput,
@@ -19,6 +20,7 @@ import {
     NumberInput,
     ReferenceArrayInput,
     ReferenceManyField,
+    ReferenceInput,
     SelectInput,
     SimpleFormIterator,
     TabbedForm,
@@ -65,6 +67,14 @@ const PostEdit = props => (
                 <ImageInput multiple source="pictures" accept="image/*">
                     <ImageField source="src" title="title" />
                 </ImageInput>
+                <ReferenceInput
+                    label="User"
+                    source="user_id"
+                    reference="users"
+                    allowEmpty
+                >
+                    <AutocompleteInput />
+                </ReferenceInput>
             </FormTab>
             <FormTab label="post.form.body">
                 <RichTextInput
--- a/examples/simple/src/posts/PostEdit.js
+++ b/examples/simple/src/posts/PostEdit.js
@@ -29,6 +29,7 @@ import {
     minValue,
     number,
     required,
+    FormDataConsumer,
 } from 'react-admin'; // eslint-disable-line import/no-unresolved
 import PostTitle from './PostTitle';
 
@@ -43,7 +44,7 @@ const EditActions = ({ basePath, data, hasShow }) => (
     </TopToolbar>
 );
 
-const PostEdit = props => (
+const PostEdit = ({ permissions, ...props }) => (
     <Edit title={<PostTitle />} actions={<EditActions />} {...props}>
         <TabbedForm defaultValue={{ average_note: 0 }}>
             <FormTab label="post.form.summary">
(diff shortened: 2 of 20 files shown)


Kmaschta:
Be aware that there is an open issue with this component, it would be awesome if this PR might help. https://github.com/marmelab/react-admin/issues/2428

fzaninotto:
This may be an improvement, but it's also a BC break, since additional props were passed to react-autosuggest. This can't be merged before 3.0.

komik966:
This autocomplete looks cleaner :) I tried it but this problem: https://github.com/mui-org/material-ui/issues/13179 still remains.

This improvement not affects much code of ra - maybe it's possible to add it as new Autocomplete and deprecate actual one before v3 release? or publish it as package (e.g. https://github.com/fizix-io/ra-customizable-datagrid)?

examples/simple/src/posts/PostEdit.js
```diff
@@ -65,6 +68,49 @@ const PostEdit = props => (
...
                 </ImageInput>
+                {permissions === 'admin' && (
+                    <ArrayInput source="authors">
+                        <SimpleFormIterator>
+                            <ReferenceInput
+                                label="User"
+                                source="user_id"
+                                reference="users"
+                            >
+                                <AutocompleteInput />
```
fzaninotto: an `ArrayInput` listing `AutocompleteInput`?? Why didn't you use an `AutocompleteArrayInput`?

packages/ra-ui-materialui/src/input/AutocompleteInput/getSuggestions.spec.js
```diff
@@ -0,0 +1,83 @@
+import getSuggestions from './getSuggestions';
+
+describe('getSuggestions', () => {
+    const choices = [
+        { id: 1, value: 'one' },
+        { id: 2, value: 'two' },
+        { id: 3, value: 'three' },
+    ];
+
+    it('should return suggestions based on choices', () => {
```
fzaninotto:
```suggestion
    it('should return all suggestions when filtered by empty string', () => {
```

packages/ra-ui-materialui/src/input/AutocompleteInput/getSuggestions.spec.js
```diff
@@ -0,0 +1,83 @@
...
+                allowEmpty: false,
+                optionText: 'value',
+                getSuggestionText: ({ value }) => value,
+                optionValue: 'id',
+                limitChoicesToValue: true,
+            })('**o')
+        ).toEqual([{ id: 1, value: '**one' }]);
+    });
+
+    it('should not filter choices based on second argument if limitChoicesToValue is false', () => {
```
fzaninotto: I don't understand this one

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
```diff
@@ -0,0 +1,396 @@
...
+ * <AutocompleteInput source="author_id" options={{ fullWidth: true }} />
+ */
+const AutocompleteInput = ({
+    classes: classesOverride,
+    allowEmpty,
+    choices,
+    fullWidth,
+    helperText,
+    label,
+    limitChoicesToValue,
```
fzaninotto: this one isn't documented and I don't know what it means

cypress/integration/edit.js
```diff
@@ -88,9 +88,6 @@ describe('Edit Page', () => {
             .clear()
             .type('Sed quo');
         cy.get('[role="tooltip"]').within(() => {
-            cy.contains(
```
fzaninotto: this is a regression

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
```diff
@@ -0,0 +1,396 @@
...
+            if (setFilter) {
+                setFilter(value);
+            }
+        },
+        [setFilter]
+    );
+
+    useEffect(() => {
+        updateFilter('');
+    }, [input.value, updateFilter]);
```
fzaninotto: I don't understand this one. Each time the value changes, the filter resets? Why? A JsDoc comment would be welcome

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
```diff
@@ -0,0 +1,396 @@
...
+
+    const getSuggestionTextFromValue = value => {
+        const currentChoice = choices.find(
+            choice => getSuggestionValue(choice) === value
+        );
+
+        return getSuggestionText(currentChoice);
+    };
+
+    const handleSuggestionSelected = suggestionText => {
```
fzaninotto: needs memoization with `useCallback`, just like most of the functions defined in this component

fzaninotto: I'm not comfortable with using the text to find the suggestion. In the previous version, this handler received the suggestion *object*, so there was no need to search for it. 

I'm thinking about cases where multiple suggestions may have the same suggestion text.

packages/ra-ui-materialui/src/input/AutocompleteInput/getSuggestions.spec.js
djhi: I suspect a regression here

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
djhi: It is in the current version: https://marmelab.com/react-admin/Inputs.html#autocompleteinput

> If you want to limit the initial choices shown to the current value only, you can set the limitChoicesToValue prop.

cypress/integration/edit.js
djhi: Yes, we talked about it and you said ok for it

examples/simple/src/posts/PostEdit.js
djhi: Because this is not the only field inside this `ArrayInput`. See the `Role` field below. However, I believe this was introduced as a way to manually test the `AutoCompleteInput` by @ThieryMichel and is not used in any automated tests. Maybe we can remove it ?

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
djhi: Good catch. Fixed

packages/ra-ui-materialui/src/input/AutocompleteInput/index.js
djhi: I added a comment about it

fzaninotto:
Argh! needs rebase.

packages/ra-ui-materialui/src/input/AutocompleteInput/getSuggestions.spec.js
djhi: Fixed

cypress/integration/edit.js
fzaninotto: Gotcha

examples/simple/src/posts/PostEdit.js
fzaninotto: we may remove it in the future, let's keep it for now

packages/ra-ui-materialui/package.json
```diff
@@ -49,6 +49,7 @@
         "classnames": "~2.2.5",
         "connected-react-router": "^6.4.0",
         "css-mediaquery": "^0.1.2",
+        "downshift": "3.2.7",
```
fzaninotto: `react-autosuggest` isn't removed because we still have AutocompletearrayInput, right?

djhi: Yes, we'll have to rewrite this component too

PR #3032, opened 2019-03-20
add missing dependencies

cortopy:
I've added a few dependencies missing from packages. Not having these dependencies breaks installations with pnpm, requiring [a fix on behalf of the developer](https://pnpm.js.org/docs/en/faq.html#pnpm-does-not-work-with-your-project-here)


fzaninotto:
what's pnpm? What's the error and the project setup?

djhi:
pnpm is an alternative to npm and yarn: https://github.com/pnpm/pnpm

fzaninotto:
Well, if it's not compatible with npm and yarn, I don't see why we should make changes in the react-admin codebase to support it.

kopax:
@fzaninotto pnpm is a replacement, a 3rd package manager choice for all developers (like yarn). We use npm for that reason, even if yarn is more convenient on many points.

I found it was strange that **pnpm** force us to list in `package.json` dependencies that are not used directly in the code base and:

**ra-core**

- `history`: https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/util/TestContext.tsx#L7

**ra-ui-materialui**

- `react-router`: https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/Layout.js#L5
- `react-router-redux`: https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/DatagridRow.js#L5

This should be merged, and a fix to prevent this would be to use https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

cortopy:
@fzaninotto it is compatible with npm and yarn. But it was created with two aims in mind:
- performance
- solve doppelgangers that both npm and yarn suffer from 

It's the default package manager in [rush](https://rushjs.io/pages/advanced/npm_doppelgangers/) for these reasons

cortopy:
@kopax I don't use those plugins as I have come to rely on pnpm to tell me about phantom dependencies. Could add it if needed though

fzaninotto:
@kopax thanks for the clarification and the links, now I understand. 

@cortopy I'd gladly merge another PR adding the linter rule to prevent such problems in the future.

packages/ra-core/package.json
```diff
@@ -36,6 +36,7 @@
         "cross-env": "^5.2.0",
         "enzyme": "~3.7.0",
         "enzyme-adapter-react-16": "~1.6.0",
+        "history": "4.7.2",
```
fzaninotto: Please use a more tolerant dependency, like `^4.7.2`

cortopy: done

PR #3036, opened 2019-03-21
Fixed wrong scalar types into variables for ra-data-graphql-simple

Artexoid:
The fix for GraphQL version 14.0.0+ related scalar value coercion. The current version doesn't work correctly. GraphQL changes: https://www.libupdate.com/libs/1c88bd1e-024d-41aa-9241-c29cd6d5f0c5

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildVariables.js
+++ b/packages/ra-data-graphql-simple/src/buildVariables.js
@@ -210,7 +210,7 @@ export default introspectionResults => (
     params,
     queryType
 ) => {
-    const preparedParams = prepareParams(params, queryType);
+    const preparedParams = prepareParams(params, queryType, introspectionResults);
 
     switch (aorFetchType) {
         case GET_LIST: {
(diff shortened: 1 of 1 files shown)


djhi:
Sorry, the tests are currently a PITA to maintain. I'll have to clear some of them.

Artexoid:
@djhi Also, I fixed a little bug for it

Artexoid:
> Thanks! Can you please run prettier on this file? There is a make command for it, and a yarn one if you can't run make

Ok, I'll do now

djhi:
Btw, I'm afraid this will be a breaking change for those who haven't upgraded their graphql package. Can you confirm ?

Artexoid:
> Btw, I'm afraid this will be a breaking change for those who haven't upgraded their graphql package. Can you confirm ?

@djhi It works well for me. I tested it with graphql@0.13.2 and graphql@14.1.1 . The update related only with scalar values coercion to correctly types. 

Started from version 14.0.0 graphQL server throws an error if the client send in variables wrong type. For example on the server args

{
  id: Int
}

And the client send

{
  "id": "24"
}

It will work before version 14.0.0, but started from version 14.0.0 the server will return error "wrong type".

Just need to complete tests with green result :-)

Artexoid:
It will works for old GraphQL packages too

fzaninotto:
Thanks!

PR #3042, opened 2019-03-22
Support URI for ImageField source

djdmbrwsk:
I was trying to use `<ImageField>` with a full URI, but was unable to.  ~After looking at the source it looks like this should be possible with lists, but there is a bug.~  For single images, it looks like it's not possible.  So two commits here:

- ~Fix src default for lists~
- Try to default src for non-lists

**Update:**

Looks like my understanding of the list logic was wrong so I revered that commit. I also updated the full URI option to be more backwards compatible and added a test.

**Side Note:** The docs on contributing might need to be updated. I'm having a hard time getting all of the tests to pass locally. Getting a handful of `ra-data-graphql` failures.

### DIFF
--- a/packages/ra-ui-materialui/src/field/ImageField.js
+++ b/packages/ra-ui-materialui/src/field/ImageField.js
@@ -1,8 +1,9 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import get from 'lodash/get';
-import { withStyles, createStyles } from '@material-ui/core/styles';
+import { createStyles, withStyles } from '@material-ui/core/styles';
 import classnames from 'classnames';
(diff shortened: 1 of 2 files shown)


djhi:
I suggest you open a question on StackOverflow first and explain your use case.

djdmbrwsk:
Good question! Our resources don't have a full image URLs readily available in them; they have file IDs. So I thought it would be nice to write a custom field component that composes the full URL from the file ID and forwards it along to a wrapped `<ImageField>`.

I guess I mostly want to wrap `<ImageField>` to benefit from some of the styles/logic.  In this case they're pretty minimal, so I guess I'll copy/paste.

PR #3043, opened 2019-03-22
Fix custom actions cannot override basePath

kopax:
Previously, it was impossible to write: `<Show basePath="/foo" actions={<ShowActions basePath="/foo/bar" />} />` because we were not tolerating `props` on `DefaultActions`, making the API/props useless/broken.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Create.js
+++ b/packages/ra-ui-materialui/src/detail/Create.js
@@ -72,6 +72,7 @@ export const CreateView = ({
                         basePath,
                         resource,
                         hasList,
+                        ...actions.props,
                     })}
                 </CardContentInner>
             )}
--- a/packages/ra-ui-materialui/src/detail/Edit.js
+++ b/packages/ra-ui-materialui/src/detail/Edit.js
@@ -92,6 +92,7 @@ export const EditView = ({
                             hasShow,
                             hasList,
                             resource,
+                            ...actions.props
                         })}
                     </CardContentInner>
                 )}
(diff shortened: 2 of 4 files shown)


kopax:
To be clear, `cloneElement` does keep props set it's written [here](https://reactjs.org/docs/react-api.html#cloneelement), but `createElement` 2nd arguments will override any props set beforehand, making impossible to customize `basePath`, resource and so on.

packages/ra-ui-materialui/src/detail/Create.js
```diff
@@ -72,6 +72,7 @@ export const CreateView = ({
                         basePath,
                         resource,
                         hasList,
+                        ...actions.props,
```
kopax: In here `basePath`, `resource`, `hasList` cannot be customized without it.

packages/ra-ui-materialui/src/detail/Edit.js
```diff
@@ -92,6 +92,7 @@ export const EditView = ({
                             hasShow,
                             hasList,
                             resource,
+                            ...actions.props
```
kopax: In here `basePath`, `resource`, `hasList`, `hasShow`, `data` cannot be customized without it.

packages/ra-ui-materialui/src/detail/Show.js
```diff
@@ -86,6 +86,7 @@ export const ShowView = ({
                             hasList,
                             hasEdit,
                             resource,
+                            ...actions.props
```
kopax: In here `basePath`, `resource`, `hasList`, `hasEdit`, `data` cannot be customized without it.

djhi:
Ah I see. Fair point!

djhi:
There are probably many places where this should be done too

fzaninotto:
Thanks!

PR #3046, opened 2019-03-23
[RFR] Allow <Query> to pass all results from dataProvider

Kmaschta:
As explained on [the documentation](https://marmelab.com/react-admin/DataProviders.html#response-format), the dataProvider can return two keys `data` and `total`:

image

As of now, `<Query>` can only return `data`, and this is very restrictive, especially when dealing with lists or references.

This PR aims to pass the prop `total` with the data, but also test and document the usage of <Query> with lists.

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -349,6 +349,34 @@ const UserProfile = ({ record }) => (
 ```
 {% endraw %}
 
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks!

PR #3047, opened 2019-03-23
[RFR] Replace withDataProvider by Query on demo dashboard

Kmaschta:
Replace all the dataprovider usage on the demo dashboard by the `<Query>` component.

**Pros**
- Make the code more readable
- Suggests a better pattern to new users
- Test the `<Query>` component on a "real" usage

**Cons**
- Do a lot more requests (but can be memoized)
- No longer suggest how to use the `withDataProvider` HOC

### DIFF
--- a/examples/demo/src/dashboard/CustomerAvatar.js
+++ b/examples/demo/src/dashboard/CustomerAvatar.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { Query, GET_ONE } from 'react-admin';
+
+import Avatar from '@material-ui/core/Avatar';
+
+const CustomerAvatar = ({ record, classes }) => {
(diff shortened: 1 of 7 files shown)


fzaninotto:
Not good: you're replacing 2 queries with 1+N queries in the orders list. This is the reason why I used `withDataProvider` in the dashboard in the first place.

Kmaschta:
I agree, and I was also wondering if it was a real issue or not.
Since react-admin cache results in its redux store (it's often the same requests), what do we show as examples in the demo: performance vs. API?

fzaninotto:
Both: performance in the dashboard, API in the review moderation buttons.

Kmaschta:
Ok, my bad!

PR #3053, opened 2019-03-25
doc(WithPermissions): Update example, using render instead of children

kopax:
https://stackoverflow.com/questions/55297559/react-admin-what-is-authparams-for?answertab=active#tab-top


djhi:
Thanks!

Issue #3040, opened 2019-03-22
Confirmation message for deletion does not pick up resource translation


straurob:
**What you were expecting:**
When deleting an item I expect the confirmation message to pick up the translated resource name. In the below screenshot that would be `Staat` or `Staaten`.

**What happened instead:**
The confirmation message shows the untranslated resource name `staat`.

![image](https://user-images.githubusercontent.com/18437455/54822093-74d56380-4ca4-11e9-9d0a-1708aadb3f68.png)

**Steps to reproduce:**
1. Setup translation messages (for the locale `de`) in `i18n.js`.
2. Provide a resource named `staat` (or any other name).
3. Open the list view.
4. Select one or more entries.
5. Click "Delete".

```
module.exports = {
  de: {
    resources: {
      staat: {
        name: "Staat |||| Staaten",
        fields: {
          // Fields go here...
        }
      }
    }
  }
}
```
**Other information:**
As far as I can see this seems to be related to `BulkDeleteWithConfirmButton.js` in the `ra-ui-materialui` package.

**Environment**
* React-admin version: 2.8.2
* React version: 16.8.4
* Browser: Firefox 66.0

PR #3058, opened 2019-03-26
[RFR] Pluralize the confirmation message in BulkDeleteWithConfirmButton

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/3040
Follows https://github.com/marmelab/react-admin/pull/2955
 
We use to singularize the confirm message in the `BulkDeleteWithConfirmButton` component. But we often select more than one component. The solution is to use [inflect](https://github.com/dreamerslab/node.inflection#inflectioninflect-str-count-singular-plural-) to select wisely between pluralize or singularize.

## Todo

- [x] Use `inflect` instead of `singularize` in the `BulkDeleteWithConfirmButton` component

Sélection_005

Sélection_006

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.js
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.js
@@ -106,7 +106,7 @@ class BulkDeleteWithConfirmButton extends Component {
                     translateOptions={{
                         smart_count: selectedIds.length,
                         name: inflection.humanize(
-                            inflection.singularize(resource),
+                            inflection.inflect(resource, selectedIds.length),
                             true
                         ),
                     }}


packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.js
```diff
@@ -106,7 +106,7 @@ class BulkDeleteWithConfirmButton extends Component {
                     translateOptions={{
                         smart_count: selectedIds.length,
                         name: inflection.humanize(
-                            inflection.singularize(resource),
+                            inflection.inflect(resource, selectedIds.length),
```
djhi: I think we should do even better. Many of users actually provide translations for their resources names. We should use them

djhi: See #3040

fzaninotto: Good idea, but this should be done across all messages. Let's merge @Luwangel's PR and do that in another PR.

PR #3061, opened 2019-03-27
Easier action override

### LINKED ISSUE #2905: RFC: Easier Save Action Customization
## API

```jsx
const handleSave = (values: any, redirect: RedirectionSideEffect) => {
    // I can override the redirect
    // I can override/add/remove data
    // I can return a custom action, it will be dispatched for me

    return myCustomAction(values, redirect);
}


const CreatePost = props => (
    <Create onSave={handleSave}>
        ....
    </Create>
);
```

If I return an action, it will be dispatched for me, avoiding the redux connect boilerplate.
If I don't return anything, nothing will be done for me. I may dispatch myself and/or do something else.

## Use cases

- I need to override the form values before sending them to my dataProvider. The [current way](https://marmelab.com/react-admin/Actions.html#altering-the-form-values-before-submitting) to do it is way too verbose and complicated.
- I need to change the redirect depending on the form values
- I need to close a dialog/drawer which contain my Create/Edit component once it is saved

## Details on current solution

Today, you have to first implement a custom connected button which already implies a lot of boilerplate, for example (and this does not even includes the code needed for a custom redux action):

```jsx
import { crudCreate, SaveButton, Toolbar } from 'react-admin';

// A custom action creator which modifies the values before calling the default crudCreate action creator
const saveWithNote = (values, basePath, redirectTo) =>
    crudCreate('posts', { ...values, average_note: 10 }, basePath, redirectTo);

class SaveWithNoteButtonView extends Component {
    handleClick = () => {
        const { basePath, handleSubmit, redirect, saveWithNote } = this.props;

        return handleSubmit(values => {
            saveWithNote(values, basePath, redirect);
        });
    };

    render() {
        const { handleSubmitWithRedirect, saveWithNote, ...props } = this.props;

        return (
            <SaveButton
                handleSubmitWithRedirect={this.handleClick}
                {...props}
            />
        );
    }
}

const SaveWithNoteButton = connect(
    undefined,
    { saveWithNote }
)(SaveWithNoteButtonView);
```

Then, you have to create a custom toolbar to use it:

```jsx
const PostCreateToolbar = props => (
    <Toolbar {...props}>
        <SaveButton
            label="post.action.save_and_show"
            redirect="show"
            submitOnEnter={true}
        />
        <SaveWithNoteButton
            label="post.action.save_with_average_note"
            redirect="show"
            submitOnEnter={false}
            variant="flat"
        />
    </Toolbar>
);
```

And finally, you have to pass this custom toolbar to the `SimpleForm` through the `toolbar` prop.

djhi:
Fixes #2905

### DIFF
--- a/packages/ra-core/src/controller/CreateController.spec.tsx
+++ b/packages/ra-core/src/controller/CreateController.spec.tsx
@@ -4,27 +4,27 @@ import { shallow } from 'enzyme';
 import { UnconnectedCreateController as CreateController } from './CreateController';
 
 describe('CreateController', () => {
-    describe('Presetting the record from the location', () => {
-        const defaultProps = {
-            basePath: '',
-            crudCreate: jest.fn(),
-            hasCreate: true,
-            hasEdit: true,
-            hasList: true,
-            hasShow: true,
-            isLoading: false,
-            location: {
-                pathname: '/foo',
-                search: undefined,
-                state: undefined,
-                hash: undefined,
-            },
-            match: { isExact: true, path: '/foo', params: undefined, url: '' },
-            resource: 'foo',
-            title: 'Foo',
-            translate: x => x,
-        };
+    const defaultProps = {
+        basePath: '',
+        dispatch: jest.fn(),
+        hasCreate: true,
+        hasEdit: true,
+        hasList: true,
+        hasShow: true,
+        isLoading: false,
+        location: {
+            pathname: '/foo',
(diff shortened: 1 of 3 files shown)


djhi:
I'm not sure we should allow our users to override all the action creators in our controllers. I think it only make sense for a few. Actually, I don't see any other candidate than the `save` method of the Create and Edit controllers

PR #3064, opened 2019-03-28
fix: expand icon overlaps actions in tree view

macrozone:
On certain material-ui versions, the expand icon stretches the whole TreeNode width, because we set it to `left: ...` but material-ui sets it to `right: ...`.

I am force to use 3.9.2, but this was already the case in the used version 1.4.0.

This is fixed in material-ui 4 in this commit: https://github.com/mui-org/material-ui/commit/7b0f7d5afae78303c43545573bd1433094921e7e

until then, setting `right: auto` on the expandIcon in the TreeNode will fix it for every version


fzaninotto:
thanks!

PR #3065, opened 2019-03-28
[RFR] Fix typescript warnings

### LINKED ISSUE #2978: Typescript Import Errors
**What you were expecting:**
Expected that a Typescript based react-admin website that had no Typescript compile errors would continue to not have any compile errors when react-admin was updated from version 2.7.2 to 2.7.3.

**What happened instead:**
An import statement like:
`import { Layout } from 'react-admin';` 

began displaying errors like:

`Module '".../web/node_modules/react-admin/esm"' has no exported member 'Layout'.`

**Other information:**
I believe this behavior was introduced in PR #2924 when the following change was made to the react-admin/package.json file:

`"declaration": true,`

Apparently, since the `ra-ui-materialui` package is not yet producing types, any types found in that package that are imported in Typescript cause errors. 

**Environment**

* React-admin version: 2.7.3
* Last version that did not exhibit the issue (if applicable): 2.7.2
* React version: 16.3.2
* Browser: Chrome Version 72.0.3626.121

fzaninotto:
Closes #2978

### DIFF
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -29,6 +29,7 @@
     "devDependencies": {
         "@types/history": "^4.7.2",
         "@types/node-polyglot": "^0.4.31",
+        "@types/react-redux": "^7.0.5",
         "@types/react-router": "^4.4.4",
         "@types/react-router-dom": "^4.3.1",
         "@types/recompose": "^0.27.0",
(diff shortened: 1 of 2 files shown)


fzaninotto:
I don't understand these TypeScript errors and I can't reproduce them in my IDE. :cry:

Bnaya:
you need to check global typecheck, the IDE won't show them,
tsc -p tsconfig.json --noEmit

fzaninotto:
Superseded by #3643

PR #3069, opened 2019-03-29
With 'daily' suffix there is one picture all the time

NikitaVlaznev:
Without 'daily' it's random.

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Login.js
+++ b/packages/ra-ui-materialui/src/auth/Login.js
@@ -155,7 +155,7 @@ Login.propTypes = {
 };
 
 Login.defaultProps = {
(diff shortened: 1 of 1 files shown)


Kmaschta:
Duplicate of #2964

But I agree, unslash is bugging :(

fzaninotto:
When you write "all the time", do you mean during more than one day? could it be because you're behind an aggressive proxy / CDN ?

Kmaschta:
@fzaninotto It's been weeks that unslash didn't changed the image at `https://source.unsplash.com/random/1600x900/daily`, and that every react admin have now the same login background.


unslash daily image

fzaninotto:
I've sent an email to the Unsplash support to determine of the feature is temporarily disabled or broken.

fzaninotto:
This seems to be fixed, closing

PR #3083, opened 2019-04-02
[RFR] Fix Export button ignores permanent filter from List

### LINKED ISSUE #2933: Exporting CSV file doesn't include the permanent filter
**What you were expecting:**
Exported CSV file to take into account the permanent filter of the List component.

**What happened instead:**
CSV file includes all entries, regardless of the permanent filter.

**Steps to reproduce:**
Create a List component with a permanent filter. For example 
`<List {...props} filter={{id: 1}} >`
Download CSV file by using the export button.

**Environment**

* React-admin version: 2.7.0
* React version: 16.4.1
* Browser: Chrome

I know I can probably achieve this by creating a custom exporter function. If this is intended behavior, it should be mentioned in the documentation.

fzaninotto:
Closes #2933

### DIFF
--- a/packages/ra-ui-materialui/src/list/List.js
+++ b/packages/ra-ui-materialui/src/list/List.js
@@ -102,6 +102,7 @@ export const ListView = ({
     // component props
     actions,
     aside,
+    filter,
     filters,
     bulkActions, // deprecated
     bulkActionButtons,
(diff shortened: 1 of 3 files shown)


djhi:
Merged too quickly. Isn't it missing some documentation?

PR #3091, opened 2019-04-03
[RFR] Migrate ra-ui-materialui to TypeScript (Part 2: fields)

Kmaschta:
Wish me luck.

## TODO
- [x] Migrate fields components to TS
- [x] Use `Identifier` instead of string everywhere an `id` is used
- [X] Use the `withTranslate` from Gildas' PR


packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -55,7 +66,7 @@ export const BooleanField = ({
         return (
             <Typography
                 component="span"
-                body1="body1"
```
Kmaschta: This was a common bug. Bad copy/paste

djhi:
I'll work on the remaining fields in a separate PR

packages/ra-ui-materialui/src/field/types.ts
```diff
@@ -0,0 +1,33 @@
...
+    label?: string;
+    sortable?: boolean;
+    className?: string;
+    cellClassName?: string;
+    headerClassName?: string;
+    textAlign?: TextAlign;
+}
+
+// Props injected by react-admin
+export interface InjectedFieldProps {
```
djhi: I think this should be exported by `ra-core`

djhi:
I'm going to revert the changes on all the Reference fields components. I believe they should leverage types declared but not yet exported from `ra-core`.

packages/ra-ui-materialui/src/detail/SimpleShowLayout.spec.js
```diff
@@ -12,7 +12,7 @@ describe('<SimpleShowLayout />', () => {
                 <TextField source="bar" />
             </SimpleShowLayout>
         );
-        const inputs = wrapper.find('pure(TextField)');
+        const inputs = wrapper.find('EnhancedTextField');
```
djhi: By setting the displayName on the `EnhancedTextField`, we can avoid testing implementation details here

Kmaschta:
There are still some `Component.propTypes` in this PR, are they still needed?

djhi:
Actually I added them back. Yes they are necessary for run time check for people not using TS

fzaninotto:
needs rebase following changes on BooleanFields

PR #3093, opened 2019-04-03
Prevent ra-data-graphql-simple from circular dependencies issue

GM-Alex:
The path prevents the `buildFields` method from circular dependencies, so that `linkedTypes` are only processed if they aren't  analyzed before.


GM-Alex:
@djhi Done. I made a new one but I think we could extend the old one because they are nearly identically.

fzaninotto:
Thanks!

PR #3096, opened 2019-04-04
[RFR] Turn data generator for demo into a public repository

fzaninotto:
## Rationale

- When learning react-admin, developers often need a fake API. jsonplaceholder isn't perfect (it doesn't accept mutations), so developers use JSONServer or json-graphql-server, but they need data.
- The demo doesn't run on CodeSandbox as is because of the dependency to an internal package. Having the demo run in CodeSandbox would ease bug reports
- Who doesn't need fake data?

### DIFF
--- a/examples/data-generator/package.json
+++ b/examples/data-generator/package.json
@@ -1,7 +1,9 @@
 {
-    "name": "data-generator",
+    "name": "data-generator-retail",
     "version": "2.7.1",
-    "private": true,
+    "homepage": "https://github.com/marmelab/react-admin/tree/master/examples/data-generator",
+    "bugs": "https://github.com/marmelab/react-admin/issues",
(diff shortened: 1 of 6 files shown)


examples/data-generator/package.json
```diff
@@ -1,7 +1,9 @@
 {
-    "name": "data-generator",
+    "name": "data-generator-retail",
```
fzaninotto: the name `data-generator` is already taken

djhi: `ra-data-generator` ?

fzaninotto: well, we may publish data generators for blogs, social network, etc in the future. Also, it's in react-admin for now, but honestly nothing forbids that it lives on its own. It's just more convenient to keep it here for now.

Issue #3073, opened 2019-03-31
ra-input-rich-text breaks when using special characters 


JakobRobijns:
**What you were expecting:**

I wanted to use  RichTextInput from 'ra-input-rich-text' like normal use. 

**What happened instead:**

When trying to insert some Spanish content text _("ARTÍCULO 1.- El presente reglamento es de observancia obligatoria")_, and make the "ARTÍCULO" bold (or another action), it keeps creating paragrafs.
After searching why, it's the 'Í' that breaks it. (It's a variant of the normal 'I' used in Spanish.

Edit: It's not only the 'I', other characters also break it.

**Steps to reproduce:**

-Create an react-admin application and use RichTextInput from 'ra-input-rich-text'.
-Try inserting content with the letter 'Í'.
-Make it bold/italic/...

**Related code:**

`
import RichTextInput from 'ra-input-rich-text';
`
`            <RichTextInput source="description" />
`

**Other information:**

[GIF of the bug](https://i.imgur.com/DJvbvOS.gif)

**Environment**

* React-admin version:  2.8.4
* Last version that did not exhibit the issue (if applicable): /
* React version: 16.8.5
* Browser: Firefox & Chrome
* Stack trace (in case of a JS error):

Kmaschta:
We are using the latest version of `quill@1.3.6` for this rich text component.
Mind you check that this bug is related to their code or to our implementation of the rich text?

Thanks

JakobRobijns:
@fzaninotto Tnx. I also tried it on other quill plugins (React-Quill Demo - https://codepen.io/alexkrolick/pen/xgyOXQ)  & (https://github.com/zenoamaro/react-quill) And I can't reproduce it.

roychoo:
@JakobRobijns @fzaninotto i think this is due to the integration of Quill,
mainly 2 issues i think,

1. after conversion of text to delta, quill seems to add p tag for sentences that has special characters
2. due to that, and the way that it is checking whether it is required to update the component, it goes into an infinite loop of adding p tags.

i have made a fix but i am not sure if that is the best way.

PR #3099, opened 2019-04-05
Fix unnecessary update of ra-input-rich-text on edit

roychoo:
Hi,

in ra-input-rich-text,

whenever the text changes, below is the sequence of events happening.
Original text `<p>test</p>`
1. User change text to `<p>test1</p>`
2. `onTextChange` fired
3. `componentDidUpdate` fired
4. update happens as previous prop value is still `<p>test</p>`, but unecessary
5. `onTextChange` fired again.
6. stops there as the prevProp input value is the same as current prop input value

2nd issue is that, it seems quilljs is always adding additional `<p>` after this method
```
this.quill.setContents(
     this.quill.clipboard.convert(this.props.input.value)
)
```
***ONLY*** for the text that has special characters, i.e ```ARTÍCULO 1.- El presente reglamento es de observancia obligatoria``` after you bold ```ARTÍCULO 1```

fixes https://github.com/marmelab/react-admin/issues/3073

### DIFF
--- a/packages/ra-input-rich-text/src/index.spec.js
+++ b/packages/ra-input-rich-text/src/index.spec.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import debounce from 'lodash/debounce';
+import {render, fireEvent, waitForElement, cleanup, getByTestId} from 'react-testing-library'
+
+import { RichTextInput } from './index';
+
+let container;
+
+jest.mock('lodash/debounce');
+describe('RichTextInput', () => {
+    beforeEach(() => {
+      container = document.createElement('div');
+      document.body.appendChild(container);
+      document.getSelection = () => {
+          return {
+            removeAllRanges: () => {},
+            getRangeAt: function() {}, 
+          };
+      };
+      jest.mock('lodash.debounce', (fn) => fn);
+    });
+
+    afterEach(() => {
+        document.body.removeChild(container);
+        container = null;
+    });
+
+    it('should call text-change event only once when editing', async () => {
+        const mockFn = jest.fn();
+        debounce.mockImplementation(fn => fn);
+        const { getByTestId, rerender } = render(
+        <RichTextInput
+            input={{
+              value: '<p>test</p>',
+                onChange: mockFn
+            }}
+            meta={{error: null}} />);
+        const quillNode = await waitForElement(() => {
+            return getByTestId('quill')
+        });
+        quillNode.__quill.setText('test1');
+        rerender(
+          <RichTextInput
+            input={{
+              value: '<p>test1</p>',
+                onChange: mockFn
+            }}
+            meta={{error: null}} />)
+
+        expect(mockFn).toHaveBeenCalledTimes(1);
+    });
+})
--- a/packages/ra-input-rich-text/src/index.spec.js
+++ b/packages/ra-input-rich-text/src/index.spec.js
@@ -27,13 +27,13 @@ describe('RichTextInput', () => {
 
     it('should call handleChange only once when editing', async () => {
         jest.useFakeTimers();
-        const mockFn = jest.fn();
+        const handleChange = jest.fn();
         debounce.mockImplementation(fn => fn);
         const { getByTestId, rerender } = render(
         <RichTextInput
             input={{
               value: '<p>test</p>',
-                onChange: mockFn
+                onChange: handleChange
             }}
             meta={{error: null}} />);
         const quillNode = await waitForElement(() => {
@@ -51,10 +51,11 @@ describe('RichTextInput', () => {
           <RichTextInput
             input={{
               value: '<p>test1</p>',
-                onChange: mockFn
+                onChange: handleChange
             }}
             meta={{error: null}} />)
 
-        expect(mockFn).toHaveBeenCalledTimes(1);
+        // jest.runOnlyPendingTimers();
+        expect(handleChange).toHaveBeenCalledTimes(1);
     });
 })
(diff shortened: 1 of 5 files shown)


djhi:
Thanks! It seems the tests are failing though

roychoo:
Hi just want to check, did anyone take the opportunity to review the code?

test-setup.js
```diff
@@ -1,4 +1,5 @@
 require('raf/polyfill');
+require('mutationobserver-shim');
```
fzaninotto: MutationObserver is used nowhere in the tests, why did you add this?

packages/ra-input-rich-text/src/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    beforeEach(() => {
+      container = document.createElement('div');
+      document.body.appendChild(container);
+      document.getSelection = () => {
+          return {
+            removeAllRanges: () => {},
+            getRangeAt: function() {}, 
+          };
+      };
+      jest.mock('lodash.debounce', (fn) => fn);
```
fzaninotto: you mocked debounce twice

packages/ra-input-rich-text/src/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+import { RichTextInput } from './index';
+
+let container;
+
+jest.mock('lodash/debounce');
+describe('RichTextInput', () => {
+    beforeEach(() => {
+      container = document.createElement('div');
+      document.body.appendChild(container);
+      document.getSelection = () => {
```
fzaninotto: what's that for? Please comment the code to explain why you need to do that.

packages/ra-input-rich-text/src/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      jest.mock('lodash.debounce', (fn) => fn);
+    });
+
+    afterEach(() => {
+        document.body.removeChild(container);
+        container = null;
+    });
+
+    it('should call text-change event only once when editing', async () => {
+        const mockFn = jest.fn();
```
fzaninotto: call it handleChange

packages/ra-input-rich-text/src/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+        debounce.mockImplementation(fn => fn);
```
fzaninotto: it seems to be the same implementation as in the beforeEach

packages/ra-input-rich-text/src/index.spec.js
```diff
@@ -0,0 +1,52 @@
...
+        <RichTextInput
+            input={{
+              value: '<p>test</p>',
+                onChange: mockFn
+            }}
+            meta={{error: null}} />);
+        const quillNode = await waitForElement(() => {
+            return getByTestId('quill')
+        });
+        quillNode.__quill.setText('test1');
```
fzaninotto: If you're using react-testing-library, this is not the way to go. You should mimick the actions of a real user, not use internal APIs.

test-setup.js
roychoo: mutationobserver is used in quilljs

roychoo: i will put in some comments on the ones that require due to quilljs

packages/ra-input-rich-text/src/index.spec.js
roychoo: actually, i did tried the normal user way, but quilljs is using `contenteditable` and not input, thus not sure how to simulate a change event. Do let me know if you have any ideas.

packages/ra-input-rich-text/src/index.spec.js
roychoo: will remove this

packages/ra-input-rich-text/src/index.spec.js
roychoo: done

packages/ra-input-rich-text/src/index.spec.js
roychoo: the one in beforeEach is not working, forget to remove it. will remove it

test-setup.js
fzaninotto: Then the polyfill should be added by quilljs themselves, or am I missing something?

packages/ra-input-rich-text/src/index.spec.js
fzaninotto: You can focus() on the element, then send key events. See https://stackoverflow.com/questions/6139107/programmatically-select-text-in-a-contenteditable-html-element

test-setup.js
roychoo: i think in default jest uses jsdom and jsdom do not have support for mutationobserver.

https://github.com/jsdom/jsdom/issues/639

fzaninotto: Right, so you should add that comment to the code so that other developers know why it's included

packages/ra-input-rich-text/src/index.spec.js
roychoo: hmm, just realised fireEvent actually works, it was due to the event loop. Had made the changes

test-setup.js
roychoo: done.

packages/ra-input-rich-text/src/index.spec.js
djhi: Are you sure ? I still see `mockFn`

roychoo: Oh sorry I misunderstood, I thought you meant the test description. Will made the required change in another 3 hours.

roychoo: @djhi done.

djhi:
Thanks!

roychoo:
Thanks for the review and merged

PR #3102, opened 2019-04-05
[RFR] Relax material ui version dependency

fzaninotto:
This allows usage of material ui v3, which is backwards compatible with v1 (Firefox 45 aside).

I tested the simple and demo example: they work perfectly. 

Now I regret we didn't do this earlier (!)

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInput.spec.js
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInput.spec.js
@@ -1,7 +1,7 @@
 import React from 'react';
-import assert from 'assert';
-import { shallow } from 'enzyme';
+import expect from 'expect';
 import { CheckboxGroupInput } from './CheckboxGroupInput';
+import { render, cleanup } from 'react-testing-library';
 
 describe('<CheckboxGroupInput />', () => {
     const defaultProps = {
@@ -13,41 +13,12 @@ describe('<CheckboxGroupInput />', () => {
             value: [],
         },
         translate: x => x,
-        muiTheme: {
-            baseTheme: {},
-            textField: {},
-            prepareStyles: () => {},
-        },
     };
 
-    it('should use a mui Checkbox', () => {
-        const wrapper = shallow(<CheckboxGroupInput {...defaultProps} />);
-        const CheckboxElement = wrapper
-            .find('WithStyles(FormControlLabel)')
-            .shallow()
-            .dive()
-            .find('WithStyles(Checkbox)');
-        assert.equal(CheckboxElement.length, 1);
-    });
+    afterEach(cleanup);
 
-    it('should use the input parameter value as the initial input value', () => {
-        const wrapper = shallow(
-            <CheckboxGroupInput
-                {...defaultProps}
-                input={{ value: [1], onChange: () => {} }}
-            />
-        );
-        const CheckboxElement = wrapper
-            .find('WithStyles(FormControlLabel)')
-            .shallow()
-            .dive()
-            .find('WithStyles(Checkbox)')
-            .first();
-        assert.equal(CheckboxElement.prop('checked'), true);
-    });
-
-    it('should render choices as mui Checkbox components', () => {
-        const wrapper = shallow(
+    it('should render choices as checkbox components', () => {
+        const { getByLabelText } = render(
             <CheckboxGroupInput
                 {...defaultProps}
                 choices={[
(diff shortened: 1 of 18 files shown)


fzaninotto:
Some tests fail because we tested the implementation rather than the feature. Time to switch to react-testing-library!

oliviertassinari:
> Time to switch to react-testing-library

👍 We are migrating from the shallow to the mount API of enzyme in Material-UI.

maicWorkGithub:
Material-UI V4 is coming......

djhi:
> Material-UI V4
It's only an alpha for now. We have time :)

oliviertassinari:
v4.0.0-beta.0 should be out this weekend, v4.0.0 should be out before the end of the month.

dzienisz:
@fzaninotto hey, when you will merge this?

ricardovf:
Tried to install it using **yarn add marmelab/react-admin#3102/head** with no success, can you help?

yarn add v1.13.0
[1/4] 🔍  Resolving packages...
error Can't add "react-admin-lerna": invalid package version undefined.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.

djhi:
@ricardovf You can't install a monorepo this way

fzaninotto:
I know that many people are waiting for this one. We’re actually working on it, but it has no release date.

andrewkslv:
@fzaninotto did you test capability on production build? I have run into the issue when classNames have wrong order after refreshing the page.

I created a demo repo for it here https://github.com/andrewkslv/react-admin-with-material-ui-next
Production website http://dead-ticket.surge.sh

fzaninotto:
@andrewkslv which issue? I don't understand your point. 

And no, we haven't tested that on production yet, as you can see the CI build is still red, we're working on it.

andrewkslv:
@fzaninotto thank you!

I recorded a video with the issue. https://i.imgur.com/xgJQZZj.mp4

fzaninotto:
The remaining warnings are deprecation warnings for material-ui. I hope they don't occur in production builds.

fzaninotto:
@andrewkslv I don't know what you're supposed to have and what you're having instead. Please use the issue template to describe your problem in a new issue.

andrewkslv:
@fzaninotto thanks.

packages/ra-ui-materialui/src/input/CheckboxGroupInput.spec.js
```diff
@@ -13,298 +13,191 @@ describe('<CheckboxGroupInput />', () => {
...
-                    translateChoice={false}
-                    meta={{ touched: false }}
+                    meta={{ touched: false, error: 'Required field.' }}
                 />
             );
-            const FormHelperTextElement = wrapper.find(
-                'WithStyles(FormHelperText)'
-            );
-            assert.equal(FormHelperTextElement.length, 0);
+            expect(container.querySelector('p')).toBeNull();
```
djhi: You're testing implentation details here. If the root tag changes for some reason the test will fails but the user won't actually see anything. Those tests may become a pain to maintain. I believe testing the `Required field.` is not visible is enough

djhi: You might use `FormHelperTextProps` to pass a `testid` prop if you really really want this

fzaninotto: This PR is already a great step in the right direction. I know I'm still testing implementation details every once in a while, but that's also because material-ui makes it somewhat hard. 

If the tests about helper text break in the future, I'll fix them. At least, tests about values and label won't break anymore. 

Many people are waiting for this change, so I suggest we merge it as is.

fzaninotto:
I'm sorry but I'll have to revert that change. There are 2 reasons:

1. When using react-admin 2.x with material-ui 3.x, material-ui throws deprecation errors in the console that can't be silenced. There errors are very noisy and hide real developer errors. Besides, depending on your CI settings, they may make the builds fail. The [mui team won't consider turning these errors into warnings](https://github.com/mui-org/material-ui/issues/15397), and we have no way to silence them in react-admin without breaking compatibility with mui 1.x.
2. When setting the dependency of react-admin to material-ui ">1.5.0 <4.0.0", even though an app depending on react-admin has a dependency on an old but compatible version (like 1.5.1), yarn installs 2 versions of material-ui (one recent, like 3.5, and one satisfying the old version requirement, like 1.5.1). I have no idea why this happens, but the duplicated libraries end up in the js bundle, almost doubling its size. This doesn't happen if an app has a dependency on a recent version (like 3.5): in that case, yarn correctly deduplicates material-ui. That's what I first did in this PR. But when I tried to downgrade the examples to get rid of the mui 3.x errors, I saw that yarn duplicated the material-ui package. A doubling of the js bundle size is a major bug that we can't publish.

In other terms, allowing react-admin to support both material-ui 1.X and 3.X is NOT possible. I'm sorry if I gave false hope, it's just that I really wanted it to work, too.

We'll just have to wait for material-ui v4, which will require a major rewrite of both react-admin and your apps based on react-admin v2 :disappointed:

eps1lon:
> The mui team won't consider turning these errors into warnings,

This is not true. Please don't spread misinformation before everybody had a chance to speak up. We don't just deploy code without any rationale. For anybody interested in why the warnings are logged the way they are I suggest reading the linked issue.

fzaninotto:
> This is not true.

@eps1lon You've lost me: I opened issue https://github.com/mui-org/material-ui/issues/15397 to ask you specifically to turn these errors into logs or warnings, and you closed it. If you actually want to turn deprecation warnings into console warnings, then please reopen the said issue.

I'm curious as to what you consider misinformation, as I linked the issue in my comment. Everyone can make up their mind.

dzienisz:
@fzaninotto can we make react-admin version that will work only on material ui v3?

fzaninotto:
@dzienisz you sure can, by forking the repo. The core team won't do it in the current repository, for reasons explained [in the FAQ](https://marmelab.com/react-admin/FAQ.html#why-doesnt-react-admin-support-the-latest-version-of-material-ui).

oliviertassinari:
> The mui team won't consider turning these errors into warnings, and we have no way to silence them in react-admin without breaking compatibility with mui 1.x.

@fzaninotto To be precise. We won't change the warning strategy in v3. We will improve the strategy in v4: [Improve the warning strategy (#15343)](https://github.com/mui-org/material-ui/issues/15343).

We want the warnings to log the **component stack trace**. It's a requirement we don't want to give up on. It really helps to identify where the wrong logic is. Handling breaking changes are painful enough, we are using all our leverages to simplify the migration. In order to do that, we have to use the prop-types module. This module happens to use `console.error` instead of `console.warn`. Turning the warnings in a *console.warn* / *console.log* would mean giving up on the components stack trace while not significantly improving the situation. I don't think that it's OK to logs deprecation messages, flooding the console when following the documentation instructions (the default).
 
In v4, we want to leverage the new React.warn() method. It logs the component stack trace with *console.warn*. We would be happy to have your input on the topic, help us shape the future of the library.

> allowing react-admin to support both material-ui 1.X and 3.X is NOT possible.

It wasn't designed for. I wouldn't try to do it. What prevents you from dropping Material-UI v1 support? As far as I know, you don't expose Material-UI directly, all the components are wrapped. Could you increase the Material-UI dependency from v1 to v3, fix all the warnings and release the change as a minor?

> We'll just have to wait for material-ui v4, which will require a major rewrite of both react-admin and your apps based on react-admin v2 😞

The v1 or v3 (it's the same) migration is detailed in https://next.material-ui.com/guides/migration-v3/. What point is specifically a "major rewrite"?

PR #3104, opened 2019-04-05
sticky appbar

astiag:
some times we have lots of content and it requires to scroll down to see them all.
in my case save button is at the end of content. on that position appbar is already hide.
since we have only one loading indicator inside appbar, after user clicking save button there is no way to see something is happening.

so I made this PR to be able to determine custom appbar with `sticky` position.

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -593,6 +593,18 @@ const MyAppBar = withStyles(styles)(({ classes, ...props }) => (
 export default MyAppBar;
 ```
 
+In addition you can pass `position="sticky"` to prevent AppBar from hiding when you scrolling down
+
+```jsx
+const MyAppBar = (props) => (
+    <AppBar {...props} position="sticky">
+       ...
+    </AppBar>
+);
+
+export default MyAppBar;
+```
+
 To use this custom `MyAppBar` component, pass it as prop to a custom `Layout`, as shown below:
 
 ```jsx
--- a/packages/ra-ui-materialui/src/layout/AppBar.js
+++ b/packages/ra-ui-materialui/src/layout/AppBar.js
@@ -57,49 +57,48 @@ const AppBar = ({
     toggleSidebar,
     userMenu,
     width,
+    position,
     ...rest
 }) => (
-    <Headroom>
-        <MuiAppBar
-            className={className}
-            color="secondary"
-            position="static"
-            {...rest}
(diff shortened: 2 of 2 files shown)


docs/Theming.md
```diff
@@ -593,6 +593,18 @@ const MyAppBar = withStyles(styles)(({ classes, ...props }) => (
 export default MyAppBar;
 ```
 
+In addition you can pass `position="sticky"` to prevent AppBar from hiding when you scrolling down
+
+```jsx
+const MyAppBar = (props) => (
+    <AppBar {...props} position="sticky">
```
fzaninotto: I'd prefer a boolean `sticky` prop

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -113,10 +112,12 @@ AppBar.propTypes = {
     toggleSidebar: PropTypes.func.isRequired,
     userMenu: PropTypes.node,
     width: PropTypes.string,
+    position: PropTypes.string,
 };
 
 AppBar.defaultProps = {
     userMenu: <UserMenu />,
+    poition: 'static',
```
fzaninotto: typo

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -132,4 +133,11 @@ const enhance = compose(
     withWidth()
 );
 
-export default enhance(AppBar);
+const EnhancedAppBar = enhance(AppBar);
+
+export default ({ position, ...props }) =>
```
fzaninotto: You should name that component to ease debugging

docs/Theming.md
```diff
@@ -593,6 +593,18 @@ const MyAppBar = withStyles(styles)(({ classes, ...props }) => (
 export default MyAppBar;
 ```
 
+In addition you can pass `position="sticky"` to prevent AppBar from hiding when you scrolling down
```
fzaninotto:
```suggestion
In addition, you can pass `position="sticky"` to prevent the `<AppBar>` from hiding when you scroll down.
```

fzaninotto:
Also, tests fail.

fzaninotto:
Any updates?

djhi:
No news for some time. Closing

PR #3115, opened 2019-04-10
[RFR] Display errors on mui BooleanInput

AntoineLelaisant:


packages/ra-ui-materialui/src/input/BooleanInput.js
```diff
@@ -51,6 +53,13 @@ export class BooleanInput extends Component {
                         />
                     }
                 />
+                {meta.error
+                    ? meta.error.map((error, index) => (
```
fzaninotto: `meta.error` is a String, not an array (see https://redux-form.com/8.2.0/docs/api/field.md/#-code-meta-error-string-code-optional)

packages/ra-ui-materialui/src/input/BooleanInput.spec.js
```diff
@@ -35,10 +37,42 @@ describe('<BooleanInput />', () => {
...
             .dive();
         assert.equal(wrapper.find('WithStyles(Switch)').prop('checked'), false);
     });
+
+    it('should displays errors', () => {
+        const wrapper = shallow(
+            <BooleanInput
+                source="foo"
+                input={{}}
+                meta={{ error: ['foo', 'bar'] }}
```
fzaninotto: error should be a string

fzaninotto:
Any updates?

PR #3119, opened 2019-04-11
[RFR] Fix coding style

AntoineLelaisant:
I just ran the `make prettier` command.


fzaninotto:
Thanks for your PR.

There was probably an update in prettier that causes these changes. The problem wit hcosmetic changes like these ones is that they break all open PRs, and cause problems when we want to merge long running branches (like next). That's why we run prettier just before or after releasing a minor version.

There will probably be more changes in the codebase between now and the new minor release. To avoid the hassle of rebasing, we'll run prettier ourselves when it's time.

PR #3123, opened 2019-04-12
Fix undefined id in RadioButtonGroupInput

johncalvinroberts:
If no `id` is passed to `RadioButtonGroupInput` component, the input is given the id `undefined_optionvalue`. Consider the case where two fields have the same choices, but different `source`. They will be given the same id, and the form control labels will interfere, which is not nice.

Updated the component to not put `undefined_optionvalue` into the id of the node, and added a note to the docs.

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -700,7 +700,6 @@ Lastly, use the `options` attribute if you want to override any of Material UI's
     labelPosition: 'right'
 }} />
 ```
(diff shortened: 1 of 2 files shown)


johncalvinroberts:
Okay, will do

johncalvinroberts:
@djhi I replaced with source and removed the ID prop, can still handle backwards compatibility.

djhi:
Thanks!

PR #3144, opened 2019-04-19
[RFR] Relax react router dependency to avoid version mismatch

### LINKED ISSUE #3138: require("history") console warnings
Hi! I'm using react-admin version 2.8.5, and I'm getting these warnings in the console. What could be wrong? I did a yarn upgrade also, and the warnings are still there.

`Warning: Please use `require("history").createBrowserHistory` instead of `require("history/createBrowserHistory")`. Support for the latter will be removed in the next major release.`

`Warning: Please use `require("history").createHashHistory` instead of `require("history/createHashHistory")`. Support for the latter will be removed in the next major release.`

`Warning: Please use `require("history").createMemoryHistory` instead of `require("history/createMemoryHistory")`. Support for the latter will be removed in the next major release.`

`Warning: Please use `require("history").PathUtils` instead of `require("history/PathUtils")`. Support for the latter will be removed in the next major release.`

fzaninotto:
Closes #3138


fzaninotto:
The warnings come from `history`, which is a dependency of `react-router-dom`. `history` is resolved to the latest version (4.9.0) on new projects. It complains that `react-router-dom` uses an outdated way of calling the history adapters. The fix consists of relaxing the `react-router-dom` version constraint, to have a more recent version that calls `history` adapters the right way.

PR #3146, opened 2019-04-19
[RFR] Fix query component does not fetch again when updated

### LINKED ISSUE #3126: Query component does not update
**Background:**
The simple user story for background here is that I wanted to select a customer from a list and then display a list of locations associated with the customer to pick from. The first reference (customer) is sorted out by a `ReferenceInput` which then triggers a `FormDataConsumer` which renders a custom component which uses a `Query` to look up the customer data and then pass the locations list of ids to a nested `Query` to get the locations for use in a SelectInput. _(Let me know if it would be helpful to see the code implementing the pattern above.)_

**What you were expecting:**
When changing customers I expected the locations `SelectInput` to be re-rendered with the new customer's locations as choices

**What happened instead:**
When setting a customer the first time it all worked fine. When changing the customer after that initial time all the machinery fired off but the data in the Query was from the initial customer.

**Related code:**
I fixed the issue by implementing `componentDidUpdate`. I copied the react-admin `Query` component and replaced `componentDidMount` with all of the code here:

```
  queryAndSetState = () => {
    const { dataProvider, type, resource, payload, options } = this.props;
    dataProvider(type, resource, payload, options)
      .then(({ data, total }) => {
        this.setState({
          data,
          total,
          loading: false
        });
      })
      .catch(error => {
        this.setState({
          error,
          loading: false
        });
      });
  };

  componentDidMount = () => this.queryAndSetState();

  componentDidUpdate = prevProps => {
    // This should be enough to ensure we update when we want and avoid an infinte loop
    //   JSON pattern should work here since we already rely on it for the
    //   simple types used for query objects. We could also compare using
    //   ids only but more parsing would be needed.
    if ( JSON.stringify(prevProps.payload) === JSON.stringify(this.props.payload) ) {
      return;
    }

    this.queryAndSetState(prevProps);
  };
```

**Environment**
* React-admin version: `^2.8.5`
* React version: `^16.8.6`

fzaninotto:
Closes #3126

### DIFF
--- a/packages/ra-core/src/util/Query.spec.tsx
+++ b/packages/ra-core/src/util/Query.spec.tsx
@@ -4,7 +4,7 @@ import {
     cleanup,
     fireEvent,
     // @ts-ignore
-    waitForDomChange
+    waitForDomChange,
 } from 'react-testing-library';
 import expect from 'expect';
 import Query from './Query';
@@ -36,7 +36,11 @@ describe('Query', () => {
                 {({ store }) => {
                     dispatchSpy = jest.spyOn(store, 'dispatch');
                     return (
-                        <Query type="mytype" resource="myresource" payload={myPayload}>
+                        <Query
+                            type="mytype"
+                            resource="myresource"
+                            payload={myPayload}
+                        >
                             {() => <div>Hello</div>}
                         </Query>
                     );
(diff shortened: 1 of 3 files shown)


RafeSacks:
Hi,

I was about to work on a PR myself and then I saw this. Thanks!!

An issue I came across while testing my original posted solution is that the promise has to be "cancelable" to avoid races with components being unmounted. Does this PR handle this issue?

After a bunch of research I ended up adding a dependency to `trashable-component`. which provides a high-level component to handle it.

I import:
```
/*
 * Reason for `trashable-react` dependency. Handle promise canceling on unmount
 * that also handles garbage collection correctly. Since this package has a
 * solution with testing that proves it, why not use it? Ref:
 *   - https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html
 *   - https://github.com/facebook/react/issues/5465
 *   - https://github.com/hjylewis/trashable
 */
import makeComponentTrashable from "trashable-react";
```

...and wrap:
```
export default makeComponentTrashable(withDataProvider(Query));
```

Then just wrap the dataProvider promise:
```
    registerPromise(dataProvider(type, resource, payload, options))
      .then(({ data, total }) => { ...
```

RafeSacks:
I was also wondering if `fetchStart` and `fetchEnd` (in promise `finally`) is needed? I found a number of places in the docs where these are used but they are not used in `Query`.

PR #3150, opened 2019-04-19
Update Tutorial.md

themalkolm:
No need to import Resource right away, minor thingy.


themalkolm:
Scratch that, looks like it is required somehow.

PR #3152, opened 2019-04-21
[RFR] Improve translation loading

### LINKED ISSUE #3139: Translations not updating using an async function as provider and not changing language key
**What you were expecting:**
When loading the translations from an async function (e.g. XHR Request) the translated keys should update.

**What happened instead:**
It only changes, when the action dispatched by the side effect has a different language key than the current selected language. If it's the same, then the TranslatedProvider does not rerender, because it's  componentDidUpdate method only checks on the language key.

So even when you change to the same language again it won't update, if you haven't changed it to another language before.

**Steps to reproduce:**
It's a rare case,  so I created a code sandbox. 
https://yjvyo9yp29.codesandbox.io/#/posts/13

Only changed the i18nProvider and added a language switcher to an aside of the post edit view, where the resource keys are visible.

**Other information:**
I think this can be easily fixed by checking if the messages object changes in the TranslationProvider componentDidUpdate method. 
*But* then still there are technical keys visible until the async function returns the translation. 

**Environment**
react-admin 2.8.5

nik-lampe:
These changes do two things:

- When receiving a new locale and messages it's not only checked if the locale has changed but also if the messages have changed. This fixes #3139 .
Object comparison might not be the best use here, but I think that the messages object won't be too large for it to still be performant.

- It will start showing the fetch spinner before loading the translations and end it afterwards. This is useful when translations are loaded async. When it takes longer, the user doesn't know that anything is happening and then suddenly the strings switch. 
This might be improved to delay it for a short while, so there is no spinner when the messages are returned immediately.

### DIFF
--- a/packages/ra-core/src/i18n/TranslationProvider.tsx
+++ b/packages/ra-core/src/i18n/TranslationProvider.tsx
@@ -56,7 +56,7 @@ class TranslationProviderView extends Component<ViewProps, State> {
     }
 
     componentDidUpdate(prevProps) {
-        if (prevProps.locale !== this.props.locale) {
+        if (prevProps.locale !== this.props.locale || prevProps.messages !== this.props.messages) {
             const { locale, messages } = this.props;
 
             const polyglot = new Polyglot({
--- a/packages/ra-core/src/sideEffect/i18n.ts
+++ b/packages/ra-core/src/sideEffect/i18n.ts
@@ -4,6 +4,8 @@ import {
     CHANGE_LOCALE,
     changeLocaleSuccess,
     changeLocaleFailure,
+    fetchStart,
+    fetchEnd,
 } from '../actions';
 
 /**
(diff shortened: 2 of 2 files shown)


djhi:
Build is failing because you did not apply eslint on the modified files.

nik-lampe:
Oops, true, fixed it. Thanks 🙌

packages/ra-core/src/i18n/TranslationProvider.tsx
```diff
@@ -56,7 +56,7 @@ class TranslationProviderView extends Component<ViewProps, State> {
     }
 
     componentDidUpdate(prevProps) {
-        if (prevProps.locale !== this.props.locale) {
+        if (prevProps.locale !== this.props.locale || prevProps.messages !== this.props.messages) {
```
djhi: You're comparing object references here for messages. This should be enough as the reference to the messages object should change when loading new ones

fzaninotto:
Thanks!

PR #3153, opened 2019-04-22
SelectArrayInput disable value for choice

JackyW83:


fzaninotto:
Please explain your use case, document and unit test the new option.

djhi:
Can we help you with this PR ?

PR #3156, opened 2019-04-23
Fix rowClick using function value returning empty redirect behavior

### LINKED ISSUE #3155: [Feature request] Prevent default redirect behavior on rowClick
The first words, thanks for your awesome project.

My feature request related to a rowClick event on Datagrid. I have some custom action when click an item on a row. Sometime, i pick a row as a template data for a create form. But after rowClick event execute, default redirect event occurs too. So that my form have been reset and destroy.

I think if we want to prevent this default behavior just return nothing in rowClick event hanlder.

image

tunglt1810:
closes #3155


fzaninotto:
Thanks!

tunglt1810:
Thanks you all!

PR #3159, opened 2019-04-23
Support nested records in source property of ReferenceManyField

aramando:
All the `...Field` components can accept a compound source property to enable the display of nested record properties, with the exception of `ReferenceManyField`. This PR fixes that.

Unfortunately I haven't been able to get tests working so far and can't spend any more time on it right now, so I'm hoping someone else can add the one extra test to test this specific capability of the source property.

I don't think any extra documentation is required, as this is merely bringing the behaviour of this field in line with all the other field components.


fzaninotto:
Thanks!

Issue #3160, opened 2019-04-24
Graphql Aliases


Oupsla:
Hello folks 👋 
I'm using react-admin with ra-data-graphql-simple as dataProvider and I'm having an issue.
When I try to use aliases ([ref](https://graphql.org/learn/queries/#aliases)), the page fails to load.
I tried to find where it comes from but without success 😞 

**What you were expecting:**
Using aliases is working great and can be use like a normal query (the field is mapped to the new name defined by the alias)

Example (for a custom query defined for `fetchType: GET_ONE`)

```
query Device($id: ID!) {
  data: Device(id: $id) {
    id
    serial
    lat: latitude
    lng: longitude
    __typename
  }
}
```

**What happened instead:**
Got a `Warning: Missing translation for key: "Cannot read property 'type' of undefined"` in the console and the page fails to load and goes back to the list view.
But the query (over http) succeeds and returns correct result !

**Steps to reproduce:**
Create a dataprovider with ra-data-graphql-simple and add an alias.
Code sandbox : https://codesandbox.io/s/pkm8051kp0

Go to the customers list and see the error

**Environment**

* React-admin version: ^2.8.6
* Last version that did not exhibit the issue (if applicable):
* React version: ^16.8.6
* Browser: Chrome on Macos
* Stack trace (in case of a JS error):

```
console.(anonymous function) | @ | index.js:1446
-- | -- | --
  | printWarning | @ | warning.js:34
  | warning | @ | warning.js:58
  | warn | @ | index.js:28
  | push../node_modules/node-polyglot/index.js.Polyglot.t | @ | index.js:395
  | Notification.render | @ | Notification.js:151
  | finishClassComponent | @ | react-dom.development.js:8747
  | updateClassComponent | @ | react-dom.development.js:8715
  | beginWork | @ | react-dom.development.js:9364
  | performUnitOfWork | @ | react-dom.development.js:12372
  | workLoop | @ | react-dom.development.js:12402
  | renderRoot | @ | react-dom.development.js:12431
  | performWorkOnRoot | @ | react-dom.development.js:13028
  | performWork | @ | react-dom.development.js:12949
  | performSyncWork | @ | react-dom.development.js:12927
  | requestWork | @ | react-dom.development.js:12817
  | scheduleWorkImpl | @ | react-dom.development.js:12691
  | scheduleWork | @ | react-dom.development.js:12650
  | enqueueSetState | @ | react-dom.development.js:6920
[...]

PR #3168, opened 2019-04-26
[RFR] Allow aliases in graphql dataprovider

Oupsla:
Fix: https://github.com/marmelab/react-admin/issues/3160

### DIFF
--- a/packages/ra-data-graphql-simple/package.json
+++ b/packages/ra-data-graphql-simple/package.json
@@ -39,7 +39,7 @@
         "ra-data-graphql": "^3.0.0-alpha.3"
     },
     "peerDependencies": {
-        "graphql": "^0.13.2",
+        "graphql": "^14.1.1",
         "ra-core": "^3.0.0-alpha.0"
     },
     "devDependencies": {
--- a/packages/ra-data-graphql-simple/src/getResponseParser.js
+++ b/packages/ra-data-graphql-simple/src/getResponseParser.js
@@ -2,17 +2,20 @@ import { TypeKind } from 'graphql';
 import { GET_LIST, GET_MANY, GET_MANY_REFERENCE } from 'ra-core';
 import getFinalType from './getFinalType';
 
-const sanitizeResource = (introspectionResults, resource) => data => {
+const sanitizeResource = (introspectionResults, resource, aliases) => data => {
     const result = Object.keys(data).reduce((acc, key) => {
         if (key.startsWith('_')) {
             return acc;
         }
 
-        const field = resource.type.fields.find(f => f.name === key);
+        const isAlias = aliases && aliases[key] ? true : false;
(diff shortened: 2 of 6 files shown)


Oupsla:
Hello @fzaninotto,
Could be possible to get a review ? 😄

djhi:
Thanks for working on this! It looks good to me. I'll have to test it against the demo though and I won't be able to do so before monday. It would be great to find a usecase for aliases in the demo too. I'll try to figure something but go ahead if you think of something first.

Oupsla:
I encounter this need for a specific problem at work 😄 

My usecase is that I got users and organisations (and those organisations got accesses to 'things' ).

So in a show view I have created tabs to show the summary of the user, his organisations and his accesses (through his organisations).

So I got a query like that (to fetch one user) (little note, I follow recommandations of graphql for paginations between object) : 
```gql

query User($id: ID!) {
    data: User(id: $id) {
      id
      email
      organisations: organisationsConnection(first: 100) {
        edges {
          id
          type
          node {
            id
            name
            identification
          }
        }
      }
accesses: organisationsConnection(first: 100) {
        edges {
          node {
            accessesConnection(first: 100) {
              edges {
                node {
                  id
                  name
                }
              }
            }
          }
        }
      }
    }
  }
```

So I need to query the same subquery 2 times (and so the need of aliases 😄 ).

djhi:
Thanks for sharing. I was looking for a use case in our demo example though :)

Oupsla:
Don't merge it right now, I have detected an issue while using it !
If the same alias is used multiple times but at different levels of the graphql query, it can fails... (it is a edge case but I have encounter it 😄 ), I'll try to fix it and commit it

Oupsla:
After working on a solution, I was wondering the goal of the method `sanitizeResource`.
I tried to removing it completely and my project is working (and with aliases 😆 ) .

@djhi (or maybe @fzaninotto), do you have more insights about this method ? (It is maybe a legacy that can be removed).

Oupsla:
I have tried another approach, not based on graphql introspection that just filter the response and add things (like `.id` and `Ids`, based on tests).
This solution allows aliases.

I think it is better cause the response payload of a graphql endpoint must respect the query so no need to check the introspection.

Already tested through my project (with aliases) and working, tell me what you think about it 😄

Oupsla:
@djhi @fzaninotto Friendly ping to get a review if you got time 😄

Oupsla:
Hello folks !
Any news about reviewing/merging this PR ?

djhi:
Sorry about that. We've been very busy working on the v3. As this is a new feature, would you mind targeting the `next` branch ? I'll take a look right after!

Oupsla:
No problem, base PR changed and rebased. Just wanted to do a friendly ping so this PR don't die 😄 
By the way really good work for the v3, really eager to try it out !

djhi:
Thanks!

Issue #3174, opened 2019-04-29
Deep props make Query send request on every update


PPierzc:
**What you were expecting:**
The Query component should send a request on update only when the props have changed.

**What happened instead:**
Sends request on every update

**Steps to reproduce:**
Create a deep payload object for the Query component.

**Related code:**
```
const payload = {
      id: record.id,
      target: 'author',
      sort: {
        order: 'DESC',
        field: 'createdAt'
      },
      pagination: {
        page: 1,
        perPage: 1
      }
}

return (
      <Query
        type='GET_MANY_REFERENCE'
        resource={constants.resources.posts}
        payload={payload}
      >
        {({ data, loading, error }) => {
          return <></>
      </Query>
)
```

**Other information:**
A solution could be to change shallowEqual in the Query's componentDidUpdate to deepEqual.

**Environment**

* React-admin version: 2.9.0
* Last version that did not exhibit the issue (if applicable): -
* React version: 16.6.3
* Browser: Google Chrome 73
* Stack trace (in case of a JS error): -

Kmaschta:
Related to https://github.com/marmelab/react-admin/pull/3146

PR #3176, opened 2019-04-29
[RFR] Fix sending request by Query with deep props on every update

PPierzc:
- Changed shallowEqual to deepEqual in Query to accommodate for possible deep props.

 - Changed the placement of `myPayload` variable.
When `myPayload` is global the objects are indeed identical and the script worked did pass. However, this assumes independency from props, which is not desired. This can easily lead to endless loops. 

Solves #3174

### DIFF
--- a/packages/ra-core/src/util/Query.tsx
+++ b/packages/ra-core/src/util/Query.tsx
@@ -1,5 +1,5 @@
 import { Component, ReactNode } from 'react';
-import { shallowEqual } from 'recompose';
+import equal from 'deep-equal';
 import withDataProvider from './withDataProvider';
 
 type DataProviderCallback = (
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/util/Query.tsx
```diff
@@ -1,5 +1,5 @@
 import { Component, ReactNode } from 'react';
-import { shallowEqual } from 'recompose';
+import equal from 'deep-equal';
```
fzaninotto: `deep-equal` isn't a direct dependency. use `lodash/isEqual` instead

djhi:
Any update?

fzaninotto:
Superseded by #3267

Issue #3178, opened 2019-04-30
RejectButton is coded as AcceptButton


ericwb:
**What you were expecting:**
I would expect the code in [RejectButton.js](https://github.com/marmelab/react-admin/blob/master/examples/demo/src/reviews/RejectButton.js) to define a class called RejectButton.

**What happened instead:**
Take a look at [RejectButton.js](https://github.com/marmelab/react-admin/blob/master/examples/demo/src/reviews/RejectButton.js) and you'll notice the code defines a class named AcceptButton.  Seems a little confusing, because there is actual code for rejection and it's using the thumbs down icon.

**Steps to reproduce:**
Code inspection of [RejectButton.js](https://github.com/marmelab/react-admin/blob/master/examples/demo/src/reviews/RejectButton.js)

**Related code:**
N/A

**Other information:**
n/a

**Environment**

* React-admin version: master
* Last version that did not exhibit the issue (if applicable): n/a
* React version: n/a
* Browser: n/a
* Stack trace (in case of a JS error): n/a

djhi:
Thanks for reporting. It looks like a bad copy/paste. Would you mind opening a PR to fix it ?

ericwb:
Sure, will do

fargito:
This issue should be closed, since a PR fixed it

PR #3182, opened 2019-04-30
Rename improperly named RejectButton class name

ericwb:
Looks like there was some copy-and-paste error, so RejectButton
was defined as AcceptButton. This commit renames references to
properly to "Reject..."

Fixes Issue #3178

Signed-off-by: Eric Brown <browne@vmware.com>


djhi:
Thanks!

PR #3188, opened 2019-05-04
[RFR] Add useTranslate hook

fzaninotto:
- [x] Add `useTranslate` Hook
- [x] Document it
- [x] Test it
- [x] Use it in simple example
- [x] Use it in demo example
- [x] upgrade ra-core to use it instead of `withTranslate` in function components
- [x] upgrade ra-ui-material-ui to use it instead of `withTranslate` in function components
- ~~use `useTranslate` in class components~~ (requires translating these components to functions, to be done sometime in the future)

```diff
// in src/MyHelloButton.js
import React from 'react';
-import { translate } from 'react-admin';
+import { useTranslate } from 'react-admin';

-const MyHelloButton = ({ translate }) => {
+const MyHelloButton = () => {
+   const translate = useTranslate();
    return (
        <button>{translate('myroot.hello.world')}</button>
    );
}

-export default translate(MyHelloButton);
+export default MyHelloButton;
```

### DIFF
--- a/docs/Translation.md
+++ b/docs/Translation.md
@@ -277,7 +277,7 @@ const App = () => (
 
 ## `useTranslate` Hook
 
-If you need to translate messages in your own components, React-admin provides a `useTranslate` hook, which simply returns the `translate` function:
+If you need to translate messages in your own components, React-admin provides a `useTranslate` hook, which returns the `translate` function:
 
 ```jsx
 // in src/MyHelloButton.js
--- a/docs/Translation.md
+++ b/docs/Translation.md
@@ -100,11 +100,9 @@ You can find translation packages for the following languages:
 
 In addition, the previous version of react-admin, called admin-on-rest, was translated in the following languages:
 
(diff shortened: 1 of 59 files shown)


fzaninotto:
Switching to RFR

examples/demo/src/categories/LinkToRelatedProducts.js
```diff
@@ -16,30 +15,29 @@ const styles = {
     },
 };
 
-const LinkToRelatedProducts = ({ classes, record, translate }) => (
```
fzaninotto: The diff looks much better with 'hide whitespace changes' turned on.

docs/Translation.md
```diff
@@ -280,25 +275,28 @@ const App = () => (
 );
 ```
 
-## Translating Your Own Components
+## `useTranslate` Hook
 
-React-admin package provides a `translate` Higher-Order Component, which simply passes the `translate` function as a prop to the wrapped component:
+If you need to translate messages in your own components, React-admin provides a `useTranslate` hook, which simply returns the `translate` function:
```
djhi: s/simply //

PR #3190, opened 2019-05-06
fixed AutocompleteInput popup to re-render after choice is selected

FACOLOMBANI:
Hi,

This is a PR to fix AutocompleteInput issue #2428. I added the prop focusInputOnSuggestionClick by default to false, and since the input loses focus on suggestion selection, i had to change the state when calling handleSuggestionSelected().

Best Regards.

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.js
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.js
@@ -15,33 +15,34 @@ import classnames from 'classnames';
 
 import { addField, translate, FieldTitle } from 'ra-core';
 
-const styles = theme => createStyles({
-    container: {
-        flexGrow: 1,
-        position: 'relative',
-    },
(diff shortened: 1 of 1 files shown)


fzaninotto:
You seem to use different prettier settings than us. Can you please avoid reformatting on save or use our prettier preset, to minimize the coding style changes?

FACOLOMBANI:
Of course, i apologize, just ran your prettier settings.

fzaninotto:
You patch doesn't work. I tested on the simple example, in a Comment detail, changing the related post doesn't work anymore.

FACOLOMBANI:
This is weird, i tested it locally on the reviews component using the customer filter and it worked fine, i'll check it again

FACOLOMBANI:
I added a new case in the handleChange function, can you please retest with this modification?

djhi:
Thanks!

PR #3213, opened 2019-05-11
[RFR] Use hooks in controllers

djhi:
I believe the first step before migrating our controller components to custom hooks is to make them use hooks in the first place.

- [x] `CreateController`
- [x] `EditController`
- [x] `ShowController`
- [x] `ListController`
- [x] Unit test the utility functions which have been extracted

### DIFF
--- a/packages/ra-core/src/controller/CreateController.tsx
+++ b/packages/ra-core/src/controller/CreateController.tsx
@@ -1,16 +1,16 @@
-import { Component, ReactNode, ComponentType } from 'react';
-import { connect } from 'react-redux';
-import compose from 'recompose/compose';
+import { ReactNode, useCallback, FunctionComponent } from 'react';
+// @ts-ignore
+import { useDispatch, useSelector } from 'react-redux';
 import inflection from 'inflection';
 import { parse } from 'query-string';
 
-import withTranslate from '../i18n/translate';
-import { crudCreate as crudCreateAction } from '../actions';
-import checkMinimumRequiredProps from './checkMinimumRequiredProps';
+import { crudCreate } from '../actions';
+import { useCheckMinimumRequiredProps } from './checkMinimumRequiredProps';
 import { Location } from 'history';
 import { match as Match } from 'react-router';
-import { Record, Translate, Dispatch } from '../types';
+import { Record, Translate, ReduxState } from '../types';
 import { RedirectionSideEffect } from '../sideEffect';
+import { useTranslate } from '../i18n';
 
 interface ChildrenFuncParams {
     isLoading: boolean;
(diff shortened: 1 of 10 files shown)


packages/ra-core/src/controller/CreateController.tsx
```diff
@@ -36,11 +36,22 @@ interface Props {
     resource: string;
 }
 
-interface EnhancedProps {
-    crudCreate: Dispatch<typeof crudCreateAction>;
-    isLoading: boolean;
-    translate: Translate;
-}
+export const getRecord = ({ state, search }, record: any = {}) =>
```
fzaninotto: this should live in the reducers directory as it's a selector (same with the test of this function)

packages/ra-core/src/controller/CreateController.tsx
```diff
@@ -83,95 +94,57 @@ interface EnhancedProps {
...
+        (state: ReduxState) => state.admin.loading > 0
+    );
 
-    constructor(props) {
-        super(props);
-        const {
-            location: { state, search },
-            record,
-        } = this.props;
-        this.record =
```
fzaninotto: I think you've broken the clone feature by not porting this

packages/ra-core/src/controller/EditController.tsx
```diff
@@ -91,123 +80,77 @@ interface EnhancedProps {
...
-            children,
-            id,
-            isLoading,
-            record,
-            resource,
-            translate,
-            version,
-        } = this.props;
+    useEffect(() => {
+        dispatch(resetForm(REDUX_FORM_NAME));
```
fzaninotto: resetForm wasn't called on mount in the previous version. Isn't there a consequence of calling it here?

packages/ra-core/src/controller/CreateController.tsx
djhi: Agreed, I will extract selectors

packages/ra-core/src/controller/CreateController.tsx
djhi: The e2e tests disagree with you ;)

packages/ra-core/src/controller/EditController.tsx
djhi: I think we have e2e tests for this case but I will check

packages/ra-core/src/controller/CreateController.tsx
djhi: And btw I ported this. This is the `getRecord` function you commented earlier

kopax:
This looks cool thanks! Will we have a part about what are controllers in the documentation (RFC style)?

djhi:
> Will we have a part about what are controllers in the documentation (RFC style)?

See the roadmap ;)

fzaninotto:
Awesome!

djhi:
> Awesome!
:hugs: 

Now we can extract custom hooks :rocket: !

PR #3215, opened 2019-05-13
Pass dispatch to function that declares resources at runtime

smeng9:
When we call a function to dynamically define the resources at app start, it would be convenient if we can dispatch additional actions to redux's store to set some states or show a notification.

### DIFF
--- a/packages/ra-core/src/CoreAdminRouter.tsx
+++ b/packages/ra-core/src/CoreAdminRouter.tsx
@@ -8,6 +8,7 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { AnyAction } from 'redux';
 import { Route, Switch } from 'react-router-dom';
 import compose from 'recompose/compose';
 import getContext from 'recompose/getContext';
@@ -47,6 +48,7 @@ interface EnhancedProps {
     authProvider?: AuthProvider;
     isLoggedIn?: boolean;
     userLogout: Dispatch<typeof userLogoutAction>;
+    dispatch: Dispatch<AnyAction>;
 }
 
 interface State {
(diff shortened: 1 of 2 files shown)


fzaninotto:
We're using the latest version of react-redux in `next`, which uses hooks. Getting access to the `dispatch` function in your components becomes as easy as calling `useDispatch` and requires no change in the react-admin core. So you won't need the change you proposed to dispatch actions.

Thanks for the patch anyway.

smeng9:
Hi @fzaninotto I think we may still need to reopen this pull request. For this specific case that modifies the official documentation [declaring resources at runtime](https://marmelab.com/react-admin/Admin.html#declaring-resources-at-runtime), I tested the `useDispatch` hook with react-admin 3.0.0 alpha, it does not seem to work.

Code snippet that does not work:
```
const fetchResources = async (permissions) => {

    // we cannot call useDispatch here!
    let dispatch = useDispatch();

    let response = await fetch('http://path.to.my.api/resources')
    let json = await response.json();
    dispatch({someAction: setReduxStates_or_triggerSideEffects});
    return knownResources.filter(resource => json.resources.includes(resource.props.name)));
}

const App = () => (
    <Admin dataProvider={simpleRestProvider('http://path.to.my.api')}>
        {fetchResources}
    </Admin>
);
```

`fetchResources` function is a regular Javascript callback function of [CoreAdminRouter.tsx.](https://github.com/marmelab/react-admin/blob/d9251a145e7259634f8e640e3fecde8752208a81/packages/ra-core/src/CoreAdminRouter.tsx#L79) Putting an `useDispatch` there does not satisfy the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) because it is neither at the top level nor a react function. If we want to follow the react trend of using hooks, we have to change `CoreAdminRouter` to a functional component and call useDispatch there and pass the dispatch to the `fetchResources` function like below.

Instead we need to pass the `dispatch` from somewhere else:
```
// dispatch is passed through here ------------v
const fetchResources = async (permissions, dispatch) => {
    let response = await fetch('http://path.to.my.api')
    let json = await response.json();
    dispatch({someAction: setReduxStates_or_triggerSideEffects});
    return knownResources.filter(resource => json.resources.includes(resource.props.name)));
}
```

The simplest way currently is to connect `dispatch` to CoreAdminRouter.tsx in this pull request and is 2.x compatible. We can consider change CoreAdminRouter.tsx to functional component in the 3.x but it will take longer time for this feature to be available.

In conclusion, we will still need to change the react-admin core to make the `dispatch` passed to `fetchResources` function because hooks does not work the other way.

Thanks for your time to take a look.

PR #3216, opened 2019-05-13
[RFR] Fix handling of deleted references

djhi:
## Problem

When a `<ReferenceManyField>` displays a `<DataGrid>` with a `DeleteButton` on each row, optimistically deleting a reference will set its row into loading state but won't actually remove it until the action is resolved.

ra-references-many-delete-bug

## Solution

Handle the `DELETE` effects in the `oneToMany` reducer by checking whether we have references related to the deleted item resource and removing the deleted item from them.

ra-references-many-delete-bug-fix

### DIFF
--- a/packages/ra-core/src/reducer/admin/references/oneToMany.ts
+++ b/packages/ra-core/src/reducer/admin/references/oneToMany.ts
@@ -23,12 +23,23 @@ const oneToManyReducer: Reducer<State> = (
     action: ActionTypes
 ) => {
     if (action.meta && action.meta.optimistic) {
+        const relatedTo = getRelatedReferences(
+            previousState,
+            action.meta.resource
+        );
+
         if (action.meta.fetch === DELETE) {
-            return removeDeletedReference(previousState, action);
+            return relatedTo.reduce(
+                removeDeletedReference(action.payload.id),
+                previousState
+            );
         }
 
         if (action.meta.fetch === DELETE_MANY) {
-            return removeDeletedReferences(previousState, action);
+            return relatedTo.reduce(
+                removeDeletedReferences(action.payload.ids),
+                previousState
+            );
         }
     }
     switch (action.type) {
@@ -130,48 +141,34 @@ export const getReferencesByIds = (
 const getRelatedReferences = (previousState, resource) =>
     Object.keys(previousState).filter(key => key.includes(resource));
 
-const removeDeletedReference = (previousState, action: ActionTypes) => {
-    const relatedTo = getRelatedReferences(previousState, action.meta.resource);
+const removeDeletedReference = removedId => (previousState, key) => {
+    const hasReferenceToRemovedId = previousState[key].ids.includes(removedId);
 
-    return relatedTo.reduce((acc, key) => {
-        const hasReferenceToRemovedId = previousState[key].ids.includes(
-            action.payload.id
-        );
+    if (!hasReferenceToRemovedId) {
+        return previousState;
+    }
 
-        if (!hasReferenceToRemovedId) {
-            return acc;
-        }
-
-        return {
-            ...acc,
-            [key]: {
-                ids: previousState[key].ids.filter(
-                    id => id !== action.payload.id
-                ),
-                total: previousState[key].total - 1,
-            },
-        };
-    }, previousState);
+    return {
+        ...previousState,
+        [key]: {
+            ids: previousState[key].ids.filter(id => id !== removedId),
+            total: previousState[key].total - 1,
+        },
+    };
 };
 
-const removeDeletedReferences = (previousState, action: ActionTypes) => {
-    const relatedTo = getRelatedReferences(previousState, action.meta.resource);
+const removeDeletedReferences = removedIds => (previousState, key) => {
+    const idsToRemove = previousState[key].ids.filter(id =>
+        removedIds.includes(id)
+    );
 
-    return relatedTo.reduce((acc, key) => {
-        const idsToRemove = previousState[key].ids.filter(id =>
-            action.payload.ids.includes(id)
-        );
-
-        return {
-            ...acc,
-            [key]: {
-                ids: previousState[key].ids.filter(
-                    id => !action.payload.ids.includes(id)
-                ),
-                total: previousState[key].total - idsToRemove.length,
-            },
-        };
-    }, previousState);
+    return {
+        ...previousState,
+        [key]: {
+            ids: previousState[key].ids.filter(id => !removedIds.includes(id)),
+            total: previousState[key].total - idsToRemove.length,
+        },
+    };
 };
 
 export const nameRelatedTo = (reference, id, resource, target, filter = {}) => {
--- a/packages/ra-core/src/reducer/admin/references/oneToMany.ts
+++ b/packages/ra-core/src/reducer/admin/references/oneToMany.ts
@@ -141,7 +141,10 @@ export const getReferencesByIds = (
 const getRelatedReferences = (previousState, resource) =>
     Object.keys(previousState).filter(key => key.includes(resource));
 
-const removeDeletedReference = removedId => (previousState, key) => {
+const removeDeletedReference = (removedId: Identifier) => (
+    previousState,
+    key
+) => {
     const hasReferenceToRemovedId = previousState[key].ids.includes(removedId);
 
     if (!hasReferenceToRemovedId) {
@@ -157,7 +160,10 @@ const removeDeletedReference = removedId => (previousState, key) => {
     };
 };
 
-const removeDeletedReferences = removedIds => (previousState, key) => {
+const removeDeletedReferences = (removedIds: Identifier[]) => (
+    previousState,
+    key
+) => {
     const idsToRemove = previousState[key].ids.filter(id =>
         removedIds.includes(id)
     );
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -13,12 +15,31 @@ interface State {
...
     | CrudGetManyReferenceSuccessAction
+    | CrudDeleteSuccessAction
     | { type: 'OTHER_ACTION'; payload: any; meta?: any };
 
 const oneToManyReducer: Reducer<State> = (
     previousState = initialState,
     action: ActionTypes
 ) => {
+    if (action.meta && action.meta.fetch === DELETE && action.meta.optimistic) {
+        const relatedTo = Object.keys(previousState).filter(key =>
```
ThieryMichel: move into a removeOrphanReference function. This way future reader will know at once what we do on an optimistic delete

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -13,12 +15,56 @@ interface State {
...
+            }),
+            previousState
+        );
+    }
+    if (
+        action.meta &&
+        action.meta.fetch === DELETE_MANY &&
+        action.meta.optimistic
+    ) {
+        const relatedTo = Object.keys(previousState).filter(key =>
```
ThieryMichel: reuse the same function

packages/ra-core/src/reducer/admin/references/oneToMany.ts
djhi: Not an orphan actually but I see what you mean

ThieryMichel: Yep, naming: the hardest part in programming

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -13,12 +15,56 @@ interface State {
...
+    | CrudDeleteSuccessAction
     | { type: 'OTHER_ACTION'; payload: any; meta?: any };
 
 const oneToManyReducer: Reducer<State> = (
     previousState = initialState,
     action: ActionTypes
 ) => {
+    if (action.meta && action.meta.fetch === DELETE && action.meta.optimistic) {
+        const relatedTo = Object.keys(previousState).filter(key =>
+            key.includes(action.meta.resource)
```
Kmaschta: How can a `Object.keys` result can be an array.. ? :o

Kmaschta: Ow. `includes` works on strings too. My bad.

djhi: ?

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -13,12 +15,24 @@ interface State {
...
 type ActionTypes =
     | CrudGetManyReferenceSuccessAction
+    | CrudDeleteSuccessAction
     | { type: 'OTHER_ACTION'; payload: any; meta?: any };
 
 const oneToManyReducer: Reducer<State> = (
     previousState = initialState,
     action: ActionTypes
 ) => {
+    if (action.meta && action.meta.fetch === DELETE && action.meta.optimistic) {
```
fzaninotto: I suggest you group the optimistic cases just like in the resource.list.ids reducer

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +129,46 @@ export const getReferencesByIds = (
...
+    return relatedTo.reduce((acc, key) => {
+        const idsToRemove = previousState[key].ids.filter(id =>
+            action.payload.ids.includes(id)
+        );
+
+        return {
+            ...acc,
+            [key]: {
+                ids: previousState[key].ids.filter(
+                    id => !action.payload.ids.includes(id)
```
fzaninotto: haven't you just done that a couple lines earlier?

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +129,46 @@ export const getReferencesByIds = (
     return Object.keys(references).length > 0 ? references : null;
 };
 
+const getRelatedReferences = (previousState, resource) =>
+    Object.keys(previousState).filter(key => key.includes(resource));
+
+const removeDeletedReference = (previousState, action: ActionTypes) => {
+    const relatedTo = getRelatedReferences(previousState, action.meta.resource);
```
fzaninotto: see the ids reducer: we test that the id exist before making the change. Nothing forbids a component from removing an element which isn't in the store. That would lead to a wrong total.

packages/ra-core/src/reducer/admin/references/oneToMany.ts
djhi: The inverse

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +129,46 @@ export const getReferencesByIds = (
...
+
+    return relatedTo.reduce(
```
ThieryMichel: in fact, move this part in the function too. Or maybe make another function.

packages/ra-core/src/reducer/admin/references/oneToMany.ts
djhi: Good catch. I did it for the many case but not here...

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +138,45 @@ export const getReferencesByIds = (
...
+            total: previousState[key].total - 1,
+        },
+    };
+};
+
+const removeDeletedReferences = (removedIds: Identifier[]) => (
+    previousState,
+    key
+) => {
+    const idsToRemove = previousState[key].ids.filter(id =>
```
ThieryMichel: Ain't the total equal to the number of ids ?
If so, you could filter to get the kept id, and returns 
```js
{
    ids: keptIds,
    total: keptIds.length
}
```
If not the test should reflect that fact.

ThieryMichel: Also refactor removeDeletedReference to use removeDeletedReferences.

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +138,45 @@ export const getReferencesByIds = (
     return Object.keys(references).length > 0 ? references : null;
 };
 
+const getRelatedReferences = (previousState, resource) =>
+    Object.keys(previousState).filter(key => key.includes(resource));
+
+const removeDeletedReference = (removedId: Identifier) => (
```
fzaninotto: this seems like a special case of the more general removeDeletedReferences.

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -13,12 +15,33 @@ interface State {
...
 ) => {
+    if (action.meta && action.meta.optimistic) {
+        const relatedTo = getRelatedReferences(
+            previousState,
+            action.meta.resource
+        );
+
+        if (action.meta.fetch === DELETE) {
+            return relatedTo.reduce(
+                removeDeletedReference(action.payload.id),
```
fzaninotto: `removeDeletedReferences([action.payload.id])`?

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,6 +138,45 @@ export const getReferencesByIds = (
...
+};
+
+const removeDeletedReferences = (removedIds: Identifier[]) => (
+    previousState,
+    key
+) => {
+    const idsToKeep = previousState[key].ids.filter(
+        id => !removedIds.includes(id)
+    );
+
```
fzaninotto: again, you need to check that there is something to do before creating another state

djhi: You mean before recreating a state equal to the previous one ?

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -114,7 +138,41 @@ export const getReferencesByIds = (
...
 
-export const nameRelatedTo = (reference, id, resource, target, filter = {}) => {
+const getRelatedReferences = (previousState: State, resource: string) =>
+    Object.keys(previousState).filter(key => key.includes(resource));
+
+const removeDeletedReferences = (removedIds: Identifier[]) => (
+    previousState: State,
+    key: string
+) => {
+    const idsToRemove = previousState[key].ids.filter(id =>
```
fzaninotto: instead of filtering twice just to count the number of elements to filter out, why don't you just count once and then compare the length to the length of the original key?

```js
    const idsToKeep = previousState[key].ids.filter(
        id => !removedIds.includes(id)
    );
    if (idsToKeep.length === previousState[key].ids.length) {
        return previousState;
    }
    return {
        ...previousState,
        [key]: {
            ids: idsToKeep,
            total: idsToKeep.length,
        },
    };
```

PR #3217, opened 2019-05-13
[RFR] Extract custom hooks from controllers

djhi:
Extract `useListParams` hook


packages/ra-core/src/controller/useList.ts
```diff
@@ -0,0 +1,233 @@
...
+        resource,
+        JSON.stringify(query),
+        JSON.stringify(location),
+    ];
+
+    const changeParams = useCallback(action => {
+        const newParams = queryReducer(query, action);
+        dispatch(
+            push({
+                ...location,
```
fzaninotto: I believe you don't have to pass the current location - the `history` package uses the current location by default. That way, if you add the `resource` to the function signature, you can pass an empty array as dependency arguments to `useCallback` and never worry about cache anymore.

packages/ra-core/src/controller/useList.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+interface Actions {
+    changeParams: (action: any) => void;
+    setPage: (page: number) => void;
+    setPerPage: (pageSize: number) => void;
+    setSort: (sort: Sort) => void;
+}
+
+const useList = ({
```
fzaninotto: I'd call it useListParams

fzaninotto:
Why didn't you move setFilter()?

djhi:
> Why didn't you move setFilter()?

I was planning to extract a separate hook for filters but it may be too soon. I'll try to transform the controller into a single hook first

packages/ra-core/src/controller/useList.ts
djhi: `useCrudList` ?

fzaninotto: we tend to use the Crud word for communication with the dataProvider. this hook does not communicate with the dataProvider, so I'm -1 for that name.

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -0,0 +1,282 @@
...
+    filterDefaultValues,
+    sort = {
+        field: 'id',
+        order: SORT_ASC,
+    },
+    perPage = 10,
+    filter,
+    debounce = 500,
+}: Props): [Query, Actions] => {
+    const [displayedFilters, setDisplayedFilters] = useState({});
```
fzaninotto: This hoos should only contain the query-related logic, so I think that displayed filters, data from the provider, and the initial effect should stay in the controller (or go to another hook).

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -0,0 +1,282 @@
...
+    changeParams: (action: any) => void;
+    setPage: (page: number) => void;
+    setPerPage: (pageSize: number) => void;
+    setSort: (sort: Sort) => void;
+    setFilters: (filters: any) => void;
+    hideFilter: (filterName: string) => void;
+    showFilter: (filterName: string, defaultValue: any) => void;
+}
+
+const useListParams = ({
```
fzaninotto: I'd love a small jsDoc explaining the return value

packages/ra-core/src/controller/useListParams.ts
djhi: About the filters, that would mean moving the filters actions out too

fzaninotto: right

djhi: Maybe too soon to do that ? What about renaming them `activeFilters` and `setActiveFilters` ?

packages/ra-core/src/controller/useListParams.ts
djhi: Added some

PR #3225, opened 2019-05-15
[RFR] Migrate simple example to hooks

zyhou:
- [x] Migrate `PostList.js` hooks
- [x] Migrate `PostCreate.js` hooks
- [x] Migrate `ResetViewsButton.js` hooks
- [x] Migrate `CommentEdit.js` to hooks
- [x] Migrate `CommentList.js` to hooks
- [x] Migrate `PostPreview.js` to hooks
- [x] Migrate `PostQuickCreate.js`hooks
- [x] Migrate `PostQuickCreateCancelButton.js` to hooks
- [x] Migrate `PostReferenceInput.js` to hooks
- [x] Migrate `UserEdit.js` hooks
- [x] `useSelector` `isLoading` (don't remove the props)

Review with "Hide whitespace changes" params

### DIFF
--- a/examples/simple/src/comments/CommentEdit.js
+++ b/examples/simple/src/comments/CommentEdit.js
@@ -1,6 +1,6 @@
 import Card from '@material-ui/core/Card';
 import Typography from '@material-ui/core/Typography';
-import { withStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/styles';
 import React from 'react';
 import {
     AutocompleteInput,
@@ -25,7 +25,7 @@ const LinkToRelatedPost = ({ record }) => (
     </Link>
 );
 
-const editStyles = {
+const useEditStyles = makeStyles({
     actions: {
         float: 'right',
     },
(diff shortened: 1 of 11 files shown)


examples/simple/src/posts/PostList.js
```diff
@@ -91,58 +96,64 @@ const PostPanel = ({ id, record, resource }) => (
...
-                />
-            }
-            medium={
-                <Datagrid rowClick={rowClick} expand={<PostPanel />}>
-                    <TextField source="id" />
-                    <TextField source="title" cellClassName={classes.title} />
-                    <DateField
-                        source="published_at"
-                        cellClassName={classes.publishedAt}
+const PostList = ({ ...props }) => {
```
fzaninotto:
```suggestion
const PostList = props => {
```

examples/simple/src/posts/PostCreate.js
```diff
@@ -24,31 +24,19 @@ import {
 const saveWithNote = (values, basePath, redirectTo) =>
     crudCreate('posts', { ...values, average_note: 10 }, basePath, redirectTo);
 
-class SaveWithNoteButtonComponent extends Component {
-    handleClick = () => {
-        const { basePath, handleSubmit, redirect, saveWithNote } = this.props;
+const SaveWithNoteButton = props => {
+    const dispatch = useDispatch();
+
+    const handleClick = () => {
```
djhi: You should wrap this into a `useCallback`

examples/simple/src/posts/ResetViewsButton.js
```diff
@@ -1,35 +1,33 @@
...
 
-class ResetViewsAction extends Component {
-    handleClick = () => {
-        const { basePath, startUndoable, resource, selectedIds } = this.props;
-        startUndoable(
-            crudUpdateMany(resource, selectedIds, { views: 0 }, basePath)
+const ResetViewsButton = props => {
+    const dispatch = useDispatch();
+
+    const handleClick = () => {
```
djhi: Same

examples/simple/src/comments/PostPreview.js
```diff
@@ -1,24 +1,24 @@
...
+                : null,
+        [props.resource, props.id],
+    );
+    const version = useSelector(state => state.admin.ui.viewVersion);
 
-const mapStateToProps = (state, props) => ({
-    record: state.admin.resources[props.resource]
-        ? state.admin.resources[props.resource].data[props.id]
-        : null,
-    isLoading: state.admin.loading > 0,
```
fzaninotto: This wasn't ported in the hooks. Even if it's not used in the render function, the fact that it changes forces a rerender. I this it was necessary.

examples/simple/src/comments/PostReferenceInput.js
```diff
@@ -15,156 +13,143 @@ import {
...
+                    { field: 'id', order: 'DESC' },
+                    {},
+                ),
+            );
+        },
+        [state.new_post_id],
+    );
 
-    handleNewClick = event => {
+    const handleNewClick = event => {
```
fzaninotto: wrap with useCallback

examples/simple/src/posts/PostCreate.js
```diff
@@ -26,14 +26,16 @@ const saveWithNote = (values, basePath, redirectTo) =>
...
-    const handleClick = () => {
-        const { basePath, handleSubmit, redirect } = props;
-
-        return handleSubmit(values => {
-            dispatch(saveWithNote(values, basePath, redirect));
-        });
-    };
+    const handleClick = useCallback(
+        () => {
+            return handleSubmit(values => {
```
fzaninotto: why do you `return` instead of using the short form?

zyhou:
@djhi Can you take a look on `PostQuickCreate.js` I don't understand why I can't use `useSelector` in this case.

If you use `useSelector`, we need to click on two times on "save" button

fzaninotto:
@zyhou I fixed the useSelector bug

examples/simple/src/comments/CommentEdit.js
```diff
@@ -1,6 +1,6 @@
 import Card from '@material-ui/core/Card';
 import Typography from '@material-ui/core/Typography';
-import { withStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/styles';
```
fzaninotto: Please use `makeStyles` from `@material-ui/core/styles` (here an in other imports). I don't remember why, but it's the one we should use.

fzaninotto:
No, useCallback wasn't the cause of the problem.

djhi:
> No, useCallback wasn't the cause of the problem.

I didn't say it was. I just think they aren't necessary unless a performance issue arise.

djhi:
I really think we are over-complicating things with those `useCallback`. They shouldn't be the default and it probably shouldn't be what we show in an example.

fzaninotto:
Let's merge it like that. We'll review it once the core performance gets on par with 2.0.

PR #3228, opened 2019-05-15
[RFR] Reference hooks

ThieryMichel:
Refactor ReferenceFieldController using hooks and introduce useReference hooks.
ReferenceMany will follow.

### DIFF
--- a/packages/ra-core/src/controller/ListController.spec.tsx
+++ b/packages/ra-core/src/controller/ListController.spec.tsx
@@ -9,7 +9,7 @@ import ListController, {
     sanitizeListRestProps,
 } from './ListController';
 
-import TestContext from '../util/TestContext';
+import renderWithRedux from '../util/renderWithRedux';
 import { CRUD_CHANGE_LIST_PARAMS } from '../actions';
 
 describe('ListController', () => {
@@ -60,24 +60,13 @@ describe('ListController', () => {
                 children: fakeComponent,
             };
 
-            let reduxStore;
-            let dispatch;
-
-            const { getByLabelText } = render(
-                <TestContext
-                    initialState={{
-                        admin: {
-                            resources: { posts: { list: { params: {} } } },
-                        },
-                    }}
-                    enableReducers
-                >
-                    {({ store }) => {
-                        reduxStore = store;
-                        dispatch = jest.spyOn(store, 'dispatch');
-                        return <ListController {...props} />;
-                    }}
-                </TestContext>
+            const { getByLabelText, dispatch, reduxStore } = renderWithRedux(
+                <ListController {...props} />,
+                {
+                    admin: {
+                        resources: { posts: { list: { params: {} } } },
+                    },
+                }
             );
             const searchInput = getByLabelText('search');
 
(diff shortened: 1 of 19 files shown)


packages/ra-core/src/controller/field/ReferenceFieldController.tsx
```diff
@@ -1,10 +1,6 @@
-import { Component, ReactNode } from 'react';
-import { connect } from 'react-redux';
-import get from 'lodash/get';
-
-import { crudGetManyAccumulate as crudGetManyAccumulateAction } from '../../actions';
-import { linkToRecord } from '../../util';
-import { Record, Dispatch } from '../../types';
+import { SFC, ReactNode, ReactElement } from 'react';
```
djhi: I think they renamed `SFC` into `FunctionalComponent` in the latest typings. Probably not a good idea to upgrade the typings in this PR though

packages/ra-core/src/controller/ListController.spec.tsx
```diff
@@ -65,7 +65,7 @@ describe('ListController', () => {
 
             const { getByLabelText } = render(
                 <TestContext
-                    store={{
+                    initialState={{
```
djhi: Why don't you use your shiny new `renderWithRedux` ? Same for all tests using TestContext

ThieryMichel: I was afraid of the rebase ....

djhi: What rebase ?

packages/ra-core/src/controller/field/ReferenceFieldController.tsx
djhi: For reference: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30364

packages/ra-core/src/util/renderWithRedux.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { render } from 'react-testing-library';
+
+import TestContext from './TestContext';
+
+export default (component, initialState) => {
```
fzaninotto: Please write down example usage in jsDoc

packages/ra-core/src/controller/field/ReferenceFieldController.tsx
```diff
@@ -54,69 +48,11 @@ interface Props {
...
-    mapStateToProps,
-    {
-        crudGetManyAccumulate: crudGetManyAccumulateAction,
-    }
-)(UnconnectedReferenceFieldController);
+export const ReferenceFieldController: SFC<Props> = ({
+    children,
+    ...props
+}) => {
+    return children(useReference(props)) as ReactElement<any>;
```
fzaninotto: missing usage of ChildrenFuncParams

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -0,0 +1,93 @@
+import { useEffect } from 'react';
+// @ts-ignore
+import { useDispatch, useSelector } from 'react-redux';
+import get from 'lodash/get';
+
+import { crudGetManyAccumulate } from '../../actions';
+import { linkToRecord } from '../../util';
+import { Record, ReduxState } from '../../types';
+
+interface Props {
```
fzaninotto: you shouldn't call it `props`, it's not a component

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Fetch reference record, and return it when avaliable
+ *
+ * The reference prop sould be the name of one of the <Resource> components
+ * added as <Admin> child.
+ *
+ * @example
+ * <ReferenceField label="User" source="userId" reference="users">
```
fzaninotto: bad example, show example usage of the hook instead

packages/ra-core/src/controller/field/useReference.ts
djhi: I think I named mine `Props` too in the `useListParams`

djhi: Btw, although I can understand this may be misleading with the React props, they are indeed props|properties of an object passed as a parameter to the hooks.

PR #3236, opened 2019-05-16
[RFR] Reference many hooks

ThieryMichel:
- [x] refactor ReferenceManyFieldCOntroller from calls component to functional component
- [x] use hooks to inject redux state
- [x] add useReferenceMany hooks
- [x] use referenceMany hooks
- [x] add doc


fzaninotto:
what's left for RFR?

packages/ra-core/src/controller/usePagination.ts
```diff
@@ -0,0 +1,47 @@
+import { useState, useEffect } from 'react';
```
fzaninotto: I'd call this hook `usePaginationState`, because `usePagination` is too generic IMO. Same for `useSort`

packages/ra-core/src/controller/useSort.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param {Object} initialSort
+ * @param {string} initialSort.resource The current resource name
+ * @param {string} initialSort.reference The linked resource name
+ * @returns {SortProps} The sort props
+ */
+export default (
+    initialSort: Sort = { field: 'id', order: 'DESC' }
+): SortProps => {
+    const [sort, dispatch] = useReducer(sortReducer, initialSort);
+    useEffect(() => dispatch(initialSort), [
```
fzaninotto: Can't we avoid dispatching that effect on mount, as we already have the initial sort in the reducer?

packages/ra-ui-materialui/src/field/ReferenceManyField.js
```diff
@@ -106,22 +106,65 @@ ReferenceManyFieldView.propTypes = {
...
-                <ReferenceManyFieldView
-                    {...props}
-                    {...{ children, ...controllerProps }}
-                />
-            )}
-        </ReferenceManyFieldController>
+        <ReferenceManyFieldView
+            {...props}
+            {...{
+                children,
```
fzaninotto: I find it less readable if you explicit all props here, as compared to what we did before.

fzaninotto:
Thanks!

PR #3238, opened 2019-05-16
[RFR] Redux-Saga Error Handling

liegri:
The goal of this change is better error handling with redux saga.

Currently, if there is an unhandled error thrown in a redux saga, **all saga execution is stopped** without the possibility to catch those errors.

In the proposed change, an `onError` handler for redux saga (see `createSagaMiddleware(options)`
 in https://redux-saga.js.org/docs/api/) can be passed through the `CoreAdmin` component to `createSagaMiddleware`. This provides a convenient error handling solution to all react-admin users.

In the Future, it might also be practical to restart saga execution after an error was handled.

### DIFF
--- a/packages/ra-core/src/CoreAdmin.tsx
+++ b/packages/ra-core/src/CoreAdmin.tsx
@@ -48,6 +48,7 @@ export interface AdminProps {
     menu?: ComponentType;
     theme?: object;
     title?: TitleComponent;
+    onSagaError?(error: Error): void;
 }
 
 interface AdminContext {
@@ -163,6 +164,7 @@ React-admin requires a valid dataProvider function to work.`);
             i18nProvider,
             initialState,
             locale,
+            onSagaError,
         } = this.props;
 
         return this.reduxIsAlreadyInitialized ? (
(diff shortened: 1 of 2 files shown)


fzaninotto:
Also, can you explain the problem it solves? Which bug did you meet that was due to a saga throwing an error?

liegri:
> Also, can you explain the problem it solves? Which bug did you meet that was due to a saga throwing an error?

This isn't a bugfix, but more a new feature. The problem is, that when you add custom sagas to your react admin page, and one of those throws an error, it bubbles up to the root saga which crashes. As described in the Redux documentation, in such a case all saga execution is stopped, without visibly crashing the page. This includes react-admin sagas and would result in extremely unexpected behavior.

This feature implements a way of displaying/logging/handling such errors.

fzaninotto:
What is the default behavior? Is the error swallowed, or is it at least logged to the console? Also, please document that new prop in the Admin documentation.

fzaninotto:
Bump, can you please document the new feature?

liegri:
> Bump, can you please document the new feature?

Yes, if I find the time. Possibly next week.

fzaninotto:
This PR hasn't been updated for too long, so I'm closing it. If the author finds time to answer the questions, we'll consider reopening it.

PR #3243, opened 2019-05-17
[RFR] Authenticate NotFound page

### LINKED ISSUE #3230: Default 404 page should forward to login when unauthenticated
**What you were expecting:**
When I'm setting up `authProvider` in `Admin` I'm expecting to not allow unauthenticated user to see internal app structure, ( especially the **Menu**), i.e not allow to get further than **Login** page. Even if I go via unexisting path, default 404 page should forward me to login, but not show the Layout.

**What happened instead:**
If user enters random url path, he is forwarded to default **404** page, and shows **Menu**, even user is unauthenticated.
This can be roughly fixed by implementing own **NotFound** page, wrapped in `<Authenticated> `HOC and pass it in `catchAll` prop, so it will forward you to Login page in this case.
This forces me to duplicate default  RA’s **NotFound** page code, if I want to keep page the same, or implement it from scratch, even if I’m OK with default.
In rare case, If you want to show **404** page for unauthenticated user for unexisting path (via `catchAll`), but hide the **Menu** in Layout - you have no option to set this as with `Route` in `customRoutes` with `noLayout` prop.

**Steps to reproduce:**
Setup Application with Auth.
Go to any unexisting url like "www.your-ra-app.com/unexistingurl" before Logging. 
See NotFount page with all Layout visible.

**Related code:**
Reproduced on default sandbox code:

**Environment**

* React-admin version: 2.8.6
* React version: 16.8.6
* Browser: Chrome

Kmaschta:
Fixes #3230

### DIFF
--- a/packages/ra-core/src/RoutesWithLayout.tsx
+++ b/packages/ra-core/src/RoutesWithLayout.tsx
@@ -71,8 +71,9 @@ const RoutesWithLayout: SFC<Props> = ({
                 />
             ) : null}
             <Route
-                render={() =>
+                render={routeProps =>
                     createElement(catchAll, {
+                        ...routeProps,
                         title,
                     })
                 }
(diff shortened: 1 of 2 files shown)


Kmaschta:
How can we see that there is anonymous-accessible pages?

I agree, but currently we have a big part of the app (layout, custom menu, etc) that are accessible by un-logged people.

fzaninotto:
Thanks!

Issue #2952, opened 2019-03-04
RFC: Decouple data actions from views


fzaninotto:
## Problem

The page components (`List`, `Edit`, etc.). dispatch crud actions on mount to recuperate the data from the data provider. These actions are data oriented, i.e. the `crudGetList` asks for a list of resources, the `crudGetOne` asks for a single record, etc. But we've added side effects to these actions (e.g. redirect to the list when the crudGetOne returns a 404) that are view oriented, i.e. they are only valid for a given view.

```ts
export const crudGetOne = (
    resource: string,
    id: Identifier,
    basePath: string,
    refresh: RefreshSideEffect = true
): CrudGetOneAction => ({
    type: CRUD_GET_ONE,
    payload: { id },
    meta: {
        resource,
        fetch: GET_ONE,
        basePath,
        onFailure: {
            notification: {
                body: 'ra.notification.item_doesnt_exist',
                level: 'warning',
            },
            redirectTo: 'list',
            refresh,
        },
    },
});
```

The problem occurs when trying to use the crud- actions for a custom component. When calling a crudGetOne to grab e.g. the basket of an order, the user doesn't want to be redirected to the list in case of 404 on the basket resource.

## Solution

The side effects of the crud actions should be configurable, and by default, there should be none. Something like:

```ts
export const crudGetOne = (
    resource: string,
    id: Identifier,
    basePath: string,
    onSuccess: {},
    onFailure: {}
): CrudGetOneAction => ({
    type: CRUD_GET_ONE,
    payload: { id },
    meta: {
        resource,
        fetch: GET_ONE,
        basePath,
        onSuccess,
        onFailure,
    },
});
```

It would be up to the caller (the controller) to specify the side effects

```diff
// in EditController.js
    updateData(resource = this.props.resource, id = this.props.id) {
-       this.props.crudGetOne(resource, id, this.props.basePath);
+      this.props.crudGetOne(resource, id, this.props.basePath, {}, {
+           notification: {
+               body: 'ra.notification.item_doesnt_exist',
+               level: 'warning',
+           },
+           redirectTo: 'list',
+           refresh: true,
+       });
    }
```

This is a great improvement in decoupling the data layer and the vue layer, and to promote the reusability of our action creators. It's also a breaking change.

Thoughts?

fzaninotto:
As of v3-alpha.4, this is done. We no longer use the action creators (which include side effects) directly, but we use hooks and the controllers decide which side effects to call.

PR #3248, opened 2019-05-19
[WIP] add useOptimisticQuery hook

fzaninotto:
Allows to group two pieces of controller logic:

- fetch the dataProvider using a Redux action
- get the result from the store

I also introduced a new reducer to store the loading state of every request, which allows to pass that information to the view. 

I updated the Edit and Show controllers to use that new hook. The `isLoading` prop is now relative to the main query (not the global loading state).

### DIFF
--- a/packages/ra-core/src/actions/dataActions/crudGetOne.ts
+++ b/packages/ra-core/src/actions/dataActions/crudGetOne.ts
@@ -10,23 +10,14 @@ import {
 export const crudGetOne = (
     resource: string,
     id: Identifier,
-    basePath: string,
-    refresh: RefreshSideEffect = true
+    meta: any
 ): CrudGetOneAction => ({
     type: CRUD_GET_ONE,
     payload: { id },
     meta: {
         resource,
         fetch: GET_ONE,
-        basePath,
-        onFailure: {
-            notification: {
-                body: 'ra.notification.item_doesnt_exist',
-                level: 'warning',
-            },
-            redirectTo: 'list',
-            refresh,
-        },
+        ...meta,
     },
 });
 
@@ -41,12 +32,7 @@ export interface CrudGetOneAction {
     readonly meta: {
         resource: string;
         fetch: typeof GET_ONE;
-        basePath: string;
-        onFailure: {
-            notification: NotificationSideEffect;
-            redirectTo: RedirectionSideEffect;
-            refresh: RefreshSideEffect;
-        };
+        [key: string]: any;
     };
 }
 
@@ -68,11 +54,9 @@ export interface CrudGetOneFailureAction {
     readonly requestPayload: RequestPayload;
     readonly meta: {
         resource: string;
-        notification: NotificationSideEffect;
-        redirectTo: RedirectionSideEffect;
-        refresh: RefreshSideEffect;
         fetchResponse: typeof GET_ONE;
         fetchStatus: typeof FETCH_ERROR;
+        [key: string]: any;
     };
 }
 
@@ -87,5 +71,6 @@ export interface CrudGetOneSuccessAction {
         resource: string;
         fetchResponse: typeof GET_ONE;
         fetchStatus: typeof FETCH_END;
+        [key: string]: any;
     };
 }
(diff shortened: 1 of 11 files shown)


packages/ra-core/src/actions/dataActions/crudGetOne.ts
```diff
@@ -10,23 +10,14 @@ import {
...
-    basePath: string,
-    refresh: RefreshSideEffect = true
+    meta: any
 ): CrudGetOneAction => ({
     type: CRUD_GET_ONE,
     payload: { id },
     meta: {
         resource,
         fetch: GET_ONE,
-        basePath,
```
fzaninotto: decoupling side effects, refs #2952

packages/ra-core/src/controller/EditController.tsx
```diff
@@ -85,23 +87,29 @@ const EditController = (props: Props) => {
...
-    );
-
     const version = useSelector(
         (state: ReduxState) => state.admin.ui.viewVersion
     );
 
+    const { data: record, loading } = useGetOne(
+        resource,
+        id,
+        {
```
fzaninotto: now it's the controller who decides the side effects, not the action creator

packages/ra-core/src/controller/useOptimisticQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+ * );
+ *
+ * @param {object} action The action to dispatch, e.g. { type 'RA/GET_ONE', payload: { id: 123 }, meta: { fetch: GET_ONE, resource: 'posts' } }
+ * @param {function} selector The selector function to get the result from the store
+ * @param {number} version The version number. Increase to force a new fetch
+ *
+ * @returns {GetOneResponse} An object with the shape { data, error, loading, loaded }
+ */
+const useOptimisticQuery = (
+    action: FetchAction,
```
fzaninotto: I'm not so sure about the signature of this one. Should it accept:
- an action returned by an action creator, as I did here, or
- a quadruplet `{ type, resource, payload, meta }` , like the `useQuery` action

In the latter case, we don't need the crud action creators anymore.

fzaninotto:
After a good night's sleep, and the discernment of a week day, I think this experimentation is useful but does not show the best solution.

1. Using the Redux store to store loading state of requests is an heresy. No component will need to access that state, so it better be local. We've already found a nice way to manage the loading state with the `useDataProvider` hook, so let's use this one.
2. Action creators, when removing side effects, are all the same. They just populate the payload with their arguments, and craft the action type based on the crud verb. So they are useless - we could replace them all by a single function. As a consequence, we shouldn't use them, and instead promote the hook alternative (`useGetOne`, `useGetList`, etc), which actually return the content. We don't need to remove them from the core (that would be another breaking change), but we should deprecate them and remove them in a future major version.
3. `useQuery` could benefit from the optimistic approach, too. It just needs a selector function to get the result from the store, and a custom reducer to store all `CUSTOM_FETCH` response payloads somewhere in the store.
4. So there wouldn't be much difference between `useQuery` and `useOptimisticQuery`. We could merge the two implementations. A non-optimistic query would just be called with an empty selector.
5. If we're using a `useQuery`-like API in our controllers, we have the benefit of a Promise API, so declarative side effects in the metas don't make much sense. I think the controllers should just `dispatch` the success and failure side effects themselves. That way, there would be only one way to dispatch side effects, and we could get rid of a lot of sagas. Also in a future major version.
6. A `useQuery` accepting an action name and a selector starts having too many arguments, so we must change its signature, to something like:

```js
useQuery(query, options);

//example
const useGetOne = (resource, id, sideEffects) => useQuery(
  { type: GET_ONE, resource: 'posts', payload: ' { id: 123 } },
  { action: CRUD_GET_ONE, selector: state => state.admin.resources[resource].data[id], meta: sideEffects }
``` 

It's not completely similar to the `dataProvider` signature anymore, but I think it's for the better.

So I'm going to close this PR and start working on a better implementation.

(note for self: do not work on weekends)

Issue #3249, opened 2019-05-20
ArrayInput does not like primitive values


Veikedo:
**What happened instead:**

**Steps to reproduce:**
I have this form 
```
<ArrayInput source='configuration.xpathexpressions' label='XPath expressions'>
  <SimpleFormIterator>
    <TextInput label='XPath expression' />
  </SimpleFormIterator>
</ArrayInput>
```

Note - I don't have source prop in `TextInput`. 
And here what I see in the browser

![image](https://user-images.githubusercontent.com/5779339/58015701-961bca00-7b04-11e9-8e2e-06f3365a99b8.png)

The problem is with this line:
https://github.com/marmelab/react-admin/blob/5d2a88df4ed31d1f61aa0fee5519ee637354bedb/packages/ra-ui-materialui/src/form/SimpleFormIterator.js#L102

If we make it just `fields.push();` everything would work as expected, even for complex objects (ie with fields with `source` props)

**CodeSandbox**

https://codesandbox.io/s/simple-cwjuz

Here is the form with array input: https://cwjuz.codesandbox.io/#/tags/5

**Other information:**
A question on this topic https://stackoverflow.com/questions/51468966/react-admin-array-input-with-array-of-strings-numbers

**Environment**

* React-admin version: ^2.9.1
* Last version that did not exhibit the issue (if applicable):
* React version: ^16.8.6

Veikedo:
Create PR https://github.com/marmelab/react-admin/pull/3250

mleczakm:
Are there any easy way to override this method ( `_this.addField` in `SimpleFormIterator`) in current implementation?

taitems:
I'm hitting this issue now. Is there any interest in firing up the PR again, or a suitable workaround? I just want to have an array of URLs.

PR #3250, opened 2019-05-20
Support for primitive values in ArrayInput

Veikedo:
PR for https://github.com/marmelab/react-admin/issues/3249

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -99,7 +99,7 @@ export class SimpleFormIterator extends Component {
     addField = () => {
         const { fields } = this.props;
         this.ids.push(this.nextId++);
-        fields.push({});
+        fields.push();
     };
 
     render() {


packages/ra-ui-materialui/src/form/SimpleFormIterator.js
```diff
@@ -99,7 +99,7 @@ export class SimpleFormIterator extends Component {
     addField = () => {
         const { fields } = this.props;
         this.ids.push(this.nextId++);
-        fields.push({});
+        fields.push();
```
djhi: Isn't this equivalent to not calling push at all?

Veikedo: No, without this it won't add anything

djhi: I think calling `array.push` without any argument does not modify the array and is equivalent to not calling the function at all. At least this is what I see when trying this in a console

Kmaschta: @djhi You're right 
image

Veikedo: It would be right if `fields` were just a standard mutating array, however it's not.
See here https://redux-form.com/8.2.0/docs/api/fieldarray.md/#-code-fields-push-value-any-function-code-

djhi:
This change breaks the current support for objects in ArrayInput. See the failing tests

smeng9:
For backward compatibility of this fix, we could check if `<SimpleFormIterator>` has an only child input element and if the child input element does not have a `source` property. If that's the case, we can push primitive values, else we will still push empty `{}` for objects in `<ArrayInput>`

djhi:
Not enough. The following wouldn't work:

```
<ArrayInput source="authors">
    <SimpleFormIterator>
        <DateInput source="identity.first_name" />
    </SimpleFormIterator>
</ArrayInput>
```

I see two options. Either introduce a `SingleValueFormIterator` or add a `getNewValue` prop on the `SimpleFormIterator`

Kmaschta:
Any news on this PR?

djhi:
No news for some time, closing. Feel free to ping us here to reopen it if needed

PR #3251, opened 2019-05-20
#3132 handle permanent filters on props update in list controller

### LINKED ISSUE #3132: Permanent filter mapped to state value doesn't filter correctly
**What you were expecting:**

I mapped a permanent list filter to a state value from the redux store (with custom reducer). Let's say that's value1.
When the state value is changed, the list should be filtered by this new value (value2).

**What happened instead:**

Instead, the list re-renders, but it is still filtered by value1. When changing the value to value3, the list gets filtered with value2 (so always one "step" back). I logged the mapped prop to the console in the list component and it is correct. Only the api parameters "stay behind".

**Related code:**
I set up a Codesandbox [here](https://codesandbox.io/s/r656z506p). Unfortunately I get some console errors there that I don't get in my real-case scenario (might be related to using a fake data provider, but I couldn't figure them out).

* React-admin version: 2.8.5
* React version: 16.7

Thank you for considering this.

tarnowsc:
Fixes #3132

### DIFF
--- a/packages/ra-core/src/controller/ListController.spec.tsx
+++ b/packages/ra-core/src/controller/ListController.spec.tsx
@@ -113,6 +113,22 @@ describe('ListController', () => {
             expect(props.changeListParams.mock.calls[0][1].filter).toEqual({});
         });
 
+        it('should update pernament filters', () => {
+            const props = {
+                ...defaultProps,
+                debounce: 200,
+                crudGetList: jest.fn(),
+                filter: { foo: 1 },
+                children: fakeComponent,
+            };
+
+            const wrapper = shallow(<ListController {...props} />);
+            wrapper.setProps({ filter: { foo: 2 } });
+            clock.tick(200);
+
+            expect(props.crudGetList.mock.calls[1][3].foo).toEqual(2);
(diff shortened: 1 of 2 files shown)


djhi:
Can you please give a bit of context for your changes? What problem did you have, linked issues, etc.?

Thanks!

tarnowsc:
The issue was described here: https://github.com/marmelab/react-admin/issues/3132
Basically when props get updated the updateData function is using previous props and it ends up in the situation that the list permanent filter state is "one click behind".
I think that the solution is more or less ok but I would appreciate feedback or better ideas to solve that.

packages/ra-core/src/controller/ListController.spec.tsx
```diff
@@ -113,6 +113,22 @@ describe('ListController', () => {
             expect(props.changeListParams.mock.calls[0][1].filter).toEqual({});
         });
 
+        it('should update pernament filters', () => {
```
Kmaschta: s/pernament/permanent/

fzaninotto:
Superseded by #3308

Issue #3108, opened 2019-04-08
[Performance]Too many request to API endpoint when using ReferenceInput and ReferenceArrayInput 


tunglt1810:
Hi, first words, thanks for your awesome project.

I have use ReferenceInput and ReferenceArrayInput in my form like this:

```
<ReferenceArrayInput source="parentRoleIds" reference="roles">
                <CheckboxGroupInput optionValue="id" optionText="name"/>
</ReferenceArrayInput>
```

But when i check the checkbox, this component send too much request to server. I think it's not necessary. 

![ReferenceInput](https://user-images.githubusercontent.com/19906349/55715748-99109e80-5a1f-11e9-8ece-fc59c1846860.gif)

After the input mounted and load reference data. It's should be not need load reference data again when changing the value of input

djhi:
Hi, and thanks for your question. As explained in the [react-admin contributing guide](https://github.com/marmelab/react-admin/blob/master/README.md#contributing), the right place to ask a "How To" question, get usage advice, or troubleshoot your own code, is [StackOverFlow](http://stackoverflow.com/questions/tagged/react-admin).

This makes your question easy to find by the core team, and the developer community. Unlike Github, StackOverFlow has great SEO, gamification, voting, and reputation. That's why we chose it, and decided to keep GitHub issues only for bugs and feature requests.

So I'm closing this issue, and inviting you to ask your question at:

http://stackoverflow.com/questions/tagged/react-admin

And once you get a response, please continue to hang out on the react-admin channel in StackOverflow. That way, you can help newcomers and share your expertise!

tunglt1810:
Hi @djhi , it's my mistake of using "word". I've update this post

djhi:
Thanks for reporting this. If you are able to illustrate the bug or feature request with an example, please provide a sample application via one of the following means:

* CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple)
* A sample application via GitHub

fzaninotto:
I can confirm the problem on the simple example:

- Open PostEdit
- Change tags input to use CheckboxGroupInput:

    ```jsx
                <ReferenceArrayInput
                    reference="tags"
                    source="tags"
                    filter={{ published: true }}
                >
                    <CheckboxGroupInput optionText="name" />
                </ReferenceArrayInput>
    ```

Each click on a checkbox calls the dataProvider, as seen on this screenshot:

[...]

PR #3252, opened 2019-05-20
[RFR] Remove duplicate API call in ReferenceArrayInput

fargito:
Fixes https://github.com/marmelab/react-admin/issues/3108

### DIFF
--- a/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
@@ -436,7 +436,7 @@ describe('<ReferenceArrayInputController />', () => {
         ).toEqual([defaultProps.reference, [6]]);
     });
 
-    it('should call crudGetMany with empty list when already fetched input value changes', () => {
+    it('should not call crudGetMany when already fetched input value changes', () => {
         const crudGetMany = jest.fn();
         const wrapper = shallow(
             <ReferenceArrayInputController
@@ -447,15 +447,15 @@ describe('<ReferenceArrayInputController />', () => {
             />
         );
         expect(
-            crudGetMany.mock.calls[crudGetMany.mock.calls.length - 1]
+            crudGetMany.mock.calls[0]
         ).toEqual([defaultProps.reference, [5, 6]]);
         wrapper.setProps({ input: { value: [6] } });
         expect(
-            crudGetMany.mock.calls[crudGetMany.mock.calls.length - 1]
-        ).toEqual([defaultProps.reference, []]);
+            crudGetMany.mock.calls.length
+        ).toEqual(1);
     });
 
-    it('should call crudGetOne and crudGetMatching when record changes', () => {
+    it('should only call crudGetOne and not crudGetMatching when only the record changes', () => {
         const crudGetMany = jest.fn();
         const crudGetMatching = jest.fn();
         const wrapper = shallow(
@@ -471,6 +471,6 @@ describe('<ReferenceArrayInputController />', () => {
         assert.equal(crudGetMany.mock.calls.length, 1);
         wrapper.setProps({ record: { id: 1 } });
         assert.equal(crudGetMatching.mock.calls.length, 2);
-        assert.equal(crudGetMany.mock.calls.length, 2);
+        assert.equal(crudGetMany.mock.calls.length, 1);
     });
 });
--- a/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
@@ -417,6 +417,44 @@ describe('<ReferenceArrayInputController />', () => {
         assert.equal(crudGetMany.mock.calls.length, 2);
     });
 
+    it('should call crudGetMany when input value changes only with the additional input values', () => {
+        const crudGetMany = jest.fn();
+        const wrapper = shallow(
+            <ReferenceArrayInputController
+                {...defaultProps}
+                input={{ value: [5] }}
+                allowEmpty
+                crudGetMany={crudGetMany}
+            />
+        );
+        expect(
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -232,7 +233,8 @@ export class UnconnectedReferenceArrayInputController extends Component<
                     'The value of ReferenceArrayInput should be an array'
                 );
             }
-            crudGetMany(reference, ids);
+            const idsToFetch = isInitialCall ? ids : ids.filter(id => !this.props.input.value.includes(id))
```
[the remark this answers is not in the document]
djhi: Please use english :wink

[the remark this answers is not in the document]
fargito: > Ca marche https://lodash.com/docs/4.17.11#difference ?

Wouldn't https://lodash.com/docs/4.17.11#difference work better ?

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -4,6 +4,7 @@ import debounce from 'lodash/debounce';
 import compose from 'recompose/compose';
 import { createSelector } from 'reselect';
 import isEqual from 'lodash/isEqual';
+import difference from 'lodash/difference'
```
khoanguyenxuan: Should we use the comma at the end of line style?

djhi: Yes, please configure your editor to apply our prettier rules

fargito: Ah yes my bad. @djhi I noticed that `make prettier` changed a whole bunch of files that don't seem to respect those rules. Would it be ok to create a separate pull request with all those changes ?

djhi: Please don't. Can you just apply prettier on the file you modified? If you can automate it and our make command change too many files, just run the `lint` make command and fix manually. Thanks!

fargito:
@djhi is it ok for you ? :)

djhi:
Looks good but I won't have time to test it before next week. I'm at react europe

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -164,7 +165,8 @@ export class UnconnectedReferenceArrayInputController extends Component<
     }
 
     componentDidMount() {
-        this.fetchReferencesAndOptions(this.props);
```
fzaninotto: Why did you change that?

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -232,7 +234,10 @@ export class UnconnectedReferenceArrayInputController extends Component<
                     'The value of ReferenceArrayInput should be an array'
                 );
             }
-            crudGetMany(reference, ids);
+            const idsToFetch = isInitialCall
```
fzaninotto: Why dod you need the isInitialCall? The difference should work in all cases.

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -417,6 +417,44 @@ describe('<ReferenceArrayInputController />', () => {
...
+        expect(
+            crudGetMany.mock.calls[crudGetMany.mock.calls.length - 1]
+        ).toEqual([defaultProps.reference, [5]]);
+        wrapper.setProps({ input: { value: [5, 6] } });
+        expect(
+            crudGetMany.mock.calls[crudGetMany.mock.calls.length - 1]
+        ).toEqual([defaultProps.reference, [6]]);
+    });
+
+    it('should call crudGetMany with empty list when already fetched input value changes', () => {
```
fzaninotto: No, it should not call crudGetMany at all in that case.

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -232,7 +234,10 @@ export class UnconnectedReferenceArrayInputController extends Component<
...
+                ? ids
+                : difference(ids, this.props.input.value);
+            crudGetMany(reference, idsToFetch);
```
fzaninotto: if `idsToFetch` is empty, the action shouldn't even be dispatched.

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
fargito: Ah ok, that's what I did at first but it broke previous tests. I will edit this

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
fargito: Actually not, when the component just mounted, if we just call `fetchReferences(this.props)`, the difference between the `props` argument of the `fetchReferences` function and `this.props` results in an empty list (they're the same object of course :D). However we still need to call `crudGetMany` with all the values I think

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
fargito: Yes, I will change this :)

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
fargito: I edited this, I think it should be cleaner now

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
fargito: Edited

fzaninotto:
Thanks!

PR #3258, opened 2019-05-21
[RFR] Factorize prettier logic to prettierrc

fargito:
Hello !

I noticed a duplication in the tslint and prettier configuration, so I factorized it in a prettierrc configuration file.

Thanks for your awesome project !


fzaninotto:
Thanks!

PR #3263, opened 2019-05-22
[RFR] Create prettier-fix command

fargito:
Hello ! 

I noticed that it was a bit painful to apply prettier rules on a single file, since calling `make prettier` apply those rules to all the files, even the ones that were not modified recently.

I though it was then profitable to create a specific command to apply the rules to only one file, which was not possible before.

Thanks :)


fzaninotto:
We use our IDEs for single file code formatting. Don't you have the "format on save" option ticked on your side?

fargito:
I haven't configured it yet, but I think it might also benefit to occasional contributor :)

fargito:
Or do you think there could be a guide in the docs on how to configure the IDEs ?

fzaninotto:
Nope, helping people configuring their IDE is not our mission. And I think there are plenty of tutorials about that, more than there would be about using our `make` commands. 

Also the problem with the current `make prettier` command is that it fixes a lot of problems in files that contributors haven't touched. That's what we should do: fix these problems upfront. That way, `make prettier` would only find problems in the code that contributors have touched.

PR #3266, opened 2019-05-22
[WIP] On fetch of single record handle invalidation of old records

### LINKED ISSUE #3227: After some minutes opening an expand clears the list
I'm seeing this locally using react-admin-2.9.1 but the issue is visible on the [demo/invoices page](https://marmelab.com/react-admin-demo/#/invoices) aswell.

The expected behaviour is the one we can see just after opening the page and clicking on the expand arrow: the expand opens and every other record on the list remains visible.

If you open the invoices list, wait for 5-10 minutes (I'm not sure about the exact timeout) and click an expand arrow the expand opens correcly but _**every other**_ item on the list disappear.

sagarbakhtar:
- [x] On CRUD_GET_ONE_SUCCESS action don't invalidate old ids
- [ ] On GET_ONE action don't invalidate old records

Fixes #3227

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/list/ids.spec.ts
+++ b/packages/ra-core/src/reducer/admin/resource/list/ids.spec.ts
@@ -65,4 +65,26 @@ describe('data addRecordIdsFactory', () => {
 
         assert.deepEqual(newState, ['record3', 'record1', 'record2']);
     });
+
+    it.only('should preserve old records when "preserveOldIds" param is true ', () => {
+        const newRecords = ['record1'];
+        const oldRecords = ['record2', 'record3'];
+        const getFetchedAt = jest.fn().mockReturnValue({
+            record1: 'date',
+        });
+        Object.defineProperty(oldRecords, 'fetchedAt', {
+            value: {
+                record2: 'date',
+                record3: 'date',
+            },
+        });
+
+        const newState = addRecordIdsFactory(getFetchedAt)(
+            newRecords,
+            oldRecords,
+            true
+        );
+
+        assert.deepEqual(newState, [...oldRecords, ...newRecords]);
+    });
 });
--- a/packages/ra-core/src/reducer/admin/resource/list/ids.spec.ts
+++ b/packages/ra-core/src/reducer/admin/resource/list/ids.spec.ts
@@ -66,7 +66,7 @@ describe('data addRecordIdsFactory', () => {
         assert.deepEqual(newState, ['record3', 'record1', 'record2']);
     });
 
-    it.only('should preserve old records when "preserveOldIds" param is true ', () => {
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/reducer/admin/resource/list/ids.spec.ts
```diff
@@ -65,4 +65,26 @@ describe('data addRecordIdsFactory', () => {
 
         assert.deepEqual(newState, ['record3', 'record1', 'record2']);
     });
+
+    it.only('should preserve old records when "preserveOldIds" param is true ', () => {
```
djhi: `only` spotted

sagarbakhtar: Thanks, removing it.

packages/ra-core/src/reducer/admin/resource/list/ids.ts
```diff
@@ -28,12 +28,18 @@ type State = IdentifierArrayWithDate;
 
 export const addRecordIdsFactory = getFetchedAtCallback => (
     newRecordIds: IdentifierArrayWithDate = [],
-    oldRecordIds: IdentifierArrayWithDate
+    oldRecordIds: IdentifierArrayWithDate,
+    preserveOldIds: boolean = false
 ): IdentifierArrayWithDate => {
-    const newFetchedAt = getFetchedAtCallback(
+    let newFetchedAt = getFetchedAtCallback(
```
fargito: Is there a reason to replace `const` by `let` ?

sagarbakhtar: @fargito yes, reassigning value to that variable.

fargito: Wouldn't it be cleaner to create another variable named for example `returnedFetchedAt` that would contain, depending on `preserveOldIds`, either `{ ...oldRecordIds.fetchedAt, ...newFetchedAt }` or `{ ...newFetchedAt }` ? And then pass it as argument to the `filter` function ?

sagarbakhtar: Yes, makes sense, will change and update it.

djhi:
Any updates ?

sagarbakhtar:
> Any updates ?

Was working on few other things, will work in this today.

djhi:
:wave: @sagarbakhtar, any chance you can finish this ?

sagarbakhtar:
@djhi will work on this for the next 2 days.

fzaninotto:
any updates?

Luwangel:
Already fixed and get old :older_adult: . Thanks for contributing.

PR #3267, opened 2019-05-22
[RFR] [Rework] Fix sending request by Query with deep props on every update

### LINKED ISSUE #3174: Deep props make Query send request on every update

fargito:
This is a rework of https://github.com/marmelab/react-admin/pull/3176

- Changed shallowEqual to deepEqual in Query to accommodate for possible deep props.

- Changed the placement of myPayload variable.
When myPayload is global the objects are indeed identical and the script worked did pass. However, this assumes independency from props, which is not desired. This can easily lead to endless loops.

Fixes #3174

Closes https://github.com/marmelab/react-admin/pull/3176

### DIFF
--- a/packages/ra-core/src/util/Query.spec.tsx
+++ b/packages/ra-core/src/util/Query.spec.tsx
@@ -213,11 +213,15 @@ describe('Query', () => {
 
     it('should not dispatch a new fetch action when updating with the same query props', () => {
         let dispatchSpy;
-        const myPayload = {};
         const { rerender } = render(
             <TestContext>
                 {({ store }) => {
                     dispatchSpy = jest.spyOn(store, 'dispatch');
+                    const myPayload = {
+                        foo: {
+                            bar: 1,
+                        },
+                    };
                     return (
                         <Query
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

PR #3269, opened 2019-05-23
[RFR] Delegate the redirection after logout to authProvider

### LINKED ISSUE #2957: [RFC] Redirection to login should be done in the authProvider
## Problem

If the authProvider delegates login and check to a third-party server (e.g. OAuth), then it should be able to manage the login form URL.

Currently, the login form URL is hardcoded. The only possibility to redirect to an external authentication server is to use window.location in AUTH_LOGOUT, which occurs *after* a first redirection to '/login' in case of voluntary disconnection. That creates a blinking screen.

extract from the current `auth` saga:

```js
case USER_CHECK: {
    try {
        yield call(authProvider, AUTH_CHECK, payload);
    } catch (error) {
        yield call(authProvider, AUTH_LOGOUT);
        yield put(
            replace({
                pathname: (error && error.redirectTo) || '/login',
                state: { nextPathname: meta.pathName },
            })
        );
    }
    break;
}
case USER_LOGOUT: {
    yield put(
        push(
            (action.payload && action.payload.redirectTo) ||
                '/login'
        )
    );
    yield call(authProvider, AUTH_LOGOUT);
    break;
}

```

## Solution

The auth saga should not do the redirection by itself, but use the return object from AUTH_LOGOUT as login URL. 

This is a breaking change.

djhi:
Fixes #2957 

- [x] Implementation
- [x] Refactor
- [x] Tests
- [x] Update documentation

### DIFF
--- a/packages/ra-core/src/sideEffect/auth.ts
+++ b/packages/ra-core/src/sideEffect/auth.ts
@@ -83,10 +83,11 @@ export const handleCheck = (authProvider: AuthProvider) =>
         try {
             yield call(authProvider, AUTH_CHECK, payload);
         } catch (error) {
-            yield call(authProvider, AUTH_LOGOUT);
+            const redirectTo = yield call(authProvider, AUTH_LOGOUT);
             yield put(
                 replace({
-                    pathname: (error && error.redirectTo) || '/login',
+                    pathname:
+                        (error && error.redirectTo) || redirectTo || '/login',
                     state: { nextPathname: meta.pathName },
                 })
             );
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/sideEffect/auth.ts
```diff
@@ -23,111 +23,115 @@ import {
...
+    function*(action) {
+        const { payload, meta } = action;
+        try {
+            yield put({ type: USER_LOGIN_LOADING });
+            const authPayload = yield call(authProvider, AUTH_LOGIN, payload);
+            yield put({
+                type: USER_LOGIN_SUCCESS,
+                payload: authPayload,
+            });
+            const redirectTo = yield meta.pathName ||
```
Kmaschta: Can we `yield` a string?
Would be more clear to me with `meta.pathName || yield select(...)`

Kmaschta: Wut why

djhi: react-admin-yield-ts

PR #3120, opened 2019-04-11
Datagrid rowClick="toggle"

possi:
Allow mass action selection with click on entire row.

PR #3270, opened 2019-05-23
[RFR] [Rework] Datagrid rowclick toggle

fargito:
This is a rework of https://github.com/marmelab/react-admin/pull/3120

* Allow `rowClick="toggleSelection"` option on DatagridRow
* Add corresponding documentation
* Add tests for all the `rowClick` options

Closes https://github.com/marmelab/react-admin/pull/3120

### DIFF
--- a/packages/ra-ui-materialui/src/list/DatagridRow.js
+++ b/packages/ra-ui-materialui/src/list/DatagridRow.js
@@ -62,13 +62,12 @@ class DatagridRow extends Component {
 
         if (!rowClick) return;
 
-        if (typeof rowClick === 'function') {
-            const path = await rowClick(id, basePath, record);
-            this.handleRedirection(path, event);
-            return;
-        }
+        const path =
+            typeof rowClick === 'function'
+                ? await rowClick(id, basePath, record)
+                : rowClick;
 
-        this.handleRedirection(rowClick, event);
+        this.handleRedirection(path, event);
     };
 
     handleRedirection = (path, event) => {
--- a/packages/ra-ui-materialui/src/list/DatagridRow.js
+++ b/packages/ra-ui-materialui/src/list/DatagridRow.js
@@ -74,21 +74,22 @@ class DatagridRow extends Component {
     handleRedirection = (path, event) => {
         const { basePath, id, push } = this.props;
 
-        if (path === 'edit') {
-            push(linkToRecord(basePath, id));
-            return;
-        }
-        if (path === 'show') {
-            push(linkToRecord(basePath, id, 'show'));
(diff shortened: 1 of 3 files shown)


djhi:
You're on :fire: ! Thanks!

djhi:
However, this is technically a new feature and should target the `next` branch which will be the `3.0`

fargito:
Oops wait how do I only keep only my commits ?

packages/ra-ui-materialui/src/list/DatagridRow.js
```diff
@@ -62,33 +62,36 @@ class DatagridRow extends Component {
...
+                push(linkToRecord(basePath, id, 'show'));
+                return;
+            case 'expand':
+                this.handleToggleExpanded(event);
+                return;
+            case 'toggleSelection':
+                this.handleToggle(event);
+                return;
+            default:
+                push(path);
```
fzaninotto: If path is empty, the previous code didn't push. You're introducing  breaking change here.

packages/ra-ui-materialui/src/list/DatagridRow.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+describe('<DatagridRow />', () => {
+    const defaultProps = {
+        id: 15,
+        basePath: '/blob',
+    };
+
+    const event = { preventDefault: () => {}, stopPropagation: () => {} };
+
+    describe('on click event, depending on rowClick prop', () => {
```
fzaninotto:
```suggestion
    describe('rowClick', () => {
```

packages/ra-ui-materialui/src/list/DatagridRow.spec.js
```diff
@@ -0,0 +1,94 @@
...
+                    {...defaultProps}
+                    rowClick={customRowClick}
+                    push={push}
+                />
+            );
+
+            await wrapper.instance().handleClick(event);
+            expect(push.mock.calls).toEqual([['/bar/foo']]);
+        });
+    });
```
fzaninotto: missing a test with a falsy value - it should do nothing

packages/ra-ui-materialui/src/list/DatagridRow.js
fargito: No because there is a condition just before the `switch` :)

fargito: However I agree that it would be clearer with this condition in the `default`

fzaninotto:
Thanks!

Issue #2722, opened 2019-01-03
Nested property filter is hidden on location change but still applied


hon2a:
**What you were expecting:**

I use `Filter` for a nested property on a `List` page, e.g.:

```javascript
<Filter {...props}>
  <TextInput source="nested.prop" />
</Filter>
```

When I use this filter (show it, fill in some value, wait for the list to filter out) navigate to a different page of the application (e.g. edit an entity from the list) and then navigate back (press the "Back" button of the browser), I expect the filter to still be applied and shown in the UI.

**What happened instead:**

When I do the above, the list is loaded with the filter still applied but the filter is no longer shown in the UI. I have to toggle it again from the filters drop-down to see it and be able to interact with it.

**Steps to reproduce:**

Here's a [demo](https://codesandbox.io/s/32l8y1pm91).

1. Click on the filters drop-down toggle and select "Nested Prop".
1. Type `one` into the filter input and wait for the list to be filtered down to just a single item.
1. Click on that item to open the editing page, then click on the browser "Back" button.
1. Note that the filter is still applied but no longer visible.

To compare with a plain (not nested) property, add a "Title" filter as well. Then type "Sed quo et et fugiat modi" into the Title filter input and "five" into the "Nested Prop" filter input and do the navigation back-and-forth again. Note how the Title filter is shown again correctly.

**Related code:**

[Demo](https://codesandbox.io/s/32l8y1pm91) already linked above.

**Environment**

* React-admin version: `2.5.3`
* React version: `16.7.0`
* Browser: Chrome `71.0.3578.98 (Official Build) (64-bit)`

donatascn:
did anyone find a workaround for this issue that can used before the bug is fixed?

fzaninotto:
Fixed by #3274

PR #3274, opened 2019-05-24
fix filter condition to work with nested object paths

donatascn:
Fixing a bug from this issue: https://github.com/marmelab/react-admin/issues/2722


donatascn:
I probably shouldn't have forced pushed the changes. Sorry!

djhi:
Thanks!

PR #3275, opened 2019-05-24
optimize getFetchedAt.ts

slecoustre:
When you have a call with many identifiers then the getFetchedAt took a lot of time.

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/data.ts
+++ b/packages/ra-core/src/reducer/admin/resource/data.ts
@@ -74,7 +74,9 @@ export const addRecords = (
     );
 
     const records = { fetchedAt: newFetchedAt };
-    Object.keys(newFetchedAt).forEach(id => (records[id] = newRecordsById[id] || oldRecords[id]));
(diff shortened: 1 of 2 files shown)


Kmaschta:
How many identifiers did you have that this reduce becomes a performance issue ? :o

slecoustre:
@Kmaschta I had 1200 id and it was between 1 to 2s just for getFetchedAt. with this optimization this is almost instantaneous

fzaninotto:
I made some perf tests and the difference is indeed abysmal. I'm surprised that array reduce is so slow. Anyway, thanks a lot for the patch!

NikitaVlaznev:
May be it's not reduce's fault.
Accumulator was recreated on each iteration by {...acc}:

    const newRecordsById = newRecords.reduce(
        (acc, record) => ({
            ...acc,
            [record.id]: record,
        }),
        {}
    );

May be this could solve the problem (comma operator just returns the last value):

    const newRecordsById = newRecords.reduce(
        (acc, record) => (
            acc[record.id] = record,
            acc
        ),
        {}
    );

PR #3277, opened 2019-05-26
add redux-form version warning

joserocha3:


fzaninotto:
Hi, and thanks for your PR. The redux-form dependency in react-admin is already set to "~7.4.0" in the `packages.json`, so this warning will already be mentioned by npm. Besides, the documentation you're modifying *is* for react-admin 2.0. 

It wouldn't make sense to double all the version requirements from the `packages.json` with warnings in the documentation. So we won't merge your PR.

Issue #3278, opened 2019-05-27
Export objects to CSV


fargito:
**What you were expecting:**
When I click on the 'export to csv' button, the dowloaded file should not display [object Object] in any field

**Steps to reproduce:**
* Open the demo app
* Go to 'posts'
* Click the 'export' button
* In the dowloaded file, in the 'backlinks' column, there are [object Object] items

fzaninotto:
Thanks for your report. 

In the general case, I don't think JSON is acceptable in a CSV. On the other hand, `[object Object]` forces the developer to design a well-crafted CSV format. So it's an acceptable default IMO.

Now I agree that the simple example should have a custom exporter. I'll mark that as a bug, but just for the demo app.

fargito:
Ok perfect, I'll create a custom exporter for the demo app

PR #3279, opened 2019-05-27
[RFR] Add exporter for posts list in demo app

fargito:
Fixes https://github.com/marmelab/react-admin/issues/3278


djhi:
Thanks!

PR #3280, opened 2019-05-27
[RFR] Make Responsive use component injection rather than element

fzaninotto:
Follows #3262

### DIFF
--- a/examples/demo/src/layout/SubMenu.js
+++ b/examples/demo/src/layout/SubMenu.js
@@ -32,11 +32,12 @@ const SubMenu = ({
 }) => {
     const translate = useTranslate();
     const classes = useStyles();
+    const Icon = icon;
     return (
         <Fragment>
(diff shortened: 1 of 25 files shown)


examples/demo/src/layout/SubMenu.js
```diff
@@ -32,11 +32,12 @@ const SubMenu = ({
 }) => {
     const translate = useTranslate();
     const classes = useStyles();
+    const Icon = icon;
```
djhi: You can simply rename it in the props instead of declaring yet another variable:
```js
const SubMenu = ({
    handleToggle,
    sidebarIsOpen,
    isOpen,
    name,
    icon: Icon,
    children,
}) => ...
```

fzaninotto:
We'll revert the migration of the injection system (see #3246 for details), so I'm closing this PR.

Issue #2288, opened 2018-09-07
ReferenceInput: accept function as linkType prop


jheiska:
Everything in my DB is related to a specific building. I have multiple tabs in BuildingShow-view showing different kinds of information, and I'd like to be able to link directly to the appropriate tab from the List views of these Resources.

**Describe the solution you'd like**
I'd like to be able to define the tab a ReferenceField links to. For example, from my "AlertsList" I could include a **linkTab="alerts"** -option:

```jsx
<ReferenceField
    label="building"
    reference="buildings"
    source="BuildingId"
    linkType="show"
    linkTab="alerts"
>
    <TextField source="BuildingName" />
</ReferenceField>
```

I could write my own Field component and define the link there. I just thought this would be a neat feature now that we have the possibility to link to specific tabs.

fzaninotto:
In fact, it would be a good idea to accept a function as `linkType` prop, just like we accept a function as `redirect` prop in forms.

fzaninotto:
Fixed by #3282

PR #3282, opened 2019-05-28
[RFR] Add link function to ReferenceField

fargito:
Fixes https://github.com/marmelab/react-admin/issues/2288

* enable linkType function for ReferenceField
* update docs

PR to `next` as it is a new feature

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -520,6 +520,15 @@ You can also prevent `<ReferenceField>` from adding link to children by setting 
 </ReferenceField>
 ```
 
+You can also use a custom `linkType` function to get a custom path for the children.
+
+```jsx
+// Custom path
+<ReferenceField label="User" source="userId" reference="users" linkType={() => '/my/path'}>
+    <TextField source="name" />
+</ReferenceField>
+```
+
 **Tip**: React-admin uses `CRUD_GET_ONE_REFERENCE` action to accumulate and deduplicate the ids of the referenced records to make *one* `GET_MANY` call for the entire list, instead of n `GET_ONE` calls. So for instance, if the API returns the following list of comments:
 
 ```jsx
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -495,10 +495,10 @@ With this configuration, `<ReferenceField>` wraps the user's name in a link to t
 </Admin>
 ```
 
-To change the link from the `<Edit>` page to the `<Show>` page, set the `linkTo` prop to "show".
+To change the link from the `<Edit>` page to the `<Show>` page, set the `link` prop to "show".
 
 ```jsx
-<ReferenceField label="User" source="userId" reference="users" linkTo="show">
+<ReferenceField label="User" source="userId" reference="users" link="show">
     <TextField source="name" />
 </ReferenceField>
 ```
@@ -511,20 +511,20 @@ By default, `<ReferenceField>` is sorted by its `source`. To specify another att
 </ReferenceField>
 ```
 
-You can also prevent `<ReferenceField>` from adding link to children by setting `linkTo` to `false`.
+You can also prevent `<ReferenceField>` from adding link to children by setting `link` to `false`.
 
 ```jsx
 // No link
-<ReferenceField label="User" source="userId" reference="users" linkTo={false}>
+<ReferenceField label="User" source="userId" reference="users" link={false}>
     <TextField source="name" />
 </ReferenceField>
 ```
 
-You can also use a custom `linkTo` function to get a custom path for the children. This function must accept `record` and `reference` as arguments. On previous versions of React-Admin, this prop was named `LinkType`, however you should migrate it to `linkTo`.
+You can also use a custom `link` function to get a custom path for the children. This function must accept `record` and `reference` as arguments.
 
 ```jsx
 // Custom path
-<ReferenceField label="User" source="userId" reference="users" linkTo={(record, reference) => `/my/path/to/${reference}/${record.id}`}>
+<ReferenceField label="User" source="userId" reference="users" link={(record, reference) => `/my/path/to/${reference}/${record.id}`}>
     <TextField source="name" />
 </ReferenceField>
 ```
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -520,11 +520,11 @@ You can also prevent `<ReferenceField>` from adding link to children by setting 
 </ReferenceField>
 ```
 
-You can also use a custom `linkType` function to get a custom path for the children.
+You can also use a custom `linkType` function to get a custom path for the children. This function must accept sourceId and reference as arguments.
 
 ```jsx
 // Custom path
-<ReferenceField label="User" source="userId" reference="users" linkType={() => '/my/path'}>
+<ReferenceField label="User" source="userId" reference="users" linkType={(sourceId, reference) => `/my/path/to/${reference}/${sourceId}`}>
     <TextField source="name" />
 </ReferenceField>
 ```
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -495,10 +495,10 @@ With this configuration, `<ReferenceField>` wraps the user's name in a link to t
 </Admin>
 ```
 
-To change the link from the `<Edit>` page to the `<Show>` page, set the `linkType` prop to "show".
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -80,6 +83,8 @@ export const useReference = ({
     const rootPath = basePath.replace(resource, reference);
     const resourceLinkPath = !linkType
         ? false
+        : typeof linkType === 'function'
+        ? linkType()
```
fzaninotto: what's the use of passing a function if that function doesn't have access to the `record` and the `referenceRecord`?

docs/Fields.md
```diff
@@ -520,6 +520,15 @@ You can also prevent `<ReferenceField>` from adding link to children by setting
 </ReferenceField>
 ```
 
+You can also use a custom `linkType` function to get a custom path for the children.
+
+```jsx
+// Custom path
+<ReferenceField label="User" source="userId" reference="users" linkType={() => '/my/path'}>
```
fzaninotto: please use a more meaningful example, e.g. by redirecting to a different path depending on the reference record.

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -50,7 +52,8 @@ export interface UseReferenceProps {
  * @param {Object} option
  * @param {boolean} option.allowEmpty do we allow for no referenced record (default to false)
  * @param {string} option.basePath basepath to current resource
- * @param {string | false} option.linkType The type of the link toward the referenced record. edit, show of false for no link (default to edit)
+ * @param {string | false | linkTypeFunction} option.linkType The type of the link toward the referenced record. 'edit', 'show' or
+ * false for no link (default to edit). Alternatively a function that returns a string
```
fzaninotto: Please don't break the comment line, I don't think all jsdoc readers handle it correctly.

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -84,7 +83,7 @@ export const useReference = ({
     const resourceLinkPath = !linkType
         ? false
         : typeof linkType === 'function'
-        ? linkType()
+        ? linkType(sourceId, reference)
```
djhi: I think @fzaninotto meant the record itself

fargito: So it should be `linkType(record)` ?

djhi: `linkType(record, reference)` where record is the item referencing, and reference the referenced item

docs/Fields.md
```diff
@@ -495,10 +495,10 @@ With this configuration, `<ReferenceField>` wraps the user's name in a link to t
 </Admin>
 ```
 
-To change the link from the `<Edit>` page to the `<Show>` page, set the `linkType` prop to "show".
+To change the link from the `<Edit>` page to the `<Show>` page, set the `linkTo` prop to "show".
```
fzaninotto: Please name it `link` instead of `linkTo`

docs/Fields.md
```diff
@@ -511,11 +511,20 @@ By default, `<ReferenceField>` is sorted by its `source`. To specify another att
...
 
 ```jsx
 // No link
-<ReferenceField label="User" source="userId" reference="users" linkType={false}>
+<ReferenceField label="User" source="userId" reference="users" linkTo={false}>
+    <TextField source="name" />
+</ReferenceField>
+```
+
+You can also use a custom `linkTo` function to get a custom path for the children. This function must accept `record` and `reference` as arguments. On previous versions of React-Admin, this prop was named `LinkType`, however you should migrate it to `linkTo`.
```
fzaninotto:
```suggestion
You can also use a custom `linkTo` function to get a custom path for the children. This function must accept `record` and `reference` as arguments.
```

fzaninotto: This should be in the upgrade guide, not in the doc.

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -7,14 +7,17 @@ import { crudGetManyAccumulate } from '../../actions';
...
+
 interface Option {
     allowEmpty?: boolean;
     basePath: string;
     record?: Record;
     reference: string;
     resource: string;
     source: string;
-    linkType: string | boolean;
+    linkTo: string | boolean | LinkToFunctionType;
```
fzaninotto: it should be optional, too

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -7,14 +7,17 @@ import { crudGetManyAccumulate } from '../../actions';
...
+    linkType?: string | boolean | LinkToFunctionType;
```
fzaninotto: please add a deprecated comment

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -78,9 +83,23 @@ export const useReference = ({
         }
     }, [sourceId, reference]);
     const rootPath = basePath.replace(resource, reference);
-    const resourceLinkPath = !linkType
-        ? false
-        : linkToRecord(rootPath, sourceId, linkType as string);
+    // backcompatibility : keep linkType but with warning
+    const getResourceLinkPath = (linkTo, linkType) => {
+        if (linkType !== undefined) {
```
fzaninotto: I'd use the new param first, then fall back to the old one

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -78,9 +83,23 @@ export const useReference = ({
...
+            console.warn("DEPRECATED : argument 'linkType' should be replaced by 'linkTo' in <ReferenceField />")
```
fzaninotto:
```suggestion
            console.warn("The 'linkType' prop is deprecated and should be named to 'link' in <ReferenceField />")
```

packages/ra-core/src/controller/field/useReference.ts
```diff
@@ -78,9 +83,23 @@ export const useReference = ({
         }
     }, [sourceId, reference]);
     const rootPath = basePath.replace(resource, reference);
-    const resourceLinkPath = !linkType
-        ? false
-        : linkToRecord(rootPath, sourceId, linkType as string);
+    // backcompatibility : keep linkType but with warning
```
fzaninotto:
```suggestion
    // Backward compatibility: keep linkType but with warning
```

packages/ra-ui-materialui/src/field/ReferenceField.js
```diff
@@ -149,14 +165,14 @@ ReferenceField.propTypes = {
     sortBy: PropTypes.string,
     source: PropTypes.string.isRequired,
     translateChoice: PropTypes.func,
-    linkType: PropTypes.oneOfType([PropTypes.string, PropTypes.bool])
-        .isRequired,
+    linkType: PropTypes.oneOfType([PropTypes.string, PropTypes.bool, PropTypes.func]),
+    linkTo: PropTypes.oneOfType([PropTypes.string, PropTypes.bool, PropTypes.func]).isRequired,
```
fzaninotto: not required, because the old code must still work

fargito: Well it was the same in the old code. It simply is `edit` by default but the prop must be there right ?

docs/Fields.md
fargito: Ok :)

packages/ra-core/src/controller/field/useReference.ts
fzaninotto: disregard: if we have a default value for the new prop, your way is correct

packages/ra-ui-materialui/src/field/ReferenceField.js
fzaninotto: you're right, as there is a default value it's OK.

docs/Fields.md
fargito: Done

fargito:
I need to change the tests as well, do you want me to test the backward compatibility ?

fargito:
@fzaninotto did you have time to review this ;) ?

fzaninotto:
Thanks!

PR #3291, opened 2019-05-31
[RFR] Show notification on USER_CHECK error

### LINKED ISSUE #2967: Show custom notification after logout in USER_CHECK
When user check fails user should be allowed to show custom notification. Currently when auth token expires we want to show a user message with custom message. This behaviour works well in case of `USER_LOGIN` however it is missing in `USER_CHECK`.

https://github.com/marmelab/react-admin/blob/ed361d34f3d35634f53a0caf08bbd1467c18d8be/packages/ra-core/src/sideEffect/auth.ts#L64

User should be allowed to set custom notification which will appear after logout.

sagarbakhtar:
Show notification on USER_CHECK error in auth.

I used google translate to convert from English to French, please correct me if I am wrong.

Fixes #2967

### DIFF
--- a/packages/ra-core/src/sideEffect/auth.ts
+++ b/packages/ra-core/src/sideEffect/auth.ts
@@ -29,6 +29,12 @@ const nextPathnameSelector = state => {
 };
 
 const currentPathnameSelector = state => state.router.location;
+const errorMessageSelector = (error, defaultMessage) =>
+    typeof error === 'string'
+        ? error
+        : typeof error === 'undefined' || !error.message
+        ? defaultMessage
+        : error.message;
 
 export default (authProvider?: AuthProvider) => {
     if (!authProvider) {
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/sideEffect/auth.ts
```diff
@@ -29,6 +29,12 @@ const nextPathnameSelector = state => {
 };
 
 const currentPathnameSelector = state => state.router.location;
+const errorMessageSelector = (error, defaultMessage) =>
```
djhi: As this is a simple function, can you rename it to something like `getErrorMessage` ?

sagarbakhtar: Yes, will update.

fzaninotto:
Thanks!

PR #3292, opened 2019-06-01
[RFR] Doc update of <CheckboxGroupInput>

### LINKED ISSUE #2883: CheckboxGroupInput Options property not overwriting materials ui checkbox props.
**What you were expecting:**

I would expect the CheckboxGroupInput to apply my "labelPosition" or "onCheck" that I pass to it in the "options" prop.
**What happened instead:**
I am attempting to use react-admin's input component: CheckboxGroupInput (https://marmelab.com/react-admin/Inputs.html#checkboxgroupinput). I wanted to overwrite the "labelPosition" to be left instead of right, however when I attempted to set this property it did not change the label position. I also wanted to overwrite the onCheck callback with my own custom callback, but ran into the same issue.
I was able to set the "checked" or "disabled" property to true in the options object, but it did not apply "labelPosition" or "onCheck" that I gave it.

Attempt to modify the "labelPosition" or "onCheck" in the CheckboxGroupInput.
**Steps to reproduce:**

Attempt to pass your own "labelPosition" or "onCheck" in the CheckboxGroupInput options prop.
**Related code:**

```
const NameField = ({ record }) => <div><div>{record.name}</div>
</div>;
class PartnersCreate extends React.Component {
    render(){
        return(
            <Create {...this.props}>
                <SimpleForm redirect="list">
                    <TextInput label="Partner Name" source="name" />
                    <ReferenceInput
                        source="catalogs"
                        reference="catalogs"
                        label="Catalog"
                    >
                        <SelectInput optionText="title" label="Catalog" />
                    </ReferenceInput>
                    <BooleanInput source="checkout" label="Checkout" />
                    <FormDataConsumer>
                    {({ formData, ...rest }) => formData.checkout &&
                    <div>
                        <CheckboxGroupInput
                            source="checked" 
                            label="Products" 
                            choices={products} 
                            optionText={<NameField />} 
                            optionValue="name" 
                            options={{labelPosition:'left'}}

                        /> 
                    
                        <DiscountPercentageInput />
                        <div>
                            <TextInput source="discountMessage" label="Discount Message" />
                        </div>
                        <div>
                            <NumberInput source="commission" label="Commission Percentage" />
                        </div>
                    </div>
                    }
                    </FormDataConsumer>

                </SimpleForm>
            </Create>
        )
    }
}
export default PartnersCreate;
```

**Other information:**

I recently added the checkout check in <FormDataConsumer>, and the behavior was the same for CheckboxGroupInput outside of this recently added change.
**Environment**

* React-admin version: 2.0.0
* React version: 16.3.1
* Browser: Latest Chrome

sagarbakhtar:
Example in the doc was using prop from the latest version of `<Checkbox>` component from Material Ui and also a link of `<Checkbox>` documentation was pointing to the latest Material Ui version, and Material Ui version used by `ra-ui-materialui` does not support many props from the latest version, so the confusion for users.

Changed both prop example and link to version (1.5.1) used by `ra-ui-materialui`.

Resolves #2883

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -456,13 +456,16 @@ Lastly, use the `options` attribute if you want to override any of Material UI's
 
 {% raw %}
 ```jsx
+import { FavoriteBorder, Favorite } from '@material-ui/icons';
+
 <CheckboxGroupInput source="category" options={{
-    labelPosition: 'right'
+    icon: <FavoriteBorder />,
+    checkedIcon: <Favorite />
 }} />
 ```
 {% endraw %}
 
-Refer to [Material UI Checkbox documentation](http://www.material-ui.com/#/components/checkbox) for more details.
+Refer to [Material UI Checkbox documentation](https://v1-5-0.material-ui.com/api/checkbox/) for more details.
 
 ## `<DateInput>`
 


djhi:
Thanks !

PR #3304, opened 2019-06-05
[RFR] Prettier

djhi:
- [x] Upgrade Prettier
- [x] Ensure Prettier Config is complete to avoid user overrides and applied to all files
- [x] Apply prettier on the project


fzaninotto:
We should not upgrade prettier before we are about to release v3. It would make the backport of the `master` changes to `next` extremely painful.

djhi:
> We should not upgrade prettier before we are about to release v3. It would make the backport of the master changes to next extremely painful.

As discussed I will make the same PR on `next` to avoid any issues

fzaninotto:
Can you revert the line width change to minimize the patch?

packages/ra-ui-materialui/src/list/List.js
```diff
@@ -98,23 +98,9 @@ const sanitizeRestProps = ({
...
-    // overridable by user
-    children,
-    className,
-    classes,
-    exporter,
-    title,
-    ...rest
-}) => {
+export const ListView = withStyles(styles)(({ // component props
+    actions, aside, filter, filters, bulkActions, bulkActionButtons, pagination, children, className, classes, exporter, title, ...rest }) => {
```
fzaninotto: weird

djhi: Looks like the comments are doing this

fzaninotto:
Thanks! Next step, the `next` branch.

PR #3308, opened 2019-06-05
[RFR] Ensure List Data Update when Default Filters Change

### LINKED ISSUE #3132: Permanent filter mapped to state value doesn't filter correctly

djhi:
Fix #3132
Supersedes #3251

### DIFF
--- a/packages/ra-core/src/controller/ListController.spec.tsx
+++ b/packages/ra-core/src/controller/ListController.spec.tsx
@@ -119,17 +119,22 @@ describe('ListController', () => {
         });
 
         it('should update data if permanent filters change', () => {
-            expect.assertions(2);
+            expect.assertions(4);
+
+            const children = jest.fn();
             const props = {
                 ...defaultProps,
                 debounce: 200,
                 crudGetList: jest.fn(),
                 filter: { foo: 1 },
-                children: fakeComponent,
+                children,
             };
 
             const wrapper = shallow(<ListController {...props} />);
+
             expect(props.crudGetList.mock.calls[0][3]).toEqual({ foo: 1 });
+            expect(children.mock.calls[0][0].displayedFilters).toEqual({});
+            expect(children.mock.calls[0][0].filterValues).toEqual({});
 
             wrapper.setProps({ filter: { foo: 2 } });
             expect(props.crudGetList.mock.calls[1][3]).toEqual({ foo: 2 });
--- a/packages/ra-core/src/controller/ListController.spec.tsx
+++ b/packages/ra-core/src/controller/ListController.spec.tsx
@@ -119,8 +119,6 @@ describe('ListController', () => {
         });
 
         it('should update data if permanent filters change', () => {
-            expect.assertions(4);
-
             const children = jest.fn();
             const props = {
                 ...defaultProps,
@@ -132,12 +130,20 @@ describe('ListController', () => {
 
             const wrapper = shallow(<ListController {...props} />);
 
+            // Check that the permanent filter was used in the query
             expect(props.crudGetList.mock.calls[0][3]).toEqual({ foo: 1 });
+            // Check that the permanent filter is not included in the displayedFilters (passed to Filter form and button)
             expect(children.mock.calls[0][0].displayedFilters).toEqual({});
+            // Check that the permanent filter is not included in the filterValues (passed to Filter form and button)
             expect(children.mock.calls[0][0].filterValues).toEqual({});
 
             wrapper.setProps({ filter: { foo: 2 } });
+            // Check that the permanent filter was used in the query
             expect(props.crudGetList.mock.calls[1][3]).toEqual({ foo: 2 });
+            // Check that the permanent filter is not included in the displayedFilters (passed to Filter form and button)
+            expect(children.mock.calls[0][0].displayedFilters).toEqual({});
+            // Check that the permanent filter is not included in the filterValues (passed to Filter form and button)
+            expect(children.mock.calls[0][0].filterValues).toEqual({});
         });
 
         afterEach(() => {
--- a/packages/ra-core/src/controller/ListController.spec.tsx
+++ b/packages/ra-core/src/controller/ListController.spec.tsx
@@ -41,6 +41,11 @@ describe('ListController', () => {
         toggleItem: jest.fn(),
         total: 100,
         translate: jest.fn(),
+        perPage: 25,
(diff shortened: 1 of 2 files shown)


djhi:
@tarnowsc Thanks for trying to fix this issue :+1:  While reviewing your pull request, I noticed opportunities to simplify our code and fix the issue which is why I created another pr.

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -207,24 +206,14 @@ export class UnconnectedListController extends Component<
...
-            nextProps.query.page !== this.props.query.page ||
-            nextProps.query.perPage !== this.props.query.perPage ||
-            !isEqual(nextProps.query.filter, this.props.query.filter) ||
+            !isEqual(nextProps.query, this.props.query) ||
             !isEqual(nextProps.filter, this.props.filter) ||
             !isEqual(nextProps.sort, this.props.sort) ||
             !isEqual(nextProps.perPage, this.props.perPage)
         ) {
-            this.updateData(
-                Object.keys(nextProps.query).length > 0
```
fzaninotto: nextProps.query isn't the same as the result of `getQuery`, which does the fallback to params. So I think you still need the params fallback here.

packages/ra-core/src/controller/ListController.spec.tsx
```diff
@@ -113,6 +118,23 @@ describe('ListController', () => {
...
+                children: fakeComponent,
+            };
+
+            const wrapper = shallow(<ListController {...props} />);
+            expect(props.crudGetList.mock.calls[0][3]).toEqual({ foo: 1 });
+
+            wrapper.setProps({ filter: { foo: 2 } });
+            expect(props.crudGetList.mock.calls[1][3]).toEqual({ foo: 2 });
+        });
+
```
fzaninotto: Could you also add a test that permanent filters don't appear in the filter form and button?

packages/ra-core/src/controller/ListController.tsx
djhi: You mean we should check if `nextProps.params` has changed ? Because we are now passing the props to `updateData`

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -302,19 +265,19 @@ export class UnconnectedListController extends Component<
...
 
-    updateData(query?: any) {
-        const params = query || this.getQuery();
-        const { sort, order, page = 1, perPage, filter } = params;
+    updateData(props = this.props) {
+        const query = this.getQuery(props);
+
+        const { sort, order, page = 1, perPage, filter } = query;
         const pagination = {
-            page: parseInt(page, 10),
```
fzaninotto: This was here for a reason. Page and perPage can be strings. You said that TypeScript complained that the params are already numbers, but I don't see where that comes from.

packages/ra-core/src/controller/ListController.tsx
```diff
@@ -244,52 +233,26 @@ export class UnconnectedListController extends Component<
...
         const query: Partial<ListParams> =
-            Object.keys(this.props.query).length > 0
-                ? this.props.query
-                : this.hasCustomParams(this.props.params)
-                ? { ...this.props.params }
-                : { filter: this.props.filterDefaultValues || {} };
+            Object.keys(props.query).length > 0
+                ? props.query
+                : hasCustomParams(props.params)
+                ? props.params
```
fzaninotto: It was cloned for a reason (I think because parmas are mutated somewhere), you shouldn't change that.

packages/ra-core/src/controller/ListController.spec.tsx
fzaninotto: bump

agnsier:
Hi, any progress on that pull request? Could you tell me when do you plan to release a new version with that fix? 

Thanks!

packages/ra-core/src/controller/ListController.spec.tsx
djhi: Done

packages/ra-core/src/controller/ListController.tsx
djhi: From the ListParams interface

packages/ra-core/src/controller/ListController.tsx
djhi: Fixed

packages/ra-core/src/controller/ListController.tsx
djhi: Done in the getQuery function

manuelamaria:
Hi guys, thanks for looking into this reported issue. This functionality became very important for our project. Is there any chance this PR might get merged soon? I've been trying to find a workaround with no success.

Thank you!

agnsier:
> Hi guys, thanks for looking into this reported issue. This functionality became very important for our project. Is there any chance this PR might get merged soon? I've been trying to find a workaround with no success.
> 
> Thank you!

We also had a problem with that permanent filter in our project. In our case, we added a key property for a List component. That helped :) 
Example:
`
  <List
    {...props}
    filters={<CustomFilters />}
    exporter={false}
    key={props.filter.customFilter}
  >
<CustomDatagrid/>
</List>
`

This {props.filter.customFilter} came from react-router Route component.

Good luck :)

packages/ra-core/src/controller/ListController.spec.tsx
djhi: Added comments on the tests to explain

manuelamaria:
Great! Thanks a lot!

Thanks to you @agnsier also, for your suggestion.

djhi:
Please not it has not been released yet though and won't be before a few weeks

PR #3309, opened 2019-06-05
[RFR] New documentation with Docusaurus

alexisjanvier:
[Docusaurus](https://docusaurus.io/docs/en/next/installation)

- [x] Add docusaurus
- [x] Update documentation to next version
- [x] Update build and deploy scripts

Capture d’écran de 2019-06-05 16-00-14
Capture d’écran de 2019-06-05 16-00-43

To test locally from this branch, you have to do an `make install` and then run the development server with `make doc`. The documentation can then be consulted on `http://localhost:3000/react-admin`


fzaninotto:
After reviewing this PR and talking with @alexisjanvier, we're not going to merge it (the PR, not @alexisjanvier).

This PR shows 3 shortcomings of Docusaurus that are blockers from my point of view:

1. Images can't be versioned. All versions share the same images. But out screenshots vary a lot between versions, so we need image versioning.
2. Images aren't located in the docs folder, so when users browse the doc with github, they don't see the inline images. A good share of our users don't use the documentation website but GitHub directly. 
3. Versioned docs are copied to a new directory in master. This significantly increases the size of the repository at each release. This is going to get worse in docusaurus 2 as they'll copy *all* the doc files, and not only the ones that changed.

I realize that Docusaurus would bring a lot, but I believe we can implement some of their features with a few lines of shell (or Node, for that matter):

1. Better search by implementing algolia search (https://github.com/algolia/docsearch)
2. Better look and feel by using their generated CSS and HTML
3. Doc versioning by copying the docs folder (including images) to a subdoc in the gh-pages branch
4. Index versioning of the index (in the sidebar) by using Jekyll includes and variables 

We still want to improve the docs and publish old versions. Docusaurus just isn't the right tool for our constraints.

Thanks @alexisjanvier anyway for the hard work!

PR #3311, opened 2019-06-05
[RFR] Feature: add 'filterSelectedFromSuggestions' prop to 'AutocompleteArrayInput'

theVeid:
This prop will allow to not to show already selected choices items in the suggestions list.

As soon as this feature is not a breaking change, I would like it to be merged into 'master'.

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.js
@@ -183,7 +183,8 @@ export class AutocompleteArrayInput extends React.Component {
     };
 
     handleSuggestionsFetchRequested = () => {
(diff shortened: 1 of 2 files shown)


fzaninotto:
Hi @theVeid, and thanks for your PR.

This is a new feature, I'm not sure we want it, it lacks documentation, and we'll be rewriting the autocomplete input in the `next` branch based on a new library. That's too many reasons to not accept your pull request.

I suggest you wait for the rewrite of this component in the `next` branch, and we'll discuss your suggestion then.

theVeid:
Hi @fzaninotto, thanks for your reply.

For the next time, if I would like to repeat pull request into `next` branch, should I also make changes into documentation, or it would be better to provide some gifs with PR, so you can see what exactly was added?

PR #3312, opened 2019-06-05
[RFR] Replace component injection with element injection

djhi:
Revert #3262


djhi:
It's green but I still have a strange warning left to fix

packages/ra-ui-materialui/src/list/List.js
```diff
@@ -1,13 +1,12 @@
-/* eslint no-console: ["error", { allow: ["warn", "error"] }] */
```
fzaninotto: can this be the cause of your warning?

djhi: Nope, it's a warning about react-router

djhi:
> There were 100 files changed in the previous PR, only 91 in this PR. Some changes may be missing.

Yes some changes are still relevant (typos, etc.)

fzaninotto:
Thanks a lot!

PR #3321, opened 2019-06-08
[RFR] Replace tslint with eslint

djhi:
- [x] Replace tslint with eslint
- [x] Adjust configuration
- [x] Replace all tslint comments with eslint ones


packages/ra-core/src/controller/EditController.tsx
```diff
@@ -133,9 +130,13 @@ const EditController = (props: Props) => {
                 dispatch(updateAction);
             }
         },
-        [id, resource, basePath, record]
+        [resource, id, record, basePath, undoable, dispatch]
     );
 
+    if (!children) {
```
djhi: Not really sure about this one but the lint rule for hook complained about us applying hooks conditionnaly

packages/ra-core/src/controller/EditController.tsx
```diff
@@ -104,7 +101,7 @@ const EditController = (props: Props) => {
 
     useEffect(() => {
         dispatch(resetForm(REDUX_FORM_NAME));
-    }, [resource, id, version]);
+    }, [resource, id, version, dispatch]);
```
djhi: Not sure about this either. It was adding by "eslint --fix" call

examples/simple/src/comments/PostPreview.js
```diff
@@ -11,7 +11,6 @@ const PostPreview = props => {
         [props.resource, props.id]
     );
     const version = useSelector(state => state.admin.ui.viewVersion);
-    const isLoading = useSelector(state => state.admin.loading > 0);
```
fzaninotto: Why did you remove this one? It forces a redraw when isLoading changes.

packages/ra-core/src/controller/EditController.tsx
fzaninotto: I'd say ignore the lint rule on that one

packages/ra-core/src/controller/ShowController.tsx
```diff
@@ -73,9 +73,6 @@ interface Props {
 const ShowController = (props: Props) => {
     useCheckMinimumRequiredProps('Show', ['basePath', 'resource'], props);
     const { basePath, children, id, resource } = props;
-    if (!children) {
```
fzaninotto: same

packages/ra-core/src/fetch/useQueryWithStore.ts
```diff
@@ -136,7 +138,7 @@ const useQueryWithStore = (
                     loaded: false,
                 });
             });
-    }, [JSON.stringify({ query, options })]); // deep equality, see https://github.com/facebook/react/issues/14476#issuecomment-471199055
```
fzaninotto: no, we need that JSON.stringify, it does a deep equal with low overhead

packages/ra-ui-materialui/src/list/DatagridLoading.js
```diff
@@ -72,10 +72,10 @@ export default ({
                             className={classes.expandIconCell}
                         >
                             <IconButton
+                                // TODO: add an aria-label
                                 className={classes.expandIcon}
                                 component="div"
                                 aria-hidden="true"
-                                role="expand"
```
fzaninotto: why?

packages/ra-ui-materialui/src/list/DatagridRow.js
```diff
@@ -177,9 +177,10 @@ export class DatagridRow extends Component {
                         ) : null
                     )}
                 </TableRow>
-                {expand && expanded && (
+                {// TODO: add proper aria attributes (aria-expanded)
+                expand && expanded && (
                     <TableRow key={`${id}-expand`}>
-                        <TableCell colSpan={colSpan} role="expand-content">
```
fzaninotto: why ?

examples/simple/src/comments/PostPreview.js
djhi: Right, it complained about the unused variable

packages/ra-ui-materialui/src/list/DatagridLoading.js
djhi: Because after looking for it, it appears to not be an existing aria role.

lint rule: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-role.md
list of valid roles: https://www.w3.org/TR/wai-aria/#role_definitions

packages/ra-ui-materialui/src/list/DatagridRow.js
djhi: Same reason. However there are other ways but it deserves a dedicated PR. I'll open an issue to track it

packages/ra-core/src/controller/EditController.tsx
djhi: Why are we doing this though? Shouldn't we throw an error about it ?

djhi: Ok but we are breaking the hooks rules here

packages/ra-ui-materialui/src/list/DatagridLoading.js
djhi: Besides, this button does nothing. It's juste here for placeholder UI while loading

packages/ra-ui-materialui/src/list/DatagridRow.js
djhi: Fixed it at it was needed for the tests

packages/ra-core/src/controller/EditController.tsx
djhi: Btw, couldn't we fallback to the guesser when we detect there are no children ?

fzaninotto: I agree, this is not a normal use case, we should throw. And mark the `children` prop as required.

Also, we can't fall back to an `EditGuesser` because there is no such thing in `ra-core` (it's in `ra-ui-material-ui`)

packages/ra-ui-materialui/src/list/DatagridLoading.js
fzaninotto: gotcha

packages/ra-data-graphcool/src/buildVariables.js
```diff
@@ -166,10 +165,11 @@ export default introspectionResults => (
                 queryType
             );
         }
-
         case DELETE:
             return {
                 id: params.id,
             };
+        default:
+            console.log('Unknown fetch verb');
```
fzaninotto: No, unknown fetch verbs are allowed and can be implemented in custom dataProviders

packages/ra-data-graphql-simple/src/buildVariables.js
```diff
@@ -275,10 +270,11 @@ export default introspectionResults => (
                 queryType
             );
         }
-
         case DELETE:
             return {
                 id: preparedParams.id,
             };
+        default:
+            console.log('Unknown fetch verb');
```
fzaninotto: Same, we don't mind unknown fetch verbs

packages/ra-data-graphcool/src/buildVariables.js
djhi: Yes, but it is a good practice to handle the default case right ?

djhi: Those dataProviders did not handle them at all and returned undefined. How should we handle that ?

PR #3322, opened 2019-06-09
[RFR] Replace tslint with eslint

djhi:


packages/ra-language-french/index.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
             show: 'Afficher',
             sort: 'Trier',
             undo: 'Annuler',
+            expand: 'Étendre',
+            close: 'Close',
```
fzaninotto: Fermer? Réduire?

PR #3323, opened 2019-06-10
[RFR] Use theme to store sidebar width

### LINKED ISSUE #2375: Use theme to store sidebar width
React-admin expects the sidebar width to be set via props, while the (material-ui) theme themes to be the perfect place for that.

djhi:
Fixes #2375


packages/ra-ui-materialui/src/layout/Sidebar.js
```diff
@@ -125,18 +117,11 @@ class Sidebar extends PureComponent {
...
     classes: PropTypes.object,
-    closedSize: PropTypes.number,
     open: PropTypes.bool.isRequired,
     setSidebarVisibility: PropTypes.func.isRequired,
-    size: PropTypes.number,
     width: PropTypes.string,
 };
 
-Sidebar.defaultProps = {
-    size: DRAWER_WIDTH,
```
fzaninotto: then the constants line 12 and 13 can be removed, too - but I think we should use them to set default values for when the theme doesn't contain the sizes.

docs/Theming.md
```diff
@@ -284,14 +284,16 @@ const App = () => (
 );
 ```
 
+Note that we use `createMuiTheme` from `react-admin`. This ensures your theme contains the custom values needed for some `react-admin` components such as the `<SideBar>`.
```
fzaninotto: I suggest that we don't make that a requirement (or it would be another BC break) and instead provide default values for these theme dimensions in the sidebar component.

docs/Theming.md
```diff
@@ -441,17 +445,23 @@ const MyAppBar = props => <AppBar {...props} userMenu={MyUserMenu} />;
 
 ### Sidebar Customization
 
-You can specify the `Sidebar` size by setting the `size` property:
+You can specify the `Sidebar` width by setting the `width` and `closedWidth` property on your custom material-ui them:
 
 ```jsx
-import { Sidebar } from 'react-admin';
+import { createMuiTheme } from 'react-admin';
```
fzaninotto: no need to use the react-admin one here as you're providing the required dimensions.

packages/ra-ui-materialui/src/createMuiTheme.ts
```diff
@@ -0,0 +1,10 @@
+import { createMuiTheme as baseCreateMuiTheme } from '@material-ui/core/styles';
```
fzaninotto: Why do we still need this file? I think it's not used anywhere anymore.

fzaninotto:
Thanks!

PR #3324, opened 2019-06-10
[RFR] Replace papaparse with a lighter library

### LINKED ISSUE #2374: Replace papaparse by something lighter
React-admin bundles the `papaparse` library for converting JSON to CSV. But 90% of the `papaparse` code is used to convert CSV to JSON. Unless we implement import, we should replace papaparse by a lighter library.

djhi:
Fixes #2374 
We saved 10kb

Before:
bundle analyze with papaparse

After:
bundle analyze with jsonexport

I took the opportunity to add bundle analyzers to the simple and demo examples. That will probably come handy later.


fzaninotto:
Excellent, thanks!

PR #3327, opened 2019-06-12
Fix AutoCompleteArrayInput not showing error message

hithacker:
AutoCompleteArrayInput was not showing up error message when there is a validation error. It shows up the input in red color but the error message is not shown. 
Fixing the code to follow the same patten as SelectInput and AutocompleteInput.


fzaninotto:
Thanks!

PR #3328, opened 2019-06-12
[RFR] Use makeStyles hooks on examples demo

zyhou:
Use `makeStyles` on demo

Review with "Hide whitespace changes" options in Github

### DIFF
--- a/examples/demo/src/products/GridList.js
+++ b/examples/demo/src/products/GridList.js
@@ -3,13 +3,13 @@ import compose from 'recompose/compose';
 import MuiGridList from '@material-ui/core/GridList';
 import GridListTile from '@material-ui/core/GridListTile';
 import GridListTileBar from '@material-ui/core/GridListTileBar';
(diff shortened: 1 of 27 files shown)


examples/demo/src/products/GridList.js
```diff
@@ -45,68 +45,71 @@ const getColsForWidth = width => {
...
+};
 
 const GridList = ({ loadedOnce, ...props }) =>
     loadedOnce ? <LoadedGridList {...props} /> : <LoadingGridList {...props} />;
 
-const enhance = compose(
-    withWidth(),
-    withStyles(styles)
-);
+const enhance = compose(withWidth());
```
fzaninotto: No need for `compose` anymore here as there is only one HOC

zyhou:
Now it's RFR, I didn't test all the pages on the demo

fzaninotto:
Thanks!

PR #3329, opened 2019-06-12
[RFR] Migrate Responsive to custom media queries

fzaninotto:
- [x] Replace usage of `<Responsive>` in react-admin by media query hooks
- [x] Replace usage of `<Responsive>` in examples by media query hooks
- [x] Document the new hooks
- [x] Fix tests

~~This PR is blocked by a regression in material-ui, which makes the tests fail with the hooks (https://github.com/mui-org/material-ui/issues/16184).~~ (fixed in https://github.com/mui-org/material-ui/releases/tag/v4.1.1)


djhi:
See https://material-ui.com/components/use-media-query/#usemediaquery-query-options-matches

Issue #3127, opened 2019-04-15
Support helperText in a BooleanInput


kctrlv:
**What you were expecting:**
I would like to be able to set a helperText in this input. I can do this in other inputs, such as
`<TextInput helperText='abc123' />` and 
`<RadioButtonGroupInput meta={ { helperText: 'abc123' } } />`
but currently there is no way to set it for a BooleanInput. 

**Other information:**
The reason is that FormHelperText is not included in the `BooleanInput.js` file. If it were, it would probably be as simple as copying lines 161-164 from `RadioButtonGroupInput.js`

fzaninotto:
Hi, nice idea for an enhancement. Feel free to submit a PR about it.

fzaninotto:
Fixed by #3339

PR #3336, opened 2019-06-13
Booleaninput helpertext

kctrlv:
This is to support passing a helperText into a BooleanInput, following the same convention as other inputs that use the meta tag, such as RadioButtonGroupInput. 

An issue was created here prior to making this PR:  https://github.com/marmelab/react-admin/issues/3127


fzaninotto:
Thanks! This is technically a new feature, so it should be a PR against the `next` branch rather than `next`. Moreover, that feature has precisely been added in #3339, so I'm closing your PR.

PR #3339, opened 2019-06-15
[RFR] Refactor validators to not return translated errors

djhi:
This is the first step to migrate from redux-form to `final-form`. Indeed, `react-final-form` validators won't receive props anymore, so our current third argument containing the `translate` function will always be empty.

This PR refactor the validators to return either a string which will be translated or an object with `message` and `args` properties.

I also introduced an unstyled `ValidationError` component which knows how to translate those errors. As it returns the resulting string inside a React Fragment, it can be defined in `ra-core` and used in any UI library.

Finally, I updated the official inputs to use this `ValidationError`. In the process, I have migrated their tests from `enzyme` to `react-testing-library` and ensured all inputs support the `helperText` prop.

- [x] Refactor validators
- [x] Introduced `ValidationError` component
- [x] Introduced `InputHelperText` component
- [x] Migrated inputs & their tests
- [x] Updated upgrade guide


djhi:
I noticed that we don't handle errors and helperText the same way for all inputs. Some will display the error instead of the helperText, some will display both.

I think it would be better to handle this in a consistant way by introducing an `InputHelperText` component which will display both the helper text and the error.

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { SFC } from 'react';
```
fzaninotto: SFC is deprecated, use `FunctionalComponent` instead

packages/ra-core/src/form/ValidationError.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+        translate.mockClear();
+    });
+
+    it('It renders the error message translated if it is a string', () => {
+        const { getByText } = render(
+            <ValidationError error="ra.validation.required" />
+        );
+
+        expect(getByText('ra.validation.required')).toBeTruthy();
+        expect(translate).toHaveBeenCalledWith('ra.validation.required', {
```
fzaninotto: Bad testing strategy IMO. You'd better wrap your component with a TranslationProvider that translates the `ra.validation.required` message, and look for the translation in the render. Same for the other test.

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {
+    ValidationErrorMessage,
+    ValidationErrorMessageWithArgs,
+} from './validate';
+import { useTranslate } from '../i18n';
+
+interface Props {
+    error: ValidationErrorMessage;
```
fzaninotto: it's optional, missing `?`

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const ValidationError: SFC<Props> = ({ error }) => {
+    const translate = useTranslate();
+
+    if ((error as ValidationErrorMessageWithArgs).message) {
+        const { message, args } = error as ValidationErrorMessageWithArgs;
+        const { _, ...allArgsButDefault } = args;
+
+        const translatedArgs = Object.keys(allArgsButDefault).reduce(
```
fzaninotto: very functional but also very hard to read IMO. Could you rework that reduce to make it easier to read (hint: imperative style is sometimes better)?

packages/ra-ui-materialui/package.json
```diff
@@ -61,6 +61,7 @@
         "react-redux": "^7.1.0-rc.1",
         "react-router": "^5.0.1",
         "react-router-dom": "^5.0.1",
+        "react-testing-library": "^7.0.0",
```
fzaninotto: it should be a `devDependency`

packages/ra-ui-materialui/src/input/ArrayInput.spec.js
```diff
@@ -1,45 +1,17 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
```
fzaninotto: is it still required?

packages/ra-ui-materialui/src/input/ArrayInput.spec.js
```diff
@@ -1,45 +1,17 @@
...
 import SimpleFormIterator from '../form/SimpleFormIterator';
 
 describe('<ArrayInput />', () => {
-    it('should render a FieldArray', () => {
-        const wrapper = shallow(<ArrayInputView source="arr" record={{}} />);
-        expect(wrapper.find('pure(FieldTitle)').length).toBe(1);
-        expect(wrapper.find('FieldArray').length).toBe(1);
-    });
-
-    it('should pass an undefined fields to child when initialized with an undefined value', () => {
```
fzaninotto: why did you remove this test?

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.js
```diff
@@ -1,32 +1,32 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import assert from 'assert';
-import { shallow, render, mount } from 'enzyme';
+import { shallow, mount } from 'enzyme';
```
fzaninotto: is it still needed?

packages/ra-core/src/form/ValidationError.tsx
djhi: It's not optional

packages/ra-core/src/form/ValidationError.spec.tsx
djhi: I don't get it. Can you explain why?

packages/ra-ui-materialui/src/input/ArrayInput.spec.js
djhi: Because it tests redux-form, not react-admin

packages/ra-core/src/form/ValidationError.spec.tsx
fzaninotto: Because your mock is testing internal method calls, instead of asserting that the result is what's expected from the end user's standpoint. In other terms, you're testing the implementation rather than the feature.

packages/ra-ui-materialui/src/input/BooleanInput.js
```diff
@@ -25,38 +27,48 @@ export class BooleanInput extends Component {
...
-                            resource={resource}
-                            isRequired={isRequired}
-                        />
-                    }
-                />
-                {meta.error && (
-                    <FormHelperText error>{meta.error}</FormHelperText>
-                )}
-            </FormGroup>
+            <FormControl className={className} {...sanitizeRestProps(rest)}>
```
fzaninotto: applying the className to Formcontrol instead of FormGroup is technically a breaking change. Can you avoid that?

packages/ra-ui-materialui/src/input/DateInput.js
```diff
@@ -41,59 +42,63 @@ const sanitizeValue = value => {
...
+    input,
+    isRequired,
+    label,
+    options,
+    source,
+    resource,
+    helperText,
+    ...rest
+}) => {
+    const handleChange = event => {
```
fzaninotto: please use useCallback to memoize the function and avoid rerenderings of TextField

packages/ra-ui-materialui/src/input/DateInput.spec.js
```diff
@@ -32,31 +32,23 @@ describe('<DateInput />', () => {
 
     describe('error message', () => {
         it('should not be displayed if field is pristine', () => {
-            const { container } = render(
```
fzaninotto: you've removed a useful test

packages/ra-ui-materialui/src/input/InputHelperText.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { SFC } from 'react';
+import { useTranslate, ValidationError, ValidationErrorMessage } from 'ra-core';
+
+interface Props {
+    helperText?: string;
+    error?: ValidationErrorMessage;
+    touched: boolean;
+}
+
+const InputHelperText: SFC<Props> = ({ helperText, touched, error }) => {
```
fzaninotto: SFC is deprecated, please use FunctionalComponent instead

packages/ra-ui-materialui/src/input/NumberInput.js
```diff
@@ -17,82 +18,89 @@ import sanitizeRestProps from './sanitizeRestProps';
...
+    source,
+    step,
+    resource,
+    helperText,
+    onBlur,
+    onFocus,
+    onChange,
+    ...rest
+}) => {
+    const handleBlur = event => {
```
fzaninotto: please memoize. Same for other callback

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
```diff
@@ -76,20 +77,30 @@ const styles = createStyles({
...
 
-    renderRadioButton = choice => {
-        const {
-            optionText,
-            optionValue,
-            translate,
-            translateChoice,
-            source,
-        } = this.props;
+    const handleChange = (event, value) => {
```
fzaninotto: please memoize with useCallback

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
```diff
@@ -147,57 +158,46 @@ describe('<RadioButtonGroupInput />', () => {
...
-                meta={{ helperText: 'Can I help you?' }}
+                helperText="Can I help you?"
             />
         );
         expect(queryByText('Can I help you?')).not.toBeNull();
     });
 
     describe('error message', () => {
         it('should not be displayed if field is pristine', () => {
-            const { container } = render(
```
fzaninotto: please keep this test

packages/ra-ui-materialui/src/input/SelectArrayInput.spec.js
```diff
@@ -153,56 +153,45 @@ describe('<SelectArrayInput />', () => {
...
-            />
+            <SelectArrayInput {...defaultProps} helperText="Can I help you?" />
         );
-        expect(queryByText('Can I help you?')).not.toBeNull();
+        expect(queryByText('Can I help you?')).toBeDefined();
     });
 
     describe('error message', () => {
         it('should not be displayed if field is pristine', () => {
-            const { container } = render(
```
fzaninotto: please keep this test

packages/ra-ui-materialui/src/input/SelectInput.js
```diff
@@ -129,44 +131,45 @@ const styles = theme =>
...
-    componentWillReceiveProps(nextProps) {
-        if (nextProps.input.value !== this.props.input.value) {
-            this.setState({ value: nextProps.input.value });
-        }
-    }
+    const [value, setValue] = useState(input.value);
+    const translate = useTranslate();
 
-    handleChange = eventOrValue => {
+    const handleChange = eventOrValue => {
```
fzaninotto: needs useCallback

packages/ra-ui-materialui/src/input/TextInput.js
```diff
@@ -19,71 +20,79 @@ import sanitizeRestProps from './sanitizeRestProps';
...
+    source,
+    type,
+    helperText,
+    FormHelperTextProps,
+    onBlur,
+    onFocus,
+    onChange,
+    ...rest
+}) => {
+    const handleBlur = eventOrValue => {
```
fzaninotto: useCallback

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
djhi: I actually kept it but added an error because you can't test it isn't here if it is not defined right ? I removed the `should not be displayed if field has been touched but is valid` test though as I don't see its point

packages/ra-ui-materialui/src/input/SelectInput.js
```diff
@@ -129,122 +131,119 @@ const styles = theme =>
...
-            : eventOrValue;
-        this.props.input.onChange(value);
+    const [value, setValue] = useState(input.value);
+    const translate = useTranslate();
 
-        // HACK: For some reason, redux-form does not consider this input touched without calling onBlur manually
-        this.props.input.onBlur(value);
-        this.setState({ value });
-    };
+    useEffect(() => {
```
djhi: Obviously a code smell. This is the implementation of the previous `componentWillReceiveProps`. Hopefully this will not be needed after migrating to final-forms

packages/ra-ui-materialui/src/input/DateInput.spec.js
djhi: I actually kept it but added an error because you can't test it isn't here if it is not defined right ? I removed the should not be displayed if field has been touched but is valid test though as I don't see its point

packages/ra-core/src/form/validate.ts
```diff
@@ -8,42 +7,47 @@ const EMAIL_REGEX = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"
...
-              ...props,
           })
-        : props.translate(message, {
-              _: message,
-              ...messageArgs,
-          });
+        : messageArgs
+        ? {
+              message,
+              args: messageArgs,
```
fzaninotto: you should pass the `message` as default translation (using the `_` arg) or users will be flooded by missing translation messages.

djhi: Hu ? This is the job of the component which displays the error and actually, it does it

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const translateArgValue = (translate: Translate, argValue: any) => {
+    if (Array.isArray(argValue)) {
+        return translateArrayArgValue(translate, argValue);
+    }
+
+    return translate(argValue.toString(), { _: argValue.toString() });
+};
+
+const translateArgs = (translate: Translate, args: { [key: string]: any }) =>
```
fzaninotto: I don't understand what you're doing here. We should not translate arguments. we should translate the message.

fzaninotto:
Thanks!

PR #3349, opened 2019-06-20
[RFR] Upgrade react-redux to version 7.1

fzaninotto:
Requires changes in useSelector due to breaking change in react-redux 7.1.0-alpha.5 (see https://github.com/reduxjs/react-redux/releases/tag/v7.1.0-alpha.5).

Without this upgrade, TypeScript will fail to build after a yarn fresh install, because the new DefinitelyTyped types for react-redux enforce the `useSelector` signature of 7.1.0 final.

### DIFF
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -19,7 +19,7 @@
         "react-admin": "^2.0.0",
         "react-app-polyfill": "^0.1.3",
         "react-dom": "~16.8.0",
-        "react-redux": "^7.1.0-rc.1",
+        "react-redux": "^7.1.0",
         "react-router": "^5.0.1",
         "react-router-dom": "^5.0.1",
         "react-scripts": "^3.0.0",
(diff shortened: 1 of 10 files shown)


examples/demo/package.json
```diff
@@ -19,7 +19,7 @@
         "react-admin": "^2.0.0",
         "react-app-polyfill": "^0.1.3",
         "react-dom": "~16.8.0",
-        "react-redux": "^7.1.0-rc.1",
+        "react-redux": "^7.1.0",
```
djhi: Shouldn't we remove this line as `ra-core` already includes it as a dependency (not a peer)?

fzaninotto: No, because some components in ra-ui-material-ui use useDispatch and useSelector directly.

djhi: And ? `ra-ui-material` also depends on `ra-core`

fzaninotto: Right, but we have to be explicit about our dependencies. If ra-ui-material-ui imports functions from recat-router, it must have react-router in dependencies. I didn't invent that rule, it's enforced by TypeScript I think. 

Also, this PR doesn't change anything regarding new dependencies - it just upgrades them.

PR #3354, opened 2019-06-22
Logout problem #1870

sarathak:
fix to Logout problem #1870


Kmaschta:
Thanks for the PR! There is 2 E2E tests that are failing about the auth redirection.

djhi:
That won't work, I'm facing the same issue in #3269. I don't have a solution yet.
Doing this in this order trigger a state update loop. I'm afraid it will require much more work

djhi:
#3269 has been merged

PR #3357, opened 2019-06-24
[RFR] Fix DashboardMenuItem className prop

Kmaschta:
## Problem

When (re)using the component `<DashboardMenuItem>`, if pass it a `className` prop it is supressed by the component and not passed to its children `MenuItem`.

## Solution

1. Pass the `className` prop to its children
2. or, let this one in the default props passed in the children

I choose the second one, because this prop isn't used anyway.


fzaninotto:
Thanks!

PR #3363, opened 2019-06-25
[RFR] Docs: Move built-in validators above custom validation

heyfife:
I put the easy way to do validation first followed by advanced, custom validation methods.

We had custom ``required()`` validators all over our code since that's the first way validation is described in the docs. Hopefully this reordering will help others avoid the extra, custom work when built-in functionality already exists to cover common cases.


heyfife:
Yes. Will do this week.

fzaninotto:
Again, this doesn't make sense, and the repetition of the validators makes the doc harder to read.

heyfife:
No, it did not. I left in the built-in validation section at the top and meant to remove it. It should be clear now.

fzaninotto:
Thanks!

PR #3368, opened 2019-06-26
[RFR] Add auth hooks

fzaninotto:
- [x] [BC Break] Store `authProvider` in modern React context instead of legacy context
- [x] Add `useAuth()` hook (replacement for `<Authenticated>`)
- [x] Add `usePermissions()` hook (replacement for `<WithPermissions>`)
- [ ] ~~use `usePermissions` instead of `WithPermissions` in `<Resource>`~~ (we need an intermediate component because hooks can't execute inside a render function)
- [x] Use the new hooks in examples
- [x] Update documentation
- [x] Update upgrade guide

```jsx
import { useAuth } from 'react-admin';

const CustomRoutes = [
    <Route path="/foo" render={() => {
        useAuth();
        return <Foo />;
    }} />,
    <Route path="/bar" render={() => {
        const { authenticated } = useAuth(
            { myContext: 'foobar' }, 
            { logoutOnFailure: false }
        );
        return authenticated ? <Bar /> : <BarNotAuthenticated />;
    }} />,
];
const App = () => (
    <Admin customRoutes={customRoutes}>
        ...
    </Admin>
);
```

```jsx
import { usePermissions } from 'react-admin';

const PostDetail = props => {
    const { loaded, permissions } = usePermissions();
    if (loaded && permissions == 'editor') {
        return <PostEdit {...props} />
    } else {
        return <PostShow {...props} />
    }
};
```


packages/ra-core/src/auth/useAuth.tsx
```diff
@@ -0,0 +1,110 @@
...
+ *              return authenticated ? <Bar /> : <BarNotAuthenticated />;
+ *          }} />,
+ *     ];
+ *     const App = () => (
+ *         <Admin customRoutes={customRoutes}>
+ *             ...
+ *         </Admin>
+ *     );
+ */
+const useAuth = (authParams = emptyParams, logoutOnFailure = true) => {
```
Kmaschta: I would suggest to name the second parameter so that we use it like:

```jsx
useAuth(context, { logoutOnFailure: false });
```

That would be more explicit for an external, and lower the bar for a new React Admin user.

fzaninotto:
Switching to RFR

docs/Authorization.md
```diff
@@ -223,83 +185,79 @@ export default ({ permissions }) => (
     </Card>
 );
 ```
-{% endraw %}
 
-## Restricting Access to Content Inside Custom Pages
+## `usePermissions()` Hook
 
-You might want to check user permissions inside a [custom pages](./Admin.md#customroutes). You'll have to use the `WithPermissions` component for that. It will ensure the user is authenticated then call the `authProvider` with the `AUTH_GET_PERMISSIONS` type and the `authParams` you specify:
+You might want to check user permissions inside a [custom page](./Admin.md#customroutes). That's the purpose of the `usePermissions()` hook,which calls the `authProvider` with the `AUTH_GT_PERMISSIONS` type on mount, and returns the result when available:
```
djhi:
```suggestion
You might want to check user permissions inside a [custom page](./Admin.md#customroutes). That's the purpose of the `usePermissions()` hook,which calls the `authProvider` with the `AUTH_GET_PERMISSIONS` type on mount, and returns the result when available:
```

djhi:
:fire:

PR #3379, opened 2019-07-02
[WIP] Refactor ra-tree

djhi:
- [x] Use [@atlaskit/tree](https://atlaskit.atlassian.com/packages/core/tree) which uses [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) for better performances
- [x] Simplify and constraint usage. The tree is meant to display data, not to edit it through forms
- [x] Don't use List or ListController
- [ ] Add a loading indicator on the expanding node when needed ?
- [x] Hook to the drag&drop feature of `@atlaskit/tree` and provide default handlers
- [x] Supports reordering items
- [ ] Add optimistic support to drag and drop mutations
- [ ] Update readme & documentation

ra-tree-drag-drop


packages/ra-tree-core/src/getTreeFromArray.ts
```diff
@@ -0,0 +1,43 @@
...
+        isExpanded: expandedNodeIds.some(id => id == item.id),
+        isChildrenLoading: false,
+        data: item,
+    };
+};
+
+/**
+ * Build a tree representation of the data returned by the List component for usage with @atlaskit/tree
+ */
+export default (data, parentSource, expandedNodeIds) => {
```
fzaninotto: typing input and output would help understanding the function better.

packages/ra-tree-ui-materialui/src/NodeView.js
```diff
@@ -36,53 +36,66 @@ export class NodeView extends Component {
...
-                        : null
-                )}
-                {actions &&
-                    cloneElement(actions, {
-                        basePath,
-                        record: node.record,
-                        resource,
-                    })}
-            </div>
+            <Card className={classes.root}>
```
fzaninotto: not a big fan of the extra shadow caused by a Card. Is it really necessary?

djhi: It allows to see the different items more clearly imo. What do you suggest ?

examples/simple/src/Layout.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Layout } from 'react-admin';
+import { withStyles, createStyles } from '@material-ui/core/styles';
+
+const styles = createStyles({
+    root: {
+        overflow: 'auto', // Required for expanding nodes on hover when dragging elements in ra-tree
```
fzaninotto: can you restrict to `overflow-x`? Should we make it the default style in the ra Layout?

package.json
```diff
@@ -4,7 +4,7 @@
     "scripts": {
         "build": "lerna run build",
         "build-demo": "cd examples/demo && cross-env REACT_APP_DATA_PROVIDER=rest yarn -s build",
-        "test-unit": "cross-env NODE_ENV=test cross-env BABEL_ENV=cjs NODE_ICU_DATA=node_modules/full-icu jest",
+        "test-unit": "cross-env NODE_ENV=test cross-env BABEL_ENV=cjs NODE_ICU_DATA=node_modules/full-icu jest --watch --onlyChanged",
```
fzaninotto: please revert, I'm using the test-nuit for full tests. You can use `npx jest --watch` instead

packages/ra-tree-ui-materialui/src/NodeView.tsx
```diff
@@ -0,0 +1,113 @@
...
+        return (
+            <Card className={classes.root} {...props}>
+                <CardHeader
+                    avatar={
+                        <div {...provided.dragHandleProps}>
+                            <DragHandleIcon />
+                        </div>
+                    }
+                    action={
+                        <>
```
fzaninotto: I think you should extract this into another component

examples/simple/src/Layout.js
djhi: > Should we make it the default style in the ra Layout?

I'm not sure yet. I'll have to check our demo pages to ensure there are no side effects

macrozone:
that looks so nice, can't wait to try it out 👍

christiaanwesterbeek:
Do you plan to merge and release this with 2.x or 3.x?

djhi:
Due to a customer request, I'll work on version for react-admin v2 first, with the same features. This should be done by next week

macrozone:
@dijhi any chance that you could revisit this? would love to have this on react admin v3!

djhi:
@macrozone I'd love but unfortunately I don't have the time.

fzaninotto:
This won't be merged in react-admin v3. Closing.

macrozone:
so sad :-(

fzaninotto:
But we'll be proposing a Tree module as part of react-admin enterprise edition. It's almost ready.

macrozone:
@djhi @fzaninotto any plans to revisit the tree view or is the enterprise version ready? I think it would add much value to react-admin. Enterprise edition is currently lacking any information on the website. I understand that you want to make business with react-admin, but the information is a bit scarce currently.

fzaninotto:
@macrozone The tree view was redeveloped, it's finished and available as part of the Enterprise version. We've just launched, without much fanfare, the React-admin Enterprise-edition website. We'll do some marketing about it on September, but you can already subscribe: https://marmelab.com/ra-enterprise/

macrozone:
@fzaninotto that looks really interesting. As we use react-admin a lot, we might consider that

hacktisch:
Not sure if I should be asking this here but does this mean that the free version of React-Admin won't receive useful updates like these anymore? I agree with the idea of a paid enterprise version but I and many others use React-Admin in a personal setting and the prices as shown on the pricing page aren't realistic for our situations. How about a single purchase possibility without customer service and with updates until a limited time in the future?
I have been waiting patiently To find out that there is now a solution which I cannot access is quite an anticlimax for me.

fzaninotto:
@hacktisch We've been investing a lot in the open-source version, and we'll continue to do so (see for instance the List improvements in the latest minor version).

However, the popularity of react-admin comes with its own challenges, and maintaining the open-source edition is expensive. It currently requires a equivalent of one full-time developer. To make this sustainable, we've decided to try and sell a product based on react-admin - that's the Enterprise Edition. If this doesn't work out, we simply won't be able to maintain the open-source core anymore. That's the general problem of open-source software: if nobody pays, the entire ecosystem is too fragile and may collapse any time.

So I'm sorry you're disappointed, but I suppose you're developing web apps for a living? That's also our case, and we don't want to go bankrupt because we spend a too large share of our time in service of others who don't pay us. If you're doing it for free for a charity, then contact us - we'll offer you a rebate. If you're doing software for free and you can afford it, then you can probably afford redeveloping a tree view component as well.

We want react-admin to be a robust and long-lived solution. There is simply no other way to achieve it.

hacktisch:
@fzaninotto 
Perfectly understandable of course, this all makes sense. And let me congratulate you on launching the enterprise version. I hope it will work out positively. And I can't thank you enough for all your efforts building React-Admin for free.

Of course I can develop modules myself for React-Admin (as I have done already) and I would gladly cooperate on continuing to develop React-Admin. But up until this point, you guys took the maintenance and moderation upon yourselves and while the project was ope-source, contributing modules was not as easy as with most open-source projects considering the way it was setup. And as you are now planning to focus mainly on the Enterprise version, I suppose it would be best if other project owners from the community would be appointed. It would be bad if now everybody forks the current free code base and start making changes and publish their own adaptations, all becoming incompatible with eachother (and with the Enterprise version - this is not even good for the Enterprise version I think). 

It would of course be the best if both free and Enterprise project would always remain compatible with eachother, so that anyone using the free version will always be able to switch to the paid version. 
I am not sure what you guys will then think of it when I or others will just 'reverse-engineer' all enterprise modules and publish it open source.
I hope there can be a situation where open source and enterprise versions can remain to benefit from eachother, whilst removing  the burden that is now on your shoulders at Marmelab.

fzaninotto:
> And as you are now planning to focus mainly on the Enterprise version

Where did you read that? That's not our intention.

> it would be best if other project owners from the community would be appointed

Are you a volunteer? We need someone willing to spend at least 8 hours per week for the next 3 years, for free. Not easy to find.

> It would be bad if now everybody forks the current free code base and start making changes and publish their own adaptations

It's always been possible, it's the spirit of open-source, and I don't see what's wrong with forking. 

>  anyone using the free version will always be able to switch to the paid version

You've misunderstood what React-admin Enterprise edition is. It's a set of modules *on top of* the open-source version of react-admin. You don't have to choose between one or the other. React-admin OSS is and will always remain at the core. 

> when I or others will just 'reverse-engineer' all enterprise modules and publish it open source

That'd be a breach of intellectual property (the enterprise modules aren't open-source), and that'd be shooting yourself in the foot. Because, if we can't make money with enterprise modules, we won't be able to maintain react-admin OSS, and your web apps based on react-admin will eventually die. 

Enough spoken about this, I'm locking the conversation.

PR #3388, opened 2019-07-05
Fix Lists/Aside sum example

### LINKED ISSUE #3387: Documentation aside example issue + fix
Hi in the Documentation > chapter 6. Lists > [aside component](https://marmelab.com/react-admin/List.html#aside-component), you wrote a code to calculate the sum, which is not working.

This gave me issues, see my [stackoverflow question](https://stackoverflow.com/questions/56878628/display-total-of-grid-column-values-in-react-admin)

The issue:
```
Total views: {ids.map(id => data[id]).reduce((sum, post) => sum + post.views)}
```

The fix:
```
Total views: {ids.map(id => data[id]).reduce((sum, post) => sum + post.views, 0)}
```

firepol:
Fixes #3387

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -637,7 +637,7 @@ const Aside = ({ data, ids }) => (
     <div style={{ width: 200, margin: '1em' }}>
         <Typography variant="title">Posts stats</Typography>
         <Typography variant="body1">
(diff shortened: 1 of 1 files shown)


djhi:
Thanks!

PR #3390, opened 2019-07-07
[RFR] Add tooltip to menu entries on the sidebar

m4theushw:
Fixes #3343. The demo uses a custom menu so I updated it to also show the tooltip. I created a PR for this before but I closed it due to have chosen the wrong branch.

PR #3393, opened 2019-07-09
[RFR] Prevents passing the sidebarIsOpen prop to the MenuItem

m4theushw:
This PR is related to #3390. I forgot to add the `sidebarIsOpen` to the sanitized props before passing them to `@material-ui/core/MenuItem`. Without the fix the prop is passed to a HTML element which causes an unknown-prop warning on the console.


fzaninotto:
thanks!

PR #3398, opened 2019-07-10
[RFR] Add useEditController hook

fzaninotto:
Removes one level of indentation in the React tree for Edit views.

I also introduced a new specialized mutation hook, `useUpdate`. This addresses #2952 for the `crudUpdate` action.

```jsx
import { useEditController } from 'react-admin';
import EditView from './EditView';

const MyEdit = props => {
     const controllerProps = useEditController(props);
     return <EditView {...controllerProps} {...props} />;
}
```

The only difference between `useEditController` and `EditController` is that the former doesn't inject `translate` (just like for `useListController`).

### DIFF
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -53,6 +53,7 @@
         "classnames": "~2.2.5",
         "connected-react-router": "^6.4.0",
         "date-fns": "^1.29.0",
+        "deepmerge": "^3.0.0",
         "inflection": "~1.12.0",
         "lodash": "~4.17.5",
         "node-polyglot": "^2.2.2",
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -53,7 +53,6 @@
         "classnames": "~2.2.5",
         "connected-react-router": "^6.4.0",
         "date-fns": "^1.29.0",
-        "deepmerge": "^3.0.0",
         "inflection": "~1.12.0",
         "lodash": "~4.17.5",
         "node-polyglot": "^2.2.2",
--- a/packages/ra-core/src/controller/useEditController.ts
+++ b/packages/ra-core/src/controller/useEditController.ts
@@ -1,0 +1,124 @@
+import { useEffect, useCallback } from 'react';
+// @ts-ignore
+import { useDispatch } from 'react-redux';
+import { reset as resetForm } from 'redux-form';
+import inflection from 'inflection';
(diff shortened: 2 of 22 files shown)


packages/ra-core/package.json
```diff
@@ -53,6 +53,7 @@
         "classnames": "~2.2.5",
         "connected-react-router": "^6.4.0",
         "date-fns": "^1.29.0",
+        "deepmerge": "^3.0.0",
```
djhi: `lodash/merge` ?

packages/ra-core/src/controller/useEditController.ts
```diff
@@ -0,0 +1,132 @@
...
+                level: 'warning',
+            },
+            redirectTo: 'list',
+            refresh: true,
+        },
+    });
+
+    useEffect(() => {
+        dispatch(resetForm(REDUX_FORM_NAME));
+    }, [resource, id, version]); // eslint-disable-line react-hooks/exhaustive-deps
```
djhi: You should pass the missing deps

packages/ra-core/src/fetch/useMutation.ts
```diff
@@ -1,4 +1,5 @@
 import { useCallback } from 'react';
+import deepmerge from 'deepmerge';
```
djhi: use the existing lodash dep?

packages/ra-core/src/controller/useEditController.ts
fzaninotto: No, because the previous code was:

```jsx
    componentWillReceiveProps(nextProps: Props & EnhancedProps) {
        if (
            this.props.id !== nextProps.id ||
            nextProps.version !== this.props.version
        ) {
            this.props.resetForm(REDUX_FORM_NAME);
            this.updateData(nextProps.resource, nextProps.id);
        }
    }
```

djhi: I fail to see how it relates to the new hook version and to what hooks actually requires

djhi: Just read this from an article by Dan:
> You may omit dispatch, setState, and useRef container values from the deps because React guarantees them to be static. But it also doesn’t hurt to specify them.)

https://overreacted.io/a-complete-guide-to-useeffect/

Maybe it's true for react-redux `dispatch` as well

PR #3415, opened 2019-07-15
Improve ReferenceArrayInput error message when given bad data

Kmaschta:
When passing a bad value to a ReferenceArrayInput, we have a cryptic
error message.
Now, we explain what data was badly formatted an its location.

## Previews

**Before**

(todo)

**After**

image


fzaninotto:
thanks!

PR #3421, opened 2019-07-18
Migrate ra-data-simple-rest to typescript 

maruware:


packages/ra-data-simple-rest/src/index.ts
```diff
@@ -25,72 +96,99 @@ import {
...
+        params: RequestParams
+    ) => {
         let url = '';
-        const options = {};
+        const options: RequestOptions = {};
         switch (type) {
             case GET_LIST: {
-                const { page, perPage } = params.pagination;
-                const { field, order } = params.sort;
+                const params_ = params as GetListRequestParams;
```
fzaninotto: This shouldn't be necessary as TypeScript knows how to infer type on discimiinated unions (cf http://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions)

packages/ra-data-simple-rest/src/index.ts
```diff
@@ -10,7 +10,78 @@ import {
     UPDATE_MANY,
     DELETE,
     DELETE_MANY,
-} from 'react-admin';
+} from 'ra-core';
+
+interface Pagenation {
```
fzaninotto: there already is a Pagination type in ra-core, use that one. Same for sort

packages/ra-data-simple-rest/src/index.ts
```diff
@@ -10,7 +10,78 @@ import {
...
+
+interface GetListRequestParams {
+    pagination: Pagenation;
+    sort: Sort;
+    filter: {
+        [key: string]: any;
+    };
+}
+
+interface IdParams {
```
fzaninotto: use the Identifier type from ra-core instead

packages/ra-data-simple-rest/src/index.ts
```diff
@@ -10,7 +10,78 @@ import {
...
+    id: number | string;
+}
+interface IdsParams {
+    ids: (number | string)[];
+}
+
+interface GetOneRequestParams extends IdParams {}
```
fzaninotto: these types should be added to ra-core as they'll be required for all data providers.

packages/ra-data-simple-rest/src/index.ts
```diff
@@ -10,7 +10,78 @@ import {
...
+}
+interface UpdateManyRequestParams extends IdsParams {
+    data: any;
+}
+interface CreateRequestParams {
+    data: any;
+}
+interface DeleteRequestParams extends IdParams {}
+interface DeleteManyRequestParams extends IdsParams {}
+type RequestParams =
```
fzaninotto: users can send their own request types, so the list should not be limited.

packages/ra-data-simple-rest/src/index.ts
maruware: Thank you for reviewing.
I think "params" need to have a property to identify the type to do that (like s.kind in cf).

packages/ra-data-simple-rest/src/index.ts
maruware: I'm sorry, I'll check it.

packages/ra-data-simple-rest/src/index.ts
maruware: I'm sorry, I'll check it.

packages/ra-data-simple-rest/src/index.ts
maruware: I agree.I'll move this.

packages/ra-data-simple-rest/src/index.ts
maruware: ra-data-simple-rest appears to accept no other params type.
What about changing name "RequestParams" to "DataSimpleRestRequestParams"?

packages/ra-data-simple-rest/src/index.ts
fzaninotto: that's the `type` on which you `switch`

packages/ra-data-simple-rest/src/index.ts
fzaninotto: That's right, ra-data-simple-rest accepts no other param, but the type that react-admin must use should allow additional request types. I suppose that means adding a `default` case in ra-data-simple-rest to throw an error in case of unsupported type.

packages/ra-data-simple-rest/src/index.ts
maruware: `type` is not a property of `params`, so I think "discimiinated unions" pattern can't be applied this.

fzaninotto: yes, it should work - read again the doc I pointed to. We've already used it in the react-admin code base, in the reducers.

maruware: It means changing to like following code?

```ts
    /**
     * @param {String} req.type One of the constants appearing at the top if this file, e.g. 'UPDATE'
     * @param {String} req.resource Name of the resource to fetch, e.g. 'posts'
     * @param {Object} req.params The data request params, depending on the type
     * @returns {Object} { url, options } The HTTP request parameters
     */
    const convertDataRequestToHTTP = (req: RequestData) => {
        let url = '';
        const options: RequestOptions = {};
        switch (req.type) {
            case GET_LIST: {
                const { page, perPage } = req.params.pagination;
                const { field, order } = req.params.sort;
```

fzaninotto:
Any updates?

fzaninotto:
No news for some time, closing. Besides, the simple rest provider will be overhauled due to #3686

PR #3423, opened 2019-07-18
update react admin to postgrest

Wellers0n:
add react admin to postgrest


fzaninotto:
Thanks for your PR.

We only keep a handful of data providers in the react-admin repositories (the ones we use and that we are able to test and maintain). For other data providers, please publish them to npm and open a pull request here to have them added to the third-party data providers list in the documentation.

PR #3425, opened 2019-07-19
[RFR] Declare fetch side effects as function

### LINKED ISSUE #3413: [RFC] Using hooks for fetch side effects
## Problem

We we wrote react-admin, there was no common solution for handling side effects (like redirection, notification, or refresh after the dataProvider answers). So we've chosen redux-saga, and declared side effects as objects in the `meta` field of redux actions, as follows:

```ts
export const crudUpdate = (
    resource: string,
    id: Identifier,
    data: any,
    previousData: any,
    basePath: string,
    redirectTo: RedirectionSideEffect = 'show'
): CrudUpdateAction => ({
    type: CRUD_UPDATE,
    payload: { id, data, previousData },
    meta: {
        resource,
        fetch: UPDATE,
        onSuccess: {
            notification: {
                body: 'ra.notification.updated',
                level: 'info',
                messageArgs: {
                    smart_count: 1,
                },
            },
            redirectTo,
            basePath,
        },
        onFailure: {
            notification: {
                level: 'warning',
            },
        },
    },
});
```

In v3, Following RFC #2952, we've removed the side effects from the action object to let them be decided by the caller. The current version in `next` looks like this:

```js
    // in ra-core/src/controllers/useEditController.ts
    const [mutate, { loading: isSaving }] = useMutation(
        { type: 'EDIT', resource, payload: {} } // payload is set later by the caller of mutate
    );

    const save = (data, redirectTo = 'list') =>
        mutate(
            null, // the mutate callback ignores the first argument as it's usually an event sent by onSubmit or onClick
            { data }, // this overrides the mutation payload
            {
                onSuccess: {
                    notification: {
                        body: 'ra.notification.updated',
                        level: 'info',
                        messageArgs: {
                            smart_count: 1,
                        },
                    },
                    basePath,
                    redirectTo,
                },
                onFailure: {
                    notification: {
                        level: 'warning',
                    },
                },
            },
        );
    );
```

So we've abstracted the action away (in the `useMutation` hook) but not the side effects. 

I find these side effect objects to be an unnecessary level of indirection. They are hard to write, they require constant lookups in the documentation for syntax, they are hard to explain. Is the error side effect key 'onError' or 'onFailure'? I always have to check the right syntax. 

And React 16.8 now offers a native way to handle side effects: hooks. 

## Proposed Solution

The `useMutation` hook uses the `useDataProvider` hook. I've designed `useDataProvider` to return a Promise, so that it's easy for the developer to execute code once the call to the data provider ends. But that ability wasn't kept in the current `useMutation` implementation. If we keep that ability (returning a Promise), then side effects can be added naturally:

```ts
    // in ra-core/src/controllers/useEditController.ts
    const [mutate, { loading: isSaving }] = useMutation(
        { type: 'EDIT', resource, payload: {} } // payload is set later by the caller of mutate
    );

    const save = (data, redirectTo = 'list') =>
        mutate(null, { data })
            .then(() => {
                // success side effects go here
            })
            .catch(() => {
                // failure side effects go here
            }),
``` 

But then, how to trigger these side effects? Let's create hooks for notification, redirection, and refresh. 

```diff
    // in ra-core/src/controllers/useCreatecontroller.ts
+   const notify = useNotification();
+   const redirect = useRedirection();
    const [mutate, { loading: isSaving }] = useMutation(
        { type: 'EDIT', resource, payload: {} } // payload is set later by the caller of mutate
    );

    const save = (data, redirectTo = 'list') =>
        mutate(null, { data })
            .then(() => {
+               notify('ra.notification.created', 'info', {
+                   messageArgs: {
+                       smart_count: 1,
+                   },
+               });
+               redirect(redirectTo, basePath);
            })
            .catch(() => {
+               notify('ra.notification.http_error', 'warning');
            });
```

## Drawback

While the changes to `useMutation` and the implementation of the new hooks aren't very hard, a problem occurs with `useQuery`. That's because this hook does not return a callback as `useMutation`, so we have nothing to attach a Promise to.

```js
// current hooks signatures
const { data, total, loading, error } = useQuery({ type, resource, payload });
const [mutate, { data, total, loading, error }] = useMutation({ type, resource, payload });
```

We need a way to trigger side effects for read queries, too (e.g. redirecting to the `list` page when a request for a record in the `edit` page returns a 404).

So I suggest we change the signatures of both `useQuery` and `useMutation` to always return the query, which is a Promise:

```js
// proposed hooks signatures
const { query,  data, total, loading, error } = useQuery({ type, resource, payload });
const { mutate, data, total, loading, error } = useMutation({ type, resource, payload });
```

That way, it becomes possible to attach success and failure side effects to read queries, too:

```js
// in useEditController.js
// before
const { data: record, loading } = useGetOne(resource, id, {
    basePath,
    version, // used to force reload
    onFailure: {
        notification: {
            body: 'ra.notification.item_doesnt_exist',
            level: 'warning',
        },
        redirectTo: 'list',
        refresh: true,
    },
});

// after
const notify = useNotification();
const redirect = useRedirection();
const refresh = useRefresh();
const { query, data: record, loading } = useGetOne(resource, id, { version })
query.catch(() => {
    notify('ra.notification.item_doesnt_exist', 'warning');
    redirect('list', basePath);
    refresh();
})
```

Considering we're still in the alpha phase, I think we can afford changing these signatures.

## Benefits

- the hooks are totally idiomatic in react 16.8. This feels like "just react" instad of something weird.
- We don't need to explain metas and sagas anymore. People don't need to open a dozen files to understand how that works.
- it's easy to add custom side effects
- the new side effects hooks can be used elsewhere (not only in request callbacks)

What's your take on that change?

fzaninotto:
This is totally backwards compatible - delerative side effect objects are still interpreted by the Redux saga.

- [x] Add `useNotify`, `useRedirect`, `useRefresh`, and `useUnselectAll` hooks
- [x] Update `useDataProvider` to execute `onSuccess` and `onFailure` functions in the context of a hook
- [x] Update List, Show, Edit and Create controllers to use hook side effects instead of object side effects (allows for custom side effects)
- [x] Update DeleteButton  to use hook side effects instead of object side effects
- [x] Update documentation

image

Closes #3413. 

This change does not reduce the code in controllers, but makes it more understandable. There are less moving parts, it's more idiomatic to modern React, and it bypasses some of the redux-saga logic.

It allows to customize side effects quite easily, e.g. to display a custom notification message in case of failed HTTP request.

```jsx
const MyEdit = props => {
   const editControllerProps = useEditController(props);
   const [update, { loading: isSaving }] = useUpdate(
        props.resource,
        props.id,
        {}, // set by the caller
        editControllerProps.record
    );

    const save = useCallback(
        (data: Partial<Record>, redirectTo = 'list') =>
            update(
                null,
                { data },
                {
                    onSuccess: () => {
                        notify(/** custom notification */);
                        redirect(/** custom redirection */);
                    },
                    onFailure: error => {
                        notify(/** custom notification */);
                        redirect(/** custom redirection */);
                    },
                    undoable,
                }
            ),
        [basePath, notify, redirect, undoable, update]
    );
    return <ListView {...props} {...controllerProps} save={save} />
}
```

This also opens the path for a pure hook-based fetch side effect handling (not using the fetch saga anymore).

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -86,7 +86,7 @@ You can destructure the return value of the `useQuery` hook as `{ data, total, e
 
 ## `useQueryWithStore` Hook
 
-Internally, react-admin uses a more powerful version of `useQuery` called `useQueryWithStore`, which has an internal cache. In practice, `useQueryWithStore` persist the response from the dataProvider in the internal react-admin store, so that result remains available if the hook is called again in the future.  
+Internally, react-admin uses a more powerful version of `useQuery` called `useQueryWithStore`, which has an internal cache. In practice, `useQueryWithStore` persist the response from the dataProvider in the internal react-admin redux store, so that result remains available if the hook is called again in the future.  
 
 You can use this hook to avoid showing the loading indicator if the query was already fetched once. 
 
@@ -164,7 +164,7 @@ export const CommentList = (props) =>
     </List>;
 ```
 
-**Tip**: For simple mutations, you can use a specialised hook like `useUpdate` instead of the more generic `useMutation`. The main benefit is that `useUpdate` will update the recod in Redux store first, allowing optimistic rendering of the UI:
+**Tip**: For simple mutations, you can use a specialised hook like `useUpdate` instead of the more generic `useMutation`. The main benefit is that `useUpdate` will update the record in Redux store first, allowing optimistic rendering of the UI:
 
 ```jsx
 import { useUpdate } from 'react-admin';
@@ -206,9 +206,12 @@ Here is how to add notifications and a redirection to the `ApproveButton` compon
 
 ```diff
 // in src/comments/ApproveButton.js
-import { useMutation, UPDATE } from 'react-admin';
+-import { useMutation, UPDATE } from 'react-admin';
++import { useMutation, useNotify, useRedirect, UPDATE } from 'react-admin';
 
 const ApproveButton = ({ record }) => {
++   const notify = useNotify();
++   const redirect = useRedirect();
     const [approve, { loading }] = useMutation(
         {
             type: UPDATE,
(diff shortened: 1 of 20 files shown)


fzaninotto:
Switching to RFR

examples/simple/src/users/UserEdit.js
```diff
@@ -35,7 +35,7 @@ const UserEditToolbar = props => {
     return (
         <Toolbar {...props} classes={classes}>
             <SaveButton />
-            <DeleteWithConfirmButton />
+            <DeleteWithConfirmButton {...props} />
```
djhi: We pass props to both the toolbar and the button ?

packages/ra-core/src/fetch/useDelete.ts
```diff
@@ -0,0 +1,44 @@
+import { CRUD_DELETE } from '../actions/dataActions/crudDelete';
+import { DELETE } from '../dataFetchActions';
+import useMutation from './useMutation';
+import { Identifier } from '../types';
+
+/**
+ * Get a callback to call the dataProvider with a DELETE verb, the result and the loading state.
```
djhi: the result ?

docs/Actions.md
```diff
@@ -206,40 +206,38 @@ Here is how to add notifications and a redirection to the `ApproveButton` compon
...
 +           },
++           onFailure: (error) => notify(`Error: ${error.message}`, 'warning'),
 +       }
     );
     return <FlatButton label="Approve" onClick={approve} disabled={loading} />;
 };
 ```
 
-React-admin can handle the following side effects:
+The `onSuccess` function is called with the response from the `dataProvider` as argument. The `onError` function is called wit hthe error returned by the `dataProvider`.
```
djhi: s/wit hthe/with the/

djhi:
```suggestion
The `onSuccess` function is called with the response from the `dataProvider` as argument. The `onError` function is called with the error returned by the `dataProvider`.
```

fzaninotto:
I don't think we need to add anything to the upgrade guide, as this adds only backward compatible changes. Any application running before that PR will continue to run after that PR.

PR #3426, opened 2019-07-19
fix(EmailField): prevent click bubbling in a datagrid with rowClick

despatates:
Similar to a54acf0152745ebb9bb262ccbec82f2862a3a011.
Let's say we have a user list with an email field.
Without this fix, clicking on the email was redirecting to user/edit
before launching the mailto action.
Now, it just execute the mailto action.


fzaninotto:
Thanks!

PR #3427, opened 2019-07-19
Removed the button property from TreeNode

### LINKED ISSUE #3062: In Tree-component, every node adds a nested Ripple-effect on hover which kills performance
**What you were expecting:**

In ra-tree-ui-materialui, nested nodes don't create additional ripple effects

**What happened instead:**

Every nested nodes gets darker and creates an additional ripple effects which uses a lot of performance

**Steps to reproduce:**

Create a nested tree with ra-tree-ui-materialui and expand it. hover over one entry

**Related code:**

-

**Other information:**

the problem is visible even on the screencast: https://github.com/marmelab/react-admin/tree/4cf148571b9ec80493bca6979b2825ab2dd5e603/packages/ra-tree-ui-materialui 

**Environment**

* React-admin version: ^2.8.3
* React version: 16.8.*
* Browser: chrome

hacktisch:
To increase the performance

Fixes #3062

### DIFF
--- a/packages/ra-tree-ui-materialui/src/TreeNode.js
+++ b/packages/ra-tree-ui-materialui/src/TreeNode.js
@@ -138,7 +138,6 @@ class TreeNode extends Component {
         return connectDropTarget(
             <div className={classes.root}>
                 <ListItem
-                    button
                     classes={{
                         root: classNames({
                             [classes.node]: node.children.length > 0,


macrozone:
@hacktisch  i tried that out, but it does not make any difference. Or am i doing it wrong?

Kmaschta:
Thanks!
Can this change have other impact than just remove the overlay?

djhi:
Please note that we are completely rewriting ra-tree for 3.0

hacktisch:
> Thanks!
> Can this change have other impact than just remove the overlay?

I have to admit I cannot guarantee this, since this is the first and only part of the source code that I changed and all seems to be working well for me (it only removed the click animation on the button and all it's parents, but the click action remained to work) but I haven't tested in combination with all the other packages for instance. I would propose a better solution is to keep some sort of click animation, but prevent it from bubbling up to the parent nodes. But I didn't find how this could be done.

Is there any ETA for this rewrite of ra-tree?

djhi:
> I would propose a better solution is to keep some sort of click animation, but prevent it from bubbling up to the parent nodes. But I didn't find how this could be done.

By checking `node.childre.length` ?

> Is there any ETA for this rewrite of ra-tree?

Not yet but I'll notify you here when we'll have a RC to test :)

fzaninotto:
#3379 supersedes this PR, so I'm closing it. As @djhi said, we're rewriting the component with another approach that should improve the performance drastically.

PR #3438, opened 2019-07-21
[WIP] Avoid warning when receiving a 204 HTTP response after delete

Kmaschta:
Some API return a 204 HTTP response after successfuly deleted a resource.

In development only, we have an inopportune warning each time we delete a resource and the API return an empty 204 HTTP response.

Note : In development, this cause an error and a `RA/HIDE_NOTIFICATION` action is dispatched.
In production, some other code might break later in the process, I'll check that further.

image

### DIFF
--- a/packages/ra-core/src/sideEffect/fetch.ts
+++ b/packages/ra-core/src/sideEffect/fetch.ts
@@ -25,6 +25,10 @@ function validateResponseFormat(
     type,
     logger = console.error // eslint-disable-line no-console
 ) {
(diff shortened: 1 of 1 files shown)


fzaninotto:
The ` MUST return a `data` entry for the `DELETE` verb - some features depend on it. If the API doesn't return anything, it's the dataProvider's job to return the input data.

Kmaschta:
Ok! I tried to do so, but the error was still displayed.
I'll check that and might open a PR on the data provider, then.

PR #3439, opened 2019-07-21
[RFR] Move renderWithRedux to another package

### LINKED ISSUE #3431: [3.0.0-alpha.0 upgrade] can't resolve react-testing-library
Thank you very much for this alpha release i think it will be very helpful

after upgrading 
react-admin => 3.0.0-alpha.0
@material-ui/core => 4.2.1
@material-ui/icons => 4.2.1

i got this error

_ra-core/esm/util/renderWithRedux.js
Module not found: Can't resolve 'react-testing-library'_

i tried to install it but i got another error 

_Error: 🚨  react-testing-library has moved to @testing-library/react. Please uninstall react-testing-library and install @testing-library/react instead, or use an older version of react-testing-library. Learn more about this change here: https://github.com/testing-library/dom-testing-library/issues/260 Thanks! :)_

fzaninotto:
Closes #3431

This is to avoid having to export a dependency for a package that should only be used in dev.


fzaninotto:
The cyclic dependency creates problems, probably with Redux which should always be a singleton. As I don't want to spend ages debugging that, I'll take the shortcut in another PR by considering developers use tree shaking or their users don't care about first load performance.

fzaninotto:
Also, react-testing-library is 468 Bytes minified, so let's not worry too much about it. 

https://bundlephobia.com/result?p=react-testing-library@8.0.1

PR #3441, opened 2019-07-21
[RFR] Support API empty response on DELETE and DELETE_MANY actions

Kmaschta:
Ok, I checked for both `DELETE` and `DELETE_MANY` actions.

- `DELETE` doesn't require to have an identifier (`{ data: { id: 'id' } }`) but needs to have a data key, which should be returned by the data provider
- `DELETE_MANY` should return a data as a array, but do not require to have something in that array. If the API doesn't return a response, the data provider will return an empty array

Note : It is possible for the data provider to return the ids or resources sent with the request thanks to the `params` argument, but since react admin does nothing with this data, I preferred to remove not require it.

### DIFF
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -679,7 +679,7 @@ export default (type, resource, params) => {
                 case CREATE:
                     return { data: { ...params.data, id: json.id } };
                 case DELETE_MANY:
-                    return { data: json ? json : [] };
+                    return { data: json || [] };
                 default:
                     return { data: json };
             }
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -467,8 +467,8 @@ Request Type         | Response format
 `CREATE`             | `{ data: {Record} }`
 `UPDATE`             | `{ data: {Record} }`
 `UPDATE_MANY`        | `{ data: {mixed[]} }` The ids which have been updated
-`DELETE`             | `{ data: {Record} }` The resource that had been deleted or nothing
-`DELETE_MANY`        | `{ data: {mixed[]} }` The ids which have been deleted or an empty array
+`DELETE`             | `{ data: {Record|null} }` The record that has been deleted (optional)
+`DELETE_MANY`        | `{ data: {mixed[]} }` The ids of the deleted records (optional)
 `GET_MANY`           | `{ data: {Record[]} }`
 `GET_MANY_REFERENCE` | `{ data: {Record[]}, total: {int} }`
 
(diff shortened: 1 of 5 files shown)


packages/ra-data-json-server/src/index.js
```diff
@@ -121,6 +121,8 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
                 };
             case CREATE:
                 return { data: { ...params.data, id: json.id } };
+            case DELETE_MANY:
+                return { data: json ? json : [] };
```
fzaninotto:
```suggestion
                return { data: json || [] };
```

packages/ra-data-simple-rest/src/index.js
```diff
@@ -126,6 +126,9 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson) => {
                 };
             case CREATE:
                 return { data: { ...params.data, id: json.id } };
+            case DELETE_MANY: {
+                return { data: json ? json : [] };
```
fzaninotto:
```suggestion
                return { data: json || [] };
```

docs/DataProviders.md
```diff
@@ -678,6 +678,8 @@ export default (type, resource, params) => {
                     };
                 case CREATE:
                     return { data: { ...params.data, id: json.id } };
+                case DELETE_MANY:
+                    return { data: json ? json : [] };
```
fzaninotto:
```suggestion
                    return { data: json || [] };
```

docs/DataProviders.md
```diff
@@ -467,8 +467,8 @@ Request Type         | Response format
 `CREATE`             | `{ data: {Record} }`
 `UPDATE`             | `{ data: {Record} }`
 `UPDATE_MANY`        | `{ data: {mixed[]} }` The ids which have been updated
-`DELETE`             | `{ data: {Record} }`
-`DELETE_MANY`        | `{ data: {mixed[]} }` The ids which have been deleted
+`DELETE`             | `{ data: {Record} }` The resource that had been deleted or nothing
```
fzaninotto:
```suggestion
`DELETE`             | `{ data: {Record} }` The record that has been deleted
```

docs/DataProviders.md
```diff
@@ -467,8 +467,8 @@ Request Type         | Response format
...
+`DELETE_MANY`        | `{ data: {mixed[]} }` The ids which have been deleted or an empty array
```
fzaninotto:
```suggestion
`DELETE_MANY`        | `{ data: {mixed[]} }` The ids of the deleted records
```

docs/DataProviders.md
Kmaschta: I would precise, to the data provider maintainers that this data isn't required.

docs/DataProviders.md
Kmaschta: Idem, I would precise, to the data provider maintainers that this data isn't required.

djhi:
Any update ?

Kmaschta:
Reviews applied and rebased

Issue #3392, opened 2019-07-09
Add an `emptyName` prop to SelectInput


tdeo:
**Is your feature request related to a problem? Please describe.**
I'd like to display a custom text when the empty entry is selected in a SelectInput, to inform the user about what the default behaviour will be

**Describe the solution you'd like**
I'd like to pass a `emptyName` prop to `SelectInput` the same way we can add `emptyValue`. Ideally, this could either be a String or a React component

**Additional context**
Not sure what props should the React Component (or arguments if the prop accepts a function) receive. Maybe the `source` ?

PR #3444, opened 2019-07-23
Add emptyText prop to SelectInput

tdeo:
To address https://github.com/marmelab/react-admin/issues/3392

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -174,7 +174,7 @@ By default the component matches choices with the current input searchText: if i
 
 If you want to limit the initial choices shown to the current value only, you can set the `limitChoicesToValue` prop.
 
-When dealing with a large amount of `choices` you may need to limit the number of suggestions that are rendered in order to maintain usable performance. The `shouldRenderSuggestions` is an optional prop that allows you to set conditions on when to render suggestions. An easy way to improve performance would be to skip rendering until the user has entered 2 or 3 characters in the search box. This lowers the result set significantly, and might be all you need (depending on your data set). 
+When dealing with a large amount of `choices` you may need to limit the number of suggestions that are rendered in order to maintain usable performance. The `shouldRenderSuggestions` is an optional prop that allows you to set conditions on when to render suggestions. An easy way to improve performance would be to skip rendering until the user has entered 2 or 3 characters in the search box. This lowers the result set significantly, and might be all you need (depending on your data set).
 Ex. `<AutocompleteInput shouldRenderSuggestions={(val) => { return val.trim().length > 2 }} />` would not render any suggestions until the 3rd character was entered. This prop is passed to the underlying `react-autosuggest` component and is documented [here](https://github.com/moroshko/react-autosuggest#should-render-suggestions-prop).
 
 `<AutocompleteInput>` renders a material-ui `<TextField>` component. Use the `options` attribute to override any of the `<TextField>` attributes:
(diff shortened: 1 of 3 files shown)


docs/Inputs.md
```diff
@@ -1007,8 +1007,7 @@ const FullNameField = ({ record }) => <span>{record.first_name} {record.last_nam
 <SelectInput source="gender" choices={choices} optionText={<FullNameField />}/>
 ```
 
-Enabling the `allowEmpty` props adds an empty choice (with a default `null` value, which you can overwrite with the `emptyValue` prop) on top of the options, and makes the value nullable:
-
+Enabling the `allowEmpty` props adds an empty choice (with a default `null` value, which you can overwrite with the `emptyValue` prop) on top of the options, and makes the value nullable. You can furthermore customize the MenuItem for the emptychoice by using the `emptyText` prop, which can receive either a string or a React Element, which doesn't receive any props.
```
djhi:
```suggestion
Enabling the `allowEmpty` props adds an empty choice (with a default `null` value, which you can overwrite with the `emptyValue` prop) on top of the options, and makes the value nullable. You can furthermore customize the `MenuItem` for the empty choice by using the `emptyText` prop, which can receive either a string or a React Element, which doesn't receive any props.
```

packages/ra-ui-materialui/src/input/SelectInput.js
```diff
@@ -155,10 +155,21 @@ export class SelectInput extends Component {
         this.setState({ value });
     };
 
+    renderEmptyItemOption = emptyText => {
+        const { translate, translateChoice } = this.props;
+        if (React.isValidElement(emptyText))
+            return React.cloneElement(emptyText);
+        return translateChoice
```
djhi: I think it not necessary to check for the `translateChoice` prop here as we pass the `emptyText` as the default translation anyway

tdeo: My thought was to mimic the behaviour of the other menuitem entries, let me know what behaviour you'd rather have

djhi: Just return `translate(emptyText, { _: emptyText })` if not an element ?

djhi:
It seems prettier did not run on the modified files. Can you check ?

djhi:
ping @fzaninotto: `master` or `next` ?

fzaninotto:
Definitely `next`

tdeo:
@fzaninotto @djhi It's properly rebased on next

packages/ra-ui-materialui/src/input/SelectInput.spec.js
```diff
@@ -117,6 +117,50 @@ describe('<SelectInput />', () => {
...
+
+        const { getByRole, getByText, queryAllByRole } = render(
+            <SelectInput
+                allowEmpty
+                emptyText={emptyText}
+                {...defaultProps}
+                choices={[]}
+            />
+        );
+        const TextFieldElement = getByRole('button');
```
djhi:
```suggestion
        const EmptyMenuElement = getByRole('button');
```

packages/ra-ui-materialui/src/input/SelectInput.spec.js
```diff
@@ -117,6 +117,50 @@ describe('<SelectInput />', () => {
...
+        fireEvent.click(TextFieldElement);
```
djhi:
```suggestion
        fireEvent.click(EmptyMenuElement);
```

packages/ra-ui-materialui/src/input/SelectInput.spec.js
```diff
@@ -117,6 +117,50 @@ describe('<SelectInput />', () => {
...
+
+        const { getByRole, getByText, queryAllByRole } = render(
+            <SelectInput
+                allowEmpty
+                emptyText={emptyText}
+                {...defaultProps}
+                choices={[]}
+            />
+        );
+        const TextFieldElement = getByRole('button');
```
djhi:
```suggestion
        const EmptyMenuElement = getByRole('button');
```

packages/ra-ui-materialui/src/input/SelectInput.spec.js
```diff
@@ -117,6 +117,50 @@ describe('<SelectInput />', () => {
...
+        fireEvent.click(TextFieldElement);
```
djhi:
```suggestion
        fireEvent.click(EmptyMenuElement);
```

djhi:
Thanks!

PR #3449, opened 2019-07-23
[RFR] Fix hide filter button isn't aligned with field

### LINKED ISSUE #3432: "hide filter" button is not aligned
Screenshot from 2019-07-21 14-49-36

fzaninotto:
Closes #3432

Before:

image

After:

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/FilterFormInput.js
+++ b/packages/ra-ui-materialui/src/list/FilterFormInput.js
@@ -13,7 +13,7 @@ const sanitizeRestProps = ({ alwaysOn, ...props }) => props;
 
 const useStyles = makeStyles(theme => ({
     body: { display: 'flex', alignItems: 'flex-end' },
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/FilterFormInput.js
```diff
@@ -3,23 +3,38 @@ import PropTypes from 'prop-types';
...
 import { useTranslate } from 'ra-core';
 
 const emptyRecord = {};
 
 const sanitizeRestProps = ({ alwaysOn, ...props }) => props;
 
-const FilterFormInput = ({ filterElement, handleHide, classes, resource }) => {
+const useStyles = makeStyles(theme => ({
+    body: { display: 'flex', alignItems: 'flex-end' },
+    spacer: { width: '1em' },
```
djhi: It would be better to use `theme.spacing` here too so that spaces are consistent

PR #3452, opened 2019-07-24
[RFR] Fix ReferenceField link

fzaninotto:
The URL was always undefined. I added a test to avoid that kind of regression in the future.


fzaninotto:
Wait, some links are still wrong

fzaninotto:
And back to RFR

PR #3457, opened 2019-07-25
[RFR] Upgrade material-ui to 4.2.1

djhi:


fzaninotto:
tests fail

package.json
```diff
@@ -26,7 +26,7 @@
             "/node_modules/",
             "/lib/",
             "/esm/",
-            "/examples/simple/"
+            "/examples/simplme/"
```
danada: Sorry for resurrecting an old PR. I think this change may have not been intentional.

PR #3463, opened 2019-07-26
[RFR] Upgrade types

### LINKED ISSUE #2978: Typescript Import Errors

fzaninotto:
- [x] Introduce `@types/react-redux`
- [x] Run `yarn upgrade` to fix errors due to duplicated type dependencies
- [x] Fix errors revealed by updated types  

Supersedes #3065
May fix #2978


packages/ra-core/src/form/withDefaultValue.tsx
```diff
@@ -10,7 +10,7 @@ export interface DefaultValueProps extends InputProps {
     initializeForm: typeof initializeFormAction;
 }
 
-export class DefaultValueView extends Component<DefaultValueProps> {
+export class DefaultValueView extends Component<any> {
```
djhi: Why ?

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -35,7 +34,7 @@ const LogoutWithRef: FunctionComponent<
     Props & MenuItemProps<'li', { button: true }> // HACK: https://github.com/mui-org/material-ui/issues/16245
 > = React.forwardRef(function Logout(props, ref) {
     const { className, redirectTo, ...rest } = props;
-    const classes = useStyles();
+    const classes = useStyles({});
```
djhi: I remember doing that in my mui upgrade PR. As this is a temporary workaround, you should add a link to the github issue

packages/ra-ui-materialui/src/layout/CardActions.tsx
```diff
@@ -20,7 +20,7 @@ const CardActions = ({ className, children, ...rest }) => {
         true,
         '<CardActions> is deprecated. Please use the <TopToolbar> component instead to wrap your action buttons'
     );
-    const classes = useStyles();
+    const classes = useStyles({});
```
djhi: Same

packages/ra-core/src/form/withDefaultValue.tsx
fzaninotto: because the `connect` call at the end of the file triggers this cryptic TypeScript error:

```
Argument of type 'typeof DefaultValueView' is not assignable to parameter of type 'ComponentType<never>'.
  Type 'typeof DefaultValueView' is not assignable to type 'ComponentClass<never, any>'.
    Types of property 'propTypes' are incompatible.
      Type '{ decoratedComponent: Requireable<ReactElementLike | ((...args: any[]) => any)>; defaultValue: Requireable<any>; initializeForm: Validator<(...args: any[]) => any>; input: Requireable<object>; source: Requireable<...>; validate: Requireable<...>; }' is not assignable to type 'never'.ts(2345)
```

And I don't know how to silence it.

packages/ra-ui-materialui/src/auth/Logout.tsx
fzaninotto: Done. I just don't understand why TypeScript complains in these 2 instances of `useStyles()` and not in the many other ones in the codebase.

PR #3468, opened 2019-07-29
[RFR] Use data provider without saga

fzaninotto:
- [x] Rename the `fetch` directory in `ra-core` to `dataProvider`
- [x] [BC Break] Add `<DataProviderContext>` and use it in `<AdminCore>`
- [x] Copy `fetch` and `undo` sagas logic inside `useDataProvider`
- [x] Bypass these sagas in `useDataProvider` by no longer including the `fetch` meta
- [x] [BC Break] Reimplementing the Undo feature using an `EventEmitter`
- [x] Fix tests
- [x] Add Upgrade guide

Overall this is completely transparent unless you're using a custom app. The benefit is that the `useDataProvider` code is much easier to follow, with less layers to look at. Also, merging `fetch` and `undo` really makes sense as the Undo feature can only undo a fetch action.

The cons are the use of the EventEmitter, which I preferred over a temporary redux state, and the removal of the possibility to hack the react-admin `dataProvider` logic by replacing the `fetch` saga. 

From that point on, technically we could remove the `fetch` and `auth` sagas, as well as the `crudXXX` action creators. However, we keep them for backward compatibility's sake.


packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -0,0 +1,294 @@
...
+ *          <Fragment>
+ *              {posts.map((post, key) => <PostDetail post={post} key={key} />)}
+ *          </Fragment>
+ *     }
+ * }
+ */
+const useDataProvider = (): DataProviderHookFunction => {
+    const dispatch = useDispatch() as Dispatch;
+    const dataProvider = useContext(DataProviderContext) || defaultDataProvider;
+    const isOptimistic = useSelector(
```
djhi: I'm not sure we need this reducer to handle optimistic detection. Can't w achieve the same with metas on the action ?

fzaninotto: No, because the use case is the following:

1. user clicks on Delete post
2. the `useDataProvider` hook optimistically deletes the record from the store and redirects to the post list
3. the list calls the dataProvider for the list of posts on mount
4. While waiting for the result, the list shows the posts from the sorte, where the deleted post has been removed
5. As the deletion wasn't performed on the server yet, the server returns results including the deleted record
6. The list shows the deleted post again

To avoid that, the 'optimistic mode' blocks queries while an optimistic update is pending, so 3 and 5 never happen.

PR #3471, opened 2019-07-30
Can use custom Icon for confirm check button.

BobDu:


fzaninotto:
that's technically a new feature, so it should be PRed against `next` rather than `master`. Also, rather than adding a prop for each of the button elements you want to override, let the developer pass an entirely custom `Button` instead.

fzaninotto:

PR #3472, opened 2019-07-30
[RFR]: Fix - DateInput allowEmpty={true} sets value to ' '

### LINKED ISSUE #3171: DateInput allowEmpty={true} set value to "" instead of null
**What you were expecting:**
Creating/editing a form with a DateInput that allows empty sets the value to '' if it has been set before

**What happened instead:**
Value is set to "" instead of `null`

**Steps to reproduce:**
1. Create a DateInput-field that allows the value to be empty
2. Set the field to a date
3. Remove the data
4. Save
5. The field is now "" instead of null

**Related code:**

```js
<DateInput source="foo" allowEmpty />
```

**Other information:**

**Environment**

* React-admin version:
* React version:
* Browser:

Kunnu01:
Fixes #3171

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateInput.js
+++ b/packages/ra-ui-materialui/src/input/DateInput.js
@@ -118,6 +118,7 @@ DateInput.propTypes = {
 };
 
 DateInput.defaultProps = {
+    allowEmpty: false,
     options: {},
 };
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
I'm not sure #3171 is a real bug, see conversation there.

PR #3473, opened 2019-07-30
[RFR] Fix Review accept / reject in demo

fzaninotto:
- [x] Fix `AcceptButton` side effects (`useMutation` *requires* that side effects are passed as a function)
- [x] Fix `BulkAcceptButton` side effects
- [x] Fix `BulkRejectButton` side effects
- [x] Fix race condition in `useMutation` and `useQuery` (a too drastic caching strategy caused callbacks formed when isOptimistic was true not to be updated)


fzaninotto:
Switching to RFR

Issue #2403, opened 2018-10-09
multiline TextField


aserrallerios:
If a record was created using `LongTextInput` it can contain multiline plain text (no HTML).

In the show view I find no satisfactory way to show this multiline plain text.

`TextField` uses `Typography` and it looks like it doesn't support multiline, so if I'm not mistaken we can either create another component or change the `TextField` implementation.

djhi:
Confirmed. We need a `LongTextField` which would display its content inside a `<pre>` tag. This can probably be achieved using the `component` prop of the `
Typography` component

burik666:
Temporarily, I used the `component` attribute.
https://github.com/marmelab/react-admin/blob/e564864ab02de1395386b618e61272006e85e0c3/packages/ra-ui-materialui/src/field/TextField.js#L11-L20

djhi:
As the `TextField` component passes all its props to the underlying `Typography`, you can achieve this with:

`<TextField component="pre" />`.

`LongTextInput` is deprecated as the same logic applies (you can pass a `multiline` prop), so we won't add a new field either

PR #3481, opened 2019-08-02
[RFR] add LongTextField component

Kunnu01:
fixes: #2403


djhi:
As we deprecated the `LongTextInput` in v3 (after all it's the same as the `TextInput` with a prop which can be set by users), I wonder whether we should do the same here. @fzaninotto, what do you think ?

djhi:
To be clear, users can achieve the same results using the `<TextField>` component:

```jsx
<TextField
    source="body"
    component="pre"
    variant="body1"
/>
```

djhi:
Would you mind opening a PR to add more customization examples in the documentation of the `TextField` and `TextInput` instead, explaining how they pass material-ui props down to their root component ?

PR #3483, opened 2019-08-02
[WIP] Replace addField HOC with useField hook

djhi:


packages/ra-core/src/form/FormField.tsx
```diff
@@ -21,20 +22,11 @@ interface Props extends Omit<FieldProps<any, HTMLElement>, 'validate'> {
 }
 
 export const FormField: SFC<Props> = ({ input, validate, ...props }) => {
```
fzaninotto: input is injected by the Field, it should not be in the props, should it?

fzaninotto: Sorry, disregard my comment

djhi: It's here to check wether we need to inject it (for backward compatibility). If present, we don't need to inject the props from the hook

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -15,7 +16,7 @@ const leftPad2 = leftPad(2);
  * @returns {String} A standardized datetime (yyyy-MM-ddThh:mm), to be passed to an <input type="datetime-local" />
  */
 const convertDateToString = v => {
-    if (!(v instanceof Date) || isNaN(v)) return '';
+    if (!(v instanceof Date) || isNaN(v.getDate())) return '';
```
Kmaschta: I'm confused about `v.getDate` here..

If `v` is a Date, this function returns the day of the month, which is always a number
If `v` is a date from moment or date-fns, it returns a Date, and is never a number

Am I missing something?

Also, the `convertDateToString` doesn't respect the ISO 8601 standard: it misses the timezone at the end (should be `Z` or `+00` something).

All thing considered, can we replacethis whole function by `date.toISOString()`, or do we need to strip the timezone?

djhi: I haven't think about it, just trying to convert those inputs to TypeScript and make them use the hook from react-final-form, without changing too many things.

djhi: If you think this is a prb, open an issue and we'll address it in a dedicated PR

djhi:
Closing this in favor of multiple smaller PRs

PR #3461, opened 2019-07-25
[WIP] makeStyles and functional components

MohammedFaragallah:
Refs #3453

current status

* [x]  `examples/demo/src/layout/Login.js`
* [ ]  `ra-input-rich-text/src/index.js`
* [x]  `ra-tree-ui-materialui/src/DragLayer.js`
* [x]  `ra-tree-ui-materialui/src/DragPreview.js`
* [x]  `ra-tree-ui-materialui/src/NodeActions.js`
* [x]  `ra-tree-ui-materialui/src/NodeForm.js`
* [x]  `ra-tree-ui-materialui/src/NodeView.js`
* [x]  `ra-tree-ui-materialui/src/RootDropTarget.js`
* [x]  `ra-tree-ui-materialui/src/Tree.js`
* [x]  `ra-tree-ui-materialui/src/TreeNode.js`
* [x]  `ra-ui-materialui/src/Link.tsx`
* [ ]  `ra-ui-materialui/src/auth/Login.tsx`
* [x]  `ra-ui-materialui/src/auth/LoginForm.tsx`
* [x]  `ra-ui-materialui/src/button/CreateButton.js`
* [x]  `ra-ui-materialui/src/button/SaveButton.js`
* [x]  `ra-ui-materialui/src/field/BooleanField.tsx`
* [x]  `ra-ui-materialui/src/field/ChipField.tsx`
* [x]  `ra-ui-materialui/src/field/FileField.tsx`
* [x]  `ra-ui-materialui/src/field/ImageField.tsx`
* [x]  `ra-ui-materialui/src/form/FormInput.js`
* [x]  `ra-ui-materialui/src/form/SimpleFormIterator.js`
* [x]  `ra-ui-materialui/src/form/TabbedForm.js`
* [x]  `ra-ui-materialui/src/form/Toolbar.js`
* [ ]  `ra-ui-materialui/src/input/AutocompleteArrayInput.js`
* [x]  `ra-ui-materialui/src/input/AutocompleteArrayInputChip.js`
* [ ]  `ra-ui-materialui/src/input/AutocompleteInput.js`
* [x]  `ra-ui-materialui/src/input/CheckboxGroupInput.js`
* [ ]  `ra-ui-materialui/src/input/FileInput.js`
* [x]  `ra-ui-materialui/src/input/FileInputPreview.js`
* [ ]  `ra-ui-materialui/src/input/ImageInput.js`
* [x]  `ra-ui-materialui/src/input/Labeled.js`
[...]

Issue #3453, opened 2019-07-24
use useStyles instead of withStyles in next


fzaninotto:
Material-ui v4 promotes the usage of hooks instead of HOCs to apply CSS styles to a component. The current code base still uses a lot of `withStyles` in `next`. If you feel like contributing, an easy task is to convert these components to `useStyles`.

We need your help! Check out #3328 for an example.

- [ ] `examples/demo/src/layout/Login.js`
- [ ] `ra-input-rich-text/src/index.js`
- [x] `ra-tree-ui-materialui/src/DragLayer.js`
- [x] `ra-tree-ui-materialui/src/DragPreview.js`
- [x] `ra-tree-ui-materialui/src/NodeActions.js`
- [x] `ra-tree-ui-materialui/src/NodeForm.js`
- [x] `ra-tree-ui-materialui/src/NodeView.js`
- [x] `ra-tree-ui-materialui/src/RootDropTarget.js`
- [x] `ra-tree-ui-materialui/src/Tree.js`
- [x] `ra-tree-ui-materialui/src/TreeNode.js`
- [ ] `ra-ui-materialui/src/Link.tsx`
- [x] `ra-ui-materialui/src/auth/Login.tsx`
- [x] `ra-ui-materialui/src/auth/LoginForm.tsx`
- [x] `ra-ui-materialui/src/button/CreateButton.js`
- [x] `ra-ui-materialui/src/button/SaveButton.js`
- [x] `ra-ui-materialui/src/field/BooleanField.tsx`
- [x] `ra-ui-materialui/src/field/ChipField.tsx`
- [x] `ra-ui-materialui/src/field/FileField.tsx`
- [x] `ra-ui-materialui/src/field/ImageField.tsx`
- [x] `ra-ui-materialui/src/form/FormInput.js`
- [ ] `ra-ui-materialui/src/form/SimpleFormIterator.js`
- [x] `ra-ui-materialui/src/form/TabbedForm.js`
- [x] `ra-ui-materialui/src/form/Toolbar.js`
- [ ] `ra-ui-materialui/src/input/AutocompleteArrayInput.js`
- [ ] `ra-ui-materialui/src/input/AutocompleteArrayInputChip.js`
- [x] `ra-ui-materialui/src/input/AutocompleteInput.js`
- [x] `ra-ui-materialui/src/input/CheckboxGroupInput.js`
- [x] `ra-ui-materialui/src/input/FileInput.js`
[...]

PR #3485, opened 2019-08-03
[RFR] Replace withStyles with useStyles

m4theushw:
Refs #3453

This PR converts the remaining components from #3461.

- [x] ra-ui-materialui/src/auth/Login.tsx
- [X] ra-ui-materialui/src/layout/Notification.js

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Login.tsx
+++ b/packages/ra-ui-materialui/src/auth/Login.tsx
@@ -1,20 +1,9 @@
-import React, {
-    Component,
-    ComponentType,
-    HtmlHTMLAttributes,
-    ReactNode,
-} from 'react';
+import React, { HtmlHTMLAttributes, ReactNode, useRef, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Card from '@material-ui/core/Card';
 import Avatar from '@material-ui/core/Avatar';
-import {
-    createMuiTheme,
-    withStyles,
-    createStyles,
-    WithStyles,
-    Theme,
-} from '@material-ui/core/styles';
+import { createMuiTheme, makeStyles, Theme } from '@material-ui/core/styles';
 import { ThemeProvider } from '@material-ui/styles';
 import LockIcon from '@material-ui/icons/Lock';
 import { StaticContext } from 'react-router';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -74,86 +62,72 @@ const styles = (theme: Theme) =>
...
-                </div>
-            </ThemeProvider>
-        );
-    }
-}
+    });
 
-const EnhancedLogin = withStyles(styles)(Login) as ComponentType<Props>;
+    return (
+        <ThemeProvider theme={createMuiTheme(theme)}>
```
djhi: You should probably initialize the theme inside a `useMemo` before passing it to the `ThemeProvider`

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -74,86 +62,72 @@ const styles = (theme: Theme) =>
  *        </Admin>
  *     );
  */
-class Login extends Component<
-    Props & WithStyles<typeof styles> & HtmlHTMLAttributes<HTMLDivElement>
-> {
-    theme = createMuiTheme(this.props.theme);
-    containerRef = React.createRef<HTMLDivElement>();
-    backgroundImageLoaded = false;
+const Login: React.SFC<Props & HtmlHTMLAttributes<HTMLDivElement>> = ({
```
djhi: Import `FunctionComponent` from `react` and use it instead of `React.SFC` which is deprecated. They are both alias to the same type but `FunctionComponent` is now the official one (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30364)

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    warning: {
+        backgroundColor: theme.palette.error.light,
+    },
+    undo: {
+        color: theme.palette.primary.light,
+    },
+}));
+
+const Notification: React.SFC<Props> = ({
```
djhi: Same as `Login`. Use `FunctionComponent` imported from 'react`

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const notification = useSelector(getNotification);
+    const dispatch = useDispatch();
+    const translate = useTranslate();
+    const styles = useStyles({});
+
+    useEffect(() => {
+        setOpen(!!notification);
+    }, [notification]);
+
+    const handleRequestClose = () => {
```
djhi: Can you wrap it inside a `useCallback` ?

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -0,0 +1,118 @@
...
+        setOpen(false);
+    };
+
+    const handleExited = () => {
```
djhi: Can you wrap it inside a `useCallback` ?

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -0,0 +1,118 @@
...
+        if (notification && notification.undoable) {
+            dispatch(complete());
+            undoableEventEmitter.emit('end', { isUndo: false });
+        }
+        dispatch(hideNotification());
+    };
+
+    const handleUndo = () => {
```
djhi: Can you wrap it inside a `useCallback` ?

djhi:
As I explained in #3461, I'm going to change all inputs for react-final-form migration. I will update them to use `makeStyle` at the same time. You should omit them to avoid conflicts later

Kunnu01:
Can I work on the remaining components ( `BulkActions`, `Layout` )?

m4theushw:
@Kunnu01 yes, of course. I will change the PR to RFR.

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    warning: {
+        backgroundColor: theme.palette.error.light,
+    },
+    undo: {
+        color: theme.palette.primary.light,
+    },
+}));
+
+const Notification: React.FunctionComponent<Props> = ({
```
djhi: I think you can add the SnackbarProps as we pass rest props to the underlying Snackbar component.

```suggestion
const Notification: React.FunctionComponent<Props & SnackbarProps> = ({
```

m4theushw: Maybe not because the SnackbarProps has a non-optional [`open` property](url), which in our case is passed by Redux inside the component.

djhi: `Omit<SnackbarPros, 'open'>` ?

djhi:
Thanks!

PR #3499, opened 2019-08-09
[RFR] Introduce useInput hook

djhi:
This is the first step to avoid the `addField` HOC in our inputs.


packages/ra-core/src/form/isRequired.ts
```diff
@@ -0,0 +1,11 @@
+const isRequired = validate => {
+    if (validate && validate.isRequired) {
+        return true;
+    }
+    if (Array.isArray(validate)) {
+        return !!validate.find(it => it.isRequired);
```
Kmaschta: Would be a good candidate for a `validate.some` :)

PR #3500, opened 2019-08-09
[RFR] Convert BooleanInput to use useInput

djhi:


packages/ra-ui-materialui/src/input/InputPropTypes.ts
```diff
@@ -0,0 +1,12 @@
+import PropTypes from 'prop-types';
+
+/**
+ * Common PropTypes for all react-admin inputs
+ */
+const InputPropTypes = {
```
Kmaschta: Is this type won't be useful exported from `ra-core` ?

I think of others UI implementations for RA

djhi: Done

PR #3502, opened 2019-08-09
[RFR] Convert CheckboxGroupInput to use useInput

djhi:


packages/ra-ui-materialui/src/input/CheckboxGroupInput.js
```diff
@@ -91,160 +83,116 @@ const styles = theme =>
...
 CheckboxGroupInput.propTypes = {
     choices: PropTypes.arrayOf(PropTypes.object),
-    classes: PropTypes.object,
     className: PropTypes.string,
     label: PropTypes.string,
     source: PropTypes.string,
     options: PropTypes.object,
-    id: PropTypes.string,
-    input: PropTypes.shape({
-        onChange: PropTypes.func.isRequired,
```
Kmaschta: This property isn't available with the new version? Is this a breaking change?
Same for question for the `meta` prop

packages/ra-ui-materialui/src/input/CheckboxGroupInput.js
```diff
@@ -91,160 +83,116 @@ const styles = theme =>
...
+                // try to convert string value to number, e.g. '123'
+                newValue = JSON.parse(event.target.value);
+            } catch (e) {
+                // impossible to convert value, e.g. 'abc'
+                newValue = event.target.value;
+            }
+            if (isChecked) {
+                finalFormOnChange([...(value || []), ...[newValue]]);
+            } else {
+                finalFormOnChange(value.filter(v => v != newValue)); // eslint-disable-line eqeqeq
```
Kmaschta: Idem, why use lose equality? I don't understand

packages/ra-ui-materialui/src/input/CheckboxGroupInputItem.js
```diff
@@ -0,0 +1,62 @@
...
+            key={get(choice, optionValue)}
+            onChange={onChange}
+            control={
+                <Checkbox
+                    id={`${id}_${get(choice, optionValue)}`}
+                    color="primary"
+                    className={classes.checkbox}
+                    checked={
+                        value
+                            ? value.find(v => v == get(choice, optionValue)) !==
```
Kmaschta: Why do you want looe equality here?

packages/ra-ui-materialui/src/input/CheckboxGroupInput.js
djhi: That's because we're using the hooks from final-form

packages/ra-ui-materialui/src/input/CheckboxGroupInput.js
djhi: It was already the case in the previous version. I try to not alter how it worked. If you want to know why it's done like this, I suggest checking the file history ;)

packages/ra-ui-materialui/src/input/CheckboxGroupInputItem.js
djhi: Same answer

packages/ra-ui-materialui/src/input/CheckboxGroupInput.js
djhi: I think that's because values in form are strings, whereas our choices might be numbers, etc.

Issue #3239, opened 2019-05-17
Infinite loading view when declaring resource at runtime with no authProvider


smeng9:
**What you were expecting:**

When I try to declare the Resource at runtime, I expect the Resource would be rendered even without authProvider
According to the documentation https://marmelab.com/react-admin/Admin.html#declaring-resources-at-runtime the authProvider is completely optional

**What happened instead:**

It shows an infinite loading page when authProvider is omitted. The last dispatched action in Redux Dev Tool is RA/USER_LOGOUT and it is trying to use the none existing authProvider.

**Steps to reproduce:**

In the simple demo application
1. Remove all Resources except posts in Admin
2. Remove authProvider of Admin, then it will show an infinite loading view.

If we put authProvider back the page will render normally. If we simply have an authProvider that return Promise.resolve() for all cases, the page will also render normally. (I use this as a temporary workaround in my code.)

**Related code:**

```
https://codesandbox.io/s/simple-p6ej7
```

**Other information:**

The reason might be related with #1650. Because authProvider is not defined, it will definitely throw on AUTH_GET_PERMISSIONS. Then it's trying to log the user out with the not defined authProvider?

**Environment**

* React-admin version: 2.9.1
* Last version that did not exhibit the issue (if applicable): Not sure
* React version: 16.9 alpha 0
* Browser: Latest Chrome and Latest Firefox
* Stack trace (in case of a JS error):

Kmaschta:
You might be right, when the children of the `<Admin>` component is a function, we should assume that the `authProvider` must be set.
I don't know if it's an expected behavior nor if it's documented.

https://github.com/marmelab/react-admin/blob/f4ad22c222611ed2b8a7b0145c46ed792345d94a/packages/ra-core/src/CoreAdminRouter.tsx#L175-L176

I'll let the core maintainers decide on what to do about it.
One workaround would be to not use a a function as children when you don't have an auth provider.

fzaninotto:
If you use a function as child of `Admin`, and this function doesn't receive the permissions (because you don't have any `authProvider`, how does your function select which resources to include or not?

smeng9:
Hi @fzaninotto, sorry for the confusion here. The codesandbox is used to illustrate the problem, but our use case is different than selecting which resources to include or not.

Our use case generates all `<Resource />` on the fly from a schemas json fetched from backend:
```
    fetchResources = async (permissions) => {
        let response = await fetch(`${apiUrl}/schemas`);
[...]

PR #3505, opened 2019-08-10
[RFR] check if authProvider is defined

Kunnu01:
fixes: #3239

### DIFF
--- a/packages/ra-core/src/CoreAdminRouter.tsx
+++ b/packages/ra-core/src/CoreAdminRouter.tsx
@@ -80,7 +80,7 @@ export class CoreAdminRouter extends Component<
         try {
             const permissions = authProvider
                 ? await authProvider(AUTH_GET_PERMISSIONS)
-                : null;
+                : undefined;
             const resolveChildren = props.children as RenderResourcesFunction;
 
             const childrenFuncResult = resolveChildren(permissions);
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/CoreAdminRouter.tsx
```diff
@@ -78,7 +78,9 @@ export class CoreAdminRouter extends Component<
     ) => {
         const { authProvider } = props;
         try {
-            const permissions = await authProvider(AUTH_GET_PERMISSIONS);
+            const permissions = authProvider
+                ? await authProvider(AUTH_GET_PERMISSIONS)
+                : null;
```
djhi: I'll wait for more people to comment but I think `undefined` would make more sense here

djhi: More détails: null could mean no permissions whereas undefined would mean, we didn't even checked

Kunnu01: So should I change it to undefined?

djhi: Just wanted to clarify :) I'd still like to wait for more people to agree or disagree

Kmaschta: Let's go for `undefined`

PR #3506, opened 2019-08-10
[RFR] Trigger refresh after crudUpdate

natrim:
related to #2425 
superseed's #2823

### DIFF
--- a/packages/ra-core/src/actions/dataActions/crudUpdate.ts
+++ b/packages/ra-core/src/actions/dataActions/crudUpdate.ts
@@ -4,6 +4,7 @@ import { FETCH_END, FETCH_ERROR } from '../fetchActions';
 import {
     NotificationSideEffect,
     RedirectionSideEffect,
+    RefreshSideEffect,
 } from '../../sideEffect';
 
 export const crudUpdate = (
@@ -12,7 +13,8 @@ export const crudUpdate = (
     data: any,
     previousData: any,
     basePath: string,
-    redirectTo: RedirectionSideEffect = 'show'
+    redirectTo: RedirectionSideEffect = 'show',
+    refresh: RefreshSideEffect = true
 ): CrudUpdateAction => ({
     type: CRUD_UPDATE,
     payload: { id, data, previousData },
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/actions/dataActions/crudUpdate.ts
```diff
@@ -12,7 +13,8 @@ export const crudUpdate = (
     data: any,
     previousData: any,
     basePath: string,
-    redirectTo: RedirectionSideEffect = 'show'
+    redirectTo: RedirectionSideEffect = 'show',
+    refresh: RefreshSideEffect = true
```
djhi: Do we need a refresh if we have a redirect ? I think the default value should be `false`

natrim: same as with delete, with undoable false and redirect to same page, we need refresh to happen if no redirect

djhi: If you choose to disable optimistic updates and to redirect to the same page, isn't it your responsability to set the refresh to `true`? This will trigger a refresh and a redirection for all default cases

djhi: I'm surprised it was accepted on delete btw

natrim: but you cannot set refresh true using provided props,
the only workaround is to create new redux action and set it there

but many ppl will not even notice it, cause they would think that refresh happens like it does if its undoable (with undoable it works with the redirect on same page)

djhi: I'll consider this a workaround until v3 is out then

natrim: this "workaround" is already in next branch for delete https://github.com/marmelab/react-admin/blob/ac43cadc9f461c4dff5c02c4f0390f884206c463/packages/ra-core/src/actions/dataActions/crudDelete.ts#L16

djhi: We'll probably won't need them anymore soon

PR #3507, opened 2019-08-11
[RFR] createAdminStore - set initialState on logout

natrim:
to prevent undefined behaviour

fixes having undefined state instead of new state

ie. if you have some inital data from localStorage, then you dont wanna forget them on logout

and if initialState is function, call it to get the state

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -502,6 +502,35 @@ const App = () => (
 
 The `initialState` prop lets you pass preloaded state to Redux. See the [Redux Documentation](http://redux.js.org/docs/api/createStore.html#createstorereducer-preloadedstate-enhancer) for more details.
 
+It accepts either a function or an object:
+
+```jsx
+const initialState = {
+    theme: 'dark',
+    grid: 5,
+};
+
+const App = () => (
+    <Admin initialState={initialState} dataProvider={simpleRestProvider('http://path.to.my.api')}>
+        // ...
+    </Admin>
+);
+```
+
+```jsx
+const initialState = () => ({
+    theme: localStorage.getItem('theme'),
+    grid: localStorage.getItem('grid'),
+});
+
+const App = () => (
+    <Admin initialState={initialState} dataProvider={simpleRestProvider('http://path.to.my.api')}>
+        // ...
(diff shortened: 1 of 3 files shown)


Kmaschta:
Since it's a new feature, you should aim to the `next` branch instead of `master`.
We're preparing the V3, which is under heavy development on that branch.

Also, it would be better to prepare an enhancement issue to explain what are you suggesting, and why do you think it's a good idea.

natrim:
well it's not really fully new feature (i just though function would be nice to have in initial state)
but the bug is in state becoming undefined after user logout

consider situation like this, user goes to the app, the app initializes with initialState that contains "important" information (ie. theme, rows per page, or some "stuff")

but then the user logouts and suddenly the app state is undefined, it should really reset to initialState and not undefined

Kmaschta:
I agree with you! That being said, only bug fixes are merged on master.

By "new feature" I meant that this is a totally new behavior of React Aadmin, and this change might even break production apps in some way.

Moreover, this PR introduces a new feature in the public API: thie `initialState` prop as function.
This PR should aims at the `next` branch.

Finally, since the initialState change, mind you update the documentation about the `<Admin>` component please?

natrim:
ok,
changed pull target to next,
added notice about the function possibility to doc

Kmaschta:
I would add a word on the `UPGRADE.md` to in order to warn about the store reset, but I'll let the core maintainers decide about that.

docs/Admin.md
```diff
@@ -502,6 +502,8 @@ const App = () => (
 
 The `initialState` prop lets you pass preloaded state to Redux. See the [Redux Documentation](http://redux.js.org/docs/api/createStore.html#createstorereducer-preloadedstate-enhancer) for more details.
 
+Can be passed as function or object. If passed as function it will be called just before passing result to Redux.
```
djhi:
```suggestion
It accepts either a function or an object:
```

djhi: An example would be nice

natrim: example added

fzaninotto:
Thanks!

Issue #2685, opened 2018-12-20
DateTimeInput not working in Firefox and Safari


tomtom16:
**What you were expecting:**

Expected Behaviour is only in Chrome --> Datepicker opens, Time can also be set
<img width="300" alt="screenshot_chrome" src="https://user-images.githubusercontent.com/23172948/50304119-656d3100-048f-11e9-9cec-4509e70487c4.png">

**What happened instead:**

String value is shown, but is not editable.
<img width="315" alt="screenshot_firefox" src="https://user-images.githubusercontent.com/23172948/50304078-4e2e4380-048f-11e9-8b94-7753e2582d8a.png">

**Steps to reproduce:**

Add DateTimeInput to your Edit Configuration and open in the different browsers

**Related code:**

```
<DateTimeInput source={property.name} />
```

**Other information:**

**Environment**

* React-admin version: ^2.4.4
* Last version that did not exhibit the issue (if applicable):
* React version: ^16.5.2
* Browser: Firefox and Safari not working; Chrome working

* Stack trace (in case of a JS error):

fzaninotto:
hi, and thanks for your bug report.

What you describe is the documented behavior (see https://marmelab.com/react-admin/Inputs.html#datetimeinput), and it's due to variable browser support of the `datetime-local` input type (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local). There is nothing more we can do.

As explained in the docs, if you need a cross-browser, material-ui styled DateTime input, check out https://github.com/vascofg/react-admin-date-inputs.

martdavidson:
The confusion is that the docs only mention Safari - not Firefox. I was bit by the same.

PR #3511, opened 2019-08-12
[RFR] Migrate DateInput to use useInput hook

djhi:


packages/ra-ui-materialui/src/input/DateInput.js
```diff
@@ -23,59 +23,69 @@ const dateFormatter = v => {
...
 };
 
+/**
+ * Converts a datetime string without timezone to a date object
+ * with timezone, using the browser timezone.
+ *
+ * @param {String} value Date string, formatted as yyyy-MM-ddThh:mm
+ * @return {Date}
+ */
+const parse = value => (value ? new Date(value) : '');
```
Kmaschta: I'm still bothered by this timezone cast..
If we don't have a timezone here, it's because the `format` stripped it.

IMO, `convertDateToString` should just use `date.toISOString()` so that we keep the right timezone all the way.

djhi: As I said before, I'm reluctant to change the input behavior in this PR. Please open an issue about it so that we don't forget though

Kmaschta: Ok, I checked that out before opening a PR, and I was wrong.

This input type has two drawbacks:
- It is not supported by some major browsers used now (Firefox and Safari) #2685
- It format the date using the non-standard `yyyy-MM-ddThh:mm` format instead of [ISO 8601 Extended Format](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/toISOString) (eg. `YYYY-MM-DDTHH:mm:ss.sssZ`). Removing support for seconds and timezone.

But, the [`datetime` input type is deprecated by the WHATWG HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime).
Moreover, [no other date or time date is supported by Firefox and Safari](https://caniuse.com/#feat=input-datetime).

Hence, the `datetime-local` seems to be the right input type choice if we don't want to use external library to manage this input.

Now one question: should the `<DateTimeInput />` return a string with the format  `yyyy-MM-ddThh:mm` or a `Date` with the current user timezone and seconds set to 00?

djhi: This PR only deals with the DateInput. I opened another one for the DateTimeInput

djhi: Besides, as said before, I don't want to change its behavior here. Just convert it to hooks. Please open an issue about date/time handling if needed

Kmaschta: Sorry, I'm not sure to understand then.
By passing a `parse` function to the `useInput` you change the behavior: DateInput returns now a Date instead of a string.
Or am I missing something?

djhi: You're right, I'll remove it for now

PR #3514, opened 2019-08-13
[RFR] Migrate DateTimeInput to TypeScript

djhi:
Follow #3513

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
+++ b/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
@@ -1,7 +1,7 @@
-import React from 'react';
+import React, { FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
-import TextField from '@material-ui/core/TextField';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -14,8 +14,8 @@ const leftPad2 = leftPad(2);
  * @param {Date} value value to convert
  * @returns {String} A standardized datetime (yyyy-MM-ddThh:mm), to be passed to an <input type="datetime-local" />
  */
-const convertDateToString = value => {
-    if (!(value instanceof Date) || isNaN(value)) return '';
+const convertDateToString = (value: Date) => {
+    if (!(value instanceof Date) || isNaN(value.getDate())) return '';
```
Kmaschta: If value is type Date, can `isNaN(value.getDate())` be true?

djhi: It can be an invalid date, so yes

PR #3515, opened 2019-08-13
[RFR] Migrate DisabledInput to use useInput

djhi:


Kmaschta:
Do we still need `<DisabledInput />`?
What's the difference with `<TextInput source="id" disabled />`?

packages/ra-ui-materialui/src/input/DisabledInput.js
```diff
@@ -1,42 +1,47 @@
...
-    />
-);
+}) => {
+    const {
+        id,
+        input: { value },
+    } = useInput({
+        resource,
+        source,
+        type: 'checkbox',
```
Kmaschta: Why default it to checkbox ? :o

djhi:
> What's the difference with <TextInput source="id" disabled />?

It's not disabled and we don't pass the event handlers (change, blur, focus)

djhi:
PS: I agree with you though. I think having a readonly textinput is weird, especially without a `readonly` nor `disabled` attribute and it's probably not good regarding accessibility.

I suggest we either deprecate it and add a `readonly` attribute. We should also print a message to explain how to migrate from it

djhi:
However, this is not the purpose of this PR. I'm just migrating components from the injection pattern to hooks usage. Let's open an issue about the DisabledInput

PR #3516, opened 2019-08-13
[RFR] Migrate TextInput and LongTextInput to use useInput

djhi:


packages/ra-ui-materialui/src/input/LongTextInput.js
```diff
@@ -1,26 +1,16 @@
 import React from 'react';
-import { addField } from 'ra-core';
-import { TextInput } from './TextInput';
+import TextInput from './TextInput';
 
 /**
  * @deprecated use <TextInput multiline /> instead
  */
 export const LongTextInput = props => {
-    console.warn(
```
Kmaschta: Meh? This is no longer deprecated?

djhi: I was sure I removed another warning...

PR #3517, opened 2019-08-13
[RFR] Remove BulkActions in V3

Kunnu01:
- [x] Remove BulkActions


UPGRADE.md
```diff
@@ -173,6 +173,7 @@ Components deprecated in 2.X have been removed in 3.x. This includes:
 * `ViewTitle` (use `Title` instead)
 * `RecordTitle` (use `TitleForRecord` instead)
 * `TitleDeprecated` (use `Title` instead)
+* `BulkActions` (use `BulkActionButtons` props instead)
```
Kmaschta:
```suggestion
* `BulkActions` (use the [`bulkActionButtons` prop](https://marmelab.com/react-admin/List.html#bulk-action-buttons) instead)
```

Kunnu01:
> Removing this component also implies updating the `List` component which used it

`List` uses a prop name `bulkActions`. Are you talking about that?

djhi:
> List uses a prop name bulkActions. Are you talking about that?

Yes 

> If yes, ListToolbar also uses it. Should I remove it from both the files?

Yes :)

packages/ra-ui-materialui/src/list/ListToolbar.js
```diff
@@ -52,7 +51,6 @@ const ListToolbar = ({
                 React.cloneElement(actions, {
                     ...rest,
                     className: styles.actions,
-                    bulkActions,
```
djhi: I think you'll have to update the default ListActions too

fzaninotto:
Neat! I think the Upgrade guide should be a bit more explicit, by showing a code snippet with a before/after diff.

djhi:
Do you mean we should remove all other deprecated components (LongTextInput for example)?

Kunnu01:
Okay, I'll update it. Do you want it in a specific format or should I just simply add code snippets saying `before` and `after`?

fzaninotto:
Check the rest of the UPGRADE file: we use snippets with the `diff` syntax. Try to do the same.

Kunnu01:
Done. Please review. :)

fzaninotto:
Thanks!

PR #3523, opened 2019-08-13
[RFR] Migrate NullableBooleanInput to TypeScript

djhi:
Follow #3522


packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -38,7 +40,7 @@ const NullableBooleanInput = ({
     validate,
     ...rest
 }) => {
-    const classes = useStyles();
+    const classes = useStyles({});
```
Kmaschta: Just curiosity, but why is the impact of the brackets as arguments of `useStyles`?

djhi: None. See https://github.com/mui-org/material-ui/issues/16867

djhi: In a nutshell, it's a workaround for TypeScript

PR #3525, opened 2019-08-13
[RFR] Migrate NumberInput to TypeScript

djhi:
Follow #3524

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.tsx
@@ -12,6 +12,10 @@ const parse = value => {
     return isNaN(float) ? null : float;
 };
 
+interface Props {
+    step: string | number;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/NumberInput.tsx
```diff
@@ -87,7 +89,7 @@ NumberInput.propTypes = {
     options: PropTypes.object,
     resource: PropTypes.string,
     source: PropTypes.string,
-    step: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+    step: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
Kmaschta: I might be wrong, but `step` is not in `InputProps<TextFieldProps>` nor in `Omit<TextFieldProps, 'label' | 'helperText'>`.

Should we add this prop in a new interface or something?

PR #3526, opened 2019-08-14
[RFR] Migrate SelectInput to use useInput

djhi:


packages/ra-core/src/form/useInput.ts
```diff
@@ -86,6 +89,16 @@ const useInput = ({
         [onFocus, customOnFocus]
     );
 
+    // If there is an input prop, this input has already been enhanced by final-form
+    if (options.input) {
```
djhi: Not sure of the performances implications of calling the `useField` from final-form at L51 but ignoring its result here. Final-form *should* be optimized and we don't have any other choice to comply to the hooks rules anyway

PR #3538, opened 2019-08-19
[RFR] Migrate RadioButtonGroupInput to use useInput

djhi:


packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
```diff
@@ -78,70 +76,41 @@ const styles = createStyles({
...
     optionText,
     optionValue,
+    resource,
+    source,
     translateChoice,
+    validate,
     ...rest
 }) => {
-    const translate = useTranslate();
+    const classes = useStyles({});
```
fzaninotto: I tend to let a way for developers to override the styles, by accepting a `classes` that I pass to `useStyles` at runtime 

See example at https://github.com/marmelab/react-admin/blob/8cdd8cedcf90967404c7e6d71a3403c3a35f8a1c/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.js#L42-L54

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
```diff
@@ -78,70 +76,41 @@ const styles = createStyles({
...
+        resource,
+        source,
+        validate,
+        ...rest,
+    });
 
     return (
         <FormControl
             component="fieldset"
-            className={className}
```
fzaninotto: why did you remove the ability to style the main component? It's an unnecessary breaking change.

packages/ra-ui-materialui/src/input/RadioButtonGroupInputItem.js
```diff
@@ -0,0 +1,38 @@
...
+            label={
+                translateChoice
+                    ? translate(choiceName, { _: choiceName })
+                    : choiceName
+            }
+        />
+    );
+};
+
+export default RadioButtonGroupInputItem;
```
fzaninotto: the previous function used `useCallback`, this component should use `React.memo` or `useMemo` to achieve the same.

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
```diff
@@ -1,123 +1,157 @@
...
         expect(input1.checked).toBeFalsy();
-        const input2 = getByLabelText('Female');
+        const input2 = getByLabelText('Mastercard');
         expect(input2.type).toBe('radio');
-        expect(input2.name).toBe('foo');
+        expect(input2.name).toBe('type');
         expect(input2.checked).toBeFalsy();
     });
 
     it('should use the input parameter value as the initial input value', () => {
```
fzaninotto: you've completely changed the tes implementation, so the test title no longer corresponds to the implementation.

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
```diff
@@ -143,60 +182,107 @@ describe('<RadioButtonGroupInput />', () => {
...
                 />
             </TranslationContext.Provider>
         );
-        expect(queryByText('**Male**')).toBeNull();
-        expect(queryByText('Male')).not.toBeNull();
+        expect(queryByText('**Mastercard**')).toBeNull();
+        expect(queryByText('Mastercard')).not.toBeNull();
     });
 
     it('should displayed helperText if prop is present in meta', () => {
```
fzaninotto:
```suggestion
    it('should display helperText if prop is present in meta', () => {
```

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
djhi: I didn't remove it, it's in the `rest` prop

packages/ra-ui-materialui/src/input/RadioButtonGroupInputItem.js
djhi: Premature optimization ?

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
djhi: I don't really get that comment. `initialValues` provides the initial input value and we check that is indeed applied to the underlying input

packages/ra-ui-materialui/src/input/RadioButtonGroupInputItem.js
djhi: I mean, we used `useCallback` because we weren't using *standard* react to render the items. We should let React do its job

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.js
djhi: I'll probably have to check all previously migrated components for this

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.js
fzaninotto: we no longer use the 'input parameter value', as the test description says

PR #3542, opened 2019-08-19
[RFR] Migrate ReferenceInput to TypeScript

djhi:


packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -226,14 +242,11 @@ export const ReferenceInputView = ({
         isRequired,
         label,
         resource,
-        meta: {
-            ...meta,
-            helperText: warning || false,
-        },
+        helperText: warning || helperText,
```
fzaninotto: is it a bug fix or a breaking change?

djhi: A breaking change we already discussed. All inputs should accept an helperText props and it should not come from the meta. Missing upgrade guide though

djhi: However, I wonder whether it makes sense for the ReferenceXXX inputs as they receive another input as their child. I think this child should be one handling the helper text

fzaninotto: indeed

djhi: By the way, the ReferenceXXX inputs used to override their child `helperText` to display a warning when a referenced item is not found. I believe they should not override anything and instead either pass the warning explicitly (new warning prop for their child) or pass it as the `error`.

fzaninotto: Can you add the upgrade guide for the helper text?

djhi: I'd rather wait that we discuss about how to handle those cases

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -226,14 +258,10 @@ export const ReferenceInputView = ({
...
         resource,
-        meta: {
-            ...meta,
-            helperText: warning || false,
-        },
+        meta: finalMeta,
         source,
         choices,
         basePath,
-        onChange,
```
fzaninotto: why that change?

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -172,28 +188,29 @@ const sanitizeRestProps = ({
...
 
 export const ReferenceInputView = ({
     allowEmpty,
     basePath,
     children,
     choices,
-    classes,
-    className,
-    error,
+    classes = undefined,
```
fzaninotto: I don't understand why you initialize the value to `undefined`. That's the default, right?

djhi: TypeScript. I could introduce yet another props interface though

UPGRADE.md
```diff
@@ -797,13 +797,53 @@ const OrderEdit = (props) => (
...
 
+If you've implemented a custom child component for `<ReferenceInput>` of `<ReferenceArrayInput>`, you must now display the `error` in your component.
+
+```diff
+const MySelectIpnut = ({
+    // ...
+-   meta: { helperText }
++   meta: { error }
+}) => (
+    if (error) {
```
djhi: I'm not sure about this sample. The error is inside the meta, just like any final-form error. Custom components should display it like any other form error

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
djhi: Because user defined `onChange` is now handled in the `ReferenceInput` component through `useInput` (ensuring that it does not prevent final-form from working).

`onChange` is then injected back through the `input` prop inside `ReferenceInputView`

PR #3543, opened 2019-08-19
Fix SimpleList showing ListItem as button even though linkType is false

b-raines:
When `linkType` is set to false, the `ListItems` are still styled as buttons. This seems like undesired behavior for a non-interactive item. Instead, do not style as buttons if linkType is false (or null).

Another solution would be to add a new prop to List to explicitly declare whether ListItems were styled as buttons if there was concern about altering existing behavior.


fzaninotto:
Thanks!

PR #3545, opened 2019-08-19
[RFR] Provide empty string to label prop of SearchInput

jaytula:
In the example/simple and example/demo app, the `SearchInput` component is used without a `label` prop and because the `label` prop has a default of `false`, this generates a translate warning.

image

This PR simply provides an empty string for the label prop of `SearchInput` whereever it is used in the example/simple and example/demo app.

### DIFF
--- a/examples/demo/src/products/ProductList.js
+++ b/examples/demo/src/products/ProductList.js
@@ -26,7 +26,7 @@ const QuickFilter = ({ label }) => {
 
 export const ProductFilter = props => (
     <Filter {...props}>
-        <SearchInput label="" source="q" alwaysOn />
+        <SearchInput source="q" alwaysOn />
         <ReferenceInput
             source="category_id"
             reference="categories"
--- a/examples/demo/src/products/ProductList.js
+++ b/examples/demo/src/products/ProductList.js
@@ -26,7 +26,7 @@ const QuickFilter = ({ label }) => {
 
 export const ProductFilter = props => (
     <Filter {...props}>
-        <SearchInput source="q" alwaysOn />
+        <SearchInput label="" source="q" alwaysOn />
         <ReferenceInput
             source="category_id"
             reference="categories"
--- a/examples/demo/src/reviews/ReviewFilter.js
+++ b/examples/demo/src/reviews/ReviewFilter.js
@@ -17,7 +17,7 @@ const ReviewFilter = props => {
     const classes = useFilterStyles();
     return (
         <Filter {...props}>
-            <SearchInput source="q" alwaysOn />
+            <SearchInput label="" source="q" alwaysOn />
             <SelectInput
                 source="status"
                 choices={[
--- a/examples/demo/src/reviews/ReviewFilter.js
+++ b/examples/demo/src/reviews/ReviewFilter.js
@@ -17,7 +17,7 @@ const ReviewFilter = props => {
     const classes = useFilterStyles();
     return (
         <Filter {...props}>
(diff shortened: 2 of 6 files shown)


djhi:
Wouldn't the proper fix be to not output a label at all in `TextInput` if `label === false` ?

jaytula:
Good point. I'll give that a try.

djhi:
Sorry, maybe not in `TextInput` but rather in `FieldTitle` which should return `null` if `label` strictly equals to `false`.

jaytula:
Ok.  So that means leave `<TextInput label={false}>` as is and change the the `FieldTitle` (nested under `ResettableTextField` label prop to `label === false ? null : label` in `TextInput.tsx`?

djhi:
Nope, I meant changing `https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/util/FieldTitle.tsx` so that it returns `null` if `label === false`

jaytula:
Ah, I see.  Thanks!  Will get to it tomorrow morning.

jaytula:
I get a TypeScript warning after making the change because `label` is not expected to be `boolean`  in the `FieldTitle` component.

image

If I change the typing to be either `string | boolean` the red squiggles move to `getFieldLabelTranslationArgs` because it also expects the `label` to be a `string`:

image

Upon inspecting `getFieldLabelTranslationArg.ts` https://github.com/marmelab/react-admin/blob/38fc6290d573da4d9128360916ff5954a9b62727/packages/ra-core/src/util/getFieldLabelTranslationArgs.ts, this function also expects the `label` prop in the `Args` interface to be a `string` as well as the the first value (i.e. `label`) of the returned array (as specified in `TranslationArguments`.

It appears that more components and functions expect the `label` to be purely a `string` or `undefined` but not `boolean`.  Please let me know your thoughts on how best to fix it.  Thanks for helping me dig and understand this better.

djhi:
Just convert the label to string if not equal to `false` ? `label.toString()` to be sure ?

fzaninotto:
I believe `SearchInput` should set the label to `''` instead of `false`, *and* that `FieldTitle` shouldn't try to call `translate` if the label is falsy.

jaytula:
@fzaninotto Setting label as `<SearchInput label="" />` works fine but not calling `translate` if `FieldTitle` is falsy results in failing unit-tests because the `source` prop is used by `translate` when `label` is falsy.  

Just to be clear, this is what I changed `FieldTitle.tsx` to:

```javascript
import React, { SFC } from 'react';
import pure from 'recompose/pure';

import useTranslate from '../i18n/useTranslate';
import getFieldLabelTranslationArgs from './getFieldLabelTranslationArgs';

interface Props {
    isRequired?: boolean;
    resource?: string;
    source?: string;
    label?: string;
}

export const FieldTitle: SFC<Props> = ({
    resource,
    source,
    label,
    isRequired,
}) => {
    return (
        <span>
            {label ? translate(
                ...getFieldLabelTranslationArgs({ label, resource, source })
            ) : ''}
            {isRequired && ' *'}
        </span>
    );
};

// wat? TypeScript looses the displayName if we don't set it explicitly
FieldTitle.displayName = 'FieldTitle';

export default pure(FieldTitle);
```
Here is the result summary of `make test-unit`:

image

fzaninotto:
This PR needs rebase.

jaytula:
> This PR needs rebase.

It looks like the original issue has been resolved in the `next` branch because the component is now provided with an empty string default for the `label` prop.

I'll go ahead and close this.

PR #3546, opened 2019-08-20
[RFR] Refactor withStyles to useStyles for 10 components

jaytula:
Refactored components are: CreateButton, FormInput, ToolBar, Labeled, RadioButtonGroupInput, NotFound, BulkActionsToolbar, DatagridHeaderCell, DatagridLoading, SimpleList

This is for issue #3453 .


packages/ra-ui-materialui/src/button/CreateButton.js
```diff
@@ -4,18 +4,17 @@ import onlyUpdateForKeys from 'recompose/onlyUpdateForKeys';
...
-import compose from 'recompose/compose';
 import classnames from 'classnames';
 import { Link } from 'react-router-dom';
 import { useTranslate } from 'ra-core';
 
 import Button from './Button';
 
-const styles = theme =>
+const useStyles = makeStyles(theme =>
     createStyles({
```
fzaninotto: you don't need `createStyles` when using `makeStyles`

packages/ra-ui-materialui/src/button/CreateButton.js
```diff
@@ -30,16 +29,17 @@ const styles = theme =>
...
 
 const CreateButton = ({
     basePath = '',
     className,
-    classes = {},
     label = 'ra.action.create',
     icon = <ContentAdd />,
     ...rest
 }) => {
+    const classes = useStyles();
```
fzaninotto: As we want to let developers override default styles, react-admin components still accept a `classes` prop that theu pass to `useStyles`. Here is an example: https://github.com/marmelab/react-admin/blob/f8a8b9f0b42f6bf88d11e4b1f6d06dbde95a934a/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.js#L42-L54

packages/ra-ui-materialui/src/button/CreateButton.js
```diff
@@ -69,15 +69,10 @@ const CreateButton = ({
...
     size: PropTypes.string,
     icon: PropTypes.element,
 };
 
-const enhance = compose(
-    onlyUpdateForKeys(['basePath', 'label', 'translate']),
-    withStyles(styles)
-);
-
+const enhance = onlyUpdateForKeys(['basePath', 'label', 'translate']);
```
fzaninotto: no need for an `enhance` variable if there is only one HOC: you can inline it with `export`

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: Thanks! Will make the changes.

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: Ah, I see. Will make the changes.

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: Makes sense. Will make the changes.

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: I've removed the un-needed `createStyles`

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: I've added the the ability to override the `classes` prop for the components.

packages/ra-ui-materialui/src/button/CreateButton.js
jaytula: I've inlined the cases where there is only a single HOC

jaytula:
I've made requested changes. Thank you again for reviewing and providing instructive advice on how to fix the shortcomings. Looking forward to a re-review.  Thanks!

fzaninotto:
Thanks!

PR #3549, opened 2019-08-20
[RFR] Remove DisabledInput

djhi:
- [x] Remove the component
- [x] Update docs
- [x] Upgrade guide


fzaninotto:
I can't find the `readOnly` prop in the material-ui documentation for `TextField`. It seems it's called `disabled`. Am I missing something?

https://material-ui.com/api/text-field/

djhi:
You're right, according to mui documentation it should be:

```
InputProps={{
     readOnly: true,
}}
```

fzaninotto:
But doesn't the `disabled` prop suffice?

djhi:
`disabled` !== `readonly` semantically.

Besaides, disabled inputs won't even be sent in the form payload.

fzaninotto:
The question is: is it our job to document this subtle distinction? Or should we encourage the simplest way (in my mind, `disabled`)?

PR #3554, opened 2019-08-21
[RFR] Change to useStyles for SaveButton component

jaytula:
Converted `SaveButton` to a function component and replace `withStyles` with `useStyles`. Thank you in advance for reviewing.


fzaninotto:
Thanks!

PR #3562, opened 2019-08-21
[RFR]: use makeStyles in Field Components

Kunnu01:
ref: #3453 
Converts following components:
- [x] BooleanField
- [x] ChipField
- [x] FileField
- [x] ImageField

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -41,13 +41,14 @@ export const BooleanField: SFC<
     Props & InjectedFieldProps & TypographyProps
 > = ({
     className,
+    classes: classesOverride,
     source,
     record = {},
     valueLabelTrue,
     valueLabelFalse,
     ...rest
 }) => {
-    const classes = useStyles({});
+    const classes = useStyles({ classes: classesOverride });
     const translate = useTranslate();
     const value = get(record, source);
     let ariaLabel = value ? valueLabelTrue : valueLabelFalse;
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -5,14 +5,14 @@ import pure from 'recompose/pure';
 import FalseIcon from '@material-ui/icons/Clear';
 import TrueIcon from '@material-ui/icons/Done';
 import Typography, { TypographyProps } from '@material-ui/core/Typography';
-import { createStyles, withStyles, WithStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/core/styles';
 import compose from 'recompose/compose';
 import { useTranslate } from 'ra-core';
 
 import { FieldProps, InjectedFieldProps, fieldPropTypes } from './types';
 import sanitizeRestProps from './sanitizeRestProps';
 
-const styles = createStyles({
+const useStyles = makeStyles({
     label: {
         // Move the text out of the flow of the container.
         position: 'absolute',
@@ -37,19 +37,17 @@ interface Props extends FieldProps {
     valueLabelFalse?: string;
 }
 
-interface EnhancedProps extends WithStyles<typeof styles> {}
-
 export const BooleanField: SFC<
-    Props & InjectedFieldProps & EnhancedProps & TypographyProps
+    Props & InjectedFieldProps & TypographyProps
 > = ({
     className,
-    classes,
     source,
     record = {},
     valueLabelTrue,
     valueLabelFalse,
     ...rest
 }) => {
+    const classes = useStyles({});
     const translate = useTranslate();
     const value = get(record, source);
     let ariaLabel = value ? valueLabelTrue : valueLabelFalse;
@@ -101,12 +99,9 @@ export const BooleanField: SFC<
 };
 
 const EnhancedBooleanField = compose<
-    Props & InjectedFieldProps & EnhancedProps & TypographyProps,
+    Props & InjectedFieldProps & TypographyProps,
     Props & TypographyProps
->(
-    pure,
-    withStyles(styles)
-)(BooleanField);
+>(pure)(BooleanField);
 
 EnhancedBooleanField.defaultProps = {
     addLabel: true,
--- a/packages/ra-ui-materialui/src/field/ChipField.spec.js
+++ b/packages/ra-ui-materialui/src/field/ChipField.spec.js
@@ -7,7 +7,7 @@ describe('<ChipField />', () => {
     afterEach(cleanup);
 
     it('should display the record value added as source', () => {
-        const { getByAttribute } = render(
+        const { getAttribute } = render(
             <ChipField
                 className="className"
                 classes={{}}
@@ -16,11 +16,11 @@ describe('<ChipField />', () => {
             />
         );
 
-        expect(getByAttribute('label')).toEqual('foo');
+        expect(getAttribute('label')).toEqual('foo');
     });
 
     it('should not display any label added as props', () => {
-        const { getByAttribute } = render(
+        const { getAttribute } = render(
             <ChipField
                 className="className"
                 classes={{}}
@@ -30,6 +30,6 @@ describe('<ChipField />', () => {
             />
         );
 
-        expect(getByAttribute('label')).toEqual('foo');
+        expect(getAttribute('label')).toEqual('foo');
     });
 });
--- a/packages/ra-ui-materialui/src/field/ChipField.spec.js
+++ b/packages/ra-ui-materialui/src/field/ChipField.spec.js
@@ -7,7 +7,7 @@ describe('<ChipField />', () => {
     afterEach(cleanup);
 
     it('should display the record value added as source', () => {
-        const { getAttribute } = render(
+        const { getByLabelText } = render(
             <ChipField
                 className="className"
                 classes={{}}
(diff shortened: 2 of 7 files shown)


Kunnu01:
@djhi @fzaninotto 
In FileField test cases, some tests are failing when comparing the `classes` of the component.
For eg -  [here](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/field/FileField.spec.js#L15) 
expected is `<div class=" "></div>` and output is `<div class="makeStyles-root-1" classes="[object Object]"></div>` because of `makeStyles`.
Similar is the issue with `ImageField` test cases.
What should I do in this case?

djhi:
Would be best to migrate these tests to testing-library instead of enzyme imo

Kunnu01:
Done, please review.

djhi:
Sorry, I updated my comment after posting it. I meant all the tests for those files, not just the ImageField.

packages/ra-ui-materialui/src/field/ImageField.spec.js
```diff
@@ -28,14 +28,14 @@ describe('<ImageField />', () => {
             />
         );
 
-        const img = wrapper.find('img');
-        assert.equal(img.prop('src'), 'http://foo.com/bar.jpg');
-        assert.equal(img.prop('alt'), 'Hello world!');
-        assert.equal(img.prop('title'), 'Hello world!');
+        const img = getByRole('img');
+        expect(img.prop('src')).toEqual('http://foo.com/bar.jpg');
```
djhi: `img` is an HTML `img` element. You can use `getAttribute` instead of `prop`

packages/ra-ui-materialui/src/field/ImageField.spec.js
```diff
@@ -28,14 +28,14 @@ describe('<ImageField />', () => {
             />
         );
 
-        const img = wrapper.find('img');
-        assert.equal(img.prop('src'), 'http://foo.com/bar.jpg');
-        assert.equal(img.prop('alt'), 'Hello world!');
-        assert.equal(img.prop('title'), 'Hello world!');
+        const img = getByRole('img');
```
djhi: use either `getByAltText` or `getByTitle`: https://testing-library.com/docs/dom-testing-library/api-queries#byalttext

djhi: That way, you can avoid the assertion on either the title or alt attribute

Kunnu01: Okay, thank you :)

packages/ra-ui-materialui/src/field/ImageField.spec.js
```diff
@@ -29,9 +29,9 @@ describe('<ImageField />', () => {
         );
 
         const img = getByRole('img');
-        expect(img.prop('src')).toEqual('http://foo.com/bar.jpg');
-        expect(img.prop('alt')).toEqual('Hello world!');
-        expect(img.prop('title')).toEqual('Hello world!');
+        expect(img.src).toEqual('http://foo.com/bar.jpg');
```
djhi: That works too :+1:

Kunnu01: How can I run these test files individually on my system? I am trying to run using jest but it giving syntax error because of `testing-library`.

djhi: Can you paste the syntax error here?

Kunnu01: TestFail

packages/ra-ui-materialui/src/field/ChipField.spec.js
```diff
@@ -1,33 +1,33 @@
...
-                    classes={{}}
-                    source="name"
-                    record={{ name: 'foo' }}
-                    label="bar"
-                />
-            ).prop('label'),
-            'foo'
-        ));
+    it('should display the record value added as source', () => {
+        const { getByLabelText } = render(
```
djhi: `getByLabelText` would return inputs from their label or elements using `aria-label`. However, you're looking for *basic* html text here: use `getByText`.

packages/ra-ui-materialui/src/field/ChipField.spec.js
```diff
@@ -1,33 +1,33 @@
...
+                classes={{}}
+                source="name"
+                record={{ name: 'foo' }}
+            />
+        );
+        expect(getByLabelText('foo')).toBeDefined();
+    });
+
+    it('should not display any label added as props', () => {
+        const { getByLabelText } = render(
```
djhi: Same here

packages/ra-ui-materialui/src/field/FileField.spec.js
```diff
@@ -79,12 +79,12 @@ describe('<FileField />', () => {
             />
         );
 
-        const link = wrapper.find('a');
-        assert.equal(link.prop('target'), '_blank');
+        const link = getByTitle('Hello world!');
```
djhi: The record does not have a title. You can add it at L76. See next test

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -37,19 +37,17 @@ interface Props extends FieldProps {
...
 }
 
-interface EnhancedProps extends WithStyles<typeof styles> {}
-
 export const BooleanField: SFC<
-    Props & InjectedFieldProps & EnhancedProps & TypographyProps
+    Props & InjectedFieldProps & TypographyProps
 > = ({
     className,
-    classes,
```
djhi: Keep it but rename it `classesOverride`:

`classes: classesOverride`

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -37,19 +37,17 @@ interface Props extends FieldProps {
...
     source,
     record = {},
     valueLabelTrue,
     valueLabelFalse,
     ...rest
 }) => {
+    const classes = useStyles({});
```
djhi: Users should be able to override classes. Pass `classesOverride` to the useStyles like this:

```suggestion
    const classes = useStyles({ classes: classesOverride });
```

packages/ra-ui-materialui/src/field/ChipField.spec.js
```diff
@@ -1,33 +1,33 @@
...
+    it('should display the record value added as source', () => {
+        const { getByText } = render(
+            <ChipField
+                className="className"
+                classes={{}}
+                source="name"
+                record={{ name: 'foo' }}
+            />
+        );
+        expect(getByText('foo')).toBeDefined();
```
djhi: Use `not.toBeNull`. `getByText` might return null and the test would pass otherwise.

packages/ra-ui-materialui/src/field/ChipField.spec.js
```diff
@@ -1,33 +1,33 @@
...
+        const { getByText } = render(
+            <ChipField
+                className="className"
+                classes={{}}
+                source="name"
+                record={{ name: 'foo' }}
+                label="bar"
+            />
+        );
+        expect(getByText('foo')).toBeDefined();
```
djhi: Same here

packages/ra-ui-materialui/src/field/ChipField.tsx
```diff
@@ -3,33 +3,37 @@ import compose from 'recompose/compose';
...
-        {...sanitizeRestProps(rest)}
-    />
-);
+export const ChipField: SFC<FieldProps & InjectedFieldProps & ChipProps> = ({
+    className,
+    source,
+    record = {},
+    ...rest
+}) => {
+    const classes = useStyles({});
```
djhi: Same as previous component, allow to override classes

packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -17,10 +17,17 @@ interface Props extends FieldProps {
...
+    className,
+    record,
+    source,
+    title,
+    src,
+    target,
+    ...rest
+}) => {
     const sourceValue = get(record, source);
+    const classes = useStyles({});
```
djhi: Allow classes override

packages/ra-ui-materialui/src/field/ImageField.tsx
```diff
@@ -23,10 +23,16 @@ interface Props extends FieldProps {
...
+export const ImageField: SFC<Props & InjectedFieldProps> = ({
+    className,
+    record,
+    source,
+    src,
+    title,
+    ...rest
+}) => {
     const sourceValue = get(record, source);
+    const classes = useStyles({});
```
djhi: Allow classes override

packages/ra-ui-materialui/src/field/FileField.spec.js
```diff
@@ -74,6 +74,7 @@ describe('<FileField />', () => {
                 {...defaultProps}
                 record={{
                     url: 'http://foo.com/bar.jpg',
+                    title: 'Hello world!',
                 }}
                 target="_blank"
```
djhi: You need to add the title source `title="title"`

fzaninotto:
Sorry, but I had to force push to `next` to avoid a complicated rebase. As a consequence, the commits from this PR are no longer in `next`. Would you mind opening a new PR with the same commits?

djhi:
As this is my fault (I shouldn't have merged this PR), I'll take care of it @Kunnu01

PR #3571, opened 2019-08-22
[RFR] Remove deprecated Link component

jaytula:
- Remove deprecated Link component
- Remove `WithStyles` wraps  `SingleFieldList.spec.js`
- Update two components to import from react-router-dom for `Link`: `ReferenceField` and `SingleFieldList`


jaytula:
Sure, will do.

packages/ra-ui-materialui/src/index.js
```diff
@@ -1,4 +1,4 @@
-import Link from './Link';
+import { Link } from 'react-router-dom';
```
djhi: I think we don't need to re-export the Link component. Ping @fzaninotto

fzaninotto: if it makes the upgrade easier... Just mark it as deprecated down where you export it.

fzaninotto:
I'm the one who deprecated Link (in #2050), but I'm afraid it breaks link styling. I'll have to checkout your branch and test.

fzaninotto:
Indeed this breaks styling of ReferenceFields. It should not have been deprecated in the first place (my bad). So I'm closing this PR and removing the deprecated mention. But thanks a lot for your contribution anyway!

PR #3572, opened 2019-08-22
[RFR] List perf improvements

fzaninotto:
Various list perf improvements following a profiling sessions.

After an initial render of the datagrid from store, avoids a rerendering of each datagrid row once an unchanged response comes from the dataprovider.

Also, I believe we'll have to support Component injection in some places to avoid forced renders due to the fact that *elements* are always different.

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetMany.ts
+++ b/packages/ra-core/src/dataProvider/useGetMany.ts
@@ -81,8 +81,10 @@ const useGetMany = (resource: string, ids: Identifier[], options: any = {}) => {
     const [state, setState] = useSafeSetState({
         data,
         error: null,
-        loading: true,
-        loaded: data.length !== 0 && !data.includes(undefined),
+        loading: ids.length !== 0,
+        loaded:
+            ids.length === 0 ||
+            (data.length !== 0 && !data.includes(undefined)),
     });
     if (!isEqual(state.data, data)) {
         setState({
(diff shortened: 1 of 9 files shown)


packages/ra-ui-materialui/src/list/DatagridRow.js
```diff
@@ -107,7 +126,6 @@ const DatagridRow = ({
                         className={classes.expandIconCell}
                     >
                         <ExpandRowButton
-                            className={classes.expandButton}
```
fzaninotto: this was causing the expand button to be always shown expanded (recent regression)

packages/ra-ui-materialui/src/list/DatagridRow.js
```diff
@@ -145,12 +163,19 @@ const DatagridRow = ({
             {expand && expanded && (
                 <TableRow key={`${id}-expand`} id={`${id}-expand`}>
                     <TableCell colSpan={nbColumns}>
-                        {cloneElement(expand, {
-                            record,
-                            basePath,
-                            resource,
-                            id: String(id),
-                        })}
+                        {isValidElement(expand)
```
fzaninotto: this is a temporary API. I'm not sure yet whether we should support components and elements everywhere, we'll see if profiling forces us to do that often.

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -81,8 +82,10 @@ const useGetMany = (resource: string, ids: Identifier[], options: any = {}) => {
     const [state, setState] = useSafeSetState({
         data,
         error: null,
-        loading: true,
-        loaded: data.length !== 0 && !data.includes(undefined),
+        loading: ids.length !== 0,
+        loaded:
+            ids.length === 0 ||
```
fzaninotto: avoids showing the loader if there is nothing to load

PR #3573, opened 2019-08-22
[RFR] Create/Edit allow changing success message

natrim:
Allow~ changing success message in Create/Edit

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -180,10 +180,7 @@ const Aside = ({ record }) => (
 
 ### Success message
 
-Once the `dataProvider` returns successfully after save, users see a generic notification ("Element creatd" / "Element updated"). You can customize this message by passing a `successMessage` prop.
-You can use this prop to change the success message.
-
-**Tip**: The message will be translated.
+Once the `dataProvider` returns successfully after save, users see a generic notification ("Element created" / "Element updated"). You can customize this message by passing a `successMessage` prop:
 
 ```jsx
 const PostEdit = props => (
@@ -191,6 +188,8 @@ const PostEdit = props => (
         ...
     </Edit>
 ```
+
+**Tip**: The message will be translated.
 
 ### Undoable
 
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -20,6 +20,7 @@ Here are all the props accepted by the `<Create>` and `<Edit>` components:
 * [`title`](#page-title)
 * [`actions`](#actions)
 * [`aside`](#aside-component)
+* [`successMessage`](#success-message)
 * [`undoable`](#undoable) (`<Edit>` only)
 
 Here is the minimal code necessary to display a form to create and edit comments:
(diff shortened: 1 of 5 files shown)


docs/CreateEdit.md
```diff
@@ -177,6 +178,20 @@ const Aside = ({ record }) => (
 
 **Tip**: Always test that the `record` is defined before using it, as react-admin starts rendering the UI before the API call is over.
 
+### Message
```
fzaninotto: the name of the prop is too generic. `successMessage` would be better.

docs/CreateEdit.md
```diff
@@ -177,6 +178,20 @@ const Aside = ({ record }) => (
...
+
+By default, on save you will get notification with generic message about change done.
```
fzaninotto:
```suggestion
Once the `dataProvider` returns successfully after save, users see a generic notification ("Element creatd" / "Element updated"). You can customize this message by passing a `successMessage` prop.
```

packages/ra-core/src/controller/useCreateController.ts
```diff
@@ -90,9 +92,13 @@ const useCreateController = (props: CreateProps): CreateControllerProps => {
                     onSuccess: onSuccess
                         ? onSuccess
                         : ({ data: newRecord }) => {
-                              notify('ra.notification.created', 'info', {
-                                  smart_count: 1,
-                              });
+                              notify(
+                                  message ? message : 'ra.notification.created',
```
fzaninotto:
```suggestion
                                  message || 'ra.notification.created',
```

packages/ra-core/src/controller/useEditController.ts
```diff
@@ -55,16 +56,21 @@ export interface EditControllerProps {
...
     const translate = useTranslate();
     const notify = useNotify();
     const redirect = useRedirect();
     const refresh = useRefresh();
     const version = useVersion();
     const { data: record, loading } = useGetOne(resource, id, {
         version, // used to force reload
-        onFailure: () => {
-            notify('ra.notification.item_doesnt_exist', 'warning');
+        onFailure: error => {
```
fzaninotto: I don't agree with this change. All around the application, if the `GET_ONE` call fails, we consider that it's a 404. If you change it here, it should be changed everywhere else. But I don't want to do that, so please revert it.

packages/ra-core/src/controller/useEditController.ts
```diff
@@ -95,7 +101,7 @@ const useEditController = (props: EditProps): EditControllerProps => {
                 {
                     onSuccess: () => {
                         notify(
-                            'ra.notification.updated',
+                            message ? message : 'ra.notification.updated',
```
fzaninotto:
```suggestion
                            message || 'ra.notification.updated',
```

packages/ra-core/src/controller/useShowController.ts
```diff
@@ -55,8 +55,13 @@ const useShowController = (props: ShowProps): ShowControllerProps => {
     const version = useVersion();
     const { data: record, loading } = useGetOne(resource, id, {
         version, // used to force reload
-        onFailure: () => {
-            notify('ra.notification.item_doesnt_exist', 'warning');
+        onFailure: error => {
```
fzaninotto: same, please revert

natrim:
updated

fzaninotto:
Thanks!

PR #3576, opened 2019-08-22
[RFR] Migrate SelectArrayInput to hooks

djhi:


packages/ra-ui-materialui/src/input/SelectArrayInput.js
```diff
@@ -122,149 +122,149 @@ const styles = theme =>
...
-        const { optionText, translate, translateChoice } = this.props;
-        if (React.isValidElement(optionText)) {
-            return React.cloneElement(optionText, {
-                record: choice,
-            });
-        }
+    const handleChange = useCallback(
+        event => {
+            input.onChange(event.target.value);
+            // HACK: For some reason, redux-form does not consider this input touched without calling onBlur manually
```
fzaninotto: redux-form?

djhi: oups

djhi: actually not needed anymore

PR #3577, opened 2019-08-22
[RFR] Improve form performance

fzaninotto:
Our forms rerender all fields each time one of the values change - this is not good. I'm proposing a better default - having SimpleForm and TabbedForm subscribe only to submissions. 

- [x] improve perf
- [x] look for side effects


djhi:
I haven't found any side effects. Inspection using the react profiler does not show any performance issues anymore

PR #3580, opened 2019-08-22
[RFR] Convert SingleFieldList component to useStyles

jaytula:
- Convert SingleFieldList component to useStyles
- Remove `WithStyles` from `WithStyles(SingleFieldList)`
- Use default import for `SingleFieldList` in `SingleFieldList.spec.js`

### DIFF
--- a/packages/ra-ui-materialui/src/list/SingleFieldList.js
+++ b/packages/ra-ui-materialui/src/list/SingleFieldList.js
@@ -1,13 +1,13 @@
-import React, { cloneElement, Component, Children } from 'react';
+import React, { cloneElement, Children } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/SingleFieldList.js
```diff
@@ -54,68 +54,67 @@ const sanitizeRestProps = ({
...
-            ids,
-            data,
-            loaded,
-            resource,
-            basePath,
-            children,
-            linkType,
-            ...rest
-        } = this.props;
+    const handleClick = () => {};
```
fzaninotto: move this function out of the component to avoid passing a new function at every render

jaytula: done

fzaninotto:
Thanks!

PR #3582, opened 2019-08-22
[RFR] Convert AppBar component to useStyles

jaytula:
Convert AppBar component to useStyles


jaytula:
Sure, will give it a go

PR #3586, opened 2019-08-23
[RFR] remove deprecated LongTextInput

Kunnu01:
- [x] Remove `LongTextInput`
- [x] Update `TextInput`
- [x] Update Docs
- [x] Upgrade guide


packages/ra-ui-materialui/src/input/TextInput.tsx
```diff
@@ -72,6 +73,8 @@ export const TextInput: FunctionComponent<
                     helperText={helperText}
                 />
             }
+            multiline={multiline}
```
djhi: Please revert all changes in this file. We don't want to replicate all props of the underlying TextField compnent

Kunnu01: Done

djhi:
Thanks!

PR #3588, opened 2019-08-23
Remove comment on withStyles in LinearProgress

jaytula:
Removed leftover reference to `withStyles` in LinearProgress component


djhi:
Thanks!

PR #3589, opened 2019-08-23
[RFR] Convert DragPreview component to useStyles

jaytula:
This is a conversion to `useStyles` but it is the first that I've done that has life-cycle method `shouldComponentUpdate` that just returns `false`.

I did not try to add any memoization because I wouldn't really be understanding whether it worked. Would very much appreciate an explanation how `shouldComponentUpdate` for this seemingly trivial case should be handled.

Thanks!


djhi:
You can wrap it inside a `React.memo`: https://reactjs.org/docs/react-api.html#reactmemo

jaytula:
Thanks!  Would something like this be the right idea for this particular component?

```javascript
export default React.memo(DragPreview, () => true);
```

PR #3591, opened 2019-08-23
[RFR] Convert DragLayer component to useStyles

jaytula:
Convert DragLayer component to useStyles #3453


jaytula:
I tried dragging the tags around and appears to be working.  It's a little tricky to know whether it was dropped into or out of an item or re-ordered.

Also temporarily set a backgroundColor on the `DragLayer` component (`clases.layer`) and tested dragging to know that the layer was activated when dragged.

djhi:
Thanks!

PR #3594, opened 2019-08-23
[RFR] Improve form look and feel

fzaninotto:
Material design no longer advocates the usage of fields denoted by a simple underline. Only the "filled" and "outline" variants are documented in [the material design guide](https://material.io/components/text-fields/).

The major bump in react-admin v3 is the right occasion to reflect that change. But we should let an escape hatch for users who absolutely don't want filled inputs. That is possible with the material ui theme, and we should make it possible to override the variant at the form level.

Also, I want to make forms more compact, so we use a `dense` margin by default.

| Before | After |
|----------|---------|
| image | image |

- [x] Provide a default theme with less ugly filled variant than mui default light
- [x] Provide a default theme with less ugly filled variant than mui default dark
- [x] Allow users to override `margin` and `variant` at the form level
- [x] Make common inputs dense and filled by default
- [x] Make other inputs dense and filled by default
- [x] Update Form documentation to mention the new props 
- [x] Add upgrade guide

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -363,6 +363,8 @@ Here are all the props accepted by the `<SimpleForm>` component:
 * [`submitOnEnter`](#submit-on-enter)
 * [`redirect`](#redirection-after-submission)
 * [`toolbar`](#toolbar)
+* [`variant`](#variant)
+* [`margin`](#margin)
 * `save`: The function invoked when the form is submitted. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
 * `saving`: A boolean indicating whether a save operation is ongoing. This is passed automatically by `react-admin` when the form component is used inside `Create` and `Edit` components.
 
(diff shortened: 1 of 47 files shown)


fzaninotto:
Switching to RFR. The failling tests also fail on `next`.

docs/CreateEdit.md
```diff
@@ -827,6 +832,34 @@ Here are the props received by the `Toolbar` component when passed as the `toolb
...
+        <SimpleForm variant="standard">
+            ...
+        </SimpleForm>
+    </Edit>
+);
+```
+
+## Margin
+
+By default, react-admin input components use the Material Design "dense" mrgin. If you want to use the "normal" or "none" margins, you can either set the `margin` prop on each Input component individually, or set the `margin` prop directly on the Form component. In that case, the Form component will transmit the `margin` to each Input.
```
djhi:
```suggestion
By default, react-admin input components use the Material Design "dense" margin. If you want to use the "normal" or "none" margins, you can either set the `margin` prop on each Input component individually, or set the `margin` prop directly on the Form component. In that case, the Form component will transmit the `margin` to each Input.
```

fzaninotto:
Thanks for searching and fixing that hard bug!

PR #3598, opened 2019-08-25
[RFR] Convert PaginationActions to hooks

jaytula:
Convert `PaginationActions` to hooks.

This was mostly straightforward conversion done step-by-step (running `make test-unit` at every step). One note though...  I removed the HOC `recompose/pure` along with the non-default export. Using the pure-wrapped component does not pass the `PaginationActions.spec`.   Let me know if I got this wrong.

### DIFF
--- a/packages/ra-ui-materialui/src/list/PaginationActions.js
+++ b/packages/ra-ui-materialui/src/list/PaginationActions.js
@@ -18,12 +18,12 @@ const styles = theme =>
         hellip: { padding: '1.2em' },
     });
 
-export class PaginationActions extends Component {
+export function PaginationActions(props) {
     /**
      * Warning: material-ui's page is 0-based
      */
-    range() {
-        const { page, rowsPerPage, count } = this.props;
+    const range = () => {
+        const { page, rowsPerPage, count } = props;
         const nbPages = Math.ceil(count / rowsPerPage) || 1;
         if (isNaN(page) || nbPages === 1) {
             return [];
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/PaginationActions.js
```diff
@@ -163,16 +154,11 @@ export class PaginationActions extends Component {
...
 };
 
-const enhance = compose(
-    pure,
-    translate,
-    withStyles(styles)
-);
-
-export default enhance(PaginationActions);
+export default PaginationActions;
```
djhi: You should wrap this component with `React.memo` (basically the same as `pure`)

packages/ra-ui-materialui/src/list/PaginationActions.js
```diff
@@ -163,16 +154,11 @@ export class PaginationActions extends Component {
...
 };
 
-const enhance = compose(
-    pure,
-    translate,
-    withStyles(styles)
-);
-
-export default enhance(PaginationActions);
+export default React.memo(PaginationActions, () => true);
```
djhi: You're preventing this component to rerender when its props change. Please don't override the isEqual function (that's what `pure` does).

packages/ra-ui-materialui/src/list/PaginationActions.js
```diff
@@ -1,29 +1,33 @@
...
 import PropTypes from 'prop-types';
-import pure from 'recompose/pure';
 import Button from '@material-ui/core/Button';
-import { withStyles, createStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/core/styles';
 import ChevronLeft from '@material-ui/icons/ChevronLeft';
 import ChevronRight from '@material-ui/icons/ChevronRight';
-import compose from 'recompose/compose';
-import { translate } from 'ra-core';
+import { useTranslate } from 'react-admin';
```
natrim: should be ra-core

Issue #2983, opened 2019-03-10
[ra-data-graphql-simple] Http 401 error is not redirecting to login page


ballad89:
**What you were expecting:**

I expected to be redirected to the login page when the graphql endpoint returns http 401 or 403

**What happened instead:**

It displays the error message, but continues and tries to parse the null data returned

**Steps to reproduce:**

- Have the graphql endpoint return 401 for a query, with an error message and null data

**Related code:**

**Other information:**

**Environment**

* React-admin version: "^2.7.3"
"ra-data-graphql-simple": "^2.7.1",
* React version: "^16.8.4"
* Browser: Google Chrome
* Stack trace (in case of a JS error):
Below is what gets printed to console.log
```js
index.js:83 POST http://localhost:3001/graphql 401 (Unauthorized)
index.js:1446 Warning: Missing translation for key: "GraphQL error: LOGIN_USER
GraphQL error: LOGIN_USER
GraphQL error: Non-nullable field was null."
index.js:1446 Warning: Missing translation for key: "GraphQL error: LOGIN_USER
GraphQL error: LOGIN_USER
GraphQL error: Non-nullable field was null."
```

fzaninotto:
As explained in the documentation, it's your responsibility to handle 401 or 403 errors in the authProvider: https://marmelab.com/react-admin/Authentication.html#catching-authentication-errors-on-the-api

This is not related to the data provider.

ballad89:
Maybe I did not explain clearly, I am saying the default behaviour explained below does not work for graphql
![image](https://user-images.githubusercontent.com/4387126/54317122-01b85700-45eb-11e9-9ab1-ebacdf347ffd.png)

fzaninotto:
You're right, the Authentication documentation is misleading. I'm reopening this issue and marking it as a documentation bug.

ballad89:
the reason I felt it is a bug is because the graphql package does not catch and throw the correct error as the other packages do. It assumes it always succeeds
https://github.com/marmelab/react-admin/blob/master/packages/ra-data-graphql/src/index.js#L100

fzaninotto:
I mistakenly pushed the fix to master. Reverting and reopening the issue.

ballad89:
Are you referring to the documentation fix or my immediate comment above ?

fzaninotto:
ah, you're referring to the `fetchJson` utility and its control of the HTTP status?

[...]

PR #3604, opened 2019-08-26
[RFR] [ra-data-graphql] Parse GraphQL errors

melvynhills:
This PR adds proper error parsing for Apollo GraphQL so they can be used in an auth provider.

Addresses the issue raised here: https://github.com/marmelab/react-admin/issues/2983

Not 100% sure on which branch I should have opened the PR (master/next)

### DIFF
--- a/packages/ra-data-graphql/src/index.js
+++ b/packages/ra-data-graphql/src/index.js
@@ -9,6 +9,7 @@ import {
     CREATE,
     UPDATE,
     DELETE,
+    HttpError,
 } from 'ra-core';
 
 import buildApolloClient from './buildApolloClient';
@@ -77,6 +78,17 @@ export default async options => {
 
     const buildQuery = buildQueryFactory(introspectionResults, otherOptions);
 
+    const parseError = ({ networkError, graphQLErrors }) => {
+        const message =
+            get(graphQLErrors, '[0].message') ||
+            get(networkError, 'result.errors[0].message') ||
+            get(networkError, 'message');
+        const status =
+            get(graphQLErrors, '[0].statusCode') ||
+            get(networkError, 'statusCode');
+        throw new HttpError(message, status);
+    };
+
     const raDataProvider = (aorFetchType, resource, params) => {
         const overriddenBuildQuery = get(
             override,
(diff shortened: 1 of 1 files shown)


packages/ra-data-graphql/src/index.js
```diff
@@ -77,6 +78,17 @@ export default async options => {
 
     const buildQuery = buildQueryFactory(introspectionResults, otherOptions);
 
+    const parseError = ({ networkError, graphQLErrors }) => {
+        const message =
+            get(graphQLErrors, '[0].message') ||
+            get(networkError, 'result.errors[0].message') ||
+            get(networkError, 'message');
+        const status =
```
djhi: Isn't this part highly dependent on the server implementation ? I'm not sure we should handle errors at all as they are many possible implementations for errors. Wouldn't it be better to add a section about error handling in the documentation?

djhi:
> Not 100% sure on which branch I should have opened the PR (master/next)

`next` branch is perfect :+1:

packages/ra-data-graphql/src/index.js
melvynhills: Oh my, just realised the `params` attribute is actually the error object in the `AUTH_ERROR` case. My bad! So we could specify that in the docs but it looks obvious now haha.

Regarding your comment, I believe this is the standard Apollo error formatting, i.e. all errors behave that way regardless of server implementations. The goal here was to reformat the error object so it would work like the other data providers.

I guess I can close the PR as all this code can be used within the auth provider implementation.

djhi: > Oh my, just realised the params attribute is actually the error object in the AUTH_ERROR case. My bad! So we could specify that in the docs but it looks obvious now haha.

A documentation PR to clarify this point would be greatly appreciated. It is not so obvious in our current documentation ;)

> all errors behave that way regardless of server implementations

Maybe true for the `graphQLErrors`. Not so sure about the `networkError` though.

melvynhills:
Now we're here, I just want to say thank you guys @marmelab for this framework. I've rarely seen such quality of code architecture, documentation and support. Keep it up!

macrozone:
@melvynhills has this been released? do you know anything?

macrozone:
@melvynhills aaah, this has not been merged, maybe reopen it?

PR #3605, opened 2019-08-26
[RFR] Ensure Query, Mutation and withDataProvider are backward compatible

### LINKED ISSUE #3593: Declarative side effects are no longer supported in `withDataProvider` and `Mutation`
in `next`, custom queries now accept side effect *functions* rather than side effect *objects*. that's fine in the new hooks (`useQuery`, `useGetList`, etc), but that's a BC break in the existing query utils (`withDataProvider`, `Query`, `Mutation`). To ease the transition to v3, these components should accept object side effects and transform them to function side effects before calling `useDataProvider`.

djhi:
Fixes #3593

- [x] Introduced `useDataProviderWithDeclarativeSideEffects` (might need a better name)
- [x] Tests


packages/ra-core/src/dataProvider/useDataProviderWithDeclarativeSideEffects.ts
```diff
@@ -0,0 +1,73 @@
...
+    const unselectAll = useUnselectAll();
+
+    return useCallback(
+        (type: string, resource: string, params: any, options: any = {}) => {
+            const convertToFunctionSideEffect = sideEffects => {
+                if (
+                    typeof sideEffects === 'undefined' ||
+                    typeof sideEffects === 'function'
+                ) {
+                    return sideEffects;
```
Luwangel: You can test `null` too. But it's not possible using `typeof`, because `null` has type 'object'`.

packages/ra-core/src/dataProvider/useDataProviderWithDeclarativeSideEffects.ts
```diff
@@ -0,0 +1,73 @@
...
+const useDataProviderWithDeclarativeSideEffects = (): DataProviderHookFunction => {
+    const dataProvider = useDataProvider();
+    const notify = useNotify();
+    const redirect = useRedirect();
+    const refresh = useRefresh();
+    const unselectAll = useUnselectAll();
+
+    return useCallback(
+        (type: string, resource: string, params: any, options: any = {}) => {
+            const convertToFunctionSideEffect = sideEffects => {
```
Luwangel: I think you can move this method at the top of the file or in another file, and write unit tests.

fzaninotto: No he can't, because the scope of the function must contain the side effects grabbed from hooks just above

packages/ra-core/src/dataProvider/useDataProviderWithDeclarativeSideEffects.ts
```diff
@@ -0,0 +1,73 @@
+import {
+    useNotify,
+    useRedirect,
+    useRefresh,
+    useUnselectAll,
+} from '../sideEffect';
+import useDataProvider, { DataProviderHookFunction } from './useDataProvider';
+import { useCallback } from 'react';
+
+const useDataProviderWithDeclarativeSideEffects = (): DataProviderHookFunction => {
```
fzaninotto: I'd add a jsDoc block explaining that this is for backwards compatibility (to be removed in a future major version)

packages/ra-core/src/dataProvider/useMutation.ts
```diff
@@ -2,7 +2,7 @@ import { useCallback } from 'react';
 import merge from 'lodash/merge';
 
 import { useSafeSetState } from '../util/hooks';
-import useDataProvider from './useDataProvider';
+import useDataProvider from './useDataProviderWithDeclarativeSideEffects';
```
fzaninotto: I don't want that `useMutation` accept declarative side effects. It's a new hook, we can enforce its syntax. I just want it for `<Mutation>`.

packages/ra-core/src/dataProvider/useQuery.ts
```diff
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 
 import { useSafeSetState } from '../util/hooks';
-import useDataProvider from './useDataProvider';
+import useDataProvider from './useDataProviderWithDeclarativeSideEffects';
```
fzaninotto: I don't want that `useQuery` accept declarative side effects. It's a new hook, we can enforce its syntax. I just want it for `<Query>`.

packages/ra-core/src/sideEffect/useUnselectAll.ts
```diff
@@ -10,10 +10,17 @@ import { setListSelectedIds } from '../actions';
...
  * unselectAll();
  */
-const useUnselectAll = resource1 => {
+const useUnselectAll = (resource1?: string) => {
     const dispatch = useDispatch();
     return useCallback(
-        resource2 => dispatch(setListSelectedIds(resource2 || resource1, [])),
+        (resource2?: string) => {
+            if (!resource2 && !resource1) {
+                console.log(
```
fzaninotto: :+1: I was bitten by this, too

packages/ra-core/src/dataProvider/useDataProviderWithDeclarativeSideEffects.ts
Luwangel: It's not possible to pass them as params?

packages/ra-core/src/dataProvider/useDataProviderWithDeclarativeSideEffects.ts
djhi: Reviewed

packages/ra-core/src/dataProvider/useMutation.spec.tsx
```diff
@@ -128,4 +134,236 @@ describe('useMutation', () => {
         await waitForDomChange({ container: testElement });
         expect(testElement.textContent).toEqual('provider error');
     });
+
+    it('supports declarative onSuccess side effects', async () => {
```
fzaninotto: Technically, you're testing `Mutation` and not `useMutation` here. I'd move the tests to avoid confusion.

packages/ra-core/src/dataProvider/useMutation.ts
```diff
@@ -95,9 +97,37 @@ const useMutation = (
...
     });
     const dataProvider = useDataProvider();
+    const dataProviderWithDeclarativeSideEffects = useDataProviderWithDeclarativeSideEffects();
+
     const mutate = useCallback(
         (event, callTimePayload = {}, callTimeOptions = {}): void => {
             setState({ loading: true });
+
+            if (options.withDeclarativeSideEffectsSupport) {
+                dataProviderWithDeclarativeSideEffects(
```
fzaninotto: instead of pasting the code, I'd create a variable function and call it once:

```js
const dataProviderWithSideEffects = options.withDeclarativeSideEffectsSupport ? dataProviderWithDeclarativeSideEffects : dataProvider;
```

packages/ra-core/src/dataProvider/useQuery.ts
```diff
@@ -94,7 +96,34 @@ const useQuery = (
         loaded: false,
     });
     const dataProvider = useDataProvider();
+    const dataProviderWithDeclarativeSideEffects = useDataProviderWithDeclarativeSideEffects();
+
     useEffect(() => {
+        if (options.withDeclarativeSideEffectsSupport) {
```
fzaninotto: same as for `useMutation`

packages/ra-core/src/sideEffect/useUnselectAll.ts
fzaninotto: maybe you should use the `warning` from `ra-core/util/warning` to hide the warning in production.

packages/ra-core/src/dataProvider/useMutation.spec.tsx
djhi: The whole file is testing `Mutation`. Should I rename it ?

PR #3613, opened 2019-08-28
[RFR] Export Form Helpers

djhi:
- [x] Export `useInitializeFormWithRecord` for usage in custom form implementations
- [x] Fix it to ensure the form remains pristine after loading the record

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.js
@@ -4,12 +4,11 @@ import { Form } from 'react-final-form';
 import arrayMutators from 'final-form-arrays';
 import { useSelector } from 'react-redux';
 import classnames from 'classnames';
-import { useTranslate } from 'ra-core';
+import { useTranslate, useInitializeFormWithRecord } from 'ra-core';
 
 import FormInput from './FormInput';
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/form/SimpleForm.js
```diff
@@ -4,12 +4,11 @@ import { Form } from 'react-final-form';
 import arrayMutators from 'final-form-arrays';
 import { useSelector } from 'react-redux';
 import classnames from 'classnames';
-import { useTranslate } from 'ra-core';
+import { useTranslate, useInitializeFormWithRecord } from 'ra-core';
 
 import FormInput from './FormInput';
 import Toolbar from './Toolbar';
 import CardContentInner from '../layout/CardContentInner';
-import useInitializeFormWithRecord from './useInitializeFormWithRecord';
```
fzaninotto: if it's not used anymore, shouldn't you remove the file?

djhi: The file has been moved to `ra-core` and is imported at L7

djhi: See previous change in github diff

PR #3621, opened 2019-08-29
Add hasBulkActions to useMemo inputs

Evgeny81:
If hasBulkActions prop changed dynamically, DataGrid renders Checkbox only in the DataGridHeader and doesn't render Checkbox in every row because useMemo ignores this prop change.


fzaninotto:
Thanks!

PR #3625, opened 2019-08-30
[RFR] CoreAdminRouter - log initializeResources errors

natrim:
sometimes when you make error in authprovider, the app just goes into cycle of login and logout

and you don't even know why

so its better to report the error to console, so you can fix the problem


fzaninotto:
Thanks!

Issue #3629, opened 2019-08-30
[alpha.3] problems with permissions in List


natrim:
aka using example from docs to hide buttons in grid

Seems like in v3 is some problem with permissions or rerendering

Showcase from the simple example: 
https://codesandbox.io/s/simple-7zyys

![Aug-30-2019 15-00-16](https://user-images.githubusercontent.com/488136/64022790-3c7d5f80-cb37-11e9-95f1-eb0b6a239a87.gif)

as you can see the buttons are showed only for admin, but if you go away from page, and return, the button's are now gone

Steps to reproduce:
1) open example, you will se buttons if you are admin
2) go to another page
3) go back, you will now see the buttons are gone even if you are admin

djhi:
Thanks for reporting and the reproduction sandbox!

djhi:
I think this is actually the same issue as #3631

fzaninotto:
Fixed by #3635

PR #3635, opened 2019-09-02
[RFR] Fix Datagrid can't handle dynamic children

### LINKED ISSUE #3631: [alpha.3] Dynamical DataGrid column changing not followed with List rerendering
**What you were expecting:**
I expect to be able to dynamically configure columns set in DataGrid

**What happened instead:**
Only header row changes its configuration, on column config change without page reload.

**Steps to reproduce:**

- Go to codesandbox
- Open Posts view
- Click first  `Toggle Columns` button
- Consider only header row config changed

**Related code:**
https://codesandbox.io/s/simple-ju5rm

Consider `Toggle Columns` button behavior - when columns configuration depends on state, when config in state changes - only header row updates configuration. 
If config stored in the cache - columns are rendered correctly on page reloading.

**Other information:**
This reproduced only on 3.0 alpha version.
Quite similar behavior observed with `ra-customizable-datagrid` module, even if this module rerenders all selected DataGrid children under the hood, instead of conditional rendering one-by-one. Previously it worked fine.
If there are some backward incompatible changes in DataGrid, I'll be glad for some guidance about them, to find a way to fix external package.

**Environment**

* React-admin version: "^3.0.0-alpha.3",
* Last version that did not exhibit the issue (if applicable): 2.9.x
* React version: ~16.8.0
* Browser: Chrome

djhi:
Fixes #3631 and #3629

### DIFF
--- a/packages/ra-ui-materialui/src/list/DatagridBody.js
+++ b/packages/ra-ui-materialui/src/list/DatagridBody.js
@@ -91,10 +91,10 @@ const areEqual = (prevProps, nextProps) => {
     return isEqual(prevPropsWithoutChildren, nextPropsWithoutChildren);
 };
 
-export const MemoDatagridBody = memo(DatagridBody, areEqual);
+export const PureDatagridBody = memo(DatagridBody, areEqual);
 // trick material-ui Table into thinking this is one of the child type it supports
-MemoDatagridBody.muiName = 'TableBody';
-MemoDatagridBody.defaultProps = {
+PureDatagridBody.muiName = 'TableBody';
+PureDatagridBody.defaultProps = {
     row: <PureDatagridRow />,
 };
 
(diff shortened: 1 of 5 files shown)


natrim:
confirming it fixes #3629

djhi:
@fzaninotto I just tried this branch again with a CPU throttle set to x6 and I haven't notice any performance issues nor differences with the version without this fix. Can you try again ?

packages/ra-ui-materialui/src/list/DatagridBody.js
```diff
@@ -91,7 +82,11 @@ DatagridBody.defaultProps = {
     row: <DatagridRow />,
 };
 
+const MemoDatagridBody = memo(DatagridBody);
```
fzaninotto: memo doesn't work here as `children` will always change

fzaninotto:
I think the default export should be DatagridBody, not MemoDatagridBody

PR #3637, opened 2019-09-02
[RFR][alpha.4] fix bad import in PaginationActions

natrim:
should be ra-core


natrim:
it's because i am not importing react-admin but ra-core and ra-ui-materialui directly and so my build will fail if something like this happens

PR #3640, opened 2019-09-02
[RFR] [BC Break] Migrate export button

### LINKED ISSUE #3623: Rewrite ExportButton using useGetMany
The ExportButton component still dispatches crud actions by hand. Switching to hooks may change the signature, so we should rewrite it before releasing the v3 beta.

fzaninotto:
As expected in #3623, using the new hooks changes the signature of the `exporter` function, but for the best.

```diff
-const exportComments = (data, fetchRelaterRecords, dispatch) => {
+const exportComments = (data, fetchRelaterRecords, dataProvider) => {
    // ...
}
```

This is a small BC break, because using the `dispatch` parameter in the exporter function was really cumbersome, and I doubt anyone has ever used it. 

Also, people wishing to customize the export more deeply (with custom side effects, etc) can easily write their own. Here is the simplified version of the `ExportButton` component:

```jsx
import {
    downloadCSV,
    useDataProvider,
    useNotify,
    GET_LIST,
} from 'react-admin';
import jsonExport from 'jsonexport/dist';

const ExportButton = ({ sort, filter, maxResults = 1000, resource }) => {
    const notify = useNotify();
    const payload = { sort, filter, pagination: { page: 1, perPage: maxResults }}
    const handleClick = dataProvider(GET_LIST, resource, payload)
        .then(({ data }) => jsonExport(data, (err, csv) => downloadCSV(csv, resource)))
        .catch(error => notify('ra.notification.http_error', 'warning'));

    return (
        <Button
            label="Export"
            onClick={handleClick}
        />
    );
};
```

Closes #3623

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -906,3 +906,54 @@ There are two breaking changes in the new `<AutocompleteInput>`:
 -   highlightFirstSuggestion={true}
 />
 ```
+
+## The `exporter` function has changed signature
+
+In a `List`, you can pass a custom `exporter` function to control the data downloaded by users when they cllick on the "Export" button.
+
+```jsx
+const CommentList = props => (
+    <List {...props} exporter={exportComments}>
+        // ...
+    </List>
+)
+```
+
+In react-admin v3, you can styll pass an `exporter` function this way, but its signature has changed:
+
+```diff
+-const exportComments = (data, fetchRelaterRecords, dispatch) => {
++const exportComments = (data, fetchRelaterRecords, dataProvider) => {
+    // ...
+}
+```
+
+If you used `dispatch` to call the dataProvider using an action creator with a `callback` side effect, you will see that the v3 version makes your exporter code much simpler. If you used it to dispatch custom side effects (like notification or redirect), we recommend that you override the `<ExportButton>` component completely - it'll be much easier to maintain.
+
+As a base, here is the simplified `ExportButton` code:
+
+```jsx
+import {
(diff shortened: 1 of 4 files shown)


UPGRADE.md
```diff
@@ -906,3 +906,54 @@ There are two breaking changes in the new `<AutocompleteInput>`:
...
+
+```jsx
+const CommentList = props => (
+    <List {...props} exporter={exportComments}>
+        // ...
+    </List>
+)
+```
+
+In react-admin v3, you can styll pass an `exporter` function this way, but its signature has changed:
```
djhi:
```suggestion
In react-admin v3, you can still pass an `exporter` function this way, but its signature has changed:
```

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -0,0 +1,170 @@
...
+    filter?: any;
+    maxResults: number;
+    resource: string;
+    onClick?: (e: Event) => void;
+    label: string;
+    icon?: JSX.Element;
+    [key: string]: any;
+}
+
+const ExportButton: FunctionComponent<Props> = ({
```
djhi: As we pass the rest props to the Button, I would also add the MUI Button Props (even though we use our own button, it is based on the MUI one):

```ts
import { ButtonProps } from '@material-ui/core/Button';

const ExportButton: FunctionComponent<Props & ButtonProps> = ({
```

PR #3648, opened 2019-09-03
[RFR] Fix margins are too small in Show views 

### LINKED ISSUE #3641: Field margin is too small on ShowView
In v3, the ShowView and TabbedShowView layouts feels too packed. This is a regression due to the Form look and feel rework. 

image

fzaninotto:
Closes #3641


djhi:
A screenshot would have been nice

PR #3651, opened 2019-09-05
[RFR][v3] SimpleFormIterator - don't display error if its object

natrim:
object means it's error's for children so we can ignore it as children will display the errors by themself's

fixes error - passing object to react


fzaninotto:
Thanks!

PR #3653, opened 2019-09-05
Allowing for more fine tuning of Quill without rewriting RichTextInput

Sheraff:
In Quill, there are a bunch of things you can't really do with *just* the options passed to the constructor. I propose adding a tiny callback for those who need more custom behaviors.

For example, if you want to restrict the formatting possibilities offered **and** customize the behavior of some of these formats, you need to do it in two steps:

```javascript
const quill = new Quill($ref, {
    modules: { 
        toolbar: ['bold', 'italic', 'link']
    }
});
quill.getModule('toolbar').addHandler('bold', function(value) {
    this.quill.format('bold', value)
});
```

With this proposed *pull request*, you'd be able to implement these behaviors very simply:

```javascript
const quillInit = quill => quill.getModule('toolbar').addHandler('bold', function (value) {
    this.quill.format('bold', value)
});

// ...

<RichTextInput source="text" toolbar={[ ['bold', 'italic', 'link'] ]} quillInit={quillInit}/>
```

The alternative is to re-implement the entire <RichTextInput> component and its style... :-(

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -961,6 +961,18 @@ You can customize the rich text editor toolbar using the `toolbar` attribute, as
 <RichTextInput source="body" toolbar={[ ['bold', 'italic', 'underline', 'link'] ]} />
 ```
 
+If you need more customization, you can access the quill object through the `quillInit` callback that will be called just after its initialization.
+
+```js
+const quillInit = quill => quill.getModule('toolbar').addHandler('bold', function (value) {
+    this.quill.format('bold', value)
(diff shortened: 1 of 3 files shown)


Sheraff:
> I'm +1 with that change, but as it is a new feature, it has to be PRed against `next`. Also, you must document the new prop and add the exemple you write in the PR description in the react-admin documentation.

Thanks! Done here: https://github.com/marmelab/react-admin/pull/3714

Issue #2953, opened 2019-03-04
Login Page should redirect to the app if the user is already logged


dharmesh7980:
**Is your feature request related to a problem? Please describe.**
When the user go to the login page if they are already logged, the login form is still displayed.

**Describe the solution you'd like**
I would like the redirect the user to the dashboard by default if they are already logged.

**Additional context**
In order to reproduce on the demo, here is the process to follow:

1. First login in react-admin https://marmelab.com/react-admin-demo/#/login
2. After successful login again visit https://marmelab.com/react-admin-demo/#/login

User sees the login page, but the they would be redirected to the dashboard for example.

dharmesh7980:
First login in react-admin
https://marmelab.com/react-admin-demo/#/login

after successful login again visit https://marmelab.com/react-admin-demo/#/login , still user access login page it should not be happened,

Kmaschta:
I rephrased your whole issue, because it was really hard to understand. Let me know if it's what you asked for.

The next time, please follow the issue templates.

macscripter:
any update?

djhi:
Not sure we want that in the core and this is achievable in user land. I'll wait for more comments from the core team

fzaninotto:
I agree, this is achievable in userland using a `<Redirect>` in `<Login>`. Closing.

PR #3655, opened 2019-09-05
[RFR] [BC break] Reimplement auth logic using hooks

fzaninotto:
- [x] Reimplement auth logic using hooks (no Redux, no saga)
- [x] Use new hooks in Login, Logout, Resource, and CoreAdminRouter
- [x] In the Login page, redirect to home page if user is authenticated (refs #2953)
- [x] Use new hooks in examples
- [x] Update auth documentation
- [x] Handle auto logout when dataProvider returns a 403 (by using useAuthProvider inside useDataProvider)
- [x] Write upgrade guide
  - no more Redux action dispatched for auth calls
  - we systematically pass the `location` to the authProvider (rather than just the pathName for certain verbs)

```jsx
import { useLogin } from 'react-admin';

const LoginButton = () => {
    const [loading, setLoading] = useState(false);
    const login = useLogin();
    const handleClick = {
        setLoading(true);
        login({ username: 'john', password: 'p@ssw0rd' }, '/posts')
            .then(() => setLoading(false));
    }
    return <button onClick={handleClick}>Login</button>;
}

import { useLogout } from 'react-admin';

const LogoutButton = () => {
    const logout = useLogout();
    const handleClick = () => logout();
    return <button onClick={handleClick}>Logout</button>;
}

import { useCheckAuth } from 'react-admin';

const MyProtectedPage = () => {
    const checkAuth = useCheckAuth();
    useEffect(() => {
        checkAuth().catch(() => {});
    }, []);
    return <p>Private content: EZAEZEZAET</p>
} // tip: use useAuthenticated() hook instead

import { useCheckAuth } from 'react-admin';

const MyPage = () => {
    const checkAuth = useCheckAuth();
    const [authenticated, setAuthenticated] = useState(true); // optimistic auth
    useEffect(() => {
        checkAuth(undefined, false)
             .then() => setAuthenticated(true))
             .catch(() => setAuthenticated(false));
    }, []);
    return authenticated ? <Bar /> : <BarNotAuthenticated />;
} // tip: use useAuthState() hook instead
```


fzaninotto:
Switching to RFR!

Issue #3551, opened 2019-08-20
Form ignores empty string values


b-raines:
**What you were expecting:**
Submitting a form with an empty string value submits that value along with the form.

**What happened instead:**
A `TextInput` value == "" is ignored altogether when submitting the form.

**Steps to reproduce:**
Create a `SimpleForm` with a `TextInput` field. Edit the field, and then delete all text (if you leave it empty, it _will_ submit a null value). Submit the form and observe the params that are sent in the request. The empty field is not present at all.

**Related code:**
Will try to get a sandbox posted shortly w/ code.

**Environment**

* React-admin version: v3.0.0-alpha.2
* Last version that did not exhibit the issue (if applicable): 2.9.5
* React version: 16.9.0
* Browser: Chrome Version 76.0.3809.100

djhi:
Most probably not an issue in react-admin. See https://github.com/final-form/react-final-form/issues/430

djhi:
If we want this behavior, I'd rather go with form wide solution explained [here](https://github.com/final-form/react-final-form/issues/130#issuecomment-493447888) in the same issue (including the next two comments).

@fzaninotto What do you think?

fzaninotto:
I agree, we should handle that at the form level if possible

fzaninotto:
Fixed by #3758

berenddeboer:
This does not seem to have fixed the simple TextInput? If I empty a text input, it gets removed from the output, and I don't get an empty string. That's what I expected would happen with this fix.

Did I miss something?

fzaninotto:
@berenddeboer I don't reproduce your bug on master. When I empty a text input, the app sends an update message to the dataProvider, with a null value for the field I emptied. 

What's your react-admin version?

berenddeboer:
Not an empty string? Null values won't work for me as the field is required (but can be empty), but that's perhaps contradictory, I could change the schema to allow nulls. But I didn't see the field listed in the variables, but hadn't looked at the actual graphql fields it sends. Will look at that to double check.

My react-admin version is 3.1.2, I believe that was the latest until recently.

sanchezweezer:
Have same issue on 3.5.5 version

b-raines:
Also seeing this seems to have regressed.

b-raines:
Upon checking, it appears that the data is being sent to my dataProvider... nothing has changed there, so very curious as to why this is now behaving differently.

b-raines:
Was able to fix on my end, so disregard.

PR #3656, opened 2019-09-06
Submit empty form values

b-raines:
react-final-form does not submit values for inputs w/ undefined values

https://github.com/marmelab/react-admin/issues/3551

### DIFF
--- a/packages/ra-ui-materialui/src/form/checkForInitialValues.js
+++ b/packages/ra-ui-materialui/src/form/checkForInitialValues.js
@@ -1,0 +1,22 @@
+import merge from 'lodash/merge';
+
+const checkForInitialValues = (initialValues, newValues) => {
+    // library issue via https://github.com/final-form/react-final-form/issues/130#issuecomment-493447888
+    const emptiedData = Object.keys(initialValues).reduce((acc, key) => {
+        if (typeof newValues[key] === 'object' && newValues[key] !== null) {
+            acc[key] = checkForInitialValues(
+                initialValues[key],
+                newValues[key]
+            );
+        } else {
+            acc[key] =
+                typeof newValues[key] === 'undefined' ? '' : newValues[key];
+        }
+        return acc;
+    }, {});
+
+    // need to deep merge to get new child properties
+    return merge(emptiedData, newValues);
+};
+
+export default checkForInitialValues;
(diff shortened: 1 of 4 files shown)


natrim:
checkForInitialValues should be export from folder to allow custom imports from outside of package to create custom forms

packages/ra-ui-materialui/src/form/checkForInitialValues.js
```diff
@@ -0,0 +1,22 @@
+import merge from 'lodash/merge';
+
+const checkForInitialValues = (initialValues, newValues) => {
```
fzaninotto: by reading that method name, I don't understand what it's supposed to do or to return. Can you add jsDoc, example usage, or change the name?

packages/ra-ui-materialui/src/form/checkForInitialValues.js
```diff
@@ -0,0 +1,22 @@
...
+    // library issue via https://github.com/final-form/react-final-form/issues/130#issuecomment-493447888
```
fzaninotto: What does that comment mean?

packages/ra-ui-materialui/src/form/checkForInitialValues.js
```diff
@@ -0,0 +1,22 @@
...
+                newValues[key]
+            );
+        } else {
+            acc[key] =
+                typeof newValues[key] === 'undefined' ? '' : newValues[key];
+        }
+        return acc;
+    }, {});
+
+    // need to deep merge to get new child properties
```
fzaninotto: I don't understand what you're doing here

packages/ra-ui-materialui/src/form/checkForInitialValues.spec.js
```diff
@@ -0,0 +1,28 @@
...
+            age: 30,
+            email: null,
+            job: { title: undefined },
+        };
+        const finalValues = checkForInitialValues(initialValues, values);
+        expect(finalValues).toStrictEqual({
+            firstName: 'John',
+            lastName: '',
+            age: 30,
+            email: null,
```
fzaninotto: I don't understand the distinction between null and undefined here

djhi: I think `undefined` is when the field has never been set whereas `null` is used explicitly

djhi:
Superseded by #3758 for rebase and review

PR #3661, opened 2019-09-09
Add redux-devtools-extension trace feature in development mode if available.

### LINKED ISSUE #3658: [RFC] Add devToolsTrace prop to Admin to allow Redux DevTools Trace to be enable.
I've added a "devToolsTrace" prop in my fork in order to enable [Trace Feature](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Features/Trace.md)
of Redux DevTools.
I don't know if its the rigth way to go or what you think about it.
The code is based on [this post](https://github.com/zalmoxisus/redux-devtools-extension/issues/320#issuecomment-285639448) to avoid a bug if devtools is not installed.

I could make a [PR].

[packages/ra-core/src/createAdminStore.ts](https://github.com/WiXSL/react-admin/commit/b9419626df2e2a38655dbeeef577ebd1622484e3)

[packages/ra-core/src/CoreAdmin.tsx](https://github.com/WiXSL/react-admin/commit/382f3f627b71f86d20d2ced1440cf31070ecaf33)

WiXSL:
Closes #3658


fzaninotto:
Why not enable it by default? I don't think adding a new prop to `Admin` makes much sense for that feature.

PR #3665, opened 2019-09-10
[RFR][v3] NumberInput - step needs to be passed in inputProps

natrim:
the new version of material-ui in v3 does not pass all props to native input

so the step needs to be passed in inputProps to work

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
@@ -28,6 +28,19 @@ describe('<NumberInput />', () => {
         expect(input.getAttribute('type')).toEqual('number');
     });
 
+    describe('props', () => {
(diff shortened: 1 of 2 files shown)


fzaninotto:
[The mui documentation](https://material-ui.com/api/text-field/#textfield-api) says otherwise (TextInput passes extra props to FormControl, which passes extra props to the native control). So either there is a bug on their side, or their documentation is false. Can you please open an issue on their tracker?

Also, for such PRs, a unit test showing the bug is necessary.

natrim:
the documentation says: 

TextField: "Any other props supplied will be provided to the root element (FormControl)."
FormControl: "Any other props supplied will be provided to the root element (native element)."

and it does, the step is passed to native element, aka. root div of the formcontrol
but we want the step passed to "input"

Snímek obrazovky 2019-09-12 v 8 18 55

fzaninotto:
Thanks!

PR #3667, opened 2019-09-10
[WIP] New AutocompleteArrayInput

djhi:
- [x] Refactor `AutocompleteArrayInput` using `downshift`
- [x] Refactor common components used by `AutocompleteInput`
- [ ] Refactor (hooks, etc.) to avoid too much code duplication between `AutocompleteArrayInput` and `AutocompleteInput`
- [ ] Update documentation
- [ ] Update Upgrade guide


packages/ra-ui-materialui/src/input/AutocompleteInput.js
```diff
@@ -281,25 +281,32 @@ const AutocompleteInput = ({
...
+                                />
+                            }
                             InputProps={getInputProps({
                                 ...InputProps,
+                                inputRef: storeInputRef,
                                 id,
                                 name: input.name,
                                 onBlur: input.onBlur,
                                 onFocus: handleFocus(openMenu),
+                                onChange: event => {
```
fzaninotto: extract to a memoized function

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
```diff
@@ -0,0 +1,619 @@
...
+                onSubmit={jest.fn()}
+                render={() => (
+                    <TranslationContext.Provider
+                        value={{
+                            translate: x => `**${x}**`,
+                        }}
+                    >
+                        <AutocompleteArrayInput
+                            {...defaultProps}
+                            translate={x => `**${x}**`}
```
fzaninotto: this should not be there

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
```diff
@@ -0,0 +1,619 @@
...
+                            }}
+                        />
+                    )}
+                />
+            );
+
+            expect(input.value).toEqual('foo');
+        });
+
+        it('should revert the searchText when allowEmpty is false', () => {
```
fzaninotto: bad test title, allowempty does not exist for that component

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
```diff
@@ -0,0 +1,619 @@
...
+            ) as HTMLInputElement;
+
+            fireEvent.focus(input);
+            fireEvent.change(input, { target: { value: 'foo' } });
+            expect(queryAllByRole('option')).toHaveLength(0);
+            fireEvent.blur(input);
+            expect(input.value).toEqual('');
+        });
+
+        it('should show the suggestions when the input value is empty and the input is focussed and choices arrived late', () => {
```
fzaninotto:
```suggestion
        it('should show the suggestions when the input value is empty and the input is focused and choices arrived late', () => {
```

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
```diff
@@ -0,0 +1,619 @@
...
+                    onSubmit={jest.fn()}
+                    render={() => (
+                        <AutocompleteArrayInput
+                            {...defaultProps}
+                            meta={{ touched: true, error: 'Required' }}
+                        />
+                    )}
+                />
+            );
+            expect(queryByText('Required')).toBeDefined();
```
fzaninotto: I'm not at ease with `toBeDefined` testing the opposite of `toBeNull`. I don't want to look for low level JS semantics, so I suggest you use `not.ToBeNull` instead

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
```diff
@@ -0,0 +1,619 @@
...
+            expect(queryAllByRole('option')).toHaveLength(2);
+            fireEvent.blur(input);
+
+            fireEvent.focus(input);
+            fireEvent.change(input, { target: { value: 'ab' } });
+            expect(queryAllByRole('option')).toHaveLength(2);
+        });
+    });
+
+    it('does not automatically select a matched choice if there are more than one', () => {
```
fzaninotto: is it useful? It never automatically selects a matched choice, right?

packages/ra-ui-materialui/src/input/getSuggestions.ts
```diff
@@ -0,0 +1,104 @@
...
+    allowEmpty: boolean;
+    optionText: Function | string;
+    optionValue: string;
+    getSuggestionText: (choice: any) => string;
+    limitChoicesToValue: boolean;
+    suggestionLimit?: number;
+    selectedItem?: any | any[];
+}
+
+export default ({
```
fzaninotto: Could you add jsDoc and example usage?

packages/ra-ui-materialui/src/input/getSuggestions.ts
```diff
@@ -0,0 +1,104 @@
...
+    choices,
+    allowEmpty,
+    optionText,
+    getSuggestionText,
+    optionValue,
+    limitChoicesToValue,
+    selectedItem,
+    suggestionLimit = 0,
+}: Options) => filter => {
+    // This is the first display case when the input already has a value.
```
fzaninotto: what is a "display case"?

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
```diff
@@ -0,0 +1,430 @@
...
+            zIndex: 1,
+            marginTop: theme.spacing(1),
+            left: 0,
+            right: 0,
+        },
+        chip: {
+            margin: theme.spacing(0.5, 0.5, 0.5, 0),
+        },
+        chipContainerFilled: {
+            margin: '27px 12px 10px 0',
```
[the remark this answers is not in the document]
djhi: Actually, I borrowed this from mui filled inputs code. They don't use spacing for it. As you can see, they don't use multiple of `8` here which is the basic in `theme.spacing`. I used the same to ensure they align well with the input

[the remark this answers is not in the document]
djhi: See https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/FilledInput/FilledInput.js#L108

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
```diff
@@ -0,0 +1,430 @@
...
+                                                tabIndex={-1}
+                                                label={getSuggestionText(item)}
+                                                className={classes.chip}
+                                                onDelete={handleDelete(item)}
+                                            />
+                                        ))}
+                                    </div>
+                                ),
+                                onBlur,
+                                onChange: event => {
```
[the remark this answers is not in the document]
djhi: I can't, it uses a function (`onChange`) from the render props of Downshift

djhi:
Reviewed. Note that I will probably refactor those components a lot along with the others which handle multiple choices in a future PR

mpaciu:
I should write this issue maybe in sandbox, but like is the same problem here reporterted, only i go to mencion it. i am testing the beta 3.0.0 version, and i am seeing that te problem with this component when working with ReferenceInput persist. Attached an image. 

In the firs image i typed the letter ka. In the second image i wrote the letters ka an the letter u. When the component filter, the filter value stay on top of the window.

I hope to be clear and to be in the right place.

Sure, thaks you a lot for all this work in this project, is great.

type k
type k and u

PR #3670, opened 2019-09-11
[RFR][v3] FormDataConsumer - add subscription prop to the FormDataConsumer

natrim:
by default subscribe's to all changes like it is now
but user could change it by passing the subscription prop

### DIFF
--- a/packages/ra-core/src/form/FormDataConsumer.tsx
+++ b/packages/ra-core/src/form/FormDataConsumer.tsx
@@ -1,5 +1,6 @@
 import React, { ReactNode, SFC } from 'react';
 import { useFormState } from 'react-final-form';
+import { FormSubscription } from 'final-form';
 import get from 'lodash/get';
 
 import warning from '../util/warning';
@@ -15,6 +16,7 @@ interface ConnectedProps {
     form?: string;
     record?: any;
     source?: string;
+    subscription?: FormSubscription;
     [key: string]: any;
 }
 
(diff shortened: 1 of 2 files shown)


djhi:
Did you notice any performance issues ? Did you measure this change actually improve performances ?

natrim:
i have big form (80 inputs, iterators, arrayinputs with tables) and get small lags spikes without this if writing to some input fast enough, so the speed up is only ~50ms, so most users will not notice any change

fzaninotto:
I think FormDataConsumer should be "works all the time" component for subscribing to the form. For more fined grained subscriptions, I think users should use `useFormState` or `<FormSpy>`. So I'm -1 on this one.

natrim:
maybe i could add subscription prop to the FormDataConsumer
it will by default subscribe to all changes like it is now
but user could change it by passing the sub object?

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -117,8 +119,8 @@ export const FormDataConsumerView: SFC<Props> = ({
     return ret === undefined ? null : ret;
 };
 
-const FormDataConsumer = (props: ConnectedProps) => {
-    const formState = useFormState();
+const FormDataConsumer = ({ subscription, ...props }: ConnectedProps) => {
+    const formState = useFormState(subscription ? { subscription } : undefined);
```
fzaninotto: I think you don't need the test. Passing `useFormState({ subscription })` will default to all subscriptions if `subscription` is undefined.

natrim: ok

fzaninotto:
Thanks!

PR #3671, opened 2019-09-11
[RFR][v3][BC break] SimpleFormIterator - use resource fields label

natrim:
aka. use same format as FieldTitle to get translated labels in SImpleFormIterator from source without the need to write label's manually


fzaninotto:
I'm not sure I understand. Can you give a use case with sample code?

natrim:
https://codesandbox.io/s/simple-79rgs

when you look to edit post, then go to miscelanous and add backlink, you will see that the fields in the iterator are not translated

Sep-12-2019 07-58-15

fzaninotto:
Thanks!

Issue #2307, opened 2018-09-11
Provide an alternative to polyglot library with a library supporting all locales (e.g. intl-messageformat)


SorJEF:
As react-admin uses [polyglot](https://github.com/airbnb/polyglot.js) library there is no good way to use pluralization feature for many locales. For instance, this is the case with the Ukrainian and Belarussian language. Polyglot simply does not support them. Furthermore, looks like polyglot is no longer actively supported because [simple pull requests](https://github.com/airbnb/polyglot.js/pulls) adding new locales to locale groups are not getting resolved for more than a half-year.

A great solution would be to switch to another library especially as the tree shaking feature has been implemented. A candidate might be [intl-messageformat](https://github.com/yahoo/intl-messageformat), which is used across many popular and production-ready libraries like [react-intl](https://github.com/yahoo/react-intl). Furthermore, it uses official [Unicode CLDR rules for pluralization](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html) as well as other official Unicode specs for locale-specific formatting.

For many use cases, there is simply no alternative ways to achieve desired behavior except forking `react-admin` and reimplementing i18n. For some languages, a developer can load messages under another locale which has the same pluralization rules as the target one. But this approach cannot be used if it is required to have both the target locale and the implemented locale. This is often the case with Ukrainian, Belarussian, and other Slavic languages. In the polyglot library, there is only 1 applicable pluralization algorithm — Russian plural forms. In Ukraine, it is often required to use both languages for the app Ukrainian and Russian (because of the mostly bilingual community). Hence, described workaround is not applicable.

SorJEF:
As I understand replacing polyglot entirely is a very difficult task. Especially considering that many translation libraries for react-admin depend on it. A good solution might be providing some way for the user to choose another formatter programmatically.

fzaninotto:
Already discussed in https://github.com/marmelab/react-admin/issues/1099#issuecomment-332563058, and the decision we made was to not implement what you ask. See the related discussion for detailed arguments.

kopax:
@SorJEF we made react-intl function that help you swap polyglot with react-intl

- GitHub: https://github.com/yeutech-lab/react-admin-intl
- Doc: https://yeutech-lab.github.io/react-admin-intl/

It work great, it rely on webpack alias to tweak the import and inject a different translation provider.

It exactly do :  

```js
[...]

PR #3672, opened 2019-09-11
[RFR] Refactor I18n layer to use hooks

fzaninotto:
- [x] [BC Break] Rewrite the `TranslationProvider` using hooks
- [x] [BC Break] Remove i18n reducer and saga
- [x] Add `useSetLocale()` and `useLocale()` hooks
- [x] Add `useLoading()` and `useUpdateLoading()` hooks (wrapper around the dispatching of loading actions)
- [x] Display a notification if locale change fails
- [x] Update documentation
- [x] Add upgrade guide

`withTranslate` is unchanged, and the i18nProvider signature stays the same. The only changes for existing apps are for getting and setting locale, and for custom apps).

But the code is now much easier to follow, and more idiomatic to modern React.

Added benefit: when using a custom app, developers can replace the `TranslationProvider` by their own and use another translate function than Polyglot (as long as it has the same signature) (refs #2307.


packages/ra-core/src/i18n/TranslationProvider.tsx
```diff
@@ -1,102 +1,119 @@
...
+                    stopLoading();
+                    setState({
+                        provider: i18nProvider,
+                        locale: newLocale,
+                        translate: polyglot.t.bind(polyglot),
+                    });
+                })
+                .catch(error => {
+                    stopLoading();
+                    notify('ra.notification.i18n_error', 'warning');
```
djhi: This is a technical error. I'm not sure we should notify the end user about it

fzaninotto: The error occurs while loading (potentially via http) a new set of messages. The problem may be on the user side (networking problem), at this stage there is no way to know. 

I think it's preferable to show an error notification when the language switch fails, otherwise a user who pressed a 'change locale' button may hit the button again and again, thinking their click didn't work.

PR #3680, opened 2019-09-12
Independent scrolling for the book summary

andrestone:
Just a little fix for when navigating through items with many sub-items (such as "Tutorial").


fzaninotto:
Reverting as it creates a bug (#3700)

Issue #758, opened 2017-06-14
Disable missing Translation Warnings?


thochi:
Is it possible to disable the translation warnings if we don't use i18n?

djhi:
Not possible currently. Polyglot does support allowing undefined translations (see its [options](https://github.com/airbnb/polyglot.js#options-overview)) however. This would requires an additional prop to disable custom translations.

This babel plugin might be useful: [babel-plugin-transform-dev-warning](https://github.com/oliviertassinari/babel-plugin-transform-dev-warning)

fzaninotto:
Which translation warnings are you seeing? It must be on your custom components, because aor components don't throw translation warnings in case of missing translation for a field label. It uses the (non-documented- default translation everywhere:

```js
translate(label, { _: label }
```

As for the interface labels ("Save", "Filter", etc), they must always be in your translation dictionary. If they are not, that's a bug, and the warning should not be silenced.

onka13:
Temporary you can override console.error.

```
var originalLog = console.error
console.error = function log(...args) {
	if (args.length > 0 && typeof args[0] === "string" && /^Warning: Missing translation/.test(args[0])) {
		return
	}
	originalLog.apply(console, args)
}
```

fvoordeckers:
Solved it by adding a custom i18nProvider that allows missing keys:
```
const i18nProvider = polyglotI18nProvider(locale => i18nMessages[locale], 'en', { allowMissing: true });

<Admin
    i18nProvider={i18nProvider}
    ...
 >
```

More details: https://marmelab.com/react-admin/Translation.html
and: https://www.npmjs.com/package/node-polyglot#options-overview

caiquejjx:
Which package gives me the "i18nMessages" instance?

fvoordeckers:
> Which package gives me the "i18nMessages" instance?

@caiquejjx  That's a dict you build containing all the locales you want to support

dkocich:
it would be good to see one line entries instead of printing the whole stack trace with each one missing translation

PR #366, opened 2017-02-25
Expose polyglot instance

yshing:
Separate the creation of polyglot instance from TranslationProvider and make it expose at
admin-on-rest/i18n/polyglot.

PR #3685, opened 2019-09-12
[RFR] I18n provider new signature

fzaninotto:
In order to allow different translation frameworks than Polyglot (#2307), or to pass options to the Polyglot constructor (#3662, #758, #366), we need to move some react-admin logic inside the `i18nProvider`.

With this PR, the i18nProvider has a signature similar to the other providers: it accepts a message type (either `I18N_TRANSLATE` or `I18N_CHANGE_LOCALE`) and a params argument.

```jsx
// react-admin 2.x
const i18nProvider = (locale) => messages[locale];

// react-admin 3.x
const i18nProvider = (type, params) => {
    const polyglot = new Polyglot({ locale: 'en', phrases: messages.en });
    let translate = polyglot.t.bind(polyglot);
    if (type === 'I18N_TRANSLATE') {
        const { key, options } = params;
        return translate(key, options);
    }
    if type === 'I18N_CHANGE_LOCALE') {
        const newLocale = params;
        return new Promise((resolve, reject) => {
            // load new messages and update the translate function
        })
    }
} 
```

But don't worry: react-admin v3 exports a function called `polyglotI18nProvider`, that you can just wrap around your old `i18nProvider` to make it compatible with the new provider signature:

```diff
import React from 'react';
-import { Admin, Resource } from 'react-admin';
+import { Admin, Resource, polyglotI18nProvider } from 'react-admin';
import englishMessages from 'ra-language-english';
import frenchMessages from 'ra-language-french';

const messages = {
    fr: frenchMessages,
    en: englishMessages,
};
const i18nProvider = locale => messages[locale];

const App = () => (
-   <Admin locale="en" i18nProvider={i18nProvider}>
+   <Admin locale="en" i18nProvider={polyglotI18nProvider(i18nProvider)}>
        ...
    </Admin>
);

export default App;
```


UPGRADE.md
```diff
@@ -1034,6 +1034,56 @@ If you had custom reducer or sagas based on these actions, they will no longer w
...
+
+const messages = {
+    fr: frenchMessages,
+    en: englishMessages,
+};
+const i18nProvider = locale => messages[locale];
+
+const App = () => (
+-   <Admin locale="en" i18nProvider={i18nProvider}>
++   <Admin locale="en" i18nProvider={polyglotI18nProvider(i18nProvider)}>
```
djhi: Why not wrap the function at L1075 ?

packages/ra-core/src/i18n/useSetLocale.tsx
```diff
@@ -30,9 +33,31 @@ import { TranslationContext } from './TranslationContext';
...
+    const { setLocale, i18nProvider } = useContext(TranslationContext);
+    const { startLoading, stopLoading } = useUpdateLoading();
+    const notify = useNotify();
+    return useCallback(
+        (newLocale: string) =>
+            new Promise(resolve => {
+                startLoading();
+                // so we systematically return a Promise for the messages
+                // i18nProvider may return a Promise for language changes,
+                resolve(i18nProvider(I18N_CHANGE_LOCALE, newLocale));
```
djhi: Can't we end up with a promise resolving to another promise here ?

packages/ra-core/src/i18n/useTranslate.ts
```diff
@@ -21,9 +22,17 @@ import { TranslationContext } from './TranslationContext';
...
  * }
  */
-const useTranslate = () => {
-    const { translate } = useContext(TranslationContext);
-    return translate;
+const useTranslate = (): Translate => {
+    const { i18nProvider, locale } = useContext(TranslationContext);
+    const translate = useCallback(
+        (key: string, options?: any) =>
+            i18nProvider(I18N_TRANSLATE, { key, options }) as string,
```
djhi: We are not accepting promises at all for this call ?

packages/ra-core/src/i18n/useSetLocale.tsx
fzaninotto: Promises are unwrapped by JS.

packages/ra-core/src/i18n/useTranslate.ts
fzaninotto: Because if translate is asynchronous, every translatable component will render twice. That's too bad for performance.

packages/ra-core/package.json
```diff
@@ -57,10 +57,10 @@
         "final-form": "^4.18.5",
         "inflection": "~1.12.0",
         "lodash": "~4.17.5",
-        "node-polyglot": "^2.2.2",
         "prop-types": "^15.6.1",
         "query-string": "~5.1.1",
         "ra-language-english": "^3.0.0-alpha.4",
+        "ra-i18n-polyglot": "^3.0.0-alpha.4",
```
djhi: Shouldn't this be in the dev dependencies only, for tests ? Otherwise we are still including polyglot for all users

fzaninotto: Just like for the english language, we need a sensible default.

PR #3692, opened 2019-09-15
Support for GraphQL Interface type

MichielDeMey:
This PR simply adds support for `Interface` types.

For example if you have a similar GraphQL schema
```
interface Device {
  ...    
}

type MotionSensor implements Device {
  ...    
}

type TemperatureSensor implements Device {
  ...
}
```

`Device` will then be picked up as a resource.


fzaninotto:
If I understand correctly, this is also a bugfix as the current graphQL provider fails when using interfaces?

MichielDeMey:
I would assume so. Haven't been able to test this with GraphCool or any other GraphQL services as I'm using a regular Apollo GraphQL server.

djhi:
Thanks!

PR #3698, opened 2019-09-16
[RFR] Introduce useGetMatching and useReferenceArrayInputController

djhi:
- [x] useGetMatching
- [x] useReferenceArrayInputController
- [x] Fix tests

### DIFF
--- a/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
@@ -1,13 +1,12 @@
 import React from 'react';
 import expect from 'expect';
-import { cleanup, wait } from '@testing-library/react';
+import { cleanup, wait, fireEvent } from '@testing-library/react';
 import ReferenceArrayInputController from './ReferenceArrayInputController';
 import { renderWithRedux } from '../../util';
-import { DataProviderContext } from '../../../lib';
+import { CRUD_GET_MATCHING, CRUD_GET_MANY } from '../../../lib';
 
 describe('<ReferenceArrayInputController />', () => {
     const defaultProps = {
-        children: jest.fn(),
         input: { value: undefined },
         record: undefined,
         basePath: '/tags',
@@ -19,8 +18,10 @@ describe('<ReferenceArrayInputController />', () => {
     afterEach(cleanup);
 
     it('should set loading to true as long as there are no references fetched and no selected references', () => {
-        const children = jest.fn(() => <div />);
-        renderWithRedux(
+        const children = jest.fn(({ loading }) => (
+            <div>{loading.toString()}</div>
+        ));
+        const { queryByText } = renderWithRedux(
             <ReferenceArrayInputController
                 {...defaultProps}
                 input={{ value: [1, 2] }}
@@ -29,12 +30,14 @@ describe('<ReferenceArrayInputController />', () => {
             </ReferenceArrayInputController>
         );
 
-        expect(children.mock.calls[0][0].loading).toEqual(true);
+        expect(queryByText('true')).not.toBeNull();
     });
 
     it('should set loading to true as long as there are no references fetched and there are no data found for the references already selected', () => {
-        const children = jest.fn(() => <div />);
-        renderWithRedux(
+        const children = jest.fn(({ loading }) => (
+            <div>{loading.toString()}</div>
+        ));
+        const { queryByText } = renderWithRedux(
             <ReferenceArrayInputController
                 {...defaultProps}
                 input={{ value: [1, 2] }}
@@ -42,17 +45,17 @@ describe('<ReferenceArrayInputController />', () => {
                 {children}
             </ReferenceArrayInputController>
         );
-        expect(children.mock.calls[0][0].loading).toEqual(true);
+        expect(queryByText('true')).not.toBeNull();
     });
 
     it('should set loading to false if the references are being searched but data from at least one selected reference was found', () => {
-        const children = jest.fn(() => <div />);
-        renderWithRedux(
+        const children = jest.fn(({ loading }) => (
+            <div>{loading.toString()}</div>
+        ));
+        const { queryByText } = renderWithRedux(
             <ReferenceArrayInputController
-                {...{
-                    ...defaultProps,
-                    input: { value: [1, 2] },
-                }}
+                {...defaultProps}
+                input={{ value: [1, 2] }}
             >
                 {children}
             </ReferenceArrayInputController>,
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -198,34 +205,15 @@ describe('<ReferenceArrayInputController />', () => {
...
-            >
-                {children}
-            </ReferenceArrayInputController>
-        );
-        assert.equal(children.mock.calls[0][0].onChange, onChange);
+        expect(children.mock.calls[0][0].warning).toEqual(undefined);
     });
 
-    it('should call crudGetMatching on mount with default fetch values', () => {
+    it.skip('should call crudGetMatching on mount with default fetch values', () => {
```
fzaninotto: skip detected, here and down under

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -198,34 +205,15 @@ describe('<ReferenceArrayInputController />', () => {
                 {children}
             </ReferenceArrayInputController>
         );
-        assert.equal(children.mock.calls[0][0].warning, undefined);
-    });
-
-    it('should pass onChange down to child component', () => {
```
fzaninotto: why did you remove this one?

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -269,207 +256,181 @@ describe('<ReferenceArrayInputController />', () => {
...
-            },
-            {
-                field: 'id',
-                order: 'DESC',
-            },
-            {
-                q: 'bar',
-            },
-        ]);
+        // wrapper.instance().setFilter('bar');
```
fzaninotto: commented out code (in a skipped test). You *really* don't want that code to run!

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
```diff
@@ -1,34 +1,19 @@
...
-    `${resource}@${source}`;
+import { Record, Sort, Pagination } from '../../types';
+import useReferenceArrayInputController from './useReferenceArrayInputController';
 
 interface ChildrenFuncParams {
     choices: Record[];
     error?: string;
+    loaded: boolean;
     loading: boolean;
-    onChange: (value: any) => void;
```
fzaninotto: I still don't understand why the new controller doesn't return this one.

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -0,0 +1,145 @@
...
+ * @param {string} option.reference The linked resource name
+ * @param {string} option.resource The current resource name
+ * @param {string} option.source The key of the linked resource identifier
+ *
+ * @return {Object} controllerProps Fetched data and callbacks for the ReferenceArrayInput components
+ */
+const useReferenceArrayInputController = ({
+    basePath,
+    filter: defaultFilter,
+    filterToQuery = searchText => ({ q: searchText }),
```
fzaninotto: Move the default value to a constant to avoid having a new function at every render, defeating your useMemo line 93

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -0,0 +1,145 @@
...
+    options,
+    reference,
+    resource,
+    source,
+}: Option): ReferenceArrayInputProps => {
+    const translate = useTranslate();
+
+    const [pagination, setPagination] = useState({ page: 1, perPage });
+    const [sort, setSort] = useState(defaultSort);
+    const [filter, setFilter] = useState('');
```
fzaninotto: the default filter should be an empty object

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -0,0 +1,145 @@
...
+        reference,
+        pagination,
+        sort,
+        finalFilter,
+        source,
+        resource,
+        options
+    );
+
+    const finalReferenceRecords = referenceRecords.filter(
```
fzaninotto: We need a comment here, like "filter out not found references - happens when the dataProvider doesn't guarantee referential integrity"

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -0,0 +1,145 @@
...
+    const finalMatchingReferences =
+        matchingReferences && matchingReferences.length > 0
+            ? (matchingReferences || []).concat(finalReferenceRecords)
+            : finalReferenceRecords.length > 0
+            ? finalReferenceRecords
+            : undefined;
+
+    const dataStatus = getDataStatus({
+        input,
+        // We merge the currently selected records with the matching ones, otherwise
```
fzaninotto: I think you should move this comment up to line 115

packages/ra-core/src/dataProvider/useGetMatching.ts
```diff
@@ -0,0 +1,113 @@
...
+import {
+    getReferenceResource,
+    getPossibleReferenceValues,
+    getPossibleReferences,
+} from '../reducer';
+
+const referenceSource = (resource, source) => `${resource}@${source}`;
+
+/**
+ * Call the dataProvider with a GET_LIST verb and return the result as well as the loading state.
```
fzaninotto: The comment is the same as useGetList... You should tweak it so that readers understand the difference

packages/ra-core/src/dataProvider/useGetMatching.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param {string} referencingResource The resource name, e.g. 'posts'. Used to build a cache key
+ * @param {Object} options Options object to pass to the dataProvider. May include side effects to be executed upon success of failure, e.g. { onSuccess: { refresh: true } }
+ *
+ * @returns The current request state. Destructure as { data, total, ids, error, loading, loaded }.
+ *
+ * @example
+ *
+ * import { useGetMatching } from 'react-admin';
+ *
+ * const LatestNews = () => {
```
fzaninotto: Bad example

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
djhi: Yes, because this PR is WIP ;) It event has a `Fix tests` in its description!

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
djhi: Because onChange override is handled at the input level. See the `useInput` hook

packages/ra-core/src/dataProvider/useGetMatching.ts
djhi: It is tweaked in the example. GET_MANY is actually sent as GET_LIST

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
djhi: No, because it goes through `filterToQuery` which then transform it into an object

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -1,475 +1,633 @@
...
+        expect(dispatch).toHaveBeenCalledWith({
+            type: CRUD_GET_MANY,
+            meta: {
+                resource: 'tags',
+            },
+            payload: { ids: [6] },
+        });
     });
 
-    it('should only call crudGetOne and not crudGetMatching when only the record changes', () => {
```
djhi: I removed this test as the component should not care about record, only about the value

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -1,475 +1,633 @@
...
-            'tags',
-            'posts@tag_ids',
-            { page: 1, perPage: 42 },
-            { field: 'foo', order: 'ASC' },
-            { foo: 'bar' },
-        ]);
-        assert.equal(crudGetMany.mock.calls.length, 1);
-    });
 
-    it('should call crudGetMany when input value changes', () => {
```
djhi: Merged this test and the next one

packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx
djhi: > Because `onChange` override is handled at the input level. See the `useInput` hook

PR #3704, opened 2019-09-17
Independent book-summary scrolling saga

andrestone:
- Added script to default.html in order to change style according to the search result display.
- Added the overflow styling back to the css.
pr2


fzaninotto:
Thanks!

PR #3705, opened 2019-09-17
[RFR] Migrate ReferenceArrayInput to use hooks

djhi:


packages/ra-ui-materialui/src/input/ReferenceArrayInput.js
```diff
@@ -209,22 +212,43 @@ ReferenceArrayInputView.propTypes = {
...
-            )}
-        </ReferenceArrayInputController>
+        <ReferenceArrayInputView
+            id={id}
+            input={input}
+            isRequired={isRequired}
+            meta={meta}
+            translate={translate}
+            {...props}
+            {...{ children, ...controllerProps }}
```
fzaninotto: why the double clone?

PR #3709, opened 2019-09-18
[RFR][v3] CoreAdminRouter - use useSafeSetState

natrim:
to prevent warning about setting state when unmounted, as the component can be unmounted before resolveChildren returns

( happens to me when i load permissions asynchronously )


fzaninotto:
Thanks!

Issue #3695, opened 2019-09-16
tab label not change to red while one of it's input  with source='post.title'(the source like x.y) is invalid


atares0223:
**What you were expecting:**
tab label change to red while one of it's input with source='post.title'  is invalid. 
**What happened instead:**
tab label not change to red, once click save, you don't know where is invalid for a TabbedForm.
**Steps to reproduce:**
create a TabbedForm with FormTabs.
create some tabs.
have one input with source='post.title' and required. 
switch to other tab , the tab have 'post.title' is not showing warning red label.
`<TextInput source="post.title" validate={required()}/>`
![image](https://user-images.githubusercontent.com/16397455/65012439-ec313a80-d949-11e9-83c4-79904c3adb77.png)
![image](https://user-images.githubusercontent.com/16397455/65012502-23075080-d94a-11e9-8629-98c2bc68b022.png)

**Related code:**
https://codesandbox.io/s/simple-ttj4i?fontsize=14

**Environment:**

React-admin version: 2.9.1 and v3 alpha
React version: 16.8.6
Browser: Chrome

djhi:
Please follow the issue template

atares0223:
@djhi please help me figure it out.  it not surpport complex source for a input.

djhi:
Your codesandbox does not use the version of react-admin you specified in the issue template: it is based on the v3 alpha and not on `v2.9.1`.

Besides, using `v2.9.6`, I cannot reproduce the issue.

Please either provide a codesandbox with the version of react-admin you are actually using or correct your issue description

atares0223:
@djhi  thanks for help , my dev version is v2.9.1 and the codesandbox is v2.9.6. it's both not working.
I think I don't make myself clearly, the  codesandbox can reproduce the issue.
steps in the  codesandbox:
1. go to post edit page.
2. make sure that the 'post.title' is empty so that is will trigger a invalid required warnning under the input.
3. switch to other tabs except SUMMARY.
4.it should trigger a event that make the label of the **SUMMARY** tab to **red 'SUMMARY'** , but it's not. since the source is changed from 'title'->'**post.title**'  (a format like **x.y**)

djhi:
Try upgrading to `2.9.6`

djhi:
I cannot reproduce your issue on the official codesandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple)

atares0223:
[...]

PR #3710, opened 2019-09-18
Fix TabbedForm does not detect errors when source is a path

djhi:
#3695 for v3


djhi:
@fzaninotto Maybe you prefer to backport the commits from `v2.9.7` (see #3711) ?

fzaninotto:
I'm OK with having it in 3.0 as the components have changed a lot anyways. But your PR needs rebasing.

PR #3712, opened 2019-09-18
Support for GraphQL Interface type, continuation of 3692

MichielDeMey:
This PR is a continuation of https://github.com/marmelab/react-admin/pull/3692, which adds Interface support.


djhi:
Thanks!

PR #3713, opened 2019-09-19
[RFR][v3] useQuery/useQueryWithStore - fix loading state not resetting

natrim:
we need to set loading to true ( like in withMutation ) to support reload

without it the loading is then always false on subsequent re-renders

and withMutation fix loaded being undefined for while


djhi:
Would you mind rebasing ?

natrim:
done

fzaninotto:
indeed, thanks!

PR #3714, opened 2019-09-19
Allowing for more fine tuning of Quill without rewriting RichTextInput

Sheraff:
- Implementing the changes mentioned in https://github.com/marmelab/react-admin/pull/3653 
- Documenting changes in ra-rich-text-input/README.md
- Documenting changes in docs/Inputs.md
- Requesting pull into `next` branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,20 +2,20 @@
 
 ## v2.9.6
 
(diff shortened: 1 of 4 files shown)


Sheraff:
> Just about the punctilious about the naming, I would suggest to rename the prop for something like `configureQuill` that would be more clear IMO

Done ;-)

fzaninotto:
Thanks!

PR #3723, opened 2019-09-20
Ensure we initialize the i18nProvider with the locale specified on Admin

### LINKED ISSUE #3722: Default language other than English is not working
Following the documentation [Change the default regional settings](https://github.com/marmelab/react-admin/blob/next/docs/Translation.md#changing-the-default-locale) when the prop locale in `<Admin>` is different to `'en'`, the initial translations don't seem works correctly.

However, the language seems to be set correctly and the language switch works perfectly.

**What you were expecting:**
That the application loads as the initial language the value from the `locale` prop of the `<Admin>` component.

**What happened instead:**
The application loads always English as the initial language, independent of the `locale` prop of the `Admin` component.

**Steps to reproduce:**
Use the demo to reproduce it and in `App.js` simply replace the prop `locale` to `'fr'` instead `'en'`.

**Environment**

* React-admin version: 3.0.0-alpha.4

djhi:
Fixes #3722


fzaninotto:
Superseded by #3755

Issue #3686, opened 2019-09-12
[RFC] Replace Providers by objects


fzaninotto:
## Problems

It's impossible to give useful Typescript types to the `dataProvider`, `authProvider`, and (since #3685) `i18nProvider` because they are functions accepting a `type` parameter as first argument, and with a return value depending on that `type`. This makes writing custom providers a bit harder. 

Besides, the `switch/case` syntax of providers is sometimes seen as a bad coding practice. 

Lastly, all the constants we use (like `GET_ONE`, `AUTH_LOGOUT`, etc) make the code heavier and longer to write. 

## Proposal

React-admin calls the Provider functions with a message type and arguments. There is another way to do that, more idiomatic to JavaScript: calling methods on an object.

I propose to transform the syntax of all 3 providers so that they can be objects. 

For instance, an `authProvider` written as a function:

```jsx
import { AUTH_LOGIN, AUTH_LOGOUT, AUTH_CHECK, AUTH_ERROR } from 'react-admin';

export default (type, params) => {
    if (type === AUTH_LOGIN) {
        const { username } = params;
        localStorage.setItem('username', username);
        // accept all username/password combinations
        return Promise.resolve();
    }
    if (type === AUTH_LOGOUT) {
        localStorage.removeItem('username');
        return Promise.resolve();
    }
    if (type === AUTH_ERROR) {
        return Promise.resolve();
    }
    if (type === AUTH_CHECK) {
        return localStorage.getItem('username')
            ? Promise.resolve()
            : Promise.reject();
    }
    return Promise.reject('Unkown method');
};
```

Can be written as an object as follows:

```jsx
export default {
    login({ username }) {
        localStorage.setItem('username', username);
        // accept all username/password combinations
        return Promise.resolve();
    },
    logout() {
        localStorage.removeItem('username');
        return Promise.resolve();
    },
    error() {
        return Promise.resolve();
    },
    check() {
        return localStorage.getItem('username')
            ? Promise.resolve()
            : Promise.reject();
    }
}
```

This is very easy to Type using TypeScript. Something like:

```ts
type AuthProvider {
    login: (args: any) => Promise<void>;
    logout: () => Promise<void>;
    error: (error: any) => Promise<void>;
    check: () => Promise<void>
}
```

## Backward Compatibility

This is not a breaking change if react-admin detects that a provider is a function rather than an object, and wraps it inside a converter function:

```js
[...]

PR #3726, opened 2019-09-23
[RFR] Convert DataProvider to Object

djhi:
Related to #3686

- [x] Add TypeScript type for DataProvider methods
- [x] Make `useDataProvider` return a proxied `dataProvider`
- [x] Remove handling of event in `useMutation` callback
- [x] Add converter for legacy (function) style dataProviders to make the change BC
- [x] Update examples
- [x] Update documentation
  - [x] DataProvider
  - [x] Querying the API
  - [x] Other chapters
- [x] Update core data providers
  - [x] `ra-data-fakerest`
  - [x] `ra-data-simple-rest`
  - [x] `ra-data-json-server`
  - [ ] ~~`ra-data-graphql-simple`~~ (too much work, the object/function wrapper is enough)
  - [ ] ~~`ra-data-graphcool`~~ (too much work, the object/function wrapper is enough)


docs/CreateEdit.md
```diff
@@ -945,3 +945,106 @@ export const UserEdit = ({ permissions, ...props }) =>
...
+        form.change('average_note', 10);
+
+        handleSubmitWithRedirect('edit');
+    }, [form]);
+
+    return <SaveButton {...props} handleSubmitWithRedirect={handleClick} />;
+};
+```
+
+2. Using react-admin hooks to run custom mutations
```
djhi: Note that this will not **trigger** a form validation pass

docs/DataProviders.md
```diff
@@ -5,31 +5,54 @@ title: "Data Providers"
...
+dataProvider
+    .getOne('posts', { id: 123 })
+    .then(response => {
+        console.log(response.data); // { id: 123, title: "hello, world" }
+    })
+```
+
+It's the Data Provider's job to turns these method calls into HTTP requests, and transform the HTTP responses to the data format expected by react-admin. In technical terms, a Data Provider is an *adapter* for an API. 
+
+And to inject a Data Provider in a react-admin application, just pass it as the `dataProvider` prop of the `<Admin>` component, as follows:
```
djhi:
```suggestion
And to inject a Data Provider in a react-admin application, pass it as the `dataProvider` prop of the `<Admin>` component, as follows:
```

docs/DataProviders.md
```diff
@@ -167,320 +183,170 @@ Now all the requests to the REST API will contain the `X-Custom-Header: foobar`
...
- *  => Promise.resolve({ data: { id: 123, title: "hello, world" } })
- *
- * @param {string} type Request type, e.g GET_LIST
- * @param {string} resource Resource name, e.g. "posts"
- * @param {Object} payload Request parameters. Depends on the action type
- * @returns {Promise} the Promise for a response
- */
-const dataProvider = (type, resource, params) => new Promise();
-```
+APIs are so diverse that quite often, none of the available Data Providers suit you API. In such cases, you'll have to write your own Data Provider. Don't worry, it usually takes a couple hours and presents no difficulty.
```
djhi:
```suggestion
APIs are so diverse that quite often, none of the available Data Providers suit you API. In such cases, you'll have to write your own Data Provider. Don't worry, it usually takes only a couple hours. 
```

fzaninotto:
Switching to RFR

djhi:
Sorry, need a rebase

packages/ra-core/src/dataFetchActions.ts
```diff
@@ -20,3 +20,28 @@ export const fetchActionsWithArrayOfRecordsResponse = [
     DELETE_MANY,
 ];
 export const fetchActionsWithTotalResponse = [GET_LIST, GET_MANY_REFERENCE];
+
+export const sanitizeFetchType = (fetchType: string) => {
```
pemoreau: can't we use types for `GET_LIST` and `getList`, ... this would make switch/case/ safer.
I could try to do it in another PR

packages/ra-core/src/dataProvider/useMutation.ts
```diff
@@ -129,11 +178,131 @@ const useMutation = (
...
+    if (callTimeQuery instanceof Event)
+        return {
+            type: query.type,
+            resource: query.resource,
+            payload: query.payload,
+            options: sanitizeOptions(options),
+        };
+    if (query)
+        return {
+            type: query.type || callTimeQuery.type,
```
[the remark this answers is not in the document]
fzaninotto: No, I don't think so: If both `query` and `callTimeQuery` are falsy, the test line 258 will throw an error. So in this line, at least one of the two is of type `Mutation`, which has a required `type`.

packages/ra-core/src/dataProvider/useMutation.ts
```diff
@@ -129,11 +178,131 @@ const useMutation = (
...
+        payload: callTimeQuery.payload,
+        options: sanitizeOptions(callTimeOptions),
+    };
+};
+
+const hasDeclarativeSideEffectsSupport = (
+    options?: MutationOptions,
+    callTimeOptions?: MutationOptions
+) => {
+    if (!options && !callTimeOptions) return false;
```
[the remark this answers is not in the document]
fzaninotto: I don't think so. Can you elaborate?

packages/ra-core/src/dataFetchActions.ts
fzaninotto: Well, the types should ba 'open' in that case, because users may pass a custom fetch action (like `placeOrder` or `login`) and we don't want to fail at compile time in this case. 

Can be tested on another PR.

packages/ra-core/src/dataProvider/useMutation.ts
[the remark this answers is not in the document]
pemoreau: May be I do not understand something, but we can have `query !== null` and `query.type` falsy. And in this case, since ` callTimeQuery` may be null, a null exception could occur.

packages/ra-core/src/dataProvider/useMutation.ts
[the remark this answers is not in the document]
pemoreau: when `(!options && !callTimeOptions)` is true, none of the two following tests can be true (since the need at least `callTimeOptions`  or `options` to exist)

Therefore, the last `return false` is sufficient. 
In other words : by default `false` is returned, except when `(callTimeOptions && callTimeOptions.withDeclarativeSideEffectsSupport)` or `(options && options.withDeclarativeSideEffectsSupport)`

packages/ra-core/src/dataFetchActions.ts
pemoreau: OK I understand.

PR #3730, opened 2019-09-24
[RFR][v3] Notification - fix text color

natrim:
the default black font color is hard to see, so use MUI defined contrastText color

be

foro


djhi:
Thanks! Would you mind adding screenshots for before and after the fix?

natrim:
ok

PR #3732, opened 2019-09-25
[RFR] Fix FilterForm Submit

### LINKED ISSUE #3731: [v3] Filter does reload whole app on Enter
Filter does not stop enter from submiting the whole form to browser as GET to current url

in v2 there was no form but only div so it did not happen

repro in basic simple example: https://codesandbox.io/s/simple-wsknw

could be probably fixed by onSubmit={e => e.preventDefault()} on < form >

steps to reproduce:

1) click to search
2) write to search something and press enter before it auto-submits
3) you will get normal browser form submit to GET
4) the whole app reloads and seems like filtering failed (well, it failed)

Sep-24-2019 16-21-35

djhi:
Fixes #3731

### DIFF
--- a/packages/ra-ui-materialui/src/list/FilterForm.js
+++ b/packages/ra-ui-materialui/src/list/FilterForm.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Form, FormSpy } from 'react-final-form';
 import classnames from 'classnames';
@@ -37,7 +37,10 @@ const sanitizeRestProps = ({
     dirtySinceLastSubmit,
     dispatch,
     displayedFilters,
+    errors,
+    filters,
     filterValues,
+    form,
     handleSubmit,
     hasSubmitErrors,
     hasValidationErrors,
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/FilterForm.js
```diff
@@ -61,71 +65,76 @@ const sanitizeRestProps = ({
...
                 throw new Error(
                     'Cannot use alwaysOn and defaultValue on a filter input. Please set the filterDefaultValues props on the <List> element instead.'
                 );
             }
         });
-    }
-
-    getShownFilters() {
-        const { filters, displayedFilters, initialValues } = this.props;
+    }, [filters]);
```
fzaninotto: `filters` will change at each render, so this effect will run way too often. Maybe not significant though.

packages/ra-ui-materialui/src/list/FilterForm.js
```diff
@@ -61,71 +65,76 @@ const sanitizeRestProps = ({
...
-                        variant={variant}
-                    />
-                ))}
-                <div className={classes.clearFix} />
-            </form>
-        );
-    }
-}
+
+    const handleHide = event => hideFilter(event.currentTarget.dataset.key);
```
fzaninotto: memoize with useCallback

PR #3740, opened 2019-09-26
Change filter extend order in ReferenceInputController

### LINKED ISSUE #3739: Changing filter in `ReferenceArrayInput` does change the filter added to the `GET_LIST` action.
**What you were expecting:**

Updating filter in `ReferenceArrayInput` should update the filter added in the `GET_LIST` API call.

**What happened instead:**

Updating filter in `ReferenceArrayInput` does not update filter in the API call. Uses the stale filter.

**Steps to reproduce:**

https://codesandbox.io/s/simple-cop17?fontsize=14

Try to create a Post `/posts/create` which has a `ReferenceArrayInput` for searching a user. You can see in the browser console action logger that the `filter` in the `GET_LIST` call is one form change old.
image

See in the screenshot above - the user input for email does not match the filter in the action.

**Other information:**

I believe this is happening because the `fetchOptions` in the line [203](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx#L203) in `ReferenceArrayInputController` is called without passing the `nextProps` as argument which means the `defaultFilter` in line [262](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/input/ReferenceArrayInputController.tsx#L262) is still the old filter.

**Environment**

* React-admin version: ^2.9.5
* React version:^16.8.6
* Browser:Chrome

TheHyphen:
Closes #3739

### DIFF
--- a/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
@@ -402,6 +402,33 @@ describe('<ReferenceArrayInputController />', () => {
         assert.equal(crudGetMany.mock.calls.length, 1);
     });
 
+    it('should call crudGetMatching with replaced filter from outside', () => {
+        const crudGetMatching = jest.fn();
+        const crudGetMany = jest.fn();
+        const wrapper = shallow(
+            <ReferenceArrayInputController
+                {...defaultProps}
+                allowEmpty
+                input={{ value: [5] }}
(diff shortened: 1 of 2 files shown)


djhi:
Thanks! Would you mind adding a test for this?

TheHyphen:
Done, added tests now.

TheHyphen:
When do you think this might make it to a release?

fzaninotto:
Thanks!

PR #3742, opened 2019-09-27
[RFR] Discriminated union type for auth action

pemoreau:
Applied discriminated union type for auth actions to make the code more uniform


pemoreau:
Thanks.
I think this is the case:
- `AuthActionType` is exported by `src/auth/types.ts`
- `src/auth/index.ts` contains `export * from './types';`
- `src/index.ts` contains `export * from './auth';`
but correct me if I do not see something

PR #3743, opened 2019-09-27
Add Content-Type header only for non GET requests.

### LINKED ISSUE #3687: Issue with the Content-Type header for GET requests
**What you were expecting:**

When React-Admin fetches the list of items for a specific model with `fetchUtils.fetchJson`, the GET request should not have a `Content-Type` header.

**What happened instead:**

A `Content-Type` header with the value `application/json` is added to the request.

**Steps to reproduce:**

Create a sample react-admin application and go to the list of items for an entity.
Check the HTTP headers of the request.

**Related code:**

I am sorry I cannot provide a sample right now but any React-Admin application using `fetchUtils.fetchJson` for your HTTP requests should show the issue. The related code adding this header for `fetchUtils.fetchJson` is:

https://github.com/marmelab/react-admin/blob/893add6e7c98be3a68f55fb42c9f8fbd7879d3d9/packages/ra-core/src/util/fetch.ts#L16-L21

**Other information:**

As per the [Mozilla documentation of the header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type), the `Content-Type` header in a request indicate the type of the body, which is relevant for `PUT` or `POST` requests but not `GET` requests.

I am particularly affected by this issue because a firewall in my company is blocking these requests. As it is not required for `GET` requests, I think it can be removed.

I am willing to work on a PR to fix this if this is OK for you.

**Environment**

* React-admin version: 2.9.6
* Last version that did not exhibit the issue (if applicable): None
* React version: 16.4.1
* Browser: All

clement-escolano:
Fixes #3687

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
         "prettier": "~1.17.1",
         "raf": "~3.4.0",
         "ts-jest": "^23.10.5",
-        "wait-on": "^2.1.0"
+        "wait-on": "^2.1.0",
+        "whatwg-fetch": "^3.0.0"
     },
     "workspaces": [
         "packages/*",
--- a/packages/ra-core/src/util/fetch.spec.ts
+++ b/packages/ra-core/src/util/fetch.spec.ts
@@ -1,5 +1,9 @@
 import assert from 'assert';
-import { queryParameters, flattenObject } from './fetch';
+import {
+    createHeadersFromOptions,
+    queryParameters,
+    flattenObject,
+} from './fetch';
 
 describe('queryParameters', () => {
     it('should generate a query parameter', () => {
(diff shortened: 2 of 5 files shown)


clement-escolano:
> Would you mind adding a test for it ?

The `fetchJson` is actually not tested at all and involve a lot of logic (network, promises, etc). I can add some tests for the use case of the PR but I don't feel comfortable with creating all the boilerplate to test `fetchJson` method.
What do you think?

djhi:
A test just for the use case of the PR would be enough :) Thanks!

fzaninotto:
Thanks!

PR #3744, opened 2019-09-27
[RFR] Upgrade React-router

djhi:


packages/ra-core/src/auth/useLogout.ts
```diff
@@ -40,42 +40,38 @@ const useLogout = (): Logout => {
 
     const store = useStore();
     const dispatch = useDispatch();
+    const history = useHistory();
 
     const logout = useCallback(
         (params = {}, redirectTo = defaultAuthParams.loginUrl) =>
             authProvider.logout(params).then(redirectToFromProvider => {
                 dispatch(clearState());
                 const currentLocation = store.getState().router.location;
```
fzaninotto: Can't you get rid of this one redux usage?

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -40,42 +40,38 @@ const useLogout = (): Logout => {
...
+                });
                 return redirectToFromProvider;
             }),
-        [authProvider, store, dispatch]
+        [authProvider, store, history, dispatch]
     );
 
     const logoutWithoutProvider = useCallback(
         _ => {
             const currentLocation = store.getState().router.location;
```
fzaninotto: and here, too.

packages/ra-ui-materialui/src/field/ArrayField.spec.js
```diff
@@ -60,9 +61,11 @@ describe('<ArrayField />', () => {
             </ArrayField>
         );
         const wrapper = mount(
-            <Provider store={createStore(() => ({}))}>
-                <Dummy />
-            </Provider>
+            <MemoryRouter>
```
fzaninotto: you can pretty much use the TestContext here

fzaninotto:
also, do we still use `connected-react-router` for anything else than sagas?

fzaninotto:
And there is the FIXME in the CloneButton, can we get rid of it?

packages/ra-ui-materialui/src/field/ArrayField.spec.js
djhi: I cleanup those tests and rewrote them using rtl. One of them was useless

fzaninotto:
> Only place left is in the saving reducer.

I think we can get rid of this one, as the `useUpdate` and `useCreate` hook pass their own `loading` state. We'd need to pass it down from  Edit to SimpleForm instead of selecting the value in SimpleForm.

As it's a BC break, we'd better do it now. 

> However, I'm not sure we should pass the record in the state. Having it in the URL is a good thing imo: explicit, allow to refresh the page, copy/paste the url, etc.

I agree. Can you check that using the `state` is never documented anywhere and remove the FIXME then?

Issue #3660, opened 2019-09-09
[alpha.4][v3] Cannot save empty item in SelectInput with allowEmpty


natrim:
SelectInput cannot be cleared when using allowEmpty

![Sep-09-2019 14-44-08](https://user-images.githubusercontent.com/488136/64531832-9be62700-d310-11e9-8b10-85484c216b0b.gif)

Reproduction sandbox:
https://codesandbox.io/s/simple-yd6sm

Reproduction steps:
1) open sandbox or some other form with select input with allowEmpty
2) try to change saved and select value to empty value
3) it will stay same

(it stays the same in PUT request)

djhi:
Reproduced. Thanks!

fzaninotto:
Fixed by #3758

PR #3758, opened 2019-10-01
[RFR] Fix Forms Remove Empty Values

### LINKED ISSUE #3551: Form ignores empty string values

djhi:
Supersedes #3656
Fixes #3551, #3660

### DIFF
--- a/packages/ra-core/src/form/sanitizeEmptyValues.ts
+++ b/packages/ra-core/src/form/sanitizeEmptyValues.ts
@@ -1,0 +1,33 @@
+import merge from 'lodash/merge';
+
+/**
+ * Because final-form removes undefined and empty string values completely
+ * (the key for the empty field is removed from the values), we have to check
+ * wether this value was initially provided so that it is correctly sent to
+ * the backend.
+ * See https://github.com/final-form/react-final-form/issues/130#issuecomment-493447888
+ *
+ * @param initialValues The initial values provided to the form
+ * @param values The current form values
+ */
+const sanitizeEmptyValues = (initialValues: object, values: object) => {
+    // For every field initialy provided, we check wether it value has been removed
+    // and set it explicitly to an empty string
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/form/sanitizeEmptyValues.ts
```diff
@@ -0,0 +1,33 @@
...
+    // For every field initialy provided, we check wether it value has been removed
+    // and set it explicitly to an empty string
+    const initialValuesWithEmptyFields = Object.keys(initialValues).reduce(
+        (acc, key) => {
+            if (typeof values[key] === 'object' && values[key] !== null) {
+                acc[key] = sanitizeEmptyValues(initialValues[key], values[key]);
+            } else {
+                acc[key] =
+                    typeof values[key] === 'undefined' ? '' : values[key];
+            }
```
pemoreau: As far as I understand you consider 3 cases:
- values[key] is undefined, in that case you initialize the value with ""
- values[key] is a non-null object, and you recursively sanitarize the values
- in all other cases, the value is kept as it is
I think you should also test that initialValues[key]!==null otherwise a recursive call may occurs and throw an error (Object.keys(null) is not defined)
Furthermore, if value[key]===nulll you keep the null value, is that the wanted behavior ?

PR #3760, opened 2019-10-01
Adds props to allow customizing the title and content of Confirm modals for delete buttons.

macklin-10x:
Adds props for customizing Confirm messages for delete actions.


fzaninotto:
I'd prefer that the main component accept a `confirm` prop that would default to `Confirm`. that way, users can customize more than the title and content (that they can already customize via translation).

macklin-10x:
Well, as far as I can tell, the message for *all* delete confirmations can be changed via translation, but it can't be customized for each different record type.  Please correct me if I'm mistaken.

How do you see the mechanics working with a `confirm` prop?  It seems like many/most of the props passed to `Confirm` involve the local state defined and managed by the main component.  It seems like the main things to be configured at all are the messages and translation options, but then the main component still passes in all of the callbacks?

fzaninotto:
You're right, most of the props currently passed to `Confirm` are computed, so it makes sense to do it as you suggested.

For BulkDeleteWithConfirm, the default `label` is passed as `defaultProps` instead of default parameter value. Would you mind making this consistent? 

I'll schedule your change for 3.1.

macklin-10x:
Fabulous!  Yes I will make that tweak.

macklin-10x:
I adjusted the defaults for the bulk delete and also added PropTypes for the new props which I had forgotten to do in the initial changeset.  Thanks for being so responsive and for maintaining a really useful framework!

macklin-10x:
One last question: do you have a rough guess as to when you think 3.1 will be released?

fzaninotto:
We give a rough estimate of the release dates in [the milestones page](https://github.com/marmelab/react-admin/milestones). V3.1 will be published in November.

fzaninotto:
Thanks!

PR #3781, opened 2019-10-06
Add redux-devtools-extension trace feature in development mode if available.

WiXSL:
Closes [#3658](https://github.com/marmelab/react-admin/issues/3658)

Creating a new PR identical to [#3661](https://github.com/marmelab/react-admin/pull/3661) but from a new branch of my fork, to avoid problems with my next branch.


djhi:
Thanks!

packages/ra-core/src/createAdminStore.ts
```diff
@@ -65,15 +65,21 @@ export default ({
...
 
+    const composeEnhancers =
+        (process.env.NODE_ENV === 'development' &&
+            typeof typedWindow !== 'undefined' &&
+            typedWindow.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ &&
+            typedWindow.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
+                trace: true,
+                traceLimit: 25,
+            })) ||
+        compose;
```
fzaninotto: I don't understand this code. It looks like you're changing the `compose` method but you're not, and using `compose` isn't even necessary since there is only one middleware.

WiXSL: Is a suggestion from [this post: [When browser doesn't have the redux devtools extension the app throws an undefined error](https://github.com/zalmoxisus/redux-devtools-extension/issues/320)

fzaninotto: ah, sorry, I misread the change:  `typedWindow.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__()` actually returns a `compose` function. Now it makes sense.

WiXSL: I've updated "CustomApp.md" file.

WiXSL:
I've updated "CustomApp.md" file.

docs/CustomApp.md
```diff
@@ -19,22 +19,44 @@ Here is the default store creation for react-admin:
...
+interface Params {
+    dataProvider: DataProvider;
+    history: History;
+    authProvider?: AuthProvider;
+    customReducers?: any;
+    customSagas?: any[];
+    i18nProvider?: I18nProvider;
+    initialState?: InitialState;
+    locale?: string;
+}
```
fzaninotto: The documentation is in JavaScript, not TypeScript

WiXSL: Ok, I think I got it right now.

WiXSL:
I've updated the docs as you requested.I hope is ok.
Let me know.

docs/CustomApp.md
```diff
@@ -52,21 +52,29 @@ export default ({
         );
     };
     const sagaMiddleware = createSagaMiddleware();
-
+    const typedWindow = window;
```
fzaninotto: no need for typedWindow in a JS world

WiXSL: Got it!

WiXSL:
Ok, I've updated the docs.

fzaninotto:
Thanks!

PR #3783, opened 2019-10-07
[RFR] Fix ReferenceInput Ignore its sort prop

### LINKED ISSUE #3642: alpha.4 ReferenceInputController ignores sort parameter
**What you were expecting:**
Setting the sort parameter in a ReferenceInput element should set the sort options forwarded to the server.

**What happened instead:**
The sort options are ignored.

**Steps to reproduce:**
Create a ReferenceInput with a sort parameter a look at the request sent to the server.

**Related code:**
```
      <ReferenceInput
        label="Away Team"
        source="thumbnailMetadata.awayTeamId"
        reference={ApiResources.teamInformation.name}
        sort={{ field: 'name', order: 'ASC' }}
        perPage={100}
      >
        <SelectInput optionText="name" />
      </ReferenceInput
```
l
**Other information:**
The problem occurs in `packages/ra-core/src/controller/input/useReferenceInputController.ts`. See line 79:

```
const useReferenceInputController = ({
    input,
    perPage = 25,
    filter = defaultFilter,
    reference,
    filterToQuery,
    referenceSource = defaultReferenceSource,
    resource,
    source,
    // sort, <-- should be here
}: Option): ReferenceInputValue => {

    const { pagination, setPagination } = usePaginationState({ perPage });
    const { sort, setSort } = useSortState(/* should set default sort state here */);
    
```

The sort option is ignored. I can set the sort order from the child element but I can't set the sort from the ReferenceInput.

**Environment**

* React-admin version: 3.0.0-alpha.4
* Last version that did not exhibit the issue (if applicable): Pre 3.0
* React version: N/A
* Browser: N/A

djhi:
Fixes #3642

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.ts
@@ -95,15 +95,6 @@ const useReferenceInputController = ({
     const { pagination, setPagination } = usePaginationState({ perPage });
     const { sort, setSort } = useSortState(sortOverride);
 
-    // We need to keep the current sort synchronized with the sort prop
-    // because the inner component may call the setSort too
-    useEffect(() => {
-        if (!isEqual(sort, sortOverride)) {
-            return;
-        }
-        setSort(sortOverride);
-    }, [setSort, sort, sortOverride]);
-
     const { filter: filterValue, setFilter } = useFilterState({
         permanentFilter: filter,
         filterToQuery,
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -84,12 +87,23 @@ const useReferenceInputController = ({
...
     const translate = useTranslate();
 
     const { pagination, setPagination } = usePaginationState({ perPage });
-    const { sort, setSort } = useSortState();
+    const { sort, setSort } = useSortState(sortOverride);
+
+    // We need to keep the current sort synchronized with the sort prop
+    // because the inner component may call the setSort too
+    useEffect(() => {
+        if (!isEqual(sort, sortOverride)) {
```
fzaninotto: why the `!`? I'd expect that the effect does nothing if the props are equal, and sets the sort if the props are different. That makes me think that you are probably missing a unit test.

Issue #1866, opened 2018-05-30
Global validation doesn't work in TabbedForm


zavadpe:
**What you were expecting:**
Global validation prevents creation of new item according set validation rules.

**What happened instead:**
Item could be created even in the situation when global validation rules were not met.

**Steps to reproduce:**
1. Set global validation function to `TabbedForm` with at least 2 tabs.
2. Add global and local validation rules, e.g. field not empty on first tab.
3. No validation rules on second tab.
4. Try to Create/Save item when on second tab.

**Related code:**
In `examples/demo/src/products/index.js`:
```
const validateProductCreate = (values) => {
  const errors = {};
  if (!values.image) {
      errors.image = ['image is required'];
  }
  ...
  return errors;
};

export const ProductCreate = withStyles(
    createStyles
)(({ classes, ...props }) => (
    <Create {...props}>
        <TabbedForm validate={validateProductCreate}>
  ...
```
Product can be created when on 'Description' tab.

**Environment**

* React-admin version: 2.0.2
* React version: 16.3.1
* Browser: any

fzaninotto:
Bug confirmed and reproduced

fzaninotto:
This seems to be caused by a bug in redux-form (see link above). Unfortunately, we can't circumvent it in react-admin... Except by rendering all tabs and hiding the ones not in focus using CSS.

PR #3785, opened 2019-10-08
Tabs performance improvement

hugofelp:
This PR attempts to improve rendering performance of tabbed forms by not rendering hidden tabs.

**case**
I have a large form, with expensive components and repeatable fields. Depending on the amount of data, the initial load and general performance of the form is poor. Besides optimising my own components, which I have done, I also split my form into smaller chunks using a TabbedForm, hoping fewer elements would be rendered. Unfortunately due to a few problems (like  https://github.com/marmelab/react-admin/issues/1956 and https://github.com/marmelab/react-admin/issues/1866) the TabbedForm renders all tabs at once, and re-renders all of them every time you switch tabs.

**solution**
It seems not rendering form elements causes issues with redux-form. So for hidden tabs we iterate over the provided input components and replace them with an empty shell from react-redux, passing along value and validation rules. That way the form is complete from redux-form perspective but not generating DOM elements.

I have tried this solution on my project and the performance gain is significant.


fzaninotto:
Hi, and thanks for your PR.

Not rendering hidden tabs prevents the validation of their inputs. Iterating over the children doesn't work for inputs decorated by another component, and we've repeatedly expressed that we want to prevent inspecting children at all costs (it's a bad mix up of responsibilities). So your solution would break other features. We can't accept that patch. 

For your use case, I recommend that you develop your own TabbedForm and use it instead of react-admin's. Every react-admin component can be replaced by your own when they don't actually match your need. that's the beauty of a component-based framework!

PR #3797, opened 2019-10-10
Fix background image is not random

tienvx:
`https://source.unsplash.com/random/1600x900/daily` always return the same image, while `https://source.unsplash.com/random/1600x900` return random image


fzaninotto:
Duplicate of #3069 and #2964.

The `daily` URL should return one new image per day. If it doesn't because of an unsplash intermittent bug, then we must change the simple example layout to use a custom background, otherwise it slows down the e2e tests a lot.

tienvx:
Thanks for the explaination

PR #3823, opened 2019-10-15
feat(file-input): use theme for dropzone background

despatates:
Background color for FileInput dropzone is hardcoded.
In dark theme, it still keeps a light background.

dropzone-input-before-dark

With this code, it will use color issued from the theme to be uniform with the rest of the design.

dropzone-input-after-dark

Of course, in light theme, it will use theme colors too.
Before:

dropzone-input-before-light

After:

dropzone-input-after-light


djhi:
Thanks!

Issue #3821, opened 2019-10-14
Invariant violation related to react-router / react-router-dom non-peer dependency


thclark:
###

First up. I'm known for concise and helpful issues. But this one gets a bit tortuous. Sorry about that. Hopefully the proposed solution is simple.

### Overview

I have an issue closely related to, but **not the same as** #3018 and its dupe #3019. It's caused by the way in which `react-router` and `react-router-dom` access their internal context (the authors are humming about a better way [here](https://github.com/ReactTraining/react-router/issues/6755) but that's too late to save my Monday :( .

#3078 is very closely related (almost certainly the same issue), but has been closed without identifying the root cause, and doesn't highlight the case where you actually import `react-router` or `react-router-dom` elsewhere outside `<Admin>`. It's left quite a few people hanging, not understanding why they can't have `react-router-dom` in their `package.json`.

**What you were expecting:**
To be able to use a `<Link>` component, from `react-router-dom`, in a page whose `<Route>` is outside of `<Admin>`, but in a valid router.

**What happened instead:**
I get a variety of errors similar to:
```
Invariant Violation: You should not use <Switch> outside a <Router>
```
even though things are inside valid routers.

**Steps to reproduce:**
This seems to be very much an edge case, dependent on npm's rules for resolving package dependencies, but I believe the following steps should reproduce:

- Add `react-router-dom` (I used 5.1.2) to your app's `package.json`
- Place your `<Admin .../>` component in a `<Route/>`, which itself resides in a `<Switch/>` inside a valid `<ConnectedRouter/>`.
  - *So far so good, this works. The router used by the Admin doesn't conflict with the outer Router even though they're nested, so you can happily use `Admin` alongside other stuff.*
- Render a `<Link>` in a different page outside the admin route.
  - *This now breaks, because to do this you have to install `react-router-dom` explicitly*

I'm sorry, to make this work on a prototype deadline I just ripped out all `<Link>`s and references to `react-router-dom` elsewhere on my site, so it's  chunk of work to go back. But I expect I'll encounter this again so the next time, I'll attach my `package-lock.json` to try and help show what's going on.

## In depth

In the case of #3018, there was a clash of version between `react-router` and `react-router-dom` - put simply, they couldn't see each other's routes. That was back when RA was using v4.x.x of RR and RRD.

[...]

PR #3825, opened 2019-10-16
Unified imports of routing components to react-router-dom

thclark:
As discussed in #3821, the import of routing components 

Not: This is a *partial* fix to #3821 in that it encourages users to import the `react-router-dom` library correctly (making occurrence of the issue less likely).

This PR does not address the other part of a fix, in which `react-router-dom` needs to be a peer dependency.

There's one difficulty here. `react-router`'s `StaticContext` is used in the typing system, but [it's not part of the public API](https://github.com/ReactTraining/react-router/issues/6733), and therefore not re-exported from `react-router-dom`. So I've had to leave that as being from `react-router`, and hope that doesn't defy the point of all this.

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -368,7 +368,7 @@ export default App;
 
 ## `customRoutes`
 
-To register your own routes, create a module returning a list of [react-router](https://github.com/ReactTraining/react-router) `<Route>` component:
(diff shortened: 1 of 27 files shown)


fzaninotto:
Thanks!

PR #3829, opened 2019-10-17
Fix JSDocs errors.

WiXSL:


packages/ra-core/src/dataProvider/Mutation.tsx
```diff
@@ -28,6 +28,7 @@ interface Props {
 /**
  * Get a callback to call the data provider and pass the result to a child function
  *
+ * @param {any} children
```
djhi: Missing a description ?

> {Function} Must be a function which will be called with the mutate callback

packages/ra-core/src/dataProvider/Query.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
 /**
  * Fetch the data provider and pass the result to a child function
  *
+ * @param {any} children
```
djhi: Same

packages/ra-core/src/sideEffect/admin.ts
```diff
@@ -11,6 +11,7 @@ import undo from './undo';
 
 /**
  * @param {Object} dataProvider A Data Provider function
+ * @param {Function} authProvider
```
djhi:
```suggestion
 * @param {Function} authProvider An Authentication Provider object
```

packages/ra-core/src/dataProvider/Query.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
 /**
  * Fetch the data provider and pass the result to a child function
  *
+ * @param {Function} children Must be a function which will be called with the mutate callback
```
djhi:
```suggestion
 * @param {Function} children Must be a function which will be called with an object containing the following keys: data, loading and error
```

djhi:
Thanks!

PR #3866, opened 2019-10-23
Fix property name for default values in upgrade guide 

straurob:
The property must not be named `initialValues` but `initialValue`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -949,7 +949,7 @@ const PostFilter = props =>
     </Filter>;
 ```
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
Are you sure? 

https://final-form.org/docs/react-final-form/types/FormProps#initialvalues

straurob:
> Are you sure?
> 
> https://final-form.org/docs/react-final-form/types/FormProps#initialvalues

Oh, man, I'm sorry... I didn't see that this is applies for the whole form. I was just looking at a single text input field.

We can delete this PR.

PR #3870, opened 2019-10-24
Fix useGetMany hook accumulatedIds filter function

heesienooi:
The issue:
ReferenceInput initialises an empty string value before any selection is made. This empty string value gets passed to useGetMany hook under the hood. And then it triggers the dataProvider `getMany` function and causes an error because the received params was an empty id string.

The fix:
Inside useGetMany hook, remove all falsey values in accumulatedIds. The current implementation just filters out null value only which is not sufficient.

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetMany.ts
+++ b/packages/ra-core/src/dataProvider/useGetMany.ts
@@ -153,11 +153,11 @@ const callQueries = debounce(() => {
         /**
          * Extract ids from queries, aggregate and deduplicate them
          *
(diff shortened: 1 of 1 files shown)


macrozone:
i also think null should be filtered, but we should make sure that Fields like SelectField or AutocompleteField sends always `null` when no value is selected and not an empty string like they do right now

fzaninotto:
any updates?

fzaninotto:
Superseded by #3989

PR #3871, opened 2019-10-24
upgrade redux-saga

skoppe:
Might be a good change to include in the upcoming 3.0.0 release.

It is a breaking change for users that currently use redux-saga in their projects. They would also need to upgrade.

Besides a few import renames there are no breaking changes in redux-saga 1.1.1


fzaninotto:
Thanks for your patch, but the `master` branch doesn't accept breaking changes. Besides, the change you're asking has already been made in the `next` branch, already released in 3.0.0-beta.

PR #3878, opened 2019-10-25
Update TabbedForm readme

tdeo:
As of material-ui v4, tabs are made scrollable by using `variant="scrollable"`
instead of `scrollable={true}`


djhi:
Thanks!

PR #3884, opened 2019-10-26
Fix error occur when use SimpleForm with ArrayInput

jinseoplee:
The issue:
Now destroyOnUnregister configuration of react-final-form has problem.
If set destroyOnUnregister to true on Form component of react-final-form, the following problems arise.

error

Of course I can use it wrapped, but I modified it myself cuz I wanted to use SimpleForm that the react-admin provides as it is.

The fixed:
Remove destroyOnUnregister configuration on SimpleForm.

success

please check and accept my pull request. :)

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.js
@@ -48,7 +48,6 @@ const SimpleForm = ({ initialValues, defaultValue, saving, ...props }) => {
             onSubmit={submit}
             mutators={{ ...arrayMutators }}
             keepDirtyOnReinitialize
-            destroyOnUnregister
             subscription={defaultSubscription}
             {...props}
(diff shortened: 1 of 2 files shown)


tdeo:
I was running into something similar inside a `TabbedForm`, should the same be done there ?

jinseoplee:
@tdeo yes right.
I checked again to whole project after looked at your comment.
As a result, they were used in SimpleForm and TabbedForm...
I just removed destroyOnUnregister on TabbedForm.

If you want to use TabbedForm or SimpleForm with ArrayInput right now, you can use it as follows until it is updated.

```
<SimpleForm destroyOnUnregister={false}></SimpleForm>

or

<TabbedForm destroyOnUnregister={false}></TabbedForm>
```

thanks for your comment!

fzaninotto:
Thanks!

PR #3889, opened 2019-10-28
[RFR] Fix DeleteWithConfirm throws exception

### LINKED ISSUE #3873: Cannot read property 'stopPropagation' of undefined when confirming record deletion
I am using react admin 3.0.0-beta.3 with dataProvider =  jsonServerProvider('http://jsonplaceholder.typicode.com'); 3.0.0-beta.3

I have a sample edit form for posts per react-admin tutorial (https://github.com/marmelab/react-admin/blob/next/docs/Tutorial.md) and added "undoable={false}" to edit form. This setting triggers confirmation prompt. When I click "confirm", I get an error "Cannot read property 'stopPropagation' of undefined". I also notice that the browser is javascript files. 

Attached screenprint shows files reloaded after I clicked confirm and the error message.
image

Source Code is attached.

[react-admin-app-3x.zip](https://github.com/marmelab/react-admin/files/3768781/react-admin-app-3x.zip)

fzaninotto:
closes #3873

### DIFF
--- a/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.js
+++ b/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.js
@@ -98,16 +98,12 @@ const DeleteWithConfirmButton = ({
         e.stopPropagation();
     };
 
-    const handleDelete = useCallback(
-        event => {
-            event.stopPropagation();
-            deleteOne();
-            if (typeof onClick === 'function') {
-                onClick();
-            }
-        },
-        [deleteOne, onClick]
-    );
+    const handleDelete = useCallback(() => {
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.js
```diff
@@ -98,16 +98,12 @@ const DeleteWithConfirmButton = ({
         e.stopPropagation();
     };
 
-    const handleDelete = useCallback(
-        event => {
-            event.stopPropagation();
```
fzaninotto: The event propagation is already stopped in `Confirm`

PR #3890, opened 2019-10-28
Autocomplete defaults to margin="dense" prop

tdeo:
For consistency with other inputs

Should it also default to not being fullWidth ? Or even leverage `ResttableTextInput` to have the same visual aspect ?

Here's a codesandbox showing both side by side: https://codesandbox.io/s/practical-chandrasekhar-869ep
image


djhi:
> Should it also default to not being fullWidth ? Or even leverage ResttableTextInput to have the same visual aspect ?

Yes & yes!

Thanks :)

tdeo:
I've actually never worked with Typescript before, and it turns out to be not that straightforward, can anyone pick this up ?

fzaninotto:
Thanks!

PR #3905, opened 2019-10-30
Fix ResettableTextField component endAdornment prop not being applied.

### LINKED ISSUE #3869: React-admin 3.0.0: Can't set resettable prop in SearchInput Component
Setting `resettable` prop to a `SearchInput` component doesn't have any effect.
The component always show the search icon on the end.

I set it to every `Filter` of this demo:
[https://codesandbox.io/s/react-admin-demo-300-0sygp](https://codesandbox.io/s/react-admin-demo-300-0sygp)

WiXSL:
Closes #3869


djhi:
Can you rebase ?

WiXSL:
I don't like the fact that the clear value icon changes the size of the Input in a SearchInput with resettable={true}.
I have no idea how to address that.

fzaninotto:
But now the tests fail... Can you take a look?

WiXSL:
I got it working!

packages/ra-ui-materialui/src/input/ResettableTextField.js
```diff
@@ -99,37 +101,50 @@ function ResettableTextField({
...
+                    ) : (
+                        endAdornment ||
+                        (resettable && (
+                            <InputAdornment
+                                position="end"
+                                classes={{
+                                    root: props.select ? selectAdornment : null,
+                                }}
+                            >
+                                <IconButton disabled />
```
fzaninotto: you're showing the reset button even though there is no content? I don't like that idea, I think it adds useless visual information.

WiXSL: It doesn't show anything, maintaining a `IconButton` prevents the input with from changing when `resettable` is  true and there is no value in the input.
If I'm not mistaken this was your code, I just added the disabled prop to avoid user interaction and the check to see if `resettable` is true.

WiXSL: I'm looking into it, you are right, plus on blur, no addorment is showing.
I'll pul something soon.

WiXSL: Done.
I don't know if is the cleanest way to  do it but it works.
Let me know.

WiXSL:
```jsx
<SearchInput source="q" alwaysOn resettable />
```
resettable

```jsx
<SearchInput source="q" alwaysOn resettable clearAlwaysVisible />
```
resettable-clearAlwaysVisible

```jsx
<SearchInput source="q" alwaysOn />
```
no-resettable

WiXSL:
@fzaninotto, do you think this is ok?

fzaninotto:
The `endAdornment` code needed even more work to acomodate another bug (#4069), so I had to redo the patch it in #4107, which supersedes this PR. Thanks for your work anyway.

PR #3907, opened 2019-10-30
[RFR] Add CoreAdminContext to ease custom apps

fzaninotto:
## Problem

Setting Resources dynamically is still very cumbersome: even if `<Admin>` accepts a function as child, this function can't execute hooks. So it's impossible, for instance, to have a dynamic list of resources based on a call to the `dataProvider` (since the `dataProvider` is only defined *after* the `Admin` component renders).

Many people use the Custom App approach just because of that, and they have to rewrite too much code.

## Solution

Separate the `CoreAdmin` component in two: 

1. The context innitialization
2. The Resources initialization

 That way, people wanting to have dynamic Resources can reuse the first component rather than rewrite it. 

## Example

```jsx
import React, { useEffect, useState } from 'react';
import { AdminContext, AdminUI, Resource, ListGuesser, useDataProvider } from 'react-admin';

const App = () =>  (
    <AdminContext dataProvider={myDataProvider}>
        <Resources/>
    </AdminContext>
);

const Resources = () => {
    const [resources, setResources] = useState([]);
    useEffect(() => {
        dataProvider.introspect().then(r => setResources(r));
    }, []);

    return (
        <AdminUI>
            {resources.map(resource => (
                <Resource name={resource.name} key={resource.key} list={ListGuesser} />
            ))}
        </AdminUI>
    );
}
```

### DIFF
--- a/packages/ra-core/src/dataProvider/Query.spec.tsx
+++ b/packages/ra-core/src/dataProvider/Query.spec.tsx
@@ -9,8 +9,7 @@ import {
 import expect from 'expect';
 
 import Query from './Query';
-import CoreAdmin from '../CoreAdmin';
-import Resource from '../Resource';
+import { CoreAdmin, Resource } from '../core';
 import renderWithRedux from '../util/renderWithRedux';
 import TestContext from '../util/TestContext';
 import DataProviderContext from './DataProviderContext';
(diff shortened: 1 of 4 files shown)


fzaninotto:
@kopax Can you elaborate on the `{...rest}`? I don't get it.

fzaninotto:
@kopax react-admin won't store arbitrary context, this can be done in userland in one line of code.

WiXSL:
Are these additions documented ?
I can't find them in the docs.

fzaninotto:
No, they're not. And you're right, they would probably simplify the custom apps documentation. Would you like to work on a documentation PR about that?

WiXSL:
Oh, that took me by surprise, hehe. 
Right now, I can't guarantee worcking on this. But i'll take it into account.

PR #3908, opened 2019-10-30
[RFR] Remove ra-realtime

fzaninotto:
The ra-realtime module is more a proof-of-concept. Trying to use it in 3.0 lead to a dead end. Is it broken? Is the documentation outdated? 

As sagas aren't the preferred way to set up side effects, the implementation is also not adapted to the v3 architecture. The right way to do it is to use a global EventListener and to interact with it from the `dataProvider`.

The v3 release is too close to refactor ra-realtime, so we'd better remove it.


kiptoomm:
hello, any chance to restore or re-implement this feature in the near future, if at all?

PR #3925, opened 2019-11-04
Adds a reference to a new data provider

jjgumucio:
I would like to add a reference to a data provider created to interact with [Moleculer Services](https://moleculer.services/) APIs.

Your feedback would be great!


fzaninotto:
Thanks!

PR #3932, opened 2019-11-05
Fix resettable SelectInput with value doesn't show options on click

tdeo:
Today, when a `SelectInput` is `resettable`, the end adorments prevents clicking on the select arrow on the right to open the dropdown (see https://codesandbox.io/s/intelligent-almeida-24cm2).

I also removed the padding on the right in order to avoid a zone where clicking isn't captured


fzaninotto:
Thanks!

PR #3938, opened 2019-11-06
Fix js vs jsx markdown styles in docs code.

WiXSL:


WiXSL:
May be I should't modify the Changelog even if there are typos since those are the commits descriptions.
Let me know.

fzaninotto:
Awesome, thanks!

Issue #3701, opened 2019-09-17
Style: React Admin components styles can't be overriden with Material UI theme


Claire-Cleany:
**Is your feature request related to a problem? Please describe.**
I would like to modify the appearance of my application using Material UI theme. For some rules, I'd like to use "overrides" key on React-Admin components, but it doesn't work. For instance, the Layout subcomponents appearance.

**Describe the solution you'd like**
I'd like the React-Admin components styles to be named when withStyles or makeStyles are used, like Material UI are naming their components with the second argument of withstyles/makeStyles.

**Describe alternatives you've considered**
Material UI tries to use Component.displayName when no style name is found, but none of the components on which makesStyles/withStyles was called seemed to have a displayName when breaking in makeStyles function.

fzaninotto:
hi @Claire-Cleany, and thanks for this report. 

What you ask is an enhancement, so if we ever implement it, it'll be in the `next` branch, using material-ui v4. This implies that every time we use `makeStyles`, we pass an `options` parameter:

```diff
const useStyles = makeStyles({
  foo: {
    textAlign: 'center',
  }
-});
+}, { name: 'FooComponent' });
```

It's a large change - it requires that we update *every* component. It can be done little by little, though.

Cf. https://github.com/mui-org/material-ui/blob/daac0562fc008a792356fc83956b5788c8893117/packages/material-ui-styles/src/makeStyles/makeStyles.js#L184

The react-admin core team won't work on that feature for the near future as we don't need it. Feel free to provide a PR if you want it to happen. If this issue receives no community contribution, it will be closed within a few months.

fzaninotto:
Fixed by #3946

PR #3946, opened 2019-11-06
Add name option to makeStyles

DanudeSandstorm:
I've begun adding the name of the component in the ra-ui-materialui components that use ``makeStyles`` as mentioned in #3701 .
Please let me know whether this is what is expected and I can perform this addition on the remaining components. 

Note: There's a difference in prettier/linting specification between ``.tsx`` and ``.js`` files, which results in differences in the makeStyles formatting.

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Logout.tsx
+++ b/packages/ra-ui-materialui/src/auth/Logout.tsx
@@ -13,12 +13,15 @@ interface Props {
     redirectTo?: string;
 }
 
-const useStyles = makeStyles((theme: Theme) => ({
-    menuItem: {
-        color: theme.palette.text.secondary,
-    },
-    icon: { minWidth: theme.spacing(5) },
-}));
+const useStyles = makeStyles(
+    (theme: Theme) => ({
+        menuItem: {
+            color: theme.palette.text.secondary,
+        },
+        icon: { minWidth: theme.spacing(5) },
+    }),
+    { name: 'LogoutWithRef' }
+);
 
 /**
  * Logout button component, to be passed to the Admin component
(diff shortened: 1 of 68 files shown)


packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -13,12 +13,15 @@ interface Props {
...
-    icon: { minWidth: theme.spacing(5) },
-}));
+const useStyles = makeStyles(
+    (theme: Theme) => ({
+        menuItem: {
+            color: theme.palette.text.secondary,
+        },
+        icon: { minWidth: theme.spacing(5) },
+    }),
+    { name: 'LogoutWithRef' }
```
fzaninotto: Why not Logout?

DanudeSandstorm: I was going by the export default. But I'll change it to the filename instead.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -21,13 +21,16 @@ const sanitizeRestProps = ({
...
-}));
+const useStyles = makeStyles(
+    theme => ({
+        content: {
+            paddingTop: theme.spacing(1),
+            paddingLeft: theme.spacing(2),
+            paddingRight: theme.spacing(2),
+        },
+    }),
+    { name: 'TabbedShow' }
```
fzaninotto:
```suggestion
    { name: 'TabbedShowLayout' }
```

WiXSL:
@DanudeSandstorm can you rebase?

DanudeSandstorm:
I rebased onto next

WiXSL:
Thank you.
Cause I need those changes for styling as well.

WiXSL:
@fzaninotto , is this PR ready to be merged?

fzaninotto:
Excellent, thanks!

PR #3948, opened 2019-11-07
EnhancedFilterForm: Respect classes property

MiMo42:
Problem: FilterForm not styleable
EnhancedFilterForm did not respect provided classes property.


fzaninotto:
Thanks!

PR #3959, opened 2019-11-08
Fix warnings caused by passing new FinalForm form state property 'dirty…

MiMo42:
> Warning: React does not recognize the `dirtyFieldsSinceLastSubmit` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `dirtyfieldssincelastsubmit` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

Theres a new form state related property 'dirtyFieldsSinceLastSubmit' in new FinalForm 4.18.6 (has been released yesterday). 
It is currently passed to form because it is not sanitized yet.
This PR sanitizes it in react-admin forms.


djhi:
Thanks!

Issue #3470, opened 2019-07-29
Datagrid: Can't set expand and optional checkboxes at the same time


WiXSL:
I need to use a Datagrid with expand attribute and also be able to show checkboxes only in cells that meet some criteria.

The Docs explain how to achieve the latter here:
[https://marmelab.com/react-admin/List.html#body-element](https://marmelab.com/react-admin/List.html#body-element)

But if you do this and set the "expand" attribute to a Datagrid, It renders incorrectly.
You can't have both characteristics.

fzaninotto:
Please follow the issue template and provide sample code, an example of what you get, and whet you expect instead. Even better, please add a link to a CodeSandbox allowing us to reproduce the problem.

WiXSL:
Ok.
Check out the "PostList" component here:
[https://codesandbox.io/s/tutorial-3mgjq?fontsize=14](url)

It has a Datagrid component with expand and a custom body:

`<Datagrid rowClick="expand" expand={<PostShow />} body={<MyDatagridBody />}>`

And it renders incorrectly. You can't expand the rows and the "select all" checkbox is miss-aligned with the rest of the rows.

WiXSL:
Well, now i see where the problem is.
The code of the custom DatagridRow doesn't  have the resolution of the expand attribute. I took the code from the Docs. Which now this will be a Docs problem report instead of a react-admin one.

Kmaschta:
I'm not sure to understand the issue. On your CodeSandbox, everything seems aligned to me.

![image](https://user-images.githubusercontent.com/1819833/62201336-7f58e580-b387-11e9-95a1-018a864c4c1d.png)

Do you have a screenshot?
Also, we really need the information that were asked by the template : what is your React Admin version and browser?

WiXSL:
@Kmaschta, you are seeing now the code with the solution I've found, which is overriding the entire DatagridRow prop plus the ExpandRowButton component and adding the "getHasCheckBox".
I couldn't clarify that I did that yesterday.

`<DatagridBody
    {...props}
    row={<MyDatagridRow getHasCheckBox={rec => !rec.selected} />}
  />`

fzaninotto:
What exactly is this `getHasCheckBox` prop you're talking about? Can you please clarify which component sets it and which component uses it?

WiXSL:
[...]

PR #3964, opened 2019-11-09
[RFR] Added canSelectRow function prop to Datagrid component.

WiXSL:
`DatagridRow` components of a `<Datagrid/>` will show a checkbox if `canSelectRow` prop returns true.
Solves the select all problem described here: https://github.com/marmelab/react-admin/issues/3470#issuecomment-551280562.

```jsx
<Datagrid
    {...props}
    canSelectRow={ record => record.id > 300 } 
>
...
</Datagrid>
```
chrome-capturee

### DIFF
--- a/packages/ra-ui-materialui/src/list/Datagrid.js
+++ b/packages/ra-ui-materialui/src/list/Datagrid.js
@@ -110,7 +110,7 @@ function Datagrid({ classes: classesOverride, ...props }) {
         setSort,
         size = 'small',
         total,
-        selectableRows = () => true,
+        canSelectRow = () => true,
         version,
         ...rest
     } = props;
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/Datagrid.js
```diff
@@ -110,6 +110,7 @@ function Datagrid({ classes: classesOverride, ...props }) {
         setSort,
         size = 'small',
         total,
+        selectableRows = () => true,
```
djhi: I think this name is confusing for a function as it does not contain a verb. Moreover, this function is called with a single row, so `Rows` is confusing too here. What about `canSelectRow` ?

WiXSL: You are right, i'll change it.

WiXSL: Done!

WiXSL:
I'm open to suggestions.

WiXSL:
@djhi did you have time to see this?

djhi:
No but I have some time today so I will :)

fzaninotto:
Thanks!

PR #3965, opened 2019-11-10
[v3] Fix pessimistic delete one does not remove item from selection

WiXSL:
Version 3.0.0 of #3926


fzaninotto:
We'll backport changes from master before releasing a new version of the next branch (as we always do). Replicating the change would make that difficult, so I prefer closing your PR. Thanks anyway.

PR #3968, opened 2019-11-11
Intuitive way of unselecting all selected items in BulkActionsToolbar

christiaanwesterbeek:
This is inspired by Google Photos that intuitively starts offering the unselect feature (top-left) once one photo or more is selected.

image

### DIFF
--- a/packages/ra-language-english/index.js
+++ b/packages/ra-language-english/index.js
@@ -5,7 +5,6 @@ module.exports = {
             add: 'Add',
             back: 'Go Back',
             bulk_actions: '1 item selected |||| %{smart_count} items selected',
-            unselect_items: 'Unselect item |||| Unselect %{smart_count} items'
             cancel: 'Cancel',
             clear_input_value: 'Clear value',
             clone: 'Clone',
@@ -23,6 +22,7 @@ module.exports = {
             show: 'Show',
             sort: 'Sort',
             undo: 'Undo',
+            unselect: 'Unselect'
             expand: 'Expand',
             close: 'Close',
         },
--- a/packages/ra-language-english/index.js
+++ b/packages/ra-language-english/index.js
@@ -22,7 +22,7 @@ module.exports = {
             show: 'Show',
             sort: 'Sort',
             undo: 'Undo',
-            unselect: 'Unselect'
+            unselect: 'Unselect',
             expand: 'Expand',
             close: 'Close',
         },
--- a/packages/ra-language-english/index.js
+++ b/packages/ra-language-english/index.js
@@ -5,6 +5,7 @@ module.exports = {
             add: 'Add',
             back: 'Go Back',
             bulk_actions: '1 item selected |||| %{smart_count} items selected',
+            unselect_items: 'Unselect item |||| Unselect %{smart_count} items'
             cancel: 'Cancel',
             clear_input_value: 'Clear value',
             clone: 'Clone',
--- a/packages/ra-language-french/index.js
+++ b/packages/ra-language-french/index.js
@@ -23,7 +23,7 @@ module.exports = {
             show: 'Afficher',
             sort: 'Trier',
             undo: 'Annuler',
-            unselect: 'Désélectionner'
+            unselect: 'Désélectionner',
             expand: 'Étendre',
             close: 'Fermer',
         },
--- a/packages/ra-language-french/index.js
+++ b/packages/ra-language-french/index.js
@@ -6,7 +6,6 @@ module.exports = {
             back: 'Retour',
             bulk_actions:
                 '%{smart_count} selectionné |||| %{smart_count} selectionnés',
-            unselect_items: 'Désélectionner %{smart_count}'
             cancel: 'Annuler',
             clear_input_value: 'Vider le champ',
             clone: 'Dupliquer',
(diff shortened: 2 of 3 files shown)


djhi:
Would mind adding a screenshot of the whole list with english language please ?

christiaanwesterbeek:
I don't have that integration readily available now. But what this PR basically adds, is a Cross Icon left of "n items selected" which unselects everything selected.

image

BTW: The export button on the right, is the BulkExportButton from https://github.com/christiaanwesterbeek/ra-extensions (a repo I recently started).

packages/ra-ui-materialui/src/list/BulkActionsToolbar.js
```diff
@@ -59,6 +63,14 @@ const BulkActionsToolbar = ({
             })}
             {...sanitizeListRestProps(rest)}
         >
+            <IconButton
```
fzaninotto: I find it very weird to propose to "unselect" *before* showing the number of selected items. I'd expect the contrary.

packages/ra-ui-materialui/src/list/BulkActionsToolbar.js
```diff
@@ -5,12 +5,15 @@ import Toolbar from '@material-ui/core/Toolbar';
...
 import { lighten } from '@material-ui/core/styles/colorManipulator';
+import IconButton from '@material-ui/core/IconButton';
+import CloseIcon from '@material-ui/icons/Close';
 import { useTranslate, sanitizeListRestProps } from 'ra-core';
 
 import TopToolbar from '../layout/TopToolbar';
 
 const useStyles = makeStyles(theme => ({
     toolbar: {
+        display: 'flex',
```
fzaninotto: Why do you need to do that? Can't you just add the "Unselect" link in the div of class title?

packages/ra-ui-materialui/src/list/BulkActionsToolbar.js
christiaanwesterbeek: It's the way of Google Photos. It makes sense to me personally. That's why I copied it.
image

packages/ra-ui-materialui/src/list/BulkActionsToolbar.js
christiaanwesterbeek: > Can't you just add the "Unselect" link in the div of class title?

I tried, but vertical alignment was off.

So I added `IconButton` as a sibling before the title div, but the effect was that the IconButton got 1/3 of the width of the entire toolbar, which I fixed with toolbar.display='flex' and title.flex=1 instead of title.flex='0 0 auto'.

packages/ra-language-english/index.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
             add: 'Add',
             back: 'Go Back',
             bulk_actions: '1 item selected |||| %{smart_count} items selected',
+            unselect_items: 'Unselect item |||| Unselect %{smart_count} items'
```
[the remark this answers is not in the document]
christiaanwesterbeek: "Unselect 3 items" is in the tooltip, not as text, so I guess that way it isn't feeling as repetitive. But we can change this text of course. Or just remove the tooltip, since the cross should make sense intuitively. It does to me at least.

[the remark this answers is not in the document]
djhi: No please keep it for accessibility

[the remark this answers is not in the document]
christiaanwesterbeek: > I think it's better to just use 'Unselect' in all cases.

Agreed. 'Unselect' without repeating the number in the Tooltip is fine.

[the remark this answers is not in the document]
djhi: btw an `aria-label` would be nice :)

packages/ra-ui-materialui/src/list/BulkActionsToolbar.js
fzaninotto: OK, I get it, thanks

fzaninotto:
Tests fail, needs rebase.

christiaanwesterbeek:
I rebased. Some tests still fail. At a glance that seemed unrelated to my changes.

packages/ra-language-english/index.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
             show: 'Show',
             sort: 'Sort',
             undo: 'Undo',
+            unselect: 'Unselect'
```
fzaninotto: tests fail because of a syntax error in this line (missing comma)

packages/ra-language-french/index.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
             show: 'Afficher',
             sort: 'Trier',
             undo: 'Annuler',
+            unselect: 'Désélectionner'
```
fzaninotto: and here, too

fzaninotto:
e2e tests fail

christiaanwesterbeek:
I'll have to get acquainted with testing react(-admin) components before I'll be able to fix that. Not sure when that will be.

WiXSL:
@christiaanwesterbeek,
I think the failing test is due to the fact that you now have two `selectAll` elements.
Try changing the following definition in file `react-admin\cypress\support\ListPage.js`:

```diff
    toggleSelectAll() {
-       cy.get(this.elements.selectAll).click();
+       cy.get(this.elements.selectAll).last().click();
    },
```
I haven't try it, just a thought given the message of the failing test.

fzaninotto:
This PR needs design tweaks, some work on the tests, and a rebase. So I'm closing it in favor of #4859.

PR #3971, opened 2019-11-12
[RFR][v3] useSuggestions - don't crash on empty value

natrim:
if default value is undefined it crashes on String.replace

> useSuggestions.js:77 Uncaught TypeError: Cannot read property 'replace' of undefined
>     at escapeRegExp (useSuggestions.js:77)
>     at useSuggestions.js:91
>     at useSuggestions.js:151
>     at Array.filter (<anonymous>)
>     at useSuggestions.js:150

### DIFF
--- a/packages/ra-core/src/form/useSuggestions.ts
+++ b/packages/ra-core/src/form/useSuggestions.ts
@@ -84,7 +84,8 @@ const useSuggestions = ({
 
 export default useSuggestions;
 
-const escapeRegExp = value => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+const escapeRegExp = value =>
+    value ? value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') : ''; // $& means the whole matched string
 
 interface Options extends UseChoicesOptions {
     choices: any[];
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/form/useSuggestions.ts
```diff
@@ -84,7 +84,8 @@ const useSuggestions = ({
 
 export default useSuggestions;
 
-const escapeRegExp = value => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+const escapeRegExp = value =>
+    value ? value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') : ''; // $& means the whole matched string
```
fzaninotto: will fail for value = 0. Would you mind testing if `typeof value === 'undefined'`?

natrim: well, value should be technically string, so any number would fail?

fzaninotto: I'm not sure I understand your point. If the user typed the string '0', than the escaped regex will be empty string, which is wrong - it should be '0'. 

Besides, I don't understand in which case the current code fails. Could you give an example of failing case, or, even better, a failing unit test?

natrim: but, string '0' will pass trough

the value ? value will only stop: undefined, null, false, ''     and replace them with ''

natrim:
@fzaninotto i added the test

fzaninotto:
Thanks!

PR #3972, opened 2019-11-12
[RFR][V3] TabbedFormTabs - validTabPaths fix index

natrim:
validTabPaths needs to contain all valid elements indexed by existence
else conditional tabs leave space and are not considered valid for active tab

problem was that Children.toArray(children) strips them before map so the indexes are different from the rendering Children.map

Nov-12-2019 10-04-11

repo: https://codesandbox.io/s/simple-i0xmz
1) edit post
2) try clicking tabs
3) you will see the index starting at 2
4) and comments not working right


fzaninotto:
Thanks!

PR #3974, opened 2019-11-12
[RFR] Fix RadioButtonGroupInput label size

### LINKED ISSUE #3955: [v3] RadioButtonGroupInput - label is different size than other labels
RadioButtonGroupInput's label size is different from other's label's ie. CheckboxGroupInput

label

repo: https://codesandbox.io/s/simple-i0xmz

steps:
1) edit post
2) see the difference

fzaninotto:
Before:

image

After: 
image

Closes #3955

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
@@ -23,7 +23,7 @@ const sanitizeRestProps = ({
 const useStyles = makeStyles(theme => ({
     root: {},
     label: {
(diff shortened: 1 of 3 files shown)


djhi:
I think this is actually an issue in material-ui. It looks like the label should be in the focused state by default in the guidelines: https://material.io/components/selection-controls/#radio-buttons. We should probably open an issue on material-ui

fzaninotto:
Not sure, we do the same kind of tailor-made label positioning for the CheckboxGroupInput.

djhi:
Yes, and I feel the same about it

PR #3975, opened 2019-11-12
SimpleForm / TabbedForm are missing PropTypes

### LINKED ISSUE #3969: [v3] SimpleForm / TabbedForm are missing PropTypes
SimpleForm and TabbedForm Components don't have propTypes defined.
SimpleFormView and TabbedFormView have them instead and is killing my intellisense ;-).
I'll add them myself, but i'm not sure which ones to include.

DanudeSandstorm:
Closes #3969

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.js
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.js
@@ -70,6 +70,20 @@ const defaultSubscription = {
     invalid: true,
 };
 
+SimpleForm.propTypes = {
+    defaultValue: PropTypes.oneOfType([PropTypes.object, PropTypes.func]), // @deprecated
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/form/SimpleForm.js
```diff
@@ -70,6 +70,20 @@ const defaultSubscription = {
     invalid: true,
 };
 
+SimpleForm.propTypes = {
+    defaultValue: PropTypes.oneOfType([PropTypes.object, PropTypes.func]), // @deprecated
+    initialValues: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+    save: PropTypes.func,
```
djhi: I'm not sure we should export those prop as they are injected by react-admin (save, saving, record and version)

fzaninotto:
Superseded by #3996

PR #3995, opened 2019-11-17
Change Class components to Function components.

WiXSL:


examples/demo/src/App.js
```diff
@@ -30,61 +30,61 @@ const i18nProvider = polyglotI18nProvider(locale => {
...
+            );
+        };
 
-    render() {
-        const { dataProvider } = this.state;
+        fetchDataProvider();
 
-        if (!dataProvider) {
-            return (
-                <div className="loader-container">
```
fzaninotto: I prefer the same loader for the data Generator waiting time and the data provider waiting time, so please don't use react-admin's Loading component in the demo.

WiXSL: Ok, done.

fzaninotto:
Thanks!

PR #4002, opened 2019-11-19
Change Tab and FormTab components to function components.

WiXSL:


packages/ra-ui-materialui/src/detail/Tab.js
```diff
@@ -58,8 +58,9 @@ const sanitizeRestProps = ({
  *     );
  *     export default App;
  */
-class Tab extends Component {
-    renderHeader = ({ className, label, icon, value, translate, ...rest }) => (
+const Tab = props => {
+    const translate = useTranslate();
+    const renderHeader = ({ className, label, icon, value, ...rest }) => (
```
fzaninotto: This function is parameters and variables in the closure. I think you should decide between a function without parameters (and dereferencing all props in the Tab component signature), or a full-fledged component (in which case no parameter can be defined outside, as you did with translate). I'd personally opt for the first option.

packages/ra-ui-materialui/src/form/FormTab.js
```diff
@@ -18,8 +18,9 @@ const sanitizeRestProps = ({
 
 const hiddenStyle = { display: 'none' };
 
-class FormTab extends Component {
-    renderHeader = ({ className, label, icon, value, translate, ...rest }) => {
+const FormTab = props => {
+    const translate = useTranslate();
+    const renderHeader = ({ className, label, icon, value, ...rest }) => {
```
fzaninotto: same remark

packages/ra-ui-materialui/src/detail/Tab.js
WiXSL: Ok, Done.

packages/ra-ui-materialui/src/form/FormTab.js
WiXSL: Ok, Done.

fzaninotto:
Thanks!

PR #4003, opened 2019-11-19
CommentShow component used in simple example instead of ShowGuesser.

WiXSL:


fzaninotto:
Thanks for the patch!

We use the simple example to test some of the features of react-admin - here, we use the Comment Show view to test the guessers. So I don't want to remove that from the simple example.

WiXSL:
Oh, ok, i didn't realise.

PR #4026, opened 2019-11-22
[RFR] Add allowDuplicates for AutocompleteArrayInput

alanpoulain:
Following https://github.com/marmelab/react-admin/pull/2912 and fixes https://github.com/marmelab/react-admin/issues/2311.

In React-admin 3, it's not possible to have duplicated values anymore for the `AutocompleteArrayInput`. However it could be useful to allow this behavior in some cases.

### DIFF
--- a/packages/ra-core/src/form/useSuggestions.ts
+++ b/packages/ra-core/src/form/useSuggestions.ts
@@ -6,6 +6,7 @@ import { useTranslate } from '../i18n';
 /*
  * Returns helper functions for suggestions handling.
  *
+ * @param allowDuplicates A boolean indicating whether a suggestion can be added several times
  * @param allowEmpty A boolean indicating whether an empty suggestion should be added
  * @param choices An array of available choices
  * @param emptyText The text to use for the empty suggestion. Defaults to an empty string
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/form/useSuggestions.ts
```diff
@@ -169,15 +174,15 @@ export const getSuggestionsFactory = ({
             // ignore the filter to show more choices
             suggestions = removeAlreadySelectedSuggestions(
                 choices,
-                selectedItem,
+                allowDuplicates ? null : selectedItem,
```
fzaninotto: if `allowDuplicates` is true, `removeAlreadySelectedSuggestions()` returns the input unchanged, so I'd prefer that you test it it's true on lone 175 and avoid calling `removeAlreadySelectedSuggestions` altogether in that case.

alanpoulain:
The related documentation has been updated. I've also added the `allowEmpty` prop, please tell me if you prefer to remove this change or have it in another PR (in the `master` branch I guess).

The code has been updated to follow your recommendation too.

And yes it replaces https://github.com/marmelab/react-admin/pull/2912.

fzaninotto:
Thanks!

PR #4050, opened 2019-11-26
Fix console errors in tests

fzaninotto:
I removed almost all errors in unit tests. I don't understand the last one, that we can leave for now. 

image


djhi:
> I removed almost all errors in unit tests. I don't understand the last one, that we can leave for now.

I think the last one occurs because the dataProvider used in TestContext  does not return anything

PR #4052, opened 2019-11-26
Fix styles for <Ready /> component

nkoshell:


fzaninotto:
Two questions:

1. What's wrong with the current mobile design?

image

2. Why do you care about this page on Mobile? I see no situation where a developer would use a mobile browser as their main browser for the initial phases of react-admin development.

nkoshell:
1. Check this page on real mobile device.
2. All right, just for your information.

PR #4055, opened 2019-11-27
[RFR] Fix AutocompleteInput

### LINKED ISSUE #4040: Default width for AutocompleteInput is different from the rest of the inputs
[The problem is described here.](https://github.com/marmelab/react-admin/issues/4011)

djhi:
Fixes #4040

Many classes weren't used at all so I removed them.

Besides, while testing this on the simple example, I noticed that `SimpleFormIterator` was adding a `div` because of the `TransitionGroup`. This `div` was inserted as the child of `ul` which breaks semantics and accessibility

image

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -8,7 +8,6 @@ import React, {
     isValidElement,
 } from 'react';
 import Downshift, { DownshiftProps } from 'downshift';
-import classNames from 'classnames';
 import get from 'lodash/get';
 import { makeStyles, TextField } from '@material-ui/core';
 import { TextFieldProps } from '@material-ui/core/TextField';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -328,13 +332,6 @@ const AutocompleteInput: FunctionComponent<
                             name={input.name}
                             InputProps={{
                                 inputRef: storeInputRef,
-                                classes: {
-                                    root: classNames(classes.inputRoot, {
```
fzaninotto: this was used to support both filled and standard variants. I'm afraid this change breaks that. Can you make sure the form displays correctly with `<SimpleForm variant="standard">`?

djhi: Confirmed. It works by setting the variant on `SimpleForm`

PR #4064, opened 2019-11-27
Added documentation for canRowSelect prop.

WiXSL:


docs/List.md
```diff
@@ -906,6 +907,20 @@ const PostList = props => (
     </List>
 )
 ```
+
+### `canRowSelect`
+
+You can customize which rows will show a selection checkbox using the `canRowSelect` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows a checkbox only for rows with an id grather than 300:
```
fzaninotto:
```suggestion
You can customize which rows will show a selection checkbox using the `canRowSelect` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows a checkbox only for rows with an id greather than 300:
```

docs/List.md
```diff
@@ -906,6 +907,20 @@ const PostList = props => (
     </List>
 )
 ```
+
+### `canRowSelect`
+
+You can customize which rows will show a selection checkbox using the `canRowSelect` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows a checkbox only for rows with an id greather than 300:
```
fzaninotto:
```suggestion
You can customize which rows will show a selection checkbox using the `canRowSelect` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows a checkbox only for rows with an id greater than 300:
```

WiXSL:
Done!
I change it to `isRowSelectable`.

fzaninotto:
Thanks!

PR #4071, opened 2019-11-28
Add types for button components

fzaninotto:
In the process of making the code more robust at compile time, I converted most Button components to TypeScript. 

It took me no less than 6 hours. Let it be said that I don't like TypeScript.


packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -38,15 +38,17 @@ const CreateButton = ({
...
+    );
     return isSmall ? (
         <Fab
             component={Link}
             color="primary"
             className={classnames(classes.floating, className)}
             to={`${basePath}/create`}
             aria-label={label && translate(label)}
-            {...rest}
+            {...rest as any}
```
fzaninotto: I had to disable type checking here because of a material-ui bug: `Button` doesn't recognize the `component` prop :/

See https://github.com/mui-org/material-ui/issues/15827

fzaninotto:
to improve readability for non-ts readers, I chose to move types to the end of each file. I did the same for secondary functions, so that a component file starts by the component definition.

packages/ra-ui-materialui/src/button/CloneButton.tsx
```diff
@@ -44,18 +35,39 @@ export const CloneButton = ({
...
+    handleSubmit,
+    handleSubmitWithRedirect,
+    invalid,
+    pristine,
+    saving,
+    submitOnEnter,
+    ...rest
+}: any) => rest;
+
+interface Props {
```
djhi: We should probably extract the two first properties in a reusable interface for all our components

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -147,14 +126,66 @@ const DeleteWithConfirmButton = ({
...
+                '@media (hover: none)': {
+                    backgroundColor: 'transparent',
+                },
+            },
+        },
+    }),
+    { name: 'RaDeleteWithConfirmButton' }
+);
+
+interface Props {
```
djhi: Lots of duplication with DeleteButton. Couldn't we improve that ?

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.tsx
```diff
@@ -120,12 +88,65 @@ const DeleteWithUndoButton = ({
...
+                '@media (hover: none)': {
+                    backgroundColor: 'transparent',
+                },
+            },
+        },
+    }),
+    { name: 'RaDeleteWithUndoButton' }
+);
+
+interface Props {
```
djhi: Same

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -81,78 +150,16 @@ interface Props {
         ) => Promise<any>,
         dataProvider: DataProvider
     ) => Promise<void>;
-    sort: Sort;
     filter?: any;
+    icon?: JSX.Element;
+    label: string;
     maxResults: number;
```
djhi: isn't maxResults optional ?

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -81,78 +150,16 @@ interface Props {
...
-    resource: string;
     onClick?: (e: Event) => void;
-    label: string;
-    icon?: JSX.Element;
-    basePath: string;
+    resource: string;
```
djhi: Optional ?

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -81,78 +150,16 @@ interface Props {
...
+    sort: Sort;
```
djhi: Optional ?

packages/ra-ui-materialui/src/button/CloneButton.tsx
fzaninotto: I think the code is hard enough to read like that, I prefer a bit of duplication to a lot of indirection.

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
fzaninotto: That'd imply extracting the common part to another file (to avoid circular dependencies), and that would make all the Delete components harder to read... Again, I prefer some duplication for clarity.

PR #4075, opened 2019-11-28
[RFR] Allow to extend useEditController

Kmaschta:
`useCreateController` accepts a onSuccess and a onFailure props, but `useEditController` don't
So, I made this PR to have the same behavior for both of them


fzaninotto:
Cool. You should PR against `next` rather than `master`.

Kmaschta:
Done :+1:

fzaninotto:
Thanks!

PR #4077, opened 2019-11-28
fix(dependencies): adding missing query-string dependency

kopax:
Used within `ra-ui-materialui` package as for example in https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/button/CloneButton.js#L6


packages/ra-ui-materialui/package-lock.json
```diff
@@ -0,0 +1,87 @@
+{
```
fzaninotto: please don't commit the packages.json

packages/ra-ui-materialui/package.json
```diff
@@ -75,6 +75,7 @@
         "jsonexport": "^2.4.1",
         "lodash": "~4.17.5",
         "prop-types": "^15.6.1",
+        "query-string": "~5.1.1",
```
fzaninotto: please relax the version constraint with a `^` instead of `~` to allow npm to avoid duplicate packages.

packages/ra-ui-materialui/package-lock.json
kopax: In my memory, this project required to commit it. Am I wrong or did it change?

fzaninotto:
Thanks!

PR #4078, opened 2019-11-28
fix(dependencies): added missing dev dependencies

kopax:
`peerDependencies` are not automatically installed in development.

This incorrect configuration would normally lead to `react` and `react-dom` import not being solved in development.

### DIFF
--- a/packages/react-admin/package-lock.json
+++ b/packages/react-admin/package-lock.json
@@ -1,0 +1,79 @@
+{
+    "name": "react-admin",
+    "version": "3.0.2",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "js-tokens": {
+            "version": "4.0.0",
+            "resolved": "https://npm.kopaxgroup.com:443/js-tokens/-/js-tokens-4.0.0.tgz",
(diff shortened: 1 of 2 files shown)


fzaninotto:
I don't see how this can be a bug: `ra-core` already has a react and react-dom as `devDependency`, so there should be no case where these deps are missing.

kopax:
You are right. I don't how `ra-ui-materialui` can be used without `ra-core`.  But `ra-core` if installed from `npm` does not bring `react` and `react-dom`, only `react-admin` does. This is more conventional (note the "would normally")

fzaninotto:
> But ra-core if installed from npm does not bring react and react-dom

I don't understand why. React is both in devDeps and peerDeps in ra-core, that should do the trick. Do you actually experience a bug?

kopax:
What I meant is `npm install ra-core --save` does not install devDeps and peerDeps. Only react-admin does. I do not experience bug because you want to install manually the devdeps. This is not a bug just a convention (meaning a reference should exist because an import of react exist in ra-ui-materialui, meaning a version is attached to ra-ui-materialui, the same as ra-core) and you don't need to follow as the dependency of ra-core is always installed and present with it's devDeps in development, it won't cause bug or error.

PR #4081, opened 2019-11-29
Migrate the Demo to Typescript

djhi:
This will help us validate/debug our typings.


examples/demo/src/react-admin.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-admin';
```
fzaninotto: what's that for?

examples/demo/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
fzaninotto: what's that for?

examples/demo/src/react-admin.d.ts
djhi: Because TS complain about react-admin not having types

examples/demo/src/react-app-env.d.ts
djhi: Added by CRA automatically

WiXSL:
Shouldn't "source-map-explorer" be in devDependencies?

PR #4085, opened 2019-11-29
fix(LoadingIndicator): remove unused isLoading props (https://material-ui.com/api/circular-progress/)

kopax:
Please double-check as I may be doing a mistake.

I haven't been able to use `isLoading`, according to `<CircularProgress />` [documentation](https://material-ui.com/api/circular-progress/), the `props.isLoading` is not used by mui nor `<LoadingComponent />`

### DIFF
--- a/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
+++ b/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
@@ -36,7 +36,6 @@ export const LoadingIndicator = ({
 LoadingIndicator.propTypes = {
     classes: PropTypes.object,
     className: PropTypes.string,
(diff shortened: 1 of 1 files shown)


kopax:
While this PR is being checked, just a side note question:

I am upgrading @bootstrap-styled/ra-ui for 3.0.2 and I am trying to work in `examples/demo` while editing `package/**`. I have started the server with `make run-demo` but it seems after a while, change does not get refreshed and I need to stop the server and `make build && make run-demo`. Any tought about this?

fzaninotto:
Thanks! As for the demo, there is no hot reloading for the changes in the react-admin packages - only for the changes made in the demo source.

Issue #4073, opened 2019-11-28
Authprovider getPermissions error


liviozanol:
**What you were expecting:**
AuthProvider authentication using OpenID should work correctly in react admin version 3.0.1. 

**What happened instead:**
React-Admin returning "TypeError: Cannot read property 'then' of undefined" on ra-core/esm/auth/usePermissions.js.

Specific line: https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/auth/usePermissions.ts#L53

**Steps to reproduce:**
I think that if you just try to use authprovider the problem will occur. If not, let me know, so that I can provide more info/code.

**Related code:**
https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/auth/usePermissions.ts#L53

**Other information:**
React Admin is a GREAT product, thank you so much.

I'm not even near an expert on react/react-admin, so be patience.

I have read the upgrade guide, and we don't use match, context or resource. Also, we don't use AUTH_GET_PERMISSIONS (don't use permissions at all). I think the issue is related to "The authProvider No Longer Receives Default Parameters" in the upgrade, but can't confirm. https://github.com/marmelab/react-admin/commit/9730e6c72a171a4d92ccb562e26c8a24394cec08#diff-2786406d872b3d7bd9273d26f8b1d303

If this is not a bug, forgive me.

**Environment**

* React-admin version: 3.0.1
* Last version that did not exhibit the issue (if applicable): 2.9.9
* React version: ^16.12.0
* Browser: chrome 78.0.3904.108
* Stack trace (in case of a JS error):

djhi:
Thanks for reporting this. If you are able to illustrate the bug or feature request with an example, please provide a sample application via one of the following means:

* CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple)
* A sample application via GitHub

liviozanol:
@djhi Sorry. Since I'm using a private openId provider, I can't simulate the issue on CodeSandBox or GitHub. I can however provide some parts of our RA code.

Will this help?

fzaninotto:
> we don't use AUTH_GET_PERMISSIONS (don't use permissions at all)

Whether or not you use permissions, your authProvider must return a Promise for the `AUTH_GET_PERMISSIONS` call (can simply be `Promise.resolve([])`).

liviozanol:
@fzaninotto Thank you very much. It worked!

[...]

PR #4088, opened 2019-11-29
Fix missing mention about AUTH_GET_PERMISSIONS in Upgrade guide

fzaninotto:
Refs #4073

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -489,7 +489,7 @@ export default (type, params) => {
 
 ## The `authProvider` Must Handle Permissions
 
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -487,6 +487,41 @@ export default (type, params) => {
 }
 ```
 
+## The `authProvider` Must Handle Permissions
+
+React-admin calls the `authProvider` to get the permissions for each page - using the `AUTH_GET_PERMISSIONS` verb. While in 2.x, implementing this `AUTH_GET_PERMISSIONS` verb was optional, it becomes compulsory in 3.0 as soon as you provide a custom `authProvider`. You can simply return a Promise resolved to an empty array to ignore permissions handling.
```
djhi: I think you can return a resolved promise with nothing in it (`undefined`) as well. I'm afraid mentioning an empty array might confuse our users about what permissions are. Indeed, the result of this call may be a string, an object, an array, anything actually

PR #4089, opened 2019-11-30
added FileInputPreview and ImageInputPreview to export

oleg-andreyev:


fzaninotto:
Why do you need that? What problem does it solve?

oleg-andreyev:
@fzaninotto trying to create form with drag'n'drop file upload and later create "preview" of uploaded files which are associcated with my object.

fzaninotto:
You can already pass a custom Field as child of FileInput. What would your solution allow that a custom Field doesn't allow?

oleg-andreyev:
I want to have RemoveCircle and as far as I see only `FileInputPreview` has it

CreateForm

```
<ImageInput accept="image/*" source="images" multiple={true} resource={props.resource}>
  <ImageField source="src" title="title"/>
</ImageInput>
```

EditForm

```
<ImageInputPreview label="Image" source="image_id" reference="images">
  <ImageField ...>
</ImageInputPreview>
```

fzaninotto:
ImageInputPreview is an alias for FileInputPreview. So whatever you can do with FileInput, you can do with ImageInput. Image previews do have a remove icon (which shows on hover).

so again, I don't understand what you're trying to achive.

oleg-andreyev:
I'm trying to import it.

```
import {withDataProvider, ImageInput, ImageField, ImageInputPreview} from 'react-admin';
import {withDataProvider, ImageInput, ImageField, FileInputPreview} from 'react-admin';
```

```
./src/PlaceEdit.js
Attempted import error: 'ImageInputPreview' is not exported from 'react-admin'.
Attempted import error: 'FileInputPreview' is not exported from 'react-admin'.
```

To make it working I have to import component directly from ra-materialui

```
import ImageInputPreview from 'ra-ui-materialui/esm/input/ImageInputPreview';
```

fzaninotto:
Ok, we're getting nowhere here, either you don't understand my questions or I don't know how to ask them.

Anyway, you want to use an internal component and you should not need to. You can already customize an ImageInput preview by passing a child. If you want more customization, write your own component.

oleg-andreyev:
@fzaninotto I did not know it was internal (haven't noticed it's usage in FileInput component). Thanks!

It may be stupid but how about marking internal components with `@internal` for better DX?

fzaninotto:
We simply don't export or document internal components.

PR #4098, opened 2019-12-02
Convert component definitions to variable holding arrow function.

WiXSL:


fzaninotto:
Not sure we want that. the current code allows stack trace to show significant function names rather than anonymous functions.

WiXSL:
ok, I'll close it

PR #4109, opened 2019-12-03
Add <BulkExportButton> component

fzaninotto:
- [x] Move export logic to ra-core
- [x] Use context instead of props to get exporter
- [x] Add `<BulkExportButton>` component
- [x] Use it in `<PostList>` in simple example
- [x] Add doc

Capture d’écran de 2019-12-03 16-11-59


packages/ra-ui-materialui/src/list/List.js
```diff
@@ -128,44 +130,47 @@ export const ListView = props => {
     const controllerProps = getListControllerProps(rest);
 
     return (
-        <div
-            className={classnames('list-page', classes.root, className)}
-            {...sanitizeRestProps(rest)}
-        >
-            <Title title={title} defaultTitle={defaultTitle} />
+        <ExporterContext.Provider value={exporter}>
```
fzaninotto: If we keep adding stuff to the List context (like filters, selected ids, etc), I wonder if we shouldn't use a ListContext instead... opinions?

PR #4113, opened 2019-12-05
[RFR] Invite to create the first record

### LINKED ISSUE #2639: Empty List Invite
If there is no result, and if there is no active filter, and if the resource has a Create page, the List page should display a special content inviting the user to create their first element.

localhost_8080_

Inspirations:

- http://emptystat.es/
- https://uxplanet.org/empty-state-mobile-app-nice-to-have-essential-f11c29f01f3
- https://twitter.com/steveschoger/status/1058398981888376832

m4theushw:
Closes #2639

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -28,7 +28,7 @@ Here are all the props accepted by the `<List>` component:
 * [`filterDefaultValues`](#filter-default-values) (the default values for `alwaysOn` filters)
 * [`pagination`](#pagination)
 * [`aside`](#aside-component)
-* [`emptyState`](#empty-state)
+* [`empty`](#empty-page)
 
 Here is the minimal code necessary to display a list of posts:
 
@@ -627,16 +627,18 @@ const Aside = ({ data, ids }) => (
 ```
 {% endraw %}
 
-### Empty state
+### Empty page
 
-When there is no result, and there is no active filter, and the resource has a create page then a special page inviting the user to create the first record is displayed. Use the `emptyState` prop to modify that page, passing your custom component:
+When there is no result, and there is no active filter, and the resource has a create page, react-admin displays a special page inviting the user to create the first record.
+
+You can use the `empty` prop to replace that page by a custom component:
 
 {% raw %}
 ```jsx
 import Button from '@material-ui/core/Button';
 import { CreateButton, List } from 'react-admin';
 
-const EmptyState = ({ basePath }) => (
+const Empty = ({ basePath, resource }) => (
     <div style={{ textAlign: 'center', margin: '1em' }}>
         <h1>No products available</h1>
         <div>Create one or import from a file</div>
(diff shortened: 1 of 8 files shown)


docs/List.md
```diff
@@ -683,6 +685,49 @@ const Aside = ({ data, ids }) => (
...
+
+You can use the `empty` prop to replace that page by a custom component:
+
+{% raw %}
+```jsx
+import Button from '@material-ui/core/Button';
+import { CreateButton, List } from 'react-admin';
+
+const Empty = ({ basePath, resource }) => (
+    <div style={{ textAlign: 'center', margin: '1em' }}>
```
fzaninotto: We try and avoid inline styles in the documentation. Try using `<Box>`, or just remove the inline style for the example.

packages/ra-ui-materialui/src/list/Empty.js
```diff
@@ -0,0 +1,51 @@
...
+import Inbox from '@material-ui/icons/Inbox';
+import { makeStyles } from '@material-ui/styles';
+import { useTranslate } from 'ra-core';
+import { CreateButton } from '../button';
+import inflection from 'inflection';
+
+const useStyles = makeStyles({
+    message: {
+        textAlign: 'center',
+        fontFamily: 'Roboto, sans-serif',
```
fzaninotto: Why do you need that? As it's in the layout, it should already have the right font.

packages/ra-ui-materialui/src/list/Empty.js
```diff
@@ -0,0 +1,51 @@
...
+            _: inflection.pluralize(resource),
+        }),
+        true
+    );
+
+    return (
+        <>
+            <div className={classes.message}>
+                <Inbox className={classes.icon} />
+                <h1>{translate('ra.page.empty', { name: resourceName })}</h1>
```
fzaninotto: use `<Typography>` instead of `<h1>` to get the right style

packages/ra-ui-materialui/src/list/Empty.js
m4theushw: The hardcoded font is because I used `<NotFound />` as inspiration. Now it's not needed anymore.

m4theushw:
PR rebased

fzaninotto:
sorry, a recent PR merge modified the List doc and requires a new rebase for this PR

m4theushw:
Ready!

docs/List.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: "The List View"
+title: 'The List View'
```
fzaninotto: Too many things have changed in this file. Can you reduce the changes to what you added?

packages/ra-language-french/index.js
```diff
@@ -40,6 +40,8 @@ module.exports = {
             loading: 'Chargement',
             not_found: 'Page manquante',
             show: '%{name} #%{id}',
+            empty: 'Pas encore de %{name}.',
+            invite: 'Voulez-vous en créer un ?',
```
fzaninotto: This will cause bad wording when the resource name is feminine (ex: promotion). But I don't see an easy way to fix that, so we'll go with that translation.

docs/List.md
m4theushw: Sorry, vscode was running prettier on docs files

packages/ra-language-french/index.js
m4theushw: We could leave to user to override the empty text, by looking for a `resources.foobar.empty` key. If there's no such key we use the default one `ra.page.empty`. 

The invite key could also be customizable, like `resources.foobar.invite`.

fzaninotto: Great idea. Would you mind adding that feature to your PR?

m4theushw: I'll work on that

PR #4116, opened 2019-12-05
Add the ability to override form layouts in a simpler way

fzaninotto:
- [x] Add ability to customize `Edit`, `Create`, and `Show` root component
- [x] Make `handleSubmitWithRedirect` optional in Toolbar 
- [x] Create `FormWithRedirect` component in `ra-core` to refactor common form code and make  custom form layouts easier
- [x] Document the way to build a custom form layout
- [x] Use the previous changes to improve use of screen real estate in demo/Customers/Edit view
- [x] Fix demo/Customer/Create form to match the Edit one

**before**

Capture d’écran de 2019-12-05 17-16-48

**after**

Capture d’écran de 2019-12-05 16-52-09


fzaninotto:
Back to WIP, I'll do the layout of the forms with `Box` instead of divs with `useStyle`. In fact, I kind of fell in love with material-ui's `Box`.

fzaninotto:
And we're back to RFR

fzaninotto:
After a good night's sleep, I think we can do a better work than inviting people to copy a large bunch of code. Back to WIP.

fzaninotto:
And Ready For Review

docs/CreateEdit.md
```diff
@@ -192,6 +193,21 @@ const PostEdit = props => (
...
+Some form layouts also use `Card`, in which case the user ends up seeing a card inside a card, which is bad UI. To avoid that, you can override the main form container by passing a `component` prop:
+
+```jsx
+const PostEdit = props => (
+    <Edit component="div" {...props}>
+        ...
+    </Edit>
+```
+
+The default value for the `component` prop is `Card`.
```
djhi: An exemple of passing a react component might be useful for beginners here so that they don't pass `component="Card"`

docs/CreateEdit.md
```diff
@@ -928,6 +944,141 @@ export const UserEdit = props => {
...
+                </Toolbar>
+            </form>
+        )}
+    />
+);
+```
+
+This custom form layout component uses the `FormWithRedirect` component, which wraps react-final-form's `Form` component to handle redirection logic. It also uses react-admin's `<SaveButton>` and a `<DeleteButton>`.
+
+**Tip**: When `Input` components have a `resource` prop, they use it to determine the input label. `<SimpleForm>` and `<TabbedForm>` inject this `resource` prop to `Input` components automatically. When you use a custom form layout, pass the `resource` prop manually - unless the `Input` has a `label` prop.
```
djhi: We really should use react Context instead :)

djhi: I know, v4 ^^

fzaninotto: indeed ^^

arkkanoid:
Hi, Just updating to v3.1.0:

./~/ra-core/esm/form/FormWithRedirect.js
Module parse failed: /usr/src/app/node_modules/ra-core/esm/form/FormWithRedirect.js Unexpected token (53:313)
You may need an appropriate loader to handle this file type.

fzaninotto:
Weird, the transpilation didn't transform `{ ...arrayMutators }` properly in this file, because it's a default value for a prop... I'll fix it and republish shortly.

Thanks for the report.

fzaninotto:
Out of curiosity, how did you get this error (webpack/parce, babel settings, browser)? I can't reproduce it on any browser on the react-admin demo, even IE11.

arkkanoid:
I upgraded RA from v2.9, after installing the new packages and run yarn start I get this error

fzaninotto:
FYI this is a bug in TypeScript (https://github.com/microsoft/TypeScript/issues/35771) triggered by a useless convention in react-final-form (https://github.com/final-form/react-final-form/issues/704), but we'll have to fix it in the react-admin code...

PR #4132, opened 2019-12-10
Make pagination more extendable

### LINKED ISSUE #4008: Can't theme pagination actions toolbar.
The `Pagination` component uses a `PaginationActions` component which renders the numbers/prev/next  items and is the only way to pass css classes, but is inaccessible outside of the `Pagination` component and the classes used there are hardcoded. 

The documentation suggests coding another component for the case of wanting a “<previous - next>” like toolbar but nothing that allows to style what is being render.

```jsx
...

const Pagination = ({
    loading,
    page,
    perPage,
    rowsPerPageOptions,
    total,
    setPage,
    setPerPage,
    ...rest
}) => {

...

return (
        <TablePagination
            ...
            ActionsComponent={PaginationActions}
            ...
        />
    );
...
```

### LINKED ISSUE #4131: Allow to customize the Pagination component
In order to change the color (primary or secondary) of the pagination buttons, in the datagrid, we have to clone the whole components `Pagination` and `PaginationActions`.

1. In the `PaginationActions` component, allow to overwrite the button props
2. In the `Pagination` component, do not use directly the PaginationAction component, but pass it as a default prop `actions = <PaginateActions />`

Today, in order to customize the pagination, we're forced to clone these components entirely.

Kmaschta:
Closes #4131
Closes #4008

### DIFF
--- a/packages/ra-ui-materialui/src/list/Pagination.js
+++ b/packages/ra-ui-materialui/src/list/Pagination.js
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import { TablePagination, Toolbar, useMediaQuery } from '@material-ui/core';
 import { useTranslate, sanitizeListRestProps } from 'ra-core';
 
-import PaginationActions from './PaginationActions';
-import PaginationLimit from './PaginationLimit';
+import DefaultPaginationActions from './PaginationActions';
+import DefaultPaginationLimit from './PaginationLimit';
 
 const emptyArray = [];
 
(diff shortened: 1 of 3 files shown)


WiXSL:
Thank you. This also closes #4008

packages/ra-ui-materialui/src/list/Pagination.js
```diff
@@ -16,6 +16,8 @@ const Pagination = ({
     total,
     setPage,
     setPerPage,
+    ActionsComponent,
```
fzaninotto: not sure about this name. I think in other places (or in material-ui), component props start with lowercase and don't include the term "component"

Kmaschta: You're right: list, create, show etc, are component (and not elements) yet their props are lowercase and without component at the end.
I submitted the change

Issue #4136, opened 2019-12-10
ImageInput onDrop callback not executed in v3


red1:
**What you were expecting:**

* ImageInput onDrop callback passed in the options should be fired and executed

**What happened instead:**
* The callback does not work

**Steps to reproduce:**
* Add a onDrop callback to any ImageInput
* try to upload an image
* the callback "onDrop" is not executed
 
**Related code:**

* CodeSandbox Example
* https://codesandbox.io/s/wizardly-haze-knyxy?fontsize=14&hidenavigation=1&module=%2Fsrc%2Fposts%2FPostEdit.js&theme=dark

```
const callback = () => {
  console.log("callback onDrop is not Working/not fired :(");
};
....
<ImageInput multiple  source="pictures" accept="image/*" options={{ onDrop: callback }}>
   <ImageField source="src" title="title" />
</ImageInput>
     
```

**Other information:**

**Environment**

* React-admin version: 3.0.0+ 
* Last version that did not exhibit the issue (if applicable): 2.9.x
* React version: 16.9.0
* Browser: All
* Stack trace (in case of a JS error): None

Kmaschta:
Reproduced! Thanks

Kmaschta:
React Admin overwrites the onDrop argument:

https://github.com/marmelab/react-admin/blob/93f653a88223958635df7df7a247644fb409c841/packages/ra-ui-materialui/src/input/FileInput.tsx#L149

A PR can be made to call `options.onDrop` if it exists on the React Admin's `onDrop`

red1:
thanks @Kmaschta

PR #4137, opened 2019-12-10
Note that ImageInput does not support anymore the onDrop callback

red1:
Note that ImageInput does not support anymore the onDrop callback


Kmaschta:
The onDrop not called is a bug from React Admin not a documentation default, see #4136

Kmaschta:
Superseded by #4140

PR #4139, opened 2019-12-10
Better prop sanitation for reference inputs

### LINKED ISSUE #4133: Console warning when using alwaysOn ReferenceInput in Filter component
**What you were expecting:**
No console warnings

**What happened instead:**
Console warning:
`Warning: React does not recognize the `alwaysOn` prop on a DOM element.`
It hit me after set alwaysOn prop on a ReferenceInput within a Filter component.

**Steps to reproduce:**
- Open modified ra simple example at https://codesandbox.io/s/simple-ciybl
- Switch to Comments
- See Warning in console

**Related code:**
https://codesandbox.io/s/simple-ciybl

**Other information:**
I only had the time to do some short investigation. 
It looks like alwaysOn prop is actually _not_ sanitized, although there is a dedicated sanitizeRestProps in FilterFormInput. The alwaysOn flag is transferred to children of FilterFormInput and applied to DOM elements further down.

**Environment**

* React-admin version: 3.0.3 
* React version: 16.12.0
* Browser: Chrome

Kmaschta:
Closes #4133

### DIFF
--- a/packages/ra-ui-materialui/src/input/ReferenceArrayInput.js
+++ b/packages/ra-ui-materialui/src/input/ReferenceArrayInput.js
@@ -6,6 +6,7 @@ import {
     useTranslate,
 } from 'ra-core';
 
+import sanitizeInputProps from './sanitizeRestProps';
 import LinearProgress from '../layout/LinearProgress';
 import Labeled from '../input/Labeled';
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks!

PR #4153, opened 2019-12-12
Incorrect useStyles exports.

WiXSL:


WiXSL:
I think is not the idea to export them.
Correct me if i am wrong.

PR #4157, opened 2019-12-12
docs: add control for `icon` in Custom Menu

m4nu56:
add a simple control for existing `resource.icon` in the resource before using it into `react.createElement()`


ThieryMichel:
It is done in the code, no need to overcomplicate the documentation example with this.

PR #4173, opened 2019-12-16
Fix duplicate query-string package

### LINKED ISSUE #1762: Failed to minify on build
I followed the tutorial and created app with create-react-app.
But when I build with `npm run build` i have this error:
```
Failed to minify the code from this file:

        ./node_modules/query-string/index.js:42

Read more here: http://bit.ly/2tRViJ9
```
**Environment**

* React-admin version: 2.0.0-RC1
* React version: 16.3.2

fzaninotto:
Refs #4081
Closes #1762

### DIFF
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -72,7 +72,7 @@
         "inflection": "~1.12.0",
         "lodash": "~4.17.5",
         "prop-types": "^15.6.1",
-        "query-string": "~6.9.0",
+        "query-string": "^5.1.1",
         "recompose": "~0.26.0",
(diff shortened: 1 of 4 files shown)


packages/ra-core/package.json
```diff
@@ -72,7 +72,7 @@
         "inflection": "~1.12.0",
         "lodash": "~4.17.5",
         "prop-types": "^15.6.1",
-        "query-string": "~6.9.0",
+        "query-string": "^5.1.1",
```
djhi: Why are we using an old version ?

djhi: nvm

Issue #4164, opened 2019-12-13
Resource aggregator for react-admin


dryhten:
Hey guys, I wrote a package that can be used in addition to react-admin v2 to aggregate resources. I was not sure how to contact you about this so I'm opening this issue.

The package is here https://github.com/dryhten/ra-resource-aggregator/

On the npm registry here https://www.npmjs.com/package/ra-resource-aggregator

It's still in its early stages, I'll probably update it a bit over the next few days. Maybe if others will find it useful you guys could add it to [miscellaneous](https://marmelab.com/react-admin/Ecosystem.html#miscellaneous)?

My use case is this: I have a PostgreSQL database and I want a non tech friendly admin over it. So I decided to go with `react-admin` + `ra-data-hasura` + hasura graphql server + postgresql. The only issues I had were that a lot of the resources/forms were highly customized and were supposed to change/create/delete data from multiple tables, so I created a small aggregator package that should help me get the job done faster.

Let me know what you think!

fzaninotto:
Feel free to open a PR to edit the Ecosystem doc adding a link to your package.

PR #4177, opened 2019-12-16
Add resource aggregator to docs

dryhten:
This was discussed here https://github.com/marmelab/react-admin/issues/4164


fzaninotto:
Thanks!

Issue #3244, opened 2019-05-17
Error message not shown on API error response for undoable actions with handled side effects.


tusmenko:
**What you were expecting:**
As mentioned in **Optimistic Rendering and Undo** part of documentation:
> If the fetch fails, react-admin shows an error notification, and forces a refresh, too.

Implementing Component that used basic (UPDATE in my case) actions using `withDataProvider`, and setting `onSuccess` and `onError` side effects handlers I expect to see corresponding error and success messages from API\dataProvider.

**What happened instead:**
Only message defined in `onSuccess` appears on click, "optimistically". 
But no Error message shown on API response with error, even if list view updates and optimistically rendered value falls back to initial state.

**Steps to reproduce:**
- Go to sandbox example
- On Posts tab click the custom Status switcher
-> Switch optimistically toggles it's value.
- After mocked dataProvider throws HttpError with error message, switch toggles back.
-> NO error message shown.

**Related code:**
https://codesandbox.io/s/47t8x

Consider my code modification on:
1) src/posts/StatusSwitch.js - custom Toggle switch component.
2) src/dataProvider.js - exception throwing on posts UPDATE.
3) src/posts/PostList.js - StatusSwitch added to row.

Custom switch powered withDataProvider configured with following options:

```js
dataProvider(
      UPDATE,
      resource,
      {
        id: record.id,
        data: updatedRecord
      },
      {
        undoable: true,
        onSuccess: {
          notification: { body: `${label} has changed` }
        },
        onError: {
          notification: {
            body: `Error: ${label} has not changed`,
            level: "warning"
          }
        }
      }
    );
```

**Environment**

* React-admin version: 2.8.6
* React version:16.8.6
* Browser:Chrome

Kmaschta:
You're right, I see two problems with your CodeSandbox :

1. After optimistic update, the real request doesn't show a error notification (but reverts the change, that is ok)
2. When clicking again on status toggle, the state doesn't change. I suspect an agressive anti re-render rule on that

I'll let core maintainers check that bug.
Note : Open the CodeSandbox on full screen to see the status toggle.

alexrosabr:
We had a similar problem here: 

1. in the List view, click Edit
2. in the Edit view, edit a field that appears in the List view and Save
3. the List view is updated with the change, 
4. after the undo delay, the API is called and returns HTTP error
5. the error message from API is correctly shown 
6. the List view is NOT updated to reverse the change

It worked on 2.9.x

[...]

PR #4179, opened 2019-12-17
Fix the Refresh of Lists when an optimistic update fails

### LINKED ISSUE #4176: List does not refresh after API error (video included)
After the "Save" returns an API error, the rejected value is reverted in the List

The API returned error and the List still shows the rejected value.

Video below.

**Environment**

* React-admin version: 3.0.3
* React version: 16.12
* Browser: Chrome 78

React Admin 2.9
ezgif-ra29

React Admin 3.0
ezgif-ra30

m4theushw:
Fixes #4176

### DIFF
--- a/cypress/integration/edit.js
+++ b/cypress/integration/edit.js
@@ -1,10 +1,12 @@
 import createPageFactory from '../support/CreatePage';
 import editPageFactory from '../support/EditPage';
+import listPageFactory from '../support/ListPage';
 
 describe('Edit Page', () => {
     const EditPostPage = editPageFactory('/#/posts/5');
     const CreatePostPage = createPageFactory('/#/posts/create');
     const EditCommentPage = editPageFactory('/#/comments/5');
+    const ListPagePosts = listPageFactory('/#/posts');
 
     describe('Title', () => {
         it('should show the correct title in the appBar', () => {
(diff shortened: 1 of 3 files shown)


fzaninotto:
You must also remove the call to the `refresh()` side effect in `useEditController` and `useShowController`, otherwise the refresh will occur twice.

m4theushw:
@fzaninotto I'm calling `refresh` only on undoable operations. I think that is not the case of `useEditController` and `useShowController` since they refresh when `getOne` fails. Am I right?

fzaninotto:
Right, but then with your patch there will still be two refreshes in case of optimistic error. 

So the right way to fix this bug is to add a refresh as onFailure side effect in the save function of useEditController and useCreateController.

m4theushw:
I didn't add a refresh in `useCreateController` because it's not undoable and this bug only occurs with optimistic saving.

JulienMattiussi:
Closes #3244

fzaninotto:
Thanks!

fzaninotto:
hm, this PR, once merged to master, breaks the e2e tests. I'll take a look.

fzaninotto:
Fixed in b23212028eb975edcf9e7e6caadfdc17bf874c66

PR #4185, opened 2019-12-18
Fix: getPermissionsWithoutProvider is not defined

### LINKED ISSUE #4184: ReferenceError: getPermissionsWithoutProvider is not defined
Because `getPermissionsWithoutProvider` is initialized with a var, it can't be placed below `useGetPermissions`.

Either, change `getPermissionsWithoutProvider` to function OR move it before `useGetPermissions`

**Environment**

* React-admin version: 3.0.4
* React version: 16.12.0
* Browser: Chrome
* Stack trace (in case of a JS error): `ReferenceError: getPermissionsWithoutProvider is not defined`

kopax:
Because getPermissionsWithoutProvider is initialized with a var, it can't be placed below useGetPermissions.

Fix #4184

### DIFF
--- a/packages/ra-core/src/auth/useGetPermissions.ts
+++ b/packages/ra-core/src/auth/useGetPermissions.ts
@@ -45,7 +45,6 @@ const useGetPermissions = (): GetPermissions => {
     return authProvider ? getPermissions : getPermissionsWithoutProvider;
 };
 
-
 /**
  * Proxy for calling authProvider.getPermissions()
  *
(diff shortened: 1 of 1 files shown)


djhi:
Sorry missed this one and @fzaninotto fixed it in https://github.com/marmelab/react-admin/issues/4184. Thanks anyway!

kopax:
No problem, you are welcome. 

By any chance, I know it is not the place but since we are into permissions, I am stuck in a v2 to v3 upgrade and I have two weird [issues](https://stackoverflow.com/questions/59408349/upgrade-react-admin-v2-to-v3-permissions-is-always-empty-and-context-is-not-con) while I was wrapping the `layout` like with a `WithPermissions`, 
 `permissions` is always `undefined` no matter if `authProvider.getPermissions()` always return an non empty array of perms.

It was working fine in v2, still investigating but that's hard to catch without knowing what change.

PR #4202, opened 2019-12-22
Set default redux state for desktop sidebar

justinr1234:
This allows redux-persist to function correctly. Otherwise, the state is being forced overridden by react-admin.

It would be better to do a dynamic media query against the active `theme` coming from MUI, but this is a more simple solution that assumes `md` is always the default.

### DIFF
--- a/packages/ra-core/src/reducer/admin/ui.ts
+++ b/packages/ra-core/src/reducer/admin/ui.ts
@@ -26,8 +26,14 @@ export interface UIState {
     readonly viewVersion: number;
 }
 
+let isDesktop = false;
+
+if (typeof window !== 'undefined') {
+    isDesktop = window.matchMedia('(min-width: 960px)').matches;
+}
+
 const defaultState: UIState = {
-    sidebarOpen: false,
+    sidebarOpen: isDesktop,
     optimistic: false,
     viewVersion: 0,
 };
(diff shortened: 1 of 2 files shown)


ThieryMichel:
Sorry, but test fails.

fzaninotto:
I don't understand why your patch makes redux-persist work better. Can you describe the problem? Also, you've patched the isDesktop and I don't know why. Again, please describe the issue it fixes (and a failing test case proving the issue).

justinr1234:
Sure thing, I’ll fix up tests and add something new.

The issue it’s solving is that the code the way it’s written now will overwrite anything that’s saved to local storage by redux persist. So in my app I want to save the settings the user has so that when they reload the page the sidebar remains either open or closed, depending what they wanted to do. If you dispatch a redux action from the component lifecycle to change the value, this will happen after redux persist has rehydrates the state. So what you end up with is:

(In desktop)
1) state closed (default)
2) store rehydrated by redux persist (can be any value based on what the user had)
3) the react admin component sets the state to open, overwriting anything.

Since there is no way to change the behavior of RA, we have no reliable way (without getting hacky) to have redux persist win.

What my change is aiming to do is to set the initial state of the store based on a media query. This means step (1) won’t always be “closed” on desktop but would instead be “open”. Then after that step (2) can run with redux persist setting the value to whatever it wants.

justinr1234:
@fzaninotto that make sense?

fzaninotto:
Why don't you use the `initialState` prop of the `<Admin>` component to set the initial state based on local storage when the app loads?

justinr1234:
@fzaninotto definitely makes sense, my biggest concern is that the redux store is being initialized to the wrong state and then imperatively changed after component mount. This can be avoided by just initializing the state to the proper default from the beginning. It seems the reason this imperative code was added, with the comment to find a better way later, was because it was a temporary fix. It seems to make sense to change the default behavior of react admin to be correct in that it sets the initial state properly based on screen size instead of starting wrongly then imperatively changing it later.

fzaninotto:
Superseded by #4677

PR #4210, opened 2019-12-24
[Feature] added rootSagaMiddleware to have an ability pass preconfigured saga-middleware

oleg-andreyev:
### Feature
Ability to register custom middlewares

#### Context
I want to connect Sentry.io as logger to my frontend.
In order to do that I need pass pre-configured middleware https://rafaelfragoso.com/catching-errors-with-redux-sagas-and-sentry/ with onError/and logger


ThieryMichel:
Sorry, but test fails. Also can you base this PR on the next branch since this is a new feature. Thanks.

fzaninotto:
We won't support adding a custom middleware through yet another parameter of the `<Admin>` component. The right way to do it is to provide your own Redux context outside of `<Admin>`, as described in the [Including in another app](https://marmelab.com/react-admin/CustomApp.html) documentation.

PR #4216, opened 2019-12-26
Fix remove props warnings in CheckboxGroupInput 

SeanBE:
- Dropping the required flag since a default value is provided when destructuring. Proptypes complains when component is child of a `ReferenceArrayInput` component.
- Pull out loaded attribute from `rest` to avoid fieldet from complaining. 
`index.js:1375 Warning: Received `true` for a non-boolean attribute "loaded".`


fzaninotto:
In fact, I think the `loaded` props should be sanitized before passing props to children. So the right fix should be the opposite: add the `loaded` prop to the `sanitizeRestProps` of ReferenceArrayInput, and remove it from the SelectArrayInput. 

Would you mind working on that patch?

PR #4223, opened 2019-12-27
[RFR] rewrite RichTextInput to TypeScript

tlaziuk:
Hi, I found that you need help with rewriting the project to TypeScript.

I'm not sure about removing the `prop-types` dependency, but the TypeScript is here so IMO it is not needed anymore, leave a comment if you have opposite opinion so I'll restore it.

Instead of default values in object spread I've added `defaultProps`, also - the full object of props is passed to `useStyles` hook.

From minor things - I had to cast the styles (*Stylsheet file) object to `any` because apparently the object is of invalid type (eg. numeric string is used instead of number) - but changing that may be considered as a breaking change because treating the values depends on used JSS engine.


packages/ra-input-rich-text/src/QuillBubbleStylesheet.ts
```diff
@@ -808,4 +810,4 @@ export default {
             visibility: 'visible',
         },
     },
-};
+} as any) as StyleRules;
```
ThieryMichel: You can remove the any here and replace the numeric string by number.

packages/ra-input-rich-text/src/QuillSnowStylesheet.ts
```diff
@@ -797,4 +799,4 @@ export default {
     '.ql-container.ql-snow': {
         border: '1px solid #ccc',
     },
-};
+} as any) as StyleRules;
```
ThieryMichel: Same: You can remove the any here and replace the numeric string by number.

tlaziuk:
@ThieryMichel done

packages/ra-input-rich-text/src/index.tsx
```diff
@@ -1,34 +1,68 @@
...
-    resource,
-    variant,
-    margin = 'dense',
-    ...rest
-}) => {
-    const classes = useStyles();
-    const quillInstance = useRef();
-    const divRef = useRef();
-    const editor = useRef();
+interface IProps {
```
fzaninotto: We iue the name "Props" in the react-admin codebase for the props interface. Can you remove the leading 'I' in your interface?

fzaninotto:
Thanks! I forgot to ask that you PR against the `next` branch rather than `master`. `master` is only for bug fixes and we want to minimize the changes of compatibility breakage there.

packages/ra-input-rich-text/src/index.tsx
```diff
@@ -123,20 +157,13 @@ const RichTextInput = ({
     );
 };
 
-RichTextInput.propTypes = {
```
fzaninotto: need to keep the `propTypes`. Compile-time type validation doesn't replace runtime type checking (unfortunately). Would you mind adding these back?

packages/ra-input-rich-text/src/index.tsx
```diff
@@ -123,20 +157,13 @@ const RichTextInput = ({
...
-        PropTypes.array,
-        PropTypes.bool,
-        PropTypes.shape({
-            container: PropTypes.array,
-            handlers: PropTypes.object,
-        }),
-    ]),
-    fullWidth: PropTypes.bool,
-    configureQuill: PropTypes.func,
+RichTextInput.defaultProps = {
```
fzaninotto: Please keep these as default values for the props instead, React is deprecating `defaultProps` for function components (https://github.com/facebook/react/pull/16210).

tlaziuk: :+1:

packages/ra-input-rich-text/src/index.tsx
tlaziuk: :+1:

packages/ra-input-rich-text/src/index.tsx
```diff
@@ -125,14 +160,6 @@ RichTextInput.propTypes = {
     label: PropTypes.string,
     options: PropTypes.object,
     source: PropTypes.string,
-    toolbar: PropTypes.oneOfType([
```
tlaziuk: I removed that because `Quill` should validate it's options itself, IMO it's a good practice because leaving that breaks the composition.

fzaninotto:
Thanks!

PR #4230, opened 2020-01-01
Add Hasura GraphQL Provider in DataProviders list

cpursley:


fzaninotto:
I don't understand: there was already on Hasura dataProvider. Why did you add another one? Listing two in the docs can only confuse the reader, so we'll only list one. Unless you provide a good reason to keep yours instead of the previous one, I'll revert that PR.

cpursley:
The existing hasura data provider is very limited and only has access to the v1 endpoint which is intended for administration as described here: 

https://github.com/hasura/ra-data-hasura/issues/13

This data provider in the pr utilizes the full GraphQL capabilities of Hasura.

fzaninotto:
Then why didn't you PR in the existing Hasura data provider instead? 

Again, we want to limit the number of options for react-admin developers, and we don't want to replace the GitHub search engine. If we have to make a choice, we'll take the one with the most contributors and the longest track record - i.e. not yours.

cpursley:
The existing provider takes an entirely different approach (it's not using graphql and very limited in it's capabilities).

PR #4236, opened 2020-01-03
Add helperText prop to ArrayInput

### LINKED ISSUE #4227: ArrayInput does not support helperText prop
**What you were expecting:**
As ArrayInput should accept the common input props, helperText should be working, too.

**What happened instead:**
HelperText prop is not handled in the input, therefore does not appear.

**Related code:**
I'd love to create a pull request, however, I am not sure how the related files are compiled. Making the prop work would only need a few lines of code into ArrayInput.tsx:

```
import {  InputLabel, FormControl, FormHelperText } from '@material-ui/core';
...
<InputLabel ... />
<FormHelperText>{helperText}</FormHelperText>
```
Besides this, adding helperText const to the props and the proptypes.

**Environment**

* React-admin version: 3.1.1

netdown:
Fixes #4227 

I was not able test - I am not sure of building -, sorry about that, but as it is only a 5 line change, there should be no problem.


packages/ra-ui-materialui/src/input/ArrayInput.tsx
```diff
@@ -85,6 +86,8 @@ export const ArrayInput = ({
                     isRequired={isRequired(validate)}
                 />
             </InputLabel>
+            {helperText &&
```
djhi: We might have validation applied to the `ArrayInput` itself, in which case the `HelperText` should display the error instead of the helper text like the other inputs do. You can find a very close implementation in `SelectArrayInput`.

netdown: Updated, however I am not sure of the props useInput requires, do all inputs receive them?

djhi:
Looks good. However you'll have to pass the `error` to the `InputLabel` too

packages/ra-ui-materialui/src/input/ArrayInput.tsx
```diff
@@ -85,6 +109,15 @@ export const ArrayInput = ({
                     isRequired={isRequired(validate)}
```
djhi: `isRequired` is already a boolean

packages/ra-ui-materialui/src/input/ArrayInput.tsx
```diff
@@ -69,6 +76,23 @@ export const ArrayInput = ({
         validate: sanitizedValidate,
         ...rest,
     });
+    
+    const {
+        id,
+        input,
+        isRequired,
+        meta: { error, touched },
+    } = useInput({
```
djhi: You don't need `useInput`, `useFieldArray` returns a `meta` which contains the `error` if any. See https://github.com/final-form/react-final-form-arrays#fieldarrayrenderprops

packages/ra-ui-materialui/src/input/ArrayInput.tsx
djhi: Ignore previous comment. As you'll not be using `useInput`, you'll have to call `isRequired(validate)`. Just not the `isRequired` from the `useInput` but the function imported

fzaninotto:
As it's a new feature, please PR against `next` rather than `master`

netdown:
> Looks good. However you'll have to pass the `error` to the `InputLabel` too for it to become red on error

Neither component passes that prop.

djhi:
> Neither component passes that prop

See https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx#L215

PS: This was merged a few hours ago (#4249)

netdown:
@fzaninotto Are you sure it should be marked as a new feature? The documentation suggests that this prop is also available on ArrayInput, therefore I'd consider it as a fix.

fzaninotto:
The current documentation in wrong, then ;). Yes, it's a new feature - it was never implemented.

netdown:
@djhi Any suggestions on CI failure? The error prop is of type any in useInput and final form: https://final-form.org/docs/react-final-form/types/FieldRenderProps#metaerror

djhi:
First, can you please rebase your branch on `next`. The diff is currently very hard to follow

fzaninotto:
Superseded by #4262

PR #4238, opened 2020-01-03
fix ra-data-graphql-simple convert SCALAR Date type to ISOString

### LINKED ISSUE #4218: DateTimeInput converts to empty Object in ra-data-graphql-simple provider
**What you were expecting:**
Dateinput should return the correct datetime string

**What happened instead:**
Variables that are connected to this input are {} in graphql mutation

**Steps to reproduce:**
Install ra-data-graphql-simple use any form that contains DateTimeInput and submit, network inspect the graphql request

* React-admin version:"^3.1.1"
* React version: "^16.12.0"
* Browser: Chrome

ThieryMichel:
fix #4218

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildVariables.js
+++ b/packages/ra-data-graphql-simple/src/buildVariables.js
@@ -35,6 +35,9 @@ const castType = (value, type) => {
         case 'SCALAR:Boolean':
             return Boolean(value);
 
+        case 'SCALAR:Date':
+            return typeof value === 'string' ? value : value.toISOString();
+
         default:
             return value;
     }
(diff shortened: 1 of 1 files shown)


packages/ra-data-graphql-simple/src/buildVariables.js
```diff
@@ -35,6 +35,9 @@ const castType = (value, type) => {
         case 'SCALAR:Boolean':
             return Boolean(value);
 
+        case 'SCALAR:Date':
```
djhi: Are you aware that people might name this scalar whatever they want? It is not supported out of the box by GraphQL and might be named `DateTime`, `IsoDate` or `Smurf` for what we know. I think this should be documented that this **example** implementation of a graphql data provider expect it to be named `Date` and be an ISO date.

ThieryMichel: Maybe my fix is not good then. We need to serialize date object to ISO to construct the query. Otherwise it get serialized to {} and the info is lost.

PR #4240, opened 2020-01-03
Remove required location property

CarsonF:
Update remaining places where location is passed in to ignore/handle it but use `useLocation` hook instead.

### DIFF
--- a/packages/ra-core/src/controller/useCreateController.ts
+++ b/packages/ra-core/src/controller/useCreateController.ts
@@ -3,7 +3,7 @@ import { useCallback } from 'react';
 import inflection from 'inflection';
 import { parse } from 'query-string';
 import { Location } from 'history';
-import { match as Match } from 'react-router-dom';
+import { match as Match, useLocation } from 'react-router-dom';
 
 import { useCheckMinimumRequiredProps } from './checkMinimumRequiredProps';
 import { useCreate } from '../dataProvider';
@@ -39,8 +39,8 @@ export interface CreateProps {
     hasEdit?: boolean;
     hasList?: boolean;
     hasShow?: boolean;
-    location: Location;
-    match: Match;
+    location?: Location;
+    match?: Match;
     record?: Partial<Record>;
     resource: string;
     successMessage?: string;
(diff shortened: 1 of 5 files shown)


fzaninotto:
Why should we use `useLocation` in components that are children of `Route` and therefore receive the `location` prop?

CarsonF:
react-router is moving away from that API with the v5.1 release. It's stated that the `component` and `render` props on `Route` will go away with the v6 release. They expect you to use the hooks instead of them cloning the element to inject props.
https://reacttraining.com/blog/react-router-v5-1/

packages/ra-core/src/controller/useCreateController.ts
```diff
@@ -39,8 +39,8 @@ export interface CreateProps {
     hasEdit?: boolean;
     hasList?: boolean;
     hasShow?: boolean;
-    location: Location;
-    match: Match;
+    location?: Location; // deprecated
```
fzaninotto: it's not deprecated, it's ignored. react-router still injects these props.

packages/ra-core/src/controller/useListController.ts
```diff
@@ -29,7 +30,7 @@ export interface ListProps {
     hasEdit?: boolean;
     hasList?: boolean;
     hasShow?: boolean;
-    location: Location;
+    location?: Location; // deprecated
```
fzaninotto: same

packages/ra-ui-materialui/src/list/List.js
```diff
@@ -90,8 +90,8 @@ List.propTypes = {
     hasEdit: PropTypes.bool.isRequired,
     hasList: PropTypes.bool.isRequired,
     hasShow: PropTypes.bool.isRequired,
-    location: PropTypes.object.isRequired,
-    match: PropTypes.object.isRequired,
+    location: PropTypes.object, // deprecated
```
fzaninotto: same

packages/ra-ui-materialui/src/layout/NotFound.js
```diff
@@ -49,13 +49,13 @@ const NotFound = ({
     className,
     classes: classesOverride,
     title,
-    location,
+    location, // not used, but don't want to pass to div
```
fzaninotto: prefer a `sanitizeRestProps` function that cleans up the props passed to the div.

packages/ra-ui-materialui/src/list/List.js
CarsonF: I'm not very good with PropTypes. Should I remove it? Won't that make React complain if it is passed in?

packages/ra-core/src/controller/useCreateController.ts
CarsonF: Would you like me to remove it from the interface instead?

fzaninotto: no, just remove the `// deprecated` comment

packages/ra-ui-materialui/src/list/List.js
fzaninotto: Don't remove the line, just the comment.

fzaninotto:
Thanks!

PR #4242, opened 2020-01-03
Small fixes for SSR

CarsonF:
It doesn't seem like SSR is supported, and the process is complicated when it involves data fetching. Maybe it will never be supported out of the box, but I'm hoping I can make some changes to make it possible with a custom setup.

These are two small changes just to allow rendering to be successful server side (even though data is incomplete).
- `ConnectedRouter` interferes with `StaticRouter` and can just be ignored server side.
  https://github.com/supasate/connected-react-router/pull/322
- `Title` component was assuming `document` exists. Now it just silently fails, which is the same as not finding the `#react-admin-title` container. 
  What we really need here is a portal implementation that works server side. There's some POCs for this, but they are just that. I'm still thinking about a better way to implement this.

### DIFF
--- a/packages/ra-core/src/core/CoreAdminContext.tsx
+++ b/packages/ra-core/src/core/CoreAdminContext.tsx
@@ -73,9 +73,13 @@ React-admin requires a valid dataProvider function to work.`);
             <AuthContext.Provider value={finalAuthProvider}>
                 <DataProviderContext.Provider value={finalDataProvider}>
                     <TranslationProvider i18nProvider={i18nProvider}>
-                        <ConnectedRouter history={finalHistory}>
-                            {children}
-                        </ConnectedRouter>
+                        {typeof window !== 'undefined' ? (
+                            <ConnectedRouter history={finalHistory}>
+                                {children}
+                            </ConnectedRouter>
+                        ) : (
+                            children
+                        )}
                     </TranslationProvider>
                 </DataProviderContext.Provider>
             </AuthContext.Provider>
--- a/packages/ra-core/src/i18n/TranslationProvider.tsx
+++ b/packages/ra-core/src/i18n/TranslationProvider.tsx
@@ -2,7 +2,6 @@ import React, {
     useCallback,
     useMemo,
     Children,
-    ReactElement,
     FunctionComponent,
 } from 'react';
 
@@ -13,7 +12,6 @@ import { I18nProvider } from '../types';
 interface Props {
     locale?: string;
     i18nProvider: I18nProvider;
-    children: ReactElement<any>;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -73,9 +73,13 @@ React-admin requires a valid dataProvider function to work.`);
...
                     <TranslationProvider i18nProvider={i18nProvider}>
-                        <ConnectedRouter history={finalHistory}>
-                            {children}
-                        </ConnectedRouter>
+                        {typeof window !== 'undefined' ? (
+                            <ConnectedRouter history={finalHistory}>
+                                {children}
+                            </ConnectedRouter>
+                        ) : (
+                            children
```
fzaninotto: I don't understand: calls to `<Route>` in react-admin components (e.g. `<Resource>`) will fail if there is no router. This can only work if developers wrap their entire Admin in a StaticRouter. We should detect that a router is already provided and throw a nice exception if there is none, otherwise the developer experience will be confusing.  

Or am I missing something?

packages/ra-core/src/i18n/TranslationProvider.tsx
```diff
@@ -13,7 +12,6 @@ import { I18nProvider } from '../types';
 interface Props {
     locale?: string;
     i18nProvider: I18nProvider;
-    children: ReactElement<any>;
```
fzaninotto: I don't understand that change

CarsonF: `FunctionComponent` has `children` defined with as a `ReactNode`
```ts
interface FunctionComponent<P = {}> {
    (props: PropsWithChildren<P>, context?: any): ReactElement | null;
    ...
}

type PropsWithChildren<P> = P & { children?: ReactNode };
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/24f1d0c82da2d898acd03fbb3e692eba3c431f82/types/react/index.d.ts#L517-L518
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/24f1d0c82da2d898acd03fbb3e692eba3c431f82/types/react/index.d.ts#L773

By defining children in the `TranslationProvider` props we are restricting it further from what React accepts. This is incorrect. Context Provider's accept anything really as their children.

The change is also needed because TypeScript complains about `TranslationProvider` usage in `CoreAdminContext` because `children` there is the same loose type instead of the strict `ReactElement`.

packages/ra-core/src/core/CoreAdminContext.tsx
CarsonF: Hmm... `react-router` will throw an error if `Route` is used outside of `Router`
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Route.js#L35

I believe this is enough and that it shouldn't be the responsibility of `AdminContext` to enforce that.
BUT say the word and I'll add a check and a different(?) error message.

CarsonF: > This can only work if developers wrap their entire Admin in a StaticRouter.

And yes, all SSR frameworks do this (or require the dev to do it) as it requires tight integration with the server's request.

fzaninotto:
Thanks!

PR #4246, opened 2020-01-05
Added missing DateTimeInput parse and format props.

WiXSL:


packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -33,7 +33,7 @@ const dateTimeRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}$/;
  *
  * @param {Date | String} value date string or object
  */
-const format = (value: string | Date) => {
+const dateTimeFormatter = (value: string | Date) => {
```
djhi: I think it would be clearer to not naming like a class. Something like `formatDateTime` ?

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -58,14 +58,15 @@ const format = (value: string | Date) => {
  * @param {String} value Date string, formatted as yyyy-MM-ddThh:mm
  * @return {Date}
  */
-const parse = (value: string) => new Date(value);
+const dateTimeParser = (value: string) => new Date(value);
```
djhi: Same, `parseDateTime` ?

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
WiXSL: Done!

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
WiXSL: Done!

fzaninotto:
Thanks!

PR #4252, opened 2020-01-06
enable hot reload for demo in dev

ThieryMichel:


fzaninotto:
can't the `npm link` technique from #4359 work here?

ThieryMichel:
> can't the `npm link` technique from #4359 work here?


Actually, yes a `yarn link` does the job nicely. Guess I will just update the README.

ThieryMichel:
> can't the `npm link` technique from #4359 work here?

Sadly no, the demo use the builded module, and so even linked we still need to build the module. Which is already the case without link.

ThieryMichel:
The correct solution was to build the packages in watch mode while launching the demo, lerna does the rest.

packages/ra-core/package.json
```diff
@@ -24,7 +24,7 @@
         "build": "yarn run build-cjs && yarn run build-esm",
         "build-cjs": "rimraf ./lib && tsc",
         "build-esm": "rimraf ./esm && tsc --outDir esm --module es2015",
-        "watch": "rimraf ./lib && tsc --watch"
```
ThieryMichel: the demo need the esm package.

PR #4254, opened 2020-01-06
Fix Remove redundant components named exports in favor of the default ones.

### LINKED ISSUE #4253: Components exported twice
Some components are exported twice, as default and as not default, I don't no if it's the idea.

- ArrayInput
- DatagridCell
- DateInput
- DateTimeInput
- FileField
- ImageField
- Labeled
- LoadingIndicator
- ReferenceArrayField
- ReferenceInput
- TextInput

- List, Edit, Show (exports its `useStyles` hooks)

WiXSL:
Fixes #4253.

### DIFF
--- a/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
+++ b/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
@@ -23,7 +23,7 @@ const useStyles = makeStyles(
     { name: 'RaDatagridHeaderCell' }
 );
 
-const DatagridHeaderCell = ({
+export const DatagridHeaderCell = ({
     className,
     classes: classesOverride,
     field,
--- a/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
+++ b/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
@@ -23,7 +23,7 @@ const useStyles = makeStyles(
     { name: 'RaDatagridHeaderCell' }
 );
 
-export const DatagridHeaderCell = ({
+const DatagridHeaderCell = ({
     className,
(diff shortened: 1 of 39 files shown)


WiXSL:
Some components test failed when importing the default version which is a `recompos.pure` or `recompos.shouldComponentUpdate` version of them.

May be I could change tat for `React.memo` in another PR.

WiXSL:
Done!

packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
```diff
@@ -23,7 +23,7 @@ const useStyles = makeStyles(
     { name: 'RaDatagridHeaderCell' }
 );
 
-export const DatagridHeaderCell = ({
+const DatagridHeaderCell = ({
```
fzaninotto: I think this one has a HOC and should still be exported

packages/ra-ui-materialui/src/list/FilterForm.js
```diff
@@ -82,7 +82,7 @@ const sanitizeRestProps = ({
     ...props
 }) => props;
 
-export const FilterForm = ({
+const FilterForm = ({
```
fzaninotto: same with this one

WiXSL: Done.

packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
WiXSL: Done.

fzaninotto:
Thanks!

PR #4258, opened 2020-01-07
[RFC] Theming documentation: MyUserMenu passed as component

michelerota:
I think that the userMenu prop for the custom AppBar MyAppBar should be passed as a component, or it will trigger an "Error: Element type is invalid"


ThieryMichel:
good catch

PR #4260, opened 2020-01-07
Update documentation regarding pre-configured props of `react-final-form` 

### LINKED ISSUE #4201: SimpleForm's `keepDirtyOnReinitialize` prop should be `false` by default as `final-forms` does.
**What you were expecting:**

https://github.com/marmelab/react-admin/blob/66e3eeb465b226d23f0096450d9b5e7182cee72c/packages/ra-core/src/form/FormWithRedirect.tsx#L39

Hello, thanks for the great framework. Our team recently spent almost full 2-days to figure out why `form.reset()` doesn't work for the dirty field. `keepDirtyOnReinitialize` property is documented as a `false` for its default value. However, in RA, it defaults to `true` going against the document of `final-form` (see https://final-form.org/docs/final-form/types/Config#keepdirtyonreinitialize). There is no special note about this property in RA's document, so devs cannot expect this behavior at all.

**What happened instead:**

The property should have been `false` as the dependency (`final-form`) does. Or, at least, we should document this in RA's document separately.

**Steps to reproduce:**

**Related code:**

```
```

**Other information:**

**Environment**

* React-admin version: 3.3.1
* React version: 16.9.0
* Browser: N/A

ohtangza:
# Summary
- ~~It closes #4201~~ (already closed by @fzaninotto). I made this PR to address the issue in #4201. Some props of `react-final-form` are pre-configured. For instance. `keepDirtyOnReinitialize` for `SimpleForm` because of better design choice. Since there is a chance where newcomers might not expect it, I made this PR to explain a little bit more detail of the idea of @fzaninotto. Please see https://github.com/marmelab/react-admin/issues/4201#issuecomment-571471919 understand @fzaninotto's thoughts on the overridden props of dependencies more in detail.
- I did not change any existing explanation but add new sentences. I hope this helps other developers in the future including my team.

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -381,7 +381,7 @@ React-admin provides guessers for the `List` view (`ListGuesser`), the `Edit` vi
 
 The `<SimpleForm>` component receives the `record` as prop from its parent component. It is responsible for rendering the actual form. It is also responsible for validating the form data. Finally, it receives a `handleSubmit` function as prop, to be called with the updated record as argument when the user submits the form.
 
-The `<SimpleForm>` renders its child components line by line (within `<div>` components). It accepts Input and Field components as children. It relies on `react-final-form` for form handling. Note that some props of `react-final-form` is given by `react-admin` as the result of a design choice. For instance, `keepDirtyOnReinitialize` is set to be `true` in `react-admin`, but it's `false` in `react-final-form` by default. However, you can override this props with any values you want because `react-admin` passes down props to to the underlying component in order to give flexibility to developers.
+The `<SimpleForm>` renders its child components line by line (within `<div>` components). It accepts Input and Field components as children. It relies on `react-final-form` for form handling. Note that some props of `react-final-form` is given by `react-admin` as the result of a design choice. For instance, `keepDirtyOnReinitialize` is set to be `true` in `react-admin`, but it's `false` in `react-final-form` by default. However, you can override this props with any values you want because `react-admin` passes down props to to the underlying component.
 
 ![post edition form](./img/post-edition.png)
 
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -381,7 +381,7 @@ React-admin provides guessers for the `List` view (`ListGuesser`), the `Edit` vi
 
 The `<SimpleForm>` component receives the `record` as prop from its parent component. It is responsible for rendering the actual form. It is also responsible for validating the form data. Finally, it receives a `handleSubmit` function as prop, to be called with the updated record as argument when the user submits the form.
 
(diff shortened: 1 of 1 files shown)


docs/CreateEdit.md
```diff
@@ -381,7 +381,7 @@ React-admin provides guessers for the `List` view (`ListGuesser`), the `Edit` vi
 
 The `<SimpleForm>` component receives the `record` as prop from its parent component. It is responsible for rendering the actual form. It is also responsible for validating the form data. Finally, it receives a `handleSubmit` function as prop, to be called with the updated record as argument when the user submits the form.
 
-The `<SimpleForm>` renders its child components line by line (within `<div>` components). It accepts Input and Field components as children. It relies on `react-final-form` for form handling.
+The `<SimpleForm>` renders its child components line by line (within `<div>` components). It accepts Input and Field components as children. It relies on `react-final-form` for form handling. Note that some props of `react-final-form` is given by `react-admin` as the result of a design choice. For instance, `keepDirtyOnReinitialize` is set to be `true` in `react-admin`, but it's `false` in `react-final-form` by default. However, you can override this props with any values you want because `react-admin` passes down props to to the underlying component.
```
ohtangza: @ssohymind Can you review this text to double-check if there is no issue in your understanding?

djhi:
```suggestion
The `<SimpleForm>` renders its child components line by line (within `<div>` components). It accepts Input and Field components as children. It relies on `react-final-form` for form handling. Note that some props of `react-final-form` are given by `react-admin` as the result of a design choice. For instance, `keepDirtyOnReinitialize` is set to be `true` in `react-admin`, but it's `false` in `react-final-form` by default. However, you can override this props with any values you want because `react-admin` passes down props to to the underlying component.
```

ssohymind: @ohtangza, I can fully understand what this text means!

ohtangza: Just applied! Thanks!

fzaninotto:
Sorry, but I think this is not bringing any more clarity. 

It says 'some' props are overridden but the reader doesn't know which ones or with which value. And it doesn't say how to clear this confusion.

Also, this arrives very early in the form documentation, at a step where the reader is trying to understand how to use this component in the general case, and not in detail.

Again, I'm sorry that you've had problems with the way we customize components from an external library, but we do it all over the place in the react-admin codebase. I don't see the point of adding a warning for every component, like: 

> Warning: TextInput overrides material-ui's TextField onBlur prop with its own default because of design decisions.

So I'm closing this PR.

ohtangza:
@fzaninotto Oh, it's all okay for not accepting this PR. I also want this to be closed if it doesn't fit into the maintainer's thought. Anyways, thanks for your time to consider this PR. I believe the existence of this issue and PR will help future contributors to understand the direction of this open-source framework.

Issue #3727, opened 2019-09-23
Caching "bug" when fetching data


natrim:
Just noticed a weird "bug" in ra-core/src/reducer/admin/resource/data.ts

When fetching data they get cached, but when one (or more) of the items is deleted on the backend, the RA cache fails to remove it from existing items.

ie. lets have data:
 [{id: 1, name: "one"}, {id:2,name:"two"}]

it gets fetched by GET_LIST and cached
then the data on api is changed, ie. by another app (mobile,db,...)

when you press refresh in RA you get new data without the deleted item
 [{id:2,name:"two"}]

but the addRecord method looks on the incoming data and says "look item one is not here, lets just return the old one" and so you have stale data

Kmaschta:
Thank you for opening this issue. In order to confirm it, we'll have to reproduce it.
As explained in the bug report template, please fork the following CodeSandbox and repeat your issue on it:

https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple

This is the simplest way to confirm a bug is related to the React Admin codebase.

Moreover, in general, please follow the template for bug report (which version of react admin are you refering to? etc)

natrim:
well you cannot reproduce it by using the codesandbox because changing the fake backend data causes whole app to reload and thus negating the RA caching

noticed it on v3 but the v2 uses same code

fzaninotto:
Confirmed, the bug exists in the `ids.ts` reducer rather than the `data.ts` reducer.

sebitoelcheater:
any news respecting this issue?

fzaninotto:
Fixed by #4261

PR #4261, opened 2020-01-07
Fix cache timeout reload list

### LINKED ISSUE #4245: Do not reload data after cache timeout.
**What you were expecting:**

After cache timeout, LIST Page should reload data from provider.

**What happened instead:**

After cache timeout, then click one item of list page,  some items in LIST Page disappear 

**Steps to reproduce:**

Open reviews list page of Demo project in react-admin  examples folder, then after ten minutes(or change  'defaultCacheDuration' value in util/getFetchedAt.ts)  without click anything, click one item in list, the problem happened:  some  items disappear .

**Related code:**

```
```

**Other information:**

**Environment**

* React-admin version: V3
* React version:
* Browser:

ThieryMichel:
fixes #4245, #3727, #3227 
To fix the issue, I add to change the reducer so as to never remove data not impacted by an action even if it is outdated.
For example a GET_ONE would update the other items of the same resources removing all outdated value.
Same for GET_REFERENCE_MANY that would replace all the previous items.

The issue is that if we display an edit form along the list.
Then by opening this edit form (that trigger a GET_ONE) Then this GET_ONE would remove all items except the edited one as well as all reference linket to these items

As an added bonus, now if we are on an edit page and return on the list after the cache has expired: The old list will first be displayed, before fetching the new data and if the data did not change, we get no rerender. As opposed to displaying the list with only the edited item, before reloading the rest.

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/data.spec.ts
+++ b/packages/ra-core/src/reducer/admin/resource/data.spec.ts
@@ -8,7 +8,7 @@ import {
     GET_MANY_REFERENCE,
 } from '../../../core';
 import getFetchedAt from '../../../util/getFetchedAt';
-import dataReducer, { replaceRecords, addOneRecord } from './data';
+import dataReducer, { replaceRecords, addRecords, addOneRecord } from './data';
 import { FETCH_END } from '../../../actions';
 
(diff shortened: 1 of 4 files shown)


fzaninotto:
Thanks!

PR #4262, opened 2020-01-07
Add helperText prop to ArrayInput

### LINKED ISSUE #4227: ArrayInput does not support helperText prop

netdown:
Fixes #4227

(port of #4236 in next branch)

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput.tsx
@@ -2,7 +2,8 @@ import React, { cloneElement, Children } from 'react';
 import PropTypes from 'prop-types';
 import { isRequired, FieldTitle, composeValidators } from 'ra-core';
 import { useFieldArray } from 'react-final-form-arrays';
-import { InputLabel, FormControl } from '@material-ui/core';
+import { InputLabel, FormControl, FormHelperText } from '@material-ui/core';
+import InputHelperText from './InputHelperText';
 
 import sanitizeRestProps from './sanitizeRestProps';
 
@@ -51,6 +52,7 @@ export const ArrayInput = ({
     className,
     defaultValue,
     label,
+    helperText,
     children,
     record,
     resource,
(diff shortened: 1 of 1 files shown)


fzaninotto:
Tests fail because of https://github.com/final-form/react-final-form-arrays/issues/78. We cant merge this one until react-final-form-arrays fixes their bug upstream.

packages/ra-ui-materialui/src/input/ArrayInput.tsx
```diff
@@ -70,21 +72,37 @@ export const ArrayInput = ({
...
+                error={touched && !!error}
+            >
                 <FieldTitle
                     label={label}
                     source={source}
                     resource={resource}
                     isRequired={isRequired(validate)}
                 />
             </InputLabel>
+            {(touched && error) || helperText ? (
```
archfz, line 97: For consistency shouldn't this be under the items? For other inputs we don't put the helper text between the label and the form control. @fzaninotto what do you think ?

netdown:
@fzaninotto What is the reason of closing?

WiXSL:
> Tests fail because of [final-form/react-final-form-arrays#78](https://github.com/final-form/react-final-form-arrays/issues/78). We cant merge this one until react-final-form-arrays fixes their bug upstream.

This has been fixed in version [3.1.2](https://github.com/final-form/react-final-form-arrays/releases/tag/v3.1.2)

PR #4265, opened 2020-01-08
Improve @material-ui includes.

WiXSL:
Refs #4253


packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ActionDelete from '@material-ui/icons/Delete';
-import colorManipulator, { makeStyles } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/core/styles';
+import { fade } from '@material-ui/core/styles/colorManipulator';
```
AleksandarFaraj: Seems like it's discouraged to do 3rd level imports, , check the link

    @material-ui/core/styles/colorManipulator
                   1     2           3
https://material-ui.com/guides/minimizing-bundle-size/

LevelImports

WiXSL: You are seen the first commit. In the last one, I change it.

I saw that problem, but unless you have a solution, I have to leave it like it was because you cannot import `colorManipulator` directly.

AleksandarFaraj: Sorry!

WiXSL: It's ok.

fzaninotto:
We won't merge the changes to the `@material-ui/core` imports. Our developer experience is also important, and the "one line per component import" is just too cumbersome. Besides, the performance issue only concerns icons. So please just change the 2 instances of `@material-ui/icons` import.

WiXSL:
Ok.
At least could I include one liners like:

```diff
- import { useMediaQuery } from '@material-ui/core';
+ import useMediaQuery from '@material-ui/core/useMediaQuery';

- import { makeStyles } from '@material-ui/core';
+ import { makeStyles } from '@material-ui/core/styles';

fzaninotto:
No thanks, I'll merge it as is. I don't want to introduce two syntax for importing from `@material-ui/core`.

fzaninotto:
Thanks!

PR #4266, opened 2020-01-08
[RFR] Convert the Menu, MenuItem and DashboardMenuItem to TypeScript

steurt:
I wanted to make an addition to the Menu component, when going into the source I discovered that it was still written in JavaScript and not yet converted to TypeScript. Therefore I have now first converted the existing component to TypeScript. Later a pull request will follow with an addition to the Menu.

I converted the Menu and its MenuItem and DashboardMenuItem child components to TypeScript. The functionality and interfaces have stayed the same.


fzaninotto:
Thanks!

PR #4275, opened 2020-01-10
Fix warning on catchall page due to improper sanitization in <NotFound> component.

WiXSL:
This could happend from `RoutesWithLayout.tsx`.
```jsx
...
<Route
    render={routeProps =>
        createElement(catchAll, {
            ...routeProps,
            title,
        })
    }
 />
...

### DIFF
--- a/packages/ra-ui-materialui/src/layout/NotFound.js
+++ b/packages/ra-ui-materialui/src/layout/NotFound.js
@@ -69,7 +69,7 @@ const NotFound = ({ className, classes: classesOverride, title, ...rest }) => {
     );
 };
 
-const sanitizeRestProps = ({ 
+const sanitizeRestProps = ({
     staticContext,
     history,
     location,
--- a/packages/ra-ui-materialui/src/layout/NotFound.js
+++ b/packages/ra-ui-materialui/src/layout/NotFound.js
@@ -69,7 +69,13 @@ const NotFound = ({ className, classes: classesOverride, title, ...rest }) => {
     );
 };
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
No, I don't think we should do that. If a route is to be used as a child of Route, it should expect these props to be passed.

Luwangel:
Hi, thanks for your contribution :)

By reading your PR, I'm wondering what issue you try to fix. I think it's the warning on the NotFound page, but you didn't mention it in the description. Could you write what you try to achieve?

In addition, you suggest another change which could be a good improvement.  Why didn't you do it?

Regards,
Adrien

WiXSL:
I tried to avoid they staticContext error thrown on NotFound component render.
They comment was to show the origin of the problem.

WiXSL:
@fzaninotto, In this  case the `sanitizeRestProps` is called only to sanitize props to be applied to a `div` element inside `NotFound` component.


```jsx
const NotFound = ({ className, classes: classesOverride, title, ...rest }) => {
    const classes = useStyles({ classes: classesOverride });
    useAuthenticated();
    return (
        <div
            className={classnames(classes.container, className)}
            {...sanitizeRestProps(rest)}
        >
    //...

fzaninotto:
Sorry, I read too fast. Reopening.

fzaninotto:
In the future, can you please post a more precise description of the bug you're trying to fix? Otherwise we have to guess and sometimes we guess wrong ;)

WiXSL:
No problem, It was my fault, , I assume it was a trivial change to understand.

fzaninotto:
Thanks!

PR #4282, opened 2020-01-13
Fix types on CoreAdminUI

CarsonF:
All "required" props had default values, which means they aren't required.


Luwangel:
It looks good to me. Thanks :)

PR #4292, opened 2020-01-15
SimpleList Skeleton for First Load

djhi:
It adapts depending on :
- left/right icon/avatar is specified 
- secondary or tertiary text is specified.

SimpleListLoading

### DIFF
--- a/packages/ra-ui-materialui/src/list/Placeholder.tsx
+++ b/packages/ra-ui-materialui/src/list/Placeholder.tsx
@@ -9,7 +9,7 @@ const useStyles = makeStyles(
             display: 'flex',
         },
     }),
-    { name: 'RaSimpleListLoading' }
+    { name: 'RaPlaceholder' }
 );
 
 interface Props {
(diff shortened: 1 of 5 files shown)


packages/ra-ui-materialui/src/list/Placeholder.tsx
```diff
@@ -0,0 +1,25 @@
...
+import classnames from 'classnames';
+
+const useStyles = makeStyles(
+    theme => ({
+        root: {
+            backgroundColor: theme.palette.grey[300],
+            display: 'flex',
+        },
+    }),
+    { name: 'RaSimpleListLoading' }
```
fzaninotto: should be RaPlaceholer

packages/ra-ui-materialui/src/list/Placeholder.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FC } from 'react';
```
fzaninotto: you should export this component in index.js

djhi: Done

Issue #2489, opened 2018-10-31
defaultValue of Inputs inside ArrayInput not working


AndresVelandia:
**What you were expecting:**
I set defaultValue of Inputs inside an ArrayInput and expect that when I add a new element it sets the default value. For example, if I have an array of posts each one with and id, title, created_at, description, and set the defaultValue of created_at DateInput to Date.now(), I expect that everytime I add a new post the created_at will be Date.now()

**What happened instead:**
It doesn't work

**Related code:**
Forked from simple demo, edit a post and try to add a new backlink in the miscellaneous tab. 
https://codesandbox.io/s/k06x14xy8v

**Environment**

* React-admin version: 2.4.1
* React version: ^16.6.0
* Browser: Chrome

fzaninotto:
Confirmed and reproduced. That's because `defaultValues` are currently only set at one level:

```js
// in ra-core/form/getDefaultValues.js
const getDefaultValues = (data = {}, defaultValue = {}, defaultValues = {}) => {
    const globalDefaultValue =
        typeof defaultValue === 'function' ? defaultValue() : defaultValue;
    return { ...globalDefaultValue, ...defaultValues, ...data };
};
```

Let's say a record has sub elements, as follows:

```
{
  id: 123,
  foos: [
    { bar: 'baz' }
 ]
}
```

Given there is an `ArrayInput` written as:

```jsx
<ArrayInput>
  <SimpleFormIterator>
    <TextInput source="bar" defaultValue="hello" />
  </SimpleFormIterator>
</ArrayInput>
```

When a user add a `foo` by way of `ArrayInput`, the default value of `foos` is set to: 

```
[
    { bar: 'baz' },
    { bar: 'hello' }
]
```

but since `foos` *already has a value*, this default value is never used...

afilp:
@fzaninotto So is this an issue that we should raise in `redux-form` repo? Or we should investigate how to do this with `react-admin` code? Maybe iterate on the fileds upon the addition of a new item and add the `default` values to it too?

fzaninotto:
react-admin does its own default value voodoo to overcome the redux-form limitations in this field. the solution must be found in react-admin.

tusmenko:
One point to add - I found that it's not working on Edit view, but works on Create page.

brian-isenpai:
We're experiencing issues with default value of NumberInput and TextInput inside ArrayInput and SimpleFormIterator. The default values work the first time you add a value, but after that it doesn't work. 

[...]

PR #4306, opened 2020-01-16
fix(simple-form-iterator): Fixes object-object when there is a single TextInput without source param.

gvillo:
This could be a good start to fix [Object object] issue on SimpleFormIterator. I was able to reproduce and fix the issue if you have a plain array (such as [''tag1, 'tag2', 'tag3'].

I didn't edited docs, but the idea is if you don't specify a `source` param in the TextInput and it's just 1 input > it's a plain string array otherwise it's an object (you could add more than just input or just name property as you wish on `source`)

#2489

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -103,9 +103,14 @@ export class SimpleFormIterator extends Component {
     };
 
     addField = () => {
-        const { fields } = this.props;
+        const { fields, children } = this.props;
         this.ids.push(this.nextId++);
-        fields.push({});
+        // Checks if there is just 1 children without source param then push string since it's not an {}
+        if (children.props && children.props.source === undefined) {
+            fields.push('');
+        } else {
+            fields.push({});
+        }
     };
 
     render() {


burnes:
Couldn't it also be of a different type than string (e.g. number)? Can't we use the `defaultValue` prop for that?

JulienMattiussi:
We'd prefer to merge this PR https://github.com/marmelab/react-admin/pull/4394/files to fix this issue
 
- The use of undefined dont fix a string type for values
- The defaultValue can be used on the Input props to set a default typed value.

Unfortunately, your test on children prop is not very relevant for the data that array shoud contains.
- A SimpleFormIteratorIterator can have many children but having a array data of raw items

``` 
<SimpleForm>
    <ArrayInput source="emails">
        <SimpleFormIterator translate={x => x}>
            <TextInput />
            <SomeCustomButton />
        </SimpleFormIterator>
    </ArrayInput>
</SimpleForm>
```

- A SimpleFormIteratorIteratorbut can have a single children whithout source but having a array data of object items

``` 
<SimpleForm>
    <ArrayInput source="emails">
        <SimpleFormIterator translate={x => x}>
            <ComplexHOCInput />
        </SimpleFormIterator>
    </ArrayInput>
</SimpleForm>
```

fzaninotto:
Superseded by #4394. Thanks for your PR, it was a great inspiration!

gvillo:
@fzaninotto awesome, sorry I didn't have much time to work on it. Glad it will be fixed shortly. Thank you all guys!

PR #4338, opened 2020-01-22
Export Reference Utilities

djhi:
This will avoid duplication in addons related to references.
@fzaninotto Should I target the `next` branch ?


fzaninotto:
Yes, this should target the `next` branch as it's not a bug fix.

PR #4339, opened 2020-01-22
Ensure useEditController specifies a default redirect to its caller 

djhi:
Like useCreateController does.

This is useful when you decorate form components and override their submit handler.

### DIFF
--- a/packages/ra-core/src/controller/useEditController.ts
+++ b/packages/ra-core/src/controller/useEditController.ts
@@ -42,6 +42,7 @@ export interface EditControllerProps {
     resource: string;
     basePath: string;
     record?: Record;
+    redirect: RedirectionSideEffect;
     version: number;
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/controller/useEditController.ts
```diff
@@ -146,8 +147,11 @@ const useEditController = (props: EditProps): EditControllerProps => {
         resource,
         basePath,
         record,
+        redirect: DefaultRedirect,
```
fzaninotto: I don't understand, the redirect target should be a  parameter of the controller, not something it returns..

djhi: It's only the default. The "real" redirect can be specified on the form itself.

fzaninotto:
As this is not a bug fix, please PR against next

PR #4342, opened 2020-01-22
Reduce the default app weight by removing the image background on the Login screen

ThieryMichel:
Replace default unsplash image for Login component by a simple gradient.

Demo login is unchanged.

Preview:

image

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Login.tsx
+++ b/packages/ra-ui-materialui/src/auth/Login.tsx
@@ -160,7 +160,6 @@ Login.propTypes = {
 };
 
 Login.defaultProps = {
-    backgroundImage: 'https://source.unsplash.com/random/1600x900/daily',
     theme: defaultTheme,
     children: <DefaultLoginForm />,
 };
--- a/packages/ra-ui-materialui/src/auth/Login.tsx
+++ b/packages/ra-ui-materialui/src/auth/Login.tsx
@@ -44,6 +44,8 @@ const useStyles = makeStyles(
             justifyContent: 'flex-start',
             backgroundRepeat: 'no-repeat',
             backgroundSize: 'cover',
+            backgroundImage:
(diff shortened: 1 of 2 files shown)


fzaninotto:
If I'm not mistaken, that's a breaking change: people having defined a custom image for the background will lose it. Can you make the change backwards compatible?

alanpoulain:
It seems to me it's even worse since the default prop has been removed from `ra-ui-materialui`. People *not* having a custom `backgroundImage` prop will have a blank page.

Also only the demo has been changed, why not using this for the framework?

Nice change btw :slightly_smiling_face:

ThieryMichel:
> If I'm not mistaken, that's a breaking change: people having defined a custom image for the background will lose it. Can you make the change backwards compatible?

I just removed the default value for the background image. Only people that specified no backround image will have a white background. Which can be considered a breaking change. I can split the PR.

- One to change the demo login page (no BC)
- One to remove the default value for Login default backgroundImage (BC)

ThieryMichel:
Well totally changed the pr to only replace the unspalsh default image with a gradient in the Login component.

fzaninotto:
this needs to be done on the `next` branch as it's not a bug fix

packages/ra-ui-materialui/src/auth/Login.tsx
```diff
@@ -44,6 +44,8 @@ const useStyles = makeStyles(
             justifyContent: 'flex-start',
             backgroundRepeat: 'no-repeat',
             backgroundSize: 'cover',
+            backgroundImage:
```
fzaninotto: suggested alternative: 'radial-gradient(circle at 50% 11em, #313264 0%, #00023b 50%, #00023b 100%)'

fzaninotto:
You should also update the Theming.md documentation:

image

PR #4343, opened 2020-01-22
Update the Catching Authentication Errors code example

ArnaudBan:
I found this small error in the documentation when adding  authentication on a project. I hope that can help somebody else.

### DIFF
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -149,7 +149,7 @@ export default {
     login: ({ username, password }) => { /* ... */ },
     logout: () => { /* ... */ },
     checkError: (error) => {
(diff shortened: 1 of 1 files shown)


fzaninotto:
I don't think that's correct. The error is what is returned by the dataProvider. The dataProvider spec doesn't say that it should contain a `response` field. 

Perhaps you had problems due to your own implementation of the dataProvider?

ArnaudBan:
Thank you for the quick response. I am using [Hydra dataProvider](https://github.com/api-platform/admin/blob/master/src/hydra/dataProvider.js). I am going to check if i can figure out where is the problem and let them know.

PR #4346, opened 2020-01-23
Fix Datagrid column headers don't update sort indicator when used in ReferenceManyField

### LINKED ISSUE #4178: Column headers for Datagrid in ReferenceManyField don't update sort indicator
**What you were expecting:**
When using a ReferenceManyField with a Datagrid child a user can re-sort the columns by clicking on the column headers. The field being used for sort and the sort order should be indicated by an arrow next to the column header. 

**What happened instead:**
The field with the initial sort has the indicator next to it but sorting on any other column doesn't update the sort field or oder indicator. The data is sorted correctly as expected. This feature used to work with RA 2.9

**Steps to reproduce:**
Go to https://marmelab.com/react-admin-demo/#/customers
Sort by "Orders" to bring a customer with multiple orders to the top of the list
Select a customer with multiple orders
Click on the "Orders" tab
Click on the column headers to sort by that column in ascending or descending order
The columns will sort correctly but the indicator won't be updated

**Other information:**

**Environment**

* React-admin version: 3.0.4
* Last version that did not exhibit the issue (if applicable): 2.9.x

fzaninotto:
Closes #4178

### DIFF
--- a/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
+++ b/packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
@@ -23,17 +23,18 @@ const useStyles = makeStyles(
     { name: 'RaDatagridHeaderCell' }
 );
 
-export const DatagridHeaderCell = ({
-    className,
-    classes: classesOverride,
-    field,
-    currentSort,
-    updateSort,
-    resource,
-    isSorting,
-    ...rest
-}) => {
-    const classes = useStyles({ classes: classesOverride });
+export const DatagridHeaderCell = props => {
+    const {
+        className,
+        classes: classesOverride,
+        field,
+        currentSort,
+        updateSort,
+        resource,
+        isSorting,
+        ...rest
+    } = props;
+    const classes = useStyles(props);
     const translate = useTranslate();
     return (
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/DatagridHeaderCell.js
```diff
@@ -23,17 +23,18 @@ const useStyles = makeStyles(
...
-    updateSort,
-    resource,
-    isSorting,
-    ...rest
-}) => {
-    const classes = useStyles({ classes: classesOverride });
+export const DatagridHeaderCell = props => {
+    const {
+        className,
+        classes: classesOverride,
```
djhi: You probably don't want to destructure this one or it won't be passed to `useStyles` preventing overrides

fzaninotto: I'm passing `props` to `useStyles`, not `rest`. Destructuring as `classesOverride` here avoids the name conflict with the `useStyles` return value.

djhi: Read too quickly...

PR #4354, opened 2020-01-24
Fix typo in ra-data-graphcool and ra-data-graphql-simple README's.

kahdo:
I was reading the docs and found this typo.


fzaninotto:
Thanks!

PR #4360, opened 2020-01-27
Fix helper place on inputs to avoid clipping/unclipping forms

JulienMattiussi:
This PR makes the helper text space always reserved in forms inputs (even if there isn't any error or help text). This prevent forms to change height brutally upon validation, and the submit buttons to move under the mouse pointer.

Forms will be a bit taller than before with this change

image


fzaninotto:
Please PR on `next` rather than `master` - this is not exactly a bug fix.

packages/ra-ui-materialui/src/input/InputHelperText.tsx
```diff
@@ -18,7 +18,9 @@ const InputHelperText: FunctionComponent<Props> = ({
         <ValidationError error={error} />
     ) : helperText ? (
         <>{translate(helperText, { _: helperText })}</>
-    ) : null;
+    ) : (
```
fzaninotto: I don't understand that change. What's wrong with `null`?

JulienMattiussi: With null, the helper component is not rendered
With the empty fragment, it is rendered with nothing inside

PR #4365, opened 2020-01-28
Add displayNull option in NullableBooleanInput and document it

### LINKED ISSUE #4362: Null value for NullableBooleanInput doesn't work
Issue originally reported here https://github.com/marmelab/react-admin/issues/4299#issuecomment-575219409

**What you were expecting:**
Value `null` for `NullableBooleanInput` to show what I have in translation for `ra.boolean.null`.

**What happened instead:**
It doesn't show anything.

**Steps to reproduce:**
Go here https://codesandbox.io/s/nifty-snowflake-mcf9e
Click on `jane`. Notice there is nothing for `null` even if translation should say Unknown.

I mentioned a workaround for this in my previous comment, replacing the `''` to `' '` (a whitespace) both in `MenuItem` and `getStringFromBoolean` in `NullableBooleanInput.tsx` seems to works.

JulienMattiussi:
Closes #4362

## Issue

When a NullableBooleanInput has a null value, we never see the custom label for this null value

## Solution

Add `displayNull` prop in NullableBooleanInput to allow it to display a custom null label

It simply override needed TextField props for this behaviour 

- SelectProps: {displayEmpty: true }
- InputLabelProps: { shrink: true }

## Documentation

image


packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -65,6 +66,21 @@ const NullableBooleanInput: FunctionComponent<
...
+              SelectProps: {
+                  ...(options && options.SelectProps),
+                  displayEmpty: true,
+              },
+              InputLabelProps: {
+                  ...(options && options.InputLabelProps),
+                  shrink: true,
+              },
+          }
+        : { ...options };
```
fzaninotto:
```suggestion
        : options;
```

packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -65,6 +66,21 @@ const NullableBooleanInput: FunctionComponent<
         source,
         validate,
     });
+
+    const enhancedOptions = displayNull
```
fzaninotto: there is something I don't really like with your syntax: `displayNull` takes precedence over `options`. 

I'd do it differently: prepare a static options object outside of the component:

```js
const displayNullOptions = { SelectProps: { displayEmpty: true }, InputLabelProps: { shrink: true } }
```

Then pass it ti TextField if `displayNull` is true:

```jsx
<TextField
   {displayNull ? ...displayNullOptions : ...{} }
   {...options}
```

JulienMattiussi: I would like this case to work fine : 
```
<NullableBooleanInput
   source="source"
   displayNull
   options={{SelectProps: { anotherCustomProp: true }}}
/>
```

I think, with your syntax, that the 
`SelectProps: { displayEmpty: true }`
is totally overrided by the 
`SelectProps: { anotherCustomProp: true }`

My code wanted to mix them in
```
SelectProps: { 
   displayEmpty: true,
   anotherCustomProp: true,
}```

JulienMattiussi: I changed for options to precede on displayNull, without erasing it

fzaninotto:
Please PR against `next` and add an automated test

sliterok:
I think you should mention in the documentation that you should add new messages as i18n provider for admin component so that it would work. Just wasted 20 mins of my life trying to figure out what's wrong.

djhi:
> I think you should mention in the documentation that you should add new messages as i18n provider for admin component so that it would work. Just wasted 20 mins of my life trying to figure out what's wrong.

Sure, maybe you can open a pull request?

sliterok:
> > I think you should mention in the documentation that you should add new messages as i18n provider for admin component so that it would work. Just wasted 20 mins of my life trying to figure out what's wrong.
> 
> Sure, maybe you can open a pull request?

Done! https://github.com/marmelab/react-admin/pull/5172

Luwangel:
Thanks @sliterok!

PR #4367, opened 2020-01-29
Fix Autocomplete selection display with custom Component

### LINKED ISSUE #4281: AutocompleteInput using a function as optionText shows [object Object] after selecting an item
**What you were expecting:**
When using an optionRenderer function as the optionText for AutocompleteInput, after selecting an item from the list, I hoped to see something useful in the input.

**What happened instead:**
Instead I see [object Object].
image

**Steps to reproduce:**
Code like this should do it:
```

const optionRenderer = choice => {
    return (
         <span>{choice.name}-{choice.id}</span>
    )
};
// ...
<ReferenceInput
    source="post_id"
    reference="posts"
    perPage={15}
    sort={{ field: 'title', order: 'ASC' }}
    fullWidth
>
    <AutocompleteInput
        optionText={optionRenderer}
        options={{ fullWidth: true }}
    />
</ReferenceInput>
```

**Related code:**
See https://codesandbox.io/s/pensive-perlman-vwg90 and navigate to comments and select a comment.

**Other information:**
There seems no other prop besides optionText to render a suggestion once it has been selected. So it kinda makes sense that [object Object] is printed, but I consider it a bug if no other means are available to render the selected item. I suppose I'm missing something.

Somewhat related issues:
- https://github.com/marmelab/react-admin/issues/4092
- https://github.com/marmelab/react-admin/issues/4183
- https://github.com/marmelab/react-admin/issues/3503

**Environment**

* React-admin version: 3.1.2
* Last version that did not exhibit the issue (if applicable): not sure
* React version: 16.12
* Browser: any

JulienMattiussi:
Closes #4281

## Issue

When an autocomplete use a custom Component as `optionText`, The selected value can't be display in the filter TextField.

## Solution

When you want to pass a custom component as `optionText` to an Autocomplete, 
you also have to pass a text rendering function returning the filter text to be displayed for your custom component
This `inputText` function must return a `string` which can be similary to the displayed component, or totally different

```jsx
const OptionRenderer = ({ record }) => (
    <span>
        {record.title}-{record.id}
    </span>
);

const inputText = record => `${record.title}-${record.id}`;

<AutocompleteInput
   matchSuggestion={(filterValue, suggestion) => true}
   optionText={<OptionRenderer />}
   inputText={inputText}
   options={{ fullWidth: true }}
/>
```
## Preview

image

## Documentation

image

image

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -249,7 +249,7 @@ Lastly, would you need to override the props of the suggestions container (a `Po
 | `matchSuggestion` | Optional | `Function` | - | Required if `optionText` is a React element. Function returning a boolean indicating whether a choice matches the filter. `(filter, choice) => boolean`
 | `optionText` | Optional | <code>string &#124; Function &#124; Component</code> | `name` | Fieldname of record to display in the suggestion item or function which accepts the correct record as argument (`(record)=> {string}`) |
 | `optionValue` | Optional | `string` | `id` | Fieldname of record containing the value to use as input value  |
-| `inputText` | Optional | <code>Function</code> | `-` | If `optionText` is a custom Component, This function is needed to determine the text displayed filter of the current selection. In Autocomplete, the current selection have to be an editable string and can not be a Component  (`(record)=> {string}`) |
+| `inputText` | Optional | <code>Function</code> | `-` | If `optionText` is a custom Component, this function is needed to determine the text displayed for the current selection. |
 | `setFilter` | Optional | `Function` | `null` | A callback to inform the `searchText` has changed and new `choices` can be retrieved based on this `searchText`. Signature `searchText => void`. This function is automatically setup when using `ReferenceInput`.  |
 | `shouldRenderSuggestions` | Optional | Function | `() => true` | A function that returns a `boolean` to determine whether or not suggestions are rendered. Use this when working with large collections of data to improve performance and user experience. This function is passed into the underlying react-autosuggest component. Ex.`(value) => value.trim() > 2` |
 
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -162,7 +162,7 @@ const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
 <AutocompleteInput source="author_id" choices={choices} optionText={optionRenderer} />
 ```
 
-`optionText` also accepts a custom Component. If you provide one, you will have to set the `inputText` function wich determine the input text filter of the current selection:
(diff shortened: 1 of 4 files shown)


docs/Inputs.md
```diff
@@ -162,6 +162,28 @@ const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
 <AutocompleteInput source="author_id" choices={choices} optionText={optionRenderer} />
 ```
 
+`optionText` also accepts a custom Component. If you do so, you will have to set the inputText function wich determine the input text filter of the current selection:
```
djhi:
```suggestion
`optionText` also accepts a custom Component. If you provide one, you will have to set the `inputText` function wich determine the input text filter of the current selection:
```

fzaninotto:
Please target `next` rather than `master` for this one

docs/Inputs.md
```diff
@@ -162,6 +162,28 @@ const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
 <AutocompleteInput source="author_id" choices={choices} optionText={optionRenderer} />
 ```
 
+`optionText` also accepts a custom Component. If you provide one, you will have to set the `inputText` function wich determine the input text filter of the current selection:
```
fzaninotto:
```suggestion
`optionText` also accepts a custom Component. However, as the underlying Autocomplete component requires that the current selection is a string, if you opt for a Component, you must pass a function as the `inputText` prop. This function should return text representation of the current selection:
```

docs/Inputs.md
```diff
@@ -225,8 +247,9 @@ Lastly, would you need to override the props of the suggestions container (a `Po
...
 | `emptyText` | Optional | `string` | `''` | The text to use for the empty element |
 | `matchSuggestion` | Optional | `Function` | - | Required if `optionText` is a React element. Function returning a boolean indicating whether a choice matches the filter. `(filter, choice) => boolean`
-| `optionText` | Optional | <code>string &#124; Function</code> | `name` | Fieldname of record to display in the suggestion item or function which accepts the correct record as argument (`(record)=> {string}`) |
+| `optionText` | Optional | <code>string &#124; Function &#124; Component</code> | `name` | Fieldname of record to display in the suggestion item or function which accepts the correct record as argument (`(record)=> {string}`) |
 | `optionValue` | Optional | `string` | `id` | Fieldname of record containing the value to use as input value  |
+| `inputText` | Optional | <code>Function</code> | `-` | If `optionText` is a custom Component, This function is needed to determine the text displayed filter of the current selection. In Autocomplete, the current selection have to be an editable string and can not be a Component  (`(record)=> {string}`) |
```
fzaninotto:
```suggestion
| `inputText` | Optional | <code>Function</code> | `-` | If `optionText` is a custom Component, this function is needed to determine the text displayed for the current selection. |
```

PR #4374, opened 2020-01-31
Fix UI for latest version of material-ui

fzaninotto:
Material-ui 4.8 changes the menu density by default. But as we added the "dense" prop in react-admin, now the menus are very small...

- [x] Make SelectInput and SelectArrayInput helperText space always reserved
- [x] Fix menu size
- [x] Upgrade yarn.lock
- [x] Fix TypeScript errors in ra-core
- [x] Fix TypeScript errors in ra-ui-materialui

### DIFF
--- a/examples/simple/src/posts/PostCreate.js
+++ b/examples/simple/src/posts/PostCreate.js
@@ -132,10 +132,7 @@ const PostCreate = ({ permissions, ...props }) => {
                 <FormSpy subscription={{ values: true }}>
                     {({ values }) =>
                         values.title ? (
-                            <NumberInput
-                                source="average_note"
-                                defaultValue={5}
-                            />
+                            <NumberInput source="average_note" />
                         ) : null
                     }
                 </FormSpy>
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -86,6 +86,7 @@ EnhancedBooleanField.defaultProps = {
 };
 
 EnhancedBooleanField.propTypes = {
+    // @ts-ignore
     ...Typography.propTypes,
     ...fieldPropTypes,
     valueLabelFalse: PropTypes.string,
--- a/packages/ra-ui-materialui/src/input/InputHelperText.tsx
+++ b/packages/ra-ui-materialui/src/input/InputHelperText.tsx
@@ -19,6 +19,11 @@ const InputHelperText: FunctionComponent<Props> = ({
     ) : helperText ? (
         <>{translate(helperText, { _: helperText })}</>
     ) : (
+        // material-ui's HelperText cannot reserve space unless we pass a single
+        // space as child, which isn't possible when the child is a component.
+        // Therefore, we must reserve the space ourselves by passing the same
+        // markup as material-ui.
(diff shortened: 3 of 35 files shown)


packages/ra-ui-materialui/src/input/InputHelperText.tsx
```diff
@@ -19,7 +19,8 @@ const InputHelperText: FunctionComponent<Props> = ({
     ) : helperText ? (
         <>{translate(helperText, { _: helperText })}</>
     ) : (
-        <> </>
+        // eslint-disable-next-line react/no-danger
+        <span dangerouslySetInnerHTML={{ __html: '&#8203;' }} />
```
djhi: A comment might help here. I don't get this.

packages/ra-ui-materialui/src/input/NullableBooleanInput.spec.tsx
```diff
@@ -26,24 +26,24 @@ describe('<NullableBooleanInput />', () => {
             />
         );
         const select = getByRole('button');
-        fireEvent.click(select);
+        fireEvent.mouseDown(select);
```
djhi: Why ? Comment maybe ?

examples/simple/src/posts/PostCreate.js
```diff
@@ -132,10 +132,7 @@ const PostCreate = ({ permissions, ...props }) => {
                 <FormSpy subscription={{ values: true }}>
                     {({ values }) =>
                         values.title ? (
-                            <NumberInput
-                                source="average_note"
-                                defaultValue={5}
-                            />
+                            <NumberInput source="average_note" />
```
Luwangel: Why have you removed the default value?

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -86,6 +86,7 @@ EnhancedBooleanField.defaultProps = {
 };
 
 EnhancedBooleanField.propTypes = {
+    // @ts-ignore
```
Luwangel: Why do we need to set `ts-ignore` everywhere?

examples/simple/src/posts/PostCreate.js
fzaninotto: Because final-form latest version doesn't take defaultValue into account if the initialValue is already defined at the form level (see https://github.com/final-form/final-form/pull/282)

packages/ra-ui-materialui/src/field/BooleanField.tsx
fzaninotto: I didn't find a better way to make latest TypeScript stop complaining ¯\_(ツ)_/¯

packages/ra-ui-materialui/src/input/NullableBooleanInput.spec.tsx
fzaninotto: That's the way the material-ui team tests their Select component. I didn't looked further when the react-admin Select-based tests stopped passing after upgrading material-ui to the latest version.

guntoroyk:
Hi @djhi @fzaninotto , I upgraded my app to the latest version of react-admin, why did the custom filter position turn out like this? Please help. Thanks.

image

guntoroyk:
Hi @djhi  @fzaninotto  , I upgraded my app to the latest version of react-admin, why did the custom filter position turn out like this? Please help. Thanks.

image

fzaninotto:
@guntoroyk see #4389

PR #3805, opened 2019-10-11
Use typescript 3.7 to emit declarations from js code, Use composite build

Bnaya:
Using composite project and the new ability of ts 3.7 of allowJs + declarations.

* very good DX & smarter build 
* No need to run build for dependencies  to edit code in editor
* lets us build d.ts files for also the js based packages, so we may bundle types with all of the 
packages.
*<s> Typescript 3.7 is not released yet, but there are no errors.
Maybe the merge should wait.</s>

I'm really not sure how good the inferred declarations are.
symbols discoverability should be good, and that's a big plus, 
but maybe using React components without clear props will confuse developers more than help them.
So we can choose to not ship the declarations inside the bundle.

Improving the types will not require full `ts` refactor, it can be done incrementally, using checkJs on the relevant package and adding typescript-jsdoc types.

The entry point for all of the typescript programs is `tsconfig.build-all.all.json`
That reference  all of the inner tsconfigs, in cjs & esm variant.
means, cjs & esm are built as part of one big project. no need a separate invocation 

Development workflow using:
```
yarn tsc --build tsconfig.build-all.all.json --incremental --watch
```
Will give very efficient dev workflow, while watching the entire codebase, the `incremental` option makes you "pay" only for your changes.

You may still run tsc --build -w on specific packages, without the need to build the packages it depends on before, as it will also build it for you as composite project.

<s>The main thing that missing: build also esm</s>

<s>Some of the `js` packages emitted d.ts are broken due to:
https://github.com/microsoft/TypeScript/issues/33626</s>
run `yarn validate-emitted-declarations` to validate the emitted d.ts files
<s>I think that as first step, it's better to shipping the emitted d.ts files for the js-based packages. </s>
I've added a workaround to emit valid d.ts, and also 33626 is expected to be fixed before 3.7 stable.

Issue #1617, opened 2018-03-08
Typescript declarations


in19farkt:
Hi. I want to write typescript types declarations for this library and could you please help me to plan how to do it. Does it's worth to cover version 1.4 or wait for version 2.0 and make types for its?

fzaninotto:
We've spoken about that in #250: we're not users of typescript ourselves and couldn't maintain these declarations.

I advise you to focus on `next`, and write the types as standalone files.

in19farkt:
Thanks. I'll start writing types for `next` and post them in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

clementgarbay:
Hi @in19farkt, did you start writing types file or not yet ? I didn't see types package for this project on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

in19farkt:
Yes, I started writing types, but it's hard and time-consuming. The draft version can be viewed [here](https://github.com/in19farkt/typed-react-admin). I did not make changes for a long time, because I copied these types into the project so that it was easier to change them.

abumalick:
I think you started to use typescript, right @fzaninotto ?

Are you planning to ship types for `react-admin` now?

fzaninotto:
We'll ship them when they are ready and complete.

user753:
@fzaninotto I think even some basic uncomplete declarations would be very helpful for typescript users. 
For example you could use typescript 3.7 and get declarations for js files before full migration to typescript.

fzaninotto:
There is a PR to update TypeScript to 3.7 (#3805), but it's not mergeable.

Vadorequest:
@fzaninotto Please re-open this issue. There are multiple issues about TypeScript definitions, but none are open (that I could find). One should be open until the issue is actually resolved, otherwise people will keep opening new ones. So I suggest you reopen this one so that we may continue to discuss this topic here.

I'd like to explain a bit TS history, has you stated you're not familiar with it (in 2018).
DefinitelyTyped is the reference of TS typings, and it doesn't matter who adds them there. Anybody can. It's used as a fallback solution by developers when a library doesn't include/release its own typings.

The most preferred solution nowadays is to include typings from the lib itself, it's keeps things centralised and works well with new releases. This process is automated when the project itself uses TypeScript.

I'm starting with react-admin today in 2020 and there are still no typings, or quite a few:
<details>

```ts
[...]

PR #4375, opened 2020-01-31
typescript declarations emit

Bnaya:
Emit typescript declarations - utilize typescript 3.7 ability to emit types for js code.
Branched out from #3805 

Minor code changes to make the types compilation pass


packages/react-admin/src/index.ts
```diff
@@ -5,5 +5,11 @@ import AdminRouter from './AdminRouter';
 import defaultI18nProvider from './defaultI18nProvider';
 
 export * from 'ra-core';
+/* 
+error - what do to? which one is the correct symbol?
+Module 'ra-core' has already exported a member named 'Notification'. Consider explicitly re-exporting to resolve the ambiguity.ts(2308)
+Module 'ra-core' has already exported a member named 'Pagination'. Consider explicitly re-exporting to resolve the ambiguity.ts(2308)
+*/
+// @ts-ignore
```
Bnaya: @fzaninotto what's the correct symbols here should be?

user753:
There is many required props like
`classes` required
```
declare function Datagrid({ classes: classesOverride, ...props }: {
    [x: string]: any;
    classes: any;
}): JSX.Element;
```
`record` required
```
declare function ReferenceField({ children, record, source, ...props }: {
    [x: string]: any;
    children: any;
    record: any;
    source: any;
}): JSX.Element;
```
ReferenceInput 
all required
```
interface Props {
    allowEmpty: boolean;
    basePath: string;
    children: ReactElement;
    classes: any;
    className: string;
    label: string;
    reference: string;
    resource: string;
    [key: string]: any;
}
```

Bnaya:
@user753 this is by no-mean full typescript coverage.
The types from the js files are based on the guesses of TSC.
But it is a great start for symbol discovery and a way to make incremental improvements to the types,
via tsx refactor or jsdocs-types in exiting js files

user753:
@Bnaya I understand that. I think It coulb be fixed if all guessed props were optional. 
Maybe there is a simple way to do that? 
Also if these types was on https://github.com/DefinitelyTyped/DefinitelyTyped we could easily fix them.

packages/ra-ui-materialui/src/layout/Loading.js
```diff
@@ -33,6 +33,9 @@ const useStyles = makeStyles(
     { name: 'RaLoading' }
 );
 
+/**
+ * @type {React.FunctionComponent<{className: string; classes: object; loadingPrimary?: string; loadingSecondary?: string;}>}
```
fzaninotto: I prefer that you turn that component to TypeScript. I don't want to maintain a lib with 3 different syntax (JS, TS, and JS-with-types)

packages/react-admin/src/index.ts
fzaninotto: I believe ra-ui-material-ui should import those types from ra-core

fzaninotto:
Also, before emitting the type definitions from ra-ui-material-ui, we must migrate the demo to TypeScript, so that we can make sure the types don't break existing apps. Would you like to tackle that task first?

Bnaya:
> migrate the demo ...Would you like to tackle that task first

I've enabled checkJs for the demo and ill go over the errors i'm getting, That's should be equivalent to typescript refactor +-
I've also deleted the `examples/demo/src/react-admin.d.ts` file, and now my VSCODE picking up the types built from my workspace.

In general, i have the feeling the component composition api RA works with not playing well with strict typescript. for example:
Screen Shot 2020-02-03 at 18 32 11
That's a valid RA code, but there is no way to tell typescript that "if your parent component has source prop you don't need one of your own"
So i'm not sure to where to take it from here

> that's because we want to be sure that the types we publish don't break existing TS apps

I think it will be very difficult to guarantee that no typescript app will not break in any way after publishing types with `react-admin`
each project has it's own hacks around types now
For me it's `@ts-ignore` all over the imports, for others its empty react admin `d.ts`
Even if our new types would be perfect, it might collide with any `d.ts` that in the consuming projects

fzaninotto:
On the example you underline, if the source can be injected, in that case it should not be required in the types.

Bnaya:
I'm willing to put effort to make that happen, but there are quite few things that are not clear to me
I would like to talk directly if possible to find the best course of action

Vadorequest:
@Bnaya Check this out https://github.com/marmelab/react-admin/issues/1617#issuecomment-595874113 :)

fzaninotto:
@Bnaya What isn't clear and what do you want to talk about?

matepaiva:
I am really looking forward to this feature! It's really helpful!

falkenhawk:
@fzaninotto since you closed this, I guess this is now superceded by https://github.com/marmelab/react-admin/issues/4505 ?

PR #4384, opened 2020-02-04
Fix RichTextInput dark theme

### LINKED ISSUE #4366: No visibile editor controls on ra-input-rich-text with dark theme
**What you were expecting:**
when switching theme from light to dark theme editor controls (ex: bold icon,italic icon,etc) should be displayed as light colored text (ex: white) 

**What happened instead:**
I think that the problem is that when you switch from light to dark theme editor controls color remains black so it will be "hidden" from dark color of layout.
If you "hover" one of the editor controls they correct appears as colored with primary color from theme.

**Steps to reproduce:**
- log in into marblelab demo application (https://marmelab.com/react-admin-demo/#/login)
- Using top right dropdown icon go to configuration and swtich theme from light to dark.
- Using the left side menu go under catalog->poster and try to create new one.
- Move to third tab labeled as "description" and you will have a rich text input but no visible editor controls.
-Try "hover" one of editor controls and it appears colored as primary theme color.

**Other information:**
I added screenshots two screenshot to better display what's happening one displays editor with "hidden" controls, the other one displays what happens when "hovering" one editor control button
hidden_editor_controls
hidden_editor_controls_hover

JulienMattiussi:
Closes #4366 

## Issue

Rich text toolbar is not visible in dark theme

## Screenshots

image

image

image

### DIFF
--- a/packages/ra-input-rich-text/src/styles.js
+++ b/packages/ra-input-rich-text/src/styles.js
@@ -1,5 +1,9 @@
 import QuillSnowStylesheet from './QuillSnowStylesheet';
 
+const darkColor = '#444';
+const lightColor = 'white';
+const selectedColor = '#304ffe';
+
 export default theme => ({
     label: {
         position: 'relative',
@@ -61,7 +65,7 @@ export default theme => ({
                     fontWeight: 500,
                 },
                 '& a': {
-                    color: '#304ffe',
+                    color: selectedColor,
                 },
                 '& ul': {
                     marginBottom: '1rem',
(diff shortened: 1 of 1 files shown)


packages/ra-input-rich-text/src/styles.js
```diff
@@ -1,5 +1,9 @@
 import QuillSnowStylesheet from './QuillSnowStylesheet';
 
+const darkColor = '#444';
```
Luwangel: Do we have a material ui theme?

JulienMattiussi: :+1:

Issue #4385, opened 2020-02-05
[RFC] Reduce dataProvider requests using a cache


fzaninotto:
## Problem

Web apps have a significant environmental footprint, partly due to their network requests (16%) and datacenter activity (19%) ([source](https://theshiftproject.org/wp-content/uploads/2019/03/Executive-Summary_Lean-ICT-Report_EN_lowdef.pdf)). 

React-admin apps make at least one network request per page, even though the local Redux store already contains the necessary data, and most of the time, this data is up to date.

## Solution

Encourage a caching solution on the client-side to reduce the number of HTTP request sent by react-admin apps (therefore reducing both network requests and datacenter activity). 

I think there are 2 ways to do that:

1. Implement caching at the dataProvider level. In fact, it's already done - the default `dataProvider` implementations use `fetch()`, which honors `Cache-Control` headers (see [fetch() doc on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache) for details). So if the API responses include `Cache-Control` headers, the `dataProvider` will know when to return a cached response (or simply revalidate it).
2. Implement caching at the react-admin level (in `useDataProvider()`) in order to have a smarter cache based on the Redux store. The idea is that a call to `getList` returns all the necessary data for a `getOne` on an id included in the `getList` response. Or, to match a usage scenario, after displaying a posts list, clicking on a post to display the post detail should not require an additional call to the server. That would reduce network requests even more than a dataProvider-based cache. 

As 1. is already possible, we just need to document it as a good practice. 

About 2., it should be opt-in and easily configurable. It requires a semi-complete, client-side database implementation based on the Redux store, so that react-admin can answer the various dataProvider read calls (`getList`, `getOne`, `getMany`, etc) without ever hitting the dataProvider. And ideally, when using the cache, `useDataProvider` should skip the "loading" state to save on performance (and avoid a double redraw).

I pushed a POC in a PR (#4386) where I played with this concept. Not only does the application reduces its environmental footprint, the user experience is also vastly improved due to an incredible performance boost.    

Please share your comments.

fzaninotto:
No comment, I'm closing this RFC. #4386 implements this idea.

PR #4386, opened 2020-02-05
Add opt-in client-side caching layer to save on network requests

fzaninotto:
See rationale in #4385

What started as a simple enhancement resulted into a pretty large refactoring and optimization. 

- [x] `useGetList` now stores data in the `list` reducer (no more in `customQueries`)
- [x] `useGetList` now has one cache for each set of parameter. This allows optimistic rendering (and cache) of lists when changing pagination, filters, and sorting
- [x] `useListController` now uses `useGetList`
- [x] `useListController` uses `resources.[resourceName].list.ids` when `resources.[resourceName].list.cachedRequest[requestSignature]` isn't defined (avoids showing an empty list when changing params)
- [x] No more Refresh when leaving optimistic mode. Instead, all dataProvider queries that were cancelled while in optimistic mode are replayed when leaving
- [x] When a dataProvider success response contains a `validUntil` key, the relevant cache in the Redux store is maked with this validity date
- [x] When a `useDataProvider` call has a valid cached response, the call to `dataProvider` is skipped altogether 
- [x] Write requests and refresh reset the cache validity and force real calls to the dataProvider

The current implementation is opt-in (and backwards compatible): the `dataProvider` responses must include a `validUntil` key to enable the cache.


fzaninotto:
As there is a `refresh` side effect after the update of the `editController`, the cache currently does not work after an edition.

fzaninotto:
Fixed the problem with the forced refresh after an undoable call

fzaninotto:
Now we cache the list of ids per request, that means that we can use the cache (and optimistic rendering) when going back to a list (paginatinon, sort, and filters included) that was seen before.

fzaninotto:
The requests made by the `<ReferenceInput>` component aren't cached because their controller still uses the `CrudGetMatchingAccumulate` action creator. That means that the `dataProvider.getList()` it makes to fetch the possible references is done by the old `fetch` saga (called by the `accumulate` saga) instead of the `useDataProvider` hook. As the cache is only implemented in the new hook, `<ReferenceInput>` doesn't use the cache. 

We need to migrate the `useReferenceInputController` code so that it doesn't use the `accumulate` saga anymore.

packages/ra-core/src/controller/useListController.ts
```diff
@@ -158,14 +160,32 @@ const useListController = (props: ListProps): ListControllerProps => {
...
-            state.admin.resources[resource]
-                ? state.admin.resources[resource].list.total
-                : null
+        // data selector
+        (state: ReduxState): Identifier[] => {
+            const resourceState = state.admin.resources[resource];
+            // if the resource isn't initialized, return null
+            if (!resourceState) return null;
+            const idsForQuery =
+                resourceState.list.idsForQuery[requestSignature];
```
ThieryMichel: Would'nt it be better to have a dataForQuery holding both ids and total? This way the requestSignature would be present only once in the state.

packages/ra-core/src/dataProvider/replyWithCache.ts
```diff
@@ -0,0 +1,61 @@
...
+    GetManyParams,
+    GetManyResult,
+} from '../types';
+
+export const canReplyWithCache = (type, payload, resourceState) => {
+    const now = new Date();
+    switch (type) {
+        case 'getList':
+            return (
+                resourceState &&
```
ThieryMichel: could use lodash/get here:
```js
get(resourceState, ['list', 'validity', JSON.stringify(payload as GetListParams)]);
```

packages/ra-core/src/dataProvider/replyWithCache.ts
```diff
@@ -0,0 +1,61 @@
...
+                resourceState.list &&
+                resourceState.list.validity &&
+                resourceState.list.validity[
+                    JSON.stringify(payload as GetListParams)
+                ] > now
+            );
+        case 'getOne':
+            return (
+                resourceState &&
```
ThieryMichel: use lodash/get

packages/ra-core/src/dataProvider/replyWithCache.ts
```diff
@@ -0,0 +1,61 @@
...
+            return false;
+    }
+};
+
+export const getResultFromCache = (type, payload, resourceState) => {
+    switch (type) {
+        case 'getList': {
+            const data = resourceState.data;
+            const requestSignature = JSON.stringify(payload);
+            const ids = resourceState.list.idsForQuery[requestSignature];
```
ThieryMichel: Strange, are we sure resourceState.list.idsForQuery is always set ?

packages/ra-core/src/reducer/admin/resource/list/validity.ts
```diff
@@ -0,0 +1,40 @@
...
+import { FETCH_END } from '../../../../actions';
+import { GET_LIST } from '../../../../core';
+
+interface ValidityRegistry {
+    [key: string]: Date;
+}
+
+const initialState = {};
+
+const validityReducer: Reducer<ValidityRegistry> = (
```
ThieryMichel: I think this would be better grouped with idsForQuery and totalForQuery to avoid having the same serialized query repeated three times though the state.

packages/ra-core/src/reducer/admin/resource/validity.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+};
+
+const addIds = (
+    ids: Identifier[] = [],
+    validUntil: Date,
+    oldValidityRegistry: ValidityRegistry
+): ValidityRegistry => {
+    const validityRegistry = { ...oldValidityRegistry };
+    ids.forEach(id => {
```
ThieryMichel:
```js
return ids.reduce((validityRegistry, id) => ({
    ...validityRegistry,
    [id]: validUntil
}), oldValidityRegistry);
```

packages/ra-core/src/reducer/admin/resource/validity.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+    return validityRegistry;
+};
+
+const removeIds = (
+    ids: Identifier[] = [],
+    oldValidityRegistry: ValidityRegistry
+): ValidityRegistry => {
+    const validityRegistry = { ...oldValidityRegistry };
+    ids.forEach(id => {
```
ThieryMichel: Idem use reduce

packages/ra-core/src/dataProvider/replyWithCache.ts
fzaninotto: Due to prettier, the code isn't super shorter, and due to lodash, it isn't more readable:

```jsx
            // with lodash
            return (
                get(resourceState, [
                    'list',
                    'validity',
                    JSON.stringify(payload as GetListParams),
                ]) > now
            );

            // without lodash
            return (
                resourceState &&
                resourceState.list &&
                resourceState.list.validity &&
                resourceState.list.validity[
                    JSON.stringify(payload as GetListParams)
                ] > now
            );
```

In this case, I prefer a bit more verbosity with a bit less indirection.

packages/ra-core/src/dataProvider/replyWithCache.ts
fzaninotto: yes: if the query key was set in `validity`, it means it was done in `idsForQuery`, too.

packages/ra-core/src/reducer/admin/resource/validity.ts
fzaninotto: The reduce version isn't shorter, and clones the registry once for every id. The original version clones the registry only once. I think the *reduce* version is slower and takes more memory. As it doesn't have the benefit of brevity, let's keep some imperative programming here. 

```js
// without reduce
const addIds = (
    ids: Identifier[] = [],
    validUntil: Date,
    oldValidityRegistry: ValidityRegistry
): ValidityRegistry =>
{
    const validityRegistry = { ...oldValidityRegistry };
    ids.forEach(id => {
        validityRegistry[id] = validUntil;
    });
    return validityRegistry;
};

// with reduce
const addIds = (
    ids: Identifier[] = [],
    validUntil: Date,
    oldValidityRegistry: ValidityRegistry
): ValidityRegistry =>
    ids.reduce(
        (validityRegistry, id) => ({
            ...validityRegistry,
            [id]: validUntil,
        }),
        oldValidityRegistry
    );
```

packages/ra-core/src/controller/useListController.ts
fzaninotto: It would be better indeed, but I don't know how to implement it properly. 

Let's say we add a `cachedRequests` key to the `list` reducer:

```js
// instead of
const listState = {
    ids: [13, 12, 11, 10, 9, 8, 7, 6, 4, 2],
    loadedOnce: true,
    params: {
        sort: null,
        order: null,
        page: 1,
        perPage: null,
        filter: {}
    },
    selectedIds: [],
    total: 13,
    idsForQuery: {
        '{"pagination":{"page":1,"perPage":10},"sort":{"field":"published_at","order":"DESC"},"filter":{}}': [13, 12, 11, 10, 9, 8, 7, 6, 4, 2],
    },
    totalForQuery: {
        '{"pagination":{"page":1,"perPage":10},"sort":{"field":"published_at","order":"DESC"},"filter":{}}': 13
    },
    validity: {
        '{"pagination":{"page":1,"perPage":10},"sort":{"field":"published_at","order":"DESC"},"filter":{}}': '2020-02-24T05:28:13.220Z'
    }
};

// group cache-related state as
const listState = {
    ids: [13, 12, 11, 10, 9, 8, 7, 6, 4, 2],
    loadedOnce: true,
    params: {
        sort: null,
        order: null,
        page: 1,
        perPage: null,
        filter: {}
    },
    selectedIds: [],
    total: 13,
    cachedRequests: {
        '{"pagination":{"page":1,"perPage":10},"sort":{"field":"published_at","order":"DESC"},"filter":{}}': {
            ids: [13, 12, 11, 10, 9, 8, 7, 6, 4, 2],
            total: 13,
            validity: '2020-02-24T05:28:13.220Z',
        }
    }
};
```

The problem is: How do I compose the `ids`, `total`, and `validity` reducers into a variable key of the `cachedRequests` reducer? We've done that for the `resources` reducer and it's hardly readable.

I'm still scratching my head about this one.

fzaninotto: I've found a way, will commit soon

fzaninotto:
switching to RFR

packages/ra-ui-materialui/src/list/Datagrid.js
```diff
@@ -93,8 +93,8 @@ const useStyles = makeStyles(
  *     </Datagrid>
  * </ReferenceManyField>
  */
-function Datagrid({ classes: classesOverride, ...props }) {
-    const classes = useStyles({ classes: classesOverride });
+const Datagrid = props => {
+    const classes = useStyles(props);
```
JulienMattiussi: According to MUI documentation, I doubt that works (or even worked before)
because we never use given props in the `makeStyles(theme => ({...}))` few lines upper

https://material-ui.com/styles/basics/#adapting-based-on-props

IMO, `useStyles();` would do the same

packages/ra-core/src/controller/useListController.ts
```diff
@@ -209,10 +205,10 @@ const useListController = (props: ListProps): ListControllerProps => {
         displayedFilters: query.displayedFilters,
         filterValues: query.filterValues,
         hasCreate,
-        // ids might be null if the resource has not been initialized yet (custom routes for example)
-        ids: ids || [],
+        hideFilter: queryModifiers.hideFilter,
+        ids: typeof ids === 'undefined' ? defaultIds : ids,
+        loaded: loaded || defaultIds.length > 0,
```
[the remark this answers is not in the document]
fzaninotto: What's wrong with defaultIds?

packages/ra-ui-materialui/src/list/Datagrid.js
fzaninotto: We're not adapting based on props, we're allowing the user to pass a `classes` props to override the classes we provide. It's undocumented in the material-ui doc, but it works.

packages/ra-core/src/controller/useListController.ts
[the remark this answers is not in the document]
JulienMattiussi: Nothing wrong. It's just that `defaultTotal > 0` seems lighter.

packages/ra-ui-materialui/src/list/Datagrid.js
JulienMattiussi: Ok.

fancyaction:
> The requests made by the `<ReferenceInput>` component aren't cached because their controller still uses the `CrudGetMatchingAccumulate` action creator. That means that the `dataProvider.getList()` it makes to fetch the possible references is done by the old `fetch` saga (called by the `accumulate` saga) instead of the `useDataProvider` hook. As the cache is only implemented in the new hook, `<ReferenceInput>` doesn't use the cache.
> 
> We need to migrate the `useReferenceInputController` code so that it doesn't use the `accumulate` saga anymore.

What's the status on this? Is this being tracked and/or worked on?

I hit a serious issue this week with our app using a new form that can load in with 20+ child line inputs (via ArrayInput/SimpleFormIterator). I implemented the new caching features but it's not a network issue -- it's a scripting issue and it looks like it's due to the old fetch saga. It really sinks the app and makes a borderline unusable user experience. I think the above issue might address what I'm dealing with. If I can find some free time, I plan to try to look into updating the useReferenceInputController and try my best.

fzaninotto:
This problem is still present, not being worked on, and tracked by issue #4531 that I just created thanks to your comment. 

If you feel like giving it a try, we'd appreciate the help!

gomcodoctor:
look like requestSignature  used in useGetList for cachedRequests is different than one sent in action in requestPayload

``    const requestSignature = JSON.stringify({ pagination, sort, filter });
``

in requestSignature  filter variable do not have navigation and sort data, but in requestPayload its there.
due to this useGetList will never return cached data

fzaninotto:
@gomcodoctor I don't know what you're talking about. The request signatures computed in useGetList and in the cachedRequests reducer are the same for me. If you've found a bug, please open a new issue with a way to reproduce it.

gomcodoctor:
I will create a new issue.. I think it's bug

AymanElarian:
> > The requests made by the `<ReferenceInput>` component aren't cached because their controller still uses the `CrudGetMatchingAccumulate` action creator. That means that the `dataProvider.getList()` it makes to fetch the possible references is done by the old `fetch` saga (called by the `accumulate` saga) instead of the `useDataProvider` hook. As the cache is only implemented in the new hook, `<ReferenceInput>` doesn't use the cache.
> > We need to migrate the `useReferenceInputController` code so that it doesn't use the `accumulate` saga anymore.
> 
> What's the status on this? Is this being tracked and/or worked on?
> 
> I hit a serious issue this week with our app using a new form that can load in with 20+ child line inputs (via ArrayInput/SimpleFormIterator). I implemented the new caching features but it's not a network issue -- it's a scripting issue and it looks like it's due to the old fetch saga. It really sinks the app and makes a borderline unusable user experience. I think the above issue might address what I'm dealing with. If I can find some free time, I plan to try to look into updating the useReferenceInputController and try my best.

did you found any solution yet , I face the same issue ?

PR #4392, opened 2020-02-05
304 noop

ThieryMichel:
Allow the dataProvider to returns a NOOP result, indicating there is no change to apply. Useful for request with status 304 (Not modified) for example.


Luwangel:
Needs a rebase (and a description) :wink:

ThieryMichel:
> When the server returns a NOOP, this skips the success side effects, so that's a problem. Also, seeing as you had to change all hooks, I wonder if this should not be a dataProvider feature, ir if it's still relevant now that we have the application cache.

I think you are right, now that there is the application cache this seems rather redundant.

PR #4394, opened 2020-02-05
Fix Default SimpleFormIterator Push Value

### LINKED ISSUE #2489: defaultValue of Inputs inside ArrayInput not working

### LINKED ISSUE #3249: ArrayInput does not like primitive values

jdemangeon:
Closes #2489
Closes #3249

When we use the `SimpleFormIterator` inside an array of scalar values (`ArrayInput`) displayed as `TextInput`, the famous `[Object object]` value is displayed on "+ Add" click. 

Example:

```
import { SimpleFormIterator } from "react-admin";

/* recipients = ["foo@bar.com", "bar@foo.com", ....] */

<ArrayInput source="recipients">
      <SimpleFormIterator>
          <TextInput />
      </SimpleFormIterator>
</ArrayInput>
```
Results to:

image

So, to avoid this, we should inject `undefined` instead of an empty `{}`.

Here is a temporary workaround for those who are having the same problem:

```
import React from "react";
import { SimpleFormIterator } from "react-admin";

export const SimpleFormIteratorWithUndefinedDefault = props => {
  const originPush = props.fields.push;
  props.fields.push = data =>
    originPush(JSON.stringify(data) === "{}" ? undefined : data);

  return <SimpleFormIterator {...props} />;
};

```

PS: This "display" is the default one used in ApiPlatform Admin for the `array` field type (https://github.com/api-platform/admin/blob/master/src/InputGuesser.js#L90)

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.js
@@ -105,7 +105,7 @@ export class SimpleFormIterator extends Component {
     addField = () => {
         const { fields } = this.props;
         this.ids.push(this.nextId++);
-        fields.push({});
+        fields.push(undefined);
     };
 
     render() {
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.spec.js
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.spec.js
@@ -1,0 +1,140 @@
+import { cleanup, fireEvent, wait, getByText } from '@testing-library/react';
+import React from 'react';
+import { renderWithRedux } from 'ra-core';
+
+import { SimpleFormIterator } from './SimpleFormIterator';
+import TextInput from '../input/TextInput';
+import { ArrayInput } from '../input';
+import SimpleForm from './SimpleForm';
+
+describe('<SimpleFormIterator />', () => {
+    afterEach(cleanup);
+
+    it('should display an add item button at least', () => {
+        const { getByText } = renderWithRedux(
+            <SimpleForm>
+                <ArrayInput source="emails">
+                    <SimpleFormIterator translate={x => x}>
+                        <TextInput source="email" />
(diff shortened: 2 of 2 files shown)


jdemangeon:
Oops, i've seen that another PR covering this issue (https://github.com/marmelab/react-admin/pull/4306/files).

PR #4390, opened 2020-02-05
Fix FilterForm layout broken by bottom margin on Inputs

fzaninotto:
Closes #4389

Before:

image

After:

image

PR #4410, opened 2020-02-11
Fix saveOnEnter : Use normal click on Save button and fix helperText linked issue 

### LINKED ISSUE #4170: Unable to submit a form using the keyboard
**What you were expecting:**
To be able to complete and submit a create form using the keyboard

**What happened instead:**
You can tab through and fill out all of the inputs but once you get to the save button pressing space or enter does not submit the form. There doesn't appear to be any way to submit the form using the keyboard.

**Steps to reproduce:**
Go to https://marmelab.com/react-admin-demo/#/customers/create
Enter the details for the new customer using the keyboard
Tab down to the "Save" button
Press space or enter. Space shows an animation like it is going to submit but nothing happens.

**Related code:**
Possibly related to this section of the RA code stopping the keyboard input from submitting the form https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/button/SaveButton.js#L85

```
// As we handle the "click" through the mousedown event, we have to make sure we cancel
    // the default click in case the issue mentionned above does not occur.
    // Otherwise, this would trigger a standard HTML submit, not the final-form one.
    const handleClick = event => {
        event.preventDefault();
        event.stopPropagation();
    };
```

**Other information:**
This worked previously with RA2. 

**Environment**

* React-admin version: 3.0.4
* Last version that did not exhibit the issue (if applicable): 2.9
* Browser: Chrome

JulienMattiussi:
Closes #4170 

- [x] Use normal click on Save button
- [x] Fix helperText place issue on ReferenceInput
- [x] Fix helperText place issue on RichTextInput
- [x] Fix tests

## Fixes and behaviour

- The Create or Edit Forms can now be validated with a click or with the Enter key
- The helperText place is reserved for all inputs to ensure the Save button won't move down if a requirement text fires at validation.
- It's possible to disable this default helperText by adding helperText={false} on inputs
- The ReferenceInputs in Filters won't have a bad margin anymore.
- ReferenceInput doesn't accept an helperText value, it should be set on the child component

## Screenshots

SubmitOnEnter

image

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -127,7 +127,7 @@ describe('Create Page', () => {
         EditPage.delete();
     });
 
-    it.only('should redirect to edit page after submit on enter', () => {
+    it('should redirect to edit page after submit on enter', () => {
         const values = [
             {
                 type: 'input',
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -127,6 +127,38 @@ describe('Create Page', () => {
         EditPage.delete();
     });
 
+    it.only('should redirect to edit page after submit on enter', () => {
+        const values = [
+            {
+                type: 'input',
+                name: 'title',
+                value: 'Test title',
+            },
+            {
+                type: 'textarea',
+                name: 'teaser',
+                value: 'Test teaser',
+            },
+            {
+                type: 'rich-text-input',
+                name: 'body',
+                value: 'Test body',
+            },
+        ];
+
+        CreatePage.setValues(values);
+        CreatePage.submitWithKeyboard();
+        EditPage.waitUntilVisible();
+        cy.get(EditPage.elements.input('title')).should(el =>
+            expect(el).to.have.value('Test title')
+        );
+        cy.get(EditPage.elements.input('teaser')).should(el =>
+            expect(el).to.have.value('Test teaser')
+        );
+
+        EditPage.delete();
+    });
+
     it('should redirect to show page after create success with "Save and show"', () => {
         const values = [
             {
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -42,7 +42,7 @@ describe('<SaveButton />', () => {
             </TestContext>
         );
 
-        fireEvent.mouseDown(getByLabelText('ra.action.save'));
+        fireEvent.click(getByLabelText('ra.action.save'));
         expect(onSubmit).toHaveBeenCalled();
     });
 
@@ -55,7 +55,7 @@ describe('<SaveButton />', () => {
             </TestContext>
         );
 
-        fireEvent.mouseDown(getByLabelText('ra.action.save'));
+        fireEvent.click(getByLabelText('ra.action.save'));
         expect(onSubmit).not.toHaveBeenCalled();
     });
 
@@ -77,7 +77,7 @@ describe('<SaveButton />', () => {
             </TestContext>
         );
 
-        fireEvent.mouseDown(getByLabelText('ra.action.save'));
+        fireEvent.click(getByLabelText('ra.action.save'));
         expect(dispatchSpy).toHaveBeenCalledWith({
             payload: {
                 message: 'ra.message.invalid_form',
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -3,6 +3,10 @@ import React from 'react';
 import expect from 'expect';
 import { TestContext } from 'ra-core';
 
+import Create from '../detail/Create';
+import SimpleForm from '../form/SimpleForm';
+import Toolbar from '../form/Toolbar';
+import TextInput from '../input/TextInput';
 import SaveButton from './SaveButton';
 
 describe('<SaveButton />', () => {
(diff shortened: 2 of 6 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -390,11 +391,13 @@ const AutocompleteInput: FunctionComponent<
                                 htmlFor: id,
                             })}
                             helperText={
-                                <InputHelperText
-                                    touched={touched}
-                                    error={error}
-                                    helperText={helperText}
-                                />
+                                disableHelperText ? null : (
```
djhi: Why do you introduce another prop for that? Wouldn't it be better to accept `false` for `helperText` ?

djhi: See https://github.com/marmelab/react-admin/pull/4390

JulienMattiussi: Didn't see this one (I was thinking all filters was always broken due to ReferenceInput issue)
Will change my PR

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -283,6 +283,7 @@ export const ReferenceInputView: FunctionComponent<ReferenceInputViewProps> = ({
         setFilter,
         setPagination,
         setSort,
+        helperText,
```
djhi: Reference inputs (all of them) should not accept an `helperText`. Their child might though. I think we should warn the user when they set the `helperText` prop on any of the reference inputs and explain they should add it on the child instead.

JulienMattiussi: ok with it

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -41,11 +45,53 @@ describe('<SaveButton />', () => {
...
+                </Create>
+            </TestContext>
+        );
+
+        const input = getByLabelText('Title');
+        fireEvent.focus(input);
+        fireEvent.change(input, { target: { value: 'test' } });
+
+        const button = getByLabelText('enterButton');
+        //console.log(button);
```
Luwangel: console.log :(

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -41,11 +45,53 @@ describe('<SaveButton />', () => {
...
+                    handleSubmitWithRedirect={onSubmit}
+                    submitOnEnter={true}
+                    saving={false}
+                />
+            </Toolbar>
+        );
+
+        const { getByLabelText } = render(
+            <TestContext>
+                <Create {...defaultCreateProps}>
```
Luwangel: You don't need a Create to test the SaveButton. It's also broken when using directly the SimpleForm.

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -41,8 +45,50 @@ describe('<SaveButton />', () => {
                 />
             </TestContext>
         );
+        fireEvent.click(getByLabelText('ra.action.save'));
+        expect(onSubmit).toHaveBeenCalled();
+    });
+
+    it('should trigger submit action when Enter key if no saving is in progress', () => {
```
fzaninotto:
```suggestion
    it('should trigger the submit action on Enter when submitOnEnter is true if no saving is in progress', () => {
```

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -41,8 +45,50 @@ describe('<SaveButton />', () => {
...
+                </Create>
+            </TestContext>
+        );
+
+        const input = getByLabelText('Title');
+        fireEvent.focus(input);
+        fireEvent.change(input, { target: { value: 'test' } });
 
-        fireEvent.mouseDown(getByLabelText('ra.action.save'));
+        const button = getByLabelText('enterButton');
```
fzaninotto: unused

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -268,6 +269,15 @@ export const ReferenceInputView: FunctionComponent<ReferenceInputViewProps> = ({
           }
         : meta;
 
+    // helperText should never be set on ReferenceInput, only in child component
+    // But in a Filter component, the child helperText have to be forced to false
+    warningLog(
+        helperText !== undefined && helperText !== false,
+        "<ReferenceInput> doesn't accept an helperText value, it should be set on the child component"
```
fzaninotto:
```suggestion
        "<ReferenceInput> doesn't accept a helperText prop. Set the helperText prop on the child component instead"
```

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
JulienMattiussi: I tried various things, the test never pass.
I don't know why
Hope that'll do it in e2e

cypress/integration/create.js
```diff
@@ -127,6 +127,38 @@ describe('Create Page', () => {
         EditPage.delete();
     });
 
+    it.only('should redirect to edit page after submit on enter', () => {
```
fzaninotto: .only detected

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -45,50 +41,8 @@ describe('<SaveButton />', () => {
                 />
             </TestContext>
         );
-        fireEvent.click(getByLabelText('ra.action.save'));
-        expect(onSubmit).toHaveBeenCalled();
-    });
-
-    it('should trigger submit action when Enter key if no saving is in progress', () => {
```
fzaninotto: Why did you remove this test?

JulienMattiussi: It never worked, And I never understood why. I replaced it by an e2e doing same

PR #4430, opened 2020-02-19
Throw an error when useGetOne is called for an undefined resource

fzaninotto:
## Problem

`useQuery` works even without a `<Resource>` because it doesn't rely on the Redux store. However, developers may want to use the more specialized hooks, like `useGetOne()`. These only work if there is a `<Resource>`, but for now if there is no resource the hook returns null - which is not a good developer experience.

## Solution

This change throws an error in case of missing Resource for useGetOne. It's a start. It's a bit harder to do for other data Provider hooks, so let's start with this baby step/


fzaninotto:
This caused a regression in 3.6.0, which should be fixed by #4990 (released in 3.6.3).

PR #4432, opened 2020-02-19
Fix error in CoreAdminUI in development mode

### LINKED ISSUE #4322: CoreAdminUI with dynamic list of Resource does not render in production build
**What you were expecting:**
I expect the production build would render in the same way as in development server

**What happened instead:**
After fresh yarn && yarn build, the production build renders as a white screen.

**Steps to reproduce:**

First I got the tutorial example from here, the tutorial is able to render in both development server and build
https://github.com/marmelab/react-admin/tree/master/examples/tutorial

Then I added the third code snippet in `CoreAdmin.tsx` to make it support dynamic list of Resources
https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/core/CoreAdmin.tsx

The final code is on code sandbox. I tried to make the example simple, so I only modified App.js by directly put the `<Resources>` as children of `<CoreAdminUI>`
https://codesandbox.io/s/ecstatic-ganguly-j4e35

The code above is able to run on codesandbox, the development server, but not able to run after build. I used `serve -s build` to serve the transpiled code.
I believe it can be reproduced and it is not a build toolchain issue since the original tutorial example is able to run after build.

**Related code:**

The issue should be reproduced locally using `yarn build`
```
https://codesandbox.io/s/ecstatic-ganguly-j4e35
```

**Other information:**

After modified my code to load a dynamic list of resources in the project, I also experienced similar translation broken issue in #3000 and #3007 for the production build but not on development server.

**Environment**

* React-admin version: ^3.0.0
* Last version that did not exhibit the issue (if applicable): 2.9.9 (This version does not have CoreAdminContext and CoreAdminUI)
* React version: ^16.9.0
* Browser: Firefox 71
* Stack trace (in case of a JS error): The console is empty

fzaninotto:
Closes #4322


examples/simple/webpack.config.js
```diff
@@ -36,6 +36,9 @@ module.exports = {
             'OptionTextElement',
             'RedirectionSideEffect',
             'RefreshSideEffect',
+            'AdminUIProps',
```
fzaninotto: We should spend some time upgrading babel, this is forced by a bug that was fixed a few months ago

PR #4447, opened 2020-02-26
Add cache to usePermissions to avoid uncessessary render

djhi:


fzaninotto:
This is not a bug fix, and as such it should be PRed against `next` rather than `master`

fzaninotto:
tests fail, needs rebase

fzaninotto:
needs to be rebased on `next`

fzaninotto (approved):

fzaninotto:
In fact, such a cache was already introduced in #5607 under another name. I don't think that we should change the default `usePermissions`, and I think we should revert this. What do you think?

PR #4450, opened 2020-02-26
Convert SimpleFormIterator to function component

WiXSL:


packages/ra-ui-materialui/src/form/SimpleFormIterator.js
```diff
@@ -64,177 +63,171 @@ const styles = theme =>
...
+    variant,
+    margin,
+    TransitionProps,
+    defaultValue,
+}) => {
+    const translate = useTranslate();
+    const classes = useStyles({ classes: classesOverride });
+
+    // we need a unique id for each field for a proper enter/exit animation
+    // but redux-form doesn't provide one (cf https://github.com/erikras/redux-form/issues/2735)
```
fzaninotto: we're not using redux-form anymore

packages/ra-ui-materialui/src/form/SimpleFormIterator.js
```diff
@@ -64,177 +63,171 @@ const styles = theme =>
...
+    // so we keep an internal map between the field position and an auto-increment id
+    let nextId =
```
fzaninotto: I don't understand how your code can be working. 

The right way to port a component property from this to a function component is to use a ref, otherwise the `nextId` and `ids` variables will be reset at each render.

WiXSL: Sorry I couldn't finish yesterday.

packages/ra-ui-materialui/src/form/SimpleFormIterator.js
WiXSL: Ok.

fzaninotto:
Thanks!

PR #4456, opened 2020-02-27
[FIX] LoadingIndicator show infinite loading after Unauthorized error

Developerius:
Requests with Unauthorized error dispatch only FETCH_START.


fzaninotto:
Good catch, thanks!

PR #4448, opened 2020-02-26
Convert ra-ui-materialui inputs, fields & layouts to typescript

josephktcheung:
1. Convert the following components to typescript:
    - `Link`
    - `ReferenceArrayField`
    - `ReferenceField`
    - `ReferenceManyField`
    - `ResettableTextField`
    - `AppBar`
    - `CardContentInner`
    - `DeviceTestWrapper`
    - `Error`
    - `Layout`
    - `LinearProgress`
    - `Loading`
    - `LoadingIndicator`
    - `NotFound`
    - `Notification`
    - `Responsive`
    - `Sidebar`
    - `Title`
    - `TitleForRecord`
    - `TopToolbar`
    - `UserMenu`
2. Export types from `ra-core`

PR #4459, opened 2020-02-28
Convert ra-ui-materialui fields to typescript

josephktcheung:
This is to break down #4448 for easier review process. I will close #4448 and please continue our discussion here, thanks.

Convert the following components into typescript:
* `ReferenceField`
* `ReferenceArrayField`
* `ReferenceManyField`


fzaninotto:
Thanks!

PR #4464, opened 2020-03-02
Allow user to override SelectInput style

josephktcheung:


packages/ra-ui-materialui/src/input/SelectInput.tsx
```diff
@@ -169,7 +170,7 @@ const SelectInput: FunctionComponent<
     ...rest
 }) => {
     const translate = useTranslate();
-    const classes = useStyles({});
+    const classes = useStyles({ classes: classesOverride });
```
fzaninotto: We've noticed that this syntax forces a rerender even when the props don't change (because the object passed to `useStyles` is always new). The right way to override styles is to pass the entire `props` object, without changing it, to `useStyles`. This will require many changes in the current codebase, and it's not the subject of this PR. But for changes like yours, we prefer starting with the right syntax right away.

josephktcheung: I see. Thanks for the explanation

josephktcheung: @fzaninotto I plan to use typescript 3.7's optional chaining & Nullish Coalescing like this
```typescript
interface Props {
    minWidth?: number;
}

export type SelectInputProps = ChoicesProps &
    InputProps<TextFieldProps> &
    Omit<TextFieldProps, 'label' | 'helperText'> &
    Props;

const useStyles = makeStyles<Theme, SelectInputProps>(
    theme => ({
        input: props => ({
            minWidth: props?.minWidth ?? theme.spacing(20),
        }),
    }),
    { name: 'RaSelectInput' }
);
```
However, prettier 1.17.1 doesn't support typescript 3.7. Would you consider upgrading prettier to 1.19.1?

fzaninotto: We won't upgrade tooling at this stage. 

Also, minWidth is a new feature, I suggest you make another PR with it, explaining your rationale.

josephktcheung: got it, thanks

PR #4469, opened 2020-03-03
Fix <SimpleListLoading> `link` style is not implemented error.

WiXSL:
You get an error using `<SimpleListLoading>` component :

```
Material-UI: the key `link` provided to the classes prop is not implemented in undefined.
You can only override one of the following: primary,tertiary. 
    in SimpleListLoading (created by SimpleList)
    in SimpleList (at licencia.js:61)
    in LicenciaList (created by WithPermissions)
    in Resource (at App.js:47)
    in App (at src/​index.js:7)`
`link` style is defined in `<SimpleList>`

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList.js
+++ b/packages/ra-ui-materialui/src/list/SimpleList.js
@@ -14,13 +14,19 @@ import SimpleListLoading from './SimpleListLoading';
 
 const useStyles = makeStyles(
     {
+        tertiary: { float: 'right', opacity: 0.541176 },
+    },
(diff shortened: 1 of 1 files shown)


fzaninotto:
The styles for `LinkOrNot` cannot be overridden by the user. So instead of having one `useStyles` hook for two components, we'd better have two styles hooks, one for each component.

WiXSL:
Well, LinkOrNot is never exported anyways, should y change it to another file and export it as well?

fzaninotto:
No, it's fine like that, thanks.

PR #4477, opened 2020-03-05
Add documentation for use of Fields inside SimpleFormIterator

### LINKED ISSUE #2500: Fields are not rendered in arrayinput with simpleformiterator
**What you were expecting:**
To see my fields

**What happened instead:**
An empty span is rendered

**Steps to reproduce:**
```jsx
        <SimpleForm>
            <ArrayInput source="fases">
                <SimpleFormIterator disableAdd disableRemove>
                    <TextField label="Fase" source="titel" />
                </SimpleFormIterator>
            </ArrayInput>
        </SimpleForm>
```

does not work but if i change into an input, it works

```jsx
        <SimpleForm>
            <ArrayInput source="fases">
                <SimpleFormIterator disableAdd disableRemove>
                    <TextInput label="Fase" source="titel" />
                </SimpleFormIterator>
            </ArrayInput>
        </SimpleForm>
```

The behavior of DateField is the same...

**Environment**

* React-admin version: 2.4.1
* React version: 16.5.2
* Browser: chrome

JulienMattiussi:
Closes #2500 

## Screenshot

image

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -108,7 +108,7 @@ import { ArrayInput, SimpleFormIterator, DateInput, TextInput } from 'react-admi
 }
 ```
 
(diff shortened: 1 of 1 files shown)


amayatsky:
Is it a good decision to document a getSource(); call and make it an 'official' way to use fields inside the form iterator?
I don't even remember what the getSource() does but I do remember it was some sort of a hack or a work around your framework: https://github.com/marmelab/react-admin/issues/2500#issuecomment-526459968

JulienMattiussi:
No issue with this IMO.
Your workaround works fine.
And as it's the only valuable solution for this need, it's better to document it than to let people trying many bad solutions.

By the way, we already use getSource() in the simple examples (i.e. in PostCreate.js

PR #4479, opened 2020-03-05
[RFR] Add color prop to AppBar

### LINKED ISSUE #4396: AppBar should not use secondary color
Currently `AppBar` uses secondary color as backgroud: 

`<MuiAppBar className={className} color="secondary" {...rest}>`

Material Design suggests:

> A secondary color provides more ways to accent and distinguish your product. Having a secondary color is optional, and should be applied sparingly to accent select parts of your UI.
> 
> Secondary colors are best for:
> 
> Floating action buttons
> Selection controls, like sliders and switches
> Highlighting selected text
> Progress bars
> Links and headlines

So I think `AppBar` should use primary color by default.

WiXSL:
Fixes #4396

### DIFF
--- a/packages/ra-ui-materialui/src/layout/AppBar.js
+++ b/packages/ra-ui-materialui/src/layout/AppBar.js
@@ -132,7 +132,7 @@ AppBar.propTypes = {
     children: PropTypes.node,
     classes: PropTypes.object,
     className: PropTypes.string,
-    color: PropTypes.oneOf(['primary', 'secondary']),
+    color: PropTypes.oneOf(['default', 'inherit', 'primary', 'secondary', 'transparent']),
     logout: PropTypes.element,
     open: PropTypes.bool,
     userMenu: PropTypes.element,
--- a/packages/ra-ui-materialui/src/layout/AppBar.js
+++ b/packages/ra-ui-materialui/src/layout/AppBar.js
@@ -56,7 +56,6 @@ const useStyles = makeStyles(
  * @param {string} color The color of the AppBar
  * @param {Object} logout The logout button component that will be pass to the UserMenu component
  * @param {boolean} open State of the <Admin/> Sidebar
- * @param {string} title The title of the AppBar
  * @param {Function} userMenu A custom UserMenu component for the AppBar. UserMenu component by default
  *
  * @example
@@ -82,7 +81,6 @@ const AppBar = ({
     color = 'secondary',
     logout,
     open,
-    title,
     userMenu,
     ...rest
 }) => {
--- a/packages/ra-ui-materialui/src/layout/AppBar.js
+++ b/packages/ra-ui-materialui/src/layout/AppBar.js
@@ -47,7 +47,35 @@ const useStyles = makeStyles(
     }),
     { name: 'RaAppBar' }
 );
-
+/**
+ * The AppBar component renders a custom MuiAppBar.
+ *
+ * @param {Function} children Must be a function which will be called with the mutate callback
+ * @param {Object} classes CSS class names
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,11 +47,40 @@ const useStyles = makeStyles(
...
 );
-
+/**
+ * The AppBar component renders a custom MuiAppBar.
+ *
+ * @param {Function} children Must be a function which will be called with the mutate callback
+ * @param {Object} classes CSS class names
+ * @param {string} className CSS class applied to the MuiAppBar component
+ * @param {string} color The color of the AppBar
+ * @param {Object} logo //
```
djhi: Would you mind removing this very old prop which is not used at all anymore ?

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,11 +47,40 @@ const useStyles = makeStyles(
...
+ *        </AppBar>
+ *    );
+ *};
+ */
 const AppBar = ({
     children,
     classes: classesOverride,
     className,
+    color = 'secondary',
     logo,
```
djhi: Please remove it too

WiXSL: Done!

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,12 +47,39 @@ const useStyles = makeStyles(
     }),
     { name: 'RaAppBar' }
 );
-
+/**
+ * The AppBar component renders a custom MuiAppBar.
+ *
+ * @param {Function} children Must be a function which will be called with the mutate callback
```
fzaninotto: No, this is wrong. Children should be a React node.

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,12 +47,39 @@ const useStyles = makeStyles(
...
+ * @param {Object} classes CSS class names
+ * @param {string} className CSS class applied to the MuiAppBar component
+ * @param {string} color The color of the AppBar
+ * @param {Object} logout The logout button component that will be pass to the UserMenu component
```
fzaninotto: Not Object, but component

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,12 +47,39 @@ const useStyles = makeStyles(
...
+ * @param {boolean} open State of the <Admin/> Sidebar
+ * @param {string} title The title of the AppBar
```
fzaninotto: no, it's not used - you can remove it.

packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -47,12 +47,39 @@ const useStyles = makeStyles(
...
+ * @param {Function} userMenu A custom UserMenu component for the AppBar. UserMenu component by default
```
fzaninotto: Not Function, but Element

packages/ra-ui-materialui/src/layout/AppBar.js
WiXSL: Ok. I've copied it from another ra components actually.

WiXSL:
Toolbar `disableGutters` prop isn't set anywhere.
Should I remove it, or take it from the AppBar props?

fzaninotto:
Thanks!

PR #4494, opened 2020-03-09
Fix ReferenceArrayInput list when filter is changed

### LINKED ISSUE #4466: Changing the value to null and setting filter on a ReferenceArrayInput does not filter choices properly
**What you were expecting:**

Changing the value of a `ReferenceArrayInput` from, let's say `1`, to `null` and changing the `filter` prop to not include `1` in the results should also remove `1` from choices.

**What happened instead:**

The choices still show `1` and it does not go away until you select a different choice. If you select `1`, the `1` never goes away.

**Steps to reproduce:**

I've provided a CodeSandbox link.
To reproduce the error:

1. Go to comments create page
2. Select a value for Post from the ReferenceArrayInput that I've added. Make sure the value is not `2` or `3` as those are used in the filter. I've shown the current value and filter right below this control.
3. Now, click on the "Add Filter" button next to the control which sets the value to `null` and the filter value to `{id: [2, 3]}`.
4. The value should now be `null` and filter should now be `{id: [2, 3]}`.
5. Click on the control to see the choices now. There will be three choices when there are only supposed to be two. The third one apart from 2 and 3 will be the one you selected before adding the filter.
6. Click on the value that you chose before adding the filter. Now, it's as if the value is selected and it does not matter that the results of the filtered choices do not include that value.

* Here's the sandbox link:
https://5kikk.csb.app/#/comments/create
https://codesandbox.io/s/sparkling-currying-5kikk

Here's a couple of screen captures of the bug:

When selecting an "invalid" choice:

invalid

When selecting a "valid" choice:

valid

**Environment**

* React-admin version: 3.2.3
* Last version that did not exhibit the issue (if applicable): ~
* React version: 16.9.0
* Browser: Chrome
* Stack trace (in case of a JS error): ~

JulienMattiussi:
Closes #4466 

## Issue

When you change the filter of a ReferenceArrayInput and reset its selected value using useForm hook, the previous selected value is always proposed in the list.

## Solution

Limit ReferenceArrayInput proposed list to: 
- effectively selected values
- matching values

## Screenshot

referenceFilter

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
@@ -60,7 +60,23 @@ const useReferenceArrayInputController = ({
         const newIdsToFetch = difference(input.value, inputValue.current);
         // Only get from store ids selected and already fetched
         const newIdsToGetFromStore = difference(input.value, newIdsToFetch);
-
+        /*
+            input.value (current)
+                +------------------------+
+                | ********************** |
+                | ********************** |  inputValue.current (old)
+                | ********** +-----------------------+
+                | ********** | ooooooooo |           |
+                | ********** | ooooooooo |           |
+                | ********** | ooooooooo |           |
+                | ********** | ooooooooo |           |
+                +---|--------|------|----+           |
+                    |        |      |                |
+                    |        |      |                |
+                    |        +------|----------------+
+                    |               |
+            newIdsToFetch    newIdsToGetFromStore
+        */
         // Change states each time input values changes to avoid keeping previous values no more selected
         if (!isEqual(idsToFetch, newIdsToFetch)) {
             setIdsToFetch(newIdsToFetch);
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -105,7 +105,10 @@ const useReferenceArrayInputController = ({
         : referenceRecordsFromStore;
 
     // filter out not found references - happens when the dataProvider doesn't guarantee referential integrity
-    const finalReferenceRecords = referenceRecords.filter(Boolean);
+    // filter out not selected items - happens when the form value is changed by script (i.e. with a filtering Button)
+    const finalReferenceRecords = referenceRecords
+        .filter(Boolean)
+        .filter(item => input.value.includes(item.id));
```
fzaninotto: I think the problem isn't there, it's in the useEffect line 58 that does not properly work in all cases.

Hemant-yadav:
Hi,

Even on ReferenceInput changing the filter doesn't correctly updates the filters.
It persists the original filters. I forked the [Sandbox](https://codesandbox.io/s/icy-sea-vtw6s)

Would you please look on #4508 which references this issue. LGTM tho

JulienMattiussi:
You're right about `ReferenceInput`
We'll try to check this issue soon #4507.

By the way, there is no technical relation between this `RefenceArrayInput` fix and the needed `ReferenceInput` fix. Both their behaviour and issues are different.

Issue #3348, opened 2019-06-19
Making default theme RTL won't apply to some components


mohas:
**What you were expecting:**
for all the components to flip when I specify RTL in theme and `dir="rtl"` on body tag

**What happened instead:**
setting rtl on theme does absolutely nothing.
![ra-rtl](https://user-images.githubusercontent.com/1816728/59797013-f8c6d800-92f3-11e9-9673-1b2c00a52825.PNG)
As you can see many components are not RTL, icons are not flipped, paddings are not flipped, there is `text-align: left;` defined on the body.
I made list and edit components rtl using a jssProvider like:

    const jss = create({ plugins: [...jssPreset().plugins, rtl()] });

    function RTL(props) {
      return (
        <JssProvider jss={jss}>
          {props.children}
        </JssProvider>
      );
    }

Then use it with list and edit and create components:

    <RTL>
      <List {...props} ...

**Steps to reproduce:**

1-Create an RTL theme:

    const theme = {
      direction: 'rtl',
      typography: {
        fontFamily: 'IranSans'
      },
      isRtl: true
    };
    const themeWithDirection = createMuiTheme({...defaultTheme, ...theme});

2-Pass it to Admin component:

    <Admin locale="fa" i18nProvider={i18nProvider} theme={theme} ...

There is very little badly written documentation and answers regarding these issues, so any help is appreciated, thanks.

**Environment**

* React-admin version: 2.9.3
* React version: 16.6.3
* Browser: any

mohas:
I've made further inspections on next, back buttons and why their icons won't flip, turns out material-ui `TablePaginationActions` is doing a correct job regarding theme direction:

    return _react.default.createElement("div", other, _react.default.createElement(_IconButton.default, (0, _extends2.default)({
            onClick: this.handleBackButtonClick,
            disabled: page === 0
          }, backIconButtonProps), theme.direction === 'rtl' ? _ref : _ref2), _react.default.createElement(_IconButton.default, (0, _extends2.default)({
            onClick: this.handleNextButtonClick,
            disabled: page >= Math.ceil(count / rowsPerPage) - 1
          }, nextIconButtonProps), theme.direction === 'rtl' ? _ref3 : _ref4));

but `PaginationActions` in react-admin does not regard theme direction when making action buttons, so my guess is the same problem lies everywhere in react-admin components so how can I introduce theme prop into those components?

mohas:
I made a gist to show how to fix it, can I make this into a pull request?
[...]

PR #4496, opened 2020-03-09
Fix pagination buttons are in the wrong direction in RTL languages

alikazemkhanloo:
this PR will change the chevron direction in PaginationActrions for RTL, as in the gist suggested in #3348. But using hooks to get the theme.

### DIFF
--- a/packages/ra-ui-materialui/src/list/PaginationActions.js
+++ b/packages/ra-ui-materialui/src/list/PaginationActions.js
@@ -127,7 +127,11 @@ function PaginationActions(props) {
                     onClick={prevPage}
                     className="previous-page"
                 >
-                    {theme.direction === "rtl" ? <ChevronRight /> : <ChevronLeft />}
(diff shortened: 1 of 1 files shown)


fzaninotto:
I'm not sure I understand. Can you please explain the problem, and paste a screenshot to show the result of your change?

kopax:
RTL means right to left, such as arabic language. 
image

fzaninotto:
I know that, I just don't understand what the problem is with pagination in RTL languages.

alikazemkhanloo:
this is what it looks like in RTL:

Screenshot from 2020-03-10 12-55-29---

JulienMattiussi:
@alikazemkhanloo 
I think your reverted ChevronLeft and ChevronRight in your PR, because you keep the current behaviour when `direction === rtl`.

alikazemkhanloo:
@JulienMattiussi Thanks.

fzaninotto:
thanks!

PR #4500, opened 2020-03-10
Add support for allowEmpty=false in Filter Inputs

### LINKED ISSUE #4499: FilterFormInput always has allowEmpty=true props
**What you were expecting:**
FilterFormInput should take prop allowEmpty from Input component but allowEmpty prop is ignore and always set true. See https://github.com/marmelab/react-admin/blob/v3.3.0/packages/ra-ui-materialui/src/list/FilterFormInput.js#L46

**What happened instead:**
FilterFormInput has always allowEmpty=true despite of input props

**Steps to reproduce:**
Set allowEmpty={false} for Input component
```
  <Filter {...props}>
    <SearchInput source="q" />
    <ReferenceInput
      source="post_id"
      reference="posts"
      allowEmpty={false}
      alwaysOn
    >
    </ReferenceInput>
  </Filter>
```

**Related code:**
CodeSandbox: https://codesandbox.io/s/ecstatic-sea-remus/
(https://remus.csb.app/#/comments)

**Environment**

* React-admin version:
3.3.0
2.9.0
* React version:
16.12
* Browser:
chrome

JulienMattiussi:
Closes #4499 

## Issue 

FilterFormInput has always allowEmpty=true despite of input props

## Solution

We should allow to have filter not emptiable.

### DIFF
--- a/packages/ra-ui-materialui/src/list/FilterFormInput.js
+++ b/packages/ra-ui-materialui/src/list/FilterFormInput.js
@@ -43,6 +43,10 @@ const FilterFormInput = ({
                 </IconButton>
             )}
             {React.cloneElement(filterElement, {
+                allowEmpty:
+                    filterElement.props.allowEmpty === undefined
+                        ? true
+                        : filterElement.props.allowEmpty,
                 resource,
                 record: emptyRecord,
                 variant,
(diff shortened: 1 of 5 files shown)


packages/ra-ui-materialui/src/list/FilterFormInput.js
```diff
@@ -43,7 +43,6 @@ const FilterFormInput = ({
                 </IconButton>
             )}
             {React.cloneElement(filterElement, {
-                allowEmpty: true,
```
fzaninotto: Not good: this is breaking existing forms. You should inspect if the child has defined allowempty first.

JulienMattiussi: :+1:

PR #4501, opened 2020-03-10
Improved PropTypes definition in FormTab component.

WiXSL:


packages/ra-ui-materialui/src/form/FormTab.js
```diff
@@ -73,15 +73,20 @@ const FormTab = ({
...
 FormTab.propTypes = {
+    basePath: PropTypes.string,
     className: PropTypes.string,
     contentClassName: PropTypes.string,
     children: PropTypes.node,
     intent: PropTypes.oneOf(['header', 'content']),
     hidden: PropTypes.bool,
     icon: PropTypes.element,
     label: PropTypes.string.isRequired,
+    margin: PropTypes.Margin,
```
fzaninotto: I think this is a mistake

WiXSL: I'll change it.

WiXSL: Done!

packages/ra-ui-materialui/src/form/FormTab.js
```diff
@@ -73,15 +73,20 @@ const FormTab = ({
...
     intent: PropTypes.oneOf(['header', 'content']),
     hidden: PropTypes.bool,
     icon: PropTypes.element,
     label: PropTypes.string.isRequired,
+    margin: PropTypes.oneOf(['none', 'dense', 'normal']),
     path: PropTypes.string,
+    record: PropTypes.object,
+    resource: PropTypes.string,
     value: PropTypes.string,
+    variant: PropTypes.oneOf(['text', 'outlined', 'contained']),
```
fzaninotto: where did you get these values from? The [material-ui doc](https://material-ui.com/api/text-field/#props) says the possible values are `standard`, `outlined`, and `filled`

WiXSL: You are right.

packages/ra-ui-materialui/src/form/FormTab.js
```diff
@@ -73,15 +73,20 @@ const FormTab = ({
...
     intent: PropTypes.oneOf(['header', 'content']),
     hidden: PropTypes.bool,
     icon: PropTypes.element,
     label: PropTypes.string.isRequired,
+    margin: PropTypes.oneOf(['none', 'dense', 'normal']),
     path: PropTypes.string,
+    record: PropTypes.object,
+    resource: PropTypes.string,
     value: PropTypes.string,
+    variant: PropTypes.oneOf(['standar', 'outlined', 'filled']),
```
fzaninotto: still wrong... 'standard' with a d

fzaninotto:
The multiple comments on this PR show me that you haven't tested the code. I suggest you add unit or integration tests.

packages/ra-ui-materialui/src/form/FormTab.js
WiXSL: You are right I didn't test the last change. Sorry.

WiXSL:
Ok. Could you give me a basic idea of a test regarding the check of the propTypes.

fzaninotto:
Add a test case in the FormTab unit case where you pass all possible props to the TabbedForm (margin, variant, etc), and check that there is no warning when it runs.

fzaninotto:
Thanks!

Issue #4473, opened 2020-03-04
useFilterState does not update permanentFilter value if changed


fancyaction:
**What you were expecting:**
If I use ReferenceInput with AutocompleteInput and set filter prop (aka permanentFilter) to a variable, data filter should update alongside the variable. 

Example: I have user select input for author_id, in the following ReferenceInput->AutoComplete Input for books I want to filter for whatever author_id was selected (pulling value either from formState.values or FormDataConsumer's formData) and filterToQuery on book's title:
`filter={ formData && formData.author_id ? {author_id: formData.author_id} : {} }`
`filterToQuery={(searchText: string) => ({ title: searchText })}`

**What happened instead:**
useFilterState is always setting permanentFilter to whatever it originally loaded in with. In the above case, it will always be undefined because formData.author_id doesn't update until user selects author. Once they select author_id=23 it will still be read from the original undefined value.

**Steps to reproduce:**
Set a RefeneceInput's filter prop to another input on form (see above example or CodePen below)

**Related code:**
Please go to posts -> PostCreate.js for code example:
https://codesandbox.io/s/focused-snyder-ereg7

It's hard to demonstrate fully with data set there but if you open console you'll see on tag_id select an initial call is made with `"filter":{"tag_id":16,"body":""}}` but every following call looks like `"filter":{"body":"test"}}` because tag_id (taken from dataForm.tag_id) is read as `undefined` in useFilterState.

```
       <ReferenceInput label="Tag" source="tag_id" reference="tags">
          <AutocompleteInput />
        </ReferenceInput>
        <FormDataConsumer>
          {({ formData, ...rest }) => (
            <ReferenceInput
              label="Comment"
              source="comment_id"
              reference="comments"
              filter={{ tag_id: formData.tag_id }}
              // filter={formData.tag_id ? { tag_id: formData.tag_id } : {}}
              filterToQuery={searchText => ({ body: searchText })}
              {...rest}
            >
              <AutocompleteInput
                shouldRenderSuggestions={val => 2 < val.trim().length}
                optionText="body"
              />
            </ReferenceInput>
          )}
        </FormDataConsumer>
```

**Other information:**
In useFilterState, replacing `permanentFilter` with `permanentFilterProp.current` (e.g. latest filter prop value) in setFilterValue seems to fix but unsure if that could cause issues elsewhere.

This may very well relate to issue #3921 

[...]

PR #4508, opened 2020-03-11
Fix useFilterState: useCallback listens on permanent filter changes

### LINKED ISSUE #4507: Dynamic filter support in ReferenceInput
**What you were expecting:**
Changing the filter value in ReferenceInput should make query with new filters.

**What happened instead:**
Changing filters doesn't work, instead the query uses the same old filter value.

**Steps to reproduce:**
Change filter of ReferenceInput at runtime by setTimeOut or by another formInput value, and pass the new filter to ReferenceInput.

**Related code:**
https://codesandbox.io/s/cold-star-f2rfi

Go to comment edit page, and check the console logs printed in dataProvider. The filter for ReferenceInput of Posts was changed after 2 seconds, still the latest query in dataProvider has old filter in params.

```
const [filter, setFilter] = React.useState({ category: "tech" });
```

```
  React.useEffect(() => {
    setTimeout(() => {
      setFilter({ category: "lifestyle" });
    }, 2000);
  }, []);
```

```
<ReferenceInput
  source="post_id"
  reference="posts"
  perPage={15}
  sort={{ field: "title", order: "ASC" }}
  fullWidth
  filter={filter}
>
  <AutocompleteInput
    matchSuggestion={(filterValue, suggestion) => true}
    optionText={<OptionRenderer />}
    inputText={inputText}
  />
</ReferenceInput>
```

**Other information:**
I have also raised a PR on my approach to solve this issue. Here is the link.
https://github.com/marmelab/react-admin/pull/4508

**Environment**

* React-admin version: 3.2.1
* React version: 16.12.0
* Browser: Chrome

tarunraj95:
Closes #4507 

Updating filter prop on referenceInput wasn't updating the filters for fetching. The last query uses the first filters provided.

Listening on permanentFilter on setFilter callback solved the issue.

### DIFF
--- a/packages/ra-core/src/controller/useFilterState.ts
+++ b/packages/ra-core/src/controller/useFilterState.ts
@@ -80,7 +80,7 @@ export default ({
             });
             latestValue.current = value;
         }, debounceTime),
-        [ permanentFilter ]
+        [permanentFilter]
     );
 
     return {
--- a/packages/ra-core/src/controller/useFilterState.ts
+++ b/packages/ra-core/src/controller/useFilterState.ts
@@ -80,7 +80,7 @@ export default ({
             });
             latestValue.current = value;
         }, debounceTime),
-        [permanentFilter]
+        [ permanentFilter ]
     );
 
     return {
--- a/packages/ra-core/src/controller/useFilterState.ts
+++ b/packages/ra-core/src/controller/useFilterState.ts
@@ -80,7 +80,7 @@ export default ({
             });
             latestValue.current = value;
         }, debounceTime),
-        []
+        [permanentFilter]
     );
 
     return {


fzaninotto:
Thanks!

fancyaction:
This isn't working for me. My input gets wiped in AutocompleteInput with ReferenceInput when using the dynamic filter. I see the filter now updates correctly, but any input in the field gets cleared. 

I'm guessing the "permenantFilter" changes every render so it updates the setFilter every render which in turn is updating the autocompletInput's onChange every render, wiping my input. I've noticed this issue come up quite a bit since using hooks in my own projects. 

I had a duplicate issue #4473 . I think my suggested fix there might work: Remove this fix (which is causing callback to change every render) and in that callback update `permanentFilterProp` in the `setFilterValue`. Not 100% on this because I can't run and test on my local ...

On a side note, can you help me set up my local env so I can work on this bug and others? I made [post here](https://stackoverflow.com/questions/60780889/react-admin-make-commands-are-failing) asking for help with the Make process. Also, having trouble getting the whole npm link / yarn link process to work while trying to debug. It connects to my local RA node_module but then doesn't update if I make changes to my local RA module -- maybe this is due to my Webpack or Babel setup? I have a feeling all of this is due to my Windows setup. 

I'd love to contribute to RA and some of its 3rd party modules (react-admin-date-inputs, specifically) but blocked by these issues. Any assistance would be greatly appreciated!

PR #4511, opened 2020-03-11
Allow SelectArrayInput to be variant=outlined

### LINKED ISSUE #4439: SelectInput and SelectArrayInput don't respect options prop when inside ReferenceInput component
**What you were expecting:**
The SelectInput and SelectArrayInput in ReferenceInput respect fullWidth, variant in options prop

**What happened instead:**
The SelectInput and SelectArrayInput in ReferenceInput  don't respect fullWidth, variant in options prop

**Steps to reproduce:**

**Related code:**

CodeSandbox:  https://codesandbox.io/s/jovial-poitras-mjg9y?fontsize=14&hidenavigation=1&theme=dark
steps to produce:
Edit any of Post, go to MISCELLANEOUS tab, The tags input still remain "filled" variant even I change to "outlined"

JulienMattiussi:
Closes #4439 

## Issue

`variant="outlined"` doesn't work on `SelectArrayInput`

## Solution

Apply MUI method to build a valid Select with `variant="outlined"`
Forked from MUI demo:
https://codesandbox.io/s/material-demo-684eu

Basic usage
```js
<SelectArrayInput
    fullWidth
    variant="outlined"
    source="tags"
    choices={[
       { name: 'Tech', id: 'tech' },
       { name: 'Lifestyle', id: 'lifestyle' },
    ]}
/>
```

ReferenceArrayInput usage
```js
<ReferenceArrayInput
    reference="tags"
    source="tags"
    fullWidth
    filter={{ published: true }}
>
    <SelectArrayInput variant="outlined" />
</ReferenceArrayInput>
```

## Screenshot

image

### DIFF
--- a/examples/simple/src/posts/PostEdit.js
+++ b/examples/simple/src/posts/PostEdit.js
@@ -20,7 +20,6 @@ import {
     ReferenceManyField,
     ReferenceInput,
     SelectInput,
-    SelectArrayInput,
     SimpleFormIterator,
     TabbedForm,
     TextField,
--- a/examples/simple/src/posts/PostEdit.js
+++ b/examples/simple/src/posts/PostEdit.js
@@ -20,6 +20,7 @@ import {
     ReferenceManyField,
     ReferenceInput,
     SelectInput,
+    SelectArrayInput,
     SimpleFormIterator,
     TabbedForm,
     TextField,
--- a/packages/ra-ui-materialui/package.json
+++ b/packages/ra-ui-materialui/package.json
@@ -26,7 +26,7 @@
         "watch": "tsc --outDir esm --module es2015 --watch"
     },
     "devDependencies": {
-        "@material-ui/core": "^4.3.3",
+        "@material-ui/core": "^4.9.5",
         "@material-ui/icons": "^4.2.1",
         "@material-ui/styles": "^4.3.3",
         "@testing-library/react": "^8.0.7",
@@ -51,7 +51,7 @@
         "rimraf": "^2.6.3"
     },
     "peerDependencies": {
-        "@material-ui/core": "^4.3.3",
+        "@material-ui/core": "^4.9.5",
         "@material-ui/icons": "^4.2.1",
         "@material-ui/styles": "^4.3.3",
         "final-form": "^4.18.5",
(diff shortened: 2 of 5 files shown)


examples/simple/src/posts/PostEdit.js
```diff
@@ -126,7 +127,7 @@ const PostEdit = ({ permissions, ...props }) => (
                     source="tags"
                     filter={{ published: true }}
                 >
-                    <AutocompleteArrayInput fullWidth />
+                    <SelectArrayInput variant="outlined" fullWidth />
```
fzaninotto: to be removed after tests

packages/ra-ui-materialui/package.json
```diff
@@ -26,7 +26,7 @@
         "watch": "tsc --outDir esm --module es2015 --watch"
     },
     "devDependencies": {
-        "@material-ui/core": "^4.3.3",
+        "@material-ui/core": "^4.9.5",
```
fzaninotto: please find the earliest revision where this is fixed

yarn.lock
```diff
@@ -1171,6 +1171,24 @@
     react-is "^16.8.0"
     react-transition-group "^4.3.0"
 
+"@material-ui/core@^4.9.5":
```
fzaninotto: Not good: you should replace a dependency, not add one

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -161,14 +165,18 @@ const SelectArrayInput: FunctionComponent<
         ...rest
     } = props;
     const classes = useStyles(props);
+    const inputLabel = useRef(null);
+    const [labelWidth, setLabelWidth] = useState(0);
+    useEffect(() => {
+        setLabelWidth(inputLabel.current.offsetWidth);
```
jdemangeon, line 171: inputLabel.current can be null at mounting. Can you add a check?

fzaninotto: I don't think so. the effect will run after the first render, and at that time, the ref will be defined.

PR #4518, opened 2020-03-12
Add a default order prop in List Fields

### LINKED ISSUE #2736: Choose the default order value for sorting

JulienMattiussi:
Closes #2736

## Issue

In Lists, columns are always sorted by `ASC` order by default
Sometimes, it's not the main desired behaviour (recent dates before, .....)

## Usage

```js
<TextField
    source="title"
    sortBy="title"
    sortByOrder="DESC"
/>
```

## Documentation

image

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -552,7 +552,6 @@ export const PostList = (props) => (
 
 By default, when the user clicks on a column header, the list becomes sorted in the ascending order. You change this behavior by setting the `sortByOrder` prop to `"DESC"`:
 
-{% raw %}
 ```jsx
 // in src/posts.js
 import React from 'react';
@@ -575,7 +574,6 @@ export const PostList = (props) => (
     </List>
 );
 ```
-{% endraw %}
 
 ### Permanent Filter
 
--- a/docs/List.md
+++ b/docs/List.md
@@ -550,7 +550,7 @@ export const PostList = (props) => (
 
 ### Specify Sort Order
 
-By default, a column is sorted by the `ASC` order. You can use `DESC` via the `sortByOrder` property:
+By default, when the user clicks on a column header, the list becomes sorted in the ascending order. You change this behavior by setting the `sortByOrder` prop to `"DESC"`:
 
 {% raw %}
 ```jsx
--- a/docs/List.md
+++ b/docs/List.md
@@ -550,7 +550,7 @@ export const PostList = (props) => (
 
 ### Specify Sort Order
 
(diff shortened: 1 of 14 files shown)


docs/List.md
```diff
@@ -548,6 +548,35 @@ export const PostList = (props) => (
 ```
 {% endraw %}
 
+### Specify Sort Order
+
+By default, a column is sorted by the `ASC` order. You can use `DESC` via the `sortByOrder` property:
+
+{% raw %}
```
fzaninotto: no need for this since the code snippet doesn't contain any double mustache

docs/List.md
```diff
@@ -548,6 +548,35 @@ export const PostList = (props) => (
 ```
 {% endraw %}
 
+### Specify Sort Order
+
+By default, a column is sorted by the `ASC` order. You can use `DESC` via the `sortByOrder` property:
```
fzaninotto:
```suggestion
By default, when the user clicks on a column header, the list becomes sorted in the ascending order. You change this behavior by setting the `sortByOrder` prop to `"DESC"`:
```

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -156,8 +156,11 @@ const useListParams = ({
     }, requestSignature); // eslint-disable-line react-hooks/exhaustive-deps
 
     const setSort = useCallback(
-        (newSort: string) =>
-            changeParams({ type: SET_SORT, payload: { sort: newSort } }),
+        (newSort: string, newOrder?: string) =>
+            changeParams({
+                type: SET_SORT,
+                payload: { sort: newSort, order: newOrder || undefined },
```
fzaninotto: no need for the `|| undefined` here

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -156,8 +156,11 @@ const useListParams = ({
     }, requestSignature); // eslint-disable-line react-hooks/exhaustive-deps
 
     const setSort = useCallback(
-        (newSort: string) =>
-            changeParams({ type: SET_SORT, payload: { sort: newSort } }),
+        (newSort: string, newOrder?: string) =>
```
fzaninotto: I'd name the params `sort` and `order`, it would make the changeParams payload as simple as `{ sort, order }`

PR #4524, opened 2020-03-13
Fix code sample in CreateEdit docs.

### LINKED ISSUE #4522: [docs] Missing segments array in FormWithRedirect code example
While experimenting with v.3 new features I have attempted to just copy code from the docs to play around with but I noticed missing part which took me a few minutes searching in the repo to get things working. 

This is the  page ->[Custom Form Component part](https://marmelab.com/react-admin/CreateEdit.html#custom-form-component)

When you copy and execute an error an error appears because the `segments ` array is missing and I think it would be nice to have it in the snippets itself. Shall I make a pull request with the edit?

WiXSL:
Fixes #4522.


fzaninotto:
Thanks!

PR #4528, opened 2020-03-14
Fix invalid JSON pass to URLS in documentation.

### LINKED ISSUE #4520: [Update documentation] ra-data-simple-rest dialect sends Valid JSON in URL params
You should update the documentation of [`ra-data-simple-rest` dialect](https://github.com/marmelab/react-admin/tree/master/packages/ra-data-simple-rest#rest-dialect) to reflect that the library does indeed send valid JSON in URL params.

Example:
```
GET http://my.api.url/posts?sort=['title','ASC']&range=[0, 24]&filter={title:'bar'}
```

Needs to be:
```
GET http://my.api.url/posts?sort=["title","ASC"]&range=[0, 24]&filter={"title":"bar"}
```

Thanks a lot for the library.

WiXSL:
Fixes #4520

### DIFF
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -133,10 +133,10 @@ Here is how this Data Provider maps react-admin calls to API calls:
 
 | Method name        | API call
 |--------------------|----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks!

PR #4529, opened 2020-03-14
[RFR] Change docs menu Fields item order with Create and Edit item.

WiXSL:
I often found hard to find intuitivelly the `<Field> Components` menu item since is in between the items of the main views.

```
Before:

<List> View
<Show> View
<Field> Components
<Create> and <Edit> Views
<Input> Components
...

After:

<List> View
<Show> View
<Create> and <Edit> Views
<Field> Components
<Input> Components
...


fzaninotto:
That's better indeed. Thanks!

PR #4536, opened 2020-03-15
Convert ArrayField component to function component.

WiXSL:


packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,33 +95,23 @@ const initialState = {
...
+    addLabel,
+    basePath,
+    children,
+    record,
+    sortable,
+    source,
+    fieldKey,
+    ...rest
+}) => {
+    const getDataAndIds = (
```
ThieryMichel: You should move this function outside of the component

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -134,39 +131,37 @@ export class ArrayField extends Component<
...
-        });
-    }
-}
-
-const EnhancedArrayField = pure<FieldProps>(ArrayField);
+    useEffect(() => {
+        const { ids, data } = getDataAndIds(record, source, fieldKey);
+        setIds(ids);
+        setData(data);
+    }, [record]);
```
ThieryMichel: add resource and fieldKey in the hook dependency

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -134,39 +131,37 @@ export class ArrayField extends Component<
...
 
+    // @ts-ignore
```
ThieryMichel: Strange, a ts-ignore shouldn't be needed here

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -134,39 +131,37 @@ export class ArrayField extends Component<
...
+    return cloneElement(Children.only(children), {
+        ids,
+        data,
+        loading: false,
+        basePath,
+        currentSort: {},
+        ...rest,
+    });
+};
+// @ts-ignore
```
ThieryMichel: same no ts-ignore needed

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -134,39 +131,37 @@ export class ArrayField extends Component<
...
+const EnhancedArrayField = memo<FieldProps>(ArrayField);
+// @ts-ignore
```
ThieryMichel: please remove

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -134,39 +131,37 @@ export class ArrayField extends Component<
...
 EnhancedArrayField.defaultProps = {
     addLabel: true,
 };
-
+// @ts-ignore
```
ThieryMichel: shouldn't be necessary

WiXSL:
I think i should add som @ts-ignore there.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: I re-add it because of travis complaint.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: I re-add it because of travis complaint.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: I re-add it because of travis complaint again.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: The same.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: I rollback to use recompose.pure insead of memo. 2 ts-ignore gone.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: I rollback to use recompose.pure insead of memo. 2 ts-ignore gone.

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,78 +117,40 @@ const initialState = {
...
+    basePath,
+    children,
+    record,
+    sortable,
+    source,
+    fieldKey,
+    ...rest
+}) => {
+    const { ids: initialIds, data: initialData } = record
+        ? getDataAndIds(record, source, fieldKey)
```
fzaninotto, line 133: getDataAndIds is computationally intensive. To avoid running it at each render, I'd do the calculation inside a callback passed to useState(), as follows:

```diff
- const { ids: initialIds, data: initialData } = record ? getDataAndIds(record, source, fieldKey)
- const [ids, setIds] = useState(initialIds);
+ const [ids, setIds] = useState(() => {
+   const list = get(record, source);
+   if (!list) return [];
+   return fieldKey ? list.map(item => item[fieldKey]) : list.map(JSON.stringify);
+ });
```

WiXSL: Done!

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,78 +117,43 @@ const initialState = {
...
-                      prev[JSON.stringify(item)] = item;
-                      return prev;
-                  }, {}),
-                  ids: list.map(JSON.stringify),
-              };
-    }
+            ? list.map(item => item[fieldKey])
+            : list.map(JSON.stringify);
+    });
+    const [data, setData] = useState(ids);
```
fzaninotto, line 139: This seems weird to me. According to the previous code, this should be something similar to the other useState code

WiXSL: Ok.

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,78 +117,56 @@ const initialState = {
...
-            source,
-            fieldKey,
-            ...rest
-        } = this.props;
-        const { ids, data } = this.state;
+    useEffect(() => {
+        const { ids, data } = getDataAndIds(record, source, fieldKey);
+        setIds(ids);
+        setData(data);
+    }, [record, source, fieldKey]);
```
fzaninotto, line 158: This will be slower than the initial version, as the effect will run at least once - so the getDataAndId logic will be executed at least twice. 

I see two options: 

1. add a ref for the record, to execute the effect only if the ref has changed
2. Remove the initial value in the `useState()` calls, so that getDataAndId gets only executed once - in the effect

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,78 +117,56 @@ const initialState = {
...
+export const ArrayField: FunctionComponent<
+    FieldProps & InjectedFieldProps & State
+> = ({
+    addLabel,
+    basePath,
+    children,
+    record,
+    sortable,
+    source,
+    fieldKey,
```
fzaninotto, line 129: it seems that `fieldKey` is present in `FieldProps` while it's specific to Arrayfield. We should remove this from `FieldProps` and add an `ArrayFieldProps` here with just that prop.

WiXSL: Done.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: Done!

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,86 +121,49 @@ const initialState = {
...
-            : {
-                  data: list.reduce((prev, item) => {
-                      prev[JSON.stringify(item)] = item;
-                      return prev;
-                  }, {}),
-                  ids: list.map(JSON.stringify),
-              };
-    }
+export const ArrayField: FunctionComponent<
+    ArrayFieldProps & FieldProps & InjectedFieldProps & State
```
fzaninotto, line 125: Why is `State` here? `ids` and `data` aren't injected.

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -88,86 +121,49 @@ const initialState = {
...
+        data,
+        loading: false,
+        basePath,
+        currentSort: {},
+        ...rest,
+    });
+};
 
-const EnhancedArrayField = pure<FieldProps>(ArrayField);
+const EnhancedArrayField = pure<ArrayFieldProps & FieldProps>(ArrayField);
```
fzaninotto, line 156: While you are updating this component to modern standards, would you mind using `React.memo` rather than `pure`?

WiXSL: I tried, and I get ts errors, I revert it back to avoid to many problematic changes. But I'll try again.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: Done.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: Done, but I don't know if is the best way to do it. Correct me if I'm wrong.
`memo<ArrayFieldProps & FieldProps>(ArrayFIeld);` throw me every posible TS error on `EnhancedArrayField` component.

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -153,15 +153,14 @@ export const ArrayField: FunctionComponent<
     });
 };
 
-const EnhancedArrayField = pure<ArrayFieldProps & FieldProps>(ArrayField);
+const EnhancedArrayField: FunctionComponent<ArrayFieldProps & FieldProps> = memo(ArrayField);
```
fzaninotto, line 156: We don't even need the `EnhancedArrayField`: place the `memo` at the `ArrayField` definition.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: Done. I hope.

packages/ra-ui-materialui/src/field/ArrayField.tsx
WiXSL: Done. I hope.

fzaninotto:
Awesome, thanks a lot for your patience!

sebitoelcheater:
I get an error using `SingleFieldList` inside `ArrayField`. This PR seems to be the responsible.
```
<ArrayField source="statements">
  <SingleFieldList>
    <ChipField source="name"/>
  </SingleFieldList>
</ArrayField>
```
throws
```
TypeError: Cannot read property 'map' of undefined
```

djhi:
@sebitoelcheater Can you please open a proper issue?

Issue #3245, opened 2019-05-18
Button inside Toolbar component complains about several props


manattersa:
**What you were expecting:**
Using a Button component inside a Toolbar to customize a SimpleForm behaviour shouldn't complain about props.

**What happened instead:**
React complains with several "React does not recognize the `<xxx>` prop on a DOM element".

**Steps to reproduce:**
I followed instructions about how to customize the Toolbar for a SimpleForm (https://marmelab.com/react-admin/CreateEdit.html#toolbar) but instead of a SaveButton I use a simple Button in it.

**Related code:**
`
const MyToolbar = props => (
    <Toolbar {...props}>
        <Button label="Test"/>
    </Toolbar>
);
`

**Other information:**
If instead I sanitize the props it works without warnings.
```
const MyButton = ({
    basePath,
    className,
    classes,
    saving,
    invalid,
    variant,
    translate,
    handleSubmit,
    handleSubmitWithRedirect,
    submitOnEnter,
    record,
    redirect,
    resource,
    locale,
    showNotification,
    undoable,
    pristine,
    ...rest
}) => <Button {...rest}/>;

const MyToolbar = props => (
    <Toolbar {...props}>
        <MyButton label="Test"/>
    </Toolbar>
);
```
**Environment**

* React-admin version: 2.9.1
* React version: 16.8.6
* Browser: Firefox, Chrome.

Kmaschta:
Instead of changing the `<Button>` component, we have to change the documentation so that it doesn't cause warnings or errors.

PR are welcome to change the doc!

fzaninotto:
Fixed by #4544

PR #4544, opened 2020-03-17
Fix Edit Toolbar docs to explain how to use a CreateButton

agent3bood:
https://github.com/marmelab/react-admin/issues/3245

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -883,6 +883,8 @@ Here are the props received by the `Toolbar` component when passed as the `toolb
 
 **Tip**: To alter the form values before submitting, you should use the `handleSubmit` prop. See [Altering the Form Values before Submitting](#altering-the-form-values-before-submitting) for more information and examples.
 
+**Tip** `<Toolbar>` will pass many props to its children, if you use `<CreateButton>` or other buttons you will get `React does not recognize the <xxx> prop on a DOM element` errors, insteade you can build you own button.
+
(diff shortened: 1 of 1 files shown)


docs/CreateEdit.md
```diff
@@ -883,6 +883,8 @@ Here are the props received by the `Toolbar` component when passed as the `toolb
 
 **Tip**: To alter the form values before submitting, you should use the `handleSubmit` prop. See [Altering the Form Values before Submitting](#altering-the-form-values-before-submitting) for more information and examples.
 
+**Tip** `<Toolbar>` will pass many props to its children, if you use `<CreateButton>` or other buttons you will get `React does not recognize the <xxx> prop on a DOM element` errors, insteade you can build you own button.
```
fzaninotto: There are several things wrong in that change

> `<Toolbar>` will pass many props to its children

Is vague, uses future tense while we try to use present tense everywhere

> if you use `<CreateButton>` or other buttons

Not true: EditButton or DeleteButton don't trigger the error

> insteade

isn't English

> you can build you own button

This isn't clear at all. And besides, the recommendation is rather to wrap an existing button inside another component that will just use the necessary props and drop the rest

agent3bood: I have updated my PR, now it is specific and has example

fzaninotto:
Thanks.

WiXSL:
Couldn't this be improve in `CreateButton.tsx` by sanitising this props?

PR #4548, opened 2020-03-17
Improve FormTab and Tab components definitions.

WiXSL:
There is no need of using functions to resolve the return component in these cases.


fzaninotto:
I disagree with this change. I think the functions make the code clearer. They play the role of comments.

In general, we try to make the code readable by an average programmer, and very long ternaries like the ones in your PR put a real burden on the developer's mind. 

So I'm closing this PR, but thanks for the suggestion anyway!

WiXSL:
I'll take that.

PR #4573, opened 2020-03-23
Fix total in optimistic

### LINKED ISSUE #4337: Editing a record throws Total from response is not a number error
**What you were expecting:**

Expect it won't log any error in console when changing filter on a `ReferenceManyField`
**What happened instead:**

`Total from response is not a number` error is shown in console when filter prop is changed on `ReferenceManyField` 
**Steps to reproduce:**

1. Go to codesandbox https://codesandbox.io/s/elastic-surf-urpjz
2. Edit a post -> go to comments tab
3. I added a body filter to filter comments in the comments tab in `PostEdit.js`. Change the body filter by typing some characters
4. `Total from response is not a number. Please check your dataProvider or the API` error is thrown in console

Screenshot 2020-01-22 at 1 25 38 PM

**Related code:**

https://codesandbox.io/s/elastic-surf-urpjz

**Other information:**

This is the line which throws the error https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/dataProvider/useQueryWithStore.ts#L121. 

I wonder if the check here https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/dataProvider/useQueryWithStore.ts#L119-L120 is suitable for `useGetManyReference` / `ReferenceManyField`.

When the filter prop is changed, `useGetManyReference` will use a new [`relatedTo`](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/dataProvider/useGetManyReference.ts#L80-L81)  to select ids & total from store, but we are comparing them against the old `relatedTo` (generated using old filter prop)'s ids & total.

**Environment**

* React-admin version: Latest
* React version:
* Browser:
```
Total from response is not a number. Please check your dataProvider or the API. 
    in Anonymous (at PostEdit.js:151)
    in PostEdit (created by WithPermissions)
    in Resource (at src/​index.js:40)
```

JulienMattiussi:
Closes #4337 

## Issue

In optimitics mode, there is an error in console after any changes (update/delete/create) that redirect to a list.

`Total from response is not a number`

## Solution

Add check of undefined total in `useQueryWithStore`

### DIFF
--- a/examples/simple/src/dataProvider.js
+++ b/examples/simple/src/dataProvider.js
@@ -34,3 +34,4 @@ const delayedDataProvider = new Proxy(sometimesFailsDataProvider, {
 });
 
 export default cacheDataProviderProxy(delayedDataProvider);
+// export default delayedDataProvider;
--- a/examples/simple/src/dataProvider.js
+++ b/examples/simple/src/dataProvider.js
@@ -34,4 +34,3 @@ const delayedDataProvider = new Proxy(sometimesFailsDataProvider, {
 });
 
 export default cacheDataProviderProxy(delayedDataProvider);
-// export default delayedDataProvider;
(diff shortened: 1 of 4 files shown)


fzaninotto:
What's left for RFR?

JulienMattiussi:
Want to finalize an e2e test to check it is ok

PR #4574, opened 2020-03-24
[RFR] Improve buttons props sanitation

WiXSL:
https://github.com/marmelab/react-admin/pull/4544#issuecomment-603241431

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
@@ -39,8 +39,8 @@ const BulkDeleteWithConfirmButton: FC<
     const {
         basePath,
         classes: classesOverride,
-        confirmTitle = '',
-        confirmContent,
+        confirmTitle = 'ra.message.bulk_delete_title',
+        confirmContent = 'ra.message.bulk_delete_content',
         icon,
         label,
         onClick,
@@ -158,8 +158,6 @@ BulkDeleteWithConfirmButton.propTypes = {
 };
 
 BulkDeleteWithConfirmButton.defaultProps = {
-    confirmTitle: 'ra.message.bulk_delete_title',
-    confirmContent: 'ra.message.bulk_delete_content',
     label: 'ra.action.delete',
     icon: <ActionDelete />,
 };
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
@@ -39,7 +39,7 @@ const BulkDeleteWithConfirmButton: FC<
     const {
         basePath,
         classes: classesOverride,
-        confirmTitle,
+        confirmTitle = '',
         confirmContent,
         icon,
         label,
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
@@ -39,8 +39,8 @@ const BulkDeleteWithConfirmButton: FC<
     const {
         basePath,
         classes: classesOverride,
-        confirmTitle = 'ra.message.bulk_delete_title',
-        confirmContent = 'ra.message.bulk_delete_content',
+        confirmTitle,
+        confirmContent,
         icon,
         label,
         onClick,
@@ -158,6 +158,8 @@ BulkDeleteWithConfirmButton.propTypes = {
 };
 
 BulkDeleteWithConfirmButton.defaultProps = {
+    confirmTitle: 'ra.message.bulk_delete_title',
+    confirmContent: 'ra.message.bulk_delete_content',
     label: 'ra.action.delete',
     icon: <ActionDelete />,
 };
--- a/packages/ra-ui-materialui/src/button/BulkExportButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkExportButton.tsx
@@ -85,7 +85,7 @@ interface Props {
     label?: string;
     onClick?: (e: Event) => void;
     selectedIds: Identifier[];
-    resource?: string;
+    resource: string;
 }
 
 export type BulkExportButtonProps = Props & ButtonProps;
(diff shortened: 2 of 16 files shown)


packages/ra-ui-materialui/src/button/BulkExportButton.tsx
```diff
@@ -85,7 +85,7 @@ interface Props {
     label?: string;
     onClick?: (e: Event) => void;
     selectedIds: Identifier[];
-    resource?: string;
+    resource: string;
```
fzaninotto: why? The `resource` is usually injected by the parent component, and we don't document the need to add it manually. This is a breaking change.

packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
```diff
@@ -39,8 +39,8 @@ const BulkDeleteWithConfirmButton: FC<
     const {
         basePath,
         classes: classesOverride,
-        confirmTitle,
-        confirmContent,
+        confirmTitle = 'ra.message.bulk_delete_title',
```
fzaninotto: Why that change? Also, why do it for confirmTitle and not for label?

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -48,29 +48,24 @@ const DeleteWithConfirmButton: FC<DeleteWithConfirmButtonProps> = props => {
...
-                setOpen(false);
-            },
-            undoable: false,
-        }
-    );
+    const [deleteOne, { loading }] = useDelete(resource, record.id, record, {
+        action: CRUD_DELETE,
+        onSuccess: () => {
+            notify('ra.notification.deleted', 'info', { smart_count: 1 });
+            redirect(redirectTo, basePath); //FIXME: redirections/RedirectionSideEffect !== useRedirect/RedirectionSideEffect
```
fzaninotto: FIXME

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -36,8 +36,8 @@ const DeleteWithConfirmButton: FC<DeleteWithConfirmButtonProps> = props => {
         icon = defaultIcon,
         label = 'ra.action.delete',
         onClick,
-        record,
-        resource,
+        record = { id: '' },
```
fzaninotto: Why? The record is sometimes undefined, in which case we should not display the button (or allow it to be clicked). It doesn't make sense to delete a record with id ''

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.tsx
```diff
@@ -23,8 +23,8 @@ const DeleteWithUndoButton: FC<DeleteWithUndoButtonProps> = props => {
         className,
         icon = defaultIcon,
         onClick,
-        resource,
-        record,
+        resource = '',
+        record = { id: '' },
```
fzaninotto: same, please do not use a default value

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -101,7 +101,7 @@ interface Props {
     label?: string;
     maxResults?: number;
     onClick?: (e: Event) => void;
-    resource?: string;
+    resource: string;
```
fzaninotto: No, it's not required at compile time, because it's injected at runtime

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -43,10 +45,10 @@ const SaveButton: FC<SaveButtonProps> = props => {
 
     const handleClick = event => {
         if (typeof onSave === 'function') {
-            setOnSave(onSave);
+            setOnSave && setOnSave(onSave);
```
fzaninotto: If the SaveButton is used outside of a FormContext, we want the save to fail.

packages/ra-ui-materialui/src/layout/Confirm.tsx
```diff
@@ -57,12 +57,12 @@ const useStyles = makeStyles(
...
     const {
-        isOpen,
+        isOpen = false,
         loading,
         title,
         content,
-        confirm,
-        cancel,
+        confirm = 'ra.action.confirm',
+        cancel = 'ra.action.cancel',
```
fzaninotto: why pass a default value for this one and not for other props?

packages/ra-ui-materialui/src/button/BulkExportButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/button/ExportButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/button/SaveButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/layout/Confirm.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.tsx
WiXSL: I was using type-script with strictMode = true. I'll change it.

WiXSL:
@fzaninotto, now it's for review.

fzaninotto:
Thanks!

PR #4590, opened 2020-03-28
Fix useState return type in demo example Dashboard

developerium:
Hi Guys
I was looking to contribute to converting more of the demo app to typescript and noticed there is a better way to enforce state type.

Please let me know how it looks.


fzaninotto:
Thanks!

PR #4604, opened 2020-03-30
Change Docs: menu item to use an instance of icon

janakact:
The existing way throws and error `Error: Expected ref to be a function, a string, an object returned by React.createRef(), or nul`. 
This fix works.


fzaninotto:
Thanks!

PR #4621, opened 2020-04-01
[Docs]: Improve buttons sanitation explanation.

WiXSL:
Because of #4574

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -1057,10 +1057,15 @@ Here are the props received by the `Toolbar` component when passed as the `toolb
 
 **Tip**: To alter the form values before submitting, you should use the `handleSubmit` prop. See [Altering the Form Values before Submitting](#altering-the-form-values-before-submitting) for more information and examples.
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
This conflicts with #4544 (already in master). Let me merge `master` into `next` and you'll change what needs to be changed.

fzaninotto:
Done, please rebase

WiXSL:
Done.

fzaninotto:
I don't understand, the previous code snippet already took care of sanitizing the extra props.

Kmaschta:
Since we fixed the CreateButton props sanitation, and this Create and Edit views documentation already have a lot of tips and examples, I'm in favor of removing this tip.
It is not the purpose of React Admin to explain how Material UI components work ?

WiXSL:
Actually the idea of #4574 was to remove #4544, Buy I didnt want to remove somebody elses work, so I improve it to a common use case  that has the same problem. A custom button could be created out of any librarte I justo use material-ui as an example.

WiXSL:
They are not white spaces, they are zero width unicode characters (in this case \u200B).
I asume someone add them accidentally.
But I'll reverted back if you want, no problems.

Please confirm about this.

fzaninotto:
You're right! How did you detect those? Some of us may have a bad setting in their IDE.

WiXSL:
I'm using a plugin for JetBrains IDEs.

https://plugins.jetbrains.com/plugin/7448-zero-width-characters-locator

fzaninotto:
Nice. I found this equivalent plugin for VSCode: https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins

WiXSL:
Oh, I'll take note of that. Thanks.

PR #4636, opened 2020-04-03
Fix error when passing custom link style to SingleFieldList

Slavvkko:
Using
```
const useSingleFieldListStyles = makeStyles({
  root: {
    display: "block",
    flexWrap: "unset",
    marginTop: 0,
    marginBottom: 0
  },
  link: {
    display: "block"
  }
});

const Demo = props => {
  const singleFieldListClasses = useSingleFieldListStyles();
  return (
    <SingleFieldList classes={singleFieldListClasses}>
      <TextField source="name"/>
    </SingleFieldList>
  );
};
```

Got an error in console:
`Material-UI: the key link provided to the classes prop is not implemented in undefined.`

Because styles in `SingleFieldList` doesn't have `link` class


fzaninotto:
Thanks!

PR #4638, opened 2020-04-04
Make useInput hook return child's input.value when parent's enhanced …

### LINKED ISSUE #4203: Initial Values are not showing up in ReferenceArrayInput on second render.
**What you were expecting:**
Expect the initialValue to be displayed on every render

**What happened instead:**
initialValue is not showing on second render onwards.

**Steps to reproduce:**
https://codesandbox.io/s/cool-dust-72uir

- Go to the comments menu.
- Click on the create button.
- Observe that the posts are auto-populated. Do not save or enter any data in the form.
- Go to comments from the menu again.
- Click on the create button.
- Observe that the post are not auto-populated.
- Click on the post's input. Do not select any value.
- The initialValue starts to appear.

**Environment**

* React-admin version: 3.1.1
* Last version that did not exhibit the issue (if applicable): 2.*
* React version: 16.8.6
* Browser: Chrome 79.0.3945.88
* Stack trace (in case of a JS error): No Error

prasanna214:
…input.value is undefined.

Closes #4203 

Issue was caused by useField hook of react-final-form returning defaultValue of field as undefined (when it is not yet available in  state) while enhancing the parent input.

useInput hook of react-admin is considering same enhanced input with value undefined while rendering the children.

### DIFF
--- a/packages/ra-core/src/form/useInput.spec.tsx
+++ b/packages/ra-core/src/form/useInput.spec.tsx
@@ -64,6 +64,43 @@ describe('useInput', () => {
         expect(inputProps.meta).toBeDefined();
     });
 
+    it('allows to override the value of enhanced input', () => {
+        let inputProps;
+        const enhancedInput: any = {
+            id: 'parent',
+            name: 'title',
+            isRequired: true,
+        };
+        const meta = {
+            touched: false,
+        };
+        render(
+            <Form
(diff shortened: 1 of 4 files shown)


fzaninotto:
I came wit han entirely different (and more lightweight) solution to fix #4203. See #4641, which supersedes this PR.

PR #4645, opened 2020-04-06
Avoid Multiple Notifications on Expired Session

djhi:


packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -40,21 +42,32 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
...
-                    notify('ra.notification.logged_out', 'warning');
-                    return true;
-                }),
+        (error?: any) => {
+            // Sometimes, a component might trigger multiple simultaneous
+            // dataProvider calls which all fail and call this function.
+            // To avoid having multiple notifications, we first verify if
+            // a checkError promise is already ongoing
+            if (!authCheckPromise) {
+                authCheckPromise = authProvider
```
fzaninotto, line 51: why do you store the entire Promise instead of just a boolean?

djhi: Because this way, anyone interested by its result will receive it ?

fzaninotto: Hm, you're exposing a private API publicly without a real use case... I'm not comfortable with it.

djhi: We were returning it before already. Nothing has changed. We just don't allow multiple simultaneous calls, that's all

fzaninotto: Gotcha, thanks for the explanation.

PR #4657, opened 2020-04-07
Added missing AppBar translation

WiXSL:


packages/ra-ui-materialui/src/layout/AppBar.js
```diff
@@ -110,7 +110,7 @@ const AppBar = props => {
                                 ? 'ra.action.close_menu'
                                 : 'ra.action.open_menu',
                             {
-                                _: 'Open/Close menu',
```
fzaninotto, line 113: No, the idea of the `_` parameter is to display a string if the actual translation is missing. This is done as a backwards compatibility measure since the text wasn't translated before.

WiXSL: Make sense. As you wish.

PR #4659, opened 2020-04-07
[Doc] Add a link to Clipboard List Field third-party package

OoDeLally:
Here is a copy-to-clipboard field.
https://github.com/OoDeLally/react-admin-clipboard-list-field
Being new to both React Admin and MaterialUI, I would be grateful if anyone can review my component and suggest improvements.

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -152,7 +152,7 @@ const TagsField = ({ record }) => (
         ))}
     </ul>
 )
-TagsField.defaultProps = { 
+TagsField.defaultProps = {
     addLabel: true
 };
 ```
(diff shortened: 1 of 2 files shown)


docs/Fields.md
```diff
@@ -894,6 +894,14 @@ PriceField.defaultProps = {
 ```
 {% endraw %}
 
+
+## Third-Party Components
```
fzaninotto, line 898: Please add it to the Ecosystem documentation instead

OoDeLally: Ok. I moved all third-part links to Ecosystem, to be consistent.
This could be part of this PR or another one, your choice.

fzaninotto: So your last changes makes me realize that what I asked you is wrong, sorry. The third party field should be added to the Fields doc, and the Ecosystem should be updated to add a link to the third party fields.

OoDeLally:
Here you go.

fzaninotto:
Thanks!

PR #4661, opened 2020-04-08
Fix EmptyText Inconsistencies

### LINKED ISSUE #4648: TextField with null value outputs "null" (as string)
**What you were expecting:**
If a TextField value is null, it should not render a value

**What happened instead:**
The null value gets converted into a string and is rendered as "null"

**Steps to reproduce:**
```
const record = { title: null };
const { queryByText } = render(
  <TextField record={record} source="title" />
 );
```

I think these changes are related to the addition of the emptyText property.   It looks like emptyText used to have a default value of '', but now it does not.

https://github.com/marmelab/react-admin/blob/054fd1c6d80a38b613e8d93c8e0aedecc98cad05/packages/ra-ui-materialui/src/field/TextField.tsx#L14

**Environment**

* React-admin version: 3.4.0
* React version: latest
* Browser: chrome latest

jdemangeon:
Closes #4648

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.spec.js
+++ b/packages/ra-ui-materialui/src/field/BooleanField.spec.js
@@ -59,18 +59,21 @@ describe('<BooleanField />', () => {
         expect(queryByTitle('ra.boolean.false')).toBeNull();
     });
 
-    it('should display the emptyText when is present and the value is null', () => {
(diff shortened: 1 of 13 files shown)


dgarvin57:
Thanks for fixing this. I can confirm it's working properly in 3.4.2.

PR #4677, opened 2020-04-10
Don't force the sidebar to be opened by default

Luwangel:
Currently, the sidebar is opened on desktop and closed on mobile. But this feature uses the material-ui [useMediaQuery](https://material-ui.com/components/use-media-query/) hook, which triggers two renders (the first one is always `false` to work on SSR). To fix it, react-admin forces the sidebar to be opened after the second render.

But as a consequence, it's impossible to define another behaviour by **changing the initial state of the application**.

The solution imagined is to only rely on the initial state of the application (and not force a reopening afterward). And to open the sidebar by default on desktop, we will use directly the `window.matchMedia` method.

## Todo

- [x] Remove the hack that forces

## Screenshots

**Opened on desktop**

Sélection_002

**Closed on mobile**

Sélection_001

### DIFF
--- a/packages/ra-core/src/reducer/admin/ui.ts
+++ b/packages/ra-core/src/reducer/admin/ui.ts
@@ -28,7 +28,7 @@ export interface UIState {
 
 // Match the medium breakpoint defined in the material-ui theme
 // See https://material-ui.com/customization/breakpoints/#breakpoints
-const isDesktop: Boolean = () =>
+const isDesktop = (): boolean =>
     // (min-width: 960px) => theme.breakpoints.up('md')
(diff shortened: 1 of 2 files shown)


djhi:
Oups, didn't notice the breaking change. Why is it considered a BC ?

Luwangel:
It will change the initial state of redux. It could have an impact I guess.

quyenlt1812:
Hi @Luwangel ! I'm using RA in SSR. So, "window" is not available during server-side rendering. Could you please add `typeof window !== 'undefined'` in the isDesktop function?

PR #4694, opened 2020-04-16
fix: Export button is not getting updated when  is updated - 4217

### LINKED ISSUE #4217: Export button is not getting updated when `total` is updated
**What you were expecting:**
Expecting the `Export` button to get re-rendered when the total is changed

**What happened instead:**
The `Export` button was not changed

**Steps to reproduce:**
visit a resource's list page directly using the link (or by refreshing it)

**Related code:**
At [this line](https://github.com/marmelab/react-admin/blob/ab614debd8a49868256c880e96f9350171f90a9b/packages/ra-ui-materialui/src/list/ListActions.js#L78) of code the `total` property is not included in the `onlyUpdateForKeys` list which causes the export button not to be refreshed when the `{data, total}` pair is updated. 
Meanwhile at [this line of code](https://github.com/marmelab/react-admin/blob/ab614debd8a49868256c880e96f9350171f90a9b/packages/ra-ui-materialui/src/list/ListActions.js#L47) the export button is `total` dependent.

Also, I've checked the 3.x version, there the total is included as written at [this line of code](https://github.com/marmelab/react-admin/blob/7cdac4db262d7e64f9529f25f9e2c01fc371d983/packages/ra-ui-materialui/src/list/ListActions.js#L48)

**Possible Solution**
Just add `total` to the list of `onlyUpdateForKeys`.

**Other information:**
None

**Environment**

* React-admin version: 2.9.5
* Last version that did not exhibit the issue (if applicable): idk
* React version: 16.9.0 
* Browser: google chrome

manishsundriyal:
Closes #4217

The `Export` button was not updating (enable/disable) when the total no. of items is changed.

`Fix:` the `total` key was missing in the `onlyUpdateForKeys` call.

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListActions.js
+++ b/packages/ra-ui-materialui/src/list/ListActions.js
@@ -81,4 +81,5 @@ export default onlyUpdateForKeys([
     'displayedFilters',
     'filterValues',
     'selectedIds',
+    'total',
 ])(Actions);


fzaninotto:
Thanks!

PR #4708, opened 2020-04-21
[RFR]: Allow SimpleList linkType prop to be resolved as a function.

WiXSL:
The main use case is to allow to decide wich action to take on each record by evaluating the record itself like the other `SimpleList` main props.
You can do that with the `Datagrid` for large screens easilly but for small screens the `SimpleList` doesn't have a simple way to do this.

```jsx
// in src/posts.js
import React from 'react';
import { useMediaQuery } from '@material-ui/core';
import { List, SimpleList, Datagrid, TextField, ReferenceField, EditButton } from 'react-admin';

export const PostList = props => {
    const isSmall = useMediaQuery(theme => theme.breakpoints.down('sm'));
    return (
        <List {...props}>
            {isSmall ? (
                <SimpleList
                    primaryText={record => record.title}
                    secondaryText={record => `${record.views} views`}
                    tertiaryText={record => new Date(record.published_at).toLocaleDateString()}
                    linkType={record => record.canEdit ? "edit" : "show"} 
                />
            ) : (
                <Datagrid>
                    ...
                </Datagrid>
            )}
        </List>
    );
}

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList.js
+++ b/packages/ra-ui-materialui/src/list/SimpleList.js
@@ -35,13 +35,18 @@ const LinkOrNot = ({
     basePath,
     id,
     children,
+    record,
 }) => {
     const classes = useLinkOrNotStyles({ classes: classesOverride });
-    return linkType === 'edit' || linkType === true ? (
+    const link = typeof linkType === 'function'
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

PR #4719, opened 2020-04-24
Fix Filter align with SearchFilter

### LINKED ISSUE #4710: Filters require to use SearchInput as first Input to render properly
**What you were expecting:**
Adding Inputs to `<Filter>` as suggested in the documentation (https://marmelab.com/react-admin/List.html#filtering-the-list) should suffice to get a nice layout

**What happened instead:**

The online demo uses `SearchInput` as the fist input. If we remove the SearchInput (in the screenshot simply removed from the DOM), add some Filters and reduce page width the layout is broken (as other inputs do not have `margin-top:32px` they have no visual padding to the title-bar)

https://github.com/marmelab/react-admin/blob/52a4ca810eebb1eea25e900392432b14024f8097/packages/ra-ui-materialui/src/input/SearchInput.tsx#L13

Link to Demo: https://marmelab.com/react-admin-demo/#/commands?displayedFilters=%7B%22customer_id%22%3Atrue%2C%22returned%22%3Atrue%2C%22date_lte%22%3Atrue%7D&filter=%7B%22status%22%3A%22ordered%22%7D&order=DESC&page=1&perPage=25&sort=date

grafik

**Environment**

* React-admin version: 3.4.2
* Last version that did not exhibit the issue (if applicable): 3.3.
* Browser: Chrome

JulienMattiussi:
Closes #4710 

## Issue

Filters are not properly aligned when there is no searchInput and not enouth space for having them all on one line.

SearchInput height is different from other inputs and make UI ugly.

## Solution

- SearchInput uses same height as other inputs for better UI (with label)
- Filters are center aligned
- Don't overload top margin of filter form div

## Screenshots

**With alwaysOn SearchInput**

image

image

**Without SearchInput**

image

image

### DIFF
--- a/packages/ra-ui-materialui/src/input/SearchInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SearchInput.tsx
@@ -1,32 +1,21 @@
 import React, { FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
 import SearchIcon from '@material-ui/icons/Search';
-import { makeStyles, InputAdornment } from '@material-ui/core';
+import { InputAdornment } from '@material-ui/core';
 import { TextFieldProps } from '@material-ui/core/TextField';
 import { useTranslate, InputProps } from 'ra-core';
 
 import TextInput from './TextInput';
-
-const useStyles = makeStyles(
-    {
-        input: {
-            marginTop: 32,
-        },
-    },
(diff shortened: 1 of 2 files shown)


fzaninotto:
Not good: the search filter is smaller on purpose.

djhi:
Superseded by #5271

PR #4720, opened 2020-04-24
Fix default total selector in useQueryWithStore

alanpoulain:
Using `useQueryWithStore` with the default parameters is not working as expected for the `total` state value (it was always `null`).


fzaninotto:
Thanks!

PR #4730, opened 2020-04-27
Set AutocompleteInput value to an empty string when it's selected item is null

WiXSL:
Most AutocompleteInput tests threw console errors regarding initial state.

```
VM74:37 Warning: A component is changing a controlled input of type text to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. 

VM74:37 downshift: A component has changed the controlled prop "inputValue" to be - uncontrolled. This prop should not switch from controlled to uncontrolled (or vice versa). Decide - between using a controlled or uncontrolled Downshift element for the lifetime of the component. - More info: https://github.com/downshift-js/downshift#control-props 
     in Downshift (created by AutocompleteInput)
     in AutocompleteInput (at horascuentascorrientes.js:90)
    in HorasCCCreate (created by WithPermissions)
     in Resource (at App.js:52)
     in App (at src/​index.js:20)

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -12,7 +12,6 @@ import { Form } from 'react-final-form';
 import { TestTranslationProvider } from 'ra-core';
 
 describe('<AutocompleteInput />', () => {
-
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

PR #4753, opened 2020-05-02
Fix Sidebar scrolling with Admin content

### LINKED ISSUE #4684: Side menu scroll with data
**What I Want**
I want to fixed side menu it should not scrolling with my data.

**Description**
I have installed react-admin in my react web App project but i am facing side menu scrolling issue. Side menu is scrolling with data. 

**Steps to reproduce:**
I have following the same example that is providing by react-admin and in this example issue can be reproduce on dashboard.
If we will scroll the dashboard then side menu is also scrolling.
**Link to reproduce**:
 [Link](https://marmelab.com/react-admin-demo/#/)

**Environment**

* React-admin version: ^3.3.1
* React version:^16.13.0
* Browser: Chrome

WiXSL:
Fixes #4684

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Sidebar.js
+++ b/packages/ra-ui-materialui/src/layout/Sidebar.js
@@ -11,7 +11,7 @@ export const CLOSED_DRAWER_WIDTH = 55;
 const useStyles = makeStyles(
     theme => ({
         drawerPaper: {
-            position: 'relative',
+            position: 'sticky',
             height: 'auto',
             overflowX: 'hidden',
             width: props =>


packages/ra-ui-materialui/src/layout/Sidebar.js
```diff
@@ -11,7 +11,7 @@ export const CLOSED_DRAWER_WIDTH = 55;
 const useStyles = makeStyles(
     theme => ({
         drawerPaper: {
-            position: 'relative',
+            position: 'sticky',
```
fzaninotto, line 14: this completely breaks the layout on iE11 (which is otherwise pretty well supported)

adfaris, line 14:
```suggestion
            position: 'fixed',
```
please take a look at this doc for explanation. https://caniuse.com/#feat=css-fixed

WiXSL: But changing 'sticky' for 'fixed' by itself doesn't do anything. I already tried it.

WiXSL: I looked at your PR. and try to add

position: 'fixed',
top: '4em',
zIndex: 100,

but it breaks the layout.

WiXSL: It works in the demo because you have a surrounding div, but not when changing the same in Sidebars's Drawer component

WiXSL: Any other suggestions?

WiXSL:
For anyone interested, I've moved to a different solution in user land:
A `Datagrid` with scrollable content indirectly solving the movement of the menus.

```jsx
import { makeStyles } from "@material-ui/core/styles";
import { Datagrid } from "react-admin";
import TableContainer from "@material-ui/core/TableContainer";

const useStyles = makeStyles(
    theme => ({
        container: {
            maxHeight: "70vh"
        }
    })
);

export const ScrollDatagrid = props => {

    return (
        <TableContainer className={classes.container}>
            <Datagrid
                stickyHeader /* <= Incompatible with IE 11 */
                {...props}
            />
        </TableContainer>
    );
};

PR #4767, opened 2020-05-05
Fix components / hooks error in react-devtools

### LINKED ISSUE #4697: React-admin components / hooks throws error in react developer tools
**What you were expecting:**

Expect choosing react-admin components / components that use `react-admin`'s data fetching hooks in react developer tool will show component state, hooks, props correctly.

**What happened instead:**

Choosing a react-admin component that use dataloader related hooks e.g. `List` will throw error in console
Custom component that uses `useDataLoader` related hooks  e.g. `useCreate`, `useGetOne`, etc. has same issue.
**Steps to reproduce:**

Prerequisite: Install React Developer Tools extension in chrome https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi

1. Run simple demo `make run-simple`
2. Go to `http://localhost:8080/#/posts`
3. Open console -> React Developer Tools, search and choose `List` component
4. Console throws following error
Screenshot 2020-04-17 at 2 38 29 PM
5. Right panel stucks at `Loading`
Screenshot 2020-04-17 at 2 40 13 PM

**Environment**

* React-admin version: master
* React version: react@16.12.0
* Browser: Chrome
```
Uncaught TypeError: uploadCapableDataProvider[name] is not a function
    at Proxy.<anonymous> (dataProvider.js:22)
    at dataProvider.js:26
```

EricJin987:
Fixes #4697

React-devtools uses `obj[Symbol.iterator] === 'function'` to decide whether an object is an iterator. If it is, react-devtools will call that `obj[Symbol.iterator]` function. Since a dataProvider proxy's GET trap always returns a function, the comparison will always be true. The real dataProvider doesn't have the `Symbol.iterator` property, so it throws an error. When `name` is `Symbol.iterator`, returning `undefined` can make react-devtools think the dataProvider is an object and won't call `obj[Symbol.iterator]`.

### DIFF
--- a/packages/ra-core/src/dataProvider/cacheDataProviderProxy.ts
+++ b/packages/ra-core/src/dataProvider/cacheDataProviderProxy.ts
@@ -26,27 +26,24 @@ export default (
         get: (target, name: string) => {
             if (typeof name === 'symbol') {
                 return;
-            } else {
-                return (resource, params) => {
-                    if (
-                        name === 'getList' ||
-                        name === 'getMany' ||
-                        name === 'getOne'
-                    ) {
-                        // @ts-ignore
-                        return dataProvider[name](resource, params).then(
-                            response => {
-                                const validUntil = new Date();
-                                validUntil.setTime(
-                                    validUntil.getTime() + duration
-                                );
-                                response.validUntil = validUntil;
-                                return response;
-                            }
-                        );
-                    }
-                    return dataProvider[name](resource, params);
-                };
             }
+            return (resource, params) => {
+                if (
+                    name === 'getList' ||
+                    name === 'getMany' ||
+                    name === 'getOne'
+                ) {
+                    // @ts-ignore
+                    return dataProvider[name](resource, params).then(
+                        response => {
+                            const validUntil = new Date();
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/dataProvider/cacheDataProviderProxy.ts
```diff
@@ -23,16 +23,30 @@ export default (
...
-                return dataProvider[name](resource, params).then(response => {
-                    const validUntil = new Date();
-                    validUntil.setTime(validUntil.getTime() + duration);
-                    response.validUntil = validUntil;
-                    return response;
-                });
+        get: (target, name: string) => {
+            if (typeof name === 'symbol') {
+                return;
+            } else {
```
fzaninotto, line 29: Could you remove the `else` in all your modifications? It's useless because of the early return, and it harms legibility as it adds one level of indentation. Besides, it would make your changes smaller and easier to review!

EricJin987: Done.

fzaninotto:
Thanks!

PR #4772, opened 2020-05-06
CheckboxGroupInputItem doesn't propagate options to Checkbox

### LINKED ISSUE #4749: CheckboxGroupInputItem not propagating options to Checkbox
**What you were expecting:**

Passing something in the `options` property of `CheckboxGroupInput` would set a corresponding property in the Material UI `Checkbox` descendant, as per docs:

> Lastly, use the options attribute if you want to override any of Material UI’s <Checkbox> attributes:
>
> import { FavoriteBorder, Favorite } from '@material-ui/icons';
> 
> ```
> <CheckboxGroupInput source="category" options={{
>   icon: <FavoriteBorder />,
>   checkedIcon: <Favorite />
> }} />
> ```

**What happened instead:**

Options are not propagated as properties to `Checkbox` elements.

**Steps to reproduce:**

Run the example from docs.

**Other information:**

`CheckboxGroupInput` does [receive and forward](https://github.com/marmelab/react-admin/blob/94b0828558ed783dab000ecbc03f8a3e0447111c/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx#L185) the `options` property to `CheckboxGroupInputItem` but `CheckboxGroupInputItem` doesn't seem to do anything with it. Instead of passing `{...rest.options}` to `Checkbox`, it [passes](https://github.com/marmelab/react-admin/blob/94b0828558ed783dab000ecbc03f8a3e0447111c/packages/ra-ui-materialui/src/input/CheckboxGroupInputItem.tsx#L54) `{...rest}` so the following doesn't work:

```jsx
<CheckboxGroupInput options={{disabled: true}} />
```

But the following does:

```jsx
<CheckboxGroupInput disabled={true} />
```

**Environment**

* React-admin version: 3.4.2
* React version: 16.13.1
* Browser: Chromium 81

WiXSL:
Fixes #4749

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInputItem.tsx
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInputItem.tsx
@@ -21,6 +21,7 @@ const CheckboxGroupInputItem = props => {
         onChange,
         optionText,
         optionValue,
+        options,
         translateChoice,
         value,
         ...rest
@@ -51,6 +52,7 @@ const CheckboxGroupInputItem = props => {
                             : false
                     }
                     value={String(getChoiceValue(choice))}
+                    {...options}
                     {...rest}
                 />
             }


fzaninotto:
Thanks @WiXSL!

Issue #4733, opened 2020-04-28
<Edit /> submit button is offering to submit while the form is still in pristine state and hasn't been edited


kopax:
**What you were expecting:**

I expect the submit button to be disabled if the form is still in `pristine` state

**What happened instead:**

The submit button is always enabled

**Steps to reproduce:**

https://marmelab.com/react-admin-demo/#/customers/129

**Related code:**

This is how a react-final-form submit button can solve the issue

```diff
<Button
-  disabled={submitting}
+  disabled={submitting || pristine}
/>
```

**Environment**

* React-admin version: 3.4.3
* Last version that did not exhibit the issue (if applicable):
* React version: 16.13.1
* Browser: chrone
* Stack trace (in case of a JS error):

fzaninotto:
Nice idea for an enhancement. Feel free to open a PR to implement it.

kopax:
The render method of `<Form />` from React Final Form is actually passing the `submitting` and `pristine` in the same render method so that shouldn't be to hard.

I am currently reading the `Button.tsx` and I found that this may do that trick, but since I am less familiar with the source, I'd like to get your opinion first:

```diff
import React, { FC, ReactElement, SyntheticEvent, ReactNode } from 'react';
import PropTypes from 'prop-types';
import {
    Button as MuiButton,
    Tooltip,
    IconButton,
    useMediaQuery,
    makeStyles,
    PropTypes as MuiPropTypes,
} from '@material-ui/core';
import { ButtonProps as MuiButtonProps } from '@material-ui/core/Button';
import { Theme } from '@material-ui/core';
import classnames from 'classnames';
import { Record, RedirectionSideEffect, useTranslate } from 'ra-core';

/**
 * A generic Button with side icon. Only the icon is displayed on small screens.
 *
 * The component translates the label. Pass the icon as child.
 * The icon displays on the left side of the button by default. Set alignIcon prop to 'right' to inverse.
 *
 * @example
 *
 * <Button label="Edit" color="secondary" onClick={doEdit}>
 *   <ContentCreate />
 * </Button>
 *
 */
const Button: FC<ButtonProps> = props => {
    const {
        alignIcon = 'left',
        children,
        classes: classesOverride,
        className,
        color,
        disabled,
        label,
        size,
        ...rest
    } = props;
    const translate = useTranslate();
    const classes = useStyles(props);
    const isXSmall = useMediaQuery((theme: Theme) =>
        theme.breakpoints.down('xs')
    );
[...]

PR #4773, opened 2020-05-06
SaveButton renders disabled if Form is pristine.

WiXSL:
Implements #4733

### DIFF
--- a/cypress/integration/create.js
+++ b/cypress/integration/create.js
@@ -302,6 +302,14 @@ describe('Create Page', () => {
     });
 
     it('should show rich text input error message when form is submitted', () => {
+        const values = [
+            {
+                type: 'input',
+                name: 'title',
+                value: 'Test title',
+            },
+        ];
+        CreatePage.setValues(values);
         CreatePage.submit();
         cy.get(CreatePage.elements.richTextInputError)
             .should('exist')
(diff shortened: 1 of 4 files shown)


fzaninotto:
This is technically a new feature, please PR against `next` rather than `master`.

kopax:
Excellent. I wonder how I could have missed the SaveButton. Anyway, I would not have added such correct test to the application. Good job thank you really much @WiXSL!

WiXSL:
@kopax The `saving` prop is already handling that issue in the `SaveButton` component.

fzaninotto:
Thanks!

kopax:
Hi guys, just wondering now that 3.6.0 is out, I am testing it in my back office app, and this is what I got:

image

This is what I have in my `node_modules`:

```bash
$ cat node_modules/react-admin/package.json | grep version
  "version": "3.6.0"
$  cat node_modules/ra-core/package.json | grep version
  "version": "3.6.0"
$ cat node_modules/ra-ui-materialui/package.json | grep version
  "version": "3.6.0"
```

Was this really shipped in 3.6.0?  What am I missing?

johnlepikhin:
Is there a way to disable this feature for some form? It breaks Clone behaviour. User just can't save exactly the same object.

PR #4774, opened 2020-05-06
Try to fix cypress tests

fzaninotto:


fzaninotto:
Th problem doesn't come from the Chrome version : Travis dosn't manage to install chrome at all

image

PR #4784, opened 2020-05-07
Fix ReferenceInput when using an object literal as filter prop

fzaninotto:
Supersedes #4688


fancyaction:
I've been finding myself using `// eslint-disable-line react-hooks/exhaustive-deps` more than I like. Not sure where to post about this, but I've been finding lately that RA components are receiving new Material UI classname every render, trigging function component update via that prop change.

Anyway, this looks like a good solution and one we've been waiting for so thanks!

fzaninotto:
@fancyaction The updated class name is a bug, please open a GitHub issue with a reproducible example.

Issue #4731, opened 2020-04-27
`createFactory ` is deprecated Warning in react 16.13.1


macrozone:
**What you were expecting:**

No warnings

**What happened instead:**

```
      Warning: React.createFactory() is deprecated and will be removed in a future ma
jor release. Consider using JSX or use React.createElement() directly instead.

      at printWarning (packages/dataprovider/node_modules/react/cjs/react.development
.js:315:30)
      at warn (node_modules/react/cjs/react.development.js:278:5)
      at createFactoryWithValidation (node_modules/react/cjs/react.development.js:1827:7)
      at node_modules/recompose/shouldUpdate.js:25:44
      at Object.pure [as default] (node_modules/recompose/pure.js:29:97)
      at Object.<anonymous> (node_modules/ra-core/lib/util/FieldTitle.js:19:33)
```

**Steps to reproduce:**

update to react 16.13.1

**Related code:**

its actually an upstream issue from recompose: https://github.com/acdlite/recompose/pull/795

unforunatly, the pull request is not merged and it does not seem to be active.

an alternative would be to use React.memo instead of pure from recompose 

**Environment**

* React-admin version: 3.4.1
* Last version that did not exhibit the issue (if applicable):
* React version: 16.13.1
* Browser: chrome 80
* Stack trace (in case of a JS error): see above

djhi:
Yes, I noticed those annoying warnings as well. Now that most of our components use hooks, we can probably remove recompose.

fzaninotto:
Fixed by #4786

PR #4786, opened 2020-05-07
Change recompose methods usage for React.memo

WiXSL:
#4731


WiXSL:
Note:
The `Enhanced...` version of the returning components pass integration tests. returning `memo(Component)` directly did not.
If this is ok, `ArrayField` component  has to be changed like that as well.

fzaninotto:
I think the problem comes from the fact that SimpleForm inspect its children props looking for an `addLabel` value. React-admin field props set this as `defaultProps`. But if you decorate with `React.memo` *after* setting the defaultProps, this no longer works. 

Therefore, I think you should set the `React.memo` call earlier in all fields:

```diff
-const TextField = props = > ( ...);
+const TextField = memo<FieldProps & TypographyProps>(props = > ( ...));
-const EnhancedTextField = memo<FieldProps & TypographyProps>(TextField);
-EnhancedTextField.defaultProps = {
-    addLabel: true,
-};
+TextField.defaultProps = {
+    addLabel: true,
+};
-export default EnhancedTextField;
+export default TextField;
```

Also, some of the tests import the non-decorated Field (`import { TextField } from 'TextField'` instead of `import TextField from 'TextField'`). This shouls no longer be necessary when using `React.memo`.

packages/ra-ui-materialui/src/button/CloneButton.tsx
```diff
@@ -57,13 +56,14 @@ CloneButton.propTypes = {
...
-const enhance = shouldUpdate(
+const PureCloneButton = memo(
+    CloneButton,
     (props: Props, nextProps: Props) =>
-        (props.record &&
-            nextProps.record &&
-            props.record !== nextProps.record) ||
-        props.basePath !== nextProps.basePath ||
-        (props.record == null && nextProps.record != null)
+        (!props.record ||
```
fzaninotto, line 62: This is hard to read. Besides, the initial condition looks dubious. It would probably be better with:

```
(props.record === nextProps.record) &&
(props.basePath === nextProps.basePath)
```

But this lacks the `label` and the `icon`...

In fact, you can omit the second parameter to `memo` completely. I don't understand this optimization (even though I made it myself). We'll re-add it if a profiling session proves that it's useful.

packages/ra-ui-materialui/src/button/ShowButton.tsx
```diff
@@ -45,13 +44,14 @@ ShowButton.propTypes = {
...
-const enhance = shouldUpdate(
+const PureShowButton = memo(
+    ShowButton,
     (props: Props, nextProps: Props) =>
-        (props.record &&
-            nextProps.record &&
-            props.record.id !== nextProps.record.id) ||
-        props.basePath !== nextProps.basePath ||
-        (props.record == null && nextProps.record != null)
+        (!props.record ||
```
fzaninotto, line 50: This is also hard to read. The idea is that we want to compare record.id instead of the entire record (which may change although the values don't change), and the null check was added in 1ec8f2e895aa9c036b5c1031559223ac5ab59385 to avoid a bug. 

The right way to do it would be

```js
(props.record && nextProps.record ? props.record.id === nextProps.record.id : props.record == nextProps.record) &&
props.basePath === nextProps.basePath
```

fzaninotto:
Thanks!

PR #4788, opened 2020-05-07
Easier customization of sidebar

### LINKED ISSUE #4407: Easier customization of Sidebar
It's currently difficult to really customize the styles of the Sidebar component as it does not pass all classes to the underlying Drawer.

adfaris:
Resolves #4407 

It resolves it by passing a CSS props from the parent to the child component. In this case, the SubMenu parent class defines the Style variable and passes it to Menu and using the passed props; I was able to style the sideBar. 

`Submenu.tsx`

`interface Props {
    dense: boolean;
    handleToggle: () => void;
    icon: ReactElement;
    isOpen: boolean;
    name: string;
    sidebarIsOpen: boolean;
    style: string;
}`
then will pass the style props, wherever styling is needed. 

`<Divider className={style} />`

In 'Menu.tsx'

Just define makeStyles and add use it to style the sidebar. 

`const useStyles = makeStyles({
    sideBarSales: {
        backgroundColor: 'red',
        fontSize: '1.5em',
    }
}
`
`const classes = useStyles();`
and use it to style sidebar menu like: 

`<SubMenu
                handleToggle={() => handleToggle('menuSales')}
                isOpen={state.menuSales}
                sidebarIsOpen={open}
                name="pos.menu.sales"
                icon={<orders.icon />}
                dense={dense}
                style={classes.sideBarSales}>`

Screen Shot 2020-05-08 at 12 28 35 PM

Is that a good direction to pursue? Is there a reference in the code on how this is done in the project?

### DIFF
--- a/examples/demo/src/layout/Menu.tsx
+++ b/examples/demo/src/layout/Menu.tsx
@@ -5,7 +5,6 @@ import LabelIcon from '@material-ui/icons/Label';
 import { useMediaQuery, Theme } from '@material-ui/core';
 import { useTranslate, DashboardMenuItem, MenuItemLink } from 'react-admin';
 import { makeStyles } from '@material-ui/core/styles';
-import { CSSProperties } from '@material-ui/core/styles/withStyles';
 
 import visitors from '../visitors';
 import orders from '../orders';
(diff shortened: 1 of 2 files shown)


fzaninotto:
We don't want to customize the sidebar of the demo. It shows one way to build a custom menu, not a component built in a generic way with an API for customization. 

As for How To questions, please use StackOverflow.

PR #4791, opened 2020-05-08
Fix Fab component console warning in empty list small screen

### LINKED ISSUE #4790: Invalid prop passed to Fab component (create button) on empty List
**What you were expecting:**

No console warnings when viewing an empty List **on a small screen**.

**What happened instead:**

Console warning appears:

```
Invalid prop `variant` of value `contained` supplied to `ForwardRef(Fab)`, expected one of ["round","extended"].

```

Screen Shot 2563-05-08 at 10 19 01

**Steps to reproduce:**

View an empty List on small screen and check console.

- Reproduce on demo app by setting one of the resources as an empty array in `examples/simple/src/data.js`

**Environment**

- React-admin version: latest
- React version: any
- Browser: any

cpv123:
Fixes #4790

### DIFF
--- a/packages/ra-ui-materialui/src/button/CreateButton.tsx
+++ b/packages/ra-ui-materialui/src/button/CreateButton.tsx
@@ -16,6 +16,7 @@ const CreateButton: FC<CreateButtonProps> = props => {
         classes: classesOverride,
         label = 'ra.action.create',
         icon = defaultIcon,
+        variant = 'text',
         ...rest
     } = props;
     const classes = useStyles(props);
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -16,6 +16,7 @@ const CreateButton: FC<CreateButtonProps> = props => {
         classes: classesOverride,
         label = 'ra.action.create',
         icon = defaultIcon,
+        variant = 'text',
```
fzaninotto, line 19: Why do you add a default value? that's a breaking change. Your code should work without a default value.

cpv123: You're right - I've removed default.

fzaninotto:
Thanks!

PR #4799, opened 2020-05-11
[Doc] Fix previous / next pagination

alanpoulain:
I've changed the previous / next pagination example with a better implementation.

In the previous one, the icons were not working and the `labelPosition` prop doesn't exist.

Now, the texts are also translated and RTL is managed.

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -1852,9 +1852,9 @@ Alternately, if you want to replace the default pagination by a "<previous - nex
 
 ```jsx
 import Button from '@material-ui/core/Button';
+import Toolbar from '@material-ui/core/Toolbar';
 import ChevronLeft from '@material-ui/icons/ChevronLeft';
 import ChevronRight from '@material-ui/icons/ChevronRight';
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks, @alanpoulain.

The snippet you modified was just an example of how to write a custom pagination component - not a full-fledged component like the ones published by material-ui. The role of this snippet is to explain the `page`, `setPage`and `total` props. The right way to build an accessible pagination component is, IMHO, outside of the react-admin responsibility. And adding the translation and theme support dilutes the main takeaway.

So I'm reluctant to merge this change, as there are other snippets around the documentation that were written using the same approach (simplify - focus on the current learning). 

But I'm open to discussion.

alanpoulain:
As you prefer. I can only fix the issues if you want.

fzaninotto:
Yes, please only fix the icons to that the example is both simple to understand, and works out of the box when copied.

alanpoulain:
Done.

fzaninotto:
Thanks!

PR #4810, opened 2020-05-13
Fix initialValue is ignored in children of ArrayInput

### LINKED ISSUE #4709: initialValue is not used on child fields of SimpleFormIterator
**What you were expecting:**
I have a property that is an array of objects. When array element exists, but a property is missing on the object, Field's initial value should work

**What happened instead:**
Field is empty (initial value is not assigned)

**Steps to reproduce:**

- Go to https://codesandbox.io/s/adoring-morning-54qgk?file=/src/posts/PostEdit.js:3462-3480
- To to post edit screen and edit record with post ID=1 (miscelaneous tab); in the sandbox, it url is https://54qgk.csb.app/#/posts/1/2
- Notice how initialValue was not applied to ZZZ field. 

**Related code:**
```js
        <ArrayInput source="backlinks">
          <SimpleFormIterator>
            <DateInput source="date" />
            <TextInput source="url" />
            <TextInput source="zzz" initialValue="ZZZZZ" defaultValue="qqqqq" />
          </SimpleFormIterator>
        </ArrayInput>
```
image

Notice, however that initialValue field is honored when creating a new record
image

**Other information:**
During troubleshooting, i noticed that FieldArray children are not honoring default values when placed into FormWithRedirect; however, I confirmed that the same code worked in native react (outside of react-admin; see attached testApp.txt file, which contains App component that can be rendered in react app) 
[testApp.txt](https://github.com/marmelab/react-admin/files/4507306/testApp.txt)

)
```js
import React from 'react';
import { FieldArray } from 'react-final-form-arrays';
import { Field } from 'react-final-form';
import { Edit, FormWithRedirect } from 'react-admin';
import { default as _ } from 'lodash';

const ReportEditForm = () => {
  return (
    <React.Fragment>
      <FieldArray name='a1'>
        {({ fields }) => {
          return fields.map((name) => (
            <div key={name}>
              <Field name={`${name}.a1`} component='input' placeholder='a1' initialValue='def.a1' />
            </div>
          ));
        }}
      </FieldArray>
    </React.Fragment>
  );
};

const ReportEdit = (props) => {
  return (
    <Edit {...props} undoable={false} title='Edit report'>
      <FormWithRedirect render={ReportEditForm} {...props} />
    </Edit>
  );
};

export default ReportEdit;
```

**Environment**

* React-admin version: 3.4.1

fzaninotto:
Closes #4709

### DIFF
--- a/packages/ra-core/src/form/useInitializeFormWithRecord.ts
+++ b/packages/ra-core/src/form/useInitializeFormWithRecord.ts
@@ -18,7 +18,7 @@ const useInitializeFormWithRecord = record => {
         // Using batch ensure we don't get rerenders until all our values are set
         form.batch(() => {
             Object.keys(record).forEach(key => {
-                // We have to check the record key is actually registered as a field
+                // We have to check that the record key is actually registered as a field
                 // as some record keys may not have a matching input
                 if (registeredFields.some(field => field === key)) {
                     if (Array.isArray(record[key])) {
@@ -33,7 +33,7 @@ const useInitializeFormWithRecord = record => {
                                     );
                                 });
                             } else {
-                                // array of scala values
+                                // array of scalar values
                                 form.change(`${key}[${index}]`, value);
                             }
                         });
--- a/packages/ra-core/src/form/useInitializeFormWithRecord.ts
+++ b/packages/ra-core/src/form/useInitializeFormWithRecord.ts
@@ -21,7 +21,26 @@ const useInitializeFormWithRecord = record => {
                 // We have to check the record key is actually registered as a field
                 // as some record keys may not have a matching input
                 if (registeredFields.some(field => field === key)) {
-                    form.change(key, record[key]);
+                    if (Array.isArray(record[key])) {
+                        // array of values
+                        record[key].forEach((value, index) => {
+                            if (value && Object.keys(value).length > 0) {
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/form/useInitializeFormWithRecord.ts
```diff
@@ -18,10 +18,29 @@ const useInitializeFormWithRecord = record => {
...
             Object.keys(record).forEach(key => {
-                // We have to check the record key is actually registered as a field
+                // We have to check that the record key is actually registered as a field
                 // as some record keys may not have a matching input
                 if (registeredFields.some(field => field === key)) {
-                    form.change(key, record[key]);
+                    if (Array.isArray(record[key])) {
+                        // array of values
+                        record[key].forEach((value, index) => {
+                            if (value && Object.keys(value).length > 0) {
```
NikitaVlaznev, line 27: This breaks up arrays with non "Dictionary" objects, like string for example

```
Object.keys("cats")
[ '0', '1', '2', '3' ]
```

NikitaVlaznev:
Causes a bug #4823.

alanpoulain:
@NikitaVlaznev see https://github.com/marmelab/react-admin/pull/4819

PR #4819, opened 2020-05-16
Fix useInitializeFormWithRecord with an array of objects

alanpoulain:
Because, sadly, in JS:
```js
const value = 'string';
console.log(Object.keys(value).length);
// --> 6
```


fzaninotto:
Thanks!

PR #4828, opened 2020-05-18
Avoid treating null-values as objects with keys

stensrud:
`typeof null === "object"`


djhi:
Thanks for contributing :) Can you please explain the issue solved by this PR?

stensrud:
My GraphQL schema has fields with `[String]` type, the string being nullable. When a fetched object has a `null`-value, ex `[null]`, the response parser will crash when trying to run `Object.keys(null).reduce(...)`.

djhi:
Thanks!

PR #4829, opened 2020-05-19
Fix refresh button breaks useGetList result with application cache

### LINKED ISSUE #4817: No data from useGetList after refresh
**What you were expecting:**

useGetList gives data even after pressing refresh button at the header toolbar

**What happened instead:**

Data is empty after refresh

**Steps to reproduce:**

Reproduces at  the simple example `examples/simple/src/customRouetLayout.js`.

1. `make run-simple`
2. Go to http://localhost:8080/#/custom2
3. "Found 13 posts!" displayed.
4. Press Refresh icon
5. "Found posts!" displayed. (no posts listed.)

**Other information:**

Passing version value of `useVersion()` hook to options parameter of `useGetList` seems to be a workaround.
```js
--- a/examples/simple/src/customRouteLayout.js
+++ b/examples/simple/src/customRouteLayout.js
@@ -1,12 +1,15 @@
 import React from 'react';
-import { useGetList, useAuthenticated, Title } from 'react-admin';
+import { useGetList, useAuthenticated, Title, useVersion } from 'react-admin';
 
 const CustomRouteLayout = () => {
     useAuthenticated();
+    const version = useVersion();
     const { ids, data, total, loaded } = useGetList(
         'posts',
         { page: 1, perPage: 10 },
-        { field: 'published_at', order: 'DESC' }
+        { field: 'published_at', order: 'DESC' },
+        {},
+        { version }
     );
```

**Environment**

* React-admin version: current master branch 70e9b48
* Browser: Chrome

fzaninotto:
Fixes #4817

### DIFF
--- a/packages/ra-core/src/dataProvider/useQueryWithStore.ts
+++ b/packages/ra-core/src/dataProvider/useQueryWithStore.ts
@@ -3,6 +3,7 @@ import { useSelector } from 'react-redux';
 import isEqual from 'lodash/isEqual';
 
 import useDataProvider from './useDataProvider';
+import useVersion from '../controller/useVersion';
 import getFetchType from './getFetchType';
 import { useSafeSetState } from '../util/hooks';
 import { ReduxState } from '../types';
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/dataProvider/useQueryWithStore.ts
```diff
@@ -120,7 +121,8 @@ const useQueryWithStore = (
     loaded: boolean;
 } => {
     const { type, resource, payload } = query;
-    const requestSignature = JSON.stringify({ query, options });
+    const version = useVersion(); // used to allow force reload
```
fzaninotto, line 124: Now even if the caller forgets to pass the version in the query options, the side effect will re-execute when a REFRESH_VIEW action is dispatched.

Issue #1510, opened 2018-02-09
[RFC] support pagination using cursors


yonahforst:
I'm in the middle of writing a client for AWS DynamoDB. One of the problems is that the DB uses  cursor-based pagination (is that what it's called?)

I'm using  a custom reducer to keep track of the previous + next keys, but I need a way to pass those from the `<Pagination/>` component to the client.

At first I thought it would be a good idea to allow the page param to be an object, but that got messy quickly. there are several places where it's assumed to be an integer.

However, filters can be objects and maybe that's a better fit anyway. What do you guys thinking about exposing `setFilters` to the `<Pagination/>` component?

It's working great for me and I have a PR lined up to submit.

fzaninotto:
Good idea. And while we are at it, we could expose a `setPerPage` method, allowing the pagination to change the number of rows per page.

yonahforst:
Well, perhaps I spoke too soon when I said 'working great'. 

 I don't think my current implementation is suitable as generic DynamoDB client that anyone can use (which I would very much like it to be).

My main problem is that I don't know the prev/next cursor keys until the server responds with the contents of the current page. Somehow I need to pass those key to `<Pagination/>`. Now I'm doing it with a custom reducer and making it a ConnectedComponent but it's really ugly and requires a bunch of config.

I'm actually considering moving back to `page` and `perPage` and just maintaining a list of ids locally to the client. (DynamoDB cursors are just ids, so it's easy to figure out prev/next)

What would you guys recommend? 

another option would be just to have a custom saga that updates the filters with the next/prev keys upon `CRUD_GET_LIST_SUCCESS` and then read them in `<Pagination/>` without connecting it

yonahforst:
using page and perPage is working well for me.
https://github.com/abiglobalhealth/aor-dynamodb-client

still working on supporting   CREATE,  UPDATE,  DELETE

mrcasablr:
@yonahforst repo link is a 404

yonahforst:
Whoops. Had to make it public. Thanks for catching that

djhi:
Can you make a PR to add it to the ecosystem section in the documentation ?

yonahforst:
Done. Pr #1518

garethsb-sony:
We'd love to see cursor-based pagination support in React Admin.

[...]

PR #4834, opened 2020-05-20
Meta for Data Providers

alanpoulain:
Add a `meta` object to be returned by any method of a data provider.

This allows to add some information about the response.

It could be useful for a lot of things, for instance to tell if a pagination is partial or not (https://github.com/api-platform/admin/pull/297, see comment from @fzaninotto https://github.com/api-platform/admin/pull/297#issuecomment-627817098) or to add cursor-based pagination information (https://github.com/marmelab/react-admin/issues/1510).

### DIFF
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -339,15 +339,15 @@ Data Providers methods must return a Promise for an object with a `data` propert
 
 Method             | Response format
 ------------------ | ----------------
-`getList`          | `{ data: {Record[]}, total: {int}, validUntil?: {Date} }`
-`getOne`           | `{ data: {Record}, validUntil?: {Date} }`
-`getMany`          | `{ data: {Record[]}, validUntil?: {Date} }`
-`getManyReference` | `{ data: {Record[]}, total: {int} }`
-`create`           | `{ data: {Record} }`
-`update`           | `{ data: {Record} }`
-`updateMany`       | `{ data: {mixed[]} }` The ids which have been updated
-`delete`           | `{ data: {Record} }` The record that has been deleted
-`deleteMany`       | `{ data: {mixed[]} }` The ids of the deleted records (optional)
+`getList`          | `{ data: {Record[]}, total: {int}, meta?: {object}, validUntil?: {Date} }`
+`getOne`           | `{ data: {Record}, meta?: {object}, validUntil?: {Date} }`
+`getMany`          | `{ data: {Record[]}, meta?: {object}, validUntil?: {Date} }`
+`getManyReference` | `{ data: {Record[]}, meta?: {object}, total: {int} }`
+`create`           | `{ data: {Record}, meta?: {object} }`
+`update`           | `{ data: {Record}, meta?: {object} }`
+`updateMany`       | `{ data: {mixed[]}, meta?: {object} }` The ids which have been updated
+`delete`           | `{ data: {Record}, meta?: {object} }` The record that has been deleted
+`deleteMany`       | `{ data: {mixed[]}, meta?: {object} }` The ids of the deleted records (optional)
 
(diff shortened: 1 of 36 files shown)


alanpoulain:
Any input on this PR? Does it seem relevant?

fzaninotto:
First of all, thanks for the work you've done on this one. 

I've reviewed about half the PR last week before switching to something else. It's a lengthy change and I couldn't do it all on one go. 

It stayed in my mind all weekend ; I was uncomfortable with the changes I reviewed but I didn't manage to understand why. 

Now I think I understand why I don't really like it: in order to fix a problem with pagination on certain API backends, you introduce a change to *all* data provider requests (and not only `getList`), even though this change brings no benefit in most cases. Also, this doesn't really solve the pagination problem, as a custom Pagination control must still somehow dig in the Redux store to work. Finally, there is currently a fair amount of logic linked to the `total` value, from triggers to display List skeletons to changes when deleting an item. This logic wouldn't work if your pagination isn't total-based. 

In other words, I think you've gone too far with the wrong abstraction.

I think a better approach would be to try to model the pagination parameters in a more flexible way. The current modeling (page + perPage + total) doesn't work in some cases. What model could work in all cases?

But I'm leaving the PR open to let the discussion continue.

alanpoulain:
I've introduced the `meta` object not just for pagination, but possibly for other things, for instance if the received data are "subscribable" (could be necessary for a https://mercure.rocks/ implementation).

However you are probably right about the pagination abstraction. The current modeling doesn't reflect all the possibilities of an API:
- the `total` is not necessarily a known value.
- the `nextPage` and `previousPage` can be a useful information.
- the pagination can be cursor-based: it needs `cursor`, `startCursor`, `endCursor` properties.
- maybe it exists other kinds of pagination, it should easily extendable to them.

I'm not sure how the modeling should be done. Some properties seem essential for pagination like `nextPage` and `previousPage`. Others are specific. Something like this then?

```ts
interface Pagination {
  nextPage: boolean;
  previousPage: boolean;
  type: Page|Cursor|string;
  meta: PageMeta|CursorMeta|object;
};

interface PageMeta {
  page: int;
  perPage: int;
  total?: int;
};

interface CursorMeta {
  cursor: string;
  startCursor: string;
  endCursor: string;
};
```

bayareacoder:
@alanpoulain I'm not sure how to implement a 'plugin for react-admin' without manually changing their existing source?

Anyway, I have something working meanwhile. Still using a customReducer to get the pagination info to a CustomPagination component as in my prior post, but not inserting that in the `filter` prop of `List` since that causes a re-render loop. 
Rather just keep history of prior GET_LIST in buildQuery file and then compare current vs prior page to determine how to set the input params for a GraphQL Relay-compatible query. Assuming `cursor` has the info from prior GET_LIST and `params` is the usual params provided by react-admin for the current GET_LIST. 
This assumes only a PREV/NEXT pagination control.
```
          let before: string | undefined;
          let after: string | undefined;
          let first: number | undefined;
          let last: number | undefined;
          if (typeof cursor?.pagination.page !== "number") {
            // no prior cursor -> read first page
            first = params.pagination.perPage;
          }
          else if (params.pagination.page > cursor?.pagination.page) {
            // page forward -> use `after` & `first`
            after = cursor.pageInfo.endCursor;
            first = params.pagination.perPage;
          } else if (params.pagination.page < cursor?.pagination.page) {
            // page backward -> use `before` & `last`
            before = cursor.pageInfo.startCursor;
            last = params.pagination.perPage;
          } else {
            first = cursor?.params.first;
            last = cursor?.params.last;
            before = cursor?.params.before;
            after = cursor?.params.after;
```

fzaninotto:
I have another need that your proposal would help solve: search facets. 

When the backend uses a search engine, list responses often contain "facets", i.e. categories allowing to further filter the list, together with the cardinality of each category. 

There is currently no way to grab these facets from a getList response, and metas would definitely help.

PR #4837, opened 2020-05-21
fix custom application route documentation

### LINKED ISSUE #4832: Documentation update for CustomApp.md
**What you were expecting:**
The example for using \<Routes\> in **Not Using the \<Admin\> Components** section should be

`<Route exact path="/posts/:id" render={(routeProps) => <PostEdit hasShow resource="posts" basePath={routeProps.match.url} id={decodeURIComponent((routeProps.match).params.id)} {...routeProps} />} />`

**What happened instead:**
Instead the hasShow attribute is put onto the \<Route\> so PostEdit may give out a warning because the routeProps does not include the hasShow attribute.
id is not decoded from URI, so everytime dataProvier will fetch an undefined id.

**Other information:**
We also need to update appropriate section in UPGRADE.md
Update useCheckMinimumRequiredProps('Edit', ['basePath', 'resource'], props) to check for 'id' in the useEditController.js file to give out a warning

**Environment**

* React-admin version: 3.5.1
* React version:
* Browser:

smeng9:
Resolves #4832

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1250,10 +1250,32 @@ const App = () => (
                     <ConnectedRouter history={history}>
                         <Switch>
                             <Route exact path="/" component={Dashboard} />
-                            <Route exact path="/posts" hasCreate render={(routeProps) => <PostList resource="posts" {...routeProps} />} />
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

PR #4843, opened 2020-05-22
useListParams hook exported.

### LINKED ISSUE #4836: useListParams not exported from react-admin
## Bug report & general question

**What you were expecting:**

`import { useListParams } from 'react-admin` to give me the `useListParams` hook.

Question: _Should this be available to import? Is there a reason it is not?_

**What happened instead:**

App fails to compile with the follow error:
`Attempted import error: 'useListParams' is not exported from 'react-admin'.`

**Related code:**

The `useListParams` hook is not exported from https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/index.ts

**Environment**

* React-admin version: 3.5.1

WiXSL:
Fixes #4836

### DIFF
--- a/packages/ra-core/src/controller/index.ts
+++ b/packages/ra-core/src/controller/index.ts
@@ -2,10 +2,6 @@ import CreateController from './CreateController';
 import EditController from './EditController';
 import ListController from './ListController';
 import ShowController from './ShowController';
-import {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

PR #4849, opened 2020-05-25
Update information for Loopback4

cycorld:
Loopback is updated from 3 to 4.
Many people are using version 3 yet. Version 4 has many break changes.
Thankfully, elmaistrenko created a provider for the new version. It works well.


fzaninotto:
Thanks!

Issue #4813, opened 2020-05-14
Support passing an element as label to the fields and inputs


djhi:
**Is your feature request related to a problem? Please describe.**
When trying to pass an element as the `label` prop for an input or field, there are multiple warning in the console (one related to translation and another related to PropTypes)

**Describe the solution you'd like**
`<FieldTitle />` should check if `label` is a valid element before trying to translate it

fzaninotto:
Fixed by #4852

PR #4852, opened 2020-05-25
Add ability to use a React element as Field label

### LINKED ISSUE #4848: Use icons as field labels in Datagrid
I want to show an icon as a column header in Datagrid, like this:
Screen Shot 2020-05-24 at 10 00 36 PM

I can achieve this with: 
```jsx
label={<StarBorderIcon fontSize={'small'} className={classes.columnIcon} />}
```
The problem is that the `FieldTitle` component always tries to translate the field's `label`, and gives this error in the console:
```console
Warning: Missing translation for key: "[object Object]" 
    at FieldTitle (http://localhost:4533/static/js/0.chunk.js:106580:21)
    at pure(FieldTitle) (http://localhost:4533/static/js/0.chunk.js:187826:9)
    at th
    at TableCell (http://localhost:4533/static/js/0.chunk.js:27368:28)
    at WithStyles (http://localhost:4533/static/js/0.chunk.js:41841:31)
    at DatagridHeaderCell (http://localhost:4533/static/js/0.chunk.js:131626:25)
...
```

Maybe we could have a new property in the Field that accepts a component as label, instead of using the `label` itself (`iconLabel`?).

Can't think of anything else.

fzaninotto:
Closes #4848


djhi:
It also closes #4813

PR #4853, opened 2020-05-25
[RFR] Support array of objects in sanitizeEmptyValues

### LINKED ISSUE #4726: Empty values of object in an array are not sanitized by `sanitizeEmptyValues`
Here is my input :

`<TextInput source={"i18n[0].content"} />`

The record looks like

```
{
    id: 1,
    i18n: [{
        language_id: "fr",
        content: "Mon texte",
    }, {
        language_id: "en",
        content: "My text",
    }]
}
```

**What you were expecting:**

When the input has a value, it is sent to dataProvider.

When it is empty, the key is not sent at all.

I'd expect the sent record to be:

```
{
    id: 1,
    i18n: [{
        language_id: "fr",
        content: "",
    }, {
        language_id: "en",
        content: "",
    }]
}
```

or

```
{
    id: 1,
    i18n: [{
        language_id: "fr",
        content: null,
    }, {
        language_id: "en",
        content: null,
    }]
}
```

**What happened instead:**

The content of the actual sent record with an empty content field is:

```
{
    id: 1,
    i18n: [{
        language_id: "fr",
    }, {
        language_id: "en",
    }]
}
```

**Steps to reproduce:**
Just configure an input with array access

I'm creating a sandbox example

armellarcier:
Fixes #4726

This will not support an array of objects with different structure AND key reordering.

### DIFF
--- a/packages/ra-core/src/form/sanitizeEmptyValues.ts
+++ b/packages/ra-core/src/form/sanitizeEmptyValues.ts
@@ -16,8 +16,16 @@ const sanitizeEmptyValues = (initialValues: object, values: object) => {
     if (!initialValues) return values;
     const initialValuesWithEmptyFields = Object.keys(initialValues).reduce(
         (acc, key) => {
-            if (values[key] instanceof Date || Array.isArray(values[key])) {
+            if (values[key] instanceof Date) {
                 acc[key] = values[key];
+            } else if (Array.isArray(values[key])) {
+                if (Array.isArray(initialValues[key])) {
+                    acc[key] = values[key].map((value, index) =>
+                        sanitizeEmptyValues(initialValues[key][index], value)
+                    );
+                } else {
+                    acc[key] = values[key];
+                }
             } else if (
                 typeof values[key] === 'object' &&
                 typeof initialValues[key] === 'object' &&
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

armellarcier:
Glad I could help. As I stated in the PR description, this fixes a problem and should be a good solution for the vast majority of use cases but an array being what it is, reordering an array of objects with different structures will lead to props with null values in objects that didn’t have the prop in the first place. This is not perfect. Maybe the `parse={identity}` option should be considered (in « userland » as you may say).

packages/ra-core/src/form/sanitizeEmptyValues.ts
```diff
@@ -16,8 +16,16 @@ const sanitizeEmptyValues = (initialValues: object, values: object) => {
     if (!initialValues) return values;
     const initialValuesWithEmptyFields = Object.keys(initialValues).reduce(
         (acc, key) => {
-            if (values[key] instanceof Date || Array.isArray(values[key])) {
```
sebitoelcheater, line 19: What if `values` is null or undefined? is it possible? shouldn't there be an if for that?

PR #4859, opened 2020-05-26
Add ability to deselect rows in one click in BulkActionsToolbar

fzaninotto:
Peek 26-05-2020 23-43

Supersedes #3968

### DIFF
--- a/cypress/support/ListPage.js
+++ b/cypress/support/ListPage.js
@@ -21,9 +21,9 @@ export default url => ({
         logout: '.logout',
         bulkActionsToolbar: '[data-test=bulk-actions-toolbar]',
         customBulkActionsButton:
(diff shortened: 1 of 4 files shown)


cypress/support/ListPage.js
```diff
@@ -21,9 +21,9 @@ export default url => ({
         logout: '.logout',
         bulkActionsToolbar: '[data-test=bulk-actions-toolbar]',
         customBulkActionsButton:
-            '[data-test=bulk-actions-toolbar] button:first-child',
+            '[data-test=bulk-actions-toolbar] button[aria-label="Reset views"]',
```
djhi, line 24: :+1: Better but it would even better to not rely on those css selectors at all, just as users do. This is something we should fix in a pr dedicated to e2e tests though

PR #4864, opened 2020-05-27
Fix react imports

### LINKED ISSUE #4857: Fix React Imports
See https://github.com/facebook/react/pull/18102

The correct way to import React is:
```
```

djhi:
Fix #4857


deksden:
Small note: some md files in code samples have code with double quotes in import, like `import * as React from "react";`. All code actually have single-quoted imports.

PR #4872, opened 2020-05-28
Update dependencies

djhi:


fzaninotto:
I'm -1 about changing the version in peerDependencies. We don't need it, and it can be considered as a breaking change. We need to check the package is forward compatible, not break its backward compatibility.

packages/react-admin/package.json
```diff
@@ -34,22 +34,22 @@
         "react-dom": "^16.9.0"
     },
     "dependencies": {
-        "@material-ui/core": "^4.3.3",
-        "@material-ui/icons": "^4.2.1",
-        "@material-ui/styles": "^4.3.3",
+        "@material-ui/core": "^4.10.0",
```
fzaninotto, line 37: the dependencies in react-admin should not change either.

yarn.lock
```diff
@@ -873,13 +873,20 @@
...
 
-"@babel/runtime@7.8.4", "@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2", "@babel/runtime@^7.3.1", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.0", "@babel/runtime@^7.4.4", "@babel/runtime@^7.4.5", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.2", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.4", "@babel/runtime@^7.8.3":
+"@babel/runtime@7.8.4", "@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2", "@babel/runtime@^7.3.1", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.0", "@babel/runtime@^7.4.4", "@babel/runtime@^7.4.5", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.2", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.4":
   version "7.8.4"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.8.4.tgz#d79f5a2040f7caa24d53e563aad49cbc05581308"
   integrity sha512-neAp3zt80trRVBI1x0azq6c57aNBqYZH8KhMm3TaB7wEI5Q4A2SHfBHE8w9gOhI/lrqxtEbXZgQIrHP+wvSGwQ==
   dependencies:
     regenerator-runtime "^0.13.2"
 
+"@babel/runtime@^7.10.0":
```
fzaninotto, line 883: This one is troublesome. Can you check that we don't double the runtime in production bundles?

djhi: After building the demo, I can confirm the runtime isn't bundled twice

djhi: In fact, the final bundle is even smaller (-2.45 KB) :smile:

Issue #4505, opened 2020-03-10
Add Typescript types


fzaninotto:
## Context

We want to migrate react-admin to TypeScript and release types with the library. 

**That doesn't mean that you won't be able to use react-admin if you're not using TypeScript**. React-admin will still publish both JS files (as today) and .d.ts files (ignored by non-typescript users). So if you don't use TypeScript, that won't change anything for you. 

We've started a migration effort a year ago. As of writing, 70% of the codebase is in already TypeScript. We use these types internally during development. We've only released the types for `ra-core` publicly.

We're working on migrating the rest of the codebase. It's a long and tedious job. It's not the core team priority. Your help is welcome.

## Risk

We don't want to emit bad types, because it would break existing TS applications. Therefore, we need to test the types thoroughly. 

## Roadmap

Migrating the demo should be the priority, as it allows us to validate our types internally before publishing them.

1. ✔️ Migrate all .js files to .ts files in the example demo (https://github.com/marmelab/react-admin/tree/master/examples/demo).
2. Migrate all `ra-ui-materialui` JS files to TypeScript. Publish the compiled types in the next ra-ui-material-ui release.
3. Convert the rest of react-admin dependencies (ra-language-english, etc) to typeScript, publish their types
4. Locally, emit the types from react-admin in development. This is when things will break. Test the demo, test on as many existing TS apps as possible. Once all the types are checked, and only then,
5. Release `react-admin` types officially
6. ???
7. Profit

All these changes must be done on the `next` branch (`master` is only for bug fixes).

## How you can help

If you know and understand the react-admin internals, and if you are an experienced TypeScript developer, please help us migate .js files to .ts in the order described above.

This is not an easy task. If you don't have enough knowledge of the react-admin internals, then please leave it to people more knowledgeable. Otherwise, the core team will spent more time answering your questions than doing the migration themselves. 

## What to do in the meantime

Some community packages for react-admin types can be found here and there. They are outdated. Disable TypeScript checking on the react-admin modules.

josephktcheung:
@fzaninotto thanks for writing this. I will split my last PR on `ra-ui-materialui` #4448 into smaller ones for review, but I think I’ll take a stab at Point 1 migrating `demo` to ts first.

iamstiil:
Happy to help :)

sergelg90:
[...]

PR #4873, opened 2020-05-28
Migrate AppBar to tsx

josephktcheung:
Migrate AppBar to tsx #4505

### DIFF
--- a/packages/ra-ui-materialui/src/layout/AppBar.tsx
+++ b/packages/ra-ui-materialui/src/layout/AppBar.tsx
@@ -1,15 +1,17 @@
-import React, { Children, cloneElement } from 'react';
+import React, { Children, cloneElement, FC, ReactElement } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 import classNames from 'classnames';
 import {
     AppBar as MuiAppBar,
+    AppBarProps as MuiAppBarProps,
     IconButton,
     Toolbar,
     Tooltip,
     Typography,
     makeStyles,
     useMediaQuery,
+    Theme,
 } from '@material-ui/core';
 import MenuIcon from '@material-ui/icons/Menu';
 import { toggleSidebar, useTranslate } from 'ra-core';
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/AppBar.tsx
```diff
@@ -147,17 +150,19 @@ const AppBar = props => {
     );
 };
 
+interface Props {
```
MohammedFaragallah, line 153: Hello @josephktcheung can you please add `title` property as `TitleComponent` from `ra-core`

josephktcheung: Hi @TheHyphen thanks for the review. Good catch on this, from what I understand `AppBar` no longer uses `title` prop, the reason why it's not removed is because [we inject `title` prop to `AppBar` in `Layout`](https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/layout/Layout.js#L124). If we don't de-structure `AppBar` props [here](https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/layout/AppBar.js#L90), we may pass a wrong `title` type to material-ui's `AppBar` by spreading the `rest` prop [here](https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/layout/AppBar.js#L101)

Therefore, I question the need of passing `title` prop when we `cloneElement` in https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/layout/Layout.js#L124. I think `title` prop can be removed from `AppBar` which we did previously https://github.com/marmelab/react-admin/commit/a9e4e0d8b3a063eb686e4a15bbae94384a1c1528 but somehow reverted back by @fzaninotto https://github.com/marmelab/react-admin/commit/0f1b4043a37af1c15579be93a18474fdbc2c7a0d#diff-bbb6c2d0f8bfb7588ebadf3e64ed2e57.

@fzaninotto please confirm if `title` prop is still relevant for `AppBar` or is it a legacy code?

PR #4875, opened 2020-05-29
convert ra-ui-materialui layout to typescript

MohammedFaragallah:
refs #4505


MohammedFaragallah:
@fzaninotto this is ready for review but if it's not clear enough I'll be able to add some comments soon

some questions
- I think we should create a separate ErrorBounday component and covert layout to a functional component
- some component destructure `classes` property out I think we can remove this pattern and typescript will warn users if they pass a non-supported property

fzaninotto:
> I think we should create a separate ErrorBounday component and convert Layout to a functional component

I agree, good idea!

> some component destructure classes property out I think we can remove this pattern and typescript will warn users if they pass a non-supported property

Not sure I understand. Can you give me an example?

packages/ra-ui-materialui/src/defaultTheme.ts
```diff
@@ -7,11 +27,6 @@ export default {
             contrastText: '#fff',
         },
     },
-    typography: {
```
fzaninotto, line 10: Why did you remove this one?

packages/ra-ui-materialui/src/layout/CardContentInner.tsx
```diff
@@ -30,11 +30,11 @@ const useStyles = makeStyles(
  * padding double the spacing between each CardContent, leading to too much
  * wasted space. Use this component as a CardContent alternative.
  */
-const CardContentInner = props => {
-    const { className, children } = props;
+const CardContentInner: FC<CardContentProps> = props => {
+    const { className, children, ...rest } = props;
```
fzaninotto, line 34: Not necessary for now, may have side effects... Please revert this change.

packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -52,7 +52,15 @@ function goBack() {
     window.history.go(-1);
 }
 
-const Error = props => {
+export interface ErrorComponentProps {
```
fzaninotto, line 55: ErrocComponentProps => ErrorProps (we never include the name "component" in other interface names)

packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -86,12 +94,8 @@ const Error = props => {
                     </ExpansionPanel>
                 )}
                 <div className={classes.toolbar}>
-                    <Button
-                        variant="contained"
-                        icon={<History />}
```
fzaninotto, line 91: I think `startIcon` is the right prop to get the alignment icon/text

packages/ra-ui-materialui/src/layout/Layout.tsx
```diff
@@ -74,11 +85,59 @@ const sanitizeRestProps = ({
...
+    className?: string;
+    customRoutes?: any[];
+    dashboard?: DashboardComponent;
+    error?: ComponentType<ErrorComponentProps>;
+    logout?: ReactNode;
+    menu?: ComponentType<MenuProps>;
+    notification?: ComponentType<NotificationProps>;
+    open?: boolean;
+    sidebar?: ComponentType<DrawerProps>;
+    title: TitleComponent;
```
fzaninotto, line 105: It's optional at build time (but required at runtime).

packages/ra-ui-materialui/src/layout/Loading.tsx
```diff
@@ -53,7 +59,6 @@ const Loading = props => {
 };
 
 Loading.propTypes = {
-    classes: PropTypes.object,
```
fzaninotto, line 56: You shouldn't remove this one. When `classes` is present in the props, useStyles() uses it to override the default styles.

packages/ra-ui-materialui/src/layout/Loading.tsx
```diff
@@ -33,7 +33,13 @@ const useStyles = makeStyles(
     { name: 'RaLoading' }
 );
 
-const Loading = props => {
+interface Props {
+    className?: string;
```
fzaninotto, line 37: you miss one prop: `classes?`

packages/ra-ui-materialui/src/layout/Menu.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement } from 'react';
+import React, { FC, ReactElement, ReactNode } from 'react';
```
fzaninotto, line 1: Then I think ReactElement is not used anymore

packages/ra-ui-materialui/src/layout/NotFound.tsx
```diff
@@ -45,8 +46,14 @@ function goBack() {
     window.history.go(-1);
 }
 
-const NotFound = props => {
-    const { className, classes: classesOverride, title, ...rest } = props;
+interface Props extends RouteComponentProps {
```
fzaninotto, line 49: make sure you export it.

packages/ra-ui-materialui/src/layout/NotFound.tsx
```diff
@@ -45,8 +46,14 @@ function goBack() {
...
+    className: string;
+    title: TitleProps['defaultTitle'];
+    classes?: object;
+}
+
+const NotFound: FC<Props> = props => {
+    const { className, title, ...rest } = props;
```
fzaninotto, line 56: you should also add `classes: classesOverride` to the destructuring, or the root `<div>` may end up with a `classes` attribute which isn't DOM compliant.

packages/ra-ui-materialui/src/layout/NotFound.tsx
```diff
@@ -45,8 +46,14 @@ function goBack() {
     window.history.go(-1);
 }
 
-const NotFound = props => {
-    const { className, classes: classesOverride, title, ...rest } = props;
+interface Props extends RouteComponentProps {
+    className: string;
+    title: TitleProps['defaultTitle'];
+    classes?: object;
```
fzaninotto, line 52: I think MUI offers a way to infer the accepted type based on the makeStyles.

packages/ra-ui-materialui/src/layout/NotFound.tsx
```diff
@@ -62,8 +69,8 @@ const NotFound = props => {
                 <div>{translate('ra.message.not_found')}.</div>
             </div>
             <div className={classes.toolbar}>
-                <Button variant="contained" icon={<History />} onClick={goBack}>
```
fzaninotto, line 65: use startIcon

packages/ra-ui-materialui/src/layout/Sidebar.tsx
```diff
@@ -45,22 +55,21 @@ const useStyles = makeStyles(
...
 
-const Sidebar = props => {
-    const {
-        children,
-        closedSize,
-        size,
-        classes: classesOverride,
-        ...rest
-    } = props;
+interface Props extends DrawerProps {
```
fzaninotto, line 58: let's call it `SidebarProps` and export it.

packages/ra-ui-materialui/src/layout/Sidebar.tsx
```diff
@@ -45,22 +55,21 @@ const useStyles = makeStyles(
...
-    useSelector(state => state.locale); // force redraw on locale change
+    const isXSmall = useMediaQuery<Theme>(theme =>
+        theme.breakpoints.down('xs')
+    );
+    const isSmall = useMediaQuery<Theme>(theme => theme.breakpoints.down('sm'));
+    const open = useSelector((state: ReduxState) => state.admin.ui.sidebarOpen);
     const handleClose = () => dispatch(setSidebarVisibility(false));
     const toggleSidebar = () => dispatch(setSidebarVisibility(!open));
-    const classes = useStyles({ ...props, open });
+    const classes = useStyles({ open });
```
fzaninotto, line 72: not good, we must pass the complete `props` object or this will cause rerenderings

packages/ra-ui-materialui/src/layout/Sidebar.tsx
```diff
@@ -45,22 +55,21 @@ const useStyles = makeStyles(
...
-const Sidebar = props => {
-    const {
-        children,
-        closedSize,
-        size,
-        classes: classesOverride,
-        ...rest
-    } = props;
+interface Props extends DrawerProps {
+    children: ReactElement;
```
fzaninotto, line 59: I think it should be `React.ReactNode`

packages/ra-ui-materialui/src/layout/TitleForRecord.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC } from 'react';
+import PropTypes from 'prop-types';
+import Title, { TitlePropType, TitleProps } from './Title';
+
+interface Props extends TitleProps {
+    record: TitleProps['record'];
+}
+
+const TitleForRecord: FC<Props> = props =>
+    props.record ? <Title {...props} /> : null;
```
fzaninotto, line 10: the previous version white listed the props it passed to the `<Title>` component. Please keep that behavior.

packages/ra-ui-materialui/src/defaultTheme.ts
MohammedFaragallah: I didn't mean to push that with this PR but I checked everywhere there is no use for variant `title`

packages/ra-ui-materialui/src/layout/Sidebar.tsx
MohammedFaragallah: ReactNode gives an error

`Type 'string' is not assignable to type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>'.`

packages/ra-ui-materialui/src/layout/NotFound.tsx
MohammedFaragallah: I don't know about it but something like ReturnType<typeof useStyles> should work?

packages/ra-ui-materialui/src/layout/NotFound.tsx
MohammedFaragallah: > > some component destructure classes property out I think we can remove this pattern and typescript will warn users if they pass a non-supported property
> 
> Not sure I understand. Can you give me an example?

that's what I was talking about, didn't know about `makeStyles` override behavior, Thank you

packages/ra-ui-materialui/src/layout/NotFound.tsx
fzaninotto: absolutely!

fzaninotto: No, wait, it won't work. the `classes` *argument* passed to `useStyles` isn't the same type as the *result* of `useStyles`, which is usually named `classes`.

I'm looking for the right way to type the `classes` override (the material-ui documentation doesn't mention it).

fzaninotto: Okay, there is no easy way to type these - see my comment in https://github.com/mui-org/material-ui/issues/17973#issuecomment-638868260.

Let's default to `any` for now.

fzaninotto: Or we can add a generic type to `ra-ui-material-ui`:

```tsx
import { CSSProperties, CreateCSSProperties } from "@material-ui/styles";

type PropsFunc<Props extends object, T> = (props: Props) => T;

export type ClassesOverride<
  UseStyles extends (props: any) => Record<string, string>,
  Props extends object = {}
> = Partial<
  {
    [p in keyof ReturnType<UseStyles>]:
      | CSSProperties
      // JSS property bag where values are based on props
      | CreateCSSProperties<Props>
      // JSS property bag based on props
      | PropsFunc<Props, CreateCSSProperties<Props>>
  }
>;
```

which allows to type our `classes` props as follows:

```tsx
interface Props {
    classes: ClassesOverride<typeof useStyles>;
    ...
}
```

What do you think?

MohammedFaragallah: Excellent but from what I understand the parent component passes a `classes` object each key is a `className` string, not `CSSProperties` object?

```jsx
import { makeStyles } from '@material-ui/core';
import React, { FC } from 'react';

const useNotFoundStyles = makeStyles(
	theme => ({
		root: {
			display: 'flex',
		},
	}),
	{ name: 'RaNotFound' },
);

type ClassesOverride<
	UseStyles extends (props: any) => Record<string, string>
> = Partial<Record<keyof ReturnType<UseStyles>, string>>;

interface NotFoundProps {
	classes?: ClassesOverride<typeof useNotFoundStyles>;
}

export const NotFound: FC<NotFoundProps> = props => {
	const classes = useNotFoundStyles(props);
	return <div className={classes.root}>DEMO</div>;
};
```

```jsx
const useLayoutStyles = makeStyles(
  theme => ({
    root: {
      display: "block"
    }
  }),
  { name: "RaLayout" }
);

interface LayoutProps {}

export const Layout: React.FC<LayoutProps> = props => {
  const classes = useLayoutStyles(props);
  return (
    <div>
      <NotFound classes={{ root: classes.root }} />
    </div>
  );
};
```

fzaninotto: Totally, you're right and I was too tired when writing my comment. Sorry for that!

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -26,6 +32,13 @@ const Title = ({ className, defaultTitle, locale, record, title, ...rest }) => {
     return createPortal(titleElement, container);
 };
 
+export interface TitleProps {
+    defaultTitle: TitleComponent;
```
fzaninotto, line 36: it's optional. Either one of `title` or `defaultTitle` should be defined.

packages/ra-ui-materialui/src/layout/TitleForRecord.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <Title
+            title={title}
+            record={record}
+            defaultTitle={defaultTitle}
+            className={className}
+        />
+    ) : null;
+
+interface Props extends TitleProps {
+    record: TitleProps['record'];
```
fzaninotto, line 21: I don't understand this type. It specializes nothing.

packages/ra-ui-materialui/src/layout/NotFound.tsx
```diff
@@ -76,13 +81,21 @@ const sanitizeRestProps = ({
...
     ...rest
-}) => rest;
+}: Partial<NotFoundProps>): Omit<
+    Partial<NotFoundProps>,
+    keyof RouteComponentProps | 'className' | 'title'
+> => rest;
+
+export interface NotFoundProps extends RouteComponentProps {
+    className: string;
+    title: TitleProps['defaultTitle'];
```
fzaninotto, line 91: Optional I think

packages/ra-ui-materialui/src/layout/NotFound.tsx
MohammedFaragallah: No worries you are working on so many things, I'll add this type to Layout components for now and start a new PR for all `ra-material-ui` other components.

packages/ra-ui-materialui/src/layout/TitleForRecord.tsx
MohammedFaragallah: to make `record` required, but to make things more clear I can add a helper type?

```ts
export type SomeRequired<T, U extends keyof T> = Omit<T, U> &
  Required<Pick<T, U>>;
```

```diff
+ interface Props extends <TitleProps,'record'> {
- interface Props extends TitleProps {
-     record: TitleProps['record'];
```

WiXSL:
Is this PR still valid?

Luwangel:
@WiXSL I don't know. Let's ask François, he'll be back soon :wink: 

@fzaninotto I'm not sure I understand why you closed that PR.

PR #4878, opened 2020-05-30
Fix mobile version of ReviewList on the example demo

alanpoulain:
Checking the `loading` prop for rendering the `SimpleList` was wrong, it should be `!loading` or `loaded`.

The `ReviewMobileList` component in the demo has also been fixed (it's coming from the `SimpleList` component).

### DIFF
--- a/examples/demo/src/reviews/ReviewListMobile.js
+++ b/examples/demo/src/reviews/ReviewListMobile.js
@@ -29,10 +29,10 @@ const useStyles = makeStyles({
     },
 });
 
-const ReviewMobileList = ({ basePath, data, ids, loading, total }) => {
+const ReviewMobileList = ({ basePath, data, ids, loaded, total }) => {
     const classes = useStyles();
     return (
-        (loading || total > 0) && (
+        (loaded || total > 0) && (
             <List className={classes.root}>
                 {ids.map(id => (
                     <Link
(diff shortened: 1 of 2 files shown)


fzaninotto:
In SimpleList, at that stage, the `loaded` prop is always true, because of the previous test:

```jsx
    if (loaded === false) {
        return (
            <SimpleListLoading
                classes={classes}
                className={className}
                hasLeftAvatarOrIcon={!!leftIcon || !!leftAvatar}
                hasRightAvatarOrIcon={!!rightIcon || !!rightAvatar}
                hasSecondaryText={!!secondaryText}
                hasTertiaryText={!!tertiaryText}
            />
        );
    }

    return (
        (loading || total > 0) && (
            <List className={className} {...sanitizeListRestProps(rest)}>
...
```

So a test for the `loaded` value is useless. 

The question is: is the test for the `loading` value useful? There is a case where both `loading` and `loaded` can be true: when fetching the data for the list after it's already been rendered once (i.e. when coming back to a list page already viewed). The current test forces the display of the list in that case even though the total is null. I don't see the point. 

So you can *remove* the test on the `loading` value altogether in SimpleList. 

I agree with the change on ReviewListMobile.

fzaninotto:
Thanks!

PR #4881, opened 2020-06-01
Add create / edit side effects and transform

fzaninotto:
## Problem

I wanted to force a refresh after save in a `<Create>` view. I found it super hard to do even though I know the react-admin internals. I realized it's probably impossible for someone who doesn't know react-admin. 

Also, there are ways to override the notification message in the Create/Edit view (`successMessage`) and ways to override the `redirect` in `SimpleForm`. These are both side effects that should be the responsibility of the main controller, not the form. These props allow a very granular customization, but don't allow adding another side effect.

Finally, the recent addition of the `onSave` prop requires too much documentation and too much boilerplate for most common use cases. 

## Solution

Allow users to declare success and failure side effects at the controller side, and to override them for each `SaveButton`.

Introduce a `transform` function to transform a submitted record before it is sent to the `dataProvider`.

## Examples

Customizing the notification based on the error from the `dataProvider`:

```jsx
import { Edit, useNotify, useRedirect } from 'react-admin';

const PostEdit = props => {
    const notify = useNotify();
    const onFailure = (error) => {
        if (error.code == 123) {
            notify('Could not save changes: concurrent edition in progress', 'warning');
        } else {
            notify('ra.notification.http_error', 'warning')
        }
        redirect('list', props.basePath);
    }
    return (
        <Edit {...props} onFailure={onFailure}>
            ...
        </Edit>
    );
}
```

Altering the form value when the user clicks on a particular save button:

```jsx
const PostCreateToolbar = props => (
    <Toolbar {...props}>
        <SaveButton submitOnEnter={true} />
        <SaveButton
            label="post.action.save_and_notify"
            transform={data => ({ ...data, notify: true })}
            submitOnEnter={false}
        />
    </Toolbar>
);

const PostCreate = (props) => (
    <Create {...props}>
        <SimpleForm toolbar={<PostCreateToolbar />}>
            // ...
        </SimpleForm>
    </Create>
);
```

## Tasks

- [x] Add `onSuccess`, `onFailure` and `transform` support in `<Create>` and `<Edit>`
- [x] Add `onSuccess`, `onFailure` and `transform` support in `<SaveButton>`
- [x] Deprecate `successMessage`
- [x] Deprecate `onSave`
- [x] Add unit & integration tests for the new props
- [x] Update documentation

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -20,9 +20,11 @@ Here are all the props accepted by the `<Create>` and `<Edit>` components:
 * [`title`](#page-title)
 * [`actions`](#actions)
 * [`aside`](#aside-component)
-* [`successMessage`](#success-message)
 * [`component`](#component)
 * [`undoable`](#undoable) (`<Edit>` only)
+* [`successMessage`](#success-message)
+* [`onSuccess`](#onSuccess)
+* [`onFailure`](#onFailure)
 
 Here is the minimal code necessary to display a form to create and edit comments:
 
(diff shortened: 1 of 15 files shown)


docs/CreateEdit.md
```diff
@@ -5,46 +5,21 @@ title: "The Create and Edit Views"
...
-* [`successMessage`](#success-message)
-* [`onSuccess`](#onSuccess)
-* [`onFailure`](#onFailure)
+```
+<Resource name="posts" create={PostCreate} edit={PostEdit} />
+                               ----------        --------
+                                    |               |
+    displayed when browsing to /posts/create        |
+                                                    |
+                    displayed when browsing to /posts/123
```
djhi, line 16: :+1:

PR #4882, opened 2020-06-01
[2.x.x] Add disabled option in SelectArrayInput

maxmalov:
This is mostly backporting of https://github.com/marmelab/react-admin/pull/2555 but for `SelectArrayInput`. Currently, we stuck at version 2.9.9, that's why I've picked branch `2.x.x` as the base branch.

If these changes are ok and good to go, I would like to "forwardport" them to the latest version a little bit later :wink:


package.json
```diff
@@ -52,12 +52,13 @@
...
         "eslint-plugin-prettier": "^3.1.0",
         "eslint-plugin-react": "^7.13.0",
-        "eslint-config-react-app": "^4.0.1",    
+        "eslint-config-react-app": "^4.0.1",
         "eslint-plugin-react-hooks": "^1.6.0",
         "express": "~4.16.3",
         "full-icu": "~1.2.1",
         "husky": "^2.3.0",
         "jest": "^23.6.0",
+        "jest-dom": "3.5.0",
```
maxmalov, line 61: I have no other idea hot to test whether a dom element (for menu item) is disabled. I've picked the old version with the purpose of having the smallest changes to `yarn.lock`

fzaninotto:

Version 2 is in maintenance mode, so we won't be adding new features to it. If you need enhancements on that version, I suggest you fork it and republish it under another organization. 

To add new features, please use the `next` branch (currently based on version 3).

alextaymx:
hi , do we still have disabled options in select array input?

Mwamitovi:
Same issue here, adding "disabled: true" doesn't seem to disable a fields within `<SelectArrayInput>`. 
Any alternatives?

PR #4901, opened 2020-06-05
Make boolean field less tall by default

fzaninotto:
## Problem

`<BooleanField>` makes all lines in a `<Datagrid>` taller. 

## Solution

Make it smaller. Also, use smaller chips in `<ReferenceManyField>` in datagrids. This makes the grid data density consistent whether it contains `<BooleanField>` components or not. 

## Before

image

## After

image

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -17,11 +17,16 @@ interface Props extends FieldProps {
     valueLabelFalse?: string;
 }
 
-const useStyles = makeStyles({
-    root: {
-        display: 'flex',
+const useStyles = makeStyles(
+    {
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -16,10 +17,16 @@ interface Props extends FieldProps {
     valueLabelFalse?: string;
 }
 
+const useStyles = makeStyles({
+    root: {
+        display: 'flex',
+    },
+});
```
WiXSL, line 24: You should add `{ name: 'RaBooleanField' }`

fzaninotto: Absolutely, thanks for the review!

PR #4904, opened 2020-06-06
Fix: pristine not applied from edit toolbar on save button

kopax:
- Fix https://github.com/marmelab/react-admin/issues/4733


djhi:
Thanks!

PR #4908, opened 2020-06-07
[RFR] feat: Added disableIndex prop for SimpleFormIterator to omit rendering of the current index

pefi1011:
Hi,
added simple prop for SimpleFormIterator to disable rendering of index.

Here is how simple demo it looks like if prop `disableIndex` is passed

Screenshot 2020-06-07 at 09 48 11

I am not sure how and where to write the test for this? Any advise or guidance would be greatly appreciated


djhi:
Hi and thanks for you pull request. I actually wonder why we added this `Typography` element instead of using an `<ol>`. Using an `<ol>` would allow users to customize the index display using simple CSS.

Ping @fzaninotto

fzaninotto:
Indeed, using an `<ol>` tag seems to be a better solution, preventing the addition of another prop to customize the look and feel. 

@pefi1011 Would you like to work on that implementation?

fzaninotto:
Any updates on this one?

pefi1011:
> Any updates on this one?

I could try to implement the OL solution after 15th of July if thats not too late.

PR #4256, opened 2020-01-06
Convert ra-data-graphql to TypeScript

CarsonF:

PR #4909, opened 2020-06-07
convert ra-data-graphql to TypeScript, update dataProvider

hgiasac:
This PR is forked from https://github.com/marmelab/react-admin/pull/4256, with dataProvider update, due to inactivity

Update breaking changes on `ra-data-graphql-simple` and `ra-data-graphcool`


packages/ra-data-graphql/src/constants.ts
```diff
@@ -18,4 +18,4 @@ export const MUTATION_TYPES = [
     UPDATE_MANY,
     DELETE_MANY,
 ];
-export const ALL_TYPES = QUERY_TYPES.concat(MUTATION_TYPES);
+export const ALL_TYPES = [...QUERY_TYPES, ...MUTATION_TYPES] as string[];
```
fzaninotto, line 21: I don't understand why you need the `as` here.

hgiasac:
> You mention breaking changes. What are they?

These libraries extend `ra-data-graphql` with old data provider interface. It needs to be update with new data provider. https://github.com/marmelab/react-admin/pull/4909/files#diff-6c733a2a555721bcbd187734093ea94a

packages/ra-data-graphql/src/constants.ts
hgiasac: I don't remember why too. I will remove it

fzaninotto:
I don't see what is a breaking change. You've updated these libs, they should work seamlessly, or am I missing something?

hgiasac:
@fzaninotto Yes, I updated them. You can ignore that comment

PR #4911, opened 2020-06-08
Set HttpError name to be HttpError

jesseshieh:
While trying to use `error.name`, I noticed that the HttpError used by react-admin has its name set to `Error` instead of `HttpError`. I believe this is probably due to https://github.com/Microsoft/TypeScript/issues/14099

Setting it explicitly seems like it would be fine in this case, but let me know if you prefer a different solution.


jesseshieh:
np, done

djhi:
Thanks!

Issue #4916, opened 2020-06-08
ArrayField with nested SingleFieldList broken in version 3.6.0


sebitoelcheater:
**What you were expecting:**
`SingleFieldList` to work properly inside `ArrayField`, as in version 3.5.5.
```
<ArrayField source="statements">
  <SingleFieldList>
    <ChipField source="name"/>
  </SingleFieldList>
</ArrayField>
```

**What happened instead:**
It throws
```
TypeError: Cannot read property 'map' of undefined
```

**Steps to reproduce:**
Visit `list` in any resource view with `List` component containing `ArrayField` with a nested `SingleFieldList`:
```
<ArrayField source="statements">
  <SingleFieldList>
    <ChipField source="name"/>
  </SingleFieldList>
</ArrayField>
```

**Related code:**
```
export const CompanyList = props => (
  <List {...props}>
    <Datagrid rowClick="edit">
      <ArrayField source="statements">
        <SingleFieldList>
          <ChipField source="name"/>
        </SingleFieldList>
      </ArrayField>
    </Datagrid>
  </List>
);
```

**Other information:**
[This PR](https://github.com/marmelab/react-admin/pull/4536) seems to be causing the error.

**Environment**

* React-admin version: 3.6.0
* Last version that did not exhibit the issue (if applicable): 3.5.5
* React version: ^16.13.1
* Browser: Chrome
* Stack trace (in case of a JS error): TypeError: Cannot read property 'map' of undefined

djhi:
Fixed by #4918

DeciusAntiochus:
Just had the same issue since I went on the 3.6.0 version, but with different error

```
Uncaught TypeError: Cannot read property 'map' of undefined
    at SingleFieldList (0.chunk.js:138755)
```

My list component:
```
export const AbonnementList = (props) => (
  <List {...props} filters={<FullFilter />}>
    <Datagrid rowClick='edit'>
      <TextField source='id' />
      <TextField source='name' />
      <TextField source='price' />
      <TextField label='Période de facturation' source='pricePeriod' />
      <TextField label="Période d'octroi des crédits" source='offeredPeriod' />

      <ArrayField source='descriptions'>
        <SingleFieldList linkType={false}>
          <TextField source='description'></TextField>
        </SingleFieldList>
      </ArrayField>
      <EditButton />
    </Datagrid>
  </List>
);```

And of course, the fix proposed in the PR didn't work.

fzaninotto:
[...]

PR #4918, opened 2020-06-08
fix ArrayField initial state value

smeng9:
Hi,

When I am using `<SingleFieldList>` component as children of `<ArrayField>`, there seems to have a race condition that will make the whole `<List>` view of a resource crash sometime.

The error message is:
> TypeError: ids is undefined

I have traced the issue to its parent component. The useState hook is not initializing ids and data properly, setting them to undefined value. The useEffect hook will run during first render, but the state update function setIds and setData are asynchronous. This will cause the undefined ids to be passed down and ids.map function cannot be called properly.

The fix is also simple, just adding initial states.

Thank you

### DIFF
--- a/packages/ra-ui-materialui/src/field/ArrayField.tsx
+++ b/packages/ra-ui-materialui/src/field/ArrayField.tsx
@@ -134,8 +134,8 @@ export const ArrayField: FunctionComponent<
         fieldKey,
         ...rest
     }) => {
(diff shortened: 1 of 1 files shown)


smeng9:
Related Issue #4916
&
Related PR #4536

djhi:
Thanks!

PR #4930, opened 2020-06-11
select the first resource which has "list page" as init point

Slavvkko:
For example, I have resources that registered differently for users by their permissions.
Imagine the "News" resource is first on the list.
For the first user "News" registered with full params (name, icon, list, create, edit).
For the second user "News" registered only with "name", for fetching by useGetList or any other.


fzaninotto:
You should update the unit tests, too.

PR #4937, opened 2020-06-13
Add generics to TranslationMessage

iamstiil:
This will allow translations to be typed and extended with custom translations.

If this change is fine, I will continue working on it to fully include it in react-admin.

### DIFF
--- a/packages/ra-core/src/i18n/TranslationMessages.ts
+++ b/packages/ra-core/src/i18n/TranslationMessages.ts
@@ -2,11 +2,9 @@ interface StringMap {
     [key: string]: StringMap | string | undefined;
 }
 
-export interface TranslationMessages extends StringMap {
+export interface RATranslationMessages {
     ra: {
         action: {
-            // for custom translation strings
-            [key: string]: StringMap | string;
             add_filter: string;
             add: string;
             back: string;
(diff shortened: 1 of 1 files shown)


fzaninotto:
Can you please explain why this change is needed (i.e. the problem it solves), and provide an example of how it works?

iamstiil:
Let's assume I'm using `ra-i18n-polyglot` aswell for my translations inside of the app, then I'll propably import the translations from `ra-language-english` and extend those with my own translations.

```ts
import enDefaults from 'ra-language-english';
import { TranslationMessage } from 'react-admin';

import CustomTranslation from './CustomTranslation.ts';

const en: TranslationMessage<CustomTranslation> = {
  ...enDefaults,
  myKey: 'myTranslation',
};

export default en;
```

with `./CustomTranslation.ts` being:
```ts
export interface CustomTranslation {
  myKey: string;
}

export default CustomTranslation;
```

Now any change to `CustomTranslation` will force all languages to be translated accordingly. Those will need to include all `ra` translations aswell as all `CustomTranslation`s.

fzaninotto:
Why don't you use a union type in that case?

iamstiil:
That's what I'm currently doing but I feel like generics would make it more clean

fzaninotto:
Thanks for your PR. 

This is a breaking change, and we won't merge any of those before the next major version (5.0). Besides, you can already force the type of your custom translation messages with a custom type. 

So we won't merge this to the core. I'm closing this PR.

PR #4941, opened 2020-06-14
Port assert to expect in tests.

WiXSL:


packages/ra-core/src/controller/field/ReferenceManyFieldController.spec.tsx
```diff
@@ -1,9 +1,20 @@
...
+// Ignore warnings about act()
+// See https://github.com/testing-library/react-testing-library/issues/281,
+// https://github.com/facebook/react/issues/14769
+const originalError = console.error;
+jest.spyOn(console, 'error').mockImplementation((...args) => {
+    if (/Warning.*not wrapped in act/.test(args[0])) {
+        return;
+    }
+    originalError.call(console, ...args);
+});
```
djhi, line 16: I'm not sure we should do this. We are probably not waiting for side effects to apply where we should.

WiXSL: Done!

djhi: If you have time and noticed those warnings, would you mind trying to add the wait/expect calls needed to make them go away ?

WiXSL: Ok. I don't get any warnings now!

packages/ra-core/src/dataProvider/fetch.spec.ts
```diff
@@ -79,11 +78,11 @@ describe('createHeadersFromOptions', () => {
         };
 
         const headersWithMethod = createHeadersFromOptions(optionsWithMethod);
-        assert.strictEqual(headersWithMethod.get('Content-Type'), null);
+        expect(headersWithMethod.get('Content-Type')).toStrictEqual(null);
```
fzaninotto, line 81: toBeNull? Same for the next one

packages/ra-core/src/reducer/admin/notifications.spec.ts
```diff
@@ -8,11 +8,12 @@ import reducer from './notifications';
 
 describe('notifications reducer', () => {
     it('should return empty notification by default', () => {
-        assert.deepEqual([], reducer(undefined, { type: 'foo' }));
+        expect([]).toEqual(reducer(undefined, { type: 'foo' }));
```
fzaninotto, line 11: I suggest we swap the two operands here, I'm not a fan of this type of assertion where the result is put first.

packages/ra-core/src/reducer/admin/ui.spec.ts
```diff
@@ -13,52 +13,47 @@ describe('ui reducer', () => {
         viewVersion: 0,
     };
     it('should return hidden sidebar by default', () => {
-        assert.deepEqual(defaultState, reducer(undefined, { type: 'foo' }));
+        expect(defaultState).toEqual(reducer(undefined, { type: 'foo' }));
```
fzaninotto, line 16: same here, we should inverse operands

packages/ra-core/src/reducer/admin/notifications.spec.ts
WiXSL: Done!

packages/ra-core/src/reducer/admin/ui.spec.ts
WiXSL: Done!

packages/ra-core/src/dataProvider/fetch.spec.ts
WiXSL: Done!

fzaninotto:
Thanks!

PR #4948, opened 2020-06-17
De-googleify e-commerce demo

fzaninotto:
## Problem

With the default material-ui theme, the demo looks too much like a Google, and turns some people away. 

## Solution

Customize the theme for the demo.  

**before**

**after**


djhi:
Is the dark theme done too ?

fzaninotto:
The dark theme hasn't changed much but it's still usable.

fzaninotto:
As for using theme spacing, you're right in theory, but:

1. we currently don't give a way to customize the default theme spacing in the demo, so it's useless
2. There are MANY other components in this demo using numbers for spacing. This would be a day of work.

PR #4957, opened 2020-06-19
Fix : Resource name case in confirm for multi-language

### LINKED ISSUE #4783: lowercased delete names
**What you were expecting:**
I wanted to customize the text of a delete message in German. Therefore I added a translation for my resource and the bulk_delete_title in `translations.ts`:

```
...
"resources": {
  "Company": {
    "name": "Firma"
  }
}
...
"message": {
  "bulk_delete_title":
          "Lösche %{name} |||| Lösche %{smart_count} %{name} Elemente",
}
...
```

I assumed the message would be `Lösche Firma` on the delete confirmation overlay.

**What happened instead:**

The displayed message was `Lösche firma` with lowercased `F`, which is wrong in German. Could not find a way to "hack" this without changing the react admin codebase.

**Reason:**

The translate options of the `<Confirm/>` object in `<DeleteWithConfirmButton/>` uses `inflection.humanize` to lowercase the resource. This may make sense for English language, but not for others.

**Proposal:**

Make this humanize optional or use a different translation message ID and do not humanize.

**Environment**

* React-admin version: 3.5.0
* React version: 16.12.0
* Browser: any

JulienMattiussi:
Closes #4783 

## Issue

Resources names are always set to lower case in confirm messages and it doesn't match with all language synxates (ie : in German, names should be capitalized)

## Solution

Allow to force case in translation file

## Documentation

Capture d’écran de 2020-06-19 10-35-26

### DIFF
--- a/examples/demo/src/i18n/en.ts
+++ b/examples/demo/src/i18n/en.ts
@@ -40,6 +40,7 @@ const customEnglishMessages: TranslationMessages = {
     resources: {
         customers: {
             name: 'Customer |||| Customers',
+            forcedCaseName: 'Customer |||| Customers',
             fields: {
                 commands: 'Orders',
                 first_seen: 'First seen',
@@ -81,6 +82,7 @@ const customEnglishMessages: TranslationMessages = {
         },
         commands: {
             name: 'Order |||| Orders',
+            forcedCaseName: 'Order |||| Orders',
             amount: '1 order |||| %{smart_count} orders',
             title: 'Order %{reference}',
             fields: {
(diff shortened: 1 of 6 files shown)


djhi:
Nice. It's a new feature though and should probably be merged on `next`

examples/demo/src/i18n/en.ts
```diff
@@ -40,6 +40,7 @@ const customEnglishMessages: TranslationMessages = {
     resources: {
         customers: {
             name: 'Customer |||| Customers',
+            forcedCaseName: 'Customer |||| Customers',
```
fzaninotto, line 43: I don't understand why you add these translations, they should not be necessary.

PR #4963, opened 2020-06-20
Add types to layout/Title

iamstiil:


packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -27,6 +30,16 @@ const Title = ({ className, defaultTitle, locale, record, title, ...rest }) => {
     return createPortal(titleElement, container);
 };
 
+export declare type TitleProp<T> = string | React.ReactElement<T>;
```
fzaninotto, line 33: why do you need `declare`? exporting it should suffice.

fzaninotto:
Superseded by #5183

PR #4952, opened 2020-06-17
Simplify the development of custom List views

fzaninotto:
## Problem

1. For the core team, passing around list controller props is cumbersome, and makes TypeScript types bloated and cryptic. 
2. It's hard to know which variables are provided by react-admin in a list
3. As a consequence, it's hard to develop a custom List layout (e.g. with metrics at the top, or with both a left and a right sidebar)

## Solution

Use a context to store controller props instead of cloning children and passing them the prop.

Continue cloning the child and passing controller props for now, to keep BC, but ignore these props in child components. 

## Target

No need to "push" the props from parent to child, the children can "pull" the props from the context. 

```jsx
const PostList = props => (
    <BaseList {...props} perPage={10}>
        <div>
             <Filter />
        </div>
        <Grid container>
            <Grid item xs={8}>
                <SimpleList primaryText={record => record.title} />
            </Grid>
            <Grid item xs={4}>
                List instructions...
            </Grid>
        </Grid>
        <div>
            Post related links...
        </div>
    </BaseList>
);
```

## Tasks

- [x] Add  `ListContext` and `useListContext` to store and read list controller variables
- [x] Add `<ListBase>` to simplify the combination of `<ListContext>` and `useListController`
- [x] Migrate `<List>` and `<ListView>` to TypeScript
- [x] Make List create a `<ListContext.Provider>`
- [x] ~~Turn `<ListView>` into a top/right/bottom/left/center layout~~ (do the layout you want with `ListBase`)
- [x] use  `useListContext` instead of props in Pagination
- [x] use  `useListContext` instead of props in Filter
- [x] use  `useListContext` instead of props in Empty
- [x] use  `useListContext`instead of props in BulkActionsToolbar
- [x] use  `useListContext` instead of props in ListToolbar
- [x] use  `useListContext` instead of props in Datagrid
- [x] use  `useListContext` instead of props in SingleFieldList
- [x] use  `useListContext` instead of props in SimpleList
- [x] use  `useListContext` instead of props in ArrayField
- [x] use  `useListContext` instead of props in ReferenceManyField (and add filtering capabilities)
- [x] use  `useListContext` instead of props in ReferenceArrayField (and add sorting, pagination and filtering capabilities)
- [x] use  `useListContext` instead of props in ListGuesser
- [x] Add tests
- [x] Add documentation (basically rewrite the List documentation)

Slight BC breaks:
- useListController no longer returns the version
- useReferenceManyFieldController returns a basePath instead of a referenceBasePath

PR #4966, opened 2020-06-21
Add types to layout/Error

iamstiil:
Required #4963

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Error.tsx
+++ b/packages/ra-ui-materialui/src/layout/Error.tsx
@@ -1,5 +1,5 @@
-import * as React from 'react';
-import { Fragment } from 'react';
+import { useTranslate } from 'ra-core';
+import React, { Fragment, FC } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Button from '@material-ui/core/Button';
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -1,5 +1,5 @@
-import * as React from 'react';
```
fzaninotto, line 1: please revert this change.

packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -100,12 +100,37 @@ const Error = props => {
...
+export declare type ErrorClassKey =
+    | 'container'
+    | 'icon'
+    | 'toolbar'
+    | 'title'
+    | 'panel'
+    | 'panelDetails';
+
+export interface ErrorProps {
+    classes?: Record<ErrorClassKey, string>;
```
fzaninotto, line 112: we prefer using ClassesOverride (see my PR #4952)

packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -100,12 +100,37 @@ const Error = props => {
...
+    className?: string;
+    error: Error;
+    errorInfo?: Record<'componentStack', string>;
```
fzaninotto, line 115: where does that come from?

packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -53,7 +53,7 @@ function goBack() {
     window.history.go(-1);
 }
 
-const Error = props => {
+const ErrorComponent: FC<ErrorProps> = props => {
```
fzaninotto, line 56: no, we don't name all our components XXXComponent

packages/ra-ui-materialui/src/layout/Error.tsx
iamstiil: It comes from https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/layout/Layout.js#L134

packages/ra-ui-materialui/src/layout/Error.tsx
iamstiil: This is done due to the fact that the JS `Error` class is overridden in the context, but is needed as a type for the props.

PR #4967, opened 2020-06-21
Add types to layout/CardContentInner

iamstiil:


packages/ra-ui-materialui/src/layout/CardContentInner.tsx
```diff
@@ -1,8 +1,9 @@
-import * as React from 'react';
+import React from 'react';
```
fzaninotto, line 1: Please revert this one, it's intended, see #4857 for details.

packages/ra-ui-materialui/src/layout/CardContentInner.tsx
```diff
@@ -40,10 +41,18 @@ const CardContentInner = props => {
     );
 };
 
+export declare type CardContentInnerClassKey = 'root';
+
+export interface CardContentInnerProps {
+    className?: string;
+    classes?: ClassNameMap<CardContentInnerClassKey>;
```
fzaninotto, line 48: We prefer another approach to typing classes, see https://github.com/mui-org/material-ui/issues/17973#issuecomment-639281445

fzaninotto:
Wrong rebase - pleas try again.

Issue #4027, opened 2019-11-22
Who is using react-admin?


fzaninotto:
We have the impression that react-admin is widely used in production. We don't have any tracker to tell us who - we don't like trackers. But the popularity of the repository (10,000 stars), the activity of the issue / PR tracker (about 4 issues and 4 PRs per day), the number of npm downloads (20,000 per week), and the companies who contact us directly (about 2 every week) let us think that you guys are not just reading the documentation because it's funny.

We'd love to advertise some well-known company names in [the react-admin landing page](https://marmelab.com/react-admin/). That'd be a good sign for other enterprise customers who benchmark libraries, and would probably accelerate the growth of the react-admin user base. 

This issue is more of a poll. If you're using react-admin in production, and you accept that we name your company as a react-admin user in the public website, **please add a comment with your company name**, and a brief description of the app you built. If you can join a screenshot or a public URL, we'd be delighted!

If you're the CTO of such a well-known company, would you accept that we publish a citation from you explaining how react-admin helped you solve the needs you had? If so, please post a comment with your name, title, company, and avatar.  

Thanks in advance for your support!

anthonycmain:
Hi Francois, 

We've embedded React Admin in the backend solution of the 4th iteration app framework. Its currently included in 12 projects in various states of deployment (of which most will be going live very soon).

We have a high profile project currently in progress which will feature it that we intend to use as a priority case study for our platform as a whole which I'd be happy to discuss.

P.S We are about to undertake a major refactor in the next few weeks and are hoping v3 will be ready in time :D (no pressure!!)

Anthony
https://thedistance.co.uk

fzaninotto:
@anthonycmain Thanks! Sounds like you're a pretty advanced user already. Could you please share the details here, as our intention is to use them publicly?

As for v3, it was released a few days ago.

loicgymlib:
Hi François,

We've been using React-Admin since August and are pretty happy with it mainly for the ease to set it up and GraphQL out of the box.

We included React-Admin in two major projets in Gymlib:
- Backoffice for internal management (Version 2.9.1)
- Partnership Dashboard (Version 2.9.1)

Additional projects may be added soon as version 3 looks pretty nice.
We are working on upgrade to version 3 which looks very promising ! (Nice job)

I would be happy to provide screenshots and help you promote React-Admin of course.

Loïc Monthorin (CTO)
https://gymlib.com

fzaninotto:
[...]

PR #4979, opened 2020-06-25
Data-driven

christiaanwesterbeek:
I would use react-admin not just for B2B (as your Github repo description says) or admin applications. But for any data-driven web application that I would start or plan to build today.


fzaninotto:
I'm not sure this is better. What's a non data-driven application?

christiaanwesterbeek:
Yeah, I'm not sure either. I read "B2B" on the github repo description and "admin" in readme. I've built a react-admin based application that looks nothing like admin and nothing like react-admin, and is used by thousands of consumers daily. I was looking for something that they have in common. I couldn't come up with anything else than data-driven.

fzaninotto:
Well, as we already say "admin" in the library title, we can afford using another buzzword in the description. Thanks for the PR.

On another subject, are you authorized to mention the app you've built with react-admin in #4027?

christiaanwesterbeek:
I will ask if it's OK if I write something in #4027 

Another idea. Maybe you could go for data-centric, because indeed what isn't data-driven these days, but react-admin does focus and shine on working with data.

PR #4983, opened 2020-06-26
[RFR] Avoid import errors when tree-shaking @material-ui

rassie:
Material-UI documents some tree-shaking performance optimizations at https://material-ui.com/guides/minimizing-bundle-size/. Any downstream project using both Material-UI and react-admin implementing these recommendations will encounter errors like

`Cannot find module: '@material-ui/core/esm/createMuiTheme'. Make sure this package is installed.`

This is apparently due to the way Material-UI re-exports styling exports via `@material-ui/core`. The official recommendation is to import via `@material-ui/core/styles`, which `react-admin` is not doing throughout. This PR is making sure `makeStyles` and `createMuiTheme` are imported correctly. This is obviously incomplete: any re-exported import from [@material-ui/core/styles](https://github.com/mui-org/material-ui/tree/next/packages/material-ui/src/styles) would probably be a potential problem, however, just including `makeStyles` and `createMuiTheme` has been sufficient to make a minimal example project with the tree-shaking optimization include compilable. 

Also included: ESLint configuration forbidding these two imports.


packages/ra-ui-materialui/src/button/Button.spec.tsx
```diff
@@ -2,7 +2,8 @@ import { render, cleanup } from '@testing-library/react';
 import * as React from 'react';
 import expect from 'expect';
 import { TestContext } from 'ra-core';
-import { createMuiTheme, ThemeProvider } from '@material-ui/core';
+import { ThemeProvider } from '@material-ui/core';
```
djhi, line 5: Shouldn't this be imported from `@material-ui/core/styles` as well? Same for all other `ThemeProvider` imports

rassie:
> I'm not sure updating the test imports will bring any benefit

Technically, you are right, however I think making an exception in the ESLint rule will probably be more complicated than using the right imports in the tests.

packages/ra-ui-materialui/src/button/Button.spec.tsx
rassie: That's an interesting case: `ThemeProvider` normally lives in `@material-ui/styles`, which means it's re-exported from `@material-ui/core` and is in theory subject to the same problem we are talking about here. However, I've failed to confirm this so far, still looking into the source code. It doesn't seem to hurt compilation anyway, at least not for a minimal project, but I'll keep digging.

rassie: OK, found it. [`@material-ui/core/styles`](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/styles/index.d.ts#L26) exports a couple of exports from `@material-ui/styles`, while [`@material-ui/core`](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/index.d.ts#L57) re-exports everything from `@material-ui/core/styles`. Now to find out why this works and e.g. `makeStyles` doesn't.

rassie: I don't think I get it fully, but it seems that `ThemeProvider` is resolved correctly with the `babel` rules because at the end of the import chain it's not a part of `@material-ui/core` but of `@material-ui/styles`, which is not getting transformed by `babel`. In theory, everything under `core/styles` should probably be part of the rule:

```
colorManipulator.js
createBreakpoints.js
createMixins.js
createMuiStrictModeTheme.js
createMuiTheme.js
createPalette.js
createSpacing.js
createStyles.js
createTypography.js
cssUtils.js
defaultTheme.js
makeStyles.js
MuiThemeProvider.js
responsiveFontSizes.js
shadows.js
shape.js
styled.js
transitions.js
useTheme.js
withStyles.js
withTheme.js
zIndex.js
```

But the question is probably how to keep up with changes. I think `createMuiTheme` and `makeStyles` could be enough for starters and we'll have to see what happens with MUI v5, this problem might disappear in the process.

djhi:
Sorry we merged something else. Do you mind rebasing?

rassie:
@djhi done, thanks!

rassie:
@djhi sorry for nagging, but it seems this PR is the last one for 3.6.3, is there anything blocking this from being merged, can I help with that in any way?

fzaninotto:
Sorry for the delay. 

There were significant changes in the `next` branch that we don't want to rebase because of this PR. So we will publish 3.6.3 without these changes, then merge `next` into `master`, then see if this PR can be merged without conflicts.

rassie:
Thanks for the update, I'll keep an eye on releases and will rebase when 3.6.3 is done.

rassie:
Or should I just rebase on top of `next`?

rassie:
I've rebase the PR on `master` after the 3.7.0 release. Finger crossed the build succeeds :)

examples/simple/src/tags/TagList.js
```diff
@@ -10,6 +10,7 @@ import {
     Collapse,
     Card,
 } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
```
fzaninotto, line 13: useless here

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -1,14 +1,9 @@
...
 import { useCallback, FunctionComponent, ReactElement } from 'react';
 import PropTypes from 'prop-types';
-import {
-    ListItemIcon,
-    MenuItem,
-    useMediaQuery,
-    makeStyles,
-} from '@material-ui/core';
+import { ListItemIcon, MenuItem, useMediaQuery } from '@material-ui/core';
 import { MenuItemProps } from '@material-ui/core/MenuItem';
```
fzaninotto, line 5: Could you move this one into the grouped import above?

fzaninotto: Sorry, it's a Prop, disregard this comment.

examples/simple/src/tags/TagList.js
rassie: Removed, no idea how it got there.

fzaninotto:
Thanks!

Issue #4913, opened 2020-06-08
Add ability to make the <SaveButton> not disabled by default


fzaninotto:
In `<Edit>` and `<Create>` forms, the `<SaveButton>` is rendered disabled until the user enters a change (i.e. as long as the form is still in `pristine` state).

In some cases (e.g. a form wizard), that's inconvenient. 

Currently, to bypass the disabled default, a developer needs to write their own `<SaveButton>`, create a custom Toolbar, and pass that to the Form. That's a lot of work to get what used to be the default behavior. 

we need to provide a simple way to disable the `pristine = disabled` feature.

UltimateForm:
hello @fzaninotto do you guys intend to work on this any time soon? This is kinda of critical to a project I'm working on since I'm using wizards and dialogs inside the form.

fzaninotto:
Fixed by #5002

PR #5002, opened 2020-07-03
Add ability to make the <SaveButton> not disabled by default

Luwangel:
Implements https://github.com/marmelab/react-admin/issues/4913

To build this feature I allow the user to pass a new prop named `alwaysEnableSaveButton` to the `<Toolbar>` which forces enabling ou disabling the `<SaveButton>`. If the user doesn't specify this prop, we keep the classic behaviour by disabling the button until the form is dirty. 

## Todo

- [x] Allow the `<SaveButton>` to be disabled by passing the `disabled` prop
- [x] Create a new prop called `alwaysEnableSaveButton` in the `<Toolbar>` and implements its mechanism
- [x] Keep the condition _pristine equals to disabled_ only if the `alwaysEnableSaveButton` prop has not been explicitly passed to the `<Toolbar>` but move the logic to the `<Toolbar>` instead of the `<SaveButton>` (possible **breaking change here**)
- [x] Fix and improve the documentation
- [x] Fix tests

## Example

How to force enabling the `<SaveButton>` without rewriting the whole `<Toolbar>`:

``` js
import {
    Create,
    SimpleForm,
    Toolbar,
} from 'react-admin';
import PostReferenceInput from './PostReferenceInput';

export const CommentCreate = props => (
    <Create {...props}>
        <SimpleForm redirect={false} toolbar={<Toolbar alwaysEnableSaveButton />}>
            // Form goes here
        </SimpleForm>
    </Create>
);
```

If you don't pass the disabled prop, the behaviour "pristine equals to disabled" will be applied as before.

## Screenshots

The following screenshots are outdated because the prop has changed. You need to use `alwaysEnableSaveButton` instead of `disabled` and invert the condition.

**Force enabling the `<SaveButton>`**

Sélection_006

**Force disabling the `<SaveButton>`**

Sélection_007

**Keep the pristine equals to disabled mechanism**

Sélection_008

### DIFF
--- a/packages/ra-ui-materialui/src/form/Toolbar.js
+++ b/packages/ra-ui-materialui/src/form/Toolbar.js
@@ -89,7 +89,7 @@ const Toolbar = props => {
                             handleSubmitWithRedirect={
                                 handleSubmitWithRedirect || handleSubmit
                             }
-                            disabled={disabled}
+                            disabled={valueOrDefault(disabled, pristine)}
                             invalid={invalid}
                             redirect={redirect}
                             saving={saving}
--- a/packages/ra-ui-materialui/src/form/Toolbar.js
+++ b/packages/ra-ui-materialui/src/form/Toolbar.js
@@ -47,6 +47,43 @@ const useStyles = makeStyles(
 const valueOrDefault = (value, defaultValue) =>
     typeof value === 'undefined' ? defaultValue : value;
 
+/**
+ * The Toolbar displayed at the bottom of forms.
+ *
+ * @example Never disable the <SaveButton />
+ *
+ * import * as React from 'react';
+ * import {
+ *     Create,
+ *     DateInput,
+ *     TextInput,
+ *     SimpleForm,
+ *     Toolbar,
+ *     required,
+ * } from 'react-admin';
+ *
+ * const now = new Date();
+ * const defaultSort = { field: 'title', order: 'ASC' };
+ *
+ * const CommentCreate = props => (
+ *     <Create {...props}>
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/form/Toolbar.js
```diff
@@ -47,8 +47,46 @@ const useStyles = makeStyles(
...
+ *     Toolbar,
+ *     required,
+ * } from 'react-admin';
+ *
+ * const now = new Date();
+ * const defaultSort = { field: 'title', order: 'ASC' };
+ *
+ * const CommentCreate = props => (
+ *     <Create {...props}>
+ *         <SimpleForm redirect={false} toolbar={<Toolbar disabled={false} />}>
```
JulienMattiussi, line 70: wrong props

Luwangel:
> Why is this a breaking change?

People who use directly the `<SaveButton>` cannot rely on the `pristine` prop anymore to disable the button.  They have to use the `disabled` property instead. It's not a breaking change if they use the `<SaveButton>` inside the `<Toolbar>` or the `<SimpleForm>`.

So most of the time it's not a breaking change. But for some isolated cases, it could be one minor breaking change. It's better to mention it just in case.

florianajir:
> People who use directly the `<SaveButton>` cannot rely on the `pristine` prop anymore to disable the button. They have to use the `disabled` property instead. It's not a breaking change if they use the `<SaveButton>` inside the `<Toolbar>` or the `<SimpleForm>`.

So if I had in my code this 
```
<SaveButton
    pristine={pristine}
    disabled={loading} />
```
I have to change to 
```
<SaveButton
    disabled={loading || pristine} />
```
Am I right ?

Issue #4629, opened 2020-04-02
Documentation about setting resources dynamically is outdated


WiXSL:
Current: 
https://marmelab.com/react-admin/doc/3.3/Admin.html#declaring-resources-at-runtime

New:
#3907

I'll see if I find some time to do this, no warranties.

Kmaschta:
That would be awesome :)

fzaninotto:
I don't understand, we don't enforce the usage of `<AdminUI>`, and the previous way still works... Can you be more specific?

WiXSL:
Of course, I don't say to remove the current way to do it but to add documentation about your solution here #3907 for cases where for example (quoting):

`... even if <Admin> accepts a function as child, this function can't execute hooks. So it's impossible, for instance, to have a dynamic list of resources based on a call to the dataProvider (since the dataProvider is only defined after the Admin component renders).
Many people use the Custom App approach just because of that, and they have to rewrite too much code.`

fzaninotto:
Understood. A PR is welcome!

PR #5004, opened 2020-07-03
Document dynamic resources

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/4629
Follows https://github.com/marmelab/react-admin/pull/3907

## Todo

- [x] Write a section in the documentation about the <AdminContext> and the <AdminUI> components

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -29,6 +29,8 @@ Here are all the props accepted by the component:
 
 - [The `<Admin>` Component](#the-admin-component)
   - [`dataProvider`](#dataprovider)
+  - [`authProvider`](#authprovider)
+  - [`i18nProvider`](#i18nprovider)
   - [`title`](#title)
   - [`dashboard`](#dashboard)
   - [`catchAll`](#catchall)
@@ -38,12 +40,10 @@ Here are all the props accepted by the component:
   - [`customReducers`](#customreducers)
   - [`customSagas`](#customsagas)
   - [`customRoutes`](#customroutes)
-  - [`authProvider`](#authprovider)
   - [`loginPage`](#loginpage)
   - [`logoutButton`](#logoutbutton)
   - [`initialState`](#initialstate)
   - [`history`](#history)
-  - [Internationalization](#internationalization)
   - [Declaring resources at runtime](#declaring-resources-at-runtime)
   - [Using react-admin without `<Admin>` and `<Resource>`](#using-react-admin-without-admin-and-resource)
 
@@ -66,6 +66,32 @@ const dataProvider = {
 ```
 
 The `dataProvider` is also the ideal place to add custom HTTP headers, authentication, etc. The [Data Providers Chapter](./DataProviders.md) of the documentation lists available data providers, and explains how to build your own.
+
+## `authProvider`
+
+The `authProvider` prop expect an object with 5 methods, each returning a Promise, to control the authentication strategy:
+
+```jsx
+const authProvider = {
+    login: params => Promise.resolve(),
+    logout: params => Promise.resolve(),
+    checkAuth: params => Promise.resolve(),
+    checkError: error => Promise.resolve(),
+    getPermissions: params => Promise.resolve(),
+};
+
+const App = () => (
+    <Admin authProvider={authProvider} dataProvider={simpleRestProvider('http://path.to.my.api')}>
+        ...
+    </Admin>
+);
+```
+
+The [Authentication documentation](./Authentication.md) explains how to implement these functions in detail.
+
+## `i18nProvider`
+
+The `i18nProvider` props let you translate the GUI. The [Translation Documentation](./Translation.md) details this process.
 
 ## `title`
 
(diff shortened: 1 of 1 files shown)


docs/Admin.md
```diff
@@ -29,8 +29,6 @@ Here are all the props accepted by the component:
 
 - [The `<Admin>` Component](#the-admin-component)
   - [`dataProvider`](#dataprovider)
-  - [`authProvider`](#authprovider)
```
fzaninotto, line 32: As these are the most important props, I prefer that they appear first.

docs/Admin.md
```diff
@@ -29,8 +29,6 @@ Here are all the props accepted by the component:
...
-  - [`i18nProvider`](#internationalization)
```
fzaninotto, line 33: I think you'd better remove the other link to internationalization

docs/Admin.md
```diff
@@ -40,12 +38,15 @@ Here are all the props accepted by the component:
...
   - [`customSagas`](#customsagas)
   - [`customRoutes`](#customroutes)
+  - [`authProvider`](#authprovider)
   - [`loginPage`](#loginpage)
   - [`logoutButton`](#logoutbutton)
   - [`initialState`](#initialstate)
   - [`history`](#history)
   - [Internationalization](#internationalization)
   - [Declaring resources at runtime](#declaring-resources-at-runtime)
+    - [Use a Function As its Child](#use-a-function-as-its-child)
```
fzaninotto, line 48: I don't think we should link to level 3 headers in this list

docs/Admin.md
```diff
@@ -559,13 +559,16 @@ The `i18nProvider` props let you translate the GUI. The [Translation Documentati
 
 ## Declaring resources at runtime
 
-You might want to dynamically define the resources when the app starts. The `<Admin>` component accepts a function as its child and this function can return a Promise. If you also defined an `authProvider`, the child function will receive the result of a call to `authProvider.getPermissions()` (you can read more about this in the [Authorization](./Authorization.md) chapter).
+You might want to dynamically define the resources when the app starts.
+
+### Use a Function As its Child
```
fzaninotto, line 564:
```suggestion
### Using a Function As `<Admin>` Child
```

docs/Admin.md
```diff
@@ -559,13 +559,16 @@ The `i18nProvider` props let you translate the GUI. The [Translation Documentati
 
 ## Declaring resources at runtime
 
-You might want to dynamically define the resources when the app starts. The `<Admin>` component accepts a function as its child and this function can return a Promise. If you also defined an `authProvider`, the child function will receive the result of a call to `authProvider.getPermissions()` (you can read more about this in the [Authorization](./Authorization.md) chapter).
+You might want to dynamically define the resources when the app starts.
```
fzaninotto, line 562: This sentence alone with the section title afterwards feels odd.

docs/Admin.md
```diff
@@ -595,6 +598,44 @@ const App = () => (
 );
 ```
 
+### Unplug the <Admin> using the `<AdminContext>` and the `<AdminUI>`
+
+Setting Resources dynamically is still very cumbersome: even if `<Admin>` accepts [a function as child](#declaring-resources-at-runtime), this function can't execute hooks.
```
fzaninotto, line 603:
```suggestion
Setting Resources dynamically using the children-as-function syntax may not be enough in all cases, because this function can't execute hooks.
```

docs/Admin.md
```diff
@@ -595,6 +598,44 @@ const App = () => (
...
+
+So it's impossible, for instance, to have a dynamic list of resources based on a call to the dataProvider (since the dataProvider is only defined after the Admin component renders).
```
fzaninotto, line 605:
```suggestion
So it's impossible, for instance, to have a dynamic list of resources based on a call to the `dataProvider` (since the `dataProvider` is only defined after the `<Admin>` component renders).
```

docs/Admin.md
```diff
@@ -595,6 +598,44 @@ const App = () => (
...
+
+To do so, you have to build your own <Admin> component using both the <AdminContext> and the <AdminUI> (as we do internally). It's straightforward.
```
fzaninotto, line 607: add backticks around components

fzaninotto, line 607:
```suggestion
To overcome this limitation, you can build your own `<Admin>` component using two lower-level components: `<AdminContext>` (responsible for putting the providers in contexts) and `<AdminUI>` (responsible for displaying the UI). Here is an example:
```

docs/Admin.md
```diff
@@ -595,6 +598,44 @@ const App = () => (
...
+        </AdminContext>
+    );
+}
+
+function AsyncResources() {
+    const [resources, setResources] = useState([]);
+    const dataProvider = useDataProvider();
+
+    useEffect(() => {
+        dataProvider.introspect().then(r => setResources(r));
```
fzaninotto, line 626: introspect isn't a legal dataProvider verb, so you'll have to be more explicit here

PR #5006, opened 2020-07-03
DatagridRow with rowClick set to "toggleSelection" and selectable set to "false" got selected.

WiXSL:
Now I'm thinking this should be merge into master.


fzaninotto:
Thanks!

fzaninotto:
It won't matter much: both 3.6.3 and 3.7 will probably be released within a day.

PR #5014, opened 2020-07-07
Update CreateButton on 'to' prop change

rkfg:
I need the `CreateButton` to reflect the filter values to prefill the new item form. The use case is like this:
- filter the data by a couple of fields
- no expected data found
- press Create
- expect those two fields filled with the filter values

It worked almost fine except I needed to refresh the page after setting the filter to make the `to` property be in sync with the filter (. After some research I found out that this button is exported via `memo` and it's not updated if the `to` prop is changed. I replaced the `TopToolbar` in `List` and the button is declared like this:
```js
    <CreateButton to={{
      pathname: '/records/create',
      search: `?source=${JSON.stringify({ organization_id: filterValues.organization_id, department_id: filterValues.department_id })}`,
    }} />
```
When I used a regular button it worked just fine. Not all buttons in `ra-ui-materialui` are wrapped into `memo` (the other two I found are `ShowButton` and `CloneButton`), I'm not sure if this is an important optimization or if it has some deeper meaning but apparently it might cause this subtle bugs.

### DIFF
--- a/packages/ra-ui-materialui/src/button/CreateButton.tsx
+++ b/packages/ra-ui-materialui/src/button/CreateButton.tsx
@@ -89,6 +89,7 @@ export default memo(CreateButton, (prevProps, nextProps) => {
     return (
         prevProps.basePath === nextProps.basePath &&
         prevProps.label === nextProps.label &&
-        prevProps.translate === nextProps.translate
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

Issue #4992, opened 2020-06-29
ReferenceInput does not pass loading prop to child component


sebashwa:
**What you were expecting:**
According to the [documentation](https://marmelab.com/react-admin/Inputs.html) `ReferenceInput` should pass down a `loading` prop. I'd like to wrap `AutocompleteInput` so it shows a loading indicator whilst loading the references.

**What happened instead:**
I can not find a loading prop passed down to the child component. I think it is extracted from the `ReferenceInput` in order to figure out wether a `<LinearProgress />` or the `<ReferenceInputView />` should be shown at all.

**Steps to reproduce:**
Wrap a component in a `ReferenceInput`

**Related code:**
I created a minimal version to reproduce the behaviour logging the props of the user autocomplete in `PostEdit`:

https://codesandbox.io/s/exciting-wilbur-jkwdo?file=/src/posts/PostEdit.js

**Environment**

* React-admin version: 3.6.2
* React version: ^16.9.0

djhi:
That's because the `<ReferenceInput />` already show a `<LinearProgress />` when loading. In order to customize that, you'll have to write your own `<ReferenceInput />` component using the underlying hook: `useReferenceInputController`

sebashwa:
Thanks for your reply. What I actually want to do is show a spinner in case references are fetched (when someone types something inside the input). I think this should be possible with the (according to the documentation) passed loading prop:

> The child component receives the following props from <ReferenceInput>:
>  loading: whether the request for possible values is loading or not

But this one does not seem to be passed along.

djhi:
Indeed. Thanks for reporting this

fzaninotto:
The `loading` prop computed by `useReferenceInputController` combines the loading state of the matching references and the loading state of the reference record. It's probably not what you want. 

In the short term, the solution to fix this issue is probably to *remove* the mention of the loading prop in the doc. 

In the longer term, we need to have `<ReferenceInput>` implement a `<ListContext.Provider>` - and this will give access to `loading` to its children. 

It's a new feature though, so this will come in a minor branch.

PR #5016, opened 2020-07-07
[RFR] Refactor <ReferenceInput>

### LINKED ISSUE #4531: Migrate useReferenceInputController to use useDataProvider instead of sagas
The requests made by the `<ReferenceInput>` component aren't cached because their controller still uses the `CrudGetMatchingAccumulate` action creator. That means that the `dataProvider.getList()` it makes to fetch the possible references is done by the old fetch saga (called by the `accumulate` saga) instead of the `useDataProvider` hook. As the cache is only implemented in the new hook, `<ReferenceInput>` doesn't use the cache.

We need to migrate the `useReferenceInputController` code so that it doesn't use the accumulate saga anymore.

Seen in #4386

### LINKED ISSUE #4601: Empty view shows after coming back from detail view
1. Open the url `/#/posts/1` on the simple example on  a new window
2. Click on the Save button
3. The user is redirected to the List view, which shows the Empty component... but it shouldn't.

Tested in v3.3.3

fzaninotto:
## Problem

The `<ReferenceInput>` code lags behind the rest of the code, it is hard to follow, and it only allows limited functionality in children. 

## Solution

- [x] Have `<ReferenceInput>` implement a `<ListContext>` (refs #4992)
- [x] Implement deduplication logic from accumulate saga into `useQueryWithStore`
- [x] Make `useReferenceInput` use `useDataProvider` instead of sagas (Closes #4531)
- [x] Fix List view shows Empty component when coming from the Edit view (closes #4601)
- [ ] ~~Move loading logic from `<ReferenceInput>` to child (to be consistent with `<List>`)~~ (even ReferenceField doesn't do that yet)


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -77,61 +60,195 @@ interface Option {
...
+        },
+        dataStatus: {
+            error: dataStatus.error,
+            loading: dataStatus.waiting,
+            warning: dataStatus.warning,
+        },
+        choices: Object.keys(finalData).map(id => finalData[id]),
+        // kept for backwards compatibility
+        // @deprecated to be removed in 4.0
         error: dataStatus.error,
```
Luwangel, line 200: Only the `error` is deprecated or it's also the following parameters?

packages/ra-core/src/controller/useListController.ts
```diff
@@ -177,7 +177,7 @@ const useListController = <RecordType extends Record = Record>(
         get(state.admin.resources, [resource, 'list', 'ids'], [])
     );
     const defaultTotal = useSelector((state: ReduxState): number =>
-        get(state.admin.resources, [resource, 'list', 'total'], 0)
+        get(state.admin.resources, [resource, 'list', 'total'])
```
Luwangel, line 180: If you don't pass 0, the result could be "undefined". It could lead to errors if the sub-components don't manage this use case.

Luwangel (changes requested):
Top! I only have a few remarks and questions.

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
```diff
@@ -75,7 +75,7 @@ const Pagination: FC<PaginationProps> = props => {
     );
 
     // Avoid rendering TablePagination if "page" value is invalid
-    if (total === 0 || page < 1 || page > totalPages) {
+    if (total === null || total === 0 || page < 1 || page > totalPages) {
```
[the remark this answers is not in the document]
fzaninotto: fortunately this is JS, where `null/25=0` with no warning.

packages/ra-core/src/controller/input/useReferenceInputController.ts
fzaninotto: it's the following parameters, but we'll need the RecordContext to truly deprecate them

packages/ra-core/src/controller/useListController.ts
fzaninotto: Normally they all do. That's why I fixed the Pagination component. 

As for user land components, yes this is a BC break, albeit minor in my opinion. And it's necessary to fix #4601

fzaninotto:
Switching to RFR

fzaninotto:
I'm getting a wrning when updating a Comment in the simple example, back to WIP

fzaninotto:
And back to RFR

packages/ra-core/src/controller/input/index.ts
```diff
@@ -1,6 +1,8 @@
 import ReferenceArrayInputController from './ReferenceArrayInputController';
 import ReferenceInputController from './ReferenceInputController';
-import useReferenceInputController from './useReferenceInputController';
+import useReferenceInputController, {
```
djhi, line 3: Probably a good opportunity to use only named exports

djhi (approved):

djhi (approved):

PR #5024, opened 2020-07-08
Add date input validators

roseline124:
## Feature Background
We can use custom DateInput with material-ui's DatePicker supporting this feature.
But, it causes additional codes and package and there might be people (like me) who prefers DateInput to DatePicker because of some reasons(UI and format bug, etc). so I add this date validators.

## Changes
- Add date, datetime validators `minDate`, `maxDate`
- Add minDate, maxDate to `validation.spec.ts`
- Add translated message (en, fr)
- Add example in `CreateEdit.md` document

## Check Lists
- [x] pass unit test
- [x] pass lint


packages/ra-language-french/src/index.ts
```diff
@@ -138,6 +138,8 @@ const frenchMessages: TranslationMessages = {
             maxLength: 'Maximum %{max} caractères',
             minValue: 'Minimum %{min}',
             maxValue: 'Maximum %{max}',
+            minDate: 'Minimum %{min} date',
```
fzaninotto, line 141:
```suggestion
            minDate: 'Au plus tôt %{min}',
```

packages/ra-language-french/src/index.ts
```diff
@@ -138,6 +138,8 @@ const frenchMessages: TranslationMessages = {
...
+            maxDate: 'Maximum %{max} date',
```
fzaninotto, line 142:
```suggestion
            maxDate: 'Au plus tard %{max}',
```

packages/ra-core/src/form/validate.ts
```diff
@@ -1,4 +1,5 @@
 import lodashMemoize from 'lodash/memoize';
+import format from 'date-fns/format';
```
fzaninotto, line 2: Browsers have an API for formatting dates (Intl), no need for an external library

packages/ra-language-french/src/index.ts
roseline124: Thanks!

packages/ra-core/src/form/validate.ts
```diff
@@ -198,15 +198,18 @@ export const minDate = memoize(
     (min, message = 'ra.validation.minDate', formatString = 'YYYY-MM-DD') => (
         value,
         values
-    ) =>
-        !isEmpty(value) && new Date(value) < new Date(min)
+    ) => {
+        const inputValue = new Date(value);
+        inputValue.setSeconds(0);
```
fzaninotto, line 203: What's the rationale of this?

roseline124:
> I'm not sure I understand the difference with `minValue` and `maxValue`. Is it only the formatting of the `min` and `max` values in the error message?

I admit `minValue` and `maxValue` can cover this feature. But, I think `minDate`, `maxDate` validators are more readable and the error message is also important, too.

1. comparing code using minDate to one not using

I I added `nowDate.setSeconds(0)` to my feature.
Because DateTimeInput can't choose seconds, the seconds is set as `0` when selecting date in DateTimeInput.
So If I had set the baseDate as `Fri Jul 10 2020 17:47:57` and selected `Fri Jul 10 2020 17:47`, the validator will throw error even If these are the same date.

```typescript
<DateTimeInput
  source="advertisementEndAt"
  validate={[minDate(new Date())]}
/>
```

```typescript
<DateTimeInput
  source="advertisementEndAt"
  validate={[
    (value, values) => {
      const nowDate = new Date()
      nowDate.setSeconds(0)
      return minValue(Number(new Date()))(Number(value), values)
    },
  ]}
/>
```

2. Error message
- minDate: `Must be at least 2020-07-10`
- minValue: `Must be at least 1594371535226`

I'm okay if you close this PR.
Thank you for reviewing my code

packages/ra-core/src/form/validate.ts
roseline124: It's written below!

packages/ra-core/src/form/validate.ts
roseline124: > Hi, these validators look great 👍
> 
> Thanks for contributing 😊.
> 
> I've one remark: Can we can consider using another name for these validators?
> 
> * Minimal date means the input should contain a date after the specified one
> * Maximal date means the input should contain a date before the specified one
> 
> So maybe you can name these methods `afterDate` and `beforeDate`. What do you think?



> Hi, these validators look great 👍
> 
> Thanks for contributing 😊.
> 
> I've one remark: Can we can consider using another name for these validators?
> 
> * Minimal date means the input should contain a date after the specified one
> * Maximal date means the input should contain a date before the specified one
> 
> So maybe you can name these methods `afterDate` and `beforeDate`. What do you think?

I read your comment. Thank you for reviewing my code!
I implemented code to allow the base date to be included.
You means that it'll be better if the code excludes the base date? or just changes the validator name?

roseline124: > Browsers have an API for formatting dates (Intl), no need for an external library

Thank you! 
I used `new Intl.DateTimeFormat().format(date)` also changed format options (DateTimeFormatOptions).
The option can be a little long like this. I don't know this is good. 
```
const options = {
    year: 'numeric',
    month: 'short',
    day: 'numeric',
    hour: 'numeric',
    minute: 'numeric',
};
```
It needs to know `DateTimeFormatOptions` and  readability doesn't look good. 
What do you think?

docs/CreateEdit.md
```diff
@@ -946,6 +949,7 @@ const validateEmail = email();
 const validateAge = [number(), minValue(18)];
 const validateZipCode = regex(/^\d{5}$/, 'Must be a valid Zip Code');
 const validateSex = choices(['m', 'f'], 'Must be Male or Female');
+const validateDate = [minDate(new Date())];
```
JulienMattiussi, line 952: Should use maxDate here to make this exemple more consistant with a Birthdate

docs/CreateEdit.md
```diff
@@ -919,6 +919,8 @@ Alternatively, you can specify a `validate` prop directly in `<Input>` component
 * `required(message)` if the field is mandatory,
 * `minValue(min, message)` to specify a minimum value for integers,
 * `maxValue(max, message)` to specify a maximum value for integers,
+* `minDate(min, message)` to specify a minimum value for date,
```
fzaninotto, line 922: this doc is now incomplete - you must add the formatting options parameter

docs/CreateEdit.md
fzaninotto: you should also add formatting options to illustrate usage

packages/ra-core/src/form/validate.ts
```diff
@@ -179,6 +179,81 @@ export const maxValue = memoize(
...
+ */
+export const minDate = memoize(
+    (min, message = 'ra.validation.minDate', formatOptions = {}) => (
+        value,
+        values
+    ) => {
+        const input = new Date(value);
+        input.setSeconds(0);
+
+        return !isEmpty(value) && input < new Date(min)
```
fzaninotto, line 204: `min` should be a `Date`, why do you create a new Date object

packages/ra-core/src/form/validate.ts
```diff
@@ -179,6 +179,81 @@ export const maxValue = memoize(
...
+ * @param {string|function} message
+ * @param {object} formatOptions
+ *
+ * @example
+ *
+ * const fooValidators = [minDate(new Date(), 'Should be after today')];
+ * <DateInput name="foo" validate={fooValidators} />
+ */
+export const minDate = memoize(
+    (min, message = 'ra.validation.minDate', formatOptions = {}) => (
```
fzaninotto, line 197: please type the args using TypeScript

packages/ra-core/src/form/validate.ts
```diff
@@ -179,6 +179,81 @@ export const maxValue = memoize(
...
+ * <DateInput name="foo" validate={fooValidators} />
+ */
+export const maxDate = memoize(
+    (max, message = 'ra.validation.maxDate', formatOptions = {}) => (
+        value,
+        values
+    ) => {
+        const input = new Date(value);
+        input.setSeconds(0);
+        return !isEmpty(value) && input > new Date(max)
```
fzaninotto, line 241: same, `max` should already be a `Date`

PR #5031, opened 2020-07-09
[RFR] Split ListContext for better performance

fzaninotto:
## Problem

Since 3.7.0, some of the performance optimizations of the List components were lost, because they all depend on the ListContext, which updates each time the user changes one of the list params (sort, filter, pagination).

## Solution

Unfortunately, React doesn't allow to bail out of a re-render triggered by a context update. They might implement it in the future (see https://github.com/reactjs/rfcs/pull/119), but it's not possible yet. 

So in addition to `ListContext`, we should add specialized contexts for sort, filter, and pagination.

```jsx
const ListContextProvider = ({ value, children }) => (
    <ListContext.Provider value={value}>
        <ListFilterContext.Provider value={usePickFilterContext(value)}>
            <ListSortContext.Provider value={usePickSortContext(value)}>
                <ListPaginationContext.Provider
                    value={usePickPaginationContext(value)}
                >
                    {children}
                </ListPaginationContext.Provider>
            </ListSortContext.Provider>
        </ListFilterContext.Provider>
    </ListContext.Provider>
);
```

To test these optimizations, we should add new components and use them on the demo.

## Tasks

- [x] Add `ListContextProvider` and use it instead of `ListContext.Provider` everywhere
- [x] Add `ListFilterContext` and `useListFilterContext`
- [x] Add `ListSortContext` and `useListSortContext`
- [x] Add `ListPaginationContext` and `useListPaginationContext`
- [x] Add `<FilterListItem>` and `<FilterLiveSearch>` (depending on `useListFilterContext`) based on the ecommerce demo
- [x] Add `<SortButton>` (depending on `useListSortContext`) and use it in the ecommerce demo
- [ ] ~~Add tests~~ (no time)
- [x] Add documentation

## Ecommerce demo changes

image


fzaninotto:
Added documentation. Tests will be added in a future PR. Ready for review / merge.

docs/List.md
```diff
@@ -1316,6 +1172,153 @@ const FilterSidebar = () => (
...
 
+### Building a Custom Filter
+
+![Filters with submit button](./img/filter_with_submit.gif)
+
+If neither the `<Filter>` button/form combo or the `<FilterList>` sidebar match your need, you can always build your own. React-admin provides shortcuts to facilitate the development of custom filters.
+
+For instance, by default, the filter button/form combo doesn't provide a submit button, and submits automatically after the user has finished interacting with the form. This provides a smooth user experience, but for some APIs, it can cause too many calls. 
+
+In that case, the solution is process the filter when users click on a submit button, rather than when they type values in form inputs. React-admin doesn't provide any component for that, but it's a good opportunity to illustrate the internals of the filter functionality. We'll actually provide an alternative implementation to the `<Filter>` button/form combo.
```
djhi, line 1127:
```suggestion
In that case, the solution is to process the filter when users click on a submit button, rather than when they type values in form inputs. React-admin doesn't provide any component for that, but it's a good opportunity to illustrate the internals of the filter functionality. We'll actually provide an alternative implementation to the `<Filter>` button/form combo.
```

PR #5033, opened 2020-07-10
Fix documentation : Remove unused import of EditButton

jpetitcolas:
`EditButton` was imported in the code sample but without any purpose


fzaninotto:
Thanks!

PR #5042, opened 2020-07-12
[RFR] Fix ESLint @material-ui/core rule configuration

rassie:
This is a continuation of #4983. Apparently, I have been overly optimistic about ESLint's capability at configuration parsing and application, which lead to error reporting only on one of two restricted imports. This PR corrects it and also fixes all the issues this caused.


Luwangel:
Thanks again!

PR #5044, opened 2020-07-13
Change to use length of values with useEffect in AutocompleteArrayInput

helenwilliamson:
as if you spread values when this is nothing in values, the second parameter to useEffect will have 1 entry.
when there is a something in values, the second parameter to useEffect will have 2 entries. By using lenghth,
there are always 2 entries in the array

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
@@ -217,7 +217,7 @@ const AutocompleteArrayInput: FunctionComponent<
     // would have to first clear the input before seeing any other choices
     useEffect(() => {
         handleFilterChange('');
-    }, [...values, handleFilterChange]);
+    }, [values.length, handleFilterChange]);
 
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
```diff
@@ -217,7 +217,7 @@ const AutocompleteArrayInput: FunctionComponent<
     // would have to first clear the input before seeing any other choices
     useEffect(() => {
         handleFilterChange('');
-    }, [...values, handleFilterChange]);
+    }, [values.length, handleFilterChange]);
```
djhi, line 220: What if the values change but not the length? For example, by using final-form change action, we could replace a value by another and the length would not change. What do you think about the following?
```suggestion
    }, [values.join(','), handleFilterChange]);
```

helenwilliamson: sure - that'd work

Issue #4635, opened 2020-04-03
DeleteButton throwing error on Show view


gvillo:
I am facing an error while accessing a show page (adding custom actions to it) directly via URL. The error is not happening if you click on Show button on admin.

**What you were expecting:**
Display show page correctly

**What happened instead:**
Throwing an error

**Steps to reproduce:**
Add custom actions (like https://marmelab.com/react-admin/Show.html), if you add a DeleteButton without checking data object, if you hit enter in the show url, you will get an error in DeleteWithUndoButton.

**Related code:**
```js
const CustomShowActions = ({ basePath, data, resource }) => (
  <TopToolbar>
    <EditButton basePath={basePath} record={data} />
    <DeleteButton basePath={basePath} record={data} resource={resource}/>
  </TopToolbar>
);

export const CustomerShow = props => (
  <Show actions={<CustomShowActions />} {...props}>
    <SimpleShowLayout>
      <TextField source="id" />
      <TextField source="name" />
    </SimpleShowLayout>
  </Show>
);
```

For some weird reason https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple is not working.

**Other information:**

Workaround: checking if data is not undefined
```js
const CustomShowActions = ({ basePath, data, resource }) => (
  <TopToolbar>
    <EditButton basePath={basePath} record={data} />
    {data &&
      <DeleteButton basePath={basePath} record={data} resource={resource}/>
    }
  </TopToolbar>
);
```

**Environment**

* React-admin version: 3.4.0
* Last version that did not exhibit the issue (if applicable): NA
* React version: 16.13.1
* Browser: Google Chrome Version 80.0.3987.163 (Official Build) (64-bit)
* Stack trace (in case of a JS error):

```
TypeError: Cannot read property 'id' of undefined
DeleteWithUndoButton
node_modules/ra-ui-materialui/esm/button/DeleteWithUndoButton.js:37
  34 | var notify = useNotify();
  35 | var redirect = useRedirect();
  36 | var refresh = useRefresh();
> 37 | var _d = useDelete(resource, record.id, record, {
     | ^  38 |     action: CRUD_DELETE,
  39 |     onSuccess: function () {
  40 |         notify('ra.notification.deleted', 'info', { smart_count: 1 }, true);
View compiled
renderWithHooks
node_modules/react-dom/cjs/react-dom.development.js:14803
  14800 |   }
  14801 | }
  14802 | 
[...]

PR #5052, opened 2020-07-15
DeleteWithConfirmButton record.id Undefined

greenfrogs:
Fixes a record.id undefined bug which occurs when using the DeleteButton with undoable set to false. Adds the DeleteButton with undoable set to false to simple for the edit integration test. This is likely related to #4635


greenfrogs:
Github is not automatically pulling changes from the master branch so I will recreate the pull request.

PR #5056, opened 2020-07-15
DeleteWithConfirmButton record.id Undefined

greenfrogs:
Fixes a record.id undefined bug which occurs when using the DeleteButton with undoable set to false. Adds the DeleteButton with undoable set to false to simple for the edit integration test (and updated tests to click on the undo=true delete button). This is likely related to #4635

### DIFF
--- a/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
+++ b/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
@@ -83,7 +83,7 @@ const DeleteWithConfirmButton: FC<DeleteWithConfirmButtonProps> = props => {
                             true
                         ),
                     }),
-                    id: record.id,
+                    id: record ? record.id : null,
                 }}
                 onConfirm={handleDelete}
                 onClose={handleDialogClose}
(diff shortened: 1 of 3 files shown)


greenfrogs:
Hi, thanks for the quick response. I explained the bug quite badly so I had attached a demo of it crashing on the simple example:

`record.id`:
react-admin-id-bug

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -83,7 +83,7 @@ const DeleteWithConfirmButton: FC<DeleteWithConfirmButtonProps> = props => {
                             true
                         ),
                     }),
-                    id: record.id,
+                    id: record ? record.id : null,
```
JulienMattiussi, line 86: I agree with @Luwangel that the button should never be built without a record id, an that the responsibility to give the needed props props belong to his parent.
But maybe you can add a explic check in the button to alert explicitly when this id prop is missing.

greenfrogs:
Thank you for all of your help, I have corrected the issues in my code. For future users, do you think it would be possible to add it / make it clearer in the documentation?

fzaninotto:
Superseded by #5110

Issue #5026, opened 2020-07-09
Documentation example for "Not Using the `<Admin>` Components" is not working


srosset81:
**What you were expecting:**
I followed the guide to [include React Admin in another app](https://marmelab.com/react-admin/CustomApp.html).

**What happened instead:**
The first part of the guide ("Using an Existing Redux Provider") worked fine, but when I tried to do the second part ("Not Using the `<Admin>` Components"), I encountered several problems:

- The example `authProvider` did not work (`const authProvider = () => Promise.resolve();`), because `authProvider` is now supposed to be an object instead of a function. I used instead my own provider.
- No `locale` variable was defined, so I set it to `"en"`.
- No `theme` was provided to the `<ThemeProvider>`, so I had to create one.
- `PropTypes` did not exist so I had to import the `prop-types` package.
- The `withContext` expected `authProvider` to be a `PropTypes.object` instead of a `PropTypes.func`

I would gladly submit a PR to fix these few documentation problems, but **I also encountered a severe bug** that I was not able to fix: when I try to edit a resource, the client is redirected to the list, but the data are not updated.

**Steps to reproduce:**
Here's a codesandbox: https://codesandbox.io/s/naughty-snowflake-l2pe2
To reproduce:
- Select a post on the list
- Edit the title and click `Save`
- The optimistic update works and the title is changed, but the data are not updated

How do I know that "the data are not updated" ?
- In my own data provider, I can see the `update` method is never called.
- In the Redux DevTools, I see it stops at `@@router/LOCATION_CHANGE` while on [working examples](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple), the `RA/COMPLETE` action is called a little bit later.
- If I try to reload the page, I get a warning from my browser that data have not been submitted. I don't have this warning when I try working examples.
- If I try to edit or create another post, I get errors. I suspect that React-Admin thinks I'm still in "optimistic update" mode.

**Related code:**
https://codesandbox.io/s/naughty-snowflake-l2pe2

**Other information:**
I wonder if it is a problem with the redux-saga ?

**Environment**
- React-admin version: 3.6.3 on the codesandbox (but it also doesn't work with the latest 3.7.0 version)
- React version: 16.9.0 on the codesandbox
- Browser: latest version of Chrome

Thanks !

srosset81:
Sorry the file `createAdminStore.js` was missing from the codesandbox, I don't know where it went.

[...]

PR #5058, opened 2020-07-16
Update "Not Using the <Admin> Components" doc

srosset81:
This update the documentation for "Not Using the `<Admin>` Components" to fix several problems.

- The example `authProvider` does not work (`const authProvider = () => Promise.resolve();`), because `authProvider` is now supposed to be an object instead of a function.
- No `locale` variable is defined.
- No `theme` is provided to the `<ThemeProvider>`.
- `PropTypes` does not exist so it must be imported from the `prop-types` package.
- The `withContext` expects `authProvider` to be a `PropTypes.object` instead of a `PropTypes.func`
- The `Notification` component is not included, while it is necessary for the data provider to work.

The last problem is probably [a bug](https://github.com/marmelab/react-admin/issues/5026) but I think it can be worth documenting it until the bug is fixed.

### DIFF
--- a/docs/CustomApp.md
+++ b/docs/CustomApp.md
@@ -150,23 +150,26 @@ Here is the main code for bootstrapping a barebone react-admin application witho
 ```diff
 // in src/App.js
 import * as React from "react";
++import PropTypes from "prop-types";
 import { Provider } from 'react-redux';
 import { createHashHistory } from 'history';
 +import { ConnectedRouter } from 'connected-react-router';
 +import { Switch, Route } from 'react-router-dom';
 +import withContext from 'recompose/withContext';
 -import { Admin, Resource } from 'react-admin';
-+import { AuthContext, DataProviderContext, TranslationProvider, Resource } from 'react-admin';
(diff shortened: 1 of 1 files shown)


docs/CustomApp.md
```diff
@@ -244,12 +248,14 @@ const App = () => (
...
-+       authProvider: PropTypes.func,
++       authProvider: PropTypes.object,
 +   },
 +   () => ({ authProvider })
 +)(App);
 ```
 
 Note that this example still uses `<Resource>`, because this component lazily initializes the store for the resource data.
 
+Currently the `<Notification>` component is required (See [issue #5026](https://github.com/marmelab/react-admin/issues/5026)).
```
djhi, line 259: The notification component will always be required for optimistic support so I think you can remove this line

srosset81: Done !
Have to say I find it strange that a component called `Notification` is required for optimistic update. If it can't be changed, maybe it should be renamed ?

fzaninotto:
Thanks!

Issue #4793, opened 2020-05-08
Deprecated findDOMNode warning in StrictMode


leonardoarroyo:
This issue was already created(#4577). When it was closed, react-transition-group had not yet been updated. There's a new release in react-transition-node which fixed the issue. Can we get that dep updated?

fzaninotto:
Fixed by #4597

rlems:
I still seem to have this issue with react-transition-group 4.3.0 installed.
I think the issue is solved in 4.4.0: https://github.com/reactjs/react-transition-group/releases/tag/v4.4.0
In the `SimpleFormIterator` there'd have to be a nodeRef provided to the `CCSTransition` and `li` component

GimpMaster:
Sorry to comment on a closed issue, but I am still seeing this issue even with react-transition-group version 4.4.1

Here is yarn.lock

```
react-transition-group@^4.4.0, react-transition-group@^4.4.1:
  version "4.4.1"
  resolved "https://registry.yarnpkg.com/react-transition-group/-/react-transition-group-4.4.1.tgz#63868f9325a38ea5ee9535d828327f85773345c9"
  integrity sha512-Djqr7OQ2aPUiYurhPalTrVy9ddmFCCzwhqQmtN+J3+3DzLO209Fdr70QrN8Z3DsglWql6iY1lDWAfpFiBtuKGw==
  dependencies:
    "@babel/runtime" "^7.5.5"
    dom-helpers "^5.0.1"
    loose-envify "^1.4.0"
    prop-types "^15.6.2"
```

Here is the error I'm seeing:

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
    in div (created by ForwardRef(Paper))
    in ForwardRef(Paper) (created by WithStyles(ForwardRef(Paper)))
    in WithStyles(ForwardRef(Paper)) (created by Transition)
    in Transition (created by ForwardRef(Grow))
    in ForwardRef(Grow) (created by Unstable_TrapFocus)
    in Unstable_TrapFocus (created by ForwardRef(Modal))
    in div (created by ForwardRef(Modal))
    in ForwardRef(Portal) (created by ForwardRef(Modal))
    in ForwardRef(Modal) (created by ForwardRef(Popover))
[...]

PR #5079, opened 2020-07-23
Use latest react-transition-group

pietro909:
Fixes https://github.com/marmelab/react-admin/issues/4793#issuecomment-648764853


pietro909:
Thank you, I've updated it @Luwangel 

Curious enough, the diff shows that it was [already in the yarn.lock](https://github.com/marmelab/react-admin/pull/5079/files#diff-8ee2343978836a779dc9f8d6b794c3b2R13573).

pietro909:
Updated. Thank you for your instructions, I hope is ok now :-)

PR #5087, opened 2020-07-27
Remove recompose dependency from ra-core and demo projects.

WiXSL:
Reference https://github.com/marmelab/react-admin/pull/5051#pullrequestreview-448934920


Luwangel:
Great. You make me think that I didn't push my `remove recompose` branch :D https://github.com/marmelab/react-admin/pull/5088

PR #5088, opened 2020-07-27
Remove recompose

Luwangel:
Follows https://github.com/marmelab/react-admin/pull/4786, https://github.com/marmelab/react-admin/pull/5051 and https://github.com/marmelab/react-admin/pull/5087

> This PR depends on changes merged in the `master` branch.

## Todo

- [x] Remove recompose from `ra-core`
- [x] Remove recompose from the demo
- [x] Remove recompose from `ra-ui-materialui` and replace its usage by the code


WiXSL:
You forgot to include changes from #5051

UPGRADE.md
```diff
@@ -1300,7 +1300,7 @@ The undo feature is partially implemented in the `Notification` component. If yo
 // in src/MyNotification.js
 import * as React from "react";
 import { connect } from 'react-redux';
-import compose from 'recompose/compose';
+import compose from 'recompose/compose'; // You should add recompose/compose to your dependencies
```
ThieryMichel, line 1303: recompose/compose is a simple compose function.
Not sure this is the case, but if this is the only needed function from recompose we better propose a lighter alternative, like lodash.compose.
Adding the whole recompose library just for the compose function is overkill.

fzaninotto:
Could you explain the breaking change in a comment in this PR? It will be easier to find the explanation when we release the next version.

Issue #5059, opened 2020-07-16
Switching order problem in ReferenceManyField


timofeyp:
**What you were expecting:**
I want to use sorting inside `ReferenceManyField`, as it works on `List` components: by clicking at column name, and then component switches its list order DESC\ASK.

**What happened instead:**
Just first click works. And there is order indicate icon just near first clicked column. No this icon on next clicked columns.

![2020-07-16 18-56-42 2020-07-16 18_57_52](https://user-images.githubusercontent.com/37977572/87694500-f3f28300-c796-11ea-905e-882edb473198.gif)

**Steps to reproduce:**
Just click at column names in app with `ReferenceManyField` component.

**Related code:**
```
`import React from 'react';
import PropTypes from 'prop-types';
import {
  Show,
  TextField,
  TabbedShowLayout,
  Tab,
  DateField,
  NumberField,
  ReferenceManyField,
  Datagrid,
  Pagination,
} from 'react-admin';
import SessionActions from 'components/Actions/SessionActions';

const View = (props) => (
  <Show
    actions={<SessionActions id={props.id} />}
    title={`Пользователь #${props.id}`}
    {...props}
  >
    <TabbedShowLayout>
      <Tab label="Инфо">
        <TextField label="Имя" source="first_name" />
        <TextField label="Отчество" source="middle_name" />
        <TextField label="Фамилия" source="last_name" />
        <NumberField label="Телефон" source="phone" />
      </Tab>
      <Tab label="Карты" path="body"></Tab>
      <Tab label="Девайсы" path="devices">
        <ReferenceManyField
          target="user_id"
          reference="devices"
          addLabel={false}
          pagination={<Pagination />}
        >
          <Datagrid>
            <TextField source="id" />
            <DateField source="created_at" />
            <DateField source="updated_at" />
            <TextField source="vendor_id" />
            <TextField source="push_token" />
          </Datagrid>
        </ReferenceManyField>
      </Tab>
      <Tab label="Сессии" path="sessions">
        <ReferenceManyField
          target="user_id"
          reference="sessions"
          addLabel={false}
          pagination={<Pagination />}
        >
          <Datagrid>
            <TextField source="id" />
            <DateField source="created_at" />
            <DateField source="updated_at" />
            <DateField source="expired_at" />
            <TextField source="useragent" />
          </Datagrid>
        </ReferenceManyField>
      </Tab>
    </TabbedShowLayout>
  </Show>
);

View.propTypes = {
[...]

PR #5094, opened 2020-07-28
Fix sorting a list using a <Datagrid> inside a <ReferenceManyField>

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/5059

Several bugs are fixed:

- Fix the switch between `ASC` and `DESC` orders inside `<ReferenceManyField>` and `<ReferenceArrayField>`
- Fix sorting using another field than the default one or the first selected one

I introduced a new prop named `field` in the `<DatagridHeaderCell>` which contains the same data as the prop named `sort` which is now deprecated.

This renaming is motivated by the confusion introduced by two similar names (sort in `<Datagrid>` and sort in `<DatagridHeaderCell>`) containing different content. The first one contains an object `{ field: 'id', order: 'ASC' }` named `sort` and the second one `{ sort: 'id', order: 'ASC' }`.

## Todo

- [x] Fix props used to memorize the `<DatagridHeaderCell>`
- [x] Fix the sorting issue of the `<Datagrid>`
- [x] Fix the updateSort method in the `<Datagrid>` to query the right order
- [x] Fix tests

## Screenshots

Peek 28-07-2020 12-27

### DIFF
--- a/packages/ra-ui-materialui/src/list/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/Datagrid.tsx
@@ -101,13 +101,15 @@ const Datagrid: FC<DatagridProps> = props => {
             event.stopPropagation();
             const newField = event.currentTarget.dataset.field;
             const newOrder =
-                currentSort.field === newField &&
-                event.currentTarget.dataset.order === 'ASC'
-                    ? 'DESC'
-                    : 'ASC';
+                currentSort.field === newField
+                    ? currentSort.order === 'ASC'
+                        ? 'DESC'
+                        : 'ASC'
+                    : event.currentTarget.dataset.order;
+
             setSort(newField, newOrder);
         },
-        [currentSort.field, setSort]
+        [currentSort.field, currentSort.order, setSort]
     );
 
     const handleSelectAll = useCallback(
--- a/packages/ra-ui-materialui/src/list/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/Datagrid.tsx
@@ -100,7 +100,7 @@ const Datagrid: FC<DatagridProps> = props => {
         event => {
             event.stopPropagation();
             setSort(
-                event.currentTarget.dataset.sort,
+                event.currentTarget.dataset.field,
                 event.currentTarget.dataset.order
             );
         },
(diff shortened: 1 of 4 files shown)


fzaninotto:
We should not introduce breaking changes at this stage. I suggest you fix the bug without renaming the arguments, and open an issue for the renaming and flag it with the breaking change tag. We'll address it for 4.0.

packages/ra-ui-materialui/src/list/Datagrid.tsx
```diff
@@ -99,12 +99,14 @@ const Datagrid: FC<DatagridProps> = props => {
...
             event.stopPropagation();
-            setSort(
-                event.currentTarget.dataset.sort,
-                event.currentTarget.dataset.order
-            );
+            const newField = event.currentTarget.dataset.field;
+            const newOrder =
+                currentSort.field === newField && currentSort.order === 'ASC'
+                    ? 'DESC'
+                    : 'ASC';
```
fzaninotto, line 106: The default sort order can de defined on a per field basis. You must rely on the `data-order` attribute.

packages/ra-ui-materialui/src/list/Datagrid.tsx
```diff
@@ -197,8 +199,8 @@ const Datagrid: FC<DatagridProps> = props => {
                             />
                         </TableCell>
                     )}
-                    {Children.map(children, (field, index) =>
-                        isValidElement(field) ? (
+                    {Children.map(children, (field, index) => {
+                        return isValidElement(field) ? (
```
fzaninotto, line 203: why this change?

Luwangel: I'm reverting this change. It comes from my investigations.

packages/ra-ui-materialui/src/list/Datagrid.tsx
fzaninotto: In this new piece of code, if the user clicks on the header of a column that isn't yet used for sorting (thus `currentSort.field !== newField`), the sort order defaults to the `ASC` sort order. In the previous piece of code, the sort order defaulted to `event.currentTarget.dataset.order`. That's a breaking change. 

You see, developers can customize the default sort order for a field using the `sortByOrder` prop on a Field - and that prop is stored in the header dataset. If you don't use that value, you're breaking the `sortByOrder` feature. 

See https://marmelab.com/react-admin/List.html#specify-sort-order

Issue #4943, opened 2020-06-15
Allow usage of BulkDeleteButton with undoable=false in non-list pages


sebitoelcheater:
**What you were expecting:**
In List component of a resource, when I select one element to delete with a BulkDeleteButton with `undoable` prop set to false, a confirmation dialog is open. When I click confirm, the dialog should disappear after the deletion.

**What happened instead:**
The dialog remains open.
![error-bulk-delete-ra](https://user-images.githubusercontent.com/2347536/84620994-7097f600-aea7-11ea-8959-3b6001e46009.gif)

**Steps to reproduce:**
```
const AssetBulkActionButtons = props => (
    <Fragment>
      <BulkDeleteButton {...props} undoable={false}/>
    </Fragment>
);

<List {...props} bulkActionButtons={<AssetBulkActionButtons/>}>
...
```

**Environment**
* React-admin version: 3.5.5
* Last version that did not exhibit the issue (if applicable): 2.x.x
* React version: 16.13.1
* Browser: Chrome
* Stack trace (in case of a JS error): -

djhi:
I can't reproduce this issue on our simple example. Please provide a sample application based on this [CodeSandbox](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple).

a-pavlov-parc:
I have the same issue with the undoable DeleteButton component. I checked the sources and I don't see it is supposed to be closed on success
![image](https://user-images.githubusercontent.com/8544449/87701276-6025b480-c7a0-11ea-83d9-6f4219913c02.png)

Luwangel:
Hello @sebitoelcheater and @a-pavlov-parc 

Just like @djhi, I cannot reproduce the described bug on the Simple Example. See my [Code Sandbox](https://codesandbox.io/s/young-sunset-p7k0x?file=/src/posts/PostList.js).

Maybe it comes from your code or from an old version of React-Admin.

I'm closing this issue for the moment. Don't hesitate to reopen it if you can reproduce it on CodeSandbox.

Regards,
Adrien

a-pavlov-parc:
@Luwangel in order to reproduce this you must perform editing on the list page, you're doing this on another one. In your case, it does redirection after successful deleting, in our case it does not because the page it's going to redirect to is the same page where we are staying at the moment

fzaninotto:
[...]

PR #5097, opened 2020-07-29
Fix bulk delete confirmation dialog does not close after success when using Datagrid in a non-list page

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/4943

It allows the usage of BulkDeleteButton with undoable=false in non-list pages

## Todo

- [x] Close the delete confirmation dialog on success

### DIFF
--- a/packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
+++ b/packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
@@ -80,18 +80,19 @@ const useDeleteWithConfirmController = ({
     const [deleteOne, { loading }] = useDelete(resource, null, null, {
         action: CRUD_DELETE,
         onSuccess: () => {
+            setOpen(false);
             notify('ra.notification.deleted', 'info', { smart_count: 1 });
             redirect(redirectTo, basePath);
             refresh();
         },
         onFailure: error => {
+            setOpen(false);
             notify(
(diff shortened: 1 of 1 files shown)


djhi:
Why are you targeting `next` ?

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -80,18 +80,19 @@ const useDeleteWithConfirmController = ({
     const [deleteOne, { loading }] = useDelete(resource, null, null, {
         action: CRUD_DELETE,
         onSuccess: () => {
+            setOpen(false);
```
ThieryMichel, line 83: A suggestion: It could be the occasion to add an `onEnd` event that would fire on both success and failure.
But yet again maybe it is overkill.

ThieryMichel, line 83: A suggestion: It could be the occasion to add an `onEnd` event that would fire on both success and failure.
Or maybe it is overkill.

Luwangel: It's not a little update because it should be implemented in all the hooks to keep the API consistent. Moreover, I'm not sure it's something we want in React-Admin. You can open an issue to discuss it :wink:.

PR #5109, opened 2020-08-04
Fix Placeholder component className prop

WiXSL:


djhi:
Would you mind explaining why it is an issue ?

WiXSL:
Just for consistency since is the only component that has that className prop order.

fzaninotto:
Thanks!

PR #5121, opened 2020-08-06
Fix Sidebar component classes

### LINKED ISSUE #4407: Easier customization of Sidebar

WiXSL:
Fixes #4407.

I needed to use custom classes and I did it like this.
I you have a better implementation I'm all ears.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Sidebar.js
+++ b/packages/ra-ui-materialui/src/layout/Sidebar.js
@@ -83,6 +83,7 @@ const Sidebar = props => {
                 className: drawerPaper,
             }}
             onClose={toggleSidebar}
+            classes={classes}
             {...rest}
         >
             {cloneElement(Children.only(children), {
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/Sidebar.js
```diff
@@ -11,6 +11,18 @@ export const CLOSED_DRAWER_WIDTH = 55;
...
+        paperAnchorLeft: {},
+        paperAnchorRight: {},
+        paperAnchorTop: {},
+        paperAnchorBottom: {},
+        paperAnchorDockedLeft: {},
+        paperAnchorDockedTop: {},
+        paperAnchorDockedRight: {},
+        paperAnchorDockedBottom: {},
+        modal: {},
         drawerPaper: {
```
Luwangel, line 26: So now there are no differences between the `drawerPaper` and the `paper` classnames? I think we can use only the `paper` key and mark the `drawerPaper` key as deprecated.

fzaninotto: That's a BC break and it brings small added value. Let's merge this PR as is.

fzaninotto:
Thanks!

PR #5126, opened 2020-08-09
Fix warning when submitting a Form with updated CheckboxGroupInput

### LINKED ISSUE #4969: Warning for unsaved changes when updating a <CheckboxGroupInput>
**What you were expecting:**

In the simple example, edit a Post, update the Notification recipients and save, then get to the list.

**What happened instead:**

After clicking on the Save button, an alert shows up and says:

> Some of your changes weren't saved. Are you sure you want to ignore them?

**Steps to reproduce:**

I cannot reproduce the problem in the CodeSandbox, but I get it when running the simple example with `make run` on Master. 

I can only reproducing when updating the Notification Recipients, not any other field. 

image

**Environment**

* React-admin version: 3.6.1
* React version: 16.12
* Browser: Chrome

fzaninotto:
Closes #4969


fzaninotto:
Something is wrong, moving to WIP

fzaninotto:
OK, this is the wrong fix, closing.

PR #5146, opened 2020-08-12
Fix Form Initialization from Record

### LINKED ISSUE #4954: Redirect from create to edit can have a wrong pristine due to optimistic rendering
**What you were expecting:**

Within a project, I have a **customer** `resources`, `<CustomerCreate />` and `<CustomerEdit />` form. 

Server side, within it's rest controller, I am :

- first saving the entities on `Stripe` (a payment system) using it's api
- then put the returned `customerStripeId` into my customer before saving the entity into my database. (The response of this controller is just the newly created id)

After creation, I have the default client redirect that brings me to the `<CustomerEdit />` form, I expect the `edit` form to have the `pristine` state with a `<Button disabled />`.

**What happened instead:**

On create redirect, I have the form in `pristine` state with the optimistic rendering, but then right after the entity fetch and the form update and lose it's `pristine` state when the `customerStripeId` is filled.

**Steps to reproduce:**

- Create a resource on a `<Create />` view that has the default redirect to edit
- Make the create controller add extra data to the entity while saving the create form.
- Notice the form values being updated and the pristine state being changed.

Pressing the application refresh button restore the `pristine=true`

**Environment**

* React-admin version: 3.6.1
* React version:  16.13.0
* Browser: Google Chrome

**Related issues**

- https://github.com/marmelab/react-admin/issues/4733
- https://github.com/marmelab/react-admin/pull/4904
- https://github.com/marmelab/react-admin/issues/4913#issuecomment-645370798

djhi:
Fixes #4954

### DIFF
--- a/packages/ra-core/src/form/useInitializeFormWithRecord.ts
+++ b/packages/ra-core/src/form/useInitializeFormWithRecord.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'react';
 import { useForm } from 'react-final-form';
-import { isObject } from '../inference/assertions';
+import { merge } from 'lodash';
 
 /**
  * Restore the record values which should override any default values specified on the form.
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/form/useInitializeFormWithRecord.ts
```diff
@@ -13,42 +13,9 @@ const useInitializeFormWithRecord = record => {
...
-                        // scalar value
-                        form.change(key, record[key]);
-                    }
-                    form.resetFieldState(key);
-                }
-            });
-        });
+        const initialValues = form.getState().initialValues;
+        const initialValuesMergedWithRecord = merge({}, initialValues, record);
+        form.initialize(initialValuesMergedWithRecord);
```
fzaninotto, line 18: This is neat but there is a difference with the previous code: you initialize values including the ones for which there is no associated input. You put that restriction in 3e56efe8caec8ab297dea9c3ccb8c293b5ff5309. Are you sure you're not reintroducing a bug?

fzaninotto:
ping

packages/ra-core/src/form/useInitializeFormWithRecord.ts
djhi: I did not find any

fzaninotto: Including the one from #3721?

djhi: I'll add some tests tomorrow

PR #5147, opened 2020-08-12
Fix addEventListener calls for compatibility with React v17.

### LINKED ISSUE #5134: Make eventListeners compatible with React 17
React 17 is around the corner, and it introduces a breaking change on event listeners (see https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues).

I think we must make a change in the 2 occurrences of `addEventListener` in the react-admin codebase.

WiXSL:
Fixes #5134


fzaninotto:
Thanks!

PR #5148, opened 2020-08-12
Fix RichTextInput paragraph margin bottom.

WiXSL:
ra


fzaninotto:
I don' understand the problem. It's normal that paragraphs have a bottom margin.

WiXSL:
But in a TextArea like component we should be able to have each line sticked together as we type enter key.

fzaninotto:
It's not a textarea, it's a rich text editor. The text should look as much as possible as the rendered text, in which paragraphs have margins.

WiXSL:
All right I'll have to override the style locally.

Issue #2741, opened 2019-01-09
Add a BackButton


sorcer1:
Adding a BackButton to do a history.go(-1) or another way will be usefull to add it in List's actions like that :

![image](https://user-images.githubusercontent.com/584256/50901802-859e6900-1419-11e9-99d2-ef0db7ef5193.png)

djhi:
Can you explain your usecase ?

sorcer1:
I will try with the demo app : https://marmelab.com/react-admin-demo/#/reviews

In the "Reviews" list we click in a customer (reference field), so we go to the customer show view.
I want to go back to the "Reviews" with a button.
In a browser is easy, but think about some devices (without physical back buttons) or destop/tablet apps in fullscreen mode.

djhi:
You can click the `Reviews` item in the menu.

We won't implement that in the core but you can create your own custom button for that by connecting it with the react router redux actions

mnlbox:
@djhi I think it's a very popular use-case and better to support in core. :thinking: 
Are you any plan to add this?

djhi:
Nope, we won't. We had one before and we explicitly removed it. Nothing prevent you to have your own default actions that you use on all your views though

djhi:
You can even create your own default views (Create, Edit, List, etc.) which just import/reexport the default one with your own default props

mnlbox:
@djhi But how we can back a route?
I'm not familiar with react-admin routing mechanism.

djhi:
Please ask this kind of questions on StackOverflow as explained in our issue template

mnlbox:
OK man :+1:

mnlbox:
@djhi  But I think maybe it's better to add a discord chat to react-admin too. It's help community grow better and someone also can find useful note there. :wink:

alexrosabr:
> Nope, we won't. We had one before and we explicitly removed it. Nothing prevent you to have your own default actions that you use on all your views though

I think a mere `history.go(-1)` has problems with circular navigation. IMHO the main use case is to simply go back to the **List** view from **Edit** and **Show** views. The framework could add an automatic "Back" button alongside with the "Edit" button in the **Show** view that simply redirects to the **List** view. A similar approach could be used to automatically add the "Back" button in the **Edit** view.

fzaninotto:
If what you need is a button that links to the List view, you can use the `<ListButton>`, as follows:

```jsx
[...]

PR #5155, opened 2020-08-13
[Doc] Explain how to add a custom back button

fzaninotto:
This is a common demand, so I think it should be documented.

Refs #2741

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -175,7 +175,7 @@ If you want this button to look like a Back button, you can pass a custom label 
 ```jsx
 import ChevronLeft from '@material-ui/icons/ChevronLeft';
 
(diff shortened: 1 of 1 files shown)


docs/CreateEdit.md
```diff
@@ -157,6 +157,32 @@ export const PostEdit = (props) => (
...
+    </TopToolbar>
+);
+```
+
+If you want this button to look like a Back button, you can pass a custom label and icon to the ListButton:
+
+```jsx
+import ChevronLeft from '@material-ui/icons/ChevronLeft';
+
+const EditActions = ({ basePath, data }) => (
```
Luwangel, line 178:
```suggestion
const PostEditActions = ({ basePath, data }) => (
```

Maybe rename this component to reuse the example defined before.

PR #5164, opened 2020-08-17
Send range header in simple REST client getList request

jpetitcolas:
Using the simple REST client, I noticed an issue on Chrome : it filters the `Content-Range` header in the response if no `Range` header is specified in the request, which sounds logic (even if perhaps too strict).


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5165, opened 2020-08-18
Allows List Empty Page Disabling

### LINKED ISSUE #5158: Disable empty state on List
I don't always want an empty state on a `List` and it's impossible to opt out

Passing `false` to the `empty` prop disable the empty state completely

Passing my `Datagrid` component as the empty state. It works but the `List` toolbar is not displayed.

jdemangeon:
Fixes #5158

It's sometimes useful to be able to display an empty list instead of displaying a special "empty page" when there's no data to display.

So, here is a small change that allows to display an empty page when the list is empty.

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -683,7 +683,7 @@ The `empty` component can call the `useListContext()` hook to receive the same p
 -   `total`,
 -   `version`,
 
-You can also set the `empty` props value to `false` to bypass empty page display and render an empty list instead.
+You can also set the `empty` props value to `false` to bypass the empty page display and render an empty list instead.
 
 ```
 import { List } from 'react-admin';
--- a/docs/List.md
+++ b/docs/List.md
@@ -683,6 +683,18 @@ The `empty` component can call the `useListContext()` hook to receive the same p
 -   `total`,
 -   `version`,
 
+You can also set the `empty` props value to `false` to bypass empty page display and render an empty list instead.
+
+```
+import { List } from 'react-admin';
+
+const ProductList = props => (
+    <List empty={false} {...props}>
+        ...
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -100,7 +100,7 @@ export const ListView: FC<ListViewProps> = props => {
             {...sanitizeRestProps(rest)}
         >
             <Title title={title} defaultTitle={defaultTitle} />
-            {shouldRenderEmptyPage
+            {shouldRenderEmptyPage && empty !== false
```
Luwangel, line 103: Why not update the `shouldRenderEmptyPage` method to use the `empty` prop?

docs/List.md
```diff
@@ -683,6 +683,18 @@ The `empty` component can call the `useListContext()` hook to receive the same p
 -   `total`,
 -   `version`,
 
+You can also set the `empty` props value to `false` to bypass empty page display and render an empty list instead.
```
djhi, line 686:
```suggestion
You can also set the `empty` props value to `false` to bypass the empty page display and render an empty list instead.
```

packages/ra-ui-materialui/src/list/ListView.tsx
jdemangeon: Because of Typescript ;)

fzaninotto (changes requested):
There are integration tests (List.spec.js) about the Empty feature, can you add one about the `false` value?

jdemangeon:
Review applied

fzaninotto (approved):

PR #5170, opened 2020-08-19
Convert Form components to typescript

fzaninotto:
Refs #4505


packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
```diff
@@ -85,7 +92,7 @@ const DefaultRemoveButton = props => {
     );
 };
 
-const SimpleFormIterator = props => {
+const SimpleFormIterator: FC<any> = props => {
```
fzaninotto, line 95: somehow I got lazy in the end

djhi: :scream:

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -150,17 +152,17 @@ const FormWithRedirect: FC<FormWithRedirectOwnProps & FormProps> = ({
 export interface FormWithRedirectOwnProps {
     defaultValue?: any;
     record?: Record;
-    save: (
+    save?: (
```
djhi, line 155: We really should rename this to `onSubmit` imo

fzaninotto: that'd be a breaking change... Perhaps in a future major.

djhi: Support both with a warning ?

djhi (approved):

PR #5176, opened 2020-08-20
Fix List subcontext lack resource data

fzaninotto:
The subcontexts of ListContext are made to avoid depending on the entire ListContext. Yet they are too tiny. the `resource`, for instance, is necessary for buttons. Adding a bit more data makes subcontexts enough for more cases, without adding more rerenders.


djhi:
We could probably introduce a `useResource` btw ;)

djhi (approved):

PR #5180, opened 2020-08-23
Add user name and avatar on the top bar

fzaninotto:
identity

Requires a new method in the `authProvider`:

```jsx
// in src/authProvider.js
const authProvider = {
    login: ({ username, password }) => { /* ... */ },
    getIdentity: () => {
        const { id, fullName, avatar } = JSON.parse(localStorage.getItem('auth'));
        return { id, fullName, avatar };
    }
    // ...
};

export default authProvider;
```

The change is backward compatible (if the `authProvider.getIdentity()` method isn't implemented, the admin shows an anonymous user avatar with no name, as before).


djhi (approved):
:ok_hand:

Issue #5096, opened 2020-07-29
Remove deprecated sort prop in <DataGridHeaderCell>


Luwangel:
Follows issue https://github.com/marmelab/react-admin/issues/5059 and PR https://github.com/marmelab/react-admin/pull/5094

**Is your feature request related to a problem? Please describe.**

The renaming has already be done. But we need to remove deprecated props.

The original issue is motivated by the confusion introduced by two similar names (`sort` in `<Datagrid>` and `sort` in `<DatagridHeaderCell>`) containing different content. The first one contains an object `{ field: 'id', order: 'ASC' }` named sort and the second one `{ sort: 'id', order: 'ASC' }`.

**Describe the solution you'd like**

Remove `sort` in favor to `field` in the `<DatagridHeaderCell>`.

Files concerned:

- `packages/ra-ui-materialui/src/list/DatagridHeaderCell.js` (line 64)

WiXSL:
Fixes #7065

PR #5182, opened 2020-08-24
Fix Deprecated Sort PropType on DatagridHeaderCell

jdemangeon:
Closing #5096


djhi:
Too soon. We don't want to do it before next major.

fzaninotto:
We don't have plans for 4.0 yet. I'll close this PR as it's a simple change and it needs rebase, but we'll keep track of it in the 4.0 roadmap.

PR #5183, opened 2020-08-24
Migrate Layout components to typescript

fzaninotto:
Refs #4505


packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
```diff
@@ -147,9 +149,9 @@ const useStyles = makeStyles(
     { name: 'RaReferenceArrayField' }
 );
 
-type ReferenceArrayFieldClassKey = 'progress';
+export type ReferenceArrayFieldClassKey = 'progress';
```
djhi, line 152: I don't get this `progress` value for the key

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -237,9 +237,9 @@ ReferenceFieldView.propTypes = {
     translateChoice: PropTypes.oneOfType([PropTypes.func, PropTypes.bool]),
 };
 
-type ReferenceFieldClassKey = 'link';
+export type ReferenceFieldClassKey = 'link';
```
djhi, line 240: Not sure if previous comment & fix is applicable here as well

fzaninotto: yes, fixed, too

packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
fzaninotto: That's an old way to type the classesOverride for mui styling. Replaced with our own ClassesOverride

djhi (approved):

PR #5202, opened 2020-08-29
Forward ref in data grid element

### LINKED ISSUE #5194: Support forwarding ref with Datagrid*
I want to embed drag and copy feature in ra data grid. But I cannot access data grid's `innerText` because `DataGrid`, `DataGridBody` does not forward `ref` to inner material-ui element.

Add `React.forwardRef` to `Datagrid.tsx`, `DatagridBody.tsx`, `DatagridCell.tsx`.

Add drag and copy feature to react-admin data grid.

jeiea:
Close #5194 

`DatagridRow` ref doesn't cover expanded row. It can be controversial in my thought.

I made this commit on next branch because react guides [use of forwardRef is breaking change](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers).

### DIFF
--- a/packages/ra-ui-materialui/src/list/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/Datagrid.tsx
@@ -63,7 +63,7 @@ import { ClassesOverride } from '../types';
  *     </Datagrid>
  * </ReferenceManyField>
  */
-const Datagrid: FC<DatagridProps> = props => {
+const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
     const classes = useDatagridStyles(props);
     const {
         optimized = false,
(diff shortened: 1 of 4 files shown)


djhi (approved):
:+1:

jeiea:
Currently `DatagridRow` is fragment, which cannot assign a ref.
I assigned ref to `TableRow` and didn't take account of the following `div` (expanded row).
For now I don't need ref of expanded row so it's fine to me, but I don't know how others think about this.
`DatagridRow`'s ref doesn't include expanded row. I thought this fact cannot be clear.

fzaninotto:
Yet I don't see any other solution - a `<tbody>` cannot contain anything other than `<tr>`, so we cannot enclose the two table rows with anything other than a fragment.  

We'll go with your change, and if the absence of ref on the expand row ever becomes a problem, we'll deal with it then.

fzaninotto (approved):

fzaninotto:
Thanks!

Issue #4412, opened 2020-02-11
BooleanInput parse and format props don't seem to work


amhouel:
**What you were expecting:**

Have the parse and format functions transform our record non-boolean values to boolean values for the input and vice-versa and see the BooleanInput toggle work accordingly on click.

**What happened instead:**

The BooleanInput is always set to true and doesn't change on click.
If the initial value is a true value, the parse and format functions successfully change the value to the false value on the 1st click but then get stuck on that value.
If the initial value is a false value, then the parse and format functions always return that same false value.

**Steps to reproduce:**

In the Posts list, click on any Edit button and inside the Edit form click on the "Title" BooleanInput.

**Related code:**

https://codesandbox.io/s/serverless-pine-dhgem

**Environment**

* React-admin version: 3.2.1
* Last version that did not exhibit the issue (if applicable): 2.9.9
* React version: 16.9
* Browser: Chrome
* Stack trace (in case of a JS error):

fzaninotto:
The `format` and `parse` functions are well called when I try this code:

```jsx
        <BooleanInput
          source="title"
          format={v => {console.log('format', v);return v}}
          parse={v =>{console.log('parse', v);return v}}
        />
```

So in my opinion, the problem is either on your side or in the final-form side.

jparkrr:
(I'm working with @amhouel)

Yes, you're right. It looks like react-final-form [doesn't support `format` for checkboxes](https://github.com/final-form/react-final-form/blob/f3c1aea93b520c0620f1b8090f5966e0da94a798/src/useField.js#L213)

fzaninotto:
Thanks for looking into it. I'm closing this issue.

pefi1011:
Hi @jparkrr @amhouel how did you solve this? I have the same requirement: transform non-boolean record (array of strings) to boolean values for the input and vice-versa and see the BooleanInput toggle work accordingly on click.

I have tried parse/format as @fzaninotto mentioned, but in the logs I see do not see `parse` logs. The app crashes with `Error: Cannot set a numeric property on an object` before the `parse` function is executed

kopax:
I have the same issue with BooleanInput.

armellarcier:
I found a workaround. You can use it like `<FormattedBooleanInput trueValue="yes" falseValue="no" source="myfield" />`

```jsx
import React from "react";
import { FormDataConsumer, BooleanInput } from "react-admin";

const FormattedBooleanInput = ({
    falseValue = "N",
[...]

PR #5203, opened 2020-08-30
fix BooleanInput parse/format

JustMonk:
## Problem
React-admin boolean input will be broken if parse/format function used.

### Description
I recently upgraded from 2.9 to the latest 3.8.2 and found my BooleanInputs was broken. My API takes on the string values "1" (if true) or "0" (if false) for these purposes I use parse `v => v ? "1" : "0"` and format `v => !!Number(v)`. On 2.9 its work, but on 3.X input freeze and dont changes value after a couple of switches.

### Step to reproduce
You can see live example here: https://codesandbox.io/s/ra-boolean-parse-format-example-yixdg
(Default final-form field with same params work, ra-boolean - does not)

### Solution
I tried to find the problem and saw the `BooleanInput` has a property `"type": "checkbox"`, in the [official documentation of final form](https://final-form.org/docs/react-final-form/types/FieldProps#type) this field is marked as optional. I think this parameter works fine inside the final form, but in react-admin context it breaks the parse/format functions, when I removed it - the problem was solved.
(P.S: since ra-boolean is based on MUI switch, DOM element will have `type="checkbox"` by default and accessibility will not be affected)

### DIFF
--- a/packages/ra-ui-materialui/src/input/BooleanInput.tsx
+++ b/packages/ra-ui-materialui/src/input/BooleanInput.tsx
@@ -32,7 +32,7 @@ const BooleanInput: FunctionComponent<
 }) => {
     const {
         id,
-        input: { onChange: finalFormOnChange, type, value, ...inputProps },
+        input: { onChange: finalFormOnChange, value, ...inputProps },
         isRequired,
         meta: { error, touched },
     } = useInput({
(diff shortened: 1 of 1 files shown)


djhi:
It fails the tests though so something is wrong

fzaninotto:
Already reported as #4412. The conclusion as that this bug must be fixed in react-final-form, not in react-admin.

PR #5212, opened 2020-08-31
Fix Optimistic Effects Handling

djhi:
- [x] Fixes optimistic effects handling
Before this fix, if you had a `<List>` with a `<DeleteButton undoable>` on every line, deleting multiples records one after the other would end up with an invalid state. Removed items would still appear in the list even if the server call was made.

- [x] Fixes `beforeunload` event handler removal

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -406,13 +406,14 @@ const replayOptimisticCalls = async () => {
     // We must perform any undoable queries first so that the effects of previous undoable
     // queries do not conflict with this one.
 
-    // We only handle all sideeffects queries if there are no more undoable queries
+    // We only handle all side effects queries if there are no more undoable queries
     if (undoableOptimisticCalls.length > 0) {
         clone = [...undoableOptimisticCalls];
+        undoableOptimisticCalls.splice(0, undoableOptimisticCalls.length);
+
         await Promise.all(
             clone.map(params => Promise.resolve(doQuery.call(null, params)))
         );
-        undoableOptimisticCalls.splice(0, undoableOptimisticCalls.length);
     } else {
         clone = [...optimisticCalls];
         optimisticCalls.splice(0, optimisticCalls.length);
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -386,13 +400,27 @@ const warnBeforeClosingWindow = event => {
...
-        )
-    );
-    optimisticCalls.splice(0, optimisticCalls.length);
+const replayOptimisticCalls = async () => {
+    let clone;
+
+    // We must perform any undoable queries first so that the effects of previous undoable
+    // queries do not conflict with this one.
+
+    // We only handle all sideeffects queries if there are no more undoable queries
```
fzaninotto, line 409:
```suggestion
    // We only handle all side effect queries if there are no more undoable queries
```

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -386,13 +400,27 @@ const warnBeforeClosingWindow = event => {
...
+    if (undoableOptimisticCalls.length > 0) {
+        clone = [...undoableOptimisticCalls];
+        await Promise.all(
+            clone.map(params => Promise.resolve(doQuery.call(null, params)))
+        );
+        undoableOptimisticCalls.splice(0, undoableOptimisticCalls.length);
```
fzaninotto, line 415: This one should move just after the clone

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -300,7 +305,10 @@ const performUndoableQuery = ({
             if (window) {
                 window.removeEventListener(
                     'beforeunload',
-                    warnBeforeClosingWindow
+                    warnBeforeClosingWindow,
+                    {
+                        capture: true,
+                    }
```
djhi, line 311: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal

fzaninotto (changes requested):
Please describe the problem this PR solves

fzaninotto (approved):

PR #5222, opened 2020-09-02
Support function for initialValues

### LINKED ISSUE #5221: Form initialValues prop as function is not working
**What you were expecting:**

As per [docs](https://marmelab.com/react-admin/CreateEdit.html#global-default-value), the value of the form `initialValues` prop is an object, or a function returning an object

**What happened instead:**

Passing function to initialValues prop does not run while passing to defaultValue works.

**Related code:**

```jsx
<SimpleForm initialValues={(values) => {
  console.log(values)
  return values
}} />
```

**Environment**

* React-admin version: 6.8.3
* React version: 16.13.1

djhi:
Fixes #5221

- [x] Fix
- [x] Tests

### DIFF
--- a/packages/ra-core/src/form/getFormInitialValues.ts
+++ b/packages/ra-core/src/form/getFormInitialValues.ts
@@ -10,19 +10,19 @@ export default function getFormInitialValues(
     }
 
     return {
-        ...getValues(defaultValue),
-        ...getValues(initialValues),
+        ...getValues(defaultValue, record),
+        ...getValues(initialValues, record),
         ...record,
     };
 }
 
-function getValues(values) {
+function getValues(values, record) {
     if (typeof values === 'object') {
         return values;
     }
 
     if (typeof values === 'function') {
-        return values();
+        return values(record);
     }
 
     return {};
--- a/packages/ra-core/src/form/getFormInitialValues.ts
+++ b/packages/ra-core/src/form/getFormInitialValues.ts
@@ -3,28 +3,27 @@ export default function getFormInitialValues(
     defaultValue,
     record
 ) {
-    let finalInitialValues = {
-        ...initialValues,
-        ...record,
-    };
-
     if (typeof defaultValue !== 'undefined') {
         console.warn(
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/form/getFormInitialValues.ts
```diff
@@ -3,28 +3,27 @@ export default function getFormInitialValues(
...
+    };
+}
+
+function getValues(values) {
+    if (typeof values === 'object') {
+        return values;
+    }
+
+    if (typeof values === 'function') {
+        return values();
```
fzaninotto, line 25: not good, we were passing the record as parameter in the previous version.

packages/ra-core/src/form/getFormInitialValues.ts
```diff
@@ -3,28 +3,27 @@ export default function getFormInitialValues(
...
-            ...defaultValue,
-            ...finalInitialValues,
-        };
-    } else if (typeof defaultValue === 'function') {
-        finalInitialValues = {
-            ...defaultValue(record),
-            ...finalInitialValues,
-        };
+    return {
+        ...getValues(defaultValue),
```
fzaninotto, line 13: `getValues` should take the `record` as parameter.

fzaninotto (changes requested):

fzaninotto (approved):

packages/ra-core/src/form/getFormInitialValues.ts
```diff
@@ -3,28 +3,27 @@ export default function getFormInitialValues(
...
-        };
-    } else if (typeof defaultValue === 'function') {
-        finalInitialValues = {
-            ...defaultValue(record),
-            ...finalInitialValues,
-        };
+    return {
+        ...getValues(defaultValue, record),
+        ...getValues(initialValues, record),
+        ...record,
```
djhi, line 15: We should probably check whether `initialValues` or `defaultValue` are functions before overriding them with the record as we pass it to them as an argument. What do you think @fzaninotto ?

fzaninotto (approved):

PR #5224, opened 2020-09-02
simpleRestProvider: using a custom http header to count items in a collection

alexisjanvier:
Historically, Simple REST Data Provider uses the http `Content-Range` header to retrieve the number of items in a collection. But this is a *hack* of the [primary role of this header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range).

And this can be problematic, for example within an infrastructure using a Varnish that may use, modify or delete this header. We also have feedback indicating that using this header is problematic when you host your application on [Versel](https://vercel.com/).

The solution is to use another http header to return the number of collection's items.

**This PR introduces a third optional parameter to the `simpleRestProvider` allowing to define this header.**

### DIFF
--- a/packages/ra-data-simple-rest/README.md
+++ b/packages/ra-data-simple-rest/README.md
@@ -105,7 +105,7 @@ Now all the requests to the REST API will contain the `Authorization: SRTRDFVESG
 
 Historically, Simple REST Data Provider uses the http `Content-Range` header to retrieve the number of items in a collection. But this is a *hack* of the [primary role of this header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range).
 
-And this can be problematic, for example within an infrastructure using a Varnish that may use, modify or delete this header. We also have feedback indicating that using this header is problematic when you host your application on [Versel](https://vercel.com/).
+However this can be problematic, for example within an infrastructure using a Varnish that may use, modify or delete this header. We also have feedback indicating that using this header is problematic when you host your application on [Vercel](https://vercel.com/).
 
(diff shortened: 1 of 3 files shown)


packages/ra-data-simple-rest/README.md
```diff
@@ -101,6 +101,40 @@ const httpClient = (url, options = {}) => {
 
 Now all the requests to the REST API will contain the `Authorization: SRTRDFVESGNJYTUKTYTHRG` header.
 
+## Note about Content-Range
+
+Historically, Simple REST Data Provider uses the http `Content-Range` header to retrieve the number of items in a collection. But this is a *hack* of the [primary role of this header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range).
+
+And this can be problematic, for example within an infrastructure using a Varnish that may use, modify or delete this header. We also have feedback indicating that using this header is problematic when you host your application on [Versel](https://vercel.com/).
```
djhi, line 108:
```suggestion
However this can be problematic, for example within an infrastructure using a Varnish that may use, modify or delete this header. We also have feedback indicating that using this header is problematic when you host your application on [Vercel](https://vercel.com/).
```

packages/ra-data-simple-rest/README.md
```diff
@@ -101,6 +101,40 @@ const httpClient = (url, options = {}) => {
...
+
+The solution is to use another http header to return the number of collection's items, the other header commonly used for this is `X-Total-Count`. So if you use `X-Total-Count`, you will have to :
```
djhi, line 110:
```suggestion
The solution is to use another http header to return the number of collection's items. The other header commonly used for this is `X-Total-Count`. So if you use `X-Total-Count`, you will have to :
```

djhi (changes requested):

djhi (approved):

PR #5243, opened 2020-09-11
Remove sanitizing __ types

maximgeerinck:
When using graphql, we rely on `__typename` for identifying the correct types in a `union` type.

imagine i have the following interface

```
interface Animal {
    id: ID!
}

type Crawler implements Animal {
    legs: Int
}

type Swimmer implements Animal {
    fins: Int
}
```

we would return a type of the form `union SomeAnimal = Crawler & Swimmer` where we rely on `__typename` to correctly display some information on the dashboard.

With this proposal I ask to not filter out `__` fields.


djhi:
Is this still needed in v5?

fzaninotto:
V3 isn't supported anymore, please migrate to v4 or v5.

Issue #5204, opened 2020-08-30
useLogin optional pathName cannot be reached


ValentinnDimitroff:
**What you were expecting:**
Using the hook useLogin("/redirectUrl") with passed optional pathName for redirect after authenticating should try to redirect ot the given path if it is registered route.

**What happened instead:**
When a user has logout and then logins again, one is redirected to the last pathName in the doesn't work when you logout.
Currently if there is history in the browser a user is unable to get redirected to a specifiic url after login.

**Steps to reproduce:**
1. Login using the useLogin("/redirectUrl") hook.
2. Navigate to a url different from the default one - "/" or the passed param "redirectUrl"
3. Logout
4. Login again through useLogin("/redirectUrl") 

As a result you are redirected to the url before logging out and not the redirectUrl path.

**Related code:**
```
inside the useLogin.ts
....
const login = useCallback(
        (params: any = {}, pathName = defaultAuthParams.afterLoginUrl) =>
            authProvider.login(params).then(ret => {
                // TODO - additional conditions
                history.push(nextPathName || pathName);
                return ret;
            }),
        [authProvider, history, nextPathName]
    );
```

**Suggestion:**
Check if pathName is not the default value and prioritize it before opting to nextPathName from history.
I suggest the following draft for the above piece of code where a // TODO is left:

```
const redirectPathName = pathName !== defaultAuthParams.afterLoginUrl 
                                                  ? pathName 
                                                  : nextPathName || pathName;
history.push(redirectPathName);
```

If I am not wrong in the way I understand how the hook works will be more to happy to make a PR.

djhi:
Thanks for reporting this. If you are able to illustrate the bug or feature request with an example, please provide a sample application by forking the following CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple).

ValentinnDimitroff:
Sure, here it is: (https://codesandbox.io/s/clever-antonelli-vhslz?file=/src/_test/Login.js) 

I have places a sample LoginPage component in the "_test "directory.  The useLogin() hook is used and as a second param I have passed the "/#/comments" path where I would like to redirect after login for example.

Steps to reporduce:

1. Login
2. Navigate to the "/#/posts" path or anything different from "/#/comments" .
3. Logout
4. Login

You are redirected to the last url before logging out and not the one passed explicitly in the LoginPage component.

djhi:
Reproduced. Thanks

[...]

PR #5248, opened 2020-09-12
Changed code to use pathName first (if exists) for redirect after login

bardeutsch:
Fixing issue #5204


jdemangeon (approved):

jdemangeon:
Merged! Thanks for your work!

PR #5258, opened 2020-09-15
allow to end undoable mutation directly in onSuccess callback

### LINKED ISSUE #5133: Allow optimistic mode exit immediately inside useMutate options.onSuccess callback
Hello, I recently played with `useMutate` hook more extensively. What I noticed that when using `options.undoable` option we have to provide `onSuccess` callback to `useMutate` hook - which is fine. What I had problem with is that I was unable to exit `optimistic` mode because even though I would dispatch `complete` action and call `undoableEventEmitter.emit('end', { isUndo: false })` my app would always stay in `optimistic` mode.

After some diggiging into source code i traced everything to this [two lines of code](https://github.com/marmelab/react-admin/blob/aeda86582e6ac1ea37f29739a8f08743694616c9/packages/ra-core/src/dataProvider/useDataProvider.ts#L292-L293). Problem is that `undoableEventEmitter.once` handler is registered after calling `onSuccess` callback. Which means that if your callback is sync and executes immediately undoableEventEmitter.emit will never register for that specific `mutate` call. 

I solved the problem by simply adding minimal timeout to my `onSuccess` callback:
```js
    onSuccess: () => {
        setTimeout(() => {
            undoableEventEmitter.emit('end', { isUndo: false })
        }, 100)
    }
```

Whit this `undoableEventEmitter.once`gets registered and after `100ms` the `emit` call is handled and app exits `optimistic` mode.

**Solution**
I think that `onSuccess` should be called after `undoableEventEmitter.once` de facto switching above mentioned two lines of code. This would allow to use `optimistic` mode and presenting (for example) updated UI to the user and then immediately exiting `optimistic` mode which would then do the API call.

I am open to implementing this in a pull request if needed.

ThieryMichel:
This allows to have optimistic behavior while sending the query immediately.
fix #5133

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -296,7 +296,6 @@ const performUndoableQuery = ({
             optimistic: true,
         },
     });
-    onSuccess && onSuccess({});
     undoableEventEmitter.once('end', ({ isUndo }) => {
         dispatch(stopOptimisticMode());
         if (isUndo) {
@@ -389,6 +388,8 @@ const performUndoableQuery = ({
             );
         }
     });
+    onSuccess && onSuccess({});
+
     return Promise.resolve({});
 };
 


djhi (commented):
It would be nice to have a test for that

fzaninotto:
As it's a new feature, please PR against `next`.

Also, we need some documentation to know how to set a call as optimistic without undo.

ThieryMichel:
The current modification is insufficient to have optimistic with no undo.
We lose the revert in case of server error.
Need further investigation.

fzaninotto:
Supersedd by #5799

PR #5263, opened 2020-09-16
Fix unexpected List Page Change

jdemangeon:
This fix avoid resetting current page (in a list) when the total is resetted (during loading phase).
This way, the demo "reviews" page don't go to the first page when we accept or reject a review.

I also removed the redirect to "/reviews" when we accept or reject reviews in bulk mode. This way we stay on the current page of reviews

### DIFF
--- a/packages/ra-core/src/controller/useListController.ts
+++ b/packages/ra-core/src/controller/useListController.ts
@@ -180,11 +180,16 @@ const useListController = <RecordType = Record>(
         get(state.admin.resources, [resource, 'list', 'total'], 0)
     );
 
+    // Since the total can be empty during the loading phase
+    // We need to override that total with the latest loaded one
+    // This way, the useEffect bellow won't reset the page to 1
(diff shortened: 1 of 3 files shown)


djhi (approved):
:+1:

packages/ra-core/src/controller/useListController.ts
```diff
@@ -180,11 +180,12 @@ const useListController = <RecordType = Record>(
         get(state.admin.resources, [resource, 'list', 'total'], 0)
     );
 
+    const finalTotal = total || defaultTotal;
```
fzaninotto, line 183: total can be 0, use typeof total === 'undefined' instead

fzaninotto:
Can you please add a description? It seems this solves 2 different bugs

packages/ra-core/src/controller/useListController.ts
jdemangeon: You're right!

jdemangeon:
Review applied

PR #5270, opened 2020-09-18
Fixes #5168 - Pagination broke after e6357bc

### LINKED ISSUE #5168: Pagination broke after e6357bc
After: https://github.com/marmelab/react-admin/commit/e6357bcc95f53d588a8b44073c1dc22fcbcc9b85
Specifically: https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/useListController.ts#L187-L189

Broke my apps pagination. I wasn't able to investigate too deeply but I think the root cause may have been that I mutate params inside it's `buildQuery`, so I think there may be a cache mismatch or something. `defaultTotal` was correct, `total` was undefined in the new change, forcing the page to continually `setPage(1)` below.

**What you were expecting:**
No breaking change from 3.7.1 to 3.8.1

**What happened instead:**
Pagination broke

**Steps to reproduce:**
I can't give a full reproduction presently, but a `dataProvider`'s `buildQuery` which mutates `params` of a `GET_LIST`

**Related code:**
Specifically: https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/useListController.ts#L187-L189

**Environment**

* React-admin version: 3.8.1
* Last version that did not exhibit the issue (if applicable): 3.7.1
* React version: ^16.9.0
* Browser: Chrome

RobertLowe:
Pretty straight forward fix. :neutral_face: :broken_heart:

### DIFF
--- a/packages/ra-core/src/controller/useListController.ts
+++ b/packages/ra-core/src/controller/useListController.ts
@@ -173,18 +173,22 @@ const useListController = <RecordType = Record>(
     // useGetList hook again. While the result of this new call is loading,
     // the ids and total are empty. To avoid rendering an empty list at that
     // moment, we override the ids and total with the latest loaded ones.
-    const defaultIds = useSelector((state: ReduxState): Identifier[] =>
-        get(state.admin.resources, [resource, 'list', 'ids'], [])
+    const defaultIds = useSelector(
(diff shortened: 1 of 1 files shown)


djhi:
Thanks! However, #5263 is already fixing that

PR #5275, opened 2020-09-19
Added buttonProps prop to FilterButton

maksimkuznetsov:
Added the ability to customize the "Add Filter" button.


Luwangel (changes requested):
Hi, thanks for contributing :)

In react-admin, we prefer to pass an element instead of properties when it's possible. In your case, I suggest passing the entire `<Button>`. Could you make the change?

You can take a look to the `<Layout>` that uses this mechanism. https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/Layout.tsx

Have a good day,
Adrien

fzaninotto:

Issue #5268, opened 2020-09-17
LoadingIndicator does not get class overrides


danejorgensen:
**What you were expecting:**
Can apply custom theme to LoadingIndicator component

**What happened instead:**
LoadingIndicator is default color.   Does not pass styles to RefreshIconButton

It does pass the styles to the CircularProgress component while isLoading is true.

**Steps to reproduce:**
```
import { LoadingIndicator } from 'react-admin';

theme => ({
  indicator: {
    color: '#F00'
  }
})

<LoadingIndicator className={classes.indicator} />
```

**Environment**

* React-admin version: 3.8.4
* Last version that did not exhibit the issue (if applicable):
* React version:
* Browser:
* Stack trace (in case of a JS error):

djhi:
Thanks for reporting. Indeed only the `CircularProgress` is passed a `className`.

Luwangel:
This feature will be part of the next release. You'll have to overide the classes passed to the `LoadingIndicator` using the key `loadedIcon`.

matboud:
just wrapp it inside a div and pass the styles you want to that div, or inside the LoadingIndicator.d.ts file add this line "const style: PropTypes.Requireable<object>; " in the namespace prototype and you u ll be fine to go

PR #5279, opened 2020-09-21
Improve LoadingIndicator component styling

WiXSL:
Implements #5268

### DIFF
--- a/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
+++ b/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
@@ -13,7 +13,7 @@ const useStyles = makeStyles(
         loader: {
             margin: 14,
         },
-        icon: {},
+        loadedIcon: {},
     },
     { name: 'RaLoadingIndicator' }
 );
@@ -32,7 +32,7 @@ const LoadingIndicator = props => {
             {...rest}
         />
     ) : (
-        <RefreshIconButton className={classNames(classes.icon, className)} />
+        <RefreshIconButton className={classes.loadedIcon} />
     );
 };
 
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/LoadingIndicator.js
```diff
@@ -31,7 +32,7 @@ const LoadingIndicator = props => {
             {...rest}
         />
     ) : (
-        <RefreshIconButton />
+        <RefreshIconButton className={classNames(classes.icon, className)} />
```
djhi, line 35: I'm afraid passing the same className as the progress component can have strange effects. Not sure how to handle this though

Luwangel: The only solution to avoid possible breaking changes is to remove the classname passed to the refresh icon. This should not be a problem for users because as customisation of this icon was not possible before, they will only use the classes override.

packages/ra-ui-materialui/src/layout/LoadingIndicator.js
```diff
@@ -13,6 +13,7 @@ const useStyles = makeStyles(
         loader: {
             margin: 14,
         },
+        icon: {},
```
Luwangel, line 16: By reading the word `icon`, we can believe that we'll change the loading indicator. We should probably rename it to `loadedIcon` in order to be more explicit.

fzaninotto:
Sorry, this was closed automatically by GitHub when I merged the `next` branch into `master`. I'm reopening it.

packages/ra-ui-materialui/src/layout/LoadingIndicator.js
WiXSL: ok

packages/ra-ui-materialui/src/layout/LoadingIndicator.js
WiXSL: Done!

Luwangel (approved):

PR #5290, opened 2020-09-22
Add documentation of Data Provider

zachrybaker:
Added a link to a data provider I added today, that is simply taking the ra-data-simple-rest provider and giving it a  configuration hash of resourceName:identityName values so that it can map between 'id' and the real identity property name in HTTP communication.  Simple.  And React-admin continues to work as it always has, but people get to avoid changing their server-side API.


djhi (approved):
Nice job. Thanks!

Issue #4933, opened 2020-06-11
Allow more configuration in introspection to determine which types should be react-admin resources


bayareacoder:
**Is your feature request related to a problem? Please describe.**
To detect a GraphQL type as a resource, the current introspection requires that there is a query in the GraphQL API with a singular name of the resource AND a plural name. While you can set the operation names of these queries in an option, you cannot change the requirement that BOTH need to be present for a type to be detected as a resource due to this fixed function in `introspection.js`:

```
  const isResource = (type) =>
    queries.some(
      (query) => query.name === options.operationNames[GET_LIST](type)
    )
    &&
    queries.some(
      (query) => query.name === options.operationNames[GET_ONE](type)
   );
```
Any whitelisting of resources via `include` is only applied in `filterTypesByIncludeOrExclude` AFTER the above. This is a problem with APIs that do not support a plural or singular version of a query.

**Describe the solution you'd like**
The `include` list should override any filtering done by `isResource`

**Describe alternatives you've considered**
You could pass in a completely custom `resolveIntrospection` to fix this, but seems overkill

**Additional context**
This goes together with issue#4932 to allow multiple GraphQL queries for a single react-admin verb. For instance, both changes are needed to support GraphQL APIs that support `GET_MANY` or event `GET_LIST` only from multiple `GET_ONE` fetches (eg in case all `id`s are known outside the API) as they would not have a plural query, and then require multiple queries to provide the data required for `GET_LIST` or `GET_MANY`

djhi:
The solution is to build your own GraphQL dataProvider. Use the existing ones as starting points.

bayareacoder:
Sure but the proposal is to make small changes to `ra-data-graphql` so it can be used as-is with these kind of APIs. The change is simply to `isReource`, in file `introspection.js`:
```
  const isResource = (type) => {
    // whitelist any type that is in `include` list
    if (filterTypesByIncludeExclude(options)) return true;
    // otherwise we require there is a GraphQL query for both the operation name of GET_LIST and GET_ONE in order
    // for the type to be a resource
    return (
      queries.some(
        (query) => query.name === options.operationNames[GET_LIST](type)
      ) &&
      queries.some(
        (query) => query.name === options.operationNames[GET_ONE](type)
      )
    );
  }
```

Kilometers42:
[...]

PR #5305, opened 2020-09-25
ra-data-graphql: Update include/exclude introspection logic

Kilometers42:
See https://github.com/marmelab/react-admin/issues/4933 for background info. But, right now all resources in the GQL data provider must implement both a GET_ONE and a GET_LIST for a resource to actually be considered a resource. This is very limiting because there is no workaround for a resource that only has a GET_ONE. 

What this PR does is move the include/exclude logic to before the check that a resource has a GET_ONE and GET_LIST. If a resource is included or excluded, it will short circuit and either include or exclude the resource, regardless of the GET_ONE/GET_LIST check. 

This solution is pretty solid in my opinion -- it is totally backward compatible, it allows users who want to include a resource that is a bit nonstandard to do so by explicitly setting a configuration object, and it feels more intuitive that a resource which is explicitly included in the configuration object is always included.

### DIFF
--- a/packages/ra-data-graphql/src/introspection.js
+++ b/packages/ra-data-graphql/src/introspection.js
@@ -4,24 +4,28 @@ import { GET_LIST, GET_ONE } from 'ra-core';
 
 import { ALL_TYPES } from './constants';
 
-export const filterTypesByIncludeExclude = ({ include, exclude }) => {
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5311, opened 2020-09-28
Allow the option labels of NullableBooleanInput to be customized

gavacho:
Hi, I would like to customize the labels of the options of the `NullableBooleanInput` control.

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -450,28 +450,18 @@ import { NullableBooleanInput } from 'react-admin';
 
 ![NullableBooleanInput](./img/nullable-boolean-input.png)
 
-`<NullableBooleanInput />` doesn't display the empty option by default. If you want to customize its label and display it, you can use the `displayNull` prop.
+The labels of the options can be customized using the `nullLabel`, `falseLabel` and `trueLabel` properties.
 
 ```jsx
 import { NullableBooleanInput } from 'react-admin';
 
-<NullableBooleanInput 
+<NullableBooleanInput
     label="Commentable"
     source="commentable"
-    displayNull
+    nullLabel="Either"
+    falseLabel="No"
+    trueLabel="Yes"
 />
-```
-
-Also you need to provide your own label for null value.
-
-```jsx
-import polyglotI18nProvider from 'ra-i18n-polyglot';
-import englishMessages from 'ra-language-english';
-
-englishMessages.ra.boolean.null = 'Null label';
-const i18nProvider = polyglotI18nProvider(() => englishMessages, 'en');
-
-<Admin i18nProvider={i18nProvider}></Admin>
 ```
 
 ![NullableBooleanInput](./img/nullable-boolean-input-null-label.png)
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -450,7 +450,21 @@ import { NullableBooleanInput } from 'react-admin';
 
 ![NullableBooleanInput](./img/nullable-boolean-input.png)
 
-The labels of the options can be customized using the `nullLabel`, `falseLabel` and `trueLabel` properties.
+The labels of the options can be customized for the entire application by overriding the translation.
+
+```jsx
+import polyglotI18nProvider from 'ra-i18n-polyglot';
+import englishMessages from 'ra-language-english';
+
+englishMessages.ra.boolean.null = 'Null label';
+englishMessages.ra.boolean.false = 'False label';
+englishMessages.ra.boolean.true = 'True label';
+const i18nProvider = polyglotI18nProvider(() => englishMessages, 'en');
+
+<Admin i18nProvider={i18nProvider}></Admin>
+```
+
+Additionally, individual instances of `NullableBooleanInput` may be customized by setting the `nullLabel`, `falseLabel` and `trueLabel` properties. Values specified for those properties will be translated by react-admin.
 
 ```jsx
 import { NullableBooleanInput } from 'react-admin';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -108,12 +101,12 @@ const NullableBooleanInput: FunctionComponent<
...
-            {...enhancedOptions}
+            {...options}
             {...sanitizeRestProps(rest)}
         >
-            <MenuItem value="">{translate('ra.boolean.null')}</MenuItem>
-            <MenuItem value="false">{translate('ra.boolean.false')}</MenuItem>
-            <MenuItem value="true">{translate('ra.boolean.true')}</MenuItem>
+            <MenuItem value="">{translate(nullLabel)}</MenuItem>
+            <MenuItem value="false">{translate(falseLabel)}</MenuItem>
+            <MenuItem value="true">{translate(trueLabel)}</MenuItem>
```
gavacho, line 109: Should I keep these call to `translate` here? Or should consumers of this component be expected to translate their own values?

packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -70,20 +77,6 @@ const NullableBooleanInput: FunctionComponent<
         validate,
     });
 
-    const enhancedOptions = displayNull
-        ? {
-              ...options,
-              SelectProps: {
-                  displayEmpty: true,
```
gavacho, line 77: The property is named `allowEmpty` so this has no effect. This component is already adding a "null" option.

fzaninotto:
Why don't you want to use the current solution (using translation messages)?

gavacho:
@fzaninotto I have two different NullableBooleanInputs in my app and want to use different labels on them.

gavacho:
@fzaninotto The default behavior of this control is to display 3 items, the item representing the null item has a blank label so the options, while present in the dropdown, is very difficult to select.

This has created confusion in the past. For example, this PR https://github.com/marmelab/react-admin/pull/4365 is addressing a problem that doesn't exist (because the null option is always rendered by NullableBooleanInput).

Maybe NullableBooleanInput should guard against the case of the translation message being an empty string.

fzaninotto:
PR #5326 should solve the bug of the null option not being visible enough.

packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
fzaninotto: you should keep the calls to `translate`

packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -30,7 +30,12 @@ const getStringFromBoolean = (value?: boolean | null): string => {
 };
 
 const NullableBooleanInput: FunctionComponent<
-    InputProps<TextFieldProps> & Omit<TextFieldProps, 'label' | 'helperText'>
+    InputProps<TextFieldProps> &
```
fzaninotto, line 33: this type is becoming increasingly complex. Anyone trying to extend `NullableBooleanInput` in TypeScript will fight trying to type its props. 

Please create a custom `NullableBooleanInputProps` interface and export it.

docs/Inputs.md
```diff
@@ -450,30 +450,20 @@ import { NullableBooleanInput } from 'react-admin';
...
 import { NullableBooleanInput } from 'react-admin';
 
-<NullableBooleanInput 
+<NullableBooleanInput
     label="Commentable"
     source="commentable"
-    displayNull
+    nullLabel="Either"
+    falseLabel="No"
+    trueLabel="Yes"
```
fzaninotto, line 463: You should mention that these labels are translated by react-admin.

docs/Inputs.md
```diff
@@ -450,30 +450,20 @@ import { NullableBooleanInput } from 'react-admin';
 
 ![NullableBooleanInput](./img/nullable-boolean-input.png)
 
-`<NullableBooleanInput />` doesn't display the empty option by default. If you want to customize its label and display it, you can use the `displayNull` prop.
+The labels of the options can be customized using the `nullLabel`, `falseLabel` and `trueLabel` properties.
```
fzaninotto, line 453: This is a bit confusing: the doc should explain that in order to customize the labels for the entire app, the right way is to override the translation. Custom labels are only necessary if one wants to have distinct labels for one given input.

packages/ra-ui-materialui/src/input/NullableBooleanInput.spec.tsx
```diff
@@ -138,4 +138,70 @@ describe('<NullableBooleanInput />', () => {
             'true'
         );
     });
+
+    it('should allow to customize the label of the null option', () => {
```
fzaninotto, line 142: 👍 Excellent, thanks for thinking of adding tests!

fzaninotto (changes requested):

gavacho:
Oops, I just realized this was opened against the `master` branch upstream. Shall I re-open against `next`?

djhi:
> Oops, I just realized this was opened against the master branch upstream. Shall I re-open against next?

The next release will be 3.9 and we already moved the related code to master so you're fine :)

fzaninotto (approved):

fzaninotto:
Perfect, thanks a lot!

PR #5312, opened 2020-09-28
Fix TabbedShowLayout resolves path incorrectly if first tab is null

### LINKED ISSUE #5294: TabbedShowLayout component resolves path incorreclly if first tab is null
Trying to set an optional `Tab` component as the first Tab inside a `TabbedShowLayout` component.
**Note**: I solve it by setting a custom path to every Tab and making my own ShowButton.

Test to reproduce the problem in `ra-ui-materialui` package:

The test fails and you get a console error.
`console.error ../../node_modules/@material-ui/core/Tabs/Tabs.js:221
    Material-UI: The value provided to the Tabs component is invalid.
    None of the Tabs' children match with `/`.
    You can provide one of the following values: //1, //2.`

If you remove the `{null}` line, the test pass without errors.

```jsx
import expect from 'expect';
import { cleanup, render } from '@testing-library/react';

import TabbedShowLayout from './TabbedShowLayout';
import Tab from './Tab';
import TextField from '../field/TextField';
import { createMemoryHistory } from 'history';
import { Router } from 'react-router-dom';

describe('<TabbedShowLayout />', () => {
    afterEach(cleanup);

    const renderWithRouter = children => {
        const history = createMemoryHistory();

        return {
            history,
            ...render(<Router history={history}>{children}</Router>),
        };
    };

    it('should display the first valid Tab component and its content', () => {
        const { queryByText, history } = renderWithRouter(
            <TabbedShowLayout basePath="/" record={{ id: 123 }} resource="foo">
                {null}
                <Tab label="Tab1">
                    <TextField label="Field On Tab1" source="field1" />
                </Tab>
                <Tab label="Tab2">
                    <TextField label="Field On Tab2" source="field2" />
                </Tab>
            </TabbedShowLayout>
        );

        expect(queryByText('Tab1')).not.toBeNull();
        expect(queryByText('Field On Tab1')).not.toBeNull();
    });
});

### PR DESCRIPTION
**Pull request by WiXSL**
Fixes #5294

### DIFF
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -98,8 +98,8 @@ const TabbedShowLayout = props => {
 
             <Divider />
             <div className={classes.content}>
-                {Children.map(nonNullChildren, (tab, index) => {
-                    return tab && isValidElement(tab) ? (
+                {Children.map(nonNullChildren, (tab, index) =>
+                    tab && isValidElement(tab) ? (
                         <Route
                             exact
                             path={escapePath(
@@ -114,8 +114,8 @@ const TabbedShowLayout = props => {
                                 })
                             }
                         />
-                    ) : null;
-                })}
+                    ) : null
+                )}
             </div>
         </div>
     );
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
@@ -88,17 +88,17 @@ const TabbedShowLayout = props => {
     } = props;
     const match = useRouteMatch();
     const classes = useStyles(props);
-    const finalChildren = Array.isArray(children)
-        ? children.filter(child => child !== null)
-        : children;
+    const nonNullChildren = Children.toArray(children).filter(
+        child => child !== null
+    );
 
     return (
         <div className={className} key={version} {...sanitizeRestProps(rest)}>
-            {cloneElement(tabs, {}, finalChildren)}
+            {cloneElement(tabs, {}, nonNullChildren)}
 
             <Divider />
             <div className={classes.content}>
-                {Children.map(finalChildren, (tab, index) => {
+                {Children.map(nonNullChildren, (tab, index) => {
                     return tab && isValidElement(tab) ? (
                         <Route
                             exact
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/ra-ui-materialui/src/detail/TabbedShowLayout.js [hunk 1]
```diff
@@ -87,17 +87,19 @@ const TabbedShowLayout = props => {
         ...rest
     } = props;
     const match = useRouteMatch();
-
     const classes = useStyles(props);
+    const finalChildren = Array.isArray(children)
```
fzaninotto, line 91: Use React.Children.toArray, because the children prop data structure isn't documented and may change in the future. 

Also, I'd call the variable `nonNullChildren`

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -87,17 +87,19 @@ const TabbedShowLayout = props => {
...
+        ? children.filter(child => child !== null)
+        : children;
 
     return (
         <div className={className} key={version} {...sanitizeRestProps(rest)}>
-            {cloneElement(tabs, {}, children)}
+            {cloneElement(tabs, {}, finalChildren)}
```
fzaninotto, line 97: what's the problem with cloning null children?

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
```diff
@@ -87,17 +87,19 @@ const TabbedShowLayout = props => {
...
 
             <Divider />
             <div className={classes.content}>
-                {Children.map(children, (tab, index) =>
-                    tab && isValidElement(tab) ? (
+                {Children.map(finalChildren, (tab, index) => {
```
fzaninotto, line 101: here, you can use the regular `children`, right? This avoids having a different value at each render, and defeating pure component optimizations.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
WiXSL: Your suggestion passes the test but throw console errors since you end up with a different amount of headers and content if you have a `null` child

packages/ra-ui-materialui/src/detail/TabbedShowLayout.js
WiXSL: I don't understand, if I go back to using `Children.map(children, (tab, index) =>` the tests won't pass.
Should I filter `nonNullChildren` with `useCallback`?

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5313, opened 2020-09-28
More TS fixes

djhi:
Many of the props are now optional because they are injected by react-admin. This was preventing usage such as passing a `Show` view inside an `expand`

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Create.tsx
+++ b/packages/ra-ui-materialui/src/detail/Create.tsx
@@ -215,17 +215,17 @@ const useStyles = makeStyles(
 const sanitizeRestProps = ({
     hasCreate = null,
     hasEdit = null,
-    history,
-    loaded,
-    loading,
-    location,
-    match,
+    history = null,
+    loaded = null,
+    loading = null,
+    location = null,
+    match = null,
     onFailure = null,
     onSuccess = null,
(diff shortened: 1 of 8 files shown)


Luwangel (approved):

packages/ra-ui-materialui/src/detail/Create.tsx
```diff
@@ -226,6 +226,6 @@ const sanitizeRestProps = ({
     permissions = null,
     transform = null,
     ...rest
-}) => rest;
+}: any) => rest;
```
fzaninotto, line 229: I'm not a fan of this: setting a null default value should catch wrong type alerts, but we need typing here to catch true alerts for props that should not appear here.

fzaninotto (changes requested):
please don't use `any` in existing `sanitizeRestProps` params signature

packages/ra-ui-materialui/src/detail/Create.tsx
djhi: we already use any in most `sanitizeRestProps` functions

djhi: Fixed it in all components

fzaninotto (approved):

PR #5328, opened 2020-09-30
Fix theming docs UserMenu example

ceracera:
I was getting this error when trying UserMenu customizaton:

`Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`

I changed how MyAppBar is passed to Layout. From `{<MyAppBar />}` to `{MyAppBar}` Seems ok now. It looks like Layout is not expecting component that way.


djhi (approved):
Thanks!

PR #5329, opened 2020-09-30
Add localStorage data provider

fzaninotto:
An admin can use localStorage for part of or all the underlying data. A localStorage dataProvider opens the possibility to build local-first admins.

```jsx
import localStorageDataProvider from 'ra-data-local-storage';
const dataProvider = localStorageDataProvider();
```

Uses `ra-data-fakerest` under the hood, synchronizing changes between the FakeRest (in-memory) and localStorage databases. It also hides warnings for undefined collections, and allows the addition of a record to an undefined collection.

Note that localStorage is limited to 5MB in size.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,10 @@ build-ra-data-json-server:
 	@echo "Transpiling ra-data-json-server files...";
 	@cd ./packages/ra-data-json-server && yarn -s build
 
+build-ra-data-localstorage:
+	@echo "Transpiling ra-data-localstorage files...";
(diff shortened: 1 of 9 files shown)


djhi:
I don't really get why it needs fakerest. Can you elaborate ?

fzaninotto:
To avoid duplicating the code needed to handle dataProvider requests in an in memory database. Basically this entire file:

https://github.com/marmelab/FakeRest/blob/master/src/Collection.js

djhi (approved):

PR #5340, opened 2020-10-02
Fix missing link to <DateTimeInput> in reference documentation

Hettomei:
I always reach https://marmelab.com/react-admin/Reference.html . Not finding DateTimeInput here was confusing.

Thanks for this package.


fzaninotto (approved):

fzaninotto:
Good catch, thatnks!

PR #5342, opened 2020-10-02
Convert Tab component to TypeScript

### LINKED ISSUE #5334: Tab component Type Definition is incorrect
**What you were expecting:**
Should be able to use `<Tab label="summary">` as shown in the documentation.

**What happened instead:**
Looks like all the possible Tab props are required:

```
TS2740: Type '{ children: Element[]; label: string; }' is missing the following properties from type '{ [x: string]: any; basePath: any; children: any; contentClassName: any; context: any; className: any; icon: any; label: any; record: any; resource: any; value: any; }': basePath, contentClassName, context, className, and 4 more.
```

Also, it would be nice if that list of props was available as an exported TabProp interface.

**Steps to reproduce:**
Try creating the TabbedShowLayout example from the docs, should be enough to show the issue.

* React-admin version: 3.9.0-beta.3
* React version: 16.13.1

fzaninotto:
Closes #5334

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Tab.tsx
+++ b/packages/ra-ui-materialui/src/detail/Tab.tsx
@@ -1,21 +1,12 @@
 import * as React from 'react';
-import { isValidElement } from 'react';
+import { isValidElement, ReactElement, ReactNode } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


Luwangel (approved):
Perfect!

PR #5345, opened 2020-10-03
Remove unused components css rules

WiXSL:


packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
```diff
@@ -24,7 +24,6 @@ const sanitizeRestProps = ({
 
 const useStyles = makeStyles(
     theme => ({
-        root: {},
```
fzaninotto, line 27: For this one, we should use it in the JSX root element instead of removing it.

fzaninotto (changes requested):
Good cleanup!

djhi (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5351, opened 2020-10-05
Fix useAuthState hook docs

WiXSL:


packages/ra-core/src/auth/useAuthState.ts
```diff
@@ -12,20 +12,15 @@ interface State {
...
  *
- * The return value updates according to the authProvider request state:
- *
- * - start:   { authenticated: false, loading: true, loaded: false }
- * - success: { authenticated: true,  loading: false, loaded: true }
- * - error:   { authenticated: false, loading: false, loaded: true }
- *
- * Useful in custom page components that can work both for connected and
- * anonymous users. For pages that can only work for connected users,
- * prefer the useAuthenticated() hook.
```
djhi, line 28: I think we should keep the initial format:

> Hook for ...
> Calls the dataProvider
> State changes
> Why and when to use it

djhi (changes requested):

packages/ra-core/src/auth/useAuthState.ts
WiXSL: Done. I hope is ok. Let me know

WiXSL:
I've added a fix for a typo in the JDoc of useCheckAuth as well. I hope is ok

djhi (changes requested):
:ok_hand: Thanks!

djhi (approved):

PR #5360, opened 2020-10-06
Allow MenuItemLink to receive TooltipPops

### LINKED ISSUE #5359: Allow MenuItemLink to have a customizable and optional tooltip
**Problem description**
I added a few MenuItemLinks in my Customized Appbar but then I wanted to customized the tooltips (I don't want them or I want to change the content), it turns that the configuration for that component was hardcoded to display the same than the primary title and to be displayed on the right side.

**The solution I'd like**
I would like to be able to pass the props to the MenuItemLink with the tooltip options, and if I don't pass any, I don't want the tooltip to be rendered.

**Alternatives I've considered**
I've considered using another component apart from the MenuItemLink, but this component is configured out of the box to keep track of the current route.

Here is a [PR](https://github.com/marmelab/react-admin/pull/5360)

gstn-caruso:
Closes #5359

### DIFF
--- a/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
+++ b/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
@@ -12,7 +12,7 @@ import { StaticContext } from 'react-router';
 import { NavLink, NavLinkProps } from 'react-router-dom';
 import MenuItem, { MenuItemProps } from '@material-ui/core/MenuItem';
 import ListItemIcon from '@material-ui/core/ListItemIcon';
-import Tooltip from '@material-ui/core/Tooltip';
+import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
 import { makeStyles } from '@material-ui/core/styles';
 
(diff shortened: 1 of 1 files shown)


djhi (changes requested):
Looks good. However, we'd like the tooltip to be displayed by default as before (using the primaryText). 

Can you please update the PR so that you can disable the tooltip by passing `false` a new `tooltip` prop? This prop should accept a label and defaults to the `primaryText` if undefined.

fzaninotto:
I don't think that disabling tooltip is a good feature. When the sidebar is closed, users need the tooltip.

So please limit your PR to allowing custom tooltip props, not removing the tooltip.

gstn-caruso:
Sure, let me change that 👌.

Despite that change I have a question @fzaninotto, is this element supposed to be used only in the sidebar? I tried to put it in a custom Appbar (that's why I wanted to remove the tooltip, or at least customize it)

fzaninotto:
Ys, it's only designed to be used in the Sidebar

fzaninotto:
Any updates on this one?

PR #5364, opened 2020-10-07
add open in Gitpod button + update config to open preview of port 8080 on workspace startup.

nisarhassan12:
This Pr updates the config to open the preview of port 8080 on Gitpod workspace startup with this change in place a workspace once started will look like the following:

image

Also I have added a Open in Gitpod button + a bit of description in the Readme so that more people can benefit from it.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,3 +1,6 @@
 tasks:
   - init: make install
     command: make run
+ports: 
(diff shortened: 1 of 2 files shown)


djhi (approved):

djhi:
Thanks!

PR #5380, opened 2020-10-11
Strip down lodash footprint by loading only the required utility

impronunciable:
Importing `lodash` instead of `lodash/merge` was having an impact in the `ra-core` import size for `examples/demo`.

This is `lodash` footprint before the change:

Screen Shot 2020-10-11 at 11 27 33

And after:

Screen Shot 2020-10-11 at 11 30 02

### DIFF
--- a/packages/ra-core/src/form/useInitializeFormWithRecord.ts
+++ b/packages/ra-core/src/form/useInitializeFormWithRecord.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'react';
 import { useForm } from 'react-final-form';
-import { merge } from 'lodash';
+import merge from 'lodash/merge';
 
 /**
  * Restore the record values which should override any default values specified on the form.


fzaninotto:
The import you changed is only one of many. You're supposed to use a bundler with tree-shaking ability to minimize your bundle size.

impronunciable:
Actually it was the only one importing the main lodash module. That's why I got the file size reduction. I think this would be beneficial for every ra user even if they don't use a bundler with tree shaking

fzaninotto:
You're right, all the other imports use the alternative syntax. So not only does your PR improve bundle size for those not using a bundler with tree shaking, it also makes the code more consistent. Let me reopen and merge it.

fzaninotto:
Thanks for this patch. Note that using a bundler with tree shaking will reduce the bundle size even more (because we use mostly the `import { XXX } from ...` syntax for material-ui).

impronunciable:
Thank you and thanks for mantaining the project!

PR #5385, opened 2020-10-12
[Doc] Fix ts code to js code in examples

WiXSL:


Luwangel (approved):
Thanks! You're right, we should display the examples in JS :)

PR #5386, opened 2020-10-12
Add missing "path" type for Tab component

nickwaelkens:
I noticed the Tab component was missing a type for the `path` prop, which still works fine but makes TypeScript complain.


fzaninotto (approved):

fzaninotto:
Indeed, thanks!

PR #5390, opened 2020-10-12
[Doc] Fix specialized dataProvider hooks usage

### LINKED ISSUE #4325: Undocumented difference between useQuery and useGetList Hooks
Hello,

When retrieving a list of entities with the useQuery hook, the data property of the response is an array of entites. But when using useGetList (or also useGetManyReference) the response has a data property which is an object with id -> entity pairs, and an ids property which includes all the IDs.

```
 const { data } = useQuery({
    type: 'getList',
    resource,
    payload: {pagination, sort, filter}
})
```

=> data is an array of objects.

```
 const { data, ids } = useGetList(
    resource,
    pagination,
    sort,
    filter,
  )
```

=> data is on object like `{ 1: entity, 2: entity, ... }` and ids is an array with IDs like `[1, 2, ...]``

I was wondering, why is it like this?
And maybe it should be documented :) (or maybe it is, but I didn't find it)

fzaninotto:
Closes #4325

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -235,20 +235,178 @@ const ApproveButton = ({ record }) => {
 };
 ```
 
-The specialized hooks based on `useQuery` execute on mount:
+The specialized hooks based on `useQuery` (`useGetList`, `useGetOne`, `useGetMany`, `useGetManyReference`) execute on mount. The specialized hooks based on `useMutation` (`useCreate`, `useUpdate`, `useUpdateMany`, `useDelete`, `useDeleteMany`) return a callback.
 
-* `useGetList(resource, pagination, sort, filter, options)`
-* `useGetOne(resource, id, options)`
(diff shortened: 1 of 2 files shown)


docs/Actions.md
```diff
@@ -235,20 +235,178 @@ const ApproveButton = ({ record }) => {
...
+         </ul>
+     );
+};
+```
+
+### `useGetManyReference`
+
+```jsx
+// syntax
+const { data, ids, total, loading, loaded, error } = useGetManyReference(resource, target, id, pagination, sort, filter, referencingResource, options)
```
Luwangel, line 308:
```suggestion
const { data, ids, total, loading, loaded, error } = useGetManyReference(resource, target, id, pagination, sort, filter, referencingResource, options);
```

Luwangel (commented):
Great!

PR #5402, opened 2020-10-14
Fix non null id types handling

### LINKED ISSUE #5400: Edit screen doesn't work with ra-data-graphql-simple if resource id is of type Int!
**What you were expecting:**

given this schema simplified schema:

```
type Query {
  Entity(id: Int!): Entity
}

type Entity {
  id: Int!
}
```

I expect to be able to click on a row of a DataGrid displaying the list of Entity and seeing an Edit screen for Entity.

**What happened instead:**

We get an error `Element does not exist` and we a redirected to the list of Entity. The error message is incorrect, the real error is that the `id` variable is given as a string instead of a number.

**Other information:**

The issue is caused by the `castType` function (buildVariable.js:27). This function should cast the `id` variable to type string but it doesn't work with non nullable types.

In the `castType` function, we should check if the type's `kind` is `NON_NULL`, and in this case use `type.ofType` to discover the real type of the argument.

**Environment**

* React-admin version: 3.9.2
* React version: 16.13.1
* Browser: Chrome

EmrysMyrddin:
If the `id` field of an entity has a non null type like `Int!`, it cannot be edited with the `Edit` screen.

The problem was the `castType` function not handling types with `NON_NULL` kind which wraps the real type under a `ofType` property.

closes #5400

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildVariables.js
+++ b/packages/ra-data-graphql-simple/src/buildVariables.js
@@ -25,7 +25,8 @@ const sanitizeValue = (type, value) => {
 };
 
 const castType = (value, type) => {
-    switch (`${type.kind}:${type.name}`) {
+    const realType = type.kind === 'NON_NULL' ? type.ofType : type;
+    switch (`${realType.kind}:${realType.name}`) {
         case 'SCALAR:Int':
             return Number(value);
 


djhi (approved):
Thanks!

PR #5403, opened 2020-10-14
Add a GET_COUNT fetch action for graphql dataProvider

### LINKED ISSUE #4566: Add the GET_COUNT data fetch action
GraphQL implementation is based on the assumption that count requests has same input type as list requests. Also all the list requests must has same input type. Else we need to make multiple count query fields - separated query per each list request. Ans one more count field with different filter type.

All this tasks may be solved in custom dataProvider when resource will has **count fetch action**. But without information about count fields and its data in introspection the data provider implementation gains meaningless complexity and instability.

To solve the problem I have to suggest to add **`GET_COUNT`** to the [dataFetchActions.ts](https://github.com/marmelab/react-admin/blob/v3.3.1/packages/ra-core/src/core/dataFetchActions.ts) and also to query types list at [ra-data-graphql/src/constants.js](https://github.com/marmelab/react-admin/blob/v3.3.1/packages/ra-data-graphql/src/constants.js#13)

This will significantly simplify implementation of the dataProvider and increase customization level at all.

EmrysMyrddin:
In the graphql provider, we have to define a query to fetch the total count of entity for a given resource. Previously, the name of this query was hardcoded.

This PR aims to allow the customization of this query name, such as every other query names.

I'm not sure about where to put the new `GET_COUNT` constant, tell me if you think there is a better place.

I have exported `GET_COUNT` from `ra-data-graphql-simple` to allow an easier use for developers, allowing to import this simbol directly from `ra-data-graphql-simple` instead of having an explicit dependence on `ra-data-graphql` in the application code.

Closes #4566

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildGqlQuery.js
+++ b/packages/ra-data-graphql-simple/src/buildGqlQuery.js
@@ -1,5 +1,5 @@
 import { GET_LIST, GET_MANY, GET_MANY_REFERENCE, DELETE } from 'ra-core';
-import { QUERY_TYPES } from 'ra-data-graphql';
+import { QUERY_TYPES, GET_COUNT } from 'ra-data-graphql';
 import { TypeKind } from 'graphql';
 import * as gqlTypes from 'graphql-ast-types-browser';
 
@@ -158,16 +158,29 @@ export const buildApolloArgs = (query, variables) => {
     return args;
 };
 
+function getCountType(resource) {
+    const countType = resource[GET_COUNT];
+
+    if (!countType) {
+        throw new Error(
+            `No query or mutation matching fetch type ${GET_COUNT} could be found for resource ${resource.type.name}`
+        );
+    }
+
+    return countType;
+}
+
 export default introspectionResults => (
     resource,
     aorFetchType,
     queryType,
     variables
 ) => {
+    const countType = getCountType(resource);
     const { sortField, sortOrder, ...metaVariables } = variables;
     const apolloArgs = buildApolloArgs(queryType, variables);
     const args = buildArgs(queryType, variables);
-    const metaArgs = buildArgs(queryType, metaVariables);
+    const metaArgs = buildArgs(countType, metaVariables);
     const fields = buildFields(introspectionResults)(resource.type.fields);
     if (
         aorFetchType === GET_LIST ||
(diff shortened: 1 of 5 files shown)


EmrysMyrddin:
I have a bug, it doesn't work if you don't provide a custom `GET_COUNT` operation name. Working on it and adding some errors with more useful informations for the developer.

EmrysMyrddin:
I have updated the unit tests but it shows that we potentially have a breaking change with this feature on the `ra-data-graphql` package.

`ra-data-graphql-simple` implicitly relies on a `_allResourcesMeta` query, but it was not the case of `ra-data-graphql`. This PR explicitly a `GET_COUNT` operation in `ra-data-graphql`, but someone was using `ra-data-graphql` instead of `ra-data-graphql-simple` for some reason, it can break.

I don't know if this is fine or not. If it's not, do you have some inputs about how to implement this without breaking change ?

packages/ra-data-graphql/src/introspection.js
```diff
@@ -70,6 +70,9 @@ export default async (client, options) => {
             ) &&
             queries.some(
                 query => query.name === options.operationNames[GET_ONE](type)
+            ) &&
+            queries.some(
+                query => query.name === options.operationNames[GET_COUNT](type)
             )
```
EmrysMyrddin, line 76: Removing this check should remove the breaking change.

```suggestion
```

EmrysMyrddin:
After more thought, there is no explicit error message in the `isResource`, so it bring nothing to make `GET_COUNT` operation required in `ra-data-graphql`.

There is no more breaking change. I have just added a more explicit error in `ra-data-graphql-simple` for missing `GET_COUNT` operation, which was implicitly required before.

EmrysMyrddin:
I have also fixed a bug where the params where not filtered out if GET_COUNT query didn't declared them

EmrysMyrddin:
Just a little update, Is there something I should do to push this PR forward ? We would like to have this feature before deploying our backoffice based on react-admin to production :-)

EmrysMyrddin:
@fzaninotto @djhi Just a friendly ping :-) Do I have something to do for this PR to be merged ?

patou:
Any chance to view this PR merged ?

mael-morel:
Hi, is it possible to merge this PR ?

fzaninotto:
it needs rebase, can only be pushed to `next`, and will have to wait until we overhaul the current GraphQL provider.

mael-morel:
ok, thanks for the reply

fzaninotto:
React-admin version 3 is feature freeze. Can you please rebase this PR against `next`, and make it compatible with v4?

EmrysMyrddin:
I'm closing this PR. I don't use this anymore and I'm not sure if it's useful to spare time on a rebase.

Don't hesitate to comment if you think that it could be interesting to work on it, I could look at it if you want to.

PR #5411, opened 2020-10-16
Speed up show & hide filter

fzaninotto:
The `useListController` hook returned show and hide filter functions based on the debounced setFilters function. No need to debounce (and to wait for 500 millisecons) to show or hide a filter!

Also, the FiltrListItems (for the ListSidebar) don't need a 500ms delay either. 

This improves filtering UX in a noticeable way.

### DIFF
--- a/packages/ra-core/src/controller/useListParams.ts
+++ b/packages/ra-core/src/controller/useListParams.ts
@@ -208,8 +208,16 @@ const useListParams = ({
     }, debounce);
 
     const setFilters = useCallback(
-        (filter, displayedFilters) =>
-            debouncedSetFilters(filter, displayedFilters),
+        (filter, displayedFilters, debounce = true) =>
+            debounce
+                ? debouncedSetFilters(filter, displayedFilters)
+                : changeParams({
+                      type: SET_FILTER,
+                      payload: {
+                          filter: removeEmpty(filter),
+                          displayedFilters,
+                      },
+                  }),
         requestSignature // eslint-disable-line react-hooks/exhaustive-deps
     );
 
--- a/packages/ra-core/src/controller/useListParams.ts
+++ b/packages/ra-core/src/controller/useListParams.ts
@@ -197,52 +197,55 @@ const useListParams = ({
     );
     const displayedFilterValues = query.displayedFilters || emptyObject;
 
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/controller/useListParams.ts
```diff
@@ -197,52 +197,63 @@ const useListParams = ({
...
+        // we don't use lodash.set() for displayed filters
+        // to avoid problems with compound filter names (e.g. 'author.name')
+        const displayedFilters = Object.keys(displayedFilterValues).reduce(
+            (filters, filter) => {
+                return filter !== filterName
+                    ? { ...filters, [filter]: true }
+                    : filters;
+            },
+            {}
+        );
```
EmrysMyrddin, line 234: Perhaps it is more easy to read with a simple loop ?

```suggestion
		const displayedFilters = {}
		for(filter in displayedFilterValues) {
			if(filter !== filterName) displayedFilters[filter] = true
		}
```

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -153,14 +153,20 @@ const FilterListItem: FC<{ label: string; value: any }> = props => {
...
-        }, {} as any);
-        setFilters({ ...filterValues, ...inverseValue }, null);
+        const keysToRemove = Object.keys(value);
+        const filters = Object.keys(filterValues).reduce(
+            (acc, key) =>
+                keysToRemove.includes(key)
+                    ? acc
+                    : { ...acc, [key]: filterValues[key] },
+            {}
+        );
```
EmrysMyrddin, line 167: If you are agree with the previous comment, this one can also be written with a loop:

```suggestion
		const filters = {}
		for(filter in filters) {
			if(keysToRemove.includes(filter)) filters[key] = filterValues[filter]
		}
```

EmrysMyrddin (approved):
This is great for the user experience, side bar filters feels a lot more responsive !

packages/ra-core/src/controller/useListParams.ts
fzaninotto: that's a matter of taste... and I don't particularly like `for` loops.

djhi (approved):

PR #5368, opened 2020-10-08
allow className attribute in SimpleFormIterator

edulix:
This is a simple PR to allow the className attribute in SimpleFormIterator.

PR #5414, opened 2020-10-17
Allow formClassName in SimpleFormIterator

audunhalland:
I saw that the `className` property had recently beed added to SimpleFormIterator, but it didn't do exactly what I was hoping for. As `className` is for the "exterior" div, I wanted a way to control the "interior" div that directly holds the inputs.

On the name of the property:
* `formClassName` is used throughout the codebase (though I'm not sure if this falls into the same pattern)
* SimpleFormIterator already uses the `form` style internally for this div.

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
@@ -100,6 +100,7 @@ const SimpleFormIterator: FC<SimpleFormIteratorProps> = props => {
         basePath,
         children,
         className,
+        formClassName,
         fields,
         meta: { error, submitFailed },
         record,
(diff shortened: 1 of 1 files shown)


djhi:
Thanks but it has already be done in #5368

audunhalland:
No, it hasn't. Please read the description again.

fzaninotto:
`formClassName` is for inputs (in your case, `ArrayInput`). If you want to override the style of an inner component of `SimpleFormIterator`, you'll have to use [material-ui's `withStyle`](https://material-ui.com/styles/advanced/#withstyles) to override the `form` class:

```jsx
import { withStyles } from '@material-ui/core/styles';
import { SimpleFormIterator } from 'react-admin';

const MyFormIterator = withStyles({
   form: { /* your custom sytles here */ }
})(SimpleFormIterator);
```

audunhalland:
Ah, thanks. I didn't know of that pattern.

Issue #5419, opened 2020-10-19
Pass false in UserMenu prop in AppBar component


mariaefi29:
**Is your feature request related to a problem? Please describe.**

Sometimes applications are not required to have any users. 

For that purpose I think it would be a great idea to pass `userMenu={false}` as a prop in a custom AppBar component.

**Describe the solution you'd like**

It would be great to be able to create your own custom AppBar component very easy like this:

`const MyAppBar = props => <AppBar {...props} userMenu={false} />;`

**Describe alternatives you've considered**

Now there is an error if your are trying to pass false, because userMenu requires JSX.Element:

`userMenu?: JSX.Element; `

There is a workaround to pass icon={false} in a custom UserMenu component and you will not see the icon of the profile page. 

```
const MyUserMenu = props => <UserMenu {...props} icon={false} />;

const MyAppBar = props => <AppBar {...props} userMenu={<MyUserMenu />} />;
```

However, the component is still there and clickable:

![image](https://user-images.githubusercontent.com/25860669/96412751-4195e600-11f3-11eb-8c93-34b742485de4.png)

Are there any other alternatives that I probably haven't found?

fzaninotto:
Fixed by #5421

mariaefi29:
Dear @fzaninotto,

Could you please provide some informations when this fix will be released?)

Thank you!

djhi:
Hi @mariaefi29, as it has been merged on the `next` branch, it will be released in `3.10.0`. As for the date of the release, we can't really say. **Should** be in a few weeks.

mariaefi29:
@djhi Thank you very much for your reply! I'll keep waiting then for 3.10.0 :)

PR #5421, opened 2020-10-19
Allow to disable the UserMenu without rewriting the AppBar

Luwangel:
Fixes https://github.com/marmelab/react-admin/issues/5419

> Sometimes applications are not required to have any users.
> For that purpose I think it would be a great idea to pass userMenu={false} as a prop in a custom AppBar component.

Using `false` without a `true` option may be confusing. So I use `null` instead.

## Todo

- [x] Allow to pass `null` to the `userMenu` prop of the `<AppBar>`
- [x] Update TS type for the `<AppBar>`

## Screenshot

```js
// examples/simple/src/Layout.js line 42
const MyAppBar = props => <AppBar {...props} userMenu={null} />;
```
Sélection_004

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -15,7 +15,7 @@ Here is an example customizing an `EditButton` component inside a `Datagrid`, us
 
 {% raw %}
 ```jsx
-import * as React from "react";
+import * as React from 'react';
 import { NumberField, List, Datagrid, TextField, EditButton } from 'react-admin';
 import { makeStyles } from '@material-ui/core/styles';
 
@@ -50,7 +50,7 @@ Here is an example using the `classes` property of the `Filter` and `List` compo
 
 {% raw %}
 ```jsx
-import * as React from "react";
+import * as React from 'react';
 import {
     BooleanField,
     Datagrid,
@@ -146,7 +146,7 @@ Sometimes you want the format to depend on the value. The following example show
 
 {% raw %}
 ```jsx
-import * as React from "react";
+import * as React from 'react';
 import { NumberField, List, Datagrid, TextField, EditButton } from 'react-admin';
 import { makeStyles } from '@material-ui/core/styles';
 import classnames from 'classnames';
@@ -189,7 +189,7 @@ Furthermore, you may extract this highlighting strategy into a Higher Order Comp
 
 {% raw %}
 ```jsx
-import * as React from "react";
+import * as React from 'react';
 import { NumberField, List, Datagrid, TextField, EditButton } from 'react-admin';
 import { makeStyles } from '@material-ui/core/styles';
 import classnames from 'classnames';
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
> Using false without a true option may be confusing. So I use null instead.

We currently use `false` for things like this everywhere in react-admin. I think we should keep it consistant

Luwangel:
> > Using false without a true option may be confusing. So I use null instead.
> 
> We currently use `false` for things like this everywhere in react-admin. I think we should keep it consistant

And if we put `true`, should we display the default `<UserMenu>`?

djhi:
> And if we put true, should we display the default <UserMenu>?

It should accept a boolean but either `false` or an element (on the TS type). If `true` is provided, an error should probably be thrown. I'll try to find an example somewhere

djhi:
See https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/ListView.tsx#L72 (bulkActionsButtons)

fzaninotto:
The default value is `undefined`, using `null` as something different will confuse users IMHO. So I'd prefer `false`, even though it sucks for TS stypes.

docs/Theming.md
```diff
@@ -438,6 +438,17 @@ const MyAppBar = props => <AppBar {...props} userMenu={<MyUserMenu />} />;
 const MyLayout = props => <Layout {...props} appBar={MyAppBar} />;
 ```
 
+You can also remove the `<UserMenu>` from the `<AppBar>` by passing `null` to the `userMenu` prop:
```
fzaninotto, line 441: this should be `false` (same in the snippet below)

fzaninotto (changes requested):

packages/ra-ui-materialui/src/layout/AppBar.tsx
```diff
@@ -179,7 +192,7 @@ export interface AppBarProps extends Omit<MuiAppBarProps, 'title' | 'classes'> {
     logout?: JSX.Element;
     open?: boolean;
     title?: string | JSX.Element;
-    userMenu?: JSX.Element;
+    userMenu?: JSX.Element | boolean;
```
djhi, line 195:
```suggestion
    userMenu?: JSX.Element | false;
```

djhi (changes requested):

packages/ra-ui-materialui/src/layout/AppBar.tsx
djhi: I think this should work

Luwangel: True is allowed because it will display the default user menu.

fzaninotto (approved):

PR #5434, opened 2020-10-22
Fix Appbar custom content flickers when loading

fzaninotto:
The size of the spinner isn't exactly the same as the size of the refresh button. As a consequence, if the AppBar has a custom content, this content moves a bit when the app loads data.

Making the size of the two components match exactly fixes the bug.

## Before

spinner2

## After

spinner1

### DIFF
--- a/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
+++ b/packages/ra-ui-materialui/src/layout/LoadingIndicator.js
@@ -2,7 +2,7 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { useSelector } from 'react-redux';
-import { makeStyles } from '@material-ui/core/styles';
+import { makeStyles, useTheme } from '@material-ui/core/styles';
 import CircularProgress from '@material-ui/core/CircularProgress';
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/LoadingIndicator.js
```diff
@@ -11,7 +11,7 @@ import RefreshIconButton from '../button/RefreshIconButton';
 const useStyles = makeStyles(
     {
         loader: {
-            margin: 14,
+            margin: 16,
```
djhi, line 14: Why don't use `theme.spacing(2)` ? Could use the `useTheme` hook in the component as well

packages/ra-ui-materialui/src/layout/LoadingIndicator.js
```diff
@@ -27,8 +27,8 @@ const LoadingIndicator = props => {
         <CircularProgress
             className={classNames('app-loader', classes.loader, className)}
             color="inherit"
-            size={18}
-            thickness={5}
+            size="1em"
```
djhi, line 30: You might want to get the theme from `useTheme` and apply the same spacing

djhi (approved):

PR #5440, opened 2020-10-26
[Doc] Group Inputs and Fields in documentation to allow better discoverability

fzaninotto:
## Problem

The fields and Inputs doc table of content are intimidating because they show too much content. Besides, the toc is too long and doesn't show the last elements on small screens

## Solution

Users use the reference and full-text search to get to a given input doc anyway, so let's use heading levels for discoverability.

Add a new heading level and group Inputs and Fields by category

## Before

image

 ## After

image

## Demo

input-toc


Luwangel (approved):
It is difficult to review the changes because of Github. So I took a quick look at it and it looks good to me.

PR #5441, opened 2020-10-26
Add ability to customize the ready screen on empty admins

fzaninotto:
```jsx
import { Admin } from 'react-admin';

const Ready = () => (
    <div>
        <h1>Admin ready</h1>
        <p>You can now add resources</p>
    </div>
)

const App = () => (
    <Admin ready={Ready}>
        ...
    </Admin>
);
```

This will help build an offline-first distribution of react-admin.


packages/ra-core/src/util/Ready.tsx
```diff
@@ -65,37 +65,41 @@ const Button = ({ img, label, href }) => {
...
+                    alt="react-admin logo"
+                />
+                <h1>Welcome to React-admin</h1>
+                <div>
+                    Your application is properly configured.
```
Luwangel, line 81: We don't translate these texts?

packages/ra-core/src/util/Ready.tsx
```diff
@@ -65,37 +65,41 @@ const Button = ({ img, label, href }) => {
...
-            <h1>Welcome to React-admin</h1>
-            <div>
-                Your application is properly configured.
-                <br />
-                Now you can add a &lt;Resource&gt; as child of &lt;Admin&gt;.
+export default () =>
+    process.env.NODE_ENV === 'production' ? (
+        <span />
+    ) : (
+        <div style={styles.root}>
```
Luwangel, line 72: It's probably better to use material-ui styles here instead of inline styles.

Luwangel (changes requested):

packages/ra-core/src/util/Ready.tsx
fzaninotto: Because at that stage it's probable that the admin hasn't any i18n. The default Ready component is meant for development purposes only.

packages/ra-core/src/util/Ready.tsx
fzaninotto: no, it's in ra-core, there is no material-ui here.

docs/Admin.md
```diff
@@ -552,6 +553,32 @@ const App = () => (
 );
 ```
 
+### `ready`
+
+When you run an `<Admin>` with no child `<Resource>`, react-admin displays a "ready" screen:
+
+![Empty Admin](./img/tutorial_empty.png)
+
+You can replace that "ready" screen your custom component by passing it as the `ready` prop:
```
Luwangel, line 562:
```suggestion
You can replace that "ready" screen by passing a custom component as the `ready` prop:
```

Luwangel (changes requested):
Just one last review about the documentation and I'll merge it :)

PR #5446, opened 2020-10-28
Introduce Resource Context

djhi:


djhi:
Seems I messed up my commits

packages/ra-core/src/controller/details/useCreateController.ts
```diff
@@ -118,6 +115,11 @@ export const useCreateController = <
     const recordToUse = getRecord(location, record);
     const version = useVersion();
 
+    const { resource } = useResource(props);
+    const { hasCreate, hasEdit, hasShow, hasList } = useResourceConfig(
```
fzaninotto, line 119: I don't think the controller should grab that data. The controller should only grab data necessary to build a context. The resource configuration data is already available in Redux. Individual components that need the resourceConfig can simply use it from the redux store (which is BC because the data already exists in the store).

packages/ra-core/src/controller/details/useCreateController.ts
```diff
@@ -213,6 +215,10 @@ export const useCreateController = <
     });
 
     return {
+        hasCreate,
```
fzaninotto, line 218: so I don't think the controller should return that data. Same for other controllers

packages/ra-core/src/controller/details/useCreateController.ts
```diff
@@ -118,6 +115,11 @@ export const useCreateController = <
     const recordToUse = getRecord(location, record);
     const version = useVersion();
 
+    const { resource } = useResource(props);
```
fzaninotto, line 118: I think the controllers should only pass the resource down for BC. since the context is already there, components down the tree can already get it

packages/ra-core/src/core/ResourceProvider.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { ReactElement, ReactNode } from 'react';
+import { ResourceContext, ResourceContextValue } from './ResourceContext';
+
+export const ResourceProvider = ({
```
fzaninotto, line 5: Should be called ResourceContextProvider for consistency

packages/ra-core/src/core/useGetResource.ts
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from 'react';
+import { ResourceContextValue } from './ResourceContext';
+import { useResources } from './useResources';
+
+export const useGetResource = (name: string): ResourceContextValue => {
```
fzaninotto, line 5: without jsDoc, this one is confusing. Why would you want to use it instead of useResourceConfig?

fzaninotto (changes requested):

PR #5447, opened 2020-10-28
Allow to set many fields as ChipField source

vibby:
Concatenate all values to do so.
Concatenation string is optionally settable.

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -202,6 +202,22 @@ import { ChipField, SingleFieldList, ReferenceManyField } from 'react-admin';
 
 Any additional props are passed to material-ui's `<Chip>` element. Check [The material-ui `<Chip>` documentation](https://material-ui.com/api/chip/) for details.
 
+#### Using many fields
+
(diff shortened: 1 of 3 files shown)


vibby:
I have been searching for tests to add some, but cannot find where they are, can you lead me that way ?

fzaninotto:
This won't work without a label, as the Datagrid uses the field source for the column header.

Also, why allow only join? I think there are many ways to combine several fields into one string.

Finally, why do it only in Chipfield? All fields could benefit from having a value computed from several sources, like, say FunctionField.

But this is not the philosophy of react-admin. The Field components we bundle are simple, and deal with 80% of the cases. For the remaining 20%, it would require more complexity, more documentation, and finally a steeper learning curve for react-admin, while it's easier to do in plain react.

So we won't merge your PR, but feel free to publish your Field component as a standalone package, and to open a PR here to add a link to it in the Fields documentation.

Thanks for your proposal anyway.

vibby:
Ok, that makes sense. Thank you for all your work and share @fzaninotto and ll :)

PR #5453, opened 2020-10-29
Use React 17 in examples to make sure react-admin works with this version 

fzaninotto:
- [x] Upgrade dependencies
- [x] Make sure the simple example works
- [x] Make sure the tutorial example works
- [x] Make sure the demo example works


fzaninotto:
Current status:

image

fzaninotto:
👍

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -446,9 +446,9 @@ const AutocompleteInput: FunctionComponent<AutocompleteInputProps> = props => {
                     color,
                     ...inputProps
                 } = getInputProps({
+                    ...InputProps,
```
fzaninotto, line 449: Why did you make this (breaking) change? It prevents developers from overriding `onBlur` and `onFocus`

PR #5463, opened 2020-10-30
Better compatibility with custom AuthProviders

leoafarias:
There are some cases with custom AuthProvider that might choose to return something on the promise.
You can see an example here:
https://github.com/benwinding/react-admin-firebase/blob/master/src/providers/AuthProvider.ts#L198

By returning void, it makes these providers incompatible and display type errors.


packages/ra-core/src/types.ts
```diff
@@ -68,8 +68,8 @@ export interface UserIdentity {
 export type AuthProvider = {
     login: (params: any) => Promise<any>;
     logout: (params: any) => Promise<void | string>;
-    checkAuth: (params: any) => Promise<void>;
-    checkError: (error: any) => Promise<void>;
+    checkAuth: (params: any) => Promise<any>;
+    checkError: (error: any) => Promise<any>;
```
djhi, line 72: Then a generic type with a default to `void` might be better don't you think?

djhi (changes requested):
Thanks :)

packages/ra-core/src/types.ts
djhi: Somehing like:
```ts
checkError: <T = void>(error: any) => Promise<T>;
```

leoafarias: @djhi I tried the following initially. However, I still get a type error when adding the authProvider
```typescript
checkError: <T = any>(error: any) => Promise<T>;
```

I get the following linting error the same applies for `void`
```
Type 'Promise<User>' is not assignable to type 'Promise<T>'.
        Type 'User' is not assignable to type 'T'.
          'T' could be instantiated with an arbitrary type which could be unrelated to 'User'
```

I could get around this by doing the following.
```tsx
 <Admin
        dataProvider={dataProvider}
        authProvider={authProvider as AuthProvider}
      >
```

It's your call, but just some AuthProviders might need to have to be overriden.

Let me know what you think.

leoafarias: At least the override would work. With `Promise<void>` the override errors out saying the types are incompatible.

packages/ra-core/src/types.ts
```diff
@@ -68,8 +68,8 @@ export interface UserIdentity {
 export type AuthProvider = {
     login: (params: any) => Promise<any>;
     logout: (params: any) => Promise<void | false | string>;
-    checkAuth: (params: any) => Promise<void>;
-    checkError: (error: any) => Promise<void>;
+    checkAuth: <T = any>(error: any) => Promise<T>;
```
fzaninotto, line 71: `checkAuth` param shouldn't be named `error`, it's misleading

fzaninotto:
TypeScript doesn't seem to like your changes...

leoafarias:
> TypeScript doesn't seem to like your changes...

Yes, that is the reason why my initial commit did not have a generic.

fzaninotto:
We won't move further with. this PR: It's the authPRovider's responsibility to respect the contract.

PR #5471, opened 2020-11-02
fix: add default value for hasShow in Show.tsx

FredericEspiau:
Prevents having `hasShow` on a `div` element, because `React` complains about this

```
Warning: React does not recognize the `hasShow` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hasshow` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (created by ShowView)
    in ShowView (created by Show)
```


djhi (approved):

djhi:
Thanks!

PR #5473, opened 2020-11-02
Fix layout component type

fzaninotto:
Supersedes #5399

### DIFF
--- a/examples/demo/src/layout/themes.ts
+++ b/examples/demo/src/layout/themes.ts
@@ -6,7 +6,7 @@ export const darkTheme = {
         secondary: {
             main: '#FBBA72',
         },
-        type: 'dark', // Switching the dark mode on is a single property value change.
+        type: 'dark' as 'dark', // Switching the dark mode on is a single property value change.
     },
     overrides: {
         MuiAppBar: {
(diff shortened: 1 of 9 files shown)


examples/demo/src/layout/themes.ts
```diff
@@ -6,7 +6,7 @@ export const darkTheme = {
         secondary: {
             main: '#FBBA72',
         },
-        type: 'dark', // Switching the dark mode on is a single property value change.
+        type: 'dark' as 'dark', // Switching the dark mode on is a single property value change.
```
djhi, line 9: :scream: :thinking:

examples/demo/src/layout/themes.ts
```diff
@@ -32,6 +32,7 @@ export const lightTheme = {
         background: {
             default: '#fcfcfe',
         },
+        type: 'light' as 'light',
```
djhi, line 35: :eyes:

djhi (approved):

examples/demo/src/layout/themes.ts
djhi: Does it mean users will have to cast this as well ?

fzaninotto: yes, that's because MUI defines the `type` type as `'dark' | 'light'`, and TypeScript is dumb. Nothing we can do to avoid that.

PR #5483, opened 2020-11-03
Fix ShowButton does not update on 'to' prop change

rkfg:
Identical to #5014, ShowButton doesn't update if 'to' prop changes. My goal was to make the edit and show buttons work consistently with tabbed layouts, when a tab is opened in the show view the edit button should open the same tab in the edit view. Due to memoizing it didn't always work and now it does.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5499, opened 2020-11-06
Extended downloadCSV helper with type paramater

ValentinnDimitroff:
Recently I needed to change the output encoding (text/csv;charset=utf-8) of the csv file. I had to reimplement it at my side because currently there is no way to  plug in at the type sport and customize the output. 

That's why I am proposing this extended parameter with default value if nothing is passed in.

### DIFF
--- a/packages/ra-core/src/export/downloadCSV.ts
+++ b/packages/ra-core/src/export/downloadCSV.ts
@@ -1,4 +1,4 @@
-export default (csv, filename, blobType = 'text/csv') => {
+export default (csv, filename, blobType = 'text/csv;charset=utf-8') => {
     const fakeLink = document.createElement('a');
     fakeLink.style.display = 'none';
     document.body.appendChild(fakeLink);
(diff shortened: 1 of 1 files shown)


fzaninotto:
Shouldn't `text/csv;charset=utf-8` be the default mime type? After all, the output always uses UTF-8, right?

ValentinnDimitroff:
Actually I also think it is better to be the default mime type. In my case it appears it was not the default behaviour because I had unreadble cyrilic symbols in the exported csv file. After the change it all started working correctly. 

I just wanted to keep everything as it was but if you don't mind I can replace the parameter's default value and update the PR?

fzaninotto:
yes, please!

fzaninotto:
I didn't make myself clear enough: I think we don't need to make the type customizable is we modify it to always be utf-8. So you could just make it a bug fix instead of an enhancement.

ValentinnDimitroff:
Yeah, sounds as the most sensible solution. Didn't catch your point the first time - I rebased to master now as it is just a bug fix.

djhi (approved):

PR #5503, opened 2020-11-09
[RFR] Fix documentation issue

janakact:
The property is not `multiLine`. It is `multiline` in the component props.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5511, opened 2020-11-10
Fix options prop not being injected to View elements

### LINKED ISSUE #5512: List component doesn't receive options prop from Resource component
Addressed here #5511
Reproduced with react-admin v3.10.0

WiXSL:
Closes #5512


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5517, opened 2020-11-11
Editing code examples errors and typos

ValentinnDimitroff:


fzaninotto:
Thanks, could you PR against `master` rather than `next`? The former is for bug/documentation fixes, the latter for new features.

ValentinnDimitroff:
Oh, sure! Didn't know that's the way to proceed but will do from now on :)

fzaninotto (approved):

PR #5525, opened 2020-11-12
[Doc] Fix SimpleList properties table is too small

WiXSL:


docs/List.md
```diff
@@ -2213,7 +2213,7 @@ const CustomList = () => {
 
 For mobile devices, a `<Datagrid>` is often unusable - there is simply not enough space to display several columns. The convention in that case is to use a simple list, with only one column per row. The `<SimpleList>` component serves that purpose, leveraging [material-ui's `<List>` and `<ListItem>` components](https://material-ui.com/components/lists/). `<SimpleList>` is an **iterator** component: it gets an array of ids and a data store from the `ListContext`, and iterates over the ids to display each record.
 
-### Properties
+#### Properties
```
fzaninotto, line 2216: I don't understand that change. What's the problem?

fzaninotto (changes requested):

docs/List.md
WiXSL: `SimpleList` html props table has smaller font-size than the other props tables.
This change should fix that

simplelist-fields

fzaninotto: But then you're changing the document structure... The proper fix would be to change the CSS, which apparently uses a different size for tables following a h4 than a h3.

WiXSL: Ok, I'll look into it

PR #5527, opened 2020-11-13
Fix useReferenceInputController sorting

### LINKED ISSUE #5526: ReferenceInput returned choices are not being sorted by sort field property
**What you were expecting:**

The choices returned by ReferenceInput should be sorted when sort property is used.

**What happened instead:**

Items are returned in default / ID order

**Steps to reproduce:**

Add the below ReferenceInput posts filter to the simple demo post list and observe SelectInput elements are ordered by ID instead of title   

**Related code:**

```
        <ReferenceInput
            label="Post"
            source="id"
            reference="posts"
            sort={{ field: 'title', order: 'ASC' }}
            alwaysOn
        >
        </ReferenceInput>
```

**Other information:**

I've looked into the file: packages/ra-core/src/controller/input/useReferenceInputController.ts and have noticed that the non-sorting issue started happening around the point [useGetList was introduced](https://github.com/marmelab/react-admin/commit/0de23cf83269cbf44874aeaa72310532b100d1e8).  

I was able to get the list sorted by mapping by the ID list supplied by useGetList (finalIds).  Have not tested with manual choices yet:  [possible fix](https://github.com/rbrynjolfson/react-admin/commit/651ca49b33fdc7d532319d3a63950405a3d0b341)

**Environment**

* React-admin version: 3.10.0
* Last version that did not exhibit the issue (if applicable): 3.9.6
* React version: 17.0.1
* Browser: Chrome

djhi:
Fixes #5526

### DIFF
--- a/examples/simple/src/comments/PostReferenceInput.js
+++ b/examples/simple/src/comments/PostReferenceInput.js
@@ -30,26 +30,12 @@ const useStyles = makeStyles({
 const PostReferenceInput = props => {
     const translate = useTranslate();
     const classes = useStyles();
-    const dispatch = useDispatch();
     const { change } = useForm();
 
     const [showCreateDialog, setShowCreateDialog] = useState(false);
     const [showPreviewDialog, setShowPreviewDialog] = useState(false);
     const [newPostId, setNewPostId] = useState('');
-
-    useEffect(() => {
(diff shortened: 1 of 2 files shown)


fzaninotto (changes requested):
Apart from the warning raised by the linter, I'm OK with these changes

fzaninotto (approved):

PR #5528, opened 2020-11-13
Add support for a custom datagrid header cell react element

edulix:


fzaninotto:
Hi @edulix, and thanks for your contribution. 

If we add a finer customization of the Datagrid, we should allow developers to inject a `datagridHeader` row elemnt, rather than customizing a cell element further down in the tree. 

Would you mind updating your PR in that sense?

edulix:
@fzaninotto thanks for the feedback. Are you sure? Looking at the code, because of the bulk actions or expand mode support that would require a bit more changes.

fzaninotto:
The design principles behind react-admin is to avoid "god" objects that allow deep customization. Instead, we allow injection of components at every level. You can see this at work in the `<Admin>` component, where we invite users to pass a custom `layout` if they want to change the `menu`.

The same principles apply here. Yes, it implies that the `datagridHeader` component will expect props like `expand` and `hasBulkActions`.

edulix:
It makes sense, thanks for the clarification. As the Datagrid component is actually small because it's a wrapper, I think I will go down the route of creating my own datagrid with yours as a base, which I think I would have to anyway for the things I want to do. Thus, I'm closing this pull request.

PR #5531, opened 2020-11-13
Document the role of the `id` field in records fixes #5530

### LINKED ISSUE #5530: Feature request: document the role of the `id` field in records better
While implementing a dataProvider with a fake `getOne` and a real `getList`, it took us a few hours to realize that our fake result from `getOne` was wrong. We expected our fake data to be displayed in the `show` page after clicking a row from the `list` page, instead we got a blank page. No useful results or messages appeared in the UI or in the console, apart from a cryptic `warning: You cannot change <Router history>` that looked completely unrelated since we never changed it and never re-render the app component (the one with `Admin` or in our case `AdminContext`) as mentionned in this [issue](https://github.com/marmelab/react-admin/issues/4998#issuecomment-653424456).

It seems that the `id` field in the URL of the `show` page must match the record that comes from the `getOne` or weird things happen. This field seems super important. 
I'd like to see at least better docs about the role of the `id` field in the records. After looking quickly it seems this place is ideal https://marmelab.com/react-admin/DataProviders.html#response-format. Below the table is a sentence saying that each `Record` must have an `id` field.

I was about to open an issue about the React router problem...

**Additional context**
Using react-admin 3.10.0.

DjebbZ:
Fixes #5530

The content is rough, it's just to give you a head start about what to write.

### DIFF
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -378,7 +378,7 @@ Data Providers methods must return a Promise for an object with a `data` propert
 | `delete`           | `{ data: {Record} }` The record that has been deleted           |
 | `deleteMany`       | `{ data: {mixed[]} }` The ids of the deleted records (optional) |
 
-A `{Record}` is an object literal with at least an `id` property, e.g. `{ id: 123, title: "hello, world" }`.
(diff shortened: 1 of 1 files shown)


fzaninotto:
As explained in #5530, I don't think that's the right fix for the problem, so I'm closing this PR.

PR #5559, opened 2020-11-19
Fix FilterListItem doesn't accept object values

mjattiot:
If the filter value is a nested object, the component is unable to find out if it is selected.


fzaninotto:
Could you give an example of the usage you're doing of nested object values?

mjattiot:
Sure. Our filters are mongo queries that are serialized to a REST API. It looks like this : 
` <FilterList
        label="resources.goldens/disclosures.fields.marketData.DsclsrRspnDdln"
        icon={<AccessTimeIcon />}
    >
        <FilterListItem
            label="dates.filters.today"
            value={{
                $and: [
                    {
                        'marketData.IssrDsclsrDdln.Dt.Dt': {
                            $gte: { $date: startOfToday().toISOString() },
                        },
                    },
                    {
                        'marketData.IssrDsclsrDdln.Dt.Dt': {
                            $lte: { $date: endOfToday().toISOString() },
                        },
                    },
                ],
            }}
        />
`

fzaninotto (approved):

fzaninotto:
Now I understand. Thanks!

PR #5560, opened 2020-11-19
Fix Loading route missing theme with LoadingPage

thcolin:
Hello !

I've seen the loading page miss my material-ui theme palette configuration and display the default "blue" color for the `<CircularProgress />` component

The issue is mainly caused because `<Loading />` component isn't wrapped in a `<ThemeUIProvider />` when used in `CoreAdminRouter`, so I created a `<LoadingPage />` handling this use case

I've fixed default theme value to `createMuiThem({})`


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5561, opened 2020-11-19
[Doc] Remove suggestion of children in Admin's ready example

WiXSL:


docs/Admin.md
```diff
@@ -575,9 +575,7 @@ const Ready = () => (
 )
 
 const App = () => (
-    <Admin ready={Ready}>
-        ...
-    </Admin>
+    <Admin ready={Ready}></Admin>
```
fzaninotto, line 578: I don't think it makes sense. The use case is more of an admin with a children-as-function that returns nothing.

WiXSL: I remove the dots because there seem confusing for the example. Like you need or may have Admin's children when using the `ready` prop.

WiXSL: I've changed it to a function as a child

fzaninotto:
I don't think your edit improves the doc. You're giving an example that isn't realistic at all.  

In fact, I don't understand the problem with the current version of the doc.

WiXSL:
Ok, maybe your right. Let's close it

PR #5575, opened 2020-11-23
Move control of sidebar width to its menu child

djhi:


fzaninotto (approved):

fzaninotto:
@djhi This PR doesn't have a description or a use case. The changelog says "Fix `<Sidebar>` width cannot be modified by the child `<Menu>`". Do you remember why we did this?

fzaninotto:
I think that's because we want to allow menus of different width (like an icon menu), so it's the menu's responsibility to define its width, not its container's.

fzaninotto:
Found it. We did this because ra-enterprise *removes* the width handling from the Sidebar to allow it to be handled by the child Menu. But when a developer uses a ra oss Menu inside a ra-enterprise Admin (and instead of the ra ee Menu), the open/close feature doesn't work. 

This PR forces the ra oss Menu to *also* handle widths based on open/closed status. And we decided that in the next major, it'd be better that the Menu alone handles the open/close status, not the sidebar.

fzaninotto:
After testing, it appears that it's not possible to have a sidebar adapt to the width of its content. So we'll have to rely on the theme to let developers customize the sidebar width depending on the menu width.

PR #5577, opened 2020-11-23
Fix logout causes error in useGetList

### LINKED ISSUE #5576: useGetList on Dashboard breaks when logging out
**What you were expecting:**
Data displayed on a dashboard page and fetched with the `useGetList` hook should remain consistent and not cause problems when logging out.

**What happened instead:**
When logging out, `ids` are still filled after re-render, but the `data` object yields an empty `Object`. This causes errors if data is rendered like in the `useGetList` example code in the documentation: https://marmelab.com/react-admin/Actions.html

IDs and data should be consistent!

**Steps to reproduce:**
In case the code sandbox link below shouldn't be available anymore: I changed the "tutorial" example to load and display data on the dashboard, like so:
```
import { useGetList } from "react-admin";
import Card from "@material-ui/core/Card";
import CardContent from "@material-ui/core/CardContent";
import CardHeader from "@material-ui/core/CardHeader";

export default () => {
  const { data, ids, loading, error } = useGetList(
    "posts",
    {
      page: 1,
      perPage: 10
    },
    { field: "published_at", order: "DESC" }
  );
  console.log("######### ids = ", ids, "\n######## data = ", data);
  return (
    <Card>
      <CardHeader title="Welcome to the administration" />
      <CardContent>Lorem ipsum sic dolor amet...</CardContent>
      {!loading && !error && (
        <>
          {ids.map((id) => (
            <p>
              {id}: {data[id].title}
            </p>
          ))}
        </>
      )}
    </Card>
  );
};
```
When logging out, `ids` are still present but `data` becomes an empty object, which breaks the data lookup. The console output looks like this:

```
######### ids =  
(10) [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]

######## data =  
{}
```

**Related code:**
See code sandbox: https://codesandbox.io/s/elegant-volhard-funkp?file=/src/Dashboard.js

**Environment**

* React-admin version: ^3.9.0

fzaninotto:
Closes #5576

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetList.ts
+++ b/packages/ra-core/src/dataProvider/useGetList.ts
@@ -82,7 +82,7 @@ const useGetList = <RecordType extends Record = Record>(
             ids: get(
                 state.admin.resources,
                 [resource, 'list', 'cachedRequests', requestSignature, 'ids'],
-                defaultIds
+                null
             ),
             allRecords: get(
                 state.admin.resources,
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Looks great but you have a warning about hook dependencies

djhi (approved):

PR #5599, opened 2020-11-27
Fix data-generator avatars

djhi:
image

### DIFF
--- a/examples/data-generator/package.json
+++ b/examples/data-generator/package.json
@@ -1,6 +1,6 @@
 {
     "name": "data-generator-retail",
-    "version": "3.9.5",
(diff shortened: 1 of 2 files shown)


examples/data-generator/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "data-generator-retail",
-    "version": "3.9.4",
+    "version": "3.9.5",
```
fzaninotto, line 3: It's normally managed by lerna, you shouldn't have to update it manually.

fzaninotto (changes requested):

fzaninotto (approved):

PR #5604, opened 2020-11-30
Make SaveContext provides access to side effects

djhi:


packages/ra-core/src/controller/details/SaveContext.tsx
```diff
@@ -81,6 +84,9 @@ export const usePickSaveContext = <
                 'setOnFailure',
                 'setOnSuccess',
                 'setTransform',
+                'onSuccess',
+                'onFailure',
+                'transorm',
```
fzaninotto, line 89: transform?

fzaninotto (changes requested):

fzaninotto (approved):

djhi:
This PR should probably target `next`, shouldn't it?

packages/ra-core/src/controller/details/SaveContext.tsx
```diff
@@ -24,7 +27,7 @@ interface SaveContextValue extends SideEffectContextValue {
     saving?: boolean;
 }
 
-export const SaveContext = createContext<SaveContextValue>({});
+export const SaveContext = createContext<SaveContextValue>(undefined);
```
fzaninotto, line 30: I'm surprised TypeScript doesn't complain about the default context value. The type should be `SaveContextValue?` I think.

fzaninotto (changes requested):
Yes, this should target `next`. Besides, the linter has some remarks about missing dependencies in your `useEffect` calls in tests.

packages/ra-core/src/controller/details/SaveContext.tsx
djhi: I was surprised as well but yet it doesn't complain about it

fzaninotto (approved):

PR #5606, opened 2020-11-30
Use getFieldLabelTranslationArgs in ReferenceArrayInput

alanpoulain:
The `ReferenceArrayInput` component is not using the `getFieldLabelTranslationArgs` function like the `FieldTitle` component (used by the other inputs: `ReferenceInput`, `TextInput`, ...).

The difference is that, if no label is given and if there is no translation, the label is not humanized by default.

Before:
image

After:
image


djhi (approved):
Good catch! Thanks!

PR #5607, opened 2020-11-30
Fix usePermissions always triggers a re-render even though the permissions are unchanged

fzaninotto:
## Problem

All react-admin routes (list, edit, create, etc) are wrapped by `WithPermissions`. This component renders the page content optimistically with empty permissions, then calls `authProvider.getPermissions()`, and passes the result to the child.

In many cases, this means the routes will render twice: once with empty permissions, and once with the permissions returned by the `authProvider`. 

## Solution

`WithPermissions` calls `authProvider.getPermissions()` with empty params. In many cases, the result of this call is always the same.

`useGetPermissions` can cache the result from the first call, and initialize its state with it. 

that way, in most cases, the return from  `authProvider.getPermissions()` will be the same as the cached one, and we'll save a re-render.

This should git a short perf boost to all apps, whether they use permissions or not.

## Before

12 renders when coming back to the Post List page

image

## After

11 renders when coming back to the Post List page

image

### DIFF
--- a/packages/ra-core/src/auth/WithPermissions.tsx
+++ b/packages/ra-core/src/auth/WithPermissions.tsx
@@ -9,7 +9,7 @@ import { Location } from 'history';
 
 import warning from '../util/warning';
 import useAuthenticated from './useAuthenticated';
-import usePermissions from './usePermissions';
+import usePermissionsOptimized from './usePermissionsOptimized';
 
 export interface WithPermissionsChildrenParams {
     permissions: any;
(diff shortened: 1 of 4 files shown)


fzaninotto:
That won't work because even though the `permissions` don't change, `usePermissions` changes the loading state once `auth.getPermissions()` returns, and that triggers a rerender. 

This is unfortunate here as we don't use the loading state.

this means we must introduce a new hook that only stores the permissions, but not the loading state.

djhi (approved):
I can't figure a better name but I'm not convinced by `usePermissionsOptimized`. I think most people will end up using it just because of the name

djhi (changes requested):
Oh and you forgot to document it ;)

fzaninotto:
It's an internal API. I'm not even sure we should export it.

PR #5613, opened 2020-12-02
Fix useless AppBar rerenders

fzaninotto:
These changers save some rerenders when profiling the simple app using the React profiler. Yet they have no perceivable effects on performance in ms.

So, let's consider this as a cleanup PR.


djhi (approved):
There's a lint warnings and tests fail though

djhi (approved):

PR #5623, opened 2020-12-03
Add better support for async validators

djhi:
It's currently possible to pass an async validation function if, and only if, this is either the only validator or the last one inside the validation functions array.

This PR aims at providing real support for async validation functions.

### TODO

- [x] Implementation
- [x] Tests
- [x] Documentation


fzaninotto (approved):

fzaninotto:
What's left for RFR?

PR #5624, opened 2020-12-04
Update NotificationType type.

hieusmiths:
Addi success type

### DIFF
--- a/packages/ra-core/src/actions/notificationActions.ts
+++ b/packages/ra-core/src/actions/notificationActions.ts
@@ -1,6 +1,6 @@
 export const SHOW_NOTIFICATION = 'RA/SHOW_NOTIFICATION';
 
-export type NotificationType = 'info' | 'warning' | 'error';
+export type NotificationType = 'success' | 'info' | 'warning' | 'error';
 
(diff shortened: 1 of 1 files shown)


hieusmiths:
With Notification, i think we should support success type for sync with material or some thing operator between info and success

djhi:
If we want to support a `success` variant, this PR should also update the custom styles in the `Notification` component as we don't ues the `Alert` component from material-ui. I'm not sure we want this variant though. Ping @fzaninotto

hieusmiths:
> If we want to support a `success` variant, this PR should also update the custom styles in the `Notification` component as we don't ues the `Alert` component from material-ui. I'm not sure we want this variant though. Ping @fzaninotto

Okay, I will update the custom style for Notification success type more.

fzaninotto:
I'm OK with adding this level

fzaninotto:
Any updates on this one?

hieusmiths:
> Any updates on this one?

Now I don't have time for an update, maybe is future.

fzaninotto:
Superseded by #5710

PR #5627, opened 2020-12-06
Fix ra-data-simple-rest delete headers part2

XuHaoJun:
should create Headers instance instead of pure object.


fzaninotto:
Thanks! PR #5628 supersedes this PR, because the unit tests don't pass in your PR.

PR #5633, opened 2020-12-07
allow common input prop textAlign=center

andrewlorenz:
Even though textAlign="center" is a perfectly valid option in MaterialUI, and also documented as an option for react-admin, it causes a warning when used as it is excluded from the permitted values in the PropTypes.  This simple PR is to extend the permitted values to include "center" and thus remove the unwarranted warning:

```
Warning: Failed prop type: Invalid prop `textAlign` of value `center` supplied to `TextField`, expected one of ["right","left"].
    at TextField

### DIFF
--- a/packages/ra-ui-materialui/src/field/types.ts
+++ b/packages/ra-ui-materialui/src/field/types.ts
@@ -2,7 +2,7 @@ import { ReactElement } from 'react';
 import { Record } from 'ra-core';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
The [material-ui TableCell documentation](https://material-ui.com/api/table-cell/#props) lists even more valid values for the `align` prop.  I suggest you use `Pick<TableCellProps, 'align'>` as type to avoid duplicating it in react-admin.

fzaninotto:
Superseded by #6152

PR #5641, opened 2020-12-08
Fix: Removed hasCreate check

### LINKED ISSUE #5637: List empty property not working as expected: empty component is not displayed when list is empty
**What you were expecting:**
When the `<List empty={}>` component is empty, the list should display the empty component.

**What happened instead:**
The provided empty component is not displayed when the list is empty, Took me a while to figure it out, it's related to checking the `hasCreate` state

**Steps to reproduce:**
Simply pass a custom component `<List empty={<CustomComponent />}>` to an empty list that doesn't have `hasCreate=true`
Seem like it's a bug related to the assumption that `hasCreate` needs to be true to display the empty component. The empty component should be displayed for any empty Lists regardless of the `hasCreate` status. One of our resources doesn't allow creation but we would still like to display custom empty components

**Related code:**
Should be an easy fix: (remove `hasCreate` check)

https://github.com/marmelab/react-admin/blob/12a40fb73cb4c8cd60feeed515aac6dccd420b55/packages/ra-ui-materialui/src/list/ListView.tsx#L91
```
    const shouldRenderEmptyPage =
        hasCreate &&
        loaded &&
        !loading &&
        total === 0 &&
        !Object.keys(filterValues).length;
```

**Environment**
* React-admin version: 3.10.2
* React version: 17.0.1

Thanks for a great framework guys!

nausif:
Removed hasCreate check from shouldRenderEmptyPage which is on ListView Component.

Closes #5637

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -88,10 +88,7 @@ export const ListView = (props: ListViewProps) => {
     );
 
     const shouldRenderEmptyPage =
-        loaded &&
-        !loading &&
-        total === 0 &&
-        !Object.keys(filterValues).length;
+        loaded && !loading && total === 0 && !Object.keys(filterValues).length;
 
     return (
         <div
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -46,6 +46,7 @@ export const ListView = (props: ListViewProps) => {
         total,
         loaded,
         loading,
+        hasCreate,
         filterValues,
         selectedIds,
     } = listContext;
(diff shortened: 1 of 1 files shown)


djhi:
Hi and thanks for taking time to make this PR. I think that's not enough though. First, there is a linter warning. Then, the linter warning made me think about what will be displayed by default when the list is empty and there is no create view configured. Indeed, the default empty component will still show a create invite and button even though it won't be possible.

I'm not sure yet about how we want to address this. Any suggestion?

fzaninotto:
I think it should be the responsiblity of the Empty component to display something different if hasCreate is false. Because currently, lists with no crate button and empty content cannot show a custom screen. 

So I'm +1 for this change, but you must add the hasCreate check in the default Empty component

djhi:
Superseded by #5688

UtsavEuler:
Why is there is a requirement for no filters, I want to show a custom component in case filters are applied and there were no results. Is there some other way to do that?

slax57:
> Why is there is a requirement for no filters, I want to show a custom component in case filters are applied and there were no results. Is there some other way to do that?

@UtsavEuler In that case you can leverage the `empty` prop on the **child component** instead (for instance `<Datagrid>`).
See the documentation for [Datagrid empty prop](https://marmelab.com/react-admin/Datagrid.html#empty).

PR #5642, opened 2020-12-08
Fix FilterListItem doesn't accept object values V2

mjattiot:
PR #5559 was incomplete after deeper testing

@fzaninotto could you tell us if we are allowed to add deepdash and if yes how to pass the build tests to make this PR green ! :)

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
@@ -8,7 +8,7 @@ import {
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import CancelIcon from '@material-ui/icons/CancelOutlined';
-import { useTranslate, useListFilterContext } from 'ra-core';
+import { useTranslate, useListContext } from 'ra-core';
 import { shallowEqual } from 'react-redux';
 import { isEmpty, isMatch, merge, omit } from 'lodash';
 import { paths } from 'deepdash-es/standalone';
@@ -145,7 +145,12 @@ const useStyles = makeStyles(theme => ({
  */
 const FilterListItem: FC<{ label: string; value: any }> = props => {
     const { label, value } = props;
-    const { filterValues, setFilters } = useListFilterContext();
+    const {
+        filterValues,
+        setFilters,
+        displayedFilters,
+        hideFilter,
+    } = useListContext();
     const translate = useTranslate();
     const classes = useStyles(props);
 
--- a/packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
@@ -10,7 +10,8 @@ import { makeStyles } from '@material-ui/core/styles';
 import CancelIcon from '@material-ui/icons/CancelOutlined';
 import { useTranslate, useListFilterContext } from 'ra-core';
 import { shallowEqual } from 'react-redux';
-import isEqual from 'lodash/isEqual';
+import { isEmpty, isMatch, merge, omit } from 'lodash';
+import { paths } from 'deepdash-es/standalone';
 
 const useStyles = makeStyles(theme => ({
     listItem: {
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -148,23 +149,21 @@ const FilterListItem: FC<{ label: string; value: any }> = props => {
     const translate = useTranslate();
     const classes = useStyles(props);
 
-    const isSelected = isEqual(value, filterValues);
```
mjattiot, line 151: isEqual is wrong because we can have several filters in filterValues. We just want to know if "value" is part of filterValues, therefore using lodash isMatch instead

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -148,23 +149,21 @@ const FilterListItem: FC<{ label: string; value: any }> = props => {
...
+    const isSelected =
+        isEmpty(value) || isEmpty(filterValues)
+            ? false
+            : isMatch(filterValues, value);
 
     const addFilter = () => {
-        setFilters({ ...filterValues, ...value }, null, false);
```
mjattiot, line 154: Problem : we should update URL properly with the nested fields of value and without overriding nodes of other filters contained in filterValues. 
Correction : using lodash deep merge instead of spread operator.

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -148,23 +149,21 @@ const FilterListItem: FC<{ label: string; value: any }> = props => {
...
-        const keysToRemove = Object.keys(value);
-        const filters = Object.keys(filterValues).reduce(
-            (acc, key) =>
-                keysToRemove.includes(key)
-                    ? acc
-                    : { ...acc, [key]: filterValues[key] },
-            {}
-        );
-
-        setFilters(filters, null, false);
```
mjattiot, line 167: Problem: we should update URL properly by removing only the leaf keys of value from filterValues.
Correction: using lodash omit and deepdash paths instead. Do you mind adding deepdash as a dependency ?

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -148,23 +149,21 @@ const FilterListItem: FC<{ label: string; value: any }> = props => {
...
+        setFilters(omit(filterValues, paths(value)), null, false);
+        const displayedFilterToRemove = Object.keys(
```
mjattiot, line 163: Problem: We found out that the displayedFilter key of the URL was not properly cleaned when remove a sidebar filter, therefore leave standard Filters shown. 
Correction: manual cleaning

mjattiot:
Adding inline comments for explanation

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -144,27 +145,30 @@ const useStyles = makeStyles(theme => ({
...
+            displayedFilters
+        )?.find(elem => paths(value).includes(elem));
+        hideFilter(displayedFilterToRemove);
```
mjattiot, line 171: Problem: We found out that the displayedFilter key of the URL was not properly cleaned when remove a sidebar filter, therefore leave standard Filters shown.
Correction: manual cleaning

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -144,27 +145,30 @@ const useStyles = makeStyles(theme => ({
...
+        hideFilter,
+    } = useListContext();
     const translate = useTranslate();
     const classes = useStyles(props);
 
-    const isSelected = isEqual(value, filterValues);
+    const isSelected =
+        isEmpty(value) || isEmpty(filterValues)
+            ? false
+            : isMatch(filterValues, value);
```
mjattiot, line 160: Problem: isEqual is wrong because we can have several filters in filterValues. We just want to know if "value" is part of filterValues.
Correction: using lodash isMatch instead

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -144,27 +145,30 @@ const useStyles = makeStyles(theme => ({
...
 
     const addFilter = () => {
-        setFilters({ ...filterValues, ...value }, null, false);
+        setFilters(merge({}, filterValues, value), null, false);
```
mjattiot, line 163: Problem : we should update URL properly with the nested fields of value and without overriding nodes of other filters contained in filterValues.
Correction : using lodash deep merge instead of spread operator.

packages/ra-ui-materialui/src/list/filter/FilterListItem.tsx
```diff
@@ -144,27 +145,30 @@ const useStyles = makeStyles(theme => ({
...
-        const filters = Object.keys(filterValues).reduce(
-            (acc, key) =>
-                keysToRemove.includes(key)
-                    ? acc
-                    : { ...acc, [key]: filterValues[key] },
-            {}
-        );
-
-        setFilters(filters, null, false);
+        setFilters(omit(filterValues, paths(value)), null, false);
```
mjattiot, line 167: Problem: we should update URL properly by removing only the leaf keys of value from filterValues.
Correction: using lodash omit and deepdash paths instead. Do you mind adding deepdash as a dependency ?

mjattiot:
@fzaninotto could you tell us if we are allowed to add deepdash and if yes how to pass the build tests to make this PR green ! :)

fzaninotto:
You will have to add unit tests showing the problems you're trying to solve. The `master` branch already contains a FilterListItemp.spec.tsx you can build upon.

Also, please avoid adding a new dependency.

fzaninotto:

PR #5643, opened 2020-12-08
Getting parsed value in FilterliveSearch

hbaptiste:
This PR from Opensense proposes a few improvements for the FilterLiveSearch. It
 - allows the component to deal with nested FilterValues
- allows, on the onSearchChange event, to use a formatted value when the 'format' prop is provided to the component
- allows the FilterLiveSearch form to be reseted when the corresponding <Filter> is removed
- allows us to provide and display our own label (current is always 'search')

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import { FC, ChangeEvent, memo } from 'react';
+import { FC, ChangeEvent, memo, useRef } from 'react';
 import { InputAdornment } from '@material-ui/core';
 import SearchIcon from '@material-ui/icons/Search';
-import { Form } from 'react-final-form';
+import { AnyObject, Form } from 'react-final-form';
 import { useTranslate, useListFilterContext } from 'ra-core';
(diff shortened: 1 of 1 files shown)


hbaptiste:
Missing dependencies

packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
```diff
@@ -22,41 +23,57 @@ import TextInput from '../../input/TextInput';
...
-                />
-            )}
+            {({ form }) => {
+                formRef.current = form;
+                return (
+                    <TextInput
+                        resettable
+                        helperText={false}
+                        source={source}
+                        label={
```
djhi, line 61: What do you think about setting a default value for `label` in the props declaration instead?

djhi (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
hbaptiste: @djhi, done. Cleaner.

fzaninotto:
Hi, and thanks for your PR. 

Please add unit tests, document your changes in the Documentation, and target the `next` branch.

fzaninotto:

PR #5644, opened 2020-12-08
Fix FilterListItem doesn't appear selected when more than one filter is applied

fzaninotto:
Fixes a regression introduced by #5559 

Added unit tests to avoid regressions in the future

## Before

filterlistitem_broken

## After

filterlistitem


mjattiot:
Indeed our last PR was not properly tested, sorry.
We have introduced a new one #5642 doing the same fix as yours, but yours seems more elegant.
Anyway we also introduce into this #5642 PR some additional improvments to handle nested values in FilterListItem

djhi (approved):

PR #5655, opened 2020-12-11
Fix "Missing translation" console error when the dataProvider fails

fzaninotto:
## Problem

Server error messages are seldom translated. And the `useNotify` hook logs a warning in the console for every missing translation. This pollutes the console a great deal and harms developer experience. 

## Solution

Provide default translation for dataProvider error notifications (the untranslated message). This will remove the warnings for untranslated server errors.

We do this in every call to useNotify in return to dataProvider errors instead of in useNotify itself, because useNotify is often called with error codes that need to be identified as mission transltions.  

## Before

image

## After

image


djhi (approved):
:+1:

PR #5657, opened 2020-12-11
Fix Filter fails to show compound filters with no default value

### LINKED ISSUE #5653: Filter with nested source does not work correctly
duplicate the issue [https://github.com/marmelab/react-admin/issues/5361](https://github.com/marmelab/react-admin/issues/5361) as how it got closed but is now actual again

> I expected that when adding a filter field to the screen (without entering any value), nothing happens to the current list. Instead, a request is sent to
> getList posts {"pagination":{"page":1,"perPage":10},"sort":{"field":"published_at","order":"DESC"},"filter":{"comment":{}}}
> and the list becomes empty.
> 
> This differs from the behavior of adding a filter for a non-nested field — until a filter value is entered, no pre-request is sent.
> The difference in behavior for the user is not obvious.
> 
> Often the fields of nested filters are named as if it were a field of the current resource and it is difficult to explain to the user why it works differently.

here's a link Codesandbox [https://codesandbox.io/s/aged-https-f8ynv?file=/src/posts/PostList.js](https://codesandbox.io/s/aged-https-f8ynv?file=/src/posts/PostList.js) with last packages versions

1. Add a filter comment.title to the Posts list view.
2. Get unwanted filtering

> no, you indicated that you fixed that in 3.9.6 https://github.com/marmelab/react-admin/releases/tag/v3.9.6
> 
> I was confused that when I updated the application a couple of weeks ago to 3.9.6 everything worked for me. I was delighted and removed some of the crutches in my backend. Everything worked in the release build. However, when yesterday they began to reassemble the release, the filters again stopped working correctly...
> They searched for a long time and determined that if you slip a yarn.lock two weeks ago, then everything works.
> Hence the conclusion - something has changed in your dependencies.

* React-admin version: 3.10.4
* Last version that did not exhibit the issue (if applicable): 3.9.6

fzaninotto:
Closes #5653

### DIFF
--- a/packages/ra-core/src/controller/useListParams.spec.tsx
+++ b/packages/ra-core/src/controller/useListParams.spec.tsx
@@ -192,7 +192,7 @@ describe('useListParams', () => {
     describe('showFilter', () => {
         it('should initialize displayed filters', () => {
             const TestedComponent = () => {
-                const [_, { showFilter }] = useListParams({
+                const [, { showFilter }] = useListParams({
                     resource: 'foo',
                     location: {} as any,
                 });
@@ -226,7 +226,7 @@ describe('useListParams', () => {
         });
         it('should initialize filters', () => {
             const TestedComponent = () => {
-                const [_, { showFilter }] = useListParams({
+                const [, { showFilter }] = useListParams({
                     resource: 'foo',
                     location: {} as any,
                 });
@@ -261,7 +261,7 @@ describe('useListParams', () => {
 
         it('should initialize displayed filters on compound filters', () => {
             const TestedComponent = () => {
-                const [_, { showFilter }] = useListParams({
+                const [, { showFilter }] = useListParams({
                     resource: 'foo',
                     location: {} as any,
                 });
@@ -296,7 +296,7 @@ describe('useListParams', () => {
 
         it('should initialize filters on compound filters', () => {
             const TestedComponent = () => {
-                const [_, { showFilter }] = useListParams({
+                const [, { showFilter }] = useListParams({
                     resource: 'foo',
                     location: {} as any,
                 });
--- a/packages/ra-core/src/controller/useListParams.spec.tsx
+++ b/packages/ra-core/src/controller/useListParams.spec.tsx
@@ -1,4 +1,11 @@
-import { getQuery, getNumberOrDefault } from './useListParams';
+import * as React from 'react';
+import expect from 'expect';
+import { render } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/controller/useListParams.ts
```diff
@@ -246,7 +246,9 @@ const useListParams = ({
             ...displayedFilterValues,
             [filterName]: true,
         };
-        const filter = set(filterValues, filterName, defaultValue);
+        const filter = defaultValue
```
fzaninotto, line 249: The problem was with compound filters (e.f. 'comment.title') with no default value. set returned `{comment: {}}`, which is wrong in that case.

packages/ra-core/src/controller/useListParams.spec.tsx
```diff
@@ -182,4 +189,144 @@ describe('useListParams', () => {
             expect(result).toEqual(0);
         });
     });
+    describe('showFilter', () => {
+        it('should initialize displayed filters', () => {
+            const TestedComponent = () => {
+                const [_, { showFilter }] = useListParams({
```
djhi, line 195: You don't need the `_`, just leave the slot it empty
```suggestion
                const [, { showFilter }] = useListParams({
```

packages/ra-core/src/controller/useListParams.spec.tsx
```diff
@@ -182,4 +189,144 @@ describe('useListParams', () => {
...
+                        order: 'ASC',
+                        page: 1,
+                        perPage: 10,
+                    }),
+            });
+        });
+
+        it('should initialize displayed filters on compound filters', () => {
+            const TestedComponent = () => {
+                const [_, { showFilter }] = useListParams({
```
djhi, line 264: Same
```suggestion
                const [, { showFilter }] = useListParams({
```

packages/ra-core/src/controller/useListParams.spec.tsx
```diff
@@ -182,4 +189,144 @@ describe('useListParams', () => {
...
+                        order: 'ASC',
+                        page: 1,
+                        perPage: 10,
+                    }),
+            });
+        });
+
+        it('should initialize filters on compound filters', () => {
+            const TestedComponent = () => {
+                const [_, { showFilter }] = useListParams({
```
djhi, line 299: Same
```suggestion
                const [, { showFilter }] = useListParams({
```

djhi (changes requested):

djhi (approved):

PR #5658, opened 2020-12-11
[Docs] Fix List examples

WiXSL:


docs/List.md
```diff
@@ -1759,8 +1759,7 @@ export const PaginationActions = props => <RaPaginationActions {...props} color=
 export const Pagination = props => <RaPagination {...props} ActionsComponent={PaginationActions} />;
 
 export const UserList = props => (
-    <List {...props} pagination={<Pagination />}>
-    </List>
+    <List {...props} pagination={<Pagination />} />
```
fzaninotto, line 1762: I'd prefer that you add "..." between the two lines, suggesting that there must be children.

Using List without children isn't supported/

fzaninotto (changes requested):

docs/List.md
WiXSL: Done!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5663, opened 2020-12-14
Fix full width TextInput in Filter bug

### LINKED ISSUE #4579: width=100% does not work for List filter text input
**What you were expecting:**
I have a list filter with one field. I am trying to make text input 100% of width as follows. However it still stays the default width.

```js
const useStylesFilterForm = makeStyles({
  form:{ backgroundColor: 'orange'}
})


const PostFilter = (props) => {
  const classes=useStylesFilterForm();
  return (
    <Filter {...props}  classes={classes} >
      <TextInput label="Search 1" source="id" alwaysOn  fullWidth={true}  />
    </Filter>
  );
}
```

**What happened instead:**
Although form control received MuiFormControl-fullWidth class which set width:100% on the element), the Text Input Element still remained the original size.

**Steps to reproduce:**
Add PostFilter  component as filter to any list

**Related code:**
React admin used the following html code to render filter (notice ```MuiFormControl-fullWidth``` applied properly)

```html
<form class="RaFilter-form-194 makeStyles-form-191 RaFilterForm-form-195">
<div data-source="id" class="filter-field RaFilterFormInput-body-197">
<div class="MuiFormControl-root MuiTextField-root MuiFormControl-marginDense MuiFormControl-fullWidth">...
.
```

I was able to force Chrome to render 100% width by adding the following styles:
- add ```flex: "1 0 auto" ``` to ```form``` div;
- add ```width:100%``` style to the parent div (div data-source="id"); however i was not able to add this style using properties passed TextInput or Filter components:

  - I also noticed that TextInput should accept ```formClassName``` property to be applied to the parent div  (per https://marmelab.com/react-admin/Inputs.html#common-input-props); However, the following code also did not work, as no styles were applied to the parent div. Additionally, helper Text did not render.

```js
const useStylesFilterForm = makeStyles({
  form:{flex: "1 0 auto"},
  formClassName:{backgroundColor:"red", width:"100%"},
  inputComponent:{width:"100%"}
})


const PostFilter = (props) => {
  const classes=useStylesFilterForm();

  return (
    <Filter {...props}  classes={{form:classes.form}} >
      <TextInput
        resettable
        label="Report Search"
        source="search:reportSearch"
        alwaysOn
        fullWidth={true}
        formClassName={classes.formClassName}
        helperText="Helper text"
        className={classes.inputComponent}
        />
    </Filter>
  );
}
```

**Environment**

* React-admin version: 3.3.3 
* React version: 16.13.1
* Material UI 4.9.7

AnkitaGupta111:
Closes #4579

### DIFF
--- a/packages/ra-ui-materialui/src/input/SearchInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SearchInput.tsx
@@ -12,7 +12,7 @@ import TextInput from './TextInput';
 const useStyles = makeStyles(
     {
         input: {
-            marginTop: 32,
+            // marginTop: 32,
         },
     },
     { name: 'RaSearchInput' }
--- a/packages/ra-ui-materialui/src/input/SearchInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SearchInput.tsx
@@ -12,7 +12,7 @@ import TextInput from './TextInput';
 const useStyles = makeStyles(
     {
         input: {
-            // marginTop: 32,
+            marginTop: 32,
         },
     },
     { name: 'RaSearchInput' }
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
@@ -14,11 +14,11 @@ import { ClassesOverride } from '../../types';
 const useStyles = makeStyles(
     theme => ({
         form: {
-            alignItems: 'center',
+            marginTop: -theme.spacing(2),
             flexGrow: 1,
-
             paddingTop: 0,
             display: 'flex',
+            alignItems: 'flex-end',
             flexWrap: 'wrap',
             minHeight: theme.spacing(10),
             pointerEvents: 'none',
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
@@ -14,10 +14,11 @@ import { ClassesOverride } from '../../types';
 const useStyles = makeStyles(
     theme => ({
         form: {
-            marginTop: -theme.spacing(2),
+            alignItems: 'center',
+            flexGrow: 1,
+
             paddingTop: 0,
             display: 'flex',
-            alignItems: 'flex-end',
             flexWrap: 'wrap',
             minHeight: theme.spacing(10),
             pointerEvents: 'none',
--- a/packages/ra-ui-materialui/src/list/filter/FilterFormInput.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterFormInput.tsx
@@ -15,6 +15,9 @@ const useStyles = makeStyles(
             alignItems: 'flex-end',
             pointerEvents: 'auto',
         },
+        fullWidthBody: {
+            width: '100%',
+        },
         spacer: { width: theme.spacing(2) },
         hideButton: {},
     }),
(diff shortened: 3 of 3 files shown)


packages/ra-ui-materialui/src/input/SearchInput.tsx
```diff
@@ -12,7 +12,7 @@ import TextInput from './TextInput';
 const useStyles = makeStyles(
     {
         input: {
-            marginTop: 32,
+            // marginTop: 32,
```
djhi, line 15: Why did you comment this? Does not seem related to the issue. Besides, we don't keep commented out code

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -14,10 +14,11 @@ import { ClassesOverride } from '../../types';
 const useStyles = makeStyles(
     theme => ({
         form: {
-            marginTop: -theme.spacing(2),
+            alignItems: 'center',
+            flexGrow: 1,
```
djhi, line 18: Why? Does not seem related to the issue

djhi (changes requested):

fzaninotto:
What bug are you trying to solve?

fzaninotto:
Nevermind, I didn't see that you referenced an issue in the title. 

The GitHub convention is to reference it in the description (I edited yours to match that convention)

AnkitaGupta111:
@djhi can you look into this pr

fzaninotto:
You didn't answer @djhi's comment (see earlier in these PR comments).

packages/ra-ui-materialui/src/input/SearchInput.tsx
AnkitaGupta111: i removed this unrelated code changes now in latest commit

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
AnkitaGupta111: i removed this unrelated code changes now in latest commit

AnkitaGupta111:
@fzaninotto i already removed  unrelated code changes @djhi pointed

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
fzaninotto: no, it's still there

AnkitaGupta111: I removed `alignItems:center` property which was not related to this issue but `flexGrow:1`  is something which helped in fixing bug. 

`filterFormInput` has a parent wrapper `form` tag in `filterForm` component and by applying style `flexGrow:1` to form ,`div` tag (inside `filterFormInput` which is parent of filter text input) style` width=100%` property is effective  and because of parent `div` can now take full-width,search text input could take up `width=100%` as well

macklin-10x:
@djhi @fzaninotto what needs to happen with this PR to get it merged?

fzaninotto:
I find the result surprising:

```jsx
const commentFilters = [
    <SearchInput source="q" alwaysOn />,
    <ReferenceInput source="post_id" reference="posts" fullWidth={true}>
    </ReferenceInput>,
];
```

image

I'd expect otherwise:
- either the ReferenceInput takes the rest of the available space at the right side of the search input,
- or it takes the full card width, including the space below the buttons

So I'm not sure the fix is correct.

fzaninotto:
This PR fixes a bug that is no longer present in 4.x and 5.x. So I'm closing it.

PR #5668, opened 2020-12-14
Optimize References Loaders Display

djhi:
When displaying a lot of references fields/inputs, for example in a List, having a lot of loaders can harm performances, especially on low-end devices. This PR applies the same optimization as for the skeletons: don't display loaders until at least a second has passed

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
@@ -15,7 +15,7 @@ import {
 import { fieldPropTypes, PublicFieldProps, InjectedFieldProps } from './types';
 import { ClassesOverride } from '../types';
 import sanitizeFieldRestProps from './sanitizeFieldRestProps';
-import { LinearProgressIfTooLong } from '../layout/LinearProgressIfTooLong';
+import { LinearProgress } from '../layout';
 
 /**
  * A container component that fetches records from another resource specified
@@ -175,7 +175,7 @@ export const ReferenceArrayFieldView: FC<ReferenceArrayFieldViewProps> = props =
     const { loaded } = useListContext(props);
 
     if (!loaded) {
-        return <LinearProgressIfTooLong className={classes.progress} />;
+        return <LinearProgress className={classes.progress} />;
     }
 
     return (
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { Children, cloneElement, FC, memo, ReactElement } from 'react';
 import PropTypes from 'prop-types';
-import { LinearProgress } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import {
     ListContextProvider,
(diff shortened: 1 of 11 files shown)


packages/ra-ui-materialui/src/layout/LinearProgressIfTooLong.tsx
```diff
@@ -0,0 +1,9 @@
+import { useTimeout } from 'ra-core';
+import * as React from 'react';
+import LinearProgress from './LinearProgress';
+
+export const LinearProgressIfTooLong = ({ timeout = 1000, ...props }) => {
```
fzaninotto, line 5: I think we don't need a second component for that. It should be the default behavior of our LinerProgress

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
```diff
@@ -16,6 +15,7 @@ import {
 import { fieldPropTypes, PublicFieldProps, InjectedFieldProps } from './types';
 import { ClassesOverride } from '../types';
 import sanitizeFieldRestProps from './sanitizeFieldRestProps';
+import { LinearProgress } from '../layout';
```
fzaninotto, line 18: useless change

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -15,11 +15,11 @@ import {
     Record,
 } from 'ra-core';
 
-import LinearProgress from '../layout/LinearProgress';
 import Link from '../Link';
 import sanitizeFieldRestProps from './sanitizeFieldRestProps';
 import { PublicFieldProps, fieldPropTypes, InjectedFieldProps } from './types';
 import { ClassesOverride } from '../types';
+import { LinearProgress } from '../layout';
```
fzaninotto, line 22: useless change

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -14,7 +14,7 @@ import {
 } from 'ra-core';
 
 import sanitizeInputRestProps from './sanitizeInputRestProps';
-import LinearProgress from '../layout/LinearProgress';
```
fzaninotto, line 17: useless change

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -17,9 +17,9 @@ import {
 } from 'ra-core';
 
 import sanitizeInputRestProps from './sanitizeInputRestProps';
-import LinearProgress from '../layout/LinearProgress';
 import Labeled from './Labeled';
 import ReferenceError from './ReferenceError';
+import { LinearProgress } from '../layout';
```
fzaninotto, line 22: useless change

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
djhi: Nope, we are now using our own LinearProgress instead of MUI

fzaninotto (approved):

PR #5674, opened 2020-12-15
Remove min width in confirmation dialog

rkfg:
The `minWidth` style attribute in `<Confirm>` dialog should be removed. It doesn't make the dialog look much better and it adds a scroll bar for the contents on mobile.
Vanilla:
Screenshot_20201215-143843
Fixed:
Screenshot_20201215-143727

I use this component in my own code too and it's even worse when the confirmation text is long enough.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5675, opened 2020-12-15
Fix ExportButton doesn't take permanent filter into account

### LINKED ISSUE #5286: List exporter does not take into account what is in the filter prop
**What you were expecting:**
I am using the `List` component like the following.

```
<List
        perPage={15}
        pagination={orderPagination()}
        actions={<OrderListActions />}
        filter={{ store_id: storeIdFilter }}
        filters={<OrderListFilter />}
        bulkActionButtons={false}
        exporter={exporter}
        className="admin-list"
        classes={otherClasses}
        sort={{ field: "placed_on", order: "DESC" }}
        {...rest}
      >
``` 
When using the exporter functionality I expect the request to contain the `store_id` filter I have defined in the `filter` prop. 

**What happened instead:**
The `GET` request does not contain the default filter `store_id` in the params

**Environment**

* React-admin version: 3.7.2
* Last version that did not exhibit the issue (if applicable): This did work at one point. Don't remember what version
* React version: 16.12.0
* Browser: All

fzaninotto:
Closes #5286

### DIFF
--- a/packages/ra-core/src/controller/useListController.ts
+++ b/packages/ra-core/src/controller/useListController.ts
@@ -61,6 +61,7 @@ export interface ListControllerProps<RecordType extends Record = Record> {
     displayedFilters: any;
     error?: any;
     exporter?: Exporter | false;
+    filter?: FilterPayload;
     filterValues: any;
     hasCreate: boolean;
     hideFilter: (filterName: string) => void;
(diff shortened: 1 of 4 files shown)


djhi (commented):
There are many eslint warnings. I know the dependencies are refs but still.. Should I merge?

fzaninotto:
The eslint warnings are on unchanged files. We need to fix them, but on another PR.

djhi (approved):

PR #5681, opened 2020-12-17
[Doc] Fix syntax in actions example for useUpdate.

abdenny:
Optimistic Rendering and Undo section ApproveButton example is missing the 4th argument for previousData in useUpdate.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5683, opened 2020-12-18
Fix TabbedShowLayout has Divider is un

hieusmiths:


fzaninotto:
I think you messed up your PR. The title is cut, there is no description, and the diff contains commits from other branches. 

Also, you already have an open PR in this repo waiting for your feedback (#5624). I suggest you focus on finishing what you've started rather than open another PR that has no chance of being merged.

PR #5690, opened 2020-12-18
Improve jsDocs

WiXSL:


packages/ra-core/src/actions/listActions.ts
```diff
@@ -71,7 +71,7 @@ export interface ToggleListItemExpandAction {
  * Action creator to toggle the expanded state of a list item
  *
  * @param {string} resource The resource name, e.g. 'posts'
- * @param {string|integer} id The record identifier, e.g. 123
+ * @param {string|int} id The record identifier, e.g. 123
```
fzaninotto, line 74: Where do you see that `int` exists? I don't see it in https://jsdoc.app/tags-param.html#names-types-and-descriptions or https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#type

If anything, the `integer` should be changed into the native JavaScript type `Number`, but it carries a lesser constraint. So I'm -1 for this change, in all the files you've edited.

packages/ra-core/src/controller/field/useReferenceArrayFieldController.ts
```diff
@@ -42,12 +42,14 @@ const defaultSort = { field: null, order: null };
  *      source: 'referenceIds';
  * });
  *
- * @param {Object} option
```
fzaninotto, line 45: The former syntax is also valid, see https://jsdoc.app/tags-param.html#parameters-with-properties and https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param

So your change isn't required nor necessary - just a matter of preference. I'd rather not change these types.

packages/ra-ui-materialui/src/button/DeleteButton.tsx
```diff
@@ -11,7 +11,7 @@ import DeleteWithConfirmButton from './DeleteWithConfirmButton';
  * Button used to delete a single record. Added by default by the <Toolbar> of edit and show views.
  *
  * @typedef {Object} Props The props you can use (other props are injected if you used it in the <Toolbar>)
- * @param {Prop} props
+ * @param {Props} props
```
fzaninotto, line 14: If you edit this one, you'd better move it out of the `typedef` definition - I don't think `@param` is valid under `@typedef`

fzaninotto (changes requested):

packages/ra-core/src/controller/field/useReferenceArrayFieldController.ts
WiXSL: Done!

packages/ra-core/src/actions/listActions.ts
WiXSL: Done!

packages/ra-ui-materialui/src/button/DeleteButton.tsx
WiXSL: Done!

packages/ra-core/src/controller/field/useReferenceArrayFieldController.ts
fzaninotto: the change is still present

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -28,12 +28,14 @@ import { useResourceContext } from '../../core';
  *      source: 'referenceIds';
  * });
  *
- * @param {Object} option
```
fzaninotto, line 31: this change isn't necessary

packages/ra-ui-materialui/src/layout/AppBar.tsx
```diff
@@ -59,13 +59,16 @@ const useStyles = makeStyles(
 /**
  * The AppBar component renders a custom MuiAppBar.
  *
- * @param {ReactNode} children React node/s to be render as children of the AppBar
```
fzaninotto, line 62: This change isn't necessary

fzaninotto (changes requested):

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
WiXSL: Are you sure? There is no param named `option` should be at least `* @param {Object} props`

packages/ra-ui-materialui/src/layout/AppBar.tsx
WiXSL: The same (I think), you have only one param named `props` here and no description for it

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
WiXSL: Done!

packages/ra-ui-materialui/src/layout/AppBar.tsx
WiXSL: Done!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5694, opened 2020-12-20
Allow TabbedForm component to pass props to the underlying MUI Tabs component

WiXSL:


WiXSL:
i just realized the same props are passed to the `form` tag container and can break the code

Issue #5696, opened 2020-12-21
SimpleForm equivalent of FormTab contentClassName


WiXSL:
`FormTab` component allows to set `contentClassName` prop to style the content of the tab which is applied to a top level `span`.
`SimpleForm` doesn't have anything like that for its content.

fzaninotto:
As seen in #5697, I don't think this is a good idea, but we can accept a PR that would allow to replace the root `component` used by `<SimpleForm>`

fzaninotto:
Fixed by #5703

PR #5697, opened 2020-12-21
Add contentClassName prop to SimpleForm component

WiXSL:
Implements #5696


fzaninotto:
I'm not sure it's a good idea. I'd prefer if SimpleForm accepted a `component` prop allowing to replace the `CardContentInner` element. 

Also, this is a new feature, so please PR against `next`.

WiXSL:
Changed for #5703

PR #5698, opened 2020-12-21
Fix select empty option in AutocompleteInput does not reset the input 

### LINKED ISSUE #4435: Reselect empty value AutocompleteInput
Greetings all.
I faced with problem when try reselect empty value on Edit form in AutocompleteInput after input some value.

**What you were expecting:**
Empty option will be selected again and the inputted value is deleted

**What happened instead:**
Input contains inputted value.

**Steps to reproduce:**
It's easy to reproduce on [RA Demo](https://marmelab.com/react-admin-demo/#/reviews?filter=%7B%7D&order=DESC&page=1&perPage=25&sort=date)
1. Select filter **Product**;
2. Select input and select **not** empty option;
3. Then select input again and select empty option; We can see empty input - correct;
4. Then input 'wa' - we can see empty and other options. Select empty option - we still see what was input.

We got report from user. "Empty option has behavior like common option, it means if we select empty we expect to see empty, but we can see what was inputted".

If it's expected behaviour please close issue. But I agree with this report because when we select some option the input.value changes to the selected even if we are talking about empty value.

Perhaps this problem applies to **react-final-form** because there is no state update after selected again the empty value.

**Environment**

* React-admin version: 3.0.3 (also tested on 3.2.2)

AnkitaGupta111:
Closes #4435

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -265,9 +265,13 @@ const AutocompleteInput: FunctionComponent<AutocompleteInputProps> = props => {
 
     const handleChange = useCallback(
         (item: any) => {
+            if (getChoiceValue(item) == null && filterValue) {
+                setFilterValue('');
+            }
+
             input.onChange(getChoiceValue(item));
         },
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

PR #5702, opened 2020-12-21
[RFR] Fix: automatic refresh

alinagaripova:
This functionality adds the ability to disable automatic page refresh


fzaninotto:
Why would you want to do that?

alinagaripova:
@fzaninotto to disable clearing of filled fields before saving

fzaninotto:
Again, I don't see the bug. Version 3.8.2 fixed an issue here (see #5152) so this shouldn't be a problem anymore. 

Please explain in more details the use case, and add a unit test to your PR to show the bug prior to fixing it.

alinagaripova:
Ok, thanks.
 I will try to update the version and solve the problem based on new fixes

PR #5703, opened 2020-12-21
Add component prop to SimpleForm

WiXSL:
Related to #5696.

So `CardContentInner` component could be overriden inside `SimpleForm` component.

I have to use `container` instead of `component` since react-final-form's `FormRenderProps` already has a `component` prop


fzaninotto:
There is no `component` in `FormRenderProps`, it's in `FormProps`. How does it prevent you from using it in `SimpleForm`?

https://final-form.org/docs/react-final-form/types/FormRenderProps

WiXSL:
You are right!
I must have been confused with a typescript error I was getting manipulating the code

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5704, opened 2020-12-22
Fix error thrown when scrollable is set {true}

### LINKED ISSUE #5695: Setting scrollable={true} in TabbedFormTabs component throws an error
Using this example from the docs and getting an error:
https://marmelab.com/react-admin/CreateEdit.html#tabbedformtabs

Sandbox:
https://codesandbox.io/s/xenodochial-torvalds-fsbet?file=/src/posts/PostEdit.js

1.- Edit a Post

screenshot

Likhitha016:
Closes #5695


fzaninotto:
No, we'll fix #5695 with documentation, not by adding a level of indirection between us and material-ui.

PR #5709, opened 2020-12-22
Fix Notification component colors

WiXSL:
Finishing #5624


WiXSL:
Change to next branch

PR #5710, opened 2020-12-22
Add support for success notification type

WiXSL:
Supersedes #5624


fzaninotto (approved):

fzaninotto:
Thanks!

fzaninotto:
I realize this changes the color of all the default notifications in a way I really don't like. It feels like a regression to me.

I'll revert this PR (and #5724) so that we can rediscuss that without blocking 3.12, which is around the corner.

fzaninotto:
In particular, the "Undo" is not visible

image

WiXSL:
Ok.
I will love to have `success`, `info`, `warning` and `error` colors anyways in the future, since right now, notifications warning and error colors are both red and success and info are both black.
Let me now if aI can make a related PR.

PR #5713, opened 2020-12-23
Fix scrollbar not appear on scrollable tabs

### LINKED ISSUE #1689: A lot of tabs shows scrollbar and crashes the application
Example app: https://pp0o4x40p0.codesandbox.io/#/posts/13

I just copied the last tab a couple of times and you'll see a scrollbar at the bottom.
What I expected: that the `Tabs` component would be scrollable and the application not to crash. 
I think this is because the Tabs parent width is implicit instead of explicit.

AnkitaGupta111:
Closes #1689

### DIFF
--- a/packages/ra-ui-materialui/src/form/TabbedFormTabs.tsx
+++ b/packages/ra-ui-materialui/src/form/TabbedFormTabs.tsx
@@ -9,6 +9,13 @@ import {
 import PropTypes from 'prop-types';
 import Tabs, { TabsProps } from '@material-ui/core/Tabs';
 import { useLocation } from 'react-router-dom';
+import { makeStyles } from '@material-ui/core/styles';
+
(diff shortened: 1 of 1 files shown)


fzaninotto:
I don't see what your PR changes or fixes. Can you elaborate?

fzaninotto:
Also, you're setting a div width to 0 while it should obviously not have zero width. Can you explain why you're doing that?

AnkitaGupta111:
https://github.com/marmelab/react-admin/issues/1689#issuecomment-464111309
i just tried using this solution and now tabs are scrollable

fzaninotto:
Blindly adding code without understanding how it works won't make a good fix. Please elaborate on the reason why we need this counter-intuitive null width.

fzaninotto:

PR #5714, opened 2020-12-23
Allow MenuItemLink to receive TooltipPops

### LINKED ISSUE #5359: Allow MenuItemLink to have a customizable and optional tooltip

WiXSL:
Supersedes #5360

Closes #5359


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5724, opened 2021-01-02
Change Notification's error color from dark to main

WiXSL:
This is a left over from #5710


fzaninotto:
Thanks!

WiXSL:
This PR targets `next` branch but is listed in `3.11.2` milestone instead of `3.12`

fzaninotto:
My mistake. Milestone updated.

fzaninotto:
As explained in #5710, the result isn't good looking, so we need to discuss that further. In the meantime, and since we must deliver 3.12, I'll be reverting the PR.

PR #5730, opened 2021-01-03
Fix Create, Edit and Show cannot be used outside of a ResourceContextProvider

### LINKED ISSUE #5662: Allow passing a custom resource prop to Create / Edit
Hi guys, loving using React-Admin! The look on my clients' faces was priceless - "Will it actually be this fast? 🤯 "

I have a feature request for a now-deprecated feature to be re-enabled, although maybe you will have a compelling clarification of why it was removed. Apologies if this has been covered elsewhere, I couldn't find another discussion pertaining to this.

In prior versions of React Admin (I was using 3.7.0), I was able to make Create/Edit forms within complex objects, for simple objects within it. For example, in the image below, a Program could open a Drawer component which allowed for adding ProgramDates or FAQs related to theProgram. As a Program is a complex object with many relations within it, this prevented the user from having to switch contexts to fill out a very simple form.
Screen Shot 2020-12-13 at 10 43 28 AM

This was achieved by the following:
```javascript
const ProgramDateDetailCreate = ({ programId, onSuccess }) => (
  <Create
    {...{onSuccess}}
    basePath='/ProgramDateDetail'
    resource='ProgramDateDetail'
    undoable={false}
  >
    <SimpleForm defaultValue={{ programId }}>
      // inputs go here
    </SimpleForm>
  </Create>
)

export default ProgramDateDetailCreate
```

In newer versions of React-Admin, the `resource` that I pass is overwritten by the resource in the URL and therefore the submit sends to the "wrong" endpoint.

The ability to override the `resource` prop in the CreateContext or EditContext. I understand there may be some issues that this causes - but I think having the option, with the caveat that some redirects etc won't work, is worthwhile. Perhaps the prop could be called `dangerouslySetResource` so that you can absolve responsibility for bugs arising from it 😅 

I attempted to use the useCreate and useEdit hooks, but they remove the convenience of SimpleForm and lead to a duplication of components.

**Additional context**
The frustrating part about this bug is that I didn't upgrade and it still affected the app - even with my dependencies locked at 3.7.0, it still broke the app when I re0built my Docker container (and re-ram `npm install`). And that's probably more of an issue with NPM, though. Luckily the app is still in QA, but a change like this that could break the an app and require a re-write of many pages makes me nervous.

**EDITS**
Grammar, adding name for override prop, adding edits

fzaninotto:
Closes #5662

Although the `<Create>`, `<Edit>` and `<Show>` components were never meant to be used outside of `<Resource>` props, developers have taken the habit to use them in standalone, passing the `resource` as prop. The fact that it worked was undocumented, but intuitive and practical.

The introduction of the `<ResourceContextProvider>` broke this. An attempt to use `<Create resource="foos">` fails - the user has to wrap the component in a `<ResourceContextProvider>`.

We recognize that `<Create>`, `<Edit>` and `<Show>` should be usable outside of `<Resource>`, so they should be able to create their own `<ResourceContextProvider>` when passed a `resource` prop.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Create.tsx
+++ b/packages/ra-ui-materialui/src/detail/Create.tsx
@@ -3,6 +3,7 @@ import { ReactElement } from 'react';
 import PropTypes from 'prop-types';
 import {
     CreateContextProvider,
+    ResourceContextProvider,
     useCheckMinimumRequiredProps,
     useCreateController,
 } from 'ra-core';
@@ -58,10 +59,18 @@ export const Create = (
 ): ReactElement => {
     useCheckMinimumRequiredProps('Create', ['children'], props);
     const controllerProps = useCreateController(props);
-    return (
+    const body = (
         <CreateContextProvider value={controllerProps}>
             <CreateView {...props} {...controllerProps} />
         </CreateContextProvider>
+    );
+    return props.resource ? (
+        // support resource override via props
+        <ResourceContextProvider value={props.resource}>
+            {body}
+        </ResourceContextProvider>
+    ) : (
+        body
     );
 };
 
@@ -74,7 +83,7 @@ Create.propTypes = {
     hasCreate: PropTypes.bool,
     hasEdit: PropTypes.bool,
     hasShow: PropTypes.bool,
-    resource: PropTypes.string.isRequired,
+    resource: PropTypes.string,
     title: PropTypes.node,
     record: PropTypes.object,
     hasList: PropTypes.bool,
--- a/packages/ra-core/src/core/useResourceContext.ts
+++ b/packages/ra-core/src/core/useResourceContext.ts
@@ -5,9 +5,36 @@ import { ResourceContext, ResourceContextValue } from './ResourceContext';
  * Hook to read the resource from the ResourceContext.
  *
  * Must be used within a <ResourceContextProvider> (e.g. as a descendent of <Resource>
- * or any reference related components).
+ * or any reference related components), or called with a resource prop.
  *
- * @returns {ResourceContextValue} The resource
+ * @example
+ *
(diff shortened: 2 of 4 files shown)


packages/ra-ui-materialui/src/detail/Create.tsx
```diff
@@ -58,11 +59,19 @@ export const Create = (
 ): ReactElement => {
     useCheckMinimumRequiredProps('Create', ['children'], props);
     const controllerProps = useCreateController(props);
-    return (
+    const body = (
         <CreateContextProvider value={controllerProps}>
             <CreateView {...props} {...controllerProps} />
         </CreateContextProvider>
     );
+    return props.resource ? (
```
fzaninotto, line 67: I'v considered xtracting this duplicated logic to a dedicated commponent, like so:

```jsx
const ConditionalWrapper = ({ condition, wrapper, children }) => 
  condition ? wrapper(children) : children;

...

<ConditionalWrapper
   condition={props.resource}
   wrapper={children => <ResourceContextProvider value={props.resource}>{children}</ResourceContextProvider>}
>
    <CreateContextProvider value={controllerProps}>
        <CreateView {...props} {...controllerProps} />
    </CreateContextProvider>
</ConditionalWrapper>
```

But I find it less legible. Besides, the duplication is only a few lines.

djhi (approved):

fzaninotto:
rebased

koaladlt:
Do we have any updates on this? Is there a way to use <Create> with a different endpoint?

fzaninotto:
@koaladlt The updates on this PR is that it's merged and published in 3.11, as the "Milestone" zone on the right column suggests. 

About your other question, please consider asking it on stackOverflow. As explained in the issue template, we don't do free support on gitHub issues.

PR #5734, opened 2021-01-04
📝 Update getIdentity signature in Authentication.md

adrien-may:


djhi:
I fail to see where `getIdentity` could be called with an error. Would you mind explaining this PR?

adrien-may:
Well, I totally agree with you, that's why I REMOVED error from the signature :)

djhi:
Indeed! My mistake :laughing: Thanks!

PR #5741, opened 2021-01-05
List - Optional Sync With Location

djhi:
Make the synchronization with the current location (URL) optional.

By default, this is enabled when a list is passed to a `<Resource>`. Otherwise the synchronization is opt-in using the `syncWithLocation` prop

- [x] Implementation
- [x] Tests


packages/ra-core/src/controller/useListParams.ts
```diff
@@ -135,7 +139,9 @@ const useListParams = ({
         perPage,
     ];
 
-    const queryFromLocation = parseQueryFromLocation(location);
+    const queryFromLocation = syncWithLocation
+        ? parseQueryFromLocation(location)
```
fzaninotto, line 143: the location is grabbed in `useListController` and passed to `useListParams`. I think we'd better call `useLocation` directly in `useListParams`.

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -161,13 +167,17 @@ const useListParams = ({
...
+                search: `?${stringify({
+                    ...newParams,
+                    filter: JSON.stringify(newParams.filter),
+                    displayedFilters: JSON.stringify(
+                        newParams.displayedFilters
+                    ),
+                })}`,
+            });
+        }
         dispatch(changeListParams(resource, newParams));
```
fzaninotto, line 181: Not sure we should keep on relying on the Redux state when we don't sync with location

fzaninotto (changes requested):
Could you add unit tests?

packages/ra-core/src/controller/useListParams.ts
djhi: Done but not visible on this comment

packages/ra-core/src/controller/useListParams.ts
djhi: Done

fzaninotto (approved):

Issue #5634, opened 2020-12-07
setFilters type definition issues.


martdavidson:
**What you were expecting:**
Documentation shows using setFilters with only one parameter. For instance, from the List View docs:

```
const onSubmit = values => {
        if (Object.keys(values).length > 0) {
            setFilters(values);
        } else {
            hideFilter("main");
        }
    };
```

Type definition is:

```
setFilters: (
        filters: any,
        displayedFilters: any,
        debounce?: boolean
    ) => void;
```

Making the `displayedFilters` param required.

**What happened instead:**
You're unable to use `setFilters` as shown in the documentation without passing `displayedFilters`. Not sure whether the type definition or the documentation incorrect.

**Steps to reproduce:**
Enable typescript, and call `setFilters` with a single parameter.

**Environment**

* React-admin version: 3.10.3
* React version: 17.0.1
* Typescript Version: 4.1.2

Thanks for the great work on types though, really looking forward to having it all locked in.

djhi:
This has been done in https://github.com/marmelab/react-admin/pull/9682

PR #5748, opened 2021-01-07
feat: improve setFilters type definition

etienne-bondot:
Improving `setFilters` type definition, answering issue #5634.


packages/ra-core/src/actions/listActions.ts
```diff
@@ -8,7 +8,7 @@ export interface ListParams {
     page: number;
     perPage: number;
     filter: any;
-    displayedFilters: any;
+    displayedFilters?: { [key: string]: boolean } | null;
```
djhi, line 11: Maybe we could extract a `DisplayedFilter` type which would be `Record<string, boolean>` ? It could be inside the root `types.ts` file? What do you think about it?

WiXSL:
@etienne-bondot, did you have time to continue this?

etienne-bondot:
@djhi @WiXSL I didn't follow the thread. I will look at this this weekend to see if I can update the PR and take the comment in account.

fzaninotto:

PR #5752, opened 2021-01-08
Introduce Form Groups

djhi:
Introducing form groups will allow us to remove the restrictions on the `FormTab` content which could accept any components and not just inputs while still tracking whether any of their inputs are invalid.

- [x] Base implementation
- [x] Cleanup
- [x] Tests
- [x] Documentation


examples/simple/src/posts/PostEdit.js
```diff
@@ -43,12 +45,25 @@ const EditActions = ({ basePath, data, hasShow }) => (
...
 
+const SanitizedBox = ({ fullWidth, basePath, ...props }) => <Box {...props} />;
+
 const PostEdit = ({ permissions, ...props }) => (
     <Edit title={<PostTitle />} actions={<EditActions />} {...props}>
         <TabbedForm initialValues={{ average_note: 0 }} warnWhenUnsavedChanges>
             <FormTab label="post.form.summary">
-                <TextInput disabled source="id" />
-                <TextInput source="title" validate={required()} resettable />
+                <SanitizedBox
```
fzaninotto, line 54: why do you need that? It seems unrelated to the PR.

packages/ra-core/src/form/FormContextProvider.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { ReactNode } from 'react';
+import { FormContextValue } from '../types';
+import { FormContext } from './FormContext';
+
+/**
+ * Provides utilities to Forms children, allowing them to change the default save function or register inputs inside a group.
```
fzaninotto, line 7:
```suggestion
 * Provides utilities to Form children, allowing them to change the default save function or register inputs to a group.
```

packages/ra-core/src/form/FormGroupContextProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const FormGroupContextProvider = ({
+    children,
+    name,
+}: {
+    children: ReactNode;
+    name: string;
+}) => {
+    const formContext = useFormContext();
+
```
fzaninotto, line 22: You should check that the formContext contains the `register` and `unregister` functions and throw a meaningful error if it doesn't. 

I'm surprised that TypeScript didn't warn you on this one.

packages/ra-core/src/form/FormGroupContext.ts
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+export const FormGroupContext = createContext<FormGroupContextValue>(undefined);
```
fzaninotto, line 3: you should add a jsdoc saying that this shouldn't be used in standalone, but through FormGroupContextProvider

packages/ra-core/src/form/FormGroupContextProvider.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import { ReactNode, useEffect } from 'react';
+import { FormGroupContext } from './FormGroupContext';
+import { useFormContext } from './useFormContext';
+
+/**
+ * This provider allows its input children to know they belong to a specific group.
```
fzaninotto, line 7:
```suggestion
 * This provider allows its input children to register to a specific group.
```

packages/ra-core/src/form/FormGroupContextProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * This enables other components to access groups properties such as its
```
fzaninotto, line 8:
```suggestion
 * This enables other components in the group to access group properties such as its
```

packages/ra-core/src/form/FormGroupContextProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * validation (valid/invalid) or whether its inputs have been updated (dirty/pristine).
+ * @param props The component props
+ * @param {ReactNode} props.children The form group content
+ * @param {String} props.name The form group name
+ */
```
fzaninotto, line 13: Please add the same example you added to the doc here

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -92,7 +93,38 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
...
+            unregisterGroup: name => {
+                delete formGroups[name];
+            },
+            registerField: (source, group) => {
+                if (group) {
+                    formGroups.current[group] = formGroups.current[group] || [];
+                    const fields = new Set(formGroups.current[group]);
+                    fields.add(source);
+                    formGroups.current[group] = Array.from(fields);
+                }
```
fzaninotto, line 112: what if group isn't defined? Shouldn't you throw an error?

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -92,7 +93,38 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
...
+            registerGroup: name => {
+                formGroups.current[name] = formGroups.current[name] || [];
+            },
+            unregisterGroup: name => {
+                delete formGroups[name];
+            },
+            registerField: (source, group) => {
+                if (group) {
+                    formGroups.current[group] = formGroups.current[group] || [];
+                    const fields = new Set(formGroups.current[group]);
```
fzaninotto, line 109: you could merge the previous line with this one using

```
const fields = new Set(formGroups.current[group] | []);
```

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -92,7 +93,38 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
...
+            unregisterField: (source, group) => {
+                if (group) {
+                    if (!formGroups.current[group]) {
+                        console.warn(`Invalid form group ${group}`);
+                    } else {
+                        const fields = new Set(formGroups.current[group]);
+                        fields.delete(source);
+                        formGroups.current[group] = Array.from(fields);
+                    }
+                }
```
fzaninotto, line 123: same, what happens when called without a group?

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -92,7 +93,38 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
...
+            getGroupFields: name => formGroups.current[name] || [],
+            registerGroup: name => {
+                formGroups.current[name] = formGroups.current[name] || [];
+            },
+            unregisterGroup: name => {
+                delete formGroups[name];
+            },
+            registerField: (source, group) => {
+                if (group) {
+                    formGroups.current[group] = formGroups.current[group] || [];
```
fzaninotto, line 108: why do you allow an input to register to a group that hasn't been registered here, while you forbid it in the following function?

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -92,7 +93,38 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
         [save]
     );
 
-    const formContextValue = useMemo(() => ({ setOnSave }), [setOnSave]);
+    const formContextValue = useMemo<FormContextValue>(
+        () => ({
+            setOnSave,
+            getGroupFields: name => formGroups.current[name] || [],
```
fzaninotto, line 99: As these functions are completely static, I suggest you define them outside of the component.

packages/ra-core/src/form/useFormContext.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { FormContext } from './FormContext';
+
+/**
+ * Retrieve the form context enabling consumers to alter its save function or to register inputs inside a form group.
+ * @returns {FormContext} The form context.
+ */
+export const useFormContext = () => {
```
fzaninotto, line 8: could be a one liner

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -0,0 +1,79 @@
...
+    pristine: boolean;
+    dirty: boolean;
+};
+
+/**
+ * Retrieve a specific form group data such as its validation status (valid/invalid) or
+ * or whether its inputs have been updated (dirty/pristine)
+ * @param {string] name The form group name
+ * @returns {FormGroupState} The form group state
+ */
```
fzaninotto, line 19: please add a usage example

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -0,0 +1,79 @@
...
+    const formContext = useFormContext();
+    const [state, setState] = useState<FormGroupState>({
+        errors: undefined,
+        valid: true,
+        invalid: false,
+        pristine: true,
+        dirty: false,
+    });
+
+    form.subscribe(
```
fzaninotto, line 31: you never unsubscribe... I think there's a leak here

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -0,0 +1,79 @@
...
+        () => {
+            const fields = formContext.getGroupFields(name);
+            const newState = Array.from(fields).reduce<FormGroupState>(
```
fzaninotto, line 34: why the Array.from()?

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -0,0 +1,79 @@
...
+                (acc, field) => {
+                    const fieldState = form.getFieldState(field);
+                    let errors = acc.errors;
```
fzaninotto, line 37: `let errors = acc.errors || [];` and remove line 40 to 42

packages/ra-core/src/form/useInput.ts
```diff
@@ -44,6 +46,19 @@ const useInput = ({
     ...options
 }: InputProps): UseInputValue => {
     const finalName = name || source;
+    const formGroupName = useFormGroupContext();
+    const formContext = useFormContext();
+
+    useEffect(() => {
+        if (!formContext) {
```
fzaninotto, line 53:
```suggestion
        if (!formContext || !formGroupName ) {
```

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -44,7 +53,9 @@ describe('<SaveButton />', () => {
             <TestContext>
                 <ThemeProvider theme={theme}>
                     <SaveContextProvider value={saveContextValue}>
-                        <SaveButton {...invalidButtonDomProps} />
+                        <FormContextProvider value={formContextValue}>
```
fzaninotto, line 56: I don't understand why this is now required and it worries me (that it's now required). this will be a gotcha for users.

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -0,0 +1,79 @@
...
+
+                    if (fieldState.error) {
+                        if (!errors) {
+                            errors = {};
+                        }
+                        errors[field] = fieldState.error;
+                    }
+
+                    const newState = {
```
fzaninotto, line 46: Please add a unit test for that logic

packages/ra-ui-materialui/src/form/TabbedForm.tsx
```diff
@@ -344,7 +339,14 @@ const sanitizeRestProps = ({
     ...props
 }) => props;
 
+export default TabbedForm;
+
 export const findTabsWithErrors = (children, errors) => {
+    // TODO: Provide documentation link
```
fzaninotto, line 345: todo

fzaninotto (changes requested):
Great work! The error detection code in the tab component is much more readable. 

I have the general feeling that this could be more unit tested.

examples/simple/src/posts/PostEdit.js
djhi: It was a test to verify that we could now have anything inside a `FormTab` and the validation would still work. I can revert it if you want

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: I lazily initialize it here at L108. I found out that inputs were registering themselves before the group anyway.

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: I don't get this comment: I do check whether a group was provided (L113) and it exists (L114)

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: I found out that inputs were registering themselves before the group

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
djhi: I actually don't get how those tests could have pass before

packages/ra-core/src/form/FormGroupContextProvider.tsx
```diff
@@ -0,0 +1,78 @@
...
+    children,
+    name,
+}: {
+    children: ReactNode;
+    name: string;
+}) => {
+    const formContext = useFormContext();
+
+    useEffect(() => {
+        if (!formContext) {
```
fzaninotto, line 60:
```suggestion
        if (!formContext || !formContext.registerGroup || !formContext.unregisterGroup) {
```

packages/ra-core/src/form/FormWithRedirect.tsx
fzaninotto: I wasn't clear. My question was: what if `registerField` is called with an empty second argument? But I understand that it's the general case (no group declared) and that in this case there is nothing to do.

packages/ra-core/src/form/FormWithRedirect.tsx
fzaninotto: Nevermind.

fzaninotto (changes requested):
Almost there!

fzaninotto (approved):

PR #5767, opened 2021-01-13
Refactor ReferenceInput to let its child handle loading

djhi:
- [x] ReferenceInput
- [x] ReferenceArrayInput


fzaninotto:
Does it fix #4992?

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -386,6 +395,7 @@ const AutocompleteInput: FunctionComponent<AutocompleteInputProps> = props => {
                     return (
                         <InputAdornment position="end">
                             <span className={classes.clearButton}>&nbsp;</span>
+                            {loading && <AutocompleteInput />}
```
fzaninotto, line 398: this seems strange. Don't you want to show the AutocompleteInputLoader instead?

fzaninotto (changes requested):
Please add more tests for the Inputs you've modified (just like you did for SelectInput)

fzaninotto (approved):

PR #5768, opened 2021-01-13
Allow to disable the ready screen

djhi:
Maybe this should target next minor version.


DjebbZ:
If I may say so, there's no doc associated to the PR. Since Ready is an official option of <Admin />, maybe the new `false` value should be documented now.

BTW, I haven't tested the code, just read it briefly.

fzaninotto:
I don't understand why this is needed, because when ready=false, the app will be blank until the children are loaded. 

Please document the use case, add a unit test, and document the new value. 

And yes, this should be PRed against `next`.

fzaninotto:
I still don't understand. Why would you use the dashboard to display something if there is no resource instead of a ready page?

My understanding is that the problem comes from the fact that an admin application in ready state doesn't render the customroutes without layout, and that prevents developers from offering a good UI from the ready screen.

When we need to do is to render the custom routes without layout together with the ready screen when there are no resources.

fzaninotto:
Needs rebase and isn't necessary, so I'm closing it.

PR #5772, opened 2021-01-14
Make NavMenu Keyboard Navigable

andrico1234:
I've tested this manually. 

I haven't added any unit tests, but I'll be happy to add an integration test that ensures that the menu item is reachable via the tab button.

https://user-images.githubusercontent.com/17087167/104582931-6960e280-5658-11eb-9a30-199813357f3b.mov

### DIFF
--- a/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
+++ b/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
@@ -16,7 +16,7 @@ import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
 import { makeStyles } from '@material-ui/core/styles';
 
 const NavLinkRef = forwardRef<HTMLAnchorElement, NavLinkProps>((props, ref) => (
-    <NavLink innerRef={ref} {...props} />
+    <NavLink innerRef={ref} {...props} tabIndex={0} />
 ));
 
 const useStyles = makeStyles(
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -16,7 +16,7 @@ import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
 import { makeStyles } from '@material-ui/core/styles';
 
 const NavLinkRef = forwardRef<HTMLAnchorElement, NavLinkProps>((props, ref) => (
-    <NavLink innerRef={ref} {...props} />
+    <NavLink innerRef={ref} {...props} tabIndex={0} />
```
andrico1234, line 19: it looks like the `{...props}` spread operator contains a `tabIndex={-1}` which would override the mine if passed through behind the spreaded props.

I'm not sure what the most appropriate way to handle this. The way I've done it is most straightforward, but it currently prevents users from enabling/disabling tabbing on their own accord. Is this configuration that we'd want to give users?

djhi: This is because material UI `MenuItem` set it to `-1` when not specified. You should add this `tabIndex` prop to the `MenuItem` component below in the same file

andrico1234: Good shout, it feels much more appropriate being set in `MenuItem`, I've pushed the changes.

fzaninotto:
Nice! Indeed, an integration test would be welcome.

djhi (approved):
Excellent, thanks!

fzaninotto (approved):

PR #5778, opened 2021-01-14
Handle submission validation errors

### LINKED ISSUE #4351: Better handling of form submission errors
The documentation is missing a section on how to display validation errors for fields that are coming from API 400 responses. Example response:

```json
{
  "errors": {
    "myField": ["error1", "error2"]
  }
}
```

Clear documentation with straightforward way of displaying validation errors from API.

I have found several issue and answers for this problem when react-admin still used redux-form. I've tried those solutions to no avail just to realize that it is now using react-final-form. Looking through the code I haven't found any easy or straightforward solution. 

One solution could be composing the `save` callback provided to the create and edit view by the `useEditController` and `useCreateController` and then saving the error somehow in state and returning it in the `validate` callback. But this solution is in no way simple and straightforwards. 

Another solution that I've implemented is using saga to capture the FETCH_ERROR and saving the error in redux. This on the other hand made a big mess on managing state as it had to be removed after displaying it, otherwise it would have affected other forms as well. 

Other than that it was hard to figure out how to actually make these errors go away after displaying them, because if they were not removed you couldn't submit the form. For this I had to also save the form state values to figure out which fields changed and which error I didn't have to display anymore.

And to put the cherry on the cake making both frontend and API errors display at the same time was again quite hard. In the end I ended up with a hook of 100 lines, and a lot more lines extract from it to different services.

```typescript
export default function useEntityFormValidator(
  schema: any,
  customErrorMessages?: ISchemaErrorToTranslationCustomMapper,
): (entity: any) => IValidationErrorMapUntranslated | void {
  const [
    schemaErrorTransformer,
    apiErrorTransformer,
    jointErrorTransformer,
    schemaValidator,
  ] = useDependencyInjections(
    ServiceTypes.TRANSFORMER_SCHEMA_ERROR_TO_TRANSLATION,
    ServiceTypes.TRANSFORMER_API_ERROR_TO_TRANSLATION,
    ServiceTypes.TRANSFORMER_TRANSLATION_ERRORS_TO_JOINT,
    ServiceTypes.VALIDATOR_SCHEMA,
  );

  const lastBadRequest = useSelector<any, IHttpError | null>(
    (state: any) => state.apiError?.lastBadRequest,
  );
  const [
    apiErrorMessages,
    setApiErrorMessages,
  ] = useState<IValidationErrorMapTranslated | null>(null);
  const [apiErrorBody, setApiErrorBody] = useState<{ [k in string]: any }>({});

  const computeRemainingApiErrorMessages = useCallback(
    (entity: any) => {
      const leftApiErrors: any = {};
      Object.keys(apiErrorMessages || {}).forEach(field => {
        if (entity[field] === apiErrorBody[field]) {
          leftApiErrors[field] = (apiErrorMessages || {})[field];
        }
      });
      return leftApiErrors;
    },
    [apiErrorMessages, apiErrorBody],
  );

  useEffect(() => {
    if (lastBadRequest?.body?.errors) {
      setApiErrorBody(lastBadRequest.requestBody);
      setApiErrorMessages(apiErrorTransformer.transform(lastBadRequest.body));

      // In order to not leave errors behind we consume it from the state.
      // Otherwise other forms could potentially display errors from wrong
      // places.
      dispatch({ type: apiErrorActions.SET_400, data: null });

      // Have to focus the body because if an input remains focused error
      // messages will not be displayed until it loses focus.
      document.body.focus();
    }
  }, [lastBadRequest]);

  return useCallback(
    (entity: any) => {
      const schemaErrors = schemaValidator.validate(entity, schema) || {};
      return jointErrorTransformer.transform(
        [
          computeRemainingApiErrorMessages(entity),
          schemaErrorTransformer.transform(schemaErrors, customErrorMessages),
        ],
        '• ',
      );
    },
    [schema, customErrorMessages, computeRemainingApiErrorMessages],
  );
}
```

### LINKED ISSUE #4703: Handle Promises from onSave in FormWithRedirect
It's possible to pass custom `save` prop to `EditView` or `CreateView`, it's useful if you want to attach some custom behavior (e.g.: upload images) to submit event (without overriding `<Toolbar/>`), but currently, it's impossible to handle form submit with `Promise API` even though `final-form` supports it
https://github.com/final-form/final-form/blob/ceaf2802184222653afd5f2d4664085e93bb18a6/src/FinalForm.js#L1069-L1092
also  "complete" https://github.com/final-form/final-form/blob/ceaf2802184222653afd5f2d4664085e93bb18a6/src/FinalForm.js#L1040 is called immediately a this leads to an issue that "Submit" button is "enabled" even if "custom logic" hasn't finished.

https://github.com/marmelab/react-admin/blob/15c8306dedaaea5f79372a19727ec7c3886f22de/packages/ra-core/src/form/FormWithRedirect.tsx#L101 should simply return value from `onSave`

This will allow returning `Promise` from `onSave` handle and attach any async logic without "completing" form submit before async is done.

**UPD**
Another benefit from this would be returning errors from the backend, if you resolve it with `errors` 
final-form will redirect "errors" for required field
https://github.com/final-form/final-form/blob/ceaf2802184222653afd5f2d4664085e93bb18a6/src/FinalForm.js#L1040-L1060

**Example 1**

```jsx
const editController = useEditController(props);

const save = (record, redirectTo, callbacks) => {
    return new Promise((resolve, reject) => {
        editController.save(record, redirectTo, {
            onSuccess: function () {
                // ... any code that you need on success ...
                resolve();
            },
            onFailure: function (error) {
                // ... any code that you need on failure ...
                reject();
            }
        });
    })
};

return (
    <EditView
        {...editController}
        save={save}
    >
        <SimpleForm validateOnBlur={true}>
            <InnerComponent/>
        </SimpleForm>
    </EditView>
);
```

**Example 2** returning errors from JSON-LD backend

```jsx
onFailure: function (error) {
    const errors = {};
    const type = error.body[0]["@type"][0];

    if (type === `${entrypoint}/docs.jsonld#ConstraintViolationList`) {
        const violations = error.body[0][`${entrypoint}/docs.jsonld#violations`];
        violations.forEach((v) => {
            const key = v[`${entrypoint}/docs.jsonld#propertyPath`][0]['@value'];
            errors[key] = v[`${entrypoint}/docs.jsonld#message`][0]['@value'];
        });
    } else if (type === `http://www.w3.org/ns/hydra/core#Error`) {
        notify(error.body[0]['http://www.w3.org/ns/hydra/core#title'][0]['@value'], 'error', {smart_count: 1});
    }

    // this will cause final-form to render errors.
    resolve(errors);
}
```

alanpoulain:
Fixes #4703.
Fixes #4351.

Returning a promise in the `submit` function is essential for handling correctly submission errors: https://final-form.org/docs/react-final-form/types/FormProps#3-asynchronous-with-a-promise.

It has been discussed in this issue too: #4351.

`useCreateController` and `useEditController` hooks are already returning a promise resolving to `undefined` for the `save` function: it will not break anything.

To handle submission errors, the custom `save` function could be implemented by using the `useDataProvider` hook, like this: https://github.com/bmihelac/react-admin/commit/1a9f02b65f1342c695a10b5aad7a6a63603fb9ba.

In order to be able to use the `useMutation` hook too, a `returnPromise` option has been added, like asked in @fzaninotto's comment: https://github.com/marmelab/react-admin/issues/4703#issuecomment-616456129.

I hope it will be OK, it would be really nice to have a way to handle submission errors easily in React-Admin and API Platform Admin!

### DIFF
--- a/packages/ra-core/src/dataProvider/useMutation.ts
+++ b/packages/ra-core/src/dataProvider/useMutation.ts
@@ -31,6 +31,7 @@ import useDataProviderWithDeclarativeSideEffects from './useDataProviderWithDecl
  * @param {Object} options
  * @param {string} options.action Redux action type
  * @param {boolean} options.undoable Set to true to run the mutation locally before calling the dataProvider
+ * @param {boolean} options.returnPromise Set to true to return the result promise of the mutation
  * @param {Function} options.onSuccess Side effect function to be executed upon success or failure, e.g. { onSuccess: response => refresh() }
  * @param {Function} options.onFailure Side effect function to be executed upon failure, e.g. { onFailure: error => notify(error.message) }
  * @param {boolean} options.withDeclarativeSideEffectsSupport Set to true to support legacy side effects e.g. { onSuccess: { refresh: true } }
@@ -52,6 +53,7 @@ import useDataProviderWithDeclarativeSideEffects from './useDataProviderWithDecl
  * - {Object} options
  * - {string} options.action Redux action type
  * - {boolean} options.undoable Set to true to run the mutation locally before calling the dataProvider
+ * - {boolean} options.returnPromise Set to true to return the result promise of the mutation
  * - {Function} options.onSuccess Side effect function to be executed upon success or failure, e.g. { onSuccess: response => refresh() }
  * - {Function} options.onFailure Side effect function to be executed upon failure, e.g. { onFailure: error => notify(error.message) }
  * - {boolean} withDeclarativeSideEffectsSupport Set to true to support legacy side effects e.g. { onSuccess: { refresh: true } }
@@ -140,7 +142,7 @@ const useMutation = (
         (
             callTimeQuery?: Mutation | Event,
             callTimeOptions?: MutationOptions
-        ): void => {
+        ): void | Promise<any> => {
             const finalDataProvider = hasDeclarativeSideEffectsSupport(
                 options,
                 callTimeOptions
@@ -156,14 +158,17 @@ const useMutation = (
 
             setState(prevState => ({ ...prevState, loading: true }));
 
-            finalDataProvider[params.type]
+            const returnPromise = params.options.returnPromise;
+
+            const promise = finalDataProvider[params.type]
                 .apply(
                     finalDataProvider,
                     typeof params.resource !== 'undefined'
                         ? [params.resource, params.payload, params.options]
                         : [params.payload, params.options]
                 )
-                .then(({ data, total }) => {
+                .then(response => {
+                    const { data, total } = response;
                     setState({
                         data,
                         error: null,
@@ -171,6 +176,9 @@ const useMutation = (
                         loading: false,
                         total,
                     });
+                    if (returnPromise) {
+                        return response;
+                    }
                 })
                 .catch(errorFromResponse => {
                     setState({
@@ -180,7 +188,14 @@ const useMutation = (
                         loading: false,
                         total: null,
                     });
+                    if (returnPromise) {
+                        throw errorFromResponse;
+                    }
                 });
+
+            if (returnPromise) {
+                return promise;
+            }
         },
         [
             // deep equality, see https://github.com/facebook/react/issues/14476#issuecomment-471199055
@@ -204,13 +219,17 @@ export interface Mutation {
 export interface MutationOptions {
     action?: string;
     undoable?: boolean;
+    returnPromise?: boolean;
     onSuccess?: (response: any) => any | Object;
     onFailure?: (error?: any) => any | Object;
     withDeclarativeSideEffectsSupport?: boolean;
 }
 
 export type UseMutationValue = [
-    (query?: Partial<Mutation>, options?: Partial<MutationOptions>) => void,
+    (
+        query?: Partial<Mutation>,
+        options?: Partial<MutationOptions>
+    ) => void | Promise<any>,
     {
         data?: any;
         total?: number;
--- a/packages/ra-core/src/form/useInitializeFormWithRecord.ts
+++ b/packages/ra-core/src/form/useInitializeFormWithRecord.ts
@@ -19,10 +19,14 @@ const useInitializeFormWithRecord = record => {
         // Disable this option when re-initializing the form because in this case, it should reset the dirty state of all fields
         // We do need to keep this option for dynamically added inputs though which is why it is kept at the form level
         form.setConfig('keepDirtyOnReinitialize', false);
-        // Ignored until next version of final-form is released. See https://github.com/final-form/final-form/pull/376
-        // @ts-ignore
-        form.restart(initialValuesMergedWithRecord);
-        form.setConfig('keepDirtyOnReinitialize', true);
+        // Since the submit function returns a promise, use setTimeout to prevent the error "Cannot reset() in onSubmit()" in final-form
+        // It will not be necessary anymore when the next version of final-form will be released (see https://github.com/final-form/final-form/pull/363)
+        setTimeout(() => {
+            // Ignored until next version of final-form is released. See https://github.com/final-form/final-form/pull/376
+            // @ts-ignore
+            form.restart(initialValuesMergedWithRecord);
+            form.setConfig('keepDirtyOnReinitialize', true);
+        });
     }, [form, JSON.stringify(record)]); // eslint-disable-line react-hooks/exhaustive-deps
 };
 
--- a/packages/ra-ui-materialui/src/detail/CreateView.tsx
+++ b/packages/ra-ui-materialui/src/detail/CreateView.tsx
@@ -66,7 +66,10 @@ export const CreateView = (props: CreateViewProps) => {
                                 ? redirect
                                 : children.props.redirect,
                         resource,
-                        save,
+                        save:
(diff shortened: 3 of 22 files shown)


fzaninotto (changes requested):
Awesome! Can you add some usage documentation in https://marmelab.com/react-admin/CreateEdit.html#validation, and a unit test in `useMutation.spec.tsx`?

packages/ra-core/src/dataProvider/useMutation.ts
```diff
@@ -180,7 +188,14 @@ const useMutation = (
                         loading: false,
                         total: null,
                     });
+                    if (returnPromise) {
+                        throw errorFromResponse;
+                    }
                 });
+
+            if (returnPromise) {
+                return promise;
```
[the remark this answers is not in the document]
alanpoulain: Fixed.

alanpoulain:
PR has been updated to add:
- documentation
- a unit test for `useMutation`
- allowing `save` to be defined in `SimpleForm`
- handle `submitError` in inputs

I've tested it on a project. With this code:
```jsx
const Create = (props) => {
  const [mutate] = useMutation()
  const save = useCallback(
    async (values) => {
      try {
        await mutate({
          type: 'create',
          resource: 'origins',
          payload: { data: values },
        }, { returnPromise: true });
      } catch (error) {
        return {
          url: 'a custom server error',
        };
      }
    },
    [mutate],
  )

  return (
    <Wrapper>
      <RACreate undoable={false} {...props}>
        <SimpleForm toolbar={<Toolbar />} save={save} redirect='/origins'>
          <InputGuesser source='name' fullWidth variant='outlined' />
          <InputGuesser source='url' fullWidth variant='outlined' />
          <InputGuesser source='comment' fullWidth variant='outlined' />
        </SimpleForm>
      </RACreate>
    </Wrapper>
  )
}
```

The result is:

https://user-images.githubusercontent.com/10920253/105028370-3b611100-5a51-11eb-8b27-0562303df507.mp4

alanpoulain:
Rebased against `next`.
The unit test `useQueryWithStore › should refetch the dataProvider on refresh` seems to be flaky.
Locally, sometimes it passes, sometimes not:
image
image

packages/ra-ui-materialui/src/detail/CreateView.tsx
```diff
@@ -66,7 +66,10 @@ export const CreateView = (props: CreateViewProps) => {
                                 ? redirect
                                 : children.props.redirect,
                         resource,
-                        save,
+                        save:
```
fzaninotto, line 69: you should do the same for EditView

packages/ra-core/src/form/useInitializeFormWithRecord.ts
```diff
@@ -19,10 +19,13 @@ const useInitializeFormWithRecord = record => {
         // Disable this option when re-initializing the form because in this case, it should reset the dirty state of all fields
         // We do need to keep this option for dynamically added inputs though which is why it is kept at the form level
         form.setConfig('keepDirtyOnReinitialize', false);
-        // Ignored until next version of final-form is released. See https://github.com/final-form/final-form/pull/376
-        // @ts-ignore
-        form.restart(initialValuesMergedWithRecord);
-        form.setConfig('keepDirtyOnReinitialize', true);
+        // Since the submit function returns a promise, use setTimeout to prevent the error "Cannot reset() in onSubmit()" in final-form
+        setTimeout(() => {
```
fzaninotto, line 23: This is really concerning to me. Any error in the form.restart() won't be caught. 

Can you explain why the error occurs, why the setTimeout removes it, and check if there is no other way? And can you add a unit test?

fzaninotto (changes requested):

packages/ra-core/src/form/useInitializeFormWithRecord.ts
alanpoulain: This is the way `final-form` is telling to do it: https://github.com/final-form/final-form/blob/ee1ef7272882a966644ca1bf0ea6f115a2492251/src/FinalForm.js#L933.
However this constraint seems to have been removed in the next version:
https://github.com/final-form/final-form/pull/363
I don't see how we could add a unit test for this.

packages/ra-ui-materialui/src/detail/CreateView.tsx
alanpoulain: Oh yes I forgot! Fixed.

packages/ra-core/src/form/useInitializeFormWithRecord.ts
alanpoulain: I've added a comment to explain that the `setTimeout` should be removed when the next version of `final-form` will be released.
And actually there is already a test for this:
https://github.com/marmelab/react-admin/blob/baaf3da901ccd68dc8d42f0a2747a9f78e8eadff/cypress/integration/edit.js#L260-L277
(it throws the linked error without the `setTimeout`)

djhi:
> The unit test useQueryWithStore › should refetch the dataProvider on refresh seems to be flaky

Indeed. Looking into it

fzaninotto (approved):

fzaninotto:
Thank you very much!

PR #5782, opened 2021-01-15
Fix return type of Exporter

ohbarye:
ref https://github.com/marmelab/react-admin/pull/5771#issuecomment-760469318

The return type of `Exporter` has been wrong, this pull request fixes it.


djhi (approved):
Indeed. Thanks!

ohbarye:
thanks for your quick review too!

PR #5787, opened 2021-01-16
Fix and improve onSuccess/onFailure typings

### LINKED ISSUE #5761: Create prop onSuccess has wrong typing
According to the [CreateProps typings](https://github.com/marmelab/react-admin/blob/872480a9a4a8491066eb8b1f0f651e48f47e9080/packages/ra-ui-materialui/src/types.ts#L50), the onSuccess callback has the type `(data: Record) => void`.

However, it seems that the true implementation in the [useCreateController](https://github.com/marmelab/react-admin/blob/872480a9a4a8491066eb8b1f0f651e48f47e9080/packages/ra-core/src/controller/details/useCreateController.ts#L167) the type `({ data } : { data: Record}) => void` is used.

Right now the two are incompatible. I would suggest to change the type as this breaks no implementation based on the current behaviour?

HendrikRoehm:
The onSuccess type declaration was not consistent with the usage in useDataProvider. Additionally, there was an erroneus definition in the material-ui package.

There were several locations, where the typings were not defined yet for onSuccess and onFailure. useDeclarativeSideEffects and useDataProvider suggested the commited types.

Fixes #5761

### DIFF
--- a/packages/ra-core/src/controller/saveModifiers.tsx
+++ b/packages/ra-core/src/controller/saveModifiers.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { createContext, useRef } from 'react';
+import { Record } from '../types';
 
 export const SideEffectContext = createContext<SideEffectContextValue>({});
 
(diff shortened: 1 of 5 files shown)


fzaninotto:
As this PR needed rebase and was incomplete, and because I had to write the types to properly understand them, I redid the job in #5853. 

thanks a lot for your contribution, it put me on the right track!

PR #5794, opened 2021-01-18
Fix Create method API call URL example

### LINKED ISSUE #5792: Create POST url documentation is inaccurate in ja-data-json-server package
image

In the create route, it says `POST http://my.api.url/${resource}/${id}`  but it should `POST http://my.api.url/${resource}` as you wouldn't have the ID-- and the actual code backs this up:

image

tjsturos:
The Create method API URL does not have an ID.  The resource is not
created yet.
Fixes #5792


djhi (approved):
Thanks!

fzaninotto (approved):

PR #5799, opened 2021-01-19
Optimistic mode

### LINKED ISSUE #5133: Allow optimistic mode exit immediately inside useMutate options.onSuccess callback

fzaninotto:
## Problem

The `undoable` mode is optimistic AND undoable. React-admin offers no way to use the dataProvider in an optimistic and non-unduable way.

## Solution

Replace the 'undoable' boolean option in useDataProvider by a fetch mode the can be one of 'regular', 'optimistic', and 'undoable'. Implement that fetch mode in `useDataProvider`.

Closes #5133, supersedes #5258

## Example usage

```tsx
const [updateMany, { loading }] = useUpdateMany(
        resource,
        selectedIds,
        { views: 0 },
        {
            action: CRUD_UPDATE_MANY,
            onSuccess: () => {
                notify('ra.notification.updated', 'info', {
                    smart_count: selectedIds.length,
                }); // note that the notification isn't undoable
                unselectAll(resource);
            },
            onFailure: error =>
                notify(
                    typeof error === 'string'
                        ? error
                        : error.message || 'ra.notification.http_error',
                    'warning'
                ),
            mode: 'optimistic', // <= that's what's new
        }
    );
```

## Tasks

- [x] Proof of concept
- [x] Improve naming
- [x] Add documentation
- [x] Add unit tests


examples/simple/src/posts/ResetViewsButton.js
```diff
@@ -34,7 +31,7 @@ const ResetViewsButton = ({ resource, selectedIds }) => {
                         : error.message || 'ra.notification.http_error',
                     'warning'
                 ),
-            undoable: true,
+            mode: 'optimistic',
```
djhi, line 34:
```suggestion
            mutationMode: 'optimistic',
```

packages/ra-core/src/dataProvider/getDataProviderCallArguments.ts
```diff
@@ -8,6 +8,7 @@ const OptionsProperties = [
     'onFailure',
     'onSuccess',
     'undoable',
+    'mode',
```
djhi, line 11:
```suggestion
    'mutationMode',
```

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -143,6 +143,7 @@ const useDataProvider = (): DataProviderProxy => {
                         undoable = false,
                         onSuccess = undefined,
                         onFailure = undefined,
+                        mutationMode = 'pessimistic',
```
djhi, line 146: You can probably initialize `mutationMode` using the `undoable` option here and avoid referencing `undoable` below

```suggestion
                        mutationMode = undoable ? 'undoable' : 'pessimistic',
```

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -180,13 +184,18 @@ const useDataProvider = (): DataProviderProxy => {
...
                     if (isOptimistic) {
                         // in optimistic mode, all fetch calls are stacked, to be
                         // executed once the dataProvider leaves optimistic mode.
                         // In the meantime, the admin uses data from the store.
-                        if (undoable) {
+                        if (
+                            undoable ||
+                            mutationMode === 'undoable' ||
+                            mutationMode === 'optimistic'
+                        ) {
```
djhi, line 198:
```suggestion
                        if (
                            undoable ||
                            mutationMode === 'undoable'
                        ) {
```

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -247,34 +257,154 @@ const doQuery = ({
...
+
+/**
+ * In optimistic mode, the hook dispatches an optimistic action and executes
+ * the success side effects right away. Then it immediately calls the dataProvider.
+ *
+ * We call that "optimistic" because the hook returns a resolved Promise
+ * immediately (although it has an empty value). That only works if the
+ * caller reads the result from the Redux store, not from the Promise.
+ */
+const performOptimisticQuery = ({
```
djhi, line 313: There is a lot of duplication between `performOptimisticQuery` and `performUndoableQuery`. Maybe we can have smaller functions to compose both behaviours?

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -89,7 +90,7 @@ export const EditView = (props: EditViewProps) => {
                             resource,
                             save,
                             saving,
-                            undoable,
+                            undoable: undoable || mutationMode === 'undoable',
```
djhi, line 93: Shouldn't we pass `mutationMode` as well ?

djhi (changes requested):
Looks pretty good!

packages/ra-core/src/dataProvider/useDataProvider.ts
fzaninotto: No, I think this change is wrong. We want optimistic calls to be replayed first, just like undoable calls, after existing optimistic mode, and before pessimistic calls. The variable `undoableOptimisticCalls` is badly named, though.

packages/ra-core/src/dataProvider/useDataProvider.ts
fzaninotto: If you look at the details, there are subtle differences here and there (e.g. handling undo and the beforeunload event listener in the performUndoableQuery). I'll extract the functions to make the comparison easier.

packages/ra-ui-materialui/src/detail/EditView.tsx
fzaninotto: yep, done

djhi (approved):
Way better!

fzaninotto:
Switching to RFR

docs/Actions.md
```diff
@@ -478,13 +478,27 @@ const ApproveButton = ({ record }) => {
...
+In the previous example, after clicking on the "Approve" button, a loading spinner appears while the data provider is fetched. Then, users are redirected to the comments list. But in most cases, the server returns a success response, so the user waits for this response for nothing. 
 
-For its own fetch actions, react-admin uses an approach called *optimistic rendering*. The idea is to handle the calls to the `dataProvider` on the client side first (i.e. updating entities in the Redux store), and re-render the screen immediately. The user sees the effect of their action with no delay. Then, react-admin applies the success side effects, and only after that, it triggers the call to the data provider. If the fetch ends with a success, react-admin does nothing more than a refresh to grab the latest data from the server. In most cases, the user sees no difference (the data in the Redux store and the data from the data provider are the same). If the fetch fails, react-admin shows an error notification, and forces a refresh, too.
+This is called **pessimistic rendering**, as all users are forced to wait because of the (usually rare) possibilit yof server failure.
```
djhi, line 483:
```suggestion
This is called **pessimistic rendering**, as all users are forced to wait because of the (usually rare) possibility of server failure. 
```

docs/Actions.md
```diff
@@ -478,13 +478,27 @@ const ApproveButton = ({ record }) => {
...
+This is called **pessimistic rendering**, as all users are forced to wait because of the (usually rare) possibilit yof server failure. 
 
-As a bonus, while the success notification is displayed, users have the ability to cancel the action *before* the data provider is even called.
+An alternative mode for mutations is **optimistic rendering**. The idea is to handle the calls to the `dataProvider` on the client side first (i.e. updating entities in the Redux store), and re-render the screen immediately. The user sees the effect of their action with no delay. Then, react-admin applies the success side effects, and only after that, it triggers the call to the data provider. If the fetch ends with a success, react-admin does nothing more than a refresh to grab the latest data from the server. In most cases, the user sees no difference (the data in the Redux store and the data from the data provider are the same). If the fetch fails, react-admin shows an error notification, and forces a refresh, too.
```
djhi, line 485:
```suggestion
An alternative mode for mutations is **optimistic rendering**. The idea is to handle the calls to the `dataProvider` on the client side first (i.e. updating entities in the Redux store), and re-render the screen immediately. The user sees the effect of their action with no delay. Then, react-admin applies the success side effects, and only after that, it triggers the call to the data provider. If the fetch ends with a success, react-admin does nothing more than a refresh to grab the latest data from the server. In most cases, the user sees no difference (the data in the Redux store and the data from the `dataProvider` are the same). If the fetch fails, react-admin shows an error notification, and forces a refresh, too.
```

docs/Actions.md
```diff
@@ -478,13 +478,27 @@ const ApproveButton = ({ record }) => {
...
 
-You can benefit from optimistic rendering when you call the `useMutation` hook, too. You just need to pass `undoable: true` in the `options` parameter:
+A third mutation mode is called **undoable**. It's like optimistic rendering, but with an added feature: after applying the changes and the side effects locally, react-admin *waits* for a few seconds before triggering the call to the data provider. During this delay, the end user sees an "undo" button that, when clicked, cancels the call to the data provider and refreshes the screen.
```
djhi, line 487:
```suggestion
A third mutation mode is called **undoable**. It's like optimistic rendering, but with an added feature: after applying the changes and the side effects locally, react-admin *waits* for a few seconds before triggering the call to the `dataProvider`. During this delay, the end user sees an "undo" button that, when clicked, cancels the call to the `dataProvider` and refreshes the screen.
```

docs/Actions.md
```diff
@@ -478,13 +478,27 @@ const ApproveButton = ({ record }) => {
...
+
+|                   | pessimistic               | optimistic | undoable  |
+|-------------------|---------------------------|------------|-----------|
+| dataProvider call | immediate                 | immediate  | delayed   |
+| local changes     | when dataProvider returns | immediate  | immediate |
+| side effects      | when dataProvider returns | immediate  | immediate |
+| cancellable       | no                        | no         | yes       |
+
+
+For the Edit view, react-admin uses the undoable mode. For the Create view, react-admin needs to wait for the response to know the id of the resource to redirect to, so the mutation mode is pessimistic.
```
djhi, line 499:
```suggestion
By default, react-admin uses the undoable mode for the Edit view. For the Create view, react-admin needs to wait for the response to know the id of the resource to redirect to, so the mutation mode is pessimistic.  
```

djhi (approved):
Awesome :)

PR #5802, opened 2021-01-20
Fix warning when passing partial props to useListContext and other view context hooks

Luwangel:
React-admin wraps lists, create, edit and show views inside contexts. The main goal is to avoid passing props from the top component to every child one by one.

For example an Edit View is composed of multiple layout elements:
- A form
- An action toolbar
- A title
...

And these layout elements are composed of atomic elements (and some of them are iterator displaying smaller elements):
- Datagrid (iterator)
- ArrayField (iterator)
- TextInput
- Select

Thanks to the modularity of react-admin, you can write your own atomic elements.

Nevertheless, from a developer perspective, it's difficult to know which props are implicitly injected from the top because there are too many levels of components. And the reason why we added the contexts.

For example, imagine you want to create a custom iterator (a customized list of elements). Thanks to the contexts, you just need to connect the `useListContext` hook and directly get the props you need like the loaded state, the data, the total number of elements.

---

When we developed the contexts, we deprecated the __props mechanism__ with the intention to remove it in the next major version (v4). During this time, we encouraged the users to change their custom code by displaying a warning:

> "Show components must be used inside a <ShowContext.Provider>. Relying on props rather than context to get Show data and callbacks is deprecated and won't be supported in the next major version of react-admin."

----

But, as a result, some smaller components became unusable without surrounding them by the appropriate context. It's the case of the `<Datagrid>` which now calls the `useListContext`.

And it's not a pleasant developer experience because it means adding a `<ListContext>` around your custom iterator. That's why we decided to rely on both props and context.

Starting from this PR, you can override the value of the `useListContext`, the `useEditContext`, the `useCreateContext and the `useShowContext` hooks.

It's as simple as passing arguments to a function, and the props will take precedence over the values of the context.

`const context = useListContext(props);`

 If there is no context, it works too like this `<Datagrid>` (see the screenshot attached):

 ``` js
<Datagrid
    // You should pass the following props to make it work without a context
    basePath=""
    currentSort={currentSort}
    data={data}
    ids={ids}
    selectedIds={selectedIds}
    loaded={loaded}
    total={total}
>
    <TextField source="id" sortable={false} />
    <TextField source="title" sortable={false} />
</Datagrid>
```

## Todo

- [x] Refact the useListContext, the `useEditContext`, the `useCreateContext` and the `useShowContext` hooks to use props 
- [x] Add an example in the simple example
- [ ] Write documentation

## Screenshot

It's now possible to use a `Datagrid` without putting it inside a `ListContext`:

Sélection_004

### DIFF
--- a/packages/ra-core/src/controller/details/useCreateContext.tsx
+++ b/packages/ra-core/src/controller/details/useCreateContext.tsx
@@ -65,7 +65,7 @@ const extractCreateContextProps = ({
     saving,
     successMessage,
     version,
-}): CreateControllerProps<RecordType> => ({
+}) => ({
     basePath,
     record,
     defaultTitle,
--- a/packages/ra-core/src/controller/details/useCreateContext.tsx
+++ b/packages/ra-core/src/controller/details/useCreateContext.tsx
@@ -1,4 +1,4 @@
-import { useContext } from 'react';
+import { useContext, useMemo } from 'react';
 import merge from 'lodash/merge';
 
 import { Record } from '../../types';
@@ -32,12 +32,16 @@ export const useCreateContext = <
         // @ts-ignore
         CreateContext
     );
-
     // Props take precedence over the context
-    // @ts-ignore
-    return props != null
-        ? merge({}, context, extractCreateContextProps(props))
-        : context;
+    return useMemo(
+        () =>
+            merge(
+                {},
+                context,
+                props != null ? extractCreateContextProps(props) : {}
+            ),
+        [context, props]
+    );
 };
 
 /**
@@ -65,7 +69,7 @@ const extractCreateContextProps = ({
     saving,
     successMessage,
     version,
-}) => ({
+}: any) => ({
     basePath,
     record,
     defaultTitle,
--- a/packages/ra-core/src/controller/details/useCreateContext.tsx
+++ b/packages/ra-core/src/controller/details/useCreateContext.tsx
@@ -1,8 +1,27 @@
 import { useContext } from 'react';
+import merge from 'lodash/merge';
+
 import { Record } from '../../types';
 import { CreateContext } from './CreateContext';
 import { CreateControllerProps } from './useCreateController';
 
+/**
+ * Hook to read the create controller props from the CreateContext.
+ *
+ * Mostly used within a <CreateContext.Provider> (e.g. as a descendent of <Create>).
+ *
+ * But you can also use it without a <CreateContext.Provider>. In this case, it is up to you
+ * to pass all the necessary props.
+ *
+ * The given props will take precedence over context values.
+ *
+ * @typedef {Object} CreateControllerProps
+ *
+ * @returns {CreateControllerProps} create controller props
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/ra-core/src/controller/details/useCreateContext.tsx [hunk 1]
```diff
@@ -14,21 +33,54 @@ export const useCreateContext = <
...
-        }
-
-        return props;
-    }
-
-    return context;
+    // Props take precedence over the context
+    // @ts-ignore
+    return props != null
+        ? merge({}, context, extractCreateContextProps(props))
```
fzaninotto, line 39: You must memoize this, otherwise the hook will return a new object at each render.

packages/ra-core/src/controller/details/useEditContext.tsx
```diff
@@ -1,35 +1,86 @@
...
-            record: props.record || props.data,
-            data: props.record || props.data,
-        };
-    }
-
-    return context;
+    // Props take precedence over the context
+    // @ts-ignore
+    return props != null
+        ? merge({}, context, extractEditContextProps(props))
```
fzaninotto, line 35: same

packages/ra-core/src/controller/details/useShowContext.tsx
```diff
@@ -1,35 +1,62 @@
...
-            record: props.record || props.data,
-            data: props.record || props.data,
-        };
-    }
-
-    return context;
+    // Props take precedence over the context
+    // @ts-ignore
+    return props != null
+        ? merge({}, context, extractShowContextProps(props))
```
fzaninotto, line 35: same

packages/ra-core/src/controller/details/useShowContext.tsx
```diff
@@ -1,35 +1,62 @@
...
-        if (process.env.NODE_ENV !== 'production') {
-            console.log(
-                "Show components must be used inside a <ShowContext.Provider>. Relying on props rather than context to get Show data and callbacks is deprecated and won't be supported in the next major version of react-admin."
-            );
-        }
-        // Necessary for actions (EditActions) which expect a data prop containing the record
-        // @deprecated - to be removed in 4.0d
-        return {
-            ...props,
-            record: props.record || props.data,
```
fzaninotto, line 29: You didn't port it to extractShowContextProps as in the Edit version. I assume this isn't required for ShowActions?

fzaninotto (changes requested):

stensrud:
@Luwangel FYI, this PR broke our production system today. Seems `selectedIds` always generates a new object on every render, and our `useEffect()` with `selectedIds` in the parameter #2 skip-array was fired in an infinite loop.

We were on `react-admin` version `3.8.3` which depends on `ra-core` version `^3.8.3` which resolved to `3.12.5`.

fzaninotto:
@stensrud Would you mind opening a new issue for your problem, following the issue template (and providing a way to reproduce the problem)?

PR #5804, opened 2021-01-21
Add skip navigation button

andrico1234:
## Notes

I messed up the rebase, and I'm just opening up a new PR with the same changes. 

## Description

I've added a button inside of `Layout.tsx` that only displays when the first action a user performs on a RA site is pressing the tab button.

This moves the focus into the main content of the website, avoiding having to tab through the top and side menus.

While the functionality is complete, I'm going to write Cypress tests for this functionality.

Notes:

Because navigating to anchors via the `#` didn't seem to work nicely with `HashRouter`, I had to install an node module that exports a custom link. [Here's a link to the discussion](https://github.com/ReactTraining/react-router/issues/394#issuecomment-220221604)

Fortunately the module has no dependencies, and it works out of the box.

I implementing this component similarly to how it's written inside of the Material-UI repo. You can view their implementation [here](https://github.com/mui-org/material-ui/blob/69eb1d66a7bab366bc6b270faff2baf3cad5b4cd/docs/src/modules/components/AppFrame.js#L194)

https://user-images.githubusercontent.com/17087167/104626969-5f59d680-568e-11eb-8999-ce665b37d484.mov

### DIFF
--- a/cypress/integration/navigation.js
+++ b/cypress/integration/navigation.js
@@ -9,12 +9,24 @@ describe('Navigation', () => {
 
             ListPage.waitUntilVisible();
 
-            cy.get('body').tab().tab().tab();
+            cy.get('body').tab().tab().tab().tab();
 
             cy.get(`${ListPage.elements.menuItems}:first-child`).should(
                 'have.class',
                 'Mui-focusVisible'
             );
+        });
+    });
+
+    describe('Skip Navigation Button', () => {
+        it('should appear when a user immediately tabs on the homepage', () => {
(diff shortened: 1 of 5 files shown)


andrico1234:
This is no longer a WIP, I believe the code is exactly how the previous MR was before I messed up the. rebase

djhi (approved):
Thanks!

fzaninotto (approved):

fzaninotto:
Thanks a lot, this is a great feature!

PR #5806, opened 2021-01-21
[Doc] Undoable is for edit only

alanpoulain:
A reference to the `undoable` prop in the `Create` component was still present in the documentation.


djhi (approved):
Thanks!

PR #5812, opened 2021-01-22
added possibility to pass isRequired as option to useInput hook

### LINKED ISSUE #5122: UseInput isRequired and meta error
I have an HOC to manage multilangue on TextInput and I would like my text input display my error message and add required on label message.

I override name of TextInput with 'source+language selected'.
I have a required validate function with default language which verify the default language is setted.
structure : 
{ fr: "mon texte", en : "my text"}
If I pass my custom required validate function to textinput : 
the value send to validator is only "mon texte" and display error all the time.

If I remove this validate, my validator works but the error is not displayed. Moreover, the label hasn't got the required option.

I would like the isRequired is initialized like input or meta in useInput :

```
 // If there is an input prop, this input has already been enhanced by final-form
    // This is required in for inputs used inside other inputs (such as the SelectInput inside a ReferenceInput)
    if (options.input) {
        return {
            id: id || source,
            input: options.input,
            meta: options.meta,
            isRequired: isRequired(validate),
        };
    }
```

Manage the display error if my text input receive error message in meta props.

**Additional context**
It's complicated to add an source example because it's a large project.

FACOLOMBANI:
Fixes #5122

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -25,6 +25,7 @@ export interface InputProps<T = any>
     resource?: string;
     source: string;
     validate?: Validator | Validator[];
+    isRequired?: boolean;
 }
 
 export interface UseInputValue extends FieldRenderProps<any, HTMLElement> {
@@ -41,6 +42,7 @@ const useInput = ({
     onBlur: customOnBlur,
     onChange: customOnChange,
     onFocus: customOnFocus,
+    isRequired: isRequiredOption,
     ...options
 }: InputProps): UseInputValue => {
     const finalName = name || source;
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/form/useInput.ts
```diff
@@ -97,7 +97,7 @@ const useInput = ({
             id: id || source,
             input: options.input,
             meta: options.meta,
-            isRequired: isRequired(validate),
+            isRequired: options.isRequired || isRequired(validate),
```
djhi, line 100: Would you mind moving the `isRequired` out of the options and directly in the `InputProps` as an optional boolean?

djhi (changes requested):

packages/ra-core/src/form/useInput.ts
FACOLOMBANI: I made the change, is it o like that?

djhi (approved):
Thanks!

PR #5820, opened 2021-01-27
Update docs with latest Hasura Data Provider

cpv123:
## Description

Hasura have replaced their original (V1) data provider with https://github.com/Steams/ra-data-hasura-graphql.

There is now only one Hasura data provider: https://github.com/hasura/ra-data-hasura


djhi:
Thanks!

PR #5822, opened 2021-01-27
Fix race condition in WithPermissions

fzaninotto:
When an app has two components calling `<WithPermissions>` on the same render phase, a race condition used to lead the second call to never return the actual permissions.

```jsx
const Test = () => (
  <>
    <WithPermissions component={Foo} />
    <WithPermissions component={Bar} />
  </>
);
```

`Foo` was rendered once with empty permissions, then rerendered with the permissions Fetched from the authProvider.  
`Bar` was only rendered once, with empty permissions. 

This was a bug due to the permissions from the authProvider being compared to the cache rather than to the initial state.

### DIFF
--- a/packages/ra-core/src/auth/usePermissionsOptimized.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissionsOptimized.spec.tsx
@@ -5,6 +5,8 @@ import usePermissionsOptimized from './usePermissionsOptimized';
 import AuthContext from './AuthContext';
 
 describe('usePermissionsOptimized', () => {
+    afterEach(cleanup);
+
     const CallPermissionsOnMount = ({ number, authParams }: any) => {
         const { permissions } = usePermissionsOptimized(authParams);
         return (
--- a/packages/ra-core/src/auth/usePermissionsOptimized.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissionsOptimized.spec.tsx
@@ -1,0 +1,109 @@
+import * as React from 'react';
+import { render, act, cleanup } from '@testing-library/react';
+import expect from 'expect';
+import usePermissionsOptimized from './usePermissionsOptimized';
+import AuthContext from './AuthContext';
+
+describe('usePermissionsOptimized', () => {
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/auth/usePermissionsOptimized.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+            </AuthContext.Provider>
+        );
+        expect(renders).toBe(1); // renders on mount
+        expect(getPermissions).toBeCalledTimes(2);
+        expect(queryByText2('admin')).not.toBeNull(); // answer from the cache
+        await act(async () => await new Promise(r => setTimeout(r)));
+        expect(renders).toBe(1); // does not rerender when the getPermissions returns the same permissions
+    });
+
+    it('can be called by two independent components', async () => {
```
fzaninotto, line 81: this test failed before the patch

packages/ra-core/src/auth/usePermissionsOptimized.ts
```diff
@@ -13,7 +13,7 @@ const emptyParams = {};
 
 // keep a cache of already fetched permissions to initialize state for new
 // components and avoid a useless rerender if the permissions haven't changed
-const alreadyFetchedPermissions = { '{}': [] };
+const alreadyFetchedPermissions = { '{}': undefined };
```
fzaninotto, line 16: This was another regression introduced in 3.11. Before, WithPermissions injected `undefined` on mount. After, it injected an empty array.

packages/ra-core/src/auth/usePermissionsOptimized.ts
```diff
@@ -71,7 +71,7 @@ const usePermissionsOptimized = (params = emptyParams) => {
                     error,
                 });
             });
-    }, [getPermissions, params, setState]);
+    }, [getPermissions, key]); // eslint-disable-line react-hooks/exhaustive-deps
```
fzaninotto, line 74: this is done on purpose, and there is no way around it

djhi (approved):

PR #5828, opened 2021-01-28
Fix onSuccess called without payload on undoable actions

### LINKED ISSUE #5819: Edit onSuccess prop doesn't work correctly
**What you were expecting:**
Expected to work as stated in the documentation

> The onSuccess function receives the response from the dataProvider call (dataProvider.create() or dataProvider.update()), which is the created/edited record (see the dataProvider documentation for details)

**What happened instead:**
Instead, onSuccess is called immediately after Save action , and the data parameter is empty

**Steps to reproduce:**
Copy the code from the documentation and apply

https://codesandbox.io/s/proud-water-10dbo?file=/src/posts/PostEdit.js
after adding code from documentation codeandbox hangs on save

* React-admin version:3.11.4

djhi:
Fixes #5819

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.spec.js
+++ b/packages/ra-core/src/dataProvider/useDataProvider.spec.js
@@ -289,6 +289,41 @@ describe('useDataProvider', () => {
             });
         });
 
+        it('should accept an onSuccess option to execute on success for optimistic actions', async () => {
+            const onSuccess = jest.fn();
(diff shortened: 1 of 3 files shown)


fzaninotto:
I don't think that's correct. You're calling the success side effect with the *request* payload, not the *response* payload.

Since we can't return the success payload for optimistic calls, we should document that, in this case, the payload is empty.

fzaninotto:
I'll update the documentation instead.

PR #5829, opened 2021-01-28
Ensure Resource Definition is Overridable with Props

djhi:


packages/ra-core/src/core/useResourceDefinition.ts
```diff
@@ -6,9 +6,34 @@ import { useResourceContext } from './useResourceContext';
...
 
+    if (
+        props.resource &&
+        (typeof props.hasCreate !== 'undefined' ||
+            typeof props.hasEdit !== 'undefined' ||
+            typeof props.hasList !== 'undefined' ||
+            typeof props.hasShow !== 'undefined')
+    ) {
+        return {
+            ...props,
```
fzaninotto, line 23: You should whitelist the props that can fit the resource definition type, otherwise the hook may return an incorrect object. Check at #5802 for an example.

fzaninotto (changes requested):

packages/ra-core/src/core/useResourceDefinition.ts
djhi: Props are typed to at least implement the required interface, isn't that enough?

fzaninotto: at compile time, yes, at runtime, no.

packages/ra-core/src/core/useResourceDefinition.ts
```diff
@@ -1,14 +1,36 @@
...
+    props: UseResourceDefinitionOptions
+): ResourceDefinition => {
     const resource = useResourceContext(props);
     const resources = useSelector(getResources);
-
-    return resources.find(r => r?.name === resource) || props;
+    const definition = resources.find(r => r?.name === resource) || props;
+    const { hasCreate, hasEdit, hasList, hasShow } = props;
+    return props != null
+        ? merge({}, definition, {
```
fzaninotto, line 18: you must memoize that, or the hook will always return a new object

packages/ra-core/src/core/useResourceDefinition.ts
```diff
@@ -1,14 +1,36 @@
...
  */
-export const useResourceDefinition = (props): ResourceDefinition => {
+export const useResourceDefinition = (
+    props: UseResourceDefinitionOptions
+): ResourceDefinition => {
     const resource = useResourceContext(props);
     const resources = useSelector(getResources);
-
-    return resources.find(r => r?.name === resource) || props;
+    const definition = resources.find(r => r?.name === resource) || props;
```
fzaninotto, line 15: the `|| props` here must be removed, I think

fzaninotto (changes requested):

fzaninotto (approved):

PR #5831, opened 2021-01-28
Add plugin events onAdd and onRemove to SimpleFormIterator

ValentinnDimitroff:
This is a feature I needed quite a few times - an option to plug in inside the SimpleFormIterator's events and supply default value for each new field (Especially needed when the children are larger amount and have nested arrays which will show blank without such option)

This is the refined version I came up with. If you find it useful but have comment on the implementation I will gladly collaborate.

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.spec.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.spec.tsx
@@ -226,232 +226,232 @@ describe('<SimpleFormIterator />', () => {
         expect(queryAllByText('ra.action.remove').length).toBe(0);
     });
 
-    it('should add children row on add button click', async () => {
-        const {
-            getByText,
-            queryAllByLabelText,
-            queryAllByText,
-        } = renderWithRedux(
-            <SaveContextProvider value={saveContextValue}>
-                <SideEffectContextProvider value={sideEffectValue}>
-                    <SimpleForm>
-                        <ArrayInput source="emails">
-                            <SimpleFormIterator>
-                                <TextInput source="email" />
-                            </SimpleFormIterator>
-                        </ArrayInput>
-                    </SimpleForm>
-                </SideEffectContextProvider>
-            </SaveContextProvider>
-        );
+    // it('should add children row on add button click', async () => {
+    //     const {
+    //         getByText,
+    //         queryAllByLabelText,
+    //         queryAllByText,
+    //     } = renderWithRedux(
+    //         <SaveContextProvider value={saveContextValue}>
+    //             <SideEffectContextProvider value={sideEffectValue}>
+    //                 <SimpleForm>
+    //                     <ArrayInput source="emails">
+    //                         <SimpleFormIterator>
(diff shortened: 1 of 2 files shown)


djhi:
Hi and thanks for your contribution. I see two new features here, not really dependant on one other. Would you mind explaining the use cases ? Besides, if we agree to this, can you add some tests and documentation ?

vdimitroff:
You are correct - the two features are not dependant on each other. I can split them into separate PRs if you prefer.

Use case scenarios:
- fieldInitialValue prop: I have a form with an array field which contains about 10 fields and a nested array. 
   1. Some of the fields are almost always with the same value which is used for calculation in combination with the rest information the user will fill out. It is convinient to load them by default and let the user enter only the  differences.
   2. I want the nested array to consist of exact number of cells and preload some of their fields also.

- onAdd prop: I want the count of the items in the array field to be restricted to some amount or upon condition. If I use the disableAdd prop for example I can restrict the user from this action but I don't have the opportunity to bring up a modal to explain the reason for the restriction or ask for reassurance and then continue. 
         Handle side effects onClick overall and embed business logic in the way the size of the array is growing.

- onRemove prop: 
   1. Analogously to the above - restricting the user or brining up a modal.
   2. For optimization purposes and business logic requrements I need to keep a cache of used values inside SelectInputs for example across all items of the array. 
    Unfortunately I can't get use of the useEffect clean up function for example because only the fields array is altered and not the component itself (If I remove an element diffrenet from the last one -> the component destroyed is not the one holding the field but the always the last one). So I need a way to trigger cache validation when an item is removed and possibly not fetching the form values each time.

I would definitely commit to providing tests and documentation to the above features.

fzaninotto:
This PR was closed automatically by GitHub when I merged `next` into `master` to prepare the 3.12 release - this was not intended. 

I'm reopening it.

djhi:
Regarding the `fieldInitialValue`, isn't the `defaultValue` on `ArrayInput` enough? We use it in the [`PostCreate`](https://github.com/marmelab/react-admin/blob/master/examples/simple/src/posts/PostCreate.js#L115) of the simple example.

The `onAdd` and `onRemove` make sense though.

vdimitroff:
Well, not really - first I thought I might have got it all wrong but after I tested with the codesandbox simple repo I am saw the proposed behaviour is different:

- defaultValue - initilizes the array as a whole with 1 member with the respective values and then each next member is undefined
- fieldInitialValue - does **not** initilize the array at first but **does** initilize every item (field) in the array when added

vdimitroff:
Hey there, could you give some update over the features so that I proceed with the ones that we decide to implement?

djhi:
Hi @ValentinnDimitroff! Thanks a lot for taking the time to implement this. It looks great. Would you mind adding some tests and updating the documentation accordingly?

ValentinnDimitroff:
Hey @djhi, I am happy that you find the suggested features useful! I would be more than pleased to provide the tests and documentation needed to support the features hopefully in a few days time!

fzaninotto:

PR #5848, opened 2021-02-02
Fix translatable inputs display

djhi:


fzaninotto (commented):
Nice! There are linter warnings, though.

fzaninotto (approved):

PR #5849, opened 2021-02-02
[RFR] Add enabled options to query hooks

### LINKED ISSUE #5827: Feature to disable useQuery and related specialised hooks
I'm building an autocomplete input for a form outside of a `Create/Edit` page and I'd like to only call the API if the search text has at least 1 character.

Here's the code I have (Autocomplete is coming from `mui-rff`):
```tsx
const CategoryAutocomplete = ({ name, label }: Props) => {
  const [searchText, setSearchText] = React.useState('')
  const debouncedSearchText = useDebounce(searchText, 300)

  const { ids, data: categories, loading: isLoading } = useGetList<Category>(
    'categories',
    {
      page: 1,
      perPage: 20,
    },
    { field: 'name', order: 'ASC' },
    {
      q: debouncedSearchText,
    }
  )
  const options = ids?.map((id) => categories?.[id]).filter(isPresent) || []
  return (
    <Autocomplete
      label={label || 'Category'}
      name={name}
      required
      loading={isLoading}
      options={options}
      getOptionValue={(option) => option.id}
      getOptionLabel={(option) => option.name}
      renderOption={(option) => <Option category={option} />}
      onInputChange={async (_e, v) => {
        setSearchText(v)
      }}
    />
  )
}
```

I'd like to be able to pass a boolean option like `enabled` that would prevent the hook from calling the API if it's `false`. This is what `react-query` is providing: https://react-query.tanstack.com/reference/useQuery#_top

image
 

I consider using `useMutation` and pass `getList` as its type but it doesn't feel idiomatic.

**Additional context**
Nothing else comes to my mind

ValentinH:
Closes #5827

# description

This PR introduces the `enabled` option to the `useQuery()`, `useQueryWithStore()` and all the corresponding specialised hooks. The goal is to be able to conditionally make a query. It can be useful for example for a query depending on another one

```ts
// fetch posts
const { ids, data: posts, loading: isLoading } = useGetList<Post>(
    'posts',
    { page: 1, perPage: 20 },
    { field: 'name', order: 'ASC' },
    {}
);

// then fetch categories for these posts
const { data: categories, loading: isLoading } = useGetMany<Category>(
    'categories',
    ids.map(id=> posts[id].category_id),
    // run only if the first query returns non-empty result
    { enabled: ids.length > 0 }
);
```

# open questions

- what should the return value contain when `enabled: false`? For now, it will contain the last value before `enabled: false`, i.e. the initial date on load and then the last result if `enabled` switches from `false` to `true`.

# todos
- [x] tests 
- [x] more documentation?

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -134,6 +134,7 @@ const useDataProvider = (): DataProviderProxy => {
                         onSuccess = undefined,
                         onFailure = undefined,
                         mutationMode = undoable ? 'undoable' : 'pessimistic',
+                        enabled = true,
                         ...rest
                     } = options || {};
 
@@ -156,6 +157,13 @@ const useDataProvider = (): DataProviderProxy => {
                         throw new Error(
                             'You must pass an onSuccess callback calling notify() to use the undoable mode'
                         );
+                    }
+                    if (typeof enabled !== 'boolean') {
+                        throw new Error('The enabled option must be a boolean');
+                    }
+
+                    if (enabled === false) {
+                        return Promise.resolve();
                     }
 
                     const params = {
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -163,7 +163,7 @@ const useDataProvider = (): DataProviderProxy => {
                     }
 
                     if (enabled === false) {
-                        return Promise.resolve();
+                        return Promise.resolve({});
                     }
 
                     const params = {
--- a/packages/ra-core/src/dataProvider/useGetList.ts
+++ b/packages/ra-core/src/dataProvider/useGetList.ts
@@ -8,15 +8,9 @@ import {
     Identifier,
     Record,
     RecordMap,
+    UseDataProviderOptions,
 } from '../types';
 import useQueryWithStore from './useQueryWithStore';
-
-interface UseGetListOptions {
-    onSuccess?: (args?: any) => void;
-    onFailure?: (error: any) => void;
-    enabled?: boolean;
-    [key: string]: any;
-}
 
 const defaultIds = [];
 const defaultData = {};
@@ -67,7 +61,7 @@ const useGetList = <RecordType extends Record = Record>(
     pagination: PaginationPayload,
     sort: SortPayload,
     filter: object,
-    options?: UseGetListOptions
+    options?: UseDataProviderOptions
 ): {
     data?: RecordMap<RecordType>;
     ids?: Identifier[];
--- a/packages/ra-core/src/dataProvider/useGetList.ts
+++ b/packages/ra-core/src/dataProvider/useGetList.ts
@@ -11,6 +11,13 @@ import {
 } from '../types';
 import useQueryWithStore from './useQueryWithStore';
 
+interface UseGetListOptions {
+    onSuccess?: (args?: any) => void;
+    onFailure?: (error: any) => void;
+    enabled?: boolean;
+    [key: string]: any;
+}
+
 const defaultIds = [];
 const defaultData = {};
 
@@ -60,7 +67,7 @@ const useGetList = <RecordType extends Record = Record>(
     pagination: PaginationPayload,
     sort: SortPayload,
     filter: object,
-    options?: any
+    options?: UseGetListOptions
 ): {
     data?: RecordMap<RecordType>;
     ids?: Identifier[];
--- a/packages/ra-core/src/dataProvider/useGetList.ts
+++ b/packages/ra-core/src/dataProvider/useGetList.ts
@@ -31,7 +31,10 @@ const defaultData = {};
  * @param {Object} pagination The request pagination { page, perPage }, e.g. { page: 1, perPage: 10 }
  * @param {Object} sort The request sort { field, order }, e.g. { field: 'id', order: 'DESC' }
  * @param {Object} filter The request filters, e.g. { title: 'hello, world' }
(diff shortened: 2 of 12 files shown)


fzaninotto:
You should implement the enabled prop only once in useDataProvider. When enabled is false, it should return an empty resolved Promise (just like for optimistic queries).

ValentinH:
I moved the implementation to the `useDataProvider` hook as you suggested.

One thing I noticed and that I'm not sure I'm happy with is that when `enabled: false`, the inital result is:
```ts
{
  data: {}
  error: null
  ids: []
  loaded: false
  loading: true
  total: undefined
}
```
I think that `loading` should be `false` at this point.

Similarly, if the query was enabled at some point and then disabled, the result is:
```ts
{
  data: {...Data}
  error: null
  ids: [...Ids]
  loaded: true
  loading: false
  total: 20
}
```

Here, I'm wondering if `loaded` should be `false`. But I'm not sure for this use case.


Regarding the tests, where would you write them? Inside `useDataProvider.spec.js` or the tests for the corresponding specialised hooks?

packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -57,7 +67,7 @@ const useGetList = <RecordType extends Record = Record>(
     pagination: PaginationPayload,
     sort: SortPayload,
     filter: object,
-    options?: any
+    options?: UseGetListOptions
```
fzaninotto, line 70: use the existing `UseDataProviderOptions` instead

packages/ra-core/src/dataProvider/useQuery.ts
```diff
@@ -143,6 +144,7 @@ export interface Query {
 
 export interface QueryOptions {
     action?: string;
+    enabled?: boolean;
```
fzaninotto, line 147: You should add options.enabled to the dependency list of the useEffect, as you did in usequeryWithStore

fzaninotto (changes requested):
For the result in case of disabled query, I'm not sure this is really important. And I have no idea of what a disabled query should return. Feel free to suggest what you think is best.

As for unit tests, please add tests for the enabled options the existing useDataProvider and useMutation tests.

packages/ra-core/src/dataProvider/useQuery.ts
ValentinH: Having `options.enabled` in the useEffect there is actually a left-over that I forgot to remove when removing the early return in the hook. As the logic is now handled in `useDataProvider` and this option is part of the `requestSignature`, I don't think this is  needed.

packages/ra-core/src/dataProvider/useGetList.ts
ValentinH: Good point, I did it as well for `useGetOne()`.

ValentinH:
The only think that seems meaningful to me regarding the `result` is having `loading: false` on start. However, I agree that this is not important and can be dealt with in the application.

I'll work on the tests later today.

ValentinH:
@fzaninotto I pushed a first test and I noticed that because we return an empty resolved Promise when enabled is false, the data provider resolves with `undefined` while the contract says that it should resolved with `GetOneResult<Record>`.

I don't think this is right.

fzaninotto:
How do you suggest we fix this Type problem? I see that react-query types the data as optional in the response type.

ValentinH:
This is actually not only a type problem but can lead to runtime error: see [this line](https://github.com/marmelab/react-admin/pull/5849/commits/7d82d1cf91a2899ff7900d2674db7f86f4331a83#diff-1328e81086fb9773afa87fb8f5faf09ea221b330d475a8b332fb452aad247008R336) in the test I created. The problem is not that `data` is missing but that the whole result is missing.

In the case of getOne, I think we should resolve with `{ data: null }`. However, it means that we need a special case for every data provider method which feels off.

ValentinH:
@fzaninotto I'm not sure how to proceed with this PR, do you have any suggestions please?

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -157,6 +158,13 @@ const useDataProvider = (): DataProviderProxy => {
                             'You must pass an onSuccess callback calling notify() to use the undoable mode'
                         );
                     }
+                    if (typeof enabled !== 'boolean') {
+                        throw new Error('The enabled option must be a boolean');
+                    }
+
+                    if (enabled === false) {
+                        return Promise.resolve();
```
fzaninotto, line 166: what if you resolve with an empty object `{}`?

ValentinH: Yes this would solve the issue shown in the unit test. If this is fine enough for you, I can do this.

fzaninotto: That's what we do e.g. in [performOptimisticQuery](https://github.com/marmelab/react-admin/blob/a478d36fbadb507b5623fce3b1b25eca577f6baa/packages/ra-core/src/dataProvider/performQuery/performOptimisticQuery.ts#L118), so you're good to go.

fzaninotto:
Caution: your PR needs rebase

packages/ra-core/src/dataProvider/useDataProvider.ts
ValentinH: Nice then! I did the change and rebased my branch. 🙂

fzaninotto:
Excellent! Now you just have to document the new option and we're good to merge.

ValentinH:
Documentation added.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5851, opened 2021-02-02
[Doc] Fix onSuccess callback signature for optimistic and undoable queries

### LINKED ISSUE #5819: Edit onSuccess prop doesn't work correctly

fzaninotto:
Closes #5819

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -515,7 +515,8 @@ const ApproveButton = ({ record }) => {
         },
         {
 +           mutationMode: 'undoable',
-            onSuccess: ({ data }) => {
+-           onSuccess: ({ data }) => {
++           onSuccess: () => {
                 redirect('/comments');
 -               notify('Comment approved');
 +               notify('Comment approved', 'info', {}, true);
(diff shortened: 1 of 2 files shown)


docs/Actions.md
```diff
@@ -527,7 +528,7 @@ const ApproveButton = ({ record }) => {
 };
 ```
 
-As you can see in this example, you need to tweak the notification for undoable calls: passing `true` as fourth parameter of `notify` displays the 'Undo' button in the notification.
+As you can see in this example, you need to tweak the notification for undoable calls: passing `true` as fourth parameter of `notify` displays the 'Undo' button in the notification. Also, as side effects are executed immediately, they can't rely on the response beins passed to onSuccess.
```
WiXSL, line 531: `beins` => `being`

djhi (approved):

docs/Actions.md
fzaninotto: Good catch, thanks!

Issue #5834, opened 2021-01-29
Consider using `unknown` instead of `any`


antoine-pous:
Hi!

I'm currently discovering React Admin and this is a really cool project, i'm really impressed :+1: 

I'm using typescript in a really strict way, and i noticed that your project use `any` type in many places instead of `unknown`. `any` is mostly a hack which disable the type checking, `unknown` remain typesafe which is what we expect in any cases.

Can you consider to refactor this part?

Have a nice day

djhi:
Very good point, thanks! Would you mind helping us?

fzaninotto:
From what I understand by reading https://stackoverflow.com/questions/51439843/unknown-vs-any/51439876#51439876, using `unknown` would require additional work by developers to set the type before using the value. I'm not sure we want that.

Could you give an concrete example of a type in the react-admin code that would benefit from `unknown`, and explain why with a concrete example?

fzaninotto:
No news for some time, closing.

PR #5853, opened 2021-02-02
[TypeScript] Fix onSuccess / onFailure types

### LINKED ISSUE #5761: Create prop onSuccess has wrong typing

fzaninotto:
Closes #5761
Supersedes #5787

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -10,7 +10,6 @@ import { Location, History, LocationState } from 'history';
 
 import { WithPermissionsChildrenParams } from './auth/WithPermissions';
 import { AuthActionType } from './auth/types';
-import { Mutation } from './dataProvider/useMutation';
 
 /**
  * data types
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -286,6 +286,8 @@ export type DataProviderProxy = {
 };
 
 export type MutationMode = 'pessimistic' | 'optimistic' | 'undoable';
+export type OnSuccess = (response?: any) => void;
+export type OnFailure = (error?: any) => void;
 
 export interface UseDataProviderOptions {
     action?: string;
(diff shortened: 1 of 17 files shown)


packages/ra-core/src/types.ts
```diff
@@ -286,6 +286,8 @@ export type DataProviderProxy = {
 };
 
 export type MutationMode = 'pessimistic' | 'optimistic' | 'undoable';
+export type OnSuccess = (response?: any) => void;
+export type OnFailure = (error?: any) => void;
```
djhi, line 290: We should probably avoid `any` in favor of `unknow`. See #5834

fzaninotto: I'm not sure I understand the benefits and shortcomings of `any` vs `unknown`. What I've read at https://stackoverflow.com/questions/51439843/unknown-vs-any tells me that there may be drawbacks for end users (like the obligation to perform type checking on an unknown value before using it).

Let's keep the two discussions separate.

djhi (changes requested):
There is an eslint warning ;)

HendrikRoehm:
Additionally, the main fix for #5761 is missing:
`OnSuccess` should be typed as `({ data } : { data: Record}) => void`

djhi:
@HendrikRoehm:

> Additionally, the main fix for #5761 is missing:

No, `data` is not always a `Record`. It could be an array of `Record` for `useGetList` or anything in the case of custom dataProvider functions

HendrikRoehm:
@djhi I guess the main problem here is that there exists at least 3 different kind of onSuccess properties, which expect different types. Plz check my issue: In the onCreateController, it **need** to be of the type if have mentioned it above. So at least the type there should be changed.

fzaninotto:
@HendrikRoehm You said:

> In the onCreateController, it need to be of the type I have mentioned above

What do you mean? Do you have a compilation error? If so, with which code? 

Also, it's only in dataProvider.create() that the onsuccess callback can expect a defined response with a data key.

djhi (approved):

PR #5857, opened 2021-02-04
Simple example vite

djhi:
Speeds up development dramatically. However, this won't work until we release the next version of Fakerest using modern bundlers (PR incoming)


examples/simple/index.html
```diff
@@ -1,19 +1,14 @@
...
-</head>
-
-<body>
-    <div id="root">
-    </div>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite App</title>
```
fzaninotto, line 6: Please change

examples/simple/index.html
```diff
@@ -1,19 +1,14 @@
 <!DOCTYPE html>
-<html>
-
-<head>
-    <title>React Admin</title>
-    <style>
-        body {
```
fzaninotto, line 7: This is required

examples/simple/package.json
```diff
@@ -1,49 +1,26 @@
...
-        "ra-data-fakerest": "^3.9.0",
-        "ra-i18n-polyglot": "^3.9.0",
-        "ra-input-rich-text": "^3.9.0",
-        "ra-language-english": "^3.9.0",
-        "ra-language-french": "^3.9.0",
-        "react": "^17.0.0",
-        "react-admin": "^3.9.0",
-        "react-dom": "^17.0.0"
-    }
+  "name": "simple-vite",
```
fzaninotto, line 2: why change the name?

examples/simple/package.json
```diff
@@ -1,49 +1,26 @@
...
+  "version": "0.0.0",
+  "scripts": {
+    "start": "vite",
```
fzaninotto, line 5: the dev script is used by codesandbox. Can you check that it can work with the vite version of the simple example?

fzaninotto (changes requested):

examples/simple/package.json
djhi: Ok codesandox will prefer `dev` over `start` but will use `start` if `dev` is not defined. However as I was unable to make Vite work on codesandbox, I added a webpack setup and a `dev` command

fzaninotto:
Needs rebase

examples/simple/index.html
```diff
@@ -0,0 +1,32 @@
...
+    <title>React Admin</title>
+    <style>
+      body {
+        margin: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script>window.global = window</script>
```
fzaninotto, line 15: why do we need this for webpack?

examples/simple/index-webpack.html
```diff
@@ -1,19 +1,19 @@
 <!DOCTYPE html>
-<html>
-
-<head>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
fzaninotto, line 5: Why did you change that?

fzaninotto (changes requested):

examples/simple/index.html
djhi: This is needed for one of our dependency, can't remember which one though. Not related to webpack

examples/simple/index-webpack.html
djhi: Because the old values didn't provide any value AFAIK. Besides, `minimum-scale` can be an accessibility issue for browsers that still respect them (those I use don't actually)

fzaninotto (approved):

PR #5858, opened 2021-02-04
Allow filters to have `null` value

jpetitcolas:
### Use Case

Our API allows to filter a list based on an `archived` attribute. By default (no `archived` filter), it returns only the non-archived entities.

Hence, we've implemented the following filter:

```js
    <Filter {...props}>
        <SelectInput
            source="archived"
            choices={[
                { id: null, name: 'All' },
                { id: false, name: 'Non-archived only' },
                { id: true, name: 'Archived only' },
            ]}
        />
    </Filter>
```
If we want to display all our entities, we need to pass the `null` value to the filter.

Before this fix, the `null` would be removed, hence replaced by an `undefined`. We would then have the non-archived only records.

### Workaround

We are using a workaround for now, setting the id to `"null"`, and replacing it in our custom `buildQuery` function. But that's sounds hack-ish, and this change may benefit to others.

### DIFF
--- a/packages/ra-core/src/util/removeEmpty.spec.ts
+++ b/packages/ra-core/src/util/removeEmpty.spec.ts
@@ -8,7 +8,7 @@ describe('removeEmpty', () => {
     });
 
     it('should remove the empty values with empty values', () => {
-        const input = { foo: '', bar: null };
+        const input = { foo: '', bar: undefined };
         expect(removeEmpty(input)).toEqual({});
     });
 
--- a/packages/ra-core/src/util/removeEmpty.ts
+++ b/packages/ra-core/src/util/removeEmpty.ts
@@ -6,10 +6,7 @@ const isObject = obj =>
 const isEmpty = obj =>
     obj instanceof Date
         ? false
-        : obj === '' ||
-          obj === null ||
-          obj === undefined ||
-          shallowEqual(obj, {});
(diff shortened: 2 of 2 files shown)


jpetitcolas:
One test is failing on the CI, yet it passes locally. Is it a flaky one? :)

fzaninotto:
Not that I know of. I relaunched all tests to see if it's flaky.

fzaninotto:
I'm afraid your patch breaks SelectInputs with the allowEmpty prop. I'll have to test it thoroughly to make sure it doesn't.

groomain:
How I could help to deploy this feature? We also have filters with null value. 
It's strange because it's working with QuickFilter and defaultValue when I arrive on the search. But then when I change a filter the null value disappear.

WiXSL:
> How could I help to deploy this feature? We also have filters with null value. It's strange because it's working with QuickFilter and defaultValue when I arrive on the search. But then when I change a filter the null value disappear.

First thing is to move this to next branch since master is not receiving new features

groomain:
hum ok I didn't know that master is not receiving new features until the release of version 4.

we will find a workaround. thanks.

fzaninotto:
Version 3 is feature freeze, and we won't accept new features there. Please open a new PR against V4.

Issue #5865, opened 2021-02-04
Create button doesn't re-render on `disabled` prop change


andrico1234:
I have a component:

```jsx
export function Actions(props) {
  const [isDisabled, setDisabledState] = useState(true);

  return (
    <TopToolbar {...sanitizeListRestProps(props)}>
      <Button
        onClick={() => {
          setDisabledState(!isDisabled);
        }}
        label="Toggle state"
      >
        <ClearAllIcon />
      </Button>
      <CreateButton disabled={isDisabled} label="Create Patient" />
    </TopToolbar>
  );
}

```

I expect the disabled state to toggle when after clicking the `Toggle state` button. Instead nothing happens, the CreateButton doesn't rerender.

This looks like a similar issue faced in this [ticket](https://github.com/marmelab/react-admin/pull/5014). I imagine the solution would be to add an additional check to see if the disabled prop has changed. 

I'm not too sure if this is a specific design choice to avoid this rerendering when this particular prop changes, but I can't see why it's not supported given [this example in the docs](https://marmelab.com/react-admin/List.html#actions) shows the `disabled` attribute being used. 

I've recreated this in the codesandbox: https://codesandbox.io/s/romantic-keldysh-bh10v?file=/src/posts/PostList.js

If you guys are fine with this, I can quickly get a PR going. I imagine it's a simple one line change

Edit: I ended up making an MR for this

PR #5866, opened 2021-02-04
rerender Create button on disabled prop change

andrico1234:
Related issue [here](https://github.com/marmelab/react-admin/issues/5865)

https://user-images.githubusercontent.com/17087167/106920462-03212a00-6703-11eb-871c-b01875605a5d.mov


djhi (approved):
Thanks!

PR #5867, opened 2021-02-04
Fix missing type for 'ra.navigation.skip_nav' translation message

bicstone:
Fix missing type for 'ra.navigation.skip_nav' translation message.

https://github.com/marmelab/react-admin/commit/d5d4a47e3f713cc1335a7032ca7ff17f27db078a#diff-d55e636245f2994de4a9a74002cca1fec0a6a2e68d1805a5439a0c885f37263b


djhi (approved):
Thanks!

PR #5880, opened 2021-02-07
docs (List): clarifies accepted List props

f-jost:
Hi, it's confusing to have two list of props defined in the docs, I think only one List props is sufficient under the l57 `Here are all the props accepted by the `<List>` component`.
Moreover in the second list it was missing `syncWithLocation` and `component` (that was missing in the two lists)


fzaninotto (approved):

fzaninotto:
You're right, thanks!

PR #5884, opened 2021-02-08
Fix DeleteWithConfirmButton Does Not Allow to Override resource

### LINKED ISSUE #5869: DeleteButton resource is not passed down
I have a resource named `answers` and another named `questions`. Within a route belonging to `answers` I have a list of `questions` displayed in a table. I want to use the `DeleteButton` to be able to delete a question from here. The `DeleteButton` accepts an attribute `resource` but it's only used to display the confirmation message. 
When the time comes that RA does the actual delete request, it gets the resource from the context, in this case the wrong resource `answers`. You can see here https://github.com/marmelab/react-admin/blob/0deb2b5d59db5d8720b8f6a2ba8468331567e19a/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx#L53 where the `resource` is no longer passed down.

To be able to pass `resource` to the `DeleteButton` and that property be passed down up until here https://github.com/marmelab/react-admin/blob/0deb2b5d59db5d8720b8f6a2ba8468331567e19a/packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx#L83 so that you are able to manually overwrite the context resource.

I managed to resolve my problem using this:
```
<ResourceContextProvider value="questions">
  <DeleteButton resource="questions" />
</ResourceContextProvider>
```
but it took me 1 hour of digging through your code, when the instinctively solution is to simply use the `resource` attribute in the button and have it pass down.

djhi:
Fixes #5869

### DIFF
--- a/packages/ra-ui-materialui/src/form/Toolbar.tsx
+++ b/packages/ra-ui-materialui/src/form/Toolbar.tsx
@@ -161,7 +161,10 @@ const Toolbar: FC<ToolbarProps> = props => {
                     Children.map(children, (button: ReactElement) =>
                         button && isValidElement<any>(button)
                             ? React.cloneElement(button, {
-                                  basePath,
+                                  basePath: valueOrDefault(
+                                      button.props.basePath,
+                                      basePath
+                                  ),
                                   handleSubmit: valueOrDefault(
                                       button.props.handleSubmit,
                                       handleSubmit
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/form/Toolbar.tsx
```diff
@@ -184,6 +184,7 @@ const Toolbar: FC<ToolbarProps> = props => {
                                       button.props.undoable,
                                       undoable
                                   ),
+                                  ...button.props,
```
fzaninotto, line 187: I don't understand this line. When we clone `button`, we clone it wit hits props. some are overridden starting at line 164, but when there is a possibility to set it directly on the button element, we use the `valueOrDefault` util. I think this is what you should do for the `resource` prop.

fzaninotto (changes requested):

fzaninotto (approved):

PR #5886, opened 2021-02-08
Add ListContext to ReferenceArrayInput

djhi:
- [x] Implementation
- [x] Ability to select rows to change the input value
- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceArrayInputContext.ts
+++ b/packages/ra-core/src/controller/input/useReferenceArrayInputContext.ts
@@ -1,4 +1,5 @@
-import { useContext } from 'react';
+import { useContext, useMemo } from 'react';
+import merge from 'lodash/merge';
 import {
     ReferenceArrayInputContext,
     ReferenceArrayInputContextValue,
@@ -13,10 +14,40 @@ export const useReferenceArrayInputContext = <
     props: T
 ): ReferenceArrayInputContextValue => {
     const context = useContext(ReferenceArrayInputContext);
-
-    if (props.choices) {
-        return props;
-    }
-
-    return context;
+    // Props take precedence over the context
+    return useMemo(
+        () =>
+            merge(
+                {},
+                context,
+                props != null
+                    ? extractReferenceArrayInputContextProps(props)
+                    : {}
+            ),
+        [context, props]
+    );
 };
+
+const extractReferenceArrayInputContextProps = <
+    T extends ReferenceArrayInputContextValue = ReferenceArrayInputContextValue
+>({
+    choices,
+    error,
+    loaded,
+    loading,
+    setFilter,
+    setPagination,
+    setSort,
+    setSortForList,
+    warning,
+}: T) => ({
+    choices,
+    error,
+    loaded,
+    loading,
+    setFilter,
+    setPagination,
+    setSort,
+    setSortForList,
+    warning,
+});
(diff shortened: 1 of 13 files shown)


fzaninotto (approved):

fzaninotto:
This should be PRed against `next`

packages/ra-core/src/controller/input/useReferenceArrayInputContext.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export const useReferenceArrayInputContext = <
+    T extends ReferenceArrayInputContextValue = ReferenceArrayInputContextValue
+>(
+    props: T
+): ReferenceArrayInputContextValue => {
+    const context = useContext(ReferenceArrayInputContext);
+
+    if (props.choices) {
+        return props;
```
fzaninotto, line 18: Please use a more fine-grained approach, as in useListContext in master

packages/ra-core/src/util/indexById.ts
```diff
@@ -0,0 +1,9 @@
+import { Record, RecordMap } from '../types';
+
+export const indexById = (records: Record[] = []): RecordMap =>
```
fzaninotto, line 3: please add jsDoc with sample input/output

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -141,19 +143,33 @@ const ReferenceArrayInput: FC<ReferenceArrayInputProps> = ({
         input,
     });
 
+    const listContext = useMemo(
+        () => ({
+            ...controllerProps,
+            setSort: controllerProps.setSortForList,
```
fzaninotto, line 149: Please add a jsDoc explaining why you do some voodoo here

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -141,19 +143,33 @@ const ReferenceArrayInput: FC<ReferenceArrayInputProps> = ({
...
-            isRequired={isRequired}
-            meta={meta}
-            translate={translate}
-            children={children}
-            {...props}
-            {...controllerProps}
-        />
+        <ResourceContextProvider value={props.reference}>
+            <ReferenceArrayInputContextProvider value={controllerProps}>
+                <ListContextProvider value={listContext}>
```
fzaninotto, line 159: I don't get it. What's the use of a ReferenceArrayInputContext if all you need is already in the ListContext?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
djhi: I agree it is mostly redundant now but it does provide the `warning` prop which indicates non "fatal" errors regarding references.

packages/ra-core/src/util/indexById.ts
djhi: Done

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
djhi: Done

fzaninotto (approved):

PR #5887, opened 2021-02-09
Handle arrays in ReferenceArrayField filter

### LINKED ISSUE #5871: Allow the filter prop of ReferenceArrayField to handle arrays
The `ReferenceArrayField` component has a `filter` prop which allows to filter results client-side. However if the filtered value is an array, the filter does not work because the component does a simple `==` comparison.
For example if I have a resource with a property `foo: ['bar', 'baz']`, then using the filter prop `{ foo: 'bar' }` will not return this property because `'bar' != ['bar', 'baz']`.

Change [this line](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/field/useReferenceArrayFieldController.ts#L177) in the `useReferenceArrayFieldController` to something like this:

```js
const recordValue = get(record, filterName);
return Array.isArray(recordValue) ? recordValue.includes(filterValue) : filterValue == recordValue;
```

Create our own component, but I think this is a feature others may want.

**Additional context**
I can do a PR if you think this is a good idea.

srosset81:
Closes #5871

### DIFF
--- a/packages/ra-core/src/controller/field/ReferenceArrayFieldController.spec.tsx
+++ b/packages/ra-core/src/controller/field/ReferenceArrayFieldController.spec.tsx
@@ -202,4 +202,86 @@ describe('<ReferenceArrayFieldController />', () => {
         expect(dispatch.mock.calls[0][0].type).toBe('RA/CRUD_GET_MANY');
         expect(dataProvider.getMany).toBeCalledTimes(1);
     });
+
+    it('should filter string data based on the filter props', () => {
+        const children = jest.fn().mockReturnValue('child');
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Thanks! Would you mind adding some tests for it?

srosset81:
@djhi Done :-)

fzaninotto:
This is a new feature, and as such you should PR against `next` rather than `master`

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5890, opened 2021-02-09
[Typescript] Add missing classes prop on SimpleFormIterator

ValentinH:
In order to customise the layout of a form in `SimpleFormIterator`, I need to use the `classes` props. It works but the TS definition was not allowing it.

This simply adds the `classes` prop.

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
@@ -21,6 +21,7 @@ import classNames from 'classnames';
 import { FieldArrayRenderProps } from 'react-final-form-arrays';
 
 import FormInput from './FormInput';
+import { ClassesOverride } from '../types';
 
 const useStyles = makeStyles(
     theme => ({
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
```diff
@@ -319,6 +319,7 @@ export interface SimpleFormIteratorProps
     extends Partial<Omit<FieldArrayRenderProps<any, HTMLElement>, 'meta'>> {
     addButton?: ReactElement;
     basePath?: string;
+    classes?: Partial<ReturnType<typeof useStyles>>;
```
djhi, line 322: Would you mind using the `ClassesOverride` type? It can be imported from the `types.ts` file

```tsx
classes: ClassesOverride<typeof useStyles>;
```

djhi (changes requested):

packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx
ValentinH: Sure no problem.

ValentinH: Done.

djhi (approved):
Thanks!

PR #5891, opened 2021-02-09
Add scroll to top on navigation

fzaninotto:
## Problem

React-router [doesn't manage scroll changes on navigation](https://reactrouter.com/web/guides/scroll-restoration). As a consequence, when users click on an item low in the menu, or on the pagination controls, the new content may reveal up hidden in the non-visible area of the page. This is disturbing for end users.

scrollToTop

On the other hand, the page should not scroll to top everywhere, and especially not when using the back button.

## Solution

- [x] Enable scroll to top by default in `<List>`, `<Create>`, `<Edit>` and `<Show>`
- [x] Allow users to disable it using the `disableScrollToTop` prop
- [ ] Add documentation  

scrollToTop2

### DIFF
--- a/examples/demo/src/products/ProductList.tsx
+++ b/examples/demo/src/products/ProductList.tsx
@@ -20,6 +20,7 @@ import {
     TopToolbar,
     useListContext,
     useTranslate,
+    useScrollRestoration,
 } from 'react-admin';
 
 import GridList from './GridList';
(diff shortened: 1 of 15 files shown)


djhi (approved):

fzaninotto:
I'll take another approach based on adding a parameter to links rather than adding a watcher in pages. It'll be more flexible.

PR #5895, opened 2021-02-10
Ensure form children can override props

djhi:
This can be useful when you want to display a component using a different resource in the form.

- [x] Implementation
- [ ] Example use case

### DIFF
--- a/packages/ra-ui-materialui/src/form/FormInput.tsx
+++ b/packages/ra-ui-materialui/src/form/FormInput.tsx
@@ -29,7 +29,9 @@ const FormInput = <RecordType extends Record | Omit<Record, 'id'> = Record>(
 ) => {
     const { input, classes: classesOverride, ...rest } = props;
     const classes = useStyles(props);
-    const { id, ...inputProps } = input ? input.props : { id: undefined };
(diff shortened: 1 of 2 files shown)


fzaninotto:
Can you add a use case?

packages/ra-ui-materialui/src/form/FormInput.tsx
```diff
@@ -29,7 +29,9 @@ const FormInput = <RecordType extends Record | Omit<Record, 'id'> = Record>(
 ) => {
     const { input, classes: classesOverride, ...rest } = props;
     const classes = useStyles(props);
-    const { id, ...inputProps } = input ? input.props : { id: undefined };
+    const { id, className, ...inputProps } = input
+        ? input.props
+        : { id: undefined, className: undefined };
```
jdemangeon, line 34: Using an empty object will do the trick. No need to assign "undefined" since things are already "undefined" in the empty object

djhi: TypeScript disagree

fzaninotto (approved):

PR #5900, opened 2021-02-11
Fix List view error after delete when using a field with no record test

fzaninotto:
This fixes a nasty bug that occurs when:

- A list contains a Field that doesn't check if the record is defined before rendering it (e.g. `FooField = ({ record }) => <>{record.foo}</>` instead of `FooField = ({ record }) => <>{record ? record.foo : null}</>`
- The user deletes a record from the list using bulk actions

image

In addition, this bug caused an additional List rerender, so this should also speed up lists a bit.

Note to self: never use both a dataProvider hook and a useSelector in the same function. Because dataProvider hooks rely on useEffect, they will return one tick later than useSelector, resulting in data inconsistencies (in this case: a list of `ids` containing the id of a record already removed from the `data` key).

### DIFF
--- a/packages/ra-core/src/controller/useListController.spec.tsx
+++ b/packages/ra-core/src/controller/useListController.spec.tsx
@@ -153,7 +153,6 @@ describe('useListController', () => {
             const props = {
                 ...defaultProps,
                 debounce: 200,
-                crudGetList: jest.fn(),
                 children,
             };
 
@@ -171,6 +170,7 @@ describe('useListController', () => {
                                     params: {},
                                     cachedRequests: {},
                                     ids: [],
+                                    selectedIds: [],
                                 },
                             },
                         },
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/controller/useListController.spec.tsx
```diff
@@ -41,6 +42,73 @@ describe('useListController', () => {
         debounce: 200,
     };
 
+    describe('data', () => {
+        it('should be synchronized with ids after delete', async () => {
```
fzaninotto, line 46: this test fails on master

packages/ra-core/src/controller/useListController.ts
```diff
@@ -181,41 +178,12 @@ const useListController = <RecordType extends Record = Record>(
         }
     );
 
-    const data = useSelector(
```
fzaninotto, line 184: Logic moved to a special hook, useGetMainList

djhi (approved):

PR #5903, opened 2021-02-12
Fix TabbedForm Does not Display Errors in Hidden Tabs on Submit

djhi:


packages/ra-ui-materialui/src/form/FormTabHeader.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
```
fzaninotto, line 1: please export that component

packages/ra-ui-materialui/src/form/FormTabHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+    icon,
+    className,
+    ...rest
+}) => {
+    const translate = useTranslate();
+    const location = useLocation();
+    const formGroup = useFormGroup(value);
+    const form = useForm();
+    const [showError, setShowError] = React.useState(false);
+    form.subscribe(
```
fzaninotto, line 21: You must do that in a useEffect, and return the unsubscribe method

fzaninotto (changes requested):

packages/ra-ui-materialui/src/form/FormTabHeader.tsx
djhi: Done

fzaninotto (approved):

PR #5916, opened 2021-02-15
Fix ReferenceArrayInputView propTypes

ValentinH:
The propTypes for `resource` was marked as required while the TS definition was marked as optional.

I'm getting a warning in my application when using `ReferenceArrayInput` without a `resource` because of this.


djhi (approved):
Thanks!

PR #5918, opened 2021-02-15
Added global css override key to UserMenu component

WiXSL:


packages/ra-ui-materialui/src/layout/UserMenu.tsx
```diff
@@ -31,7 +34,7 @@ const UserMenu = props => {
     const handleClose = () => setAnchorEl(null);
 
     return (
-        <div className={classes.user}>
```
fzaninotto, line 34: this is a breaking change. Can you please revert it?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/layout/UserMenu.tsx
WiXSL: Done!

fzaninotto (approved):

djhi:
I think this should target the `master` branch

fzaninotto:
But the buttons doc is only in next... So let's keep it like that.

fzaninotto:
Thanks!

PR #5927, opened 2021-02-17
ReferenceField: fix label proptype

ValentinH:
Passing a React element is totally fine and is already allowed by the TS types.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5928, opened 2021-02-17
Fix simple example doesn't run on CodeSandbox

ValentinH:
The goal is to fix the [Codesandbox](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple) for the Simple example which is currently broken.

Here's an example codesandbox with the fix implemented in this PR: https://codesandbox.io/s/funny-colden-7p1in


examples/simple/package.json
```diff
@@ -42,6 +42,7 @@
         "ra-input-rich-text": "^3.9.0",
         "ra-language-english": "^3.9.0",
         "ra-language-french": "^3.9.0",
+        "ra-ui-materialui": "^3.9.0",
```
djhi, line 45: `ra-ui-materialui` is a dependency of `react-admin`, this shouldn't be necessary

djhi (changes requested):
I think we should update all the other react-admin dependencies to target `3.12.4` instead

examples/simple/package.json
ValentinH: Yes that's what I thought as well. I'll try to update all the other dependencies then .

fzaninotto:
I think the problem is in the CodeSandbox side; If you clone the defautl react-admin codesandbox, it works...

ValentinH:
I don't really know what was wrong but after updating all ra-related packages, it works fine. I've updated my PR.

fzaninotto (approved):

fzaninotto:
Except some of the packages don't have a version 3.12...

ValentinH:
Indeed 🙈 
I fixed that by putting all the latest versions.

fzaninotto:
your JSON is invalid. I suggest you test your patch before pushing it.

ValentinH:
Sorry for the inconvenience. 😞

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5929, opened 2021-02-17
Fix resource propType definitions in components using ResourceContext

WiXSL:


packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -32,9 +33,10 @@ const FilterButton = (props: FilterButtonProps): JSX.Element => {
         showFilter,
         classes: classesOverride,
         className,
-        resource,
+        resource: propsResource,
```
djhi, line 36: I think you can just remove the line

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
WiXSL: But the `resource` prop is going to be injected into the div at line 76

djhi (approved):

PR #5931, opened 2021-02-17
Fix useGetManyReference loading state detection

djhi:


packages/ra-core/src/dataProvider/useGetManyReference.ts
```diff
@@ -100,7 +100,8 @@ const useGetManyReference = (
                 defaultData
             ),
         }),
-        (state: ReduxState) => getTotal(state, relatedTo)
+        (state: ReduxState) => getTotal(state, relatedTo),
+        (data: any) => data.ids && data.ids.length > 0
```
fzaninotto, line 104: Why >0 ? The dataProvider my return an empty list

packages/ra-core/src/reducer/admin/references/oneToMany.ts
```diff
@@ -58,8 +58,9 @@ const oneToManyReducer: Reducer<OneToManyState> = (
 };
 
 export const getIds = (state: ReduxState, relatedTo: string) =>
-    state.admin.references.oneToMany[relatedTo] &&
-    state.admin.references.oneToMany[relatedTo].ids;
+    (state.admin.references.oneToMany[relatedTo] &&
+        state.admin.references.oneToMany[relatedTo].ids) ||
+    null;
```
fzaninotto, line 63: if the problem is `undefined` vs `null`, I'd rather you change the `isDataLoaded` to use `!= null` instead of `!== null` - less risk of breaking things.

packages/ra-core/src/controller/field/ReferenceManyFieldController.spec.tsx
```diff
@@ -7,6 +7,76 @@ import renderWithRedux from '../../util/renderWithRedux';
...
+                reference="bar"
+                target="foo_id"
+                basePath=""
+            >
+                {props => <ComponentToTest {...props} />}
+            </ReferenceManyFieldController>,
+            {
+                admin: {
+                    references: {
+                        oneToMany: { ids: [] },
```
fzaninotto, line 62: same. Also, you test with a list of empty ids, you miss a test wit ha list of non-empty ids

packages/ra-core/src/controller/field/ReferenceManyFieldController.spec.tsx
```diff
@@ -7,6 +7,76 @@ import renderWithRedux from '../../util/renderWithRedux';
...
+                reference="bar"
+                target="foo_id"
+                basePath=""
+            >
+                {props => <ComponentToTest {...props} />}
+            </ReferenceManyFieldController>,
+            {
+                admin: {
+                    references: {
+                        oneToMany: {},
```
fzaninotto, line 27: this is not the normal shape of the state. there is a key representing the reference between `reference` and `oneToMany`

fzaninotto (changes requested):

fzaninotto (approved):

Issue #4905, opened 2020-06-06
Extend SimpleFormIterator to enable/disable rendering of index


pefi1011:
**Is your feature request related to a problem? Please describe.**

The problem is UX-based - the index of an array is not always an useful/required/needed information

```
<ArrayInput >
          <SimpleFormIterator>
         ...
          </SimpleFormIterator>
   </ArrayInput>
```

**Describe the solution you'd like**

Conditional rendering of [typography element](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/form/SimpleFormIterator.js#L141)
```
                            <Typography
                                variant="body1"
                                className={classes.index}
                            >
                                {index + 1}
                            </Typography>
```
 

It would be in the same fashion like `disableRemove` and  `disableAdd`
```
<ArrayInput >
          <SimpleFormIterator disableIndex >
         ...
          </SimpleFormIterator>
   </ArrayInput>
```

**Describe alternatives you've considered**
As far as I understood, to disable rendering of Index I would need to implement my own SimpleFormIterator iterator

pefi1011:
Here is the PR: https://github.com/marmelab/react-admin/pull/4908

Any feedback will be highly appreciated

fzaninotto:
Fixed by #6305

PR #5932, opened 2021-02-17
[WIP] Added disableIndex prop to SimpleFormIterator to omit rendering of children indexes

WiXSL:
Implements #4905
Supersides #4908


WiXSL:
More time than expected.
I'll take a look at it later

PR #5934, opened 2021-02-18
Improve data provider types

andrico1234:
Only 3 param types required this

`UpdateParams`
`UpdateManyParams`
`CreateParams`

This small change allows us to pass through our data's types if we know what they'll be at this point, and defaulting to `any` if no type is passed through. The current types treat the `data` object as `any`.

As for naming, I'm not too keen on giving single letter names to the generic's arguments. I'll be happy to change `T` to `Data` or whatever you feel is most appropriate

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -173,9 +173,9 @@ export interface GetManyReferenceResult<RecordType = Record> {
     validUntil?: ValidUntil;
 }
 
-export interface UpdateParams {
+export interface UpdateParams<T = any> {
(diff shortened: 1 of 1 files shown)


andrico1234:
FYI I had trouble linking RA using the instructions in the `.README`  so I wasn't able to test the types in my own repo. AFAIK there aren't any uses of the dataProvider's `update`, `updateMany`, and `create` methods in any of the test repos. 

I might have completely missed it though!

fzaninotto (approved):

Issue #5888, opened 2021-02-09
Add isRowExpandable prop for DataGrid component for better user experience


gradolol:
I often have cases when I need to display information about a row with expand, but when I open it, I see empty data. 
For example, a list of events with children and parents. For a child event, I don't want to show "expand" because I know that this row doesn't have any data.

My solution is implement prop like isRowSelectable.

fzaninotto:
Fixed by #5941

PR #5941, opened 2021-02-21
[RFR] Added isRowExpandable prop to Datagrid component

WiXSL:
Implements #5888

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -1978,6 +1978,7 @@ Here are all the props accepted by the component:
 * [`rowStyle`](#row-style-function)
 * [`rowClick`](#rowclick)
 * [`expand`](#expand)
+* [`isRowExpandable`](#isrowexpandable)
 * [`isRowSelectable`](#isrowselectable)
 * [`optimized`](#performance)
 
@@ -2188,6 +2189,31 @@ const PostEdit = props => (
 const PostList = props => (
     <List {...props}>
         <Datagrid expand={<PostEdit />}>
+            <TextField source="id" />
+            <TextField source="title" />
+            <DateField source="published_at" />
+            <BooleanField source="commentable" />
+            <EditButton />
+        </Datagrid>
+    </List>
+)
+```
+
+### `isRowExpandable`
+
+You can customize which rows will allow to show an expandable panel below them using the `isRowExpandable` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows an expand button only for rows with an id greater than 300:
+
+```jsx
+const PostPanel = ({ id, record, resource }) => (
+    <div dangerouslySetInnerHTML={{ __html: record.body }} />
+);
+
+const PostList = props => (
+    <List {...props}>
+        <Datagrid 
+            expand={<PostPanel />}
+            isRowExpandable={row => row.id > 300}    
+        >
             <TextField source="id" />
             <TextField source="title" />
             <DateField source="published_at" />
--- a/docs/List.md
+++ b/docs/List.md
@@ -2201,7 +2201,7 @@ const PostList = props => (
 
 ### `isRowExpandable`
 
-You can customize which rows will allow to show an expandable panel below them using the `isRowExpandable` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows an expand button only for rows with an id greater than 300:
+You can customize which rows will allow to show an expandable panel below them using the `isRowExpandable` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows an expand button only for rows that has a detail to show:
 
 ```jsx
 const PostPanel = ({ id, record, resource }) => (
@@ -2212,7 +2212,7 @@ const PostList = props => (
     <List {...props}>
         <Datagrid 
             expand={<PostPanel />}
-            isRowExpandable={row => row.id > 300}    
+            isRowExpandable={row => row.has_detail}    
         >
             <TextField source="id" />
             <TextField source="title" />
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
@@ -140,6 +140,10 @@ const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
     } = useListContext(props);
     const version = useVersion();
 
+    const contextValue = useMemo(() => ({ isRowExpandable }), [
+        isRowExpandable,
+    ]);
+
     const updateSort = useCallback(
         event => {
             event.stopPropagation();
@@ -242,10 +246,6 @@ const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
     const all = isRowSelectable
         ? ids.filter(id => isRowSelectable(data[id]))
         : ids;
-
-    const contextValue = useMemo(() => ({ isRowExpandable }), [
-        isRowExpandable,
-    ]);
 
     /**
      * After the initial load, if the data for the list isn't empty,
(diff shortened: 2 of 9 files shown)


WiXSL:
Note: The `expand` prop is passed down to every row. I don't know if this is ok or should I pass `null` for rows for which `isRowExpandable` returns `false`

WiXSL:
**Note:**

With this implementation `expand` prop gets passed down to every `DatagridRow` regardless of the value of `isRowExpandable`.
If this is not ok, I could change it.
Now that I think about it is much more clean with a code like:

 `expand: (!isRowExpandable || isRowExpandable(data[id])) && expand,` in the `DatagridBody`, thus leaving untouched `DatagridRow` component and avoiding adding the `expandable` prop.

packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
```diff
@@ -57,6 +58,8 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                         record: data[id],
                         resource,
                         rowClick,
+                        expandable:
```
WiXSL, line 61: With this implementation expand prop gets passed down to every DatagridRow regardless of the value of isRowExpandable.
If this is not ok, I could change it.
Now that I think about it is much more clean with a code like:

expand: (!isRowExpandable || isRowExpandable(data[id])) && expand, in the DatagridBody, thus leaving untouched DatagridRow component and avoiding adding the expandable prop.

djhi (changes requested):
Thanks! I think you forgot to update the `computeNbColumns` function. Adding a test would be nice :)

djhi:
Please ignore the previous comment except for the test.

docs/List.md
```diff
@@ -2198,6 +2199,31 @@ const PostList = props => (
 )
 ```
 
+### `isRowExpandable`
+
+You can customize which rows will allow to show an expandable panel below them using the `isRowExpandable` prop. It expects a function that will receive the record of each `<DatagridRow>` and returns a boolean expression.  For instance, this code shows an expand button only for rows with an id greater than 300:
```
fzaninotto, line 2204: the example (id > 300) isn't very representative of a real-world problem. Please look for a better example.

fzaninotto (changes requested):
We're removing props injection little by little, and replace them with contexts. I think this is a good occasion to introduce a DatagridContext, which will contain only the isRowExpandable function for a start. That way, you don't need to update the underlying components to sanitize the prop.

docs/List.md
WiXSL: Ok

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
```diff
@@ -0,0 +1,14 @@
+import { useContext } from 'react';
+import DatagridContext, { DatagridContextValue } from './DatagridContext';
+
+export const useDatagridContext = (): DatagridContextValue => {
```
djhi, line 4: What we did to keep backward compatibility until v4 is make sure all hooks of this kind also accept props. If props containing the expected values are provided, we return them instead of the context.

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -246,83 +248,87 @@ const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
...
-                            <Checkbox
-                                className="select-all"
-                                color="primary"
-                                checked={
-                                    selectedIds.length > 0 &&
-                                    all.length > 0 &&
-                                    all.every(id => selectedIds.includes(id))
-                                }
-                                onChange={handleSelectAll}
+        <DatagridContextProvider value={{ isRowExpandable }}>
```
djhi, line 251: You should probably extract the context value in a `useMemo` before passing it to the Provider

djhi (changes requested):
Looks good so far!

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
djhi: You can look at the `useListContext` hook in the core package for inspiration

WiXSL: Done!

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
WiXSL: Done!

djhi (approved):

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.spec.tsx
```diff
@@ -51,9 +54,11 @@ describe('<DatagridRow />', () => {
     describe('rowClick', () => {
         it("should redirect to edit page if the 'edit' option is selected", () => {
             const { getByText, history } = renderWithRouter(
-                <DatagridRow {...defaultProps} rowClick="edit">
-                    <TitleField />
-                </DatagridRow>
+                <DatagridContextProvider value={dataGridContextValue}>
```
fzaninotto, line 57: You shouldn't need to add a context in your tests. It's a sign that you've broken backwards compatibility.

Instead, set the default context value to something that allows the component to work even when the developer hasn't defined the DatagridContext

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
```diff
@@ -0,0 +1,28 @@
...
+import DatagridContext, { DatagridContextValue } from './DatagridContext';
+import merge from 'lodash/merge';
+import { warning } from 'ra-core';
+
+export const useDatagridContext = (
+    props?: DatagridProps
+): DatagridContextValue => {
+    const context = useContext(DatagridContext);
+
+    warning(
```
fzaninotto, line 12: As the default context can be enough, and as users can override the context with props, we don't want to log a warning here.

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
```diff
@@ -0,0 +1,28 @@
...
+    );
+
+    return useMemo(
+        () =>
+            merge(
+                {},
+                context,
+                props != null
+                    ? { isRowExpandable: props.isRowExpandable }
+                    : { isRowExpandable: undefined }
```
fzaninotto, line 24:
```suggestion
                    : { }
```

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
WiXSL: OK

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.spec.tsx
WiXSL: Done!

fzaninotto: You haven't actually addressed this - you still modify existing tests to make them pass, and it's the sign of a breaking change.

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
```diff
@@ -0,0 +1,21 @@
+import { useContext, useMemo } from 'react';
+import { DatagridProps } from './Datagrid';
+import DatagridContext, { DatagridContextValue } from './DatagridContext';
+import merge from 'lodash/merge';
+import { warning } from 'ra-core';
```
fzaninotto, line 5: linter warning

packages/ra-ui-materialui/src/list/datagrid/useDatagridContext.ts
```diff
@@ -0,0 +1,21 @@
...
+export const useDatagridContext = (
+    props?: DatagridProps
+): DatagridContextValue => {
+    const context = useContext(DatagridContext);
+
+    return useMemo(
+        () =>
+            merge(
+                {},
+                context,
```
fzaninotto, line 16: the default context value is `undefined`. I think it's wrong, it should be an empty object.

fzaninotto (changes requested):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #5951, opened 2021-02-24
useQuery does not work with GET_LIST

Shumuu:
Using `GET_LIST` as `type` for `useQuery` does not work. It needs `getList`


djhi (approved):
Indeed, thanks!

PR #5953, opened 2021-02-24
Fix CheckboxGroup error style when in ReferenceArrayInput

### LINKED ISSUE #5952: CheckboxGroupInput doesn't display error in red when wrapped in a ReferenceArrayInput
**What you were expecting:**
I am expecting the label and the error message to be in red if there is an an error.

**What happened instead:**
Both label and error message remain grey, which makes error spotting more complicated.  

**Steps to reproduce:**
Simply wrap a `<CheckboxGroupInput>` element into a `<ReferenceArrayInput>` element, as follows:
```jsx
<ReferenceArrayInput source="roomsIds" resource="RoomOffer" reference="Room" helperText="Test">
  <CheckboxGroupInput label="My label" validate={required()} />
</ReferenceArrayInput>
```

Here is the problem:
Capture d’écran 2021-02-24 à 10 41 19

**Explanations:**
It seems the <ReferenceArrayInput> element is passing an `error` props to all children element. As it is not sanitized in `<CheckboxGroupInput>` element, it will remains always at `null` for the `<FormControl>` element.

**Current workaround:**
My workaround is to sanitize props passed to `<CheckboxGroupInput>`, by creating a new component and simply removing the error prop:
```jsx
const SanitizedCheckboxGroupInput = props => {
  const { error, ...rest } = props
  return <CheckboxGroupInput {...rest} />
}
```

**Environment**
* React-admin version: 3.11.1
* React version: 17.0.1
* Browser: any

djhi:
Fixes #5952
image


fzaninotto:
Can you please add a screenshot showing the style error?

fzaninotto (approved):

PR #5958, opened 2021-02-25
[Doc] Improve List's custom pagination sample

WiXSL:
The previous code sample toolbar only shows with more than one page.
This sample code is based in enabling/disabling buttons rater than hiding/showing toolbar and buttons


fzaninotto:
I don't think that's an improvement. Why show buttons when they are disabled?

WiXSL:
I'm showing the toolbar with the buttons at all times wich I think is visually more consistent tha having no toolbar (if having one page of data), then suddenlly a toolbar appears with one button (lets say you change a filter).

fzaninotto:
Allow me to disagree ;) I don't think it matters much, though - this is an example, and developers can choose to follow your way or the current way based on what they learn with the current snippet.

PR #5960, opened 2021-02-26
[Doc] Fix scrollable TabbedForm example

WiXSL:
It is true there is still a problem whit this (#5695) but at least the docs should point to the right direction plus avoiding an error


fzaninotto:
Thanks!

PR #5962, opened 2021-02-26
Reset submission error when the field is changed

### LINKED ISSUE #5938: Server validation errors not removed after field change
**What you were expecting:**

After this merged PR
https://github.com/marmelab/react-admin/pull/5778
I'm expecting that, after an error is returned from server, it will be shown on the field.
That error must disappear after the field value change, because it must be checked again on the server to validate

**What happened instead:**

After an error is returned from server, it appears correctly on the related field.
But that error **remains** after the field value change, leaving the field in error state

**Steps to reproduce:**

1) Add an error handler like this suggested in documentation:
https://marmelab.com/react-admin/CreateEdit.html#submission-validation
2) Having a requested field on server, leave it blank, submit the form and trigger the error.
3) The field is now correctly in error state
4) Try to change the content
5) The fields remain in error state
6) Fill the server-side required field
7) If you submit now, with the required field filled, the form gets submitted correctly, **but an error snackbar opens** (because there are errors on the form on the submission time)
**Environment**

* React-admin version: 3.12.4
* React version: 16.13.1
* Browser: Chrome 88.0.4324.150 on Windows 10

alanpoulain:
Fixes #5938.

Use https://github.com/ignatevdev/final-form-submit-errors to reset the submission error when the field is changed.

I'm not using the `SubmitErrorsSpy` component but a new hook instead to make it work in the `FormWithRedirect` component.

I've also opened a PR in https://github.com/ignatevdev/final-form-submit-errors to add the hook there: https://github.com/ignatevdev/final-form-submit-errors/pull/6.

### DIFF
--- a/packages/ra-core/src/form/FormWithRedirect.spec.tsx
+++ b/packages/ra-core/src/form/FormWithRedirect.spec.tsx
@@ -41,7 +41,7 @@ describe('FormWithRedirect', () => {
 
         expect(renderProp.mock.calls[1][0].pristine).toEqual(true);
         expect(getByDisplayValue('Foo')).not.toBeNull();
-        expect(renderProp).toHaveBeenCalledTimes(3);
+        expect(renderProp).toHaveBeenCalledTimes(2);
     });
 
     it('Does not make the form dirty when reinitialized from a different record', () => {
@@ -75,6 +75,6 @@ describe('FormWithRedirect', () => {
 
         expect(renderProp.mock.calls[1][0].pristine).toEqual(true);
         expect(getByDisplayValue('Foo')).not.toBeNull();
-        expect(renderProp).toHaveBeenCalledTimes(3);
+        expect(renderProp).toHaveBeenCalledTimes(2);
     });
 });
--- a/packages/ra-core/src/form/FormWithRedirect.spec.tsx
+++ b/packages/ra-core/src/form/FormWithRedirect.spec.tsx
@@ -41,7 +41,7 @@ describe('FormWithRedirect', () => {
 
         expect(renderProp.mock.calls[1][0].pristine).toEqual(true);
         expect(getByDisplayValue('Foo')).not.toBeNull();
-        expect(renderProp).toHaveBeenCalledTimes(2);
+        expect(renderProp).toHaveBeenCalledTimes(3);
     });
 
     it('Does not make the form dirty when reinitialized from a different record', () => {
@@ -75,6 +75,6 @@ describe('FormWithRedirect', () => {
 
         expect(renderProp.mock.calls[1][0].pristine).toEqual(true);
         expect(getByDisplayValue('Foo')).not.toBeNull();
-        expect(renderProp).toHaveBeenCalledTimes(2);
+        expect(renderProp).toHaveBeenCalledTimes(3);
     });
 });
--- a/packages/ra-core/src/form/FormWithRedirect.tsx
+++ b/packages/ra-core/src/form/FormWithRedirect.tsx
@@ -53,10 +53,7 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
     initialValues,
     initialValuesEqual,
     keepDirtyOnReinitialize = true,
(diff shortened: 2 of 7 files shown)


packages/ra-core/src/form/FormWithRedirect.spec.tsx
```diff
@@ -41,7 +41,7 @@ describe('FormWithRedirect', () => {
 
         expect(renderProp.mock.calls[1][0].pristine).toEqual(true);
         expect(getByDisplayValue('Foo')).not.toBeNull();
-        expect(renderProp).toHaveBeenCalledTimes(2);
+        expect(renderProp).toHaveBeenCalledTimes(3);
```
alanpoulain, line 44: Using `useFormState` triggers a rerender.

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -51,7 +53,10 @@ const FormWithRedirect: FC<FormWithRedirectProps> = ({
     initialValues,
     initialValuesEqual,
     keepDirtyOnReinitialize = true,
-    mutators = arrayMutators as any, // FIXME see https://github.com/final-form/react-final-form/issues/704 and https://github.com/microsoft/TypeScript/issues/35771
+    mutators = {
```
fzaninotto, line 56: I'd create a `defaultMutators` outside of the component, to avoid that the `mutators` value change at each render

packages/ra-core/src/form/useResetSubmitErrors.ts
```diff
@@ -0,0 +1,25 @@
+import { useRef } from 'react';
+import { useForm, useFormState } from 'react-final-form';
+
+/**
+ * Reset the submission error when the corresponding field changes.
+ * final-form doesn't do this by default.
+ */
+const useResetSubmitErrors = () => {
+    const { mutators } = useForm();
+    const { values } = useFormState({ subscription: { values: true } });
```
fzaninotto, line 10: I find it weird that you need to call `useFormState` twice, especially since you've already called `useForm`. Maybe that's why you had to update the `FormWithRedirect` unit test.

I assume you could do the same with  `form.subscribe()`, without even calling `useFormState` once. 
Could you try that and see if it removes the extra render?

packages/ra-core/src/form/FormWithRedirect.spec.tsx
fzaninotto: That's a problem... This feature impacts the performance of all forms to allow a feature that isn't so common...

fzaninotto (changes requested):
Encouraging!

alanpoulain:
@fzaninotto is it better now?

fzan:
sorry to jump in the discussion, but i wanted to show my appreciation to your work @alanpoulain 
excellent imho

djhi (approved):
Very nice! Thanks!

vladgoryachev:
Hi all,
After upgrading to a version that includes this change I am getting errors in a form with FormWithRedirect that is using mutators. 
_TypeError: form.mutators.resetSubmitErrors is not a function._

The error disappears when I remove mutators.

Should I open a new bug or there is another solution?

Vgor07:
I am getting an error message after upgrading to a version with this change.

```
TypeError: form.mutators.resetSubmitErrors is not a function

    in ReactFinalForm (created by FormWithRedirect)
    in FormContextProvider (created by FormWithRedirect)
    in FormWithRedirect
```

I am using mutators in FormWithRedirect to implement some data manipulation logic, and the form doesn't work anymore. Should I open a bug or there is a solution?

djhi:
If you added custom mutators, make sure you include the one we need as well. We should document that though

vladgoryachev:
Is there a mandatory mutator now? Which one is that?

Best regards,
Vlad

________________________________

djhi:
We need array mutators and https://github.com/ignatevdev/final-form-submit-errors which has been introduced in this very PR

PR #5972, opened 2021-02-28
Fix npm package name in installation

Kiailandi:
As stated on the [npm package page](https://www.npmjs.com/package/ra-data-local-storage) the correct name of the package is ra-data-local-storage not ra-data-localstorage as stated in Installation. Current command makes the installation fail.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #5973, opened 2021-03-01
Fix dataProvider: rm HTTP req body from deleteMany

tbrrt11:
Remove unnecessary body from request. It even raises a TS error: `Property 'data' does not exist on type 'DeleteManyParams'.ts(2339)`


fzaninotto:
Thanks!

PR #5977, opened 2021-03-01
Fix ReferenceArrayInput props Drilling

djhi:


packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -169,7 +169,12 @@ const ReferenceArrayInput = ({
                         translate={translate}
                         children={children}
                         {...props}
-                        {...sanitizeListRestProps(controllerProps)}
```
fzaninotto, line 172: linter warning

fzaninotto (changes requested):

Issue #5428, opened 2020-10-20
TabbedForm validation differences between development and production build


ronaldtb:
**What you were expecting:**
I was expecting the FormTab instances in a TabbedForm to turn red in case any of the child inputs are invalid. This is working as expected during development. The className "RaTabbedForm-errorTabButton-41" is being added to the FormTab.

Opening the page containing the 'Create' and 'TabbedForm' component shows all tabs as invalid even before pressing the save button.

**What happened instead:**
After running `yarn build` and deploying this production build the tabs are not being marked as invalid. The inspector is showing a className "jss40", but this class can't be found in the generated sources.

The tabs are not turning red on validation errors.

**Other information:**
Dev environment (yarn start)
![ReactAdminDev](https://user-images.githubusercontent.com/1379883/96588751-51412780-12e4-11eb-970a-ad23c9f32d3f.gif)

Prod environment (yarn build)
![ReactAdminProd](https://user-images.githubusercontent.com/1379883/96588754-51d9be00-12e4-11eb-8504-36658c708211.gif)

**Environment**

* React-admin version: 3.9.4
* Last version that did not exhibit the issue (if applicable):
* React version: 16.14.0
* Browser: Google Chrome 86.0.4240.80
* Stack trace (in case of a JS error):

djhi:
Our demo (https://marmelab.com/react-admin-demo/#/products/create) is a production build and the tabbed forms do show invalid tabs with correct styles so I believe there is an issue with your setup. We had similar issues in the past and they were all related to multiple conflicting versions of a dependency. Please double check yours, especially packages related to react-admin and material-ui.

ronaldtb:
Thanks for the quick response and I will check the demo and our dependency versions.

fzaninotto:
This issue has gone silent. Without any way to reproduce the issue, we have to close it. Feel free to comment with new info if you want it reopened.

donovantc:
@ronaldtb did you discover which packages were conflicting? I'm having a similar issue. Except the css is there but the MUI css is overriding the jss (react-admin css classes).

donovantc:
I've been struggling with this issue for a while now, so I decided to create a very basic setup using `create-react-app` and add `react-admin` as mentioned in the [react-admin tutorial](https://marmelab.com/react-admin/Tutorial.html).

I added a create component which has 2 tabs, and set all fields to _required_. 
[...]

PR #5984, opened 2021-03-03
Fix FormTabHeader Error State Display

djhi:
ra-tabbed-forms-errors


packages/ra-ui-materialui/src/form/FormTabHeader.tsx
```diff
@@ -44,6 +44,12 @@ export const FormTabHeader = ({
     );
 };
 
+const UseFormStateOptions = {
```
guilbill, line 47: Why not calling this only `FormStateOptions`? Is it by convention?

guilbill (approved):

packages/ra-ui-materialui/src/form/FormTabHeader.tsx
djhi: Because that's the options of the `useFormState` hook. No convention

fzaninotto (approved):

fzaninotto:
Closes #5428

PR #5988, opened 2021-03-03
[Doc] Fix missing DataProviderContext in Querying the API chapter

fzaninotto:


docs/Actions.md
```diff
@@ -50,7 +50,45 @@ const UserProfile = ({ userId }) => {
 };
 ```
 
-**Tip**: The `dataProvider` returned by the hook is actually a *wrapper* around your Data Provider. This wrapper dispatches Redux actions on load, success and failure, which keeps track of the loading state.
+**Tip**: The `dataProvider` returned by the hook is actually a *wrapper* around your Data Provider. This wrapper updates the Redux store on success, and keeps track of the loading state. In case you don't want to update the Redux store (e.g. when implementing an autosave feature), you should access the raw, non-wrapper Data Provider from the `DataProviderContext`:
```
djhi, line 53:
```suggestion
**Tip**: The `dataProvider` returned by the hook is actually a *wrapper* around your Data Provider. This wrapper updates the Redux store on success, and keeps track of the loading state. In case you don't want to update the Redux store (e.g. when implementing an autosave feature), you should access the raw, non-wrapped Data Provider from the `DataProviderContext`:
```

djhi (changes requested):

djhi (approved):

PR #5989, opened 2021-03-03
Fix forward compatibility between react-admin packages

### LINKED ISSUE #5979: Install react-admin 3.10.1 but get ra-core latest 3.13.0
**What you were expecting:**
 Has no exported member 'TestContext'. Did you mean 'ListContext'
**What happened instead:**

**Steps to reproduce:**

- Delete node_modules
- Yarn install

Package.json
`"react-admin": "3.10.1"`

**Related code:**

```
import { TestContext } from "react-admin";
```

**Other information:**

**Environment**

* React-admin version: 3.10.1
* React version: 16.13.1
* Browser: chrome

fzaninotto:
Closes #5979


djhi (commented):
I think we have more packages depending on `ra-core`. Shouldn't we fix them as well?

fzaninotto:
I don't want to, because then we hit problems when a package hasn't been updated since long (e.g. `ra-data-localstorage` depends on `ra-core@^3.9.0`). I think putting a `~` in react-admin dependencies and keeping a `^` in cross-package dependencies is the good balance between security and ease of publication.

PR #5991, opened 2021-03-03
[DOC] Improve design on search modal 

zyhou:
With the last version of Algolia ([docsearch package](https://github.com/algolia/DocSearch/tree/next)), we can easily improve the search experience.

We can use the modal component that allows us to bookmark a search and see the history of your searches. The modal can be open with shortcut (`ctrl+l` or `/`), no need to go back to the top of the page to do a search.

## Before

Empty search: image

"react context" search:
image

## After

The text color is white now, the contrast note is higher. Empty search:
 image

When I click on search button, the algolia modal appears:
image

When I search "react context": 
image

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -37,6 +37,10 @@
             rel="stylesheet"
             href="https://cdnjs.cloudflare.com/ajax/libs/tocbot/4.12.0/tocbot.css"
         />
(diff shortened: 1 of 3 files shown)


fzaninotto (approved):

fzaninotto:
Awesome!

PR #5994, opened 2021-03-04
[Doc] Synchronize Readme files

WiXSL:
I've just made some fixes, but the three README files are rather different.
I'm confused which data is the most updated or whether the three README files should have the same content
Let me know


fzaninotto (approved):
The readme at the root and under the react-admin package should be the same. The readme under docs is designed to be the root of the doc website, and therefore can be different.

fzaninotto:
Thanks!

PR #5995, opened 2021-03-04
Add ability to use record from context in Field components

fzaninotto:
The `record` prop injection in fields is magical, and a source of incomprehension for react-admin learners. Why do `<TextField source="title" />` work as a `<Datagrid>` child, but not in a custom Show layout?

We've recently introduced the `RecordContext`. It's time we enable support for this context in Field components, to remove the incomprehension.

This change is totally backwards compatible. It also prepares v4, where we'll remove the `record` prop injection.

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -25,17 +25,20 @@ export const PostList = (props) => (
 );
 ```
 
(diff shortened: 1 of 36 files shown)


fzaninotto:
Back to WIP: I forgot to set the correct record in Reference fields

fzaninotto:
And back to RFR

djhi (approved):

PR #5996, opened 2021-03-04
Add possibility to globally select variant and margin for MuiTextField

Mishastir:


packages/ra-ui-materialui/src/input/ResettableTextField.tsx
```diff
@@ -26,8 +26,8 @@ function ResettableTextField(
         value,
         resettable,
         disabled,
-        variant = 'filled',
-        margin = 'dense',
+        variant,
```
fzaninotto, line 29: there are other places where we set the default variant, e.g. in `SelectArrayInput`. Your current PR is incomplete and will break existing apps.

packages/ra-ui-materialui/src/input/ResettableTextField.tsx
```diff
@@ -155,10 +155,7 @@ function ResettableTextField(
             classes={restClasses}
             value={value}
             InputProps={{
-                classes:
-                    props.select && variant === 'filled'
```
fzaninotto, line 159: Why did you change that?

fzaninotto (changes requested):
I don't understand your purpose. There are other ways to change the default variant and margin of inputs (like exporting your own wrapper of each input with the props you need). Also, the Form elements already allow to override the input type. Why should we add a third way to define the input variant?

PR #5998, opened 2021-03-04
Add missing field in interface of <DeleteButton> props

DjebbZ:
`<DeleteButton>`, when passed `undoable={false}`, calls its child `<DeleteWithConfirmButton>` and passes it the props.

Problem: `<DeleteButton>` `Props` inferface is missing (at least) 2 props that `<DeleteWithConfirmButton>` accepts, namely `confirmTitle` and `confirmContent`. This prevents code from typechecking.

Please find an quick fix in this PR.


djhi (approved):
Thanks!

Issue #5985, opened 2021-03-03
<CheckboxGroupInput /> CSS Api inconsistent with its type


andrico1234:
According to the doc [here](https://marmelab.com/react-admin/Inputs.html#css-api-7),

the `CheckboxGroupInput`'s `classes` prop should be able accept the `label` key. In the screenshot below which is done in the `CheckboxGroupInput.spec.tsx` file, I'm through these props results and am getting a TypeScript error.

![image](https://user-images.githubusercontent.com/17087167/109790941-cfee8f80-7c09-11eb-823d-dfeda47ea7bb.png)

I didn't add an actual repro case because I felt the screenshot to be sufficient enough, since it's reproducible by adding a single line in the RA codebase.

~~Also, ignoring the TS error results in the styles not being applied at all.~~ I was wrong here

andrico1234:
Fixed here: https://github.com/marmelab/react-admin/pull/6001

Note that the label styles were always working, but I misunderstood the effect to be on each individual checkbox label, and not the top level label.

PR #6001, opened 2021-03-05
Bugfix: Prevent TS error on CheckboxGroupInput label styles

andrico1234:
Closes https://github.com/marmelab/react-admin/issues/5985

https://user-images.githubusercontent.com/17087167/110091540-75cc0680-7d90-11eb-9086-3b9d86d3b1d7.mov

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
@@ -16,6 +16,7 @@ import InputHelperText from './InputHelperText';
 import classnames from 'classnames';
 import Labeled from './Labeled';
 import { LinearProgress } from '../layout';
+import { ClassesOverride } from '../types';
 
 /**
  * An Input component for a checkbox group, using an array of objects for the options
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
```diff
@@ -80,7 +81,8 @@ import { LinearProgress } from '../layout';
  * The object passed as `options` props is passed to the material-ui <Checkbox> components
  */
 const CheckboxGroupInput: FunctionComponent<
-    ChoicesInputProps<CheckboxProps> & FormControlProps
+    ChoicesInputProps<CheckboxProps> &
+        FormControlProps & { classes?: ClassesOverride<typeof useStyles> }
```
djhi, line 85: At this point, I think we should extract this into a `CheckboxGroupInputProps` interface and export it (don't forget the index file)

djhi (changes requested):
Thanks for working on this!

packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
andrico1234: Done

fzaninotto:
Thanks!

PR #6003, opened 2021-03-05
Debounce AutocompleteArrayInput setFilter call

### LINKED ISSUE #5904: AutocompleteArrayInput should be debounced
**What you were expecting:**
If I type quickly 'test' in an `AutocompleteArrayInput`, only one request to the backend should be made with `q=test`.

**What happened instead:**
4 requests are sent, i.e. one per keystroke.
image

**Steps to reproduce:**
- Go to this sandbox: https://codesandbox.io/s/admiring-monad-d5707?file=/src/posts/TagReferenceInput.js
- edit a post
- go to the Miscellaneous tab
- type in the "tags" field

Observe the console and see that many requests are sent.

**Other information:**
A good practice for this kind of type-ahead input is too debounce the queries to avoid spamming the backend api.

**Environment**

* React-admin version: 3.12.2
* React version: 17.0.1
* Browser: Brave

djhi:
Fixes #5904


packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
```diff
@@ -18,7 +18,9 @@ import {
     ChoicesInputProps,
     useSuggestions,
     warning,
+    useWhyDidYouUpdate,
```
WiXSL, line 21: Unused import

fzaninotto (approved):

PR #6005, opened 2021-03-05
Fix Form Loose Dirty Fields Values After Cancelling Navigation

### LINKED ISSUE #5081: [SimpleForm, TabbedForm] warnWhenUnsavedChanges not working on a real api / data lost on cancel
**What you were expecting:**
From the docs:
React-admin keeps track of the form state, so it can detect when the user leaves an Edit or Create page with unsaved changes. To avoid data loss, you can use this ability to ask the user to confirm before leaving a page with unsaved changes.
**What happened instead:**
When the user clicks "cancel" the Form State is resetted to the server values, so the user looses his data anyway.

**Steps to reproduce:**

1. Click on posts
2. Edit any post
 *  I've added a datagrid of users via a reference field to the Post Edit
3. Click on a row to edit a user
4. Change a value in the user form
5. Leave the form without saving by clicking "Back button" !important
6. You get informed of not saved values.
7. Click cancel.
8. BUG: Form is resetted to server values, and the user is loosing his changes.

I've set up a codesandbox based on the react admin tutorial operating on the real json server api to reproduce it.
 
**Related code:**
https://codesandbox.io/s/peaceful-monad-0q6r5

**Other information:**

I suppose it has something to do with the way useWarnWhenUnsavedChanges Hook is working.
The previous view gets loaded, effectively refetching the data of the referenced data.
 
**Environment**

* React-admin version: 3.7.1
* React version: 16.13.1
* Browser: Chrome

djhi:
Fixes #5081

Tried locally with the same setup as in the issue


packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -1,17 +1,18 @@
 import * as React from 'react';
 import { FC, useRef, useCallback, useEffect, useMemo } from 'react';
```
fzaninotto, line 2: Linter warning

fzaninotto (changes requested):
Needs rebase

fzaninotto:
Needs rebase again

fzaninotto (approved):

fzaninotto:
This change likely caused a regression in the handling of default values in Edit forms (see #6198)

PR #6006, opened 2021-03-05
Export SimpleFormView Component

WiXSL:
It Is handy to be able to use the `SimpleFormView` when overriding the entire `SimpleForm`


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6009, opened 2021-03-07
Fix/redirect cant catch search params

vctqs1:
Since we manually add `state: { _scrollToTop: true }`. `history` cannot catch search params

### DIFF
--- a/packages/ra-core/src/sideEffect/useRedirect.ts
+++ b/packages/ra-core/src/sideEffect/useRedirect.ts
@@ -30,13 +30,13 @@ export type RedirectionSideEffect = string | boolean | RedirectToFunction;
  * redirect((redirectTo, basePath, is, data) => ...)
  */
 
-const useRedirect = ({ baseURL = '' }: { baseURL?: string }) => {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Hi, and thanks for your patch. I don't understand what it fixes though. Can you pleas describe what you're trying to achieve and what happens instead?

vctqs1:
> Hi, and thanks for your patch. I don't understand what it fixes though. Can you pleas describe what you're trying to achieve and what happens instead?

image


I tried to redirect to URL `/chapters/:id?parentId=:parentID`, But router cannot catch `search params`. Router understand `params id` about `id` and `search params`: `id: "fa93d683-823e-4716-b1f1-54e791f4dc7d?parentId=ee8431`

vctqs1:
@fzaninotto could you help check again?

My project's block at this :((

fzaninotto:
I still don't understand. Could you try again and include the code you're writing and that doesn't work?

fzaninotto:
Superseded by #6145

Issue #5898, opened 2021-02-11
Add an option to make ReferenceArrayInput lazy


ValentinH:
**Is your feature request related to a problem? Please describe.**
I'm building an edit form where one of the input is an `ArrayInput`. Each item of this input is a `SimpleFormIterator` with a few fields. Some of these fields are using a `ReferenceArrayInput` with an  `AutocompleteArrayInput`.

Here's an example:
```tsx
const Demo = (props: EditProps) => {
  return (
    <Edit {...props}>
      <SimpleForm redirect="/suspicious-articles-rules">
        <TextInput source="title" label="Title" validate={[required()]} />
        <TextInput source="description" multiline rows={4} />
        <ArrayInput source="items">
          <SimpleFormIterator>
            <ReferenceArrayInput
              reference="categories"
              source="categories.include"
              label="Include categories"
            >
              <AutocompleteArrayInput
                optionText={(option: Category | undefined) =>
                  option ? `${option.name} (${option.id})` : ''
                }
                shouldRenderSuggestions={(value: string) => value.trim().length >= 2}
              />
            </ReferenceArrayInput>
            <ReferenceArrayInput
              reference="categories"
              source="categories.exclude"
              label="Exclude categories"
            >
              <AutocompleteArrayInput
                optionText={(option: Category | undefined) =>
                  option ? `${option.name} (${option.id})` : ''
                }
                shouldRenderSuggestions={(value: string) => value.trim().length >= 2}
              />
            </ReferenceArrayInput>
            <NumberInput label="Price (from)" source="price_range.from" fullWidth />
          </SimpleFormIterator>
        </ArrayInput>
      </SimpleForm>
    </Edit>
  )
}
```

The `categories` resource has a lot of items and so I'd like to fetch them only if the user enters at least 2 characters. However, even if the autocomplete doesn't show the option, each `ReferenceArrayInput` is fetching the data with no `q` param (`/categories?_end=25&_order=DESC&_sort=id&_start=0`).

Another issue I noticed is that requests from the `ReferenceArrayInput` are not batched. In the above example, if I'm editing a resource with 3 items, there are 6 requests sent to `/categories?_end=25&_order=DESC&_sort=id&_start=0` when opening the form.

**Describe the solution you'd like**
Ideally, a `lazy` option would be nice. Its purpose would be to only fetch the data for an autocomplete if the `shouldRenderSuggestions` returns `true`.

**Describe alternatives you've considered**
[...]

PR #6013, opened 2021-03-08
[RFR] Add option to make ReferenceInput choices lazy

ValentinH:
This is actually implementing a similar feature than the one requested in #5898 but for `<ReferenceInput`.

# Why
The goal is to be able to control when the `useGetList` is `enabled` to avoid over-fetching the choices. In our application, we have a user autocomplete component and, as we have hundred of thousands of possible choices, we only want to start showing/fetching the choices if the user inputs at least 2 characters.

# Solution
This PR introduces an `isEnabled` option that is an optional function we use to set the `enabled` option of the `useGetList()`. This function is passed the `filterValues` array.

Example of usage:
```tsx
<ReferenceInput
  label="example"
  reference="users"
  source="user"
  isEnabled={({ q }) => {
    return !!q && q.length > 2
  }}
>
  <AutocompleteInput ... />
</ReferenceInput>
```

# Alternatives
Today, in order to work-around this, we pass a permanent filter `filter={{ is_autocomplete: true }}` that we use on our backend api to return early to avoid hitting the DB. However, we are still making calls for nothing.

<hr/>

I also looked to add a similar prop for the `ReferenceArrayInput` but, as it's using `useGetMatching`, it's a bit more complex.

If you think this option makes sense, I'd be happy to dig more to also add support for it for the `ReferenceArrayInput`.


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -121,7 +122,10 @@ export const useReferenceInputController = (
         loaded: possibleValuesLoaded,
         loading: possibleValuesLoading,
         error: possibleValuesError,
-    } = useGetList(reference, pagination, sort, filterValues);
+    } = useGetList(reference, pagination, sort, filterValues, {
+        action: 'CUSTOM_QUERY',
+        enabled: isEnabled && isEnabled(filterValues),
```
fzaninotto, line 127: This won't work when isEnabled isn't set

fzaninotto (changes requested):
Nice! This will be a welcome addition. 

You need to add unit tests and documentation to have this PR ready to merge.

packages/ra-core/src/controller/input/useReferenceInputController.ts
ValentinH: I actually relied on the `enabled` option default value being `true`. However, I can make it `true` explicitly if you think this is clearer.

ValentinH:
OK, happy that you like it. I'll add the tests and docs then.

What do you think about my comments regarding `ReferenceArrayInput` ?

docs/Inputs.md
```diff
@@ -1646,6 +1648,16 @@ The child component receives the following props from `<ReferenceInput>`:
 - `setPagination`: : function to call to update the pagination of the request for possible values
 - `setSort`: function to call to update the sorting of the request for possible values
 
+**Tip** You can make the `getList()` call lazy by using the `isEnabled` prop. This can be useful when using an `AutocompleteInput` on a resource with a lot of data. The following example only starts fetching the options when the query has at least 2 characters:
```
fzaninotto, line 1651: I think that 'isEnabled' is too vague a name. ReferenceInput does a lot of things, fetching choices is only one of them. I'd prefer something like 'enableGetChoices'

You should also explain what value this prop expects.

fzaninotto (changes requested):
Nice! Feel free to do the same for ReferencearrayInput, too.

docs/Inputs.md
ValentinH: Done

ValentinH:
From what I see `useReferenceArrayInputController()` uses `useGetMatching()` and the `enabled` option is actually not working properly there: when the `enabled` flag is false, the `loading` prop is stuck to `true` which will show the loading indicator. I need to dig more into this and probably fix the `enabled` prop behaviour for `useGetMatching`.

If you agree, I'll do this in a follow-up PR.

ValentinH:
I just pushed the support for the `enableGetChoices` prop on the `ReferenceArrayInput`.

One small thing that I find misleading is that the `filterValues` always contain the `q` param for `ReferenceInput` but not for `ReferenceArrayInput`. This enforces checking for `q` being present in the  `ReferenceArrayInput` `enableGetChoices` function:

image

Do you think this is something that should be taken care of or is fine like this?

ValentinH:
@fzaninotto do you think this PR is ready to be merged or is it still missing something?

ValentinH:
Why is this being closed?

fzaninotto:

ValentinH:
I think this PR was closed again by Github. I believe it is still valid.

fzaninotto:
You're right, I'm reopening it.

dgkm:
It's a great feature.  Any idea in which release this PR is getting merged and will be available?

djhi (approved):

ValentinH:
Thanks for merging this PR! 🎉

djhi:
Thanks!

PR #6016, opened 2021-03-08
Introduce useResourceLabel hook

djhi:
- [x] Introduce hook
- [x] Use the hook where possible (menu, page titles, etc.)


packages/ra-core/src/core/useGetResourceLabel.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param resource The resource name
+ * @param pluralize A boolean indicating whether the resource label should be pluralized.
+ * @param lowFirstLetter A boolean indicating whether the first letter of the resource label should be lower cased.
+ */
+export const useGetResourceLabel = (): GetResourceLabel => {
+    const resources = useSelector(getResources);
+    const translate = useTranslate();
+
+    return (resource: string, pluralize = true): string => {
```
fzaninotto, line 17: Why is it a higher-order function and why isn't the hook simply returning the label? This needs to be explained in the jsDoc I think.

packages/ra-core/src/core/useResourceContext.ts
```diff
@@ -11,10 +11,7 @@ import { ResourceContext, ResourceContextValue } from './ResourceContext';
  *
  * const ResourceName = (props) => {
  *   const { resource } = useResourceContext(props);
- *   const resourceName = translate(`resources.${resource}.name`, {
- *      smart_count: 1,
- *      _: inflection.humanize(inflection.singularize(resource)),
- *   });
+ *   const resourceName = useResourceLabel(resource, false);
```
fzaninotto, line 14: bad syntax: it returns a function

packages/ra-core/src/core/useGetResourceLabel.ts
```diff
@@ -0,0 +1,44 @@
+import inflection from 'inflection';
+import { useSelector } from 'react-redux';
+import { getResources } from '../reducer';
+import { useTranslate } from '../i18n';
+
+/**
+ * A hook which will return a translated resource name. It will use the label option of the `Resource` component if provided.
+ *
```
fzaninotto, line 8: could you add a usage example?

packages/ra-ui-materialui/src/list/Empty.tsx
```diff
@@ -38,15 +42,10 @@ const Empty: FC<EmptyProps> = props => {
     const classes = useStyles(props);
     const translate = useTranslate();
 
+    const getResourceLabel = useGetResourceLabel();
     const resourceName = translate(`resources.${resource}.forcedCaseName`, {
         smart_count: 0,
-        _: inflection.humanize(
-            translate(`resources.${resource}.name`, {
-                smart_count: 0,
```
fzaninotto, line 45: Was it an error? smart_count zero seems wrong to me

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/Empty.tsx
djhi: It's not. See https://airbnb.io/polyglot.js/#pluralization. Some languages have special rules when there no elements at all.

packages/ra-core/src/core/useGetResourceLabel.ts
djhi: Done

packages/ra-core/src/core/useResourceContext.ts
djhi: Fixed

fzaninotto (changes requested):
Seing the code is use, I find the API perfectible. The second parameter, `true`, isn't very readable if one doesn't know the signature of the hook. Why not pass an options object instead?  

```diff
-getResourceLabel(resource, true)
+getResourceLabel(resource, { smart_count: 1 })
```

packages/ra-core/src/core/useResourceContext.ts
```diff
@@ -11,11 +11,8 @@ import { ResourceContext, ResourceContextValue } from './ResourceContext';
  *
  * const ResourceName = (props) => {
  *   const { resource } = useResourceContext(props);
- *   const resourceName = translate(`resources.${resource}.name`, {
- *      smart_count: 1,
- *      _: inflection.humanize(inflection.singularize(resource)),
- *   });
- *   return <>{resourceName}</>;
+ *   const getResourceLabel = useGetResourceLabel();
+ *   return <>{getResourceLabel(resource, false)}</>;
```
fzaninotto, line 15: Wrong syntax, the second argument is now an integer

packages/ra-core/src/core/useGetResourceLabel.spec.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { renderWithRedux } from 'ra-test';
+import { useGetResourceLabel } from './useGetResourceLabel';
+import { TestTranslationProvider } from '../i18n';
+
+describe('useResourceLabel', () => {
+    test.each([
+        [2, 'Posts'],
+        [1, 'Post'],
```
fzaninotto, line 9: could you please add a test for the 0 case, too?

fzaninotto (changes requested):
Needs another rebase...

fzaninotto:
needs rebase

fzaninotto (approved):

Issue #6022, opened 2021-03-09
Allow BooleanField component to evaluate its values in a non strict way


WiXSL:
Currently, the `BooleanField` evaluates its value strictly as `boolean` in order to show something.
My  `BooleanField` value may be `0` or `1` instead of `true` or `false` and the component is not showing anything in that case.
I'would like to be able to tell the `BooleanField` to loosen up a little bit.

fzaninotto:
Fixed by #6027

PR #6023, opened 2021-03-09
[RFR] Allow BooleanField component to evaluate its value in a non strict way

WiXSL:
Implements #6022


WiXSL:
This is not good enough.
Any value other than a falsy one will show a check

docs/Fields.md
```diff
@@ -126,12 +126,13 @@ import { BooleanField } from 'react-admin';
...
-| `FalseIcon`       | Optional | SvgIconComponent | `@material-ui/icons/Clear` | Icon to show for the falsy value  |
+| Prop              | Required | Type             | Default                    | Description                                                      |
+| ----------------- | -------- | ---------------- | -------------------------- | ---------------------------------------------------------------- |
+| `valueLabelTrue`  | Optional | string           | 'true'                     | Aria label for the truthy value                                  |
+| `valueLabelFalse` | Optional | string           | 'false'                    | Aria label for the falsy value                                   |
+| `TrueIcon`        | Optional | SvgIconComponent | `@material-ui/icons/Done`  | Icon to show for the truthy value                                |
+| `FalseIcon`       | Optional | SvgIconComponent | `@material-ui/icons/Clear` | Icon to show for the falsy value                                 |
+| `strictTypeCheck` | Optional | boolean          | `true`                     | If `true` the field's value is evaluated strictly as a `boolean` |
```
fzaninotto, line 135: I'd prefer if the prop worked the other way around: st it to true to *enable* loose values. Besides, the second part of the prop name is wrong, it's not a check and it's not about type, but value. I'd call the prop `looseValue`.

fzaninotto (changes requested):
This is a new feature and should be PRed against `next`

PR #6027, opened 2021-03-11
[RFR] Allow BooleanField component to evaluate its value in a non strict way

WiXSL:
Implements #6022
Follow up of #6023

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -43,10 +43,11 @@ export const BooleanField: FC<BooleanFieldProps> = memo<BooleanFieldProps>(
         const translate = useTranslate();
         const classes = useStyles(props);
         const value = get(record, source);
+        const isTruthyValue = value === true || (looseValue && value);
         let ariaLabel = value ? valueLabelTrue : valueLabelFalse;
 
         if (!ariaLabel) {
-            ariaLabel = !value ? 'ra.boolean.false' : 'ra.boolean.true';
+            ariaLabel = isTruthyValue ? 'ra.boolean.true' : 'ra.boolean.false';
         }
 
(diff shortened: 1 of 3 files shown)


WiXSL:
Is this ok to merge?

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -58,7 +58,7 @@ export const BooleanField: FC<BooleanFieldProps> = memo<BooleanFieldProps>(
                     {...sanitizeFieldRestProps(rest)}
                 >
                     <Tooltip title={translate(ariaLabel, { _: ariaLabel })}>
-                        {value === true ? (
+                        {value ? (
```
fzaninotto, line 61: no, it should be `value === true || (looseValue && value)`, otherwise, when `looseValue` is `false` and `value` is truthy but not true, it will show a TrueIcon, and that's a BC break.

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -45,11 +46,10 @@ export const BooleanField: FC<BooleanFieldProps> = memo<BooleanFieldProps>(
         let ariaLabel = value ? valueLabelTrue : valueLabelFalse;
 
         if (!ariaLabel) {
-            ariaLabel =
-                value === false ? 'ra.boolean.false' : 'ra.boolean.true';
+            ariaLabel = !value ? 'ra.boolean.false' : 'ra.boolean.true';
```
fzaninotto, line 49: The test should be more precise and use `looseValue`, because otherwise I think it's a breaking change.

fzaninotto (changes requested):

fzaninotto (approved):

fzaninotto:
Thanks!

Issue #6019, opened 2021-03-09
appBar in custom Layout does not accept AppBarProps with typescript strict:true


rowr111:
I am using Typescript (strict: true)  and trying to pass a custom appbar to Layout:

`const CustomLayout = (props: LayoutProps) => <Layout {...props} appBar={CustomAppBar} />;`

- it gives me an error:

```Type '(props: AppBarProps) => JSX.Element' is not assignable to type 'ComponentType<{ title?: string | ReactElement<any, string | JSXElementConstructor<any>> | undefined; open?: boolean | undefined; logout?: ReactNode; }> | undefined'.
  Type '(props: AppBarProps) => JSX.Element' is not assignable to type 'FunctionComponent<{ title?: string | ReactElement<any, string | JSXElementConstructor<any>> | undefined; open?: boolean
```

Even passing the standard react-admin AppBar component to appBar is not accepted.  The error doesn't happen if I have typescript set with strict: false and then everything works, but that is not a desired solution.

**Related code:**
check this sandbox:
https://codesandbox.io/s/exciting-driscoll-w9fuv?file=/src/App.tsx:917-923

**Environment**

* React-admin version: 3.13.2
* Last version that did not exhibit the issue (if applicable): N/A
* React version:  17.0.1
* Browser: chrome

PR #6035, opened 2021-03-14
Change Layout interface to use AppBarProps interface

yasharzolmajdi:
### Problem
---
appBar prop on `Layout` component did not accept `AppBarProps` interface

Fixes marmelab/react-admin#6019


djhi (changes requested):
Thanks! Can you fix the linter warning?

djhi (approved):
Thanks!

PR #6039, opened 2021-03-15
Fix Go to definition goes to the compiled code in VSCode

fzaninotto:
## Problem 

Using Ctrl+click (or "Go to definition") in VSCode when hovering on a function name leads to the compiled Types file. It's really slowing us down when developing.

go_to_definition_broken

## Solution

Output declarationMaps 

go_to_definition_fixed


djhi (approved):
:heart_eyes:

PR #6040, opened 2021-03-15
[Doc] Fix missing mention of <Labeled> in Field doc

fzaninotto:


docs/Fields.md
```diff
@@ -1341,20 +1363,43 @@ PriceField.defaultProps = {
...
+
+```jsx
+import { Labeled } from 'react-admin';
+
+const MyShowLayout = ({ record }) => (
+    <div>
+        <Labeled label="Name">
+            <FullNameField record={record} />
+        </Label>
+    </div>
```
djhi, line 1399: useless `div` ?

djhi: nevermind, that's for full line display

PR #6041, opened 2021-03-15
Add ability to omit basePath in buttons

fzaninotto:
In the general case, we can infer the basePath from the resource, which is in the context. 

```diff
                        <CardActions className={classes.cardActions}>
-                           <EditButton
-                               resource="posts"
-                               basePath={basePath}
-                               record={data[id]}
-                           />
-                           <ShowButton
-                               resource="posts"
-                               basePath={basePath}
-                               record={data[id]}
-                           />
+                           <EditButton record={data[id]} />
+                           <ShowButton record={data[id]} />
                        </CardActions>
```

### DIFF
--- a/packages/ra-core/src/controller/field/getResourceLinkPath.ts
+++ b/packages/ra-core/src/controller/field/getResourceLinkPath.ts
@@ -53,7 +53,7 @@ const getResourceLinkPath = ({
     reference,
     link = 'edit',
     record = { id: '' },
-    basePath = `/${resource}`,
+    basePath = '',
     linkType,
 }: Option): string | false => {
     if (linkType !== undefined) {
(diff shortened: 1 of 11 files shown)


packages/ra-core/src/controller/field/getResourceLinkPath.ts
```diff
@@ -62,7 +62,9 @@ const getResourceLinkPath = ({
         );
     }
     const sourceId = get(record, source);
-    const rootPath = basePath.replace(resource, reference);
+    const rootPath = basePath
```
fzaninotto, line 65: this is a change I wanted to push in the previous PR, but @djhi was too fast to merge it

djhi (commented):
There is a lint error

packages/ra-core/src/controller/field/getResourceLinkPath.ts
djhi: It was RFR! :joy:

djhi (approved):

PR #6051, opened 2021-03-17
Add rowAuxClick to DatagridRow component

WiXSL:


packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -128,6 +129,43 @@ const DatagridRow: FC<DatagridRowProps> = React.forwardRef((props, ref) => {
...
+            if (!rowAuxClick) return;
+            event.preventDefault();
+
+            const effect =
+                typeof rowAuxClick === 'function'
+                    ? await rowAuxClick(id, basePath, record)
+                    : rowAuxClick;
+            switch (effect) {
+                case 'edit':
+                    history.push(linkToRecord(basePath, id));
```
fzaninotto, line 143: the middle click is supposed to open the link in a new tab. I don't see that in your PR.

fzaninotto (changes requested):

fzaninotto:
Also, this is a new feature, and it should b PRed against `next` instead of `master`

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -150,6 +188,7 @@ const DatagridRow: FC<DatagridRowProps> = React.forwardRef((props, ref) => {
                 style={style}
                 hover={hover}
                 onClick={handleClick}
+                onAuxClick={handleAuxClick}
```
djhi, line 191: I don't think this event handle the CTRL + click combo. But this combo might not be universal cross OSs. Any thoughts?

WiXSL: Doesn't work in Safari
https://caniuse.com/?search=auxclick

djhi: The new IE :joy:

WiXSL: Hehe. What should I do? Is this ok, or I should change the implementation?

fzaninotto: why don't you just modify the `DatagridRow::handleClick` method to detect if the ctrlKey is pressed, and switch from history.push() to window.open in that case?

WiXSL: Ok

WiXSL: I can't detect the button clicked using `onClick` handler, i have to use `onMosueDown`

lefuturiste:
@fzaninotto The PR was again closed automatically (I assume). Can you reopen it?

PR #6054, opened 2021-03-18
Update refresh strategy to avoid empty page while refreshing

fzaninotto:
## Problem

On pages built by hand using the `useQueryWithStore`-based hooks (e.g. `useGetList`), a refresh triggered by the Refresh button or by a side effect causes the page content to briefly disappear. This is confusing. 

refrsh_broken

The root cause is that a refresh triggers many effects:

1. it forces a rerender of the root component (because we use `key={version}`)
2. it forces a refetch of all the `useDataProvider`-based hooks used in the screen (because the `version` is in the effect signature)
3. it empties the Redux store to force a refetch even though the application cache is on (the application cache skips the call to the actual dataProvider when a valid response from a previous cal lis in the Redux store)

The 3rd effect causes the page blank: for a brief moment, the store has no data.

## Solution

The 3rd effect is too strong. We don't need to actually empty the store to empty the application cache. We just need to remove the validity dates for all recorded data.

So I replaced the refresh by a softRefresh, which does 1, 2 but not 3 - instead, it only empties the application cache.

Default refresh calls (side effects, RefreshButton) now use the softRefresh. Users can still call the ancient ('hard') refresh by dispatching the right action.

refresh_fixed

This change isn't completely backwards compatible, but I fail to see legitimate use of the old (broken) code. See my comments in the code.

### DIFF
--- a/packages/ra-core/src/reducer/admin/resource/list/cachedRequests.ts
+++ b/packages/ra-core/src/reducer/admin/resource/list/cachedRequests.ts
@@ -1,7 +1,11 @@
 import { Reducer } from 'redux';
 
 import { Identifier } from '../../../../types';
-import { FETCH_END, REFRESH_VIEW } from '../../../../actions';
+import {
+    FETCH_END,
+    REFRESH_VIEW,
+    SOFT_REFRESH_VIEW,
+} from '../../../../actions';
 import {
     GET_LIST,
     CREATE,
@@ -34,6 +38,17 @@ const cachedRequestsReducer: Reducer<State> = (
     if (action.type === REFRESH_VIEW) {
         // force refresh
         return initialState;
+    }
+    if (action.type === SOFT_REFRESH_VIEW) {
+        // remove validity only
+        const newState = {};
+        Object.keys(previousState).forEach(key => {
+            newState[key] = {
+                ...previousState[key],
+                validity: undefined,
+            };
+        });
+        return newState;
     }
     if (action.meta && action.meta.optimistic) {
         if (
--- a/packages/ra-core/src/reducer/admin/resource/list/cachedRequests.ts
+++ b/packages/ra-core/src/reducer/admin/resource/list/cachedRequests.ts
@@ -1,11 +1,7 @@
 import { Reducer } from 'redux';
 
 import { Identifier } from '../../../../types';
-import {
-    FETCH_END,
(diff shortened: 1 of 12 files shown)


packages/ra-core/src/sideEffect/useRefresh.ts
```diff
@@ -1,19 +1,31 @@
...
  * const refresh = useRefresh();
+ * // soft refresh
  * refresh();
+ * // hard refresh
+ * refresh(true)
  */
 const useRefresh = () => {
     const dispatch = useDispatch();
     return useCallback(
-        (doRefresh = true) => doRefresh && dispatch(refreshView()),
```
fzaninotto, line 16: this doRefresh parameter was undocumented and unused. Also, it's useless - one can replace it easily:

```diff
-refresh(doRefresh);
+doRefresh && refresh()
```

packages/ra-core/src/sideEffect/useRefresh.ts
```diff
@@ -1,19 +1,31 @@
...
+        (hard?: boolean) => {
+            dispatch(hard ? refreshView() : softRefreshView());
```
fzaninotto, line 27: I wondered if there should be two different actions or one action with a variable payload. I'm open to discussion on this point

packages/ra-core/src/reducer/admin/resource/list/cachedRequests.ts
```diff
@@ -35,6 +39,17 @@ const cachedRequestsReducer: Reducer<State> = (
         // force refresh
         return initialState;
     }
+    if (action.type === SOFT_REFRESH_VIEW) {
```
fzaninotto, line 42: This is the key difference: a soft_refresh doesn't empty the cached requests, only invalidates them

packages/ra-core/src/sideEffect/useRefresh.ts
djhi: I think a single action is enough, with a type option which can be either `hard` or `soft` (for now)

djhi (commented):
Do we still have use cases for hard refreshes ?

fzaninotto:
> Do we still have use cases for hard refreshes ?

I don't have any in mind, but as it was the previous default, I think we must keep that possibility.

djhi (approved):

PR #6056, opened 2021-03-18
Add ability to call useGetList without pagination, sort, or filter params

fzaninotto:
## Problem

`useGetList` requires that users pass sort, pagination, and filter params. It's cumbersome.

## Solution

Use sensible defaults

```diff
-const { ids, data, loaded } = useGetList('posts', { page: 1, perPage: 25 }, { field: 'id', order: 'DESC' }, {});
+const { ids, data, loaded } = useGetList('posts');
```


djhi (commented):
:tada:

PR #6059, opened 2021-03-18
Fix typos and doc anchors

WiXSL:


djhi (approved):
El Corrector strikes again! Thanks!

PR #6060, opened 2021-03-18
fix #6050 - type definition for Datagrid rowClick property doesn't allow for functions that return Promise<string>

### LINKED ISSUE #6050: type definition for Datagrid rowClick property doesn't match the docs.
**What you were expecting:**
The documentation for Datagrid claims you can pass a function that returns a promise as the rowClick property:

> **Tip**: If you pass a function, it can also return a promise allowing you to check an external API before returning a path. For example:
> 
> ```js
> import fetchUserRights from './fetchUserRights';
> 
> const postRowClick = (id, basePath, record) => fetchUserRights().then(({ canEdit }) => canEdit ? 'edit' : 'show');
> ```
> 

But the type definition for RowClickFunction does not allow that
```js
type RowClickFunction = (
    id: Identifier,
    basePath: string,
    record: Record
) => string;
```

**What happened instead:**
Typescript error TS2322:
```
  Type '(id: Identifier, basePath: string, record: Record) => Promise<string>' is not assignable to type 'RowClickFunction'.
    Type 'Promise<string>' is not assignable to type 'string'.  TS2322
```

**Steps to reproduce:**
From a typescript .TSX file try to pass a function that returns a `Promise<string>` to `<Datagrid rowclick={onclick}/>`

**Environment**

* React-admin version: 3.13.14

jvert:
I wasn't sure why there were three separate definitions of RowTypeFunction so I removed two of the duplicates and fixed the remaining one to allow returning Promise<string>

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
@@ -265,7 +265,7 @@ export type RowClickFunction = (
     id: Identifier,
     basePath: string,
     record: Record
-) => string;
+) => string | Promise<string>;
 
 const areEqual = (prevProps, nextProps) => {
     const { children: _1, expand: _2, ...prevPropsWithoutChildren } = prevProps;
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -265,7 +265,7 @@ export type RowClickFunction = (
     id: Identifier,
     basePath: string,
     record: Record
-) => string;
```
djhi, line 268: Can you also export it from the index file ?

djhi (changes requested):
Thanks!

djhi (approved):
Thanks!

PR #6063, opened 2021-03-19
Fix stop propagation of click event on DeleteWithConfirmButton component

### LINKED ISSUE #5815: DeleteButton's "redirect" redirects 2 times
I'm using:
* react-admin: 3.11.4
* react-admin-firebase: 3.2.17
* react: 17.0.1
* react-dom: 17.0.1

My data model:
* 2 collections: `pages` and `items`
* `items` has `parentId` field to point to parent page

When deleting an item, I want to redirect to its parent page (not item list page).

My custom toolbar code:

```typescript
function EditToolbar(props: any) {
  const redirectDelete = () => {
    const target = `/pages/${props.record.parentId}`
    return target
  }

  return (
    <Toolbar {...props}>
      <SaveButton />
      <DeleteButton redirect={redirectDelete} />
    </Toolbar>
  )
}
```

My form code:

```typescript
const redirectForm = (basePath: any, id: any, data: any) => `/pages/${data.parentId}`

<SimpleForm redirect={redirectForm} toolbar={<EditToolbar />}>
  ...
</SimpleForm>
```

Expected behavior:
* DeleteButton's "redirect" redirects only 1 time, to `/pages/:pageId`.

Current buggy behavior:
* DeleteButton's "redirect" redirects 2 times.
* 1st time: it redirects to `/pages/:pageId`.
* 2nd time: it redirects to `/items`, and displays notification error: `Element does not exist`

In browser console, there's error log like this:

```
useDataProvider.js:16 Uncaught (in promise) {status: 200, message: "Error: Error getting id: XZHlFGGyZihM55lr7SII from collection: items", json: undefined}
```

The bug is probably related to:
https://github.com/marmelab/react-admin/issues/5541 (after delete & redirect from Edit page, react-admin still does a refresh on edit page calling `getOne` on the deleted object)

Currently, I have to use this `setTimeout` workaround to redirect another time (3rd time! And the error `Element does not exist` is still displayed to end users):

```typescript
const redirectDelete = () => {
  const target = `/pages/${props.record.parentId}`
  setTimeout(() => redirect(target), 6000)
  return target
}
```

WiXSL:
Fixes #5815

Note: `useDeleteWithUndoController` hook already runs `event.stopPropagation()`

### DIFF
--- a/packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
+++ b/packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
@@ -133,6 +133,7 @@ const useDeleteWithConfirmController = (
 
     const handleDelete = useCallback(
         event => {
+            event.stopPropagation();
             deleteOne({
                 payload: { id: record.id, previousData: record },
             });


djhi (approved):
Thanks!

Issue #4644, opened 2020-04-06
Use real data provider for testing


sebashwa:
**Is your feature request related to a problem? Please describe.**
I would like to enable my data provider for testing, so I can test my custom components and their side effects better (i.e. requests that are triggered). I'm using jest and react-testing-library and by default when I wrap the components I'd like to test using the `<TestContext enableReducers>` component, all data provider requests always respond with `{ data: null }`.

**Describe the solution you'd like**
I think it would be ideal to be able to pass the real data provider to the `TestContext` component so it will be used by the test application.

**Describe alternatives you've considered**
I tried to rewrite the `TestContext` component and used my data provider. It never triggered any requests though and failed inside the react admin data reducer (for a create request inside the addOneRecord function, with newRecord set to null)

I'm not sure why my data provider wasn't triggering any requests and ended up inside the react admin data reducer before doing so. With some guidance why this happened, I would be happy to implement this feature if it is doable.

Kmaschta:
It's a good idea, feel free to open a PR, I would love to help you on this one :)

fzaninotto:
The `<TestContext>` doesn't deal with `dataProvider`. You can already write a test with a custom dataProvider by using `DataProviderContext`:

```jsx
const TextApp = () => (
<DataProviderContext.Provider value={myDataProvider}>
    <TestContext>
        {({ store, history }) => {
            ...
        }}
    </TestContext>
</DataProviderContext.Provider>
)
```

See for instance [Mutation.spec.js](https://github.com/marmelab/react-admin/blob/eb381f4f53811fef77bb29ca091ada2fd5e2f4ac/packages/ra-core/src/dataProvider/Mutation.spec.tsx#L64) for an example of this in the react-admin codebase. 

That being said, this piece of advice should be in the Unit Testing documentation. Marking this issue as Documentation.

sebashwa:
Thanks for your input @fzaninotto I managed to write my own `TestContext` component today which basically does the same thing as you suggest. I forgot to provide the `DataProviderContext` yesterday, which is why it was not working:

I would have submitted a solution along those lines, but I agree that it is easy enough to wrap the TestContext itself to achieve the same thing:

```JS
// imports omitted

const fakeDataProvider = () => Promise.resolve({ data: null });

export default class TestContext extends Component {
    storeWithDefault = null;
    history = null;
    finalDataProvider = convertLegacyDataProvider(fakeDataProvider);

    constructor(props) {
      super(props);
[...]

PR #6067, opened 2021-03-19
Add customReducers to TestContext

ValentinnDimitroff:
Added support for customReducers inside TestContext and renderWithRedux mentioned in #4644

### DIFF
--- a/packages/ra-test/README.md
+++ b/packages/ra-test/README.md
@@ -65,6 +65,20 @@ testUtils = render(
 
 This means that reducers will work as they will within the app.
 
+### Passing your custom reducers
+
+If your component relies on customReducers which are passed originally to the `<Admin/>` component, you can plug them in the TestContext using the `customReducers` props:
+
+```jsx
+testUtils = render(
+    <TestContext enableReducers customReducers={myCustomReducers}>
+        <MyCustomEditView />
+    </TestContext>
+);
+```
+
+Note you should also enable the default react-admin reducers in order to supply the custom ones. 
+
 ### Spying on the store 'dispatch'
 
 If you are using `useDispatch` within your components, it is likely you will want to test that actions have been dispatched with the correct arguments. You can return the `store` being used within the tests using a `renderProp`.
--- a/packages/ra-test/README.md
+++ b/packages/ra-test/README.md
@@ -100,6 +100,37 @@ it('should send the user to another url', () => {
 });
 ```
 
+### Using the 'renderWithRedux' wrapper function
+
+Instead of wrapping the component under test with the `TestContext` by yourself you can get use of all of the above options and test your components almost natively looking like using just `@testing-library/react` thanks to the `renderWithRedux` wrapper function.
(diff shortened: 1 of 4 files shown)


packages/ra-test/README.md
```diff
@@ -100,6 +100,37 @@ it('should send the user to another url', () => {
 });
 ```
 
+### Using the 'renderWithRedux' wrapper function
+
+Instead of wrapping the component under test with the `TestContext` by yourself you can get use of all of the above options and test your components almost natively looking like using just `@testing-library/react` thanks to the `renderWithRedux` wrapper function.
```
djhi, line 105:
```suggestion
Instead of wrapping the component under test with the `TestContext` by yourself you can use of all of the above options and test your components almost like using just `@testing-library/react` thanks to the `renderWithRedux` wrapper function.
```

packages/ra-test/README.md
```diff
@@ -100,6 +100,37 @@ it('should send the user to another url', () => {
...
+// to the way using
```
djhi, line 106: I don't get this line

packages/ra-test/README.md
```diff
@@ -100,6 +100,37 @@ it('should send the user to another url', () => {
...
+
+It will return the same output as `@testing-library/react`'s  `render` method with added `dispatch` and `reduxStore` helpers
```
djhi, line 108:
```suggestion
It will return the same output as the `render` method from `@testing-library/react` but will add the `dispatch` and `reduxStore` helpers.
```

djhi (changes requested):

packages/ra-test/README.md
ValentinnDimitroff: My point was to explain that using renderWithRedux is actually doing the same stuff as wrapping your component in TestContext but more convinient. Line 106 is pushed by mistake, it should have been deleted.

packages/ra-test/README.md
ValentinnDimitroff: I didn't find a way to edit your suggestion so I commited the changes as a separate commit. I agree with the suggested edits and believe the 'of' after 'by yourself you can use' part is not needed.

packages/ra-test/README.md
ValentinnDimitroff: I have added one more commit where it is deleted

ValentinnDimitroff:
@djhi Is there  still something left that needs to be fixed?

ValentinnDimitroff:
@djhi  Sorry if I am getting irritating but I would love it if we can complete the feature so that I can take it off my chest. If some future edits are needed - just let me know, thanks!

fzaninotto (approved):

fzaninotto:
Sorry for the delay, many thanks for your contribution!

ValentinnDimitroff:
All is good, my pleasure :) Do you consider this sufficiently resolving #4644 ? I though we might mark it as a Linked Issue

PR #6076, opened 2021-03-23
Fix Datagrid's rowClick prop not able to return a promise

### LINKED ISSUE #6050: type definition for Datagrid rowClick property doesn't match the docs.

WiXSL:
Fixes #6050

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
@@ -36,6 +36,7 @@ import DatagridBody, { PureDatagridBody } from './DatagridBody';
 import useDatagridStyles from './useDatagridStyles';
 import { ClassesOverride } from '../../types';
 import DatagridContextProvider from './DatagridContextProvider';
+import { RowClickFunction } from './DatagridRow';
 
 /**
  * The Datagrid component renders a list of records as a table.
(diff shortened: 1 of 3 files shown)


djhi (commented):
Duplicate of #6060 ?

WiXSL:
Absolutely!
I didn't see the PR

PR #6080, opened 2021-03-25
Fix ReferenceArrayField passes empty data to child when loaded

fzaninotto:
## Problem

`<ReferenceArrayField>` relies on `dataProvider.getMany()` to retrieve the references, and to pass it them to its child (e.g. a `<SingleFieldList>`. When `getMany()` returns, for a brief moment, `loaded` is `true` and `data` is empty. If the child of `<SingleFieldList>` doesn't check for the existence of its injected `record` before looking up a field, it fails. 

```jsx
const WeakField = ({ record }: any) => <div>{record.title}</div>; // fails when record is undefined
...
<ReferenceArrayField
  record={{ id: 123, barIds: [1, 2] }}
  className="myClass"
  resource="foos"
  reference="bars"
  source="barIds"
  basePath="/foos"
>
  <SingleFieldList linkType={false}>
    <WeakField />
  </SingleFieldList>
</ReferenceArrayField>
```

## Solution

The bug comes from a race condition in `useReferenceArrayFieldController`. This hook processes the data in a `useEffect` when `loaded` becomes true. As a consequence, the hook timeline is:

1. returns `{ loaded: false, data: {} }`
2. `dataProvider.getMany()` returns
3. returns `{ loaded: true, data: {} }` 
4. Processes the `useEffect`
5. returns `{ loaded: true, data: { 1: { id: 1, title: foo } }` 

At 3, the hook state is incoherent.

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
@@ -1,10 +1,13 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render } from '@testing-library/react';
+import { render, act } from '@testing-library/react';
+import { renderWithRedux } from 'ra-test';
 import { MemoryRouter } from 'react-router-dom';
-import { ListContextProvider } from 'ra-core';
+import { ListContextProvider, DataProviderContext } from 'ra-core';
 
-import { ReferenceArrayFieldView } from './ReferenceArrayField';
+import ReferenceArrayField, {
+    ReferenceArrayFieldView,
+} from './ReferenceArrayField';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
```diff
@@ -203,4 +206,42 @@ describe('<ReferenceArrayField />', () => {
         );
         expect(container.getElementsByClassName('myClass')).toHaveLength(1);
     });
+
+    it('should have defined data when loaded', async () => {
```
fzaninotto, line 210: this test fails without the fix

djhi (approved):

PR #6082, opened 2021-03-25
Fix typo in FormWithRedirectRenderProps type

WiXSL:
Fixes #6081

PR #6085, opened 2021-03-26
Fix FormWithRedirect Props

### LINKED ISSUE #6081: Typo: `chilren` instead of `children`
I suppose this is really a typo. Even if it's intended, it's really confusing.

https://github.com/marmelab/react-admin/blob/4c62d18f2cbde22349f66ef9d857c36414201d1f/packages/ra-core/src/form/FormWithRedirect.tsx#L193

djhi:
Supersedes #6082
Fixes #6081

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleFormView.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleFormView.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Children, FC, ReactElement, ReactNode } from 'react';
+import { Children, ReactElement, ReactNode } from 'react';
 import classnames from 'classnames';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/form/SimpleFormView.tsx
```diff
@@ -1,13 +1,13 @@
 import * as React from 'react';
-import { Children, FC, ReactElement } from 'react';
+import { Children, FC, ReactElement, ReactNode } from 'react';
```
fzaninotto, line 2: Linter warning

fzaninotto (changes requested):

fzaninotto (approved):

PR #6086, opened 2021-03-26
Export TS types

tdnl:
We have been trying to use some of these types but ended up copying them in our codebase because they were not exported.
This PR solves this problem and will hopefully be useful to some developers.


djhi (approved):
Awesome! Thanks!

PR #6091, opened 2021-03-29
Fixing docs theming

danangekal:
Hi All,
I fixed the props button on a document theming custom error page

Thanks.


djhi (approved):
Thanks!

PR #6112, opened 2021-04-01
Fix customRoutes aren't used when the resources are empty

### LINKED ISSUE #6079: Lazy initialization of resources stuck in loading if no resources
**What you were expecting:**
Our goal is to create a page with no authentication needed (a forgot password page).  We created a route with a noLayout props and we made sure that our `getPermissions` function in the `authProvider` did not throw an exception if the user has none.  Then we give an empty array of resources to the admin component.

**What happened instead:**
When I go on the custom route, the admin is stuck on loading.

**Steps to reproduce:**
Pass an empty array as children to the admin component.

**Related code:**
```ts
<RA.Admin
    customRoutes={[
                <Router.Route exact path="/test" component={Test} noLayout />
    ]}
    ...
>
    {(role: string) => {
        switch (role) {
            case 'admin':
                return [<RA.Resource ...];
            case 'visitor':
                // Here is the bug :
                return [];
   
```

**Environment**

* React-admin version: 3.13.1

fzaninotto:
Closes #6079

### DIFF
--- a/examples/simple/src/index.tsx
+++ b/examples/simple/src/index.tsx
@@ -1,6 +1,6 @@
 /* eslint react/jsx-key: off */
 import * as React from 'react';
-import { Admin, Resource } from 'react-admin'; // eslint-disable-line import/no-unresolved
+import { Admin, Resource, CustomRoute } from 'react-admin'; // eslint-disable-line import/no-unresolved
 import { render } from 'react-dom';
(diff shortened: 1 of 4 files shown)


djhi (approved):
:tada:

DjebbZ:
Yay!

Issue #5893, opened 2021-02-10
Dynamic window title


jtomaszewski:
Currently all our admin app browser tabs are named in the same way; and the names in the browser history aren't really helpful neither:

![image](https://user-images.githubusercontent.com/1862580/107463208-b9788980-6bb1-11eb-9e1b-d09e0ca540c6.png)

Maybe we could add some `useEffect` code to https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/Title.tsx that updates the `document.title` whenever new `title` is set? WDYT?

MicroJackson:
How I do it with my React Admin: 
```
class SettingsList extends React.Component {
    constructor(props) {
        super(props);
    }

    componentDidMount() {
        document.title = "Some new document title.";
    }

    ......
```
And it shows that in the history and as the title.
But maybe a MaterialUI update version is better.

djhi:
@jtomaszewski Thanks for the suggestion! Would you mind creating a PR for it?

andrico1234:
Is someone currently working on this? If not, I'd be happy to give it a shot

andrico1234:
@djhi  Got carried away 😅 

https://github.com/marmelab/react-admin/pull/6118

fzaninotto:
Fixed by #6119

andrico1234:
@fzaninotto can we re-open this issue as the change was reverted. I'm going to give @wmwart's suggestion,

fzaninotto:
Yep, the fix was reverted in #6357, so I'm reopening it.

panfiva:
FYI, users can customize browser page title using react-helmet. See example code that can be placed in Show, Edit, List or Create components.

```tsx
  <Helmet>
      <title>New page title</title>
  </Helmet>
```

btoro:
> FYI, users can customize browser page title using react-helmet. See example code that can be placed in Show, Edit, List or Create components.
> 
> ```tsx
>   <Helmet>
>       <title>New page title</title>
>   </Helmet>
> ```

It would be great if the <Title/> component automatically sent the title over to Helmet, rather than having users add the title to each page / view.

You can use react portals to inject the title into the <title> component of react-helmet, but unfortunately, RA adds an additional <span> around the text, which messes up the header title.

It would be great if we can override the Title component as part of the customization / layout

lefuturiste:
@fzaninotto I'm interested to work on this issue. Before opening a new PR, I will discuss the details here.
[...]

PR #6118, opened 2021-04-02
Feature: Update window title on page change

andrico1234:
Closes: https://github.com/marmelab/react-admin/issues/5893

Updates the window title to match with the title of the current page. This solves an issue with the browser history simply displaying the same title for every single page.

Note: I created a new hook called `useDocumentTitle` this is a way to set the document title. I added it to the hooks folder in `ra-core` and exported it. To me it doesn't feel like it should live there. Practically it would only be used in this Title component, so I'm not sure if we should keep the function located in `ra-core`, but there aren't any hooks exported in ra-ui-materialui`. 

Saying that, if someone chooses to pass through their own Title component, they should be able to manage their document.title however they see fit, which means they should also be able to import `useDocumentTitle`.

tl;dr, `useDocumentTitle` lives in `ra-core` but i'll be happy to place it in `ra-ui-materialui` if you see it fit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,53 @@
 # Changelog
 
+## 3.14.1
+
+* Fix performance regression ([6096](https://github.com/marmelab/react-admin/pull/6096)) ([fzaninotto](https://github.com/fzaninotto))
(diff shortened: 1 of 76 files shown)


djhi (changes requested):
Nice! As this is a new feature, can you please target the `next` branch?

PR #6119, opened 2021-04-02
Feature: Update window title on page change

andrico1234:
Closes: #5893

Updates the window title to match with the title of the current page. This solves an issue with the browser history simply displaying the same title for every single page.

Note: I created a new hook called useDocumentTitle this is a way to set the document title. I added it to the hooks folder in ra-core and exported it. To me it doesn't feel like it should live there. Practically it would only be used in this Title component, so I'm not sure if we should keep the function located in ra-core, but there aren't any hooks exported in ra-ui-materialui`.

Saying that, if someone chooses to pass through their own Title component, they should be able to manage their document.title however they see fit, which means they should also be able to import useDocumentTitle.

tl;dr, useDocumentTitle lives in ra-core but i'll be happy to place it in ra-ui-materialui if you see it fit


andrico1234:
@djhi I can't get my head around moving the tip of a branch (e.g. moving the last commit from the tip of master over to next), so I just made a new pull request

If you have a git helper doc to hand for the above issue, i'd be eternally grateful!

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -23,6 +23,10 @@ const Title: FC<TitleProps> = ({
         typeof document !== 'undefined'
             ? document.getElementById('react-admin-title')
             : null;
+
+    const documentTitle = typeof title === 'string' ? title : defaultTitle;
+    useDocumentTitle(translate(documentTitle, { _: documentTitle }));
```
fzaninotto, line 28: I think you shouldn't set the document title when the titleProp is an element. In that case, it's the responsibility of the element to define the document title. 

Also, the defaultTitle isn't translated on line 35, but in your case it is. This will lead to different document and page title.

fzaninotto (changes requested):

examples/simple/src/posts/PostTitle.tsx
```diff
@@ -1,13 +1,11 @@
...
-    return (
-        <span>
-            {record
-                ? translate('post.edit.title', { title: record.title })
-                : ''}
-        </span>
-    );
+    const title = translate('post.edit.title', { title: record.title });
+
+    useDocumentTitle(title);
```
andrico1234, line 8: Updated the example to demonstrate handling a title update manually

examples/simple/src/tags/TagEdit.tsx
```diff
@@ -32,7 +32,7 @@ const TagEdit = props => (
                 basePath="/posts"
                 resource="posts"
                 filter={{ tags: [props.id] }}
-                title=" "
+                title={<span />}
```
andrico1234, line 35: Worth noting @fzaninotto  : The " " caused the document title to get messed up due to the page having two separate titles. Passing through an empty span means the page title and the UI title only shows the `Tags #` title

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -11,6 +11,17 @@ export interface TitleProps {
     title?: string | ReactElement;
 }
 
+interface TitleTextProps {
+    text: string;
+    className: string;
+}
+
+function TitleText({ text, ...rest }: TitleTextProps) {
```
andrico1234, line 19: This ensures both the UI title and document title are identical, and we only run the hook if title is a string.

andrico1234:
Morning folks! I'm wondering if there are any other changes you'd like me to make for this pull request. @fzaninotto @djhi

packages/ra-core/src/util/hooks.ts
```diff
@@ -62,3 +62,11 @@ export function useTimeout(ms = 0) {
 
     return ready;
 }
+
+export function useDocumentTitle(title: string) {
```
fzaninotto, line 66: I'm fine with having this hook in ra-core, but I think it deserves its own file and a jsDoc block with example usage. You can keep it in util/.

packages/ra-ui-materialui/src/layout/Title.tsx
fzaninotto: Nice. I think this component (and its type) should be exported, so it's better to put it in its own file.

fzaninotto (changes requested):
Awesome!

This feature needs basic documentatoin, e.g. in https://marmelab.com/react-admin/CreateEdit.html#page-title, where you should explain that, when using an element as title, you have to set the document title by hand - and show an example usage of the hook.

andrico1234:
Thanks for all the feedback and suggestions, @fzaninotto , I've pushed the fixes and resolved them all. I'll behappy to make any other changes as needed

packages/ra-core/src/util/index.ts
```diff
@@ -10,7 +10,8 @@ import Ready from './Ready';
 import resolveRedirectTo from './resolveRedirectTo';
 import warning from './warning';
 import useWhyDidYouUpdate from './useWhyDidYouUpdate';
-import { useSafeSetState, useTimeout, useDocumentTitle } from './hooks';
+import useDocumentTitle from './useDocumentTitle';
```
fzaninotto, line 13: We're deprecating `export default` in source files little by little. For new files, you should just export the hooks, components and types, and reexport them as is with:

```tsx
export * from './useDocumentTitle';
```

packages/ra-core/src/util/useDocumentTitle.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+
+/**
+ * A hook that updates the document's title whenever the title argument changes
```
fzaninotto, line 4:
```suggestion
 * A hook that updates the document title whenever the argument changes
```

packages/ra-ui-materialui/src/layout/index.ts
```diff
@@ -18,6 +18,7 @@ import Notification from './Notification';
 import Responsive, { ResponsiveProps } from './Responsive';
 import Sidebar, { SidebarProps } from './Sidebar';
 import Title, { TitleProps, TitlePropType } from './Title';
+import TitleText, { TitleTextProps } from './TitleText';
```
fzaninotto, line 21: same, please use `export * from`

fzaninotto (changes requested):

andrico1234:
Thanks for the re-review @fzaninotto , I've implemented the requested changes!

andrico1234:
Hi @fzaninotto , are there any other changes you'd like?

examples/simple/src/posts/PostTitle.tsx
```diff
@@ -1,13 +1,11 @@
...
 export default ({ record }: TitleProps) => {
     const translate = useTranslate();
-    return (
-        <span>
-            {record
-                ? translate('post.edit.title', { title: record.title })
-                : ''}
-        </span>
-    );
+    const title = translate('post.edit.title', { title: record.title });
```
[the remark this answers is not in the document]
andrico1234: I can add a check just in case, but based on the `TitleForRecord` component, it won't render if `record` doesn't exist.

andrico1234:
Hi @fzaninotto , I've made the required changes. Is there anything else I should change before hand?

fzaninotto (approved):

fzaninotto:
Thanks!

fzaninotto:
We haven't noticed, but this is a breaking change. If a developer has set the document title to "Acme" globally, it's replaced by the page title (e.g. "Posts"), without any simple way to insert the main app name globally. 

So I'm going to revert this PR as is must provide an additional feature without removing the current abilities. 

Feel free to open another PR to discuss possible implementations.

andrico1234:
Thanks for the headsup, i'll see if there's a way in which we can implement this without a breaking change

wmwart:
```
import { useEffect, useRef } from "react";


export const tryToSetDocumentTitle = (title) => {
    try {
        document.title = title;
    } catch (e) {
        console.warn("Document is undefined");
    }
};


const useDocumentTitle = ({
    title,
    revertOnUnmount = true,
}) => {

    const isBrowser = typeof document !== "undefined";
    const prevPageTitleRef = useRef(isBrowser ? document.title : null);

    useEffect(() => {
        if (isBrowser) {
            prevPageTitleRef.current = document.title;
        }
    }, [isBrowser]);

    useEffect(() => {
        if (isBrowser) {
            tryToSetDocumentTitle(title);
        }

        return () => {
            if (isBrowser && revertOnUnmount && prevPageTitleRef.current !== null) {
                tryToSetDocumentTitle(prevPageTitleRef.current);
            }
        };
    }, [isBrowser, title, revertOnUnmount]);
};

export default useDocumentTitle
```

andrico1234:
thanks @wmwart for the excellent example , i'll give this a shot and see how it goes!

yelworc:
@andrico1234 thanks for your work on this – I'd really appreciate this feature!

yelworc:
Just for reference, for anybody passing by – this is my stripped down version of @wmwart's hook above, adjusted for react-admin v5:

```js
export const useDocTitle = (title, { revertOnUnmount = true } = {}) => {
  const baseTitle = useDefaultTitle();
  useEffect(() => {
    document.title = `${baseTitle} › ${title}`;
    return () => {
      if (revertOnUnmount) document.title = baseTitle;
    };
  }, [baseTitle, title, revertOnUnmount]);
};
```

fzaninotto:
For anyone passing by, React 19 supports the `<title>` tag now, so you can change the title of each of your pages with a single tag.

PR #6122, opened 2021-04-03
Fix hasBulkActions prop type passed to DatagridRow

RoBYCoNTe:
Prop hasBulkActions passed to DatagridBody was computed as "array" instead of boolean generating this console error:

`Warning: Failed prop type: Invalid prop hasBulkActions of type array supplied to ForwardRef, expected boolean.`

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
@@ -49,7 +49,7 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                             [classes.clickableRow]: rowClick,
                         }),
                         expand,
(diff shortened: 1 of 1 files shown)


OoDeLally:
Same here.

It happens because `hasBulkActions` is guarded by the propType:
```
hasBulkActions: PropTypes.bool.isRequired
```

but `DatagridBody` is carelessly using JS's cast-to-boolean:
```
hasBulkActions: hasBulkActions && selectedIds,
```

so `hasBulkActions` ends up being an array.

Solution 1 (easy):

```
hasBulkActions: Boolean(hasBulkActions && selectedIds),
```


Solution 2 (harder but better): drop propTypes altogether and rely on Typescript, which already seems here anyway.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6129, opened 2021-04-06
[RFR] Fix missing query string after successful login

### LINKED ISSUE #6117: Filter values, sort params etc are lost after successful logging-in in custom login page
**What you were expecting:**
After successful login, the filter values are kept

**What happened instead:**
User is redirected to resource but without query string

**Steps to reproduce:**
1. Create custom login page
2. Use `useLogin()` hook
3. Add resource with filters
4. Send to user a link to resource with filters set e.g `domain.com/#/my-component?filter=%7B%22filterA%22%3A%22foo%22%2C%22filterB%22%3A%22bar%22%7D&order=ASC&page=1&perPage=10&sort=foo`
5. Unauthenticated user is redirected to login page
6. User successfully logged-in
7. User is redirected to `domain.com/#/my-component`

Expected:
User is redirected to `domain.com/#/my-component?filter=%7B%22filterA%22%3A%22foo%22%2C%22filterB%22%3A%22bar%22%7D&order=ASC&page=1&perPage=10&sort=foo`

Actual:
User was redirected to `domain.com/#/my-component`

**Environment**
* React-admin version: 3.14.1
* React version: 16.14.0
* Browser: Chrome Canary 91.0.4443.0

makbol:
I tested this solution with my existing project and it worked fine with custom login page and external identity provider. I hope this quick fix will be sufficient. 

Fixes #6117

### DIFF
--- a/cypress/integration/auth.js
+++ b/cypress/integration/auth.js
@@ -30,4 +30,21 @@ describe('Authentication', () => {
         ListPage.navigate();
         cy.url().then(url => expect(url).to.contain('/#/posts'));
     });
+
+    it('should redirect to initial url keeping query string', () => {
+        let urlBeforeLogout;
+
+        ListPage.navigate();
+        ListPage.addCommentableFilter();
+        cy.url().then(url => {
+            urlBeforeLogout = url;
+        });
+        ListPage.setAsNonLogged();
+        cy.reload();
+        LoginPage.login('login', 'password');
+        cy.url().then(urlAfterLogin => {
+            expect(urlAfterLogin).to.contain(urlBeforeLogout);
+        });
+        ListPage.commentableFilter().should('exist');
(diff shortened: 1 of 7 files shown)


djhi (changes requested):
Thanks! Some additional tests validating the search value would be nice.

makbol:
@djhi Alright, I'll try. Can you give me some tips how should I approach testing that? From what I see you don't have functional tests set. My perspective is testing that would require passing from unlogged state through login process to redirect to app and check if query string stays. Maybe there's a simpler way of testing that?

makbol:
Okay, functional test added. @djhi let me know if this is sufficient.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6130, opened 2021-04-06
Add ability to refetch a query in useQuery and useQueryWithStore

djhi:


packages/ra-core/src/dataProvider/useQuery.ts
```diff
@@ -113,19 +123,21 @@ const useQuery = (
                     : [payload, otherOptions]
             )
             .then(({ data, total }) => {
-                setState({
+                setState(prev => ({
+                    ...prev,
```
fzaninotto, line 127: I remember that we used to do that but we had problems in corner cases. I suggest you pass `refetch` explicitly  instead.

fzaninotto (changes requested):
Looking forward to reading the documention!

packages/ra-core/src/dataProvider/useQueryWithStore.ts
```diff
@@ -114,19 +115,26 @@ const useQueryWithStore = <State extends ReduxState = ReduxState>(
...
     totalSelector: (state: State) => number = defaultTotalSelector(query),
     isDataLoaded: (data: any) => boolean = defaultIsDataLoaded
-): {
-    data?: any;
-    total?: number;
-    error?: any;
-    loading: boolean;
-    loaded: boolean;
-} => {
+): StateResult => {
```
fzaninotto, line 118: rename to UseQueryWithStoreValue

fzaninotto (changes requested):

packages/ra-core/src/dataProvider/useQueryWithStore.ts
djhi: It would be a breaking change, right?

fzaninotto: Why? The type wasn't named or exported before this PR.

packages/ra-core/src/dataProvider/index.ts
```diff
@@ -25,12 +24,9 @@ import useDeleteMany from './useDeleteMany';
 import useRefreshWhenVisible from './useRefreshWhenVisible';
 import useIsAutomaticRefreshEnabled from './useIsAutomaticRefreshEnabled';
 
-export type {
-    QueryOptions,
-    UseMutationValue,
-    UseQueryValue,
-    UseGetOneHookValue,
-};
+export * from './useQueryWithStore';
```
fzaninotto, line 27: This won't work: the `Query` type will conflict with the `Query` component.

The `Query` type name is too generic; Something like `DataProviderQuery` would be better, WDYT?

fzaninotto (changes requested):

packages/ra-core/src/dataProvider/useQuery.ts
```diff
@@ -138,12 +151,6 @@ const useQuery = (
     return state;
 };
 
-export interface Query {
-    type: string;
-    resource?: string;
-    payload: object;
-}
-
 export interface QueryOptions {
```
fzaninotto, line 154: This one should be named `UseQueryOptions`. Sorry to find the problems one by one, but the fact that you decided to expose all exports in this file created new problems

fzaninotto (changes requested):

fzaninotto (approved):

Issue #6128, opened 2021-04-06
Crash when adding validation messages to inputs within ArrayInputs


andrico1234:
**What you were expecting:**
Since React 3.14.x, if an Inputs that lives inside ArrayInputs has an error message assigned it it, the app crashes on Save.

here's a video of this happening

https://user-images.githubusercontent.com/17087167/113691881-d5397100-96c4-11eb-9e7c-55eaf6b2102b.mov

I've tested this with both TextInputs and SelectInputs and they break, leading me to believe it's an issue with the `ArrayInput`. 

I've created a repro in [here](https://github.com/andrico1234/react-admin/tree/repro/nested-errors-in-arrays), you can run it and test the `PostEdit` page. 

This is the code causes the break

```tsx
import * as React from 'react';
import {
    ArrayInput,
    Edit,
    FormTab,
    SimpleFormIterator,
    TextInput,
    TabbedForm,
} from 'react-admin'; // eslint-disable-line import/no-unresolved

function validate(values: any) {
    const { authors } = values;
    const errors: any = {
        authors: [],
    };

    // works fine
    errors.singleRole = ['A role already exists with this author.'];

    // this doesn't works
    if (authors) {
        authors.forEach((author: any, i: number) => {
            if (!author) return;

            const error: any = {};

            if (author.role) {
                error.role = ['A role already exists with this author.'];
                errors.authors = [];
                errors.authors[i] = error;
            }
        });
    }

    return errors;
}

const PostEdit = ({ permissions, ...props }) => (
    <Edit {...props}>
        <TabbedForm warnWhenUnsavedChanges validate={validate}>
            <FormTab label="post.form.summary">
                {/* works fine */}
                <TextInput source="singleRole" />

                {/* doesn't work */}
                <ArrayInput source="authors">
                    <SimpleFormIterator>
                        <TextInput source="role" />
                    </SimpleFormIterator>
                </ArrayInput>
            </FormTab>
        </TabbedForm>
    </Edit>
);

export default PostEdit;

```

**Environment**

* React-admin version: 3.14.1+
* Last version that did not exhibit the issue (if applicable): 3.14.0
* React version: 17
* Browser: Chrome
* Stack trace (in case of a JS error):

```
Error: Objects are not valid as a React child (found: object with keys {role}). If you meant to render a collection of children, use an array instead.

[...]

PR #6133, opened 2021-04-07
Bugfix: Don't crash when validating inputs in Array 

andrico1234:
Fixes: https://github.com/marmelab/react-admin/issues/6128

Looks like the problem was brought on with this commit: https://github.com/marmelab/react-admin/commit/9e547eba388e3a041298e42126763943e3d4541e

The helperText treated the child input's error as a top level error. I think this threw an error because errors inside of Array inputs take the shape of their children, wheras `InputHelperText` expected anerror message as a string

I've kept the helperText logic in, but removed any of the logic involving the errors. I don't think it's necessary considering that the children display their respective error messages


djhi:
Unfortunately, we might have validators on the ArrayInput itself as well, such as requiring at least X elements. We must find a way to include the errors

PR #6140, opened 2021-04-09
Fix validation documentation & Submission Errors Cannot Have Translatable Error Objects

djhi:
In `react-admin`, errors might be objects with `message` and `args` properties which are used for translation.

The original `final-form-submit-errors` mutator was considering those errors as form nested fields such as `name.message` and `name.args`.

I introduced our own version of this mutator which detects those objects.

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -1215,12 +1215,17 @@ The value of the form `validate` prop must be a function taking the record as in
 const validateUserCreation = (values) => {
     const errors = {};
     if (!values.firstName) {
-        errors.firstName = ['The firstName is required'];
+        errors.firstName = 'The firstName is required';
     }
     if (!values.age) {
-        errors.age = ['The age is required'];
+        // You can return translation keys
+        errors.age = 'ra.validation.required';
     } else if (values.age < 18) {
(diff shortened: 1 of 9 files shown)


djhi:
@fzaninotto Should I target the `next` branch ?

docs/CreateEdit.md
```diff
@@ -1281,10 +1286,11 @@ export const UserCreate = (props) => (
             <TextInput label="Email" source="email" validate={validateEmail} />
             <TextInput label="Age" source="age" validate={validateAge}/>
             <TextInput label="Zip Code" source="zip" validate={validateZipCode}/>
-            <SelectInput label="Sex" source="sex" choices={[
+            <SelectInput label="Genre" source="genre" choices={[
```
WiXSL, line 1289: Typo `Genre` => `Gender`

djhi: You don't even wait for merged docs anymore :joy: Thanks!

djhi: Tried to be more inclusive and ended up being even less than before :scream:

fzaninotto (approved):

PR #6142, opened 2021-04-09
Add translateOptions as props in DeleteWthConfirmButton.tsx

vllorca:
Allow the creation of a custom message during deletion


fzaninotto:
Can you please explain your use case?

Also, this is a new feature, and as such it requires tests, documentation (at least jsDoc), and should be PRed against `next`.

PR #6144, opened 2021-04-09
Fix useLogoutIfAccessDenied may notify more than once

fzaninotto:
## Problem

When the dataProvider returns with an auth error (e.g. a 403 response), react-admin triggers a redirection to the login page and a notification message that the auth params are no longer valid.

> Your session has ended, please reconnect

image

This happens when the API invalidates the user credentials.

Now, if a page needs more than one dataProvider Request, this will cause several logouts (not a problem) and several notifications about auth params (that's a problem).

## Solution

Check that the user is authenticated before notifying that their auth params are no longer valid

### DIFF
--- a/examples/simple/src/dataProvider.tsx
+++ b/examples/simple/src/dataProvider.tsx
@@ -8,6 +8,12 @@ const dataProvider = fakeRestProvider(data, true);
 const uploadCapableDataProvider = addUploadFeature(dataProvider);
 const sometimesFailsDataProvider = new Proxy(uploadCapableDataProvider, {
     get: (target, name) => (resource, params) => {
+        // set session_ended=true in localStorage to trigger an API auth error
+        if (localStorage.getItem('session_ended')) {
+            const error = new Error('Session ended') as ResponseError;
+            error.status = 403;
+            return Promise.reject(error);
+        }
         // add rejection by type or resource here for tests, e.g.
         // if (name === 'delete' && resource === 'posts') {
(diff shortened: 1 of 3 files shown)


examples/simple/src/dataProvider.tsx
```diff
@@ -8,6 +8,12 @@ const dataProvider = fakeRestProvider(data, true);
 const uploadCapableDataProvider = addUploadFeature(dataProvider);
 const sometimesFailsDataProvider = new Proxy(uploadCapableDataProvider, {
     get: (target, name) => (resource, params) => {
+        // set session_ended=true in localStorage to trigger an API auth error
+        if (localStorage.getItem('session_ended')) {
```
fzaninotto, line 12: this gives a way to simulate an API which invalidates credentials

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -9,13 +9,27 @@ import useLogout from './useLogout';
 import useNotify from '../sideEffect/useNotify';
 import { AuthProvider } from '../types';
 
-jest.mock('./useLogout');
-jest.mock('../sideEffect/useNotify');
+let loggedIn = true;
 
-const logout = jest.fn();
```
fzaninotto, line 15: this was a bad idea and made me loose 1 hour scratching my head. Using authProvider.logout() in useLogout is way more natural.

djhi (approved):

PR #5905, opened 2021-02-12
Add scrollToTop on key navigation links

fzaninotto:
## Problem

React-router [doesn't manage scroll changes on navigation](https://reactrouter.com/web/guides/scroll-restoration). As a consequence, when users click on an item low in the menu, or on the pagination controls, the new content may reveal up hidden in the non-visible area of the page. This is disturbing for end users.

scrollToTop

On the other hand, the page should not scroll to top everywhere, and especially not when using the back button, applying filters, or switching tabs.

## Solution

- [x] Add a global scrollToTop listener in CoreAdminRouter (resists to layout change)
- [x] Update Edit, Show, Pagination and Menu buttons to enable scrollToTop
- [x] Add scrollToTop to useRedirect
- [x] Add documentation
- [x] Document other buttons, and start fixing stuff down the rabbit hole
- [x] ?!!
- [x] Profit 

scrollToTop2

Supersedes #5891

PR #6145, opened 2021-04-09
Fix useRedirect does not handle query strings

### LINKED ISSUE #6141: Redirecting to create page with useRedirect try to load the edit page
**What you were expecting:**
I needed to pass some custom query parameter to the create page. 
For that I used useRedirect with a simple Button, and so added the needed params.
It worked until I changed version from 3.12.0 to 3.14.2.

**What happened instead:**
With the new version clicking on the button the page attempt to load and it display a `Element does not exists` notification and return on the list page.
In the network tab, I see that react-admin attempt to fetch a resource with the id create.
Seems like react-admin believe I want to edit.

**Steps to reproduce:**
Simply create a button to redirect to a creation page using useRedirect.

```js
const CustomCreateButton = () => {
  const onClick = () => redirect("posts/create?foo=bar");
  return (
    <Button color="primary" variant="outlined" onClick={onClick}>
      redirect to create
    </Button>
  );
};
```

**Related code:**

Here is a code sand box
https://codesandbox.io/s/jovial-glitter-of0yi?file=/src/posts/PostList.tsx

Simply go to the list of post and click on the `redirect to create` at the top.

**Other information:**
It is possible to avoid the issue by using a Link instead of useRedirect.

**Environment**

* React-admin version: 3.14.2
* Last version that did not exhibit the issue (if applicable): 3.12.0
* React version: 17.0.1
* Browser: chrome

fzaninotto:
## Problem

PR #5905 broke `useRedirect` when using a query string:

```jsx
// ...
redirect('/foo?bar=baz');
```

Because of the following change:

```diff
-           history.push(resolveRedirectTo(redirectTo, basePath, id, data));
+           history.push({
+               pathname: resolveRedirectTo(redirectTo, basePath, id, data),
+               state: { _scrollToTop: true },
+           });
```

## Solution

Transform the redirection target to a react-router `location` object.

Closes #6141
Supersedes #6009

### DIFF
--- a/packages/ra-core/src/sideEffect/useRedirect.spec.tsx
+++ b/packages/ra-core/src/sideEffect/useRedirect.spec.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react';
+import { useEffect } from 'react';
+import expect from 'expect';
+import { renderWithRedux } from 'ra-test';
+import { createMemoryHistory } from 'history';
+import { Router } from 'react-router-dom';
+
+import useRedirect from './useRedirect';
+
+const Redirect = ({ redirectTo, basePath = '', id = null, data = null }) => {
+    const redirect = useRedirect();
+    useEffect(() => {
+        redirect(redirectTo, basePath, id, data);
+    }, []);
+    return null;
+};
+
+describe('useRedirect', () => {
+    it('should redirect to the path with query string', () => {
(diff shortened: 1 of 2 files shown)


djhi (approved):
There are linter warnings though

djhi (approved):

PR #6153, opened 2021-04-12
restore deprecated undoable prop original behavior

ThieryMichel:
For retrocompatibility undoable=false should be on pessimistic mode, currently it is on undoable mode.

### DIFF
--- a/packages/ra-ui-materialui/src/button/DeleteButton.tsx
+++ b/packages/ra-ui-materialui/src/button/DeleteButton.tsx
@@ -1,39 +1,16 @@
 import * as React from 'react';
-import { FC, ReactElement, SyntheticEvent, useEffect } from 'react';
+import { FC, ReactElement, SyntheticEvent } from 'react';
 import PropTypes from 'prop-types';
-import { Record, RedirectionSideEffect, MutationMode } from 'ra-core';
+import {
+    Record,
+    RedirectionSideEffect,
+    MutationMode,
+    getMutationMode,
+} from 'ra-core';
 
 import { ButtonProps } from './Button';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/button/DeleteButton.tsx
```diff
@@ -1,12 +1,40 @@
...
-import { FC, ReactElement, SyntheticEvent } from 'react';
+import { FC, ReactElement, SyntheticEvent, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Record, RedirectionSideEffect, MutationMode } from 'ra-core';
 
 import { ButtonProps } from './Button';
 import DeleteWithUndoButton from './DeleteWithUndoButton';
 import DeleteWithConfirmButton from './DeleteWithConfirmButton';
 
+const useMutationMode = (mutationMode, undoable) => {
```
djhi, line 10: Can you extract this in a dedicated file ?

djhi (changes requested):

packages/ra-ui-materialui/src/button/DeleteButton.tsx
```diff
@@ -1,12 +1,40 @@
...
+    useEffect(() => {
```
fzaninotto, line 11: I'm -1 with this useEffect and the related warnings. A simple function, rather than a hook, will be enough.

fzaninotto (changes requested):

fzaninotto:
bump, could you please apply the review?

djhi (commented):
There is a linter warning about performances

fzaninotto (approved):

PR #6160, opened 2021-04-13
docs (Actions): fix `useDataProvider` instead of `useProvider`

f-jost:
fix typo in Legacy Components section l748


djhi (approved):
Thanks!

Issue #6087, opened 2021-03-26
Show header row for empty Datagrids


JoeJErnst:
**Is your feature request related to a problem? Please describe.**
I don't understand why the decision was made to return null for the whole Datagrid if there are no rows to return.  I would expect to still see the header row.

**Describe the solution you'd like**
Offer a property that allows me to specify whether or not to hide empty Datagrids.

**Describe alternatives you've considered**
It could be a boolean `hideWhenEmpty` (default to true for backwards compatibility).  

**Additional context**
https://github.com/marmelab/react-admin/blob/5b44d973e778a93efbfaf2a94f6cd25c6dc7e60f/packages/ra-ui-materialui/src/list/Datagrid.js#L176

fzaninotto:
The `<List>` component exposes an `empty` prop exactly for this use case. 

https://marmelab.com/react-admin/List.html#empty-empty-page-component

JoeJErnst:
The `empty` prop on the `<List>` component is far too restrictive for this use case, for three separate reasons:

1. It only works if there's NOT an active filter
2. It only works if there's a `create`  page for the resource
3. It still won't show the column headers

I have several Datagrids in my app that will have default filters enabled, and no `create` page.  I still need my users to see that there's an empty grid, not no grid at all.  There is value in seeing what the columns are, even if no data exists.

I feel this should be true in general, not just because my app has special requirements.  I stand by my enhancement request for the `Datagrid` component.  I'll see if I can get a pull request put together with my suggested implementation, in case you want to reconsider.

fzaninotto:
I disagree that showing a header line with no body is better in general, but I understand that it may be better in some cases.

About your points:

1. Right, the empty page is designed as the initial state, not for displaying an empty filtered list
2. no, the `<List empty>` component displays even though there is no create page  
3. it's up to you to pass the component of your choice to display just a list of headers in this case

Also, the `<Datagrid>` could accept an `empty` prop to display a custom component instead of `null` when the list is empty because the filters are too narrow. We can accept a PR for that (and I think it's trivial).

PR #6164, opened 2021-04-13
Feature: Add empty Datagrid component

andrico1234:
Implements @fzaninotto suggestion from [here](https://github.com/marmelab/react-admin/issues/6087)

Adds `empty` component to `Datagrid` so we can display a custom component if there are no records to render.

Example
```tsx
<Datagrid empty={<div>No records to show</div>} hasBulkActions>
  <TitleField />
</Datagrid>
```


fzaninotto (approved):

fzaninotto:
Thanks!

Issue #6143, opened 2021-04-09
Specify data provider methods as generic (not just their arguments) for type safety


bayareacoder:
**Is your feature request related to a problem? Please describe.**
When manually calling data provider methods, or when they are called by react-admin behind the scenes, there is no type safety for input data, or returned data.

**Describe the solution you'd like**
1/ Provide the option to make all data provider methods type-safe by making them generic, so that at least when they are called manually it can be type-safe.
2/ Provide type-safety also when react-admin internally calls these methods. This would lead to a great DX where you could create a form for object create and it would err when a required property is forgotten, or a form for object update that would err if a property that cannot be updated is included.

**Describe alternatives you've considered**
We have coded 1/ ourselves in our data providers and how we call them manually, but it would probably be useful to enhance the docs with this (now that you have #5934)

eg. for `create` with `ResourceName` as a union of literal types which also ensures no non-existing resource names can be passed:
```
async create<T extends any = any, R extends RaRecord = RaRecord>(
    resource: ResourceName,
    params: CreateParams<T>,
  ): Promise<CreateResult<R>> {
// ...
}
```
and then call, for resource `Shop` as:
```
create<CreateShopData, Shop>('Shop', { data });
```

Then if you add overloaded declarations in your data provider for each of the narrowed resource types:
```
create(resource: 'Shop', params: CreateParams<CreateShopData>): Promise<CreateResult<Shop>>;
create(resource: 'Account', params: CreateParams<CreateAccountData>): Promise<CreateResult<Account>>;
// ...
async create<T extends any = any, R extends RaRecord = RaRecord>(
    resource: ResourceName,
    params: CreateParams<T>,
  ): Promise<CreateResult<R>> {
// ...
}
```
you can call without generics on the call site and get full type safety:
``` 
create('Shop', { data });
```

To avoid all the overloaded definitions, I tried first to do this with some kind of mapped/conditional type but that that wasn't successful so far.

**Additional context**
This builds further upon #5934 which introduced generics for the arguments to the data provider methods but not the methods themselves.

fzaninotto:
I'm not sure to understand where you want to add these generics.

We already have them in `useDataProvider`:

```jsx
interface Product extends Record {
    name: string;
    sku: string;
}

const dataProvider = useDataProvider();
dataProvider.getOne<Product>('resources', 123)
   .then(({ data }) => {
        // TS sees data as Product
    })
```

[...]

PR #6168, opened 2021-04-15
Add ability to call specialized dataProvider hooks with specialized parameters

### LINKED ISSUE #6047: Add support for same call time signature in custom useMutation queries
## Problem

Specialized `useMutation` queries, like `useCreate` and `useUpdate`, return a callback. The signature of the callback and the one of the initial hook are different. E.g:

```js
// build-time query
const [update] = useUpdate('likes', record.id, diff, record)
// call-time query
update({
  resource: 'likes',
  payload: { id: record.id, data: diff, previousData: record }
})
```

But the callbacks are more often used with call-time params, and this is cumbersome.

## Solution

In `useMutation::mutate`, if the first parameter is a string, then build a custom runtime payload based on the params. The query shape can be defined in the options. Something like:

```js
const useUpdate = (
    resource: string,
    id: Identifier,
    data?: any,
    previousData?: any,
    options?: MutationOptions
) =>
    useMutation(
        { type: 'update', resource, payload: { id, data, previousData } },
        {
           payloadFromParams: params => ({ id: params[0], data: params[1], previousData: params[2] }),
           ...options
        }
    );
```

that way users can do:

```js
// build-time query
const [update] = useUpdate()
// call-time query
update('likes', record.id, diff, record)
```

without breaking backwards compatibility

Caution: TypeScript might get in the way

fzaninotto:
```jsx
// build-time query
const [update] = useUpdate()
// call-time query
update('likes', record.id, diff, record)
```

Also, made all specialized mutation hooks generic:

```jsx
const [update, { data }] = useUpdate<Product>('products', record.id, diff, record)
                   \- data is Product
```

Closes #6047 
Refs #6143

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -442,13 +442,33 @@ const BulkResetViewsButton = ({ selectedIds }) => {
 ```jsx
 // syntax
 const [deleteOne, { data, loading, loaded, error }] = useDelete(resource, id, previousData, options);
+```
 
-// example
+The `deleteOne()` function can be called in 3 different ways:
+ - with the same parameters as the `useDelete()` hook: `deleteOne(resource, id, previousData, options)`
+ - with the same syntax as `useMutation`: `deleteOne({ resource, payload: { id, previousData } }, options)`
+ - with no parameter (if they were already passed to `useDelete()`): `deleteOne()`
+
+```jsx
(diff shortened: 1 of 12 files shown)


djhi (approved):
:fire:

PR #6176, opened 2021-04-16
Fix warning in UrlField when value is nullish and emptyText is empty

OoDeLally:
Choosing between 
- displaying a link or a static text, 
- showing the value or the default,
are two distinct orthogonal decisions.

In my base my `value` was `undefined` and I didnt provide an `emptyText`.
Because of that `UrlField` tried to create a `Link` leading to an annoying console warning, because it doesnt like that `children` is undefined.

In the case the value is null, it is better to display an empty static text than an empty link.


fzaninotto (approved):

fzaninotto:
Thanks!

Issue #6174, opened 2021-04-16
Props to control HideOnScroll  behavior of AppBar


pkspyder007:
**Is your feature request related to a problem? Please describe.**
I have a fixed sidebar that is toggled by the App bar for expanding and collapsing. But after scrolling down there is no button to toggle the side menu until the user scrolls back up.

https://user-images.githubusercontent.com/331353/114960721-a2664a00-9e35-11eb-8e3b-76e9dce47d1e.mov

**Describe the solution you'd like**
Ability to control the hideOnScroll behaviour through a prop on AppBar component would be really helpful for enhancing user experience.

**Describe alternatives you've considered**
Adding a second toggle button behind the appbar in side menu, but that would be redundant.

pkspyder007:
@djhi We are using React-admin a lot at [Navidrome](https://github.com/navidrome/navidrome). There are few suggestions I would like to give.
- Having the ability for fixed/sticky side menu 
- Changing Menu icon in AppBar on toggling

React-Admin provides the ability to provide a custom implementation for these use-cases but deviates us from the framework. Having the ability to change such properties based on props or some other method will help a lot.
I appreciate the work that you people have put in creating such an amazing technology 🙌

fzaninotto:
Fixed by #6178

PR #6178, opened 2021-04-16
Add container prop to AppBar

WiXSL:
Implements #6174

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -740,9 +740,22 @@ const App = () => (
 
 ## Replacing The AppBar
 
-For more drastic changes of the top component, you will probably want to create an `<AppBar>` from scratch instead of just passing children to react-admin's `<AppBar>`.
+By default, React-admin uses [Material-ui's `<AppBar>` component](https://material-ui.com/api/app-bar/) together with a custom container that internally uses a [Slide](https://material-ui.com/api/slide) to hide the `AppBar` on scroll. Here is an example of how to change this container with any component:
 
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
As we want to keep the default behavior, I think having to pass `false` explicitly is not very intuitive. What do you think about adding a `stickOnScroll` prop which would be `false` by default? Maybe there is a better name for it though ;)

fzaninotto:
Why don't you call this prop `container` and let users pass a Fragment instead of HideOnScroll?

docs/Theming.md
```diff
@@ -740,9 +740,22 @@ const App = () => (
...
 
-By default, React-admin uses [Material-ui's `<AppBar>` component](https://material-ui.com/api/app-bar/) together with [react-headroom](https://github.com/KyleAMathews/react-headroom) to hide the `AppBar` on scroll. Here is an example top bar rebuilt from scratch to remove the "headroom" effect:
+```jsx
+// in src/MyAppBar.js
+import * as React from 'react';
+import { Fragment } from 'react';
+import { AppBar } from 'react-admin';
+
+const MyAppBar = props => (
+    <AppBar {...props} conatainer={Fragment} />
```
djhi, line 752:
```suggestion
    <AppBar {...props} container={Fragment} />
```

Achievement unlocked: Corrected El Corrector!

djhi (changes requested):
:joy: :tada:

djhi (approved):

fzaninotto (approved):

fzaninotto:
thanks!

PR #6182, opened 2021-04-19
[TypeScript] Fix useMutation requires payload argument

jtomaszewski:
When you call `useMutation`, `payload` argument is already optional. However TS type definition forces you to pass `payload: {}` . This commit removes that requirement.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6189, opened 2021-04-19
Fix Notification component undo color

WiXSL:
Fixes #6173

PR #6190, opened 2021-04-19
Improve Notification component colors definitions

WiXSL:
This change matches `info`, `error` and `warning` CSS rules with material-ui's for the `Notification` component.

notification-colors


fzaninotto:
You've already proposed a similar change in #5710, and I said then that we don't want it in core. Why should we merge it now?

Also, this can be achieved in userland through themes.

WiXSL:
Ok.
This PR was after #6189 was merged but before you revert it, plus I thought the main problem of this type of change was the color of the undo button text which was fixed (before reverting) in #6189. So I move on with this proposal. 
Now I understand.
Closing

PR #6191, opened 2021-04-19
[RFR] Add Notification multiLine prop

WiXSL:
This change enables `Notification` component to show multi-line messages.


fzaninotto:
Can you please document usage, unit test it, and attach a screenshot about the rendered notification?

WiXSL:
Absolutely

WiXSL:
notification

packages/ra-core/src/sideEffect/useNotify.ts
```diff
@@ -28,13 +28,15 @@ const useNotify = () => {
             type: NotificationType = 'info',
             messageArgs: any = {},
             undoable: boolean = false,
-            autoHideDuration?: number
+            autoHideDuration?: number,
+            multiLine?: boolean
```
fzaninotto, line 32: At this stage, it becomes obvious that the signature of the `notify` callback isn't right. As we'r adding new options, it makes calls super weird. For instance, to enable multiline without changing other parameters:

```jsx
notify('foo', undefined, undefined, undefined, undefined, true)
```

We must switch to a simpler signature: `notify(message, options)`. This can be done in a backwards-compatible way by inspecting the type of the second argument at runtime. TypeScript may get in the way, but it supports multiple call signatures.

fzaninotto (changes requested):

fzaninotto:

WiXSL:
This was closed, but valid

fzaninotto:
Right, sorry. Reopening.

fzaninotto:
I see that you've updated this PR, but you didn't take my review into account. Could you please introduce the new signature together with the multiline option?

WiXSL:
I believe I'm getting errors from failing tests related to code that is in `master` but not ported to `next` branch

WiXSL:
> I'm getting errors from failing tests related to code that is in `master` but not ported to `next` branch

Now is ok

djhi:
Sorry @WiXSL, this can now target the master branch as it will be released as `3.19`

PR #6197, opened 2021-04-21
fix(RaThemeOptions): Make sidebar width options optional

jtomaszewski:
This just fixes TS type definition. The width options are already optional ( https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/Sidebar.tsx ).


djhi (approved):
Thanks!

PR #6204, opened 2021-04-22
Publish data-generator typings

floo51:
Publish `data-generator` typings so that it can be used directly within typescript without having to add own typings.

Worth adding `declarationMap` too?


fzaninotto:
yep, please add declarationMaps, too

djhi (approved):

PR #6206, opened 2021-04-22
[DOC] Add OData data provider

jvert:
I have just open-sourced an OData data provider that we are using for our react-admin implementation. It would be great if you could add it to the documentation.


djhi (approved):
Thanks!

PR #6163, opened 2021-04-13
Add Initial Support For Creation in Choices Inputs

djhi:
- [x] Add support for creation in `SelectInput`
- [x] Add support for creation in `AutocompleteInput`
- [x] Check support for creation in `ReferenceInput`
- [x] Add support for creation in `AutocompleteArrayInput`
- [x] Check support for creation in `ReferenceArrayInput`
- [x] Add support for creation in `SelectArrayInput`
- [x] Tests
- [x] Documentation

ra-quick-create

ra-quick-create-autocomplete-modal-reference

ra-quick-create-autocompletearray-modal-reference

PR #6215, opened 2021-04-27
Add Initial Support For Creation in Choices Inputs

djhi:
Supsedes #6163

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx
@@ -1,10 +1,11 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, waitFor } from '@testing-library/react';
 import { Form } from 'react-final-form';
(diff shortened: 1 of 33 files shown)


packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx
```diff
@@ -303,17 +304,131 @@ describe('<SelectArrayInput />', () => {
...
-                    )}
+                    render={() => <SelectArrayInput {...defaultProps} />}
                 />
             );
 
             expect(queryByRole('progressbar')).toBeNull();
         });
     });
+
+    test('should support creation of a new choice through the onCreate event', async () => {
```
fzaninotto, line 315: `test` used instead of `it` again... Please make sure there aren't any remaining in your PR

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx
djhi: We use both in the whole codebase btw

fzaninotto (approved):

Issue #6212, opened 2021-04-26
Allow passing children to ListAction component


WiXSL:
In order to add custom `actions` to the `List` component, one has to replace the entire default ´ListActions´ component with something else.

Couldn't it be possible to re-use `react-admin`'s `ListActions` component and pass children to it?

fzaninotto:
As explained in #6218, I don't think this is a good idea. I agree that the current way of overriding the List actions isn't very user friendly, but the solution that you propose isn't a clear win and brings confusion.

fzaninotto:
Fixe by #6218

PR #6218, opened 2021-04-27
[Doc] Improve ListActions override

WiXSL:
Implements #6212

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -90,7 +90,37 @@ The title can be either a string or an element of your own.
 
 ![Actions Toolbar](./img/actions-toolbar.png)
 
-You can replace the list of default actions by your own element using the `actions` prop:
+You can add extra actions to a `<List>` by using the `actions` prop:
+
+{% raw %}
+```jsx
+import * as React from 'react';
+import { ListActions, Button } from 'react-admin';
+import IconEvent from '@material-ui/icons/Event';
+
+const CustomListActions = (props) => {
+    return (
+        <ListActions {...props}>
+            {/* Add your custom actions */}
+            <Button
+                onClick={() => { alert('Your custom action'); }}
+                label="Show calendar"
+            >
+                <IconEvent />
+            </Button>
+        </ListActions>
+    );
+};
+
+export const PostList = (props) => (
+    <List {...props} actions={<CustomListActions />}>
+        ...
+    </List>
+);
+```
+{% endraw %}
+
+If you want to have control over default actions as well, you can replace the entire `actions` component by your own:
 
 {% raw %}
 ```jsx
--- a/docs/List.md
+++ b/docs/List.md
@@ -120,7 +120,7 @@ export const PostList = (props) => (
 );
 ```
 
-If you want to have control over default actions as well, you can replace the entire `actions` component by your own:
+This allows you to further control how the default actions behaves. For example, you could disable the `<ExportButton>` when the list is empty:
 
 {% raw %}
 ```jsx
--- a/docs/List.md
+++ b/docs/List.md
@@ -136,39 +136,17 @@ import IconEvent from '@material-ui/icons/Event';
 const ListActions = (props) => {
     const {
         className,
-        exporter,
-        filters,
         maxResults,
         ...rest
     } = props;
     const {
-        currentSort,
-        resource,
-        displayedFilters,
-        filterValues,
-        hasCreate,
-        basePath,
-        selectedIds,
-        showFilter,
         total,
     } = useListContext();
     return (
         <TopToolbar className={className} {...sanitizeListRestProps(rest)}>
-            {filters && cloneElement(filters, {
-                resource,
-                showFilter,
-                displayedFilters,
-                filterValues,
-                context: 'button',
-            })}
-            <CreateButton basePath={basePath} />
-            <ExportButton
-                disabled={total === 0}
-                resource={resource}
-                sort={currentSort}
-                filterValues={filterValues}
-                maxResults={maxResults}
-            />
+            {cloneElement(filters, { context: 'button' })}
+            <CreateButton />
+            <ExportButton disabled={total === 0} maxResults={maxResults} />
             {/* Add your custom actions */}
             <Button
                 onClick={() => { alert('Your custom action'); }}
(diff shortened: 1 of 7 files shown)


djhi:
:+1:  Maybe we should do the same for the other action components as well.

fzaninotto (commented):
Thanks for your patch. However, I'm not comfortable with this change, and for 2 reasons:

- it creates 2 ways to do one simple thing. This is confusing, which way should a developer use and why? 
- your API "hides" that some other buttons are included. I think that will lead to WTF moments for developers. 

Overall, I don't think we should implement that.

WiXSL:
Ok, how about move the logic of the actions to hooks and you always override the actions prop like now, but having control of default actions by calling this hooks (or not).

Example: 

```jsx
import Button from '@material-ui/core/Button';
import { TopToolbar, Edit, useEditDefaultActions } from 'react-admin';

const PostEditActions = (props) => {
     const DefaultActions = useEditDefaultActions(props);

    return <TopToolbar>
        <DefaultActions/> 
        {/* Add your custom actions */}
        <Button 
            color="primary"
            onClick={() => { alert('Your custom action'); }}
        >
            Custom Action
        </Button>
    </TopToolbar>;
};

    <Edit actions={<PostEditActions />} {...props}>
        ...
    </Edit>
);
```

WiXSL:
@fzaninotto,
What do you think of this idea?

djhi:
Why a hook here? We could just have a component rendering a fragment with the default actions right?

WiXSL:
You are right.

fzaninotto:
When I think about it, only the `<ListActions>` are hard to override. Maybe this is where we should do the change. 

First, the documentation could be simplified since we now use the various context:

```diff
const ListActions = (props) => {
    const {
        className,
-       exporter,
-       filters,
-       maxResults,
        ...rest
    } = props;
    const {
-       currentSort,
-       resource,
-       displayedFilters,
-       filterValues,
-       hasCreate,
-       basePath,
-       selectedIds,
-       showFilter,
        total,
    } = useListContext();
    return (
        <TopToolbar className={className} {...sanitizeListRestProps(rest)}>
-           {filters && cloneElement(filters, {
-               resource,
-               showFilter,
-               displayedFilters,
-               filterValues,
-               context: 'button',
-           })}
+           {cloneElement(filters, { context: 'button' })}
-           <CreateButton basePath={basePath} />
+           <CreateButton />
-           <ExportButton
-               disabled={total === 0}
-               resource={resource}
-               sort={currentSort}
-               filterValues={filterValues}
-               maxResults={maxResults}
-           />
+           <ExportButton disabled={total === 0} maxResults={maxResults} />
            {/* Add your custom actions */}
            <Button
                onClick={() => { alert('Your custom action'); }}
                label="Show calendar"
            >
                <IconEvent />
            </Button>
        </TopToolbar>
    );
};
```

For a simpler use case, the doc is even simpler:

```jsx
const ListActions = () => (
    <TopToolbar >
        {cloneElement(filters, { context: 'button' })}
        <CreateButton />
        <ExportButton />
        {/* Add your custom actions */}
        <Button
            onClick={() => { alert('Your custom action'); }}
            label="Show calendar"
        >
            <IconEvent />
        </Button>
    </TopToolbar>
);
```

I think this is simple enough that we don't have to introduce a new component for that.

WiXSL:
So, I should revert everything and update the only documentation?

docs/List.md
```diff
@@ -90,7 +90,37 @@ The title can be either a string or an element of your own.
 
 ![Actions Toolbar](./img/actions-toolbar.png)
 
-You can replace the list of default actions by your own element using the `actions` prop:
+You can add extra actions to a `<List>` by using the `actions` prop:
```
djhi, line 93: The original sentence was fine imo

docs/List.md
```diff
@@ -90,7 +90,37 @@ The title can be either a string or an element of your own.
...
+);
+
+export const PostList = (props) => (
+        <List {...props} actions={<ListActions/>}>
+          ...
+        </List>
+);
+```
+
+If you want to have control over default actions as well, you can replace the entire `actions` component by your own:
```
djhi, line 123:
```suggestion
This allows you to further control how the default actions behaves. For example, you could disable the `<ExportButton>` when the list is empty:
```

djhi (changes requested):
Thanks!

docs/List.md
```diff
@@ -92,6 +92,36 @@ The title can be either a string or an element of your own.
...
+);
+
+export const PostList = (props) => (
+        <List {...props} actions={<ListActions/>}>
+          ...
+        </List>
+);
+```
+
+This allows you to further control how the default actions behaves. For example, you could disable the `<ExportButton>` when the list is empty:
```
djhi, line 123:
```suggestion
This allows you to further control how the default actions behave. For example, you could disable the `<ExportButton>` when the list is empty:
```

Achievement unlocked: tricked El Corrector :laughing:

djhi (changes requested):

djhi (approved):

docs/List.md
```diff
@@ -104,45 +134,25 @@ import {
     ExportButton,
     Button,
     sanitizeListRestProps,
+    useListContext
```
fzaninotto, line 137: useListContext is already included line 131

packages/ra-ui-materialui/src/list/ListActions.tsx
```diff
@@ -15,6 +15,47 @@ import { ToolbarProps } from '@material-ui/core';
...
+ *          basePath,
+ *          selectedIds,
+ *          showFilter,
+ *          total,
+ *     } = useListContext();
+ *
+ *     const PostListActions = ({ basePath }) => (
+ *         <TopToolbar>
+ *             <CreateButton basePath={basePath} />
+ *             <ExportButton
```
fzaninotto, line 42: why  not use the simpler version?

```
{cloneElement(props.filters, { context: 'button' })}  
<CreateButton/>
<ExportButton/>
```

fzaninotto (changes requested):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6219, opened 2021-04-28
Export `PublicFieldProps` and `InjectedFieldProps` types (so devs can use them to define their own field components)

jtomaszewski:
In our admin app, we create a lot of our own custom `*Field` components. When defining them, we need to add typescript definitions for props like `addLabel?: boolean`, which aren't used directly by the field component, but by the components wrapping it (e.g. `Labeled`).

All props like that are currently nicely defined in `react-admin` as `PublicFieldProps` or `InjectedFieldProps`, but unfortunately we can't import them, as they aren't exported by the library.

This PR exports them, so they can be used.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6220, opened 2021-04-28
Export ConvertedDataProvider interface

hieusmiths:
Sometimes we should export ConvertedDataProvider.

For some case, we should call API without effect like redux dispatch loading.

Thanks


fzaninotto:

I don't understand your use case, and you messed up your PR (it contains commits that aren't from you). 

Can you please give an example use case, and fix your PR?

fzaninotto:
No news for a month, closing.

PR #6223, opened 2021-04-28
[Doc] Fix DatagridRow example

### LINKED ISSUE #6222: DataGrid overriding DataGridBody example in documentation is incorrect
**What you were expecting:**
Disabled checkboxes where isRowSelectable criteria is false and the remaining rows remain selectable

**What happened instead:**

Error when clicking a checkbox due to missing event parameter
**Steps to reproduce:**

Follow example from here: https://marmelab.com/react-admin/List.html#body-element

I fixed it by making the following changes:
Adding selectable prop instead of record.selectable which doesn't exist
 and added event parameter as onToggleItem expects this
```
const MyDatagridRow = ({ record, resource, id, onToggleItem, children, selected, selectable, basePath }) => (
    <TableRow key={id}>
        {/* first column: selection checkbox */}
        <TableCell padding="none">
            <Checkbox
                disabled={selectable}
                checked={selected}
                onClick={(event) => onToggleItem(id, event)}
            />
        </TableCell>
        {/* data columns based on children */}
        {React.Children.map(children, field => (
            <TableCell key={`${id}-${field.props.source}`}>
                {React.cloneElement(field, {
                    record,
                    basePath,
                    resource,
                })}
            </TableCell>
        ))}
    </TableRow>
);```

WiXSL:
Fixes #6222

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -2021,14 +2021,14 @@ import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import Checkbox from '@material-ui/core/Checkbox';
 
-const MyDatagridRow = ({ record, resource, id, onToggleItem, children, selected, basePath }) => (
+const MyDatagridRow = ({ record, resource, id, onToggleItem, children, selected, selectable, basePath }) => (
     <TableRow key={id}>
         {/* first column: selection checkbox */}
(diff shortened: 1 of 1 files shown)


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6226, opened 2021-04-29
Fix hideFilter called repeatedly only registers the last call

### LINKED ISSUE #6055: Not possible to hide a few filters
**What you were expecting:**
3 filters are hidden

**What happened instead:**
hidden only first filter

**Steps to reproduce:**

**Related code:**

```
hideFilter('expiringCertificates');
hideFilter('caName');
hideFilter('date');
```

**Other information:**
It's better if we will have possibility to pass array of filters to hideFilter function

**Environment**

* React-admin version: 3.13.4
* React version: 16.14.0
* Browser: Chrome
* Stack trace (in case of a JS error): No error in console

fzaninotto:
Closes #6055

### DIFF
--- a/packages/ra-core/src/controller/useListParams.ts
+++ b/packages/ra-core/src/controller/useListParams.ts
@@ -56,7 +56,6 @@ const defaultSort = {
 };
 
 const defaultParams = {};
-let tempParams;
 
 /**
  * Get the list parameters (page, sort, filters) and modifiers.
@@ -126,6 +125,7 @@ const useListParams = ({
                 : defaultParams,
         shallowEqual
     );
+    const tempParams = useRef<ListParams>();
 
     const requestSignature = [
         location.search,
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/controller/useListParams.ts
```diff
@@ -229,38 +240,18 @@ const useListParams = ({
     );
 
     const hideFilter = useCallback((filterName: string) => {
-        // we don't use lodash.set() for displayed filters
```
fzaninotto, line 232: I moved the logic to the queryReducer for easier testability, but that has no effect on the bug.

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -229,38 +240,18 @@ const useListParams = ({
...
         changeParams({
-            type: SET_FILTER,
-            payload: { filter, displayedFilters },
+            type: HIDE_FILTER,
+            payload: filterName,
         });
     }, requestSignature); // eslint-disable-line react-hooks/exhaustive-deps
 
     const showFilter = useCallback((filterName: string, defaultValue: any) => {
-        // we don't use lodash.set() for displayed filters
```
fzaninotto, line 250: Same, this change is just cosmetic

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -163,21 +164,31 @@ const useListParams = ({
     }, []); // eslint-disable-line
 
     const changeParams = useCallback(action => {
```
fzaninotto, line 166: Here is the fix: since changeParams can be called several times in the same tick, we keep the result if the reducer locally so that we can start from it if another call comes immediately after. And we change the location (or the local state) only at the next tick, when all the sync changes were applied.

packages/ra-core/src/controller/useListParams.ts
```diff
@@ -56,6 +56,7 @@ const defaultSort = {
 };
 
 const defaultParams = {};
+let tempParams;
```
fzaninotto, line 59: This should be a ref

djhi: Why not making it a ref then?

djhi (approved):

packages/ra-core/src/controller/useListParams.ts
fzaninotto: I added the note for myself. it's a ref now.

PR #6230, opened 2021-04-30
Fix logout button appears in two different menus

### LINKED ISSUE #6228: the Mobile menu doesn't need the logout button
## Problem

On Mobile, the Logout button appears twice: once in the main menu, and once in the user menu.

logout

It's confusing, and has no UX justification. 

The reason is that it was added to the navigation menu before react-admin had a user menu. But when we added it to the user menu, we didn't remove it form the navigation menu.

## Solution

Don't render the logout menu item in the navigation menu.

fzaninotto:
Closes #6228

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -856,7 +856,7 @@ const App = () => (
 );
 ```
 
-**Tip**: You can generate the menu items for each of the resourc emages by reading the Resource configurations from the Redux store: 
+**Tip**: You can generate the menu items for each of the resources by reading the Resource configurations from the Redux store: 
 
 ```jsx
 // in src/Menu.js
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -916,7 +916,7 @@ import Badge from '@material-ui/core/Badge';
 } />
 ```
 
-The `icon` prop allows to set the menu left icon.
+The `letfIcon` prop allows to set the menu left icon.
 
 Additional props are passed down to [the underling material-ui `<MenuItem>` component](https://material-ui.com/api/menu-item/#menuitem-api).
 
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -807,9 +807,56 @@ Check [the `ra-preferences` documentation](https://marmelab.com/ra-enterprise/mo
 
 ## Using a Custom Menu
 
-By default, React-admin uses the list of `<Resource>` components passed as children of `<Admin>` to build a menu to each resource with a `list` component.
+By default, React-admin uses the list of `<Resource>` components passed as children of `<Admin>` to build a menu to each resource with a `list` component. If you want to reorder, add or remove menu items, for instance to link to non-resources pages, you have to provide a custom `<Menu>` component to your `Layout`.
 
-If you want to add or remove menu items, for instance to link to non-resources pages, you can create your own menu component:
+### Custom Menu Example
+
+You can create a custom menu component using the `<DashboardMenuItem>` and `<MenuItemLink>` components:
+
+```jsx
+// in src/Menu.js
+import * as React from 'react';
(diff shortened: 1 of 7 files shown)


packages/ra-ui-materialui/src/layout/Menu.tsx
```diff
@@ -89,12 +80,9 @@ const Menu: FC<MenuProps> = props => {
                         leftIcon={
                             resource.icon ? <resource.icon /> : <DefaultIcon />
                         }
-                        onClick={onMenuClick}
```
fzaninotto, line 92: I refactored the `<MenuItemLink>` to let it pull the props it needs (rather than expecting them to be pushed by the `<Sidebar>`). 

This greatly simplifies the creation of custom menus.

docs/Theming.md
```diff
@@ -843,71 +888,88 @@ const Menu = ({ onMenuClick, logout }) => {
...
-export default MyLayout;
+<MenuItemLink to="/custom-route" primaryText={
+    <Badge badgeContent={4} color="primary">
+        Notifications
+    </Badge>
+} />
 ```
 
-Then, use this layout in the `<Admin>` `layout` prop:
+The `icon` prop allows to set the menu left icon.
```
WiXSL, line 919: `icon` => `leftIcon`

fzaninotto:
Wait, since #5575 it's the menu's responsibility to reduce width when the sidebar is closed. I'm not sure why, but if it's the case this should be documented.

fzaninotto:
OK, it's both the Sidebar AND the Menu's responsibility to reduce width, in order to support one use case: using a react-admin oss `<Menu>` inside a react-admin ee app. So we need to keep that in the ra oss code, but people who customize the menu in a re oss app don't need to implement open/closed state handling as it's already handled by the Sidebar. 

So this PR is ready.

docs/Theming.md
```diff
@@ -807,9 +807,56 @@ Check [the `ra-preferences` documentation](https://marmelab.com/ra-enterprise/mo
...
+import { Layout }  from './Layout';
+
+const App = () => (
+    <Admin layout={Layout} dataProvider={simpleRestProvider('http://path.to.my.api')}>
+        // ...
+    </Admin>
+);
+```
+
+**Tip**: You can generate the menu items for each of the resourc emages by reading the Resource configurations from the Redux store:
```
djhi, line 859: Something weird here

examples/demo/src/layout/Menu.tsx
```diff
@@ -1,9 +1,8 @@
 import * as React from 'react';
 import { FC, useState } from 'react';
 import { useSelector } from 'react-redux';
-import SettingsIcon from '@material-ui/icons/Settings';
 import LabelIcon from '@material-ui/icons/Label';
-import { useMediaQuery, Theme, Box } from '@material-ui/core';
+import { Box } from '@material-ui/core';
```
djhi, line 5: I thought we should avoid `Box`

examples/demo/src/layout/Menu.tsx
```diff
@@ -22,17 +21,13 @@ import { AppState } from '../types';
 
 type MenuName = 'menuCatalog' | 'menuSales' | 'menuCustomers';
 
-const Menu: FC<MenuProps> = ({ onMenuClick, logout, dense = false }) => {
+const Menu: FC<MenuProps> = ({ dense = false }) => {
```
djhi, line 24: Good opportunity to migrate away from `FC`

packages/ra-ui-materialui/src/layout/Menu.tsx
djhi: Indeed :+1:

djhi (changes requested):

djhi (approved):

Issue #6233, opened 2021-05-02
useQuery hook sets loaded prop to false when throws error


paulo9mv:
**What you were expecting:**

When `useQuery()` throws an error, i'm expecting `loaded` to be **true**, according to the docs:
`error: { error: [error from response], loading: false, loaded: true }`

**What happened instead:**

`loaded` is **false** after throwed error.

**Steps to reproduce:**

In simple demo, call useQuery() hook in a way that it throw an error. It will set ``loaded`` to **false**.

**Related code:**

https://github.com/marmelab/react-admin/blob/9a2fc59fb81b3e2e014eeecc194babb340ff833e/packages/ra-core/src/dataProvider/useQuery.ts
```
// useQuery.ts
.catch(error => {
                setState({
                    error,
                    loading: false,
                    loaded: false,  // <--- this line seems wrong
                    refetch,
                });
            });
```

fzaninotto:
The problem lies in the documentation, not in the code. `useQuery` (and other query hooks) should return `loaded: false` in case of error, because the data isn't loaded.

PR #6234, opened 2021-05-02
fix: change loaded prop false to true in useQuery catch

paulo9mv:
Fixes https://github.com/marmelab/react-admin/issues/6233


fzaninotto:
thanks for you PR. But as explained in #6233, the fix should be in the documentation, not in the code.

paulo9mv:
thanks for the clarification :)

PR #6235, opened 2021-05-02
Fix useGetOne and useGetList don't expose the refetch method

fzaninotto:
but useGetMany does (!)


fzaninotto:
This will have to be PRed against `next` as the change must bubble up until `useListController`, and therefore changes the `ListContext` signature.

PR #6247, opened 2021-05-05
Add the posibility to pass false to the toolbar prop

cpestour:
This is not really possible to remove the toolbar from a form component unless by passing an empty component.
I added the possibility to pass false to be able to not display the toolbar:

```jsx
import { Edit, SimpleForm } from 'react-admin';

    <Edit {...props}>
        <SimpleForm toolbar={false}>
            // ...
        </SimpleForm>
    </Edit>
);
```

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -1684,6 +1684,21 @@ export const PostEdit = (props) => (
 );
 ```
 
+In case you don't want to have a toolbar to be displayed, you can set `false` to the `toolbar` prop:
+
+```jsx
+import * as React from "react";
+import { Edit, SimpleForm } from 'react-admin';
+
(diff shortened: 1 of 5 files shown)


fzaninotto (commented):
Thanks for the PR. Could you tell us more about your use case? Why wouldn't you want a toolbar in the edit / creation form? And how will users validate the form without a submit button?

cpestour:
Yes sure. I have an Edit form with a single `SelectInput`. The save is trigger by the `onChange` of the input. Therefore I don't need the toolbar

fzaninotto:
Then the SimpleForm is overkill in your case. I suggest you use the FormWithRedirect with a custom render function, or a basic Form that you build by yourself using react-final-form.

PR #6251, opened 2021-05-06
Fix Size prop in AutocompleteInput

gebeto:


gebeto:
`size` prop should be passed in `options.InputProps` prop

PR #6266, opened 2021-05-10
[DX] Thrown an error when using a Reference field without the associated Resource

### LINKED ISSUE #6258: Thrown an error when using a Reference field without the associated Resource
Many newcomers don't understand why their ReferenceField or referenceManyField doesn't display anything - generally, they have forgotten to declare the related Resource. 

It's documented, but many newcomers don't read the documentation.

We can detect that such components are used without the right resource, we should throw an explicit error in that case.

fzaninotto:
Closes #6258

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render, act } from '@testing-library/react';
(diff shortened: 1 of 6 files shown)


djhi (approved):

djhi (commented):
Why is targeting the `next` branch?

fzaninotto:
Because it's not a bug fix.

djhi:
Right but it does not really add any feature, right?

fzaninotto:
Right, but a bugfix release can only contain bug fixes, so...

PR #6267, opened 2021-05-10
Create UserMenuProps interface for ra-ui-materialui/src/layout/UserMenu.tsx

Koli14:


packages/ra-ui-materialui/src/layout/UserMenu.tsx
```diff
@@ -37,7 +37,7 @@ const TransformOrigin: PopoverOrigin = {
     horizontal: 'right',
 };
 
-const UserMenu = props => {
+const UserMenu: FC<UserMenuProps> = props => {
```
djhi, line 40: Please avoid `FC` and type the props directly. See https://github.com/facebook/create-react-app/pull/8177. We're trying to avoid it for new code and migrate old code  progressively.

djhi (changes requested):

packages/ra-ui-materialui/src/layout/UserMenu.tsx
```diff
@@ -37,7 +37,7 @@ const TransformOrigin: PopoverOrigin = {
     horizontal: 'right',
 };
 
-const UserMenu = props => {
+const UserMenu: React.FC<UserMenuProps> = props => {
```
djhi, line 40:
```suggestion
const UserMenu = (props: UserMenuProps) => {
```

djhi (changes requested):

djhi (approved):
Thanks!

packages/ra-ui-materialui/src/layout/UserMenu.tsx
```diff
@@ -122,4 +122,13 @@ UserMenu.defaultProps = {
     icon: <AccountCircle />,
 };
 
+export interface UserMenuProps {
+    children?: React.ReactNode;
+    classes?: object;
+    className?: string;
+    label?: string;
+    logout?: React.ReactNode;
+    icon?: ReactNodeLike;
```
fzaninotto, line 131: Why is this one a ReactNodeLike, while children and logout, which also expect a node, are React.ReactNode? What's the difference?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/layout/UserMenu.tsx
Koli14: I have no idea, I checked some other source from RA, what they use by icon.

fzaninotto: React-admin doesn't use ReactNodeLike. 

And if you open a PR with code you don't understand, I don't want to merge it.

PR #6268, opened 2021-05-10
Add `component` prop to TabbedForm component

WiXSL:
Like `SimpleForm` has.
This allows to have access to the children inside the form without rewriting everything.


djhi (approved):
:+1:

fzaninotto (changes requested):

fzaninotto:
I realize that with my requested changes, your PR is now a breaking change. A developer can no longer override the content className by overriding the `classes` prop.

So I don't think we'll merge this before 4.0 (which is not yet schedules). 

Besides, something bothers me in this PR. I don't understand the use case. What does this PR allow that isn't yet possible with the CSS API?

WiXSL:
It also allows to insert logic with the use `react-final-form` hooks, since we are inside the `<form>`

fzaninotto:
> It also allows to insert logic with the use of `react-final-form` hooks, since we are inside the `<form>`

Can you elaborate?

WiXSL:
I'll give you a more concrete use case.
A custom React context to show/hide progress blocking form inputs programmatically.


```jsx
const ProgressProvider = ({ children }) => {
    const [open, setOpen] = useState(false);

    const context = useMemo(
        () => ({
            open,
            showProgress: () => setOpen(true),
            hideProgress: () => setOpen(false)
        }),
        [open]
    );

    return (
        <ProgressContext.Provider value={context}>
            {children}
        </ProgressContext.Provider>
    );
};

const ComponentInner = props => {
    const cls = useStyles(props);
    const context = useContext(MyContext);

    return (
        <CardContentInner className={context.open ? cls.block : ''}>
            {context.open && <CircularProgress
                className={cls.progress}
                thickness={4}
                size={50}
            />}
            {props.children}
        </CardContentInner>
    );
};

const ChildrenContainer = props => (
    <ProgressProvider>
        <ComponentInner {...props} />
    </ProgressProvider>
);

<TabbedForm component={ChildrenContainer}/>

fzaninotto:
In your use case, why can't you wrap the entire TabbedForm in your context?

WiXSL:
Because I only want to visually block the inputs visual area, on the entire view, plus this allows me to use `react-final-form` hooks in my `ComponentInner` since I'm inside the form

fzaninotto:

WiXSL:
This one was closed automatically but valid

WiXSL:
Is it too late to add this to `next`?

WiXSL:
Implemented by #7348

PR #6276, opened 2021-05-12
[Doc] Fix TestContext JsDocs

WiXSL:


packages/ra-test/src/TestContext.tsx
```diff
@@ -55,7 +55,7 @@ const dataProviderDefaultResponse = { data: null };
  * @example
  * // in a react testing-library test, using jest.
  * const utils = render(
- *     <TestContext initialState={{ admin: { resources: { post: { data: { 1: {id: 1, title: 'foo' } } } } } }}>
+ *     <TestContext initialState={{ admin: { resources: { posts: { data: { 1: {id: 1, title: 'foo' } } } } } }}>
```
djhi, line 58: Why?

WiXSL: `posts` are the resources, `post` indicate only one resource

djhi: I guess that's more coherent with the other examples but nothing prevent you to name your resources `post`.

djhi (approved):
Thanks!

PR #6277, opened 2021-05-13
DOC: Adds missing record props in title-handling

pamuche:


WiXSL:
@pamuche,
It is not necessary to pass the record to the Title component. 
The `Edit` view will pass the record internally

djhi:
@WiXSL is right. You don't need to pass the `record` prop. Besides, the `Title` component will soon get the current record from the `RecordContext`

PR #6282, opened 2021-05-14
Fix missing margin prop in several Labeled components

WiXSL:


packages/ra-ui-materialui/src/input/SelectInput.tsx
```diff
@@ -115,6 +115,7 @@ export const SelectInput = (props: SelectInputProps) => {
         label,
         loaded,
         loading,
+        margin = 'dense',
```
djhi, line 118: Why a default here and not in the others ? Besides, I think the Form components already have this default.

djhi (commented):
Thanks!

packages/ra-ui-materialui/src/input/SelectInput.tsx
WiXSL: I copy the default from the other inputs.
If you want I can remove all the defaults for margin?

WiXSL: But I only see margin's default value only being set for most Inputs like that

djhi: I forgot we did this, nvm :)

djhi (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6285, opened 2021-05-17
Fix performance regression causing unnecessary redraws

fzaninotto:
## Problem

Our context hooks (e.g. `useListContext`) always return new items, even though nothing has changed. 

The problem is that they use lodash/merge, which is recursive. When merging complex objects, the result is almost always a new object.

## Solution

~~Use lodash/extend instead~~. 
Use lodash/defaults instead (because extend considers undefined values as an override, see https://stackoverflow.com/a/33247597/1333479)

## Before

image

## After

image


djhi (approved):
:fire:

djhi (approved):

PR #6288, opened 2021-05-19
Disable <SaveButton/> while Inputs are being validated asynchronously

### LINKED ISSUE #4557: Async validation on field level and form submit button loading
I am trying to implement an async validation for a field with react-final-form. All is good but the submit button is not displaying loading animation when this validation is ongoing. When I click the button it will start validating but won't show the loading animation until the async validation is done only after that it will start showing the loading animation as it starts the submission. It's bad UX that when a user clicks on the button does not know what is actually happening, he might think the form is broken.

The best solution would be to use the react-final-form validating form state boolean. This in turn should only be used if the submit is actually clicked and if the validating flag is true then we should show the loading button. 

My current workaround is that I override the FormToolbar and send custom `saving` prop to the original Toolbar, but it would be nice having this in core.

```typescript
  const formState = useFormState();
  const [isSaving, setIsSaving] = useState(false);

  const customHandleSubmit = useCallback((event: any) => {
    setIsSaving(formState.validating || saving);
    handleSubmit(event);
  }, [saving, handleSubmit, formState]);

  const customHandleSubmitWithRedirect = useCallback((event: any) => {
    setIsSaving(formState.validating || saving);
    handleSubmitWithRedirect(event);
  }, [saving, handleSubmitWithRedirect, formState]);

  useEffect(() => setIsSaving(saving), [saving]
```

WiXSL:
Fixes #4557

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -362,14 +362,6 @@ describe('<SaveButton />', () => {
             update: (_, { data }) => Promise.resolve({ data }),
         } as unknown) as DataProvider;
 
-        const defaultSubscription = {
-            submitting: true,
-            pristine: true,
-            valid: true,
-            invalid: true,
-            validating: true, // Necessary for async validations
-        };
-
         const validateAgainstServer = value =>
             new Promise(resolve => setTimeout(() => value, 400));
 
@@ -385,7 +377,7 @@ describe('<SaveButton />', () => {
             <DataProviderContext.Provider value={dataProvider}>
                 <ThemeProvider theme={theme}>
                     <Edit {...defaultEditProps}>
-                        <SimpleForm subscription={defaultSubscription}>
+                        <SimpleForm>
                             <TextInput
                                 source="title"
                                 validate={validateAsync}
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -371,7 +371,7 @@ describe('<SaveButton />', () => {
         };
 
         const validateAgainstServer = value =>
-            new Promise(resolve => setTimeout(() => value !== 'ipsum', 400));
+            new Promise(resolve => setTimeout(() => value, 400));
 
         const validateAsync = async (value, allValues) => {
             const isUnique = await validateAgainstServer(value);
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -352,4 +352,60 @@ describe('<SaveButton />', () => {
             });
         });
     });
+
+    it('should disable <SaveButton/> if an input is being validated asynchronously', async () => {
+        const dataProvider = ({
+            getOne: () =>
+                Promise.resolve({
+                    data: { id: 123, title: 'lorem' },
+                }),
+            update: (_, { data }) => Promise.resolve({ data }),
+        } as unknown) as DataProvider;
+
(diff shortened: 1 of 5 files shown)


djhi (approved):
Thanks!

fzaninotto:
Please add a unit test to prove that it fixes the problem.

WiXSL:
In order for this behavior to  work, the react-final-form Form must be subscribed to `validating`

WiXSL:
@fzaninotto, I've added a test as requested

djhi (approved):

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -337,4 +337,52 @@ describe('<SaveButton />', () => {
...
+        const dataProvider = ({
+            getOne: () =>
+                Promise.resolve({
+                    data: { id: 123, title: 'lorem' },
+                }),
+            update: (_, { data }) => Promise.resolve({ data }),
+        } as unknown) as DataProvider;
+
+        const validateAgainstServer = value =>
+            new Promise(resolve => setTimeout(() => value, 400));
```
fzaninotto, line 351: this promise never resolves

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -337,4 +337,52 @@ describe('<SaveButton />', () => {
...
+        await waitFor(() => {
+            expect(queryByDisplayValue('lorem')).toBeDefined();
+        });
+
+        // change one input to enable the SaveButton (which is disabled when the form is pristine)
+        fireEvent.change(getByLabelText('resources.posts.fields.title'), {
+            target: { value: 'ipsum' },
+        });
+
+        expect(getByLabelText('ra.action.save')['disabled']).toEqual(true);
```
fzaninotto, line 386: I'm not sure you're testing the right thing. 

The SaveButton will be disabled at start because no value was changed. Then, it will be disabled because the validation is ongoing. Finally, when the validation has finished, it should either be enabled (if the validation is ok) or disabled (if the validation has failed).

Your test should reflect that and test all cases.

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -337,4 +337,52 @@ describe('<SaveButton />', () => {
...
         });
     });
+
+    it('should disable <SaveButton/> if an input is being validated asynchronously', async () => {
+        const dataProvider = ({
+            getOne: () =>
+                Promise.resolve({
+                    data: { id: 123, title: 'lorem' },
+                }),
+            update: (_, { data }) => Promise.resolve({ data }),
```
fzaninotto, line 347: I don't think you need that for your test to work.

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -337,4 +337,52 @@ describe('<SaveButton />', () => {
...
+        } as unknown) as DataProvider;
+
+        const validateAgainstServer = value =>
+            new Promise(resolve => setTimeout(() => value, 400));
+
+        const validateAsync = async (value, allValues) => {
```
fzaninotto, line 353: I don't understand why you need 2 separate functions (validateAgainstServer and validateAsync). It make sthe code harder to read.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
WiXSL: Actually during async validation the button should be disabled, after that should be enabled again, unless the form is pristine

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6292, opened 2021-05-20
add optional labelFn prop to SimpleFormIterator

iamstiil:


djhi (commented):
Can you please describe the usecase?

iamstiil:
I created a custom input, with nested iterators. I would like to update the label in the nested iterators to `1.1`, `1.2`, etc. so that I can differentiate between `1.3` and `2.3` for example.

djhi (changes requested):
Can you please add tests and update the documentation?

iamstiil:
Sure 👍

PR #6314, opened 2021-05-27
Improve documentation main container's width

### LINKED ISSUE #6313: Improve documentation container's width
Documentation central container is visually too narrow, specially for props tables.
By adding `max-width: fit-content` to `main > .container` class inside `docs/css/style-v11.css` you can improve this.

Before:
before

After:
after

WiXSL:
Fixes #6313


fzaninotto:
I don't think this is a good idea. this makes lines way too long on large screens, to the point that they are difficult to read. 

image

PR #6321, opened 2021-06-01
Introduce useList

djhi:
Introduces a new hook that applies list filtering, sorting and pagination on the provided data, either in memory or through the provided function. This can be useful for example when you can't change some API routes which return enough data to make those features necessary but you can't or don't want to implement them in your dataProvider.

### DIFF
--- a/packages/ra-core/src/controller/field/ReferenceArrayFieldController.spec.tsx
+++ b/packages/ra-core/src/controller/field/ReferenceArrayFieldController.spec.tsx
@@ -200,7 +200,7 @@ describe('<ReferenceArrayFieldController />', () => {
         });
     });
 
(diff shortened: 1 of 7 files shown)


fzaninotto (approved):

fzaninotto:
I merged it a bit too fast. Could you open a new PR with the following adjustments:

- Remove the ids and data params - they don't make sens in that context
- remove the 'initial' prefix in all arguments. 
- update JSDoc accordingly

This should allow a simpler call, like:

```jsx
 const listContext = useList({
    data,
    ids,
    filter: { is_published: truev},
    sort: { field: 'title', order: 'ASC' },
    page: 1,
    perPage: 10,
    basePath: '/resource';
    resource: 'resource';
 });

Issue #5882, opened 2021-02-07
Allow auth providers to handle authorization errors without logging out


josepedrocorreia:
Hi there,

We started using our own custom implementation of an `AuthProvider` to handle authentication and authorization in our app. A problem I am facing is related to the following scenario.

We want to implement diferentiated access where users can have different access to different resources. Say user A has access to resource "One" but not to resource "Two". Our backend would reply with a 403 if user A sent for example a list request for resource "Two". I would like to catch these errors and redirect the user to an "Unauthorized" page within the app, but allow the user to continue using it normally. They could, for example, navigate to the list for resource "One".

The problem I have is that, in the implementation of the `checkError` function, I have two options: either return `Promise.resolve()` that send the user to the page they requested, or return a `Promise.reject()` which optionally can redirect the user to another page _after_ logging him out. Neither solution serves my purpose, since if I `resolve` the user gets to the a page that fails to render because the backend returned a 403, and if I `reject` the user gets logged out and cannot continue using the app normally without logging in again.

Two suggestions:
1. Adding an optional field `logoutUser?: boolean` to the error format of the `checkError` function that controls whether the user gets logged out or not (default to `true` for backwards compatibility)
2. Allowing `checkError` to return `{ redirectTo?: string, message?: string }`, where these values would be interpreted in the same manner as for the error fomat, i.e. `Promise.resolve({ redirectTo: "/unauthorized", message: false })` would redirect the user to the `/unauthorized` page and omit the error notification message

What do you guys think?

fzaninotto:
I'm OK for the second suggestion. Would you mind opening a PR to implement it?

josepedrocorreia:
Hi @fzaninotto, I would love to contribute, but I'm not familiar with the code and unfortunately I don't have much time, sorry :-/

andrico1234:
I think this would be really useful, I'd be happy to pick this up if someone hasn't already @fzaninotto

fzaninotto:
Fixed by #6326

PR #6326, opened 2021-06-02
Feature: Offer logout option in auth provider

andrico1234:
Closes: https://github.com/marmelab/react-admin/issues/5882

Handles additional options to the `checkError`'s `Promise.reject`.

https://user-images.githubusercontent.com/17087167/120549111-05fb0580-c3eb-11eb-944b-9f150da6fa0d.mov

##  Notes: 

- Do we want to rename the `useLogoutIfAccessDenied`, since it can now sometimes not log the user out if access is denied.
- Should we move the redirect building logic out to a reusable function, since it’s very similar to the one used in `useLogout`. 
- Do we also want to add similar logic to the checkAuth function?


packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -67,11 +79,12 @@ describe('useLogoutIfAccessDenied', () => {
     });
 
     it('should not logout if passed no error', async () => {
-        const { queryByText } = render(
+        const { queryByText } = renderComponent(
```
andrico1234, line 82: We can either mock the `useHistory`, or provide a Router. Doing neither results in this: 
image

I chose to wrap them all in a router because

a) I prefer to avoid mocking if I don't need to
b) we'd have to wrap the new test in a router regardless, and so we'd be wrangling with both sets logic in the suite

andrico1234:
@djhi @fzaninotto hey folks! wondering if there's any update for this ticket? It's good for a review

docs/Authentication.md
```diff
@@ -186,6 +186,26 @@ export default {
...
+
+
+```js
+// in src/authProvider.js
+export default {
+    login: ({ username, password }) => { /* ... */ },
+    checkError: (error) => {
+        const status = error.status;
+        if (status === 401 || status === 403) {
+            localStorage.removeItem('auth');
```
djhi, line 199: I think you can remove this line

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -76,11 +81,33 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
...
+                            newLocation.state = {
+                                nextPathname: history.location.pathname,
+                                nextSearch: history.location.search,
+                            };
+                        }
+
+                        if (redirectToParts[1]) {
+                            newLocation.search = redirectToParts[1];
+                        }
+                        history.push(redirectTo);
```
djhi, line 103: We should probably extract this from here and from `useLogout`. @fzaninotto ?

djhi (changes requested):
Thanks!

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
andrico1234: I thought the same too, but wasn't sure where the right place for it it is. What do you both think?

docs/Authentication.md
andrico1234: removed

andrico1234:
I've resolved the requested changes, @djhi

docs/Authentication.md
```diff
@@ -491,40 +510,40 @@ const authProvider = {
...
 
-| Method           | Usage                                           | Parameters format  |
-| ---------------- | ----------------------------------------------- | ------------------ |
-| `login`          | Log a user in                                   | `Object` whatever fields the login form contains |
-| `checkError`     | Check if a dataProvider error is an authentication error  | `{ message: string, status: number, body: Object }` the error returned by the `dataProvider` |
-| `checkAuth`      | Check credentials before moving to a new route  | `Object` whatever params passed to `useCheckAuth()` - empty for react-admin default routes |
-| `logout`         | Log a user out                                  |                    |
-| `getIdentity`    | Get the current user identity                   |                    | 
-| `getPermissions` | Get the current user credentials                | `Object` whatever params passed to `usePermissions()` - empty for react-admin default routes |
+| Method           | Usage                                                    | Parameters format                                                                            |
```
fzaninotto, line 513: please revert these table formatting changes. They make the doc harder to read when not translated into Markdown, and they don't bring any value.

docs/Authentication.md
```diff
@@ -491,40 +510,40 @@ const authProvider = {
...
 `authProvider` methods must return a Promise. In case of success, the Promise should resolve to the following value:
 
-| Method           | Resolve if                        | Response format |
-| ---------------- | --------------------------------- | --------------- |
-| `login`          | Login credentials were accepted   | `void`          |
-| `checkError`     | Error is not an auth error        | `void`          |
-| `checkAuth`      | User is authenticated             | `void`          |
-| `logout`         | Auth backend acknowledged logout  | `string | false | void` route to redirect to after logout, defaults to `/login` |
-| `getIdentity`    | Auth backend returned identity    | `{ id: string | number, fullName?: string, avatar?: string }`  | 
+| Method           | Resolve if                        | Response format                                                                                |
```
fzaninotto, line 526: same

docs/Authentication.md
```diff
@@ -491,40 +510,40 @@ const authProvider = {
...
 | `getPermissions` | Auth backend returned permissions | `Object | Array` free format - the response will be returned when `usePermissions()` is called |
 
 ### Error Format
 
 When the auth backend returns an error, the Auth Provider should return a rejected Promise, with the following value: 
 
-| Method           | Reject if                                 | Error format |
-| ---------------- | ----------------------------------------- | --------------- |
-| `login`          | Login credentials weren't accepted        | `string | { message?: string }` error message to display |
+| Method           | Reject if                                 | Error format                                                                                                                     |
```
fzaninotto, line 539: same

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -58,6 +60,16 @@ const notify = jest.fn();
 //@ts-expect-error
 useNotify.mockImplementation(() => notify);
 
+function renderComponent(children) {
```
fzaninotto, line 63: please rename to `renderInRouter`

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
fzaninotto: Shouldn't this be `history.push(newLocation)`? Otherwise the code above is useless.

Also, setting `nextPathName` and `nextSearch` is currently only useful for the Login page. I don't see any point in having them here. And since we don't need that, we don't need to manipulate the `redirectTo`.

fzaninotto (changes requested):
Nice!

docs/Authentication.md
andrico1234: I think it's related to my prettier config, this happens every time! I'll revert

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
andrico1234: It's been a few weeks since I've gone over this code, and I'll sanity check just in case, but I believe the `history.push(redirectTo)` is the correct behaviour. In essence, we want to perform the same behaviour but as the `logout({}, redirectTo)`but to circumvent the logout functionality.

andrico1234: @fzaninotto , does the video in the PR's description clarify things?

andrico1234: ignore me, you're completely right!

andrico1234:
@djhi @fzaninotto 

I'd like to get a Cypress test added for this behaviour, but can't figure out an appropriate way to do it.

Is there a way to monkey patch in a slightly modified authProvider? If not, I could:
- create a really basic RA-app, but that seems like overkill.
- create a new resource that specifically logs the user when clicking through to it (but we'd still need to alter the authProvider)

I added a unit test, but it didn't catch the problem when I called `history.push(redirectTo)` instead of `history.push(newLocation)`.

What do you guys think?

Other than that, i've resolved the rest of the comments

andrico1234:
Note: I added a notification if a user was redirected but not logged out. I added a french translation that was made via google translate, since I don't speak French! Please let me know what you'd like the message to be 😄

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -76,11 +91,20 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
                             : error && error.redirectTo
                             ? error.redirectTo
                             : undefined;
-                    logout({}, redirectTo);
+
+                    if (logoutUser) {
+                        logout({}, redirectTo);
+                    } else {
+                        const newLocation: LocationDescriptorObject = {
```
fzaninotto, line 98: Why do you create a new object? This forbids developers to pass a location object themselves. 

I'd pass the `redirectTo` directly, like so:

```
history.push(redirectTo);
```

packages/ra-language-french/src/index.ts
```diff
@@ -143,6 +143,8 @@ const frenchMessages: TranslationMessages = {
                 'Erreur de chargement des traductions pour la langue sélectionnée',
             canceled: 'Action annulée',
             logged_out: 'Votre session a pris fin, veuillez vous reconnecter.',
+            not_authorized:
+                "Vous n'êtes pas autorisé à accéder à cette ressource.",
```
fzaninotto, line 147:
```suggestion
                "Vous n'êtes pas autorisé(e) à accéder à cette ressource.",
```

fzaninotto (changes requested):
Awesome!

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
andrico1234: How do you recommend we do this then?

Specify the `redirectTo` object be a type of `LocationDescriptorObject`, and invert control up to the authProvider?

andrico1234: Ignore the above comment, i had some issue with testing previously. I've pushed your suggestions and it works fine. 

I've tested using a regular path like `/unauthorized` and also adding query params. Are there any other manual test cases you recommend I try?

fzaninotto (approved):

fzaninotto:
Thanks a lot!

Issue #6312, opened 2021-05-27
ra-data-graphql: Upgrade to graphql@15


ad-si:
**Is your feature request related to a problem? Please describe.**

GraphQL 15 has been available for quite some time, and I'm getting problems with peer dependencies which need this version.

fzaninotto:
Nice idea, feel free to open a PR for that.

PR #6327, opened 2021-06-03
[RFR] chore(apollo): migrate to apollo 3

Oupsla:
Hello folks 👋 

Apollo packages are now deprecated and everything has been centralized to single package on npm (@apollo/client)

Little note : IntrospectionFragmentMatcher has been removed totally ([link](https://github.com/apollographql/apollo-client/issues/5423#issuecomment-828524753)) , so I have just removed it (tested with my project without errors, so I think the issue linked in the code is gone with apollo3)

### DIFF
--- a/examples/demo/src/dataProvider/graphql.ts
+++ b/examples/demo/src/dataProvider/graphql.ts
@@ -1,4 +1,4 @@
-import { ApolloQueryResult } from 'apollo-client';
+import { ApolloQueryResult } from '@apollo/client';
 import buildApolloClient, {
     buildQuery as buildQueryFactory,
 } from 'ra-data-graphql-simple';
--- a/packages/ra-data-graphql/package.json
+++ b/packages/ra-data-graphql/package.json
@@ -32,8 +32,7 @@
         "watch": "tsc --outDir esm --module es2015 --watch"
     },
     "dependencies": {
-        "apollo-client": "^2.6.3",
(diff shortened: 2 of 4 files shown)


djhi:
Thanks! We'll have to migrate the `json-graphql-server` package as well though

Oupsla:
> Thanks! We'll have to migrate the `json-graphql-server` package as well though

Do I need to do something else in this PR ?

djhi:
I don't think so but we won't merge before being able to test it and that means we'll have to migrate `json-graphql-server` first.

julien1619:
Hello! I've started working on the same thing a week ago after seeing #6312 but didn't see your PR until today.

You can remove `graphql-tag` too and use `import { gql } from "@apollo/client"` instead as I did here: https://github.com/marmelab/react-admin/pull/6347/files

Like I said in the other PR, before being merged it would need an upgrade of ra-data-graphql-simple.
I tried to upgrade it too, but it mainly depends on a dead fork of a dead project (last commit 2018), depending on an outdated version of graphql (0.11).

julien1619:
Is there anything I can do to help here?

larsivi:
I am starting to have other dependencies that also rely on Apollo 3, supporting both in the application is not easy. I am also willing to help here if necessary to expedite this issue.

Oupsla:
Coming back from vacation, will have a look at it asap

PR #6329, opened 2021-06-03
add TS generic support for usePermissions hook

andrei9669:
there are probably other hooks as well that could use this modification but this is the one that bugs me the most in my project.
I would have made the default value "unknown", to make it more strict, but that would have probably broken a lot of code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,3 @@ examples/**/static
 examples/**/dist
 cypress/videos
 cypress/screenshots
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ examples/**/static
 examples/**/dist
 cypress/videos
 cypress/screenshots
+.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ examples/**/static
 examples/**/dist
 cypress/videos
 cypress/screenshots
+.idea
```
djhi, line 16: Please revert this

djhi (changes requested):
Thanks!

.gitignore
andrei9669: I have read in some older PRs that this should be in user gitignore, but why is that so? Shouldn't the repository be responsible for what is ignored and what isn't?

djhi: You can have a root `.gitignore` in your home directory for things specific to your tools.

djhi (approved):
Thanks!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6331, opened 2021-06-03
feat: allow classes overriding on CreateButton

etienne-bondot:
As a developer we want to be able to override the `CreateButton` style. It seems that the `classesOverride` is not properly applied to the `Button` component.


packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -62,6 +62,7 @@ const CreateButton: FC<CreateButtonProps> = props => {
         <Button
             component={Link}
             to={location}
+            classes={classesOverride}
```
djhi, line 65: You should pass `classes` here as it actually merge the `classesOverride` and the default classes. This is why we pass the props (and not `rest`) to the `useStyles` hook

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/button/CreateButton.tsx
etienne-bondot: You right! Sorry I will update it.

djhi (approved):
:+1: Thanks!

djhi (changes requested):
Sorry, this is actually not enough. We must find a way to extend the original button classes. Besides, the `floating` class should not be passed to the `Button`, only to the `Fab` button.

fzaninotto:

PR #6345, opened 2021-06-09
Export ui default state for custom apps to override some values of it

smeng9:
When I want to make the sideBar default to closed by setting the `admin.ui` initialState of `<Admin/>`, some other properties are also get dropped in the redux.
e.g. initialState = { admin: { ui: { sideBarOpen: false } } } // missing the default automaticRefreshEnabled, etc...

It would be more convenient to export the default state as an object so I can override some of the values before passing it into the `<Admin/>` component.

### DIFF
--- a/packages/ra-core/src/index.ts
+++ b/packages/ra-core/src/index.ts
@@ -30,5 +30,7 @@ export {
     nameRelatedTo,
 } from './reducer/admin/references/oneToMany';
 
+export { uiDefaultState } from './reducer/admin/ui';
+
 export * from './sideEffect';
 export * from './types';
--- a/packages/ra-core/src/reducer/admin/ui.ts
+++ b/packages/ra-core/src/reducer/admin/ui.ts
@@ -42,7 +42,7 @@ const isDesktop = (): boolean =>
         ? window.matchMedia('(min-width:960px)').matches
         : false;
 
-const defaultState: UIState = {
+export const uiDefaultState: UIState = {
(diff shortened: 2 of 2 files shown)


fzaninotto:
Thanks for your PR. I'm uncomfortable with it for 2 reasons:

1. It has no documentation, so it'll probably never be used
2. It solves a very specific problem with a very specific solution. I believe the problem is wider (overriding the initial state) and probably deserves a more generic solution.
3. It has no test. Yes, I know, I said there were 2 reasons.

smeng9:
@fzaninotto Thanks for the feedback, I agree this problem is too specific and your suggestion makes sense. May I propose the following alternative solutions to address your concern?

Instead of exposing the default state, we can use the `lodash.merge` function to merge the user-passed initial override state with the default state. We will need to check whether the `action.type` is `@@INIT` in each reducer and perform the merge. This will make the solution general.

We also need to add the documentation https://marmelab.com/react-admin/Admin.html#initialstate to document user can partially overwrite some of the redux states.

If this is good I can open a different pull request to replace this one. Thanks.

smeng9:
After some search reacting to @@INIT is anti-pattern, and @@INIT may be concatenated with random strings. See: https://stackoverflow.com/questions/41305492/what-is-the-purpose-of-the-init-action-in-react-redux

So a better solution may be returning { ...defaultState, ...previousState } so we can populate undefined fields in the initialState

Issue #6350, opened 2021-06-10
title type mismatch between ListProps and EditProps, CreateProps, ShowProps


megantaylor:
seems like in https://github.com/marmelab/react-admin/pull/6119, `false` was added as a possible type for `title` in ListProps. 

For such cases, we should add a support for `false` value in the `<Title>` component, which wouldn't create a portal in that case

_Originally posted by @fzaninotto in https://github.com/marmelab/react-admin/pull/6119#discussion_r620960154_

but it was not added for EditProps, CreateProps, ShowProps, which causes a type mismatch error. 

Can `false` be added a possible type for `title` for those components?

djhi:
Make sense. Thanks for bringing that up. Can you open a PR?

PR #6351, opened 2021-06-10
[TypeScript] Fix type mismatch error on title prop for page components

megantaylor:
closes: https://github.com/marmelab/react-admin/issues/6350

adds `false` to `title` types for Create, Edit and Show components to match `title` type on List component (was added here: https://github.com/marmelab/react-admin/pull/6119)


djhi (approved):
Thanks!

PR #6354, opened 2021-06-13
Add should to ra-test docs test case's explanation

ValentinnDimitroff:
Just added a 'should' to the case's explanation for consistency and better reading


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6355, opened 2021-06-13
identity.fullName text node

omerschapiro:
I tried changing the style of identity.fullName in the UserMenu and found it very difficult, since it's just a text node and not an element.
I ended up having to create a custom UserMenu, which is basically identical to the original one.
While this solution works, I think it's a shame. A much easier fix exists - to wrap the existing identity.fullName in a tag.
In this change I used a span for convenience and simplicity, however it could also be MaterialUI's Typography tag which could accept typography props for even further customization.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/UserMenu.tsx
+++ b/packages/ra-ui-materialui/src/layout/UserMenu.tsx
@@ -80,7 +80,7 @@ const UserMenu = (props: UserMenuProps) => {
                     }
                     onClick={handleMenu}
                 >
-                    {identity.fullName}
+                    <span>{identity.fullName}</span>
                 </Button>
             ) : (
                 <Tooltip title={label && translate(label, { _: label })}>


fzaninotto:
Why not add a fixed className to the enclosing button instead?

omerschapiro:
I'm not sure how that would help, it's still a text node.

omerschapiro:
any new thoughts on this?

djhi:
Replace the default UserMenu with your own ?

omerschapiro:
that's what I did, but it was just to add one span and I think it's a shame

fzaninotto:
I don't think we should add extra markup to accommodate your use case, which is not very common IMHO.  

So the default `UserMenu` doesn't match your use case, but you can replace it by your own component. It may be a bit cumbersome, but it's consistent with our design guidelines (https://marmelab.com/react-admin/Architecture.html#component-composition).

Thanks for your PR, but we'll pass on this one.

PR #6356, opened 2021-06-14
[Doc] Add Architecture decisions

fzaninotto:
## Problem

The react-admin codebase has evolved over the yeras, and sometimes it's hard to understand why a feature was designed that way.

While we try to have explicit Pull Request descriptions, it doesn't give the Big Picture about the code organization

## Solution

Add a new Architecture section in the documentation.

### DIFF
--- a/docs/Architecture.md
+++ b/docs/Architecture.md
@@ -67,7 +67,7 @@ export const PostEdit = (props) => (
 
 This allows overriding only part of the logic of a component by composing it with another component.
 
-So many react-admin components can be overridden by passing custom components as children or via props.
+Many react-admin components can be overridden by passing custom components as children or via props.
 
 The drawback is that react-admin sometimes forces you to override several components just to enable one feature. For instance, to override the Menu, you must pass it to a custom `<Layout>`, and pass the custom `<Layout>` to `<Admin>`:
 
--- a/docs/Architecture.md
+++ b/docs/Architecture.md
@@ -1,0 +1,276 @@
+---
+layout: default
+title: "Architecture"
+---
+
+# Architecture
+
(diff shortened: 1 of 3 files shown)


docs/Architecture.md
```diff
@@ -0,0 +1,276 @@
...
+export const PostEdit = (props) => (
+    <Edit actions={<PostEditActions />} {...props}>
+        ...
+    </Edit>
+);
+```
+
+This allows overriding only part of the logic of a component by composing it with another component.
+
+So many react-admin components can be overridden by passing custom components as children or via props.
```
djhi, line 70:
```suggestion
Many react-admin components can be overridden by passing custom components as children or via props.
```

docs/Readme.md
```diff
@@ -107,13 +116,18 @@ React-admin uses an adapter approach, with a concept called *Data Providers*. Ex
 
 See the [Data Providers documentation](./DataProviders.md) for details.
 
-## Batteries Included But Removable
+## Architecture: Batteries Included But Removable
+
+React-admin is designed as a library of loosely coupled React components built on top of [material-ui](https://material-ui.com/), in addition to React hooks allowing to reuse the logic with a custom UI. 
+
+It is very easy to replace one part of react-admin with your own, e.g. to use a custom Datagrid, GraphQL instead of REST, or Bootstrap instead of Material Design.
```
djhi, line 123:
```suggestion
You may replace one part of react-admin with your own, e.g. to use a custom Datagrid, GraphQL instead of REST, or Bootstrap instead of Material Design.
```

Let's avoid simple, easy, etc.

djhi (changes requested):

djhi:
Nice!

PR #6361, opened 2021-06-15
Fixed number input label issue

### LINKED ISSUE #6352: Many of inputs don't support label={false}
**What you were expecting:**
NumberInput, AutoCompleteInput and all inputs support label={false} same as TextInput to hide label

**What happened instead:**
NumberInput and many input cause error when label prop is false

**Steps to reproduce:**
Try <AutocompleteArrayInput label={false} />

* React-admin version: 3.14.3

VikrantShirvankar:
Closes #6352

Fixed label issue for Number input.

**Before** - with `label={false}`

Screenshot 2021-06-15 at 11 27 48 PM


packages/ra-ui-materialui/src/input/NumberInput.tsx
```diff
@@ -79,12 +79,15 @@ const NumberInput: FunctionComponent<NumberInputProps> = ({
...
             label={
-                <FieldTitle
-                    label={label}
-                    source={source}
-                    resource={resource}
-                    isRequired={isRequired}
-                />
+                label !== '' &&
+                label !== false && (
+                    <FieldTitle
```
djhi, line 84: It would probably be better to handle this in `FieldTitle` directly. Then we would only have to update the propTypes in all the other components

djhi (changes requested):

djhi:
Thanks!

packages/ra-ui-materialui/src/input/NumberInput.tsx
VikrantShirvankar: Handled label issue directly inside `FieldTitle`. Updated propType for NumberInput field.
I will raise separate PR to update label propType in other components. Will it be fine?
Let me know if any  changes required.
Thanks..!

packages/ra-core/src/util/FieldTitle.tsx
```diff
@@ -18,9 +18,15 @@ export const FieldTitle: FunctionComponent<Props> = ({
     isRequired,
 }) => {
     const translate = useTranslate();
+
+    if (label === false && label === '') {
```
fzaninotto, line 22: this doesn't make sense

packages/ra-core/src/util/FieldTitle.spec.tsx
```diff
@@ -82,4 +82,15 @@ describe('FieldTitle', () => {
         const { container } = render(<FieldTitle label="foo" isRequired />);
         expect(container.firstChild.textContent).toEqual('foo *');
     });
+
+    it('should return null if label is false', () => {
+        const { container } = render(<FieldTitle label={false} isRequired />);
+        expect(container.firstChild).toBeNull();
+    });
+
+    it('should return null if label is empty string', () => {
```
fzaninotto, line 91: This isn't the desired behavior. An empty string should lead to an empty label (not the absence of label)

fzaninotto (changes requested):
Thanks for your PR.

You should also refactor TextInput, SelectInput and RichTextInput to remove the test on `label !== false`, which is now redundant.

You should document that feature in the Input documentation. 

Finally, this is technically a new feature, and as such it should be PRed against `next` rather than `master`.

packages/ra-ui-materialui/src/input/NumberInput.tsx
fzaninotto: Please update the propTypes and TypeScript types in this PR

packages/ra-core/src/util/FieldTitle.tsx
VikrantShirvankar: Ok. So the desired behavior is to return null only if label is false. Right? Correct me if I am wrong.

VikrantShirvankar: I just followed below existing code - 
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/TextInput.tsx#L65
As per this code,  `FieldTitle` will not get rendered if string is empty.

fzaninotto: my bad, you're right. Both `''` and `false` should disable the title.

VikrantShirvankar:
- Raised separate PR against `next` branch.
- Updated Input documentation
Please refer this PR - https://github.com/marmelab/react-admin/pull/6381

PR #6367, opened 2021-06-17
Add links to the Error page to help resolution

fzaninotto:
## Before

image

## After

image

The links only appear in development, just like the error and trace. Besides, the Error page is still overridable.


djhi (approved):
:+1:

PR #6368, opened 2021-06-17
Add ability to define filters as an array of Inputs

fzaninotto:
## Problem

`<ListActions>` is hard to override, partly because of [the filter button/form combo](https://marmelab.com/react-admin/List.html#the-filter-buttonform-combo). The `<Filter>` element renders twice, which isn't a common practice, and creates a few problems.

Refs #6212

## Solution

Allow `<List filters>` to be an array of inputs instead of a React element. 

```diff
-const PostFilter = props => (
-    <Filter {...props}>
-        <SearchInput source="q" alwaysOn />
-        <TextInput
-            source="title"
-            defaultValue="Qui tempore rerum et voluptates"
-        />
-        <QuickFilter
-            label="resources.posts.fields.commentable"
-            source="commentable"
-            defaultValue
-        />
-    </Filter>
-);
+const postFilters = [
+    <SearchInput source="q" alwaysOn />,
+    <TextInput source="title" defaultValue="Qui tempore rerum et voluptates" />,
+    <QuickFilter
+        label="resources.posts.fields.commentable"
+        source="commentable"
+        defaultValue
+    />,
+];

const PostList = props => {
         <List
             {...props}
             bulkActionButtons={<PostListBulkActions />}
-            filters={<PostFilter />}
+            filters={postFilters}
             sort={{ field: 'published_at', order: 'DESC' }}
             exporter={exporter}
         >
```

This allows to put the filters in a context, and to use `<FilterForm>` and `<FilterButton>` directly:

```jsx
const  ListActions = () => (
    <TopToolbar>
        <FilterButton />
        <CreateButton />
        <ExportButton />
    </TopToolbar>
);
```

- [x] Add support for array filters in `<List>`
- [x] Write tests
- [x] Update documentation

The change is backwards compatible: `<Filter>` continues to exist and still works as before.


fzaninotto:
Switching to RFR

docs/Inputs.md
```diff
@@ -5,7 +5,7 @@ title: "Input Components"
 
 # Input Components
 
-An `Input` component displays an input, or a dropdown list, a list of radio buttons, etc. Such components allow to edit a record property, and are common in the `<Edit>`, `<Create>`, and `<Filter>` views.
+An `Input` component displays an input, or a dropdown list, a list of radio buttons, etc. Such components allow to edit a record property, and are common in the `<Edit>`, `<Create>` component, and in the List Filters.
```
djhi, line 8:
```suggestion
An `Input` component displays an input, or a dropdown list, a list of radio buttons, etc. Such components allow to edit a record property, and are common in the `<Edit>` and `<Create>` components, and in the List Filters.
```

djhi (changes requested):
Awesome! I remember having eslint issues about missing keys with the array syntax for the `<Resource>` though.

fzaninotto:
@djhi I have had no eslint warning, and GitHub doesn't show any... Would you have a reproducible case?

djhi:
Nope, it was just something I remembered while working with the `<Resource>` in a render function. All good then!

djhi (approved):

PR #6375, opened 2021-06-21
Fix useInput resetting NumberInput when value = 0

andion:
This issue was making every NumberInput change its value to its defaultValue or initialValue when it reaches zero. 

If you try to do this: `<NumberInput source="test" min={-10} max={10} defaultValue={6} />`

It will go back to 6 when it reaches 0, like this

l6tTZFrzAg

I don't know if there are more side-effects not considered, but this fixes the said issue.

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -115,17 +115,17 @@ const useInput = ({
     const form = useForm();
     const recordId = record?.id;
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


djhi (commented):
Thanks but I actually opened a PR at the same time, with tests (#6374)

andion:
Great! Thank you

andion:
Already fixed in https://github.com/marmelab/react-admin/pull/6374

PR #6378, opened 2021-06-22
Apply review on useList

djhi:
Applied the the following adjustments:

- Remove the ids and data params - they don't make sens in that context
- remove the 'initial' prefix in all arguments. 
- update JSDoc accordingly

This should allow a simpler call, like:

```jsx
 const listContext = useList({
    data,
    ids,
    filter: { is_published: truev},
    sort: { field: 'title', order: 'ASC' },
    page: 1,
    perPage: 10,
    basePath: '/resource';
    resource: 'resource';
 });

### DIFF
--- a/packages/ra-core/src/controller/useList.ts
+++ b/packages/ra-core/src/controller/useList.ts
@@ -75,13 +75,7 @@ export const useList = (props: UseListOptions): UseListValue => {
         ids: Identifier[];
     }>(() => {
         return {
-            data: ids.reduce((acc, id) => {
-                const record = data.find(r => !!r && r.id === id);
-                if (record) {
-                    acc[id] = record;
-                }
-                return acc;
-            }, {}),
+            data: indexById(data),
             ids,
         };
     });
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/ra-core/src/controller/useList.ts [hunk 1]
```diff
@@ -38,42 +44,47 @@ import { ListControllerProps } from '.';
...
     const [finalItems, setFinalItems] = useSafeSetState<{
         data: RecordMap;
         ids: Identifier[];
-    }>(() => ({
-        data: indexById(initialData),
-        ids: initialIds,
-    }));
+    }>(() => {
+        return {
+            data: ids.reduce((acc, id) => {
```
fzaninotto, line 78: why not use indexById here? It's exactly what it does.

packages/ra-core/src/controller/useList.ts
```diff
@@ -156,13 +167,9 @@ export const useList = (props: UseListOptions): UseListValue => {
...
     useEffect(() => {
         if (!loaded) return;
-        // Assume that if setFilters is provided then so are methods for pagination and sorting
-        if (props.setFilters) {
-            return;
-        }
 
         // 1. filter
-        let tempData = initialData.filter(record =>
+        let tempData = Object.values(data).filter(record =>
```
fzaninotto, line 172: I don't understand why you need Object.values, since `data` is already an array

fzaninotto (changes requested):

fzaninotto (approved):

PR #6387, opened 2021-06-23
[Doc] Fix wrong URL to the CRM app entrypoint

Cornul11:
As the title says.


fzaninotto:
Nice, but I prefer to switch the App source file to tsx instead. I'll open a PR for that. 

thanks for pointing the problem anyway.

PR #6388, opened 2021-06-23
[Doc] Improve the side effect hooks documentation

fzaninotto:
image

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -627,7 +627,7 @@ Fetching data is called a *side effect*, since it calls the outside world, and i
 
 ### `useNotify`
 
-This hook returns a function that displays a notification in the bottom banner of the page.
+This hook returns a function that displays a notification in the bottom of the page.
 
 ```jsx
 import { useNotify } from 'react-admin';
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -620,10 +620,152 @@ const ApproveButton = ({ record }) => {
 
 Fetching data is called a *side effect*, since it calls the outside world, and is asynchronous. Usual actions may have other side effects, like showing a notification, or redirecting the user to another page. React-admin provides the following hooks to handle most common side effects:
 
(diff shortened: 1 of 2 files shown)


docs/Actions.md
```diff
@@ -620,10 +620,152 @@ const ApproveButton = ({ record }) => {
...
-- `useRefresh`: Return a function to force a rerender of the current view (equivalent to pressing the Refresh button).
-- `useUnselectAll`: Return a function to unselect all lines in the current `Datagrid`. Pass the name of the resource as argument.
+- [`useNotify`](#usenotify): Return a function to display a notification. 
+- [`useRedirect`](#useredirect): Return a function to redirect the user to another page. 
+- [`useRefresh`](#userefresh): Return a function to force a rerender of the current view (equivalent to pressing the Refresh button).
+- [`useUnselectAll`](#useunselectall): Return a function to unselect all lines in the current `Datagrid`. 
+
+### `useNotify`
+
+This hook returns a function that displays a notification in the bottom banner of the page.
```
djhi, line 630: bottom banner ?

docs/Actions.md
```diff
@@ -620,10 +620,152 @@ const ApproveButton = ({ record }) => {
...
+
+```jsx
+import * as React from 'react';
+import { useNotify, Edit, SimpleForm } from 'react-admin';
+
+const PostEdit = props => {
+    const notify = useNotify();
+
+    const onSuccess = () => {
+        notify(`Changes saved`, undefined, undefined, true);
```
djhi, line 672: We should probably ask for a single option object instead of 5 args imo. Maybe we can support both until we deprecate the 5 args signature?

djhi (changes requested):
:+1:

docs/Actions.md
fzaninotto: Nice idea, but let's do it in another PR

djhi (approved):

PR #6395, opened 2021-06-26
Added file Source for the snippet in documentation Authentication section

neps-in:
Added file Source for the snippet in documentation Authentication section. A small contrib.


fzaninotto:
Thanks, but in my opinion this PR doesn't solve a problem or add any value. I'll pass.

neps-in (commented):
I guess you did not see this commit 8422ff4ceb7063ee2fac601d760b4dbf59b85add
The file name was not there, I have added the filename, Its very confusing where the code should go. Hence added teh file name.

PR #6401, opened 2021-06-30
Fix DateTimeInput does not include timezone

### LINKED ISSUE #6400: DateTimeInput do not set timezone on initialValue
**What you were expecting:**

When using a DateTimeInput with initialValue, I would like the value to be time zoned when I submit the form.
```json
"published_at":"2021-06-29T09:55:00.000Z"
```

**What happened instead:**

When validating the form without changing the default value I have : 
```json
"published_at":"2021-06-29T11:29",
```

**Steps to reproduce:**

Add a DateTimeInput in a simpleForm with a default value. Submit the form without changing the default value.

**Related code:**

* https://codesandbox.io/s/dank-sound-8u2we?file=/src/posts/PostCreate.js:2565-2646

```js
<DateTimeInput source="published_at" defaultValue={dateDefaultValue} />
```

**Other information:**

**Environment**

* React-admin version: 3.16.4
* React version: 16.9.0
* Browser: Chromium

djhi:
Fixes #6400

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
+++ b/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
@@ -59,7 +59,7 @@ const formatDateTime = (value: string | Date) => {
  * @param {string} value Date string, formatted as yyyy-MM-ddThh:mm
  * @return {Date}
  */
-const parseDateTime = (value: string) => new Date(value);
+const parseDateTime = (value: string) => new Date(value).toISOString();
 
 /**
  * Input component for entering a date and a time with timezone, using the browser locale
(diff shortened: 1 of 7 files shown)


packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -59,7 +59,7 @@ const formatDateTime = (value: string | Date) => {
  * @param {string} value Date string, formatted as yyyy-MM-ddThh:mm
  * @return {Date}
  */
-const parseDateTime = (value: string) => new Date(value);
+const parseDateTime = (value: string) => new Date(value).toISOString();
```
fzaninotto, line 62: This doesn't match the jsDoc

fzaninotto (changes requested):
Tests fail

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
djhi: And it's wrong anyway

fzaninotto (changes requested):

PR #6415, opened 2021-07-06
Allow overriding label for input inside TranslatableInputs

### LINKED ISSUE #6414: Input label ignored inside TranslatableInputs
**What you were expecting:**
Custom label text (via the `label` prop) should appear on an input inside TranslatableInputs

**What happened instead:**
The default label text (derived from the name of the field) appeared instead, with seemingly no way to override it.

**Steps to reproduce:**
Open the edit view of any tag (or navigate to `/#/tags/1`).
Look at the label on the name field.

**Related code:**
https://codesandbox.io/s/eloquent-fermi-jjxte?file=/src/tags/TagEdit.tsx

**Other information:**
[This line](https://github.com/marmelab/react-admin/blob/45921e2d3b479648598535c90b2e970722da593d/packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx#L57) overwrites `child.props.label`

**Environment**

* React-admin version: 3.16.4
* React version: 17.0.2
* Browser: Chrome 91

mjomble:
Fixes #6414

### DIFF
--- a/packages/ra-core/src/i18n/TranslatableContext.ts
+++ b/packages/ra-core/src/i18n/TranslatableContext.ts
@@ -13,5 +13,5 @@ export interface TranslatableContextValue {
 }
 
 export type GetTranslatableSource = (field: string, locale?: string) => string;
-export type GetTranslatableLabel = (field: string) => string;
+export type GetTranslatableLabel = (field: string, label?: string) => string;
 export type SelectTranslatableLocale = (locale: string) => void;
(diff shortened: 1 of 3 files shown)


fzaninotto:
I think you should update the useTranslatable instead so that getLabel accepts both the source and the label, and pass the correct label to getFieldLabelTranslationArgs

mjomble:
@fzaninotto done. Also found a mistake in a comment and fixed it.

djhi (approved):

PR #6418, opened 2021-07-06
Add support for image path value in <SimpleList leftAvatar>

fzaninotto:
## Problem

By default, SimpleList expect avatar values to be initials. When passing a avatar URL, the avatar shows the end of the URL instead of the image. 

```jsx
const UserList = (props) => (
    <List {...props}>
        <SimpleList
            primaryText={record => record.name}
            secondaryText={record => record.role}
            leftAvatar={record => record.avatar}
        />
    </List>
);
```

image

To set an image as avatar in SimpleList, developers must return an `` element. This isn't super intuitive.

```diff
const UserList = (props) => (
    <List {...props}>
        <SimpleList
            primaryText={record => record.name}
            secondaryText={record => record.role}
-           leftAvatar={record => record.avatar}
+           leftAvatar={record => <img alt="" src={record.avatar} />}
        />
    </List>
);
```

Note: `user` records look like:

```
    [
        {
            id: 1,
            name: 'Logan Schowalter',
            role: 'admin',
            avatar: 'https://marmelab.com/posters/avatar-1.jpeg',
        },
        {
            id: 2,
            name: 'Breanna Gibson',
            role: 'user',
            avatar: 'https://marmelab.com/posters/avatar-2.jpeg',
        },
        {
            id: 3,
            name: 'Annamarie Mayer',
            role: 'user',
            avatar: 'https://marmelab.com/posters/avatar-3.jpeg',
        },
    ],
```

## Solution

Avatar accepts both text (displays initials) or image paths (displays image)

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList.tsx
+++ b/packages/ra-ui-materialui/src/list/SimpleList.tsx
@@ -107,7 +107,10 @@ const SimpleList = <RecordType extends Record = Record>(
         avatarCallback: FunctionToElement<RecordType>
     ) => {
         const avatarValue = avatarCallback(data[id], id);
-        if (typeof avatarValue === 'string' && avatarValue.startsWith('http')) {
+        if (
+            typeof avatarValue === 'string' &&
+            (avatarValue.startsWith('http') || avatarValue.startsWith('data:'))
+        ) {
             return <Avatar src={avatarValue} />;
         } else {
             return <Avatar>{avatarValue}</Avatar>;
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/SimpleList.tsx
```diff
@@ -102,6 +102,18 @@ const SimpleList = <RecordType extends Record = Record>(
         );
     }
 
+    const renderAvatar = (
+        id: Identifier,
+        avatarCallback: FunctionToElement<RecordType>
+    ) => {
+        const avatarValue = avatarCallback(data[id], id);
+        if (typeof avatarValue === 'string' && avatarValue.startsWith('http')) {
```
djhi, line 110: What about base64 images ?

djhi (approved):

PR #6449, opened 2021-07-19
[Doc] Fix code examples

WiXSL:


docs/Resource.md
```diff
@@ -130,6 +130,9 @@ to read the current resource name, use the `useResourceContext()` hook.
 For instance, the following component displays the name of the current resource:
 
 ```jsx
+import { Datagrid, DateField, TextField, List, useResourceContext } from 'react-admin';
```
fzaninotto, line 134: While you are at it, I realize that there is a mistake on line 126: we're mentioning RecordContext but it should be ResourceContext. Would you mind making the change?

fzaninotto (changes requested):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6457, opened 2021-07-22
[Demo] Improve learning experience by keeping component names in React DevTools

fzaninotto:
## Problem

Create-React-App uses the TerserPlugin to minimize the production build. One side effect is that it removes the React component names, which makes a poor learning experience in our demos since the React DevTools show all components as "Anonymous":

Capture d’écran 2021-07-22 083435
Capture d’écran 2021-07-22 084136

## Solution

Override the Webpack TerserPlugin configuration to disable name mangling.

While this change increases the build size a bit (about +35kB), we don't really care as our demos are intended for learning. 

Capture d’écran 2021-07-22 083644
Capture d’écran 2021-07-22 084212

Bonus: I enabled the React Profiler in production mode, too.

These changes have already been pushed to the demos:
- e-commerce: https://marmelab.com/react-admin-demo/#/
- CRM: https://marmelab.com/react-admin-crm/#/

### DIFF
--- a/examples/crm/package.json
+++ b/examples/crm/package.json
@@ -29,17 +29,18 @@
     "@types/react": "^16.9.13",
     "@types/react-beautiful-dnd": "^13.0.0",
     "@types/react-dom": "^16.9.9",
+    "rewire": "^5.0.0",
     "source-map-explorer": "^2.0.0",
     "typescript": "^4.0.2",
     "web-vitals": "^1.0.1"
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "node ./build.js",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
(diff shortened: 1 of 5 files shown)


examples/crm/package.json
```diff
@@ -29,17 +29,18 @@
...
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "node ./build.js",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
-  "homepage": "https://marmelab.com/react-admin-crm",
+  "homepage": ".",
```
fzaninotto, line 43: This allows the built site to be served from any location. It's more flexible than restricting it to one URL, and works better locally without breaking the production build.

See https://stackoverflow.com/a/44048328/1333479

djhi (approved):
Nice!

PR #6462, opened 2021-07-23
Overriding some of UI initial states

smeng9:
Replaces #6345 based on the feedback from previous reviews

### DIFF
--- a/packages/ra-core/src/reducer/admin/ui.ts
+++ b/packages/ra-core/src/reducer/admin/ui.ts
@@ -76,7 +76,7 @@ const uiReducer: Reducer<UIState> = (
         case STOP_OPTIMISTIC_MODE:
             return { ...previousState, optimistic: false };
         default:
-            return { ...defaultState, ...previousState};
+            return { ...defaultState, ...previousState };
     }
 };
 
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/reducer/admin/ui.ts
```diff
@@ -76,7 +76,7 @@ const uiReducer: Reducer<UIState> = (
         case STOP_OPTIMISTIC_MODE:
             return { ...previousState, optimistic: false };
         default:
-            return previousState;
+            return { ...defaultState, ...previousState };
```
fzaninotto, line 79: This is a big performance issue: Every action will update the UI state, which will cause a redraw of the components depending on it (e.g. sidebar). So this is not an acceptable solution. 

I advise you to check the createAdminStaore instead, to allow partial initialState to be used.

fzaninotto (changes requested):

packages/ra-core/src/reducer/admin/ui.ts
smeng9: The UI initial states are now overridden in the createAdminStore by the partial initialState passed by the user

smeng9:
Closing this since redux is removed in latest alpha

PR #6463, opened 2021-07-24
Fix no-code-demo with correct config

### LINKED ISSUE #6432: [Lab] no-code example cannot start in codesandbox
Hi, It was nice there are new packages to make a try. 🚀

However when I run the no-code example in codesandbox, it is always showing a white screen.
I am not sure if there are some config files not set up correctly

**What you were expecting:**

The demo should run successfully

**What happened instead:**

The demo cannot start

**Steps to reproduce:**

https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/no-code

**Related code:**

**Environment**

* React-admin version: 3.17
* React version: 17.0.2
* Browser: Edge 91

smeng9:
Closes #6432 

Here is an example codesandbox project with the changed config file that is working:
https://codesandbox.io/s/affectionate-allen-0xqij?file=/vite.config.ts:0-187

### DIFF
--- a/examples/no-code/vite.config.ts
+++ b/examples/no-code/vite.config.ts
@@ -1,54 +1,7 @@
 import { defineConfig } from 'vite';
 import reactRefresh from '@vitejs/plugin-react-refresh';
-import path from 'path';
 
 // https://vitejs.dev/config/
 export default defineConfig({
     plugins: [reactRefresh()],
-    alias: [
-        {
-            find: /^react-admin$/,
-            replacement: path.resolve(
-                __dirname,
-                '../../packages/react-admin/src'
-            ),
-        },
-        {
-            find: /^ra-core$/,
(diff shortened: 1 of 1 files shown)


djhi (changes requested):
Thanks! However, this change would make local development cumbersome. It would be nice to [detect we are on codesandbox](https://codesandbox.io/docs/environment#client-environments) and avoid aliases only in this case

smeng9:
Hi @djhi, with environment variable detection it works as expected. Thanks.

https://codesandbox.io/s/github/smeng9/react-admin/tree/no-code-demo-fix/examples/no-code?file=/vite.config.ts

WiXSL (approved):

djhi (approved):
Thanks!

WiXSL:
@djhi, you forgot to add it to the Milestone

PR #6465, opened 2021-07-26
Fix props.options.labelWidth not being applied in SelectArrayInput

### LINKED ISSUE #6341: SelectArrayInput to respect `{options: {labelWidth}}` and not override it
SelectArrayInput sometime shows a narrow drop-down which squeeze the label on initial render. Current implementation only adjust the size of the `Select` based on selection items width on opening up the `Select` component.

Currently it pass own its own `labelWidth` prop to the `Select` component.

If `{options: {labelWidth}}` props is provided, to respect that as the initial `labelWidth` instead of the calculated one.

**Additional context**
Alternative solution: It checks the size of the `labelText` and calculates the `labelWidth`.

WiXSL:
Fixes #6341


djhi (approved):

djhi:
Thanks!

PR #6470, opened 2021-07-27
Fix SearchInput border style when variant is outlined

### LINKED ISSUE #6468: Using "outlined" variant break Input border
**What you were expecting:**
Up until version 3.15, using the "outlined" variant in Input fields worked as expected.

**What happened instead:**
After upgrading to 3.17, if I click on an Input that has `variant='outlined'`, it cuts the border above the first character of the field.

**Steps to reproduce:**
Just click the field and you can see the border changing:

https://user-images.githubusercontent.com/331353/127083307-2a557c5f-3a59-46ee-9290-1873d4bcba8f.mov

**Related code:**
Just add the "outlined" variant, PostList.js:44
```jsx
const postFilter = [
  <SearchInput source="q" alwaysOn variant="outlined" />, 
  <TextInput source="title" defaultValue="Qui tempore rerum et voluptates" />,
  <QuickFilter
    label="resources.posts.fields.commentable"
    source="commentable"
    defaultValue
  />
];
```
Code Sandbox: https://codesandbox.io/s/gracious-frog-eknfj?file=/src/posts/PostList.tsx:1093-1362

**Environment**

* React-admin version: 3.17.0 (but was able to reproduce locally with 3.16.5)
* Last version that did not exhibit the issue (if applicable): 3.15 didn't have this bug
* React version: 17
* Browser: Chrome Version 91.0.4472.114 (Official Build) (x86_64)

susiyaki:
Close #6468 

Screenshot from 2021-07-28 01-15-46

## Causes

If label is falsy value, FieldTitle component returned null value. But it isn't string value, it's react element. So MuiTextField don't interpret label is empty.

## Resolve

In TextInput component, pass null value when label is falsy value.

### DIFF
--- a/packages/ra-ui-materialui/src/input/TextInput.tsx
+++ b/packages/ra-ui-materialui/src/input/TextInput.tsx
@@ -3,6 +3,8 @@ import { FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
 import { useInput, FieldTitle, InputProps } from 'ra-core';
 import { TextFieldProps } from '@material-ui/core/TextField';
+import { Theme } from '@material-ui/core';
+import { makeStyles, Styles } from '@material-ui/styles';
 
 import ResettableTextField from './ResettableTextField';
 import InputHelperText from './InputHelperText';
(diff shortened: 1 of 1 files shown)


djhi (approved):
Thanks!

djhi:
It looks the tests are failing though. Can you check?

susiyaki:
@djhi 
I'll check it

packages/ra-ui-materialui/src/input/TextInput.tsx
```diff
@@ -25,20 +27,37 @@ export type TextInputProps = InputProps<TextFieldProps> &
...
+        '& .MuiOutlinedInput-notchedOutline': {
+            '& legend': {
+                width: ({ label }) =>
+                    label === false || label === '' ? '0.01px' : 'auto',
+            },
+        },
+    },
+};
+
+const useStyles = makeStyles(textInputStyles, { name: 'RaTextInput' });
```
djhi, line 42: Can you put all that at the bottom of the file and put the comment back above the `TextInput` declaration? Thanks!

djhi (changes requested):

packages/ra-ui-materialui/src/input/TextInput.tsx
djhi: Btw, why did you extract the styles into `textInputStyles` instead of inlining them in the `makeStyles` call as we do everywhere else?

susiyaki: Okay!
I refer to ResettableTextField components. I also wondered why it, so I change this code.

susiyaki: I fixed it, thank you!

fzaninotto:
Thanks for your work, but as explained in the issue, the right fix is to rollback the change that caused the problem in the first place. 

Superseded by #6492

PR #6472, opened 2021-07-28
Fix delete buttons props types

djhi:
Both buttons don't need the undoable prop


WiXSL:
@djhi I could do this can of maintenance for the rest of the components if you want
Let me know

djhi:
@WiXSL That wasn't the point of the PR. I just do it whenever I edit a file which still uses `FC`. However, if you have the courage to do so, be my guest!

fzaninotto (approved):

PR #6485, opened 2021-08-03
Add a template for create-react-app

djhi:


fzaninotto:
No sure about the value for now as we already have the simple example in CodeSandbox and the basic install takes one `yarn install` command and a copy/paste. closing for now, we may revise it in the future.

PR #6498, opened 2021-08-10
[Doc] Fix Syntax Error in linkToRecord code snippet

### LINKED ISSUE #6497: Docs: linkToRecord example code showing undefined variable
Hello, 

From the [linkToRecord documentation](https://marmelab.com/react-admin/Fields.html#linking-to-other-records) there is a line of code which returning `<Link>` with `{seller.username}` as it's value. Unfortunately, it's showing in the editor: 
Screen Shot 2021-08-10 at 21 28 30
which I think the `seller` variable is not undefined yet.

I know it is just an example code that we need to change to match with our production code. But is it a good idea to make the variable more clear to the reader?

Thanks.

fzaninotto:
Closes #6497

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -1621,7 +1621,7 @@ import { Link } from 'react-router-dom';
 
 const AuthorField = () => {
     const post = useRecordContext(props);
-    const { data, loaded } = useGetOne('users',  post.user_id);
+    const { data, loaded } = useGetOne('users', post.user_id);
     const userShowPage = linkToRecord('/users', post.user_id, 'show');
 
     return loaded ? <Link to={userShowPage}>{data.username}</Link> : null;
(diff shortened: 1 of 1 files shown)


docs/Fields.md
```diff
@@ -1611,19 +1611,20 @@ const UserShow = props => (
...
 ```js
-import { linkToRecord, useRecordContext } from 'react-admin';
+import { linkToRecord, useRecordContext, useGetOne } from 'react-admin';
 
-const MyCustomField = () => {
+const AuthorField = () => {
     const post = useRecordContext(props);
-    const linkToUser = linkToRecord('/users', post.user_id, 'show');
+    const { data, loaded } = useGetOne('users',  post.user_id);
```
mddarmawan, line 1624: I think there is a double space between `useGetOne('users',` and `post.user_id);`

fzaninotto: Well spotted!

PR #6500, opened 2021-08-11
Fix useList pagination total

yksflip:
The current implementation to calculate the total in useList is broken. `finalItems.ids.length` changes once pagination is rendered ([useList.ts#L192](https://github.com/marmelab/react-admin/blob/74606bf8d5c1c72ac3c54ba22be8c1fa7b592fd2/packages/ra-core/src/controller/useList.ts#L192)) . So the value of total equals perPage.
I think using the `ids.length` instead would be appropriate to fix this.

### DIFF
--- a/packages/ra-core/src/controller/useList.ts
+++ b/packages/ra-core/src/controller/useList.ts
@@ -73,9 +73,11 @@ export const useList = (props: UseListOptions): UseListValue => {
     const [finalItems, setFinalItems] = useSafeSetState<{
         data: RecordMap;
         ids: Identifier[];
+        total: number;
     }>(() => ({
         data: indexById(data),
         ids,
+        total: 0,
     }));
 
     // pagination logic
@@ -174,6 +176,9 @@ export const useList = (props: UseListOptions): UseListValue => {
                 return result;
             })
         );
+
+        const filteredLength = tempData.length;
+
         // 2. sort
         if (sort.field) {
             tempData = tempData.sort((a, b) => {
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

packages/ra-core/src/controller/useList.ts
```diff
@@ -241,7 +241,7 @@ export const useList = (props: UseListOptions): UseListValue => {
         setPerPage,
         setSort,
         showFilter,
-        total: finalItems.ids.length,
+        total: ids.length,
```
fzaninotto, line 244: no, the total should take into account the filtering, so you need to introduce a new variable to get the length after filterring but before pagination

fzaninotto (changes requested):

packages/ra-core/src/controller/useList.ts
```diff
@@ -73,9 +73,11 @@ export const useList = (props: UseListOptions): UseListValue => {
     const [finalItems, setFinalItems] = useSafeSetState<{
         data: RecordMap;
         ids: Identifier[];
+        total: number;
     }>(() => ({
         data: indexById(data),
         ids,
+        total: 0,
```
fzaninotto, line 80: I'd initialize the total to ids.length, for consistency

fzaninotto (changes requested):
Great!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6504, opened 2021-08-11
Fix filter set via URL isn't persistent if set on the list page

### LINKED ISSUE #6502: When changed via the URLs, filters aren't persistent
**What you were expecting:**

Linking to a filtered list should work the same as filtering the list using filter controls 

**What happened instead:**

Filters enabled via URL parameters aren't "persisted" in the store, i.e. the list isn't filtered anymore when the user comes back to it.

**Steps to reproduce:**

1. Go to the eCommerce demo at https://marmelab.com/react-admin-demo/#/
2. Go to the Posters list https://marmelab.com/react-admin-demo/#/products
3. Filter the Posters list by adding the following query params: `?filter=%7B"category_id"%3A0%7D&order=ASC&page=1&perPage=20&sort=reference`
4. The list only displays Animals posters
5. Go to the Customers list https://marmelab.com/react-admin-demo/#/customers
6. Go back to the Posters list  https://marmelab.com/react-admin-demo/#/products
7. The list shows all posters - the Animals filter wasn't persisted

Note that this doesn't happen if redirecting to the filtered list from another page:

1. Go to the eCommerce demo at https://marmelab.com/react-admin-demo/#/
2. Go to the filtered Posters list https://marmelab.com/react-admin-demo/#/products?filter=%7B"category_id"%3A0%7D&order=ASC&page=1&perPage=20&sort=reference
3. The list only displays Animals posters
4. Go to the Customers list https://marmelab.com/react-admin-demo/#/customers
5. Go back to the Posters list  https://marmelab.com/react-admin-demo/#/products
6. The list shows only animal posters

image

**Related code:**

This is probably due to https://github.com/marmelab/react-admin/blob/7fdaa72a52a7a6645b11907a2d425af508dc02d2/packages/ra-core/src/controller/useListParams.ts#L156-L164

The `changeListParams` action is only dispatched on mount, not when params in the URL change.

**Environment**

* React-admin version: 3.17
* Last version that did not exhibit the issue (if applicable): none - the bug was always here
* React version: v17.0.2
* Browser: Chrome

fzaninotto:
Closes #6502

### DIFF
--- a/packages/ra-core/src/controller/useListParams.ts
+++ b/packages/ra-core/src/controller/useListParams.ts
@@ -3,7 +3,6 @@ import { useSelector, useDispatch, shallowEqual } from 'react-redux';
 import { parse, stringify } from 'query-string';
 import lodashDebounce from 'lodash/debounce';
 import pickBy from 'lodash/pickBy';
-import isEqual from 'lodash/isEqual';
 import { useHistory, useLocation } from 'react-router-dom';
 
 import queryReducer, {
--- a/packages/ra-core/src/controller/useListParams.ts
+++ b/packages/ra-core/src/controller/useListParams.ts
@@ -3,6 +3,7 @@ import { useSelector, useDispatch, shallowEqual } from 'react-redux';
 import { parse, stringify } from 'query-string';
 import lodashDebounce from 'lodash/debounce';
 import pickBy from 'lodash/pickBy';
+import isEqual from 'lodash/isEqual';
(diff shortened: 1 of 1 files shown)


diegocasmo:
It looks like we chose to tackle this issue almost at the same time 🙂 (https://github.com/marmelab/react-admin/pull/6505). I'll go ahead and close my PR in favor of yours, as this one also removes an unnecessary `dispatch`  which I didn't caught.

PR #6509, opened 2021-08-13
[RFR] solved bug with useGetMany

### LINKED ISSUE #6420: useGetMany loading/loaded state does not change when query updated
**What you were expecting:**
Using useGetMany, whenever the query parameter changes (the list of ids), the loading/loaded state should change.

**What happened instead:**
The loading/loaded state of useGetMany is only true/false on the first run of the query. If you update the list of ids you want to get, loading/loaded remains false/true even though you are getting new data. This means any checks for loaded state will return true even though the data isn't actually loaded yet.

**Steps to reproduce:**
* Use a useGetMany hook to call a resource and pass in a stateful variable as the list of ids.
* update the stateful variable to trigger a refresh of the data
* check the state of loading/loaded

**Related code:**
https://codesandbox.io/s/react-admin-usegetmany-issue-tucjj

Go to the url /custom2 in the CodeSandbox browser.

All code is in the customRouteLayout.tsx file

Sample of the primary code at issue
```javascript
const [postIds, setPostIds] = React.useState([1, 2]);

const { data, loaded } = useGetMany("posts", postIds, {
    enabled: postIds.length > 0
});
console.log("loaded", loaded);
console.log("data", data);

const updatePostIds = () => {
  setPostIds([1, 2, 3, 4, 5, 6]);
};
```

**Other information:**
I think the loading/loaded state might just be getting set too late. It appears as though the redux store is populated with the a number of undefined objects based on the number of ids passed to useGetMany. Perhaps the loading/loaded state are getting set after this? Problem is that then your component sees a bunch of undefined items it tries to process and inevitably leads to errors because it can't access any properties on those objects because they haven't actually been loaded yet.

Can the loading/loaded state be set before the undefined elements are added to the redux store?

**Environment**

* React-admin version: 13.16.4
* React version: 17.0.2
* Browser: Chrome

gianpieropa:
fixes #6420
This is my first pull request in an open source project, please be kind :)

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetMany.ts
+++ b/packages/ra-core/src/dataProvider/useGetMany.ts
@@ -219,6 +219,16 @@ const callQueries = debounce(() => {
                 }
             });
             return;
+        } else {
+            // for every query if the ids are not all set to null set loading to true and loaded to false.
+            queries.forEach(({ ids, setState }) => {
+                if (ids.findIndex(v => v != null && v !== '') !== -1)
+                    setState(prevState => ({
+                        ...prevState,
+                        loading: true,
+                        loaded: false,
+                    }));
+            });
         }
         dataProvider
             .getMany(resource, { ids: accumulatedIds }, DataProviderOptions)


WiXSL:
Superseded by #6509

PR #6514, opened 2021-08-18
Remove FC usage from Link component

WiXSL:


packages/ra-ui-materialui/src/button/BulkDeleteButton.tsx
```diff
@@ -31,12 +30,15 @@ import BulkDeleteWithUndoButton, {
...
-    undoable ? (
-        <BulkDeleteWithUndoButton {...props} />
+const BulkDeleteButton = (props: BulkDeleteButtonProps) => {
+    const { undoable, ...rest } = props;
+
+    return undoable ? (
+        <BulkDeleteWithUndoButton {...rest} />
     ) : (
-        <BulkDeleteWithConfirmButton {...props} />
+        <BulkDeleteWithConfirmButton {...rest} />
```
djhi, line 39: Nope, we should actually pass `props` here, as you can have a button which will ask for confirmation and still display an undoable notification. See https://github.com/marmelab/react-admin/pull/6472/files#diff-cf6d8ea9685fb8b58acfbc7f9afd4e47f13fd410266dcdb1c5f3bd642d247ecdR117

packages/ra-ui-materialui/src/button/DeleteButton.tsx
```diff
@@ -49,24 +49,20 @@ import { DeleteWithConfirmButton } from './DeleteWithConfirmButton';
...
 
     return mode === 'undoable' ? (
-        <DeleteWithUndoButton record={record} {...props} />
+        <DeleteWithUndoButton record={record} {...rest} />
     ) : (
         <DeleteWithConfirmButton
             mutationMode={mode}
             record={record}
-            {...props}
+            {...rest}
```
djhi, line 65: Here, we should pass `undoable` as well

packages/ra-ui-materialui/src/button/BulkDeleteButton.tsx
djhi: This is something we'll probably revisit later though

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -118,4 +113,4 @@ export interface BooleanFieldProps
     looseValue?: boolean;
 }
 
-export default BooleanField;
+export default memo<BooleanFieldProps>(BooleanField);
```
djhi, line 116: :+1:

packages/ra-ui-materialui/src/layout/DashboardMenuItem.tsx
```diff
@@ -28,7 +25,7 @@ export interface DashboardMenuItemProps {
     onClick?: () => void;
     dense?: boolean;
     /**
-     * @deprcated
+     * @deprecated
```
djhi, line 28: :+1:

djhi (changes requested):
Thanks for taking the time to do this!

WiXSL:
> Thanks for taking the time to do this!

You are welcome!
Some components where missing do to some errors, I'll change them in future PRs

PR #6515, opened 2021-08-18
[RFR] Remove FC usage from `ra-core` components

WiXSL:


djhi (approved):
:+1: Thanks!

lukascivil:
Sorry for the question, guys, but why? what are the real benefits?

djhi:
We should also rename and export many of the current `Props` interfaces

PR #6519, opened 2021-08-19
[RFR] Remove FC usage from examples

WiXSL:


examples/demo/src/dashboard/Dashboard.tsx
```diff
@@ -73,9 +67,9 @@ const Dashboard: FC = () => {
             }
         );
         const aggregations = recentOrders
-            .filter(order => order.status !== 'cancelled')
+            .filter((order: Order) => order.status !== 'cancelled')
```
djhi, line 70: This shouldn't be necessary as `recentOrders` is already an `Order[]` thanks to `getList`

examples/demo/src/dashboard/Dashboard.tsx
```diff
@@ -73,9 +67,9 @@ const Dashboard: FC = () => {
...
             .reduce(
-                (stats: OrderStats, order) => {
+                (stats: OrderStats, order: Order) => {
```
djhi, line 72: Same

examples/demo/src/dashboard/Dashboard.tsx
```diff
@@ -138,7 +132,7 @@ const Dashboard: FC = () => {
         const { data: customers } = await dataProvider.getMany<Customer>(
             'customers',
             {
-                ids: pendingReviews.map(review => review.customer_id),
+                ids: pendingReviews.map((review: Review) => review.customer_id),
```
djhi, line 135: Same

djhi (changes requested):
Thanks!

djhi (approved):

PR #6523, opened 2021-08-21
Added link to Azure Active Directory  example implementation

victorp13:


fzaninotto:
Your repo doesn't contain an actual, reusable authProvider, it's rather a proof-of-concept. I'd prefer if you could publish your code as a npm package that people could reuse.

victorp13:
I am very new to JS so am being careful to make npm packages. However, I consider my example to be very clean, and hence still a valuable addition to your documentation. I've seen people asking questions (to you, but also to the MSAL team) on how to do this type of integration: My repo can help them. 

Even without a ready-made package/authProvider my example can be useful. I will look at creating an authProvider, but it's up to you.

PR #6525, opened 2021-08-23
[Doc] Fix emptyText prop description in Fields documenation

### LINKED ISSUE #6179: emptyText prop on ArrayField has no effect.
**What you were expecting:**
Expected emptyText to be output from an ArrayField when there are no items in that array.

**What happened instead:**
Instead, we get an empty <DataGrid> (or I assume whatever child) with the table headers, and no data.

**Steps to reproduce:**
Can do something like:

```
<ArrayField
      label={'YouTube Videos'}
      source={'source'}
      emptyText={'None'}
    >
      <Datagrid>
        <TextField source={'videoId'} />
        <NumberField source={'weight'} />
      </Datagrid>
    </ArrayField>
```

And the result is:

image

Instead of the empty text.

**Environment**

* React-admin version: 3.14.3
* React version: 17.0.1
* Browser: Chrome 89.0.4389.128 on Mac.

fzaninotto:
Closes #6179


olliebennett (approved):
Great - I think this reads well 👍

PR #6540, opened 2021-08-29
Fix Yarn

quentingarcia:
react-admin is a monorepo

To specify the topology to Lerna, we need to include packages in devDependencies and peerDependencies.
This will help Lerna to understand the dependency.
When possible, we prefer to use "devDependencies" and "peerDependencies" instead of "dependencies"

Edit multiple package.json dependencies to fix yarn install/build errors.


djhi (approved):
Tested locally, this indeed fixes the build workflow when not using the make commands. Thanks!

PR #6548, opened 2021-09-01
[RFR] Fix DataResult type

### LINKED ISSUE #6530: Inconsistent type for DeleteResult
**What you were expecting:**
I returned `{}` from my custom data provider's `delete` method and expected it to work because the `data` property is typed as optional:
https://github.com/marmelab/react-admin/blob/b007fb7176d2ca4a779acddccd5b0316987b1d03/packages/ra-core/src/types.ts#L208

**What happened instead:**
I got this error: "The response to 'delete' must be like { data: ... }, but the received response does not have a 'data' key. The dataProvider is probably wrong for 'delete'."

**Steps to reproduce:**
Use a custom dataProvider with the following delete method: `delete: () => ({})`, then try deleting a record.

**Related code:**
https://codesandbox.io/s/morning-thunder-ofouy?file=/src/dataProvider.tsx

**Other information:**
This line requires the `data` property to be present for all responses:
https://github.com/marmelab/react-admin/blob/b007fb7176d2ca4a779acddccd5b0316987b1d03/packages/ra-core/src/dataProvider/validateResponseFormat.ts#L17
It should probably have an exception for delete operations.

I can submit a PR that adds this exception, but maybe there are other pieces of code that also require this property to always be present. A comment from someone more familiar with the codebase would be helpful.

**Environment**

* React-admin version: 3.17.1
* React version: 17.0.2
* Browser: Chrome 92 on Windows 10

WiXSL:
Fixes #6530

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -205,7 +205,7 @@ export interface DeleteParams {
     previousData: Record;
 }
 export interface DeleteResult<RecordType = Record> {
-    data?: RecordType;
+    data: RecordType;
 }
 
 export interface DeleteManyParams {


fzaninotto (approved):

fzaninotto:
Thanks!

Issue #6549, opened 2021-09-02
BooleanField no longer displays label by default


flouc001:
**What you were expecting:**
When using the `BooleanField` the label should be displayed by default due to the `defaultProps` setting `addLabel` to `true`.

**What happened instead:**
`BooleanField` no longer displays the label without explicitly specifying the `addLabel` prop, this is breaking.

**Steps to reproduce:**
1. Add a `BooleanField` component to a view with the `label` prop set.
2. Observe the field being rendered without the label.
3. Add the `addLabel` prop.
4. Observe the label being rendered.

**Related code:**
https://codesandbox.io/s/eloquent-silence-1te8b

Observe the tick with no label on this page: https://1te8b.sse.codesandbox.io/#/posts/13/show/2

**Other information:**
I think it's down to the memo being moved directly to the default export in this commit: https://github.com/marmelab/react-admin/commit/b79787a65d82da82e3862e7e9c7fadaf82525061#diff-d7c4dd12e78cb28e0370dc53068aa45ec475f6fe54b5b40db238547e6278f5aaR119

It's no longer respecting the `defaultProps`.

**Environment**

* React-admin version: 3.17.3
* Last version that did not exhibit the issue (if applicable): 3.17.2
* React version: 17
* Browser: Chrome
* Stack trace (in case of a JS error): N/A

flouc001:
@WiXSL what's the motivation to moving away from FC typing? I'm guessing you moved the memo because Typescript was throwing an error.

WiXSL:
> @WiXSL what's the motivation to moving away from FC typing? I'm guessing you moved the memo because Typescript was throwing an error.

Here: https://github.com/marmelab/react-admin/pull/6515#issuecomment-906702131

One solution could be to use default values for props inside the component instead of `BooleanField.defaultProps`

flouc001:
@WiXSL thank you for the response - the motivation makes sense! I raised a PR with an idea for maybe moving the responsibility to the displaying components. Let me know what you think, just throwing some ideas out there.

WiXSL:
We have to ping @fzaninotto and @djhi on how to solve this

flouc001:
Agreed, that's why I was thinking of moving the logic to the component with the responsibility of displaying the component. That way it fixes all of the fields that may encounter this issue rather than just `BooleanInput`.

eamahanna:
I also encountered this issue with the `DateField` and `ArrayField`.

PR #6550, opened 2021-09-02
[RFR] Remove reliance on defaultProps for adding labels

flouc001:
#6549 sees that adding a label to fields relies on the `addLabel` default prop being set. This reliance is compromised when wrapping the component with `useMemo`.

This code moves the reliance from `defaultProps` set on the field to the components displaying them all-the-while respecting the prop if it's being set by the user.

### DIFF
--- a/packages/ra-ui-materialui/src/detail/SimpleShowLayout.tsx
+++ b/packages/ra-ui-materialui/src/detail/SimpleShowLayout.tsx
@@ -74,7 +74,8 @@ const SimpleShowLayout = ({
                         field.props.className
                     )}
                 >
-                    {field.props.addLabel ? (
+                    {field.props.addLabel ??
+                    (field.props.label || field.props.source) ? (
                         <Labeled
                             record={record}
                             resource={resource}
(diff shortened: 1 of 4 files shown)


djhi:
I'm -1 on this. We are thinking about removing the `addLabel` prop completely in v4 and remove the responsability of adding labels from the forms and layouts. Every field would be responsible for displaying its label but would have access to its context (form, readonly view, list).

flouc001:
@djhi - I think the direction of removing inspection of child component props is absolutely the right way to go. This is less of a long term philosophy change, more of a fix what's broken for now.

Would you consider introducing it as an interim solution?

fzaninotto:
Your PR is indeed a BC break, so it doesn't only solve your problem, it creates other problems. I'm also -1 on this one. Replacing the inspection of one child prop by the inspection of another child prop will only move the problem. The proper solution is the one described by @djhi, and it's scheduled for v4. In the meantime, we'll have to deal with the problem. 

thanks for your PR anyway, feel free to suggest other changes!

PR #6552, opened 2021-09-02
Add record context in SimpleList

djhi:
SimpleList now wrap each item inside a record context and support React elements directly for the `primaryText`, `secondaryText` and `tertiaryText` props.

This allows to use any of the react-admin fields and make it way easier to display referenced records data:

```jsx
// in src/posts.js
import { List, SimpleList } from 'react-admin';

const postRowStyle = (record, index) => ({
    backgroundColor: record.nb_views >= 500 ? '#efe' : 'white',
});

    <List {...props}>
        <SimpleList
            primaryText={<TextField source="title" />}
            secondaryText={record => `${record.views} views`}
            tertiaryText={
                <ReferenceField reference="categories" source="category_id">
                    <TextField source="name" />
                </ReferenceField>
            }
            linkType={record => record.canEdit ? "edit" : "show"}
            rowStyle={postRowStyle}
        />
    </List>
);
```

It also fixes a semantic issue as the SimpleList display an `ul` but the children were anchors instead of `li`

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -2499,7 +2499,7 @@ export const PostList = (props) => (
 
 For each record, `<SimpleList>` executes the `primaryText`, `secondaryText`, `linkType`, `rowStyle`, `leftAvatar`, `leftIcon`, `rightAvatar`, and `rightIcon` props functions, and creates a `<ListItem>` with the result.
 
-The `primaryText`, `secondaryText` and `tertiaryText` functions can return a React element. This means you can use any react-admin fields, including reference fields:
+The `primaryText`, `secondaryText` and `tertiaryText` functions can return a React element. This means you can use any react-admin field, including reference fields:
 
 ```jsx
 // in src/posts.js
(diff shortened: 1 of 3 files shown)


docs/List.md
```diff
@@ -2499,6 +2499,34 @@ export const PostList = (props) => (
 
 For each record, `<SimpleList>` executes the `primaryText`, `secondaryText`, `linkType`, `rowStyle`, `leftAvatar`, `leftIcon`, `rightAvatar`, and `rightIcon` props functions, and creates a `<ListItem>` with the result.
 
+The `primaryText`, `secondaryText` and `tertiaryText` functions can return a React element. This means you can use any react-admin fields, including reference fields:
```
fzaninotto, line 2502:
```suggestion
The `primaryText`, `secondaryText` and `tertiaryText` functions can return a React element. This means you can use any react-admin field, including reference fields:
```

fzaninotto (changes requested):
Awesome!

fzaninotto (approved):

PR #6553, opened 2021-09-03
Fix BooleanField doesn't show label when used in a Show view

### LINKED ISSUE #6549: BooleanField no longer displays label by default

WiXSL:
Fixes #6549

If this is ok, I can do the same for the other fields

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -27,15 +27,16 @@ const useStyles = makeStyles(
 
 const BooleanField = (props: BooleanFieldProps) => {
     const {
+        addLabel = true,
         className,
         classes: classesOverride,
         emptyText,
         source,
         valueLabelTrue,
         valueLabelFalse,
-        TrueIcon,
-        FalseIcon,
-        looseValue,
+        TrueIcon = DoneIcon,
+        FalseIcon = ClearIcon,
+        looseValue = false,
         ...rest
     } = props;
     const record = useRecordContext(props);
@@ -82,13 +83,6 @@ const BooleanField = (props: BooleanFieldProps) => {
             {emptyText}
         </Typography>
     );
-};
-
-BooleanField.defaultProps = {
-    addLabel: true,
-    TrueIcon: DoneIcon,
-    FalseIcon: ClearIcon,
-    looseValue: false,
 };
 
 BooleanField.propTypes = {
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -25,18 +25,17 @@ const useStyles = makeStyles(
     }
 );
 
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -27,15 +27,16 @@ const useStyles = makeStyles(
 
 const BooleanField = (props: BooleanFieldProps) => {
     const {
+        addLabel = true,
```
djhi, line 30: I believe this needs to be set on the exported component (the memo)

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/field/BooleanField.tsx
WiXSL: Any hint on how to do this since React.memo does not support defaultProps, only the inner component does.

flouc001: This was how I landed on moving the default logic to the component that clones the component because there didn't appear to be a good way of doing this whilst maintaining the memo.

WiXSL:
Maybe I should revert this changes in Field components until version 4, what do you think?

fzaninotto:
@WiXSL yes, please revert the change you made in the first place that moved the memo call - but please keep the removal of FC!

WiXSL:
> @WiXSL yes, please revert the change you made in the first place that moved the memo call - but please keep the removal of FC!

@fzaninotto, the prblem is that i get `TS2339: Property 'propTypes' does not exist on type 'NamedExoticComponent '.` error on `defaultProps` and `propTypes`

fzaninotto:
does @ts-ignore work?

WiXSL:
> does @ts-ignore work?

Sorry, yes, It seems to work

fzaninotto:
Suprseded by #6564

fzaninotto (approved):

Issue #6386, opened 2021-06-23
Can not use ra-data-graphql-simple with React function components


Panzki:
**What you were expecting:**
We want to use the `ra-data-graphql-simple` package with React function components and hooks. (The documentation only contains an example with class components [here](https://github.com/marmelab/react-admin/tree/master/packages/ra-data-graphql-simple#usage).)

**What happened instead:**
When building the data provider from the `ra-data-graphql-simple` package in a function component the React application crashes. The error is: 
```
Unhandled Rejection (Error): Unknown resource undefined. Make sure it has been declared on your server side schema. Known resources are Post
```

**Steps to reproduce:**
1. Clone the repository containing a minimal example: `git clone git@github.com:Panzki/ra-simple-graphql-bug-report.git`
2. Install dependencies: `yarn`
3. Start the GraphQL server: `npx json-graphql-server db.js --p 4000`
4. Start the React application (in another terminal): `yarn start`
5. Open: http://localhost:3000

**Related code:**
The code snippet below uses the useEffect hook in a function component instead of `componentDidMount`. Here is GitHub repository containing a minimal example:  https://github.com/Panzki/ra-simple-graphql-bug-report
```jsx
import buildGraphQLProvider from "ra-data-graphql-simple";
import React from "react";
import { Admin, ListGuesser, Resource } from "react-admin";

function App() {
  const [dataProvider, setDataProvider] = React.useState<any | null>(null);

  React.useEffect(() => {
    buildGraphQLProvider({
      clientOptions: { uri: "http://localhost:4000" },
    }).then((provider) => setDataProvider(provider));
  }, []);

  if (!dataProvider) {
    return <div>Loading</div>;
  }
  return (
    <Admin dataProvider={dataProvider}>
      <Resource name="Post" list={ListGuesser} />
    </Admin>
  );
}

export default App;
```

**Other information:**

**Environment**

* React-admin version: 3.16.3
* Last version that did not exhibit the issue (if applicable):
* React version: 17.0.2
* Browser: FireFox 89.0.1 (64-bit)
* Stack trace (in case of a JS error):
![image](https://user-images.githubusercontent.com/12871965/123110010-001da080-d43c-11eb-81e1-926b32cd0911.png)

djhi:
That's most probably because your resource is named `posts` in your `db.js` file instead of `Post` (the resource name in `Resource`

Panzki:
Hi,
[...]

PR #6555, opened 2021-09-06
[RFR] Updated README of ra-data-graphql-simple.

Panzki:
This PR updates the example in the README of the `ra-data-graphql-simple` data provider to use React function components as discussed in https://github.com/marmelab/react-admin/issues/6386.

There are no code changes at all, so no disruptive/breaking behavior is expected.

### DIFF
--- a/packages/ra-data-graphql-simple/README.md
+++ b/packages/ra-data-graphql-simple/README.md
@@ -34,7 +34,7 @@ import { Admin, Resource } from 'react-admin';
 
 import { PostCreate, PostEdit, PostList } from './posts';
 
-const App = function () {
+const App = () => {
 
     const [dataProvider, setDataProvider] = React.setState(null);
     React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/ra-data-graphql-simple/README.md
```diff
@@ -27,36 +27,30 @@ The `ra-data-graphql-simple` package exposes a single function, which is a const
...
-class App extends Component {
-    constructor() {
-        super();
-        this.state = { dataProvider: null };
-    }
-    componentDidMount() {
-        buildGraphQLProvider({ clientOptions: { uri: 'http://localhost:4000' }})
-            .then(dataProvider => this.setState({ dataProvider }));
-    }
+const App = function () {
```
djhi, line 37:
```suggestion
const App = () => {
```

djhi (changes requested):
Thanks!

djhi (approved):

PR #6567, opened 2021-09-09
Fix EmailField font size differs from other fields

### LINKED ISSUE #6566: EmailField showing larger text than all other Fields
EmailField is showing text at 16px when all other fields are 14px. Was this intentional?

It seems like the field is missing `MuiTypography-root MuiTypography-body2` on the `span` tag and the other have this.

fzaninotto:
Closes #6566

### DIFF
--- a/packages/ra-ui-materialui/src/field/EmailField.tsx
+++ b/packages/ra-ui-materialui/src/field/EmailField.tsx
@@ -34,9 +34,15 @@ const EmailField = memo<EmailFieldProps>((props: EmailFieldProps) => {
             className={className}
             href={`mailto:${value}`}
             onClick={stopPropagation}
-            {...sanitizeFieldRestProps(rest)}
         >
-            {value}
+            <Typography
+                component="span"
+                variant="body2"
+                className={className}
+                {...sanitizeFieldRestProps(rest)}
+            >
+                {value}
+            </Typography>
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/field/EmailField.tsx
```diff
@@ -34,9 +34,15 @@ const EmailField = memo<EmailFieldProps>((props: EmailFieldProps) => {
...
         >
-            {value}
+            <Typography
+                component="span"
+                variant="body2"
+                className={className}
+                {...sanitizeFieldRestProps(rest)}
+            >
+                {value}
+            </Typography>
```
djhi, line 45: I think you can pass these props directly to the `Link` component. See the example above this link: https://material-ui.com/components/links/#api

djhi (changes requested):

packages/ra-ui-materialui/src/field/EmailField.tsx
fzaninotto: Why should I? The props have to end up in the Typography element anyway.

fzaninotto: nevermind, you're right, I'll change that.

djhi (approved):

PR #6569, opened 2021-09-09
Fix  `displayName` prop in Field components

WiXSL:


WiXSL:
The fact that TypeScript loses the displayName if we don't set it explicitly

fzaninotto:
How can I reproduce this problem?

WiXSL:
I just through the `Email Field` field has to have the same treatment as the `UrlField` in that respect.
But you're right, I'm not actually fixing a reproducible problem

fzaninotto:
I have reproduced the problem: using the ReactDevTools, EmailField renders as "Anonymous" while the other fields are properly labeled:

image

I think it's the result of using `memo`. That means that other fields are also impacted. would you mind generalizing your change to all Field components using `memo`?

fzaninotto:
There is something I don't understand: your 3rd commit should create a conflict, as the Field components you've patched have already been fixed in master. 

Could you rebase on master?

WiXSL:
Done

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6575, opened 2021-09-10
Fix resolveRedirectTo linkToRecord utilization

ValentinnDimitroff:
As I was going through the redirect flow got wondering at one point why should we construct the same type of path at two places and as linkToRecord is already used and coupled - why not utilize to the fullest?


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6576, opened 2021-09-10
Rename and export `Mutation` props

WiXSL:


djhi (approved):

djhi:
Thanks!

packages/ra-core/src/dataProvider/index.ts
```diff
@@ -35,6 +35,7 @@ export {
     fetchUtils,
     HttpError,
     Mutation,
+    MutationProps,
```
fzaninotto, line 38: Don't export types inside an `export`, export them under an `export type`, or else the simple example compilation fails. Cf. bd11fe0d728785a587844eab81e1c3e618acee9d for the fix for this PR, and e3dbfea69c72ac00b72702f04b92961a63be53f3 that I did silently before releasing 3.18, which was broken by 2 of your PRs. 

I don't understand why the build step doesn't catch the problem, though.

PR #6577, opened 2021-09-10
Remove `FunctionComponent` usage from `RichTextInput`

WiXSL:


packages/ra-input-rich-text/src/index.tsx
```diff
@@ -45,7 +39,7 @@ interface Props {
     [key: string]: any;
 }
 
-const RichTextInput: FunctionComponent<Props> = props => {
```
djhi, line 48: Would you mind renaming the props and exporting them too? Thanks!

djhi (approved):

packages/ra-input-rich-text/src/index.tsx
WiXSL: Done!

djhi (approved):
Thanks!

PR #6538, opened 2021-08-27
[RFR] Remove `FC` usage from `Field` components

WiXSL:

PR #6585, opened 2021-09-13
[TypeScript] Fix Field components types doesn't expose defaultProps & propTypes

fzaninotto:
Partially reverts #6538

### DIFF
--- a/packages/ra-ui-materialui/src/field/ChipField.tsx
+++ b/packages/ra-ui-materialui/src/field/ChipField.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { memo } from 'react';
+import { memo, FC } from 'react';
 import get from 'lodash/get';
 import Chip, { ChipProps } from '@material-ui/core/Chip';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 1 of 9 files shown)


packages/ra-ui-materialui/src/field/ChipField.tsx
```diff
@@ -17,7 +17,7 @@ const useStyles = makeStyles(
     { name: 'RaChipField' }
 );
 
-export const ChipField = memo<ChipFieldProps>((props: ChipFieldProps) => {
+export const ChipField: FC<ChipFieldProps> = props => {
```
WiXSL, line 20: You forgot to maintain the `memo` call

packages/ra-ui-materialui/src/field/DateField.tsx
```diff
@@ -42,7 +42,7 @@ const toLocaleStringSupportsLocales = (() => {
  * // renders the record { id: 1234, new Date('2012-11-07') } as
  * <span>mercredi 7 novembre 2012</span>
  */
-export const DateField = memo<DateFieldProps>((props: DateFieldProps) => {
+export const DateField: FC<DateFieldProps> = props => {
```
WiXSL, line 45: Same

WiXSL (changes requested):

packages/ra-ui-materialui/src/field/ChipField.tsx
WiXSL: Originally was: 
`export const ChipField: FC<ChipFieldProps> = memo<ChipFieldProps>(props => {`

djhi (approved):

PR #6589, opened 2021-09-15
Fix datagrid usage inside reference array input

### LINKED ISSUE #6531: "Unselect all" broken on Datagrid inside ReferenceArrayInput
I experimented with using a Datagrid inside a ReferenceArrayInput as a more detailed alternative to CheckboxGroupInput.
I was prepared for this use case not to be supported, but actually got it working quite well.
Except for one little thing.

**What you were expecting:**
When all rows are selected and I click on the checkbox in the header row, all rows should become unselected

**What happened instead:**
Nothing changes, all rows are still selected.
(Using the checkbox to select all rows does work, it's only unselecting that doesn't)

**Steps to reproduce:**
I recreated a simpler version of my solution here:
https://codesandbox.io/s/gracious-golick-sylvu?file=/src/posts/TagReferenceInput.tsx

Edit a post and open the 'miscellaneous' tab to see the Datagrid in the tags input.

**Related code:**
When the checkbox is unchecked, this line is executed:
https://github.com/marmelab/react-admin/blob/b007fb7176d2ca4a779acddccd5b0316987b1d03/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx#L181

Which works well in a regular List context, but `useReferenceArrayInputController` has unusual logic:
https://github.com/marmelab/react-admin/blob/b007fb7176d2ca4a779acddccd5b0316987b1d03/packages/ra-core/src/controller/input/useReferenceArrayInputController.ts#L122-L130

Instead of using `newIds` as the new value, it is added to the existing value. So we get `all ids` + `[]` = `all ids`.

I'm not sure why it does that, but I'm assuming we can't just change it to `form.change(input.name, newIds);` without breaking something else.

Alternatively, Datagrid could use `onUnselectItems()` instead of `onSelect([])`.

I might be able to submit a PR if I get some feedback on this.

I'll also understand if this scenario isn't meant to be supported 🙂

**Environment**

* React-admin version: 3.17.1
* React version: 17.0.2
* Browser: Chrome 92 on Windows 10

djhi:
Fixes #6531

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
@@ -59,7 +59,7 @@ export const useReferenceArrayInputController = (
     // only the missing references when the input value changes
     const inputValue = useRef(input.value);
     const [idsToFetch, setIdsToFetch] = useState(input.value);
-    const [idsToGetFromStore, setIdsToGetFromStore] = useState([]);
+    const [idsToGetFromStore, setIdsToGetFromStore] = useState(EmptyArray);
     const referenceRecordsFromStore = useSelector((state: ReduxState) =>
         idsToGetFromStore.map(id => state.admin.resources[reference].data[id])
     );
@@ -121,18 +121,23 @@ export const useReferenceArrayInputController = (
     const form = useForm();
     const onSelect = useCallback(
         (newIds: Identifier[]) => {
+            // This could happen when user unselect all items using the datagrid for instance
+            if (newIds.length === 0) {
+                form.change(input.name, EmptyArray);
+                return;
+            }
+
             const newValue = new Set(input.value);
             newIds.forEach(newId => {
                 newValue.add(newId);
             });
-
             form.change(input.name, Array.from(newValue));
         },
-        [form, input.name, input.value]
+        [form, input.value, input.name]
     );
 
     const onUnselectItems = useCallback(() => {
-        form.change(input.name, []);
+        form.change(input.name, EmptyArray);
     }, [form, input.name]);
 
     const onToggleItem = useCallback(
@@ -244,7 +249,7 @@ export const useReferenceArrayInputController = (
         data: referenceRecordsFetched,
         loaded,
         refetch: refetchGetMany,
-    } = useGetMany(reference, idsToFetch || []);
+    } = useGetMany(reference, idsToFetch || EmptyArray);
 
     const referenceRecords = referenceRecordsFetched
         ? referenceRecordsFetched.concat(referenceRecordsFromStore)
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -121,14 +121,9 @@ export const useReferenceArrayInputController = (
     const form = useForm();
     const onSelect = useCallback(
         (newIds: Identifier[]) => {
-            const newValue = new Set(input.value);
-            newIds.forEach(newId => {
-                newValue.add(newId);
-            });
-
-            form.change(input.name, Array.from(newValue));
+            form.change(input.name, newIds);
```
fzaninotto, line 124: why did you remove the set? it's used to avoid selecting the same thing twice

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -316,7 +311,7 @@ export const useReferenceArrayInputController = (
         perPage,
         refetch,
         resource,
-        selectedIds: input.value,
+        selectedIds: input.value || [],
```
fzaninotto, line 314: the [] should be extracted to be a true constant

fzaninotto (changes requested):

fzaninotto:
Also, could you reverse the TypeScript version bump? We've already done that in `next` and this may create a conflict

examples/no-code/package.json
```diff
@@ -1,20 +1,20 @@
...
+    },
+    "dependencies": {
+        "ra-data-local-storage": "^3.14.0",
+        "ra-no-code": "^3.15.0",
+        "react": "^17.0.0",
+        "react-dom": "^17.0.0"
+    },
+    "devDependencies": {
+        "@vitejs/plugin-react-refresh": "^1.3.1",
+        "typescript": "^4.0.2",
```
fzaninotto, line 17: Ther is still a useless change here

examples/simple/package.json
```diff
@@ -28,7 +28,7 @@
         "ignore-not-found-export-plugin": "^1.0.1",
         "react-app-polyfill": "^1.0.4",
         "style-loader": "~1.2.1",
-        "typescript": "^4.1.3",
+        "typescript": "^4.0.2",
```
fzaninotto, line 31: and here

yarn.lock
```diff
@@ -3734,20 +3734,22 @@
...
     "@types/prop-types" "*"
-    csstype "^2.2.0"
+    "@types/scheduler" "*"
+    csstype "^3.0.2"
 
-"@types/react@^16.9.56":
-  version "16.9.56"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-16.9.56.tgz#ea25847b53c5bec064933095fc366b1462e2adf0"
-  integrity sha512-gIkl4J44G/qxbuC6r2Xh+D3CGZpJ+NdWTItAPmZbR5mUS+JQ8Zvzpl0ea5qT/ZT3ZNTUcDKUVqV3xBE8wv/DyQ==
+"@types/react@^16", "@types/react@^16.9.13", "@types/react@^16.9.56":
```
fzaninotto, line 3746: there should be no change in the lock file

fzaninotto (changes requested):

fzaninotto (approved):

PR #6594, opened 2021-09-17
SimpleList: Add support for custom url

### LINKED ISSUE #6592: Allowing custom link in SimpleList
Hi!

On the application I currently working on, I would like to redirect my user to `/#/reports/X/details` instead of `/#/reports/X` or `/#/reports/X/show` when I click on one of the SimpleList line.

However, whereas the SimpleList allows to provide a custom function in `linkType`, and as it's name suggests, it only allow the user to return a "link type" (show or edit) as returning value.

It would be cool to be able to provide any link we want. (or any action we want, just like the `rowClick` function from RA.Datagrid.

Note: Unfortunately, the `linkToRecord` function is called without the type arg here (https://github.com/marmelab/react-admin/blob/c0cf7d626e44c29f9a1e1a2eab7b640aa6027c97/packages/ra-ui-materialui/src/list/SimpleList.tsx#L285). We concat the "/show" at the end.

Thanks by advance

djhi:
Closes #6592

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList.tsx
+++ b/packages/ra-ui-materialui/src/list/SimpleList.tsx
@@ -8,6 +8,7 @@ import {
     ListItem,
     ListItemAvatar,
     ListItemIcon,
+    ListItemProps,
     ListItemSecondaryAction,
     ListItemText,
 } from '@material-ui/core';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/SimpleList.tsx
```diff
@@ -271,35 +280,52 @@ const LinkOrNot = (props: LinkOrNotProps) => {
         id,
         children,
         record,
+        resource,
     } = props;
     const classes = useLinkOrNotStyles({ classes: classesOverride });
-    const link =
-        typeof linkType === 'function' ? linkType(record, id) : linkType;
+    const [effect, setEffect] = useState<string>(() =>
+        linkType === 'edit' || linkType === true
```
fzaninotto, line 287: if linktype was false, there was no link. Your change breaks that.

packages/ra-ui-materialui/src/list/SimpleList.tsx
```diff
@@ -271,35 +280,52 @@ const LinkOrNot = (props: LinkOrNotProps) => {
...
-            className={classes.link}
-        >
+    useEffect(() => {
+        if (typeof linkType !== 'function') {
+            return;
+        }
+        const getEffect = async () => {
+            if (typeof linkType === 'function') {
+                setEffect(
+                    await linkType(record, id, basePath || `/${resource}`)
```
fzaninotto, line 299: why assume that this is async? It's not the case in Datagrid and it complexifies everything.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/SimpleList.tsx
djhi: https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx#L113

packages/ra-ui-materialui/src/list/SimpleList.tsx
djhi: Indeed, I'll fix it

djhi:
I also noticed the `SimpleList` might render a `li` item inside another `li` item which is of course wrong. I'll fix it as well

packages/ra-ui-materialui/src/list/SimpleList.tsx
fzaninotto: Then why don't you do it in a click handler as in the datagrid row?

djhi: Wanted to keep the link

packages/ra-ui-materialui/src/list/SimpleList.tsx
```diff
@@ -131,7 +131,8 @@ const SimpleList = <RecordType extends Record = Record>(
                                 record={data[id]}
                             >
                                 <ListItem
-                                    button={!!linkType as any}
+                                    // Ensure we don't have li items inside our own LI at L126
+                                    component="div"
```
fzaninotto, line 135: Could you revrt this one? This change removes the hand mouse pointer on list items, and I think that the UX is more important than the DOM validation.

fzaninotto (approved):

PR #6596, opened 2021-09-20
Added expandedPanel to Datagrid styles

mjomble:
I have a project where I needed to use expandable Datagrid rows and have a different background color on the expanded panel row that appears.

For this purpose, it would be very useful to be able to override the CSS class of these rows, which this PR enables.

Not quite sure about the name `expandedPanel`. I also thought about `expandedRow`, but that seemed ambiguous as it could also mean the original always-visible row the user clicked on in order to make the second row appear.

Also, while styling the row element was necessary in my case, other people may need to style the cell instead. Maybe we should make both overridable, but I didn't do that yet.


fzaninotto:
Nice! As this is a new feature, could you please open a PR against the `next` branch instead?

mjomble:
Closing in favor of #6606 targeting the `next` branch

Issue #6597, opened 2021-09-20
[ra-test] No way to use custom dataProvider as shown in the examples


DjebbZ:
In the [ra-test](https://github.com/marmelab/react-admin/tree/master/packages/ra-test#testing-permissions) README, the last examples declare a custom dataProvider but then never use them. And `<TestContext />` don't provide a way to pass a custom dataProvider (I've read the code).

Is this an unimplemented possibility ? A bug in the documentation ? Something missing ?

djhi:
This is indeed a documentation issue. `TestContext` does not even wrap anything with a DataProvider context. You have to do it yourself. Until we fix the documentation and/or provide more tools around this, you can refer to some our tests. For instance: https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/useListController.spec.tsx#L59

DjebbZ:
Thanks for the pointer !

It kinda works because we now move the `loaded`/`loading` booleans correctly, but the `total` received inside our list is still `undefined`. I need to wrap my head around what's happening in `useGetMainList`.

WiXSL:
Fixed by #6600

PR #6600, opened 2021-09-20
Fix examples of ra-test with fake dataProviders

DjebbZ:
You helped me in https://github.com/marmelab/react-admin/issues/6597, so I figured I could make the docs better for future me and everyone.

I haven't tested the code to see if it works. Also, a few notes:
- contrarily to some test I've seen in the react-admin code base, I didn't need the `await act(async () => await new Promise(r => setTimeout(r)));` dance. It seems that simply `await`ing was enough to follow properly the event loop.
- I've simplified the `jest.fn().resolve` since the examples weren't using the mocked function for some assertions
- I've replace `queryBy` with `findBy` which is the asynchronous equivalent. In my own code, it's required or the rendered content is the inital one before the dataProvider was called.

Feel free to adjust the code before merging, there's no guarantee it's correct in the context of the examples.

### DIFF
--- a/packages/ra-test/README.md
+++ b/packages/ra-test/README.md
@@ -142,7 +142,7 @@ Here is an example with Jest and TestingLibrary, which is testing the [`UserShow
 // UserShow.spec.js
 import * as React from "react";
 import { render } from '@testing-library/react';
-import { Tab, TextField } from 'react-admin';
+import { Tab, TextField, DataProviderContext } from 'react-admin';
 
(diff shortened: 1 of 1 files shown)


DjebbZ:
Furthermore, `.resolve()` doesn't exist on `jest.fn()`.

fzaninotto (approved):

fzaninotto:
Excellent, thanks!

DjebbZ:
You're welcome!

PR #6605, opened 2021-09-21
Support custom DataProvider type with useDataProvider

djhi:
image


packages/ra-core/src/dataProvider/useQueryWithStore.ts
```diff
@@ -201,6 +204,7 @@ export const useQueryWithStore = <State extends ReduxState = ReduxState>(
             queriesThisTick[requestSignature] = new Promise<PartialQueryState>(
                 resolve => {
                     dataProvider[type](resource, payload, options)
+                        // @ts-ignore
```
djhi, line 207: There's one thing I wasn't able to address: ensuring that a dataProvider method always returns a Promise. Hence the ignore

fzaninotto (approved):

PR #6606, opened 2021-09-22
Added expandedPanel to Datagrid styles

mjomble:
Same as #6596, but targeting the `next` branch instead


mjomble:
@djhi Should I reopen the original PR that was targeting master?

djhi:
I just did. Thanks for the reminder!

PR #6613, opened 2021-09-23
[RFR] Fix redirection after login behaviour when using fetch actions

BrunoGodefroy:
This pull requests aims at fixing the redirect after login behaviour when using a fetch action such as `fetchError`.

At the moment `fetchError` always logs the user out independently of the `authProvider.checkError` result. And the redirection after that logout is broken since the `nextPathname` is incorrectly fetched from the redux store.

I have tested the changes with different scenarios on the simple example and could not see any problem.

### DIFF
--- a/packages/ra-core/src/actions/fetchActions.ts
+++ b/packages/ra-core/src/actions/fetchActions.ts
@@ -18,9 +18,13 @@ export const FETCH_ERROR = 'RA/FETCH_ERROR';
 
 export interface FetchErrorAction {
     readonly type: typeof FETCH_ERROR;
+    readonly error: any;
 }
 
-export const fetchError = (): FetchErrorAction => ({ type: FETCH_ERROR });
+export const fetchError = (error: any): FetchErrorAction => ({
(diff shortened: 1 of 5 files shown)


WiXSL:
I don't really understand the problem you are trying to solve.
Could you please provide a code-sandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple) reproducing the problem

fzaninotto:
The redirection caused by the API failure isn't normal. The content of the state.nextPathname shouldn't be an object but a string. 

To understand the problem, we really need a reproduction. Could you please for [the Simple example CodeSandbox](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple) so as to show this strange LOCATION_CHANGE action?

yelworc:
Thanks for the quick response and the confirmation that `nextPathname` should be a string, Francois.
I'll see if I can come up with a codesandbox repro, but since this is by no means a critical problem for me (just an occasional minor annoyance), I can't make any promises ;)

PR #6616, opened 2021-09-25
Fix refresh view on undoable bulk actions failure

### LINKED ISSUE #6598: BulkDeleteButton doesn't do `refresh()` on failure
**What you were expecting:**
After deleting any items in `<List />` by bulk actions, when those items are unsuccessfully deleted on backend they should be restored on List view.

**What happened instead:**

Unsuccessfully deleted items are gone, and user needs to refresh whole page to see that those items are still in List.

**Steps to reproduce:**
try to delete some post from BulkActions - after error deleted elements aren't restored
https://codesandbox.io/s/react-admin-issue-bulkdeletebutton-2qezd?file=/src/dataProvider.js

**Related code:**
https://codesandbox.io/s/react-admin-issue-bulkdeletebutton-2qezd?file=/src/dataProvider.js

**Other information:**
IMO there is missing `refresh()` in `onFailure` function in `BulkDeleteButton`

**Environment**

* React-admin version: ^3.9.0
* React version: ^16.9.0
* Browser: chrome

WiXSL:
Fixes #6598

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx
@@ -60,7 +60,7 @@ const BulkDeleteWithUndoButton = (props: BulkDeleteWithUndoButtonProps) => {
             unselectAll(resource);
             refresh();
         },
-        onFailure: error =>
+        onFailure: error => {
             notify(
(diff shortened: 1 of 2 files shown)


djhi (approved):

djhi:
Thanks!

PR #6617, opened 2021-09-25
Fix remove `FC` usage from `ArrayInput` component

WiXSL:


WiXSL:
Is this ok to be merged?

fzaninotto:
I'm a bit unsure about these FC removal PRs as we've had regressions (refs #6585). To be sure that they don't break anything, we need to build all the packages and all the demos. This takes a lot of time to test them, for a low value, so we don't consider these PRs as first priority. 

We'll merge it in the future, but please don't be in a hurry ;)

WiXSL:
Of course, I understand, I didn't mean to hurry you

fzaninotto (approved):

Issue #5197, opened 2020-08-27
Ability to override default onFailure function within useShowController


panfiva:
**Is your feature request related to a problem? Please describe.**
We are building a "Show" view that should be displaying different error components when API calls fail based on error status code: "Not found" (when receive 404) , "Access denied" (when receive 403), or "Request failed" (error with any other status)

Current implementation of useShowController will just display ra.notification.item_doesnt_exist message and redirect back to list.

We would like to be able to provide a custom onFailure function to useShowController to be able to override notifications and redirection, and also to return error object in the controller so we can status code to pick appropriate component to render.

**Describe the solution you'd like**
I would like to change the current implementation
```js

// current
const useShowController = (props: ShowProps): ShowControllerProps => {
    useCheckMinimumRequiredProps('Show', ['basePath', 'resource'], props);
    const { basePath, id, resource } = props;
    const translate = useTranslate();
    const notify = useNotify();
    const redirect = useRedirect();
    const refresh = useRefresh();
    const version = useVersion();
    const { data: record, loading, loaded } = useGetOne(resource, id, {
        action: CRUD_GET_ONE,
        onFailure: () => {
            notify('ra.notification.item_doesnt_exist', 'warning');
            redirect('list', basePath);
            refresh();
        },
    });

    const resourceName = translate(`resources.${resource}.name`, {
        smart_count: 2,
        _: inflection.humanize(inflection.pluralize(resource)),
    });
    const defaultTitle = translate('ra.page.list', {
        name: resourceName,
    });

   return {
        loading,
        loaded,
        defaultTitle,
        resource,
        basePath,
        record,
        version,
    };
```

to the following:
```js
// proposed
const useShowController = (props: ShowProps): ShowControllerProps => {
    useCheckMinimumRequiredProps('Show', ['basePath', 'resource'], props);
    const { basePath, id, resource } = props;
    const translate = useTranslate();
    const notify = useNotify();
    const redirect = useRedirect();
    const refresh = useRefresh();
    const version = useVersion();

  // moved
   const resourceName = translate(`resources.${resource}.name`, {
        smart_count: 2,
        _: inflection.humanize(inflection.pluralize(resource)),
    });

    // moved
    const defaultTitle = translate('ra.page.list', {
        name: resourceName,
    });

   // new code
    const onFailureCb =  props.onFailure ||
      ((props) => {
        notify('ra.notification.item_doesnt_exist', 'warning')
        redirect('list', props.basePath)
        refresh()
      });

[...]

PR #6625, opened 2021-09-28
override useShowController onFailure behavior and return error object

panfiva:
See https://github.com/marmelab/react-admin/issues/5197

### DIFF
--- a/docs/Show.md
+++ b/docs/Show.md
@@ -148,7 +148,7 @@ const PostShow = props => (
 ```
 {% endraw %}
 
-You can use the `useRecordContext` hook to display non-editable details about the current record in the aside component:
+You can use the `useRecordContext` hook to display non-editable details about the current record in the aside component: 
 
 {% raw %}
 ```jsx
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Hi and thanks for taking the time to open a PR!

Although adding the `onFailure` is ok, it's a new feature so your PR should target the `next` branch. However, please revert the addition of the `error` returned by `useGetOne` and `useShowController`. Indeed, overriding the `onFailure` should be enough as it will called with the error as its first argument even though we don't use it in our default handler.

Finally, can you please revert unrelated changes in the documentation?

Thanks!

panfiva:
@djhi ,

The reason i added "error" property is to be able to support the following usecase:
```js
const showController = useShowController<T>({
    resource,  basePath, id,     
    onFailure: () => {}, // prevent redirect to list and user notification on errors
  })

const { loaded, record, error } = showController

if (error) return <Error error={error} />
else if (loaded) return <Render record={record} />
else return <Loading />

```

This way, I do not have to create custom `onFailure` function that uses `useState` to render `<Error>` component

panfiva:
@djhi ,

can we commit with `error`, since it is a very useful feature to have?

djhi:
Thanks! You can do that in the same PR if you want.

djhi:
Thanks for starting the work on this! I finished it in another PR as we want to release `3.19`

PR #6635, opened 2021-09-30
Fix custom menu

### LINKED ISSUE #6630: Custom menu doesn't collapse
I'm trying to add a custom menu following the documentation here: https://marmelab.com/react-admin/Theming.html#using-a-custom-menu. The menu gets displayed ok but when it is collapsed, the text stays visible behind the main content

Open menu: 
image

Closed menu with text visible behind main content:
image

Codesandbox example:
https://codesandbox.io/s/youthful-fog-h607k?file=/src/App.js

I would expect the collapsing of the menu to be handled by the MenuItemLink component as described here: https://marmelab.com/react-admin/Buttons.html#menuitemlink

I'm not sure what causes the current beahviour

djhi:
Closes #6630

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: "Theming"
+title: 'Theming'
 ---
 
 # Theming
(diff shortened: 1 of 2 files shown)


fzaninotto:
please revert the formatting changes

fzaninotto (approved):

PR #6638, opened 2021-09-30
Fix export TranslationMessages `StringMap` type

WiXSL:
Because of the missing `StringMap` type you get a type error trying to use `TranslationMessages`.
You can see this in the `simple` project


djhi (approved):
Thanks!

PR #6646, opened 2021-10-04
Feature: Blur input on suggestion create

andrico1234:
- Checks to see if the filtervalue matches one of the choices and removes the 'create' button if it does
- Blurs the autocomplete input after creating a new suggestion


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -221,6 +221,16 @@ export const AutocompleteInput = (props: AutocompleteInputProps) => {
         [input.value, getSuggestionFromValue]
     );
 
+    const doesQueryMatchSuggestion = useMemo(() => {
+        if (typeof optionText !== 'string') {
+            return false;
+        }
```
djhi, line 227: I think we should leverage `matchSuggestion`. Extract from the documentation:

> `matchSuggestion`: Required if `optionText` is a React element. Function returning a boolean indicating whether a choice matches the filter. `(filter, choice) => boolean`

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: If `optionText` is not a React element but a function, we should call it and apply the same logic to its result (check if it is an element or a string and call `matchSuggestion` if needed?)

andrico1234: Great, i've updated the PR. I've added videos for all scenarios, but please note that using `matchSuggestion` seemed to break the `useSuggestion` behaviour. Is this expected, or did I do something wrong?

I've left notes in the MR description

andrico1234: One of the E2E tests failed, which flagged the issue I mentioned in my previous comment. I've resolved the issue, and the tests pass!

PR #6719, opened 2021-10-22
Fix warn when unsaved changes

### LINKED ISSUE #6678: Input resets to defaultValue when empty
**What you were expecting:**

The input field should remain empty when the characters are cleared.

**What happened instead:**

As the characters are cleared from the input, it resets back to whatever `defaultValue` is set.
defaultValueBug

Tried on both `TextInput` and `NumberInput`

**Steps to reproduce:**
- add an `TextInput` component
- set `defaultValue` to "default text"
- render in simple form
- press backspace until the value is cleared

**Related code:**

```jsx
          <NumberInput min={0} source="max_users" defaultValue={10} />
          <TextInput
            source="test_source"
            label="Text Input"
            defaultValue="default text"
          />
```

**Other information:**

**Environment**

* React-admin version: 3.18.3
* React version: 17.0.2
* Browser:  Version 94.0.4606.81 (Official Build) (64-bit)

djhi:
Closes #6678.

## Tasks

- [x] Add unit tests for form subpaths

## Description

The linked issue above was introduced by #6272 that fixed the issue ##6198 introduced by #6005.

However, all those workarounds were in fact necessary because of our implementation for warning users about potential data loss  if they tried to leave a page with a form with modified fields. Indeed, the `useWarnWhenUnsavedChanges` was saving the form changes when the form was unmounted because a location change happened. If users cancelled the navigation, then the hook would renavigate to the previous location and reapplied the saved changes.

However, re-navigating to the previous location means we refetched the data, reapplied initial values for the loaded record (in case of edit) then reapplied the saved changes. All this lead to several workarounds to manage dirty states that introduced more problems.

This PR uses the history API to trigger the warning. This is better because it happens **before** the location change so we don't have to save changes and reapply them. This allows us to remove the workarounds.

### DIFF
--- a/packages/ra-core/src/form/FormWithRedirect.spec.tsx
+++ b/packages/ra-core/src/form/FormWithRedirect.spec.tsx
@@ -14,7 +14,7 @@ describe('FormWithRedirect', () => {
 
     it('Does not make the form dirty when reinitialized from a record', () => {
         const renderProp = jest.fn(() => (
-            <Input source="name" defaultValue="Bar" />
+            <Input source="name" initialValue="Bar" />
         ));
         const { getByDisplayValue, rerender } = renderWithRedux(
             <FormWithRedirect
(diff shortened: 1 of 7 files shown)


fzaninotto (approved):

GTuooa:
react-router v6 removed useHistory  history.block is not work

PR #6742, opened 2021-11-02
Fix List actions examples

erakli:
There were examples with `cloneElement()` function that had been used for creation of "Add Filters" button. While this might be ok for other cases these examples haven't work locally.

This PR simplifies examples by adding pure `FilterButton` component.


WiXSL:
This should fix #6777

djhi (approved):
Thanks!

PR #6743, opened 2021-11-02
Fix Labeled puts label in wrong position

fzaninotto:
## Problem

Following the MUI migration, labels aren't properly positioned in show views, are truncated. 

## Before

image

## After

image

### DIFF
--- a/packages/ra-ui-materialui/src/detail/Tab.tsx
+++ b/packages/ra-ui-materialui/src/detail/Tab.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { isValidElement, ReactElement, ReactNode } from 'react';
-import PropTypes, { any } from 'prop-types';
+import PropTypes from 'prop-types';
 import { Link, useLocation } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


fzaninotto:
My 2c: `<Labeled>` shouldn't be a component for inputs, but for fields. As such, support for errors and required attributes should be ropped IMHO.

packages/ra-ui-materialui/src/detail/Tab.tsx
```diff
@@ -132,6 +134,7 @@ Tab.propTypes = {
     icon: PropTypes.element,
     label: PropTypes.string.isRequired,
     path: PropTypes.string,
+    spacing: any,
```
djhi, line 137: Why not `PropTypes.any`?

fzaninotto: The question is, really: how did it compile?

Fixed in the next commit.

djhi (changes requested):
There is a eslint warning left

djhi (approved):

PR #6745, opened 2021-11-04
Add label field controller

kevinmamaqi:
It is necessary to have the label parameter to not throw an error on typescript as it is necessary in cases like this:
```
<ReferenceFieldController
        basePath="/"
        reference="friends"
        resource="/"
        source="friendsId"
        link={false}
        // @ts-ignore <-- needed as typescript throws an error
        label="Family member"
      >
        {({ referenceRecord, ...props }) => {
          return (
            <ReferenceField
              resource="/"
              reference="family"
              source="familyId"
              record={referenceRecord || undefined}
              link={false}
              addLabel={false}
            >
              <TextField source="name" />
            </ReferenceField>
          );
        }}
      </ReferenceFieldController>
```


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6753, opened 2021-11-05
[Doc] Explain how to use dataProvider hooks in TestContext

fzaninotto:
It's very hard to test a component that uses the dataProvider hooks with TestContext, because of a Gotcha about Resource registration that no one knows about.

The first step is to document it. The second step is to move away from Redux for the dataProvider optimistic strategy (to be done in the future).

### DIFF
--- a/packages/ra-core/src/core/CoreAdminContext.tsx
+++ b/packages/ra-core/src/core/CoreAdminContext.tsx
@@ -82,11 +82,8 @@ React-admin requires a valid dataProvider function to work.`);
     const [store] = useState(() =>
         !reduxIsAlreadyInitialized
             ? createAdminStore({
-                  authProvider: finalAuthProvider,
                   customReducers,
-                  dataProvider: finalDataProvider,
                   initialState,
-                  history: finalHistory,
               })
             : undefined
     );
--- a/packages/ra-core/src/core/createAdminStore.ts
+++ b/packages/ra-core/src/core/createAdminStore.ts
@@ -1,12 +1,6 @@
 import { createStore, StoreEnhancer } from 'redux';
-import { History } from 'history';
 
-import {
-    AuthProvider,
-    DataProvider,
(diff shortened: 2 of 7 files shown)


packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -82,11 +82,8 @@ React-admin requires a valid dataProvider function to work.`);
     const [store] = useState(() =>
         !reduxIsAlreadyInitialized
             ? createAdminStore({
-                  authProvider: finalAuthProvider,
```
fzaninotto, line 85: these weren't used anyway

packages/ra-core/src/core/createAdminStore.ts
```diff
@@ -15,23 +9,12 @@ interface Window {
...
-    dataProvider: DataProvider;
-    history: History;
-    authProvider?: AuthProvider;
     customReducers?: any;
-    i18nProvider?: I18nProvider;
     initialState?: InitialState;
-    locale?: string;
 }
 
-export default ({
```
fzaninotto, line 27: while I was debugging the data provide rhooks issue, I messed with the Redux store, and noticed that this method signature was oudated and in dire need for a cleanup

packages/ra-test/src/TestContext.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Book = ({ id }) => {
+    const { data, loaded } = useGetOne('books', id);
+    return loaded ? <span>{data.title}</span> : <span>loading</span>;
+};
+
+export const WithDataProvider = () => (
+    <TestContext
+        enableReducers={true}
+        // FIXME: Resources must be initialized to allow dataProvider hooks to work
```
fzaninotto, line 26: That's the heart of the problem. To be fixed by moving away from Redux for query cache

PR #6758, opened 2021-11-08
Replace Labeled with FieldWithLabel in show views

fzaninotto:
`<FieldWithLabel>` adds a label on top of its child if the child exposes a `label` prop, or if it's a field with a `source` prop. It's used in Show layouts, to display both a field and its label. 

## Usage

`<FieldWithLabel>` is mostly an internal component - you don't need it if you're using `<SimpleShowLayout>` or `<TabbedshowLayout>`. But, in a custom layout, if you want to display a label on top of a field value, just wrap the Field component with `<FieldWithLabel>`. 

```jsx
import { Show, FieldWithLabel, TextField } from 'react-admin';
import { Card, Stack } from '@mui/material';

const BookShow = () => (
   <Show>
      <Card>
         <Stack>
            <FieldWithLabel>
               <TextField source="title" />
            </FieldWithLabel>
         </Stack>
     </Car>
  </Show>
);
```

### DIFF
--- a/cypress/support/ShowPage.js
+++ b/cypress/support/ShowPage.js
@@ -1,7 +1,7 @@
 export default (url, initialField = 'title') => ({
     elements: {
         body: 'body',
-        field: name => `.ra-field-${name} > div > span`,
+        field: name => `.ra-field-${name} > p > span`,
         fields: `.ra-field`,
         snackbar: 'div[role="alertdialog"]',
         tabs: `.show-tab`,
--- a/cypress/support/ShowPage.js
+++ b/cypress/support/ShowPage.js
@@ -1,7 +1,7 @@
 export default (url, initialField = 'title') => ({
     elements: {
         body: 'body',
-        field: name => `.ra-field-${name} > div > div > span`,
+        field: name => `.ra-field-${name} > div > span`,
         fields: `.ra-field`,
         snackbar: 'div[role="alertdialog"]',
         tabs: `.show-tab`,
--- a/docs/FieldWithLabel.md
+++ b/docs/FieldWithLabel.md
@@ -1,0 +1,77 @@
+---
+layout: default
+title: "FieldWithLabel"
+---
+
+# `<FieldWithLabel>`
+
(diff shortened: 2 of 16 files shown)


cypress/support/ShowPage.js
```diff
@@ -1,7 +1,7 @@
 export default (url, initialField = 'title') => ({
     elements: {
         body: 'body',
-        field: name => `.ra-field-${name} > div > div > span`,
+        field: name => `.ra-field-${name} > p > span`,
```
djhi, line 4: We really should stop this and use `@testing-library/cypress`. Not in this PR though

docs/FieldWithLabel.md
```diff
@@ -0,0 +1,77 @@
...
+     </Car>
+  </Show>
+);
+```
+
+## Label
+
+`<FieldWithLabel>` uses the humanized `source`, and renders it above the field value. So for the previous example, the label would be `"Title"`.
+
+`<FieldWithLabel>` can also use an explicit `label` prop:
```
djhi, line 35: Might be worth reminding people it supports translation keys

docs/ShowGuesser.md
```diff
@@ -30,7 +30,7 @@ It also dumps the components it has guessed in the console, where you can copy i
 
 `<ShowGuesser>` doesn't expose any prop, as it's not designed to be customized. You should replace it with a custom `<Show>` component as soon as you need to customize it.
 
-React-admin provides guessers for the `List` view (`ListGuesser`), the `Edit` view (`EditGuesser`), and the `Show` view (`ShowGuesser`).
+React-admin provides guessers for the `List` view ([`<ListGuesser>`](./List.md#the-listguesser-component)), the `Edit` view ([`<EditGuesser>`](./CreateEdit.md#the-editguesser-component)), and the `Show` view (`ShowGuesser`).
```
djhi, line 33: Missing the `Show` and `ShowGuesser`

djhi (changes requested):
Great! There are a few warnings left though

docs/ShowGuesser.md
fzaninotto: Bur we are in the ShowGuesser doc page...

djhi (approved):

PR #6759, opened 2021-11-09
Remove props injection in edit related components

djhi:
Applies solution from #6746 to Edit related components


packages/ra-core/src/core/Resource.tsx
```diff
@@ -92,32 +92,14 @@ const ResourceRoutes = (props: ResourceProps) => {
...
+                        </Route>
                     )}
                     {edit && (
-                        <Route
-                            path={`${basePath}/:id`}
-                            render={routeProps => (
-                                <WithPermissions
-                                    component={edit}
-                                    basePath={basePath}
-                                    id={decodeURIComponent(
```
fzaninotto, line 112: I think we lost something here: what happens if the identifier contains URI encodable characters? We should add a unit test for that.

fzaninotto (changes requested):
Great!

fzaninotto (approved):

PR #6760, opened 2021-11-09
Fix Don't show a warning if only one argument is passed to `useNotify`

WiXSL:
This PR avoids throwing a warning if only a message is used in `useNotify` callback call, since everything else is optional.

```js
const  notify = useNotify();

// Warning: This way of calling useNotify callback is deprecated. Please use the new syntax passing notify("[Your message]", { ...restOfArguments })
notify('Showing a simple message');

### DIFF
--- a/packages/ra-core/src/sideEffect/useNotify.ts
+++ b/packages/ra-core/src/sideEffect/useNotify.ts
@@ -52,9 +52,11 @@ const useNotify = () => {
                         }
                     )
                 );
-            } else {
+            } else if (type) {
                 const { type: messageType, ...options } = type;
                 dispatch(showNotification(message, messageType, options));
+            } else {
+                dispatch(showNotification(message));
             }
         },
         [dispatch]
--- a/packages/ra-core/src/sideEffect/useNotify.ts
+++ b/packages/ra-core/src/sideEffect/useNotify.ts
@@ -37,7 +37,7 @@ const useNotify = () => {
         ) => {
             if (typeof type === 'string' || !type) {
                 warning(
-                    true,
+                    typeof type === 'string',
                     'This way of calling useNotify callback is deprecated. Please use the new syntax passing notify("[Your message]", { ...restOfArguments })'
                 );
                 dispatch(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/ra-core/src/sideEffect/useNotify.ts [hunk 1]
```diff
@@ -37,7 +37,7 @@ const useNotify = () => {
         ) => {
             if (typeof type === 'string' || !type) {
                 warning(
-                    true,
+                    typeof type === 'string',
```
djhi, line 40: I don't get this fix. If you call `notify` with only one argument then `type` isn't a string as it would be undefined, right?

WiXSL: Sure, since `type` will be `undefined`, the warning will not show, since now the truthy condition is set to be thrown when `type` is passed as a string

djhi: Yes, but we check this condition at L38 already

djhi: If you call `notify` with only one argument (the message), then the condition at L38 is false and we won't each the warning, right?

WiXSL: My idea is to not show a warning if `type` prop is not passed. Do you have any suggestions to improve this?

djhi: np :)

WiXSL: I was incorrectly trying to avoid warnings from within react-admin tests, but that is another pr

djhi: Yes, we have to migrate all usages of `useNotify` in our tests

WiXSL: Hold on,  if I call `notify` with only one argument (the message), the condition in line 38 will be true, because `!type` is true, and I will see a warning, right?

djhi: Right, let's just remove the second condition then

WiXSL: Ok

djhi (approved):

djhi:
Thanks!

Issue #6573, opened 2021-09-10
`parse` function not returning data correctly for `DateInput` component


FernandoKGA:
**What you were expecting:**

While trying to use the `DateInput` component, I observed that it was returning an empty string for the date, however I needed it to return `null` so I could send it through the `dataProvider`. Then I used the `parse` property of `DateInput` to make it return `null` when the field were empty.

**What happened instead:**

When using the parsing function that returns the data I wanted (`null` when the field is empty), it is returning an empty string `""` to the field  `rejected_at`.

**Steps to reproduce:**

1. Let the console tab of the sandbox open to see logs.
2. Access `/posts/1` which is where I mocked the `rejected_at` (inside `data.tsx`) field, in the post with id 1.
3. Remove the date from the `Rejected at` field.
4. Save the post.
5. Find the log that stand for the post object (probably would be the first log) and see that the `rejected_at` field is an `""` and not `null`.

**Related code:**

* CodeSandbox: https://codesandbox.io/s/sweet-mountain-0q3j4?file=/src/posts/PostEdit.tsx

Parsing function located at `PostEdit.tsx`:
```js
const dateParser = (v) => {
  // v is a string of "YYYY-MM-DD" format
  const match = /(\d{4})-(\d{2})-(\d{2})/.exec(v);

  if (match !== null) {
    const d = new Date(match[1], parseInt(match[2], 10) - 1, match[3]);
    if (isNaN(d.valueOf())) return;
    return d;
  } else {
    return null;
  }
};
```

**Other information:**

I managed to bypass the problem using it along with `format` property and passing the original default code for the `format` function (`format = getStringFromDate` at the parameters) from `DateInput.tsx`, but instead of returning `""` at the first `if`, I have changed it to return `value`. I passed it to the `format` function and then it returned the `null` value that I needed.

File `ra-ui-materialui/src/input/DateInput.tsx`:
```js
const convertDateToString = (value) => {
  if (!(value instanceof Date) || isNaN(value.getDate())) return "";
  const pad = "00";
  const yyyy = value.getFullYear().toString();
  const MM = (value.getMonth() + 1).toString();
  const dd = value.getDate().toString();
  return `${yyyy}-${(pad + MM).slice(-2)}-${(pad + dd).slice(-2)}`;
};

const dateRegex = /^\d{4}-\d{2}-\d{2}$/;

const getStringFromDate = (value) => {
  // null, undefined and empty string values should not go through dateFormatter
[...]

Issue #6459, opened 2021-07-22
Bug: DateInput & ArrayInput - formatOnBlur triggers undefined ref in react-final-form


SchaeVHS:
Dear Team,
many thanks for react-admin and your work!

**Version**
"react": "^16.7.0",
"react-admin": "^3.17.0",
"react-scripts": "4.0.3",

**Code Basis of bug**
```
<ArrayInput source="positions" >
                    <SimpleIterator > 
                        <DateInput source="delivery_date"/>
                        <DateInput source="payment_date"/>
                    </SimpleIterator>
  </ArrayInput>
```

**Problem**
If you have multiple array elements and delete an array element in the middle by clicking on the delete button,
the EDIT form is not submitted/saved when clicking on the EDIT save button.

**Observation**
The browser console reports
![grafik](https://user-images.githubusercontent.com/84381106/126605021-c57d0349-11bd-4907-ab32-4fbb4719a54e.png)

Looking at the code it is that following *react-final-form* code which is triggered:
https://github.com/final-form/react-final-form/blob/2d9e7129f40efd39450c1d581956e8cfce17a20a/src/useField.js#L68

This only happens in react-admin for datefields, as datefields set "formatOnBlur":
https://github.com/marmelab/react-admin/blob/cec6894bfc2a978357a46cc0b3b4ef372644cdeb/packages/ra-ui-materialui/src/input/DateInput.tsx#L68

**Summary**
I think it is a mixed problem:

1. React final form doesn't check the return value of getFielState()
2. react-admin SimpleFormIterator somehow deletes the fields in a way, that react-final-form still wants to iterate over it,
although they are deleted?

I would really appreciate a fix, I have no idea how to proceed. With your help I can also offer to fix it myself,
but I'm lost. Is it react-final-form or is it react-admin which causes this?

Regards,
Daniel

djhi:
Thanks for reporting this. As I can't reproduce this issue on our simple example (see the posts edit which has an ArrayInput with DateInputs), please provide a sample application showing the issue by forking the following CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple).

SchaeVHS:
Many Many thanks for your template! This helped me alot!

https://codesandbox.io/s/infallible-goldberg-4pgb7?file=/src/posts/PostEdit.tsx

Steps to reproduce
- Go to posts
- Select the only post
- Delete the middle backlink
[...]

PR #6763, opened 2021-11-09
Fix dateinput break simpleformiterator

### LINKED ISSUE #6584: Problem in saving form when remove a dateInput in ArrayInput / SimpleFormIterator
**What you were expecting:**

I should be able to save a form after removing dateInput items (with - button) in SimpleFormItterator

**What happened instead:**

image

**Steps to reproduce:**

-> We have an arrayInput of dateInput, with two fields (with a remove button). 
-> We remove the first item of the arrayInput. 
-> But, when we try to save the global form, we have an error (see the previous screenshot).

**Related code:**

Main form : 
```
<Edit title={<PostTitle/>} {...props} >  
  <TabbedForm {...props} validate={formValidator}>  
    <FormTab label="Closing days">  
      <ExceptionalClosingDay/>  
    </FormTab>  
  </TabbedForm>  
</Edit>  
```

ExceptionalClosingDay component : 
```
<ArrayInput source={props.arraySource}>
  <SimpleFormIterator>
    <DateInput label={props.dateInputLabel} source="" parse={dateParser}/>
  </SimpleFormIterator>
</ArrayInput>
```

**Other information:**

We investigated the problem. In simpleFormIterator component, we have two arrays ("ids" and "fields"). When we remove a field, we call removeField on simpleFormIterator. We splice from the index one element, from ids current (the indexes of existing fields elements), then we remove the field from fieldArray. Next, we update ids current with the new length of fields array, but we observe that fields length has not changed before updating ids arrays. So, when we save, we have ids table with more elements in ids array than fields array.
[Code](https://github.com/marmelab/react-admin/blob/5c65a4fd3a0fe849c5d97892113a3d4bb6d970cd/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx#L73)

Example : 
ids => [0|1|2]
fields => [0|1|2]
When we remove the second element (index 1) before saving:
ids => [0|2]
fields => [0|1|2] (doesn't change when debugging)
When we update ids array, we use [nextId](https://github.com/marmelab/react-admin/blob/5c65a4fd3a0fe849c5d97892113a3d4bb6d970cd/packages/ra-ui-materialui/src/form/SimpleFormIterator.tsx#L57) const (fields array length) but fields array length doesn't change. So, ids array contains [0|1|2] instead of [0|1]
After removing the field, we have:
ids => [0|1|2]
fields => [0|1]
So, when we save, we have problem with field of index 2 (it doesn't exist). We can't save.

**Environment**

* React-admin version: 3.16.3
* React version: 17.0.2
* Browser: Chrome (v92.0.4515.159)

djhi:
Fixes #6584, #6459, #6573

We introduced the `formatOnBlur` on `DateInput` to fix issues with Safari but this was the cause of the SimpleFormIterator issue. I checked with BrowserStack that the DateInput was still working on the latest safari.

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateInput.tsx
+++ b/packages/ra-ui-materialui/src/input/DateInput.tsx
@@ -87,7 +87,6 @@ const DateInput = ({
     const { id, input, isRequired, meta } = useInput({
         defaultValue,
         format,
-        formatOnBlur: true,
         initialValue,
         name,
         onBlur,


fzaninotto:
You've broken the add / remove animations in the Post backlinks edition

djhi:
It was already broken

PR #6766, opened 2021-11-10
Remove props injection list

djhi:
Follow #6765 

- [x] Remove props injection
- [x] Update examples
- [x] Update documentation


fzaninotto (approved):

UPGRADE.md
```diff
@@ -74,6 +74,17 @@ If you need to access a route parameter, use react-router's `useParams` hook ins
 };
 ```
 
+We used to inject a `syncWithLocation` prop set to `true` to the `List` components used in a `Resource`. As we removed the props injection, we also inverted this prop and renamed it `disableSyncWithLocation`. This means you now have to explicitly opt out the synchronization of the list parameters with the browser location:
```
fzaninotto, line 77: This isn't very clear. I'm putting myself in the shows of someone who has never seen this prop and I wonder if I need to add it or not...

fzaninotto (changes requested):
Huge work, judos for that!

UPGRADE.md
djhi: Better now?

UPGRADE.md
```diff
@@ -74,6 +74,19 @@ If you need to access a route parameter, use react-router's `useParams` hook ins
 };
 ```
 
+We used to inject a `syncWithLocation` prop set to `true` to the `<List>` components used in a `<Resource>`. This instructed the `<List>` to synchronize its parameters (such as pagination, sorting and filters) with the browser location.
+
+As we removed the props injection, we enabled this synchronization by default for all `<List>`, used in a `<Resource>` or not. As a consequence, we also inverted this prop and renamed it `disableSyncWithLocation`. This means you now have to explicitly opt out the synchronization of the list parameters with the browser location, for example because you have multiple `<List>` inside a dashboard:
```
fzaninotto, line 79:
```suggestion
As we removed the props injection, we enabled this synchronization by default for all `<List>`, used in a `<Resource>` or not. As a consequence, we also inverted this prop and renamed it `disableSyncWithLocation`. This doesn't change anything if you only used `<List>` components inside `<Resource list>`. 

However, if you had multiple `<List>` components inside used a single page, or if you used `<List>` outside of a `<Resource>`, you now have to explicitly opt out the synchronization of the list parameters with the browser location:
```

fzaninotto (changes requested):

fzaninotto (approved):

PR #6768, opened 2021-11-10
Fix notify function when no type is passed

### LINKED ISSUE #6767: Can't call notify function without `type` argument
**What you were expecting:**
Be able to not pass optional argument `type` to `useNotify` returned function

**What happened instead:**
Code breaks because destructuring `undefined` throws an error.

**Steps to reproduce:**
```
const notify = useNotify();
notify();
```

**Environment**

* React-admin version: 3.19.0 and I think will happen on 3.19.1 too.

nidebo:
#6713 was fixing this problem but #6760 makes it happen again

Close #6767

### DIFF
--- a/packages/ra-core/src/sideEffect/useNotify.ts
+++ b/packages/ra-core/src/sideEffect/useNotify.ts
@@ -52,9 +52,11 @@ const useNotify = () => {
                         }
                     )
                 );
-            } else {
+            } else if (type) {
                 const { type: messageType, ...options } = type;
                 dispatch(showNotification(message, messageType, options));
+            } else {
+                dispatch(showNotification(message));
             }
         },
         [dispatch]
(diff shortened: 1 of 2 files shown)


WiXSL:
I've had added the exact change in #6760, but reverted for some reason, It seams ok.

packages/ra-core/src/sideEffect/useNotify.ts
```diff
@@ -52,9 +52,11 @@ const useNotify = () => {
                         }
                     )
                 );
-            } else {
+            } else if (type) {
                 const { type: messageType, ...options } = type;
                 dispatch(showNotification(message, messageType, options));
+            } else {
+                dispatch(showNotification(message));
```
djhi, line 59: Thanks!

```suggestion
            } else {
                const { type: messageType, ...options } = type || {};
                dispatch(showNotification(message, messageType, options));
            }
```

djhi (changes requested):

packages/ra-core/src/sideEffect/useNotify.ts
nidebo: Applied, thanks!

djhi (approved):
Thanks!

PR #6778, opened 2021-11-13
[DOC] Add GeoServer data provider

sergioedo:
I would like to share with the react-admin community a data provider for [GeoServer](http://geoserver.org/). This allows you to manage spatial data (geometry features) through the standard WFS API offered by GeoServer. I think it can be useful in managing entities with location data (points, lines, polygons...).


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6779, opened 2021-11-14
[BC Break] Introduce react-query in useGetOne

fzaninotto:
Using react-query will allow use to throw away a low of the code wrapping the dataProvider to manage optimistic rendering, and open new possibilities like infinite lists.

But this will be a long job, so I figured we'd better do it little by little.

So here is my plan:

1. Add a react-query client in the admin by default
2. update useXXX dataProviders one by one to call the dataProvider via react-query's useQuery, *without removing the redux layer*
3. Update the components relying on each hook 
4. Remove the reducers related to data fetching
5. ?%*
6. Profit

I started with the easiest hook, `useGetOne`, and wit hthe components depending on it, mostly `<Show>` and `<Edit>` page components.

- [x] [BC Break] `useGetOne` now uses react-query's `useQuery` (it's a breaking change for integration tests, which must be wrapped inside a `<QueryClientProvider>`)
- [x] [BC Break] `useGetOne` now returns `useQuery` params (i.e. `isLoading` instead of `loading`)
- [x] [BC Break] Rename `<show onFailure>` to `<Show queryOptions={{ onFailure }}>`
- [x] Fix tests
- [x] Update documentation
- [x] Write upgrade guide

The big breaking changes are temporary: 

- the Edit and Show pages are no longer initialised from the List data. But once we rewrite `useGetList` the same way, it should solve the problem.
- The undoable mode is broken, in that sense that a locally applied mutation is immediately replaced by the (unmodified) data fetched from the dataProvider. But this mod will need a totally different implementation.


packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -78,7 +82,7 @@ describe('<Edit />', () => {
     });
 
     describe('mutationMode prop', () => {
-        it('should be undoable by default', async () => {
+        it.skip('should be undoable by default', async () => {
```
fzaninotto, line 85: I'm not sure about the future of our mutation modes in v4, so I'm skipping these for now as they don't pass anymore

djhi:
I think we'll have to revisit this hook to set its initial data after migrating the `useGetList` hook.

fzaninotto:
> I think we'll have to revisit this hook to set its initial data after migrating the `useGetList` hook.

indeed

djhi (approved):
:fire:

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -188,52 +184,13 @@ const useDataProvider = <
                         logoutIfAccessDenied,
                         allArguments,
                     };
-                    if (isOptimistic) {
```
fzaninotto, line 191: I had to remove that piece of logic, and it breaks undoable mode because read queries now execute immediately after an optimistic mutation and ovewrite the local mutation. 

But we had no tests for that, and I'm not yet sure how to do undoable mode with react-query. One think is sure: it won't be in useDataProvider.

djhi (approved):

PR #6854, opened 2021-11-15
Refactor error boundaries

djhi:
There's a downside. We used to display the error **inside** our layout. It's now **outside** it. I may revisit this later to make the integration nicer but this is on the critical path for react-router-v6 as there's no `withRouter` HOC anymore

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -259,6 +259,36 @@ export const PostShow = () => (
 );
 ```
 
+## Error component signature has changed
+
+The `Error` component you can provide to the `<Layout>` used to receive two props: `error` and `errorInfo`. As we now use [react-error-boundary](https://github.com/bvaughn/react-error-boundary), this component now receive the two following props:
+
+- `error`: the error object.
+- `resetErrorBoundary`: a function you can call to reset the error boundary state.
+
+It is now the `Error` component responsability to call the `resetErrorBoundary` function whenever the browser location changes:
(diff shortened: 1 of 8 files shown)


UPGRADE.md
```diff
@@ -259,6 +259,36 @@ export const PostShow = () => (
...
+- `error`: the error object.
+- `resetErrorBoundary`: a function you can call to reset the error boundary state.
+
+It is now the `Error` component responsability to call the `resetErrorBoundary` function whenever the browser location changes:
+
+```jsx
+import { useRef } from 'react';
+import { useLocation } from 'react-router';
+
+export const Error = (props: ErrorProps) => {
```
WiXSL, line 275: JS => No types allowed

WiXSL (changes requested):

WiXSL (approved):

docs/Theming.md
```diff
@@ -1035,12 +1035,23 @@ import Button from '@material-ui/core/Button';
 import ErrorIcon from '@material-ui/icons/Report';
 import History from '@material-ui/icons/History';
 import { Title, useTranslate } from 'react-admin';
+import { useLocation } from 'react-router';
 
 const MyError = ({
     error,
-    errorInfo,
```
fzaninotto, line 1041: errorInfo was used inside the component JSX, how do we get it now?

PR #6856, opened 2021-11-15
Domain availability form

Oltda:
I have added a component with a form that fetches data from a SOAP API to check the availability of a domain name

### DIFF
--- a/packages/react-admin/src/CheckDomain.css
+++ b/packages/react-admin/src/CheckDomain.css
@@ -1,0 +1,49 @@
+.form-wrapper {
+  position: relative;
+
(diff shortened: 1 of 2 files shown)


fzaninotto:
Hi @Oltda, and thanks for your PR. 

I can't merge that PR for several reasons:

- `master` is feature freeze. We only push bugfixes to that branch, New features should come in `next`
- Your PR has no test and no documentation
- Your PR doesn't explain the problem it's supposed to solve
- I don't think that the feature you add should be in react-admin, as it's not a common need of B2B apps and it can be done in userland.

PR #6873, opened 2021-11-17
Migrate to React router v6

djhi:
- [x] Migrate to react-router v6
- [x] Fix tabbed layout and form 
- [x] Bring back custom routes (find a solution for noLayout routes)
- [x] Update types
- [x] Update examples
- [x] Fix tests
- [x] Update docs
- [x] Update Upgrade guide
- [x] Add stories
- [x] More tests on custom routes


packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -65,14 +65,33 @@ React-admin requires a valid dataProvider function to work.`);
             ? convertLegacyDataProvider(dataProvider)
             : dataProvider;
 
-    const finalHistory = history || createHashHistory();
+    let historyRef = React.useRef<History>();
```
fzaninotto, line 68: I think you've lost the ability to pass a custom history.

packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -65,14 +65,33 @@ React-admin requires a valid dataProvider function to work.`);
...
+        historyRef.current = createHashHistory({ window });
+    }
+
+    let finalHistory = historyRef.current;
+    let [historyState, setHistoryState] = React.useState({
+        action: finalHistory.action,
+        location: finalHistory.location,
+    });
+
+    React.useLayoutEffect(() => finalHistory.listen(setHistoryState), [
```
fzaninotto, line 79: I don't understand why you need that state. Can't you pass only the history?

packages/ra-core/src/core/RoutesWithLayout.tsx
```diff
@@ -20,60 +25,124 @@ export interface RoutesWithLayoutProps {
...
+        children,
+        customRoutes,
+        dashboard,
+        title,
+    } = props;
     const childrenAsArray = React.Children.toArray(children);
+    const registerResource = useRegisterResource();
+    const resources = createResourcesFromChildren(children);
+
+    resources.forEach(resource => {
```
fzaninotto, line 110: This should be in a useeffect

packages/ra-core/src/core/useRegisterResource.ts
```diff
@@ -0,0 +1,19 @@
+import { useDispatch, useSelector } from 'react-redux';
```
fzaninotto, line 1: 👍 will be easier to move that logic out of the Redux store now that it's behind a custom hook

packages/ra-core/src/core/useScrollToTop.tsx
```diff
@@ -23,18 +23,10 @@ import { useHistory } from 'react-router-dom';
  * );
  */
 export const useScrollToTop = () => {
-    const history = useHistory<{ _scrollToTop?: boolean }>();
-    useEffect(
-        () =>
-            history.listen((location, action) => {
-                if (
-                    action !== 'POP' &&
```
fzaninotto, line 31: This was done to preserve scroll on back button, why did you remove it?

packages/ra-core/src/core/useScrollToTop.tsx
```diff
@@ -23,18 +23,10 @@ import { useHistory } from 'react-router-dom';
...
-                }
-            }),
-        [history]
-    );
+    const location = useLocation();
+    useEffect(() => {
+        if (location.state?._scrollToTop && typeof window != 'undefined') {
+            window.scrollTo(0, 0);
+        }
+    }, [location]);
```
fzaninotto, line 31: this is less effective than the previous system, which only registered the listener once. Is it no longer possible?

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -98,8 +97,13 @@ export const MenuItemLink = forwardRef((props: MenuItemLinkProps, ref) => {
     const renderMenuItem = () => {
         return (
             <StyledMenuItem
-                className={classnames(MenuItemLinkClasses.root, className)}
-                activeClassName={MenuItemLinkClasses.active}
+                // @ts-ignore
+                className={({ isActive }) => {
+                    console.log({ to: rest, isActive });
```
fzaninotto, line 102: console.log spotted

packages/ra-ui-materialui/src/layout/Menu.tsx
```diff
@@ -25,10 +25,8 @@ export const Menu = (props: MenuProps) => {
                     .map(resource => (
                         <MenuItemLink
                             key={resource.name}
-                            to={{
-                                pathname: `/${resource.name}`,
-                                state: { _scrollToTop: true },
-                            }}
+                            to={`/${resource.name}`}
+                            state={{ _scrollToTop: true }}
```
fzaninotto, line 29: I'm surprised this works

fzaninotto (commented):
Partial review

packages/ra-core/src/core/CoreAdminContext.tsx
djhi: I just copied the code from react-router. See https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L133 & https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L172

packages/ra-core/src/core/useScrollToTop.tsx
djhi: There's no way to retrieve the current history anymore

packages/ra-ui-materialui/src/layout/Menu.tsx
djhi: It does

packages/ra-core/src/core/useScrollToTop.tsx
djhi: I don't get it, there's no listener at all. We just react to location changes

UPGRADE.md
```diff
@@ -1,5 +1,113 @@
...
+                </CustomRoutes>
+            </>
+        )}
+    </Admin>
+)
+```
+
+## React Router Upgraded to v6
+
+This should be mostly transparent or you unless:
```
WiXSL, line 105: This should be mostly transparent **for** you unless:

WiXSL (changes requested):

examples/simple/src/index.tsx
```diff
@@ -23,25 +23,43 @@ render(
             i18nProvider={i18nProvider}
             title="Example Admin"
             layout={Layout}
-            customRoutes={[
-                <RouteWithoutLayout
-                    exact
+        >
+            <CustomRoutes noLayout>
```
WiXSL, line 27: Running the `simple` project, If you :

1. Remove the `noLayout` prop here
2. Logout
3. At the Login page, manually add `/custom` in the browser's address bar

Result: You get access to the project.

Is this expected?

docs/Admin.md
```diff
@@ -664,6 +579,81 @@ function AsyncResources() {
 }
 ```
 
+## Adding Custom Pages
+
+To register your own routes, pass one or several `<CustomRoutes>` elements as children of `<Admin>`. Declare as many [react-router-dom](https://reacttraining.com/react-router/web/guides/quick-start) `<Route>` as you want inside them:
```
WiXSL, line 584: (https://reacttraining.com/react-router/web/guides/quick-start) leads to v5 website
did you mean something like (https://reactrouter.com/docs/en/v6) ?

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -40,7 +39,9 @@ const useLogout = (): Logout => {
      * To avoid that, we read the location directly from history which is mutable.
```
WiXSL, line 39: This comment is still valid?

WiXSL (changes requested):

packages/ra-core/src/core/useRegisterResource.ts
```diff
@@ -0,0 +1,23 @@
...
+import isEqual from 'lodash/isEqual';
+
+import { registerResource } from '../actions';
+import { ReduxState, ResourceDefinition } from '../types';
+
+export const useRegisterResource = () => {
+    const dispatch = useDispatch();
+    const knowResources = useSelector<ReduxState>(
```
WiXSL, line 9: `knowResources` => `knownResources`

WiXSL (changes requested):

examples/simple/src/index.tsx
djhi: Yes because this route is not protected automatically. I may had some stories to showcase the possibilities, including protected routes without layout

packages/ra-core/src/auth/Authenticated.spec.tsx
```diff
@@ -1,24 +1,41 @@
...
+                    dataProvider={() => Promise.resolve()}
+                >
+                    <Authenticated>
+                        <Foo />
+                    </Authenticated>
+                </CoreAdminContext>
+            </Provider>
         );
-        expect(queryByText('Foo')).not.toBeNull();
+        expect(screen.queryByText('Foo')).not.toBeNull();
```
fzaninotto, line 38: The rest of this test is wrong. We're testing that redux isn't called... which tests a side effect (the logout notification) of what we actually want to test. 

In fact, the test should be as follows:

1. render the component
2. see that it's content is rendered
3. trigger the authProvider.checkauth response
4. see that the conetn is still rendered

packages/ra-core/src/auth/Authenticated.spec.tsx
```diff
@@ -32,36 +49,53 @@ describe('<Authenticated>', () => {
...
+            );
+        };
+        render(
+            <Provider store={store}>
+                <CoreAdminContext
+                    authProvider={authProvider}
+                    dataProvider={() => Promise.resolve()}
+                    history={history}
+                >
+                    <Routes>
```
fzaninotto, line 72: why are you rendering routes here and not in the previous test? Fon consistency, the two tests should have the same children IMO

packages/ra-core/src/auth/Authenticated.spec.tsx
```diff
@@ -32,36 +49,53 @@ describe('<Authenticated>', () => {
...
             );
             expect(dispatch.mock.calls[1][0]).toEqual({
                 type: 'RA/CLEAR_STATE',
             });
-            expect(history.location.pathname).toEqual('/login');
-            expect(history.location.state).toEqual({
-                nextPathname: '/',
-                nextSearch: '',
-            });
+            expect(screen.getByLabelText('nextPathname').innerHTML).toEqual(
```
fzaninotto, line 96: this test sin't readable. I understand you're testing that we are on the login page, but it's really not apparent here.

packages/ra-core/src/auth/useAuthenticated.spec.tsx
```diff
@@ -40,27 +46,52 @@ describe('useAuthenticated', () => {
...
+                    dataProvider={() => Promise.resolve()}
+                >
+                    <Authenticated>
+                        <Foo />
+                    </Authenticated>
+                </CoreAdminContext>
+            </Provider>
         );
-        expect(queryByText('Foo')).toBeDefined();
+        expect(screen.queryByText('Foo')).toBeDefined();
```
fzaninotto, line 94: same as for the Authenticated component

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -27,20 +26,8 @@ import { LocationDescriptorObject } from 'history';
...
-     * would be rebuilt each time the user changes location. Consequently, that
-     * would force a rerender of all components using this hook upon navigation
-     * (CoreAdminRouter for example).
-     *
-     * To avoid that, we read the location directly from history which is mutable.
-     * See: https://reacttraining.com/react-router/web/api/history/history-is-mutable
-     */
-    const history = useHistory();
+    const navigate = useNavigate();
+    const location = useLocation();
```
fzaninotto, line 30: Major perf problem, as locaiton is in the useCallback signature, the logout callback will change each time the user navigates, and the Logout button will be rerendered.

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -27,20 +26,8 @@ import { LocationDescriptorObject } from 'history';
 const useLogout = (): Logout => {
     const authProvider = useAuthProvider();
     const dispatch = useDispatch();
-
-    /**
-     * We need the current location to pass in the router state
-     * so that the login hook knows where to redirect to as next route after login.
-     *
-     * But if we used useLocation to get it, the logout function
```
fzaninotto, line 35: This is still valid, I don't unbderstrand why you removed it

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -59,40 +46,50 @@ const useLogout = (): Logout => {
...
                     redirectToCurrentLocationAfterLogin &&
-                    history.location &&
-                    history.location.pathname
+                    location &&
+                    location.pathname
                 ) {
-                    newLocation.state = {
-                        nextPathname: history.location.pathname,
-                        nextSearch: history.location.search,
+                    newLocationOptions = {
```
fzaninotto, line 59: There can be other options in the location. You're erasing them. setting only the state didn't have this drawback.

packages/ra-core/src/auth/useLogout.ts
```diff
@@ -59,40 +46,50 @@ const useLogout = (): Logout => {
...
+                    state: {
+                        nextPathname: location && location.pathname,
+                    },
+                }
+            );
             dispatch(clearState());
             return Promise.resolve();
         },
-        [dispatch, history]
+        [dispatch, location, navigate]
```
fzaninotto, line 92: same problem as above

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -79,56 +76,67 @@ describe('useLogoutIfAccessDenied', () => {
...
-            <AuthContext.Provider value={authProvider}>
-                <TestComponent />
-            </AuthContext.Provider>
-        );
+        render(<Route path="/" element={<TestComponent />} />, {
+            wrapper: TestWrapper,
+        });
 
         await waitFor(() => {
             expect(authProvider.logout).toHaveBeenCalledTimes(0);
```
fzaninotto, line 84: In fact, we should check that checkauth has been called before

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -79,56 +76,67 @@ describe('useLogoutIfAccessDenied', () => {
...
     });
 
     it('should not log out if passed an error that does not make the authProvider throw', async () => {
-        const { queryByText } = renderInRouter(
-            <AuthContext.Provider value={authProvider}>
-                <TestComponent error={new Error()} />
-            </AuthContext.Provider>
+        render(
+            <Route path="/" element={<TestComponent error={new Error()} />} />,
+            {
```
fzaninotto, line 93: is prettier tired or did you intentionally make it less readable? ;)

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -142,76 +150,118 @@ describe('useLogoutIfAccessDenied', () => {
...
-            </AuthContext.Provider>
+        render(
+            <MemoryRouter>
+                <AuthContext.Provider value={delayedAuthProvider}>
+                    <Routes>
+                        <Route
+                            path="/"
+                            element={
+                                <>
+                                    <TestComponent />
```
fzaninotto, line 161: There were 2 forbidden component in the test you replaced

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -79,56 +76,67 @@ describe('useLogoutIfAccessDenied', () => {
     });
 
     it('should not logout if passed no error', async () => {
-        const { queryByText } = renderInRouter(
-            <AuthContext.Provider value={authProvider}>
-                <TestComponent />
-            </AuthContext.Provider>
-        );
+        render(<Route path="/" element={<TestComponent />} />, {
```
fzaninotto, line 79: Why do we need a route?

fzaninotto (changes requested):
Partial review (1/3rd of files reviewed)

examples/crm/src/deals/DealList.tsx
```diff
@@ -45,12 +49,8 @@ export const DealList = () => {
             >
                 <DealListContent />
             </List>
-            <Route path="/deals/create">
-                {({ match }) => <DealCreate open={!!match} />}
-            </Route>
-            <Route path="/deals/:id/show">
-                {({ match }) => (!!match ? <DealShow open={!!match} /> : null)}
-            </Route>
+            <DealCreate open={!!matchCreate} />
```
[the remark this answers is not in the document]
djhi: We are but the way to do it has changed

examples/simple/src/comments/CommentEdit.tsx
```diff
@@ -112,32 +115,25 @@ const CreatePost = () => {
...
                 <Title defaultTitle={`Comment #${record ? record.id : ''}`} />
                 <div className={classes.actions}>
-                    <EditActions
-                        basePath={basePath}
-                        resource={resource}
-                        data={record}
-                        hasShow
-                        hasList
-                    />
+                    <TopToolbar>
```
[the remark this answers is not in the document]
djhi: For the cypress tests

examples/simple/src/posts/PostEdit.tsx
```diff
@@ -80,22 +81,22 @@ const CreateCategory = ({
...
     <TopToolbar>
         <CloneButton
             className="button-clone"
-            basePath={basePath}
+            basePath={`/${resource}`}
             record={data}
         />
-        {hasShow && <ShowButton basePath={basePath} record={data} />}
+        {hasShow && <ShowButton basePath={`/${resource}`} record={data} />}
+        <RaButton component={Link} label="Create post" to="/posts/create" />
```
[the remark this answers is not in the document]
djhi: I am

UPGRADE.md
```diff
@@ -1,5 +1,113 @@
 # Upgrade to 4.0
 
+## The Way To Providing Custom Routes Has Changed
```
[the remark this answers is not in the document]
djhi: That's not possible in the current implementation. That would requires us to support a `basename` prop on the `Admin` that would be passed to the `Router`

examples/demo/src/layout/Menu.tsx
```diff
@@ -81,21 +81,17 @@ const Menu = ({ dense = false }: MenuProps) => {
                 dense={dense}
             >
                 <MenuItemLink
-                    to={{
-                        pathname: '/commands',
-                        state: { _scrollToTop: true },
-                    }}
+                    to="/commands"
+                    state={{ _scrollToTop: true }}
```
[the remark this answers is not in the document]
djhi: I'm not sure

examples/demo/src/reviews/ReviewList.tsx
```diff
@@ -56,69 +56,57 @@ const ReviewList = () => {
...
+    const location = useLocation();
+    const navigate = useNavigate();
 
     const handleClose = useCallback(() => {
-        history.push('/reviews');
-    }, [history]);
+        navigate('/reviews');
+    }, [navigate]);
+
+    const match = matchPath('/reviews/:id', location.pathname);
```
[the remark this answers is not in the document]
djhi: That would be true if we had a create view for the reviews

examples/demo/src/reviews/ReviewList.tsx
```diff
@@ -56,69 +56,57 @@ const ReviewList = () => {
...
 
     return (
         <Root className={classes.root}>
-            <Route path="/reviews/:id">
```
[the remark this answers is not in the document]
djhi: See previous answser

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -1,107 +1,111 @@
...
+    const [hasFunctionChild, setHasFunctionChild] = useSafeSetState(false);
     useScrollToTop();
+
+    useEffect(() => {
+        setResources(createResourcesFromChildren(props.children));
+    }, [setResources, props.children]);
+
+    useEffect(() => {
+        resources.forEach(resource => {
+            registerResource(resource);
```
fzaninotto, line 53: if you rerender the Router, the same resource will be registered twice

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -1,107 +1,111 @@
...
+        });
+        setFirstResource(resources.length > 0 ? resources[0] : null);
+    }, [registerResource, resources, setFirstResource]);
+
+    useEffect(() => {
```
fzaninotto, line 58: This is so hard to understand. You're already calling setResources line 48, and you're doing it again here. Why? Can you refactor your code and add comments to explain it better?

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -1,107 +1,111 @@
...
-    children?: AdminChildren;
-    customRoutes?: CustomRoutes;
-    loading: LoadingComponent;
-    ready?: ComponentType;
-}
-
-type State = ResourceElement[];
-
-const CoreAdminRouter = (props: AdminRouterProps) => {
+export const CoreAdminRouter = (props: AdminRouterProps) => {
```
fzaninotto, line 29: too many useeffect in this component, I can't wrap my mind around it.

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -115,111 +119,150 @@ const CoreAdminRouter = (props: AdminRouterProps) => {
...
+            // Ignore non-elements. This allows people to more easily inline
+            // conditionals in their route config.
+            return null;
+        }
+        if (element.type === Fragment) {
+            return renderCustomRoutes(element.props.children);
+        }
+
+        if (element.type !== CustomRoutes) {
+            return null;
```
fzaninotto, line 257: you should add a comment here saying that it's likely a Resource in this case

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -115,111 +119,150 @@ const CoreAdminRouter = (props: AdminRouterProps) => {
...
+            icon: resourceElement.props.icon,
+        };
+
+        resources.push(resource);
+    });
+
+    return resources;
+};
+
+const renderCustomRoutes = (children: ReactNode) => {
```
fzaninotto, line 245: I think you should rename this finction to show that youll only render custom routes with no layout.

packages/ra-core/src/core/CoreAdminRouter.tsx
```diff
@@ -115,111 +119,150 @@ const CoreAdminRouter = (props: AdminRouterProps) => {
...
+    return Children.map(children, element => {
```
fzaninotto, line 246: A render function that returns an array? Is it legal?

packages/ra-core/src/core/useScrollToTop.tsx
fzaninotto: Then when I'm going back to a list from a show view, I don't return to the scroll position I was before? This is a big breaking change, we absolutely need that feature.

packages/ra-core/src/dataProvider/useUpdate.spec.tsx
```diff
@@ -35,12 +34,13 @@ describe('useUpdate', () => {
...
-            renderWithRedux(
-                <QueryClientProvider client={new QueryClient()}>
-                    <DataProviderContext.Provider value={dataProvider}>
-                        <Dummy />
-                    </DataProviderContext.Provider>
-                </QueryClientProvider>
+            render(
+                <CoreAdminContext
+                    dataProvider={dataProvider}
+                    queryClient={new QueryClient()}
```
fzaninotto, line 40: No need for this one. Il all the tests in this file.

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -103,9 +104,25 @@ const FormWithRedirect = ({
...
     );
 
+    const subscribers = useRef<{
+        [key: string]: FormGroupSubscriber[];
+    }>({});
+
     const formContextValue = useMemo<FormContextValue>(
         () => ({
             setOnSave,
+            subscribe: (group, subscriber) => {
```
fzaninotto, line 114: Can you please explain what you're doing in a comment?

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -115,9 +132,17 @@ const FormWithRedirect = ({
             },
             registerField: (source, group) => {
                 if (group) {
-                    const fields = new Set(formGroups.current[group] || []);
-                    fields.add(source);
-                    formGroups.current[group] = Array.from(fields);
+                    if (!(formGroups.current[group] || []).includes(source)) {
```
fzaninotto, line 135: you've added more code but not more tests, I'm a bit scared

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -67,24 +67,33 @@ export const useFormGroup = (name: string): FormGroupState => {
...
+            }
+
+            return oldState;
+        });
+    }, [form, formContext, name]);
+
     useEffect(() => {
+        // Whenever the group content changes (input are added or removed)
+        // we must update its state
+        formContext.subscribe(name, () => {
```
fzaninotto, line 91: subscribe returns an unsubscribe method that you never call. I suspect a memory leak here

packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
```diff
@@ -13,36 +13,31 @@ const useWarnWhenUnsavedChanges = (
     enable: boolean,
     formRootPathname?: string
 ) => {
-    const history = useHistory();
+    const navigator = useContext(UNSAFE_NavigationContext).navigator as History;
```
fzaninotto, line 16: We definitely a comment here, this UNSAFE stuff is super scary

packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
```diff
@@ -1,7 +1,7 @@
-import { useEffect, useRef } from 'react';
```
fzaninotto, line 1: did you backport the fix we recently did in master?

packages/ra-core/src/sideEffect/useRedirect.ts
```diff
@@ -66,15 +73,17 @@ const useRedirect = () => {
...
-                    state: { _scrollToTop: true, ...state },
-                });
+                    {
+                        state: { _scrollToTop: true, ...state },
+                    }
+                );
             }
         },
-        [dispatch, history]
+        [dispatch, location, navigate]
```
fzaninotto, line 86: not good: the return function will be different each time we navigate, forcing to rerender every component using it. Think about buttons used outside of a tabbed form that will rerender when the user switches tabs.

packages/ra-core/src/storybook/FakeBrowser.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { ReactNode, useEffect, useState } from 'react';
+import {
+    createMemoryHistory,
+    History,
+    Location,
+    createPath,
+    parsePath,
+} from 'history';
+
+export const FakeBrowserDecorator = (Story, context) => {
```
fzaninotto, line 11: could you exlain how to use it in jsDoc?

fzaninotto (changes requested):
Reviewed 2/3rds (all ra-core)

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -30,25 +31,22 @@ describe('useEditController', () => {
...
-                                <div>{record && record.title}</div>
-                            )}
-                        </EditController>
-                    </SaveContextProvider>
-                </DataProviderContext.Provider>
-            </QueryClientProvider>
+        render(
+            <CoreAdminContext
+                dataProvider={dataProvider}
+                queryClient={new QueryClient()}
```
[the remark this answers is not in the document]
djhi: The issue is that the client in `CoreAdminContext` is declared in the module, not in the component. This makes it shared accross all tests which leads to issues

packages/ra-no-code/src/ui/ResourceMenuItem.tsx
```diff
@@ -64,5 +62,5 @@ export const ResourceMenuItem = (
 };
 
 const NavLinkRef = forwardRef<HTMLAnchorElement, NavLinkProps>((props, ref) => (
-    <NavLink innerRef={ref} {...props} />
```
fzaninotto, line 67: I think this is wrong

packages/ra-test/src/TestContext.tsx
```diff
@@ -63,7 +63,19 @@ export const TestContext = (props: TestContextProps) => {
...
         customReducers = {},
     } = props;
-    const history = useRef<History>(props.history || createMemoryHistory());
+    let historyRef = React.useRef<History>();
+    if (historyRef.current == null) {
+        historyRef.current = createMemoryHistory();
+    }
+
+    let finalHistory = historyRef.current;
+    let [historyState, setHistoryState] = React.useState({
```
fzaninotto, line 72: You're doing the same strange stuff as in CoreAdminContext. Why not reuse that component?

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -42,9 +48,11 @@ export const Button = (props: ButtonProps) => {
         label,
         color = 'primary',
         size = 'small',
+        to: locationDescriptor,
```
fzaninotto, line 51: I don't understand this change. Why do you need to decomponse the to prop now?

packages/ra-core/src/auth/Authenticated.spec.tsx
```diff
@@ -1,24 +1,41 @@
...
+            getPermissions: () => Promise.reject('bad method'),
+        };
+        const store = createAdminStore();
+        const dispatch = jest.spyOn(store, 'dispatch');
+
+        render(
+            <Provider store={store}>
+                <CoreAdminContext
+                    authProvider={authProvider}
+                    dataProvider={() => Promise.resolve()}
```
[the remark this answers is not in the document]
djhi: I introduced a `testDataProvider` function that we can reuse everywhere we need one without any specific implementation. It accept a partial dataProvider if you need to override any of the methods

packages/ra-ui-materialui/src/detail/Show.spec.tsx
```diff
@@ -39,25 +37,27 @@ describe('<Show />', () => {
...
         render(
-            <TestContext
+            <CoreAdminContext
+                dataProvider={dataProvider}
                 history={history}
-                enableReducers
                 initialState={{
                     admin: { resources: { books: { props: {}, data: {} } } },
                 }}
+                queryClient={new QueryClient()}
```
fzaninotto, line 46: not required. Same for other tests in this file.

packages/ra-ui-materialui/src/detail/TabbedShowLayout.tsx
```diff
@@ -108,48 +105,78 @@ export const TabbedShowLayout = (props: TabbedShowLayoutProps) => {
...
-                                cloneElement(tab, {
+                                        })}
+                                    />
+                                ) : null
+                            )}
+                        </Route>
+                    </Routes>
+                ) : (
+                    <>
+                        {cloneElement(
```
fzaninotto, line 153: I thinbk you should extract that to a render function to avoid duplication

packages/ra-ui-materialui/src/detail/TabbedShowLayoutTabs.tsx
```diff
@@ -11,12 +11,11 @@ export const TabbedShowLayoutTabs = ({
...
     ...rest
 }: TabbedShowLayoutTabsProps) => {
-    const location = useLocation();
-    const match = useRouteMatch();
+    const params = useParams();
 
-    // The location pathname will contain the page path including the current tab path
-    // so we can use it as a way to determine the current tab
-    const tabValue = location.pathname;
+    // params wil include eventual parameters from the root pathname and * for the remaining part
```
fzaninotto, line 16:
```suggestion
    // params will include eventual parameters from the root pathname and '*' for the remaining part
```

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render, waitFor } from '@testing-library/react';
-import { MemoryRouter } from 'react-router-dom';
+import { render, screen, waitFor } from '@testing-library/react';
```
fzaninotto, line 3: linter warning

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: No, we're testing that the child is rendered even though we have not called the authProvider yet

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: Because in this test, we want to check we actually redirected users to the login page?

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: Any suggestion?

packages/ra-ui-materialui/src/layout/Layout.tsx
```diff
@@ -26,15 +25,15 @@ import { SkipNavigationButton } from '../button';
...
+        appBar: AppBar = DefaultAppBar,
         children,
         className,
         dashboard,
         error: errorComponent,
         logout,
-        menu = DefaultMenu,
-        notification = DefaultNotification,
-        sidebar = DefaultSidebar,
+        menu: Menu = DefaultMenu,
```
fzaninotto, line 34: 👍 much, much better like this!

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -95,12 +94,18 @@ export const MenuItemLink = forwardRef((props: MenuItemLinkProps, ref) => {
...
 
+    const match = useMatch(
+        typeof props.to === 'string' ? props.to : props.to.pathname
+    );
+
     const renderMenuItem = () => {
         return (
             <StyledMenuItem
-                className={classnames(MenuItemLinkClasses.root, className)}
-                activeClassName={MenuItemLinkClasses.active}
```
fzaninotto, line 102: The detection of active links is no longer built-in react-router?

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -166,13 +169,13 @@ const StyledMenuItem = styled(MenuItem, { name: PREFIX })(({ theme }) => ({
...
         color: theme.palette.text.primary,
     },
 
     [`& .${MenuItemLinkClasses.icon}`]: { minWidth: theme.spacing(5) },
 }));
 
-const NavLinkRef = forwardRef<HTMLAnchorElement, NavLinkProps>((props, ref) => (
-    <NavLink innerRef={ref} {...props} />
+const LinkRef = forwardRef<HTMLAnchorElement, LinkProps>((props, ref) => (
+    <Link ref={ref} {...props} />
```
fzaninotto, line 180: why did you change ref by innerRef?

fzaninotto (changes requested):
Huge work, kudos for that!

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: As discussed, I didn't change the original tests. Testing more things si out of scope for this PR

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: As discussed, I didn't change the original tests. Testing more things si out of scope for this PR

packages/ra-core/src/auth/Authenticated.spec.tsx
djhi: As discussed, I didn't change the original tests. Testing more things si out of scope for this PR

packages/ra-core/src/auth/useAuthenticated.spec.tsx
djhi: As discussed, I didn't change the original tests. Testing more things si out of scope for this PR

packages/ra-core/src/auth/useLogout.ts
djhi: Fixed using a location ref

packages/ra-core/src/auth/useLogout.ts
djhi: Reverted and updated to the new mechanism

packages/ra-core/src/auth/useLogout.ts
djhi: I don't get this. You mean `replace` and others? The `navigate` function signature isn't the same as the old history push

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
djhi: prettier

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
djhi: Because `useLogoutIfAccessDenied` calls `useNavigate` and that requires to be inside a Router

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -611,6 +666,7 @@ describe('useEditController', () => {
             data: { foo: 'bar', transformed: true },
             previousData: undefined,
         });
+        unmount();
```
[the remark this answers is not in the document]
djhi: Just forgot to remove this one

packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -70,15 +68,32 @@ React-admin requires a valid dataProvider function to work.`);
             ? convertLegacyDataProvider(dataProvider)
             : dataProvider;
 
-    const finalHistory = history || createHashHistory();
+    let historyRef = React.useRef<History>();
+    if (historyRef.current == null) {
```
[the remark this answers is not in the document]
djhi: Replaced by a useMemo

packages/ra-core/src/core/CustomRoutes.authenticated.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export default {
+    title: 'ra-core/Admin/CustomRoutes/Authenticated',
+    decorators: [FakeBrowserDecorator],
+    parameters: {
+        initialEntries: ['/authenticated'],
+    },
+};
+
+export const AuthenticatedCustomRoute = (argsOrProps, context) => {
```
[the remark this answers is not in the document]
djhi: No, stories are not components. They're functions. They receive theirs args if you specified some and a context that can be populated by decorators

packages/ra-core/src/core/RoutesWithLayout.tsx
```diff
@@ -1,80 +1,111 @@
...
-                    )}
-                />
-            ) : firstChild ? (
-                <Route
-                    exact
-                    path="/"
-                    render={() => <Redirect to={`/${firstChild.props.name}`} />}
-                />
-            ) : null}
+        if (element.type === CustomRoutes) {
```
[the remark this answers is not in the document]
djhi: We're testing a reference to a function. I don't see how it could fail.

packages/ra-core/src/core/CoreAdminRouter.tsx
djhi: Unless registerResource takes care of it, which it does

packages/ra-core/src/core/CoreAdminRouter.tsx
djhi: I'll try

packages/ra-core/src/core/RoutesWithLayout.tsx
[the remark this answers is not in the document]
djhi: Double checked. It's fine

packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
djhi: Nope, should I?

packages/ra-no-code/src/ui/ResourceMenuItem.tsx
djhi: ?

packages/ra-test/src/TestContext.tsx
djhi: Because I believe we're going to remove `ra-test`

packages/ra-ui-materialui/src/button/Button.tsx
djhi: I wanted to reduce the breaking changes for our users. The `to` prop of react-router Link has changed. It does not include `state` anymore, nor the options such as `replace`

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
djhi: Not compatible with mui styled components

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
djhi: Because they did

WiXSL (approved):

packages/ra-core/src/controller/edit/useEditController.spec.tsx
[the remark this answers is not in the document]
djhi: Fixed by using useMemo to initialize the queryclient

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
djhi: No this test is actually useless as it tests that the `TestComponent` does not call `logoutIfAccessDenied` if not passed an error

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
djhi: Fixed

packages/ra-core/src/core/CoreAdminRouter.tsx
djhi: Extracted the logic inside a custom hook and simplified it

packages/ra-core/src/core/CoreAdminRouter.tsx
djhi: Refactored since but yes, when you pass multiple children to React node, you actually pass an array of children

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: I will extract all this inside a custom hook in another PR

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: Because the behavior has not changed, hence the tests related to form groups still pass (`TabbedForm` and `TranslatableInput`)

WiXSL (commented):
This is much understandable now, great work!

examples/simple/src/comments/CommentEdit.tsx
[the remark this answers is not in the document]
fzaninotto: Then add a FIXME comment in the code to explain it. We must revisit it before the v4 stable release, as I think there is an underlying bug.

examples/simple/src/posts/PostEdit.tsx
```diff
@@ -31,6 +32,7 @@ import {
     useCreateSuggestionContext,
     EditActionsProps,
     usePermissions,
+    Button as RaButton,
```
fzaninotto, line 35: Linter warnings

packages/ra-core/src/auth/useLogout.ts
fzaninotto: you should use the loacation ref here

packages/ra-core/src/storybook/FakeBrowser.tsx
djhi: Done

examples/simple/src/comments/CommentEdit.tsx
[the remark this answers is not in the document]
djhi: There's definitely an underlying bug. When navigating away from a dirty form using the browser address bar, the navigation is prevented but without a confirmation dialog. Looking into it

packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
fzaninotto: no, let's do it later

fzaninotto (approved):

PR #6877, opened 2021-11-17
Fix include side effects in `DeleteButton` props type

### LINKED ISSUE #6735: DeleteButton props type does not include side effects functions
But both underlying buttons (`DeleteWithUndoButton` and `DeleteWithConfirmButton`) accept those props.

WiXSL:
Fixes #6735


djhi (approved):
:+1:

Issue #6699, opened 2021-10-18
[4.0.0-alpha.1] Notify component appears at the bottom left of the canvas


WiXSL:
![test](https://user-images.githubusercontent.com/8268610/137744176-3c371a9b-ee03-4d56-9874-8c2406c70ee8.png)

oguhpereira:
Well, the material-ui has a property to configure which position the snackbar should be displayed.

Now it would be interesting to understand whether you need to keep this fixed or dynamic.

![image](https://user-images.githubusercontent.com/24482087/142301400-79d7dc08-9465-4f8e-b0e1-0aee72780ca2.png)

WiXSL:
Fixed by #6878

PR #6878, opened 2021-11-18
Fix Notify component appears at the bottom left of the canvas

oguhpereira:
to fix it I'm adding the anchorOrigin property to the snackbar

image

issue #6699


packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -86,6 +86,7 @@ export const Notification = (
                     </StyledButton>
                 ) : null
             }
+            anchorOrigin={{ vertical: 'bottom', horizontal: 'center' }}
```
djhi, line 89: I suggest to make the `NotificationProps` interface extends the `SnackbarProps` one from MUI. 
Then declare the `anchorOrigin` prop at line 25 and set its default to a this value.

And by the way, can you extract the `anchorOrigin` default value as a const outside the component?

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/layout/Notification.tsx
oguhpereira: Yes of course. I think this is a good option!

I even commented on the issue

oguhpereira: I made the change https://github.com/marmelab/react-admin/pull/6878/commits/f1e0e79ce503ac0b89939bd40834122d3df06772, but it's also worth noting whether it makes sense now to add some prop in showNotification

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -22,6 +22,7 @@ export const Notification = (
         classes: classesOverride,
         type = 'info',
         autoHideDuration = 4000,
+        anchorOrigin = { vertical: 'bottom', horizontal: 'center' },
```
fzaninotto, line 25: You should create a constant outside of the component wit hthat value, otherwise the value will be recreated at each render and cause unnecessary renders of the underlying Snackbar component.

fzaninotto (changes requested):
Please document the additional prop in the commponent's JSDoc

packages/ra-ui-materialui/src/layout/Notification.tsx
oguhpereira: @fzaninotto  I'm not sure I understood your proposal, but do you expect me to remove this property from props?


```js
    const {
        classes: classesOverride,
        type = 'info',
        autoHideDuration = 4000,
        className,
        ...rest
    } = props;
...
    const anchorOrigin =  { vertical: 'bottom', horizontal: 'center' }
```

or

```js
    const {
        classes: classesOverride,
        type = 'info',
        autoHideDuration = 4000,
        className,
        anchorOrigin
        ...rest
    } = props;
...
    const { vertical, horizontal } =  anchorOrigin
```

Well, it was validating it can also be done this way 

```js
Notification.defaultProps = {
    anchorOrigin: { vertical: 'bottom', horizontal: 'center' },
};
```

this way I could even remove the prop anchorOrigin in line 90, because from what I was looking has `{...rest}`



> Please document the additional prop in the commponent's JSDoc

Where is this docs? Theming.md? If not, where should I create?

djhi: No, he meant to declare a const outside the component:

```
const defaultAnchorOrigin = { vertical: 'bottom', horizontal: 'center' }
```

and to use that as the default for the prop:
```
anchorOrigin = defaultAnchorOrigin
```

oguhpereira: Done @djhi @fzaninotto  https://github.com/marmelab/react-admin/pull/6878/commits/cfaaa01312e91590f4879b48cd6a92c2a5f1077b

Where is JSDoc?

oguhpereira:
@fzaninotto do you expect me to create documentation in the component or is there already any file for me to edit?

packages/ra-ui-materialui/src/layout/Notification.tsx
oguhpereira: I will mark it as resolved because I changed this logic https://github.com/marmelab/react-admin/pull/6878#discussion_r755164822

WiXSL:
> @fzaninotto do you expect me to create documentation in the component or is there already any file for me to edit?

There is no jsDoc for this component, I guess you would have to create it

WiXSL (approved):

WiXSL:
> Please document the additional prop in the commponent's JSDoc

I don't think it's necessary, since `anchorOrigin` is a prop from MUI's `Snackbar`

fzaninotto:
Thanks!

PR #6880, opened 2021-11-18
Fixed the width of the filter input in tutorial

### LINKED ISSUE #6700: [4.0.0-alpha.1] Tutorial: Filter component appears too narrow
filter

CoreyB26:
Use styled components to fix the width of the input. A more elegant way would be nice. Closes #6700 
Screen Shot 2021-11-17 at 11 47 42 PM

### DIFF
--- a/examples/tutorial/src/posts.js
+++ b/examples/tutorial/src/posts.js
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { styled } from '@mui/material/styles';
 import {
     Show,
     ShowButton,
@@ -13,16 +14,21 @@ import {
     TextField,
     EditButton,
     ReferenceInput,
+    SearchInput,
     SelectInput,
     SimpleForm,
     TextInput,
 } from 'react-admin';
 
+const StyledReferenceInput = styled(ReferenceInput)({
+    width: 250,
+});
+
 const postFilters = [
-    <TextInput source="q" label="Search" alwaysOn />,
-    <ReferenceInput source="userId" label="User" reference="users" allowEmpty>
+    <SearchInput source="q" alwaysOn />,
+    <StyledReferenceInput label="User" source="userId" reference="users">
         <SelectInput optionText="name" />
-    </ReferenceInput>,
+    </StyledReferenceInput>,
 ];
 
 export const PostList = props => (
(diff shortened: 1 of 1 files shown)


fzaninotto:
The bug should be fixed in the ReferenceInput IMO, not in the simple example.

wattroll:
Hi @CoreyB26. Thanks for putting effort into fixing it.

 I've run into that issue in my project and dealt with it a similar way as you did here (by setting `min-width` individually on each of my filters to a value that makes sense for that specific filter). This allowed me to proceed with development, and is an acceptable fix. It's reasonable for a project to state what is the minimal desired width of a particular element.

Unfortunately this fix doesn't address the issue at hand (#6700 "Filter component appears too narrow"). The core problem is not with particular input being too narrow, but the whole `Filter` view not taking the available space (which causes the inputs inside it to be crammed). The crammed up `ReferenceInput` in the tutorial is just a symptom of this problem.

However I think you are onto something here. It would make sense that the input components don't submit to being crammed into unreasonably tiny space. I don't know what's the best way to achieve it, but I would expect a component to try take enough space to fit the label. It's not a straightforward styling issue, but I feel that there should be a simple solution that would work well for most cases (and when it doesn't styles can be overridden). One might argue that it's responsibility of the container (e.g. `Filter`) to give enough space to its children, so maybe this behavior could be adjusted with a flag (akin to `fullWidth` in MUI).

If you want to take this further I suggest:

1. Figure out how to do what you already did here without hard-coding `min-width: 100`. Can you make it so that the `ReferenceInput` in the tutorial takes as much space as the label needs to fit no matter what that label is?
2. Move that solution into the code of `ReferenceInput` itself.
3. Find a way to apply that solution to other input components.

All of that is irrelevant to the bug described in #6700, but would be a great enhancement that will be applicable not only in filters, but everywhere where inputs are used. E.g. if there is a create form with multiple fields in a row, it's nice to have them wrap into several rows when the screen is narrow instead of becoming too small to be usable. I have such a form in my project, and have to put each input in a `Box` with an appropriate sizing/flexing properties. Even though it's the right way to get a specific inputs in the form to be just perfect for the task at hand at any screen size,  but it would be better if doing that wasn't necessary to get a usable form on the screen.

fzaninotto:
Superseded by #7152

PR #6891, opened 2021-11-22
Update useUpdate to use react-query

fzaninotto:
- [x] [BC Break] update useUpdate signature
- [x] pessimistic mode
- [x] optimistic mode
- [x] undoable mode
- [x] hook time and calltime parameters and options
- [x] Add unit tests
- [x] [BC Break] Update useEditController to use it
- [x] [BC Break] Removed deprecated `<Edit successMessage>` prop
- [x] [BC Break] Update onSuccess and onFailure callbacks props and signature for `<Edit>` and `<SaveButton>`
- [x] Fix tests
- [x] Update demos
- [x] Update documentation

It turns out react-query's handling of cache, invalidation and stale queries is completely backwards in comparison with http logic, so this was not a piece of cake. I'm not sure we end up with less code to maintain than with our current Redux implementation.

```jsx
import { useUpdate } from 'react-admin';

const IncreaseLikeButton = ({ record }) => {
    const diff = { likes: record.likes + 1 };
    const [update, { isLoading, error }] = useUpdate();
    const handleClick = () => {
        update('likes', { id: record.id, data: diff, previousData: record })
    }
    if (error) { return <p>ERROR</p>; }
    return <button disabled={isLoading} onClick={handleClick}>Like</div>;
};
```


packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -55,12 +55,12 @@ export const useEditController = <RecordType extends Record = Record>(
 ): EditControllerResult<RecordType> => {
     const {
         id: propsId,
-        successMessage,
         onSuccess,
         onFailure,
         mutationMode = 'undoable',
         transform,
         queryOptions = {},
+        mutationOptions = {},
```
fzaninotto, line 63: I need to remove onSuccess and onFailure callbacks in favor of this one

packages/ra-core/src/dataProvider/performQuery/doQuery.ts
```diff
@@ -28,19 +28,7 @@ export const doQuery = ({
     store,
     mutationMode,
 }: DoQueryParameters) => {
-    const resourceState = store.getState().admin.resources[resource];
-    if (canReplyWithCache(type, payload, resourceState)) {
```
fzaninotto, line 32: I had to disable application cache because it's incompatible with react-query. React-query gives an equivalent, albeit not controlled by the server. It's a BC break we should accept IMO, as application cache is a minor feature not widely used.

packages/ra-core/src/dataProvider/useUpdate.spec.tsx
```diff
@@ -1,158 +1,301 @@
...
-            resource: 'foo',
-            payload: {
+            renderWithRedux(
+                <QueryClientProvider client={new QueryClient()}>
+                    <DataProviderContext.Provider value={dataProvider}>
+                        <Dummy />
+                    </DataProviderContext.Provider>
+                </QueryClientProvider>
+            );
+            localUpdate({
```
fzaninotto, line 45: Mutations created by react-query's useMutation only accept one parameter. So we can't keep the previous signature

```js
update(resource, { id, data, previousdata }) 
```

which was similar to the dataProvider function. Instead, the callback signature must be

```js
mutate({ resource, id, data, previousdata }) 
```

to avoid confusion, I also changed the hook signature the same way:

```diff
-const mutation = useUpdate(resource, { id, data, previousdata }) 
+const mutation = useUpdate({ resource, id, data, previousdata }) 
```

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+        // because setQueryData doesn't accept a staletime option
+        const updatedAt =
+            mutationMode === 'undoable' ? Date.now() + 1000 * 5 : Date.now();
+        await queryClient.setQueryData(
+            [resource, 'getOne', id],
+            (old: RecordType) => ({
+                ...old,
+                ...data,
+            }),
+            { updatedAt }
```
fzaninotto, line 86: I'm not happy with this one but I didn't find any better solution

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+                    data: callTimeData,
+                    previousData: callTimePreviousData,
+                })
+                .then(({ data }) => data);
         },
-        [mutate] // eslint-disable-line react-hooks/exhaustive-deps
+        {
+            ...reactMutationOptions,
+            onMutate: async (variables: UpdateParams<RecordType>) => {
+                // Return a context object with the snapshotted value
```
fzaninotto, line 133: this comment is out of place

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+        }
+
+        const { onSuccess, onSettled, onError } = callTimeOptions;
+        const {
+            resource: callTimeResource = resource,
+            id: callTimeId = id,
+            data: callTimeData = data,
+        } = variables;
+
+        // optimistic update as documented in https://react-query.tanstack.com/guides/optimistic-updates
```
fzaninotto, line 261: The react-query doc advises to do this in onMutate, but in that case it's impossible to fire the success side effects right away. I had to use a wrapper for `mutateAsync` (and therefore `mutate`) instead.

Besides, the wrapper was necessary for the undoable mode, which shouldn't call the `mutate` callback until the action is confirmed.

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+        });
+
+        context = { previousGetOne, previousGetList };
+
+        // run the success callbacks during the next tick
+        if (onSuccess) {
+            setTimeout(
+                () =>
+                    onSuccess(
+                        queryClient.getQueryData([
```
fzaninotto, line 298: to be replaced by previousGetOne

packages/ra-core/src/dataProvider/useUpdate.spec.tsx
fzaninotto: I've changed my mind. I've broken the signature, but in a way that makes it look like the dataProvider one:

```diff
-const [update] = useUpdate('posts', id, data, previousDate, options);
+const [update] = useUpdate('posts', { id, data, previousDate }, options);
-update(resource, id, data, previousVersion, options);
-update(resource, { id, data, previousVersion }, options);
```

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +61,355 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+            );
+        }
+
+        const {
+            id: callTimeId = id,
+            data: callTimeData = data,
+        } = callTimeParams;
+
+        // optimistic update as documented in https://react-query.tanstack.com/guides/optimistic-updates
+        // except we wo it in a mutate wrapper instead of the onMutete callback
```
djhi, line 290:
```suggestion
        // except we do it in a mutate wrapper instead of the onMutate callback
```

djhi (changes requested):

fzaninotto:
And we're Ready For Review!

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+                    );
+                }
+            },
+            onSuccess: (
+                data: RecordType,
+                variables: Partial<UseUpdateParams<RecordType>> = {},
+                context: unknown
+            ) => {
+                if (mutationMode === 'pessimistic') {
+                    // Optimistically update to the new value in getOne
```
djhi, line 197: Optimistically when pessimistic?

packages/ra-core/src/dataProvider/useUpdate.ts
djhi: Not applied?

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+        variables: Partial<UseUpdateParams<RecordType>> = {},
+        callTimeOptions: MutateOptions<
+            RecordType,
+            unknown,
+            Partial<UseUpdateParams<RecordType>>,
+            unknown
+        > = {}
+    ) => {
+        if (mutationMode === 'pessimistic') {
+            return mutation.mutateAsync(variables, callTimeOptions);
```
djhi: If I understand correctly, we don't merge the declaration time options with the calltime ones?

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+        }
+
+        const { onSuccess, onSettled, onError } = callTimeOptions;
```
djhi: We ignore the declaration time options?

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -45,74 +55,319 @@ import useMutation, { MutationOptions, Mutation } from './useMutation';
...
+
+        if (mutationMode === 'optimistic') {
+            // call the mutate without success side effects
+            return mutation.mutateAsync(variables, {
+                onSettled,
+                onError,
+            });
+        } else {
+            // undoable mutation: register the mutation for later
+            undoableEventEmitter.once('end', ({ isUndo }) => {
```
djhi, line 355: How can we be sure this event was dispatched for this specific query? What if there are multiple queries sent in a short timeframe?

djhi (changes requested):
Awesome work

packages/ra-core/src/dataProvider/useUpdate.ts
fzaninotto: this is done by the mutation callback (the first argument of useMutation)

packages/ra-core/src/dataProvider/useUpdate.ts
fzaninotto: We can't. That's the way it'as always worked. 

But it's not a real problem as, in order to perform a new mutation, users have to click somewhere, which dismisses the MUI Notification and commits the initial update.

packages/ra-core/src/dataProvider/useUpdate.ts
fzaninotto: No, they're taken care of by the useMutation callback

packages/ra-core/src/dataProvider/useUpdate.ts
djhi: Ok. That's only true for our implementation of the ui though

djhi (approved):

PR #6893, opened 2021-11-22
Configure jest to avoid rebuilding packages dependencies

djhi:
Until now, when working for example on both `ra-core` and `ra-ui-materialui`, we had to stop the tests and rebuild `ra-core` for any changes to impact the `ra-ui-materialui` tests.

This is the jest equivalent of webpack aliases.

I configured it so that `ra-core` `ra-ui-materialui` and `ra-test` are aliased to their sources.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,28 @@
+const path = require('path');
+
+module.exports = {
+    globalSetup: './test-global-setup.js',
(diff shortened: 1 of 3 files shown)


fzaninotto:
Can you xplain the problem it fixes?

fzaninotto (approved):

Issue #6511, opened 2021-08-16
initialValue on BooleanInput inside ArrayInput resets value of BooleanInput on Edit


DjebbZ:
**What you were expecting:**
When editing a record that contains a `BooleanInput` inside an `ArrayInput`, the `BooleanInput` should have the last known value.

**What happened instead:**
When editing, the `BooleanInput` displays the `initialValue`. Which means that after saving the `BooleanInput` is reset to its `initialValue`.

**Steps to reproduce:**
In the CodeSandbox below :
1. Click on the SHOW of the first record in the Posts list
2. Notice that the boolean value of the first element of Backlinks array is set to `true`
3. Click on EDIT
4. Notice that the `BooleanInput` is not set to `true` but to `false`
5. Go back to SHOW without saving and notice the value is `true` so it should be true in the edit view
6. Now click on EDIT, then SAVE without changing anything
7. Notice that the boolean field is now `false` instead of `true`.

**Related code:**
https://codesandbox.io/s/brave-dhawan-l5sfk

**Other information:**

**Environment**

* React-admin version: 3.17.0
* Last version that did not exhibit the issue (if applicable): Don't know
* React version: 17
* Browser: Tested in Chrome, Firefox, Brave.
* Stack trace (in case of a JS error): No stack trace.

Issue #6900, opened 2021-11-24
Missing items in ArrayInput with SimpleIterator


horaklukas:
**What you were expecting:**
When I open posts list and then open post detail with iterated data (backlinks) using `ArrayInput` and `SimpleFormIterator`, all backlinks are visible.

**What happened instead:**
No backlink is visible. After page refresh all backlinks are back there.

**Steps to reproduce:**
1. Go to the Posts list - api return list of posts **without `backlinks`** property, because it's not needed in the list
2. Open Post with backlinks (eg. id `3`) and go to tab "Miscellaneous" - api return detail of post **containing backlinks**
3. You should see list of backlinks but none are there until you refresh the page

**Related code:**
https://codesandbox.io/s/frosty-tess-qzbxb?file=/src/posts/PostEdit.tsx

I did some debugging and found out that problem is probably inside `SimpleFormIterator` at lines 59-75. When the detail is loading and `fields` is empty, the nextId is initialized with `0` value which means `ids` become empty array. Once the detail is loaded, `nextId` nor `ids` is reevaluated and remain as it is which means `ids` is empty array, but `fields` contains backlinks

```typescript
// We need a unique id for each field for a proper enter/exit animation
    // so we keep an internal map between the field position and an auto-increment id
    const nextId = useRef(
        fields && fields.length
            ? fields.length
            : defaultValue
            ? defaultValue.length
            : 0
    );

    // We check whether we have a defaultValue (which must be an array) before checking
    // the fields prop which will always be empty for a new record.
    // Without it, our ids wouldn't match the default value and we would get key warnings
    // on the CssTransition element inside our render method
    const ids = useRef(
        nextId.current > 0 ? Array.from(Array(nextId.current).keys()) : []
    );
```

**Other information:**

I suppose it's related to these issues

* https://github.com/marmelab/react-admin/issues/5621
* https://github.com/marmelab/react-admin/issues/5289

**Environment**

* React-admin version: 3.19.2
* Last version that did not exhibit the issue (if applicable): Seems like it was working in 3.18.1
* React version: 17.0.2
* Browser: any

I know the another cause of the issue is that the entity (post in this case) has different shape for list and for detail, but it's a common use case for use, because some entities can be huge and we need only few simple fields for distinguishing them in a list.

**Is it an antipattern for RA to have different entity shapes between endpoints?**

WiXSL:
@horaklukas, thanks for the report.

Did you change anything in the code sandbox? 
[...]

PR #6908, opened 2021-11-26
Bug in nested ArrayInput with initialValue(s)

DjebbZ:
**What you were expecting:**
When I open the Edit view of a resource that has nested ArrayInput and use some `initialValue(s)` both at the form level and the field, the inputs for this part of the data shows the value from the dataProvider.

**What happened instead:**
Instead, the inputs display the initialValue at the field level, making it easy for editor that modify another part of the record to override the existing data!

**Steps to reproduce:**
1. Open the code sandbox https://codesandbox.io/s/reproduce-nested-arrayinput-default-value-bfpnr
2. Open the Edit view of the post "Accusantium2 qui nihil voluptatum" (its url in the internal browser is https://bfpnr.sse.codesandbox.io/#/posts/1/2)
3. Look at the "Related Pictures" inputs : they're all empty because the inputs have `initialValue` set to `""` but `data.tsx` (should be `data.ts` BTW) has actual data.  

**Related code:**
The sandbox https://codesandbox.io/s/reproduce-nested-arrayinput-default-value-bfpnr has all the code, and I created this pull request to highlight the small set of changes I made to reproduce the bug.

**Other information:**
It may be related to a similar issue we created related to `<BooleanIinput />` in nested ArrayInput, not sure (https://github.com/marmelab/react-admin/issues/6511)

**Environment**

* React-admin version: 3.19.0, 3.19.2
* Last version that did not exhibit the issue (if applicable): N/A, we're not sure when the problem manifested itself.
* React version: 17
* Browser: Chrome, Brave, Firefox

### DIFF
--- a/examples/simple/src/posts/PostEdit.tsx
+++ b/examples/simple/src/posts/PostEdit.tsx
@@ -248,6 +248,7 @@ const PostEdit = ({ permissions, ...props }) => {
                                             <TextInput
                                                 source={getSource('url')}
                                                 initialValue=""
+                                                label="Url"
                                             />
                                             <ArrayInput
                                                 source={getSource(
@@ -258,6 +259,7 @@ const PostEdit = ({ permissions, ...props }) => {
                                                     <TextInput
                                                         source="name"
                                                         initialValue=""
+                                                        label="Author name"
                                                     />
                                                 </SimpleFormIterator>
                                             </ArrayInput>
--- a/examples/simple/src/posts/PostEdit.tsx
+++ b/examples/simple/src/posts/PostEdit.tsx
@@ -231,12 +231,40 @@ const PostEdit = ({ permissions, ...props }) => {
                     <BooleanInput source="commentable" defaultValue />
                     <ArrayInput source="pictures">
                         <SimpleFormIterator>
-                            <TextInput source="url" initialValue="" />
-                            <ArrayInput source="metas.authors">
-                                <SimpleFormIterator>
-                                    <TextInput source="name" initialValue="" />
-                                </SimpleFormIterator>
-                            </ArrayInput>
+                            <FormDataConsumer>
+                                {({ getSource, scopedFormData }) => {
+                                    if (
+                                        getSource &&
+                                        !scopedFormData?.metas?.authors
+                                    ) {
+                                        _.assign(scopedFormData, {
+                                            metas: {
+                                                authors: [],
+                                            },
+                                        });
+                                    }
+                                    return (
+                                        <>
+                                            <TextInput
+                                                source={getSource('url')}
+                                                initialValue=""
+                                            />
+                                            <ArrayInput
+                                                source={getSource(
+                                                    'metas.authors'
+                                                )}
+                                            >
+                                                <SimpleFormIterator>
+                                                    <TextInput
+                                                        source="name"
+                                                        initialValue=""
+                                                    />
+                                                </SimpleFormIterator>
+                                            </ArrayInput>
+                                        </>
+                                    );
+                                }}
+                            </FormDataConsumer>
                         </SimpleFormIterator>
                     </ArrayInput>
                     <TextInput disabled source="views" />
--- a/examples/simple/src/posts/PostEdit.tsx
+++ b/examples/simple/src/posts/PostEdit.tsx
@@ -40,6 +40,7 @@ import {
     DialogContent,
     TextField as MuiTextField,
 } from '@material-ui/core';
+import _ from 'lodash';
 
 import PostTitle from './PostTitle';
 import TagReferenceInput from './TagReferenceInput';
@@ -105,7 +106,7 @@ const PostEdit = ({ permissions, ...props }) => {
     return (
         <Edit title={<PostTitle />} actions={<EditActions />} {...props}>
             <TabbedForm
-                initialValues={{ average_note: 0 }}
+                initialValues={_.merge({}, { average_note: 0, pictures: [] })}
                 warnWhenUnsavedChanges
             >
                 <FormTab label="post.form.summary">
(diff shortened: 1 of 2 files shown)


WiXSL:
Could you describe the bug you are trying to fix?

DjebbZ:
Done, I updated the PR (I created the PR from CodeSandbox without a description nor a proper title)

DjebbZ:
(Sorry if I created a PR instead of an issue, but the diff directly in GitHub looked like a good idea to me)

DjebbZ:
It may be related to this issue #6900, not sure though.

DjebbZ:
I tracked it down to `useInput` inside of RA's `<TextField />`, that return a value of `""` (which is the `initialValue`, not the actual value from the record) and passes it to the `ResettableTextField` underlying component. It gets this value from `useInput`, that itself uses `useField` from `react-final-form`. Maybe the problem is over there ?

djhi:
Hi @DjebbZ,
Unfortunately, your code formater settings make it very difficult to see what you actually changed. Can you revert the unnecessary changes?

examples/simple/src/posts/PostShow.tsx
```diff
@@ -1,117 +1,128 @@
...
+            <ArrayField source="pictures">
+              <Datagrid>
+                <ImageField source="url" />
+                <ArrayField source="metas.authors">
+                  <Datagrid>
+                    <TextField source="name" />
+                  </Datagrid>
+                </ArrayField>
+              </Datagrid>
+            </ArrayField>
```
DjebbZ, line 100: This is what I added to the `simple` example (+ updating react-admin in package.json)

DjebbZ:
> Hi @DjebbZ, Unfortunately, your code formater settings make it very difficult to see what you actually changed. Can you revert the unnecessary changes?

Replied via a comment in the diff, because it's too tedious to revert just the code styles. No idea why your prettier config wasn't caught by my IDE.

examples/simple/src/posts/PostEdit.tsx
```diff
@@ -1,253 +1,240 @@
...
+          <ArrayInput source="pictures">
+            <SimpleFormIterator>
+              <TextInput source="url" initialValue="" />
+              <ArrayInput source="metas.authors">
+                <SimpleFormIterator>
+                  <TextInput source="name" initialValue="" />
+                </SimpleFormIterator>
+              </ArrayInput>
+            </SimpleFormIterator>
+          </ArrayInput>
```
DjebbZ, line 218: Sorry, I also added this part (this is in fact the most important part of the bug report)

djhi:
Got it, reproduced

examples/simple/src/posts/PostEdit.tsx
```diff
@@ -1,253 +1,240 @@
...
-                        validate={required()}
-                        resettable
-                    />
-                    <CheckboxGroupInput
-                        source="notifications"
+  console.log("PostEdit", { props });
+  return (
+    <Edit title={<PostTitle />} actions={<EditActions />} {...props}>
+      <TabbedForm
+        initialValues={_.merge({}, { average_note: 0, pictures: [] })}
```
DjebbZ, line 106: I forgot to say that I also I added this part : a global object for the initialValues. Since the ArrayInput field prevents us from specifying here the shape of the `pictures` field, I had to also add `initialValue` at the field level. This global `initialValues` may be important for the bug.

djhi:
So, I tried to reproduce it locally both on `master` and the `v3.19.2` tag branch but couldn't. I can only reproduce it on codesandbox

DjebbZ:
Have you looked at my last comment ? (`initialValues` at the form level). We commented at ~the same time, maybe you missed it when trying to reproduce ?

djhi:
I did and tried in both branches

DjebbZ:
Me neither, I'm trying to figure out what's missing and will update the PR.

DjebbZ:
I managed to reproduce, that's quite a sneaky one.

In CodeSandbox, according to [the docs](https://codesandbox.io/docs/environment#container-environment), it will run the `dev` script so the `simple` example is run through webpack. You can see it clearly in the Terminal output in CodeSandbox `$ ./node_modules/.bin/webpack-dev-server --progress --color --hot --watch --mode development` (which corresponds to the `start` script in the simple example's package.json).

But when trying to reproduce locally, I use (and I'm you did too) the `make run-simple` command, which runs the `start` command of the simple example, that use `vite`, not `webpack`.

And in our application, we use `create-react-app` which uses `webpack` internally.

If you run the simple example locally with `yarn run dev` inside `examples/simple`, you'll see the problem.

DjebbZ:
I also updated the PR code with my local reproduction case, now the diff is very clean and easy to read.

DjebbZ:
Regarding `Vite` vs `Webpack`: that might mean that the way Webpack transpiles the code produces the bug, whereas with Vite it doesn't happen. I've also noticed after a quick glance that the `vite.config.js` does a few things that `webpack.config.js` doesn't do, i.e. pointing `react-admin` to the local version, not the one stored in node_modules. Maybe the bug comes from those kind of differences in the build config?

djhi:
Thanks for the details! I'll dig

DjebbZ:
Any news ? Any way I can help investigate the issue ? Any pointer welcome to help you debug this, I don't really know where to start.

djhi:
I can now reproduce it locally even with vite if I remove the aliases.

djhi:
Found it! It looks like a regression in final-form. I still have to find the exact source of this issue though.
Workaround for now, install `final-form@4.20.2` in your project

djhi:
Actually, it shouldn't have worked before. Final-form array fields are dynamic by nature and the documentation clearly state that the `initialValue` specified on a field overrides the form initial value. We use this form initial value to initialize the form with the current record. I have a fix that will be published in `3.19.3`.

> Are you capable of telling since which version of RA this regression caused the bug? I want to estimate the potential damage in our app

Nope

DjebbZ:
Thanks for detailed answer.

DjebbZ:
Yes ! So this should be in the next release, hopefully on Friday right ?

DjebbZ:
Oh didn't see 3.19.4 !

PR #6909, opened 2021-11-26
[4.0.0-alpha.1] Fix typos, lints and docs anchors references

WiXSL:


djhi (changes requested):
:heart_eyes: You missed a lint though ;)

WiXSL:
> 😍 You missed a lint though ;)

If you are referring to `RecordType` lint related error, now, I don't get it locally, plus is a type, I can't include it in the dependencies array

djhi (approved):
You're right, that's weird

PR #6916, opened 2021-11-29
Migrate useGetList to react-query

fzaninotto:
- [x] [BC Break] have useGetList use react-query (changes useGetList signature)
- [x] [BC Break] Update useListController and ListContext to use new useGetList signature removes ids, changes data from hash to array, changes loading to isLoading)
- [x] [BC Break] Remove useGetMainList and associated ids reducers
- [x] [BC Break] Remove prop injection in Edit page components
- [x] Fix components depending on ListContext (`<Datagrid>`, `<SimpleList>`, `<SingleFieldList>`, etc.)
- [x] Fix tests
- [x] Fix demos
- [x] Add upgrade guide
- [x] Update component JSDoc


fzaninotto:
I've fixed about 70 tests and I'm exhausted. That's it for today.

packages/ra-no-code/src/ui/ImportResourceDialog.tsx
```diff
@@ -52,7 +53,7 @@ export const ImportResourceDialog = (props: ImportResourceDialogProps) => {
                     if (resourceAlreadyExists) {
                         // If we imported more records for an existing resource,
                         // we must refresh the list
-                        refresh();
+                        queryClient.refetchQueries([resource, 'getList']);
```
djhi, line 56: Why?

djhi (commented):
Awesome

packages/ra-no-code/src/ui/ImportResourceDialog.tsx
fzaninotto: Because useGetList isn't affected by refresh anymore (I've removed the version trick). We need to use react-query to refresh the list of records after importing new ones.

cypress/integration/list.js
```diff
@@ -187,7 +187,9 @@ describe('List Page', () => {
             cy.contains('1-10 of 13'); // wait for data
             ListPagePosts.toggleSelectSomeItems(3);
             ListPagePosts.applyDeleteBulkAction();
-            cy.contains('1-10 of 10');
+            // FIXME once dataProvider hookls are migrated to react-query
```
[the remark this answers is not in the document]
fzaninotto: after all hooks are migrated

fzaninotto:
Switching to RFR

docs/List.md
```diff
@@ -302,12 +302,11 @@ export const PostList = () => (
 
 **Tip**: You can also disable bulk actions altogether by passing `false` to the `bulkActionButtons` prop. When using a `Datagrid` inside a `List` with disabled bulk actions, the checkboxes column won't be added.
 
-Bulk action button components receive several props allowing them to perform their job:
+Bulk action button components can use the `useListContext` to get the elements they need to perform their job:
```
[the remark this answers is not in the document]
djhi: Besides, the first `ResetViewsButton` should probably be updated as well. It still expect props

docs/List.md
```diff
@@ -361,14 +360,16 @@ But let's say you need a customized bulkAction button, here is an example levera
...
     useUpdateMany,
     useRefresh,
     useNotify,
     useUnselectAll,
 } from 'react-admin';
 import { VisibilityOff } from '@material-ui/icons';
 
-const CustomResetViewsButton = ({ selectedIds }) => {
+const CustomResetViewsButton = () => {
+    const { selectedIds } = useListContext();
```
djhi, line 372: Alignment is weird at L376

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -246,37 +266,23 @@ export const useReferenceArrayInputController = (
...
+
     const {
         data: matchingReferences,
-        ids: matchingReferencesIds,
         total,
+        error: errorGetList,
+        isLoading: isLoadingGetList,
         refetch: refetchGetMatching,
-    } = useGetMatching(
+    } = useGetList(
```
djhi, line 282: Does it mean we will remove the `getMatching` method completely?

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -42,7 +42,7 @@ export const useGetOne = <RecordType extends Record = Record>(
 ): UseGetOneHookValue<RecordType> => {
     const dataProvider = useDataProvider();
     return useQuery<RecordType, unknown, RecordType>(
-        [resource, 'getOne', id],
+        [resource, 'getOne', String(id)],
```
djhi, line 45: Out of curiosity, why do we need to stringify it?

packages/ra-no-code/src/ui/ImportResourceDialog.tsx
djhi: Ok but why did you remove the version trick? Shouldn't it still be affected by the version?

packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -22,82 +19,78 @@ import { ListActions as DefaultActions } from './ListActions';
...
                         </BulkActionsToolbar>
                     )}
                     {children &&
                         // @ts-ignore-line
                         cloneElement(Children.only(children), {
-                            ...controllerProps, // deprecated, use ListContext instead, to be removed in v4
                             hasBulkActions: bulkActionButtons !== false,
                         })}
-                    {pagination && cloneElement(pagination, listContext)}
+                    {pagination !== false && cloneElement(pagination)}
```
djhi, line 72: We don't need to clone it anymore, right?

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -103,10 +102,9 @@ export const SingleFieldList = (props: SingleFieldListProps) => {
                 }
 
                 return (
-                    <RecordContextProvider value={data[id]} key={id}>
+                    <RecordContextProvider value={record} key={record.id}>
                         {cloneElement(Children.only(children), {
-                            key: id,
-                            record: data[id],
+                            record,
```
djhi, line 107: I think we don't need to clone the children anymore

djhi (changes requested):

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
fzaninotto: There is no getMatching method in dataProvider - only getManyReference.

packages/ra-core/src/dataProvider/useGetOne.ts
fzaninotto: Because the react-query cache is type-sensitive, and sometimes the id comes as a string (e.g. when read from the URL in a Show view), sometimes as a number (e.g. when read from a Record in useGetList response).

But since you're asking the question, I realize I need to add that comment to the code.

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
djhi: True. Are we removing `useGetMatching` then? And by the way, is `getManyReference` really necessary?

docs/List.md
fzaninotto: It's GitHub's fault.

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
fzaninotto: useGetMatching is removed in this PR. 

getManyReference is stille useful for APIs with a hierarchical structure, e.g. `/posts/123/comments`

fzaninotto: Sorry, useGetMatching isn't removed yet, I'll remove it.

packages/ra-ui-materialui/src/list/ListView.tsx
fzaninotto: Right

packages/ra-no-code/src/ui/ImportResourceDialog.tsx
fzaninotto: react-query provides a way to force refetch already. We'll remove the version reducer, but first we must make views work without version.

fzaninotto:
Review applied

packages/ra-no-code/src/ui/ImportResourceDialog.tsx
djhi: Ok but then I think we should find a way to hide this behind a custom hook

fzaninotto: Yes, we'll rewrite the useRefresh hook but after this PR, which is large enough already.

djhi:
You can rebase to get the storybook

packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -142,23 +136,19 @@ export const useReferenceInputController = (
     });
 
     // add current value to possible sources
-    let finalIds: Identifier[],
-        finalData: RecordMap<Record>,
-        finalTotal: number;
-    if (!referenceRecord || possibleValuesIds.includes(input.value)) {
-        finalIds = possibleValuesIds;
+    let finalData: Record[], finalTotal: number;
+    if (!referenceRecord || possibleValuesData.includes(input.value)) {
```
wattroll, line 140: Looks like a bug to me: `input.value` is of primitive type and `possibleValuesData` is an array of objects. Won't ever match.

This solves it in a pinch:

```diff
-    if (!referenceRecord || possibleValuesData.includes(input.value)) {
+    if (!referenceRecord || possibleValuesData.find(({ id }) => id === input.value) {
```

But I lack context to confidently suggest it as a fix. So just bringing it to your attention @fzaninotto. Let me know if it's as simple as that and I can put together a pull request.

If this doesn't quite make sense I can open a proper issue, but since it's `next` and actively being worked on I thought just dropping a comment here might be nicer.

fzaninotto: You're right, this is a bug! And the fix you suggest is correct. Would you mind opening a PR?

Commenting here is fine, it gives the proper context. The best would be to comment before the PR is merged ;)

wattroll: Thanks for quick response. #6987

PR #6917, opened 2021-11-29
Add support for router basename

djhi:
- [x] Add support for router basename prop
- [x] Documentation

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -473,11 +473,15 @@ const App = () => (
 
 ## `basename`
 
-Use this prop to make all routes and links in your Admin relative to a "base" portion of the URL pathname that they all share. This is useful when including the Admin inside a larger app. See https://reactrouter.com/docs/en/v6/api#router for more information.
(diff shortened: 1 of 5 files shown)


docs/Admin.md
```diff
@@ -470,7 +471,17 @@ const App = () => (
 );
 ```
 
-**Caution**: Do not use the 5.x version of the `history` package. It's currently incompatible with another dependency of react-admin, `connected-react-router`. `history@4.10.1` works fine. 
+## `basename`
+
+Use this prop to make all routes and links in your Admin relative to a "base" portion of the URL pathname that they all share. This is useful when including the Admin inside a larger app. See https://reactrouter.com/docs/en/v6/api#router for more informations.
```
WiXSL, line 476: `informations.` => **information**.

WiXSL (changes requested):

docs/Admin.md
djhi:
```suggestion
Use this prop to make all routes and links in your Admin relative to a "base" portion of the URL pathname that they all share. This is useful when including the Admin inside a larger app. See https://reactrouter.com/docs/en/v6/api#router for more information.
```

djhi:
With or without the basename, the app is served the same way and works

fzaninotto:
So what's the use of the basename prop?

djhi:
Actually, it only matters when using the browser history and not serving the app from the domain root.

fzaninotto (approved):

fzaninotto:
Then you should change the doc accordingly, to explain that it's only necessary when using browser history (and modify the example to include the history).

fzaninotto (approved):

PR #6921, opened 2021-11-30
Remove permissions prop injection from main views

djhi:


packages/ra-core/src/auth/useAuthenticated.ts
```diff
@@ -28,9 +26,21 @@ const emptyParams = {};
...
-        checkAuth(params).catch(() => {});
-    }, [checkAuth, params]);
+        if (enabled) {
+            checkAuth(params).catch(() => {});
+        }
+    }, [checkAuth, enabled, params]);
+};
+
+export type UseAuthenticatedOptions<ParamsType> = {
+    enabled?: boolean;
```
WiXSL, line 42: It would be great to have documentation of the new `enabled` prop

packages/ra-core/src/auth/useCheckAuth.ts
```diff
@@ -40,7 +40,7 @@ import useNotify from '../sideEffect/useNotify';
  *     return authenticated ? <Bar /> : <BarNotAuthenticated />;
  * } // tip: use useAuthState() hook instead
  */
-const useCheckAuth = (): CheckAuth => {
+export const useCheckAuth = (): CheckAuth => {
```
WiXSL, line 43: Missing export for type `CheckAuth`

WiXSL (changes requested):

packages/ra-core/src/auth/useAuthenticated.ts
```diff
@@ -28,9 +26,21 @@ const emptyParams = {};
  *         </Admin>
  *     );
  */
-export default (params: any = emptyParams) => {
+export const useAuthenticated = <ParamsType = any>(
```
fzaninotto, line 29: this is a breaking change for users who used to pass params, and should be mentioned in the upgrade guide

packages/ra-core/src/controller/create/useCreateController.spec.tsx
```diff
@@ -1,60 +1,49 @@
...
+    describe('getRecordFromLocation', () => {
+        const location: Location = {
+            key: 'a_key',
             pathname: '/foo',
             search: undefined,
             state: undefined,
             hash: undefined,
         };
 
-        it('should return an empty record by default', () => {
```
fzaninotto, line 20: why did you remove this one ?

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -49,20 +51,23 @@ export const useCreateController = <
     props: CreateControllerProps = {}
 ): CreateControllerResult<RecordType> => {
     const {
-        record = {},
-        successMessage,
+        disableAuthentication,
```
fzaninotto, line 54: You should update the Auth doc to mention this prop. Same for other controllers.

fzaninotto (changes requested):
The Resource component is becoming smaller and smaller... If we ever remove the registration phase, it'll just be routes!

packages/ra-core/src/controller/create/useCreateController.spec.tsx
djhi: Because that's not the responsability of this function anymore

docs/Authentication.md
```diff
@@ -701,11 +701,15 @@ const MyPage = () => {
 export default MyPage;
 ```
 
-If you call `useAuthenticated()` with a parameter, this parameter is passed to the `authProvider` call as second parameter. that allows you to add authentication logic depending on the context of the call:
+`useAuthenticated` accepts an options object as its only argument, with the following properties:
+- `enabled`: whether it should check for an authenticated user
```
WiXSL, line 705: - `enabled`: whether it should check for an authenticated user (`true` by default)

WiXSL (changes requested):

docs/Authentication.md
djhi: Make it a suggestion next time ;)
```suggestion
- `enabled`: whether it should check for an authenticated user (`true` by default)
```

WiXSL (approved):

docs/Authentication.md
```diff
@@ -1165,3 +1169,30 @@ const Menu = ({ onMenuClick, logout }) => {
     );
 }
 ```
+
+### Allowing Anonymous Access to Custom Views
+
+You might have custom views that leverages our components or hooks such as:
```
fzaninotto, line 1175:
```suggestion
You might have custom views that leverage react-admin components or hooks such as:

```

fzaninotto (changes requested):

fzaninotto (approved):

PR #6922, opened 2021-12-01
Faster ci

djhi:


.github/workflows/test.yml
```diff
@@ -22,8 +22,23 @@ jobs:
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Use Node.js LTS
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
+    - uses: bahmutov/npm-install@v1
+      with:
+        install-command: yarn --frozen-lockfile --non-interactive --silent
     - name: Lint
```
fzaninotto, line 42: shouldn't we put that in the typecheck phase, too?

djhi: It's more about having all jobs taking approximately the same time

djhi: unit-test is already the fastest job so it make sense to keep the linting there

fzaninotto:
Awesome!

PR #6924, opened 2021-12-02
Introduce a new AutocompleteInput based on MUI Autocomplete

djhi:
- [x] Avoid breaking change for `optionText` and leverage the `inputText` prop we had in previous version


packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -23,23 +23,8 @@ describe('<AutocompleteInput />', () => {
         resource: 'users',
     };
 
-    it('should use a Downshift', () => {
```
djhi, line 26: Removed because we don't want to test mui autocomplete

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -645,37 +632,6 @@ describe('<AutocompleteInput />', () => {
         await waitFor(() => expect(queryAllByRole('option').length).toEqual(1));
     });
 
-    it('passes options.suggestionsContainerProps to the suggestions container', () => {
```
djhi, line 648: mui autocomplete provide its own ways to customize the underlying components

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -694,60 +650,6 @@ describe('<AutocompleteInput />', () => {
         });
     });
 
-    it('should not render a LinearProgress if loading is true and a second has not passed yet', () => {
```
djhi, line 697: Mui autocomplete has its own loading mechanism. Maybe we should check we don't its loading prop unless a second has passed

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
     UseChoicesOptions,
-    mergeRefs,
     useSuggestions,
+    useTimeout,
     useTranslate,
     warning,
 } from 'ra-core';
-
-import { InputHelperText } from './InputHelperText';
-import { AutocompleteSuggestionList } from './AutocompleteSuggestionList';
```
fzaninotto, line 33: I assume that these are still required for AutocompleteArrayInput, and that's why you didn't delete them in this PR, right?

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
-import { AutocompleteSuggestionItem } from './AutocompleteSuggestionItem';
-import { AutocompleteInputLoader } from './AutocompleteInputLoader';
 import {
     SupportCreateSuggestionOptions,
     useSupportCreateSuggestion,
 } from './useSupportCreateSuggestion';
+import { InputHelperText } from './InputHelperText';
 
-/**
```
fzaninotto, line 41: Please keep a JSDoc section

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
         id: idOverride,
         input: inputOverride,
-        inputText,
         isRequired: isRequiredOverride,
         label,
-        limitChoicesToValue,
         loaded,
         loading,
-        margin = 'dense',
+        limitChoicesToValue,
```
fzaninotto, line 39: This one needs documentation

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
         helperText,
         id: idOverride,
         input: inputOverride,
-        inputText,
         isRequired: isRequiredOverride,
         label,
-        limitChoicesToValue,
         loaded,
         loading,
-        margin = 'dense',
```
fzaninotto, line 128: You should still support it, or it's a breaking change

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
         resource,
         setFilter,
-        shouldRenderSuggestions: shouldRenderSuggestionsOverride,
+        shouldRenderSuggestions,
         source,
         suggestionLimit,
-        translateChoice = true,
+        translateChoice,
         validate,
-        variant = 'filled',
```
fzaninotto, line 157: you should still support it, or it's a breaking change

fzaninotto:
I see a UI bug in the Storybook: in the "inside Reference Input" srtory, when typing text, the choices are briefly emptied before being filled with matching choices. I think you should use the keepPreviousData option in useGetList (once it's handled by react-query, of course)

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -1,193 +1,63 @@
...
         ...rest
     } = props;
 
-    if (isValidElement(optionText) && !inputText) {
```
fzaninotto, line 161: Why did you remove these warnings?

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -208,48 +78,34 @@ export const AutocompleteInput = (props: AutocompleteInputProps) => {
...
-        [input.value, getSuggestionFromValue]
+        () =>
+            choices?.find(choice => get(choice, optionValue) === input.value) ||
+            null,
+        [choices, input.value, optionValue]
     );
 
-    const doesQueryMatchSuggestion = useMemo(() => {
+    useEffect(() => {
         if (isValidElement(optionText)) {
```
fzaninotto, line 89: why put that in an effect?

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -135,9 +135,9 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
     });
 
     const handleChange = useCallback(
-        (event, newItem) => {
-            if (newItem) {
-                input.onChange([...input.value, getChoiceValue(newItem)]);
+        event => {
```
fzaninotto, line 138: Not sure about what you did here, can you elaborate?

packages/ra-ui-materialui/src/input/SelectInput.tsx
```diff
@@ -178,14 +178,14 @@ export const SelectInput = (props: SelectInputProps) => {
...
     const handleChange = useCallback(
-        async (event: React.ChangeEvent<HTMLSelectElement>, newItem) => {
-            if (newItem) {
-                const value = getChoiceValue(newItem);
-                input.onChange(value);
-                return;
+        async (event: any) => {
+            let choice = event;
+            // eslint-disable-next-line eqeqeq
+            if (event?.target?.value != undefined) {
```
fzaninotto, line 184: what that code does is cryptic to me. Would you mind adding a comment explaining it?

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
fzaninotto: I see that you did it, am I wrong?

fzaninotto (changes requested):
Great! 💯

cypress/integration/edit.js
```diff
@@ -78,6 +78,12 @@ describe('Edit Page', () => {
             cy.get(EditPostPage.elements.input('backlinks[0].url')).blur();
 
             cy.contains('Required');
+            // FIXME: We navigate away from the page and confirm the unsaved changes
```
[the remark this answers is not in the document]
djhi: I'm not sure yet

examples/simple/src/comments/CommentEdit.tsx
```diff
@@ -152,15 +152,19 @@ const CommentEdit = props => {
                                         suggestion
                                     ) => {
                                         const title = `${suggestion.title} - ${suggestion.id}`;
-
                                         return title.includes(filterValue);
                                     }}
-                                    optionText={<OptionRenderer />}
+                                    optionText={record =>
```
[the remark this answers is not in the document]
djhi: This is the unnecessary breaking change I was talking about. Will be fixed soon

examples/simple/src/comments/CommentEdit.tsx
```diff
@@ -152,15 +152,19 @@ const CommentEdit = props => {
...
+                                        `${record?.title} - ${record?.id}`
+                                    }
                                     inputText={inputText}
                                     options={{
-                                        InputProps: {
-                                            fullWidth: true,
-                                        },
+                                        renderOption: (props, record) => (
```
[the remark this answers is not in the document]
djhi: This is the unnecessary breaking change I was talking about. Will be fixed soon

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: Exactly

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: I actually migrated it without thinking but I fail to see its usefulness

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: This is the unnecessary breaking change I was talking about. Will be fixed soon

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: Because we rely on props and we want the warning to be displayed if needed when they change?

examples/simple/src/comments/CommentEdit.tsx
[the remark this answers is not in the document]
djhi: Fixed

examples/simple/src/comments/CommentEdit.tsx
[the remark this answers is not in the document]
djhi: Fixed

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: Added

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
djhi: I'll revisit once the AutocompleteArrayInput is migrated to use Autocomplete

WiXSL:
Editing a Post In simple example:

WiXSL (commented):
Editing a Post In simple example:

UPGRADE.md
```diff
@@ -755,6 +755,14 @@ test('MyComponent', () => {
...
+
+We migrated the `AutocompleteInput` so that it leverage Material UI [`<Autocomplete>`](https://mui.com/components/autocomplete/). If you relied on [Downshift](https://www.downshift-js.com/) options, you'll have to update your component.
+
+Besides, some props supported by the previous implementation aren't anymore:
+- `clearAlwaysVisible`: the clear button is now always either while hovering the input or when it has focus. You can hide it using the `<Autocomplete>` `disableClearable` prop though.
+- `resettable`: Removed for the same reason as `clearAlwaysVisible`
+
 ## `useAuthenticated` Signature has Changed
 
 `useAuthenticated` uses to accept only the parameters passed to the `authProvider.checkAuth` function. It now accept an option object with two properties:
```
WiXSL, line 768:
```suggestion
`useAuthenticated` uses to accept only the parameters passed to the `authProvider.checkAuth` function. It now accepts an option object with two properties:
```

WiXSL (changes requested):

djhi:
> Editing a Post In simple example:
> 
> 
> 
> I just set a value, and then re-open the AutocompleteInput

Good catch!

UPGRADE.md
```diff
@@ -755,9 +755,17 @@ test('MyComponent', () => {
 });
 ```
 
+## AutocompleteInput Now Uses Material UI Autocomplete
+
+We migrated the `AutocompleteInput` so that it leverage Material UI [`<Autocomplete>`](https://mui.com/components/autocomplete/). If you relied on [Downshift](https://www.downshift-js.com/) options, you'll have to update your component.
+
+Besides, some props supported by the previous implementation aren't anymore:
+- `clearAlwaysVisible`: the clear button is now always either while hovering the input or when it has focus. You can hide it using the `<Autocomplete>` `disableClearable` prop though.
```
WiXSL, line 763:
```suggestion
- `clearAlwaysVisible`: the clear button is now always visible, either while hovering the input or when it has focus. You can hide it using the `<Autocomplete>` `disableClearable` prop though.
```

WiXSL (changes requested):

docs/Inputs.md
```diff
@@ -662,17 +667,6 @@ import { AutocompleteInput, ReferenceInput } from 'react-admin';
 </ReferenceInput>
 ```
 
-Lastly, would you need to override the props of the suggestion's container (a `Popper` element), you can specify them using the `options.suggestionsContainerProps`. For example:
```
WiXSL, line 665: Since `suggestionsContainerProps` was introduced by react-admin, shouldn't we clarify that now, `@mui` way of overriding the `Popper` container should be used?

djhi: That's in the MUI documentation, I don't think we should include it

WiXSL:
The [vercel deployment](https://react-admin-6xv4rxrxy-marmelab.vercel.app/#/posts) doesn't have the authors ArrayInput in Post form and lacks the User menu

djhi:
> The [vercel deployment](https://react-admin-6xv4rxrxy-marmelab.vercel.app/#/posts) doesn't have the authors ArrayInput in Post form and lacks the User menu

You have to log in with admin/password

WiXSL:
> I was trying it locally and it caught my attention, you are right

I was trying it locally and it caught my attention, you are right

WiXSL (approved):

UPGRADE.md
```diff
@@ -755,9 +755,17 @@ test('MyComponent', () => {
 });
 ```
 
+## AutocompleteInput Now Uses Material UI Autocomplete
+
+We migrated the `AutocompleteInput` so that it leverage Material UI [`<Autocomplete>`](https://mui.com/components/autocomplete/). If you relied on [Downshift](https://www.downshift-js.com/) options, you'll have to update your component.
```
fzaninotto, line 760:
```suggestion
We migrated the `AutocompleteInput` so that it leverages Material UI [`<Autocomplete>`](https://mui.com/components/autocomplete/). If you relied on [Downshift](https://www.downshift-js.com/) options, you'll have to update your component.
```

fzaninotto (approved):

PR #6926, opened 2021-12-02
Fix List bulkActionButtons prop does not handle the value true

### LINKED ISSUE #6925: List bulkActionButtons prop does not handle the value true
**What you were expecting:**
I expect I can pass a boolean to this prop (for example by checking permissions) and it would display the default bulk actions if I pass `true`.

**What happened instead:**
If I pass `true`, no actions are displayed after selecting rows.

WiXSL:
Fixes #6925

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Children, cloneElement, ReactElement } from 'react';
+import { Children, cloneElement, isValidElement, ReactElement } from 'react';
 import PropTypes from 'prop-types';
 import Card from '@material-ui/core/Card';
 import classnames from 'classnames';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -68,10 +68,15 @@ export const ListView = (props: ListViewProps) => {
                     })}
                     key={version}
                 >
-                    {bulkActionButtons !== false && bulkActionButtons && (
-                        <BulkActionsToolbar {...controllerProps}>
-                            {bulkActionButtons}
-                        </BulkActionsToolbar>
+                    {bulkActionButtons === true ? (
+                        <DefaultBulkActionButtons />
+                    ) : (
```
djhi, line 73: I believe `DefaultBulkActionButtons` should be passed as the `BulkActionsToolbar` child

djhi (changes requested):

packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -68,10 +68,17 @@ export const ListView = (props: ListViewProps) => {
...
-                            {bulkActionButtons}
+                            <DefaultBulkActionButtons />
                         </BulkActionsToolbar>
+                    ) : (
+                        bulkActionButtons !== false &&
+                        bulkActionButtons && (
+                            <BulkActionsToolbar {...controllerProps}>
+                                {bulkActionButtons}
+                            </BulkActionsToolbar>
+                        )
```
djhi, line 81: What do you think about:

```jsx
{bulkActionButtons !== false ? (
    <BulkActionsToolbar {...controllerProps}>
         {isValidElement(bulkActionButtons) ? bulkActionButtons : <DefaultBulkActionButtons />}
    </BulkActionsToolbar>
) : null}
```

WiXSL: Even better!

djhi (approved):

PR #6932, opened 2021-12-03
Fix ArrayInput issue with initialValue

djhi:
This issue happens because of how final-form handles the initialValue prop. It takes precedence over the form level initialValue that we set to the record when editing one. However, as ArrayInput dynamically adds new final-form fields, their initialValue overrides the value they got from the form.

The solution is to update the useInput hook so that it ignores the initialValue if a record is available through the RecordContext and has a value for this field.

Closes #6908


packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -216,7 +219,7 @@ export type FormWithRedirectSave = (
 export interface FormWithRedirectOwnProps {
     defaultValue?: any;
     formRootPathname?: string;
-    record?: RaRecord;
+    record?: Partial<RaRecord>;
```
djhi, line 222: Partial here because `Create` could pass a record without an `id`

packages/ra-ui-materialui/src/input/BooleanInput.spec.tsx
```diff
@@ -90,10 +92,10 @@ describe('<BooleanInput />', () => {
     });
 
     it('should be checked if the value is true and initialValue is false', () => {
-        const { getByLabelText } = render(
-            <Form
+        const { getByLabelText } = renderWithRedux(
```
WiXSL, line 95: Shouldn't most tests that use react-final-form's `Form` with `initialValues` be change like this one?

djhi: You're right

djhi: Not all of them though. Only those where we use both the initialValues of `Form` and initialValue of the field

djhi: So, that was actually the only test where we needed that

packages/ra-core/src/form/useInput.ts
```diff
@@ -67,8 +70,12 @@ const useInput = ({
         ? composeValidators(validate)
         : validate;
 
+    // Fetch the initialValue from the record if available or apply the provided initialValue.
+    // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
+    // We don't do this for the form level initialValues so that it works as it should in final-form
+    // (ie. field level initialValue override form level initialValues for this field).
     const { input, meta } = useFinalFormField(finalName, {
-        initialValue,
+        initialValue: get(record, source, initialValue),
```
DjebbZ, line 78: If I may, this makes sense to me.

djhi: Yes, we were actually lucky it worked before XD

packages/ra-ui-materialui/src/input/BooleanInput.spec.tsx
WiXSL: 👍

WiXSL (approved):

packages/ra-core/src/form/FormWithRedirect.tsx
```diff
@@ -67,6 +67,7 @@ const FormWithRedirect = ({
     sanitizeEmptyValues: shouldSanitizeEmptyValues = true,
     ...props
 }: FormWithRedirectProps) => {
+    const record = useRecordContext(props);
```
fzaninotto, line 70: it feels very weird that you read a context to create it again. 

What I understand is that you want to wrap a RecordContext if a record param is passed. That's what OptionalRecordContextProvider does in next. Maybe the right fix is to use that new component?

packages/ra-core/src/form/useInput.spec.tsx
```diff
@@ -208,30 +209,31 @@ describe('useInput', () => {
...
-                                        aria-label="Views"
-                                        {...input}
-                                    />
-                                );
-                            }}
-                        </Input>
-                    );
-                }}
-            />
+            <RecordContextProvider value={{ id: 1, views: 0 }}>
```
fzaninotto, line 212: I don't understand why you need that change as you're adding a context in FormWithRedirect

fzaninotto (changes requested):

packages/ra-core/src/form/FormWithRedirect.tsx
djhi: That would be the right fix but I didn't want to introduce a new component

packages/ra-core/src/form/useInput.spec.tsx
djhi: Forgot to fix this test after introducing the RecordContext in FormWithRedirect

djhi: Not really important though

fzaninotto (approved):

PR #6933, opened 2021-12-03
[Doc] Fix authProvider example

### LINKED ISSUE #5200: TypeError: authProvider.checkAuth is not a function
**What you were expecting:**
login page
**What happened instead:**
TypeError: authProvider.checkAuth is not a function
**Steps to reproduce:**
Only added App.js and authProvider.js 
**Related code:**

```
import authProvider from './authProvider';
import { Admin, Resource, ListGuesser } from 'react-admin';


const dataProvider = jsonServerProvider('https://jsonplaceholder.typicode.com');
const App = () => (
    <Admin dataProvider={dataProvider} authProvider={authProvider}>
        <Resource name="users" list={ListGuesser} />
    </Admin>
);

export default App;
```

**Other information:**

**Environment**

* React-admin version:^3.8.2
* React version:^16.13.1
* Browser:Chrome
```TypeError: authProvider.checkAuth is not a function
(anonymous function)
E:/code/my-app/node_modules/ra-core/esm/auth/useCheckAuth.js:49
  46 | if (params === void 0) { params = {}; }
  47 | if (logoutOnFailure === void 0) { logoutOnFailure = true; }
  48 | if (redirectTo === void 0) { redirectTo = defaultAuthParams.loginUrl; }
> 49 | return authProvider.checkAuth(params).catch(function (error) {
     | ^  50 |     if (logoutOnFailure) {
  51 |         logout({}, error && error.redirectTo
  52 |             ? error.redirectTo

```

WiXSL:
Fixes #5200

### DIFF
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -104,6 +104,12 @@ const authProvider = {
                 throw new Error('Network error')
             });
     },
+    checkAuth: () => {
+        return localStorage.getItem('auth') ? Promise.resolve() : Promise.reject();
+    },
+    getPermissions: () => {
+        return Promise.resolve();
+    },
     // ...
 };
 
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -105,9 +105,11 @@ const authProvider = {
             });
     },
     checkAuth: () => {
-        return localStorage.getItem('auth') ? Promise.resolve() : Promise.reject();
+        // Required for the authentication to work
+        return Promise.resolve();
     },
     getPermissions: () => {
+        // Required for the authentication to work
         return Promise.resolve();
     },
     // ...


docs/Authentication.md
```diff
@@ -104,6 +104,12 @@ const authProvider = {
                 throw new Error('Network error')
             });
     },
+    checkAuth: () => {
+        return localStorage.getItem('auth') ? Promise.resolve() : Promise.reject();
```
fzaninotto, line 108: The problem is that the Authentication chapter mentions that method later ('checking credentials during navigation'). So basically we're adding code without explaining it here, and we're re-adding it later. It doesn't really make sense. 

I'd prefer that you add resolved promises for both, with a comment explaining that these methods are required for the authentication system to work.

djhi (approved):

fzaninotto (approved):

PR #6934, opened 2021-12-04
Fix useList ids and data state

WiXSL:
- [x] if `data` is passed to `useList` and `ids` are not, `ids` should be taken from `data`
- [x] `loading` and `loaded` props made optional and defaulted to `loaded`=`true` and `loading`=`false` (like in next branch)
       I know is a BC, but it shouldn't be of any consequence
- [x] Updated tests
- [x] Update documentation

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -2019,15 +2019,9 @@ const data = [
     { id: 2, name: 'Sylvester' },
     { id: 3, name: 'Jean-Claude' },
 ]
-const ids = [1, 2, 3];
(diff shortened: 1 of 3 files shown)


fzaninotto:
What bug does it fix? It seems like a new feature to me, and we've already overhauled useList in v4.

WiXSL:
It fixes the first item,
If data is passed to `useList` and `ids` are not, `ids` should be taken from `data`

PR #6935, opened 2021-12-04
【Bugfix】accept prop of FileInput and ImageInput  can't receive string array type

tkow:
DropzoneOptions props's accept can receive string[] type but, FileInputProps is can't because the redefined type is wrong.
I replace that as it uses inherited DropzoneOptions' props types  if these are redefined  of them.


tkow:
I close this PR for my mistake of derived branch.
I migrate this to https://github.com/marmelab/react-admin/pull/6940.

PR #6938, opened 2021-12-05
Fix lerna config to build simple project

WiXSL:
On a clean install and build of the project, you don't get the `examples/simple/dist` folder needed by cypress to run the tests


djhi (approved):
:+1:

PR #6940, opened 2021-12-06
【Bugfix】accept prop of FileInput and ImageInput can't receive string array type 

tkow:
DropzoneOptions props's accept can receive string[] type but, FileInputProps is can't because the redefined type is wrong.
I replace that as it uses inherited DropzoneOptions' props types if these are redefined of them.


djhi (approved):
Thanks!

PR #6941, opened 2021-12-06
[RFR] Adds RaFilterListItem class

kristelvdakker:
Adds a class to RaFilterListItem to be able to override the makeStyles with my own team


djhi (approved):
Thanks!

PR #6946, opened 2021-12-06
Fix shorten imports

WiXSL:
Use directory import when `index.ts` is available (node-style module resolution)


fzaninotto:
Let's avoid that type of change on master. It will make backporting changes to next harder.

We should only commiut bugfixes on master.

PR #6949, opened 2021-12-06
Fix types in tests

WiXSL:


fzaninotto (approved):
Thanks, but we shouldn't spend time on this on `master`. I'll merge this one, though.

PR #6954, opened 2021-12-07
Fix Warn about unsaved changes when modifying CheckGroupInput and ArrayInput

### LINKED ISSUE #6939: Warn about unsaved changes when modifying `CheckGroupInput` and `ArrayInput`
This problem was fixed in #5776 (`v3.11.3`) and is back since version `3.19.1`
You can reproduce it editing a Post record in the simple example, modifying one of these controls and trying to submit the form.

I'll try to see exactly what the issue is.

WiXSL:
Fixes #6939

### DIFF
--- a/packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
+++ b/packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
@@ -28,7 +28,7 @@ const useWarnWhenUnsavedChanges = (
         }
 
         const release = history.block(location => {
-            const isInsideForm = location.pathname.startsWith(
+            const isInsideForm = history.location.pathname.startsWith(
                 initialLocation.current
             );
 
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
```diff
@@ -28,7 +28,7 @@ const useWarnWhenUnsavedChanges = (
         }
 
         const release = history.block(location => {
-            const isInsideForm = location.pathname.startsWith(
+            const isInsideForm = history.location.pathname.startsWith(
```
WiXSL, line 31: The problem is that `location.pathname` is pointing to the wrong path, so I use `history.location.pathname`.
I have no clue why this happens

WiXSL, line 31: This is wrong, I'm removing the feature

PR #6955, opened 2021-12-07
made loading indicator in appbar optional

shreyas-jadhav:
Loading Indicator in the `<AppBar>` is made optional. we can be hide by passing `showLoadingIndicator={false}`


fzaninotto:
Thanks for your PR!

The master branch is feature freeze as we work on 4.0 in the next branch. 

Besides, users can already override the AppBar at will. I don't think this feature should end up in the core. 

So I'm closing it, but feel free to request new pulls in the future!

PR #6956, opened 2021-12-07
Add ApolloError in error body when handling errors

### LINKED ISSUE #6953: ra-data-graphql: Missing original error as body in handleError
**What you were expecting:**

When mutations fail, the specific error (HttpError body) should be accessible like shown in https://marmelab.com/react-admin/CreateEdit.html#submission-validation

**What happened instead:**

The HttpError body is not set.

**Steps to reproduce:**

* Create new project
* Add ra-data-gaphql data provider
* Implement submission validation like shown in https://marmelab.com/react-admin/CreateEdit.html#submission-validation
* `error.body` is not set and thus errors cannot be accessed

**Environment**

* React-admin version: 3.19.2

hlubek:
Fixes #6953

I added a test to check for correct error handling when executing a mutation.

### DIFF
--- a/packages/ra-data-graphql/src/index.ts
+++ b/packages/ra-data-graphql/src/index.ts
@@ -221,7 +221,6 @@ export default async (options: Options): Promise<DataProvider> => {
 };
 
 const handleError = (error: ApolloError) => {
-    console.error({ error });
     if (error?.networkError as ServerError) {
         throw new HttpError(
             (error?.networkError as ServerError)?.message,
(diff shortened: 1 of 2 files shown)


packages/ra-data-graphql/src/index.ts
```diff
@@ -221,15 +221,14 @@ export default async (options: Options): Promise<DataProvider> => {
 };
 
 const handleError = (error: ApolloError) => {
-    console.error({ error });
```
hlubek, line 224: I think this console.error should not be here. Since `handleError` re-throws, it should be logged / catched at higher levels.

WiXSL:
Thanks for your contribution!
Seems ok

fzaninotto (approved):

fzaninotto:
Thanks!

PR #6960, opened 2021-12-08
[BC Break] Update useGetOne signature to reflect the getOne signature

fzaninotto:
To make it consistent with `useGetList`, `useUpdate`, and other dataProvider hooks to come. 

```diff
-const { data, isLoading } = useGetOne('posts', 123);
+const { data, isLoading } = useGetOne('posts', { id: 123 });
```

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -291,7 +291,7 @@ The specialized hooks based on `useQuery` (`useGetList`, `useGetOne`, `useGetMan
 **Tip**: If you use TypeScript, you can specify the record type for more type safety:
 
 ```jsx
-const { data, isLoading } = useGetOne<Product>('products', 123);
+const { data, isLoading } = useGetOne<Product>('products', { id: 123 });
 //        \- type of data is Product
 ```
 
@@ -329,13 +329,13 @@ This hook calls `dataProvider.getOne()` when the component mounts.
 
 ```jsx
 // syntax
-const { data, isFetching, isLoading, error, refetch } = useGetOne(resource, id, options);
+const { data, isFetching, isLoading, error, refetch } = useGetOne(resource, { id }, options);
 
 // example
 import { useGetOne } from 'react-admin';
 
 const UserProfile = ({ record }) => {
-    const { data, isLoading, error } = useGetOne('users', record.id);
+    const { data, isLoading, error } = useGetOne('users', { id: record.id });
     if (isLoading) { return <Loading />; }
     if (error) { return <p>ERROR</p>; }
     return <div>User {data.username}</div>;
--- a/docs/DataProviders.md
+++ b/docs/DataProviders.md
@@ -767,7 +767,7 @@ Here is a glimpse of the `useGetOne` hook usage:
 import { useGetOne } from 'react-admin';
 
 const UserProfile = ({ record }) => {
-    const { data, isLoading, error } = useGetOne('users', record.id);
+    const { data, isLoading, error } = useGetOne('users', { id: record.id });
     if (isLoading) { return <Loading />; }
     if (error) { return <p>ERROR</p>; }
     return <div>User {data.username}</div>;
--- a/packages/ra-core/src/dataProvider/useGetOne.ts
+++ b/packages/ra-core/src/dataProvider/useGetOne.ts
@@ -20,12 +20,12 @@ import useDataProvider from './useDataProvider';
  * @param {Options} options Options object to pass to the react-query queryClient.
  *
  * @typedef Params
- * @property id a resource identifier, e.g. 123
+ * @prop id a resource identifier, e.g. 123
  *
  * @typedef Options
- * @property enabled Flag to conditionally run the query. If it's false, the query will not run
- * @property onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
- * @property onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
+ * @prop enabled Flag to conditionally run the query. If it's false, the query will not run
+ * @prop onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
+ * @prop onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
  *
  * @returns The current request state. Destructure as { data, error, isLoading, refetch }.
  *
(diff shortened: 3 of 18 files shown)


docs/Actions.md
```diff
@@ -329,13 +329,13 @@ This hook calls `dataProvider.getOne()` when the component mounts.
 
 ```jsx
 // syntax
-const { data, isFetching, isLoading, error, refetch } = useGetOne(resource, id, options);
+const { data, isFetching, isLoading, error, refetch } = useGetOne(resource, { id }, options);
 
 // example
 import { useGetOne } from 'react-admin';
```
WiXSL, line 335:
```suggestion
import { useGetOne, Loading } from 'react-admin';
```

docs/DataProviders.md
```diff
@@ -767,7 +767,7 @@ Here is a glimpse of the `useGetOne` hook usage:
 import { useGetOne } from 'react-admin';
```
WiXSL, line 767:
```suggestion
import { useGetOne, Loading } from 'react-admin';
```

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -1,4 +1,4 @@
-import { Identifier, Record } from '../types';
+import { Identifier, Record, GetOneParams } from '../types';
```
WiXSL, line 1:
```suggestion
import { Record, GetOneParams } from '../types';
```

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -29,15 +33,15 @@ import useDataProvider from './useDataProvider';
  * import { useGetOne } from 'react-admin';
```
WiXSL, line 33:
```suggestion
 * import { useGetOne, Loading } from 'react-admin';
```

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -16,11 +16,15 @@ import useDataProvider from './useDataProvider';
...
- * @param id The resource identifier, e.g. 123
- * @param {Object} options Options object to pass to the react-query queryClient.
- * @param {boolean} options.enabled Flag to conditionally run the query. If it's false, the query will not run
- * @param {Function} options.onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
- * @param {Function} options.onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
+ * @param {Params} params The getOne parameters { id }, e.g. { id: 123 }
+ * @param {Options} options Options object to pass to the react-query queryClient.
+ *
+ * @typedef Params
+ * @props params.id a resource identifier, e.g. 123
```
WiXSL, line 23:
```suggestion
 * @prop params.id A resource identifier, e.g. 123
```

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -16,11 +16,15 @@ import useDataProvider from './useDataProvider';
...
+ * @typedef Options
+ * @pprops options.enabled Flag to conditionally run the query. If it's false, the query will not run
```
WiXSL, line 25:
```suggestion
 * @prop options.enabled Flag to conditionally run the query. If it's false, the query will not run
```

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -16,11 +16,15 @@ import useDataProvider from './useDataProvider';
...
+ * @pprops options.onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
```
WiXSL, line 26:
```suggestion
 * @prop options.onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
```

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -16,11 +16,15 @@ import useDataProvider from './useDataProvider';
...
+ * @pprops options.onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
```
WiXSL, line 27:
```suggestion
 * @prop options.onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
```

WiXSL (changes requested):

docs/Actions.md
fzaninotto: No, it's just an example - we don't have a Loading component for that.

packages/ra-core/src/dataProvider/useGetOne.ts
fzaninotto: In fact, according to the jsDoc spec, it should be property... https://jsdoc.app/tags-typedef.html

WiXSL: But it could be @prop as well [synonym](https://jsdoc.app/index.html#:~:text=%40property%20(synonyms%3A%20%40prop))

djhi: Why do we need a @typedef when we already use TS? You can put those comments as jsdoc on the type and its properties directly, no?

packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -16,11 +16,16 @@ import useDataProvider from './useDataProvider';
...
+ * @param {Params} params The getOne parameters { id }, e.g. { id: 123 }
+ * @param {Options} options Options object to pass to the react-query queryClient.
+ *
+ * @typedef Params
+ * @property id a resource identifier, e.g. 123
+ *
+ * @typedef Options
+ * @property enabled Flag to conditionally run the query. If it's false, the query will not run
+ * @property onSuccess Side effect function to be executed upon success, e.g. { onSuccess: { refresh: true } }
+ * @property onError Side effect function to be executed upon failure, e.g. { onError: error => notify(error.message) }
```
djhi, line 28: Why a typedef when we have TS?

fzaninotto: because everyione doesn't have TS, and I think the jsDoc renders *before* type definitions in IDEs.

packages/ra-core/src/dataProvider/useGetOne.ts
fzaninotto: @WiXSL you're totally right, I'm pushing an update

WiXSL (approved):

djhi (approved):

PR #6961, opened 2021-12-08
Revert main views don't render a Card

fzaninotto:
I think we should revert the changes we made to detail views (Edit, Create and Show) in v4 to move the Card decoration to the content element (SimpleForm) instead of the container, and to remove the aside prop.

The reasoning behind the original change was to allow developers to put whatever they want as children of Show or List, letting them have 2-columns layouts without an extra prop. But in practice, it doesn't work, because the result of a view with several children looks bad because the actions (and pagination in the List view) still take the full width. 

image

partially reverts #6746

### DIFF
--- a/docs/Show.md
+++ b/docs/Show.md
@@ -11,7 +11,7 @@ The `<Show>` component handles the logic of the Show page:
 - it computes the default page title
 - it creates a `ShowContext` and a `RecordContext`,
 - it renders the page layout with the correct title and actions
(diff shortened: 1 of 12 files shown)


fzaninotto:
Switching to RFR

djhi (changes requested):
There is an eslint warning left. Otherwise :+1:

djhi (approved):

PR #6967, opened 2021-12-09
Fix List Filter Menu is not accessible

### LINKED ISSUE #6964: List Filter Menu is not accessible
**What you were expecting:**
I expect using the keyboard to open the filter menu and select an item to work

**What happened instead:**
I can use the keyboard to open the filter menu but not to select one of its items

**Steps to reproduce:**
Open the [demo](https://marmelab.com/react-admin-demo/#/commands) and use the keyboard to focus the filter menu and try to select an item.

**Environment**

* React-admin version: 3.19.3
* React version: 17

WiXSL:
Fixes #6964

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
@@ -91,12 +91,13 @@ const FilterButton = (props: FilterButtonProps): JSX.Element => {
                 anchorEl={anchorEl.current}
                 onClose={handleRequestClose}
             >
-                {hiddenFilters.map((filterElement: JSX.Element) => (
+                {hiddenFilters.map((filterElement: JSX.Element, index) => (
                     <FilterButtonMenuItem
                         key={filterElement.props.source}
                         filter={filterElement}
                         resource={resource}
                         onShow={handleShow}
+                        autoFocus={index === 0}
                     />
                 ))}
             </Menu>
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
@@ -82,6 +82,7 @@ const FilterButton = (props: FilterButtonProps): JSX.Element => {
             <Button
                 className="add-filter"
                 label="ra.action.add_filter"
+                aria-haspopup="true"
                 onClick={handleClickButton}
             >
                 <ContentFilter />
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -91,12 +91,13 @@ const FilterButton = (props: FilterButtonProps): JSX.Element => {
...
                 onClose={handleRequestClose}
             >
-                {hiddenFilters.map((filterElement: JSX.Element) => (
+                {hiddenFilters.map((filterElement: JSX.Element, index) => (
                     <FilterButtonMenuItem
                         key={filterElement.props.source}
                         filter={filterElement}
                         resource={resource}
                         onShow={handleShow}
+                        autoFocus={index === 0}
```
djhi, line 100: I don't get why this is necessary to fix the keyboard navigation. Can you explain?

djhi (commented):
While you're on this file, would you mind adding the missing `aria-haspopup="true"` on the `Button` element?

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
WiXSL: I couldn't get it to work otherwise, I'm forcing the focus on the first menu item, do you have a suggestion with aria attributes?

djhi: No but I fail to find a difference between our code and the MUI v4 demos for the menu

WiXSL: > No but I fail to find a difference between our code and the MUI v4 demos for the menu

What I've noticed is that they don't have a forwardRef `MenuItem` inside a `Menu` in their examples.
It works out of the box without this, but of course, we need it.

WiXSL: For example, this works:

```jsx
<Menu
    open={open}
    anchorEl={anchorEl.current}
    onClose={handleRequestClose}
>
    {hiddenFilters.map((filterElement: JSX.Element, index) => (
        <MenuItem
            key={filterElement.props.source}
            //filter={filterElement}
            //resource={resource}
            //onShow={handleShow}
            //autoFocus={index === 0}
        >
            <FieldTitle
                label={filterElement.props.label}
                source={filterElement.props.source}
                resource={resource}
            />
        </MenuItem>
    ))}
</Menu>

WiXSL: Of course it would be a BC

fzaninotto (approved):

PR #6969, opened 2021-12-10
Allow "false" prop for SimpleList toolbar, removing element

ZachSelindh:
Because the toolbar in the SimpleFormView renders based on a logical AND operator, it is possible to remove that element by passing "false" to the toolbar prop on a SimpleForm component, for the purpose of using RA Input components wrapped in a custom form that doesn't require the traditional toolbar.

However, passing false will throw a type error:

Type 'false' is not assignable to type 'ReactElement<any, string | JSXElementConstructor> | undefined'.ts(2322)
SimpleForm.d.ts(68, 5): The expected type comes from property 'toolbar' which is declared here on type 'IntrinsicAttributes & SimpleFormProps'

This PR alters the PropTypes and interface defined Types to allow for false as a value for "toolbar."

Thank you!

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.tsx
@@ -60,10 +60,7 @@ SimpleForm.propTypes = {
     save: PropTypes.func,
     saving: PropTypes.bool,
     submitOnEnter: PropTypes.bool,
-    toolbar: PropTypes.oneOfType([
-        PropTypes.element,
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Hi and thanks for the PR! Can you fix the prettier related warnings?

ZachSelindh:
@djhi Looks like it passes now! Thanks.

djhi (approved):

djhi:
Thanks!

Issue #6974, opened 2021-12-10
Cannot use a SimpleList as a child of an ArrayField


Luwangel:
**What you were expecting:**

I want to use a `<SimpleList>` inside an `<ArrayField>`, but it's not working.

``` 
const UserShow = ({ permissions, ...props }) => (
  <Show {...props}>
        <TextField source="id" />
        <TextField source="name" />
        <ArrayField source="skills">
          <SimpleList primaryText={(record) => record.name} />
        </ArrayField>
        <ArrayField source="skills">
          <SingleFieldList>
            <ChipField source="name" />
          </SingleFieldList>
        </ArrayField>
  </Show>
);
```

**What happened instead:**

The `<SimpleList>` is rendered empty. It's not the case with the `<SingleFieldList>`.

![Sélection_004](https://user-images.githubusercontent.com/5584839/145606539-9c26788b-8a3a-4b49-896e-1bae1e1ef430.png)

**Steps to reproduce:**

I've written an example on CodeSandbox: https://codesandbox.io/s/sweet-mendeleev-jq98y?file=/src/users/UserShow.tsx.

**Related code:**

The `<ArrayField>` computes a total of `null`. But the `<SimpleList>` is not rendered if the total is not > 0.

https://github.com/marmelab/react-admin/blob/bd409a5942d8a20b04493b922be23f548734eac9/packages/ra-ui-materialui/src/field/ArrayField.tsx#L168

https://github.com/marmelab/react-admin/blob/77be726cca9c3512fdc4ddee65b32acb0988d16f/packages/ra-ui-materialui/src/list/SimpleList.tsx#L123

**Environment**

* React-admin version: 3.19.3
* Last version that did not exhibit the issue (if applicable): ?
* React version: /
* Browser: /
* Stack trace (in case of a JS error): /

WiXSL:
Fixed by #6975

PR #6975, opened 2021-12-10
Make possible to use a SimpleList as a child of an ArrayField

Luwangel:
## Description

The `<ArrayField>` builds a list context by passing null as total. The `SimpleList` expects a total > 0 to be rendered. Both are incompatible.

I studied two solutions:

1. Compute the total using ids (chosen)
2. Remove the total > 0 condition

To not introduce breaking changes, and because it's cleaner, I choose the first solution.

## Related Issue

Fixing https://github.com/marmelab/react-admin/issues/6974

### DIFF
--- a/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
@@ -6,6 +6,7 @@ import ArrayField from './ArrayField';
 import NumberField from './NumberField';
 import TextField from './TextField';
 import Datagrid from '../list/datagrid/Datagrid';
+import SimpleList from '../list/SimpleList';
 
 describe('<ArrayField />', () => {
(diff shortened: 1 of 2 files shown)


WiXSL (approved):

djhi (changes requested):
Can you add a test for the ArrayField and SimpleList?

fzaninotto:
Why don't you use useList? It's the perfect use case for it.

https://github.com/marmelab/react-admin/blob/next/docs/useList.md

Luwangel:
> Why don't you use useList? It's the perfect use case for it.
> 
> https://github.com/marmelab/react-admin/blob/next/docs/useList.md

The fix here is to make the ArrayField / SimpleList work together. But you're right, there are other ways to build this kind of feature.

fzaninotto (approved):

PR #6918, opened 2021-11-29
update docs & example : createMuiTheme is out of date

AntoineCrb:
https://v4.mui.com/customization/theming/

https://github.com/mui-org/material-ui/pull/25992

When using the code in DOCS a warning appears, `CreateTheme` should be the function used instead.

PR #6987, opened 2021-12-13
Fix ReferenceInput: use id to add current value to possible sources

wattroll:
`input.value` should be compared with `record.id` instead of `record`.

Bug introduced in #6918.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #6989, opened 2021-12-13
Fix ListActions useMemo dependencies

### LINKED ISSUE #6988: Incorrect exporter memoization in ListActions
**What you were expecting:**
When I switch resource, I get correct list actions with export button

**What happened instead:**
When I switch resource, export button is missing because the result of `ListActions` component is memoized

**Steps to reproduce:**
I have two resource, both has list with export action.
I have a tiny wrapper around `ListActions` which render export action conditionally according to the user permissions which are loaded asynchronously.

1. On load I can see export button in resource1
2. When I switch to resource2 I can also see export button
3. Switch back on resource1 cause export button to disappear
4. When I switch to resource2, back button is also missing

**Related code:**

This is an excerpt from my code:

```typescript
import { usePermissions } from 'react-admin';

export const useAppPermissions = () => usePermissions<RolePermissions>();

export const useHasAccess = (resource: ResourceKey, ...askedPermissions: PermissionAction[]) => {
    const { permissions } = usePermissions()

    return hasAccess(permissions, resource, ...askedPermissions);
};


/**
 * ListActions wrapper for conditional rendering of some actions according to the permissions
 */
import type { ListActionsProps } from 'react-admin';

export interface ListWithPermissionProps extends ListActionsProps {}

export const ListActionsWithPermissions = ({ filters, exporter, ...props }: ListWithPermissionProps) => {
    const resource = useResourceContext() as ResourceKey;
    const access = {
        search: useHasAccess(resource, PermissionAction.SEARCH),
        export: useHasAccess(resource, PermissionAction.EXPORT),
    };
    console.log('exporter', access.export ? exporter : false)
    return (
        <ListActions
            {...props}
            filters={access.search ? filters : undefined}
            exporter={access.export ? exporter : false}
        />
    );
};
```

**Other information**

I suppose I've also found the source of the issue. There is a [missing `exporter` dependency in `useMemo`](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/ListActions.tsx#L85) hook inside `ListActions` so even when the `exporter` value changes, the old value is used.

**Environment**

* React-admin version: 3.18.1
* React version: 17.0.2
* Browser: Chrome

WiXSL:
Fixes #6988

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListActions.tsx
+++ b/packages/ra-ui-materialui/src/list/ListActions.tsx
@@ -82,7 +82,18 @@ const ListActions = (props: ListActionsProps) => {
                 )}
             </TopToolbar>
         ),
-        [resource, displayedFilters, filterValues, selectedIds, filters, total] // eslint-disable-line react-hooks/exhaustive-deps
+        [
+            resource,
+            displayedFilters,
+            filterValues,
+            selectedIds,
+            filters,
+            total,
+            className,
+            currentSort,
+            exporter,
+            hasCreate,
+        ] // eslint-disable-line react-hooks/exhaustive-deps
     );
 };
 
--- a/packages/ra-ui-materialui/src/list/ListActions.tsx
+++ b/packages/ra-ui-materialui/src/list/ListActions.tsx
@@ -86,7 +86,7 @@ const ListActions = (props: ListActionsProps) => {
             resource,
             displayedFilters,
             filterValues,
-            selectedIds,
+            showFilter,
             filters,
             total,
             basePath,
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/ListActions.tsx
```diff
@@ -82,7 +81,19 @@ const ListActions = (props: ListActionsProps) => {
...
+            filterValues,
+            showFilter,
+            filters,
+            total,
+            basePath,
+            className,
+            currentSort,
+            exporter,
+            hasCreate,
+        ] // eslint-disable-line react-hooks/exhaustive-deps
```
fzaninotto, line 96: this comment isn't at the right place: the linter still emits a warning for missing props.

fzaninotto (changes requested):

djhi (approved):

fzaninotto (approved):

PR #6991, opened 2021-12-13
feat(ra-data-graphql-simple): add return types to functions

tupe12334:


djhi:
What issue are you fixing? Can you please describe the problem and the solution?

djhi:
No news for some time, closing

PR #6993, opened 2021-12-14
Build with esbuild

djhi:
I haven't noticed any difference in the build output for the demo. I checked its bundle analysis and tree shaking seems to work fine


packages/ra-input-rich-text/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './RichTextInput';
```
djhi, line 1: Breaking change as we now have a named export. However, we plan to replace this input with the tiptap version

WiXSL (approved):
I'll try it and builds everything without problem, even though I don't see an important improvement in time

fzaninotto (approved):

fzaninotto:
That's awesome, but I find the build output too verbose:

image

Could you make it shorter?

djhi:
Only completely silent

fzaninotto:
sounds good to me

fzaninotto (approved):

PR #7002, opened 2021-12-16
Fix ra-data-graphql-simple update methods

### LINKED ISSUE #6979: [Bug] updateMany broken in ra-data-graphql-simple in 3.19.x
**What you were expecting:**

 Calling updateMany in the 3.19.x version ra-data-graphql-simple correctly calls the graphql mutation. 
**What happened instead:**

Instead, you get an error that ID isn't defined.

**Related code:**
The issue here is that the location of the `id` field got moved when the dataprovider was updated as part of 3.19. In the previous version it was a child of `data` : https://github.com/marmelab/react-admin/blob/0861691abcb70aeb33515ad959c9c23c20dd2e57/packages/ra-data-graphql-simple/src/index.js#L44 But in the current version `id` is a sibling of the `data`:  https://github.com/marmelab/react-admin/blob/master/packages/ra-data-graphql-simple/src/index.ts#L47

This causes the issue because the function buildCreateUpdateVariables expects id to be a child of data: https://github.com/marmelab/react-admin/blob/851f2da9f8de2505ad17073f896947ad698bc75e/packages/ra-data-graphql-simple/src/buildVariables.ts#L291 So no `id` is put in the `variables` object which means Apollo gets sad that one of the required fields isn't there and throws an error.

**Environment**

* React-admin version:  3.19.1 version of ra-data-graphql-simple
* Last version that did not exhibit the issue (if applicable): 3.18.3 of ra-data-graphql-simple

djhi:
Closes #6979

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildVariables.ts
+++ b/packages/ra-data-graphql-simple/src/buildVariables.ts
@@ -291,34 +291,37 @@ const buildGetListVariables = (introspectionResults: IntrospectionResult) => (
 const buildCreateUpdateVariables = (
     resource: IntrospectedResource,
     raFetchMethod,
-    params: any,
+    { id, data }: any,
     queryType: IntrospectionField
 ) =>
-    Object.keys(params.data).reduce((acc, key) => {
-        if (Array.isArray(params.data[key])) {
-            const arg = queryType.args.find(a => a.name === `${key}Ids`);
+    Object.keys(data).reduce(
+        (acc, key) => {
+            if (Array.isArray(data[key])) {
+                const arg = queryType.args.find(a => a.name === `${key}Ids`);
 
-            if (arg) {
-                return {
-                    ...acc,
-                    [`${key}Ids`]: params.data[key].map(({ id }) => id),
-                };
+                if (arg) {
+                    return {
+                        ...acc,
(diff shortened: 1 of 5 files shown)


WiXSL (approved):

packages/ra-data-graphql-simple/src/buildVariables.ts
```diff
@@ -291,34 +291,37 @@ const buildGetListVariables = (introspectionResults: IntrospectionResult) => (
...
-            ...acc,
-            [key]: params.data[key],
-        };
-    }, {});
+            return {
+                ...acc,
+                [key]: data[key],
+            };
+        },
+        { id }
```
fzaninotto, line 326: on a create call, id isn't required. I'm surprised you return it in that case

djhi: Because it doesn't matter, it will be undefined. Tried it on the demo without issues

djhi: However, I'll include some fixes for the gql version of the demo, don't merge

djhi:
I wasn't able to fix the customer creation. There are some differences between the schema generated by our dataProvider and the one from json-graphql-server

fzaninotto (approved):

PR #7003, opened 2021-12-16
[New Feature] Add onSubmitRemove prop to FileInput for handling remove files with confirming

### LINKED ISSUE #6998: [Feature Request] Add onRemove props to FileInput so that we can confirm delete action

[FileInput](https://github.com/marmelab/react-admin/blob/4da50fd4901e0724d03ce3699e34f034e5492b7c/packages/ra-ui-materialui/src/input/FileInput.tsx#L148-L161) always delete preview components state. So, we can't implement it to confirm logic though we can remove files immediately before submitting form through options.onRemove prop. Also, we can't handle appropriately errors outer of FileInput if we have  removing files' errors, because we can't restore previous ui state having state removed completely before handling errors.

A solution is that we could add onRemove props FileInput it is different from options.onRemove for back compatibility and this function allow to return boolean or Promise\<boolean\>. This function is called before file input state removed, then when this returned value is false, interrupt removing state like this.

```tsx
   const {
    onRemove: onRemoveFromProp,
     ...others
   } = props

    const onRemove = file => async () => {
        const removeCancel = !(await onRemoveFromProp(file))
        if(removeCancel) return

        if (multiple) {
            const filteredFiles = files.filter(
                stateFile => !shallowEqual(stateFile, file)
            );
            onChange(filteredFiles as any);
        } else {
            onChange(null);
        }

        if (options.onRemove) {
            options.onRemove(file);
        }
    };
```

Enable to replace all children components of FileInput.

**Additional context**

tkow:
closes #6998

Add onSubmitRemove prop to FileInput for handling remove files with confirming.
We can add confirm process with immediately removing files  cases.
See the detail in #6998.


WiXSL (approved):

WiXSL:
Looks pretty good!

packages/ra-ui-materialui/src/input/FileInput.spec.tsx
```diff
@@ -247,6 +251,208 @@ describe('<FileInput />', () => {
         });
     });
 
+    describe('should stop to remove file field onSubmitReview return false', () => {
```
fzaninotto, line 254:
```suggestion
    describe('should call onSubmitReview on removal to allow developers to conditionally prevent the removal', () => {
```

packages/ra-ui-materialui/src/input/FileInput.tsx
```diff
@@ -33,6 +33,7 @@ export const FileInput = (
         maxSize,
         minSize,
         multiple = false,
+        onSubmitRemove,
```
fzaninotto, line 36: I'd call it `validateFileRemoval`, and make it return either an empty promise (validated) or a rejected promise (failed).

packages/ra-ui-materialui/src/input/FileInput.tsx
```diff
@@ -111,7 +112,11 @@ export const FileInput = (
         }
     };
 
-    const onRemove = file => () => {
+    const onRemove = file => async () => {
+        if (onSubmitRemove) {
+            const isSuccess = await onSubmitRemove(file);
+            if (!isSuccess) return;
```
fzaninotto, line 118:
```suggestion
            try {
                await validateFileRemoval(file);
            } catch(e) {
                return;
            }
```

docs/Inputs.md
```diff
@@ -278,6 +278,7 @@ To override the style of all instances of `<ImageInput>` using the [material-ui
...
+| `onSubmitRemove`   | Optional | `Function`          | -                               | Set remove button click handler. Main difference from options.onRemove is able to cancel to remove the file's list item if it returns false `(file) => boolean \| Promise<boolean>`                                                                                                                                                                                                                                                                         |
 | `options`       | Optional | `Object`             | `{}`                            | Additional options passed to react-dropzone's `useDropzone()` hook. See [the react-dropzone source](https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js)  for details .                                                                                                                                                          |
 
 `<FileInput>` also accepts the [common input props](./Inputs.md#common-input-props).
```
fzaninotto, line 284: You probably need to give an example usage for the new prop (kind of what you added to the GitHub issue)

packages/ra-ui-materialui/src/input/FileInput.spec.tsx
```diff
@@ -247,6 +251,208 @@ describe('<FileInput />', () => {
...
+            expect(onSubmit.mock.calls[0][0]).toEqual({
+                image: {
+                    src: 'test.png',
+                    title: 'cats',
+                },
+            });
+        });
+        it('error occurs', async done => {
+            const onSubmit = jest.fn();
+            // NOTE: We couldn't handle UnhandledPromiseRejection and make
```
fzaninotto, line 337: this should be fixed by the approach I suggest

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/FileInput.spec.tsx
tkow: This test is no longer needed thanks to your suggestion. It is enough  only to test when throwing exception.

tkow:
@fzaninotto 
I have fixed almost problems you have reviewed. Thank you.

tkow:
@fzaninotto
Could you check when you have time?

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7008, opened 2021-12-16
Move theme management responsability to react-admin

djhi:


packages/ra-ui-materialui/src/layout/Theme/useSetTheme.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react';
+import { DeprecatedThemeOptions } from '@mui/material';
+import { ThemeContext, ThemeSetter } from './ThemeContext';
+
+export const useSetTheme = (
```
fzaninotto, line 5: Please add jsDoc for usage

packages/ra-ui-materialui/src/layout/Theme/ThemeProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import {
+    createTheme,
+    ThemeProvider as MuiThemeProvider,
+    Theme,
+} from '@mui/material/styles';
+import { DeprecatedThemeOptions } from '@mui/material';
+import { ThemeContext, ThemeSetter } from './ThemeContext';
+
+export const ThemeProvider = ({
```
fzaninotto, line 12: Please add JSDoc for usage

packages/ra-core/src/core/CoreAdmin.tsx
```diff
@@ -3,7 +3,7 @@ import { ComponentType } from 'react';
 
 import CoreAdminContext from './CoreAdminContext';
 import CoreAdminUI from './CoreAdminUI';
-import { AdminProps } from '../types';
+import { CoreAdminProps } from '../types';
```
fzaninotto, line 6: I think we should colocate this type with the component

packages/ra-ui-materialui/src/layout/Theme/ThemeContext.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from 'react';
+import { DeprecatedThemeOptions } from '@mui/material';
```
fzaninotto, line 2: Frightening. Isn't there a better alternative?

packages/react-admin/package.json
```diff
@@ -36,14 +36,12 @@
         "@emotion/styled": "^11.3.0",
         "@mui/icons-material": "^5.0.1",
         "@mui/material": "^5.0.2",
-        "final-form": "^4.20.2",
+        "final-form": "^4.20.4",
```
fzaninotto, line 39: seems unrelated to this PR

fzaninotto (changes requested):
Much cleaner like that!

But your PR lacks a section in the upgrade guide, and a section in the Theming doc

packages/react-admin/package.json
djhi: That's just a line that moved, most probably automatically

packages/ra-ui-materialui/src/layout/Theme/useSetTheme.ts
djhi: Added

packages/ra-core/src/core/CoreAdmin.tsx
djhi: Done but a lot of other types could be collocated as well

packages/ra-ui-materialui/src/layout/Theme/ThemeProvider.tsx
djhi: This is almost an internal component. Do you want an example that would be a copy of `react-admin` `Admin` component?

packages/react-admin/src/Admin.tsx
```diff
@@ -125,32 +127,37 @@ const Admin = (props: AdminProps) => {
...
-                disableTelemetry={disableTelemetry}
-                menu={menu}
-                catchAll={catchAll}
-                theme={theme}
-                title={title}
-                loading={loading}
-                loginPage={loginPage}
-                logout={authProvider ? logoutButton : undefined}
-                ready={ready}
+        <ThemeProvider theme={theme}>
```
fzaninotto, line 130: I think the ThemeProvider should be in the AdminContext. It would facilitate our tests, as we'd only have to wrap mui components with AdminContext (just like we wrap core components with CoreAdminContext).

packages/ra-ui-materialui/src/layout/Theme/ThemeProvider.tsx
fzaninotto: No, that's fine.

packages/ra-ui-materialui/src/layout/Theme/ThemeProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </ThemeContext.Provider>
+    );
+};
+
+export interface ThemeProviderProps {
+    children: ReactNode;
+    theme: ThemeOptions;
+}
+
+const useTheme = (themeOverride?: ThemeOptions): [Theme, ThemeSetter] => {
```
fzaninotto, line 37: This one should be in its own file, with an example usage in jsDoc

fzaninotto (changes requested):
Still missing docs and upgrade instructions

packages/react-admin/src/Admin.tsx
djhi: Sure

packages/ra-ui-materialui/src/layout/Theme/ThemeProvider.tsx
djhi: We don't even export it. It's only used internally by the `ThemeProvider`

fzaninotto: I think I'm confused because you didn't document how to read and set the theme ;)

UPGRADE.md
```diff
@@ -1209,6 +1209,35 @@ Besides, some props supported by the previous implementation aren't anymore:
 
 `useGetMainList` was a modified version of `useGetList` designed to keep previous data on screen upon navigation. As [this is now supported natively by react-query](https://react-query.tanstack.com/guides/paginated-queries#better-paginated-queries-with-keeppreviousdata), this hook is no longer necessary and has been removed. Use `useGetList()` instead.
 
+## The MUI `<ThemeProvider>` is not set by `<Layout>` anymore
+
+This shouldn't impact your code unless you had a completely custom `<Layout>` component. If you do but still uses the default `<Admin>` component, you can safely remove the `ThemeProvider` from your `Layout`:
```
fzaninotto, line 1214: This is a bit misleading, so developers may wonder if they need to manually add the ThemeProvider themselves. You should explain that ThemeProvider is set by AdminContext rather than Layout

docs/Theming.md
```diff
@@ -332,6 +332,34 @@ const App = () => (
 );
 ```
 
+## Changing the theme programmatically
```
fzaninotto, line 335: use Title Case

fzaninotto (changes requested):
tests fail

UPGRADE.md
djhi: Not sure why github shows the old version but I added a sentence before this one as requested

PR #7011, opened 2021-12-18
precise id type when possible

### LINKED ISSUE #7009: Imprecise id type inference in some react-admin hooks (in particular in useListContext)
**What you were expecting:**

My resources use number ids so I usually type my resource like this:

```typescript
import { Record } from 'react-admin'
 
interface Tag extends Record {
  id: number
}
```

So when using `useListContext`, I expect the ids and selectedIds to be list of numbers.

```typescript
const { ids, selectedIds } = useListContext<Tag>() // I expect ids and selectedIds to be typed as number[] instead of Identifier[]
```

**What happened instead:**

Instead inferred type is Identifier[]

**Related code:**

https://github.com/marmelab/react-admin/blob/034a7ba0d546bc62b582cf0b0071c4f522f91309/packages/ra-core/src/controller/useListController.ts#L53-L86

I suggest to change the definition by the following code:

```typescript
export interface ListControllerProps<RecordType extends Record = Record> {
    ids: RecordType['id'][];
    onSelect: (ids: RecordType['id'][]) => void;
    onToggleItem: (id: RecordType['id']) => void;
    selectedIds: RecordType['id'][];
   ...
}
```

This change should probably be done everywhere where a type variable extend Record and Identifier is used in the scope of the extension.

nicgirault:


fzaninotto (approved):

fzaninotto:
Thanks a lot!

PR #7014, opened 2021-12-20
Fixed githu -> github in Changelog links

ZachSelindh:
Links to fixed issues on Changelog 3.19.4 update contained "githu" rather than "github", this commit fixes those links.


fzaninotto:
Thanks!

PR #7018, opened 2021-12-21
[BC Break] Cleanup Redux from now useless reducers

fzaninotto:
Now thet useGetMany, useGetManyReference, and useGetList rely on react-query rather than Redux, the reducers are no longer in use.


UPGRADE.md
```diff
@@ -548,11 +548,15 @@ const CommentGrid = () => {
...
-## Removed the `oneToMany` Reducer
+## Removed Reducers
 
-React-admin no longer relies on Redux to fetch one-to-many relationships. Instead, the cache of previously fetched relationships is managed by react-query.
+React-admin no longer relies on Redux to fetch relationships. Instead, the cache of previously fetched relationships is managed by react-query.
 
-If you need to get the records related to the current one via a one-to-many relationship (e.g. to fetch all the books of a given author), you can use the `useGetManyReference` hook instead.
+If you need to get the records related to the current one via a one-to-many relationship (e.g. to fetch all the books of a given author), you can use the `useGetManyReference` hook instead of the `oneToMany` reducer.
+
+If you need to get possible values for a relatinoship, use the `useGetList` hook instead of the `possibleValues` reducer.
```
WiXSL, line 557:
```suggestion
If you need to get possible values for a relationship, use the `useGetList` hook instead of the `possibleValues` reducer.
```

WiXSL (changes requested):

WiXSL (approved):

PR #7020, opened 2021-12-22
[BC Break] Update useUpdateMany to use react-query

fzaninotto:
- [x] Refactor `useUpdate` for better readability and portability
- [x] Update `useUpdateMany` to use react-query
- [x] Fix tests, docs and examples

```diff
-const [updateMany, { loading }] = useUpdateMany(
-   'posts',
-   [123, 456],
-   { likes: 12 },
-);
-updateMany(resource, id, data, previousData, options);
+const [updateMany, { isLoading }] = useUpdateMany(
+   'posts',
+   {
+       ids: [123, 456],
+       data: { likes: 12 },
+   }
+);
+updateMany(resource, { ids, data }, options);
```

### DIFF
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -481,7 +481,7 @@ const IncreaseLikeButton = ({ record }) => {
 const [updateMany, { data, isFetching, isLoading, error }] = useUpdateMany(resource, { ids, data }, options);
 ```
 
-The `update()` method can be called with the same parameters as the hook:
+The `updateMany()` method can be called with the same parameters as the hook:
 
 ```jsx
 updateMany(resource, { ids, data }, options);
--- a/docs/Actions.md
+++ b/docs/Actions.md
@@ -470,7 +470,7 @@ const IncreaseLikeButton = ({ record }) => {
     const diff = { likes: record.likes + 1 };
     const [update, { isLoading, error }] = useUpdate('likes', { id: record.id, data: diff, previousData: record });
     if (error) { return <p>ERROR</p>; }
-    return <button disabled={isLoading} onClick={update}>Like</button>;
+    return <button disabled={isLoading} onClick={() => update()}>Like</button>;
 };
 ```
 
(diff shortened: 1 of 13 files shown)


docs/Actions.md
```diff
@@ -470,42 +470,43 @@ const IncreaseLikeButton = ({ record }) => {
...
 // syntax
-const [updateMany, { data, loading, loaded, error }] = useUpdateMany(resource, ids, data, options);
+const [updateMany, { data, isFetching, isLoading, error }] = useUpdateMany(resource, { ids, data }, options);
 ```
 
-The `updateMany()` function can be called in 3 different ways:
- - with the same parameters as the `useUpdateMany()` hook: `update(resource, ids, data, options)`
- - with the same syntax as `useMutation`: `update({ resource, payload: { ids, data } }, options)`
- - with no parameter (if they were already passed to `useUpdateMany()`): `updateMany()`
+The `update()` method can be called with the same parameters as the hook:
```
WiXSL, line 484:
```suggestion
The `updateMany()` method can be called with the same parameters as the hook:
```

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -81,87 +82,57 @@ export const useUpdate = <RecordType extends Record = Record>(
...
+    const updateCache = ({ resource, id, data }) => {
         // hack: only way to tell react-query not to fetch this query for the next 5 seconds
         // because setQueryData doesn't accept a stale time option
-        const updatedAt =
-            mode.current === 'undoable' ? Date.now() + 1000 * 5 : Date.now();
-        await queryClient.setQueryData(
+        const now = Date.now();
+        const updatedAt = mode.current === 'undoable' ? now + 5 * 1000 : now;
+
+        const updateColl = (old: RecordType[]) => {
```
WiXSL, line 93: Rename `updateColl` to `updateCall`
```diff
-const updateColl = (old: RecordType[]) => {
+const updateCall = (old: RecordType[]) => {

**Review by WiXSL — CHANGES_REQUESTED** [review]

File: packages/ra-core/src/dataProvider/useUpdate.ts [hunk 3]
```diff
@@ -357,41 +289,43 @@ export const useUpdate = <RecordType extends Record = Record>(
...
+         *
+         * @see https://react-query.tanstack.com/reference/QueryClient#queryclientgetqueriesdata
+         */
+        snapshot.current = queryKeys.reduce(
+            (prev, curr) => prev.concat(queryClient.getQueriesData(curr)),
+            [] as Snapshot
+        );
+
+        // Cancel any outgoing re-fetches (so they don't overwrite our optimistic update)
+        await snapshot.current.map(([key]) => queryClient.cancelQueries(key));
```
WiXSL, line 325: ~~await~~ snapshot.current.map([key]) => queryClient.cancelQueries(key));

`await` has no effect here

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -1,108 +1,421 @@
...
+         *
+         * @see https://react-query.tanstack.com/reference/QueryClient#queryclientgetqueriesdata
+         */
+        snapshot.current = queryKeys.reduce(
+            (prev, curr) => prev.concat(queryClient.getQueriesData(curr)),
+            [] as Snapshot
+        );
+
+        // Cancel any outgoing re-fetches (so they don't overwrite our optimistic update)
+        await snapshot.current.map(([key]) => queryClient.cancelQueries(key));
```
WiXSL, line 326:
```suggestion
        snapshot.current.map(async ([key]) => await queryClient.cancelQueries(key));
```

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useUpdate.ts
WiXSL: It was a `forEach` before, If not using the result, why change it to map?

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -1,108 +1,421 @@
...
+                        callTimeIds,
+                        { resource: callTimeResource, ...callTimeParams },
+                        { snapshot: snapshot.current }
+                    ),
+                0
+            );
+        }
+
+        if (mode.current === 'optimistic') {
+            // call the mutate without success side effects
```
WiXSL, line 360:
```suggestion
            // call the mutate method without success side effects
```

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -1,108 +1,421 @@
...
+        } else {
+            // undoable mutation: register the mutation for later
+            undoableEventEmitter.once('end', ({ isUndo }) => {
+                if (isUndo) {
+                    // rollback
+                    snapshot.current.forEach(([key, value]) => {
+                        queryClient.setQueryData(key, value);
+                    });
+                } else {
+                    // call the mutate without success side effects
```
WiXSL, line 374:
```suggestion
                    // call the mutate method without success side effects
```

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useUpdate.ts
fzaninotto: its short for "updateCollection"... But maybe it's not clear enough. I preferred the abbreviation to the more explicit "updateCollection" to avoid triggering more line breaks, which harms legibility.

WiXSL: Ok, my bad 😄

packages/ra-core/src/dataProvider/useUpdate.ts
fzaninotto: I'll use Promise.all() instead

WiXSL (approved):

PR #7025, opened 2021-12-23
[BC Break] Migrate useCreate to react-query

fzaninotto:
- [x] Update `useCreate` to use react-query
- [x] Replace `onSuccess` and `onFailure` `<Create>` props with `mutationOptions`
- [x] Fix tests, docs and examples

```diff
// useCreate
-const [create, { loading }] = useCreate(
-   'posts',
-   { title: 'hello, world!' },
-);
-create(resource, data, options);
+const [create, { isLoading }] = useCreate(
+   'posts',
+   { data: { title: 'hello, world!' } }
+);
+create(resource, { data }, options);
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -179,6 +179,18 @@ For queries:
 For mutations:
 
 ```diff
+// useCreate
+-const [create, { loading }] = useCreate(
+-   'posts',
+-   { title: 'hello, world!' },
+-);
+-create(resource, data, options);
++const [create, { isLoading }] = useCreate(
++   'posts',
++   { data: { title: 'hello, world!' } }
++);
++create(resource, { data }, options);
+
 // useUpdate
 -const [update, { loading }] = useUpdate(
 -   'posts',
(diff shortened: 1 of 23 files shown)


WiXSL:
💪

WiXSL:
e2e create test failing

WiXSL (changes requested):
`crm` project build fails because `onSuccess={onSuccess}` on  `examples/crm/src/deals/DealCreate.tsx`

WiXSL (changes requested):
The demo throws several errors showing the create view of a product

fzaninotto:
For a reason that I don't understand, the "Save and add" feature starts breaking the e2e tests in this branch, while it seems to be also broken in `next`...

fzaninotto:
OK, the "Save and add" feature is broken on next but the e2e tests don't see it because they don't *navigate* to the creation form via the menu, they enter the create URL in the address bar. This makes a big difference: the location state contains a _scrollToTop key when navigating with the mouse, which changes the behavior of useRedirect.

Anyway, this is not related to this PR, more a regression in next. I'll open a bug so that we remember to fix it in v4.

WiXSL (approved):

PR #7028, opened 2021-12-23
Fast cache lookup for useGetMany(Aggregate) placeholderData

wattroll:
**Problem:**

`ReferenceField` render can take up to several milliseconds when react-query cache contains large number of records (hundreds, thousands).

**Steps to reproduce:**

Load a datagrid with large number of `ReferenceField`s (e.g. more than a hundred per page), switch between pages to populate the cache. When the query cache grows to several hundreds/thousands records switching between pages will start feeling sluggish. `ReferenceField`s that refer to an already cached data will take visible amount of time to update in response to page changes.

**What's causing this:**

`useGetMany` and `useGetManyAggregate` prepare `placeholderData` to be displayed why `useQuery` is running. To do that they call `QueryClient.getQueryData(queryKey)`. This is reasonable and idiomatic usage, but unfortunately [`getQueryData(queryKey)`] works like this: 

```javascript
queryCache.queries.find(query => (query.hashQueryKey || hashQueryKey)(queryKey) == query.queryHash))
```

Not only it walks through the ever-growing list, but on each element it recalculates the constant `hashQueryKey(queryKey)` which is a not so fast `JSON.stringify` call.

[`getQueryData(queryKey)`]: https://github.com/tannerlinsley/react-query/blob/b977cf4ddd912d2ff369ef06683e462be4155ebd/src/core/queryClient.ts#L110

**Solution:**

To satisfy our access pattern it's enough to do this instead:

```javascript
queryCache.get(queryHash(queryKey))
```

which is as fast as it can reasonably be
```javascript
queryCache.get = (queryHash) => queryCache.queriesMap[queryHash]
```

Ideally this optimization should be done in `react-query`. For the time being I think it's an acceptable solution and the code change is not too hacky.

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetManyAggregate.ts
+++ b/packages/ra-core/src/dataProvider/useGetManyAggregate.ts
@@ -4,6 +4,7 @@ import {
     useQueryClient,
     useQuery,
     UseQueryOptions,
+    hashQueryKey,
 } from 'react-query';
 import union from 'lodash/union';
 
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -66,15 +67,13 @@ export const useGetManyAggregate = <RecordType extends Record = Record>(
 ): UseGetManyHookValue<RecordType> => {
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
+    const queryCache = queryClient.getQueryCache();
     const { ids } = params;
     const placeholderData = useMemo(() => {
-        const records = ids.map(id =>
-            queryClient.getQueryData<RecordType>([
```
fzaninotto, line 72: see the linter warning, you've added a new dependency to useMemo

wattroll: Amended. Will remember to lint in the future.

wattroll:
Thanks for a quick review.

I was thinking that it would be nice to wrap up the duplicated code, but decided to keep this change as small as possible and don't refactor things much until react-query integration is finished. Then it might be nice to add a `getCachedData`/`useCache` function.

fzaninotto (approved):

fzaninotto:
Thanks again for this one! I'm glad that you already gave `next` a try.

About the refactoring, let's wait until we need to do this more than twice, because splitting the logic in different places makes the code harder to read and maintain (see e.g. the dataProvider logic in `master`). Also, I'd really love if react-query fixed this issue themselves. Could you please open an issue on their repo?

PR #7031, opened 2021-12-23
Fix `SelectInput` create optionText

WiXSL:
Same fix as in #7030


packages/ra-ui-materialui/src/input/SelectInput.tsx
```diff
@@ -219,18 +243,6 @@ export const SelectInput = (props: SelectInputProps) => {
         );
     }
 
-    const renderCreateItem = () => {
```
WiXSL, line 222: remove `renderCreateItem`, `renderMenuItem` renders all the items

fzaninotto (approved):

PR #7032, opened 2021-12-23
Submit disabled prop from filter -> MenuItem

ZachSelindh:
The MUI MenuItem being used to render FilterButtonMenuItem accepts the prop "disabled", which will prevent the user from selecting that menu item if true.

This PR allows submitting disabled as a prop to filters, which will display them as an option, but prevent the user from selecting/adding that filter.

Our use case: We want to conditionally allow access to certain fields depending on user permissions, and allowing setting this prop on filters will prevent the user from creating filters using fields they don't have access to, while showing that the field still exists.


fzaninotto:
Hi, and thanks for your PR. It is indeed a nice addition and a valid use case. 

React-admin version 3, which is the `master` branch, is feature-freeze. All new features (and your PR is a new feature) should land in the `next` branch, which will be react-admin v4. 

Could you please change the base branch of your PR?

fzaninotto:
I think you failed something in rebasing, this PR contains commits that aren't yours...

ZachSelindh:
> I think you failed something in rebasing, this PR contains commits that aren't yours...

Looks like manually changing the target branch didn't work as I would have expected! Thank you, I'll try a new PR based on 'next'.

PR #7044, opened 2021-12-30
Fix loading indicator not updated on permissions error when logoutUser is false

### LINKED ISSUE #6750: Progress Indicator/Refresh button spins constantly on checkError reject with logoutUser: false
Once rejected with the refresh button spins until page refresh.

Screen Shot 2021-11-05 at 10 31 28 AM

```js
checkError: (error: any) => {
    const status = error.status;
    if (status === 401 || status === 403) {
      if (status === 401)
      {
        localStorage.removeItem('auth');
        localStorage.removeItem('permissions');
        return Promise.reject();
      }

      return Promise.reject({ redirectTo: '/', logoutUser: false });
    }
    return Promise.resolve();
  }
```

WiXSL:
Fixes #6750

### DIFF
--- a/packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
+++ b/packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
@@ -4,6 +4,8 @@ import useAuthProvider from './useAuthProvider';
 import useLogout from './useLogout';
 import { useNotify } from '../sideEffect';
 import { useHistory } from 'react-router';
+import { fetchEnd } from '../actions';
+import { useDispatch } from 'react-redux';
 
 let timer;
 
@@ -43,6 +45,8 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
     const logout = useLogout();
     const notify = useNotify();
     const history = useHistory();
+    const dispatch = useDispatch();
+
     const logoutIfAccessDenied = useCallback(
         (error?: any, disableNotification?: boolean) =>
             authProvider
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -92,12 +96,13 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
                     if (logoutUser) {
                         logout({}, redirectTo);
                     } else {
+                        dispatch(fetchEnd());
```
fzaninotto, line 99: i don't understand why this is necessary, as performXXXQuery already dispatches a FAILURE action with a FETCH_ERROR meta, that the loading reducer should use to decrease the loading state. 

Could you elaborate in a comment?

Also, if a dataProvider action is necessary, I'd prefer it to be dispatched in performXXXQuery rather than in this hook.

fzaninotto (changes requested):

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
WiXSL: The problem is that callback always returns `true`. If I make it return `logoutUser`, the user gets logged out when `logoutUser` is `false`

fzaninotto: you're right. then the change should be done in each of the 3 performXXXQuery scripts instead: if `loggedOut` is true, the method should dispatch a FETCH_END before returning.

fzaninotto (approved):

PR #7051, opened 2022-01-04
[BC Break] Remove Resource initialization 

### LINKED ISSUE #5856: Resource with icon, conflict with Redux (non-serializable value)
**What you were expecting:**
I followed official documentation setting my own <Resource> with icons (from @material-ui/icons)
Using <Resource ... icon={MyIcon} ...>, no warning raises from redux.
 
**What happened instead:**
Using the app, I get several times messages in my console like the following:

```
A non-serializable value was detected in the state, in the path: `admin.resources.my_resource.props.icon.$$typeof`. Value: Symbol(react.memo)
Take a look at the reducer(s) handling this action type: RA/FETCH_START
```

(this react.memo is the Icon being stored)
(It happens in many actions, `FETCH_START` is just an example.)

**Steps to reproduce:**
I haven't tried to replicate this issue on an empty react-admin project yet. If you never saw that before, it may definitely comes from my setup, feel free to close this issue.

A while ago, using plain javascript, I was "cheating" storing icon as a "string" and retrieving icon in my custom menu by hand. That worked for a while. I'm switching to typescript and you Resource.icon types expects a ComponentType, not a string. Your solution with icon={MyIcon} seems much cleaner anyway. The number of warning in the web console makes me think we're doing something wrong ?

**Environment**

* React-admin version: 3.12 (but I already had this problem on earlier versions)
* React: 16

fzaninotto:
- [x] Move resource-dependent reducers up so they don't need initialization
- [x] Move Resource Definition from Redux to a custom context
- [x] Fix tests
- [x] Write upgrade guide

Closes #5856


packages/ra-core/src/core/useRegisterResource.ts
```diff
@@ -1,23 +1,14 @@
...
-import { ReduxState, ResourceDefinition } from '../types';
+import { ResourceDefinitionContext } from './ResourceDefinitionContext';
+import { ResourceDefinition } from '../types';
 
 export const useRegisterResource = () => {
-    const dispatch = useDispatch();
-    const knownResources = useSelector<ReduxState>(
-        state => state.admin.resources
-    );
+    const [_, setResourceConfiguration] = useContext(ResourceDefinitionContext);
```
djhi, line 7:
```suggestion
    const [, setResourceConfiguration] = useContext(ResourceDefinitionContext);
```

djhi (changes requested):
Great!

djhi (approved):

PR #7058, opened 2022-01-06
Remove no longer necessary HACK around a @mui/material problem

wattroll:
## Problem

A strange typing of `MenuItemProps` has been employed to work around mui-org/material-ui#16245

This was causing the following code to fail with a strange type error:

```typescript
createElement(AdminUI, {
  layout: Layout,
  dashboard: Dashboard,
  catchAll: NotFound,
  loginPage: Login as LoginComponent,
  logout: Logout,
})
```

## Solution

The issue has been resolved. The hack can be removed. Type error in my code was resolved, by removing it from `Logout`. I also found the same hack in `MenuItemLinkProps` and did the same change there, but haven't tested its effect.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #7060, opened 2022-01-07
[BC Break] Add ReferenceOneField

fzaninotto:
## Problem

I started documenting the react-admin ReferenceXXXfFields, and found ReferenceOne to be a missing spot in our strategy.

## Solution

- [x] Remove props injection in ReferenceField
- [x] Add ReferenceOneField (and the `useReferenceOneFieldController`), reusing the view from `<ReferenceField>`
- [x] Add documentation for relationship types and the react-admin solutions to handle them

### DIFF
--- a/docs/FieldsForRelationships.md
+++ b/docs/FieldsForRelationships.md
@@ -24,11 +24,11 @@ To fetch the books of an author with react-admin, you can use:
 
 - [`<ReferenceManyField>`](#referencemanyfield) when the API uses a foreign key (e.g. each book has an `author_id` field)
 - [`<ReferenceArrayField>`](#referencearrayfield) when the API uses an array of foreign keys (e.g. each author has a `book_ids` field)
-- [`<ArrayField>`](#arrayfield) when the API embeds an array of records (e.g. each author has an `books` field)
+- [`<ArrayField>`](#arrayfield) when the API embeds an array of records (e.g. each author has a `books` field)
 
 ## Many-To-One
 
-On the other hand, **many-to-one relationships** are the opposite of one-to-many relationships (e.g. each book has one author). To fetch the the author of a book, you can use:
+On the other hand, **many-to-one relationships** are the opposite of one-to-many relationships (e.g. each book has one author). To fetch the author of a book, you can use:
 
 - [`<ReferenceField>`](#referencefield) when the API uses a foreign key (e.g. each book has an `author_id` field)
 - [Deep Field Source](#deep-field-source), when the API embeds the related record (e.g. each book has an `author` field containing an object)
@@ -37,7 +37,7 @@ Other kinds of relationships often reduce to one-to-many relationships.
 
 ## One-To-One
 
-For instance, **one-to-one relationships** (e.g. a book has one book_detail) are a special type of one-to-many relationship with a cardinality of 1. To fetch the details of a book, you can use:
+For instance, **one-to-one relationships** (e.g. a book has one `book_detail`) are a special type of one-to-many relationship with a cardinality of 1. To fetch the details of a book, you can use:
 
 - [`<ReferenceOneField>`](#referenceonefield) when the API uses a foreign key (e.g. each `book_detail` has a `book_id` field)
 - [`<ReferenceField>`](#referencefield) when the API uses a reverse foreign key (e.g. each `book` has a `book_detail_id` field)
@@ -167,7 +167,7 @@ const BookShow = () => (
 
 `<ReferenceField>` creates a `RecordContext` with the reference record, so you can use any component relying on this context (`<TextField>`, `<SimpleShowLayout>`, etc.).
 
-**Tip**: You don't need to worry about the fact that this components calls `<ReferenceField>` twice on the same table. React-admin will only make one call to the API.
+**Tip**: You don't need to worry about the fact that these components calls `<ReferenceField>` twice on the same table. React-admin will only make one call to the API.
 
 This is fine, but what if you need to display the author details for a list of books?
 
@@ -231,7 +231,7 @@ const AuthorShow = () => (
 
 `<ReferenceManyField>` creates a `ListContext` with the related records, so you can use any component relying on this context (`<Datagrid>`, `<SimpleList>`, etc.).
 
-**Tip**: For many APIs, there is no difference between `dataProvider.getList()` and `dataProvider.getManyReference()`. The latter is a specialized version of the former, with a predefined `filter`. But some APIs expose related records as a subroute, and therefore need a special method to fetch them. For instance, the boks of an author can be exposed via the following endpoint: 
+**Tip**: For many APIs, there is no difference between `dataProvider.getList()` and `dataProvider.getManyReference()`. The latter is a specialized version of the former, with a predefined `filter`. But some APIs expose related records as a sub-route, and therefore need a special method to fetch them. For instance, the books of an author can be exposed via the following endpoint: 
 
 ```
 GET /authors/:id/books
@@ -405,7 +405,7 @@ const AuthorShow = () => (
 );
 ```
 
-`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the same the foreign key (`book_id`). Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` fetch the related details, and takes the first one.
+`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the foreign key (`book_id`). Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` to fetch the related details, and takes the first one.
 
 `<ReferenceOneField>` creates a `RecordContext` with the reference record, so you can use any component relying on this context (`<TextField>`, `<SimpleShowLayout>`, etc.).
 
--- a/docs/FieldsForRelationships.md
+++ b/docs/FieldsForRelationships.md
@@ -127,6 +127,8 @@ const AuthorShow = () => (
 );
 ```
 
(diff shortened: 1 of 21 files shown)


djhi (approved):

WiXSL:
Reviewing...

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+
+To fetch the books of an author with react-admin, you can use:
+
+- [`<ReferenceManyField>`](#referencemanyfield) when the API uses a foreign key (e.g. each book has an `author_id` field)
+- [`<ReferenceArrayField>`](#referencearrayfield) when the API uses an array of foreign keys (e.g. each author has a `book_ids` field)
+- [`<ArrayField>`](#arrayfield) when the API embeds an array of records (e.g. each author has an `books` field)
+
+## Many-To-One
+
+On the other hand, **many-to-one relationships** are the opposite of one-to-many relationships (e.g. each book has one author). To fetch the the author of a book, you can use:
```
WiXSL, line 31:
```suggestion
On the other hand, **many-to-one relationships** are the opposite of one-to-many relationships (e.g. each book has one author). To fetch the author of a book, you can use:
```

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+        </SimpleShowLayout>
+    </Show>
+);
+```
+
+`<ReferenceField>` uses the current `record` (a book in this example) to read the id of the reference using the foreign key (`author_id`). Then, it uses `dataProvider.getOne('authors', { id })` fetch the related author.
+
+`<ReferenceField>` creates a `RecordContext` with the reference record, so you can use any component relying on this context (`<TextField>`, `<SimpleShowLayout>`, etc.).
+
+**Tip**: You don't need to worry about the fact that this components calls `<ReferenceField>` twice on the same table. React-admin will only make one call to the API.
```
WiXSL, line 170:
```suggestion
**Tip**: You don't need to worry about the fact that these components calls `<ReferenceField>` twice on the same table. React-admin will only make one call to the API.
```

WiXSL (changes requested):

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+
+## One-To-Many
+
+When one record has many related records, this is called a one-to-many relationship. For instance, if an author has written several books, `authors` has a one-to-many relationship with `books`. 
+
+To fetch the books of an author with react-admin, you can use:
+
+- [`<ReferenceManyField>`](#referencemanyfield) when the API uses a foreign key (e.g. each book has an `author_id` field)
+- [`<ReferenceArrayField>`](#referencearrayfield) when the API uses an array of foreign keys (e.g. each author has a `book_ids` field)
+- [`<ArrayField>`](#arrayfield) when the API embeds an array of records (e.g. each author has an `books` field)
```
WiXSL, line 27:
```suggestion
- [`<ArrayField>`](#arrayfield) when the API embeds an array of records (e.g. each author has a `books` field)
```

WiXSL (changes requested):

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+On the other hand, **many-to-one relationships** are the opposite of one-to-many relationships (e.g. each book has one author). To fetch the the author of a book, you can use:
+
+- [`<ReferenceField>`](#referencefield) when the API uses a foreign key (e.g. each book has an `author_id` field)
+- [Deep Field Source](#deep-field-source), when the API embeds the related record (e.g. each book has an `author` field containing an object)
+
+Other kinds of relationships often reduce to one-to-many relationships. 
+
+## One-To-One
+
+For instance, **one-to-one relationships** (e.g. a book has one book_detail) are a special type of one-to-many relationship with a cardinality of 1. To fetch the details of a book, you can use:
```
WiXSL, line 40:
```suggestion
For instance, **one-to-one relationships** (e.g. a book has one `book_detail`) are a special type of one-to-many relationship with a cardinality of 1. To fetch the details of a book, you can use:
```

WiXSL (changes requested):

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+        </SimpleShowLayout>
+    </Show>
+);
+```
+
+`<ReferenceManyField>` uses the current `record` (an author in this example) to build a filter for the list of books on the foreign key field (`author_id`). Then, it uses `dataProvider.getManyReference('books', { target: 'author_id', id: book.id })` fetch the related books.
+
+`<ReferenceManyField>` creates a `ListContext` with the related records, so you can use any component relying on this context (`<Datagrid>`, `<SimpleList>`, etc.).
+
+**Tip**: For many APIs, there is no difference between `dataProvider.getList()` and `dataProvider.getManyReference()`. The latter is a specialized version of the former, with a predefined `filter`. But some APIs expose related records as a subroute, and therefore need a special method to fetch them. For instance, the boks of an author can be exposed via the following endpoint:
```
WiXSL, line 234:
```suggestion
**Tip**: For many APIs, there is no difference between `dataProvider.getList()` and `dataProvider.getManyReference()`. The latter is a specialized version of the former, with a predefined `filter`. But some APIs expose related records as a sub-route, and therefore need a special method to fetch them. For instance, the books of an author can be exposed via the following endpoint: 
```

WiXSL (changes requested):

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+            <ReferenceOneField label="ISBN" reference="book_details" target="book_id">
+                <TextField source="ISBN" />
+            </ReferenceManyToManyField>
+            <EditButton />
+        </SimpleShowLayout>
+    </Show>
+);
+```
+
+`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the same the foreign key (`book_id`). Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` fetch the related details, and takes the first one.
```
WiXSL, line 408:
```suggestion
`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the foreign key (`book_id`). Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` to fetch the related details, and takes the first one.
```

docs/FieldsForRelationships.md
```diff
@@ -0,0 +1,414 @@
...
+
+`<ReferenceOneField>` creates a `RecordContext` with the reference record, so you can use any component relying on this context (`<TextField>`, `<SimpleShowLayout>`, etc.).
+
+**Tip**: As with `<ReferenceField>`, you can call `<ReferenceOneField>` as many times as you need in the same component, react-admin will only make one call to `dataProvider.getManyReference()`.
```
WiXSL, line 412:
```suggestion
**Tip**: As with `<ReferenceManyField>`, you can call `<ReferenceOneField>` as many times as you need in the same component, react-admin will only make one call to `dataProvider.getManyReference()`.
```

WiXSL (changes requested):

WiXSL (changes requested):
`ColoredNumberField` component in the demo is also using `record` as a prop instead of taking it from the context

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return !record || (!isLoading && referenceRecord == null) ? (
+        emptyText ? (
+            <Typography component="span" variant="body2">
+                {emptyText}
+            </Typography>
+        ) : null
+    ) : (
+        <ResourceContextProvider value={reference}>
+            <ReferenceFieldView
+                isLoading={isLoading}
```
WiXSL, line 69: I get a typescript error here: 
`Property 'isLoading' does not exist on type 'IntrinsicAttributes & ReferenceFieldViewProps & { children?: ReactNode; }'.`

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -169,21 +167,12 @@ const stopPropagation = e => e.stopPropagation();
 
 export const ReferenceFieldView: FC<ReferenceFieldViewProps> = props => {
     const {
-        basePath,
         children,
         className,
         error,
```
WiXSL, line 172: I get Typescript errors:

`TS2339: Property 'error' does not exist on type 'ReferenceFieldViewProps & { children?: ReactNode; }'.`
`TS2339: Property 'isLoading' does not exist on type 'ReferenceFieldViewProps & { children?: ReactNode; }'.`
`TS2339: Property 'referenceRecord' does not exist on type 'ReferenceFieldViewProps & { children?: ReactNode; }'.`

WiXSL (changes requested):

WiXSL:
Finish reviewing.
The new documentation explaining in detail the relationships fields looks great!

packages/ra-ui-materialui/src/field/ReferenceField.tsx
fzaninotto: Weird. I don't see it and neither does the TypeScript compilation step in the CI. Did you rebuild the ra-core types?

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
fzaninotto: same

WiXSL (approved):

packages/ra-ui-materialui/src/field/ReferenceField.tsx
WiXSL: You are right. I re-build everything and it works

PR #7063, opened 2022-01-09
Fix dead link in DataProviders.md

Fabious:
Hi!
Fixing dead link as we have migrated to TypeScript =]

Thanks!


fzaninotto:
Thanks!

PR #7064, opened 2022-01-09
Fix missing type for TranslationMessages

bicstone:
Fix missing type for TranslationMessages.

(`ra.action.open`)

https://github.com/marmelab/react-admin/commit/f94b669d9db3a0f13cb092666f3b11bd7ca424f5#diff-d55e636245f2994de4a9a74002cca1fec0a6a2e68d1805a5439a0c885f37263b


djhi (approved):
Thanks!

PR #7069, opened 2022-01-11
Fix propTypes.label for reference fields

wattroll:
## Problem:

Prop types of reference fields specify `label: PropTypes.string`, even though `type PublicFieldProps` says `label?: string | ReactElement | boolean`. A big red error is printed to the console when using an element as label.

## Solution:

Use common `fieldPropTypes.label` definition in each propType definition.

---

I am being a bit sloppy here, as this just fixes the warnings that bug me. But it makes the codebase a bit better and should be easy to merge. A bigger refactoring of propTypes is needed. The same issue is with inputs.

I didn't investigate why `Reference*Field` propTypes don't extend the `fieldPropTypes` like other fields do, probably they have to override something.


djhi (approved):
Thanks!

PR #7077, opened 2022-01-13
Migrate to Yarn 2

djhi:
First part of the migration. Another PR will enable PnP if possible


.gitignore
```diff
@@ -16,3 +16,10 @@ cypress/videos
 cypress/screenshots
 /public
 /storybook-static
+.pnp.*
+.yarn/*
```
fzaninotto, line 20: yet there is a file in the .yarn repository added in this PR

fzaninotto (changes requested):

.gitignore
djhi: Yes, look at the following lines ;-)

djhi: I follower the yarn 2 documentation: https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

fzaninotto: right, thanks!

fzaninotto (approved):

PR #7078, opened 2022-01-13
[BC Break] Rename Record type to RaRecord

### LINKED ISSUE #5074: Change TypeScript's name of "Record"

No problem. It's just a suggestion.

I found that `ra-core` is using the name `Record` on the master branch.
https://github.com/marmelab/react-admin/blob/ecd60f6476a8394bf2bdf5f470713b0b87581777/packages/ra-core/src/types.ts#L17

It conflicts with the same naming UtilityType which provided by TypeScript own. Can we change it?

(About `Record` in TypeScript doc: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt )

So, it's not a critical issue, but a bit confusing to TypeScript users.

Change `Record` and other related naming types to other names, how about `Data`? or `Entity`?

We can import `Record` as an arbitrary name by `import type { Record as AdminRecord } from 'ra-core'

**Additional context**

I already mentioned the same thing in #4505, but I decided to post it as an issue; https://github.com/marmelab/react-admin/issues/4505#issuecomment-662362962
Thanks to @Tymek.

fzaninotto:
- [x] [BC Break] Rename type
- [x] Make it build
- [x] Add Upgrade instructions

Closes #5074

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1804,7 +1804,7 @@ Data Provider methods used to return records with a generic `Record` type, unles
 
 We've renamed that type to `RaRecord` to avoid any confusion.
 
-If you've declared custom Record types, you'll need to upgrade your codea s follows:
+If you've declared custom Record types, you'll need to upgrade your code as follows:
 
 ```diff
 -import { Record } from 'react-admin';
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1668,9 +1668,6 @@ const Toolbar = (props) => {
 }
 ```
 
-
-
-
 We removed the `handleSubmit` and `handleSubmitWithRedirect` props completely. Instead, when provided any of the side effects props, the `<SaveButton>` will render a simple button and will call the `save` function with them. It also take care of preventing the default form submit.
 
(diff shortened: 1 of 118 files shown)


fzaninotto:
Note: I'm not particularly fond of `RaRecord`, but I'm not sure the alternatives are better:

- `Entity` would be the correct term, but it doesn't clic if you didn't study DDD, and it conflicts wit hthe term 'record' that we use all arount the documentation 
- `Data` is too generic
- `RecordData` is... meh
- `RecordType` doesn't follow the pattern we use (never append or prepend 'type'  or 'interface'). Same for `IRecord`.
- `AdminRecord` doesn't mean anything

If someone comes up with a better name before we release 4.0, the renaming should be straightforward, so please voice your opinion!

djhi (changes requested):

WiXSL (approved):
Other that the conflicts, everything looks ok.

djhi (approved):

WiXSL (commented):
Hold on, I'm trying the demo, I think I've found something...

PR #7082, opened 2022-01-13
[BC Break] Use context to store notifications instead of Redux

fzaninotto:
- [x] Store notifications in a context rather than Redux
- [x] Include `<Notification />`in `<AdminUI>` so that users don't need to add it to their layout and login
- [x] Remove notification reducer and actions
- [x] Fix tests
- [x] Write Upgrade guide


packages/ra-core/src/notification/NotificationContextProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }, []);
+
+    return (
+        <NotificationContext.Provider
+            value={{
+                notifications,
+                addNotification,
+                takeNotification,
+                resetNotifications,
+            }}
```
djhi, line 33: You should put that inside a `useMemo`

UPGRADE.md
```diff
@@ -716,6 +716,26 @@ const PostEdit = () => {
 };
 ```
 
+## `useNotify` Now Takes An Options Object
+
+When a component has to display a notification, developers may want to tweak the type, duration, translatino arguments, or the ability to undo the action. The callback returned by `useNotify()` used to accept a long series of argument, but the syntax wasn't very intuitive. To improve the developer experience, these options are now part of an `options` object, passed as second argument.
```
djhi, line 721: s/translatino/translation

djhi (changes requested):
There are a few eslint warnings but otherwise :+1:

docs/useNotify.md
```diff
@@ -61,3 +63,26 @@ const PostEdit = () => {
     );
 }
 ```
+
+## `autoHideDuration` Option
+
+You can define a custom delay for hiding for a given notification.
```
WiXSL, line 69:
```suggestion
You can define a custom delay for hiding a given notification.
```

WiXSL (changes requested):

UPGRADE.md
```diff
@@ -716,6 +716,26 @@ const PostEdit = () => {
...
+    const notify = useNotify();
+    const handleClick = () => {
+-       notify(`Comment approved`, 'success', undefined, true);
++       notify(`Comment approved`, { type: 'success', undoable: true });
+    }
+    return <button onClick={handleClick}>Notify</button>;
+};
+```
+
+Check [the `useNotify`documentation](https://marmelab.com/react-admin/useNotify.html) for more information.
```
WiXSL, line 737:
```suggestion
Check [the `useNotify` documentation](./useNotify.md) for more information.
```

WiXSL (changes requested):

UPGRADE.md
fzaninotto: The upgrade guide won't necessarily be published with the rest of the documentation, so I prefer to be explicit.

WiXSL (approved):
Other than the conflicts, seems ok

djhi (approved):

PR #7084, opened 2022-01-14
SimpleList: add a divider prop

wattroll:
Adds a `divider` prop that goes straight onto mui's `ListItem`.

MUI documentation:
- https://mui.com/components/dividers/#list-dividers
- https://mui.com/api/list-item/#props

### DIFF
--- a/docs/SimpleList.md
+++ b/docs/SimpleList.md
@@ -44,6 +44,7 @@ It accepts the following props:
 * [`rightAvatar`](#rightavatar)
 * [`rightIcon`](#righticon)
 * [`rowStyle`](#rowstyle)
+* [`divider`](#divider)
 
 ## `leftAvatar`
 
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Hi and thanks for opening a PR. I think a better solution would be to accept a `ListItemProps` prop just like MUI does for many components. Otherwise, we'll end up duplicating all `ListItem` props. Besides, you declared the `divider` prop but you don't pass it to `ListItem`.

wattroll:
@djhi :+1: That makes a lot of sense. I'll close this and open a new PR.

---

>  Besides, you declared the `divider` prop but you don't pass it to `ListItem`.

Right :facepalm:. Excuse my very weak JSX (I haven't ever used it in my own code). 

```xml
<ListItem disablePadding divider>
```

By seeing `disablePadding` I assumed that JSX supports property punning and this will become
```javascriptcreateElement(ListItem, { disablePadding, divider })`

wattroll:
Thanks for a quick and very to-point response.

PR #7087, opened 2022-01-14
[BC Break] React Hook Form

djhi:
Supersedes #6730

- [x] Migrate FormWithRedirect
- [x] Migrate useInput
- [x] Migrate form groups
- [x] Migrate forms
- [x] Migrate inputs
- [x] Reintroduce form level validation
- [x] Fix warnings in tests
- [x] Fix examples & demos
- [x] Update documentation
- [x] Update upgrade guide


packages/ra-core/src/form/Form.tsx
```diff
@@ -0,0 +1,208 @@
...
+            notify('ra.message.invalid_form', { type: 'warning' });
+        }
+    }, [isInvalid, notify]);
+    useInitializeFormWithRecord(defaultValues, record);
+    useWarnWhenUnsavedChanges(warnWhenUnsavedChanges, formRootPathname);
+
+    const handleSubmit = async (event: BaseSyntheticEvent) => {
+        // Prevent outer forms to receive the event
+        event.stopPropagation();
+        formHandleSubmit(event);
```
WiXSL, line 175:
```suggestion
        await formHandleSubmit(event);
```

WiXSL: Not sure if this is right

WiXSL (changes requested):

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+            required={isRequired}
+            {...rest}
+        />
+    );
+};
+```
+
+## `allowEmpty` Has Been Removed From `SelectInput`, `AutocompleteInput` and `AutocompleteArrayInput`
+
+This is necessary for the underlying MaterialUI inputs as they require that the current input value is actually one of the available option. Those components now always accept an empty value (an empty string by default). If you require the input to have a non empty value, use the `required` validation. You can safely remove this prop.
```
WiXSL, line 2065:
```suggestion
This is necessary for the underlying MaterialUI inputs as they require that the current input value is actually one of the available option. Those components now always accept an empty value (an empty string by default). If you require the input to have a non-empty value, use the `required` validation. You can safely remove this prop.
```

WiXSL (changes requested):

examples/demo/src/visitors/SegmentsInput.tsx
```diff
@@ -22,7 +22,6 @@ const SegmentsInput = ({ addField, ...rest }: Props) => {
 };
 
 SegmentsInput.defaultProps = {
-    addField: true,
```
WiXSL, line 25: `addField`: Must be removed from line 10 as well in order for the demo to build

WiXSL (changes requested):

WiXSL (commented):
How about adding new stories related to forms (In another PR maybe? I can help you 🦾 )

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
 }
 ```
 
+## react-final-form Has Been Replaced By react-hook-form
+
+### `<FormWithRedirect>` Has Been Renamed to `<Form>`
+
+The form components don't handle redirection at all anymore as it happens in side effects, the `<FormWithRedirect>` has been renamed to `<Form>`.
+
+To upgrade simply replace all occurrences of `FormWithRedirect` with  `Form`.
```
fzaninotto, line 1948:
```suggestion
To upgrade, replace all occurrences of `<FormWithRedirect>` with  `<Form>`.
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+-    <FormWithRedirect
++    <Form
+        render={() => ...}
+    />
+);
+```
+
+### `<FormWithRedirect>` Props Have Changed
+
+`<FormWithRedirect>` used to accept [`react-final-form` `<Form>` props](https://final-form.org/docs/react-final-form/types/FormProps). It now accepts [`react-hook-form` `useForm` props](https://react-hook-form.com/api/useform). This also affects the other form components (`SimpleForm`, `TabbedForm`, etc.)
```
fzaninotto, line 1964:
```suggestion
`<FormWithRedirect>` used to accept [`react-final-form` `<Form>` props](https://final-form.org/docs/react-final-form/types/FormProps). It now accepts [`react-hook-form` `useForm` props](https://react-hook-form.com/api/useform). This also affects the other form components (`<SimpleForm>`, `<TabbedForm>`, etc.)
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+
+export const MyForm = () => (
+-    <FormWithRedirect
++    <Form
+        render={() => ...}
+    />
+);
+```
+
+### `<FormWithRedirect>` Props Have Changed
```
fzaninotto, line 1962:
```suggestion
### Form Props Have Changed
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+
+`<FormWithRedirect>` used to accept [`react-final-form` `<Form>` props](https://final-form.org/docs/react-final-form/types/FormProps). It now accepts [`react-hook-form` `useForm` props](https://react-hook-form.com/api/useform). This also affects the other form components (`SimpleForm`, `TabbedForm`, etc.)
+
+The most commonly used prop is probably `initialValues` which is now named `defaultValues`:
```
fzaninotto, line 1966:
```suggestion
The most commonly used prop is probably `initialValues`, which is now named `defaultValues`:
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+-            initialValues={{ title: 'A default title' }}
++            defaultValues={{ title: 'A default title' }}
+        >
+            ...
+        </SimpleForm>
+    </Create>
+)
+```
+
+However, we kept the `validate` function prop which we automatically translate as a custom [`react-hook-form` `resolver`](https://react-hook-form.com/api/useform#validationResolver).
```
fzaninotto, line 1981:
```suggestion
We kept the `validate` function prop, which we automatically translate to a custom [`react-hook-form` `resolver`](https://react-hook-form.com/api/useform#validationResolver). So even if it's not technically a react-hook-form prop, you can still use `validate` as before.
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+
+This also means you can now use [`yup`](https://github.com/jquense/yup), [`zod`](https://github.com/colinhacks/zod), [`joi`](https://github.com/sideway/joi), [superstruct](https://github.com/ianstormtaylor/superstruct), [vest](https://github.com/ealush/vest) or any [resolvers](https://react-hook-form.com/api/useform#validationResolver) supported by `react-hook-form` to apply schema validation.
```
fzaninotto, line 1983:
```suggestion
This also means you can now use [`yup`](https://github.com/jquense/yup), [`zod`](https://github.com/colinhacks/zod), [`joi`](https://github.com/sideway/joi), [superstruct](https://github.com/ianstormtaylor/superstruct), [vest](https://github.com/ealush/vest) or any [resolver](https://react-hook-form.com/api/useform#validationResolver) supported by `react-hook-form` to apply schema validation.
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+
+### `<FormWithRedirect>` Render Function Arguments Have Changed
```
fzaninotto, line 1985: I thought we got rid of the render prop?

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+
+The `<FormWithRedirect>` render function used to be called with an object containing parts of the `final-form` form state (`valid`, `invalid`, `pristine`, `dirty`). It now only receive the `handleSubmit` function which must be passed to the `onSubmit` prop of the underlying form. If you need access to the form state, call [`react-hook-form` `useFormState` hook](https://react-hook-form.com/api/useformstate):
```
fzaninotto, line 1987:
```suggestion
`<FormWithRedirect>` used to call its child function with an object containing parts of the `final-form` form state (`valid`, `invalid`, `pristine`, `dirty`). It now only passes the `handleSubmit` function, which must be passed down to the `onSubmit` prop of the underlying form. If you need to access the form state, call [the react-hook-form `useFormState` hook](https://react-hook-form.com/api/useformstate):
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+    return (
+-        <button disabled={disabled} {...props} />
++        <button disabled={!isDirty || !isValid} {...props} />
+    );
+}
+```
+
+### `useInput` Signature Changed
+
+`useInput` used to accept both a `initialValue` and a `defaultValue` prop. It now only accept `defaultValue`:
```
fzaninotto, line 2021:
```suggestion
`useInput` used to accept both a `initialValue` and a `defaultValue` prop. It now only accepts `defaultValue`:
```

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
++const SubmitButton = (props) => {
++    const { isDirty, isValid } = useFormState();
+    return (
+-        <button disabled={disabled} {...props} />
++        <button disabled={!isDirty || !isValid} {...props} />
+    );
+}
+```
+
+### `useInput` Signature Changed
```
fzaninotto, line 2019: You're listing changes to all Input components. Maybe the title of the section should be changed.

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
++        field,
+-        meta: { touched, error },
++        fieldState: { isTouched, invalid, error },
++        formState: { isSubmitted }
+        isRequired
+    } = useInput(props);
+
+    return (
+        <TextField
+            {...field}
```
fzaninotto, line 2052: there should be a diff here

UPGRADE.md
```diff
@@ -1939,6 +1939,131 @@ If you've declared custom Record types, you'll need to upgrade your code as foll
...
+            required={isRequired}
+            {...rest}
+        />
+    );
+};
+```
+
+## `allowEmpty` Has Been Removed From `SelectInput`, `AutocompleteInput` and `AutocompleteArrayInput`
+
+This is necessary for the underlying MaterialUI inputs as they require that the current input value is actually one of the available option. Those components now always accept an empty value (an empty string by default). If you require the input to have a non-empty value, use the `required` validation. You can safely remove this prop.
```
fzaninotto, line 2065: Not super clear, please elaborate and add a diff.

cypress/integration/auth.js
```diff
@@ -17,7 +17,7 @@ describe('Authentication', () => {
         ListPage.navigate();
         cy.url().should('contain', '/#/login');
     });
-    it('should not login with incorrect credentials', () => {
+    it.only('should not login with incorrect credentials', () => {
```
fzaninotto, line 20: .only spotted!

cypress/integration/create.js
```diff
@@ -60,7 +60,7 @@ describe('Create Page', () => {
                 value: 'foo',
             },
         ]);
-        cy.get(CreatePage.elements.submitButton).click();
+        //cy.get(CreatePage.elements.submitButton).click();
```
fzaninotto, line 63: commented out code

cypress/integration/create.js
```diff
@@ -300,11 +300,13 @@ describe('Create Page', () => {
         );
     });
 
-    it('should not show rich text input error message when field is untouched', () => {
+    // Skipped as we are going to replace the RichTextInput with the tip tap version
```
fzaninotto, line 303: please add a FIXME prefix

docs/ArrayInput.md
```diff
@@ -40,7 +40,7 @@ import { ArrayInput, SimpleFormIterator, DateInput, TextInput } from 'react-admi
 
 ## Usage
 
-`<ArrayInput>` expects a single child, which must be a *form iterator* component. A form iterator is a component accepting a `fields` object as passed by [react-final-form-array](https://github.com/final-form/react-final-form-arrays#fieldarrayrenderprops), and defining a layout for an array of fields. For instance, the `<SimpleFormIterator>` component displays an array of react-admin Inputs in an unordered list (`<ul>`), one sub-form by list item (`<li>`). It also provides controls for adding and removing a sub-record (a backlink in this example).
+`<ArrayInput>` expects a single child, which must be a *form iterator* component. A form iterator is a component accepting a `fields` object as passed by [react-hook-form](https://react-hook-form.com/api/usefieldarray), and defining a layout for an array of fields. It also receive several functions to manipulate the array values. For instance, the `<SimpleFormIterator>` component displays an array of react-admin Inputs in an unordered list (`<ul>`), one sub-form by list item (`<li>`). It also provides controls for adding and removing a sub-record (a backlink in this example).
```
fzaninotto, line 43:
```suggestion
`<ArrayInput>` expects a single child, which must be a *form iterator* component. A form iterator is a component accepting a `fields` object as passed by [react-hook-form](https://react-hook-form.com/api/usefieldarray), and defining a layout for an array of fields. It also receives several functions to manipulate the array values. For instance, the `<SimpleFormIterator>` component displays an array of react-admin Inputs in an unordered list (`<ul>`), one sub-form by list item (`<li>`). It also provides controls for adding and removing a sub-record (a backlink in this example).
```

docs/CreateEdit.md
```diff
@@ -1467,7 +1465,32 @@ export const UserCreate = () => (
...
+    .required();
+
+const CustomerCreate = () => (
+    <Create>
+        <SimpleForm resolver={yupResolver(schema)}>
+            <TextInput source="name" />
+            <NumberInput source="age" />
+        </SimpleForm>
+    </Create>
+    );
```
fzaninotto, line 1492:
```suggestion
);
```

docs/CreateEdit.md
```diff
@@ -1467,7 +1465,32 @@ export const UserCreate = () => (
...
+    .object()
+    .shape({
+        name: yup.string().required(),
+        age: yup.number().required(),
+    })
+    .required();
+
+const CustomerCreate = () => (
+    <Create>
+        <SimpleForm resolver={yupResolver(schema)}>
```
fzaninotto, line 1487: I think you must document this resolver prop somewhere

docs/Inputs.md
```diff
@@ -124,12 +124,12 @@ const dateParser = v => {
...
 Edition forms often contain linked inputs, e.g. country and city (the choices of the latter depending on the value of the former).
 
-React-admin relies on [react-final-form](https://final-form.org/docs/react-final-form/getting-started) for form handling. You can grab the current form values using react-final-form [useFormState](https://final-form.org/docs/react-final-form/api/useFormState) hook.
+React-admin relies on [react-hook-form](https://react-hook-form.com/) for form handling. You can grab the current form values using react-hook-form [useWatch](https://react-hook-form.com/api/usewatch) hook.
 
 ```jsx
 import { Edit, SimpleForm, SelectInput } from 'react-admin';
-import { useFormState } from 'react-final-form';
+import { useWatch } from 'react-hook-form';
```
fzaninotto, line 132: I think this is common enough to be documented in the upgrade guide

docs/Inputs.md
```diff
@@ -389,9 +389,9 @@ const LatLngInput = () => (
 
 **Tip**: Material-ui's `<TextField>` component already includes a label, so you don't need to use `<Labeled>` in this case.
 
-`useField()` returns two values: `input` and `meta`. To learn more about these props, please refer to the [useField](https://final-form.org/docs/react-final-form/api/useField) hook documentation in the react-final-form website.
+`useController()` returns three values: `field` and `fieldState` and `formState`. To learn more about these props, please refer to the [useController](https://react-hook-form.com/api/usecontroller) hook documentation.
```
fzaninotto, line 392:
```suggestion
`useController()` returns three values: `field`, `fieldState`, and `formState`. To learn more about these props, please refer to the [useController](https://react-hook-form.com/api/usecontroller) hook documentation.
```

examples/demo/src/reviews/ReviewEditToolbar.tsx
```diff
@@ -30,7 +30,7 @@ const StyledMuiToolbar = styled(MuiToolbar)(({ theme }) => ({
 }));
 
 const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
-    const { invalid, resource, saving } = props;
```
fzaninotto, line 33: seems a bc break worth documenting

examples/demo/src/visitors/SegmentsInput.tsx
```diff
@@ -1,18 +1,18 @@
 import * as React from 'react';
-import { useTranslate, SelectArrayInput, InputProps } from 'react-admin';
+import { useTranslate, SelectArrayInput, CommonInputProps } from 'react-admin';
```
fzaninotto, line 2: Why the rename of the type? It introduces a useless BC break

examples/simple/src/comments/PostQuickCreate.tsx
```diff
@@ -1,5 +1,12 @@
 import * as React from 'react';
 import { styled } from '@mui/material/styles';
+import {
+    Button,
```
fzaninotto, line 4: linter warnings in this file

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -141,13 +130,9 @@ const PostCreate = () => {
...
                 <TextInput source="teaser" fullWidth={true} multiline={true} />
                 <RichTextInput source="body" validate={required()} />
-                <FormSpy subscription={{ values: true }}>
-                    {({ values }) =>
-                        values.title ? (
-                            <NumberInput source="average_note" />
-                        ) : null
-                    }
-                </FormSpy>
+                <DependantField dependency="title">
```
fzaninotto, line 133: Seems like something we could add to core (I'd rename Field To Input)

examples/simple/src/posts/PostShow.tsx
```diff
@@ -21,28 +21,18 @@ import {
...
-                pathname: createPath({
-                    resource: 'comments',
-                    type: 'create',
-                }),
-                state: { record: { post_id: record?.id } },
-            }}
-        >
-            Add comment
-        </Button>
+        <CloneButton
```
fzaninotto, line 31: 👍 smart

examples/simple/src/posts/TagReferenceInput.tsx
```diff
@@ -57,14 +57,14 @@ const TagReferenceInput = ({
     source: string;
     label?: string;
 }) => {
-    const { change } = useForm();
+    const { setValue } = useFormContext();
```
fzaninotto, line 60: this should be documented in the upgrade guide

examples/simple/src/users/UserCreate.tsx
```diff
@@ -24,23 +24,18 @@ const UserEditToolbar = ({ permissions, ...props }) => {
...
                         onSuccess: data => {
-                            notify('ra.notification.created', 'info', {
-                                smart_count: 1,
+                            notify('ra.notification.created', {
+                                type: 'info',
+                                messageArgs: {
+                                    smart_count: 1,
+                                },
                             });
                             redirect(false);
```
fzaninotto, line 40: I think this redirect(false) isn't required anymore. But you need to add the form reset effect.

fzaninotto (changes requested):
Partial review so far (36/159)

packages/ra-core/src/form/Form.tsx
```diff
@@ -0,0 +1,214 @@
...
+export type ValidateForm = (
+    data: FieldValues
+) => FieldValues | Promise<FieldValues>;
+
+export type FormProps = FormOwnProps &
+    Omit<UseFormProps, 'onSubmit'> & {
+        validate?: ValidateForm;
+    };
+
+export type FormWithRedirectRenderProps = Omit<
```
WiXSL, line 131: Rename `FormWithRedirectRenderProps` to `FormRenderProps`

WiXSL (changes requested):

packages/ra-core/src/form/Form.tsx
```diff
@@ -0,0 +1,214 @@
...
+    FormViewProps,
+    'children' | 'isValid' | 'render' | 'onSubmit' | 'sanitizeEmptyValues'
+>;
+
+export type FormWithRedirectRender = (
```
WiXSL, line 136: Rename `FormWithRedirectRender` to `FormRender`

WiXSL (changes requested):

WiXSL (changes requested):
[UPGRADE.md line 1706](https://github.com/marmelab/react-admin/blob/ecd3ec5a6053366b46ec943c978d9f8d1b3461e0/UPGRADE.md?plain=1#L1706)

The upgrade includes `FormWithRedirect` instead of `Form`

UPGRADE.md
djhi: Not yet and I'm not sure we actually can

cypress/integration/auth.js
djhi: :scream:

docs/CreateEdit.md
djhi: We already mention we pass all props to the useForm hook. We didn't document final-form props before and I don't think we should for react-hook-form

examples/demo/src/visitors/SegmentsInput.tsx
djhi: It's not the same type. InputProps is from ra-core, this one contains UI and IMO ra-ui-material ui only props. It extends `InputProps`

examples/simple/src/posts/PostCreate.tsx
djhi: Maybe

examples/simple/src/posts/TagReferenceInput.tsx
djhi: Are we going to write an how to migrate from final-form to react-hook-form article ?

djhi:
> The upgrade includes FormWithRedirect instead of Form

Yes but this section comes before the upgrade instructions about react-hook-form. I'm not sure it should change

WiXSL (commented):
I had all source of errors running the demo.

Standing at http://localhost:3000/#/reviews :

errors

wattroll:
@WiXSL 
> I had all source of errors running the demo.

I've run into this as well. Got it when building my app, then tried on a bare-bones sandbox and didn't manage to get it running. It's a react-hook-form issue, a "strange one": [link 1](https://github.com/react-hook-form/react-hook-form/discussions/3894), [link 2](https://github.com/react-hook-form/react-hook-form/discussions/6989), [link 3](https://github.com/react-hook-form/react-hook-form/discussions/5150).

Apparently the issue appears when the Form context provider/consumer are split between the modules in a certain way. [`FormProvider` implementation](https://github.com/react-hook-form/react-hook-form/blob/master/src/useFormContext.tsx) seems a bit suspicious to me, but I haven't yet managed to figure out what's exactly causing it.

Things I tried so far:
- stepping  around with breakpoints at the consumer/provider calls, it does appear that react calls the context consumer (which is in the children of provider), before calling the provider
- replaced react-admin's references to `FormProvider` with a function call (to avoid creating intermediate `FormProvider` component, skipping straight to `HookFormContext.Provider` avoiding giving react an opportunity to defer its rendering)
- ensured that all of the dependencies are in-line (e.g. that there is only one bundled copy of React, react-dom, and react-hook-form) and so on
- tried building it with different bundlers (vite, parcel)
- tried compiling the app with exact same typescript options that react-admin is compiled, then with no typescript at all

Haven't tried yet:
- including `react-hook-form` files into the tree. So that they wouldn't be received as a bundle with all the definitions in the same `index.esm.mjs` scope, but as separate modules.

I have a hunch that the most reliable way to work this around would be to define our own context for keeping the form methods and pass it to `react-hook-form` functions whenever it needs it.

wattroll:
> it does appear that react calls the context consumer (which is in the children of provider)

That's not intuitive, but not unreasonable, right? Seems like a great way to optimize and something that can be expected as react has complete control over the execution order of hooks/components and to my knowledge doesn't provide a guarantee that parent's render/hooks would always be called before child's render.

Whatever is causing this, I think that `react-hook-form`'s `useController` and `useFormContext` are unnecessarily brittle and should be fixed.

```typescript
// https://github.com/react-hook-form/react-hook-form/blob/a44874a3dc9b7661c2f2ca85b5f6a968394239da/src/useFormContext.tsx#L11
const HookFormContext = React.createContext<UseFormReturn | null>(null);

const useFormContext = () => React.useContext(HookFormContext) as unknown as UseFormReturn<TFieldValues>;
```

So, it's valid for `React.useContext(HookFormContext)` to return `null`, but `useFormContext` hides it from its type and acts as there will always be a value. 

Then that nullable value gets destructured which fails badly with the error that doesn't make it obvious that `FormProvider` wasn't used. 

```typescript
// https://github.com/react-hook-form/react-hook-form/blob/a44874a3dc9b7661c2f2ca85b5f6a968394239da/src/useController.ts#L27

const methods = useFormContext<TFieldValues>();
const { name, control = methods.control, shouldUnregister } = props;
```

This could be worked around if we passed `{ control }` ourselves to `useController` though, and could keep it in our own context.

wattroll:
If there is nothing particularly wrong (I checked, but it must be checked again) with our usage of `react-hook-form` (e.g. we do wrap every call to `useController` inside `FormProvider`), then clearly we need to patch `react-hook-form` to at least fix the brittleness I described above. Or just handle the context ourselves and pass values directly to `react-hook-form`'s functions.

djhi:
Thanks for the reviews guys. Back to WIP!

wattroll:
> to further simplify our Form component and get rid of its render prop

I think that's very sensible. This would encourage custom forms to use react-admin's hooks instead of importing and calling the underlying form library, which in my opinion is better, and might simplify the upgrading to v3. E.g. the upgrade guide would advice to replace the calls to `react-final-form` with calls to `react-admin` not to `react-hook-form`.

djhi:
> I think that's very sensible. This would encourage custom forms to use react-admin's hooks instead of importing and calling the underlying form library, which in my opinion is better, and might simplify the upgrading to v3.

We understand why you think that but we really want to avoid adding yet another level of indirection. I'll do my best to solve this issue without hiding react-hook-form.

packages/ra-core/src/form/sanitizeEmptyValues.spec.ts
```diff
@@ -1,77 +0,0 @@
-import expect from 'expect';
-import sanitizeEmptyValues from './sanitizeEmptyValues';
-
-describe('sanitizeEmptyValues', () => {
```
fzaninotto, line 4: We should have unit tests on the Form component to make sure the behavior stays the same without this sanitization

packages/ra-core/src/form/setSubmissionErrors.ts
```diff
@@ -0,0 +1,40 @@
+import { FieldValues, UseFormSetError } from 'react-hook-form';
+
+export const setSubmissionErrors = (
```
fzaninotto, line 3: Could you please add some JSDoc to explain what it does and how to use it?

packages/ra-core/src/form/useGetValidationErrorMessage.ts
```diff
@@ -0,0 +1,22 @@
...
+    ValidationErrorMessage,
+    ValidationErrorMessageWithArgs,
+} from './validate';
+import { useTranslate } from '../i18n';
+
+/**
+ * This hook returns a function that can translate an error message. It handles simple string errors and those which have a message and args.
+ * @see ValidationErrorMessage
+ * @see ValidationErrorMessageWithArgs
+ */
```
fzaninotto, line 11: could you please add an example ?

packages/ra-core/src/form/useInput.spec.tsx
```diff
@@ -18,243 +21,209 @@ describe('useInput', () => {
...
-
-        const { getByLabelText } = render(
-            <Form
-                onSubmit={jest.fn()}
-                render={({ form }) => {
-                    formApi = form;
 
-                    return (
+        render(
+            <CoreAdminContext dataProvider={testDataProvider()}>
```
fzaninotto, line 81: Note that you don't need to pass a dataProvider if you need one, CoreAdminContext has default one since a few days

packages/ra-core/src/form/useInput.ts
```diff
@@ -71,73 +84,68 @@ const useInput = ({
...
-    const handleFocus = useCallback(
-        event => {
-            onFocus(event);
-            if (typeof customOnFocus === 'function') {
-                customOnFocus(event);
-            }
-        },
-        [onFocus, customOnFocus]
-    );
+    useApplyInputDefaultValues(props);
```
fzaninotto, line 110: There is something I don't understant (and that would merit a comment): you're already setting the defaultValue L93, why do you need to apply it again?

packages/ra-core/src/form/useInput.ts
```diff
@@ -71,73 +84,68 @@ const useInput = ({
...
 
-    // If there is an input prop, this input has already been enhanced by final-form
+    // If there is an field prop, this input has already been enhanced by react-hook-form
```
fzaninotto, line 112:
```suggestion
    // If there is a field prop, this input has already been enhanced by react-hook-form
```

packages/ra-core/src/form/useSuggestions.ts
```diff
@@ -196,10 +194,22 @@ export const getSuggestionsFactory = ({
...
         } else {
-            suggestions = choices;
+            suggestions = [...choices];
         }
     } else {
-        suggestions = choices.filter(choice => matchSuggestion(filter, choice));
+        suggestions = choices.filter(
+            choice =>
+                matchSuggestion(filter, choice) ||
+                (selectedItem != null &&
```
fzaninotto, line 203: not sure why you added that, it doesn't seem related to react-final-form

packages/ra-data-simple-rest/package.json
```diff
@@ -19,8 +19,8 @@
     "bugs": "https://github.com/marmelab/react-admin/issues",
     "license": "MIT",
     "scripts": {
-        "build": "tsup src/index.ts --silent --clean --format cjs,esm --minify --keep-names --metafile --sourcemap --dts --legacy-output",
-        "watch": "tsup src/index.ts --silent --clean --format cjs,esm --minify --keep-names --metafile --sourcemap --dts --legacy-output --watch"
+        "build": "tsup --external react-hook-form src/index.ts --silent --clean --format cjs,esm --minify --keep-names --metafile --sourcemap --dts --legacy-output",
```
fzaninotto, line 22: do we need to add it even in packages that don't have a dependency on react-hook-form?

fzaninotto (changes requested):
Partial review (84/169)

packages/ra-core/src/form/Form.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    const handleSubmit = (event: BaseSyntheticEvent) => {
+        // Prevent outer forms to receive the event
+        event.stopPropagation();
+        formHandleSubmit(event);
+        return;
+    };
+
+    return (
+        <FormGroupsProvider>
```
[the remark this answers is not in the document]
djhi: Why?

packages/ra-core/src/form/useSuggestions.ts
djhi: It was necessary for the choices inputs

packages/ra-data-simple-rest/package.json
djhi: I'd rather be safe than sorry

djhi: Removed from packages where it was unnecessary

packages/ra-core/src/form/useInput.ts
djhi: Added a lengthy comment

packages/ra-core/src/form/useGetValidationErrorMessage.ts
djhi: Added

packages/ra-core/src/form/Form.tsx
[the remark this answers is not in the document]
djhi: It's quite length and the features are well contained this way.

packages/ra-core/src/form/Form.tsx
```diff
@@ -0,0 +1,214 @@
...
+    warnWhenUnsavedChanges?: boolean;
+}
+
+interface FormViewProps
+    extends Omit<FormOwnProps, 'onSubmit' | 'sanitizeEmptyValues'> {
+    handleSubmit: (e?: BaseSyntheticEvent) => void;
+    warnWhenUnsavedChanges?: boolean;
+}
+
+const FormView = (props: FormViewProps) => {
```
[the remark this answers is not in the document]
djhi: Done

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -64,28 +70,45 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
+        async event => {
+            if (onClick) {
+                onClick(event);
+            }
+            if (event.defaultPrevented) {
+                return;
+            }
+            if (hasSideEffects) {
+                event.preventDefault();
+                const isFormValid = await form.trigger();
```
fzaninotto, line 89: I think this logic should be in the form rather than in the SaveButton. Otherwise, will it be triggered when users submit the form using the Enter key? Or when using another Save button?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -49,45 +49,24 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
         variant,
         margin,
         TransitionProps,
-        defaultValue,
```
fzaninotto, line 52: linter warnings in this file

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -129,16 +105,16 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
                     className
                 )}
             >
-                {submitFailed && typeof error !== 'object' && error && (
+                {/* {submitFailed && typeof error !== 'object' && error && (
```
fzaninotto, line 108: commented out code

packages/ra-ui-materialui/src/input/BooleanInput.spec.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import { render, fireEvent } from '@testing-library/react';
-import { Form } from 'react-final-form';
-import { renderWithRedux } from 'ra-test';
-import { FormWithRedirect } from 'ra-core';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { testDataProvider } from 'ra-core';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
```
fzaninotto, line 4: linter warnings in this file

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -11,18 +11,14 @@ import {
...
-import {
-    FieldTitle,
-    useInput,
-    InputProps,
-    ChoicesProps,
-    useChoices,
-} from 'ra-core';
+import { FieldTitle, useInput, ChoicesProps, useChoices } from 'ra-core';
 import { InputHelperText } from './InputHelperText';
 import { SelectProps } from '@mui/material/Select';
```
fzaninotto, line 16: linter warning

packages/ra-ui-materialui/src/input/TranslatableInputs.spec.tsx
```diff
@@ -1,13 +1,15 @@
 import * as React from 'react';
 import expect from 'expect';
-import { fireEvent, screen, waitFor } from '@testing-library/react';
-import { TranslatableInputs } from './TranslatableInputs';
 import { TextInput } from './TextInput';
-import { FormWithRedirect, required, useTranslatableContext } from 'ra-core';
-import { renderWithRedux } from 'ra-test';
-import { TranslatableInputsTab } from './TranslatableInputsTab';
+import { required, testDataProvider, useTranslatableContext } from 'ra-core';
```
fzaninotto: linter warning

packages/ra-ui-materialui/src/input/useSupportCreateSuggestion.tsx
```diff
@@ -90,18 +90,13 @@ export const useSupportCreateSuggestion = (
                     : translate(createLabel, { _: createLabel })
             );
         },
-        handleChange: async eventOrValue => {
+        handleChange: async (eventOrValue: MouseEvent | any) => {
             const value = eventOrValue?.target?.value || eventOrValue;
-
             const finalValue = Array.isArray(value) ? [...value].pop() : value;
-            if (eventOrValue?.preventDefault) {
```
fzaninotto: Why did you remove this one?

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -346,6 +346,9 @@ export interface DatagridProps<RecordType extends RaRecord = any>
 }
 
 const injectedProps = [
+    'field',
```
fzaninotto: When are these fields injected? It doesn't make sense

fzaninotto (changes requested):
I have finished reviewing all the files. Will review again once you push the fixes.

fzaninotto:
I see regressions on filter forms:

- Simple Example / Post list / Add title filter: the list gets filters even though nothing was added
- Simple Example / User list / add filter role, enter "admin", remove filter, add filter name, enter "Logan", the role filter comes back

packages/ra-ui-materialui/src/button/SaveButton.tsx
djhi: Nope, it's also in the Form but has to be there too for SaveButton that are not submit

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
djhi: Because the Datagrid can be used inside a reference input

fzaninotto:
Regression on filter forms in the demo:

- Orders list, add customer filter, select first item => no filter
- Reviews list, add status filter, select first item => no filter

packages/ra-ui-materialui/src/input/useSupportCreateSuggestion.tsx
djhi: Because it was unnecessary

WiXSL (commented):
There are some references of `react-final-form` that should be changed to `react-hook-form` in the Readme.md files

WiXSL (changes requested):
In the demo:
Clearing lists filter via `SearchInput` clear icon doesn't change the filter

WiXSL (changes requested):
Demo => products

`SearchInput` doesn't filter the List at all.

examples/demo/src/layout/Login.tsx
```diff
@@ -165,23 +78,21 @@ const Login = () => {
...
+                                        validate={required()}
                                     />
                                 </div>
                                 <div className={classes.input}>
-                                    <Field
-                                        name="password"
-                                        // @ts-ignore
-                                        component={renderInput}
+                                    <TextInput
+                                        source="password"
```
WiXSL, line 91:
```suggestion
                                        source="password"
                                        fullWidth
```

examples/demo/src/layout/Login.tsx
```diff
@@ -165,23 +78,21 @@ const Login = () => {
                             </div>
                             <div className={classes.form}>
                                 <div className={classes.input}>
-                                    <Field
+                                    <TextInput
                                         autoFocus
-                                        name="username"
-                                        // @ts-ignore
-                                        component={renderInput}
+                                        source="username"
```
WiXSL, line 83:
```suggestion
                                        source="username"
                                        fullWidth
```

WiXSL (changes requested):

WiXSL (changes requested):
Demo => OrderEdit => CustomerDetails must use get the record from useRecordContext<Customer> in order to work

Demo => OrderEdit => CustomerAdress must use get the record from useRecordContext<Customer> in order to work

WiXSL (changes requested):
If you look for  `= ({ record }:` in the demo you'll find more components that need to use the record from the context.

Note: I'm just the messenger 😢

WiXSL (changes requested):
Demo => categories => Edit => Reference column link is not working

WiXSL (changes requested):
CRM also has some components that needs to use the record from the context

examples/simple/src/data.tsx
```diff
@@ -213,7 +213,9 @@ export default {
             tags: [4, 3],
             category: 'tech',
             subcategory: 'computers',
-            pictures: null,
+            // FIXME: Final form does not support a null value for arrays
```
WiXSL, line 216: Don't you mean react-hooks-form instead of final form?

WiXSL (changes requested):

fzaninotto:
@WiXSL the problems with the record context aren't linked to this PR, so we won't deal with them in ths PR. Would you mind opening an issue with them so we don't forget them?

fzaninotto (approved):
This is AWESOME!!! 💯 👍 🎉 🦘 

Thanks @djhi for this tremendous work.

WiXSL:
Great! We have react-hooks-form working. Thank you @djhi !!!

examples/simple/src/data.tsx
djhi, line 216:
```suggestion
            // FIXME: react-hook-form does not support a null value for arrays
```

fzaninotto:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

EPIC PR

anthonycmain:
Out of interest why is this a breaking change prior to v4?

djhi:
It's **for** v4 (that's the `next` branch)

anthonycmain:
> It's **for** v4 (that's the `next` branch)

Oh the Tweet I read said "This was the last big breaking change before V4" so I assume that mean v3.X

afilp:
Hello, does this affect the creation of custom inputs, because I get an error when I replaced
`import { Field } from 'react-final-form';`
with
import { Field } from 'react-hook-form';

This is the error I am getting:

```
ERROR in ./src/helpers/OurDateInput.js 146:30-35

export 'Field' (imported as 'Field') was not found in 'react-hook-form' (possible exports: Controller, FormProvider, appendErrors, get, set, useController, useFieldArray, useForm, useFormContext, useFormState, useWatch)
```

I did not find something different in UPGRADE doc or other docs on how to create a custom input field. Thanks!

fzaninotto:
@afilp indeed, it changes the syntax of custom inputs. We've updated the Inputs.md section about "Writing Your Own Input Component", but we've forgotten to add a diff in the README. We'll fix that soon.

afilp:
@fzaninotto Thank you!

erikras:
Thank you to this project for providing an awful lot of downloads and users to React Final Form. As the [lapsed] maintainer of said library, I'm both sad and happy to see you go. ❤️ 👋

djhi:
It wasn't an easy decision. Thank you for redux-form, final-form and centered! :heart:

PR #7096, opened 2022-01-17
Fix no title separation

DDobrowolski:
### Description
Fixed no title separation issue. When we're trying to render ShowView inside ListView, titles are being joined without any separation when ShowView title is default. Added space character between.

### How to reproduce
- Replace `Datagrid` props to `<Datagrid rowClick="expand" expand={<PostShow />} optimized>` in `examples\simple\src\posts\PostList.tsx`
- Remove title prop from `ShowView` in `examples\simple\src\posts\PostShow.tsx`

### Before
image

### After
image

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Title.tsx
+++ b/packages/ra-ui-materialui/src/layout/Title.tsx
@@ -28,15 +28,13 @@ const Title: FC<TitleProps> = ({
 
     warning(!defaultTitle && !title, 'Missing title prop in <Title> element');
 
-    const titleSeparator = container.childNodes.length ? ' ' : '';
-
     const titleElement = !title ? (
         <span className={className} {...rest}>
-            {titleSeparator + defaultTitle}
+            {defaultTitle + ' '}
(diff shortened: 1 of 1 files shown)


fzaninotto:
Hi, and thanks for your feedback. The composite title that results from your PR ("Posts Post #10 Post #12") isn't informative at all, and keeps getting longer and longer as users open new rows. So this is not a desirable feature. 

Instead, please [follow the documentation](https://marmelab.com/react-admin/List.html#expand) and disable the title on the embedded Show view:

```jsx
const PostShow = props => (
    <Show
        {...props}
        /* disable the app title change when shown */
        title=" "
    >
        <SimpleForm>
            <RichTextInput source="body" />
        </SimpleForm>
    </Show>
);

const PostList = props => (
    <List {...props}>
        <Datagrid expand={<PostEdit />}>
            <TextField source="id" />
            <TextField source="title" />
            <EditButton />
        </Datagrid>
    </List>
)
```

PR #7098, opened 2022-01-17
[4.0.0-alpha.2] Fix `AutocompleteInput` invalid empty option

### LINKED ISSUE #7061: [4.0.0-alpha.2]: MUI warning when using AutocompleteInput
**What you were expecting:**
Use AutocompleteInput without warnings.

**What happened instead:**

Warning in the console
```
MUI: The value provided to Autocomplete is invalid. None of the options match with `""`. You can use the `isOptionEqualToValue` prop to customize the equality test.
```

Note the `""`, probably that's the offending part.

**Steps to reproduce:**
Seems to appear in pretty much any `AutocompleteInput`. Can be seen at posters galore demo, e.g. by going to Orders resource and adding a Customer filter.

---

I bisected it down to [this commit](https://github.com/marmelab/react-admin/pull/6971/commits/7de027dce7c3dff1fa11ca1210dfc7f7083b8858):

```
7de027dce7c3dff1fa11ca1210dfc7f7083b8858 is the first bad commit
commit 7de027dce7c3dff1fa11ca1210dfc7f7083b8858
Author: Travis CI <travis@travis-ci.org>
Date:   Wed Dec 15 14:44:42 2021 +0100

    Merge both autocomplete
```

I haven't managed to spot the cause of a problem by a quick look-through. @djhi, would you have time to have a look at this? :pray:

WiXSL:
Fixes #7061

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -290,7 +290,8 @@ If you provided a React element for the optionText prop, you must also provide t
     }, [getOptionLabel, multiple, selectedChoice]);
 
     const isOptionEqualToValue = (option, value) => {
-        return getChoiceValue(option) === getChoiceValue(value);
+        // eslint-disable-next-line eqeqeq
+        return getChoiceValue(option) == getChoiceValue(value);
     };
 
     const handleInputChange = (event: any, newInputValue: string) => {


wattroll:
Thanks! I have tested it in my app, and unfortunately I am still getting the same ``None of the options match with `""``` warning.

Then I tried it in the demo-app, and in there the warning is indeed fixed. However here is something I have noticed:

2022-01-17-204810_1920x1080_scrot


There is an empty-stringed option in there, which doesn't look right, but satisfies the MUI's attempt to match with `""` :sob:

WiXSL:
That is a different issue, I will look into it tomorrow

djhi (approved):
:+1:

colinscruggs:
> Thanks! I have tested it in my app, and unfortunately I am still getting the same ``None of the options match with `""``` warning.
> 
> Then I tried it in the demo-app, and in there the warning is indeed fixed. However here is something I have noticed:
> 
> 2022-01-17-204810_1920x1080_scrot
> 
> There is an empty-stringed option in there, which doesn't look right, but satisfies the MUI's attempt to match with `""` 😭

I'm having the same exact issue as you when prepending the options list with an empty string option, both regarding the console warnings and the blank option itself being rendered at a small height. 

Referencing [this issue](https://github.com/mui/material-ui/issues/29727#issuecomment-1091963830), it looks like the devs are considering removing the warning all together, which would be great in my option. I've tried my hardest to leverage the `isOptionEqualToValue` function, but to no avail. 

To fix the issue with the blank option ("") not rendering correctly, I used the renderOption prop like so: (note, in my implementation, the options array is primitive strings, not objects)
```
renderOption={(props, option) => {
    // for optional fields, render a blank option
    return (
        <MenuItem {...props} value={option}>
            {option === '' ? <>&nbsp;</> : option}
        </MenuItem>
    );
}}
```

PR #7100, opened 2022-01-17
[BC Break] Add support mounting admin inside a sub path

fzaninotto:
## Problems

React-router v6 doesn't support nesting routers. So it will break for users who use react-router to mount the admin in a sub path (e.g. "admin").

The way react-admin v3 used to support sub path relied on passing the `basePath` everywhere, which is not very good DX.

We need to reintroduce the ability to mount react-admin in a sub path, without using `basePath`.

## Use case

```jsx
const App = () => (
    <BrowserRouter>
        <Routes>
            <Route
                path="/admin/*"
                element={
                    <Admin dataProvider={dataProvider} basename="/admin">
                        <Resource name="comments" {...comments} />
                    </Admin>
                }
            />
        </Routes>
    </BrowserRouter>,
);
```

## Solution

- [x] Do not add a Router if the app mounts inside an existing router
- [x] Add a BasenameContext to support passing a custom basename
- [x] Update link generation to take into account the basename context
- [x] Fix useRedirect
- [x] Fix Buttons (Show, Edit, Create, List, Clone)
- [x] Fix Tab routing (TabbedForm, TabbedShowLayout)
- [x] Fix Menu routing (MenuItemLink, Menu, DashboardMenu
- [x] Fix List iterator routing (Datagrid, SingleFieldList, SimpleList)
- [x] Fix Reference routing (ReferenceField, ReferenceOneField)
- [x] Fix custom links in simple demo (not in other demos, which aren't supposed to be mounted inside another router)
- [x] Add tests & stories
- [x] Fix tests
- [x] Add routing doc chapter
- [x] [BC Break] Remove `basePath` prop everywhere
- [x] Write upgrade guide


packages/ra-core/src/core/Resource.tsx
```diff
@@ -10,10 +10,32 @@ export const Resource = (props: ResourceProps) => {
...
             <Routes>
-                {Create && <Route path="create/*" element={<Create />} />}
-                {Show && <Route path=":id/show/*" element={<Show />} />}
-                {Edit && <Route path=":id/*" element={<Edit />} />}
-                {List && <Route path="/*" element={<List />} />}
+                {Create && (
+                    <Route
+                        path="create/*"
+                        element={
+                            React.isValidElement(Create) ? Create : <Create />
```
fzaninotto, line 17: Not related, but this allows to use elements in Resource:

```jsx
<Resource name="posts" list={<PostList />} />
```

In addition to using components:

```jsx
```

packages/ra-core/src/core/index.ts
```diff
@@ -3,7 +3,6 @@ export * from './components';
 export * from './ResourceContext';
 export * from './ResourceContextProvider';
 export * from './ResourceDefinitionContext';
-export * from './useScrollToTop';
```
fzaninotto, line 6: moved to the 'routing' folder

packages/ra-core/src/routing/useBasename.ts
```diff
@@ -0,0 +1,5 @@
+import { useContext } from 'react';
+
+import { BasenameContext } from './BasenameContext';
+
+export const useBasename = () => useContext(BasenameContext);
```
fzaninotto, line 5: add jsDoc

packages/ra-core/src/sideEffect/index.ts
```diff
@@ -1,7 +1,4 @@
-import useRedirect, { RedirectionSideEffect } from './useRedirect';
```
fzaninotto, line 1: moved to the routing directory

packages/ra-core/src/util/index.ts
```diff
@@ -3,11 +3,9 @@ import FieldTitle, { FieldTitleProps } from './FieldTitle';
 import getFetchedAt from './getFetchedAt';
 import getFieldLabelTranslationArgs from './getFieldLabelTranslationArgs';
 import ComponentPropType from './ComponentPropType';
-import linkToRecord from './linkToRecord';
```
fzaninotto, line 6: moved to the routing directory

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -34,24 +38,19 @@ const CreateButton = (props: CreateButtonProps) => {
...
     );
-    const location = useMemo(
-        () => ({
-            pathname: `/${resource}/create`,
-            state: { _scrollToTop: scrollToTop },
-        }),
-        [resource, scrollToTop]
-    );
+
     return isSmall ? (
```
fzaninotto, line 47: todo: useMemo

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -119,40 +117,40 @@ export interface ReferenceFieldProps<RecordType extends RaRecord = any>
     resource?: string;
     source: string;
     translateChoice?: Function | boolean;
-    /**
-     * @deprecated use link instead
-     */
-    linkType?: LinkToType;
```
fzaninotto, line 125: To be added to the upgrade guide

packages/ra-core/src/routing/useCreateInternalLink.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Identifier } from '../types';
+import { useBasename } from './useBasename';
+
+export const useCreateInternalLink = () => {
+    const basename = useBasename();
+    return useCallback(
+        ({ resource, id, type }: CreateInternalLinkParams): string => {
+            switch (type) {
+                case 'list':
+                    return `${basename}/${resource}`;
```
fzaninotto, line 12: in some old apps, users will use basepath instead of resource. The result will be a path like '//posts'. Easy fix: remove double slashes here to facilitate backwards compatibility.

packages/react-admin/src/AdminRouter.tsx
```diff
@@ -1,11 +0,0 @@
-import * as React from 'react';
```
fzaninotto, line 1: useless component: LoadingPage is already set as default prop in AdminUI

wattroll:
Looks great. I was setting up my own `AdminContext` to use `BrowserRouter` (passing `history: createBrowserHistory()` wasn't working for some reason). So apart from subpath mounting, this will bring a smooth way of changing the default router. Thanks.

---

Is `history` and `HistoryRouter` still needed per-se? From what I understood the main need for it came from using `react-connected-router` which isn't used anymore. Of course it's a question of BC breakage (as `history` is a prop of `Admin`), but it won't be big breakage if the fix is to just wrap `Admin` with `HistoryRouter { history }` or straight up with the preferred `Router` implementation.

---

Could the docs provide advice to switch away from `HashRouter`? I totally understand it being the default to facilitate getting started quickly, but e.g.  [`react-router` docs](https://reactrouter.com/docs/en/v6/api#hashrouter) say "We strongly recommend you do not use HashRouter unless you absolutely have to.".

packages/ra-core/src/routing/useCreateInternalLink.ts
```diff
@@ -0,0 +1,33 @@
+import { useCallback } from 'react';
+
+import { Identifier } from '../types';
+import { useBasename } from './useBasename';
+
+export const useCreateInternalLink = () => {
```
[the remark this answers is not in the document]
djhi: I vote `useCreateInternalTo`

examples/crm/package.json
```diff
@@ -18,8 +18,8 @@
         "react-beautiful-dnd": "^13.0.0",
         "react-dom": "^17.0.0",
         "react-error-boundary": "^3.1.4",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2",
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0",
```
djhi, line 22: v6.2.1 is out: https://github.com/remix-run/react-router/releases

examples/demo/package.json
```diff
@@ -34,8 +34,8 @@
         "react-dom": "^17.0.0",
         "react-final-form": "^6.5.2",
         "react-redux": "^7.1.0",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2",
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0",
```
djhi, line 38: Same: https://github.com/remix-run/react-router/releases

examples/simple/package.json
```diff
@@ -28,8 +28,8 @@
         "react-dom": "^17.0.0",
         "react-final-form": "^6.5.2",
         "react-query": "^3.32.1",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2"
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0"
```
djhi, line 32: Same: https://github.com/remix-run/react-router/releases

packages/ra-core/package.json
```diff
@@ -38,8 +38,8 @@
         "react-dom": "^17.0.0",
         "react-final-form": "^6.5.7",
         "react-redux": "^7.1.0",
-        "react-router": "^6.0.2",
```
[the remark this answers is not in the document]
djhi: As previously mentionned, there's v6.2.1 now: https://github.com/remix-run/react-router/releases

packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -36,6 +36,9 @@ export interface AdminContextProps {
     dashboard?: DashboardComponent;
     dataProvider?: DataProvider | LegacyDataProvider;
     queryClient?: QueryClient;
+    /**
+     * @deprecated Wrap your Admin inside a Router to change the routing strategy
+     */
```
djhi, line 41: Why ? Shouldn't we add this warning in v3 and remove this prop completely here?

packages/ra-core/src/routing/linkToRecord.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+export const linkToRecord = (resource, id, linkType = 'edit') => {
+    const link = `${resource}/${encodeURIComponent(id)}`;
+
+    if (linkType === 'show') {
+        return `${link}/show`;
+    }
+
+    return link;
+};
```
djhi, line 12: Should be deprecated in next v3 and removed here

packages/ra-core/src/routing/resolveRedirectTo.ts
```diff
@@ -0,0 +1,42 @@
+import { To } from 'react-router-dom';
+
+/**
+ * @deprecated use useCreatePath instead
```
djhi, line 4: Same

packages/ra-no-code/package.json
```diff
@@ -30,8 +30,8 @@
         "ra-test": "^4.0.0-alpha.1",
         "react": "^17.0.0",
         "react-dom": "^17.0.0",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2",
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0",
```
djhi, line 34: Same, more recent version available

packages/ra-test/package.json
```diff
@@ -37,8 +37,8 @@
         "react": "^17.0.0",
         "react-dom": "^17.0.0",
         "react-redux": "^7.1.0",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2",
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0",
```
djhi, line 41: There too

packages/ra-ui-materialui/package.json
```diff
@@ -43,8 +43,8 @@
         "react-final-form": "^6.5.7",
         "react-final-form-arrays": "^3.1.3",
         "react-redux": "^7.1.0",
-        "react-router": "^6.0.2",
-        "react-router-dom": "^6.0.2",
+        "react-router": "^6.1.0",
+        "react-router-dom": "^6.1.0",
```
djhi, line 47: Again

djhi (changes requested):
Neat! This is so much better :muscle:

UPGRADE.md
```diff
@@ -1352,6 +1355,106 @@ export const PostEdit = (props) => (
...
++       <ShowButton />
+        {/* Add your custom actions */}
+        <Button color="primary" onClick={customAction}>Custom Action</Button>
+    </TopToolbar>
+);
+```
+
+Keeping the `basePath` prop may result in unrecognized DOM props warnings, but your app will still work flawlessly even if you don't remove them. If you're using TypeScript, your code will not compile unless you remove all the `basePath` props.
+
+When a function (not a component) received the `basePath` as argument, it now receives the `resource` instead. For instance, the `<Datagrid rowClick>`prop used to accept a function:
```
WiXSL, line 1429:
```suggestion
When a function (not a component) received the `basePath` as argument, it now receives the `resource` instead. For instance, the `<Datagrid rowClick>` prop used to accept a function
```

WiXSL (changes requested):

WiXSL (commented):
Impressive work!
I couldn't been able to find problems 💪

WiXSL (approved):

packages/ra-core/src/core/CoreAdminContext.tsx
fzaninotto: This is deprecated because the only reason you may want to pass a custom history was to use another router. Now you can use anoyther router.

As for preparing the v4 migration in v3, it's a general matter that doesn't concern this PR. Besides, we couldn't deprecate it in v3 without supporting the new way, which isn't possible in v3 which uses the old react router.

packages/ra-core/src/routing/resolveRedirectTo.ts
fzaninotto: Same answer: we don't remove it to avoid a more painful upgrade.

fzaninotto:
@djhi Let's upgrade the packages to their latest version in another PR (and rather at the end of the 4.0 roadmap), to avoid multiple rebases due to yarn.lock conflicts.

fzaninotto:
@wattroll 

> Is history and HistoryRouter still needed per-se? From what I understood the main need for it came from using react-connected-router which isn't used anymore. Of course it's a question of BC breakage (as history is a prop of Admin), but it won't be big breakage if the fix is to just wrap Admin with HistoryRouter { history } or straight up with the preferred Router implementation.

No, `<Admin history>` is no longer necessary, but we don't need to remove it. I've marked it as deprecated.

Should the docs provide advice to switch away from HashRouter? I totally understand it being the default to facilitate getting started quickly, but e.g. react-router docs say "We strongly recommend you do not use HashRouter unless you absolutely have to.".

> I added a section about it, see Routing.md

djhi (approved):

PR #7102, opened 2022-01-18
Add previous data on transform

### LINKED ISSUE #7054: [Feature Request] Making previousData accessible from save and transform method.

It may be better to access previous data state(to be exact, before saved state) from save and transform method than it is only accessible from dataProvider.
Now, as I know with reading code useEditController and react-admin document, DataProvider is only way (except for customizing default component) to access previousData, but as more resource depended logics are implemented though intrinsically they are loosely coupled, it is fat soon. 

If it can be accessed from transform and save methods,  it's convenient for limiting scope narrower and  modifying request parameter logic form layer can be completed, in the result, we will no longer implement conditional logic using previousData in dataProvider , then more agnostic feature against resource type.

Add more the previousData argument and assign indirect  record 
[this line](https://github.com/marmelab/react-admin/blob/43c4fafc8bcedebc386c7d3dc3b63cfd56420a17/packages/ra-core/src/controller/details/useEditController.ts#L176)
and pass record value second parameter [https://github.com/marmelab/react-admin/blob/43c4fafc8bcedebc386c7d3dc3b63cfd56420a17/packages/ra-core/src/controller/details/useEditController.ts#L178](this code).

I read [https://marmelab.com/react-admin/CreateEdit.html#usecreatecontroller-and-useeditcontroller](this section of document).
Edit props completely rewritable and provide alternative controller props to store state of previous  data. So, if this features implements not desired for framework, we can make this feature.

**Additional context**

tkow:
Closes #7054 

Add previous data argument transform method. My tests added is redundant because previousData always is passed and  asserted in every tests but, it's noting worth to keep description about expecting behavior, ready for this behavior is modified (however if you don't need it, I'll remove them).

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -542,6 +542,22 @@ The `transform` function can also return a `Promise`, which allows you to do all
 
 **Tip**: The message will be translated.
 
+**Tip**: `<Edit>`'s transform prop function also get the `previousData` in its second argument:
+
+```jsx
+export const UserEdit = (props) => {
+    const transform = (data, { previousData }) => ({
+        ...data,
+        avoidChangeField: previousData.avoidChangeField
+    });
+    return (
+        <Edit {...props} transform={transform}>
+            ...
+        </Edit>
+    );
+}
+```
+
 ## Prefilling a `<Create>` Record
 
 Users may need to create a copy of an existing record. For that use case, use the `<CloneButton>` component. It reads the `record` from the current `RecordContext`.
@@ -2283,6 +2299,32 @@ const dataProvider = {
         }));
     },
 }
+```
+
+**Tip**: `<Edit>`'s transform prop function also get the `previousData` in its second argument:
+
+```jsx
+const PostEditToolbar = props => (
+    <Toolbar {...props}>
+        <SaveButton submitOnEnter={true} />
+        <SaveButton
+            label="post.action.save_and_notify"
+            transform={(data, { previousData }) => ({
+                ...data,
+                avoidChangeField: previousData.avoidChangeField
+            })}
+            submitOnEnter={false}
+        />
+    </Toolbar>
+);
+
+const PostEdit = () => (
+    <Edit>
+        <SimpleForm toolbar={<PostEditToolbar />}>
+            // ...
+        </SimpleForm>
+    </Edit>
+);
 ```
 
 ### Grouping Inputs
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -179,6 +179,7 @@ export const useEditController = <RecordType extends RaRecord = any>(
             resource,
             transform,
             update,
+            recordCached.previousData,
         ]
     );
 
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -99,7 +99,7 @@ export const useEditController = <RecordType extends RaRecord = any>(
         record,
     });
 
-    const [update, { isLoading: saving }] = useUpdate<RecordType>(
+    const [update, { isLoading: saving }, previousData] = useUpdate<RecordType>(
         resource,
         { id, previousData: record },
         { ...otherMutationOptions, mutationMode }
@@ -116,9 +116,9 @@ export const useEditController = <RecordType extends RaRecord = any>(
         ) =>
             Promise.resolve(
                 transformFromSave
-                    ? transformFromSave(data)
+                    ? transformFromSave(data, { previousData })
                     : transform
-                    ? transform(data)
+                    ? transform(data, { previousData })
                     : data
             ).then((data: Partial<RecordType>) =>
                 update(
@@ -173,6 +173,7 @@ export const useEditController = <RecordType extends RaRecord = any>(
             resource,
             transform,
             update,
+            previousData,
         ]
     );
 
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -99,9 +99,11 @@ export const useEditController = <RecordType extends RaRecord = any>(
         record,
     });
 
-    const [update, { isLoading: saving }, previousData] = useUpdate<RecordType>(
+    const recordCached = { id, previousData: record };
+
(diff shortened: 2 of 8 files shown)


packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -382,7 +382,7 @@ export const useUpdate = <RecordType extends RaRecord = any>(
         }
     };
 
-    return [update, mutation];
```
djhi, line 385: You don't need to pass refs in dependencies

djhi (changes requested):

packages/ra-core/src/dataProvider/useUpdate.ts
tkow: @djhi 
Sorry, I marked draft and I intended to revise but, I couldn't figure out how useUpdate cache previousData.
Now, I understand lifecycle in useQuery makes record changed and form's record nothing relevant, so at least one changes by getOne, so we can use record variable.
But, I thought this is very confused so, I make the variable with clearer name as cachedData. I removed draft sign. So could you review again?

docs/CreateEdit.md
```diff
@@ -5,7 +5,7 @@ title: "The Create and Edit Views"
 
 # The Create and Edit Views
 
-`<Resource>` maps URLs to components - it takes care of *routing*. When you set a component as the `create` prop for a Resource, react-admin renders that component when users go to the `/[resource]/create` URL. When you set a component as the `edit` prop for a resource, react-admin renders that component when users go to the `/[resource]/:id` URL. 
+`<Resource>` maps URLs to components - it takes care of *routing*. When you set a component as the `create` prop for a Resource, react-admin renders that component when users go to the `/[resource]/create` URL. When you set a component as the `edit` prop for a resource, react-admin renders that component when users go to the `/[resource]/:id` URL.
```
djhi, line 8: There are a lot of unrelated and unnecessary changes here due to your markdown formatter probably. Can you please revert them?

docs/CreateEdit.md
```diff
@@ -541,6 +541,27 @@ The `transform` function can also return a `Promise`, which allows you to do all
 
 **Tip**: The message will be translated.
 
+**Tip**: `<Edit>`'s transform prop can recieve below variables from second argument.
+
+|  field         | Description                                                                                |
+| -------------- | ------------------------------------------------------------------------------------------ |
+| `previousData` | Previous data fetched or last updated on `dataProvider.update`                             |
+
```
djhi, line 549:
```suggestion
**Tip**: `<Edit>`'s transform prop function also get the `previousData` in its second argument:
```

docs/CreateEdit.md
```diff
@@ -2289,6 +2310,37 @@ const dataProvider = {
 }
 ```
 
+**Tip**: `<SaveButton>`'s transform prop inside `<Edit>` can recieve below variables from second argument.
+
+|  field         | Description                                                                                |
+| -------------- | ------------------------------------------------------------------------------------------ |
+| `previousData` | Previous data fetched or last updated on `dataProvider.update`                             |
```
djhi, line 2317:
```suggestion
**Tip**: `<Edit>`'s transform prop function also get the `previousData` in its second argument:
```

djhi (changes requested):

tkow:
@djhi 
Now, I revised them, and please review again.

`Vercel – react-admin-storybook`'s deploy is failed. Is that no problem?

tkow:
@djhi I'm sorry, but I couldn't find why Vercel - react-admin-storybook deployment has failed because I have no idea to know without vercel logs. It can be build in my local environment. 
Could you help me when you have time?

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -116,9 +118,13 @@ export const useEditController = <RecordType extends RaRecord = any>(
         ) =>
             Promise.resolve(
                 transformFromSave
-                    ? transformFromSave(data)
+                    ? transformFromSave(data, {
```
fzaninotto, line 121: check the linter warning for missing dependency

docs/CreateEdit.md
```diff
@@ -542,6 +542,22 @@ The `transform` function can also return a `Promise`, which allows you to do all
 
 **Tip**: The message will be translated.
 
+**Tip**: `<Edit>`'s transform prop function also get the `previousData` in its second argument:
+
+```jsx
+export const UserEdit = (props) => {
+    const transform = (data, { previousData }) => ({
```
fzaninotto, line 549: I'd prefer that the transform function signature be `(data, previousData)`

fzaninotto (changes requested):

fzaninotto:
The react-admin-storybook deployment randomly fails, don't worry about It. I redeployed it manually.

docs/CreateEdit.md
tkow: @fzaninotto 

Sorry, but I'd prefer my way, because it rarely  breaks user code when we need to add more optionalData compared to push it to the tail of arguments though the change may be seldom.
Increasing argument have risks when we want to add new argument, intermediate arguments can't change order without breaking change and even if they are not used in some cases, we may need to assign undefined when we call  or when we set no used variables' specification (like, (data, _, someEditControllerParameter) as event hander.

Argument addition as option is better way for me if the argument is really optional value. In this case, users can ignore previousData so, it's optional value.
But, I can't say that previousData must be optional form because I haven't found that we need more additional parameter in the future. My point is opposite to YAGNI, and  previousData may be just enough parameter in all edit controller context parameters to satisfy user demands. So, I'll fix that  if you still want to do nonetheless you have read this comment.

packages/ra-core/src/controller/edit/useEditController.ts
tkow: Thank you I've not noticed. It's green now.

CryptoGodfatherVA2 (approved):
VAR5

CryptoGodfatherVA2 (approved):

fzaninotto:
we'll wait for #7087 to be merged to merge this one, as I prefer that you rebase your small PR rather than asking @djhi to rebase his mega-jumbo-huge-not-so-small PR.

tkow:
I rebased my commits to remove no diff next branch merge commits.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7104, opened 2022-01-18
[Doc] customizing and disabling item reordering for ArrayInput

vaizki:
ArrayInput / FormIterator has buttons to reorder items but the documentation is lacking details about the customization / disabling of these buttons. In cases where order doesn't matter they only confuse users.

This is a pure documentation update, the functionality was already there.


WiXSL (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7115, opened 2022-01-20
Fix error messages can't be copied

### LINKED ISSUE #6749: Unable to copy the error message in clipboard
**What you were expecting:**

I expect the error message to be copyable in the clipboard.

image

**What happened instead:**

When I start selecting the error message, it toggles the accordion.

**Related code:**

https://codesandbox.io/s/confident-sanderson-o4igs?file=/src/posts/PostList.tsx

**Environment**

* React-admin version: 3.19.0
* React version: 17.0.0
* Browser: Chrome 95

WiXSL:
Fixes #6749

Before;
140480134-5697c857-b012-4517-b943-4d19c76b5da9

After:
error

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Error.tsx
+++ b/packages/ra-ui-materialui/src/layout/Error.tsx
@@ -45,6 +45,7 @@ const useStyles = makeStyles(
             maxWidth: '60em',
         },
         panelDetails: {
+            flexDirection: 'column',
             whiteSpace: 'pre-wrap',
         },
         toolbar: {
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/Error.tsx
```diff
@@ -84,11 +84,14 @@ const Error = (props: ErrorProps): JSX.Element => {
                 <div>{translate('ra.message.error')}</div>
                 {process.env.NODE_ENV !== 'production' && (
                     <>
+                        <Typography className={classes.panel}>
```
fzaninotto, line 87: you shouldn't use the same class for two distinct elements.

fzaninotto (changes requested):
Not fan of the design.

Here is an alternative suggestion: the AccordionSummary has a fixed title (like "Error Stack") in your PR, and the AccordionDetails contains both the error (in bold) and the error stack. 

It should make it copyable, no?

WiXSL:
> Not fan of the design.
> 
> Here is an alternative suggestion: the AccordionSummary has a fixed title (like "Error Stack") in your PR, and the AccordionDetails contains both the error (in bold) and the error stack.
> 
> It should make it copyable, no?

The thing with this idea is that you are showing an error and there is no actual error to display, until the user interacts with it

WiXSL:
closed-error
opened-error

How do you like it?

WiXSL:
The other option is to add css `user-select : 'all'` to the summary to solve the problem.

chrome-capture

fzaninotto (approved):
I prefer that solution! 

Note that we've adopted another solution in `next` (see 19d7a69efb0b903e0dcdd5ef43454e46bf253851), but I think your solution is better and should be backported.

PR #7116, opened 2022-01-20
Add support for meta parameter in queries

fzaninotto:


WiXSL (changes requested):

djhi (approved):
Nice! Are we going to add support for a `meta` prop on our components as well?

alanpoulain:
It reminds me of something :sweat_smile: (I know it's hard to maintain this framework, don't worry! I appreciate your work).
https://github.com/marmelab/react-admin/pull/4834

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -72,7 +73,7 @@ export const useUpdateMany = <RecordType extends RaRecord = any>(
 ): UseUpdateManyResult<RecordType> => {
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
-    const { ids, data } = params;
+    const { ids, data, meta } = params;
     const { mutationMode = 'pessimistic', ...reactMutationOptions } = options;
     const mode = useRef<MutationMode>(mutationMode);
     const paramsRef = useRef<Partial<UpdateManyParams<Partial<RecordType>>>>(
```
WiXSL, line 79: Initialized with empty object, instead of `params` ?
Same in `useDelete`

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -213,11 +218,13 @@ export const useUpdateMany = <RecordType extends RaRecord = any>(
                     const {
                         resource: callTimeResource = resource,
                         ids: callTimeIds = ids,
+                        meta: callTimeMeta = meta,
```
WiXSL, line 221: 'callTimeMeta' is assigned a value but never used

WiXSL (changes requested):

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -213,11 +218,13 @@ export const useUpdateMany = <RecordType extends RaRecord = any>(
...
                     } = variables;
                     updateCache({
                         resource: callTimeResource,
                         ids: callTimeIds,
                         data,
+                        meta,
```
WiXSL, line 227: Here it should be `callTimeMeta` instead of `meta`

WiXSL (changes requested):

docs/Actions.md
```diff
@@ -103,6 +103,19 @@ const { data: user, isLoading, error } = useGetOne(
 );
 ```
 
+## `meta` Parameter
+
+All Data Provider methods accept a `meta` parameter. React-admin doesn't set this parameter by default in its queries, but it's a good way to pass special arguments or metadata to an API call.
+
+```jsx
```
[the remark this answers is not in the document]
fzaninotto: Why? We're using jsx elsewhere. 

I know there is no jsx here, but the js and jsx highlight colors aren't the same. If we don't use jsx everywhere, this will look odd.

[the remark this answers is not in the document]
WiXSL: Got it

alanpoulain:
> We're adding meta to the `request` in this PR, not in the `response`.
> 
> As for the response, we're currently working on adding support for partial pagination (i.e. list responses with no `total` but with a `hasNextPage` boolean) rather than a generic meta.

OK that's nicer to have a well-defined API for the partial pagination, but I still think having a `meta` parameter in the response is needed, especially for the metadata-oriented API (JSON-LD/Hydra) or when custom information are sent in headers.
React-admin cannot think of all the use cases (partial pagination was one of them).

fzaninotto:
> OK that's nicer to have a well-defined API for the partial pagination, but I still think having a `meta` parameter in the response is needed, especially for the metadata-oriented API (JSON-LD/Hydra) or when custom information are sent in headers. React-admin cannot think of all the use cases (partial pagination was one of them).

With the new Query API (using react-query), if you ever need to call your data provider with custom params and read the full response, it's super easy. I don't think we should handle that in core.

alanpoulain:
> With the new Query API (using react-query), if you ever need to call your data provider with custom params and read the full response, it's super easy. I don't think we should handle that in core.

That's good to know, it will be feasible to add custom behavior to components using data providers then, but I still think that it would be great if data are analyzed and returned with a special meta parameter by data providers, since it's their responsibility.
Anyway maybe more use cases will appear in the future and maybe you will reconsider it at this time.

RWOverdijk:
> Nice! Are we going to add support for a `meta` prop on our components as well?

I am curious about this as well. Our API only returns relations when asked (_embed) since we're using ra-data-json-server which uses _embed for that as well because that's the spec. Without that we can't use <Edit /> (and more specifically ReferenceArrayInput) and need a custom form.

fzaninotto:
You don't need an embed to use ReferenceArrayInput. ReferenceArrayInput fetches the related data itself.

RWOverdijk:
@fzaninotto It doesn't seem to do that for edit. I mean, it fetches the options, sure. But it doesn't change anything to the getOne of the <Edit> that tells the api to include the currently linked ids.

PR #7119, opened 2022-01-21
Add onSuccess and onFailure handlers to bulk delete buttons

martinlamas:
In this PR I add the `onSuccess` and `onFailure` handlers to the bulk delete buttons. This is the same behavior in the `DeleteButton` component.

Thanks!


fzaninotto:
Hi Martin, and thanks for your PR.

The `master` branch is feature-freeze, and doesn't accept PRs for new features. 

You can open a new PR against `next`, but beware that we are in the middle of a major refactoring and [the conventions have changed](https://github.com/marmelab/react-admin/blob/next/UPGRADE.md#onsuccess-and-onfailure-props-have-moved).

PR #7120, opened 2022-01-21
Partial pagination

### LINKED ISSUE #6300: Partial pagination support (related to api-platform/admin)

Hi! In api-platform/admin there is a support for partial pagination, it's very useful when we want to display `List` from a very big table where `count(*)` may take few minutes to be calculated. The problem is it stopped working. I've reported the problem here: https://github.com/api-platform/admin/issues/379. Partial pagination has been implemented by keeping information about it in `total`, where:
 - `total = -1` – partial pagination is enabled, and we're on the last page;
 - `total = -2` – partial pagination is enabled, and there is next page after current page;
 - `total = -3` – partial pagination is enabled, and we don't know whether there is next page.

This solution stopped working when react-admin in version 3.7.2 started to validate if `query.page > totalPages`.

The simplest solution would be to allow `total < 0` in https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/useListController.ts#L189, but that seems a bit ugly.

An alternative would be to add some `meta` information to pagination. It'd allow us to implement partial pagination, but also other things, e.g. if we'd like to add some kind of `approxTotal`, and approximate total results instead of performing direct `count(*)` – I'm sure we'd like to distinguish `total` from `approxTotal` in the user interface.

**Additional context**
For more info, please check: https://github.com/api-platform/admin/issues/379.

fzaninotto:
## Problem

The `getList()` and `getManyReference()` methods return paginated responses. Sometimes, executing a "count" server-side to return the `total` number of records is expensive.

## Solution

In this case, you can omit the `total` property in the response, and pass a `pageInfo` object instead, specifying if there are previous and next pages:

```js
dataProvider.getList('comments', {
    pagination: { page: 1, perPage: 6 },
    sort: { field: 'created_at', order: 'DESC' },
})
.then(response => console.log(response));
// {
//     data: [
//         { id: 126, body: "allo?", author_id: 12 },
//         { id: 127, body: "bien le bonjour", author_id: 12 },
//         { id: 124, body: "good day sunshine", author_id: 45 },
//         { id: 123, body: "hello, world", author_id: 67 },
//         { id: 125, body: "howdy partner", author_id: 89 },
//         { id: 138, body: "nice sweater!", author_id: 78 },
//     ],
//     pageInfo: {
//         hasPreviousPage: false,    
//         hasNextPage: true,
//     }
// }
```

React-admin's `<Pagination>` component will automatically handle the `pageInfo` object and display the appropriate pagination controls.

https://user-images.githubusercontent.com/99944/151117525-f103fd6f-252f-474e-a692-914a7685df2a.mov

And while we are at it, time to upgrade to the new MUI `<Pagination>` component!

https://user-images.githubusercontent.com/99944/151118348-269d8254-bc99-4fd3-8e40-34f38ed72d07.mov

- [x] add `pageInfo` in the response signature for `getList` and `getManyReference`
- [x] add support for `pageInfo` in `useGetList` 
- [x] add support for `pageInfo` in `useGetManyReference`
- [x] add support for `pageInfo` in `useListController`
- [x] add support for `pageInfo` in `useList`
- [x] add support for `pageInfo` in `useReferenceManyFieldController`
- [x] add support for `pageInfo` in `useReferenceInputController`
- [x] add support for `pageInfo` in `useReferenceArrayInputController`
- [x] Add new component to render Pagination controls for partial pagination
- [x] Migrate the `<PaginationActions>` to the new MUI `<Pagination>` component
- [x] Add tests
- [x] Add documentation

Closes #6300


packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
```diff
@@ -106,6 +110,12 @@ export const Pagination: FC<PaginationProps> = memo(props => {
         );
     }
 
+    const ActionsComponent = actions
+        ? actions // overridden b ycaller
```
WiXSL, line 114:
```suggestion
        ? actions // overridden by caller
```

WiXSL (changes requested):

packages/ra-language-english/src/index.ts
```diff
@@ -111,9 +111,15 @@ const englishMessages: TranslationMessages = {
...
             page_out_from_begin: 'Cannot go before page 1',
             page_range_info: '%{offsetBegin}-%{offsetEnd} of %{total}',
+            partial_page_range_info:
+                '%{offsetBegin}-%{offsetEnd} of more than %{offsetEnd}',
+            current_page: 'Page %{page}',
+            page: 'Go to page %{page}',
+            first: 'Go to first page',
+            last: 'Go to last page',
+            next: 'Go to next page',
+            previous: 'Go to previous page',
```
WiXSL, line 121: BC detected: You change `prev` for `previous`

packages/ra-language-french/src/index.ts
```diff
@@ -115,9 +115,15 @@ const frenchMessages: TranslationMessages = {
...
+                '%{offsetBegin}-%{offsetEnd} sur plus de %{offsetEnd}',
             page_rows_per_page: 'Lignes par page :',
-            next: 'Suivant',
-            prev: 'Précédent',
+            current_page: 'Page %{page}',
+            page: 'Aller à la page %{page}',
+            first: 'Aller à la première page',
+            last: 'Aller à la dernière page',
+            next: 'Aller à la prochaine page',
+            previous: 'Aller à la page précédente',
```
WiXSL, line 126: Same

WiXSL (changes requested):

packages/ra-ui-materialui/src/list/pagination/PaginationActions.tsx
```diff
@@ -1,177 +1,67 @@
...
+        <Root>
+            <Pagination
+                size={size}
+                count={nbPages}
+                // <TablePagination>, the parent, uses 0-based pagination
+                // while <Pagination> uses 1-based pagination
+                page={page + 1}
+                onChange={(e: any, page) => onPageChange(e, page - 1)}
+                {...sanitizeRestProps(rest)}
+                getItemAriaLabel={getItemAriaLabel}
```
WiXSL, line 54: `getItemAriaLabel` cannot be overridden

WiXSL (changes requested):

packages/ra-ui-materialui/src/list/pagination/PaginationActions.tsx
fzaninotto: that's on purpose. The parent, TablePagination, passes its own getItemAriaLabel function, which is compatible with TablePaginationActions, but not with Pagination (the two signatures are different).

packages/ra-language-english/src/index.ts
fzaninotto: indeed, I need to add an entry to the Upgrade guide

WiXSL (approved):

WiXSL:
e2e test failing

djhi (changes requested):
Few linter warnings left, otherwise :muscle:

djhi (approved):

PR #7133, opened 2022-01-24
fix: initialValue prop not used when input value is null

### LINKED ISSUE #7132: DateField: initialValue not working when value is null
**What you were expecting:**
The `DateInput` component should use the value provided by the `initialValue` prop when the value is null or undefined.

As state in the docs for the [DateInput](https://marmelab.com/react-admin/Inputs.html) component:

Prop | Required | Type | Default | Description
-- | -- | -- | -- | --
initialValue | Optional | mixed | - | Value to be set when the property is null or undefined

**What happened instead:**
When the value is `null` the `DateInput` ignores the value provided by the `initialValue` prop.
It does work however when the value is `undefined` the `DateInput` component correctly shows the value provided by the `initialValue` prop.

**Steps to reproduce:**

**Related code:**
I have forked the example sandbox and modified the data of the first two posts.
The first post now has `null` for date published.
The second post now has `undefined` for date published.

Next, I have adjusted the edit form for a post so that the `published_at` value now has  an `initialValue`  prop with a date.

```javascript
<DateInput
            source="published_at"
            options={{ locale: "pt" }}
            initialValue={new Date(2100, 0, 1)}
 />
```

See the sandbox: https://codesandbox.io/s/compassionate-bush-s772t
Click the first post and open the misc. tab. See the value of the published field.
**Expected**: 01/01/2100
**Actual**: mm/dd/yyy

Click the second post and open the misc. tab. See the value of the published field.
**Expected**: 01/01/2100
**Actual**: 01/01/2100

**Other information:**

**Environment**

* React-admin version: 3.19.7
* Last version that did not exhibit the issue (if applicable): I believe 3.19.5
* React version: 17.0.0
* Browser: Version 97.0.4692.71 (Official Build) (64-bit)
* Stack trace (in case of a JS error): n/a

I have traced down the source code to the following line:
https://github.com/marmelab/react-admin/blob/a1c602c37428d0e5449945f9bc64c6237b14b877/packages/ra-core/src/form/useInput.ts#L78

Where lodash `get` method is used to get the value for a given property from the record object. However, this only returns the defaultValue if the actual value is 'undefined'. See docs for [lodash get method](https://www.geeksforgeeks.org/lodash-_-get-method/).

thdk:
fixes #7132

I have replaced the lodash get method so that it is easier for maintainers / contributors to see what happens.
Lodash `get` will not use the default value when the result is `null`. While we need the initialValue to be returned for both null and undefined values as written in react admin docs. See docs for, the DateInput component.

### DIFF
--- a/packages/ra-core/src/form/useInput.spec.tsx
+++ b/packages/ra-core/src/form/useInput.spec.tsx
@@ -206,6 +206,38 @@ describe('useInput', () => {
         expect(queryByDisplayValue('foo')).not.toBeNull();
     });
 
+    it('applies the initialValue when input has null value', () => {
+        const { queryByDisplayValue } = renderWithRedux(
+            <FormWithRedirect
+                record={{
+                    title: null,
+                }}
+                onSubmit={jest.fn()}
+                render={() => {
+                    return (
+                        <Input
+                            source="title"
+                            resource="posts"
+                            initialValue="foo"
+                        >
+                            {({ id, input }) => {
+                                return (
+                                    <input
+                                        type="text"
+                                        id={id}
+                                        aria-label="Title"
+                                        {...input}
+                                    />
+                                );
+                            }}
+                        </Input>
+                    );
+                }}
+            />
+        );
+        expect(queryByDisplayValue('foo')).not.toBeNull();
+    });
+
     it('does not apply the initialValue when input has a value of 0', () => {
         const { queryByDisplayValue } = renderWithRedux(
             <FormWithRedirect
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -5,6 +5,7 @@ import {
     FieldRenderProps,
     FieldInputProps,
 } from 'react-final-form';
+import get from 'lodash/get';
 import { Validator, composeValidators } from './validate';
 import isRequired from './isRequired';
 import { useFormGroupContext } from './useFormGroupContext';
@@ -73,7 +74,7 @@ const useInput = ({
     // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level initialValues so that it works as it should in final-form
     // (ie. field level initialValue override form level initialValues for this field).
-    let inputInitialValue = record ? record[source] : undefined;
+    let inputInitialValue = get(record, source);
     if (inputInitialValue === null || inputInitialValue === undefined) {
         inputInitialValue = initialValue;
     }
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -74,12 +74,9 @@ const useInput = ({
     // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level initialValues so that it works as it should in final-form
     // (ie. field level initialValue override form level initialValues for this field).
-    let inputInitialValue = get(record, source);
-    if (inputInitialValue === null || inputInitialValue === undefined) {
-        inputInitialValue = initialValue;
-    }
+    const inputInitialValue = get(record, source, initialValue);
     const { input, meta } = useFinalFormField(finalName, {
-        initialValue: inputInitialValue,
+        initialValue: inputInitialValue ?? initialValue,
         defaultValue,
         validate: sanitizedValidate,
         ...options,
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -5,7 +5,6 @@ import {
     FieldRenderProps,
     FieldInputProps,
 } from 'react-final-form';
-import get from 'lodash/get';
 import { Validator, composeValidators } from './validate';
 import isRequired from './isRequired';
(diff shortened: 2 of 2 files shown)


packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +73,12 @@ const useInput = ({
     // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level initialValues so that it works as it should in final-form
     // (ie. field level initialValue override form level initialValues for this field).
+    let inputInitialValue = record ? record[source] : undefined;
```
WiXSL, line 76:
```suggestion
    let inputInitialValue = record ? get(record, source) : undefined;
```

WiXSL: We still need to use lodash to support nested source

WiXSL (changes requested):

packages/ra-core/src/form/useInput.ts
WiXSL: Thank you for your contribution!

packages/ra-core/src/form/useInput.spec.tsx
```diff
@@ -206,6 +206,38 @@ describe('useInput', () => {
         expect(queryByDisplayValue('foo')).not.toBeNull();
     });
 
+    it('applies the initialValue when input has null value', () => {
```
WiXSL, line 209:
```suggestion
    it('should apply initialValue when input has null value', () => {
```

WiXSL (changes requested):

WiXSL (approved):

thdk:
@WiXSL I've just found out how to test is with the examples/simple project. Maybe wait with review untill I confirmed it is working as expected there.

thdk:
@WiXSL Yep, I can confirm this is now working as expected!

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
     // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level initialValues so that it works as it should in final-form
     // (ie. field level initialValue override form level initialValues for this field).
+    let inputInitialValue = get(record, source);
```
WiXSL, line 77:
```suggestion
    const inputInitialValue = get(record, source, initialValue);
```

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
...
+    if (inputInitialValue === null || inputInitialValue === undefined) {
```
WiXSL, line 78:
```suggestion
```

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
...
+        inputInitialValue = initialValue;
```
WiXSL, line 79:
```suggestion
```

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
...
+    }
     const { input, meta } = useFinalFormField(finalName, {
-        initialValue: get(record, source, initialValue),
+        initialValue: inputInitialValue,
```
WiXSL, line 82:
```suggestion
	initialValue: inputInitialValue ?? initialValue,
```

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
     // This ensure dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level initialValues so that it works as it should in final-form
     // (ie. field level initialValue override form level initialValues for this field).
+    let inputInitialValue = get(record, source);
+    if (inputInitialValue === null || inputInitialValue === undefined) {
+        inputInitialValue = initialValue;
+    }
```
WiXSL, line 80:
```suggestion
```

WiXSL (changes requested):

packages/ra-core/src/form/useInput.ts
WiXSL: Maybe is simpler like this

packages/ra-core/src/form/useInput.ts
```diff
@@ -74,8 +74,12 @@ const useInput = ({
...
     const { input, meta } = useFinalFormField(finalName, {
-        initialValue: get(record, source, initialValue),
```
WiXSL, line 78:
```js
const inputInitialValue = get(record, source, initialValue);

const { input, meta } = useFinalFormField(finalName, {
	initialValue: inputInitialValue ?? initialValue,

WiXSL: Maybe you can rewrite it like this.

WiXSL (changes requested):

WiXSL (approved):

WiXSL:
@thdk, So after discussion the solution is to fix the documentation, not the inputs.
So `null` value should be handled in userland.

fzaninotto:
Superseded by #7139

PR #7136, opened 2022-01-25
Fix failing admin test when running all tests together

thdk:
I noticed that on my machine this test always failed with a timeout exception when running `yarn test-unit` in the root project.

It did not fail however when running `yarn test-unit --watch` because then it took just below the default 5000ms timeout setting of jest.

I have increased the timeout for this specific test and applied some best practices for this test.

I've also fixed the incorrect intellisense for the test files by explicitly setting `jest` in the required types in tsconfig.json so it doesn't look at mochas typescript definition files any longer.


WiXSL:
Thanks, I really like this one!

WiXSL (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7143, opened 2022-01-26
[RFC] Can define `variant` and `margin` for Inputs globaly with Material UI theming

guillemfondin:
## Description

For some reason, in enterprise project we need to keep ReactAdmin v2 `variant: 'standard'` for all inputs. 

Our forms are very complex, with Material Ui `<SteppedForm>`, and with a lot (some much) inputs. We can't change `variant` and `margin` in `<SimpleForm>` because we don't use it and define it in each `<Input>` add so much lines of duplication code.

To resolve it, I suggest the possibility to can define `variant` and `margin` globaly with theming system.

## Usage

```jsx
import { createTheme } from '@material-ui/core/styles';

const theme = createTheme({
    components: {
        MuiTextField: { // define the props of all instances of this component
            defaultProps: {
                variant: 'standard',
                margin: 'normal',
            }
        },
        MuiFormControl: {} // Same properties
    },
});

const App = () => (
    <Admin theme={theme}>
        // ...
    </Admin>
);

```

### DIFF
--- a/docs/CreateEdit.md
+++ b/docs/CreateEdit.md
@@ -1771,6 +1771,36 @@ export const PostEdit = () => (
 );
 ```
 
+**Tip**: If, for some reason, you need to define the `variant` globaly for all your app, you can creat your own theme, following [Material UI themes documentation](https://material-ui.com/customization/themes/).
+
+
+```jsx
+import { defaultTheme } from 'react-admin';
+import { createTheme } from '@mui/material/styles';
+
+const theme = createTheme({
+    ...defaultTheme,
+    components: {
+        MuiTextField: { // define the props of all instances of this component
+            defaultProps: {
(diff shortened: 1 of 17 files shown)


djhi:
Hi and thank you for the PR. This is a very hacky solution though and we won't merge it. You can however do that in your project by re-exporting wrappers around our components and applying styles using the theme as you do in this PR. Thanks anyway!

djhi:
Sorry about the tone of my previous message. 

I wanted to add that we are planning to remove the `variant` and `margin` props from the form components in v4. Instead, we will document that you can set a different value for these props through the theme as you did. The only difference is that you'll have to do it for each input and not just the Mui TextField.

guillemfondin:
Hi, 
Yes very hacky solution, the [RFC] tag was here for this reason :) 
Ok, thanks for information !

PR #7146, opened 2022-01-26
Builds the demos on CI

djhi:
Build both the e-commerce and crm demos on the CI to catch more issues

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,8 +79,6 @@ jobs:
       run: make build
     - name: Build e-commerce
       run: make build-demo
-    - name: Build crm
-      run: make build-crm
 
   crm:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fzaninotto:
We should reuse workflows to avoid cloning and building the code 3 times

https://docs.github.com/en/actions/using-workflows/reusing-workflows

djhi:
Reusing workflows just reuse the steps, not their output. Reusing the output doesn't make sense here as we would have to wait for the `typecheck` task to finish before building the demos. That would take more time than rebuilding the packages before building the demos.

With the current setup, all our tasks take approximatelly the same time so I think it's fine

fzaninotto:
I don't understand your point. Both your jobs run make build, which does typechecking. So they do have to wait for it.

fzaninotto (approved):

PR #7149, opened 2022-01-27
[4.0.0alpha.2] Demos fix useRecordContext usage

### LINKED ISSUE #7145: [4.0.0-alpha.2] Demos upgrade issues
Some CRM app components lack upgrade on accessing a record from `useRecordContext` instead of the `record` prop

WiXSL:
Fixes #7145

### DIFF
--- a/examples/crm/src/companies/CompanyCard.tsx
+++ b/examples/crm/src/companies/CompanyCard.tsx
@@ -59,7 +59,7 @@ const StyledMuiLink = styled(MuiLink)(({ theme }) => ({
     },
 }));
 
-export const CompanyCard = (props: { record: Company }) => {
+export const CompanyCard = (props: { record?: Company }) => {
     const [elevation, setElevation] = useState(1);
     const createPath = useCreatePath();
     const record = useRecordContext<Company>(props);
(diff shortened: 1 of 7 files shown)


examples/crm/src/companies/CompanyCard.tsx
```diff
@@ -59,9 +59,12 @@ const StyledMuiLink = styled(MuiLink)(({ theme }) => ({
     },
 }));
 
-export const CompanyCard = ({ record }: { record: Company }) => {
+export const CompanyCard = (props: { record: Company }) => {
     const [elevation, setElevation] = useState(1);
     const createPath = useCreatePath();
+    const record = useRecordContext<Company>(props);
```
djhi, line 65: I think we should you should put a RecordContextProvider in `GridList` if we really want to use the record context. Seems overkill to me as we manually pass the record

examples/crm/src/contacts/TagsList.tsx
```diff
@@ -36,8 +36,10 @@ const ColoredChipField = (props: any) => {
     );
 };
 
-export const TagsList = ({ record }: { record: Contact }) => {
+export const TagsList = (props: { record: Contact }) => {
```
djhi, line 39: Same remark for this one

djhi (approved):

PR #7155, opened 2022-01-28
Cleanup useless Redux actions

fzaninotto:
Just reemoving some unused code


packages/ra-core/src/dataProvider/withDataProvider.tsx
```diff
@@ -1,64 +0,0 @@
-import * as React from 'react';
```
fzaninotto, line 1: oops, this one was documented, wo we need to add aline in the upgrade guide about it.

djhi (approved):
Should I wait for the upgrade guide update before merging?

packages/ra-core/src/dataProvider/withDataProvider.tsx
WiXSL: And remove it from `Reference.md`

WiXSL (changes requested):

WiXSL (approved):

djhi (approved):

PR #7161, opened 2022-01-29
Export field type mapping

smeng9:
We used to import field type mappings using `import raListFieldTypes from "ra-ui-materialui/lib/list/listFieldTypes";` from the /lib/ folder of the built package. It seems in v4 alpha2 the /lib/ folder disappeared in the npm package, do we plan to add them back? Previously the /lib/ folder contains compiled js file for each ts file.

We noticed there is a change in v4, some of them are re-exported under "react-admin" https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/list/index.ts#L10 while edit and show field types are not exported.

This pull request adds the exports.


smeng9:
There seems to be flaky tests in useLogoutIfAccessDenied.spec.tsx

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7162, opened 2022-01-29
Speed up development on demos by using esbuild and aliases

djhi:
## Problem

Working on the e-commerce and crm demos is a pain because of the default CRA config:
- You have to build the ra packages at least once before running the demos
- You have to rebuild the ra packages whenever they change
- You have to restart the demo to actually see the changes
- The e-commerce demo takes forever to start and load

## Solution

For development **only**, leverage esbuild and aliases using craco

### Webpack:
- need to build ra packages at least once and rebuild them each time they change: 1m21
- first run of the demo: 1m until usable
- subsequent runs of the demo: 15s

### Craco with Esbuild and aliases (dev only):
- no need to build/rebuild the packages ever
- first run of the demo: 30s until usable
- subsequent runs of the demo: 10s

## Limitations

While we don't have to build ra packages anymore while working on the demos, I haven't been able to make fast refresh handle the ra packages changes so you still have to refresh the page manually. We didn't have this feature with the default CRA either though so we still get some great benefits.

### DIFF
--- a/examples/crm/craco.config.js
+++ b/examples/crm/craco.config.js
@@ -1,0 +1,85 @@
+const CracoEsbuildPlugin = require('craco-esbuild');
+const CracoAlias = require('craco-alias');
+const fs = require('fs');
+const path = require('path');
+
+const packages = fs.readdirSync(path.resolve(__dirname, '../../packages'));
+const aliases = packages.reduce((acc, dirName) => {
+    const packageJson = require(path.resolve(
+        __dirname,
+        '../../packages',
+        dirName,
+        'package.json'
+    ));
+    acc[packageJson.name] = path.resolve(
+        __dirname,
+        `../../packages/${packageJson.name}/src`
+    );
+    return acc;
+}, {});
+
+const findWebpackPlugin = (webpackConfig, pluginName) =>
+    webpackConfig.resolve.plugins.find(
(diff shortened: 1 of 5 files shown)


lukascivil:
Very good, but a doubt, why not use Vite? same as [example/simple](https://github.com/marmelab/react-admin/tree/master/examples/simple) ?

WiXSL (dismissed):

WiXSL (commented):
- no need to build/rebuild the packages ever => ✔️ 
- first run of the demo: 30s until usable  => ✔️ 
- subsequent runs of the demo: 10s  => ✔️

WiXSL (commented):
Could it be that fast refresh is working in CRM but not in Demo ?

I get an `Unmatched GET to /main.839e76f6a35f6f79fe0c.hot-update.json`

error

fzaninotto:
Also, does it typecheck the demos?

djhi:
At build time, yes. Otherwise, it happens in your editor just like the packages

djhi:
We already uses yarn workspaces and esbuild (through tsup). Lerna is only used for publishing

antdking:
alas, it's not supported yet: https://github.com/egoist/tsup/issues/571

WiXSL:
Superseded by #8317

Issue #7079, opened 2022-01-13
[4.0.0-alpha.1]: Style overrides for ReactAdmin components


emulienfou:
**What you were expecting:**

Should be able to overrides style for ReactAdmin components

**What happened instead:**

Nothing since to be happening when applying the overrides. I presume this is not implemented yet!

**Steps to reproduce:**

1. Create a new theme options object
2. Add components override for RaSidebar by example
3. Override background color of the sidebar

**Related code:**

```
import { PaletteMode, Components as MuiComponents } from '@mui/material'
import { RaThemeOptions } from 'react-admin'

interface Components extends MuiComponents {
  RaSidebar?: {
    styleOverrides?: any;
    drawerPaper?: any;
  };
  RaMenuItemLink?: {
    styleOverrides?: any;
  }
}

interface ThemeOptions extends RaThemeOptions {
  components: Components
}

export const themeOptions = (mode: PaletteMode): ThemeOptions => ({
  components: {
    RaSidebar: {
      styleOverrides: { root: { backgroundColor: 'red' } }
    },
    RaMenuItemLink: {
      styleOverrides: { root: { padding: 0 } }
    }
  }
})
```

**Other information:**

> Should be nice to have an updated version of the type `RaThemeOptions` including ReactAdmin components as well (like I did in the code bellow)!

**Environment**

* React-admin version: 4.0.0-alpha.1
* Last version that did not exhibit the issue (if applicable):
* React version: 17.0.2
* Browser: Opera/Chrome
* Stack trace (in case of a JS error):
* Typescript version: 4.5.4

fzaninotto:
Weird, we use style overrides in the ecommerce and crm demos, and they definitely work in `next`.
 
https://github.com/marmelab/react-admin/blob/998e254fe09121892b4fc28a97639e50f9c7483b/examples/demo/src/layout/themes.ts#L16-L21

Could you add a link to a CodeSandbow demonstrating the issue?

emulienfou:
@fzaninotto here is a Codesandbox if you want to check: https://codesandbox.io/s/react-admin-typescript-styles-overrides-tplop

I'm trying to overrides Ra components

djhi:
Closing this one as duplicate. Thanks for the report

PR #7166, opened 2022-01-31
Fix theming regressions

### LINKED ISSUE #7164: [v4 α2] Regression on theme styleOverrides
After fixing dependency issue in latest alpha demo, there seems to be some regression on theme overrides.

**What you were expecting:**

The Posters Galore Application
For example, in v3 the MenuItemLink should have a left border when active
reactadminbug

**What happened instead:**

In v4 there is no left border even when we manually override it.
Also AppBar background color is also not effective.
react-adminnobug

**Steps to reproduce:**

Clone the https://github.com/marmelab/react-admin/tree/next/examples/demo and resolve all dependency issue in codesandbox

**Related code:**

```
https://codesandbox.io/s/exciting-bhabha-82u9j?file=/src/layout/themes.ts:1928-1950
```

**Other information:**

**Environment**

* Last version that did not exhibit the issue (if applicable): 3.19.7
* React version: 17.0.2
* Browser: Edge 97

djhi:
Adding theme override support is more complicated than we thought and require more code.

We'll add more components once we have an answer to the following SO question: https://stackoverflow.com/questions/70927246/theme-override-support-for-custom-components

Fixes #7164, #7079

### DIFF
--- a/examples/demo/src/App.tsx
+++ b/examples/demo/src/App.tsx
@@ -7,6 +7,7 @@ import authProvider from './authProvider';
 import { Login, Layout } from './layout';
 import { Dashboard } from './dashboard';
 import englishMessages from './i18n/en';
+import { lightTheme } from './layout/themes';
 
 import visitors from './visitors';
 import orders from './orders';
@@ -40,6 +41,7 @@ const App = () => {
             layout={Layout}
             i18nProvider={i18nProvider}
             disableTelemetry
+            theme={lightTheme}
         >
             <CustomRoutes>
                 <Route path="/configuration" element={<Configuration />} />
--- a/packages/ra-ui-materialui/src/layout/AppBar.tsx
+++ b/packages/ra-ui-materialui/src/layout/AppBar.tsx
@@ -143,7 +143,14 @@ export const AppBarClasses = {
     title: `${PREFIX}-title`,
 };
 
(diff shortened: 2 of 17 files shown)


smeng9:
Saw your question on stackoverflow

Is it possible that we don't use overridesResolver and do following:
(Haven't tried that, just a thought)
```
const PREFIX = 'RaMenuItemLink';

export const MenuItemLinkClasses = {
    root: `${PREFIX}-root`,
    active: `${PREFIX}-active`,
    icon: `${PREFIX}-icon`,
};

const StyledMenuItem = styled(MenuItem, {
    name: PREFIX,
})(({ theme }) => ({
    [`&.${MenuItemLinkClasses.root}`]: {
        color: theme.palette.text.secondary,
        ...theme.components[PREFIX].styleOverrides.root
    },

    [`&.${MenuItemLinkClasses.active}`]: {
        color: theme.palette.text.primary,
        ...theme.components[PREFIX].styleOverrides.active
    },

    [`& .${MenuItemLinkClasses.icon}`]: { minWidth: theme.spacing(5), ...theme.components[PREFIX].styleOverrides.icon },
}));
```

djhi:
Hi @smeng9! Thanks for the suggestion but it still requires us to revisit all our components. Anyway, @oliviertassinari answered me on SO and it looks like we indeed have to set this `overridesResolver` everywhere to support theme override :sob:

fzaninotto (approved):

PR #7168, opened 2022-01-31
Ensure react-admin has a dependency to react-hook-form

### LINKED ISSUE #7160: [v4 alpha2] Missing react-hook-form dependency in demo code
**What you were expecting:**
Examples should run without issue

**What happened instead:**
Seems we did not bundle the react-hook-form in react-admin

ra-core, ra-ui-materialui should include a dependency

Did we just exclude it here by making it external? https://github.com/marmelab/react-admin/blob/next/packages/ra-core/package.json#L23

**Temporary solution:**
`yarn add react-hook-form` should do this trick to resolve the missing dependency for now

**Related code:**

```
[example/tutorial](https://codesandbox.io/s/github/marmelab/react-admin/tree/next/examples/tutorial?file=/src/App.js)
```

**Other information:**

**Environment**

* Last version that did not exhibit the issue (if applicable): N/A react-hook-form is newly introduced
* React version: 17.0.2
* Browser: Firefox 96
```
DependencyNotFoundError

Could not find dependency: 'react-hook-form' relative to '/node_modules/ra-core/dist/esm/index.js'

Suggested solutions:

Add react-hook-form as dependency
```

djhi:
Should fix the dependency issue on tutorial.

Fixes #7160

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25469,6 +25469,7 @@ __metadata:
     ra-i18n-polyglot: ^4.0.0-alpha.2
     ra-language-english: ^4.0.0-alpha.2
     ra-ui-materialui: ^4.0.0-alpha.2
+    react-hook-form: ^7.25.0
     react-redux: ^7.1.0
     react-router: ^6.1.0
     react-router-dom: ^6.1.0


WiXSL (commented):
It works ok

fzaninotto (approved):

PR #7174, opened 2022-02-02
Simplify Theming And Uniformize Classes

djhi:


packages/ra-ui-materialui/src/layout/TopToolbar.tsx
```diff
@@ -23,27 +15,21 @@ TopToolbar.propTypes = {
 export default TopToolbar;
 const PREFIX = 'RaTopToolbar';
 
-export const TopToolbarClasses = {
-    root: `${PREFIX}-root`,
-};
-
 const StyledToolbar = styled(Toolbar, { name: PREFIX })(({ theme }) => ({
```
WiXSL, line 18:
```suggestion
const StyledToolbar = styled(Toolbar, { name: PREFIX, overridesResolver: (props, styles) => styles.root })(({ theme }) => ({
```

WiXSL (changes requested):

packages/ra-ui-materialui/src/layout/CardActions.tsx
```diff
@@ -1,45 +0,0 @@
...
-import { styled } from '@mui/material/styles';
-import PropTypes from 'prop-types';
-import { warning } from 'ra-core';
-import classnames from 'classnames';
-
-export const CardActions = props => {
-    const { className, children, ...rest } = props;
-    warning(
-        true,
-        '<CardActions> is deprecated. Please use the <TopToolbar> component instead to wrap your action buttons'
```
fzaninotto, line 11: needs a mention in the upgrade guide

fzaninotto (changes requested):

packages/ra-ui-materialui/src/layout/CardActions.tsx
djhi: Done

WiXSL (approved):

fzaninotto (approved):

PR #7183, opened 2022-02-03
[4.0.0-alpha3] Fix SelectInput proptypes

### LINKED ISSUE #7179: [4.0.0-alpha.2] Unexpected behavior since allowEmpty as been removed, empty value always present
**What you were expecting:**

Empty MenuItem should not be present since `allowEmpty: false`

**What happened instead:**

An empty MenuItem is present and cannot be removed.
I followed the upgrade guide here: https://github.com/marmelab/react-admin/blob/next/UPGRADE.md#allowempty-has-been-removed-from-selectinput-autocompleteinput-and-autocompletearrayinput
However, according to the upgrade guide I should NOT have this empty value unless I add: `validate={[required()]}` to the component.

**Steps to reproduce:**

```tsx
<SelectInput source="type" choices={ [
        { id: 'Full Time', name: 'Full Time' },
        { id: 'Past Time', name: 'Past Time' }
      ] }/>
```

Here is the current render using the code bellow:
image

**Environment**

* React version: 17.0.2
* Browser: Chrome 97

WiXSL:
Fixes #7179

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectInput.tsx
@@ -296,7 +296,6 @@ export const SelectInput = (props: SelectInputProps) => {
 };
 
 SelectInput.propTypes = {
-    allowEmpty: PropTypes.bool,
     emptyText: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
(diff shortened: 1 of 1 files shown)


djhi (approved):
:+1:

PR #7184, opened 2022-02-03
[4.0.0-alpha.3] Remove `classes` prop from components

WiXSL:


WiXSL (commented):
Should I remove this from `types.tsx` ?
```js
export type ClassesOverride<
    UseStyles extends (props: any) => Record<string, string>
> = Partial<Record<keyof ReturnType<UseStyles>, string>>;

Issue #7170, opened 2022-02-01
`resolveBrowserLocale` not returning full locale


FernandoKGA:
**What you were expecting:**

When using `resolveBrowserLocale`, it would get the correctly locale and translate to it.

**What happened instead:**

The function returned only the first part of the locale. In my case, the locale is **"pt-BR"**, which returned as **"pt"** only.

**Steps to reproduce:**

1. Use `resolveBrowserLocale` at `polyglotI18nProvider` `initialLocale` paremeter or other function that receives a locale from `resolveBrowserLocale`.
2. Use a language at the browser that has a region locale (example: `pt-BR`).
3. Create a translation file for the locale.
4. Observe that the translation won't work if you use the string with region.

**Related code:**

https://codesandbox.io/s/ecstatic-clarke-ksfqx?file=/src/i18nProvider.tsx:354-374

**Other information:**

Reporting this problem because when your application deals with regions from the same language ("pt-PT" and "pt-BR"), this function cannot be used.

**Suggestions:**
- Removing the split from [TranslationUtils.ts](https://github.com/marmelab/react-admin/blob/b71d7ddb3d7ad3379ed97819ac1cb4abe38ca74e/packages/ra-core/src/i18n/TranslationUtils.ts#L39), but this can cause many bugs across applications.
- Inserting an option to allow full string from locale to be return at [TranslationUtils.ts](https://github.com/marmelab/react-admin/blob/b71d7ddb3d7ad3379ed97819ac1cb4abe38ca74e/packages/ra-core/src/i18n/TranslationUtils.ts#L39)

```js
export const resolveBrowserLocale = (defaultLocale = DEFAULT_LOCALE, allowFullLocale = false) => {
    // Rely on the window.navigator object to determine user locale
    const {
        language,
        browserLanguage,
        userLanguage,
    } = window.navigator as AllNavigatorLanguage;
    const language = (language || browserLanguage || userLanguage || defaultLocale);
    return allowFullLocale ? language : language.split('-')[0];
```

**Environment**

* React-admin version: 4.0.0-alpha.2 (paired with master)
* React version: `^17.0.0`
* Browser: Firefox 96.0 (64-bits) Mozilla Firefox for Ubuntu canonical - 1.0

FernandoKGA:
So... I've been messing around with the function and we can improve it in two ways:

1. Adding an another default parameter, not breaking the function interface, and not becoming a breaking change.
[...]

PR #7194, opened 2022-02-05
Fix resolveBrowserLocale tests

FernandoKGA:
Noticed a bug at `resolveBrowserLocale` tests during #7170 investigation to implement a new feature (#7193).

**Problem**:
As `navigator.language` is a **read-only property** (see [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) about it), it wasn't being affected by the `window.navigator = {}`. You can see that at `should return default locale if there's no available locale in browser` test, it is an invalid test.

Captura de tela em 2022-02-05 18-09-45

**Solution**:
Implement correctly using `jest.spyOn` function, see this answer: https://stackoverflow.com/a/52870312
Captura de tela em 2022-02-05 17-35-07

Should this fix enter before #7193 @WiXSL ?

### DIFF
--- a/packages/ra-core/src/i18n/TranslationUtils.spec.ts
+++ b/packages/ra-core/src/i18n/TranslationUtils.spec.ts
@@ -7,20 +7,22 @@ import {
 
 describe('TranslationUtils', () => {
     describe('resolveBrowserLocale', () => {
+        let languageGetter;
         beforeEach(() => {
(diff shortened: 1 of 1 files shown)


WiXSL:
Thank you for taking the time to work on this.

djhi (approved):

djhi:
Thanks!

PR #7199, opened 2022-02-06
[4.0.0-alpha.3] Fix missing inputs event handlers

WiXSL:


packages/ra-ui-materialui/src/input/BooleanInput.tsx
```diff
@@ -40,6 +40,7 @@ export const BooleanInput = (props: BooleanInputProps) => {
         parse,
         resource,
         source,
+        onChange,
```
djhi, line 43:
```suggestion
        onBlur,
        onChange,
```

djhi, line 43:
```suggestion
        onBlur,
        onChange,
```

PR #7203, opened 2022-02-08
Avoid cloning elements in Authenticated

djhi:


packages/ra-core/src/auth/Authenticated.tsx
```diff
@@ -36,17 +30,15 @@ export interface AuthenticatedProps {
...
-        location, // kept for backwards compatibility, unused
-        ...rest
-    } = props;
+export const Authenticated = (props: AuthenticatedProps) => {
+    const { authParams, children } = props;
     useAuthenticated({ params: authParams });
     // render the child even though the useAuthenticated() call isn't finished (optimistic rendering)
     // the above hook will log out if the authProvider doesn't validate that the user is authenticated
-    return cloneElement(children, rest);
+    return children;
```
fzaninotto, line 38: i'd prefer that you return `<>{children}</>`, this makes TypeScript happier. 

also, please type `children` as `ReactNode` instead of `ReactElement` to allow multiple children, e.g. 

```jsx
const Foo = () => (
  <Authenticated>
    <Bar />
    <Baz />
  </Authenticated>
);
```

fzaninotto (changes requested):

PR #7205, opened 2022-02-08
Avoid cloning elements buttons

djhi:


packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -112,24 +105,11 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
-            ) : (
-                cloneElement(icon, {
-                    className: classnames(
-                        SaveButtonClasses.leftIcon,
-                        SaveButtonClasses.icon
-                    ),
-                })
-            )}
-            {displayedLabel}
+            {saving ? <CircularProgress size={18} thickness={2} /> : icon}
```
WiXSL, line 112: Aren't we missing `className={SaveButtonClasses.leftIcon}` here?

djhi: Good catch

djhi: Actually not an issue as the margin is applied on the button label

WiXSL (approved):

fzaninotto (approved):

PR #7207, opened 2022-02-08
Avoid cloning elements in ArrayField and ReferenceManyField views

djhi:


fzaninotto (approved):

fzaninotto:
Do we still need ReferenceManyFieldView? It could be moved back into ReferenceManyField.

PR #7209, opened 2022-02-08
Restore previous SaveButton to Avoid Button Responsive Behavior

djhi:


fzaninotto:
Since tests don't pass, why don't you use https://mui.com/components/buttons/#loading-button?

djhi:
Because it's in the labs

fzaninotto (approved):

PR #7213, opened 2022-02-09
Fix pagination state when perPage changes

### LINKED ISSUE #7204: ReferenceManyField pagination defect
**What you were expecting:**
Pagination query of ReferenceManyField is similar to List

**What happened instead:**
No results found

**Steps to reproduce:**
A bug was founded using pagination in ReferenceManyField data grid related to the page statement and the count of rows selected from perPage drop down.
To reproduce the bug we should have a total number of records more then minimal displayed records per page. Then select the last page from the pagination toolbar and increase count of records per page. "No results found" message will be displayed.
For example, we have 26 records in data grid and 25 of them are displaying on the first page, navigate to second page and increase a count of records up to 50. The query will be created with parameters: "page: 2, perPage: 50" as a result, there are no records to show due to this request.

**Related code:**
```
    <ReferenceManyField
      reference={listResource}
      target={target}
      filter={{ some filter }}
      sort={{ field: "isEnabled", order: "DESC" }}
      pagination={<Pagination rowsPerPageOptions={[25, 50, 100]} />}
      {...props}
    >
      <Datagrid
        expand={<Component />}
        isRowExpandable={boolean}
      >
     Some fields....
      </Datagrid>
    </ReferenceManyField>
```

**Environment**

* "react-admin": "3.17.3"
* "react": "17.0.1"
* Version 96.0.4664.110 (Official Build) (64-bit)

WiXSL:
Fixes #7204

### DIFF
--- a/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
@@ -794,7 +794,7 @@ describe('<ReferenceArrayInputController />', () => {
         });
     });
 
-    it('should props compatible with the ListContext', async () => {
+    it('should props be compatible with the ListContext', async () => {
         const children = ({
             setPage,
             setPerPage,
@@ -869,7 +869,7 @@ describe('<ReferenceArrayInputController />', () => {
                 },
                 payload: {
                     pagination: {
-                        page: 2,
+                        page: 1,
                         perPage: 50,
                     },
                     sort: {
--- a/packages/ra-core/src/controller/usePaginationState.spec.ts
+++ b/packages/ra-core/src/controller/usePaginationState.spec.ts
@@ -77,4 +77,38 @@ describe('usePaginationState', () => {
             perPage: 100,
         });
     });
+
+    it('should reset the current page to 1 when perPage state changes', () => {
+        const { hookValue, childrenMock } = renderHook(() =>
+            usePaginationState()
+        );
+        expect(hookValue.pagination).toEqual({ page: 1, perPage: 25 });
+
+        act(() => hookValue.setPerPage(100));
+
+        expect(childrenMock).toBeCalledTimes(2);
+
+        act(() => hookValue.setPage(2));
+
+        expect(childrenMock.mock.calls[1][0].pagination).toEqual({
+            page: 1,
+            perPage: 100,
+        });
+    });
+
+    it('should reset the current page to initial page when perPage state changes', () => {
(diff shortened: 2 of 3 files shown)


packages/ra-core/src/controller/usePaginationState.ts
```diff
@@ -53,7 +53,11 @@ export default (
     });
     const isFirstRender = useRef(true);
 
-    const setPerPage = useCallback(perPage => setPagination({ perPage }), []);
+    const setPerPage = useCallback(
+        perPage =>
+            setPagination({ perPage, page: initialPagination.page || 1 }),
```
fzaninotto, line 58: Why would you want to set the page to `initialPagination` on `perPage` change? I fail to see the rationale.

packages/ra-core/src/controller/usePaginationState.spec.ts
```diff
@@ -77,4 +77,38 @@ describe('usePaginationState', () => {
...
+
+        act(() => hookValue.setPage(2));
+
+        expect(childrenMock.mock.calls[1][0].pagination).toEqual({
+            page: 1,
+            perPage: 100,
+        });
+    });
+
+    it('should reset the current page to initial page when perPage state changes', () => {
```
fzaninotto, line 99: I think it shouldn't

packages/ra-core/src/controller/usePaginationState.spec.ts
```diff
@@ -77,4 +77,38 @@ describe('usePaginationState', () => {
...
         });
     });
+
+    it('should reset the current page to 1 when perPage state changes', () => {
+        const { hookValue, childrenMock } = renderHook(() =>
+            usePaginationState()
+        );
+        expect(hookValue.pagination).toEqual({ page: 1, perPage: 25 });
+
+        act(() => hookValue.setPerPage(100));
```
fzaninotto, line 87: I don't understand this test. You should have a state where the page isn't 1, then change the perPage, then see that the state has a page of 1. You're doing something completely different here.

packages/ra-core/src/controller/input/ReferenceArrayInputController.spec.tsx
```diff
@@ -794,7 +794,7 @@ describe('<ReferenceArrayInputController />', () => {
         });
     });
 
-    it('should props compatible with the ListContext', async () => {
+    it('should props be compatible with the ListContext', async () => {
```
fzaninotto, line 797: thios sentence isn't English

fzaninotto (changes requested):

fzaninotto (approved):

PR #7222, opened 2022-02-10
[4.0.0-alpha.3] Replace classnames library dependency with clsx

WiXSL:


yarn.lock
```diff
@@ -10447,7 +10438,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"classnames@npm:2.3.1":
+"classnames@npm:^2.2.5":
```
djhi, line 10441: For information, `classnames` is still there because the demo needs `recharts` which uses it. It won't be installed by react-admin users unless they use a library depending on it

WiXSL: Yep, I so it

fzaninotto (approved):

Issue #5879, opened 2021-02-06
[typescript] Make DataProvider generic over resource names


fbessou:
**Is your feature request related to a problem? Please describe.**
I would like the compiler to complain when I call a dataProvider with a wrong resource name.

**Describe the solution you'd like**
Make DataProvider generic over the available resource names which constrains the first argument of all provider's methods:

```typescript
const resources = ['User', 'Book', 'City'] as const;
const dataProvider : DataProvider<typeof resources[number]>;
const users = await dataProvider.getList("User", params).data;
const towns = await dataProvider.getList("Town", params).data; // 2345: Argument of type '"Town"' is not assignable to parameter of type '"User" | "Book" | "City"
```

The type parameter could default to `string` to avoid introducing a breaking change :

```ts
export type DataProvider<ResourceTypes extends string = string> = {
  getList: <RecordType extends Record = Record>(
    resource: ResourceTypes,
    params: GetListParams
  ) => Promise<GetListResult<RecordType>>;
  [...]
}

```

**Describe alternatives you've considered**
My workaround is currently to make a data provider which doesn't respect the DataProvider interface :

```ts
function createDataProvider<ResourceNames>() {
   return {
    getList(resource: ResourceNames, params) {
      // implementations details
    }
  }
}
```
But it has two problems:
* I don't benefit from type checking against the DataProvider interface when writing my custom data provider (I don't get an error if I make a typo in a method name for example)
* when passing the provider as props to the `Admin` component, I have to cast it to DataProvider since my custom methods are stricter than the  ones provided by DataProvider : `<Admin dataProvider={dataProvider as DataProvider}>...`

fzaninotto:
Fixed by #7225

PR #7225, opened 2022-02-10
[4.0.0-alpha.3] Make DataProvider type generic over resource

WiXSL:
Implements #5879


djhi (changes requested):
I'd love to see the usage in the examples

PR #7229, opened 2022-02-11
[4.0.0-alpha.3] Fix SaveButton disable value based on form state

WiXSL:
SaveButton in simple project's forms are always enabled.


djhi (commented):
We're going to fix that in the buttons themselves as we want to avoid cloning

PR #7234, opened 2022-02-14
Fix BC break v4: useCreate and useUpdate can return a promise

alanpoulain:
Related PR: https://github.com/marmelab/react-admin/pull/5778.

To be able to handle correctly submission validation errors, the returned functions from `useCreate` and `useUpdate` need to be able to return a promise.
Only then doing like it is described in the documentation would be possible:
https://github.com/marmelab/react-admin/blob/efc89692fdf839d1267e030595a52a1127b23d71/docs/CreateEdit.md?plain=1#L1524-L1531


fzaninotto (approved):

fzaninotto:
Thanks a lot!

PR #7242, opened 2022-02-15
Fix DateField show wrong date on negative time zones

### LINKED ISSUE #5578: DateField/DateInput off by one w/ optimistic rendering
**What you were expecting:**
When PUT/POST (update/create) with DateInput, optimistic rendering renders the date as the same date inputted

**What happened instead:**
Optimistic Rendering rendered the date off by one (EX: 4/5/2020 -> 4/4/2020) which results in the check to crash and error to pop up

**Steps to reproduce:**
Use a DateInput and save from an edit/create form a date and look at how optimistic rendering renders the date.

**Related code:**
```
export const BlankEdit = props => (
    <Edit {...props}}>
        <SimpleForm warnWhenUnsavedChanges redirect="show">        
			<DateInput source="date" locales="en-US"/>
		</SimpleForm>
    </Edit>
);
```

**Other information:**

**Environment**

* React-admin version: 3.10.1
* React version: 16.13.1
* Browser: Firefox

fzaninotto:
Closes #5578

### DIFF
--- a/packages/ra-ui-materialui/src/field/DateField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/DateField.spec.tsx
@@ -31,6 +31,19 @@ describe('<DateField />', () => {
         expect(queryByText(date)).not.toBeNull();
     });
 
+    it('should render a date string', () => {
+        const { queryByText } = render(
+            <DateField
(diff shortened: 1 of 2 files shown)


WiXSL (commented):
I tried it in simple project, and now it shows the date correctly

PR #7248, opened 2022-02-16
Fix prevent `CheckboxGroupInput` from changing selected values type

### LINKED ISSUE #6516: CheckboxGroupInput overrides the type of the selected value
**What you were expecting:**
As a backend I use a GraphQLServer. in it, the values of all fields are strongly typed. The type specified for the resource identifier is String.

For selection inside the `CheckboxGroupInput`, the backend returned one of options with a record `{id: "54367564", ///...}`. When I chose this option and sent the mutation to the server, it gave me a type error.

It turned out that instead of a string value, the input converted the identifier to a number.

**What happened instead:**
Expected `CheckboxGroupInput `to leave the identifier unchanged

**Steps to reproduce:**
1) Set resource id as a string with numbers `{id: "54367564", ///...}`
2) Try to save your changes
3) Observe `{ids: [54367564, ...]}` entry as the value in the dataprovider

**Related code:**
here is a piece of code `CheckboxGroupInput` that does it.
In my use case, I don't understand why it is needed

```
 const handleCheck = useCallback(
        (event, isChecked) => {
            let newValue;
            try {
                // try to convert string value to number, e.g. '123'
                newValue = JSON.parse(event.target.value);
            } catch (e) {
                // impossible to convert value, e.g. 'abc'
                newValue = event.target.value;
            }
            if (isChecked) {
                finalFormOnChange([...(value || []), ...[newValue]]);
            } else {
                finalFormOnChange(value.filter(v => v != newValue)); // eslint-disable-line eqeqeq
            }
            finalFormOnBlur(); // HACK: See https://github.com/final-form/react-final-form/issues/365#issuecomment-515045503
        },
        [finalFormOnChange, finalFormOnBlur, value]
    );
```

* React-admin version: 3.17.0
* React version:
* Browser:

WiXSL:
Fixes #6516

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
@@ -138,13 +138,19 @@ const CheckboxGroupInput: FunctionComponent<CheckboxGroupInputProps> = props => 
     const handleCheck = useCallback(
         (event, isChecked) => {
             let newValue;
-            try {
-                // try to convert string value to number, e.g. '123'
-                newValue = JSON.parse(event.target.value);
-            } catch (e) {
-                // impossible to convert value, e.g. 'abc'
+
+            if (value.every(item => typeof item !== 'number')) {
                 newValue = event.target.value;
+            } else {
+                try {
+                    // try to convert string value to number, e.g. '123'
+                    newValue = JSON.parse(event.target.value);
+                } catch (e) {
+                    // impossible to convert value, e.g. 'abc'
+                    newValue = event.target.value;
+                }
             }
+
             if (isChecked) {
                 finalFormOnChange([...(value || []), ...[newValue]]);
             } else {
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
```diff
@@ -138,13 +138,19 @@ const CheckboxGroupInput: FunctionComponent<CheckboxGroupInputProps> = props =>
     const handleCheck = useCallback(
         (event, isChecked) => {
             let newValue;
-            try {
-                // try to convert string value to number, e.g. '123'
-                newValue = JSON.parse(event.target.value);
-            } catch (e) {
-                // impossible to convert value, e.g. 'abc'
+
+            if (value.every(item => typeof item !== 'number')) {
```
fzaninotto, line 142: no, you should iterate on choices, not value (which isn't an array by the way).

Also, the condition ('every item is not of tyype number') is harder to read than the opposite ('every item is a number'), so prefer the latter.

fzaninotto (changes requested):

fzaninotto (approved):

PR #7259, opened 2022-02-17
Fix Form save handler return type

WiXSL:


fzaninotto:
I'm not sure about this one, can you elaborate ?

fzaninotto:
TypoeScript doesn't seem to like your types, though

PR #7260, opened 2022-02-17
Fix bad imports

WiXSL:
I don't know if this is used


packages/ra-core/src/dataProvider/getFetchType.ts
```diff
@@ -8,7 +8,7 @@ import {
     UPDATE_MANY,
     DELETE,
     DELETE_MANY,
-} from '../core';
+} from '.';
```
fzaninotto, line 11: please import from './dataFetchActions' instead (otherwise it's weird).

fzaninotto (changes requested):

PR #7261, opened 2022-02-17
Fix <Admin theme> needs createTheme

fzaninotto:


docs/Theming.md
```diff
@@ -251,16 +251,14 @@ export const PostList = () => {
 
 ## Using a Predefined Theme
 
-Material UI also supports [complete theming](https://material-ui.com/customization/themes) out of the box. Material UI ships two base themes: light and dark. React-admin uses the light one by default. To use the dark one, pass it to the `<Admin>` component, in the `theme` prop (along with `createTheme()`).
+Material UI also supports [complete theming](https://material-ui.com/customization/themes) out of the box. Material UI ships two base themes: light and dark. React-admin uses the light one by default. To use the dark one, pass it to the `<Admin>` component, in the `theme` prop.
```
djhi, line 254: We probably should mention that using `createTheme` is fine too

examples/demo/src/App.tsx
```diff
@@ -37,7 +37,7 @@ const App = () => {
             )}
             authProvider={authProvider}
             dashboard={Dashboard}
-            loginPage={Login}
+            loginPage={<Login />}
```
djhi, line 40: This is weird

fzaninotto: It does work, though. I removed it so your eyes don't bleed.

docs/Theming.md
fzaninotto: I think it would be confusing for newcomers.

packages/ra-core/src/types.ts
```diff
@@ -278,10 +278,7 @@ export type AdminChildren = RenderResourcesFunction | ReactNode;
 export type TitleComponent = string | ReactElement<any>;
 export type CatchAllComponent = ComponentType<{ title?: TitleComponent }>;
 
-export interface LoginComponentProps {
-    title?: TitleComponent;
-}
-export type LoginComponent = ComponentType<LoginComponentProps>;
+export type LoginComponent = ComponentType<{}> | ReactElement<any>;
 export type DashboardComponent = ComponentType<WithPermissionsChildrenParams>;
```
djhi, line 282: Shouldn't we do the same for the Dashboard ?

fzaninotto: yes, in due time

PR #7270, opened 2022-02-18
Update datagrid style overrides [RFR]

smeng9:
The datagrid override seems is forgotten in the first beta version.

This pull request also adds a table wrapper, so we can set a width for table and make a scroll bar appear right below the table.

(Previously in examples, the scrollbar always appears on the bottom of the page, when scrolling horizontally datagrid may cover the menu)

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -679,6 +679,7 @@ The `<Datagrid>` component accepts the usual `className` prop. You can also over
 | Rule name                      | Description                                      |
 | ------------------------------ | ------------------------------------------------ |
 | `& .RaDatagrid-table`          | Applied to the table element                     |
+| `& .RaDatagrid-tableWrapper`   | Applied to the div that wraps table element      |
 | `& .RaDatagrid-thead`          | Applied to the table header                      |
 | `& .RaDatagrid-tbody`          | Applied to the table body                        |
 | `& .RaDatagrid-headerCell`     | Applied to each header cell                      |
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -678,8 +678,8 @@ The `<Datagrid>` component accepts the usual `className` prop. You can also over
 
 | Rule name                      | Description                                      |
 | ------------------------------ | ------------------------------------------------ |
-| `& .RaDatagrid-table`          | Applied to the table element                     |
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks for the overrideResolver, which was indeed forgotten.

I don't uderstand the need for a wrapper though:

1- you can add it to your markup directly

```jsx
const PostList = () => (
  <List>
     <Wrapper>
       <Datagrid>
           //..
       </Datagrid>
     </Wrapper>
  </List>
);

2- you can normally fix the height of the table through the `<Datagrid sx>` prop

**Comment by smeng9** [comment]
Hi @fzaninotto Thanks for your reply. I will explain why we need a wrapper for table.

We are trying to achieve the following css effect by setting the datagrid `{{width: "100%", overflowX: "auto"}}` so it will not overflow parent container and have the table scrollable horizontally. See the visualization attached.
scrollbar

We had thought about your two methods.
For the first method
```
const PostList = () => (
  <List>
     <div style={{width: "100%", overflowX: "auto"}}>
       <Datagrid>
           //..
       </Datagrid>
     </div>
  </List>
);
```
This will break the bulkactions because `overflowx`=auto and `overflowy` will interfere with each other (This is a CSS "feature", See: https://www.brunildo.org/test/Overflowxy2.html) So it will make `overflowy` hidden as well. The way we added bulkactions is setting a negative transform to it, but it is considered overflow, so it will not be visible. See the attached example.
bulkaction

For the second method, adding sx property to datagrid also does not work well.
The table will not respect width: "100%" rule after overflow because it is a table, not a div. It will only have 100% when there are not too many columns. Forcing it to `display: block` also breaks the layout when there is not enough columns. That is the reason we need a wrapper div just for the table to make the table horizontally scrollable. See attached screenshot for problem.
table100


This CSS problem does not appear in v3 because bulkaction was not part of the datagrid in v3, but now it seems gets moved into the datagrid in v4.

smeng9:
Hi @fzaninotto any further questions on moving this PR forward?

fzaninotto (changes requested):
Thanks. You also need to add the doc for this change. Could you update the table of available CSS class names in Datagrid.md?

smeng9:
documentation is done

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7272, opened 2022-02-18
Fix AutocompleteInput and ReferenceArrayInput stories

WiXSL:


packages/ra-ui-materialui/src/input/AutocompleteArrayInput.stories.tsx
```diff
@@ -399,20 +399,8 @@ const BookEditWithReferenceAndCreationSupport = () => (
         }}
     >
         <SimpleForm>
-            <ReferenceArrayInput reference="authors" source="author" fullWidth>
-                <AutocompleteArrayInput
-                    create={<CreateAuthor />}
-                    options={{
```
WiXSL, line 405: I remove the code syntax error, but I don't know if the Idea was to render custom options here. Seems out of the scope of this story

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -403,20 +403,7 @@ const BookEditWithReferenceAndCreationSupport = () => (
     >
         <SimpleForm>
             <ReferenceInput reference="authors" source="author">
-                <AutocompleteInput
-                    create={<CreateAuthor />}
-                    options={{
```
WiXSL, line 404: I remove the code syntax error, but I don't know if the Idea was to render custom options here. Seems out of the scope of this story

packages/ra-ui-materialui/src/input/ReferenceArrayInput.stories.tsx
```diff
@@ -25,24 +27,24 @@ const dataProvider = testDataProvider({
...
+        <Form
+            onSubmit={() => {}}
+            defaultValues={{ tag_ids: [5] }}
+            render={() => (
+                <ReferenceArrayInput
+                    reference="tags"
+                    resource="posts"
+                    source="tag_ids"
+                >
+                    <DatagridInput rowClick="toggleSelection" sx={{ mt: 6 }}>
```
WiXSL, line 43: Action buttons where hidden from view

djhi: The DatagridInput is very WIP

djhi (approved):

PR #7283, opened 2022-02-21
Fix missing `ReferenceField` children's link color

### LINKED ISSUE #7278: [v4 β1] ChipField text color regression when used as a child of ReferenceField
**What you were expecting:**
The text color should be default anchor link color

**What happened instead:**
The text color is black, so it does not respect anchor link color even though it is a link.

**Steps to reproduce:**
See codesandbox below.

Just replace some TextField in ReferenceField with ChipField.

**Related code:**

```
[3.19.10](https://codesandbox.io/s/jolly-hopper-egf7x8?file=/src/invoices/InvoiceList.tsx:1538-1581)
[4.0.0-beta.1.0](https://codesandbox.io/s/nostalgic-frog-ct4r1f?file=/src/posts.js:689-725)
```

**Other information:**

chipfield
color2

Suggest: to make color inherit from parent. Still not sure why in v3 even without inheriting the link text color it works out of the box.

**Environment**

* React-admin version: 4.0.0-beta.1.0
* Last version that did not exhibit the issue (if applicable): 3.19.10
* React version: 18.0.0-rc.0
* Browser: Edge 98
* Stack trace (in case of a JS error): N/A for css regression

WiXSL:
Fixes #7278

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceField.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceField.tsx
@@ -256,7 +256,7 @@ const Root = styled('span', {
     overridesResolver: (props, styles) => styles.root,
 })(({ theme }) => ({
     [`& .${ReferenceFieldClasses.link}`]: {
-        '& > :first-child': {
+        '& > *': {
             color: theme.palette.primary.main,
         },
     },
--- a/packages/ra-ui-materialui/src/field/ReferenceField.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceField.tsx
@@ -255,5 +255,9 @@ const Root = styled('span', {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
 })(({ theme }) => ({
-    [`& .${ReferenceFieldClasses.link}`]: {},
+    [`& .${ReferenceFieldClasses.link}`]: {
+        '& :first-child': {
+            color: theme.palette.primary.main,
+        },
+    },
 }));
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -255,5 +255,9 @@ const Root = styled('span', {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
 })(({ theme }) => ({
-    [`& .${ReferenceFieldClasses.link}`]: {},
+    [`& .${ReferenceFieldClasses.link}`]: {
+        '& :first-child': {
```
fzaninotto, line 259: I'd use the child selector (`> *`) instead to select all children if there are more than one.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -255,5 +255,9 @@ const Root = styled('span', {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
 })(({ theme }) => ({
-    [`& .${ReferenceFieldClasses.link}`]: {},
+    [`& .${ReferenceFieldClasses.link}`]: {
+        '& > :first-child': {
```
fzaninotto, line 259: I really meant `'& > *'`

fzaninotto (approved):

Issue #7284, opened 2022-02-21
[v4 β1] useConfigureAdminRouterFromChildren does not allow API Platform Admin ResourceGuesser component


alanpoulain:
CoreAdminRoutes component is using the [useConfigureAdminRouterFromChildren](https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx) hook.
This hook filters the passed children by their types:
https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx#L241-L243

Since API Platform Admin is using its [own special Resource component](https://github.com/api-platform/admin/blob/main/src/ResourceGuesser.tsx), it seems impossible to make API Platform Admin compatible with React-admin v4.

Would it be possible to provide an alternative way to filter the children? Maybe by their props? I think it will also improve the compatibility of this piece of code (comparing an element type is known to create some issues with some loader/transpiler/compiler).

fzaninotto:
Agreed, this is a regression - I'm marking it as a bug. 

We need to detect child type based on a custom property, like material-ui does with `muiName` (cf: https://github.com/mui/material-ui/blob/master/packages/mui-utils/src/isMuiElement.ts).

A PR is welcome.

PR #7288, opened 2022-02-22
fix: allow composition of child elements by using raName

alanpoulain:
Fixes https://github.com/marmelab/react-admin/issues/7284.

Like [mui does](https://mui.com/guides/composition/#wrapping-components), introduce a `raName` static on some child elements (`CustomRoutes` and `Resource`) in order to filter them in `useConfigureAdminRouterFromChildren`.

It will allow to compose components (for instance API Platform Admin [ResourceGuesser component](https://github.com/api-platform/admin/blob/main/src/ResourceGuesser.tsx)) by setting the same static property.

### DIFF
--- a/packages/ra-core/src/core/CustomRoutes.tsx
+++ b/packages/ra-core/src/core/CustomRoutes.tsx
@@ -11,6 +11,8 @@ export const CustomRoutes = (props: CustomRoutesProps) => {
     return null;
 };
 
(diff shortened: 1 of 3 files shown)


dunglas (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7297, opened 2022-02-24
Fix move common props from ReferenceInputs to its children in demos

WiXSL:


examples/crm/src/deals/DealCreate.tsx
```diff
@@ -20,6 +20,7 @@ import { Deal } from '../types';
 export const DealCreate = ({ open }: { open: boolean }) => {
     const redirect = useRedirect();
     const dataProvider = useDataProvider();
+    const validateRequired = required();
```
djhi, line 23: Why? If you want to extract, you should move it outside the component body as well imo :)

djhi (changes requested):

examples/crm/src/deals/DealCreate.tsx
WiXSL: You're right, that was the idea

djhi (approved):

fzaninotto (approved):

PR #7301, opened 2022-02-24
Fix Wrong Default Value Handling in FilterForm

### LINKED ISSUE #7191: v4 a2 Regression on alwaysOn filters with empty string
**What you were expecting:**

The list is supposed to render there, it should have the same result as no filter is applied
Screenshot from 2022-02-04 17-53-02

**What happened instead:**

All the list get filtered out, even if the filter is not set
Screenshot from 2022-02-04 17-52-26

**Steps to reproduce:**
1. Fix the dependency react-hook-form of example/totorial in next branch
2. Add `<TextInput source="title" label="Title" alwaysOn />` to postFilters
3. Make a clean render of the page (make sure no encoded filters in url)
4. Go to posts collection through the menu
5. The page will say no results found

This is supposed to be a bug because if a filter input is empty as the initial state, it should not be included in the filter.
Otherwise everything will be filtered out.

However, after type in something and the remove everything typed in the TextInput filter. The list view seems to work correctly.
This is a user experience regression.

**Related code:**

```
https://codesandbox.io/s/amazing-http-0ibjn?file=/src/posts.js:300-355
```

**Other information:**

**Environment**

* Last version that did not exhibit the issue (if applicable): 3.19.7
* React version: 17.0.2
* Browser: FireFox 96

djhi:
Fixes #7191

Also remove the `margin` and `variant` props injection

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1554,7 +1554,7 @@ const MyMenuItem = forwardRef((props, ref) => {
 });
 ```
 
-## No More Props Injection In `<SimpleForm>` and `<TabbedForm>`
+## No More Props Injection In `<SimpleForm>`, `<TabbedForm>` and `<FilterForm>`
 
 The form components used to clone their children, inspect their props to handle labels and inject the `margin` and `variant` props.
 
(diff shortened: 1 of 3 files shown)


fzaninotto:
Is it possible to add a unit test, to avoid that regression in the future?

PR #7321, opened 2022-02-28
Fix change ShowGuesser for CommentShow

WiXSL:


djhi (approved):

fzaninotto:
The ShowGuesser was actually used on purpose here (otherwise, we don't have any easy way to test the guessers). I think you should revert this one.

PR #7350, opened 2022-03-06
Avoid cloning the toolbar in TabbedForm

### LINKED ISSUE #7322: `Edit` component doesn't pass down `mutationMode` prop
Modifying simple project's PostEdit like this:
```diff
- <Edit title={<PostTitle />} actions={<EditActions />}>
+ <Edit mutationMode="pessimistic" title={<PostTitle />} actions={<EditActions />}>
```
Doesn't affect the behavior of the default `DeleteButton`

WiXSL:
Fixes #7322

- [X] Avoid cloning the toolbar inside `TabbedForm` component
- [X] Updated Upgrade Guide to instruct how to use orphan toolbar props in forms layouts
- [X] Add `mutationMode` prop to Edit, Create and Save contexts
- [X] Enforce use of contexts in `DeleteButton` (which makes it easier to override the container Toolbar)
- [X] Remove `submitOnEnter` prop
- [x] Improve documentation of the usage of custom side effect in forms


UPGRADE.md
```diff
@@ -1860,13 +1860,20 @@ const {
 
 Or in a component:
 ```diff
-export const PostEdit = (props) => (
```
WiXSL, line 1863: I change to a different example here to avoid confusion

WiXSL:
Problem:
`mutationMode` is used by `useEditController`, so is valid to pass it as a prop of an `Edit`/`EditBase` but it won't get to the toolbar's `DeleteButton`

UPGRADE.md
```diff
@@ -2666,6 +2673,24 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
...
 
+Since the `<Toolbar>` component is not cloned anymore by react-admin layouts, properties like `mutationMode` and `submitOnEnter` are not passed down from parent components. So, if you rely on this behaviour, you should create your own custom toolbar instead
+
+```diff
+import { Toolbar, SimpleForm, Edit, TextInput } from 'react-admin';
+
++const MyToolbar = props => <Toolbar {...props} mutationMode="pessimistic" submitOnEnter={false} />;
+
+export const PostEdit = (props) => (
+-    <Edit {...props} mutationMode="pessimistic">
```
WiXSL, line 2684: `mutationMode` is used by `useEditController`, so is valid to pass it as a prop of an `Edit`/`EditBase` but it won't get to the toolbar's `DeleteButton`.

UPGRADE.md
```diff
@@ -2666,6 +2666,23 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
 };
 ```
 
+Since the `<Toolbar>` component is not cloned anymore by react-admin layouts, properties like `mutationMode` and `submitOnEnter` are not passed down from parent components. So, if you rely on this behaviour, you should create your own custom toolbar instead
+
+```diff
+import { Toolbar, SimpleForm, Edit, TextInput } from 'react-admin';
+
++const MyToolbar = props => <Toolbar {...props} mutationMode="pessimistic" submitOnEnter={false} />;
```
djhi, line 2674: Actually, I think the Toolbar shouldn't clone and inject props either

djhi (changes requested):

fzaninotto:
Wait, does that mean that the DeleteButton in the SimpleForm doesn't repsect the `mutationMode` anymore? If so, that's a serious regression .

WiXSL:
> Wait, does that mean that the DeleteButton in the SimpleForm doesn't repsect the `mutationMode` anymore? If so, that's a serious regression .

This already happens in `SimpleForm` #7322

UPGRADE.md
WiXSL: So, the solution should be to pass a custom SaveButton/DeleteButton inside a custom Toolbar?

fzaninotto:
We must fix the underlying issue of mutationMode not being applied before releasing 4.0. What if we added it to the `EditContext`?

WiXSL:
> We must fix the underlying issue of mutationMode not being applied before releasing 4.0. What if we added it to the `EditContext`?

Could be. I don't see any problems with it.

WiXSL:
Now that I think about it, We can't use the EditContext inside the Toolbar, you don't know if you are creating or editing

djhi:
I think it makes sense to add that to the `SaveContext`

fzaninotto:
So let's add `mutationMode` to the EditContext and SaveContext. For the create view, it will default to "pessimistic", and will not be used (because there is no DeleteButton in the toolbar of Creation forms).

WiXSL:
OK, and what about `submitOnEnter`? Is the same problem

djhi:
I think we shouldn't care about `submitOnEnter`. Our forms shouldn't accept nor pass down this prop

WiXSL:
Is now ready for review

WiXSL:
Is now ready for review

docs/CreateEdit.md
```diff
@@ -1550,12 +1550,19 @@ export const UserCreate = () => {
 
 ## Submit On Enter
 
-By default, pressing `ENTER` in any of the form fields submits the form - this is the expected behavior in most cases. However, some of your custom input components (e.g. Google Maps widget) may have special handlers for the `ENTER` key. In that case, to disable the automated form submission on enter, set the `submitOnEnter` prop of the form component to `false`:
+By default, pressing `ENTER` in any of the form fields submits the form - this is the expected behavior in most cases. However, some of your custom input components (e.g. Google Maps widget) may have special handlers for the `ENTER` key. In that case, to disable the automated form submission on enter, set the `submitOnEnter` prop of the `SaveButton` component to `false`:
```
djhi, line 1553: I'm not sure this is good advice. Custom inputs that have Enter key handlers should stop the event propagation. This would allow other inputs to still submit the form on Enter

packages/ra-core/src/controller/create/CreateContextProvider.tsx
```diff
@@ -38,7 +38,12 @@ export const CreateContextProvider = ({
     value: CreateControllerResult;
 }) => (
     <CreateContext.Provider value={value}>
-        <SaveContextProvider value={usePickSaveContext(value)}>
+        <SaveContextProvider
+            value={{
+                ...usePickSaveContext(value),
+                mutationMode: 'pessimistic',
```
djhi, line 44: Why don't you get the mutationMode inside `usePickSaveContext`?

packages/ra-ui-materialui/src/form/Toolbar.tsx
```diff
@@ -65,19 +66,20 @@ export const Toolbar = <
         children,
         className,
         resource,
-        saving,
-        submitOnEnter = true,
-        mutationMode,
+        saving: saveingProp,
```
djhi, line 69:
```suggestion
        saving: savingProp,
```

packages/ra-ui-materialui/src/form/Toolbar.tsx
```diff
@@ -65,19 +66,20 @@ export const Toolbar = <
...
+        submitOnEnter: submitOnEnterProp,
+        mutationMode: mutationModeProp,
         ...rest
     } = props;
     const record = useRecordContext(props);
+    const saveContext = useSaveContext();
```
djhi, line 75: Don't forget to pass the props as they may override the context

djhi (changes requested):
Nice! There are some linter warnings left

packages/ra-core/src/controller/create/CreateContextProvider.tsx
WiXSL: Because I can't tell if I'm being called from a CreateContext ('pessimistic') or an EditContext ('value.mutationMode')

packages/ra-ui-materialui/src/form/Toolbar.tsx
WiXSL: No, `useSaveContext` doesn`t have this feature like `useRecordContext` have. Should I add it?

packages/ra-core/src/controller/create/CreateContextProvider.tsx
djhi: You can still do it in `usePickSaveContext` but override it for create like you already did

packages/ra-ui-materialui/src/form/Toolbar.tsx
djhi: Then the Toolbar doesn't need to accept the `saving` prop

UPGRADE.md
```diff
@@ -2666,6 +2666,28 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
...
++const MyToolbar = props => (
+    <Toolbar {...props}> 
+        <SaveButton submitOnEnter={false} />
+        <DeleteButton />
+    </Toolbar>
+);
+
+export const PostEdit = () => (
+    <Edit>
+-       <SimpleForm>
```
fzaninotto, line 2683:
```suggestion
-       <SimpleForm submitOnEnter>
```

UPGRADE.md
```diff
@@ -2666,6 +2666,28 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
...
+
+```diff
+import { Toolbar, SimpleForm, Edit, TextInput, SaveButton, DeleteButton } from 'react-admin';
+
++const MyToolbar = props => (
+    <Toolbar {...props}> 
+        <SaveButton submitOnEnter={false} />
+        <DeleteButton />
+    </Toolbar>
+);
```
fzaninotto, line 2679:
```suggestion
+const MyToolbar = props => (
+   <Toolbar {...props}> 
+       <SaveButton submitOnEnter={false} />
+       <DeleteButton />
+   </Toolbar>
+);
```

UPGRADE.md
```diff
@@ -2666,6 +2666,28 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
 };
 ```
 
+Since the `<Toolbar>` component is not cloned anymore by react-admin layouts, `submitOnEnter` property is not passed down to the `SaveButton` from parent components. So, if you rely on this behaviour, you should pass this prop yourself.
```
fzaninotto, line 2669: I think you should be more explicit and add a section titled 

> `<SimpleForm>` and `<TabbedForm>` don't support `submitOnEnter` anymore

and mention tat it also concerns Tabbedform

fzaninotto (changes requested):
Much, much better like that!

docs/CreateEdit.md
```diff
@@ -1550,18 +1550,44 @@ export const UserCreate = () => {
...
     <Edit>
-        <SimpleForm submitOnEnter={false}>
+        <SimpleForm toolbar={<MyToolbar/>}>
             ...
         </SimpleForm>
     </Edit>
 );
 ```
 
+However, some of your custom input components (e.g. Google Maps widget) may have special handlers for the `ENTER` key. In that case, you should stop the event  on those inputs. This would allow other inputs to still submit the form on Enter:
```
djhi, line 1572:
```suggestion
However, some of your custom input components (e.g. Google Maps widget) may have special handlers for the `ENTER` key. In that case, you should prevent the default handling of the event on those inputs. This would allow other inputs to still submit the form on Enter:
```

djhi (commented):
I now wonder what's the point of the `submitOnEnter` prop. Shouldn't the `SaveButton` simply pass the existing `type` prop (`submit` or `button`) down to the Mui button? In that case, we could set it to `submit` by default and check its value in the click handler

fzaninotto:
> I now wonder what's the point of the `submitOnEnter` prop. Shouldn't the `SaveButton` simply pass the existing `type` prop (`submit` or `button`) down to the Mui button?

You're totally right. Moving the prop from the form to the SaveButton made it obvious. We don't need an additional prop, `type` already does the job. We just need to properly document it.

@WiXSL Good for you?

WiXSL:
Sure, you are right!
I'll try to update this tonight

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -65,8 +67,14 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
     const saveContext = useSaveContext();
-    const hasSideEffects = !!mutationOptions || !!transform;
-    const type = !submitOnEnter || hasSideEffects ? 'button' : 'submit';
+
+    warning(
+        type === 'submit' &&
+            ((mutationOptions &&
+                (mutationOptions.onSuccess || mutationOptions.onError)) ||
+                transform),
+        'SaveButton of type "submit" recived mutationOptions. You need to pass type="button" to the SaveButton component or set mutationOptions in your main view'
```
fzaninotto, line 76:
```suggestion
        'Cannot use <SaveButton mutationOptions> props on a button of type "submit". To override the default mutation options on a particular save button, set the <SaveButton type="button"> prop, or set mutationOptions in the main view component (<Create> or <Edit>).'
```

fzaninotto (changes requested):
Much simpler! You still need to update the doc and the upgrade guide

UPGRADE.md
```diff
@@ -2659,21 +2659,29 @@ const ReviewEditToolbar = (props: ToolbarProps<Review>) => {
...
+### `submitOnEnter` Prop Has Been Removed
 
+The following components no longer accept this prop:
+
+- `<SimpleForm>`
+- `<TabbedForm>`
+- `<Toolbar>`
+- `<SaveButton>`
+
+To determine wheter `<SimpleForm>` and `<TabbedForm>` will submit if `Enter` key is pressed, you have to relay on the `<SaveButton>`'s `type` prop. Setting it to "button" will prevent submittion on `Enter`.
```
fzaninotto, line 2677:
```suggestion
By default, `<SimpleForm>` and `<TabbedForm>` submit when the user presses `Enter`. To disable this behavior, you must now turn the `<SaveButton>` (which renders as a `<input type="submit" />` by default) into an `<input type="button">` element, by setting the `type` prop to "button".

If you didn't have a custom form toolbar, you'll have to create one to set the `<SaveButton type="button" />` and prevent submission on enter. 
```

fzaninotto (changes requested):

fzaninotto (approved):

Issue #7326, opened 2022-03-01
make DataProvider an interface


antdking:
**Is your feature request related to a problem? Please describe.**
Currently, to extend the DataProvider, we have to be very explicit about generics, and do a fair amount of hackery to expose Models/Resources everywhere.
The aim here is to build an abstraction around the pieces in react-admin, to make wiring up APIs/Models/Views a lot more seamless within editor environments (fully static with plain JS for users).

**Describe the solution you'd like**
Currently, `DataProvider` is a TypeAlias, which are not injectable.
By allowing us to inject the `DataProvider`, we can cause `resourceName -> RaRecord` associations to propagate to all the hooks (`useGetList('cats')` should automatically find my Cat Model).

For now, changing it from `type` to `interface` might be enough, however I'm unsure of the fallout. I'll spin up a PR with a Demo to verify how it all goes, and so you have something physical to see.

I suspect this will be beneficial for the `no-code` project, however I've yet to do a dive into what that is in detail.

**Describe alternatives you've considered**
Wrapping + explicit types == Code too noisy

**Additional context**
n/a

PR #7351, opened 2022-03-07
feat: Improve type hints around DataProvider

antdking:
This moves `DataProvider` to be an interface, which allows for end-users to have better inference.

Some immediate wins included:
- FakeServer has been updated to infer types for end-users, speeding up prototyping
- LocalStorage can now discard its type definitions, cleaning the implementation (example of how this change improves life for end-users)
- Ability to inject known records (docs todo)

Example from the CRM project:
image

Future PR:
- wire up hooks to fetch their types from the `DataProvider` definition
  _This might cause problems if you have parallel Admin panels running, as types are global, whereas DP is scoped to a component's context provider. If wanted, we can setup a registry-registry to represent distinct entrypoints of the DP. Sounds messy, but let me know if this is a known usecase._

closes: #7326

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -9,10 +9,10 @@ import { AuthActionType } from './auth/types';
 
 export type Identifier = string | number;
 
-export type RaRecord = {
+export interface RaRecord {
     id: Identifier;
     [key: string]: any;
-};
+}
 
 export interface SortPayload {
     field: string;
@@ -42,12 +42,12 @@ export type I18nProvider = {
     [key: string]: any;
 };
 
-export type UserIdentity = {
+export interface UserIdentity {
     id: Identifier;
     fullName?: string;
     avatar?: string;
     [key: string]: any;
-};
+}
 
 /**
  * authProvider types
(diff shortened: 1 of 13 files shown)


antdking:
Feel free to have a quick review of this. I'm playing with supporting `.method<RecordType>()`

packages/ra-core/src/types.ts
```diff
@@ -9,10 +9,10 @@ import { AuthActionType } from './auth/types';
 
 export type Identifier = string | number;
 
-export interface RaRecord {
+export type RaRecord = {
```
djhi, line 12: Why?

packages/ra-core/src/types.ts
```diff
@@ -42,12 +42,12 @@ export type I18nProvider = {
     [key: string]: any;
 };
 
-export interface UserIdentity {
+export type UserIdentity = {
```
djhi, line 45: Why?

antdking: accidental

packages/ra-core/src/types.ts
antdking: Having it as a type alias seems more appropriate, as we're just describing the shape of a Type, and so it isn't a separate entity itself.
Doesn't make much difference outside of errors talking in terms of the shape, instead of the type `RaRecord`.

I can change it back if you want.

antdking:
@djhi This is ready for review as-is.

This PR is largely targetting the internals of the DataProvider, though do provide standalone benefits for users.

There will be a follow-up PR to fully implement this into the Hooks.

packages/ra-core/src/types.ts
```diff
@@ -72,54 +72,210 @@ export type LegacyAuthProvider = (
  * dataProvider types
  */
 
-export type DataProvider<ResourceType extends string = string> = {
-    getList: <RecordType extends RaRecord = any>(
+// https://stackoverflow.com/questions/51465182/how-to-remove-index-signature-using-mapped-types/68261113#68261113
+type RemoveIndex<T> = {
+    [P in keyof T as string extends P
```
antdking, line 77: Fixed by updating Prettier: #7365

fzaninotto: To make this mergeable in standalone, can you add an ignore comment?

fzaninotto (commented):
Thanks alot for looking into this!

I don't understand the benefits so far, as the direct use of dataProvider methods isn't encouraged. Could you elaborate?

Also, I'm afraid this comes a bit late for 4.0. We are working on fixing bugs and ironing out the documentaiton, and this seems something that will require a lot of testing (I'm thinking about error messages, which are usually incomprehensible in TS) and tweaking to provide a good DX.

So we'll keep this open, but probably not merge it soon.

fzaninotto:
This type of change would be breaking:

```diff
-create: <RecordType extends RaRecord = any>(
+create<
+       ResourceType extends ResourceTypes<Definition>,
+       RecordType extends RaRecord = Definition[ResourceType]
+   >(
        resource: ResourceType,
        params: CreateParams
    ): Promise<CreateResult<RecordType>>;
```

And I still don't understand how it would be beneficial. 

Besides, your PR contain a lot of unrelated code style changes that don't pass our prettier config.

This PR has become stale, so I'm closing it.

PR #7352, opened 2022-03-07
fix: expose combineDataProviders to 'react-admin'

antdking:
As per the docs


fzaninotto (approved):

fzaninotto:
Thanks for spotting this!

PR #7354, opened 2022-03-07
Add Saved Queries

djhi:


packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -23,25 +24,18 @@ describe('<FilterButton />', () => {
         filterValues: {},
     };
 
-    describe('filter button', () => {
-        it('should not be rendered, if all filters are already being displayed', () => {
-            const { queryByText } = render(<FilterButton {...defaultProps} />);
-            expect(queryByText('ra.action.add_filter')).toBeNull();
-        });
-    });
```
djhi, line 31: Removed because if we don't have any filter to show, then some filtering has been applied and we should show the button to allow users to persist it

WiXSL (approved):

WiXSL:
Everything seems all right! 👍

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -9,21 +8,43 @@ import {
...
+    const navigate = useNavigate();
+    const {
+        displayedFilters = {},
+        filterValues,
+        perPage,
+        showFilter,
+        sort,
+    } = useListContext(props);
+    const hasFilterValues = !isEqual(filterValues, {});
+    const hasSavedCurrentQuery = savedQueries.some(savedQuery =>
```
fzaninotto, line 41: make sure savedQueries is an array of the given format before using it.

packages/ra-ui-materialui/src/list/filter/RemoveSavedQueryDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+        filterValues,
+        sort,
+        perPage,
+        displayedFilters,
+    } = useListContext();
+
+    const [savedQueries, setSavedQueries] = useSavedQueries(resource);
+
+    const removeQuery = (): void => {
+        const index = savedQueries.findIndex(savedFilter =>
```
fzaninotto, line 37: same

packages/ra-ui-materialui/src/list/filter/SavedQueriesList.tsx
```diff
@@ -0,0 +1,154 @@
...
+        resource,
+        filterValues,
+        displayedFilters,
+        sort,
+        perPage,
+    } = useListContext();
+
+    const [savedQueries] = useSavedQueries(resource);
+
+    const hasSavedCurrentFilterValue = savedQueries.some(savedQuery =>
```
fzaninotto, line 69: same

packages/ra-ui-materialui/src/list/filter/SavedQueriesList.tsx
```diff
@@ -0,0 +1,154 @@
...
+        isEqual(savedQuery.value, {
+            filter: filterValues,
+            sort,
+            perPage,
+            displayedFilters,
+        })
+    );
+    const hasFilterValues = !isEqual(filterValues, {});
+
+    // note: we don't use react-admin's FilterList because it doesn't accept a default translation
```
fzaninotto, line 79: I don't understand this one

fzaninotto (changes requested):

docs/FilteringTutorial.md
```diff
@@ -225,69 +225,17 @@ export default {
 
 ## Saved Queries: Let Users Save Filter And Sort
 
-[![Saved Queries in FilterList](https://marmelab.com/ra-enterprise/modules/assets/ra-preferences-SavedQueriesList.gif)](https://marmelab.com/ra-enterprise/modules/assets/ra-preferences-SavedQueriesList.gif)
+[![Saved Queries in FilterList](./img/SavedQueriesList.gif)](./img/SavedQueriesList.gif)
 
-Saved Queries are an [Enterprise Edition](https://marmelab.com/ra-enterprise)<img class="icon" src="./img/premium.svg" /> feature letting users save a combination of filters and sort parameters into a new, personal filter. Saved queries persist between sessions, so users can find their custom queries even after closing and reopening the admin. Saved queries are available both for the Filter Button/Form combo and for the `<FilterList>` Sidebar:
+Saved Queries let users save a combination of filters and sort parameters into a new, personal filter. Saved queries persist between sessions, so users can find their custom queries even after closing and reopening the admin. Saved queries are available both for the Filter Button/Form combo and for the `<FilterList>` Sidebar.
```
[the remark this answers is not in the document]
djhi: > Saved queries are available both for the Filter Button/Form combo and for the `<FilterList>` Sidebar.

There's nothing to say about how to use it, it's enabled and usable without any custom code

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { styled } from '@mui/material/styles';
```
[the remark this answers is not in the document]
djhi: Because it remains focused which is actually a good thing regarding accessibility

packages/ra-ui-materialui/src/list/filter/AddSavedQueryDialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+        setQueryName(event.target.value);
+    };
+
+    const handleFormSubmit = (e: FormEvent<HTMLFormElement>): void => {
+        e.preventDefault();
+        addQuery();
+    };
+
+    const addQuery = (): void => {
+        setSavedQueries(
```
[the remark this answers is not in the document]
djhi: Why wouldn't it be an array and what should I do if it's not?

packages/ra-ui-materialui/src/list/filter/SavedQueriesList.tsx
djhi: Wrongly copy/pasted from ra-preferences. Refactored

docs/FilteringTutorial.md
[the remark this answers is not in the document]
djhi: Added

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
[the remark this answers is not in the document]
fzaninotto: I disagree. The first added filter should receive the focus. If you don't know which element to focus, I suggest to keep the behavior of the previous FilterButton.

packages/ra-ui-materialui/src/list/filter/AddSavedQueryDialog.tsx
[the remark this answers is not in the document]
fzaninotto: It can be something else if anotrher piece of code messes up with it. If it's not the right format, ignore the value and fall back to the default value. 

See https://github.com/marmelab/react-admin/blob/next/docs/Store.md#forward-compatibility for an explanation

packages/ra-ui-materialui/src/list/filter/FilterList.tsx
```diff
@@ -40,22 +40,31 @@ import { useTranslate } from 'ra-core';
...
             </Box>
             <List dense disablePadding>
                 {children}
             </List>
-        </>
+        </Root>
     );
 };
+
+const Root = styled('div', {
```
fzaninotto, line 59: That's basically a `Box`

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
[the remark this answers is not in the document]
djhi: The thing is, I haven't done anything to make it behave that way so I don't know what to revert

packages/ra-core/src/store/useStore.ts
```diff
@@ -46,25 +47,62 @@ import { useStoreContext } from './useStoreContext';
...
+    defaultValue?: T,
+    validate?: ValidateStoreValue<T>
+): UseStoreResult<T> => {
     const { getItem, setItem, subscribe } = useStoreContext();
+    if (
+        typeof validate === 'function' &&
+        typeof defaultValue !== 'undefined' &&
+        !validate(defaultValue)
+    ) {
+        console.warn(`Invalid default value for store key ${key}`);
```
fzaninotto, line 59: I think it should throw in that case. Also, please do that only once (in a useEffect, or in the callback of useState), because otherwise you'll be revalidating the default value at each render, which may be expensive.

packages/ra-core/src/store/useStore.ts
```diff
@@ -46,25 +47,62 @@ import { useStoreContext } from './useStoreContext';
...
+    }
     const [value, setValue] = useState(() => getItem(key, defaultValue));
 
+    useEffect(() => {
```
fzaninotto, line 63: In the following scenario, this will not work:

1. A user calls setStore with an invalid value
2. the state setter updates the state
3. the components relying on that state rerender, possibly with an invalid state shape
4. the effect runs, logs a warning and sets the value back to default. 

Too late: the error already occured in step 3

Besides, you already check validation in the set method, I don't see any case where this effect is necessary.

packages/ra-core/src/store/useStore.ts
```diff
@@ -46,25 +47,62 @@ import { useStoreContext } from './useStoreContext';
...
+            if (typeof validate === 'function') {
+                if (!validate(newValue)) {
+                    console.warn(`Invalid value for store key ${key}`);
+
+                    if (!validate(runtimeDefaultValue)) {
+                        console.warn(
+                            `Invalid default value for store key ${key}`
+                        );
+                    }
+                }
```
fzaninotto, line 104: you have to fall back to the default value if the validation fails

fzaninotto (changes requested):
You have to document the validate method in the store hook

packages/ra-core/src/store/useStore.ts
djhi: The idea was to validate the initial value from the store but you're right, it's incorrect

fzaninotto:
You still haven't fixed the focus p^roblem on the FilterButton after saving a query

djhi:
> You still haven't fixed the focus problem on the FilterButton after saving a query

Because I haven't find the fix yet

fzaninotto (changes requested):

docs/FilterList.md
```diff
@@ -214,32 +214,31 @@ You can place these `<FilterList>` anywhere inside a `<List>`. The most common c
...
+        sx={{
+            display: {
+                xs: 'none',
+                sm: 'block'
+            },
             order: -1, // display on the left rather than on the right of the list
             width: '15em',
-            marginRight: '1em',
-        },
-        [theme.breakpoints.down('sm')]: {
```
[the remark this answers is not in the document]
djhi: It is, read carefully ;-)

docs/FilterList.md
```diff
@@ -214,32 +214,31 @@ You can place these `<FilterList>` anywhere inside a `<List>`. The most common c
...
 import { LastVisitedFilter, HasOrderedFilter, HasNewsletterFilter, SegmentFilter } from './filters';
 
-const Card = withStyles(theme => ({
-    root: {
-        [theme.breakpoints.up('sm')]: {
+const FilterSidebar = () => (
+    <Box
+        sx={{
+            display: {
+                xs: 'none',
```
djhi, line 225: Hidden here

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -87,7 +133,66 @@ export const FilterButton = (props: FilterButtonProps): JSX.Element => {
                         autoFocus={index === 0}
                     />
                 ))}
+                {savedQueries.map((savedQuery, index) =>
```
fzaninotto, line 136: you need to be more defensive here, too

packages/ra-ui-materialui/src/list/filter/SavedQueriesList.tsx
```diff
@@ -0,0 +1,141 @@
...
+                />
+            ) : (
+                <Tooltip
+                    title={translate('ra.saved_queries.help')}
+                    className={SavedQueriesListClasses.floatingTooltip}
+                >
+                    <HelpIcon />
+                </Tooltip>
+            )}
+            {savedQueries.map((savedQuery, index) => (
```
fzaninotto, line 100: here, too

PR #7360, opened 2022-03-07
Make e-commerce demo canonical 

fzaninotto:
- [x] Invoices
- [x] Orders
- [x] Finishing touches
- [x] Fix UserMenu on small screens
- [x] Fix AutocompleteInput doesn't have a minimum width (and doesn't support `sx`)

### DIFF
--- a/packages/ra-core/src/dataProvider/useIsDataLoaded.ts
+++ b/packages/ra-core/src/dataProvider/useIsDataLoaded.ts
@@ -1,0 +1,40 @@
+import { useEffect, useState } from 'react';
+import { useQueryClient, QueryObserver } from 'react-query';
+
+/**
+ * Check if react-query has already fetched data for a query key.
+ *
+ * This hook is reactive.
+ *
+ * @example
+ * const isCustomerLoaded = useIsDataLoaded(['customers', 'getOne', { id: customerId }]);
+ *
+ * @returns {boolean} true if the data is loaded, false otherwise
+ */
+export const useIsDataLoaded = (
+    queryKey: any,
+    options: { enabled?: boolean } = {}
+) => {
+    const { enabled = true } = options;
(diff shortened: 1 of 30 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -517,6 +519,21 @@ If you provided a React element for the optionText prop, you must also provide t
...
+};
+
+const StyledAutocomplete = styled(Autocomplete, {
+    name: PREFIX,
+    overridesResolver: (props, styles) => styles.root,
+})(({ theme }) => ({
+    [`& .${AutocompleteInputClasses.textField}`]: {
+        minWidth: theme.spacing(20),
+    },
+}));
```
djhi, line 535: You have to update the upgrade guide and the docs about this.

djhi (changes requested):

WiXSL (changes requested):
Customers: In the following gif you can see two issues, The moving inputs when selecting items and the cut of avatar on the app bar

chrome-capture (1)

djhi (approved):
:fire:

packages/ra-core/src/dataProvider/useIsDataLoaded.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Check if react-query has already fetched data for a query key.
+ *
+ * This hook is reactive.
+ *
+ * @example
+ * const isCustomerLoaded = useIsDataLoaded(['customers', 'getOne', { id: customerId }]);
+ *
+ * @returns {boolean} true if the data is loaded, false otherwise
+ */
+export const useIsDataLoaded = (
```
djhi, line 14: It would be nice to document this

Issue #7361, opened 2022-03-07
Dependency Cycle between ra-core and ra-i18n-polyglot


antdking:
**What you were expecting:**

No dependency cycle

**What happened instead:**

```sh
$ yarn lerna run build
lerna notice cli v4.0.0
lerna info Executing command in 16 packages: "npm run build"
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE ra-core -> ra-i18n-polyglot -> ra-core
```

**Steps to reproduce:**

run `yarn build`

**Related code:**

N/A

**Other information:**

`ra-core` has tests that depend on `ra-i18n-polyglot`. This means that `ra-i18n-polyglot` needs to be included in the `ra-core` `devDependencies`.

---

While this is functioning for Lerna (with a warning), it fails for `yarn workspaces foreach`.

Suggest finding another way to test interpolation support.

**Environment**

* React-admin version: next
* Last version that did not exhibit the issue (if applicable):
* React version:
* Browser:
* Stack trace (in case of a JS error):

fzaninotto:
Fixed by #7362

PR #7362, opened 2022-03-07
fix(core): Dependency loop on ra-i18n-polyglot

antdking:
Core us used by most, so remove the ra-i18n-polyglot dev dependency.

This causes a dependency cycle while trying to build.

Implement a basic way of doing interpolation so we can test variables get passed correctly.

Closes: #7361


fzaninotto:
Thanks a lot!

PR #7366, opened 2022-03-09
Add Store invalidation system 

fzaninotto:
## Problem

Code relyoing on stored values to have a certain shape may fail if used with an old value with a different shape

## Solution

Add ability to specify a version number for the store, and invalidate the store if the version from the code differs from the version from the store.

### DIFF
--- a/docs/Store.md
+++ b/docs/Store.md
@@ -55,7 +55,7 @@ React-admin components don't access the store directly ; instead, they use purpo
 
 Using specialized hooks avoids depending on a store key.
 
-## Foward Compatibility
+## Forward Compatibility
 
 If you store complex objects in the Store, and you change the structure of these objects in the application code, the new code relying on the new object structure may fail when running with an old stored object.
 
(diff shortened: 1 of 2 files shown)


docs/Store.md
```diff
@@ -55,6 +55,77 @@ React-admin components don't access the store directly ; instead, they use purpo
 
 Using specialized hooks avoids depending on a store key.
 
+## Foward Compatibility
```
djhi, line 58:
```suggestion
## Forward Compatibility
```

docs/Store.md
```diff
@@ -55,6 +55,77 @@ React-admin components don't access the store directly ; instead, they use purpo
...
+const STORE_VERSION = 2
+
+const App = () => (
+    <Admin dataProvider={dataProvider} store={localStorageStore(STORE_VERSION)}>
+        <Resource name="posts" />
+    </Admin>
+);
+```
+
+Increase the version number eachj time you push code that isn't compatible with the stored values.
```
djhi, line 127:
```suggestion
Increase the version number each time you push code that isn't compatible with the stored values. 
```

djhi (changes requested):

djhi (approved):

PR #7367, opened 2022-03-09
Fix issue with create

djhi:


packages/ra-core/src/form/useIsFormInvalid.ts
```diff
@@ -20,9 +20,15 @@ export const useIsFormInvalid = () => {
         if (submitCount > submitCountRef.current) {
             submitCountRef.current = submitCount;
 
-            setIsInvalid(!isValid);
+            // For some reason, the validation state might not be sync yet on first submit
```
WiXSL, line 23: Isn't this related to the fact that `isValid` is affected by mode at useForm. This state is only applicable with onChange, onTouched, and onBlur mode.

https://react-hook-form.com/api/useform

djhi: Good catch, the fix is ok though as it will work in any mode (which users can change)

WiXSL (approved):

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -60,28 +58,28 @@ export const useCreateController = <
...
 
-    const [create, { isLoading: saving }] = useCreate(
+    const [create, { isLoading: saving }] = useCreate<RecordType>(
         resource,
         undefined,
         otherMutationOptions
     );
 
     const save = useCallback(
-        (
```
fzaninotto, line 70: I don't understand these changes. Do they change anything? If not, why is the new code better?

djhi: The async keywords were added while debugging. I'll remove them. However, we now return the promise at L77, just like in `useEditController`

fzaninotto (approved):

PR #7377, opened 2022-03-10
Allow usage with react 18

fzaninotto:
- [x] Update dependencies to use react-18 (not in this PR)
- [x] Update the simple example to use createRoot and test that it works (not in this PR)
- [x] Update the ecommerce example to use createRoot and StrictMode and test that it works (not in this PR)
- [x] Relax react version dependency

### DIFF
--- a/examples/demo/src/fakeServer/index.ts
+++ b/examples/demo/src/fakeServer/index.ts
@@ -1,8 +1,20 @@
+// only install the mocks once
+// this is necessary with react@18 in StrictMode
+let fakeServer: any;
+
 export default (type: string) => {
-    switch (type) {
-        case 'graphql':
-            return import('./graphql').then(factory => factory.default());
-        default:
-            return import('./rest').then(factory => factory.default());
+    if (!fakeServer) {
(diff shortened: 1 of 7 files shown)


djhi (approved):

stof:
is it expected to mark versions 1 to +INF as supported in the peerDependencies ? Previously, the code had a 16.9.0 lower bound so I doubt older versions work, and you only tested it with version 19, not with 19+

fzaninotto:
I must admit that I initially made the following change:

```diff
-"react": "^16.9.0 || ^17.0.0",
+"react": "^16.9.0 || ^17.0.0" || "rc",
```

And prettier updated it to 

```diff
"react": "*",
```

I didn't give it more thought, but you're right, this is too broad.

However, if there is a prettier rule that doesn't accept 3 versions, there must be a reason, so this needs investigation. 

Would you like to take a look at the problem?

jchook:
Maybe a version range is appropriate here? e.g. `"react": ">=16.9.0 <19.0.0"`

PR #7382, opened 2022-03-11
fix: register resource inside the component

alanpoulain:
This is a proposal to register the resource inside the component itself rather than in `useConfigureAdminRouterFromChildren`.

The main issue I was solving was to make the [ResourceGuesser component from API Platform Admin](https://github.com/api-platform/admin/blob/main/src/ResourceGuesser.tsx) works with react-admin 4: since the list/create/edit/show props are not explicitly given to the component, it is not correctly registered (and for instance the menu is not correctly filled since `hasList` is always false).

I don't have a workaround to make it work in the current implementation of react-admin 4.

However I think my proposal is interesting for you too: I don't think it is the `useConfigureAdminRouterFromChildren` responsibility to watch the resources and register them, having it in the Resource component itself seems a better idea.

I hope my proposal will satisfy you, if not, if you have any idea to make it work differently please tell me.

### DIFF
--- a/packages/ra-core/src/core/Resource.tsx
+++ b/packages/ra-core/src/core/Resource.tsx
@@ -1,34 +1,12 @@
 import * as React from 'react';
-import { isValidElement, useEffect } from 'react';
+import { isValidElement } from 'react';
 import { Route, Routes } from 'react-router-dom';
 
 import { ResourceProps } from '../types';
 import { ResourceContextProvider } from './ResourceContextProvider';
-import { useRegisterResource } from './useRegisterResource';
 
 export const Resource = (props: ResourceProps) => {
-    const registerResource = useRegisterResource();
-    const {
-        create: Create,
-        edit: Edit,
-        icon,
-        list: List,
-        name,
-        options,
-        show: Show,
-    } = props;
-
-    useEffect(() => {
-        registerResource({
-            name: name,
-            options: options,
-            hasList: !!List,
-            hasCreate: !!Create,
-            hasEdit: !!Edit,
-            hasShow: !!Show,
-            icon: icon,
-        });
-    }, [registerResource, name, options, List, Create, Edit, Show, icon]);
+    const { create: Create, edit: Edit, list: List, name, show: Show } = props;
 
     return (
         <ResourceContextProvider value={name}>
@@ -63,3 +41,21 @@ export const Resource = (props: ResourceProps) => {
 };
 
 Resource.raName = 'Resource';
+
+Resource.registerResource = ({
+    create,
+    edit,
+    icon,
+    list,
+    name,
+    options,
+    show,
+}: ResourceProps) => ({
+    name,
+    options,
+    hasList: !!list,
+    hasCreate: !!create,
+    hasEdit: !!edit,
+    hasShow: !!show,
+    icon: icon,
+});
--- a/packages/ra-core/src/core/Resource.tsx
+++ b/packages/ra-core/src/core/Resource.tsx
@@ -1,12 +1,34 @@
 import * as React from 'react';
-import { isValidElement } from 'react';
+import { isValidElement, useEffect } from 'react';
 import { Route, Routes } from 'react-router-dom';
 
 import { ResourceProps } from '../types';
 import { ResourceContextProvider } from './ResourceContextProvider';
+import { useRegisterResource } from './useRegisterResource';
 
 export const Resource = (props: ResourceProps) => {
-    const { create: Create, edit: Edit, list: List, name, show: Show } = props;
+    const registerResource = useRegisterResource();
+    const {
+        create: Create,
+        edit: Edit,
+        icon,
+        list: List,
+        name,
+        options,
+        show: Show,
+    } = props;
+
+    useEffect(() => {
+        registerResource({
+            name: name,
+            options: options,
+            hasList: !!List,
+            hasCreate: !!Create,
+            hasEdit: !!Edit,
+            hasShow: !!Show,
+            icon: icon,
+        });
+    }, [registerResource, name, options, List, Create, Edit, Show, icon]);
 
     return (
         <ResourceContextProvider value={name}>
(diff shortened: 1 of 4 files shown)


alanpoulain:
I doesn't seem to work when the `Resource` component is not rendered at all, for instance when it is inside a conditional like this:
```jsx
<Admin>
   {permissions => (
       <>
           {permissions ? <Resource name="users" {...users} /> : null}
       </> 
   )}
</Admin>
```

I've found an ugly workaround by transforming `ResourceGuesser` components into `Resource` ones with props explicitly set: https://github.com/api-platform/admin/blob/0d44df98c5340cc0de86408687ab07cc0489941f/src/getTransformedGuesserChildren.tsx.
I hope to find a better way than this.

fzaninotto:
I'm OK with the idea, but you'll have to find a way to make it work when there is a resource as function. Perhaps @djhi has an insight?

djhi:
The main issue here is that we pass the resource elements as the `element` prop of their main route in `CoreAdminRouter`. Moving the registration inside the `Resource` component means only the rendered resource is registered. You can see in the deployed simple example for your PR that we don't have the comments menu anymore. I don't have a solution yet but I'll investigate

djhi:
So after some research and discussions, we see only two options:

- Leverage the fact that you can register an already registered resource and the new config will overwrite the old one. You could do that in your custom `ResourceGuesser` without any changes on our side. However, you would face the same limitation I explained above: only the rendered resources would be using the correct configuration.
- We could add a `registerResource` prop to the `Resource` component. It would accept a function that should return the resource definition. Our `useConfigureAdminRouterFromChildren` would call this function allowing you to provide the config you need.

fzaninotto:
> The resource can register itself by defining a default registerResource but an outside component can also change the way the resource is registered, which shouldn't be the case IMO. 

I don't understand your point. Maybe we're not talking about the same thing. 

To give you an example, what we suggest is the following:

```jsx
const App = () => (
    <Resource name="posts" list={list} />
    <ResourceGuesser registerResource={() => ({ hasList: true, hasEdit: false, ... })} />
  </Admin>
);
```

I see no better way, so if this doesn't fit the API Platform use case, you'll be reduced to the hacks you've described above.

alanpoulain:
> To give you an example, what we suggest is the following:
> 
> ```js
> const App = () => (
>   <Admin dataProvider={dataProvider}>
>     <Resource name="posts" list={list} />
>     <ResourceGuesser registerResource={() => ({ hasList: true, hasEdit: false, ... })} />
>   </Admin>
> );
> ```
> 
> I see no better way, so if this doesn't fit the API Platform use case, you'll be reduced to the hacks you've described above.

It does fit API Platform Admin, since we can do something like this for the `ResourceGuesser` component:
```js
const ResourceGuesser = ({
  list = ListGuesser,
  edit = EditGuesser,
  create = CreateGuesser,
  show = ShowGuesser,
  registerResource = () => ({ hasList: true, hasEdit: true, hasCreate: true, hasShow: true })
  ...props
}: ResourceProps) => (
  <Resource
    {...props}
    create={create}
    edit={edit}
    list={list}
    show={show}
  />
);
```
It just seems weird for react-admin / API Platform users to be able to use this prop outside the `Resource` / `ResourceGuesser` component, like your example.

But if this solution is chosen, it will solve all the issues.

> Yes, there was an issue I fixed locally but forgot to mention it.

Nice! Choosing this solution means that users would not be able to use a `ResourceGuesser` component in a conditional since it will not be rendered.
Personally I prefer this way API-wise, but users-wise it could be not a good idea.

alanpoulain:
> No need. **You** will use this prop in the `ResourceGuesser`, not them.

Are you sure about this? How would you be able to get the `registerResource` prop from the `Resource` component inside the `ResourceGuesser` one in `useConfigureAdminRouterFromChildren`?
But if you can, it's OK for me.

djhi:
You're right, you would have to set it in the `defaultProps`

alanpoulain:
> You're right, you would have to set it in the `defaultProps`

Does it work in the `defaultProps`? Can you retrieve the `registerResource` prop even if the component is not rendered? It could be a solution if it's the case, but doesn't React want to deprecate this property anyway?
WDYT of my last solution (using a `registerResource` property directly on the component)?

djhi:
> Does it work in the defaultProps?

It does. However, you are also right about React probably deprecating the `defaultProps` someday.
I'll try your solution and report back

djhi:
Ok so I have a slight variation I tested locally:

```jsx
const ResourceGuesser = ({
  list = ListGuesser,
  edit = EditGuesser,
  create = CreateGuesser,
  show = ShowGuesser,
  ...props
}: ResourceProps) => (
  <Resource
    {...props}
    create={create}
    edit={edit}
    list={list}
    show={show}
  />
);

ResourceGuesser.registerResource = (props: ResourceProps) => ({
    hasList: true,
    hasEdit: true,
    hasCreate: true,
    hasShow: true
});
```

alanpoulain:
It seems alright to me, thank you for having tested this solution! Out of curiosity, why the need for the `props` argument?

If this solution is accepted, what will be the next course of action? Will you make another PR or would you want me to modify this one?

djhi:
> Out of curiosity, why the need for the props argument?

You get the props and can decide what to return based on them?

packages/ra-core/src/core/Resource.tsx
```diff
@@ -41,3 +41,21 @@ export const Resource = (props: ResourceProps) => {
...
+    options,
+    show,
+}: ResourceProps) => ({
+    name,
+    options,
+    hasList: !!list,
+    hasCreate: !!create,
+    hasEdit: !!edit,
+    hasShow: !!show,
+    icon: icon,
```
fzaninotto, line 60:
```suggestion
    icon,
```

fzaninotto (changes requested):

djhi:
Superseded by #7392. Thanks!

PR #7392, opened 2022-03-16
Custom resource element

djhi:
Supersedes #7382

Thanks @alanpoulain for reporting this issue and suggesting the solution!

# Problem

API Platform needs a custom Resource element to configure default views when users don't provide their own. However, because of how react-router v6 works, we have to register the resource definitions outside of the Resource components as they are rendered inside a route matching their name and registering their definition in their render method would only work for the rendered resource. This means we have to inspect the Resource props so views provided by a custom Resource element wouldn't be available at the registration time.

# Solution

`Resource` elements must now provide a static `registerResource` method which is called with the props provided by the user and must return the `ResourceDefinition`. This allows custom `Resource` elements to alter the definition which would be inferred by react-admin

```jsx
const ResourceGuesser = ({
  list = ListGuesser,
  edit = EditGuesser,
  create = CreateGuesser,
  show = ShowGuesser,
  ...props
}: ResourceProps) => (
  <Resource
    {...props}
    create={create}
    edit={edit}
    list={list}
    show={show}
  />
);

ResourceGuesser.registerResource = (props: ResourceProps) => ({
    name: props.name,
    hasList: true,
    hasEdit: true,
    hasCreate: true,
    hasShow: true,
    icon: props.icon,
    options: props.options,
});
```

### DIFF
--- a/packages/ra-core/src/core/Resource.tsx
+++ b/packages/ra-core/src/core/Resource.tsx
@@ -57,5 +57,5 @@ Resource.registerResource = ({
     hasCreate: !!create,
     hasEdit: !!edit,
     hasShow: !!show,
-    icon: icon,
+    icon,
 });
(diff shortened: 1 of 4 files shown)


fzaninotto (approved):

alanpoulain (approved):
Thank you!

PR #7393, opened 2022-03-16
Change Form syntax from rendreProps to children

fzaninotto:
- [x] Change Form syntax
- [x] Update tests and examples
- [x] Add introduction


packages/ra-ui-materialui/src/form/SimpleFormView.tsx
```diff
@@ -49,7 +47,7 @@ const DefaultComponent = ({ children, sx, ...props }) => (
 );
 const DefaultToolbar = <Toolbar />;
 
-export interface SimpleFormViewProps extends FormRenderProps {
+export interface SimpleFormViewProps extends FormRenderProps, StackProps {
```
djhi, line 50: What happens if you provided the `component` prop? I think you need a generic here that defaults to `StackProps`

djhi, line 50: What happens if you provided the `component` prop? I think you need a generic here that defaults to `StackProps`

djhi (changes requested):

packages/ra-ui-materialui/src/form/SimpleFormView.tsx
fzaninotto: I think the Component should only override the wrapper (CardContent in that case) and not the Stack, because otherwise it's an entirely different Layout.

docs/EditTutorial.md
```diff
@@ -0,0 +1,324 @@
...
+};
+```
+
+This form displays 3 inputs (two text inputs and one select input), and redirects to the book list view upon successful submission. It doesn't even contain default values, validation, or dependent inputs. Yet, it's already quite verbose.
+
+It's a super common component. In fact, many of its features could be extracted for reuse in other pages. Let's see how to improve the code and the developer experience in the same process. 
+
+## `<SimpleForm>`: Form Logic In A Simple Layout
+
+React-hook-form requires developers to create a form `control`, and to wrap each input with a `<Controller>` tag. We can put this logic into a custom Form component to avoid repeating it for every form. This is what [the `<SimpleForm>` component](./SimpleForm.md) does:
```
djhi, line 86: It's actually not true. You could just `register` most MUI fields without a controller

djhi (changes requested):
Very nice. The new form component is way simpler and the hook is great! There are some linter warning left

docs/EditTutorial.md
djhi: Did you see this comment?

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -12,16 +12,30 @@ export const useApplyInputDefaultValues = (props: Partial<InputProps>) => {
...
-            get(getValues(), source) == null &&
-            defaultValue != null
-        ) {
-            resetField(source, {
-                defaultValue: get(record, source, defaultValue),
-            });
+    useEffect(() => {
+        if (defaultValue == null) return;
+        if (formValue == null && recordValue == null) {
+            // special case for ArrayInput: since we use get(record, source),
```
djhi, line 21: :+1:

packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -49,50 +49,44 @@ export const LoginForm = (props: LoginFormProps) => {
...
-                        />
-                        <TextInput
-                            source="password"
-                            label={translate('ra.auth.password')}
-                            type="password"
-                            autoComplete="current-password"
-                            validate={required()}
-                            fullWidth
-                        />
+        <Form onSubmit={submit} mode="onChange" noValidate>
```
djhi, line 52: You still have to make `Root` a styled version of `Form` to allow children styling

djhi (changes requested):

docs/EditTutorial.md
fzaninotto: > Most UI libraries are built to support only controlled components, such as [Material-UI](https://github.com/mui-org/material-ui) and [Antd](https://github.com/ant-design/ant-design).

https://react-hook-form.com/advanced-usage#ControlledmixedwithUncontrolledComponents

Is the react-hook-form doc wrong?

fzaninotto: OK, I'll update our doc then

djhi (changes requested):
Awesome! There are still a few eslint warnings though

djhi (approved):

PR #7395, opened 2022-03-16
Fix List exporter prop override

### LINKED ISSUE #7394: [ra4 beta 3] - Exporter appears on <List> even when set to false
**What you were expecting:**
Using the <List> component and passing the prop `exporter={false}' should remove the export option.

**What happened instead:**
Export option still appears.

**Related code:**

```javascript
const App = () => (
    <Resource name="users" list={UsersList} />
  </Admin>
);

const UsersList = () => (
  <List exporter={false}>
    <Datagrid>
      <TextField source="id" />
      <TextField source="username" />
    </Datagrid>
  </List>
);
```

**Other information:**
I tried to track back as much in the source as I could. From what I can tell, it looks like the ListView component is not pulling the exporter information from the useListContext.

```javascript
const {
        defaultTitle,
        data,
        total,
        isLoading,
        filterValues,
    } = useListContext<RecordType>(props);
```
/packages/ra-ui-materialui/src/list/ListView.tsx - line 39-45

It should be passing the exporter info from useListContext into the ListToolbar as a prop.

**Environment**

* React-admin version: 4.0.0-beta.3
* React version:
* Browser:

WiXSL:
Fixes #7394


packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -39,6 +39,7 @@ export const ListView = <RecordType extends RaRecord = any>(
     const {
         defaultTitle,
         data,
+        exporter,
```
fzaninotto, line 42: I think the change should be in ListActions, which should read the `exporter` from the `ListContext` rather than the props.

fzaninotto (changes requested):
This deserves an integration test

djhi (approved):

fzaninotto (approved):

PR #7404, opened 2022-03-18
fix: use right index for the transition in SimpleFormIterator

alanpoulain:
Before:
image
After:
image

It's probably because the `id` of the fields is changing (`id` is generated from [useFieldArray](https://react-hook-form.com/api/usefieldarray/)), and the `ref` needs to be changed as well (see http://reactcommunity.org/react-transition-group/transition/#Transition-prop-nodeRef).

Using the previous code from master is needed to solve this issue.
I needed to revert https://github.com/marmelab/react-admin/pull/7123 as well, I don't understand how the change in this PR can work.

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
@@ -99,7 +99,7 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
                     {fields.map((member, index) => (
                         <CSSTransition
                             nodeRef={nodeRef}
-                            key={member.id}
+                            key={index}
                             timeout={500}
                             classNames="fade"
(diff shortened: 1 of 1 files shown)


WiXSL:
@alanpoulain, thanks for working on this.
It seems to work, although, I see some other problems like adding items after removing all items creates empty items.
I belive, this could be a react-hook-form's `useFieldArray` problem

WiXSL (approved):

djhi:
Thanks!

djhi:
I should have remembered why I used react-hook-form id instead of our own. This introduces weird issues that you can see on the deployed simple example for the PR.

Open https://react-admin-az7z4sx9x-marmelab.vercel.app/#/posts/13

Add two backlink, remove one, add another:
array-input-issue

alanpoulain:
Maybe the alternative is to make sure the `id` from `useFieldArray` doesn't change then.

PR #7406, opened 2022-03-19
[doc] typos fix on doc

smeng9:
Add dynamic resource creation/update/delete example to Admin.md
Update some menu items in documentation to indicate some props are depreacted.


docs/Admin.md
```diff
@@ -507,4 +507,66 @@ function AsyncResources() {
...
+            .then((resources) => resources.forEach((r) => addResource(r)));
+    }, []);
+
+    return (
+        <AdminUI>
+            {Object.keys(resources).map((resource) => (
+                <Resource
+                    name={resources[resource].name}
+                    key={resources[resource].name}
+                    list={schemaToListView(resources[resource])}
```
fzaninotto, line 556: This isn't defined

docs/Admin.md
```diff
@@ -507,4 +507,66 @@ function AsyncResources() {
...
+                />
+            ))}
+            <Resource
```
fzaninotto, line 559: Adding a schemas Resource is overkill in this example IMHO

fzaninotto (changes requested):

docs/Admin.md
```diff
@@ -507,4 +507,59 @@ function AsyncResources() {
 }
 ```
 
+### Dynamically create, update, and delete `<Resource>`
```
fzaninotto, line 510: Now that I think about it, I don't think this addition is relevant. We already explain how to add Resources dynamically based on an HTTP request earlier in this chapter. Your addition only explains another way to do it, which is only useful if people want to recreate ra-no-code. And we don't want to explain too many ways to do things, to encourage reimplementing existing modules, or to solve every possible userland requirement in the documentation.

fzaninotto (changes requested):

docs/Admin.md
smeng9: Ok. I can remove this part for now.

Are we going to have the `ra-no-code` documentation mentioned somewhere else?

fzaninotto: When we feel that it's ready, yes. For now, it's not yet ready.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7413, opened 2022-03-23
docs: remove deprecated props and unused imports

takayukioda:
`sidebarIsOpen` in _MenuItemLink_ component and `onMenuClick` in _Menu_ component has been deprecated in https://github.com/marmelab/react-admin/pull/6230 and they're no longer needed to create custom menu components.
Also there were some unused import statements, so I've clean them all.

Note that I've only updated the place where I referred, so there might be more deprecated / unused imports in the another lines of this file.

### DIFF
--- a/docs/Theming.md
+++ b/docs/Theming.md
@@ -830,16 +830,12 @@ const App = () => (
 ```jsx
 // in src/Menu.js
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


fzaninotto (approved):

fzaninotto:
Thanks!

PR #7416, opened 2022-03-23
Preventing navigation while setting List filters

### LINKED ISSUE #7368: Race condition when applying List filter and changing page
## Steps to reproduce

1. Open the [Simple example codesandbox](https://iuxnw.sse.codesandbox.io/)
2. In the post List, enable the title filter
3. Type a few characters in the title input
4. Immediately after, click on the Comments menu
5. The Title filter is applied oni the Comments list (which is wrong, as comments have no title

## Analysis

The `setFilters` called by `<FilterForm>` is debounced by default. This means that it changes the URL 300ms after the user applies a filter. But if the user has already moved to another page, the redirection happens on the wrong page. 

## Possible solutions

1. in `useListController`, keep track of calls to `setFilters`, and cancel them on unmount
2. in `useListParams`, only call `navigate` if the location maches the current resource

## Environment

* React-admin version: v3 and v4
* Last version that did not exhibit the issue (if applicable): none, the bug has always been here
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #7368

### DIFF
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -2,7 +2,7 @@ import { useCallback, useMemo, useEffect, useState, useRef } from 'react';
 import { parse, stringify } from 'query-string';
 import lodashDebounce from 'lodash/debounce';
 import pickBy from 'lodash/pickBy';
-import { useNavigate, useLocation } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


WiXSL:
Not sure if I like this approach or to cancel `setFilters` calls better

WiXSL:
Too much failing tests, this can't be the right approach

fzaninotto:
Superseded by #7443

PR #7422, opened 2022-03-23
[RFR] fix(types): missing resource options label property

soullivaneuh:
This aim to solve two things:

1. Using object as type is not recommended as we can not access a
   property without having Typescript yelling this one can not exist on
   object (ts2339)
2. As specified on the documentation, the label options is available and
   use by the react-admin ecosystem. It should be explicitly defined.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #7426, opened 2022-03-24
Remove Transitions From SimpleFormIterator

djhi:
Fixes #7381

PR #7428, opened 2022-03-24
Use framer-motion for SimpleFormIterator

alanpoulain:
Following https://github.com/marmelab/react-admin/pull/7426.

Use framer-motion to handle reorder animations. It also enables a drag feature.

### DIFF
--- a/packages/ra-ui-materialui/package.json
+++ b/packages/ra-ui-materialui/package.json
@@ -58,6 +58,7 @@
     "dependencies": {
         "autosuggest-highlight": "^3.1.1",
         "clsx": "^1.1.1",
+        "framer-motion": "^6.2.8",
         "inflection": "~1.12.0",
         "jsonexport": "^3.2.0",
         "lodash": "~4.17.5",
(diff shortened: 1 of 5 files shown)


fzaninotto:
Framer-motion is 138kb, it seems very expensive for animating 1 component

alanpoulain:
There are some considerations [in this page](https://www.framer.com/docs/guide-reduce-bundle-size/) to take into account.
Have you another idea for a component handling this animation? Or do you want to give up the idea of animating it?

fzaninotto:
It's be interesting to see how much weights it adds to a real app. Can you check on the simple example with `cd examples/simple && yarn build`?

alanpoulain:
With `framer-motion`:

image

Without `framer-motion`:

image

It seems it does add at least 100Kb.

fzaninotto:
that's too much for a single use

CiberNin:
What about react-dnd or react-smooth-dnd? They are about a third-fourth the size. And I do think dnd would be a very nice feature to have for SimpleFormIterator, or a nice incentive to shell out for enterprise ReferenceManyInput. Maybe potentially editable datagrid, ReferenceArrayInput, or DualListInput.

djhi:
That will be easier to do in userland with v5 so let's close this

PR #7439, opened 2022-03-28
Add missing exports

### LINKED ISSUE #7436: Ability to override default redirects inside Show, Create, Edit and Delete components (RA 4.0)
I've create a custom tabbed component that is supposed to render different resources inside MUI tabs
```js
<Tabbed route="search">
  <Tab name="home">
    {SearchHomePage}
  </Tab>
  <Tab name="products">
    <Resource name="products" list={ProductList} show={ProductShow} />
  </Tab>
  <Tab name="parts">
    <Resource name="parts" list={PartsList} show={PartsShow} />
  </Tab>
</Tabbed>
```

`Tabbed` component automatically registers the following routes:
- /search/home/*
- /search/products/* (all  `<Resource>`  are registered as child routes)
- /search/parts/* (all  `<Resource>`  are registered as child routes)

The issue is that react-admin's `Show`, `Create`, `Edit` and `Delete` controllers use `useRedirect`, `useCreatePath` and `useBasePath` to redirect  to `/<resourceName>` instead of `/search/<resourceName>` once actions are completed or failed

I would like to be able to force `Show`, `Create`, `Edit` and `Delete` components to use a different base path for default actions

- update `useRedirect`, `useCreatePath` and `useBasePath` to access options `{ basename: string}` and use this parameter instead of context value; also forward this parameter to `useCreatePath` call inside `useRedirect`
- update `Show`, `Create`, `Edit` and `Delete` controller to add  { basename:string } option, and forward this parameter to  `useRedirect`, `useCreatePath` and `useBasePath`  functions inside `useShowController`, `useEditController`, `useCreateController`, `useDeleteWithUndoController`, and `useDeleteWithConfirmController`

This solution will not impact current functionality, and will provide me with ability to easily influence where users are redirected to when actions are successful or failed.

Alternative is to replicate `Show`, `Create`, `Edit` and `Delete` components and supporting controllers in my project, and make changes directly to them; this option, however, is error-prone and is not maintainable.

WiXSL:
Fixes #7436

### DIFF
--- a/packages/ra-core/src/routing/index.ts
+++ b/packages/ra-core/src/routing/index.ts
@@ -1,6 +1,5 @@
 export * from './AdminRouter';
 export * from './BasenameContextProvider';
-export * from './HistoryRouter';
 export * from './linkToRecord';
 export * from './resolveRedirectTo';
 export * from './useBasename';
--- a/packages/ra-core/src/routing/index.ts
+++ b/packages/ra-core/src/routing/index.ts
@@ -1,4 +1,6 @@
 export * from './AdminRouter';
+export * from './BasenameContextProvider';
+export * from './HistoryRouter';
 export * from './linkToRecord';
 export * from './resolveRedirectTo';
 export * from './useBasename';


packages/ra-core/src/routing/index.ts
```diff
@@ -1,4 +1,6 @@
 export * from './AdminRouter';
+export * from './BasenameContextProvider';
+export * from './HistoryRouter';
```
djhi, line 3: Not sure we want to export that as react-router should release it

djhi (changes requested):

djhi (approved):

PR #7440, opened 2022-03-28
Doc improvement: add filters in list component caveat

slax57:
- [x] Add tips in doc to make it clearer that the filters should ideally be created _outside_ a List component, to avoid rendering loops
- [x] Add FAQ entry to give a typical example of problematic code, and the solution
- [ ] Continue investigation to see if we can detect this case programmatically, in order to avoid the loop or raise an error automatically


fzaninotto:
Superseded by #7368

PR #7445, opened 2022-03-29
Move `SaveButton` disable logic inside of it

WiXSL:


docs/SaveButton.md
```diff
@@ -133,6 +134,20 @@ const PostCreateToolbar = () => (
 
 Note that if you call `event.preventDefault()` in `onClick`, the form will not be submitted.
 
+## `alwaysEnable`
+
+By default, the `<SaveButton>` renderd by `<Toolbar>` is disabled until there is something to save, i.e. until the user changes at least one input. Set `alwaysEnable` to `true` to always enable the button.
```
fzaninotto, line 139:
```suggestion
By default, the `<SaveButton>` rendered by `<Toolbar>` is disabled until there is something to save, i.e. until the user changes at least one input. Set `alwaysEnable` to `true` to always enable the button.
```

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -57,16 +58,24 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
         ...rest
     } = props;
     const translate = useTranslate();
     const notify = useNotify();
     const form = useFormContext();
     const saveContext = useSaveContext();
+    const { isValidating } = useFormState();
+    // Use form isDirty, isValidating and form context saving to enable or disable the save button
+    // if alwaysEnable is undefined
+    const disabled = !valueOrDefault(
```
fzaninotto, line 75: the double negation is hard to reason about (!!disabledProp). Can you inverse the logic to make it more readable?

packages/ra-ui-materialui/src/form/Toolbar.tsx
```diff
@@ -92,10 +78,8 @@ export const Toolbar = <
         >
             {Children.count(children) === 0 ? (
                 <div className={ToolbarClasses.defaultToolbar}>
-                    <SaveButton disabled={disabled} />
-                    {record && typeof record.id !== 'undefined' && (
```
fzaninotto, line 96: you need to keep this condition to hide the DeleteButton on create screens

fzaninotto (changes requested):

packages/ra-ui-materialui/src/form/Toolbar.tsx
WiXSL: This condition already exists inside the `DeleteButton` itself, is duplicated here

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -57,16 +58,24 @@ export const SaveButton = <RecordType extends RaRecord = any>(
         onClick,
         mutationOptions,
         saving,
-        disabled,
+        disabled: disabledProp,
         type = 'submit',
         transform,
         variant = 'contained',
+        alwaysEnable,
```
fzaninotto, line 65: since you're testing -== false down there, you should give it a default value of false here

PR #7449, opened 2022-03-30
Update yarn.lock

WiXSL:


fzaninotto:
This is not enough, you also need to update the dependencies in the packages.json

packages/ra-data-graphql/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "ra-data-graphql",
-    "version": "4.0.0-beta.2",
```
fzaninotto, line 3: this one either

fzaninotto (changes requested):

djhi (changes requested):
I think the following packages need to be updated as well:
- data-generator
- no-code
- tutorial
- ra-data-simple-rest
- ra-input-rich-text
- ra-ui-materialui

djhi (approved):

PR #7454, opened 2022-03-30
Add `expandSingle` prop to Datagrid

WiXSL:
This prop forces the `Datagrid` to have one row expanded at a time

chrome-capture (1)

### DIFF
--- a/packages/ra-core/src/controller/list/useExpanded.tsx
+++ b/packages/ra-core/src/controller/list/useExpanded.tsx
@@ -45,7 +45,7 @@ export const useExpanded = (
                 ? [id]
                 : [...ids, id];
         });
-    }, [setExpandedIds, id]);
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -37,7 +38,11 @@ export const useExpanded = (
             }
             const index = ids.findIndex(el => el == id); // eslint-disable-line eqeqeq
             return index > -1
-                ? [...ids.slice(0, index), ...ids.slice(index + 1)]
+                ? single
```
fzaninotto, line 41: this creates a linter warning for missing dependency

fzaninotto (changes requested):

fzaninotto (approved):

PR #7455, opened 2022-03-31
Export register resource hook

smeng9:
Some of CRM is able to create resource schema definitions dynamically

We sometimes will need to manually register resource under different components so it would be best to export this hook


fzaninotto (approved):

fzaninotto:
Thanks!

panfiva:
@smeng9 , can you provide an example of how you register resources under different components? Thanks

smeng9:
Hi @panfiva, sorry for my missed message.

We have an endpoint called  `/schemas`, and inside there is a form to allow user to create/edit schemas for resources.

Upon successful submission of the form, we call registerResource to register it to react-admin in onSubmit

PR #7411, opened 2022-03-21
make datagrid row not clickable if rowClick returns false

smeng9:
Resolves issue in #7407

Issue #7407, opened 2022-03-21
[v4b3] Datagrid row cursor CSS


smeng9:
**What you were expecting:**

I hope for datagrid rows with `rowClick="expand"` but `isRowExpandable` evaluated to falsy does not show a `cursor:pointer` on that row, it should show the default cursor

**What happened instead:**

Currently even if the row is not expandable, there is still a pointer, which is confusing for some of user.
Also there seems is not a good way to override this besides pulling datagrid apart by implementing our own datagrid row logic just for a style change.

**Steps to reproduce:**

**Related code:**

* Preferably, a CodeSandbox forked from https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple
* A link to a GitHub repo with the minimal codebase to reproduce the issue

```
<Datagrid
        rowClick="expand"
        expand={<Panel />}
        isRowExpandable={(row) => shouldExpand(row)}   // some of rows is not expandable but still shows a cursor: pointer
>
    {children}
</Datagrid>
```

**Other information:**

Although everything conforms to the `sx` documentation for `& .RaDatagrid-clickableRow`, the user experience may not be very good.

I hope we can change how the CSS API `sx` here https://github.com/marmelab/react-admin/blob/next/docs/Datagrid.md?plain=1#L692 to if rowClick is a truthy string or the function evaluated to be truthy

```
<Datagrid
        rowClick={(id, resource, record) => shouldExpand(record) ? "expand" : null}
        expand={<Panel />}
--     /* isRowExpandable={(row) => shouldExpand(row)} */   // we probably even can deprecate the need for this prop
>
    {children}
</Datagrid>
```

This would require us to change this line https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx#L43 `[DatagridClasses.clickableRow]: rowClick`

So we check if `rowClick` is a function and we call the function to see the return value to determine whether this row is clickable.

In this way we will not show a cursor: pointer for non-clickable row.

**Environment**

* React-admin version: v4b3
* Last version that did not exhibit the issue (if applicable):
* React version: 18.0.0-rc.2
* Browser: Edge 99
* Stack trace (in case of a JS error):

fzaninotto:
Fixed by #7411

WiXSL:
> Also there seems is not a good way to override this besides pulling datagrid apart by implementing our own datagrid row logic just for a style change.

You can do something like this:

```jsx
[...]

PR #7465, opened 2022-03-31
Prevent Datagrid rowClick function to be called on mount 

### LINKED ISSUE #7463: 4.0.0-rc.0 Datagrid rowClick function called on mount
**What you were expecting:**
The rowClick function passed as prop to Datagrid is called only when the user presses a row.

**What happened instead:**
The rowClick function passed as prop to Datagrid is called on mount for each rendered row and when the user presses a row.

**Steps to reproduce:**
1. Open the [v4 codesandbox](https://codesandbox.io/s/github/marmelab/react-admin/tree/next/examples/simple?file=/src/posts/PostList.tsx)
2. Open the /src/posts/PostList.tsx file
3. In the `rowClick` function, add the following : `console.log('rowClick');console.log(record);`
4. Refresh the sandbox and open the console
5. See all the records in the datagrid being printed in the console.

**Environment**

* React-admin version: 4.0.0-rc.0
* Last version that did not exhibit the issue (if applicable): 4.0.0-beta.3
* React version: 17.0.1
* Browser: Firefox

WiXSL:
Fixes #7463
Reverts #7411

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -676,25 +676,25 @@ export const PostList = () => (
 
 The `<Datagrid>` component accepts the usual `className` prop. You can also override many styles of the inner components thanks to the `sx` property. This property accepts the following subclasses:
 
-| Rule name                      | Description                                                     |
-| ------------------------------ | --------------------------------------------------------------- |
-| `& .RaDatagrid-tableWrapper`   | Applied to the div that wraps table element                     |
-| `& .RaDatagrid-table`          | Applied to the table element                                    |
-| `& .RaDatagrid-thead`          | Applied to the table header                                     |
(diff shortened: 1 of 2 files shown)


slax57 (approved):

smeng9:
The previous PR is added to avoid the confusion to user if row is not clickable but renders the cursor to be clickable. https://github.com/marmelab/react-admin/issues/7407 Should we also reopen that issue?

WiXSL:
> The previous PR is added to avoid the confusion to user if row is not clickable but renders the cursor to be clickable. #7407 Should we also reopen that issue?

I answer in that issue how you could do what you are asking for, with styles

smeng9:
Thanks

fzaninotto (approved):

PR #7467, opened 2022-03-31
Prevent call to dataProvider.getMany from useGetManyAggregate if ids is empty

### LINKED ISSUE #7458: Loader keeps spinning after applying filter
**What you were expecting:**

The loader icon on the appbar should stop spinning after the dataProvider answers

**What happened instead:**

The loader icon keeps spinning. it's probably related to #7438

**Steps to reproduce:**

0. launch the simple example (https://codesandbox.io/s/github/marmelab/react-admin/tree/next/examples/simple) *in desktop mode* (because this seems related to the tags column being rendered)
1. Go to the posts list
2. type "lorem" in the search box
3. The loader never stops spinning

**Environment**

* React-admin version: 4.0.0-rc.0
* Last version that did not exhibit the issue (if applicable): 3
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #7458

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
+++ b/packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
@@ -85,7 +85,7 @@ describe('useGetManyAggregate', () => {
         expect(dataProvider.getMany).toHaveBeenCalledTimes(1);
     });
 
-    it.only('should not call dataProvider.getMany() if ids is empty', async () => {
+    it('should not call dataProvider.getMany() if ids is empty', async () => {
         const { rerender } = render(
             <CoreAdminContext dataProvider={dataProvider}>
                 <UseGetManyAggregate resource="posts" ids={[]} />
--- a/packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
+++ b/packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
@@ -85,6 +85,23 @@ describe('useGetManyAggregate', () => {
         expect(dataProvider.getMany).toHaveBeenCalledTimes(1);
     });
 
+    it.only('should not call dataProvider.getMany() if ids is empty', async () => {
+        const { rerender } = render(
+            <CoreAdminContext dataProvider={dataProvider}>
+                <UseGetManyAggregate resource="posts" ids={[]} />
+            </CoreAdminContext>
+        );
+        await new Promise(resolve => setTimeout(resolve));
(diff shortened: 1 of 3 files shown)


fzaninotto:
Shouldn't this be implemented in useGetMany instead?

WiXSL:
> Shouldn't this be implemented in useGetMany instead?

Don't you mean useGetManyAggregate?
Plus, now I noticed this is made inside useReferenceArrayInputController hook, when calling useGetManyAggregate as well.
But you tell me

fzaninotto:
Yes, I meant useGetManyAggregate

slax57 (approved):

packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
```diff
@@ -85,6 +85,23 @@ describe('useGetManyAggregate', () => {
...
     });
 
+    it('should not call dataProvider.getMany() if ids is empty', async () => {
+        const { rerender } = render(
+            <CoreAdminContext dataProvider={dataProvider}>
+                <UseGetManyAggregate resource="posts" ids={[]} />
+            </CoreAdminContext>
+        );
+        await new Promise(resolve => setTimeout(resolve));
+        expect(dataProvider.getMany).toHaveBeenCalledTimes(0);
```
fzaninotto, line 95: I'm curious about the value of data and isLoading in that case. Does that mean that the result is always loading that would be false, as the dataProvider is not even called.

packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -73,6 +73,7 @@ export const useGetManyAggregate = <RecordType extends RaRecord = any>(
     const queryClient = useQueryClient();
     const queryCache = queryClient.getQueryCache();
     const { ids, meta } = params;
+    const { enabled = ids.length > 0, ...restOptions } = options;
```
fzaninotto, line 76: There is still a mystery here: how come the ReferenceManyFieldController calls useGetMany with an empty array?

packages/ra-core/src/dataProvider/useGetManyAggregate.spec.tsx
fzaninotto: Also, if there are more than one request to useGetManyAggregate and one has an empty array, does it block the aggregation?

fzaninotto:
I think using `enabled` is the wrong fix. The hook should return an empty array directly when receiving an empty list of ids. I made another PR in #7477, reusing your tests, that supersedes this PR. 

Thanks for showing me the path!

PR #7468, opened 2022-04-01
[Datagrid] Fix non clickable row can't be styled

### LINKED ISSUE #7407: [v4b3] Datagrid row cursor CSS

fzaninotto:
We're adding classes to each datagrid row to let developers style them differently whether they're clickable, selectable, expandable, or not. 

Ideally, we should change the cursor by default when a row isn't clickable, selectable, or expandable, but this is for another PR.

Closes #7407

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
@@ -40,7 +40,6 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                         className: clsx(DatagridClasses.row, {
                             [DatagridClasses.rowEven]: rowIndex % 2 === 0,
                             [DatagridClasses.rowOdd]: rowIndex % 2 !== 0,
-                            [DatagridClasses.clickableRow]: rowClick,
                         }),
                         expand,
                         hasBulkActions: hasBulkActions && !!selectedIds,
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
@@ -143,7 +143,7 @@ const DatagridRow: FC<DatagridRowProps> = React.forwardRef((props, ref) => {
                     [DatagridClasses.expandable]: expandable,
                     [DatagridClasses.selectable]: selectable,
                     [DatagridClasses.clickableRow]:
-                        typeof rowClick === 'function' ? false : rowClick,
+                        typeof rowClick === 'function' ? true : rowClick,
                 })}
                 key={id}
                 style={style}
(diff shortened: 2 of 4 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
```diff
@@ -40,7 +40,6 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                         className: clsx(DatagridClasses.row, {
                             [DatagridClasses.rowEven]: rowIndex % 2 === 0,
                             [DatagridClasses.rowOdd]: rowIndex % 2 !== 0,
-                            [DatagridClasses.clickableRow]: rowClick,
```
fzaninotto, line 43: moved to DatagridRow

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -139,7 +139,12 @@ const DatagridRow: FC<DatagridRowProps> = React.forwardRef((props, ref) => {
         <RecordContextProvider value={record}>
             <TableRow
                 ref={ref}
-                className={className}
+                className={clsx(className, {
+                    [DatagridClasses.expandable]: expandable,
+                    [DatagridClasses.selectable]: selectable,
+                    [DatagridClasses.clickableRow]:
+                        typeof rowClick === 'function' ? false : rowClick,
```
slax57, line 146: I believe this should rather be

```suggestion
                        typeof rowClick === 'function' ? true : rowClick,
```

slax57 (changes requested):

slax57 (approved):

WiXSL (approved):

PR #7475, opened 2022-04-01
Add <Admin requireAuth> to hide the app until auth is checked

### LINKED ISSUE #5229: Allow Resources to wait for checkAuth resolve before rendering
This is a similar issue as the one reported here https://github.com/marmelab/react-admin/issues/4006
The gif image in that issue is still applicable (except no data is actually displayed - just the skeleton of the dashboard)

**What you were expecting:**
When user opens react-admin for the first time, the dashboard should display the login page

**What happened instead:**
When user opens react-admin for the first time, the dashboard is displayed momentarily (without data) until the checkAuth is rejected, which is when user is then redirected to login page

I looked at the demo app provided and noticed that there is a workaround of setTimeout of 500ms to avoid this problem, 
https://github.com/marmelab/react-admin/blob/master/examples/demo/src/dataProvider/rest.ts

But that does not feel correct - is there a way to disable the optimistic rendering for the Admin (i.e. don't display dashboard and then logout after checkAuth fails, but simply just don't display it until it resolves true?)

I'm using latest version 3.8.3 together with react-admin-firebase

Thanks

fzaninotto:
## Problem

Auth check is optimistic by default.  In v3, one could turn auth check to pessimistic (i.e. display nothing until `dataProvider.checkAuth()` resolves) by implementing a custom `<Resource>`. But in v4, the authentication isn't in `<Resource>` anymore, but in the page controllers. As a consequence, it is not possible to make auth pessimistic. 

## Solution

Add an option to `<Admin>` to make the app require authentication. 

This is something we must do before releasing v4, because otherwise it would prevent an entire class of features (and break ra-rbac).

Closes #5229

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -43,6 +43,7 @@ Here are all the props accepted by the component:
 - [`history`](#history)
 - [`basename`](#basename)
 - [`ready`](#ready)
+- [`requireAuth`](#requireAuth)
 - [`store`](#store)
 
 ## `dataProvider`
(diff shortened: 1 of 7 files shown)


djhi (approved):
:+1:

slax57 (approved):

jeremydrichardson:
I'm wondering if this may have affected the ability to access custom routes anonymously.

Issue here - https://github.com/marmelab/react-admin/issues/7489

PR #7491, opened 2022-04-06
Fix custom routes not accessible by anonymous users

### LINKED ISSUE #7489: Custom routes not accessible by anonymous users
**What you were expecting:**
Custom routes would be accessible by users who are _not_ logged in.

**What happened instead:**
When trying to access a custom route directly, it automatically redirects you back to the login page unless you are logged in.

**Steps to reproduce:**
Enter the url of the custom page directly in the url bar when not logged in. eg. `/#/mypage`

**Related code:**
https://codesandbox.io/s/react-adminv4-customroute-noaccess-h5sdg6

```
<CustomRoutes noLayout>
  <Route path="/mypage" element={<MyPage />} />
</CustomRoutes>
```

**Other information:**

**Environment**

* React-admin version: 4.0.0-rc.1
* Last version that did not exhibit the issue (if applicable): 4.0.0-rc.0
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #7489

### DIFF
--- a/packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
+++ b/packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
@@ -346,5 +346,40 @@ describe('<CoreAdminRoutes>', () => {
                 expect(screen.queryByText('Login')).not.toBeNull()
             );
         });
+        it('should render custom routes when the user is not authenticated and requireAuth is false', async () => {
+            const authProvider = {
+                login: jest.fn().mockResolvedValue(''),
+                logout: jest.fn().mockResolvedValue(''),
+                checkAuth: () => Promise.reject('Not authenticated'),
+                checkError: jest.fn().mockResolvedValue(''),
+                getPermissions: jest.fn().mockResolvedValue(''),
+            };
+
+            const history = createMemoryHistory();
+            render(
+                <CoreAdminContext
+                    authProvider={authProvider}
+                    dataProvider={testDataProvider()}
+                    history={history}
+                >
+                    <CoreAdminRoutes
+                        layout={Layout}
+                        loading={Loading}
+                        catchAll={CatchAll}
+                    >
+                        <CustomRoutes noLayout>
+                            <Route path="/custom" element={<i>Custom</i>} />
+                        </CustomRoutes>
+                        <Resource name="posts" list={() => <i>PostList</i>} />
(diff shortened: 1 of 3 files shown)


slax57 (commented):
I approve theses changes 🙂 

But shouldn't we add a test case to cover this in `CoreAdminRoutes.spec.tsx` as well?

packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
```diff
@@ -346,5 +346,40 @@ describe('<CoreAdminRoutes>', () => {
...
+                        <Resource name="posts" list={() => <i>PostList</i>} />
+                    </CoreAdminRoutes>
+                </CoreAdminContext>
+            );
+            expect(screen.queryByText('PostList')).not.toBeNull();
+            expect(screen.queryByText('Loading')).toBeNull();
+            history.push('/custom');
+            await waitFor(() =>
+                expect(screen.queryByText('Custom')).not.toBeNull()
+            );
```
slax57, line 382:
```suggestion
                    <CoreAdminRoutes
                        layout={Layout}
                        loading={Loading}
                        catchAll={CatchAll}
                    >
                        <CustomRoutes noLayout>
                            <Route path="/custom" element={<i>Custom</i>} />
                            <Route path="/login" element={<i>Login</i>} />
                        </CustomRoutes>
                        <Resource name="posts" list={() => <i>PostList</i>} />
                    </CoreAdminRoutes>
                </CoreAdminContext>
            );
            expect(screen.queryByText('PostList')).not.toBeNull();
            expect(screen.queryByText('Loading')).toBeNull();
            history.push('/custom');
            await new Promise(resolve => setTimeout(resolve, 1100));
            await waitFor(() =>
                expect(screen.queryByText('Custom')).not.toBeNull()
            );
```

I suggest the above changes, otherwise, the test is passing even without the fix

slax57 (changes requested):

packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
WiXSL: Nice catch!
Thanks

slax57 (approved):

fzaninotto (approved):

PR #7505, opened 2022-04-08
Doc fix: remove render prop from new Form component examples

### LINKED ISSUE #7504: render prop on <Form> seems missing
When working on upgrading from V3 to V4, I came across this in the upgrade guide:

[](https://marmelab.com/react-admin/doc/4.0/Upgrade.html#formwithredirect-render-function-arguments-have-changed)

- this explains that the arguments to the render() function has changed.
However, it seems like the render() function is completely removed from the <Form> component. And that it instead uses child components.

Indeed, the form displays nothing when using the render() function, but works as expected when placing the contents as childs of <Form>

This is on the v4.0.0-rc.1

slax57:
Remove `render` prop from the new `Form` component examples in the v4 migration guide.

Fixes #7504

### DIFF
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -1816,16 +1816,53 @@ The form components don't handle redirection anymore, as redirections are now ha
 
 To upgrade, replace all occurrences of `<FormWithRedirect>` with  `<Form>`.
 
+Also, the `render` prop has been removed, and the form content is now passed as child of the `<Form>` element.
+
 ```diff
 - import { FormWithRedirect } from 'react-admin';
 + import { Form } from 'react-admin';
 
 export const MyForm = () => (
 -    <FormWithRedirect
-+    <Form
-        render={() => ...}
-    />
+-        render={() => ...}
+-    />
++    <Form>
++        ...
++    </Form>
 );
+```
+
+If you need to access the form state (`valid`, `invalid`, `pristine`, `dirty`), you can call [the react-hook-form `useFormState` hook](https://react-hook-form.com/api/useformstate) instead:
+
+```diff
+import { FormWithRedirect } from 'react-admin';
++ import { Form } from 'react-admin';
++ import { useFormState } from 'react-hook-form';
+
+const MyCustomForm = () => {
+    return (
+-        <FormWithRedirect
++        <Form>
+-            render={({ valid, dirty, handleSubmit }) => (
+-               <form onSubmit={handleSubmit}>
+                    ...
+-                    <SubmitButton disabled={!dirty || !valid}>Save</SubmitButton>
++                    <SubmitButton>Save</SubmitButton>
+-               </form>
+            )}
+-       />
++       </Form>
+    );
+};
+
+-const SubmitButton = ({ disabled, ...props }) => {
++const SubmitButton = (props) => {
++    const { isDirty, isValid } = useFormState();
+    return (
+-        <button disabled={disabled} {...props} />
++        <button disabled={!isDirty || !isValid} {...props} />
+    );
+}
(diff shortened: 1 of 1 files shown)


docs/Upgrade.md
```diff
@@ -2121,42 +2123,32 @@ If you relied on the `handleSubmitWithRedirect` to change the redirection:
...
-+        event.preventDefault(); // Prevent the default form submission
-+        const values = form.getState().values;
-+        save(values, {
-+            onSuccess: (data) => redirect('show', '/posts', data.id)
-+        })
-+    };
     return (
 -        <button onClick={() => handleSubmitWithRedirect('show')}>Save</button>
-+        <button type="button" onClick={handleClick}>Save</button>
++        <button type="submit">Save</button>
```
fzaninotto, line 2145: I don't understand this change. This SaveButton has the defautl behavior - it's not customized

docs/Upgrade.md
```diff
@@ -2121,42 +2123,32 @@ If you relied on the `handleSubmitWithRedirect` to change the redirection:
...
     );
 );
 ```
 
+You can also use a function to handle the redirection target, as described in [this section](./EditTutorial.md#redirection-after-submission).
```
fzaninotto, line 2150:
```suggestion
You can also use a function to handle the redirection target, as described in [the redirection after submittion documentation](./EditTutorial.md#redirection-after-submission).
```

Don't use 'this thing' as support of a hyperlink, but rather the title of the this you link to.

docs/Upgrade.md
```diff
@@ -1816,18 +1816,55 @@ The form components don't handle redirection anymore, as redirections are now ha
...
++       </Form>
+    );
+};
+
+-const SubmitButton = ({ disabled, ...props }) => {
++const SubmitButton = (props) => {
++    const { isDirty, isValid } = useFormState();
+    return (
+-        <button disabled={disabled} {...props} />
++        <button disabled={!isDirty || !isValid} {...props} />
```
fzaninotto, line 1863: this doesn't work - you need at least a type="submit"

docs/Upgrade.md
```diff
@@ -1816,18 +1816,55 @@ The form components don't handle redirection anymore, as redirections are now ha
...
 
 export const MyForm = () => (
 -    <FormWithRedirect
-+    <Form
-        render={() => ...}
-    />
+-        render={() => ...}
+-    />
++    <Form>
++        ...
```
fzaninotto, line 1830: this isn't explicit enough. You should replace the `...` in the diff by a dummy component to make the change clearer

fzaninotto (changes requested):

docs/Upgrade.md
slax57: This example was about showing how to use `handleSubmitWithRedirect` to redirect to the **show** page after submission.

So in the new example I encouraged to make the following change:
```diff
-        <Create>
+        <Create redirect="show">
```
And then you don't need to customize the Save button anymore.

What is troubling you? Did you miss the `redirect="show"` I added, or is it because the `redirect` prop is more limited than what you could do with `handleSubmitWithRedirect` and you would rather have me add a more complete example?

docs/Upgrade.md
slax57: Isn't `type="submit"` the default for a button? I'll test it...

slax57: I confirm it's working without `type="submit"`.

Do you want me to add it anyway? In this case, should I also add it to the "before" line, to make it clear that the diff is not about this?

fzaninotto (approved):

PR #7507, opened 2022-04-08
Fix premium image shadow on mobile

WiXSL:
Before
before

After
after


slax57 (approved):
Love those 1-line fixes 😍

fzaninotto (approved):

PR #7510, opened 2022-04-08
Fix list actions don't trigger a refetch

### LINKED ISSUE #7494: React 18 strict mode breaks URL query params
**What you were expecting:**

That react admin works normally with React 18 strict mode

**What happened instead:**

When I click on a column header to sort the url doesn't update with query params, so nothing happens.

**Steps to reproduce:**

In the code sandbox try to sort a column, see it doesn't work. In index.js, comment out the React 18 section and uncomment the React 17 mode section, sorting should work now. 

**Related code:**

https://codesandbox.io/s/sweet-fast-1pkgk5?file=/src/index.js

React 18 mode (broken):

```
import App from "./App";
import { StrictMode } from "react";
import * as ReactDOMClient from "react-dom/client";

const rootElement = document.getElementById("root");
const root = ReactDOMClient.createRoot(rootElement);

root.render(
  <StrictMode>
    <App />
  </StrictMode>
);
```

React 17 mode (working):

```
import App from "./App";
import { render } from "react-dom";
const container = document.getElementById("root");
render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
  container
);
```

**Environment**

* React-admin version: 4
* React version: 18
* Browser: Latest Chrome and Safari

fzaninotto:
Closes #7494

### DIFF
--- a/packages/ra-core/src/util/hooks.ts
+++ b/packages/ra-core/src/util/hooks.ts
@@ -73,6 +73,7 @@ export function useTimeout(ms = 0, key = '') {
 export function useIsMounted() {
     const isMounted = useRef(true);
     useEffect(() => {
+        isMounted.current = true;
         return () => {
             isMounted.current = false;
         };


packages/ra-core/src/util/hooks.ts
```diff
@@ -73,6 +73,7 @@ export function useTimeout(ms = 0, key = '') {
 export function useIsMounted() {
     const isMounted = useRef(true);
```
WiXSL, line 74: Shouldn't be initialized with `false` ?

WiXSL (changes requested):

packages/ra-core/src/util/hooks.ts
WiXSL, line 74: Shouldn't be initialized with `false` ?

WiXSL (changes requested):

packages/ra-core/src/util/hooks.ts
fzaninotto: No, as it would trigger a useless re-render

WiXSL (approved):

djhi (approved):

PR #7519, opened 2022-04-11
Fix SelectInput / SelectArrayInput onChange handler

### LINKED ISSUE #7518: [v4.0.0-rc.1] SelectInput onChange type incorrect
**What you were expecting:**
The typing for the onChange handler for SelectInput is shown as:
React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement> | undefined

This should return an event of type: React.ChangeEvent<HTMLInputElement | HTMLTextAreaAlement>

**What happened instead:**
A string is returned with the value that the select was changed to.

**Steps to reproduce:**
While using Typescript, add a SelectInput to an Edit -> SimpleForm and then add an onChange handler. Typescript marks onChange as above, but the return value doesn't match.

**Related code:**

```
const countryChanged: ChangeEventHandler<HTMLInputElement> = (ev) => {
  console.log("ev); // e.g. "UK"
  console.log(typeof ev); // string
};

<Edit>
  <SimpleForm>
	  <SelectInput
		  label="Country"
		  source="contactAddress.countryName"
		  choices={countries}
		  optionText="title"
		  optionValue="id"
		  onChange={countryChanged}
	  />
  </SimpleForm>
</Edit>
```

Happy to fix and do a PR, but not sure whether the typing is incorrect and a string should be returned, or whether the the typing is correct and an event should be returned.

**Environment**

* React-admin version: v4.0.0-rc.1
* React version: 17.0.2
* Browser: Edge v100.0.1185.36

WiXSL:
Fixes #7518

SelectInput / SelectArrayInput  change event handler should receive the event object unless the change was fired due to the creation of a new item, in that case it should receive the new value

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
@@ -145,7 +145,6 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
 
     const handleChange = useCallback(
         (eventOrChoice: ChangeEvent<HTMLInputElement> | RaRecord) => {
-            console.log(eventOrChoice);
             // We might receive an event from the mui component
             // In this case, it will be the choice id
             if (eventOrChoice?.target) {


fzaninotto:
Also, it seems the accepted `onChange` arguments are fuzzy: event or value? We should choose one, and expect it everywhere. I don't mind if we support both for BC, but the types (and documentation) must be clear about the idiomatic way.

WiXSL:
> I approve the changes, but I have a few questions:
> 
> * Are `SelectArrayInput` and `SelectInput` the only components affected by this? What about `TextInput` and the others for example?

The only other inputs that do not pass the event object are `AutocompleteInput` and `AutocompleteArrayInput` (that I know of)
  According to @djhi, this comes from problems with final-form, maybe It could be improved now that we use react-hook-form 

> * Why do we need to do a difference with what we send to the `onChange` callback when we add an item vs when we select another? Is it documented somewhere?

As of this PR, my intention was just to maintain the functionality of `master` preventing a BC.
But I think the answer is in the fact that you can create new items, and in that case you don't have the value in the event object.
Maybe this can be changed, to pass two arguments, the event object and the new item to the onChange event handler.

I don't think all of this behavior is properly documented.

WiXSL:
> Also, it seems the accepted `onChange` arguments are fuzzy: event or value? We should choose one, and expect it everywhere. I don't mind if we support both for BC, but the types (and documentation) must be clear about the idiomatic way.

Ok, so, to be clear, do you want me to use this PR to change the onChange event handlers of these inputs to always pass the event object?
In the case of newly created items, a second arguments must be used then, since the event object doesn't include this new value

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -143,12 +144,12 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
     });
 
     const handleChange = useCallback(
-        (eventOrChoice: any) => {
+        (eventOrChoice: ChangeEvent<HTMLInputElement> | RaRecord) => {
+            console.log(eventOrChoice);
```
slax57, line 148: console.log spotted

slax57 (changes requested):

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
slax57: otherwise I'm happy with the code 😅

PR #7524, opened 2022-04-12
docs(Admin): document i18nprovider and store for AdminContext

### LINKED ISSUE #7520: v4: custom <Admin /> removes all labels (i18n?)
When using the AdminContext instead of `<Admin />` all labels change into what I assume are keys.

Upon checking the code [I think this is the issue](https://github.com/marmelab/react-admin/blob/v4.0.0-rc.1/packages/react-admin/src/Admin.tsx#L146,L149):

```js
Admin.defaultProps = {
    i18nProvider: defaultI18nProvider,
    store: localStorageStore(),
};
```

These aren't mentioned in the docs. I think this should be added to the docs, or should be moved into `<AdminContext />`.

**What you were expecting:**

I was expecting nothing to change and everything to look the same.

**What happened instead:**

After following this guide: https://marmelab.com/react-admin/doc/4.0/Admin.html#unplugging-the-admin-using-admincontext-and-adminui

I got the following result:

image

**Related code:**

My <App /> looks like this now (removed most resources and the irrelevant bits for brevity):

```tsx
function App() {
  return (
    <AdminContext authProvider={authProvider} dataProvider={withUpload}>
      <AsyncResources />
    </AdminContext>
  );
}

function AsyncResources() {
  return (
    <AdminUI layout={MyLayout} loginPage={SignIn} disableTelemetry={true}>
      <Resource name="profiles" list={ProfileList} show={ProfileShow} edit={ProfileEdit} create={ProfileCreate} />
      <Resource name="users" list={UserList} show={UserShow} edit={UserEdit} create={UserCreate} />
    </AdminUI>
  );
}
```

**Environment**

* React-admin version: 4.0.0-rc.1
* React version: 17.0.2
* Browser: Chrome 

RWOverdijk:
Update docs & docblock.

fixes #7520

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -488,16 +488,26 @@ Setting Resources dynamically using the children-as-function syntax may not be e
 
 So it's impossible, for instance, to have a dynamic list of resources based on a call to the `dataProvider` (since the `dataProvider` is only defined after the `<Admin>` component renders).
 
-To overcome this limitation, you can build your own `<Admin>` component using two lower-level components: `<AdminContext>` (responsible for putting the providers in contexts) and `<AdminUI>` (responsible for displaying the UI). Here is an example:
+To overcome this limitation, you can build your own `<Admin>` component using two lower-level components: `<AdminContext>` (responsible for putting the providers in contexts) and `<AdminUI>` (responsible for displaying the UI). Through this approach you'll have to bring your own i18n provider and store. Luckily react-admin provides easy to use defaults for you. Here is an example:
 
 ``` jsx
 import * as React from 'react';
 import { useEffect, useState } from 'react';
-import { AdminContext, AdminUI, Resource, ListGuesser, useDataProvider } from 'react-admin';
+import {
+    AdminContext,
+    AdminUI,
+    defaultI18nProvider,
+    localStorageStore,
+    Resource,
+    ListGuesser,
+    useDataProvider,
+} from 'react-admin';
+
+const store = localStorageStore();
 
 function App() {
     return (
-        <AdminContext dataProvider={myDataProvider}>
+        <AdminContext dataProvider={myDataProvider} i18nProvider={defaultI18nProvider} store={store}>
             <AsyncResources />
         </AdminContext>
     );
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

fzaninotto:
Thanks!

RWOverdijk:
@fzaninotto I did find that changing auth status doesn't reload the resources. The render gets called but navigation doesn't change. Mentioning this here before going through the issue process.

For example, before logging in all my resources return null (no permission, because I don't know who you are). After logging in I get the [`ready` component](https://marmelab.com/react-admin/doc/4.0/Admin.html#ready) because resources don't get registered.

Code helps, so here's my implementation:

```ts
function App() {
  return (
    <AdminContext
      authProvider={authProvider}
      dataProvider={withUpload}
      i18nProvider={defaultI18nProvider}
      store={store}
    >
      <GuardedResources />
    </AdminContext>
  );
}

function GuardedResources() {
  const { isLoading, guard } = usePermissionGuard();
  const composedResources = useMemo(() => {
    if (isLoading) return null;

    return resources.map(({ permissionResource: r, components, ...props }) => {
      if (!guard.isAllowed(r)) return null;

      return <Resource key={`r_${r}`} {...guard.filterComponents(r, components)} {...props} />;
    });
  }, [isLoading, guard]);
  
  return (
    <AdminUI layout={MyLayout} loginPage={SignIn} disableTelemetry={true} requireAuth>
      {composedResources}
    </AdminUI>
  );
}

export default App;
```

`isAllowed` is linked to the custom permission system which is simply a map of resources and actions. The `composedResources` memo gets called but the UI doesn't update and `ready` still shows up.

------

TL;DR; the resources don't get reloaded when logging in causing me to see the Admin.ready component until I refresh the page. Is this a bug or just shitty implementation on my side?

fzaninotto:
@RWOverdijk I'm not sure I understand the problem. As far as I know, the current implementation of dynamic resources based on permissions works (it's backed by unit tests). If you can provide a repro based on the simple example showing that it doesn't work, please open a new issue about it.

RWOverdijk:
Will do.

PR #7526, opened 2022-04-12
Allow arrays as field values in validator resolver

### LINKED ISSUE #7500: RA v4 blocks validation due to empty array
**What you were expecting:**
Before RA v4, returning an empty array from the validator meant that RA treated the form as being valid. After the new update it seems that this is no longer the case, and we get a popup saying that form isn't valid.

Not sure if this is a bug, or if it's new behaviour that needs to be documented. 

You can test this out [here](https://codesandbox.io/s/quizzical-chatterjee-7cozbr?file=/src/posts/PostCreate.tsx)

You can enter some text into the array field. 

If the value is less than 2 characters, it will add an array to the backlinks' error array. if it's larger, then it'll just return the empty array.

What do you suggest, is this a bug or just a change in behaviour?

**What happened instead:**

**Steps to reproduce:**

**Related code:**

* Preferably, a CodeSandbox forked from https://codesandbox.io/s/github/marmelab/react-admin/tree/next/examples/simple

```
```

**Other information:**

**Environment**

* React-admin version:
* React version:
* Browser:

WiXSL:
Fixes #7500

### DIFF
--- a/packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
+++ b/packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
@@ -1,0 +1,66 @@
+import { getSimpleValidationResolver } from './getSimpleValidationResolver';
+
+describe('getSimpleValidationResolver', () => {
+    const validator = getSimpleValidationResolver(errors => errors);
+
+    it('should recieve a flattened errors object', async () => {
+        const errors = await validator({
+            title: 'title too short',
+            backlinks: [
+                { url: 'url too short', id: 'missing id' },
+                { url: 'url too short', id: 'missing id' },
+            ],
+        });
+
+        expect(errors).toEqual({
+            values: {},
+            errors: {
+                title: { type: 'manual', message: 'title too short' },
+                'backlinks.0.url': {
+                    type: 'manual',
+                    message: 'url too short',
+                },
+                'backlinks.0.id': {
+                    type: 'manual',
+                    message: 'missing id',
+                },
+                'backlinks.1.url': {
+                    type: 'manual',
+                    message: 'url too short',
+                },
+                'backlinks.1.id': {
+                    type: 'manual',
+                    message: 'missing id',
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/form/getSimpleValidationResolver.ts
```diff
@@ -30,15 +44,16 @@ export const getSimpleValidationResolver = (validate: ValidateForm) => async (
     if (!errors || Object.getOwnPropertyNames(errors).length === 0) {
         return { values: data, errors: {} };
     }
+    const flattenErrors = flattenKeys(errors);
```
fzaninotto, line 47: I'd name the variable `flattenedErrors` (otherwise it starts with a verb and looks like a callback)

packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { getSimpleValidationResolver } from './getSimpleValidationResolver';
+
+describe('getSimpleValidationResolver', () => {
+    const validator = getSimpleValidationResolver(errors => errors);
```
fzaninotto, line 4: this is misleading: the parameter passed to the resolver callback parameter should be named values, not errors

packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    it('should recieve a flattened errors object', async () => {
+        const errors = await validator({
```
fzaninotto, line 7: that's really confusing: validator should receive an object of values, and here the tests makes it look like it expects errors... Could you find a way to test it that it clearer? Perhaps by unit testing the flattenKeys function

fzaninotto (changes requested):

fzaninotto (approved):

PR #7537, opened 2022-04-13
Fix default NumberInput parser

### LINKED ISSUE #7530: NumberInput does not handle zero after comma correctly
**What you were expecting:**
Being able to input a number like "2.03" inside a NumberInput

**What happened instead:**
The value gets transformed to "23"

**Steps to reproduce:**
With the basic codesandbox: https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple
Edit a Post, and in the "Average note" field type in "2.03"

**Related code:**
Problem comes from NumberInput's default parse function:

```ts
const convertStringToNumber = value => {
    const float = parseFloat(value);

    return isNaN(float) ? null : float;
};
```

According to [this](https://stackoverflow.com/questions/62564304/javascript-cannot-enter-zeros-after-a-decimal-point), the use of `parseFloat` is erroneous and needs to be replaced by `parseFloat(prevValStr.concat(value)).toFixed(numbersAfterDecimal)`

**Environment**

* React-admin version: 3.19.11 and 4.0.0-rc.1
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #7530

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.tsx
@@ -118,8 +118,11 @@ export interface NumberInputProps
 
 const convertStringToNumber = value => {
     const float = parseFloat(value);
+    if (isNaN(float)) return null;
 
-    return isNaN(float) ? null : float;
+    return value.indexOf('.') === -1
+        ? float
+        : float.toFixed(value.split('.')[1].length);
 };
 
 const convertNumberToString = value => {


slax57 (commented):
Love it ❤️ 

Maybe we should add a test case covering it?

fzaninotto:
I'm not convinced it's the right fix. n fact, we can't foresee all the possible formats. 

I see two fixes:

1. only format on blur
2. use [the `valueAsNumber` attribute](https://www.w3.org/TR/2011/WD-html5-20110525/common-input-element-attributes.html#dom-input-valueasnumber) of the native Input element (but I'm not sure about browser support)

afilp:
Thanks! 
But will this work for a locale that uses "," instead of "." for decimals? Doesn't seem so.

WiXSL:
> I'm not convinced it's the right fix. n fact, we can't foresee all the possible formats.
> 
> I see two fixes:
> 
> 1. only format on blur
> 2. use [the `valueAsNumber` attribute](https://www.w3.org/TR/2011/WD-html5-20110525/common-input-element-attributes.html#dom-input-valueasnumber) of the native Input element (but I'm not sure about browser support)
>    Edit: browser support seems to be good: https://caniuse.com/?search=valueasnumber

1- Mmm, this seems tricky, how I could go about this?
2- I've tried using `valueAsNumber` but I get the same problem described in the bug report.

afilp:
Ok, this may work for all cases I think:

We first convert any "number string" to "en-US" locale, for which we know that "." is the decimal sign (and "," the grouping sign):

```
var valueFormatted = new Intl.NumberFormat('en-US', numberFieldOptions).format(value)
```

Then we remove all "," (grouping), and use parseFloat safely:

```
parseFloat(valueFormatted.replace(",",""))
```

fzaninotto:
Superseded by #7673

PR #7539, opened 2022-04-13
Fix logout doesn't reset resource registration

### LINKED ISSUE #7527: v4: resources not reloading after login or auth change
**What you were expecting:**

When logging in I expect the resources to reload. The reason for this is that the auth state change comes with a permissions change, which decides which resources the user can view.

**What happened instead:**

The ready screen:

image

**Steps to reproduce:**

I followed this: https://marmelab.com/react-admin/doc/4.0/Admin.html#declaring-resources-at-runtime

It's really pretty simple to reproduce. Create a custom AdminContext and based on the auth state either render null or the resources (pretending you don't have permissions to see the other pages).

I [made a sandbox](https://codesandbox.io/s/strange-greider-8eieup) based on the example in the repo (bnase sandbox: https://codesandbox.io/s/github/marmelab/react-admin/tree/next/examples/simple) but it's pretty difficult to get it to run on codesandbox (a dependency issue somewhere?). What I did:

1. Download the sandbox
2. delete vite.config.js
3. npm i --force (force is needed depending on your node version, there's a conflict somewhere in the tree)
4. npm run start

Then, when in the browser follow these steps:

1. You have to log out
2. You have to hard-refresh the browser before logging in
3. Log in
4. You see the screen from my screenshot up 👆

This is an example. I tried other things like forcing a re-render. It does rerender, it just doesn't register the resources.

Video of the issue demonstrated:

https://user-images.githubusercontent.com/781745/162996302-5d81823e-ed3b-43ab-b0c0-667ce52a32d0.mov

**Related code:**

https://codesandbox.io/s/strange-greider-8eieup

Specifically in index.js:

```tsx
    if (authLoading || !authenticated) return null;

    return (
      <>
        <Resource name="comments" {...comments} />
        <Resource name="users" {...users} />
        <Resource name="tags" {...tags} />
      </>
    );
```

This is the most basic example of what I'm _actually_ trying to do which is load in permissions async and show resources based on them. Users are allowed to change their role while logged in (different profiles, like for example netflix). But something as basic as logging in already doesn't work and properly illustrates my issue I think.

**Environment**

* React-admin version: react-admin@4.0.0-rc.1
* React version: 17.0.2
* Browser: Chrome

fzaninotto:
Closes #7527 and #7586

- [x] Unregister resource definitions when resources change
- [x] Cover with unit tests


slax57 (approved):

WiXSL (changes requested):
I can still reproduce the bug described in https://github.com/marmelab/react-admin/issues/7527 locally, in the simple project. Just copy the `index.tsx` code of the proposed sandbox

fzaninotto:
After thinking it through, I believe we should unregister resources when unmounting `<Resource>` rather than on logout. Because `<Admin>` accepts a function as child, the resources can vary according to external state.

fzaninotto:
I made sure the last test doesn't pass if we don't unregister the resource. The PR is ready for review.

fzaninotto:
> I can still reproduce the bug described in #7527 locally, in the simple project. Just copy the `index.tsx` code of the proposed sandbox to the simple project's `index.tsx`

I've tested it and the ready screen only shows up briefly. I've updated the docs and code to prevent it.

docs/Admin.md
```diff
@@ -533,4 +532,4 @@ function AsyncResources() {
 }
 ```
 
-In this example, we override the `<AdminUI ready>` component to prevent the admin from displaying [the ready screen](#ready) while the list of resources is loading.
+In this example, we override the `<AdminUI ready>` component to prevent the admin from displaying [the ready screen](#ready) in development while the list of resources is loading. You can also set `ready` prop to `<Loading>` to show that the resources afe loading.
```
WiXSL, line 535:
```suggestion
In this example, we override the `<AdminUI ready>` component to prevent the admin from displaying [the ready screen](#ready) in development while the list of resources is loading. You can also set `ready` prop to `<Loading>` to show that the resources are loading.
```

docs/Admin.md
```diff
@@ -524,7 +523,7 @@ function AsyncResources() {
     }, []);
 
     return (
-        <AdminUI ready={Loading}>
+        <AdminUI ready={() => <span />}>
```
WiXSL, line 526: I don't get why you change it by a `span`, It seems to work with the `Loading` component

WiXSL (changes requested):

WiXSL (approved):

djhi (approved):
:fire:

PR #7545, opened 2022-04-14
Fix the link to the ra-rbac module

artnest:
Current link (https://marmelab.com/ra-rbac) leads to a page that doesn't exist.

`GET https://marmelab.com/ra-rbac 404`

**NOT FOUND**
You just hit a route that doesn't exist... the sadness.


fzaninotto:
Thanks!

PR #7548, opened 2022-04-14
Fix duplicate key error

### LINKED ISSUE #7541: ArrayField + Datagrid: if embedded data doesn't have an `id`, rows won't have a `key` and will cause a React warning
**What you were expecting:**
Using an ArrayField + Datagrid for embedded data that doesn't have an `id` prop should work without console errors

**What happened instead:**
I get this error in the console: `Warning: Each child in a list should have a unique "key" prop.`

**Steps to reproduce:**
Given this Show component for an "album" record:

```jsx
export const AlbumShow = () => (
  <Show>
    <SimpleShowLayout>
      <ArrayField source="features">
        <Datagrid>
          <TextField source="name" />
          <TextField source="value" />
        </Datagrid>
      </ArrayField>
    </SimpleShowLayout>
  </Show>
);
```

Where an album has embedded features that look like this (note that `features` don't have an `id` prop):
```json
{
  "id": 0,
  "features": [
    { "name": "title", "value": "Test Album" }
  ]
}
```

**Related code:**
You can see repro this in the [simple example](https://github.com/marmelab/react-admin/tree/master/examples/simple) (checked out and run locally; I don't see warnings in codesandbox's console)
Go to a post show page that has backlinks, e.g. http://localhost:8080/#/posts/12/show

**Other information:**
This is where the key is set based on the id, which doesn't exist in this case: https://github.com/marmelab/react-admin/blob/a408de0db2c6db0281b3d39e27c1f2de3deed44e/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx#L48

**Environment**

* React-admin version: 4.0.0
* React version: 18.0.0
* Browser: Chrome

WiXSL:
Fixes #7541

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
@@ -45,7 +45,7 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                         hasBulkActions: hasBulkActions && !!selectedIds,
                         hover,
                         id: record.id,
-                        key: record.id,
+                        key: `row${rowIndex}`,
                         onToggleItem,
                         record,
                         resource,


packages/ra-ui-materialui/src/list/datagrid/DatagridBody.tsx
```diff
@@ -45,7 +45,7 @@ const DatagridBody: FC<DatagridBodyProps> = React.forwardRef(
                         hasBulkActions: hasBulkActions && !!selectedIds,
                         hover,
                         id: record.id,
-                        key: record.id,
+                        key: `row${rowIndex}`,
```
fzaninotto, line 48: I think that isn't a good idea for the 90% of the cases where we actually render records. You should only fall back to rowIndex if the record id is undefined.

fzaninotto (changes requested):

fzaninotto (approved):

PR #7551, opened 2022-04-14
Fix parsed date/date-time filter resets component

### LINKED ISSUE #7540: DateTimeInputs not working correctly as Filters
**What you were expecting:**
DateTimeInput components to work correctly as list filter components

**What happened instead:**
Month and date/time values do not register properly when values change, and often reset entirely while picker is active.

**Steps to reproduce:**
Add a DateTimeInput as a Filter component, then try to use the date picker to select a value.

**Related code:**
https://codesandbox.io/s/cool-sound-pdlsh4 Check the published_at filter on the Posts list view.

**Environment**

* React-admin version: 4.0.0.
* Last version that did not exhibit the issue (if applicable): 3.19.10
* React version: 17.0.2
* Browser: Chrome

WiXSL:
Fixes #7540

The problem is in the `getFilterFormValues` return value, if a field is a Date, and has a `parse` function that returns a date (this is why it happens by default in DateTimeInputs and not in DateInputs), its value gets treated like a regular object, and ends up returning an empty object for that field, instead of the string representation of the date

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
@@ -237,7 +237,13 @@ const getInputValue = (
     key: string,
     filterValues: Record<string, any>
 ) => {
+    if (formValues[key] === undefined || formValues[key] === null) {
+        return '';
+    }
     if (Array.isArray(formValues[key])) {
+        return lodashGet(filterValues, key, '');
+    }
+    if (formValues[key] instanceof Date) {
         return lodashGet(filterValues, key, '');
     }
     if (typeof formValues[key] === 'object') {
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -237,9 +237,15 @@ const getInputValue = (
     key: string,
     filterValues: Record<string, any>
 ) => {
+    if (formValues[key] === undefined || formValues[key] === null) {
```
WiXSL, line 240: You get `null` when a DateInput/DateTimeInput  clear button is click by the user

fzaninotto (approved):

PR #7556, opened 2022-04-15
Fix BooleanInput props

### LINKED ISSUE #7550: <BooleanInput> - cannot set custom icons
In v3 I had customized <BooleanInput> to use specific icons/colors, like the manual says it should be done:

`import { BooleanInput } from 'react-admin';
import FavoriteIcon from '@mui/icons-material/Favorite';

<BooleanInput
    source="favorite"
    options={{
        checkedIcon: <FavoriteIcon />,
    }}
/>`

It worked as expected in v3.

In v4, it does not work anymore.

Looking at the source of BooleanInput.tsx, it looks like the 'options' is not passed on to the underlying <Switch> component. And that 'options' is removed by the call to sanitizeInputRestProps() ?

WiXSL:
Fixes #7550

### DIFF
--- a/docs/BooleanInput.md
+++ b/docs/BooleanInput.md
@@ -24,12 +24,7 @@ You can use the `options` prop to pass any option supported by the MUI's `Switch
 import { BooleanInput } from 'react-admin';
 import FavoriteIcon from '@mui/icons-material/Favorite';
 
-<BooleanInput
-    source="favorite"
-    options={{
-        checkedIcon: <FavoriteIcon />,
-    }}
-/>
+<BooleanInput source="favorite" checkedIcon={<FavoriteIcon />} />
 ```
 {% endraw %}
 
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -2636,21 +2636,6 @@ import { BooleanInput } from 'react-admin';
 />
 ```
 
-### `BooleanInput` `options` Prop Has Been Removed
-
-In previous versions BooleanInput's `options` prop was used for passing props to the underlying MUI's `Switch` component. Now you just pass them directly to the `BooleanInput`.
-
-```diff
-import { BooleanInput } from 'react-admin';
-import FavoriteIcon from '@mui/icons-material/Favorite';
-
-<BooleanInput 
-    source="favorite"
--   options={{ checkedIcon: <FavoriteIcon /> }}
-+   checkedIcon={<FavoriteIcon />}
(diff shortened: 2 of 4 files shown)


packages/ra-ui-materialui/src/input/BooleanInput.tsx
```diff
@@ -62,7 +63,7 @@ export const BooleanInput = (props: BooleanInputProps) => {
     return (
         <FormGroup
             className={clsx('ra-input', `ra-input-${source}`, className)}
-            {...sanitizeInputRestProps(rest)}
+            row={row}
```
WiXSL, line 66: `row` and `sx` are really the only props to pass to `FormGroup` component.
And I'm not sure about setting `sx` to both `FormGroup` and `Switch`, so I choose the latter

slax57 (approved):

docs/Upgrade.md
```diff
@@ -2636,6 +2636,21 @@ import { BooleanInput } from 'react-admin';
 />
 ```
 
+### `BooleanInput` `options` Prop Has Been Removed
```
fzaninotto, line 2639: It's a bit too late to introduce breaking change. So either we wait for 5.0, or we do it in a backwards compatible way.

packages/ra-ui-materialui/src/input/BooleanInput.tsx
fzaninotto: I think you should put the `sx` on the FormGroup. That way, the developer can customize both the FormGroup and the Switch styles

fzaninotto (changes requested):

docs/BooleanInput.md
```diff
@@ -24,12 +24,7 @@ You can use the `options` prop to pass any option supported by the MUI's `Switch
 import { BooleanInput } from 'react-admin';
 import FavoriteIcon from '@mui/icons-material/Favorite';
 
-<BooleanInput
-    source="favorite"
-    options={{
-        checkedIcon: <FavoriteIcon />,
-    }}
-/>
+<BooleanInput source="favorite" checkedIcon={<FavoriteIcon />} />
```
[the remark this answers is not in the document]
WiXSL: No, It's perfectly fine, I have to add the `options` prop for backward compatibility. The definition of `BooleanInputProps` includes `SwitchProps`

fzaninotto (approved):

PR #7557, opened 2022-04-15
Fix Form validation when SaveButton type is "button"

### LINKED ISSUE #7555: Form submission buttons with type button should show validation errors on pristine fields
**What you were expecting:**
Clicking on a save button with `type="button"`, like "Save and Add", should show validation errors, even on pristine fields

**What happened instead:**
The notification "The form is not valid. Please check for errors" is displayed, but the validation errors are not shown below the fields.

**Steps to reproduce:**
Go to https://codesandbox.io/s/relaxed-cerf-88qx9i?file=/src/posts/PostCreate.tsx , change something in the form and then click "Save and Add".

**Environment**

* React-admin version: 4.0.0
* Last version that did not exhibit the issue (if applicable): 3.9.10 (probably)
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #7555

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.tsx
@@ -98,7 +98,6 @@ export const SaveButton = <RecordType extends RaRecord = any>(
             if (type === 'button') {
                 // this button doesn't submit the form, so it doesn't trigger useIsFormInvalid in <FormContent>
                 // therefore we need to check for errors manually
-                event.preventDefault();
                 const isFormValid = await form.trigger();
                 if (!isFormValid) {
                     event.preventDefault();
--- a/packages/ra-ui-materialui/src/button/SaveButton.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.tsx
@@ -12,7 +12,6 @@ import {
     RaRecord,
     TransformData,
     UpdateParams,
-    useNotify,
     useSaveContext,
     useTranslate,
     warning,
@@ -67,7 +66,6 @@ export const SaveButton = <RecordType extends RaRecord = any>(
         ...rest
     } = props;
     const translate = useTranslate();
-    const notify = useNotify();
     const form = useFormContext();
     const saveContext = useSaveContext();
     const { isDirty, isValidating } = useFormState();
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -101,7 +112,8 @@ export const SaveButton = <RecordType extends RaRecord = any>(
                 event.preventDefault();
                 const isFormValid = await form.trigger();
                 if (!isFormValid) {
-                    event.preventDefault();
+                    event.stopPropagation();
+                    await form.handleSubmit(handleSubmitWithErrors)(event);
```
slax57, line 116: Just realizing we have done exactly the same fix @WiXSL , didn't know you were working on #7555  too 😅 

Anyway, I did it almost the same way as you, albeit with one difference: actually I use the `handleSubmit` callback both for the error case and the case when the form is valid.
This removes the need to duplicate code like `notify('ra.message.invalid_form', { type: 'warning' });` or `saveContext?.save(values, {...})`

Below is my implementation:

```tsx
    const handleSubmit = useCallback(
        async values => {
            console.log('here1');
            let errors;
            if (saveContext?.save) {
                console.log('here2');
                errors = await saveContext?.save(values, {
                    ...mutationOptions,
                    transform,
                });
            }
            if (errors != null) {
                console.log('here3');
                setSubmissionErrors(errors, form.setError);
            }
        },
        [form, saveContext, mutationOptions, transform]
    );
```

```tsx
    const handleClick: MouseEventHandler<HTMLButtonElement> = useCallback(
        async event => {
            if (onClick) {
                onClick(event);
            }
            if (event.defaultPrevented) {
                return;
            }
            if (type === 'button') {
                // Prevent outer forms to receive the event
                event.stopPropagation();
                form.handleSubmit(handleSubmit)(event);
                return;
            }
        },
        [onClick, type, form, handleSubmit]
    );
```

What do you think?

WiXSL: Yes, actually, I was going to improve it like you did it, without calling `form.trigger()`

slax57 (approved):

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -99,20 +114,11 @@ export const SaveButton = <RecordType extends RaRecord = any>(
                 // this button doesn't submit the form, so it doesn't trigger useIsFormInvalid in <FormContent>
                 // therefore we need to check for errors manually
                 event.preventDefault();
-                const isFormValid = await form.trigger();
-                if (!isFormValid) {
-                    event.preventDefault();
-                    notify('ra.message.invalid_form', { type: 'warning' });
```
fzaninotto, line 105: I no longer see this notification in the new code

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -99,20 +114,11 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
-                    return;
-                }
-                const values = form.getValues();
-                saveContext?.save(values, {
-                    ...mutationOptions,
-                    transform,
-                });
+                event.stopPropagation();
```
fzaninotto, line 117: why do you both preventDefault and stopPropagation?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/button/SaveButton.tsx
WiXSL: Because the `form.handleSubmit(handleSubmit)(event);` call already shows the same notification when needed

PR #7558, opened 2022-04-15
Add queryClient prop to <Admin />

joshq00:
The `queryClient` prop is not passed into the application.

Merging this change will propagate the `queryClient`


fzaninotto (approved):

fzaninotto:
Thanks!

PR #7561, opened 2022-04-16
Fix RaRecord example usage

### LINKED ISSUE #7560: Docs reference Record type which does not exist
**What you were expecting:**

See this example: https://marmelab.com/react-admin/doc/4.0/Upgrade.html#list-views

```ts
import { useGetList, Record } from 'react-admin';
```

I expected the `Record` type to exist. Note: It did exist in v3. I could do `import { Record } from 'ra-core';`.

**What happened instead:**

```ts
import { useGetList, Record } from 'react-admin';
```

generates this error: `'"react-admin"' has no exported member named 'Record'`

I am pretty sure the type should be `RaRecord`, according to the code here: https://github.com/marmelab/react-admin/blob/953716c693c78695180e1c3c1ff38e7062c89264/packages/ra-core/src/types.ts#L12-L15

**Environment**

* React-admin version: v4
* Last version that did not exhibit the issue (if applicable): v3
* React version: NA
* Browser: NA
* Stack trace (in case of a JS error): `'"react-admin"' has no exported member named 'Record'`

WiXSL:
Fixes #7560

### DIFF
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -1380,7 +1380,7 @@ So the `RecordMap` type is no longer necessary and was removed. TypeScript compi
 
 ```diff
 -import { useGetList, RecordMap } from 'react-admin';
-+import { useGetList, Record } from 'react-admin';
++import { useGetList, RaRecord as Record } from 'react-admin';
 
 const PostListContainer = () => {
 -   const { data, ids, loading } = useGetList(


docs/Upgrade.md
```diff
@@ -1380,7 +1380,7 @@ So the `RecordMap` type is no longer necessary and was removed. TypeScript compi
 
 ```diff
 -import { useGetList, RecordMap } from 'react-admin';
-+import { useGetList, Record } from 'react-admin';
++import { useGetList, RaRecord as Record } from 'react-admin';
```
djhi, line 1383: We should avoid renaming it imo

WiXSL: #7563 is correct. I'm closing this one in favor of that PR

WiXSL:
#7563 is correct. I'm closing this one in favor of that PR

PR #7568, opened 2022-04-19
Update DataProviderList.md

mshd:
I would like to add new dataProvider developed by @ogroppo
This DataProvider implementation is optimised for prisma and does not use graphql dependency.


docs/DataProviderList.md
```diff
@@ -45,6 +45,8 @@ If you can't find a Data Provider for your backend below, no worries! [Writing a
 * **[PostgREST](https://postgrest.org/)**: [raphiniert-com/ra-data-postgrest](https://github.com/raphiniert-com/ra-data-postgrest)
 * **[Prisma](https://github.com/weakky/ra-data-prisma)**: [weakky/ra-data-prisma](https://github.com/weakky/ra-data-prisma)
 * **[Prisma Version 2](https://www.prisma.io/)**: [panter/ra-data-prisma](https://github.com/panter/ra-data-prisma)
+* **[Prisma](https://www.npmjs.com/package/ra-data-simple-prisma)**: [codeledge/ra-prisma](https://github.com/codeledge/ra-prisma)
+
```
ogroppo, line 49: should not have impact on the markdown output but perhaps better not having a new line for consistency

mshd: ok

djhi:
What's the difference between your package and the other two already in the list?

ogroppo:
Hi, it does not rely on graphql so the less boilerplate and setup, it's also intended to be nextjs friendly!

djhi:
I see. In this case, would you mind finding a more explicit title so that users know the difference ? Maybe `Prisma REST`? Thanks!

mshd:
@djhi We renamed it to Prisma NextJS because it's optimized for NextJs

mshd:
Any update on the PR @djhi ?

fzaninotto:
I don't see anything related to NextJs in https://github.com/codeledge/ra-prisma-nextjs/tree/main/packages/ra-data-simple-prisma. Perhaps you should underline the absence of dependency on GraphQL instead.

mshd:
> I don't see anything related to NextJs in https://github.com/codeledge/ra-prisma-nextjs/tree/main/packages/ra-data-simple-prisma. Perhaps you should underline the absence of dependency on GraphQL instead.

Ok good point @fzaninotto, it's optimzed for NextJs but obviously isn't restricted to it. I renamed it to Prisma Simple, to highlight the fact that it's a simpler implementation compared to the other packages.

fzaninotto:
I updated the other elements to make the difference clearer

fzaninotto (approved):

PR #7585, opened 2022-04-22
Fix RichTextInput Link Button Stay Disabled When Selection Updates

### LINKED ISSUE #7582: Default buttons in custom toolbar for RichTextInput not re-rendering
When customizing the toolbar to provide a custom set of the default buttons, it would seem that a change in the state of the editor does not cause the button to re-render. It becomes impossible to add/modify a link through the Toolbar for example.

|Package|Version|
|------|-------|
|react-admin|4.0.1|
|react|17.0.2|

I've made an example forking the sandbox example : 
https://codesandbox.io/s/keen-brook-vkws1r?file=/package.json

The example is using the `PostEdit.tsx` `RichTextInput` for the body and uses the following reproduction scenario :
```jsx
<RichTextInput
    source="body"
    label=""
    validate={required()}
    fullWidth
    toolbar={<RichTextInputToolbar />}
/>
```

This example should result in the same thing as the default `<RichTextInput />` without a toolbar prop, but somehow, it's not the case. You can see in the screenshot below that despite having selected some text, the link button is not enabled.

image

djhi:
Fixes #7582

### DIFF
--- a/packages/ra-input-rich-text/src/buttons/LinkButtons.tsx
+++ b/packages/ra-input-rich-text/src/buttons/LinkButtons.tsx
@@ -4,16 +4,12 @@ import InsertLink from '@mui/icons-material/InsertLink';
 
 import { useTranslate } from 'ra-core';
 import { useTiptapEditor } from '../useTiptapEditor';
+import { useEditorSelection } from './useEditorSelection';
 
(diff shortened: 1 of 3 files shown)


slax57 (approved):
Just one question: do you think we can cover this with tests?

djhi:
I tried and failed when I implemented this input. Will try again later

PR #7595, opened 2022-04-28
Fix #7593, choices should indeed be `any[]`

### LINKED ISSUE #7593: SelectInput throw choices type error after upgrade to v4
After upgrade to v4, SelectInput throw type error: `Type '{ name: string; value: number; }[]' is not assignable to type 'RaRecord[]'.   Property 'id' is missing in type '{ name: string; value: number; }' but required in type 'RaRecord'.`

**What you were expecting:**

The `choices` is not limited `RaRecord[]` type .

**What happened instead:**

The prop `choices` of SelectInput is limited to `RaRecord[]`

**Steps to reproduce:**

1. Create an array of non special structure object as choices. But not include `id` as key for each choices
2. pass this array as SelectInput choices
3. throw Type Error during build.

**Related code:**

```
const choices = [{ name: 'first', value: 100 }, { name: 'second', value: 200}]

<SelectInput
  label="order"
  source="order"
  optionText="name"
  optionValue="value"
  choices={choices}
/>
```

**Other information:**

**Environment**

* React-admin version: v4
* React version: 17.0.2
* Browser: edge 100

bingluen:
See #7593

### DIFF
--- a/packages/ra-core/src/form/useChoices.tsx
+++ b/packages/ra-core/src/form/useChoices.tsx
@@ -13,7 +13,7 @@ export type OptionTextFunc = (choice: any) => string | OptionTextElement;
 export type OptionText = OptionTextElement | OptionTextFunc | string;
 
 export interface ChoicesProps {
-    choices?: RaRecord[];
+    choices?: any[];
     isFetching?: boolean;
     isLoading?: boolean;
     optionValue?: string;


slax57 (approved):

slax57:
Thanks! 🙂

PR #7609, opened 2022-04-29
Allow To Create An Admin Without Resources

### LINKED ISSUE #7598: CustomRoutes don't work without resources
It took me a while to reproduce this one but I think I found it.

When you have no resources (either because you have no permissions to see them or simply because you don't) CustomRoutes don't get registered outside of the permission function inside of `<Admin />`

It's easier to just show what I mean: https://codesandbox.io/s/sparkling-wind-5sq5u9?file=/src/index.tsx

In render() replace AppWorking with AppBroken and hit the refresh icon in the preview on the right (make sure the preview url path is set to /custom)

_**Note:** I used `<Admin />` and had the same result. I just left in AdminContext and AdminUI to more accurately reflect my actual code._

**What you were expecting:**

I was expecting custom routes to be registered regardless of permission/resource state.

**What happened instead:**

No route registered.

**Steps to reproduce:**

https://codesandbox.io/s/sparkling-wind-5sq5u9?file=/src/index.tsx

In render() replace AppWorking with AppBroken and hit the refresh icon in the preview on the right.

**Related code:**

[See sandbox](https://codesandbox.io/s/sparkling-wind-5sq5u9?file=/src/index.tsx)

**Context:**

My permissions are loaded async. React-admin doesn't handle this very well so hacked in a reload of the page after login to make sure the proper resources show up.

I'll be sending out invite emails to users. The link should go to /accept-invite, which doesn't require you to be logged in. It lives independent of the rest of the app.

When it's the first time visiting the dashboard you're not logged in and so there are no resources. This causes the "ready" component to show.

**Workaround/hack:**

The only way I can work around this is by adding an empty resource like so:

`<Resource name="hack" />`

djhi:
Fixes #7598


packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
```diff
@@ -181,6 +180,7 @@ export const useConfigureAdminRouterFromChildren = (
         };
     }, [register, unregister, resources]);
 
+    console.log({ status });
```
fzaninotto, line 183: console.log spotted

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
```diff
@@ -60,28 +60,27 @@ export const useConfigureAdminRouterFromChildren = (
...
 
     const [status, setStatus] = useSafeSetState<AdminRouterStatus>(() =>
-        !!getSingleChildFunction(children)
-            ? 'loading'
-            : resources.length > 0
-            ? 'ready'
-            : 'empty'
+        getStatus({
+            children,
+            resources,
```
fzaninotto, line 65: I suggest you use the same `routesandResources`  as in the effect, so that the reader doesn't try to understand why there is a difference

fzaninotto (changes requested):

slax57 (approved):
Okay actually it was not as complicated as I imagined 😅

fzaninotto (approved):

PR #7613, opened 2022-05-02
[Doc] Remove deprecated fieldKey

smeng9:
Make it consistent with upgrade guide here https://marmelab.com/react-admin/Upgrade.html#arrayfield-does-not-accept-a-fieldkey-prop-anymore


djhi (approved):
Thanks!

Issue #7604, opened 2022-04-29
Custom theme: styleOverrides on Ra* components are not applied


sweco-sedalh:
**What you were expecting:**

It being possible to override styles of Ra* components in the theme as in v3.

**What happened instead:**

Styles specified in the theme are not applied. Styles on Mui* components are, however, applied as expected.

**Steps to reproduce:**

1. Create a custom theme and pass it to <Admin>
2. Add `styleOverrides` on react-admin components (see below)
3. Open in browser

**Related code:**

https://codesandbox.io/s/determined-wilson-k6ifbu?file=/src/index.tsx:722-1068

```typescript
const theme = _.merge({}, defaultTheme, {
  components: {
    MuiIconButton: {
      styleOverrides: {
        root: {
          // works as expected
          color: "red"
        }
      }
    },
    RaList: {
      styleOverrides: {
        content: {
          // does not work
          backgroundColor: "red"
        }
      }
    }
  }
});
```

**Other information:**

**Environment**

* React-admin version: 4.0.1
* Last version that did not exhibit the issue (if applicable): 3.19.7
* React version: 17.0.2
* Browser: Chrome 100.0.4896.127
* Stack trace (in case of a JS error):

sweco-sedalh:
It's, or at least should be, the same thing. Though I realize my original example wasn't clear on that, I've updated the sandbox and here is a new snippet:

```typescript
const theme = _.merge({}, defaultTheme, {
  components: {
    MuiButton: {
      styleOverrides: {
        root: {
          // works as expected
          border: "5px solid red"
        },
        startIcon: {
          // works as expected
          color: "red"
        }
      }
    },
    RaList: {
      styleOverrides: {
        root: {
          // works as expected
          backgroundColor: "green"
        },
        content: {
          // does not work
          backgroundColor: "red"
        }
      }
    }
  }
});
```

See also some of the examples from the MUI documentation: https://mui.com/material-ui/customization/theme-components/#global-style-overrides

Your code works as a workaround, however this issue should definitely not be closed as it stands IMO.

sweco-sedalh:
I haven't fully understood how MUI wants it now in v5, but from what I understand the `overridesResolver` (for example https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/ListView.tsx#L171) would need to be adjusted, see for a somewhat related example: https://mui.com/system/styled/#custom-components.

djhi:
[...]

PR #7616, opened 2022-05-02
Fix style overrides for Ra* components

sweco-sedalh:
This pull request allows using `styleOverrides` from MUI themes to override the styles of all classes in react-admin components, as opposed to just the root class previously.

Fixes: #7604


packages/ra-ui-materialui/src/detail/makeOverridesResolver.tsx
```diff
@@ -0,0 +1,8 @@
+export function makeOverridesResolver(classes: { [key: string]: string }) {
```
djhi, line 1: Would you mind moving this in directly under the `src` directory?

djhi (changes requested):

packages/ra-ui-materialui/src/detail/makeOverridesResolver.tsx
sweco-sedalh: Done

packages/ra-ui-materialui/src/makeOverridesResolver.tsx
```diff
@@ -0,0 +1,8 @@
+export function makeOverridesResolver(classes: { [key: string]: string }) {
+    return (props, styles) => [
+        styles.root,
+        ...Object.entries(classes).map(([name, clazz]) => ({
```
djhi, line 4:
```suggestion
        ...Object.entries(classes).map(([name, cssClassName]) => ({
```

packages/ra-ui-materialui/src/makeOverridesResolver.tsx
```diff
@@ -0,0 +1,8 @@
...
+            [`& .${clazz}`]: styles[name],
```
djhi, line 5:
```suggestion
            [`& .${cssClassName}`]: styles[name],
```

djhi (changes requested):
Almost there! Sorry I did not see the last point that immediately

djhi:
Have you tried those modifications? I tried converting the demo theme accordingly and it does not seem to work

packages/ra-ui-materialui/src/makeOverridesResolver.tsx
```diff
@@ -0,0 +1,8 @@
+export function makeOverridesResolver(classes: { [key: string]: string }) {
+    return (props, styles) => [
+        styles.root,
+        ...Object.entries(classes).map(([name, cssClassName]) => ({
+            [`& .${cssClassName}`]: styles[name],
```
djhi, line 5: This is not always right. Sometimes, the CSS class is applied on the element directly: `&.${cssClassName}` (note the space has been removed)

sweco-sedalh:
> Have you tried those modifications? I tried converting the demo theme accordingly and it does not seem to work

I have tested them by using basically the same code as in the sandbox on #7604 applied to the simple example:

```typescript
const theme = _.merge({}, defaultTheme, {
    components: {
      MuiButton: {
        styleOverrides: {
          root: {
            // works as expected
            border: "5px solid red"
          },
          startIcon: {
            // works as expected
            color: "red"
          }
        }
      },
      RaList: {
        styleOverrides: {
          root: {
            // works as expected
            backgroundColor: "green"
          },
          content: {
            // does not work
            backgroundColor: "red"
          }
        }
      }
    }
  });
```

I hadn't seen the demo theme before, will try to update it as well (probably should be part of this PR).

sweco-sedalh:
Now the demo theme is updated and works.

fzaninotto:
Thanks for your contribution. But I'm -1 for this change, for the reasons I explained in #7604.

fzaninotto:
As explained in #7604, we won't merge that change in master. Thanks for your contribution anyway!

PR #7619, opened 2022-05-03
Fix FileInput name And id Props

### LINKED ISSUE #7572: FileInput no longer passes source to `name` attribute, and instead `id` attribute
FileInput no longer receives a `name` attribute. Instead it gets passed an `id` attribute instead.

The main problem for us is that it causes our E2E tests to fail (though this is something we will happily update), it also means that there are elements with duplicate IDs (HTML validation issue). e.g.:

image

This is a problem that exists in the CodeSandbox without any changes needed. 

Just jump into the `Create Post` page. Open the element inspector and move to the `input type="file"` element, and observe the problem

**Environment**

* React-admin version: 4.0.1
* React version:
* Browser:

djhi:
Fixes #7572

### DIFF
--- a/packages/ra-ui-materialui/src/input/FileInput.tsx
+++ b/packages/ra-ui-materialui/src/input/FileInput.tsx
@@ -149,7 +149,7 @@ export const FileInput = (props: FileInputProps) => {
 
     return (
         <StyledLabeled
-            id={id}
+            htmlFor={id}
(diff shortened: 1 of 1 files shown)


andrico1234:
I've noticed that this problem exists in ImageInput too

fzaninotto (approved):

PR #7627, opened 2022-05-03
Fix <Button/> type

smeng9:
Per MUI documentation https://mui.com/material-ui/api/button/#props `icon` is no longer a prop

Also we are passing icon as children


fzaninotto (approved):

fzaninotto:
Right, thanks!

PR #7633, opened 2022-05-04
Add possibility to disable redirect in useCreateController

waltheri:
Currently if `redirect` prop of useCreateController is set to false, it is ignored. This is probably not intentional and it makes difficult to disable redirects.


slax57 (approved):
Thanks!

djhi (approved):
Thanks!

PR #7636, opened 2022-05-04
Use theme to set input variant

### LINKED ISSUE #7635: Cannot override defaultProps using variants
**What you were expecting:**

Being able to set all inputs as outlined using a MUI theme:

```jsx
const theme = {
    components: {
        MuiTextField: {
            defaultProps: {
                variant: 'outlined',
            },
        },
    },
};
```

**What happened instead:**

The inputs keep the `filled` variant.

image

**Steps to reproduce:**

I forked the simple example codesandbox to give it a custom theme:

https://codesandbox.io/s/cranky-bhabha-7xqjwy?file=/src/index.tsx

Forms and filters should use the outlined variant, but they don't

**Related code:**

In fact, `defaultProps` from the theme only work if there is no default value in the component code. But *all* our inputs define their variants and margins. 

for instance, in `ResettableTextField`:

```tsx
        const {
            clearAlwaysVisible,
            InputProps,
            value,
            resettable,
            disabled,
            variant = 'filled',
            margin = 'dense',
            className,
            ...rest
        } = props;
````

The default values for `variant`, `margin` and `size` should be defined in the default theme, not in the components. 

**Other information:**

**Environment**

* React-admin version: 4.0.2
* React version: 17
* Browser: Chrome

### PR DESCRIPTION
**Pull request by fzaninotto**
Closes #7635

- [x] remove default values for `variant`, `margin`, `size` in all Input components
- [x] set the default `variant`, `margin`, `size` in the default theme
- [x] set the default `variant`, `margin`, `size` in the demos theme
- [x] update the documentation about overriding the theme
- [x] Add mention in the upgrade guide for custom themes

This PR introduces a small BC break, but I think we cannot wait the next major for that, plus it's an easy fix.

### CODE REVIEW

**Comment by fzaninotto** [comment]
The compilation error seems to come from MUI... https://github.com/mui/material-ui/issues/32659

**Comment by fzaninotto** [comment]
Switching to RFR

File: docs/Theming.md [hunk 1]
```diff
@@ -92,9 +110,289 @@ This example results in:
...
+    </Admin>
+);
+```
+
+There are 2 important gotchas here:
+
+- Don't forget to merge your custom style overrides with the ones from react-admin's `defaultTheme`, otherwise the application will have the default MUI theme (most notably, outlined text inputs)
+- Custom style overrides must live under a `root` key. Then, the style override syntax is the same as the one used for the `sx` prop.
+
+Note thet you don't need to call `createTheme` yourself. React-admin will do it for you.
```
slax57, line 240:
```suggestion
Note that you don't need to call `createTheme` yourself. React-admin will do it for you.
```

docs/Theming.md
```diff
@@ -92,9 +110,289 @@ This example results in:
...
+                variant: 'outlined' as const,
+            },
+        },
+    }
+};
+```
+
+## Writing a Custom Theme
+
+MUI theming also allows to change the default palette, typography, colors, etc. This is very useful to change the react-admin style to match the branding of your commpany.
```
slax57, line 292:
```suggestion
MUI theming also allows to change the default palette, typography, colors, etc. This is very useful to change the react-admin style to match the branding of your company.
```

docs/Upgrade.md
```diff
@@ -3081,6 +3081,51 @@ If you created a custom app (without the `<Admin>` component), you may have used
...
+            root: {
+                backgroundColor: "Lavender",
+                "& .RaDatagrid-headerCell": {
+                    backgroundColor: "MistyRose",
+                },
+            }
+        }
++       MuiTextField: {
++           defaultProps: {
++               variant: 'filled',
```
slax57, line 3122: don't we need `as const` here?

PR #7638, opened 2022-05-04
Fix IntrospectionOptions exported type

arjunyel:
operationNames is set by default https://github.com/marmelab/react-admin/blob/03a0980ad801d753faa92d3a74ed18be43b926ec/packages/ra-data-graphql/src/index.ts#L60


djhi (approved):
Thanks!

PR #7642, opened 2022-05-05
Simplify useConfigureAdminRouterFromChildren permissions

### LINKED ISSUE #7637: Cannot log in to the e-commerce demo
on master, when I'm not logged in, entering anything in the login form has no effect. 

It seems I'm redirected to the dashboard and then immediately afterwards back to the login page. 

this doesn't happen in the simple example. 

latest version that did not show the bug: v4.0.2 - regression caused by #7609 or #7606 IMO

image

djhi:
Fixes #7637

### DIFF
--- a/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
+++ b/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
@@ -10,7 +10,7 @@ import {
     useEffect,
     useState,
 } from 'react';
-import { useLogout, useGetPermissions, useAuthState } from '../auth';
+import { useLogout, usePermissions, useAuthState } from '../auth';
 import { useSafeSetState } from '../util';
 import {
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
```diff
@@ -43,7 +43,7 @@ import { useResourceDefinitionContext } from './useResourceDefinitionContext';
 export const useConfigureAdminRouterFromChildren = (
     children: AdminChildren
 ): RoutesAndResources & { status: AdminRouterStatus } => {
-    const permissions = usePermissionsWithLogout();
```
djhi, line 46: `usePermissions` already handle the case when there is no authProvider

slax57 (approved):
works like a charm 🙂

fzaninotto (approved):

Issue #7644, opened 2022-05-05
`disableRemove` doesn't pass down record object


andrico1234:
It seems that the following code will always log out an empty object in the console.

```
<ArrayInput source="resourcePools" label="CP Resource Pools" level={2}>
        <SimpleFormIterator
          disableRemove={record => {
            console.log(record);
            return true
          }}
        >
            ...
        </SimpleFormIterator>
</ArrayInput>
```

**What you were expecting:**
To be given the record of the current ArrayInput iterable

**Steps to reproduce:**
If you go to the [repro case](https://codesandbox.io/s/upbeat-thunder-3j4pdk?file=/src/posts/PostEdit.tsx)

- Navigate to the PostEdit page
- Add a new post_id
- Click save
- Click back into the record

You'll observe that the input is populated, but no data is being passed through to the `disableRemove` function

**Related code:**

* Preferably, a CodeSandbox forked from 
  - https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple (v4)
  - https://codesandbox.io/s/github/marmelab/react-admin/tree/3.x/examples/simple (v3)
* A link to a GitHub repo with the minimal codebase to reproduce the issue

```
insert short code snippets here
```

**Other information:**

**Environment**

* React-admin version:
* Last version that did not exhibit the issue (if applicable):
* React version:
* Browser:
* Stack trace (in case of a JS error):

slax57:
Reproduced, thanks!

andrico1234:
I'll be happy to take a pass at this!

PR #7645, opened 2022-05-05
Pass `record` through to `disableRemove` function

andrico1234:
Fixes: https://github.com/marmelab/react-admin/issues/7644

I hope this is the right way to handle this problem, though happy to make any changes as needed

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
@@ -12,7 +12,7 @@ import { styled } from '@mui/material';
 import clsx from 'clsx';
 import get from 'lodash/get';
 import PropTypes from 'prop-types';
-import { RaRecord } from 'ra-core';
+import { RaRecord, useRecordContext } from 'ra-core';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -46,6 +45,7 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
         getItemLabel = DefaultLabelFn,
     } = props;
     const { append, fields, move, remove } = useArrayInput(props);
+    const record = useRecordContext(props.record);
```
djhi, line 48:
```suggestion
    const record = useRecordContext(props);
```

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
andrico1234: 🤦🏽 

thanks for spotting!

djhi (approved):

qburst-niji:
Can you please add this fix to 3.19.11. Thanks

PR #7652, opened 2022-05-06
Make ShowBase, CreateBase and EditBase components support resource override via props

### LINKED ISSUE #7651: CreateBase  doesn't set the ResourceContext like Create does
**What you were expecting:**

When I use ```<CreateBase resource="payments">```, I expect to have the ResourceContext set.

**What happened instead:**

I don't have resourceContext set inside a CreateBase,
I have to wrap it with the ```<ResourceContextProvider value='payments'>```

* React-admin version:4.0.2

slax57:
Fixes #7651


djhi (approved):

packages/ra-ui-materialui/src/detail/Create.tsx
```diff
@@ -59,19 +54,10 @@ export const Create = <RecordType extends RaRecord = any>(
...
-    return props.resource ? (
-        // support resource override via props
-        <ResourceContextProvider value={props.resource}>
-            {body}
-        </ResourceContextProvider>
-    ) : (
-        body
+    return (
+        <CreateBase {...props}>
+            <CreateView {...props} />
```
fzaninotto, line 59: I'm not a fan of passing all props both to the controller and to the view. We should split them IMO (as we do in Show and List). 

The same goes for the Edit component.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/detail/Create.tsx
```diff
@@ -54,9 +54,23 @@ export const Create = <RecordType extends RaRecord = any>(
     props: CreateProps<RecordType> & { children: ReactElement }
 ): ReactElement => {
     useCheckMinimumRequiredProps('Create', ['children'], props);
+    const {
+        resource,
```
fzaninotto, line 58: disableAuthentication, hasEdit, hasShow are missing

packages/ra-ui-materialui/src/detail/Edit.tsx
```diff
@@ -55,9 +55,27 @@ export const Edit = <RecordType extends RaRecord = any>(
     props: EditProps<RecordType> & { children: ReactNode }
 ) => {
     useCheckMinimumRequiredProps('Edit', ['children'], props);
+    const {
+        resource,
```
fzaninotto, line 59: disableAuthentication is missing

fzaninotto (changes requested):

packages/ra-ui-materialui/src/detail/Create.tsx
slax57: I omitted them because none of these are declared in the type of `props`

```
Property 'disableAuthentication' does not exist on type 'CreateProps<RecordType> & { children: ReactElement<any, string | JSXElementConstructor<any>>; }'
```

Should I add them?

fzaninotto: yes, please

slax57: done

djhi (approved):

fzaninotto (approved):

PR #7658, opened 2022-05-07
[Doc] Update <SaveButton/> prop in <Form/> example

smeng9:
Per MDN documentation https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form

Buttons placed outside a form should accept `form` as the attribute instead of `formId` to associate with a form

### DIFF
--- a/docs/Form.md
+++ b/docs/Form.md
@@ -79,7 +79,7 @@ export const PostCreate = () => (
 
 ## `id`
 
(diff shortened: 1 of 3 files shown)


slax57 (approved):
Thanks for this contribution! 🙂 
I agree this is a little bit better, even though none of the examples provided in the doc seem to actually work out of the box (because `<Create>` only accepts one child) 🙁 
But we can already merge this change 🙂

djhi (approved):
Thanks!

Issue #7661, opened 2022-05-09
required validation not working  correctly with DateTimeInput


bingluen:
**What you were expecting:**

Show error message after trigger validation when clear `DateTimeInput`, if `required` is include in its validate props.

**What happened instead:**

Did NOT Show error message after trigger validation when clear `DateTimeInput`, if `required` is include in its validate props.

If `DateTimeInput` had been set a value, the `required` validation will always be passed.

**Steps to reproduce:**

1. Give a datetime on `DateTimeInput`. (Use browser UI or just typing, whatever)
2. Make it blur (Click other field or anywhere)
3. Click `DataTimeInput` which just give a datetime.
4. Clear it (Use browser UI or just typing, whatever)
5. Trigger validation (onChange or onSubmit, whatever)
6. The error message that should have appeared did not appeared.

**Related code:**

[`DateTimeInput.tsx#L18`](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/DateTimeInput.tsx#L18)
```typescript
const parseDateTime = (value: string) => new Date(value);
```
After clear field, `parseDateTime` will receive `''` (empty string) as value, and return `Invalid Date` with type is `Date` object.

[`validate.ts#L145`](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/form/validate.ts#L145) and [`validate.ts#L7`](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/form/validate.ts#L7)

```typescript
export const required = memoize((message = 'ra.validation.required') =>
    Object.assign(
        (value, values) =>
            isEmpty(value)
                ? getMessage(message, undefined, value, values)
                : undefined,
        { isRequired: true }
    )
);
```

```typescript
const isEmpty = (value: any) =>
    typeof value === 'undefined' ||
    value === null ||
    value === '' ||
    (Array.isArray(value) && value.length === 0);
```

`isEmpty` return `false` if `Invalid Date` (`Date` object) as `value` pass into.

**Environment**

* React-admin version: 4.0.2
* React version: 17.0.2
* Browser: Microsoft Edge 100.0.1185.50 / Firefox 99 / Firefox 100

slax57:
reproduced, thanks!

PR #7662, opened 2022-05-09
:bug: Fix DateTimeInput is not working correctly. #7661

bingluen:
Return empty string if `parseDateTime` function receive empty string as value instead `Invalid Date`(`Date` object)

Reference #7661

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
+++ b/packages/ra-ui-materialui/src/input/DateTimeInput.tsx
@@ -15,7 +15,7 @@ import { InputHelperText } from './InputHelperText';
  * @param {string} value Date string, formatted as yyyy-MM-ddThh:mm
  * @return {Date}
  */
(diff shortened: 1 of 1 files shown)


bingluen:
Or return `undefined` is closer to the semantics clear/reset

packages/ra-ui-materialui/src/input/DateTimeInput.tsx
```diff
@@ -15,7 +15,7 @@ import { InputHelperText } from './InputHelperText';
  * @param {string} value Date string, formatted as yyyy-MM-ddThh:mm
  * @return {Date}
  */
-const parseDateTime = (value: string) => new Date(value);
+const parseDateTime = (value: string) => value ? new Date(value) : value;
```
bingluen, line 18: Or return `undefined` is closer to the semantics clear/reset

djhi:
Thanks !

PR #7664, opened 2022-05-09
Fix type of Button's "component" prop

okovpashko:
The "component" prop allows passing the tag or component that will be rendered as a button
`ReactNode` refers to a string, number, null, undefined, boolean, React element, fragment, or portal
`ElementType` refers to an HTML tag name, React class or functional component


djhi (approved):

djhi:
Thanks!

PR #7665, opened 2022-05-09
Proposal: add meta to EditProps

RWOverdijk:
Hey!

I currently have the following changes in my own project, and I've had them for a while now. This is an example PR to share the diff.

I noticed that there was no nice way for me to pass meta through to the data provider without creating a lot of conditionals in the data provider.

An example of how I use this:

```tsx
export function TopicEdit(props: any) {
  return (
    <Edit meta={{ embed: 'media' }} {...props}>
      <TopicForm />
    </Edit>
  );
}
```

The change is relatively minor since everything after the controller already accepts meta.

Thoughts?

### DIFF
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -49,6 +49,7 @@ export const useEditController = <RecordType extends RaRecord = any>(
     const {
         disableAuthentication,
         id: propsId,
+        meta = {},
         mutationMode = 'undoable',
         mutationOptions = {},
         queryOptions = {},
@@ -69,7 +70,7 @@ export const useEditController = <RecordType extends RaRecord = any>(
         RecordType
     >(
         resource,
-        { id },
+        { id, meta },
         {
             onError: () => {
                 notify('ra.notification.item_doesnt_exist', {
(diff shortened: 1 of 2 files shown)


fzaninotto:
This is definitely something we'll want to add. The question is: should we add a new prop for that, or use the `queryOptions`? Especially in controllers that do more than one query (e.g. the `<Edit>` component), developers may want to specify meta for both the query and the mutation. 

Note that for a PR to be accepted, you should add tests and documentation.

RWOverdijk:
@fzaninotto I thought about that, but my understanding was that queryOptions goes straight to react query (I think that's the name, don't quote me on that).

About the pr being accepted, I wasn't expecting it to be (I know the high standards you have for this repo, so I wouldn't dare haha).

But a similar solution would be cool. What I liked about using meta is that it's meta so you can put in whatever you want, for example:

```ts
const meta = { findOne: { meta: 'for find one' }, update: { meta: 'for update' } };
```

This idea might be a bit shortsighted of me, but I think (hope) the concept is clear.

fzaninotto:
> it's meta so you can put in whatever you want, for example:
> 
> ```ts
> const meta = { findOne: { meta: 'for find one' }, update: { meta: 'for update' } };
> ```

Not a fan, especially as it introduces a new way to target either the query or the mutation. 

I agree that the `queryOptions` currently go straight to react-query, so the change would require some plumbing. But I can't find a better solution than to put the `meta` in it.

PR #7666, opened 2022-05-09
`sort` in <ReferenceArrayInput> doesn't work

DjebbZ:
Whatever I put in the `sort` props of the `<ReferenceArrayInput>` component is not taken into account, and it still filters the content by id.

The only difference I noticed is that by default, it sorts the referenced content by `id` DESC, and when I use the sort option it sorts by `id` ASC.

Reading the source code, it seems that the `useReferenceArrayInputController` doesn't do its job properly, but from a quick glance I didn't find the problem.

AFAIK, the bug exists in both the latest 3.x and 4.x versions.

### DIFF
--- a/examples/simple/src/posts/PostList.tsx
+++ b/examples/simple/src/posts/PostList.tsx
@@ -24,6 +24,8 @@ import {
     TextField,
     TextInput,
     useTranslate,
+    ReferenceArrayInput,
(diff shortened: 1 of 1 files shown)


slax57:
See https://github.com/marmelab/react-admin/issues/7668#issuecomment-1122082449

PR #7667, opened 2022-05-09
Rich text as json

RWOverdijk:
Hello again, I wanted to share another suggestion! I have this change locally and it works.

**Context**

The reason I use this:

Our front end (api consumer) is a mobile app. Instead of a webview (which looks kind of lame on a mobile phone) I use json to parse it into native components. To do this I need json instead of html.

Luckily, tiptap supports this! This tiny change allows me to store json instead of html.

**BC**

I made it backwards compatible by defaulting to html and explicitly requiring `json` to activate the new behaviour. Even if someone uses this prop it won't break.

Thoughts?

### DIFF
--- a/docs/RichTextInput.md
+++ b/docs/RichTextInput.md
@@ -144,3 +144,22 @@ export const MyEditorOptions = {
 	],
 };
 ```
+
+## Changing the content type
+
+By default the RichTextInput will store and load the content as html. You can change this to json instead by passing in a prop `contentType` set to `json` if you prefer. This will export the input data as json instead. You can read more about this in the [tiptap docs](https://tiptap.dev/guide/output#option-1-json).
+
+Example:
+
+```jsx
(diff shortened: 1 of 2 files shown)


RWOverdijk:
Well that was fun 😅

fzaninotto:
It's a nice addition! Could you PR against `next` rather than `master`, as the latter is only for bug fixes?

Also, please update the documentation and add a unit test to demonstrate the new feature.

RWOverdijk:
@fzaninotto Sure, I'll do that tomorrow 😄

RWOverdijk:
@fzaninotto Where are the unit tests for text input?

RWOverdijk:
@fzaninotto why did you reopen this? I need to reapply the changes on next, this is based on master

fzaninotto:
right, sorry!

PR #7671, opened 2022-05-09
Fix documentation for RichTextInputToolbar's LevelSelect

friday:
`ra-input-rich-text` doesn't export any `RichTextInputLevelSelect`. It should be just `LevelSelect`.


djhi (approved):
Thanks!

Issue #7625, opened 2022-05-03
NumberInput sends empty string instead of null when no value is set


weissu42:
**What you were expecting:**
NumberInput sends "null" when left empty

**What happened instead:**
NumberInput sends an empty string when left empty

I was able to fix this by setting the defaultValue of the NumberInput to "null", however I find it a bit unexpected that I have to do this.

**Environment**

* React-admin version: 4.0.1
* Last version that did not exhibit the issue (if applicable): -
* React version: 17.0.2
* Browser: Firefox, Chrome

slax57:
reproduced, thanks!

weissu42:
Thanks for the quick confirmation of the bug.

I today stumbled upon similar behaviour with the `NullableBooleanInput` component:
* when I used it inside of an `Edit` and deleted a previously present value (i.e. selecting the empty option), "null" was send - this is working fine
* when I used it inside of a `Create` and didn't select a value, the property wasn't included in the request at all - which is not really what I expected, I expected to see "null" here, too. Again, I was able to fix this by setting `defaultValue={null}` explicitly.

Can this also be considered a bug?
If yes, should it be a separate issue?

slax57:
Thanks for these clarifications.
I can't say yet if this is linked or not. Let's keep it in the same issue for now.

PR #7673, opened 2022-05-09
Fix `<NumberInput>` edge cases

### LINKED ISSUE #7530: NumberInput does not handle zero after comma correctly

fzaninotto:
## Problem

`<NumberInput>` does not handle zero after comma correctly (cf #7530)

## Solution

A controlled Number input cannot handle "transitive" values (e.g. "1.0" while typing "1.02") properly.

An uncontrolled input cannot handle reinitialization properly (this leads to problems e.g. when browsing from one record to the other in the same Edition view)

So the solution is to manage the string value locally, and only modify the form value on blur. This is the path taken by [react-aria's `useNumberField`](https://react-spectrum.adobe.com/react-aria/useNumberField.html#controlled), and I believe the only viable solution. 

Closes #7530, #7625
Supersedes #7537

## Tasks

- [x] Add storybook for `NumberInput` to test it locally
- [x] Use `event.target.valueAsNumber` instead of home-made `parse`
- [x] Manage string value with `useState`, and set form value on blur
- [x] Add more tests
- [x] Fix tests 
- [x] Update NumberInput doc to mention update in blur

Maybe we should also mention that `<input type="number" />` is problematic, [as pointed by this BBC study](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/).

### DIFF
--- a/docs/NumberInput.md
+++ b/docs/NumberInput.md
@@ -17,7 +17,7 @@ import { NumberInput } from 'react-admin';
 
 `<NumberInput>` converts the input value to a number (integer or float) *on blur*. This is because if the input updates the form value on every keystroke, it will prevent users from entering certain float values. For instance, to enter the number `1.02`, a user would type `1.0`, that JavaScript converts to the number `1`.
 
-If you need the form value to update on change instead of on blur (for instane to update another input based on the number input value), you can build your own number input using `<TextInput>`, and the `format` and `parse` props. But be aware that this only works for integer values. 
+If you need the form value to update on change instead of on blur (for instance to update another input based on the number input value), you can build your own number input using `<TextInput>`, and the `format` and `parse` props. But be aware that this only works for integer values. 
 
 ## Properties
 
--- a/docs/NumberInput.md
+++ b/docs/NumberInput.md
@@ -5,17 +5,19 @@ title: "The NumberInput Component"
 
 # `<NumberInput>`
 
-`<NumberInput>` translates to an HTML `<input type="number">`.
-
-![NumberInput](./img/number-input.gif)
-
-It is necessary for numeric values because of a [known React bug](https://github.com/facebook/react/issues/1425), which prevents using the more generic [`<TextInput>`](./TextInput.md) in that case.
+`<NumberInput>` translates to an HTML `<input type="number">`, and converts the user input to a number.
 
 ```jsx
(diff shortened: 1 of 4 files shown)


djhi (changes requested):
:+1: There is an eslint warning left

docs/NumberInput.md
```diff
@@ -5,18 +5,20 @@ title: "The NumberInput Component"
...
 import { NumberInput } from 'react-admin';
 
 <NumberInput source="nb_views" />
 ```
 
+![NumberInput](./img/number-input.gif)
+
+`<NumberInput>` converts the input value to a number (integer or float) *on blur*. This is because if the input updates the form value on every keystroke, it will prevent users from entering certain float values. For instance, to enter the number `1.02`, a user would type `1.0`, that JavaScript converts to the number `1`.
+
+If you need the form value to update on change instead of on blur (for instane to update another input based on the number input value), you can build your own number input using `<TextInput>`, and the `format` and `parse` props. But be aware that this only works for integer values.
```
djhi, line 20:
```suggestion
If you need the form value to update on change instead of on blur (for instance to update another input based on the number input value), you can build your own number input using `<TextInput>`, and the `format` and `parse` props. But be aware that this only works for integer values. 
```

djhi (changes requested):

PR #7681, opened 2022-05-10
Fix button color warning

smeng9:
We should support additional colors and custom palettes. 
Screenshot from 2022-05-10 15-25-40


djhi (approved):
Thanks!

PR #7682, opened 2022-05-11
Fixing textinput changing 'on' into 'false'

Andonyaa:
Hello,
I came across a bizarre issue while implementing the last version of react-admin. Indeed, if the content of a any Input containing text is "on" it will be replaced by "false" (the bug was case sensitive, it can be verified on the [demo login page](https://marmelab.com/react-admin-demo/#/login) for example).
I tried and fixed it in my personal repository, so i'm submitting this pull request to help other encoutering this issue.


slax57 (approved):
Nice catch! Thanks!
I'll just wait for other reviewers to confirm they too agree with the fix you chose

djhi (approved):
Thanks!

PR #7701, opened 2022-05-17
Add middlewares support to create and edit controllers

djhi:


slax57 (approved):
👏

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -152,6 +161,8 @@ export const useCreateController = <RecordType extends RaRecord = RaRecord>(
         resource,
         record: recordToUse,
         redirect: finalRedirectTo,
+        addMiddleware: addMiddleware,
```
fzaninotto, line 164: I'd call it `registerMutationMiddleware` and the other `unregisterMutationMiddleware`

packages/ra-core/src/controller/saveContext/useAddMiddleware.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect } from 'react';
+import { Middleware } from './useMutationMiddlewares';
+import { useSaveContext } from './useSaveContext';
+
+/**
+ * Internal hook that registers a middleware for the save function in the current SaveContext.
+ * @param callback The middleware function.
+ */
+export const useAddMiddleware = <
```
fzaninotto, line 9: "middleware" is generic. Middleware for what?

I'd call it `useRegisterMutationMiddleware`

packages/ra-core/src/controller/saveContext/useMutationMiddlewares.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const removeMiddleware = useCallback((callback: Middleware<MutateFunc>) => {
+        callbacks.current = callbacks.current.filter(cb => cb !== callback);
+    }, []);
+
+    const getMutateWithMiddlewares = useCallback((fn: MutateFunc) => {
+        return (...args: Parameters<MutateFunc>) => {
+            let index = callbacks.current.length - 1;
+
+            const next = (...newArgs: any) => {
```
fzaninotto, line 23: image

This magic code is a bot hard to reread, would you mind adding a bit of comments?

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -31,11 +28,26 @@ describe('useEditController', () => {
...
-                </SaveContextProvider>
+                <EditController {...defaultProps}>
+                    {({
+                        record,
+                        save,
+                        saving,
+                        addMiddleware,
+                        removeMiddleware,
+                    }) => (
+                        <SaveContextProvider
```
fzaninotto, line 39: Why do we need all that? We don't register anything, it shouldn't be required.

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -61,13 +73,26 @@ describe('useEditController', () => {
...
-                                    {({ record }) => (
+                            <EditController resource="posts">
+                                {({
+                                    record,
+                                    save,
+                                    saving,
+                                    addMiddleware,
+                                    removeMiddleware,
+                                }) => (
+                                    <SaveContextProvider
```
fzaninotto, line 84: Same. Why would we add a SeveContext if we don't use it?

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -751,4 +914,72 @@ describe('useEditController', () => {
...
+            .fn()
+            .mockImplementationOnce((_, { id, data }) =>
+                Promise.resolve({ data: { id, ...data } })
+            );
+        const dataProvider = testDataProvider({
+            // @ts-ignore
+            getOne: () => Promise.resolve({ data: { id: 12 } }),
+            update,
+        });
+        const middleware: Middleware<ReturnType<typeof useUpdate>[0]> = jest.fn(
```
fzaninotto, line 930: You should document that type in the hook jsDoc, or we'll look for it every time

fzaninotto (changes requested):

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -31,11 +28,15 @@ describe('useEditController', () => {
...
             <CoreAdminContext dataProvider={dataProvider}>
-                <SaveContextProvider value={saveContextValue}>
-                    <EditController {...defaultProps}>
-                        {({ record }) => <div>{record && record.title}</div>}
-                    </EditController>
-                </SaveContextProvider>
+                <EditController {...defaultProps}>
+                    {({
+                        record,
+                        save,
```
fzaninotto, line 34: why do you unstructure save, saving and the others if you don't use them? I don't see the purpose of this change. Same for the following tests.

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -224,17 +224,21 @@ describe('useEditController', () => {
...
-                            saveCallback = save;
-                            return <>{JSON.stringify(record)}</>;
-                        }}
-                    </EditController>
-                </SaveContextProvider>
+                <EditController {...defaultProps} mutationMode="pessimistic">
+                    {({ save, saving, record }) => {
+                        saveCallback = save;
+                        return (
+                            <SaveContextProvider
```
fzaninotto, line 231: you don't need a SaveContext for this test. Same for some of the following tests

fzaninotto (approved):

fzaninotto:
\o/

PR #7707, opened 2022-05-18
Fix ReferenceInput Stuck On Loading State

djhi:


packages/ra-core/src/controller/useReference.ts
```diff
@@ -1,12 +1,14 @@
...
 
-interface UseReferenceProps {
+interface UseReferenceProps<RecordType extends RaRecord = RaRecord> {
     id: Identifier;
     reference: string;
+    options?: UseQueryOptions<RecordType[], Error>;
 }
 
-export interface UseReferenceResult<RecordType extends RaRecord = any> {
+export interface UseReferenceResult<RecordType extends RaRecord = RaRecord> {
```
fzaninotto, line 11: why did you change that? The signature of all our other hooks uses `any` as default value.

fzaninotto (changes requested):

packages/ra-core/src/controller/useReference.ts
fzaninotto: we did the opposite when refactoring for 4.0 everywhere... this component shouldn't take another direction on its own.

fzaninotto: cf https://github.com/marmelab/react-admin/pull/7078

fzaninotto (approved):

PR #7722, opened 2022-05-20
fix(typescript): missing use mutation error typing definition

soullivaneuh:
Missing parts of https://github.com/marmelab/react-admin/pull/7698


djhi (approved):
Thanks!

PR #7724, opened 2022-05-20
docs(create): missing button type on custom save button

### LINKED ISSUE #7717: Form input are not cleaned up anymore since v4
I created a special resource to send SMS that does contains only a create component. This one is configured on the `list` property so I have it as a main page.

**What you were expecting:**

When I submit the form, I expect the form fields to be cleaned up as on v3.

**What happened instead:**

The form data is process but the value are still present on the inputs.

**Steps to reproduce:**

1. `make run-simple`
2. Go to http://localhost:8080/#/comments
3. Fill the form with valid data
4. Submit

**Related code:**

https://github.com/pocivaneuh/react-admin/pull/3

**Other information:**

The easiest way would be to trigger the same behavior as before.

As an another option that can lead to a global feature, we should expose a bit the returned value of [`useForm`](https://react-hook-form.com/api/useform) in order to access to some controls like the [`reset`](https://react-hook-form.com/api/useform/reset) function.

**Environment**

* React-admin version: v4.1
* Last version that did not exhibit the issue (if applicable): v3
* React version: v17

soullivaneuh:
closes #7717

### DIFF
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -448,6 +448,7 @@ const PostCreateToolbar = () => {
     return (
         <Toolbar>
             <SaveButton
+                type="button"
                 label="post.action.save_and_add"
                 variant="text"
                 mutationOptions={{
@@ -478,3 +479,5 @@ const PostCreate = () => {
 {% endraw %}
 
 You can also leave the choice to the user, by supplying two submit buttons: one with a redirect, and one with a form reset. The same technique applies: use the `mutationOptions` prop on the `<SaveButton>` component.
+
+Note: In order to get the `mutationOptions` being considered, you have to set the `type` prop of the `SaveButton` to `button`.


djhi (changes requested):
Thanks! Can you please revert the unnecessary changes ?

soullivaneuh:
@djhi The other changes are just unnecessary space removal. They are generated by my IDE and I put that on a separated commit: https://github.com/marmelab/react-admin/pull/7724/commits/fbbb648158d7c9b4f0d79735baf67ff151c1f5d1

I may remove it, but why not keep the cleanup?

By the way, it would be a nice addition to have a tool to automatically remove them, it will avoids that kind of changes.

soullivaneuh:
Commit is dropped.

djhi (approved):

djhi:
Thanks!

Issue #7697, opened 2022-05-16
Global and input level validation can not be combined


soullivaneuh:
This is a follow up of https://github.com/marmelab/react-admin/issues/6270, closed because of a v3 environment bug report: https://github.com/marmelab/react-admin/issues/6270#issuecomment-1126962443

> I'll be happy to close this issue and re-open if I notice this behaviour in v4

@andrico1234 the behavior I reported before is for v4, so here is a new one! :-)

**What you were expecting:**

Using both global and input level validations on a same form should merge the behavior.

**What happened instead:**

Only the global validation is executed

**Steps to reproduce:**

1. `make run-simple`
2. Go to http://localhost:8080/#/comments/create
3. Fill the "Author" field with `foo`
4. Submit the form

The "Author" form should trigger an error because of `minLength(10)` input level configuration.

**Related code:**

https://github.com/pocivaneuh/react-admin/pull/2

**Other information:**

**Environment**

* React-admin version: v4
* Last version that did not exhibit the issue (if applicable): v3
* React version: v17

slax57:
Thank you for taking the time to update this issue.

Actually, it appears that this is a known limitation of the react-hook-form library, and this is documented in the [migration guide here](https://marmelab.com/react-admin/Upgrade.html#validation-form-level--input-level-are-mutually-exclusive):
> Validation: Form Level & Input Level Are Mutually Exclusive

But, I have to admit this does not appear clearly enough (IMO) in the [Form Validation doc](https://marmelab.com/react-admin/Validation.html).
I'll do a quick PR to have this highlighted in this section of the docs as well.

soullivaneuh:
@slax57 Well, this is not really convenient and reduce a lot the interest of the input level validation.

Here is my use case with a discount code form:

```typescript
const validate: SimpleFormProps['validate'] = (values) => {
  const errors: any = {};

  if (values.beginAt && values.endAt && values.beginAt >= values.endAt) {
    errors.endAt = 'Doit être supérieure à la date de début';
  }

  return errors;
};

export const DiscountCodeForm: FC<Omit<SimpleFormProps, 'children'>> = (props) => (
  <SimpleForm
    {...props}
    validate={validate}
  >
    <TextInput
      source="code"
      validate={[
        required(),
        minLength(5),
        maxLength(20),
        regex(/^[a-z0-9]+$/i, 'Uniquement des lettres et des chiffres'),
      ]}
    />
    <TextInput
      source="label"
      validate={[
[...]

PR #7726, opened 2022-05-20
Doc: Fix Async Validation per Input example

slax57:
Related to https://github.com/marmelab/react-admin/issues/7697

### DIFF
--- a/docs/Validation.md
+++ b/docs/Validation.md
@@ -282,7 +282,7 @@ const emailValidators = [required(), validateEmailUnicity];
 
 export const UserCreate = () => (
     <Create>
-        <SimpleForm validate={validateUserCreation}>
+        <SimpleForm>
             ...
             <TextInput label="Email" source="email" validate={emailValidators} />
             ...


docs/Validation.md
```diff
@@ -282,7 +282,7 @@ const emailValidators = [required(), validateEmailUnicity];
 
 export const UserCreate = () => (
     <Create>
-        <SimpleForm validate={validateUserCreation}>
```
djhi, line 285: The `validateUserCreation` does not exist in this example

slax57: ??

soullivaneuh:
@slax57 In my opinion, this should not reference the issue as `fixes` because it will lead to auto-close. However, we agreed that the validator combination should be a think: https://github.com/marmelab/react-admin/issues/7697#issuecomment-1131519902

docs/Validation.md
djhi: nvm

djhi (approved):

PR #7727, opened 2022-05-20
Fix global theme overrides example in docs

mediafreakch:
Fixes a missing key in global theme overrides.

relates to #7604


djhi (approved):
Thanks!

fzaninotto:
Thanks!

PR #7741, opened 2022-05-23
Fix Form Level Validation Support For Complex Messages

### LINKED ISSUE #7690: The validate errors message are not display when passed as translatable object
**What you were expecting:**

With a the following simple `validate` function:

```typescript
validate={() => ({
    body: {
        message: 'ra.validation.minValue',
        args: { min: 42 },
    },
})}
```

I should have the translation of the `ra.validation.minValue` key with the `min` argument correctly placed.

Related doc: https://marmelab.com/react-admin/Form.html#validate

**What happened instead:**

The error is not displayed on the input help section.

Note: works with string only error keys.

**Steps to reproduce:**

1. `make install run-simple`
2. Go to http://localhost:8080/#/comments/create
3. Submit the form regardless the data you put on it
4. Submit

https://github.com/pocivaneuh/react-admin/pull/1

**Other information:**

**Environment**

* React-admin version: 4.0.3
* React version: 18

djhi:
Fixes #7690

### DIFF
--- a/packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
+++ b/packages/ra-core/src/form/getSimpleValidationResolver.spec.ts
@@ -1,22 +1,50 @@
 import {
     getSimpleValidationResolver,
-    flattenKeys,
+    flattenErrors,
 } from './getSimpleValidationResolver';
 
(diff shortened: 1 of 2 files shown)


slax57 (approved):
👌

PR #7750, opened 2022-05-24
Simplify Buttons Props

djhi:


slax57 (approved):

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -38,7 +32,9 @@ export type LocationDescriptor = Partial<Path> & {
  * </Button>
  *
  */
-export const Button = (props: ButtonProps) => {
+export const Button = <RecordType extends RaRecord = RaRecord>(
```
fzaninotto, line 35: should default to any

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -102,7 +98,7 @@ export const Button = (props: ButtonProps) => {
     );
 };
 
-interface Props {
+interface Props<RecordType extends RaRecord = RaRecord> {
```
fzaninotto, line 101: Same, I think this is a breaking change

packages/ra-ui-materialui/src/button/Button.tsx
djhi: Actually, I failt to see why there is a `record` prop. This should be handled by the specialized buttons only imo

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -114,13 +110,16 @@ interface Props {
     size?: 'small' | 'medium' | 'large';
     redirect?: RedirectionSideEffect;
     variant?: string;
-    // May be injected by Toolbar
-    record?: RaRecord;
+    // May be provided manually
+    record?: RecordType;
```
fzaninotto, line 114: In fact, I fail to see what `record` does here. It's not even used in the code. This somehow seems to forbid usage of the generic `Button` outside of a `record` context, e.g. in a List, a dialog, a sidebar, or a custom page. I wonder why, as `Button` is purely a UI component.

fzaninotto:
Could you please update the description to explain what problem this PR fixes?

PR #7751, opened 2022-05-24
chg: :memo: Remove label from ReferenceArrayInput

adrien-may:
ReferenceArrayInput does not have a "label" props anymore as stated above on that same file.

### DIFF
--- a/docs/ReferenceArrayInput.md
+++ b/docs/ReferenceArrayInput.md
@@ -97,7 +97,6 @@ You can tweak how this component fetches the possible values using the `page`, `
 
 ```jsx
 <ReferenceArrayInput
-    label="Tags"
     reference="tags"
     source="tags"
     enableGetChoices={({ q }) => (q ? q.length >= 2 : false)}


djhi (approved):
Indeed. Thanks!

adrien-may:
Wait wait wait !

I just realized label is not really deprecated on this field (see ra-material-ui props on that field).
In fact if you use it in a filter form, you still need to provide a label if you want a custom label when it’s stacked under the « add filter » button. Unless there are incoming changes to ra-material-ui but the « add filter » options would still need this.

Anyway sorry for the confusion, I was about to close my PR.

djhi:
My fault! Can you make another PR then?

adrien-may:
I'm not quite sure if it should be reverted. It leads to misdirection.
Label should live in its children if we want to provide a label for this input. This is clearly stated in the documentation and I'm ok with that.
However, a label is still needed on the ReferenceArrayInput IF we want a custom label in the "Add Filter" options. This is a bit weird and maybe this problem needs to be address in an other way. (Or just stay as it is)

PR #7764, opened 2022-05-30
[TypeScript] Fix Button component props accepts a record

fzaninotto:
## Problem

The `<Button>` component accepts a `record` prop for historical reasons (it used to be injected by the Toolbar), but it doesn't make sense since it is purely a presentational component and it can be used outside of a Record context (e.g. for list view buttons)

## Solution

Remove the `record` prop from the `<Button>` component, and re-add it to the descendants of `<Button>` that require it.


slax57 (approved):
Otherwise, that's good for me! 💯

packages/ra-ui-materialui/src/button/ShowButton.tsx
```diff
@@ -58,14 +61,17 @@ const defaultIcon = <ImageEye />;
...
-    record?: RaRecord;
+    record?: RecordType;
     scrollToTop?: boolean;
 }
 
-export type ShowButtonProps = Props & ButtonProps;
+export type ShowButtonProps<RecordType extends RaRecord = any> = Props<
+    RecordType
+> &
+    ButtonProps;
```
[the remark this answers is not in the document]
fzaninotto: ButtonProps already extends MuiButtonProps

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -110,22 +103,16 @@ interface Props<RecordType extends RaRecord = RaRecord> {
     size?: 'small' | 'medium' | 'large';
     redirect?: RedirectionSideEffect;
```
djhi, line 104: This should probably be removed too

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -110,22 +103,16 @@ interface Props<RecordType extends RaRecord = RaRecord> {
...
     variant?: string;
-    // May be provided manually
-    record?: RecordType;
     resource?: string;
     mutationMode?: MutationMode;
```
djhi, line 107: Those two as well

djhi (changes requested):

packages/ra-ui-materialui/src/button/CloneButton.tsx
```diff
@@ -61,12 +56,12 @@ const sanitizeRestProps = ({
 }: Omit<CloneButtonProps, 'label' | 'scrollToTop' | 'icon'>) => rest;
 
 interface Props {
-    record?: Partial<RaRecord>;
+    record?: any;
```
djhi, line 59: We should probably prefer `Record<string, unknown>` instead of `any`. Same everywhere we accept a `RecordType`

djhi (changes requested):

djhi (approved):

PR #7774, opened 2022-06-01
Fix `Fields` doc reference to `getMany()` usage in v2.x

mfaywu:
Same issue as https://github.com/marmelab/react-admin/pull/7773 

GET_MANY is [0] `posts?filter={"ids":[123,124,125]}` and ReferenceField uses GET_MANY, so the ReferenceField doc [1] shouldn't `posts?ids=[789,735]`

[0] https://github.com/marmelab/react-admin/blob/2.x/docs/DataProviders.md?plain=1#L343
[1] https://github.com/marmelab/react-admin/blob/2.x/docs/Fields.md?plain=1#L543

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -540,7 +540,7 @@ You can also prevent `<ReferenceField>` from adding link to children by setting 
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


mfaywu:
Hm, why does it say I'm trying to deploy? I'm just trying to open a PR against the 2.x branch

fzaninotto:
Thanks!

PR #7775, opened 2022-06-01
Fix `Fields` doc reference to `getMany()` usage in v1.x

mfaywu:
Same issue as https://github.com/marmelab/react-admin/pull/7773 

GET_MANY is [0] `posts?filter={"ids":[123,124,125]}` and ReferenceField uses GET_MANY, so the ReferenceField doc [1] shouldn't `posts?ids=[789,735]`

[0] https://github.com/marmelab/react-admin/blob/1.x/docs/RestClients.md?plain=1#L68
[1] https://github.com/marmelab/react-admin/blob/1.x/docs/Fields.md?plain=1#L432

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -429,7 +429,7 @@ You can also prevent `<ReferenceField>` from adding link to children by setting 
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks! Although, as we don't maintain react-admin v1 anymore, this change will not be published.

PR #7789, opened 2022-06-06
Add premium icon to DualListInput

davidhenley:


slax57 (approved):

slax57:
Indeed we missed that one, thanks!

Issue #7782, opened 2022-06-03
Warning when using null value on TextInput


mediafreakch:
**What you were expecting:**

When my API returns `null` as the value for a `<TextInput />`, it should not lead to any warnings.

**What happened instead:**

When a record is returned by the API with a value of `null`, this leads to a warning when used on a text input:

> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.

and 

> Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component.

**Steps to reproduce:**

1. Go to the sandbox provided
2. Open any post to edit
3. Observe console output (you may need to filter for `should not be null`)

**Related code:**

https://codesandbox.io/s/hopeful-worker-u1dxlq?file=/src/posts/PostEdit.tsx

**Other information:**

- Setting the `defaultValue` of the input, or `defaultValues` of the form to an empty string (`''`), doesn't make the warning go away
- From the migration docs it's clear that the behaviour changed in regard of how `null` values are treated by react-form-hook compared to react-final-form. But it's unclear to me how to mitigate this warning.

**Environment**

* React-admin version: 4.1.2
* Last version that did not exhibit the issue (if applicable): 3.19.x
* React version: 17

septentrion-730n:
Thanks for your for submitting this.

winniecarole:
Hi I would like to solve this issue .Can I try it?

slax57:
> Hi I would like to solve this issue .Can I try it?

@winniecarole Of course! We gladly accept PRs 🙂 . Just make sure to read the [contribution guidelines](https://github.com/marmelab/react-admin#contributing) if you haven't already 😉

mediafreakch:
Maybe someone else should take this bug?

slax57:
Taking some time to come back on this issue, I noticed that if your dataProvider returns `undefined` instead of `null`, then the warning disappears and everything works as expected (you don't need to add `someOptionalInput: ''` in your `defaultValues`).
In the end I'm more and more convinced this is not a bug (the warning is self-explanatory and justified).
Maybe we could document this gotcha in the migration guide though.
Hence I'll relabel this as a documentation issue.

mediafreakch:
[...]

PR #7800, opened 2022-06-08
Warning when using null value on TextInput in EditPost Issue #7782

winniecarole:
Fixes https://github.com/marmelab/react-admin/issues/7782

## Screenshots

image

### DIFF
--- a/docs/InputNullValue.md
+++ b/docs/InputNullValue.md
@@ -1,0 +1,22 @@
+
+## Null value on Text INput
+
+### go to useEditController.ts
+
+look at the current path .
+
+check that the input data is not empty
+
+#### example for EditPost
+
+const postEditArgument;
+
+if (!postEditArgument && context.snapshot[0][0][0] == 'posts') {
+notify('ra.notification.inputError', {
+type: 'warning',
+messageArgs: { smart_count: 1 },
+undoable: mutationMode === 'undoable',
+});
+}
+
+
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -130,11 +130,21 @@ export const useEditController = <
                     : data
             ).then((data: Partial<RecordType>) => {
                 const mutate = getMutateWithMiddlewares(update);
+                let postEditArgument;
+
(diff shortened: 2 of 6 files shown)


docs/InputNullValue.md
```diff
@@ -0,0 +1,22 @@
...
+
+const postEditArgument;
+
+if (!postEditArgument && context.snapshot[0][0][0] == 'posts') {
+notify('ra.notification.inputError', {
+type: 'warning',
+messageArgs: { smart_count: 1 },
+undoable: mutationMode === 'undoable',
+});
+}
```
slax57, line 20: What is the purpose of this file? :thinking: 

You did not add any new components so there is no need for a dedicated documentation file.

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -130,29 +130,49 @@ export const useEditController = <
...
                     { id, data },
                     {
                         onSuccess: async (data, variables, context) => {
+                            if (context.snapshot[0][0][0] == 'posts') {
+                                postEditArgument =
+                                    data.title !== ' ' &&
+                                    data.teaser !== ' ' &&
+                                    data.title.trim().length !== 0 &&
+                                    data.teaser.trim().length !== 0;
+                            }
```
slax57, line 146: How can you expect this code to work on any other page that the PostEdit page of the demo? :thinking:

slax57 (changes requested):
Sorry but you're heading to the wrong direction imo with this PR. The fix should be around the TextInput component, and should not be specific to the PostEdit page!

Besides, please do not commit IDE specific files:
- .idea/misc.xml
- .idea/workspace.xml

slax57:
> Please tell me where the textInput component is because I can't find it. If possible the file where I have to make the modification

@winniecarole Sorry I do not know this upfront. Event I would have to do somme digging to find out where to make the modification.

Dealing with form inputs is not easy in react-admin. If you are struggling with this fix don't worry, maybe you can start with easier issues to begin with.

WiXSL:
Shouldn't #8188 make this PR unnecessary?

fzaninotto:

PR #7804, opened 2022-06-09
[Doc] Fix custom menu items example

davidhenley:


davidhenley:
`resource` was not available in the current context. You might have to turn this into a component? Not exactly sure.

slax57 (approved):

PR #7812, opened 2022-06-10
Add ability to accept more than one child in `<Reference>` Fields

fzaninotto:
## Problem

Developers often need to render more than one field of the reference table (e.g. if the `users` table has a `first_name` and a `last_name` field) in a `<ReferenceField>`. And `<FunctionField>` is a bit cumbersome. 

## Solution 

Let `<ReferenceField>` accept more than one child, so that developers can use as many `<Field>` as they like:

```jsx
import { Show, SimpleShowLayout, ReferenceField, TextField, DateField, FunctionField } from 'react-admin';

export const PostShow = () => (
    <Show>
        <SimpleShowLayout>
            <TextField source="id" />
            <TextField source="title" />
            <ReferenceField label="Author" source="user_id" reference="users">
                <TextField source="first_name" />{' '}
                <TextField source="last_name" />
            </ReferenceField>
        </SimpleShowLayout>
    </Show>
);
```

- [x] `<ReferenceField>` code, doc and tests
- [x] `<ReferenceManyField>` code, doc and tests
- [x] `<ReferenceOneField>` code, doc and tests
- [x] `<ReferenceArrayField>` code, doc and tests


docs/ReferenceField.md
```diff
@@ -5,101 +5,236 @@ title: "The ReferenceField Component"
...
-| `& .RaReferenceField-link` | Applied to each child element |
+```jsx
+{/* default label is 'User Id', or the translation of 'resources.posts.fields.user_id' if it exists */}
+<ReferenceField source="user_id" reference="users" link="show">
+    <TextField source="name" />
+</ReferenceField>
+```
 
-To override the style of all instances of `<ReferenceField>` using the [MUI style overrides](https://mui.com/customization/globals/#css), use the `RaReferenceField` key.
+That's why you often need to set an explicit `label` on a `<RefezrenceField>`:
```
slax57, line 79:
```suggestion
That's why you often need to set an explicit `label` on a `<ReferenceField>`:
```

slax57 (changes requested):
Otherwise, that's great! And we can probably do the same with ReferenceManyField too

docs/ReferenceField.md
```diff
@@ -5,101 +5,236 @@ title: "The ReferenceField Component"
...
-import { List, Datagrid, ReferenceField, TextField, EditButton } from 'react-admin';
+<ReferenceField source="user_id" reference="users">
+    <TextField source="name" />
+</ReferenceField>
+```
 
-export const PostList = () => (
-    <List>
-        <Datagrid>
+A `<ReferenceField>` displays nothing on its own, it just fetches the data, pust it in a [`RecordContext`](./useRecordContext.md), and lets its child to render it. Usual child components for `<ReferenceField>` are other `<Field>` components (e.g. [`<TextField>`](./TextField.md)).
```
djhi, line 27:
```suggestion
A `<ReferenceField>` displays nothing on its own, it just fetches the data, puts it in a [`RecordContext`](./useRecordContext.md), and lets its children render it. Usual child components for `<ReferenceField>` are other `<Field>` components (e.g. [`<TextField>`](./TextField.md)).
```

djhi (changes requested):

slax57 (approved):
Build is failing though :slightly_frowning_face:

fzaninotto:
Back to WIP to address other reference field components

fzaninotto:
back to RFR

docs/ReferenceArrayField.md
```diff
@@ -29,24 +63,147 @@ export const PostList = () => (
...
+| `filter`     | Optional | `Object`            | -        | Filters to use when fetching the related records (the filtering is done client-side)                         |
+| `pagination` | Optional | `Element`           | -        | Pagination element to display pagination controls. empty by default (no pagination)                          |
+| `perPage`    | Optional | `number`            | 1000     | Maximum number of results to display                                                                         |
+| `sort`       | Optional | `{ field, order }`  | `{ field: 'id', order: 'DESC' }` | Sort order to use when displaying the related records (the sort is done client-side) |
 
 `<ReferenceArrayField>` also accepts the [common field props](./Fields.md#common-field-props), except `emptyText` (use the child `empty` prop instead).
 
+## `filter`
+
+`<ReferenceArrayField>` fetches all the related records, and displays them all, too. You can use the `filter` prop to filter the list of related records to dispaly (this works by filtering the records client-side, after the fetch).
```
djhi, line 85:
```suggestion
`<ReferenceArrayField>` fetches all the related records, and displays them all, too. You can use the `filter` prop to filter the list of related records to display (this works by filtering the records client-side, after the fetch).
```

djhi (changes requested):
:100:

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
```diff
@@ -138,6 +136,7 @@ ReferenceManyField.defaultProps = {
     source: 'id',
 };
 
+// FIXME kept for backwards compatibility, unused, to be removed in v5
```
slax57, line 139: Shouldn't we add some JSDoc to warn that this component is deprecated?

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, ReactNode } from 'react';
```
slax57, line 1: `ReactElement` is unused

Issue #7806, opened 2022-06-09
Create ImageButtons and ColorButtons for RichTextInputToolbar (as previously available in Quill re v3)


bladx:
**Is your feature request related to a problem? Please describe.**
This is not an issue, more like a regression I would say as we lost 2 interesting options Quill was offering back in ra v3: Image and Color (font and highlight).

**Describe the solution you'd like**
Same as there are already some buttons defined for the RichTextInputToolbar (FormatButtons, AlignmentButtons, etc.), I would like to have the ColorButtons and ImageButtons.

**Describe alternatives you've considered**
I implemented it myself using the TipTap extensions guide + ra source code for existing buttons.
I would love to share the files I humbly wrote.

**Additional context**
Please find screenshots of what I have as a result.
![Capture d’écran 2022-06-09 à 23 48 50](https://user-images.githubusercontent.com/4518629/172951275-89da7cb8-997c-4ca0-b587-948dded162ee.png)
![Capture d’écran 2022-06-09 à 23 50 01](https://user-images.githubusercontent.com/4518629/172951283-fa23f510-a70f-423a-bd9f-7134a1f26018.png)

bladx:
Hey, do you know if both the french translations and the flex-wrap/responsiveness will be include in the next release?
Also what is the process of merging `next` into `master` branch?
When can I expect to use that in my project?
Thanks a lot for the code reviews, I was really happy to contribute a bit.

slax57:
@bladx 
- French translations: can't say yet if this is going to be included to 4.2.0, but this is in our backlog
- flex-wrap/responsiveness: this on the other hand wasn't yet on our backlog. But I don't think it would be too hard to do. Would you like to open another PR for this? :slightly_smiling_face: 
- We don't yet have a release date for v4.2.0, but usually we release a new minor version approx. each month.

bladx:
Thanks for the quick answer!
I just submitted my PR.

PR #7816, opened 2022-06-11
[RFR] Add rich text input toolbar buttons

bladx:
As discussed in https://github.com/marmelab/react-admin/issues/7806, here is the PR to have both ColorButtons (including font and highlight color) and ImageButtons components for RichTextInput toolbar.

This is my first PR ever on any non-private repository, so please feel free to make me some feedbacks (english or french).

Plus, I noticed that the translation of all buttons (existing ones as well) do not exist, so even in french the titles are in english (default).

### DIFF
--- a/packages/ra-input-rich-text/src/buttons/ColorButtons.tsx
+++ b/packages/ra-input-rich-text/src/buttons/ColorButtons.tsx
@@ -21,50 +21,59 @@ import {
     purple,
 } from '@mui/material/colors';
 
-/**
- * Hook that listens clicks outside of the passed ref
- */
-const useOutsideListener = (
-    ref: React.MutableRefObject<any>,
-    onClick: () => void
-) => {
-    React.useEffect(() => {
-        const handleClickOutside = (event: MouseEvent) => {
-            if (ref.current && !ref.current.contains(event.target)) {
-                onClick();
-            }
-        };
-        // Bind the event listener
-        document.addEventListener('mousedown', handleClickOutside);
-        return () => {
-            // Unbind the event listener on clean up
-            document.removeEventListener('mousedown', handleClickOutside);
-        };
-    }, [ref, onClick]);
-};
+export const ColorButtons = (props: Omit<ToggleButtonProps, 'value'>) => {
+    const translate = useTranslate();
+    const editor = useTiptapEditor();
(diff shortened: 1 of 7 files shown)


bladx:
I also just noticed that the toolbar is not responsive (missing the flex-wrap so the screen has an horizontal scroll bar).

packages/ra-input-rich-text/src/buttons/ColorButtons.tsx
```diff
@@ -0,0 +1,194 @@
...
+ * Hook that listens clicks outside of the passed ref
+ */
+const useOutsideListener = (
+    ref: React.MutableRefObject<any>,
+    onClick: () => void
+) => {
+    React.useEffect(() => {
+        /**
+         * Alert if clicked on outside of element
+         */
```
slax57, line 34: This comment seems out of place, I think you can simply remove it

slax57 (changes requested):
Otherwise this is a really great contribution, thanks! :pray:

bladx:
Thanks for the kind feedback :)

djhi:
> Plus, I noticed that the translation of all buttons (existing ones as well) do not exist, so even in french the titles are in english (default).

Yes, thank you for this reminder. We're going to include english and french translations directly inside this package.

packages/ra-input-rich-text/src/buttons/ColorButtons.tsx
```diff
@@ -0,0 +1,191 @@
...
+                            onClick={() => selectColor(color[shade])}
+                        ></Box>
+                    ))}
+                </Box>
+            ))}
+        </Card>
+    );
+};
+
+export const ColorButtons = (props: Omit<ToggleButtonProps, 'value'>) => {
```
djhi, line 138: Would you mind moving this component at the top of the file ? We usually put first the most important things.

djhi (changes requested):
Thanks a lot!

slax57 (approved):

fzaninotto:
Could you please fix the linter warnings?

bladx:
> Could you please fix the linter warnings?

Sorry I introduced them in my last code review commit and did not pay attention.
This should be alright now.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #7833, opened 2022-06-15
[Docs] Fix Creating New Choices examples

davidhenley:
This has been bugging me for a while now and I finally figured it out! The data passed to onSuccess should not be deconstructed from props.


davidhenley:
The other documentation examples, SelectInput, AutocompleteArrayInput, etc... are probably wrong too. This is just the only one I use.

djhi:
Thanks!

Issue #7785, opened 2022-06-03
`meta` params in `useListController`


chris-beedie:
**Is your feature request related to a problem? Please describe.**
I would like to be able to pass `meta` params to my data provider when using higher level hooks, specifically `useListController`

**Describe the solution you'd like**
Pass through of `meta` params on call from `useListController` to the inner `useGetList` call

**Describe alternatives you've considered**
calling `useGetList` directly with `meta`, but this means recreating the rest of the `useListController` functionality. 

**Additional context**
N/A

PR #7841, opened 2022-06-16
Add ability to set `meta` in page components

### LINKED ISSUE #7357: Pass down custom graphql query via Edit, Show, Create Component
React admin is a great library. We have built a large app out of it

We are changing our app from REST to GraphQL. We have lots of custom views, and **approx. 100 REST API resources.** A view (Front end UI) maybe shown differently to each person depending on some logic.

While moving from REST to GraphQL, we found two problems, 
1. We did not find any props to pass custom GraphQL Query or Mutation down to the DataProvider for Edit, Create and Show Component. 
2. In the [ra-simple-graphql](https://github.com/marmelab/react-admin/tree/master/packages/ra-data-graphql-simple#overriding-a-specific-query) it is advised to use resource and type of DataProvider to handle custom queries. **If we use this approach then we may need to write Number of resources x Number of DataProvider Methods,  If/else Conditions.
 (100 * 6 = 600) in my case.**

There for we were hoping that we could call custom GraphQL Query which will be passed down via Edit or Show Component. Then we may need not write to 600 Conditions, It will be much less i.e only 6 conditions or 10 conditions.
image

For example

```

const customGQLQuery = gql`

query User{
  User {
   id,
   address
   }

`

<Edit customGQLQuery = {customGQLQuery}>
<Edit />


const getOne = async (resource,params) => {

if(params.customGQLQuery){
  // execute custom gql Query

}
else {
 // use default query builder

}

```

Since we could not find any ways to pass props down to DataProvider, we have to write lot of code otherwise graphql will call it's default query and bring fields we may not need for the UI
 

Ability to pass custom prop via for custom GraphQL Query<Edit/>, <Create/> and <Show/>

For List component I have used filters to pass custom graphql query.

**Additional context**

It is also possible that I have missed something and my problems solution is already available. In that case pls point out to me specific resource

fzaninotto:
- [x] Add support for `meta` in `<Create>`
- [x] Add support for `meta` in `<Show>`
- [x] Add support for `meta` in `<List>`
- [x] Add support for `meta` in `<Edit>`
- [x] Add documentation
- [x] Add tests

```jsx
import { Edit, SimpleForm } from 'react-admin';

const PostEdit = () => (
    <Edit mutationOptions={{ meta: { foo: 'bar' } }}>
        <SimpleForm>
            ...
        </SimpleForm>
    </Edit>
);
```

Closes #7357
Refs #7785, #7665

### DIFF
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -152,7 +152,23 @@ const PostCreate = () => (
 
 ## `mutationOptions`
 
-You can customize the options you pass to react-query's `useMutation` hook, e.g. to override success or error side effects, by setting the `mutationOptions` prop. Refer to the [useMutation documentation](https://react-query.tanstack.com/reference/useMutation) in the react-query website for a list of the possible options.
+You can customize the options you pass to react-query's `useMutation` hook, e.g. to pass [a custom `meta`](./Actions.md#meta-parameter) to the `dataProvider.create()` call.
+
+{% raw %}
+```jsx
+import { Create, SimpleForm } from 'react-admin';
+
+const PostCreate = () => (
+    <Create mutationOptions={{ meta: { foo: 'bar' } }}>
+        <SimpleForm>
+            ...
+        </SimpleForm>
+    </Create>
+);
+```
+{% endraw %}
+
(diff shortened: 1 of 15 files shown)


djhi (approved):

gengue:
This is great 🥇 
It would be really useful to have this `meta` prop in Reference inputs as well, most of the time we just need `id` and `name` to be rendered in a `Select` or `Autocomplete`, but the dataProvider will fetch the default list

Issue #7839, opened 2022-06-16
User panel displays a misleading "Logout" invite when user is not logged in.


septentrion-730n:
**What you were expecting:**
User panel should display a "Login" invite when user is not logged in.

**What happened instead:**
User panel displays a misleading "Logout" invite when user is not logged in.

**Steps to reproduce:**

(https://codesandbox.io/s/stupefied-volhard-zeedw3)

<img width="614" alt="Capture d’écran 2022-06-16 à 05 49 40" src="https://user-images.githubusercontent.com/102964006/173995115-b7559a17-59f7-404e-8bfc-036ddb90ce8f.png">

- Ensure to be logged out
- If necessary logout then use history to go back to home (posts list opened with [disableAuthentication](https://marmelab.com/react-admin/Authentication.html#allowing-anonymous-access))

**Environment**
* React-admin version: 4.1.3

PR #7842, opened 2022-06-16
Fix Hide `<Logout />` button if irrelevant.

septentrion-730n:
Closes https://github.com/marmelab/react-admin/issues/7839

Hide logout button if irrelevant.

### DIFF
--- a/examples/simple/src/posts/PostList.tsx
+++ b/examples/simple/src/posts/PostList.tsx
@@ -100,7 +100,6 @@ const PostList = () => {
     const isSmall = useMediaQuery<Theme>(theme => theme.breakpoints.down('md'));
     return (
         <List
-            disableAuthentication
             filters={postFilter}
             sort={{ field: 'published_at', order: 'DESC' }}
             exporter={exporter}
--- a/examples/simple/src/posts/PostList.tsx
+++ b/examples/simple/src/posts/PostList.tsx
@@ -100,6 +100,7 @@ const PostList = () => {
     const isSmall = useMediaQuery<Theme>(theme => theme.breakpoints.down('md'));
     return (
         <List
+            disableAuthentication
             filters={postFilter}
             sort={{ field: 'published_at', order: 'DESC' }}
             exporter={exporter}
--- a/packages/ra-ui-materialui/src/auth/Logout.spec.tsx
+++ b/packages/ra-ui-materialui/src/auth/Logout.spec.tsx
@@ -1,0 +1,37 @@
+import * as React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import { AdminContext } from '../AdminContext';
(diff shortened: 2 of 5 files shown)


examples/simple/src/posts/PostList.tsx
```diff
@@ -100,6 +100,7 @@ const PostList = () => {
     const isSmall = useMediaQuery<Theme>(theme => theme.breakpoints.down('md'));
     return (
         <List
+            disableAuthentication
```
fzaninotto, line 103: Please add a dedicated story instead of modifying the simple example to test it

packages/ra-ui-materialui/src/auth/Logout.tsx
```diff
@@ -34,7 +34,9 @@ export const Logout: FunctionComponent<
         redirectTo,
         logout,
     ]);
-    return (
+    const { authenticated } = useAuthState();
+
+    return !authenticated ? null : (
```
fzaninotto, line 39: Use the *leave early* pattern instead of a ternary to make it clearer.

```jsx
if (!authenticated) return null;
```

fzaninotto (changes requested):
It seems you're missing a unit test...

packages/ra-ui-materialui/src/auth/Logout.spec.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import { AdminContext } from '../AdminContext';
+import { Logout } from './Logout';
+import { UserAuthenticated, UserUnauthenticated } from './Logout.stories';
+
+export default { title: 'ra-ui-materialui/input/AutocompleteArrayInput' };
+
+const MinimalAdmin = authProvider => {
```
djhi, line 9: It seems this is not used

djhi (changes requested):

fzaninotto:
Awesome! Just fix the linter warning and we're good to merge.

packages/ra-ui-materialui/src/auth/Logout.spec.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import { AdminContext } from '../AdminContext';
+import { Logout } from './Logout';
+import { UserAuthenticated, UserUnauthenticated } from './Logout.stories';
+
+export default { title: 'ra-ui-materialui/input/AutocompleteArrayInput' };
```
slax57, line 7: This is wrong + you don't need this unless you're writing a story

slax57 (changes requested):

slax57 (approved):

djhi (approved):

PR #7847, opened 2022-06-17
Fix `<AutocompleteInput>` storybook does not select newly created option

slax57:
Fix `<AutocompleteInput>` storybook does not select newly created option.

Applies the same fix as https://github.com/marmelab/react-admin/pull/7833 in the storybook.


djhi (approved):

davidhenley:
Sorry about that I forgot that one.

PR #7851, opened 2022-06-17
Add support for <ReferenceField emptyText> when the reference is missing

### LINKED ISSUE #7688: Support emptyText for missing referenced entity as well on ReferenceField
I want to display empty text in case where reference is set but the referenced entity is missing. This is easily a case in systems where soft deleting is allowed.

For now I am using the same `emptyText` as is used in the same case where the reference is not set. This works for me, but maybe best is to have a separate prop. Here is the patch:

```patch
diff --git a/node_modules/ra-ui-materialui/esm/field/ReferenceField.js b/node_modules/ra-ui-materialui/esm/field/ReferenceField.js
index 3dc814e..f44ccae 100644
--- a/node_modules/ra-ui-materialui/esm/field/ReferenceField.js
+++ b/node_modules/ra-ui-materialui/esm/field/ReferenceField.js
@@ -85,7 +85,7 @@ var ReferenceField = function (props) {
     if (!isReferenceDeclared) {
         throw new Error("You must declare a <Resource name=\"" + props.reference + "\"> in order to use a <ReferenceField reference=\"" + props.reference + "\">");
     }
-    return get(record, source) == null ? (emptyText ? (React.createElement(Typography, { component: "span", variant: "body2" }, emptyText)) : null) : (React.createElement(NonEmptyReferenceField, __assign({}, rest, { record: record, source: source })));
+    return get(record, source) == null ? (emptyText ? (React.createElement(Typography, { component: "span", variant: "body2" }, emptyText)) : null) : (React.createElement(NonEmptyReferenceField, __assign({}, rest, { emptyText, record: record, source: source })));
 };
 ReferenceField.propTypes = {
     addLabel: PropTypes.bool,
@@ -157,7 +157,7 @@ export var ReferenceFieldView = function (props) {
         return React.createElement(LinearProgress, null);
     }
     if (!referenceRecord) {
-        return null;
+        return props.emptyText || null;
     }
     if (resourceLinkPath) {
         return (React.createElement(RecordContextProvider, { value: referenceRecord },
```

N\A

**Additional context**
N\A

fzaninotto:
Closes #7688

### DIFF
--- a/docs/ReferenceField.md
+++ b/docs/ReferenceField.md
@@ -66,11 +66,27 @@ With this configuration, `<ReferenceField>` wraps the user's name in a link to t
 | `source`    | Required | `string`            | -        | Name of the property to display |
 | `reference` | Required | `string`            | -        | The name of the resource for the referenced records, e.g. 'posts' |
 | `children`  | Required | `ReactNode`         | -        | One or more Field elements used to render the referenced record |
+| `emptyText` | Optional | `string`            | ''       | Defines a text to be shown when the field has no value or when the reference is missing |
 | `label`     | Optional | `string | Function` | `resources.[resource].fields.[source]`   | Label to use for the field when rendered in layout components  |
 | `link`      | Optional | `string | Function` | `edit`   | Target of the link wrapping the rendered child. Set to `false` to disable the link. |
 | `sortBy`    | Optional | `string | Function` | `source` | Name of the field to use for sorting when used in a Datagrid |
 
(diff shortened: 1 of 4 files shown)


djhi (changes requested):
:100: Aren't we missing some documentation?

fzaninotto:
> 💯 Aren't we missing some documentation?

Not anymore

djhi (approved):

PR #7852, opened 2022-06-17
[TypeScript] Fix ra-core missing null checks

fzaninotto:
`ra-core` is still far from compiling with `strictNullChecks`, but we're progressing

### DIFF
--- a/packages/ra-core/src/store/localStorageStore.ts
+++ b/packages/ra-core/src/store/localStorageStore.ts
@@ -53,11 +53,11 @@ export const localStorageStore = (version: string = '1'): Store => {
     // Whenever the local storage changes in another document, look for matching subscribers.
     // This allows to synchronize state across tabs
     const onLocalStorageChange = (event: StorageEvent): void => {
-        if (event.key.substring(0, prefixLength) !== RA_STORE) {
+        if (event.key?.substring(0, prefixLength) !== RA_STORE) {
             return;
         }
         const key = event.key.substring(prefixLength + 1);
-        const value = tryParse(event.newValue);
+        const value = event.newValue ? tryParse(event.newValue) : undefined;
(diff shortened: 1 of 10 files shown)


packages/ra-core/src/store/localStorageStore.ts
```diff
@@ -93,11 +93,12 @@ export const localStorageStore = (version: string = '1'): Store => {
             }
         },
         getItem<T = any>(key: string, defaultValue?: T): T {
-            const valueFromStorage = tryParse(
-                getStorage().getItem(`${RA_STORE}.${key}`)
-            );
+            const valueFromStorage = getStorage().getItem(`${RA_STORE}.${key}`);
+
             // eslint-disable-next-line eqeqeq
```
djhi, line 98: There is a configuration option for that rule that will never warn about `== null`

djhi (approved):

packages/ra-core/src/store/localStorageStore.ts
fzaninotto: nice! let's enable it in another Pr (I didn't add this line)

packages/ra-core/src/store/useRemoveFromStore.ts
```diff
@@ -21,8 +21,18 @@ import { useStoreContext } from './useStoreContext';
...
+        (key?: string) => {
+            if (
+                typeof key === 'undefined' &&
+                typeof hookTimeKey === 'undefined'
+            ) {
+                throw new Error(
+                    'You must provide a key to remove an item from the store'
+                );
+            }
+            // @ts-ignore
```
fzaninotto, line 33: TypeScript is dumb. It should know that either one of the two variables is defined, but its doesn't

djhi (approved):

PR #7858, opened 2022-06-19
add ra-language-german package

grandmember:


djhi:
Hi and thanks for your contribution. We won't merge it though as we can only maintain french and english packages. Besides, it seems you included many unrelated commits.

slax57:
If you wish, you can contribute to the German translation project that can be found in the [Supported Locales page](https://marmelab.com/react-admin/TranslationLocales.html): https://github.com/greenbananaCH/ra-language-german

Issue #7832, opened 2022-06-15
FilterLiveSearchProps should extend TextInputProps


KilianB:
In the FilterLiveSearch component rest props are forwarded to the TextInput field, but since they are not declared typescript is complaining

**What you were expecting:**

Typescript not complaining about the label prop

````ts
<FilterLiveSearch source='user_id' label='Sample label' />
````

**What happened instead:**

````
Type '{ source: string; label: string; }' is not assignable to type 'IntrinsicAttributes & { source?: string | undefined; sx?: SxProps<{}> | undefined; variant?: "filled" | "outlined" | undefined; }'.
  Property 'label' does not exist on type 'IntrinsicAttributes & { source?: string | undefined; sx?: SxProps<{}> | undefined; variant?: "filled" | "outlined" | undefined; }'.
````

farrukhrashid1997:
Hi @septentrion-730n, can I pick this up?

septentrion-730n:
> Hi @septentrion-730n, can I pick this up?

Sorry for the late response, I wanted to check the path  to follow with the team. It ended to be the simplest!

PR #7859, opened 2022-06-20
[TypeScript] Fix FilterLiveSearchProps should extend TextInputProps

septentrion-730n:
Fixes https://github.com/marmelab/react-admin/issues/7832

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
@@ -23,9 +23,15 @@ import { TextInput } from '../../input';
  * );
  */
 export const FilterLiveSearch = memo((props: FilterLiveSearchProps) => {
-    const { source = 'q', variant, ...rest } = props;
     const { filterValues, setFilters } = useListFilterContext();
     const translate = useTranslate();
+
+    const {
+        source = 'q',
+        variant,
+        label = translate('ra.action.search'),
+        ...rest
+    } = props;
 
     const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
         if (event.target) {
@@ -44,8 +50,6 @@ export const FilterLiveSearch = memo((props: FilterLiveSearchProps) => {
     );
 
     const onSubmit = () => undefined;
-    let label = translate('ra.action.search');
-
     return (
         <Form defaultValues={initialValues} onSubmit={onSubmit}>
             <TextInput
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
```diff
@@ -74,9 +78,10 @@ export const FilterLiveSearch = memo((props: FilterLiveSearchProps) => {
     );
 });
 
-export interface FilterLiveSearchProps {
+export interface FilterLiveSearchProps extends Omit<TextInputProps, 'source'> {
     source?: string;
     sx?: SxProps;
+    label?: string;
```
djhi, line 84: I don't think we should override the label prop as we forward it as is to the `TextInput`. We only provide a default

djhi (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
slax57: I would tend to agree with @septentrion-730n because his solution still provides a default value and also supports being overridden by the user directly in the FilterLiveSearch component.
Previously the only way you could customize the label was through localization.

However, it could be nice to handle translation on this field either way.

djhi: The `TextInput` already support all these use cases: https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/TextInput.tsx#L66 via `FieldTitle`. 

Btw, we don't event need to translate here, we could just pass the translation key directly as `label` and it will be translated

packages/ra-ui-materialui/src/list/filter/FilterLiveSearch.tsx
```diff
@@ -23,10 +23,16 @@ import { TextInput } from '../../input';
...
  */
 export const FilterLiveSearch = memo((props: FilterLiveSearchProps) => {
-    const { source = 'q', variant, ...rest } = props;
     const { filterValues, setFilters } = useListFilterContext();
     const translate = useTranslate();
 
+    const {
+        source = 'q',
+        variant,
+        label = translate('ra.action.search'),
```
djhi, line 32: You don't even need to translate here, just pass the translation key directly to the `label` prop. `<TextInput>` will handle the translation

djhi: Nevermind, accepting react node for this label would complicate the code as we only use it as the `TextInput` label when the `variant` is `outlined`. Otherwise, we pass it as the `placeholder` prop

djhi (approved):

Issue #7868, opened 2022-06-21
Prop types does not accept boolean for label props


thibault-barrat:
**What you were expecting:**
As described here https://marmelab.com/react-admin/Inputs.html#common-input-props, I want to hide the label on an `ImageInput` field by setting `label={false}`

**What happened instead:**
The label is hidden but I have this error in my browser : 
![image](https://user-images.githubusercontent.com/41492953/174772424-b156026c-32e2-417d-a2e7-8486d9a0e0ff.png)

**Steps to reproduce:**
Set the props `label={false}` on an `ImageInput` field

**Related code:**
See `PostEdit.tsx` in https://codesandbox.io/s/laughing-chaum-phq8f0

**Environment**

* React-admin version: 4.1.3

slax57:
Reproduced, thanks!

PR #7869, opened 2022-06-21
Fix `FileInput` propTypes for `label` prop

slax57:
Fixes https://github.com/marmelab/react-admin/issues/7868


djhi (approved):

packages/ra-ui-materialui/src/input/FileInput.tsx
```diff
@@ -214,7 +214,7 @@ FileInput.propTypes = {
     className: PropTypes.string,
     id: PropTypes.string,
     isRequired: PropTypes.bool,
-    label: PropTypes.string,
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
djhi, line 217: If I'm not mistaken, this can also be an element

PR #7870, opened 2022-06-21
[Docs] Fix syntax error in code examples

WiXSL:


docs/Inputs.md
```diff
@@ -530,12 +530,12 @@ const theme = {
         ...defaultTheme.components,
         MuiTextField: {
             defaultProps: {
-                variant: 'outlined' as const,
+                variant: 'outlined',
```
fzaninotto, line 533: Good catch! but as the TypeScript validation causes problem, I suggest you add a tip just after this snipper for TS users.

fzaninotto (changes requested):

fzaninotto (approved):

fzaninotto:
Thanks a lot, glad to see you back!

PR #7872, opened 2022-06-21
[fix] TabbedForm passes invalid props to its root element

### LINKED ISSUE #7871: TabbedFormView passes warnWhenUnsavedChanges to styled div, leads to react error
**What you were expecting:**
No console errors when using `<TabbedForm warnWhenUnsavedChanges>`.

**What happened instead:**
The following console error appears:

>Warning: React does not recognize the `warnWhenUnsavedChanges` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `warnwhenunsavedchanges` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    at div
    at http://cp.localhost:3001/static/js/bundle.js:36211:66
    at TabbedFormView (http://cp.localhost:3001/static/js/bundle.js:229006:15)
    at form
    at FormGroupsProvider (http://cp.localhost:3001/static/js/bundle.js:218155:15)
    at FormProvider (http://cp.localhost:3001/static/js/bundle.js:353840:5)
    at LabelPrefixContextProvider (http://cp.localhost:3001/static/js/bundle.js:213242:13)
    at RecordContextProvider (http://cp.localhost:3001/static/js/bundle.js:216310:15)
    at OptionalRecordContextProvider (http://cp.localhost:3001/static/js/bundle.js:216538:12)
    at Form (http://cp.localhost:3001/static/js/bundle.js:218346:15)
    at TabbedForm (http://cp.localhost:3001/static/js/bundle.js:229099:11)
    ...

**Steps to reproduce:**
Use the following code somewhere and open the tab:
```
      <Edit id={id}>
        <TabbedForm warnWhenUnsavedChanges>
          <FormTab label="General" path="general"></FormTab>
        </TabbedForm>
      </Edit>
```

**Related code:**
I don't even have to fork the default codesandbox, just open the Posts Edit view in the simple ra example to see the error:

**Other information:**
The issue is that TabbedForm passes the prop to the styled div, where it is not valid: https://github.com/marmelab/react-admin/blob/d8737f2234eddd6424419ed7e8a816558b8f9e54/packages/ra-ui-materialui/src/form/TabbedFormView.tsx#L62-L64
A solution could be to extend the sanitizeRestProps function to exclude warnWhenUnsavedChanges:
https://github.com/marmelab/react-admin/blob/d8737f2234eddd6424419ed7e8a816558b8f9e54/packages/ra-ui-materialui/src/form/TabbedFormView.tsx#L152

**Environment**

* React-admin version: 4.1.2
* React version: 18.1.0
* Browser: Chrome

magnattic:
closes #7871 

This resolves the issue where `TabbedForm` props like `warnWhenUnsavedChanges` end up on a styled div within `TabbedForm`, which leads to react errors.

I fixed it by sanitizing the props that get passed to the `TabbedFormView`. 
I also removed the sanitizing of the `record` prop within the `TabbedFormView` because from my understanding this should never appear here now anymore (it was already sanitized on the level above).

How to test:
- Run the simple example
- Open the Edit view of a Post
- There should be no console errors anymore

### DIFF
--- a/packages/ra-ui-materialui/src/form/TabbedForm.tsx
+++ b/packages/ra-ui-materialui/src/form/TabbedForm.tsx
@@ -87,7 +87,10 @@ export const TabbedForm = (props: TabbedFormProps) => {
 
     return (
         <Form formRootPathname={formRootPathname} {...props}>
-            <TabbedFormView formRootPathname={formRootPathname} {...props} />
+            <TabbedFormView
+                formRootPathname={formRootPathname}
+                {...sanitizeRestProps(props)}
+            />
         </Form>
     );
 };
@@ -153,3 +156,12 @@ export const findTabsWithErrors = (children, errors) => {
         return acc;
     }, []);
 };
+
+const sanitizeRestProps = ({
+    record,
+    resource,
+    warnWhenUnsavedChanges,
+    defaultChecked,
+    defaultValues,
+    ...rest
+}: TabbedFormProps) => rest;
--- a/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
+++ b/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
@@ -149,4 +149,4 @@ const Root = styled('div', {
     },
 }));
 
-const sanitizeRestProps = ({ record, resource, ...rest }: any) => rest;
+const sanitizeRestProps = ({ resource, ...rest }: TabbedFormViewProps) => rest;


packages/ra-ui-materialui/src/form/TabbedForm.tsx
```diff
@@ -153,3 +156,12 @@ export const findTabsWithErrors = (children, errors) => {
         return acc;
     }, []);
 };
+
+const sanitizeRestProps = ({
+    record,
+    resource,
```
djhi, line 162: Those two should be passed as we want to let users override them through props if needed and some child components call `useResourceContext` and `useRecirdContext`

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/form/TabbedForm.tsx
magnattic: You are right, `resource` is still accessed in `TabbedFormView`. What about `record`, though?
I could not find any usage in TabbedFormView, wouldn't it make sense to not pass it then?

magnattic: I also don't quite understand why we are passing a bunch of unrelated props into `useResourceContext`, instead of just the `resource`? I guess this is for convenience? But at the same time it makes it much harder to follow the actual intent of the code and to see which props are actually needed and used there.

cduran-seeri:
Does this also fixes the issue related to `defaultValues`?

image

fzaninotto:
I noticed the bug while preparing the 4.2 release, and fixed it directly in master (396a1b1e380fe4d0f9c932a2dc9796099db33839). I should have checked the PRs, sorry! I'll close your PR.

I noticed I sanitized the `resource` prop, so I'll open a small PR to re-add it in 4.2.1.

PR #7879, opened 2022-06-22
[Doc] Add Next.js integration tutorial

fzaninotto:
Based on https://marmelab.com/blog/2022/02/02/bootstrap-your-react-admin-project-with-nextjs.html

I could also add the authentication part, but it's for another PR.

No direct link from the sidebar yet, I'll wait until we have more integrations to make it public.


docs/NextJs.md
```diff
@@ -0,0 +1,181 @@
...
+```jsx
+// in pages/api/admin/[[...slug]].ts
+export default async function handler(req, res) {
+  // get the incoming request URL, e.g. 'posts?limit=10&offset=0&order=id.asc'
+  const requestUrl = req.url.substring("/api/admin/".length);
+  // build the CRUD request based on the incoming request
+  const url = `${process.env.SUPABASE_URL}/rest/v1/${requestUrl}`;
+  const options = {
+    method: req.method,
+    headers: {
```
djhi, line 138: Something I realized in my remix example is that you also need to forward all the original request headers and override only those needed. Otherwise, some mutations do not work

djhi (changes requested):

djhi (approved):

docs/NextJs.md
djhi: We'll fix the example when and if needed

PR #7882, opened 2022-06-23
Fix useListParams might navigate to an old pathname

### LINKED ISSUE #7874: adding a filter sometimes sets the wrong url
**What you were expecting:**
When I add a filter on a list, the url params should update with the filters and the path should not change

**What happened instead:**
Sometimes when I add a filter on a list, the path changes in addition to the url params. The path change seems to related to an older path.

**Steps to reproduce:**
We have implemented drawers that open from lists similar to those in https://marmelab.com/react-admin-demo/#/reviews/
We have reproduced the behavior in the demo as well.

1.  go to:  https://marmelab.com/react-admin-demo/#/reviews/
2. click on a review to open the drawer
3. close the drawer
4. add a filter
5. url now contains review id of previously viewed review (which is what opens the drawer) and drawer is open

the behavior is somewhat inconsistent so you may have to repeat steps 2-4 a few times

**Environment**

* React-admin version: latest

djhi:
Fixes #7874

### DIFF
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -128,45 +128,50 @@ export const useListParams = ({
         }
     }, [location.search]); // eslint-disable-line
 
-    const changeParams = useCallback(action => {
-        // do not change params if the component is already unmounted
-        // this is necessary because changeParams can be debounced, and therefore
-        // executed after the component is unmounted
-        if (!isMounted.current) return;
+    const changeParams = useCallback(
+        action => {
+            // do not change params if the component is already unmounted
+            // this is necessary because changeParams can be debounced, and therefore
+            // executed after the component is unmounted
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -128,64 +128,69 @@ export const useListParams = ({
...
+    );
 
     const setSort = useCallback(
         (sort: SortPayload) =>
             changeParams({
                 type: SET_SORT,
                 payload: sort,
             }),
-        requestSignature // eslint-disable-line react-hooks/exhaustive-deps
+        [...requestSignature, changeParams] // eslint-disable-line react-hooks/exhaustive-deps
```
fzaninotto, line 182: I think this means you can remove the eslint disable comment. Same below

fzaninotto (changes requested):

PR #7886, opened 2022-06-23
Update docs for new third party color input / field

alexgschwend:
This [third-party color input] (https://github.com/vascofg/react-admin-color-input) no longer works with react-admin v4 and is no longer maintained. So I ported the library and updated the react-admin documentation.


fzaninotto:
Thanks!

PR #7899, opened 2022-06-26
Fix <ReferenceArrayField /> throw error when its value is not an array

shreyas-jadhav:
if `record.source` is not an array it throws `ids.map is not a function`  error. i think it should be same behavior as "no records matched" and use show empty array in this case.

it is possible same apps have different data type of a field and not always an array. 
but it shouldn't crash entire list component because of that


packages/ra-core/src/controller/field/useReferenceArrayFieldController.ts
```diff
@@ -56,7 +56,8 @@ export const useReferenceArrayFieldController = (
         source,
     } = props;
     const notify = useNotify();
-    const ids = get(record, source) || emptyArray;
+    const value = get(record, source);
+    const ids = Array.isArray(value) ? value : emptyArray;
```
djhi, line 60: Would you mind adding a console.warning about it too?

djhi (changes requested):

djhi (approved):

djhi:
Thanks!

shreyas-jadhav:
Sorry! should pass now

septentrion-730n (approved):

PR #7900, opened 2022-06-26
[Doc] Add documentation for the Layout component

fzaninotto:
image


docs/Layout.md
```diff
@@ -236,4 +236,45 @@ This property accepts the following subclasses:
 
 To override the style of `<Layout>` using the [MUI style overrides](https://mui.com/customization/theme-components/), use the `RaLayout` key.
 
-**Tip**: If you find yourself tweaking the `<Layout sx>` prop, you're probably better off [writing your own layout component](./Theming.md#layout-from-scratch). 
\ No newline at end of file
+**Tip**: If you need to override global styles (like the default font size or family), you should [write a custom theme](./Theming.md#theming) rather than override the `<Layiyt sx>` prop. And you you need to tweak the default layout to add a right column or move the menu to the top, you're probably better off [writing your own layout component](./Theming.md#layout-from-scratch).
```
djhi, line 239:
```suggestion
**Tip**: If you need to override global styles (like the default font size or family), you should [write a custom theme](./Theming.md#theming) rather than override the `<Layout sx>` prop. And if you need to tweak the default layout to add a right column or move the menu to the top, you're probably better off [writing your own layout component](./Theming.md#layout-from-scratch). 
```

docs/Layout.md
```diff
@@ -236,4 +236,45 @@ This property accepts the following subclasses:
...
+export const MyLayout = (props) => (
+    <TenantContext.Provider value={getCookie('tenant')}>
+        <Layout {...props} />
+    </TenantContext.Provider>
+);
+```
+
+## Adding Debug Tools
+
+A custom layout is also the ideal place to add debug tools, e.g. [react-query devtools](https://react-query.tanstack.com/devtools) or [react-hook-form devtools](https://react-hook-form.com/dev-tools):
```
djhi, line 267: The `react-hook-form` needs access to either the form control or the form context. Therefore, it should be placed directly inside a form component.

djhi (changes requested):

djhi (approved):

Issue #7902, opened 2022-06-27
AutocompleteInput create text is undefined when using a function as option text


Guy-Adler:
**What you were expecting:**

When passing the `create` property to `AutocompleteInput`, the default text for the creation option should be `ra.action.create_item`.
![image](https://user-images.githubusercontent.com/44903310/175990045-7dd19aa9-4fcd-43cb-a425-0dedac9b114c.png)

**What happened instead:**

The text was `undefined`.
![image](https://user-images.githubusercontent.com/44903310/175989994-8ea9be37-155e-4df7-89d9-3672bc68ad97.png)

**Steps to reproduce:**

1. Create an AutocompleteInput where the `optionText` prop is a function.
2. Start typing a text which isn't an option.

**Related code:**

Code sandbox: https://codesandbox.io/s/fervent-breeze-muqxj4?file=/src/posts/TagReferenceInput.tsx

The change was made on lines 48:
```ts
optionText={(choice) => choice.name[locale]}
```

**Other information:**

This is probably caused by line 84 of  https://github.com/marmelab/react-admin/blob/58f9841d1fd7dbb3132376b6c50a2b0dd3ca4b8b/packages/ra-ui-materialui/src/input/useSupportCreateSuggestion.tsx#L69-L77
when passing optionText as a function, the string is always saved under `name`, which makes it impossible to use a combination of fields as the option text.

**Environment**

* React-admin version: 4.2.0
* Last version that did not exhibit the issue (if applicable):
* React version: 17
* Browser: Google Chrome 102.0.5005.115 
* Stack trace (in case of a JS error):

Guy-Adler:
As a workaround, one can use
```ts
const buildOption = (choice: any) => {
  if (choice.name !== undefined) {
    return choice.name;
  }
  return whatever you want;
};
```
however, this is very confusing

septentrion-730n:
Bug reproduced, thanks for your report  👍

PR #7908, opened 2022-06-28
Fix `AutocompleteInput`create text is undefined when using a function as option text

septentrion-730n:
Fix https://github.com/marmelab/react-admin/issues/7902

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -169,6 +169,41 @@ describe('<AutocompleteInput />', () => {
         });
     });
 
+    it('should not use optionText defined with a function value on the "create new item" option', async () => {
+        const choices = [
+            { id: 'ang', fullname: 'Angular' },
+            { id: 'rea', fullname: 'React' },
+        ];
+        const optionText = jest.fn(choice => choice.fullname);
+
+        const handleCreate = filter => ({
+            id: 'newid',
+            fullname: filter,
+        });
+
+        render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <SimpleForm mode="onBlur" onSubmit={jest.fn()}>
+                    <AutocompleteInput
+                        source="language"
+                        resource="posts"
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -439,34 +474,6 @@ describe('<AutocompleteInput />', () => {
         expect(setFilter).toHaveBeenCalledWith('');
     });
 
-    it('should allow customized rendering of suggesting item', () => {
```
slax57, line 442: Not sure why you would need to remove this test. To me it should still work fine.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -319,11 +319,16 @@ If you provided a React element for the optionText prop, you must also provide t
...
             }
+
             // Value selected with enter, right from the input
             if (typeof option === 'string') {
                 return option;
             }
 
+            if (option?.id === '@@ra-create') {
+                return option?.name;
+            }
```
slax57, line 330: I mostly agree with the fix you suggest, however:
- `option.id` is not guaranteed to have value `'@@ra-create'`: you can change it using the `createValue` prop
- the label is not guaranteed to be in `option.name`, it depends on the value of the `optionText` prop

I am not sure how we can work around this though... Maybe we should have the `useSupportCreateSuggestion` hook return the actual `id` it used for the "create" item (and the label field can be deduced from the `optionText` I suppose...)

slax57 (changes requested):

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
septentrion-730n: You were very right.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
septentrion-730n: 👍

PR #7909, opened 2022-06-29
Fix AutocompleteInput Should Show Options After Selection

### LINKED ISSUE #7761: AutocompleteInput inside ReferenceInput shows "No options" once a value is assigned
**What you were expecting:**
When clicking on an AutocompleteInput inside a ReferenceInput I always expect to see list of suggestions regardless if that AutocompleteInput already has a value assigned/selected.

**What happened instead:**
When the AutocompleteInput has no value assigned/selected yet, it correctly show the list of suggestions.
As soon as the AutocompleteInput has a value assigned/selected, it shows empty list with "No options".

AutocompleteInput-inside-ReferenceInput

**Steps to reproduce:**
- make sure to login as _admin_
- go to the edit page of a post
- click on the Add button at bottom of the form to add an author
- click on the User autocomplete input
- notice all the available options are shown
- select Breanna Gibson
- click on the User autocomplete input again
- notice it shows "No options" instead of the complete list

**Related code:**
It is reproducable in your `simple` application. See CodeSandbox https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple (v4)

**Other information:**
It does work as expected when the AutocompleteInput is not inside a ReferenceInput.
It seems that the input now immediately uses the already set input to query for autocompletion. Where before you saw the complete suggestion list and it only queried for autocompletion when you started to modify the value that was set.

**Environment**

* React-admin version: 4.1.0
* Last version that did not exhibit the issue (if applicable): 3.19.7 (maybe later version of v3 as well)
* React version: 17
* Browser: Chrome version 101.0.4951.67

djhi:
Fixes #7761

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -439,6 +439,40 @@ describe('<AutocompleteInput />', () => {
         expect(setFilter).toHaveBeenCalledWith('');
     });
 
+    it('should reset filter when users selected a value', async () => {
+        const setFilter = jest.fn();
+        render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <SimpleForm onSubmit={jest.fn()}>
+                    <AutocompleteInput
+                        {...defaultProps}
+                        setFilter={setFilter}
+                        choices={[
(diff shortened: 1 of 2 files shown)


slax57 (approved):
:100:

PR #7917, opened 2022-07-01
Add components to show only time

### LINKED ISSUE #7914: Add time field and input
React admin offers components to display/edit dates and datetimes ([`<DateField/>`](https://marmelab.com/react-admin/doc/4.2/DateField.html ), [`<DateInput/>`](https://marmelab.com/react-admin/doc/4.2/DateInput.html ), [`<DateTimeInput/>`](https://marmelab.com/react-admin/doc/4.2/DateTimeInput.html )), but nothing for time.

[HTML `<input type="time">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#browser_compatibility) seems to be well supported across browsers, so it should be safe to create a new components to edit time like date and datetime. To display time, JS [toLocaleTimeString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#browser_compatibility) seems also safe to use.

For the moment, I wrote my own implementations, but it would be nice to see those components by default in react-admin.

To edit times, I use the  [`<DateTimeInput/>`](https://marmelab.com/react-admin/doc/4.2/DateTimeInput.html ) with the prop `type="time"`. For the format function, I use `convertDateToString` but only returning the time part of the string. 

To display times, I use the following code:

```jsx
const timeString = new Date(<RECORD_VALUE>).toLocaleTimeString();
return (
  <Typography component="span" variant="body2">
    {timeString}
  </Typography>
```

Which is basically the same as [`<DateField/>`](https://marmelab.com/react-admin/doc/4.2/DateField.html) but using `toLocaleTimeString` instead of `toLocaleDateString`.

If you want I can make PRs to add those new components.

arnaudvergnet:
This PR adds a brand new component: `<TimeInput/>`. It behaves much like the `<DateTimeInput/>`, but only shows the time using an input with `type=time`.

Along with adding a new component, this PR also adds a new prop `showDate` to the `<DateField/>` component. This prop, much like `showTime`, allows the user to control whether to display the date in the field. When used together with `showTime`, users can show only the time using `toLocaleTimeString`. If `showTime` and `showDate` are both false, the component throws an error.

Tests and documentation have been updated.

I noticed a lot of the code was copy-pasted between `<DateInput/>`, `<DateTimeInput/>` and now `<TimeInput/>`. Would it be better write the logic in another file and reuse it in each component, or do you want to keep it this way to make understanding each component easier for users?

Closes #7914


docs/TimeInput.md
```diff
@@ -0,0 +1,21 @@
+---
+layout: default
+title: "The TimeInput Component"
+---
+
+# `<TimeInput>`
+
+An input for editing time. `<TimeInput>` renders a standard browser [Time Picker](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time), so the appearance depends on the browser.
+
```
slax57, line 9: Could you add a gif image demonstrating the component right under this paragraph? :slightly_smiling_face:

docs/navigation.html
```diff
@@ -161,6 +161,7 @@
   <li {% if page.path == 'SelectArrayInput.md' %} class="active" {% endif %}><a class="nav-link" href="./SelectArrayInput.html"><code>&lt;SelectArrayInput&gt;</code></a></li>
   <li {% if page.path == 'SelectInput.md' %} class="active" {% endif %}><a class="nav-link" href="./SelectInput.html"><code>&lt;SelectInput&gt;</code></a></li>
   <li {% if page.path == 'TextInput.md' %} class="active" {% endif %}><a class="nav-link" href="./TextInput.html"><code>&lt;TextInput&gt;</code></a></li>
+  <li {% if page.path == 'TimeInput.md' %} class="active" {% endif %}><a class="nav-link" href="./TimeInput.html"><code>&lt;TimeInput></code></a></li>
```
slax57, line 164:
```suggestion
  <li {% if page.path == 'TimeInput.md' %} class="active" {% endif %}><a class="nav-link" href="./TimeInput.html"><code>&lt;TimeInput&gt;</code></a></li>
```

packages/ra-ui-materialui/src/input/TimeInput.tsx
```diff
@@ -0,0 +1,154 @@
...
+ * @returns {String} A standardized time (hh:mm), to be passed to an <input type="time" />
+ */
+const convertDateToString = (value: Date) => {
+    if (!(value instanceof Date) || isNaN(value.getDate())) return '';
+    const hh = leftPad2(value.getHours());
+    const mm = leftPad2(value.getMinutes());
+    return `${hh}:${mm}`;
+};
+
+// yyyy-MM-ddThh:mm
```
slax57, line 128:
```suggestion
// hh:mm
```

slax57 (changes requested):
Great contribution, thank you so much! :slightly_smiling_face:

slax57:
One important thing though, since is is a new feature, could you please rebase your branch and target your PR to the `next` branch instead of `master`?

slax57:
Lastly, regarding your question about whether we need to factorize the code between `<DateInput/>`, `<DateTimeInput/>` and `<TimeInput/>` a little more, IMO the amount of duplicated code is still small an manageable, and keeps the components easy to understand. I'd keep them the way they are :wink:

arnaudvergnet:
@slax57 everything is fixed

arnaudvergnet:
@slax57 everything is fixed. My work was based on next but I forgot to change the target to next as well.

slax57 (changes requested):
Sorry I took so long to review, I was too busy to do it before.

I have one last request, and then we will be good to merge this! :slightly_smiling_face: 

I just think it would be better to showcase the time picker from the browser, which demonstrates a little better imo the benefits of this component.

I made this screencast for this purpose:
time-input


Would you mind using this one in the docs?
Thanks

arnaudvergnet:
I added both gifs in a table to make sure users understand render will change based on the browser.

arnaudvergnet:
@slax57 is everything good for you?

septentrion-730n:
Sorry for the delay. 
I just tested it in the storybook through `make storybook` and ran into a bug, I can not set the value of any `TimeInput` component. Do you experience the same issue?

arnaudvergnet:
Strange I just tested on Firefox on Linux and it works fine. Under the hood it uses an HTML `input` component with `type="time"`, which is supported by all major browsers: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#browser_compatibility

septentrion-730n:
I understand it is at least surprising. Here is a screencast on Chrome on which I use `TimeInput` then `DateTimeInput`.
I will try to get a closer look at it.

https://user-images.githubusercontent.com/102964006/181047539-01280193-6e8f-4e09-b9ec-38ebd8e62754.mov

arnaudvergnet:
Do you only get this behavior on macos? Because I don't have the hardware so I cannot test.

arnaudvergnet:
Can you open the console to check for errors? Maybe there is an issue with the function `toLocaleTimeString` on your system.

septentrion-730n:
Hi Arnaud, 

We didn't forget your job but the holidays postponed our feedback. We are coming back to you as soon as we can.

slax57 (approved):

slax57:
We tried to reproduce @septentrion-730n 's issue on several other machines, but none of them had the issue.
I believe it is safe to (finally!) merge your PR @arnaudvergnet 
Again, sorry for the delay, and many thanks again for your contribution! :slightly_smiling_face:

packages/ra-ui-materialui/src/input/TimeInput.tsx
```diff
@@ -0,0 +1,154 @@
...
+ * Converts a time string without timezone to a date object
+ * with timezone, using the browser timezone.
+ *
+ * @param {string} value Date string, formatted as hh:mm
+ * @return {Date}
+ */
+const parseTime = (value: string) => {
+    const today = new Date();
+    return value ? new Date(today.toLocaleDateString() + ' ' + value) : value;
+};
```
septentrion-730n, line 21:
```suggestion
const parseTime = (value: string) => {
    if (!value) return value;
    const timeTokens = value.split(':').map(v => parseInt(v));
    const today = new Date();
    today.setHours(timeTokens[0] ?? 0);
    today.setMinutes(timeTokens[1] ?? 0);
    return today;
};
```

septentrion-730n (changes requested):

packages/ra-ui-materialui/src/input/TimeInput.tsx
septentrion-730n: Here it is! 🙈

arnaudvergnet: What was the source of the issue? Was the `toLocaleDateString` function not supported on your machine? Was the string given to the `Date` object not understood?

Seems strange it would fail only on some macos devices.

Anyways your fix looks good and achieves the same functionality without complicating the code too much so let's use this!

I can't make the commit myself right now so feel free to push to the PR.

Thanks for fixing this!

septentrion-730n: The error was nesting in the Date parsing which was returning an invalid date.
`new Date(today.toLocaleDateString() + ' ' + value)` => `new Date('18/08/2022 10:22')`

slax57: Unfortunately it appears we do not have push permission on your branch @arnaudvergnet :slightly_frowning_face: 

Could you commit these changes when you have time?

Again, sorry for taking so long, and thank you so much for your patience. :slightly_smiling_face:

arnaudvergnet: > it was nested in the Date parsing which was returning an invalid date.

Oh maybe the parsing failed because you had your computer locale in  French so the date parsing behaved differently from mine (my computer was in English).

arnaudvergnet: > Could you commit these changes when you have time?

Can't find the option to allow edits by maintainers and I cannot make the commit right now so this will have to wait a few days if that's okay with you.

slax57: > > Could you commit these changes when you have time?
> 
> Can't find the option to allow edits by maintainers and I cannot make the commit right now so this will have to wait a few days if that's okay with you.

That's perfectly fine. :slightly_smiling_face: 
Besides, how could we urge you when we ourselves took ages to review your PR in the first place :joy:

arnaudvergnet:
@slax57 @septentrion-730n applied the changes to the first commit to keep the history clean.

slax57 (approved):
Thank you so much! :slightly_smiling_face:

PR #7925, opened 2022-07-04
Fix simple dataProvider create does not return server data

### LINKED ISSUE #7912: useCreateResult data contains the update parameters and not the response data
**What you were expecting:**
useCreateResult.data will contain all fields returned from the API, just like useUpdateResult.data does.

**What happened instead:**
useCreateResult.data contains the data sent to the API instead of the data returned from it.

**Steps to reproduce:**
Any API that returns a calculated value in one of the fields (i.o.w. the field is not part of the data sent to the API, but rather generated by the API for consumption by the RA app) and a custom data provider method in RA that consumes the value, similar to the logo upload example. I ran into this when doing logo uploads using S3 signed URL.

**Related code:**

```
export const CustomDataProvider = {
	...baseDataProvider,
		updateWithUpload: async (resource: string, record: any) => {
		        const { logoUrl, logoName, ...rest } = record;
			// call our API to get signedUrl property
			const updateResult = await baseDataProvider.update(resource, {
				id: record.id,
				data: {
					...rest,
					logoName: logoUrl.rawFile.path,
				},
				previousData: { id: record.id },
			});
			if (updateResult.data.signedUrl) {
			  // this works
			}
		},
		createWithUpload: async (resource: string, record: any) => {
		        const { logoUrl, logoName, ...rest } = record;
			// call our API first to get signedUrl property
			const createResult = await baseDataProvider.create(resource, {
				data: {
					...rest,
					logoName: logoUrl.rawFile.path,
				},
			});
			// upload logo directly to s3
			if (createResult.data.signedUrl) {
			   // we never reach this code
			}
		}
```

**Other information:**
N/A

**Environment**

* React-admin version: 3.19.5
* React version: 16.14.0
* Browser: Chrome latest on Mac

dylanlt:
Update useCreateResult in the Simple REST Provider to return the API response data.

Fixes #7912

### DIFF
--- a/packages/ra-data-simple-rest/src/index.ts
+++ b/packages/ra-data-simple-rest/src/index.ts
@@ -159,9 +159,7 @@ export default (
         httpClient(`${apiUrl}/${resource}`, {
             method: 'POST',
             body: JSON.stringify(params.data),
-        }).then(({ json }) => ({
-            data: { ...params.data, id: json.id },
-        })),
+        }).then(({ json }) => ({ data: json })),
 
     delete: (resource, params) =>
         httpClient(`${apiUrl}/${resource}/${params.id}`, {


djhi (approved):
Thanks!

PR #7926, opened 2022-07-05
assign variable before export default

sunil-sharma-999:
fix eslint warnings
- Assign object to a variable before exporting as module default
- Assign arrow function to a variable before exporting as module default

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -699,12 +699,14 @@ By default, react-admin displays the list page of the first `Resource` element a
 import * as React from "react";
 import { Card, CardContent, CardHeader } from '@mui/material';
 
-export default () => (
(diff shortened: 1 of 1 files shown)


fzaninotto:
We tend to use named exports rather than default exports in react-admin. I think the tutorial should be updated accordingly.

sunil-sharma-999:
@fzaninotto 

same, was thinking of changing them into named exports but then i thought maybe other people like default exports. should i update it?

fzaninotto:
yes, please, but make sure that the other code snippets, where we import the components, are updated, too.

sunil-sharma-999:
@fzaninotto 
   
Should i change the structure of files in tutorial or not?

PR #7938, opened 2022-07-07
Add server side validation support

djhi:
## Problem

- Server side validation is cumbersome to set up
- When using middlewares and as the SaveButton relied on the SaveContext.saving prop, the SaveButton was only disabled for while the main mutation was loading. However, additional work may still be running.

## Solution

For `pessismistic` mode only, we now transparently support server side validation. When a server validation occurs, the DataProvider should throw an error that contains an `errors` object matching the form shape.

Besides, we now await the mutation in `useCreateController` and `useEditController`. In `pessimistic` mode, that means the `SaveButton` will stay disabled until the mutation is resolved including its middlewares if any.

That does not change anything for the other mutation modes.


docs/Validation.md
```diff
@@ -320,42 +320,46 @@ const CustomerCreate = () => (
...
+    }
+}
+```
 
-**Tip**: The shape of the returned validation errors must correspond to the form: a key needs to match a `source` prop.
+**Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop.
 
 **Tip**: The returned validation errors might have any validation format we support (simple strings or object with message and args) for each key.
+
+**Tip**: If your data provider leverages our [`httpClient`](https://marmelab.com/react-admin/DataProviderWriting.html#example-rest-implementation), this will be handled automatically when your server returns an invalid response with a json body contains the `errors` key.
```
septentrion-730n, line 339:
```suggestion
**Tip**: If your data provider leverages our [`httpClient`](https://marmelab.com/react-admin/DataProviderWriting.html#example-rest-implementation), this will be handled automatically when your server returns an invalid response with a json body containing the `errors` key.
```

examples/simple/src/dataProvider.tsx
```diff
@@ -80,7 +80,13 @@ const sometimesFailsDataProvider = new Proxy(uploadCapableDataProvider, {
             params.data &&
             params.data.title === 'f00bar'
         ) {
-            return Promise.reject(new Error('this title cannot be used'));
+            return Promise.reject(
+                new HttpError('The form is invalid', 404, {
```
septentrion-730n, line 84:
```suggestion
                new HttpError('The form is invalid', 400, {
```

septentrion-730n: 400 Bad Request ?

septentrion-730n (approved):

docs/Validation.md
```diff
@@ -320,42 +320,46 @@ const CustomerCreate = () => (
...
-                <TextInput label="First Name" source="firstName" />
-                <TextInput label="Age" source="age" />
-            </SimpleForm>
-        </Create>
-    );
-};
 ```
-{% endraw %}
+{
+    body: {
```
fzaninotto, line 327: why the "body" key? Why can't the dataProvider only reject with `({ errors: { source: 'error message' } }`?

Also, your snippet is both a type description and an example, and so it's not enough informative. I'd use a more complete example:

```
{ 
  errors: {
    title: 'An article with this title already exists. The title must be unique.',
    date: 'The date is required',
    tags: { message: "The tag 'agrriculture' doesn't exist" },
  }
}
```

docs/Validation.md
```diff
@@ -320,42 +320,46 @@ const CustomerCreate = () => (
...
+    }
+}
+```
 
-**Tip**: The shape of the returned validation errors must correspond to the form: a key needs to match a `source` prop.
+**Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop.
 
 **Tip**: The returned validation errors might have any validation format we support (simple strings or object with message and args) for each key.
+
+**Tip**: If your data provider leverages our [`httpClient`](https://marmelab.com/react-admin/DataProviderWriting.html#example-rest-implementation), this will be handled automatically when your server returns an invalid response with a json body containing the `errors` key.
```
fzaninotto, line 339: this isn't super clear

packages/ra-core/src/controller/create/useCreateController.spec.tsx
```diff
@@ -502,4 +503,34 @@ describe('useCreateController', () => {
             expect.any(Function)
         );
     });
+
+    it('The save function should return errors from the create call', async () => {
```
fzaninotto, line 507: this 'it' description doesn't make sense. Perhaps you should embed it in a describe?

```js

describe('save', () => {
   it('should return errors from the create call', () => {
```

docs/Validation.md
```diff
@@ -320,42 +320,46 @@ const CustomerCreate = () => (
...
+
+Here's an example of a dataProvider not using our `httpClient`:
```
fzaninotto, line 341: again, this is confusing. You don't give an eample of the general case but you give one of a particular case.

packages/ra-core/src/controller/create/useCreateController.spec.tsx
```diff
@@ -502,4 +503,34 @@ describe('useCreateController', () => {
...
+                </CreateController>
+            </CoreAdminContext>
+        );
+        await new Promise(resolve => setTimeout(resolve, 10));
+        let errors;
+        await act(async () => {
+            errors = await saveCallback({ foo: 'bar' });
+        });
+        expect(errors).toEqual({ foo: 'invalid' });
+        await new Promise(resolve => setTimeout(resolve, 10));
```
fzaninotto, line 531: I don't understand why you need to wait here. If you have errors, two lines above, it's sure that the dataProvider was already called, right?

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -91,55 +95,67 @@ export const useCreateController = <
...
+                                      );
+                                  },
+                        }
+                    );
+                } catch (error) {
+                    if ((error as HttpError).body?.errors != null) {
+                        return (error as HttpError).body.errors;
                     }
-                );
+                }
```
fzaninotto, line 158: and otherwise, rethrow!

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -881,4 +881,40 @@ describe('useEditController', () => {
...
+                <EditController {...defaultProps} mutationMode="pessimistic">
+                    {({ save, record }) => {
+                        saveCallback = save;
+                        return <>{JSON.stringify(record)}</>;
+                    }}
+                </EditController>
+            </CoreAdminContext>
+        );
+        await new Promise(resolve => setTimeout(resolve, 10));
+        screen.getByText('{"id":12}');
```
fzaninotto, line 906: remove the await above, and instead replace this line by

```jsx
await screen.findByText('{"id":12}');
```

That way you'll never wait longer than necessary

packages/ra-core/src/controller/edit/useEditController.spec.tsx
```diff
@@ -881,4 +881,40 @@ describe('useEditController', () => {
...
+        let errors;
+        await act(async () => {
+            errors = await saveCallback({ foo: 'bar' });
+        });
+        expect(errors).toEqual({ foo: 'invalid' });
+        await new Promise(resolve => setTimeout(resolve, 10));
```
fzaninotto, line 912: same, why do you need to wait here?

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -134,57 +139,65 @@ export const useEditController = <
...
+                                              },
+                                          }
+                                      );
+                                  },
+                        }
+                    );
+                } catch (error) {
+                    if ((error as HttpError).body?.errors != null) {
+                        return (error as HttpError).body.errors;
                     }
```
fzaninotto, line 199: otherwise, rethrow

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -91,55 +95,67 @@ export const useCreateController = <
...
+                                                          : undefined,
+                                              },
+                                          }
+                                      );
+                                  },
+                        }
+                    );
+                } catch (error) {
+                    if ((error as HttpError).body?.errors != null) {
+                        return (error as HttpError).body.errors;
```
fzaninotto, line 156: This is weird: the save promise *resolves* when there is a validation error in the save process? I'd expect it to reject.

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -127,7 +127,10 @@ export const useCreate = <
             unknown
         > & { returnPromise?: boolean } = {}
     ) => {
-        const { returnPromise, ...reactCreateOptions } = createOptions;
+        const {
+            returnPromise = options.returnPromise,
```
fzaninotto, line 131: please test and document that option

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -431,7 +431,7 @@ export type UseUpdateOptions<
     RecordType,
     MutationError,
     Partial<UseUpdateMutateParams<RecordType>>
-> & { mutationMode?: MutationMode };
+> & { mutationMode?: MutationMode; returnPromise?: boolean };
```
fzaninotto, line 434: please test and document that option

fzaninotto (changes requested):

docs/Validation.md
djhi: Because that's the current shape of our `HttpError`. Our data providers put the response json inside the `body` property of the error they throw

packages/ra-core/src/controller/create/useCreateController.ts
djhi: Nope, this is the only type of errors we want to handle here. Otherwise, errors will be handled through the `onError` side effect function

packages/ra-core/src/controller/edit/useEditController.ts
djhi: Nope, this is the only type of errors we want to handle here. Otherwise, errors will be handled through the onError side effect function

packages/ra-core/src/controller/create/useCreateController.ts
djhi: This is how the `useAugmentedForm` has been implemented. If the submit function returns something, we consider this is an error object. This was done like this probably because it worked that way in `final-form`.

packages/ra-core/src/dataProvider/useCreate.ts
djhi: The `returnPromise` option already existed. However we only accepted at call time. I just added it at declaration time

packages/ra-core/src/dataProvider/useUpdate.ts
slax57: It is used in the `useEditController` in pessimistic mode, so it is tested. I'm not sure we really need to document this option as it's pretty advanced, and the option was not documented in the `updateOptions` type in the first place.

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -54,7 +54,6 @@ export const SaveButton = <RecordType extends RaRecord = any>(
         label = 'ra.action.save',
         onClick,
         mutationOptions,
-        saving,
```
slax57, line 57: `saving` should be removed from the props type as well (I believe we can consider it's not a BC because it was not documented in the first place)

slax57 (changes requested):

docs/Validation.md
```diff
@@ -362,52 +362,88 @@ const CustomerCreate = () => (
...
+            date: 'The date is required',
+            tags: { message: "The tag 'agrriculture' doesn't exist" },
+        }
+    }
+}
+```
+
+**Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop.
+
+**Tip**: The returned validation errors might have any validation format we support (simple strings or object with message and args) for each key.
```
slax57, line 381: This is, according to my tests, not entirely true. Translation strings are supported, either returned directly or as the `message` prop, but translation **args** do not seem to work.

Anyway, fixing this would fall beyond the scope of this PR IMO, so I'll just remove this part from the docs.

slax57 (changes requested):

packages/ra-core/src/controller/edit/useEditController.ts
slax57: I agree with you @fzaninotto . This might even be the origin of the failing E2E test. I'll have a look.

slax57: Okay so after verification, @djhi  is correct, both the `onError` block and the `catch` block will be executed. So we don't need to rethrow the error.

docs/Validation.md
slax57: I tried to make this section clearer. Please let me know if that's enough.

docs/Validation.md
slax57: same

slax57 (approved):

djhi (commented):
:100:

packages/ra-core/src/controller/create/useCreateController.ts
fzaninotto: can't we improve that in a backward compatible way?

packages/ra-core/src/controller/edit/useEditController.ts
fzaninotto: Than please add a comment explaining why we don't rethrow and that this is not a mistake

slax57 (approved):

PR #7940, opened 2022-07-07
[Fix] Filters list overflow on mobile

septentrion-730n:
The result is far from perfect but corrects major responsive display errors.

https://user-images.githubusercontent.com/102964006/177950099-d3d986de-209c-4bba-8c3d-185646a6bbea.mov

## With debug borders (don't appear for real)

https://user-images.githubusercontent.com/102964006/177950131-49fd0804-1072-41d0-8065-b268a9a4e001.mov


djhi:
Can you show how it looks with the post filter removed so that we can see the filter button?

djhi (approved):

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -206,14 +206,20 @@ const StyledForm = styled('form', {
...
-    minHeight: theme.spacing(8),
     display: 'flex',
+    flex: '0 1 auto',
     alignItems: 'flex-end',
     flexWrap: 'wrap',
+    minHeight: theme.spacing(9),
     pointerEvents: 'none',
-
+    [theme.breakpoints.up('md')]: {
+        padding: `${theme.spacing(0.5)}`,
```
fzaninotto, line 216: why do you need to use a template for that?

fzaninotto:
I don't like this solution, as it pushes the content down on desktop:

## Before
localhost_8080_ (1)

## After
localhost_8080_ (2)

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
septentrion-730n, line 216:
```suggestion
        padding: theme.spacing(0.5),
```

fzaninotto:
I noticed that the Edit and Show views no longer have the same top margin as the List view. This regression is not caused by your PR, but it's recent. As you're working on the design, could you take a look at it?

packages/ra-ui-materialui/src/layout/TopToolbar.tsx
```diff
@@ -20,20 +20,20 @@ const StyledToolbar = styled(Toolbar, {
...
     gap: theme.spacing(1),
-    [theme.breakpoints.up('sm')]: {
-        minHeight: theme.spacing(6),
-    },
-    [theme.breakpoints.up('xs')]: {
-        paddingLeft: 0,
-        paddingRight: 0,
+    whiteSpace: 'nowrap',
+    flex: '0 1 auto',
+    minHeight: 'unset !important',
```
fzaninotto, line 26: Can't we avoid this?

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -200,22 +201,30 @@ const PREFIX = 'RaFilterForm';
...
-    marginTop: theme.spacing(-0.5),
-    marginBottom: theme.spacing(0.5),
-    minHeight: theme.spacing(8),
-    display: 'flex',
-    alignItems: 'flex-end',
-    flexWrap: 'wrap',
-    pointerEvents: 'none',
-
-    [`& .${FilterFormClasses.clearFix}`]: { clear: 'right' },
+    '&': {
```
fzaninotto, line 211: Why don't you define these styles in the root?

septentrion-730n:
As discussed, top toolbar height remains consistent between Edit, Create and List templates

https://user-images.githubusercontent.com/102964006/180400251-4eeb5753-be7d-4bf3-aa55-49c55951aa91.mov

packages/ra-ui-materialui/src/layout/TopToolbar.tsx
septentrion-730n: I removed that, as agreed,  toolbar keeps (again) consistent height on List, Edit and View templates

fzaninotto (approved):

PR #7947, opened 2022-07-08
Fix AppBar jiggles when scrolling down rapidly

### LINKED ISSUE #7944: AppBar has a weird transition while scrolling down
**What you were expecting:**
AppBar hides smoothly while the user scrolling down

**What happened instead:**
AppBar bounces up and down 

**Steps to reproduce:**
1. Go to any list of the admin and increase the pagination to get a scrollable page
2. Scroll down faster

**Related code:**
It can be reproduced in the default demo: [https://codesandbox.io/embed/github/marmelab/react-admin/tree/master/examples/simple?hidenavigation=1&editorsize=0](https://codesandbox.io/embed/github/marmelab/react-admin/tree/master/examples/simple?hidenavigation=1&editorsize=0)

**Other information:**
Screenshot GIF

**Environment**

* React-admin version: 4.2.1
* Last version that did not exhibit the issue (if applicable): 4.2.0
* React version: 18.2.0
* Browser: All

fzaninotto:
Closes #7944

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Sidebar.tsx
+++ b/packages/ra-ui-materialui/src/layout/Sidebar.tsx
@@ -2,7 +2,13 @@ import * as React from 'react';
 import { styled } from '@mui/material/styles';
 import { ReactElement } from 'react';
 import PropTypes from 'prop-types';
-import { Drawer, DrawerProps, useMediaQuery, Theme } from '@mui/material';
+import {
+    Drawer,
+    DrawerProps,
+    useMediaQuery,
+    Theme,
+    useScrollTrigger,
+} from '@mui/material';
 import lodashGet from 'lodash/get';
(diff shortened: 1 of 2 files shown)


septentrion-730n (approved):

packages/ra-ui-materialui/src/layout/Sidebar.tsx
```diff
@@ -29,22 +35,13 @@ export const Sidebar = (props: SidebarProps) => {
         >
             {children}
         </StyledDrawer>
-    ) : isSmall ? (
```
fzaninotto, line 32: This ternary returns the same result in both cases since 9c965a8d725977d2b48bf755ffe57274a4e6731b (1 year ago), so we can remove it.

PR #7957, opened 2022-07-08
Fix <Pagination> cannot be used outside a ListContext for 3.x

### LINKED ISSUE #7955: Error with Pagination props
**What you were expecting:**

I need to create a custom list component using hooks to deal with requests. So, I don't want to use the `<List>` component.
I saw in the documentation that it is possible to use `<Datagrid> With A Custom Query`. [The doc can be read here](https://marmelab.com/react-admin/doc/3.19/List.html#rendering-datagrid-with-a-custom-query)
As you can see in the doc, we can use the `<Pagination>` component after `<Datagrid>` passing these props:
image

So, I have been expecting to use that without error.

**What happened instead:**

However, when I use these props in `<Pagination>` with Typescript, I have an error, because `<Pagination>` just receives these props:
image

The error is:
image

**Steps to reproduce:**

**You can see the error here:** https://codesandbox.io/s/heuristic-ganguly-jhsu68?file=/src/posts/PostList.tsx

image

**Related code:**

* React-admin version: 3.19.10
* React version: 17.0.2
* Browser: All

schirrel:
Fixes #7955  for 3.x version based on #7956

### DIFF
--- a/packages/ra-core/src/controller/index.ts
+++ b/packages/ra-core/src/controller/index.ts
@@ -1,7 +1,9 @@
 import ListController from './ListController';
 import ListContext from './ListContext';
 import ListFilterContext from './ListFilterContext';
-import ListPaginationContext from './ListPaginationContext';
+import ListPaginationContext, {
+    ListPaginationContextValue,
+} from './ListPaginationContext';
 import ListSortContext from './ListSortContext';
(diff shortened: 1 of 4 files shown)


fzaninotto:
Thanks!

Issue #7805, opened 2022-06-09
useTheme returns undefined as value


bladx:
**What you were expecting:**
Getting the theme value with the react-admin hook: useTheme().

**What happened instead:**
useTheme() hook returns undefined as value.

**Steps to reproduce:**
Just go to the configuration page of the demo right after a login (cache initially cleared).
https://marmelab.com/react-admin-demo/#/configuration
I reproduce every time (not that evident as the primary and secondary colors are close).

**Related code:**
```
const [theme] = useTheme();
console.log(theme);
```
**Environment**

* React-admin version: ^4.1.0
* Last version that did not exhibit the issue (if applicable): I used ^3.14.1 before so I would say the bug was introduce with v4 (when the hook has been created actually)
* React version: ^18.1.0
* Browser: Firefox 100.0.2 (64-bit), macOS Monterey Version 12.2.1

fzaninotto:
Confirmed, thanks for the report.

bladx:
Also, do you know how I can make the stored theme in local storage not to be cleaned at log out?
I found out how not to store the theme in local storage using: https://marmelab.com/react-admin/Store.html#transient-store
But I would like that the theme selected with ToggleThemeButton is not flushed at log out anymore.

PR #7960, opened 2022-07-09
Fix `useTheme` returns undefined as value

septentrion-730n:
Fix https://github.com/marmelab/react-admin/issues/7805

React-Admin `useTheme` hook use the store, and when no data is available there, returns the default value specified as hook parameter.  As this default theme parameter is not available in theme data consumption, the best solution was to use MUI `useTheme` hook for reading purpose. This PR aims at automating this process and avoid misuses.


djhi (approved):
:100:

djhi (commented):
It would be nice to test that we return the theme from the settings when available

PR #7967, opened 2022-07-12
Fix useChoicesContext props should take precedence over context

### LINKED ISSUE #7965: Inputs inside ReferenceInput's child's create component receive improper props
**What you were expecting:**
Inputs inside a "create" form component passed to a SelectInput or an AutocompleteInput inside a ReferenceInput to behave the same as those form elements in and standalone use-case

**What happened instead:**
Inputs receive odd props; choices, labels, etc. seem to be inherited from the ReferenceInput parent component instead of the props passed to the form inputs themselves.

**Steps to reproduce:**
Pass a number of form inputs to a "create" component and observe their labels/choices.

**Related code:**
https://codesandbox.io/s/compassionate-herschel-5re5vg?file=/src/posts/PostCreate.tsx 
Navigate to the Post create form and attempt to create a user at the "Author" input. Observe that the "type" SelectInput has inherited incorrect choices and label.

**Environment**

* React-admin version: 4.2.2
* React version: 17
* Browser: Chrome

djhi:
Fixes #7965

### DIFF
--- a/examples/simple/src/posts/PostCreate.tsx
+++ b/examples/simple/src/posts/PostCreate.tsx
@@ -24,8 +24,11 @@ import {
     useNotify,
     usePermissions,
     useRedirect,
+    useCreate,
+    useCreateSuggestionContext,
 } from 'react-admin';
 import { useFormContext, useWatch } from 'react-hook-form';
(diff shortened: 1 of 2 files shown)


arimet (approved):
It looks good to me

PR #7968, opened 2022-07-12
fix the type of `id` prop

tmarcinkowski-logitech:


fzaninotto:
No, this is incorrect. The form id MUST be a string. I don't know why you think otherwise, and as you didn't provide a description for this PR, I have to close it.

tmarcinkowski-logitech:
Yeah, sorry for not providing the context for this PR.

It came up as an issue reported by TypeScript when I wanted to make use of FormProps in my own form component that was intended to be used as a child of `Edit`.  
It's possible that I didn't get something right during migration from RA 3 to 4 (this migration is about to be one of the worst experiences in my whole carrier) but passing props from the edit view to the edit form was my previous approach (as of RA3) and something I started with.

Eventually, I ditched that idea and in favor of using the record context, but the original code was along these lines:

```
export const CustomEdit: React.FC<EditProps> = (props) => {
        // some code

	return (
		<Edit>
			<CustomForm {...props} />
		</Edit>
	);
};
```

`EditProps` contain `id?: Identifier;` (and that is something I keep seeing along RA codebase) but `FormProps` accepts only a string. Are you really sure that this is done by design?

djhi:
If I'm not mistaken, the `FormProsp` `id` is not the record id but the id for the form element

tmarcinkowski-logitech:
Well, that would've explain it.

PR #7993, opened 2022-07-21
Add `requireAuth` to <Authenticated /> component

smeng9:
The `useAuthenticated()` hook and old `<Authenticated />` component is always optimistic, other auth hooks cannot be used directly inside `<Route element>` and causes boilerplate, and `<Admin requireAuth>` completely disallows anonymous access to all the pages.

This feature gives user finer granularity of control of which custom routes can be anonymously accessed, which can be optimistically rendered, and which should not leak the UI and does not have flash of UI.

### DIFF
--- a/docs/Authenticated.md
+++ b/docs/Authenticated.md
@@ -5,7 +5,7 @@ title: "The Authenticated Component"
 
 # `<Authenticated>`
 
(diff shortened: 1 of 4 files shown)


smeng9 (commented):
fix linter

djhi (changes requested):
Thanks! Would you mind renaming the `pessimistic` prop to `requireAuth` just like in the `Admin` component?

djhi (approved):
Thanks!

PR #7994, opened 2022-07-21
[doc]: fix `<WithPermissions />` and `<IfCanAccess />` doc

smeng9:
The previous docs has broken examples and broken links, which causes some confusion.


djhi (approved):
Indeed, we forgot to update this. Thanks!

PR #7995, opened 2022-07-21
Fix error in useReferenceParams

septentrion-730n:
Corrects `Can't perform a React state update on an unmounted component error` while using ReferenceArrayInput in list filter.

Error appears in `CreateDialog` or `EditDialog` when dialog box is closed faster than the filters debounce delay of any autocomplete input.

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceParams.ts
+++ b/packages/ra-core/src/controller/input/useReferenceParams.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState, useRef } from 'react';
+import { useCallback, useMemo, useState, useRef, useEffect } from 'react';
 import lodashDebounce from 'lodash/debounce';
 
 import { SortPayload, FilterPayload } from '../../types';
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/controller/input/useReferenceParams.ts
```diff
@@ -127,20 +127,28 @@ export const useReferenceParams = ({
...
+                    filter: removeEmpty(filter),
+                    displayedFilters,
+                },
+            });
+        }, debounce)
+    );
+    useEffect(() => {
+        return () => {
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+            debouncedSetFilters.current.cancel();
```
fzaninotto, line 144: As this effect runs every time the hoo runs, I'm afraid this will cancel all filters

fzaninotto:
Could you describe in which condition the error occurs - or better, add a unit test ?

packages/ra-core/src/controller/input/useReferenceParams.ts
septentrion-730n: Shame on me.

fzaninotto: Now that makes sense: you clear the debounce on unmount. Got it!

fzaninotto (approved):

PR #7998, opened 2022-07-21
Fix `<SelectInput>` loading UI to avoid visual jump

### LINKED ISSUE #7975: <SelectInput /> inside <ReferenceInput /> not showing a <LinearProgress /> when loading
**What you were expecting:**
See a `<LinearProgress />` when `<ReferenceInput />` if fetching options.

**What happened instead:**
It only show the label without the `<input />` at first glance and then the `<input />` appears, making the form changing size if the fields are vertically aligned.

**Steps to reproduce:**
* visit that demo https://marmelab.com/react-admin-crm/#/companies/create
* look at the **Account manager** field when refreshing the page
* it happens on my forms too

**Related code:**
probably related around this as if `isLoading` is true before options are (I suppose) https://github.com/marmelab/react-admin/blob/5feeda4bb9dec99f31a174448ff4df5f8aaa8318/packages/ra-ui-materialui/src/input/SelectInput.tsx#L236-L248

**Environment**

* React-admin version: 4.2.2

fzaninotto:
Closes #7975

## Before

https://user-images.githubusercontent.com/99944/180303089-c9d9620d-5cd6-471c-a777-be1ad5975622.mov

## After

https://user-images.githubusercontent.com/99944/180305200-47dcbdf0-bd7e-4689-a757-72e39298fc8c.mov

### DIFF
--- a/packages/ra-ui-materialui/src/input/LoadingInput.tsx
+++ b/packages/ra-ui-materialui/src/input/LoadingInput.tsx
@@ -1,0 +1,86 @@
+import * as React from 'react';
+import { CircularProgress } from '@mui/material';
+import { styled, SxProps } from '@mui/material/styles';
+import { useInput, FieldTitle, useTimeout } from 'ra-core';
+
+import { ResettableTextField } from './ResettableTextField';
+
+/**
+ * An input placeholder with a loading indicator
+ *
+ * Avoids visual jumps when replaced by a form input
+ */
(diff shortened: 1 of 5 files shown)


fzaninotto:
I still have to suppress the jump due to the helperText

septentrion-730n (approved):

djhi (approved):

PR #7999, opened 2022-07-22
Add support for custom login redirection

### LINKED ISSUE #6583: Unwanted redirect after useLogin()
**What you were expecting:**

Calls to login() don't do an implicit redirect when custom authProvider involves a redirect flow.

**What happened instead:**

When calling the login function returned by `useLogin()`, react-admin redirects to the homepage. This causes a noticable flicker in the browser page because shortly after react-admin's redirect our authProvider (using auth0) does it's own redirect.

**Steps to reproduce:**

```
const auth0Client = new Auth0Client(...)
const customAuthProvider = {
  ...
  login: async () => {auth0Client.loginWithRedirect()},
  ...
}

...

const login = useLogin()
...
<button onClick={login}>Login</button>
```

**Suggestion**

Similarly to `authProvider.logout()` where we can return a String for the route to redirect to or `false` to skip redirecting, `authProvider.login()` could have the same return semantics so that `return false` from `authProvider.login()` doesn't redirect.

**Environment**

* React-admin version: 13.7.1
* React version: 17.0.2
* Browser: Firefox

fzaninotto:
Closes #6583

### DIFF
--- a/docs/AuthProviderWriting.md
+++ b/docs/AuthProviderWriting.md
@@ -123,6 +123,32 @@ Once the promise resolves, the login form redirects to the previous page, or to 
 
 **Tip**: It's a good idea to store credentials in `localStorage`, as in this example, to avoid reconnection when opening a new browser tab. But this makes your application [open to XSS attacks](https://www.redotheweb.com/2015/11/09/api-security.html), so you'd better double down on security, and add an `httpOnly` cookie on the server side, too.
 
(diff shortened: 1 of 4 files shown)


e-shaan:
Hello, I am still facing this issue of a Homepage flicker on Login, how do I fix it?

fzaninotto:
This PR should fix the problem. It should be bundled in version 4.3, to be published end of August.

e-shaan:
> This PR should fix the problem. It should be bundled in version 4.3, to be published end of August.

Right, thank you for the response.

PR #8004, opened 2022-07-23
[doc] Fix: Incorrect example code block of ReferenceInput

ApolloRisky:
+  Providing prop `label` to ReferenceInput doesn't take any effect. It should be passed to its child *Input component.


fzaninotto (approved):

fzaninotto:
You're right, thanks!

PR #8005, opened 2022-07-23
fix(menu): add unique key

DerYeger:
Using the `dashboard` property of the `Admin` component results in an error with the `Menu` component.
The `DashboardMenuItem` does not have a `key` set, hence React complains about missing unique-keys.

This PR defines a key (`default-dashboard-menu-item`) for the dashboard's menu item.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8009, opened 2022-07-25
[Doc] Fix useQuery example snippet

fzaninotto:


docs/Actions.md
```diff
@@ -152,8 +152,8 @@ const UserProfile = ({ userId }) => {
 
     return (
         <ul>
-            <li>Name: {data.name}</li>
-            <li>Email: {data.email}</li>
+            <li>Name: {data.data.name}</li>
```
septentrion-730n, line 155: Shouldn't we try to set this more elegantly with: 

```
const { data: result, isLoading, error } = useQuery(
...
<li>Name: {result.data.name}</li>
<li>Name: {result.data. email}</li>
``` 
```?

septentrion-730n (approved):

docs/Actions.md
fzaninotto: Not sure it's better...

PR #8011, opened 2022-07-26
Add default Record representation

### LINKED ISSUE #4899: Allow to customize default display field per resource (use name instead of id)

By default, react-admin uses the `id` field to represent a row (for example as the first column in the list view, as the displayed attribute in ReferenceFields, in the navbar of the edit view, etc.).

This behavior range from not very helpful when using integers pks (`Client #3` => doesn't give much info) to not usable when using uuids (`Client #00c8b66f-4f01-4db7-b0c6-ad588c7842ba` => destroys the layout and looks like a bug).

I'd love to be able to specify the field to be used to represent the row like this :
```html
<Resource name="clients" display="fullname" list={ListGuesser} (...) />
```
(alternative naming ideas instead of `display` : `repr`, `mainAttribute`, `titleAttribute`, ...)

Not using Guessers, and creating title componenents for all my resources. This works well, but is verbose, and you loose a bit the out-of-the-box-awesomeness of react-admin.

**Additional context**

Thanks for react-admin, it's really awesome :-)

fzaninotto:
## Problem

React-admin forces developers to define a child `<Field>` for `<ReferenceField>` to represent the related record. This is cumbersome.

## Solution

Add the ability to define a `recordRepresentation` at the `<Resource>` level, and let `<ReferenceField>`. Use that knowledge to render without a child. 

```diff
-<ReferenceField source="user_id" reference="users">
-    <TextField source="name" />
-</ReferenceField>
+<ReferenceField source="user_id" reference="users" />
```

- [x] Update `<Resource>` and `ResourceDefinitionContext` to carry the `recordRepresentation`
- [x] Add a `useGetRecordRepresentation` hook
- [x] Use `recordRepresentation` in `<ReferenceField>` and let it work without children
- [x] Use `recordRepresentation` in `<ReferenceOneField>` and let it work without children
- [x] Use `recordRepresentation` in `<Edit>` and `<Show>` page titles
- [x] Use `recordRepresentation` in `<SelectInput>` and let it work without `optionText` when used in `<ReferenceInput>`
- [x] Use `recordRepresentation` in `<AutocompleteInput>` and let it work without `optionText` when used in `<ReferenceInput>`
- [x] Use `recordRepresentation` in `<ReferenceInput>` and let it work without children
- [x] Update `<ListGuesser>`, `<EditGuesser>`  and `<ShowGuesser>` to output child-less References
- [x] Update tutorial
- [x] Add doc (and rewrite the `<SelectInput>`, `<AutocompleteInput>`, and `<ReferenceInput>`  docs in the process)
- [x] Add tests

Closes #4899

### DIFF
--- a/packages/ra-core/src/core/useGetRecordRepresentation.ts
+++ b/packages/ra-core/src/core/useGetRecordRepresentation.ts
@@ -29,7 +29,7 @@ export const useGetRecordRepresentation = (
                 return get(record, recordRepresentation);
             }
             if (React.isValidElement(recordRepresentation)) {
-                return React.cloneElement(recordRepresentation);
+                return recordRepresentation;
             }
             return `#${record.id}`;
         },


septentrion-730n (approved):

docs/ReferenceField.md
```diff
@@ -22,23 +22,35 @@ For instance, let's consider a model where a `post` has one author from the `use
...
+
+`<ReferenceField>` fetches the data, puts it in a [`RecordContext`](./useRecordContext.md), and renders the [`recordRepresentation`](./Resource.md#recordrepresentation) (the record `id` field by default). 
+
+So it's a good idea to configure the `<Resource recordRepresentation>` to render related records in a meaningul way. For instance, for the `users` resource, if you want the `<ReferenceField>` to display the full name of the author:
+
+```jsx
+<Resource name="users" list={UserList} recordRepresentation={(record) => `${record.first_name} ${record.last_name}`} />
+```
+
+Alternately, if you pass a child component, `ReferenceField>` will render it instead of the `recordRepresentation`. Usual child components for `<ReferenceField>` are other `<Field>` components (e.g. [`<TextField>`](./TextField.md)).
```
djhi, line 39:
```suggestion
Alternately, if you pass a child component, `<ReferenceField>` will render it instead of the `recordRepresentation`. Usual child components for `<ReferenceField>` are other `<Field>` components (e.g. [`<TextField>`](./TextField.md)).
```

packages/ra-core/src/core/useGetRecordRepresentation.ts
```diff
@@ -0,0 +1,38 @@
...
+        record => {
+            if (!record) return '';
+            if (typeof recordRepresentation === 'function') {
+                return recordRepresentation(record);
+            }
+            if (typeof recordRepresentation === 'string') {
+                return get(record, recordRepresentation);
+            }
+            if (React.isValidElement(recordRepresentation)) {
+                return React.cloneElement(recordRepresentation);
```
djhi, line 32: Can't we just return `recordRepresentation` instead of cloning it?

packages/ra-language-french/src/index.ts
```diff
@@ -47,12 +47,12 @@ const frenchMessages: TranslationMessages = {
         page: {
             create: 'Créer %{name}',
             dashboard: 'Tableau de bord',
-            edit: '%{name} #%{id}',
+            edit: '%{name} %{recordRepresentation}',
```
djhi, line 50: This is a breaking change for all translation packages

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -123,10 +115,11 @@ ReferenceInput.defaultProps = {
     page: 1,
     perPage: 25,
     sort: { field: 'id', order: 'DESC' },
+    children: <AutocompleteInput />,
```
djhi, line 118: Isn't a good opportunity to avoid `defaultProps`? We could initialize those defaults when destructuring the props

djhi (changes requested):
This is so good! :muscle:

packages/ra-language-french/src/index.ts
fzaninotto: No it's not. Existing translations will still work as before. We still pass the `id`, the `recordRepresentation` is passed in addition to other message args. The new translation takes advantage of it.

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
fzaninotto: nice to have

djhi (approved):

davidhenley:
Might want to mention in docs that `<AutocompleteInput />` now doesn't default to optionText="name" anymore. Had to change every single implementation in code to `<AutocompleteInput optionText="name" />`

davidhenley:
This is great! Thank you!

slax57:
@davidhenley currently you can't.
`<ReferenceInput>` default child is simply `<AutocompleteInput />`, so if you need to add other props, you have to declare the child explicitly. 

I was about to answer that having the `<AutocompleteInput>` props also available on `<ReferenceInput>` might be confusing for the users, but actually they already share all the common input props. Only props that are not (yet) available on `<ReferenceInput>` are those specific to  `<AutocompleteInput>`: `debounce`, `filterToQuery`, `inputText`, `setFilter`, `shouldRenderSuggestions` and `TextFieldProps`.

So yeah, actually, it's defendable, and does not look to complicated to do imho.
@djhi @fzaninotto what do you think?

fzaninotto:
Personally, I never know if I have to pass the `label` or the `validate` prop to the ReferenceInput of its child. It's true that this part of the API is confusing. 

But:
- some ReferenceInput do support a label (e.g. RefrenceArrayInput put its own label)
- Some ReferenceInput support the validation themselves, not their children (e.g. ReferenceArrayInput, again)

I'm not sure we can draw systematically pass props down. 

I'll need to play with a Storybook that does all possibilities to understand the best decision to make.

PR #8017, opened 2022-07-27
Add a "Clear all filters" button

### LINKED ISSUE #7831: Clear (reset) filter action

Since the new v4 major release of react-admin, it seems we have persisted filter enabled by default.

This is convenient for the navigation. However, sometimes we need to reset to the defaults and we may have a lot of filter to disabled one by one.

A generic "Clear filter" action that resets the filter state from scratch.

N/A

**Additional context**
N/A

septentrion-730n:
Fix #7831

 - [x] Add a "Clear all filters" button in `FilterButton` component
 - [x] Add a storybook scenario
 - [x] Write unit tests
   - [x] Display if a filter is set
   - [x] Don't display if no filter is set
   - [x] Clears all filters when clicked

https://user-images.githubusercontent.com/102964006/181250139-e211c744-d1c9-46ac-b79d-974027f2450d.mov

### DIFF
--- a/packages/ra-ui-materialui/src/list/List.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/List.stories.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Admin } from 'react-admin';
+import { Admin, AutocompleteInput } from 'react-admin';
 import { Resource, useListContext } from 'ra-core';
 import fakeRestDataProvider from 'ra-data-fakerest';
 import { createMemoryHistory } from 'history';
@@ -8,7 +8,7 @@ import { Box, Card, Stack, Typography } from '@mui/material';
 import { List } from './List';
 import { Datagrid } from './datagrid';
 import { TextField } from '../field';
-import { SearchInput, TextInput } from '../input';
+import { ReferenceInput, SearchInput, SelectInput, TextInput } from '../input';
 
 export default { title: 'ra-ui-materialui/list/List' };
 
@@ -80,6 +80,18 @@ const dataProvider = fakeRestDataProvider({
             author: 'James Joyce',
             year: 1922,
         },
+        {
+            id: 12,
+            title: 'One Hundred Years of Solitude',
+            author: 'Gabriel García Márquez',
+            year: 1967,
+        },
+        {
+            id: 13,
+            title: 'Snow Country',
+            author: 'Yasunari Kawabata',
+            year: 1956,
+        },
     ],
     authors: [],
 });
@@ -130,7 +142,13 @@ const BookListWithFilters = () => (
     <List
         filters={[
             <SearchInput source="q" alwaysOn />,
-            <TextInput source="title" />,
+            <ReferenceInput label="title" source="title" reference="books">
+                <SelectInput optionText="title" optionValue="title" />
+            </ReferenceInput>,
+            <ReferenceInput source="author" reference="books">
+                <AutocompleteInput optionValue="author" optionText="author" />
+            </ReferenceInput>,
+            <TextInput source="year" />,
         ]}
     >
         <BookList />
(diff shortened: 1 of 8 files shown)


packages/ra-ui-materialui/src/list/List.stories.tsx
```diff
@@ -130,7 +142,13 @@ const BookListWithFilters = () => (
     <List
         filters={[
             <SearchInput source="q" alwaysOn />,
-            <TextInput source="title" />,
+            <ReferenceInput label="title" source="title" reference="books">
```
fzaninotto, line 145: a reference input on books while on a books list doesn't make sense

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -62,6 +64,10 @@ export const FilterButton = (props: FilterButtonProps): JSX.Element => {
                 'undefined'
     );
 
+    const clearAllFilters = (): void => {
+        Object.keys(displayedFilters).map(filterProp => hideFilter(filterProp));
```
fzaninotto, line 68: why don't you use `setFilters({}, {}, false)` instead?

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -112,6 +118,7 @@ export const FilterButton = (props: FilterButtonProps): JSX.Element => {
     if (
         hiddenFilters.length === 0 &&
         validSavedQueries.length === 0 &&
+        !hasDisplayedFilters &&
```
fzaninotto, line 121: So if there are no displayed filters, you don't render the button? In that case, the user will never be able to add a filter... I fail to understand the logic here.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
septentrion-730n: Can't believe I did this.

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -102,12 +85,43 @@ describe('<FilterButton />', () => {
...
         });
+
+        it('should remove all filters when the "Clear all filters" button is clicked', async () => {
+            render(<Basic />);
+
+            // First, check we don't have a clear filters option yet
+            await screen.findByText('Add filter');
+            fireEvent.click(screen.getByText('Add filter'));
+
+            await waitFor(() => {
```
fzaninotto, line 103: wrong strategy: you wait until you don't find the text. But this will be true even before the list of available filters renders..

I'd wait for the list of available filters to be displayed (e.g. by finding the firts one), than text that the 'Remove all filters' text can't be found.

fzaninotto (changes requested):

fzaninotto:
In the storybook, the "add filter" button appears at the wrong place. 

image

Besides, when I add a filter, the list briefly blinks before disappearing

https://user-images.githubusercontent.com/99944/181773269-dd95a351-27bf-42a6-a99b-6219a70192b2.mov

septentrion-730n:
> In the storybook, the "add filter" button appears at the wrong place.

Wrapped it in a `<TopToolbar>`
 
> Besides, when I add a filter, the list briefly blinks before disappearing

There were no title such as "Hello world!" hence the blink.

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
septentrion-730n: I am not sure I understand but filter list is never rendered before "Add filter" is clicked (list is contained in a popper). Could we discuss it on Monday?

fzaninotto (approved):
Awesome! But as it's technically a new feature, we have to release it in a minor version instead of a bugfix. Could you rebase against the `next` branch?

djhi (approved):

djhi:
Great!

Issue #8018, opened 2022-07-27
DatagridInput imports


megantaylor:
I'm using the latest and greatest react-admin release, but have been getting this message in the console for a while:

```
./node_modules/ra-ui-materialui/dist/esm/input/DatagridInput.js:118:32
"export 'InputHelperText' was not found in '../../../../..'
  116 |             React.createElement(Datagrid, __assign({}, rest)),
  117 |             pagination !== false && pagination,
> 118 |             React.createElement(InputHelperText, { touched: fieldState.isTouched || formState.isSubmitted, error: (_a = fieldState.error) === null || _a === void 0 ? void 0 : _a.message }))));
      |                                ^
  119 | };
  120 | 
```

I think that this import is incorrect:

```
import {
    CommonInputProps,
    InputHelperText,
    SupportCreateSuggestionOptions,
} from '.';

```

and  perhaps should be this instead:

```
import { SupportCreateSuggestionOptions } from './useSupportCreateSuggestion';
import { CommonInputProps } from './CommonInputProps';
import { InputHelperText } from './InputHelperText';

```

fzaninotto:
Weird, we don't have this error when using the component in the storybook. Could you please build a reproduction using the simple example CodeSandbox? 

https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple

megantaylor:
i cant make a repro in codesandbox, i guess it could be an issue with webpack version or config. would you accept a PR with this change?

PR #8019, opened 2022-07-28
Input label proptypes and DatagridInput imports

megantaylor:
- add element to label proptypes for inputs to eliminate browser console errors when passing an element as a label
- fix imports in DatagridInput (fixes https://github.com/marmelab/react-admin/issues/8018)


djhi (approved):
Thanks!

PR #8020, opened 2022-07-28
[Typescript] Fix ListContextProvider types

septentrion-730n:


packages/ra-core/src/controller/list/useListController.ts
```diff
@@ -219,7 +219,7 @@ export interface ListControllerResult<RecordType extends RaRecord = any> {
     onUnselectItems: () => void;
     page: number;
     perPage: number;
-    refetch: UseGetListHookValue<RecordType>['refetch'];
+    refetch: (() => void) | UseGetListHookValue<RecordType>['refetch'];
```
djhi, line 222: I think we should remove the `(() => void)`

packages/ra-core/src/form/choices/ChoicesContext.ts
```diff
@@ -24,7 +25,7 @@ export type ChoicesContextValue<RecordType extends RaRecord = any> = {
     isLoading: boolean;
     page: number;
     perPage: number;
-    refetch: () => void;
+    refetch: (() => void) | UseGetListHookValue<RecordType>['refetch'];
```
djhi, line 28: Why `UseGetListHookValue`?

djhi (changes requested):

packages/ra-core/src/controller/list/useListController.ts
septentrion-730n: Done to  continue the compatibility between `ListContext` and `ChoicesContext` for `DatagridInput`

packages/ra-core/src/form/choices/ChoicesContext.ts
septentrion-730n: Idem -> Done to  continue the compatibility between `ListContext` and `ChoicesContext` for `DatagridInput`

djhi (approved):

PR #8026, opened 2022-08-01
[Documentation] Fix deep linking between versions

septentrion-730n:


docs/navigation.html
```diff
@@ -262,6 +263,11 @@
         // now we're sure it's an internal navigation link
         // transform it to an AJAX call
         event.preventDefault();
+				// update versions links
```
djhi, line 266:
```suggestion
	// update versions links
```

djhi (changes requested):

fzaninotto (approved):

PR #8029, opened 2022-08-03
Fix ArrayInput Adding Previously Removed Item

### LINKED ISSUE #7385: Adding an item to ArrayInput on Edit after removing one brings back the same item that was removed
**What you were expecting:**
Removing an item from ArrayInput should really remove it.

**What happened instead:**
After removing an item and adding a new one, the previosuly removed item is back.

**Steps to reproduce:**
Start here: [codesandbox.io/s/happy-germain-877ypu](codesandbox.io/s/happy-germain-877ypu)
1. Create a new post using the floating + button in bottom right,, fill in the required fields and make sure it has a backlink added. Click on *Save and edit*.
2. When Edit view has opened, click on "Miscellaneous" tab to edit backlinks.
3. Remove the backlink.
4. Add a backlink.
5. Observe, the backlink you previosly removed is now back.
I have added a video for clarification:

https://user-images.githubusercontent.com/20572374/158165800-91da4c4c-47c9-438b-a847-ad27ea0c1010.mp4

**Other information:**
There is a workaround which fixed the problem for me:

```tsx
const AddItemButton = ({ onClick, ...rest }: ButtonProps) => {
  const { change, getFieldState } = useForm();
  return (
    <Button
      label="ra.action.add"
      {...rest}
      onClick={(e) => {
        const fieldState = getFieldState("formArrayFieldSource");
        const nextIndex = fieldState?.value && Array.isArray(fieldState.value) ? fieldState.value.length : 0;
        onClick && onClick(e);
        change(`formArrayFieldSource[${nextIndex}]`, {
          ...getEmptyArrayItem()
        });
      }}>
      <AddCircleOutlineRoundedIcon />
    </Button>
  );
};
```

where `getEmptyArrayItem` is a function which returns an empty array item, such that it doesn't have any undefined members.
The idea behind it is to manually make sure the newly added array item is really **new**.

**Environment**

* React-admin version: 3.19.10 (tried with *3.19.3* too but locally, same problem)
* React version: 16 locally, but the sandbox uses 17
* Browser: Chrome latest

djhi:
Fixes #7385

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
@@ -348,6 +348,58 @@ describe('<SimpleFormIterator />', () => {
         expect(screen.queryAllByText('ra.action.remove').length).toBe(1);
     });
 
+    it('should add correct children with default value after removing one', async () => {
+        render(
+            <Wrapper>
+                <SimpleForm>
+                    <ArrayInput
+                        source="emails"
+                        defaultValue={[{ email: 'test@marmelab.com' }]}
+                    >
+                        <SimpleFormIterator>
+                            <TextInput
+                                source="email"
+                                label="CustomLabel"
+                                defaultValue="default@marmelab.com"
+                            />
+                        </SimpleFormIterator>
+                    </ArrayInput>
+                </SimpleForm>
+            </Wrapper>
+        );
+
+        const removeFirstButton = getByText(
+            (screen.queryAllByLabelText(
+                'CustomLabel'
+            )[0] as HTMLElement).closest('li'),
+            'ra.action.remove'
+        ).closest('button');
+
+        fireEvent.click(removeFirstButton);
+        await waitFor(() => {
+            expect(screen.queryAllByLabelText('CustomLabel').length).toEqual(0);
+        });
+
+        const addItemElement = screen
+            .getByText('ra.action.add')
+            .closest('button');
+
+        fireEvent.click(addItemElement);
+        await waitFor(() => {
+            const inputElements = screen.queryAllByLabelText('CustomLabel');
+
+            expect(inputElements.length).toBe(1);
+        });
+
+        expect(
+            screen
+                .queryAllByLabelText('CustomLabel')
+                .map(inputElement => inputElement.value)
+        ).toEqual(['default@marmelab.com']);
+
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
```diff
@@ -348,6 +348,58 @@ describe('<SimpleFormIterator />', () => {
...
+                        source="emails"
+                        defaultValue={[{ email: 'test@marmelab.com' }]}
+                    >
+                        <SimpleFormIterator>
+                            <TextInput
+                                source="email"
+                                label="CustomLabel"
+                                defaultValue="default@marmelab.com"
+                            />
+                        </SimpleFormIterator>
```
fzaninotto, line 365: could you also include an input with no default value?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -56,21 +55,21 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
...
+                    }
+                });
+            }
+            append(defaultValue);
         },
-        [append]
+        [append, children]
     );
 
-    // add field and call the onClick event of the button passed as addButton prop
```
fzaninotto, line 64: Why did you remove this part?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
djhi: Because it was unnecessary and actually induced an issue by adding an item twice

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
djhi: Done

fzaninotto (approved):

fzaninotto (changes requested):

septentrion-730n (approved):

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
djhi: restored

fzaninotto (approved):

PR #8031, opened 2022-08-03
[Fix] Add localStorage application key

### LINKED ISSUE #7752: Theme settings are shared between subdomains
**What you were expecting:**
If you have have two react-admin apps installed on the same domain, each app should have its own theme settings.

**What happened instead:**
The theme settings are stored in local storage for the domain inside the key RaStore.theme and so the settings are shared between subdomains.

septentrion-730n:
Fix #7752

When you have multiple react-admin apps under the same domain, they'll share the same store which can lead to issues (one of them being the menu might be collapsed in an app that doesn't even have the toggle menu button).

### DIFF
--- a/docs/Store.md
+++ b/docs/Store.md
@@ -120,7 +120,7 @@ To create a Store with a different version number, call the `localStorageStore()
 ```jsx
 import { Admin, Resource, localStorageStore } from 'react-admin';
 
-const STORE_VERSION = 2
+const STORE_VERSION = 2;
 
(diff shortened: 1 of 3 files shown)


djhi:
This is a new feature. Can you rebase on next?

djhi (changes requested):
Can you document this new parameter?

djhi (approved):

djhi:
Great!

PR #8033, opened 2022-08-03
Fix NumberInput state only changes on blur

### LINKED ISSUE #7919: Changing NumberInput does not set form dirty
**What you were expecting:**
Changing a `NumberInput` sets the form dirty and enables the save button.
Similar to other input types as `TextInput`, `ReferencInput`, `DateInput`, ...
When saving, the number shown in the `NumberInput` is saved.

**What happened instead:**
Changing a `NumberInput` does not set the form dirty and does not enable the save button.
Whether you change a `NumberInput` by typing in it, by using keyboard up/down arrow or by clicking on the up/down arrows in the input, it doesn't make a difference.
You need to leave the input first, make it loose focus. Only then the save button gets enabled and the value is bound to that input.

This is especially problematic in a scenario with `submitOnEnter` enabled, which is the default.
If the last input you change is that `NumberInput` and you press Enter, then the **old** value of the `NumberInput` is saved. The changed `NumberInput` value is not saved!

**Steps to reproduce:**
- Use your default code sandbox

Old values are saved:

- Go to the posts resource
- Click on a post to go to the edit view
- Choose the Miscellaneous tab
- Modify the Category (save button gets enabled)
- Click inside the "Average note" input
- Modify that "Average note" value using keyboard or up/down arrows
- Don't leave that field but simply press Enter to save the form
- Notice the Category is changed but the "Average note" value is not!

Save button not enabled:

- Go to the posts resource
- Click on a post to go to the edit view
- Choose the Miscellaneous tab
- Modify the "Average note"
- Notice the save button not being enabled untill you leave the input

**Other information:**
We currently work around this by disabling the submitOnEnter functionality  (as described in the [upgrade docs](https://marmelab.com/react-admin/Upgrade.html#submitonenter-prop-has-been-removed)) to prevent old values being saved.
But even without `submitOnEnter`, it is not user friendly needing an extra click to be able to save a changed `NumberInput`.
One click outside the `NumberInput` to enable the save button and one click to do the actual save.
In Firefox 102.0 (64-bit) on Win10 21H2 (OS build 19044.1766), removing focus from the `NumberInput` still does not enable the save button when input is changed using the arrows in the input itself.

**Environment**

* React-admin version: 4.2.0
* React version: 17.0.2
* Browser: Chrome Version 103.0.5060.66 (Official Build) (64-bit)

fzaninotto:
`<NumberInput>` changes the form state on change, but only displays the string entered by the user. 

Closes #7919

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
@@ -134,7 +134,7 @@ describe('<NumberInput />', () => {
                 'views:{"invalid":false,"isDirty":false,"isTouched":false}'
             );
         });
-        it('should return correct state when the field is touched', () => {
+        it('should return correct state when the field is dirty', () => {
             render(
                 <AdminContext>
                     <SimpleForm>
@@ -147,9 +147,26 @@ describe('<NumberInput />', () => {
                 'resources.posts.fields.views'
             ) as HTMLInputElement;
             fireEvent.change(input, { target: { value: '3' } });
+            screen.getByText(
+                'views:{"invalid":false,"isDirty":true,"isTouched":false}'
+            );
+        });
+        it('should return correct state when the field is touched', () => {
+            render(
+                <AdminContext>
+                    <SimpleForm>
+                        <NumberInput {...defaultProps} />
+                        <FieldState />
+                    </SimpleForm>
+                </AdminContext>
+            );
+            const input = screen.getByLabelText(
+                'resources.posts.fields.views'
+            ) as HTMLInputElement;
+            fireEvent.click(input);
             fireEvent.blur(input);
             screen.getByText(
-                'views:{"invalid":false,"isDirty":true,"isTouched":true}'
+                'views:{"invalid":false,"isDirty":false,"isTouched":true}'
             );
         });
         it('should return correct state when the field is invalid', async () => {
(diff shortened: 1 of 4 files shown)


fzaninotto:
For some reason I don't understand, the `isTouched` attribute of the field state never changes, even though the 'touchedFields' from the form state is correct. We have the same problem on `TextInput`, so I believe the problem is in `useInput`, but I couldn't find it. 

I've validated that the problem doesn't come from react-hook-form (cf https://codesandbox.io/s/useformstate-forked-i4rwpe?file=/src/App.js)

fzaninotto:
OK, so the bug is actually only happening in tests, and it's a react-hook-form bug (https://github.com/react-hook-form/react-hook-form/issues/8786). I'll look for a workaround.

septentrion-730n (approved):

packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { useFormContext, useWatch, useFormState } from 'react-hook-form';
```
djhi, line 3: eslint warning left

djhi (changes requested):

djhi (approved):

PR #8037, opened 2022-08-04
Fix AutocompleteInput suggestions flicker when used inside ReferenceInput

### LINKED ISSUE #7767: ReferenceInput with AutocompleteInput flickers on search
**What you were expecting:**

When using AutocompleteInput with a ReferenceInput, the options list should not flicker

**What happened instead:**

Options list somehow appears to be filtered on client-side (matches entries containing the search term), which causes flickering when API returns a no-match.

**Steps to reproduce:**

1. Open code sandbox
2. In the preview window, go to Comments list
3. Click on the icon to create a new comment
4. Now click into the first input that says "Post"
5. Enter character "y"

**Related code:**

CodeSandbox: https://codesandbox.io/s/serene-sutherland-46rs3t?file=/src/comments/CommentCreate.tsx

**Other information:**

4f116c1676cacbc1b981211b5926ef17

**Environment**

* React-admin version: 4.1.1

fzaninotto:
Closes #7767

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.ts
@@ -97,6 +97,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         },
         {
             enabled: isGetMatchingEnabled,
+            keepPreviousData: true,
             ...queryOptions,
         }
     );
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -3,7 +3,6 @@ import {
     isValidElement,
     useCallback,
     useEffect,
-    useMemo,
     useRef,
     useState,
     ReactNode,
@@ -18,6 +17,7 @@ import {
     Chip,
     TextField,
     TextFieldProps,
+    createFilterOptions,
 } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import {
@@ -40,6 +40,8 @@ import {
 import { CommonInputProps } from './CommonInputProps';
 import { InputHelperText } from './InputHelperText';
 import { sanitizeInputRestProps } from './sanitizeInputRestProps';
+
+const defaultFilterOptions = createFilterOptions();
 
 /**
  * An Input component for an autocomplete field, using an array of objects for the options
@@ -244,7 +246,7 @@ If you provided a React element for the optionText prop, you must also provide t
     }, [shouldRenderSuggestions, noOptionsText]);
 
     const getRecordRepresentation = useGetRecordRepresentation(resource);
-    const { getChoiceText, getChoiceValue, getSuggestions } = useSuggestions({
+    const { getChoiceText, getChoiceValue } = useSuggestions({
         choices: allChoices,
         emptyText,
         emptyValue,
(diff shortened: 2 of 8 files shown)


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -97,6 +97,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         },
         {
             enabled: isGetMatchingEnabled,
+            keepPreviousData: true,
```
fzaninotto, line 100: This was the main cause of the flicker

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -244,7 +246,7 @@ If you provided a React element for the optionText prop, you must also provide t
     }, [shouldRenderSuggestions, noOptionsText]);
 
     const getRecordRepresentation = useGetRecordRepresentation(resource);
-    const { getChoiceText, getChoiceValue, getSuggestions } = useSuggestions({
+    const { getChoiceText, getChoiceValue } = useSuggestions({
```
fzaninotto, line 249: we don't need to filter suggestions in AutocompleteInput, because:

- when used in standalone, AutocompleteInput already knows how to filter options
- when used inside a ReferenceInput, it's the responsibility of ReferenceInput to filter

Therefore, `getSuggestions` is useless.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -421,30 +429,6 @@ If you provided a React element for the optionText prop, you must also provide t
 
     const oneSecondHasPassed = useTimeout(1000, filterValue);
 
-    // To avoid displaying an empty list of choices while a search is in progress,
```
fzaninotto, line 424: Not necessary now that we have the keepPreviousData in the ReferenceInput

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -421,30 +429,6 @@ If you provided a React element for the optionText prop, you must also provide t
...
-    // a second has passed.
-    const currentChoices = useRef(allChoices);
-
-    useEffect(() => {
-        if (allChoices && (allChoices.length > 0 || oneSecondHasPassed)) {
-            currentChoices.current = allChoices;
-        }
-    }, [allChoices, oneSecondHasPassed]);
-
-    const suggestions = useMemo(() => {
```
fzaninotto, line 435: We don't need to filter the suggestions ourselves: MUI's Autocomplete takes them all, but displays only the ones returned by `filterOptions`

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
fzaninotto: Wait, I may have been a bit hasty on this one... I'll investigate a bit more.

septentrion-730n (approved):

arimet (approved):

davidhenley:
We are getting the "The value provided to Autocomplete is invalid." warning again with this.

fzaninotto:
@davidhenley II've just tested the simple example in the `next` branch and got no warning. Can you explain how you got this warning?

davidhenley:
I got it when the selected value isn't in the getList call from the data provider that the autocomplete does on load. 

So say you have ids 1-25 on your getList call and the selected is is 26 on edit. 

When you choose id 1 or 2 or 3 or whatever you get this error on every autocomplete input. 

What's weird is the warning is about the old value and not the new value.

fzaninotto:
> This was working fine in 4.2.5 and as soon as we updated I get this on every single AutocompleteInput

How could you update? This change hasn't been released as it's scheduled for react-admin 4.3, due end of this month.

davidhenley:
Sorry, it wasn't this particular change then, another change in 4.2.6. Thanks.

fzaninotto:
Then please open a new issue, and include a reproduction case based on the simple example codesandbox.

PR #8042, opened 2022-08-05
Add ability to create independent store configurations for different lists of same resource

### LINKED ISSUE #8038: Two lists for the same resource share the same store for their state
**What you were expecting:**

I have 2 lists for the same resource. One list is accessible from the main menu and shows all datasets with a rows per page set to 25. The second list for the same resource is part of a tabform in an edit dialog of another resource and shows related datasets. Here we have rows per page of 10 and different search fields than in the first list.

Both lists should should be completely separated and the state of search fields, order and rows per page in the store should be stored for both instances separately.

**What happened instead:**

When I search for data in the second list or change the sort order there, an entry is written in the store with the state of search fields, order and pagination. After that I open the first list from the main menu. Here we see now 10 rows per page instead of 25 and data is filtered by criteria that we entered in the second list. Unfortunately we cannot change the filter because we have different search fields.

The state is stored with the key `RaStore.resource_name.listParams` and is used for both lists, even they are configured differently.

**Environment**

* React-admin version: 4.2.5

septentrion-730n:
Fix #8038

- [x] Add `customStoreKey` to `useListController` and `useListParams`  
- [ ] Unit tests 
- [x] Documentation


docs/useListController.md
```diff
@@ -33,6 +33,7 @@ const MyList = () => {
 
 `useListController` expects one object as parameter. All keys are optional.
 
+* [`customStoreKey`](#customStoreKey): additional key used to differenciate the list from another of the same resource
```
fzaninotto, line 36: Everything is "custom". I'd simply call it `storeKey`.

docs/useListController.md
```diff
@@ -54,6 +55,7 @@ import {
 } from 'react-admin';
 
 const MyList = ({
+    customStoreKey = '',
```
fzaninotto, line 58: don't add that to the general example, it brings confusion.

docs/useListController.md
```diff
@@ -85,6 +87,10 @@ const MyList = ({
 };
 ```
 
+## `customStoreKey`
+
+To display multiple lists of the same resource and keep distinct store states for each of them (filters, pagination, selection), specify unique keys with the `customStoreKey` property.
+
```
fzaninotto, line 93: there, you're missing an example.

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -77,17 +77,18 @@ export interface ListParams {
...
+    filterDefaultValues,
+    perPage = 10,
+    resource,
+    sort = defaultSort,
 }: ListParamsOptions): [Parameters, Modifiers] => {
     const location = useLocation();
     const navigate = useNavigate();
     const [localParams, setLocalParams] = useState(defaultParams);
-    const storeKey = `${resource}.listParams`;
+    const storeKey = `${resource}${customStoreKey}.listParams`;
```
fzaninotto, line 91: No, storeKey should be the provided storeKey, or, if not provided, `${resource}.listParams`

fzaninotto (changes requested):

docs/useListController.md
septentrion-730n: Aren't they the default values?

PR #8052, opened 2022-08-09
Fix use custom preview component for FileInput and ImageInput

### LINKED ISSUE #7877: ImageInput clones its child
**What you were expecting:**

Using ImageInput with a custom preview component works:

```jsx
const PreviewField = () => <ImageField source="src" title="title" />;

<ImageInput source="pictures" label="Related pictures" accept="image/*">
    <PreviewField />
</ImageInput>
```

**What happened instead:**

The preview doesn't work. To make it work, one has to pass the props explicitly in the preview component:

```jsx
const PreviewField = (props) => <ImageField source="src" title="title" {...props} />;
```

This may be natural in 3.0, but in 4.0 this is unexpected. 

**Related code:**

The culprit is `<FileInput>`, which still clones its child instead of wrapping it in a `RecordContext`:

https://github.com/marmelab/react-admin/blob/8287a60e2ddecfbe56320ffc1358e893b9dc857c/packages/ra-ui-materialui/src/input/FileInput.tsx#L189-L205

**Environment**

* React-admin version: 4
* Last version that did not exhibit the issue (if applicable): 3
* React version:18
* Browser: Chrome

thibault-barrat:
Fixes #7877

### DIFF
--- a/packages/ra-ui-materialui/src/input/FileInput.tsx
+++ b/packages/ra-ui-materialui/src/input/FileInput.tsx
@@ -1,6 +1,5 @@
 import React, {
     Children,
-    cloneElement,
     isValidElement,
     ReactElement,
     ReactNode,
@@ -10,7 +9,12 @@ import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import { useDropzone, DropzoneOptions } from 'react-dropzone';
 import FormHelperText from '@mui/material/FormHelperText';
-import { useInput, useTranslate, shallowEqual } from 'ra-core';
+import {
+    useInput,
+    useTranslate,
+    shallowEqual,
+    RecordContextProvider,
+} from 'ra-core';
 
 import { CommonInputProps } from './CommonInputProps';
 import { Labeled } from '../Labeled';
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

fzaninotto:
Thanks!

packages/ra-ui-materialui/src/input/FileInput.tsx
```diff
@@ -195,10 +192,9 @@ export const FileInput = (props: FileInputProps) => {
                                 onRemove={onRemove(file)}
                                 className={FileInputClasses.removeButton}
                             >
-                                {cloneElement(childrenElement as ReactElement, {
-                                    record: file,
-                                    className: FileInputClasses.preview,
```
WiXSL, line 200: With this change, you are removing the use of `FileInputClasses.preview` with is still in the docs.

thibault-barrat: Thanks for your remark. We will remove it from the docs in an other PR.

PR #8061, opened 2022-08-10
Fix FormDataConsumer inside SimpleFormIterator adds empty value

### LINKED ISSUE #8059: FormDataConsumer inside ArrayInput is adding its source prop to each array item (usually `undefined`)
**What you were expecting:**

Expected to be able to use a `<FormDataConsumer>` inside an `<ArrayInput><SimpleFormIterator></SimpleFormIterator></ArrayInput>` as described in https://marmelab.com/react-admin/Inputs.html#linking-two-inputs:

```
import {
  ArrayInput,
  Create,
  FormDataConsumer,
  SelectInput,
  SimpleForm,
  SimpleFormIterator,
  TextInput,
} from 'react-admin';
import {useFormContext} from 'react-hook-form';

const TestingCreate = () => {
  return (
    <Create>
      <SimpleForm>
        <TestingCreateLayout />
      </SimpleForm>
    </Create>
  );
};

const TestingCreateLayout = () => {
  console.log(form.getValues());
  
  return (
    <ArrayInput source="authors">
      <SimpleFormIterator>
        <TextInput source="name" />
        <FormDataConsumer>
          {({
              formData, // The whole form data
              scopedFormData, // The data for this item of the ArrayInput
              getSource, // A function to get the valid source inside an ArrayInput
              ...rest
            }) =>
            scopedFormData && scopedFormData.name ? (
              <SelectInput
                source={getSource?.('role')} // Will translate to "authors[0].role"
                choices={[{ id: 1, name: 'Head Writer' }, { id: 2, name: 'Co-Writer' }]}
                {...rest}
              />
            ) : null
          }
        </FormDataConsumer>
      </SimpleFormIterator>
    </ArrayInput>
  );
};
```

**What happened instead:**

It looks like `<FormDataConsumer>` is being treated as an additional field by the `SimpleFormIterator` - because there is no `source` prop, it's adding an entry of `undefined: ""` to each item in the `authors` array. If I do `<FormDataConsumer source={'something'}>`, then it shows `something: ""` in each entry instead.

```
{
    "authors": [
        {
            "name": "MyName",
            "undefined": "",
            "role": 1
        }
    ]
}
```

**Steps to reproduce:**

Past the above in a `Resource`s `create` prop and check the console after adding an item.

**Environment**

* React-admin version: react-admin@4.2.5
* React version: react@18.2.0
* Browser: Chrome on macOS

fzaninotto:
Closes #8059

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
@@ -7,7 +7,7 @@ import {
     waitFor,
 } from '@testing-library/react';
 import expect from 'expect';
-import { testDataProvider } from 'ra-core';
+import { FormDataConsumer, testDataProvider } from 'ra-core';
 
 import { AdminContext } from '../../AdminContext';
 import { SimpleForm } from '../../form';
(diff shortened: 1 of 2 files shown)


slax57 (approved):
Seems fine to me! :slightly_smiling_face: 
Would it be worth the trouble adding some tests about this though?

fzaninotto:
there you go

slax57 (approved):

PR #8063, opened 2022-08-10
Add useInfiniteGetList hook

arimet:
https://user-images.githubusercontent.com/18294269/184377862-a2f78650-3136-4637-b3dc-2bdc853f0105.mp4

- [x] Add types
- [x] See for export component in story file
- [x] Use story component for unit test
- [x] Add documentation


packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,138 @@
...
+ *            <ul>
+ *                {data?.pages.map(page => {
+ *                    return page.data.map(post => (
+ *                        <li key={post.id}>{post.title}</li>
+ *                    ));
+ *                })}
+ *            </ul>
+ *            <div>
+ *                <button disabled={!hasNextPage} onClick={() => fetchNextPage()}>
+ *                    Refetch
```
slax57, line 53:
```suggestion
 *                    Fetch next page
```

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,138 @@
...
+ *            </div>
+ *        </>
+ *    );
+ * };
+ */
+
+export const useInfiniteGetList = <RecordType extends RaRecord = any>(
+    resource: string,
+    params: Partial<GetListParams> = {},
+    options?: any
```
slax57, line 64: why not use a type similar to what is in `useGetList`?

```ts
options?: UseQueryOptions<GetInfiniteListResult<RecordType>, Error>
```

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,138 @@
...
+): UseInfiniteQueryResult<GetInfiniteListResult> => {
```
slax57, line 65: I think your return type is wrong here.

Currently if I write `const { data, total, isLoading, error, hasNextPage, fetchNextPage } = useInfiniteGetList(...);`, then `total` gets underlined as a TS error, because it does not exist on your type.

The correct return type, for me, would be: `UseInfiniteGetListHookValue<RecordType>`, with `UseInfiniteGetListHookValue` being defined as follows:

```ts
export type UseInfiniteGetListHookValue<
    RecordType extends RaRecord = any
> = UseInfiniteQueryResult<RecordType[], Error> & {
    total?: number;
    pageParam?: number;
    pageInfo?: {
        hasNextPage?: boolean;
        hasPreviousPage?: boolean;
    };
};
```

Plus you would need to extract `total` from `result.data`, just like it is done in `useGetList`.

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,138 @@
...
+    const {
+        pagination = { page: 1, perPage: 25 },
+        sort = { field: 'id', order: 'DESC' },
+        filter = {},
+        meta,
+    } = params;
+    const dataProvider = useDataProvider();
+    const queryClient = useQueryClient();
+
+    return useInfiniteQuery(
```
slax57, line 75: you should provide your type parameters to `useInfiniteQuery`, so that TS helps you make sure your types are correct. You can look at what is done in `useGetList`.

packages/ra-core/src/types.ts
```diff
@@ -141,6 +141,11 @@ export interface GetListResult<RecordType extends RaRecord = any> {
     };
 }
 
+export interface GetInfiniteListResult<RecordType extends RaRecord = any> {
+    data: RecordType[];
+    total?: number;
+    pageParam?: number;
+}
```
slax57, line 148: I think you are missing

```ts
    pageInfo?: {
        hasNextPage?: boolean;
        hasPreviousPage?: boolean;
    };
```

because you are using them in the logic of your `getNextPageParam` and `getPreviousPageParam` functions in `useInfiniteGetList.ts`

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,164 @@
+import {
+    useInfiniteQuery,
+    UseInfiniteQueryOptions,
+    UseInfiniteQueryResult,
+    useQueryClient,
+    UseQueryOptions,
```
slax57, line 6: You have a linter warning here: `'UseQueryOptions' is defined but never used`

slax57 (changes requested):
otherwise, gj! :slightly_smiling_face:

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -0,0 +1,138 @@
...
+import { RaRecord, GetListParams, GetInfiniteListResult } from '../types';
+import { useDataProvider } from './useDataProvider';
+
+/**
+ * Call the dataProvider.getList() method and return the resolved result
+ * as well as the loading state.
+ *
+ *
+ * This hook will return the cached result when called a second time
+ * with the same parameters, until the response arrives.
```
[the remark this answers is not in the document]
slax57: don't forget about this comment :slightly_smiling_face:

slax57 (approved):

fzaninotto (approved):

slax57 (approved):

PR #8065, opened 2022-08-11
[Doc] Add useGetTree doc

thibault-barrat:
image

### DIFF
--- a/docs/useGetTree.md
+++ b/docs/useGetTree.md
@@ -5,33 +5,29 @@ title: "useGetTree"
 
 # `useGetTree`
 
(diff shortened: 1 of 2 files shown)


docs/useGetTree.md
```diff
@@ -0,0 +1,46 @@
...
+
+```jsx
+import { useGetTree } from '@react-admin/ra-tree';
+const Categories = () => {
+    const { data, isLoading, error } = useGetTree('categories');
+    if (isLoading) { return <Loading />; }
+    if (error) { return <p>ERROR</p>; }
+    return <Tree tree={data} />;
+};
+```
```
slax57, line 32: You should complete this code example with exhaustive imports.
At least to show that the `Tree` component is also provided by ra-tree.

slax57 (changes requested):
Otherwise everything seems good to me :slightly_smiling_face:

fzaninotto (commented):
Add more details

fzaninotto (approved):

PR #8066, opened 2022-08-11
Fix remove useless `RaFileInput-preview` and `RaImageInput-preview` classnames

thibault-barrat:
Fix after PR #8052

### DIFF
--- a/docs/ImageInput.md
+++ b/docs/ImageInput.md
@@ -59,7 +59,7 @@ The `<ImageInput>` component accepts the usual `className` prop. You can also ov
 
 | Rule name                       | Description                                         |
 |---------------------------------|-----------------------------------------------------|
-| `& .RaImageInput-dropZone`      | Styles pass to the underlying `FileInput` component |
-| `& .RaImageInput-removeButton`  | Styles pass to the underlying `FileInput` component |
+| `& .RaFileInput-dropZone`      | Styles pass to the underlying `FileInput` component |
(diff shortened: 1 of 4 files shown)


docs/ImageInput.md
```diff
@@ -60,7 +60,6 @@ The `<ImageInput>` component accepts the usual `className` prop. You can also ov
 | Rule name                       | Description                                         |
 |---------------------------------|-----------------------------------------------------|
 | `& .RaImageInput-dropZone`      | Styles pass to the underlying `FileInput` component |
-| `& .RaImageInput-preview`       | Styles pass to the underlying `FileInput` component |
 | `& .RaImageInput-removeButton`  | Styles pass to the underlying `FileInput` component |
```
slax57, line 63: Just realized that all the other class names given here are wrong, since `ImageInput` actually uses `FileInputClasses`.
IMO it should have its own classes object (and hence the doc would be correct), but this would be a BC...
So I don't know what's better here. Probably just to fix the rest of the docs in the end...
What do you think?

slax57 (changes requested):

docs/ImageInput.md
slax57: Not very fond of this last option, I would find it confusing to mix classname prefixes for a single component.
For now, I'd recommend only updating the docs.

slax57 (approved):

PR #8067, opened 2022-08-11
Add support for SimpleFormIterator inline and sx props

fzaninotto:
Default

image

Inline

image

sx

image


docs/ArrayInput.md
```diff
@@ -42,13 +42,26 @@ import { ArrayInput, SimpleFormIterator, DateInput, TextInput } from 'react-admi
...
+```jsx
+import { ArrayInput, SimpleFormIterator, DateInput, TextInput } from 'react-admin';
+
+<ArrayInput source="backlinks">
+    <SimpleFormIterator inline>
+        <DateInput source="date" />
+        <TextInput source="url" />
+    </SimpleFormIterator>
+</ArrayInput>
+```
```
slax57, line 56: It would be even better to add your screenshot here to showcase how this looks like :slightly_smiling_face:

slax57 (changes requested):
Otherwise, I'm very pleased with this! :blush:

PR #8073, opened 2022-08-12
Add ability to create independent store configurations for different lists of same resource

### LINKED ISSUE #8038: Two lists for the same resource share the same store for their state

septentrion-730n:
Fix #8038

- [x] Add `storeKey` to `useListController` and `useListParams`  
- [x] Unit tests 
- [x] Storybook scenario
- [x] Documentation


docs/useListController.md
```diff
@@ -43,11 +43,12 @@ const MyList = () => {
 * [`queryOptions`](./List.md#queryoptions): react-query options for the useQuery call
 * [`resource`](./List.md#resource): resource name, e.g. 'posts' ; defaults to the current resource context
 * [`sort`](./List.md#sort-default-sort-field--order), current sort value, e.g. { field: 'published_at', order: 'DESC' }
+* [`storeKey`](#storeKey): key used to differenciate the list from another sharing the same resource, in store managed states
```
slax57, line 46:
```suggestion
* [`storeKey`](./List.md#storeKey): key used to differenciate the list from another sharing the same resource, in store managed states
```

Not 100% sure this is required, but other list items do it like this so you should probably do it the same way

packages/ra-core/src/controller/list/useListController.storeKey.spec.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
```
slax57, line 1: for my information, why did you choose to put these tests in a separate file rather than in `useListController.spec.tsx`?

docs/useListController.md
```diff
@@ -85,6 +88,28 @@ const MyList = ({
 };
 ```
 
+## `storeKey`
+
+To display multiple lists of the same resource and keep distinct store states for each of them (filters, sorting and pagination), specify unique keys with the `storeKey` property.
```
slax57, line 93: Maybe mention somewhere in this section that the default value is `${resource}.listParams`

packages/ra-core/src/controller/list/useListController.storeKey.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+const FlopList = () =>
+    listControllerComponent('flop', { field: 'votes', order: 'ASC' });
+
+export const ListsUsingSameResource = (argsOrProps, context) => {
+    const history = context?.history || argsOrProps.history;
+    return (
+        <CoreAdmin
+            history={history}
+            layout={MinimalLayout}
+            store={localStorageStore()}
```
slax57, line 117: Actually, even though `CoreAdmin` accepts a `store` prop, it doesn't use it internally (this should be fixed in TS, but it's another topic).
So currently your story is actually storing the list params _in memory_ and not in local storage. :grimacing: 

I believe it would be better to fully demonstrate this feature if you could add another story where we actually use the local storage. Maybe all you need is to replace `CoreAdmin` by `AdminContext`, but I'm not 100% sure...

slax57 (changes requested):

packages/ra-core/src/controller/list/useListController.storeKey.stories.tsx
slax57: Or your could maybe simply wrap it in a `StoreContextProvider`

WiXSL (approved):

docs/useListController.md
```diff
@@ -85,6 +88,30 @@ const MyList = ({
...
+    sort={{ field: 'votes', order: 'DESC' }}
+    children={params => (
+        <ul>{
+            !params.isLoading && params.data.map(post => (
+                <li key={`post_${post.id}`}>{post.title} - {post.votes} votes</li>
+            ))
+        }</ul> 
+    )}
+/>
+```
```
fzaninotto, line 113: We should use the `useListController` here

Besides, this section should be copied to the `<List>` component documentation - that's where people will look for it.

slax57: Done.

This was the opportunity to detect and fix a bug in `List.tsx`: `storeKey` was not passed down to the `useListController`.

I also updated the `useListController` and `List` storybooks to have stories closer to real use-cases (IMO).

Let me know what you think!

crates: Now that every List component can have its own key, can we turn off storing/reading pagination details using query string parameters? Making the query string parameters mandatory for each list view means that you can't functionally have more than one List on a given page and also turn on pagination, without breaking every List component on the page. Filtering has the same problem. Adding column sorting, pagination or filtering to any List component will apply the same settings to every other List on that page.

slax57: @crates  See my comment [here](https://github.com/marmelab/react-admin/issues/8038#issuecomment-1245004316)

crates: @slax57 - Thanks for following up, Jean-Baptiste! That prop was exactly what I was looking for, and solved everything except the part that this upcoming PR will resolve. You guys rock!

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -77,24 +77,27 @@ export interface ListParams {
...
-    perPage = 10,
     debounce = 500,
     disableSyncWithLocation = false,
+    filterDefaultValues,
+    perPage = 10,
+    resource,
+    sort = defaultSort,
+    storeKey,
 }: ListParamsOptions): [Parameters, Modifiers] => {
+    if (!storeKey) storeKey = `${resource}.listParams`;
```
fzaninotto, line 88: This should raise a warning: you're modifying a function argument.

Please create another constant (like `finalStoreKey`) instead. 

```jsx
const finalStoreKey = storeKey ?? `${resource}.listParams`;
```

fzaninotto (changes requested):

fzaninotto (approved):

PR #8080, opened 2022-08-17
Fix <ArrayInput> validation does not prevent submission

### LINKED ISSUE #7878: ArrayInput validation does not prevent submission or update isValid
**What you were expecting:**
Validation passed directly to ArrayInput to work properly as with other inputs on the form.

**What happened instead:**
This is probably related to issue #7591, but that issue points directly to validation for ArrayInput's children, while this issue deals with the validation passed to ArrayInput specifically. Validation does not prevent submission and does not update the form context, such as "isValid".

**Steps to reproduce:**
Pass validation such as required() to an ArrayInput.

**Related code:**
https://codesandbox.io/s/condescending-colden-vic0fn?file=/src/posts/PostCreate.tsx

Navigate to the Post create form, and observe in the console that the Backlinks ArrayInput does not prevent "isValid" from updating to "true" when the other required inputs are filled, whether there is a value for Backlinks or not.

Set variable RETURN_REQUIRED to "false" and observe that my custom validation receives the form values properly, and even updates the helperText if a Backlink is added and removed, but does not return the error to the form to prevent submission.

**Environment**

* React-admin version: 4.1.5
* React version: 17
* Browser: Chrome

arimet:
https://user-images.githubusercontent.com/18294269/185149406-226254d8-1567-4b8b-8f63-4b64a7a863eb.mp4

FIX #7878 

We display the error of the input array only after a first submit to stay consistent with the other inputs
Is this considered a breaking change?


septentrion-730n (approved):

slax57 (changes requested):
I am very pleased with the code you provided :blush: 

HOWEVER

When I test it locally, I do not observe the same behaviour as you do :slightly_frowning_face: 

array-input-validation-issue


Note in this video how I am able to create a Post with no backlinks although they are marked as required.

Could you check again if this happens by your side with the latest code revision ?

@septentrion-730n could you try it as well?

Thanks.

EDIT: I have run `make install` and `make build` prior to running `make run`

slax57 (approved):

septentrion-730n:
Behaviour is fine for me.

djhi (approved):
Finally! That's great!

djhi:
Please rebase on `next` though

PR #8108, opened 2022-08-26
Fix use of context when accessing the resource in FieldTitle component

WiXSL:


slax57 (approved):
Great, thanks!

And welcome back!! :relaxed: 

Tests seem to be failing though, could you check out why?

packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
```diff
@@ -130,7 +130,10 @@ describe('<FilterForm />', () => {
                     <ListContext.Provider value={defaultProps}>
                         <Filter>
                             <ReferenceInput source="post_id" reference="posts">
-                                <SelectInput optionText="title" />
+                                <SelectInput
+                                    optionText="title"
+                                    resource="comments"
+                                />
```
slax57, line 136: Now this is worrying to me. I'm afraid that, if you needed to change this, then this change is actually a BC.

Actually, it seems that `FieldTitle` is using `useTranslateLabel`, which itself uses the `ResourceContext` to have both: the `resource` provided at call-time and the `resourceFromContext`. Depending on these 2 parameters, the `getFieldLabelTranslationArgs` function then chooses to use one of the two. Apparently the call-time resource takes precedence over the resource from context, see https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/util/getFieldLabelTranslationArgs.ts#L40

So, given all that, this actually makes me wonder, what was the case where you figure there was an issue in the first place? :sweat_smile:  This could help figure out if we need to apply a fix somewhere else.

slax57 (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
WiXSL: `So, given all that, this actually makes me wonder, what was the case where you figure there was an issue in the first place? 😅 This could help figure out if we need to apply a fix somewhere else.`

- I was trying to use a `FieldTitle` by itself without passing the resource as prop but letting the component use it from context

PR #8123, opened 2022-08-31
Fix AutocompleteInput `clearOnBlur` has no effect

### LINKED ISSUE #8096: Improper input value selection process for Autocomplete/ReferenceInput combo
**What you were expecting:**
Consistent behavior from AutocompleteInputs as children of ReferenceInputs.

**What happened instead:**
The AutocompleteInput will save incomplete text inputs without having made a dropdown selection. 

The ReferenceInput does not seem to be making the appropriate calls when given the prop "enableGetChoices"

**Steps to reproduce:**
Add an AutocompleteInput with common props such as filterToQuery and shouldRenderSuggestions as a child of a ReferenceInput with common props such as enableGetChoices.

**Related code:**
https://codesandbox.io/s/ecstatic-ptolemy-64jbky?file=/src/posts/PostCreate.tsx

**Environment**

* React-admin version: 4.3.0
* Last version that did not exhibit the issue (if applicable): 4.2.7
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #8096

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -215,11 +215,11 @@ export const AutocompleteInput = <
     });
 
     const finalOnBlur = useCallback((): void => {
-        if (clearOnBlur && !(!!create || !!onCreate)) {
+        if (clearOnBlur) {
             setFilterValue('');
         }
         field.onBlur();
-    }, [clearOnBlur, field, create, onCreate]);
+    }, [clearOnBlur, field]);
 
     const selectedChoice = useSelectedChoice<
         OptionType,
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -219,7 +219,7 @@ export const AutocompleteInput = <
             setFilterValue('');
         }
         field.onBlur();
-    }, [field]);
+    }, [clearOnBlur, field, create, onCreate]);
 
     const selectedChoice = useSelectedChoice<
         OptionType,
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -127,6 +127,7 @@ export const AutocompleteInput = <
     const {
         choices: choicesProp,
         className,
+        clearOnBlur = true,
         clearText = 'ra.action.clear_input_value',
         closeText = 'ra.action.close',
         create,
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -213,6 +214,13 @@ export const AutocompleteInput = <
         ...rest,
     });
 
+    const finalOnBlur = useCallback((): void => {
+        if (clearOnBlur && !(!!create || !!onCreate)) {
+            setFilterValue('');
+        }
```
slax57, line 220: I don't understand, why clear the filter only if you don't have `create` or `onCreate` props?

WiXSL: Is our custom `freeSolo` condition ( `freeSolo={!!create || !!onCreate}`).
At first I was asking `if (clearOnBlur && !props.freeSolo) {`

slax57: Not sure I understand :grimacing: 
According to https://mui.com/material-ui/react-autocomplete/#creatable , we should have `clearOnBlur=true` when `freeSolo=true`.

Since having `clearOnBlur=true` on MUI's `<Autocomplete>` does not seem to work, to me the condition should be `if (clearOnBlur || props.freeSolo)` rather than `if (clearOnBlur && !props.freeSolo)`
Or simply `if (clearOnBlur)` since we encourage `clearOnBlur=true` by default anyway.

Again, not sure I understand your point completely, but I tested your branch with the simple example, and it feels weird to still have something inside the Autocomplete when I click away and did not create a new item.

TLDR; I'd simply go with `if (clearOnBlur)`

slax57 (approved):

Now I just would like to have @fzaninotto 's opinion on this before merging, just to be sure

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -213,6 +214,13 @@ export const AutocompleteInput = <
         ...rest,
     });
 
+    const finalOnBlur = useCallback((): void => {
+        if (clearOnBlur) {
+            setFilterValue('');
```
fzaninotto, line 219: I've tested it with the storybook, and it feels weird. Because if I blur the field without selecting anything, the field appears empty *but the actual value hasn't changed*, so the SaveButton isn't enabled. 

We should set the filter value to the current value in this case.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -233,6 +226,14 @@ export const AutocompleteInput = <
         optionValue,
     });
 
+    const finalOnBlur = useCallback((): void => {
+        if (clearOnBlur) {
+            const optionLabel = getOptionLabel(selectedChoice);
+            setFilterValue(optionLabel);
+        }
+        field.onBlur();
+    }, [clearOnBlur, field, selectedChoice]);
```
fzaninotto, line 235: linter warning

fzaninotto (approved):

PR #8124, opened 2022-08-31
Improve SimpleFormIterator UI

fzaninotto:
- [x] Remove line index by default
- [x] Use the same type of button for all actions
- [x] Show row actions only on hover (except on Mobile)
- [x] Add a `fullWidth` prop to push actions to the row end
- [x] Emphasize the `inline` layout in the documentation
- [x] Fix label position
- [x] Fix layout for nested ArrayInput
- [x] Fix tests
- [x] Add more stories

## Before

Capture d’écran 2022-08-31 à 17 24 31

## After

ezgif com-gif-maker (5)

The docs introduce this component with a nicer screenshot:

array-input


fzaninotto:
The rationale for separating the remove button from the others in the first place was that it's destructive. Maybe we should keep that. Let me think again about the best compromise.

Back to WIP.

slax57:
> The rationale for separating the remove button from the others in the first place was that it's destructive. Maybe we should keep that. Let me think again about the best compromise.
> 
> Back to WIP.

The different colors still make it clear enough, imho

WiXSL:
Woow this is great!

fzaninotto:
OK, I've made up my mind, this is back to RFR!

docs/SimpleFormIterator.md
```diff
@@ -26,6 +26,26 @@ import {
...
+                    <NumberInput source="price" helperText={false} />
+                    <NumberInput source="quantity" helperText={false} />
+                </SimpleFormIterator>
+            </ArrayInput>
+        </SimpleForm>
+    </Edit>
+);
+```
+
+In this example, the inputs for each row appear inline, with no helper text. This dense layout is adapted to arrays with many items. If you need more room, omit the `inline` prop to use the default default layout, where each input is displayed in a separate row.
```
WiXSL, line 46:
```suggestion
In this example, the inputs for each row appear inline, with no helper text. This dense layout is adapted to arrays with many items. If you need more room, omit the `inline` prop to use the default layout, where each input is displayed in a separate row.
```

WiXSL (changes requested):

docs/SimpleFormIterator.md
WiXSL, line 46: Maybe is better `in the example above...` since this line is between two examples

WiXSL (changes requested):

WiXSL (changes requested):
Looking at the `ArrayInput` stories, specially `Nested Inline`, the inputs look to dense.
And the separation line between items gets too close to the inputs

fzaninotto:
> Looking at the ArrayInput stories, specially Nested Inline, the inputs look to dense.
> And the separation line between items gets too close to the inputs

This story shows the "densest" density we can get. I still think it's not dense enough ;)

You can see the default density in the `Basic` story, and I don't think there is a density problem there. On the contrary, I find it barely usable because everything is so far from everything. 

image

As for the separator line, I'm not 100% satisfied with the result either, but the alternatives I tested were worse.

WiXSL (approved):

slax57 (approved):

PR #8125, opened 2022-08-31
export useReferenceParams

RWOverdijk:
I need to disable the getMany done by ReferenceArrayInput.

I couldn't find a way to do that and finally decided to create my own. The challenge with that is that this export is missing.


RWOverdijk:
Ah yes, cool.

PR #8131, opened 2022-09-01
fix: 🐛 return consistent data types from dataprovider on logout

### LINKED ISSUE #8130: useGetManyAggregate throws "filter is not a function" when getting 401
**What you were expecting:**
When getting a 401, useGetManyAggregate should not throw errors 

**What happened instead:**
useGetManyAggregate throws "filter is not a function" when getting 401

**Steps to reproduce:**
Log to a react-admin application where you use useGetManyAggregate ; delete your auth cookie manually and see all the errors in the console

**Related code:**
This is due to useDataProvider that does not return the correct type of data on logout (l.112 of useDataProvider.ts)

**Other information:**
I will provide a PR to solve this in a few minnutes

**Environment**

* React-admin version: 4.3.0

matthieuMay:
Fixes #8130

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
+++ b/packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
@@ -173,4 +173,67 @@ describe('useDataProvider', () => {
 
         expect(customVerb).toHaveBeenCalledWith({ id: 1 }, ['something']);
     });
+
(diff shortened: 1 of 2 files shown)


fzaninotto:
Could you please add a unit test to prove your fix?

fzaninotto (approved):

fzaninotto:
Thanks!

PR #8138, opened 2022-09-05
Fix ensure react-hook-form `formState` update is always enable

WiXSL:


djhi (approved):

fzaninotto:
I see lots of eslint warnings, please fix them. 

Also, Could you add unit tests to prove that it fixes a problem?

Issue #7588, opened 2022-04-25
`useRecordContext` may return `undefined` (v4.0.1)


meteohr:
**What you were expecting:**
`useRecordContext` returns `RaRecord`, so it can be used without undefined check

**What happened instead:**
Error was shown, because record data was undefined

**Steps to reproduce:**
Use typescript and the `useRecordContext` hook. The typing doesn't show the return type can be also `undefined`

**Other information:**
[This PR](https://github.com/marmelab/react-admin/pull/7495) recently fixed this for version 3.19.12 but it seems like this didn't make it into a 4.x release.
[This issue](https://github.com/marmelab/react-admin/issues/7243) is also addressing the issue.

**Environment**

* React-admin version: 4.0.1
* React version: 17.0.2

slax57:
Hi,
Thank your for submitting this!
Indeed we need to port this improvement to v4 as well.
Would you like to make a PR for this?

meteohr:
Hi @slax57 ,

thanks for the quick answer. After having a more closer look into the [source code](https://github.com/marmelab/react-admin/blob/v4.0.1/packages/ra-core/src/controller/record/useRecordContext.ts#L37), it seems like the change from https://github.com/marmelab/react-admin/pull/7495 actually made it into the 4.x version, but since there was no release for ra-core lately, this is not yet reflected in the `ra-core/dist/index.d.ts` file.
So in my opinion there doesn't need to be any more code changes, but rather just a new release (for ra-core), which builds the `ra-core/dist/index.d.ts` again with new type definitions.

slax57:
Okay, so actually this has been ported to v4 already by @WiXSL with #7498 , and it's actually there since the v4.0.0.
What I can't figure out though is why the `index.d.ts` file is not up to date. I'll have to investigate.
Thanks for pointing this out!

martdavidson:
Wish I had time to contribute to the `strict` work - but happy to see this ticket here and an explanation, ran across this issue tonight.

Koeroesi86:
Changing to `const context = useContext<RecordType|undefined>(RecordContext);` could solve issue with useRecordContext in https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/record/useRecordContext.ts

djhi:
Fixed by https://github.com/marmelab/react-admin/pull/7498

gauthierm:
[...]

PR #8141, opened 2022-09-05
Use "strictNullChecks": true in `ra-core`

sixers:
This PR enables `strictNullChecks` mode in `ra-core`, which is required by many useful TypeScript packages (e.g. https://github.com/colinhacks/zod or https://trpc.io/)

The current implementation is just bare minimum to make it compile, and has a handful of ugly type casts. I submitted it to open a a discussion on the best approach to resolve issues - there are a few alternatives.

I added some comments which explain the issues and lists possible solutions.

### DIFF
--- a/packages/ra-core/src/auth/useGetIdentity.ts
+++ b/packages/ra-core/src/auth/useGetIdentity.ts
@@ -5,7 +5,6 @@ import { useSafeSetState } from '../util/hooks';
 
 const defaultIdentity = {
     id: '',
-    fullName: null,
 };
 
 /**
@@ -45,21 +44,20 @@ const useGetIdentity = () => {
     const authProvider = useAuthProvider();
     useEffect(() => {
         if (authProvider && typeof authProvider.getIdentity === 'function') {
-            const callAuthProvider = async () => {
-                try {
-                    const identity = await authProvider.getIdentity();
+            authProvider
+                .getIdentity()
+                .then(identity => {
                     setState({
                         isLoading: false,
                         identity: identity || defaultIdentity,
                     });
-                } catch (error) {
+                })
+                .catch(error => {
                     setState({
                         isLoading: false,
                         error,
                     });
-                }
-            };
-            callAuthProvider();
+                });
         } else {
             setState({
                 isLoading: false,
--- a/packages/ra-core/src/dataProvider/useCreate.ts
+++ b/packages/ra-core/src/dataProvider/useCreate.ts
@@ -91,7 +91,7 @@ export const useCreate = <
             meta: callTimeMeta = paramsRef.current.meta,
         } = {}) =>
             dataProvider
-                .create<RecordType>(callTimeResource, {
+                .create<RecordType>(callTimeResource as string, {
                     data: callTimeData,
                     meta: callTimeMeta,
                 })
@@ -117,8 +117,8 @@ export const useCreate = <
         }
     );
 
-    const create = (
-        callTimeResource: string = resource,
+    const create = async (
+        callTimeResource: string | undefined = resource,
         callTimeParams: Partial<CreateParams<RecordType>> = {},
         createOptions: MutateOptions<
             RecordType,
--- a/packages/ra-core/src/dataProvider/useDelete.ts
+++ b/packages/ra-core/src/dataProvider/useDelete.ts
@@ -15,6 +15,7 @@ import {
     DeleteParams,
     MutationMode,
     GetListResult as OriginalGetListResult,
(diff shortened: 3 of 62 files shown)


packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -45,21 +44,20 @@ const useGetIdentity = () => {
     const authProvider = useAuthProvider();
     useEffect(() => {
         if (authProvider && typeof authProvider.getIdentity === 'function') {
-            const callAuthProvider = async () => {
-                try {
-                    const identity = await authProvider.getIdentity();
```
sixers, line 50: This complained about authProvider.getIdentity possibly being undefined. There is a check, but it's in outer scope, and this is async function which can be executed later, so this check doesn't apply to the inner function. Therefore, I changed it to call authProvider.getIdentity() immediately in the current scope and simply use .then() / .catch() to store the results.

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -91,7 +91,7 @@ export const useCreate = <
             meta: callTimeMeta = paramsRef.current.meta,
         } = {}) =>
             dataProvider
-                .create<RecordType>(callTimeResource, {
+                .create<RecordType>(callTimeResource as string, {
```
sixers, line 94: This is a bit of a hack. I assume the intention is that you either have to provide `resource` in `useCreate()` invocation, or you need to pass it to returned mutation function:

1)
```javascript
const [create] = useCreate('product')
const doCreateProduct = useCallback((data) => create({data}));
```

2)
```javascript
const [create] = useCreate();
const doCreateProduct = useCallback((data) => create({resource: 'product', data}));
```

Only these 2 cases should be allowed. If that's the intention, then I see 2 options:
1) Make `UseCreateResult` result conditional based on the provided `resource`, so that it's required param in the returned mutation function if it's `undefined`, or optional if it's not `undefined`
2) Just do a runtime check and don't call `dataProvider` if `callTimeResource` is undefined.

WDYT?

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -162,8 +162,8 @@ export const useDelete = <
             meta: callTimeMeta = paramsRef.current.meta,
         } = {}) =>
             dataProvider
-                .delete<RecordType>(callTimeResource, {
-                    id: callTimeId,
+                .delete<RecordType>(callTimeResource as string, {
+                    id: callTimeId as Identifier,
```
sixers, line 166: Similar to the hack above, and we have similar options: 1) `callTimeId` should only be optional if `params.id` is not undefined, and required otherwise 2) just do a runtime check

slax57:
This is really a massive work you started here!
This was not in our top priorities at the moment, so I don't know when we will have time to review and discuss about your PR, but it will certainly help that you started to tackle the work this far, and I definitely appreciate the initiative!
Thanks!

fzaninotto:
I can foresee this will be very hard to review - I've seen nasty impacts with that type of changes in the past. I'd prefer if you split it into several smaller PRs.

sixers:
> I can foresee this will be very hard to review - I've seen nasty impacts with that type of changes in the past. I'd prefer if you split it into several smaller PRs.

I understand that. This PR is not ready to be fully reviewed and merged, it just highlights all the places that have to be updated to support `strictNullChecks`. And it also surfaces key decisions that have to be made to fully support `strictNullChecks`. I believe there is value in looking at all the changes in aggregate, to make sure the API is consistent. It would be great if you could do a high-level review and if we could agree on the general direction, and then I can split the final set of changes into a few smaller PRs.

If it makes things easier, I identified the following main "buckets" for changes.

- A. Update default values of  custom `Context`s (e.g. `ListContext`)
- B. Update types of async query hooks which return data (e.g. `useListContext`)
- C. Update types of async mutation hooks, which expects certain parameters to be passed either to a hook itself, or returned mutation function (e.g. `useCreate`)
- D. Bugs, incorrect implicit types, etc. (e.g. in `useGetIdentity`)

I added a comment for each of the buckets, but I'm also happy to describe each bucket / issue  in detail.

If you insist on splitting this work upfront, I can do that as well.

> I've seen nasty impacts with that type of changes in the past.

Could you give me an example? It would be helpful for me to get some context and make sure this change doesn't cause issues.

WiXSL:
Related to #7588

djhi:
Thanks for the PR but as @fzaninotto said, we'll work on it in smaller chunks

PR #8150, opened 2022-09-07
Fix FileInput helperText display when validation prop was set

### LINKED ISSUE #8126: ImageInput and FileInput required() is not working
**What you were expecting:**

When ImageInput has `validate={[required()]}` it should show the message in **red**
When FileInput has `validate={[required()]}` it should show the message

**What happened instead:**

ImageInput shows the message in gray (similar as helperText)
FileInput doesn't show the message

**Steps to reproduce:**

- Open my fork: https://codesandbox.io/s/friendly-allen-7txqn2?file=/src/posts/PostCreate.tsx:2833-3028
- Navigate to Post Create page
- Fill other field like Title and press any of the save buttons

**Related code:**

Add validate={[required()]} to the FileInput tag
Add 
```js
        <ImageInput
          source="imagefile"
          label="Image"
          validate={[required()]}
          accept="image/*"
        >
          <ImageField source="src" />
        </ImageInput>
```

**Other information:**
image

**Environment**

* React-admin version: 4.3.0
* React version: 17.0.0
* Browser: Chrome 100+

WiXSL:
Fixes #8126

### DIFF
--- a/packages/ra-ui-materialui/src/input/FileInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/FileInput.stories.tsx
@@ -7,6 +7,7 @@ import { Create } from '../detail';
 import { SimpleForm } from '../form';
 import { FileInput } from './FileInput';
 import { FileField } from '../field';
+import { required } from 'ra-core';
 
 export default { title: 'ra-ui-materialui/input/FileInput' };
 
@@ -21,6 +22,14 @@ export const Basic = () => (
 export const LimitByFileType = () => (
     <Wrapper>
         <FileInput source="attachment" accept="application/pdf">
+            <FileField source="src" title="title" />
+        </FileInput>
+    </Wrapper>
+);
+
+export const Required = () => (
+    <Wrapper>
+        <FileInput source="attachment" isRequired validate={required()}>
             <FileField source="src" title="title" />
         </FileInput>
     </Wrapper>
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/FileInput.stories.tsx
```diff
@@ -60,7 +69,7 @@ export const FullWidth = () => (
 
 export const Disabled = () => (
     <Wrapper>
-        <FileInput source="attachment" disabled>
+        <FileInput source="attachment" inputProps={{ disabled: true }}>
```
slax57, line 72: I don't see the input disabled on my end. Does this work for you?

I can suggest 
```suggestion
        <FileInput source="attachment" options={{ disabled: true }}>
```
which actually works but really does not provide enough visual feedback IMO (but this is beyond the scope of this PR anyway)

packages/ra-ui-materialui/src/input/ImageInput.stories.tsx
```diff
@@ -60,7 +61,15 @@ export const FullWidth = () => (
 
 export const Disabled = () => (
     <Wrapper>
-        <ImageInput source="attachment" disabled>
+        <ImageInput source="attachment" inputProps={{ disabled: true }}>
```
slax57, line 64: Same here

slax57 (changes requested):

slax57 (approved):

PR #8152, opened 2022-09-08
add "expand_all" toggle to datagrid header

hiaselhans:
this adds a "collapse all" Button in the DataGridHeader

Behavior:

- if one or more rows are expanded, the CollapseButton shows an opened state.
  - onClick collapses all shown rows
- if all rows are collapsed onClick all rows are expanded

### DIFF
--- a/packages/ra-core/src/controller/list/useExpanded.tsx
+++ b/packages/ra-core/src/controller/list/useExpanded.tsx
@@ -49,3 +49,24 @@ export const useExpanded = (
 
     return [expanded, toggleExpanded];
 };
+
+export const useExpandedMultiple = (
+    resource: string,
+    ids: Identifier[]
+): [boolean, () => void] => {
+    const [expandedIds, setExpandedIds] = useStore<Identifier[]>(
+        `${resource}.datagrid.expanded`,
+        []
+    );
+
+    const expanded = expandedIds.some(id => ids.some(id2 => id2 == id));
+
+    const setExpanded = useCallback(() => {
+        const filtered = expandedIds.filter(
+            expanded_id => !ids.some(id => id == expanded_id)
+        );
+        setExpandedIds(expanded ? filtered : filtered.concat(ids));
+    }, [expandedIds, setExpandedIds, expanded, ids]);
+
+    return [expanded, setExpanded];
+};
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridHeader.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridHeader.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Children, isValidElement, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import {
+    useExpandedMultiple,
     useListContext,
     useResourceContext,
     Identifier,
@@ -12,6 +13,7 @@ import {
 import { Checkbox, TableCell, TableHead, TableRow } from '@mui/material';
 import clsx from 'clsx';
 
+import ExpandRowButton from './ExpandRowButton';
 import DatagridHeaderCell from './DatagridHeaderCell';
 import { DatagridClasses } from './useDatagridStyles';
 
@@ -51,6 +53,11 @@ export const DatagridHeader = (props: DatagridHeaderProps) => {
     );
 
     const updateSort = setSort ? updateSortCallback : null;
+
+    const [expanded, toggleExpanded] = useExpandedMultiple(
+        resource,
+        data.map(record => record.id)
+    );
 
     const handleSelectAll = useCallback(
         event =>
@@ -93,7 +100,15 @@ export const DatagridHeader = (props: DatagridHeaderProps) => {
                             DatagridClasses.headerCell,
                             DatagridClasses.expandHeader
                         )}
-                    />
+                    >
+                        <ExpandRowButton
+                            className={clsx(DatagridClasses.expandIcon, {
+                                [DatagridClasses.expanded]: expanded,
+                            })}
+                            expanded={expanded}
+                            onClick={toggleExpanded}
+                        />
+                    </TableCell>
                 )}
                 {hasBulkActions && selectedIds && (
                     <TableCell


hiaselhans:
demo: [storybook](https://react-admin-storybook-7ntlwg2rl-marmelab.vercel.app/?path=/story/ra-ui-materialui-list-datagrid--expand)

fzaninotto (commented):
Awesome! Thanks for that.

Would you mind adding a unit test based on the storybook?

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,24 @@ export const useExpanded = (
...
+export const useExpandedMultiple = (
+    resource: string,
+    ids: Identifier[]
+): [boolean, () => void] => {
+    const [expandedIds, setExpandedIds] = useStore<Identifier[]>(
+        `${resource}.datagrid.expanded`,
+        []
+    );
+
+    const expanded = expandedIds.some(id => ids.some(id2 => id2 == id));
```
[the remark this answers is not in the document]
hiaselhans: true, thanks!

hiaselhans:
thanks @fzaninotto !

I added the tests to cypress.
Hope that works, not sure how to add tests to the storybook...

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,38 @@ export const useExpanded = (
...
```
fzaninotto, line 76: linter warning

fzaninotto: Never trust the type of a stored value - users can mess up with their localStorage, and we don't want to break the app because of that. See line 29 for an example. 

https://marmelab.com/react-admin/Store.html#forward-compatibility

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,38 @@ export const useExpanded = (
...
+
+    const setExpanded = useCallback(() => {
+        const filtered = expandedIds.filter(
```
fzaninotto, line 79: what's `filtered`? As the code is hard to understand (filter with a not some??), please choose a better variable name, like "notYetExpandedIds"

fzaninotto:
For the tests, please write a react-testing-library test rather than an e2e test. You can leverage the Storybook as we did e.g. in https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/field/ReferenceOneField.spec.tsx

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: i chose unaffectedIds now, because it is the ids not on the current page whose state shouldn't be touched.

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: done.

hiaselhans:
ok, so i added [expandAllButton.spec.tsx](https://github.com/marmelab/react-admin/pull/8152/files#diff-8c302c80afdab5eb09d5f3dfab707695fe91c1a61314d0d067f559181b56d948) with best intentions. Hope that works now.
I would still not remove the e2e test, maybe it proves useful at some point

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+import IconButton, { IconButtonProps } from '@mui/material/IconButton';
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import { useTranslate, useExpandedMultiple } from 'ra-core';
+import { DatagridClasses } from './useDatagridStyles';
+import clsx from 'clsx';
+
+interface ExpandMultipleButtonProps {
+    resource: string;
+    ids: string[] | number[];
+}
```
slax57, line 12:
```suggestion
interface ExpandAllButtonProps {
    resource: string;
    ids: Identifier[];
}
```

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <ExpandMoreIcon fontSize="inherit" />
+        </IconButton>
+    );
+};
+
+export interface ExpandRowButtonProps extends IconButtonProps {
+    component?: ElementType;
+    expanded: boolean;
+    expandContentId?: string;
+}
```
slax57, line 42: You should remove this

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,42 @@ export const useExpanded = (
 
     return [expanded, toggleExpanded];
 };
+
+/**
+ * State-like hook for controlling the expanded state of many list items
+ * expanded state is true when at least one item from ids is expanded.
+ *
+ * @param {string} resource The resource name, e.g. 'posts'
+ * @param {string[]|integer[]} ids A list of record identifiers
```
slax57, line 58:
```suggestion
 * @param {Identifier[]} ids A list of record identifiers
```

hiaselhans (commented):
done! Thanks

slax57 (approved):
Thanks!
I'll give some time to @fzaninotto if he wants to re-review it, then we'll be able to merge :slightly_smiling_face:

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,42 @@ export const useExpanded = (
...
+ * @param {Identifier[]} ids A list of record identifiers
+ * @returns {Object} Destructure as [expanded, toggleExpanded].
+ *
+ * @example
+ *
+ * const [expanded, toggleExpanded] = useExpandedMultiple('posts', [123, 124, 125]);
+ * const expandIcon = expanded ? ExpandLess : ExpandMore;
+ * const onExpandClick = () => toggleExpanded();
+ */
+export const useExpandedMultiple = (
```
fzaninotto, line 67: I'd call it `useExpandAll` for better readability

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,42 @@ export const useExpanded = (
...
+    resource: string,
+    ids: Identifier[]
+): [boolean, () => void] => {
+    const [expandedIds, setExpandedIds] = useStore<Identifier[]>(
+        `${resource}.datagrid.expanded`,
+        []
+    );
+
+    const expanded = Array.isArray(expandedIds)
```
fzaninotto, line 76: This variable name is still not explicit enough IMO. We already have an `expandedIds`, so this `expanded` is confusing. 

Why not rename `expandedIds` to `allExpandedId`, and `expanded` into `displayedExpandedIds`, or something like that?

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,42 @@ export const useExpanded = (
...
+        ? // eslint-disable-next-line eqeqeq
+          expandedIds.some(id => ids.some(id2 => id2 == id))
+        : false;
+
+    const toggleExpanded = useCallback(() => {
```
fzaninotto, line 81: rename to `toggleExpandAll` for clarity

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,42 @@ export const useExpanded = (
...
+        const unaffectedIds = expandedIds.filter(
```
fzaninotto, line 82: these are the `notDisplayedExpandedIds`, right?

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+            className={clsx(DatagridClasses.expandIcon, {
+                [DatagridClasses.expanded]: expanded,
+            })}
+            aria-label={translate(
+                expanded ? 'ra.action.close' : 'ra.action.expand'
+            )}
+            aria-expanded={expanded}
+            tabIndex={-1}
+            aria-hidden="true"
+            component="div"
```
fzaninotto, line 29: Why do you need it to be a div?

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+    return (
+        <IconButton
+            className={clsx(DatagridClasses.expandIcon, {
+                [DatagridClasses.expanded]: expanded,
+            })}
+            aria-label={translate(
+                expanded ? 'ra.action.close' : 'ra.action.expand'
+            )}
+            aria-expanded={expanded}
+            tabIndex={-1}
```
fzaninotto, line 27: Why this tabindex?

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+
+import { Expand } from './Datagrid.stories';
+
+describe('ExpandAllButton', () => {
+    it('should expand all rows at once', async () => {
+        const rendered = render(<Expand />);
+        const button = rendered.container.querySelector(
```
fzaninotto, line 9: instead of `container`, we tend to use aria labels or data-testid to locate elements without text - just like a screen reader would do. See https://testing-library.com/docs/guiding-principles

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+            '.RaDatagrid-expandHeader>div'
+        );
+        const expectExpandedRows = (count: number) => {
+            expect(
+                rendered.container.querySelectorAll('.RaDatagrid-expandedPanel')
```
fzaninotto, line 14: Same

fzaninotto (changes requested):

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: this variable name is literally as explicit as in useExpanded.
I also don't see a reason to rename `expandedIds` just because the context is different.

Example:

``` typescript
function square(x: number) {
    const exponent = 2;
    return Math.pow(x, exponent);
}

function squareAll(x: number[]) {
    const exponent = 2;
    return x.map(value => Math.pow(value, exponent));
}
```

would you rename exponent to exponentAll?

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
hiaselhans: this is just a plain copy from ExpandRowButton.
You want it gone?

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
hiaselhans: dito

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: as a compromise, i'd say `unaffectedExpandedIds`, since in this context we don't have displayed or hidden ids :imp:

packages/ra-core/src/controller/list/useExpanded.tsx
fzaninotto: You named this variable `expanded`, but it's *a filtered list of expanded ids*. The name you chose is misleading and not precise enough.

This may look like nitpicking, but we'll have to maintain this code for years. We have to be confident we understand it well.

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
fzaninotto: if it's not necessary, yes. Don't just blindly copy code without making sure you need it.

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { render, fireEvent, screen } from '@testing-library/react';
```
fzaninotto, line 2: linter warning

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { Expand } from './Datagrid.stories';
+
+describe('ExpandAllButton', () => {
+    it('should expand all rows at once', async () => {
+        const rendered = render(<Expand />);
+        const expand = () => {
+            const button = rendered.getAllByLabelText('ra.action.expand')[0];
```
fzaninotto, line 10: We use the `screen.getAllByLabelText` syntax in most of the codebase - that's the way react-testing-library advises to do tests

packages/ra-core/src/controller/list/useExpanded.tsx
fzaninotto: Sorry, I re-read your code and it seems this variable contains a boolean, not a selection of ids. To make it explicit, please name it `isExpanded`, or better, `hasSomeExpanded`

packages/ra-core/src/controller/list/useExpanded.tsx
```diff
@@ -49,3 +49,44 @@ export const useExpanded = (
...
+        []
+    );
+
+    const expanded = Array.isArray(expandedIds)
+        ? // eslint-disable-next-line eqeqeq
+          expandedIds.some(id => ids.some(id2 => id2 == id))
+        : false;
+
+    const toggleExpandedAll = useCallback(() => {
+        const unaffectedExpandedIds = expandedIds.filter(
```
fzaninotto, line 82: I don't understand this "unaffected".

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
hiaselhans: thanks

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
hiaselhans: since ExpandRowButtons are not tab-selectable, the least i would do here is to stay consistent and keep tabindex = -1.

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: Don't get me wrong. It's your project so you decide.

I just wasn't very happy with your suggestion and since we mimic a similar behavior as in useExpanded, my personal take is to use a similar naming convention.
I went for isExpanded now, since hasSomeExpanded gives more confusion in the toggleExpandedAll function

packages/ra-ui-materialui/src/list/datagrid/expandAllButton.spec.tsx
hiaselhans: I saw both usages in the codebase but most of the tests used `const { getAllByLabelText } = render(...)`.
I changed it in this PR.

packages/ra-core/src/controller/list/useExpanded.tsx
hiaselhans: possible synonyms:

- unChanged
- unAltered
- unTouched

fzaninotto (approved):

fzaninotto:
Thanks!

hiaselhans:
thanks! and keep up the good work! :)

Issue #8153, opened 2022-09-08
ArrayInput does not show validation errors if using schema validation (with yup)


tommasogangemi:
**Expected result**

The `ArrayInput` shows the related error message when using a resolver for schema validation, as it happens when using the `validate` prop.

**Actual result**

When submitting the form the `ArrayInput`'s label turns red, and there is no message shown where it should be (in the place of the helper text).

**Steps to reproduce:**

Open up [this](https://stackblitz.com/edit/github-mccsfc?file=src%2Ftags%2FTagCreate.tsx,src%2Fcomments%2FCommentCreate.tsx) stackblitz project, forked from your [simple example](https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple). 

The `TagCreate` component has a form with an array input that is validated via the `validate` prop, and correctly shows the error when trying to submit.

The `CommentCreate` component has a form with an array input that is validated by a resolver that uses a yup schema, as suggested by [your docs](https://marmelab.com/react-admin/Validation.html#schema-validation). Thevalidation works as expected, but the error message does not correctly pop up under the input, that only shows a red label.

Both the forms extract the error from the field the same way as it's done in the source code of [ArrayInput](https://github.com/marmelab/react-admin/blob/6503cb3c6aa50f5048edf7bdcb5ee22fc3ce6c10/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx#L130), and the error is logged in the console. Here we can observe that the `error` retrieved has a different shape in the 2 cases.

**Related code:**

In [this example](https://stackblitz.com/edit/github-mccsfc-uuuvrh?file=src%2Ftags%2FTagCreate.tsx,src%2Fcomments%2FCommentCreate.tsx) I try to validate the array with a yup native array schema method and a standard `validate` prop

In [this example](https://stackblitz.com/edit/github-mccsfc?file=src%2Ftags%2FTagCreate.tsx,src%2Fcomments%2FCommentCreate.tsx) I try to validate the array input based on a custom validation function mirrored in the native example and the yup example

Both of them result in the same exact misbehaviour

**Other information:**

I have recently upgraded the react-admin version from version 4.2.6 to version 4.3.1. The problem was not present in version 4.2.6 and the error message used to show up as expected

**Environment**

* React-admin version: 4.3.1 (react-hook-form 7.34.2)
[...]

PR #8154, opened 2022-09-08
Fix ArrayInput helperText not showing error message

WiXSL:
Fixes https://github.com/marmelab/react-admin/issues/8153


slax57 (approved):
Wow, great find!! :smiley: 
I believe this aims at fixing https://github.com/marmelab/react-admin/issues/8153 ? (if so, we should mention it in the PR description)

One last thing still puzzles me: how could this work in RA v4.2.6? :thinking:

slax57:
Okay I get it now, this regression was caused by https://github.com/marmelab/react-admin/pull/8080/files#diff-d599f69bbbccae699c58ac0c3ab76dde15cb75ae9280b8ecc4d206f650a0fc99R175

So basically your PR is the right way to fix it IMO. Again, well done :slightly_smiling_face:

PR #8159, opened 2022-09-13
feat: add prevent validateResponseFormat() from validating custom routes

### LINKED ISSUE #8083: validateResponseFormat triggers error when calling a dataProvider custom method
If you implement a data provider with a custom method, then call that method as documented [here](https://marmelab.com/react-admin/Actions.html#calling-custom-methods) then `validateResponseFormat()` throws an error if the returned object doesn't contain a 'data' key

`The response to 'action' must be like { data: ... }, but the received response does not have a 'data' key. The dataProvider is probably wrong for 'action'.`

I think if it's a custom method, there should not be any validation of the returned object (how could react-admin know?) 

Otherwise, maybe the type of `DataProvider` should be typed better than `[key: string]: any`

oguhpereira:
This pr makes an adjustment so that validateResponseFormat() is not used in dataProvider custom routes

- Code ✔️ 
- Test ❌ 
- Documentation ❌ 

---
Closes #8083

Obs: I need to understand what is needed to document and test

### DIFF
--- a/packages/ra-core/src/dataProvider/dataFetchActions.ts
+++ b/packages/ra-core/src/dataProvider/dataFetchActions.ts
@@ -21,6 +21,11 @@ export const fetchActionsWithArrayOfRecordsResponse = [
 ];
 export const fetchActionsWithTotalResponse = ['getList', 'getManyReference'];
 
+export const fetchActions = [
+    ...fetchActionsWithRecordResponse,
+    ...fetchActionsWithArrayOfRecordsResponse,
+];
+
 export const sanitizeFetchType = (fetchType: string) => {
     switch (fetchType) {
         case GET_LIST:
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -5,6 +5,7 @@ import { defaultDataProvider } from './defaultDataProvider';
 import validateResponseFormat from './validateResponseFormat';
 import { DataProvider } from '../types';
 import useLogoutIfAccessDenied from '../auth/useLogoutIfAccessDenied';
+import { fetchActions } from './dataFetchActions';
 
 /**
  * Hook for getting a dataProvider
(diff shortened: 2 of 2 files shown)


packages/ra-core/src/dataProvider/dataFetchActions.ts
```diff
@@ -21,6 +21,11 @@ export const fetchActionsWithArrayOfRecordsResponse = [
 ];
 export const fetchActionsWithTotalResponse = ['getList', 'getManyReference'];
 
+export const fetchActions = [
```
fzaninotto, line 24: I'd name it `reactAdminFetchActions`

fzaninotto (changes requested):
The existing response validation logic is currently not tested, but I'd like that you add at least one test for this change. I'd put it in `useDataProvider.spec.tsx`, and I'd test 3 cases:

1. a call to `dataProvider.getList()` returning a response with the right format doesn't return any warning
2. a call to `dataProvider.getList()` returning a response without the right format does return any warning
3. a call to `dataProvider.custom()` returning a response without the right format doesn't return any warning

No need to document anything.

packages/ra-core/src/dataProvider/dataFetchActions.ts
oguhpereira: Done https://github.com/marmelab/react-admin/pull/8159/commits/448a8eca6631632713b4f14513c951f5af0d4b97

oguhpereira:
I updated the unit tests but when running the console log is showing in the test log

image

slax57:
@oguhpereira you can use something like this to discard `console.error()` during your test:

```ts
jest.spyOn(console, 'error').mockImplementationOnce(() => {});
```

oguhpereira:
> @oguhpereira you can use something like this to discard `console.error()` during your test:
> 
> ```ts
> jest.spyOn(console, 'error').mockImplementationOnce(() => {});
> ```

Done https://github.com/marmelab/react-admin/pull/8159/commits/c0bc6bb10130f9dc93a5efa2cdfad072d54d198a

oguhpereira:
> @oguhpereira you can use something like this to discard `console.error()` during your test:
> 
> ```ts
> jest.spyOn(console, 'error').mockImplementationOnce(() => {});
> ```

Done https://github.com/marmelab/react-admin/pull/8159/commits/c0bc6bb10130f9dc93a5efa2cdfad072d54d198a @slax57

fzaninotto (approved):

fzaninotto:
Thanks!

PR #8161, opened 2022-09-14
Fix localStorageStore.reset() does not remove all items

fzaninotto:
I have a unit test for it in the configurable branch - I haven't added ift here to avoi!d conflicts

### DIFF
--- a/packages/ra-core/src/store/localStorageStore.ts
+++ b/packages/ra-core/src/store/localStorageStore.ts
@@ -117,9 +117,11 @@ export const localStorageStore = (
         reset(): void {
             const storage = getStorage();
             Object.keys(storage).forEach(key => {
-                storage.removeItem(key);
-                const publishKey = key.substring(prefixLength + 1);
-                publish(publishKey, undefined);
+                if (key.startsWith(prefix)) {
+                    storage.removeItem(key);
+                    const publishKey = key.substring(prefixLength + 1);
+                    publish(publishKey, undefined);
+                }
             });
         },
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/store/localStorageStore.ts
```diff
@@ -116,15 +116,11 @@ export const localStorageStore = (
         },
         reset(): void {
             const storage = getStorage();
-            for (let i = 0; i < storage.length; i++) {
-                if (storage.key(i)?.substring(0, prefixLength) === prefix) {
```
fzaninotto, line 120: The problem was that once you remove an item, the index of the next item decreases... So `storage.key(i)` 'jumps' one item each time we delete one.

packages/ra-core/src/store/localStorageStore.ts
```diff
@@ -116,15 +116,11 @@ export const localStorageStore = (
...
-                    // @ts-ignore
-                    storage.removeItem(storage.key(i));
-                    publish(key, undefined);
-                }
-            }
+            Object.keys(storage).forEach(key => {
+                storage.removeItem(key);
+                const publishKey = key.substring(prefixLength + 1);
+                publish(publishKey, undefined);
+            });
```
slax57, line 123: I agree with the analysis and the fix
But why did you also remove the `if (storage.key(i)?.substring(0, prefixLength) === prefix)` part? This should be kept IMO

slax57 (changes requested):

slax57 (approved):

PR #8167, opened 2022-09-15
fix parse  doc

Cimanel:
- [x] remove `typeof v === 'string' && v.length === 0 ` because v is always a string


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8169, opened 2022-09-15
Fix layout prop of Admin component in upgrade example

herndlm:
The upgrade example incorrectly uses an element compared to e.g. https://marmelab.com/react-admin/Theming.html#using-a-custom-logout-button


WiXSL (approved):

WiXSL:
Thanks again!

PR #8170, opened 2022-09-15
Add missing props in docs layout examples

herndlm:
Brings the examples in sync with the rest, e.g. https://marmelab.com/react-admin/Layout.html#appbar
This also seems to be needed to avoid rendering errors about components not rendering anything / missing returns


WiXSL (approved):

WiXSL:
Thanks!

PR #8183, opened 2022-09-21
Fix ReferenceInput fetching error makes AutocompleteInput unusable

fzaninotto:


WiXSL (changes requested):
Shouldn't the `SelectArrayInput` have the same treatment as `AutocompleteInput` and `SelectInput` ?

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -247,7 +253,9 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                     autoWidth
                     labelId={`${label}-outlined-label`}
                     multiple
-                    error={(isTouched || isSubmitted) && invalid}
+                    error={
+                        fetchError || ((isTouched || isSubmitted) && invalid)
```
WiXSL, line 257:
```suggestion
                        fetchError || ((isTouched || isSubmitted) && invalid)
                        !!fetchError || ((isTouched || isSubmitted) && invalid)
```

WiXSL: I'm just suggesting based on the `SelectInput` change

WiXSL (changes requested):

WiXSL (changes requested):
lint in `ReferenceArrayInput.spec.tsx`

packages/ra-ui-materialui/src/input/ReferenceArrayInput.stories.tsx
```diff
@@ -29,8 +32,136 @@ const dataProvider = testDataProvider({
...
+                resource="posts"
+                source="tag_ids"
+            >
+                <SelectArrayInput optionText="name" />
+            </ReferenceArrayInput>
+        </Form>
+    </AdminContext>
+);
+
+export const ErrorSelect = () => (
```
slax57, line 84:
```suggestion
export const ErrorSelectArray = () => (
```

PR #8162, opened 2022-09-14
Implement `emptyText` and `emptyValue` in `AutocompleteInput`

WiXSL:
Fixes #8091

- [X] Implement `emptyText` usage
- [x] Add tests
- [X] Add story
- [X] Add documentation
- [X] Add back `emptyValue` with no support for `null` or `undefined`
- [X] Fix `SelectInput`'s `emptyValue` docs

PR #8188, opened 2022-09-22
Add automated empty values sanitization

fzaninotto:
## Problem

All form inputs have an empty value of `''` (empty string). React-hook-form doesn't sanitize these values by default. This leads to unexpected `create` and `update` payloads like:

```
{
  id: 123,
  title: '',
  author: '',
}
```

The currently documented solution is to [use `parse` on individual inputs](https://marmelab.com/react-admin/Inputs.html#transforming-input-value-tofrom-record), and/or [`transform` at the controller level](https://marmelab.com/react-admin/SimpleForm.html#cleaning-up-empty-strings).

But none of these solutions is particularly convenient when developers have to use them everywhere. The input-level solution is super repetitive, and the controller-one is at the wrong level - the problem is caused by the form, it should be handled by the form. 

## Solution

Introduce an opt-in `<Form sanitizeEmtyValues>` prop

```const PostEdit = () => (
  <Edit>
    <SimpleForm sanitizeEmptyValues>
      <TextInput source="title" />
    </SimpleForm>
  </Edit>
);
```

## Todo

- [x] Create an empty value sanitizer and test it
- [x] Add the support for `sanitizeEmptyValues in `<Form>`, `<SimpleForm>`, and `<TabbedForm>``
- [x] Document the new prop
- [x] Remove advice to use `transform` to sanitize empty values 
- [x] In Reference inputs, sanitize empty values by default


djhi (approved):

docs/Form.md
```diff
@@ -47,6 +47,7 @@ Here are all the props you can set on the `<Form>` component:
 * [`id`](#id)
 * [`noValidate`](#novalidate)
 * [`onSubmit`](#onsubmit)
+* [`sanitizeEmptyValues`](#sanitizeemptyvalues)
```
slax57, line 50: Upgrade guide still says `sanitizeEmptyValues` prop has been removed -> we should update that

docs/ReferenceInput.md
```diff
@@ -116,6 +142,32 @@ const filters = [
...
+
+If you want to change this behavior, you have to pass a custom `parse` prop to the `<ReferenceInput>` *child component*, because  **`<ReferenceInput>` doesn't have a `parse` prop**. It is the responsibility of the child component to parse the input value.
+
+For instance, if you want to transform an option value before submission, and the selection control is an `<AutocompleteInput>` (the default), set [the `<AutocompleteInput parse>` prop](./Inputs.md#parse) as follows:
+
+```jsx
+import { ReferenceInput, AutocompleteInput } from 'react-admin';
+
+<ReferenceInput source="post_id" reference="posts">
+    <AutocompleteInput parse={value => value === 'not defined' ? undefined : null} />
```
slax57, line 157: Same here. I would have expected rather

```suggestion
    <AutocompleteInput parse={value => value === 'not defined' ? null : value} />
```

slax57 (changes requested):

packages/ra-core/src/form/useAugmentedForm.ts
```diff
@@ -82,6 +91,8 @@ export const useAugmentedForm = (props: UseAugmentedFormProps) => {
         shouldUseNativeValidation,
     });
 
+    const formRef = useRef(form);
```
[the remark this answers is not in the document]
fzaninotto: To avoid memoization issues in useCallback

docs/Form.md
fzaninotto: Right, fixed

docs/ReferenceInput.md
```diff
@@ -116,6 +142,32 @@ const filters = [
 ];
 ```
 
+## `parse`
+
+By default, children of `<ReferenceInput>` transform the empty form value (an empty string) into `null` before passing ot to the `dataProvider`.
```
WiXSL, line 147:
```suggestion
By default, children of `<ReferenceInput>` transform the empty form value (an empty string) into `null` before passing it to the `dataProvider`. 
```

WiXSL (changes requested):

WiXSL (changes requested):
Can you rebase to include #8162?
We may have to change the note in the docs of `AutocompleteInput` and `SelectInput` regarding `emptyValue` prop being null

fzaninotto:
Back to WIP as this makes #7782 more pregnant

fzaninotto:
Back to RFR - we'll fix #7782 in a bug fix release, no need to delay the next minor.

docs/AutocompleteInput.md
```diff
@@ -96,10 +96,43 @@ When used inside a `<ReferenceInput>`, `<AutocompleteInput>` doesn't need a `cho
...
+        { id: 456, name: 'Jane Austen' },
+    ]}
+    validate={required()}
+/>
+```
 
 ## `emptyValue`
 
-An empty choice is always added (with a default `''` value, which you can override with the `emptyValue` prop) on top of the options. You can furthermore customize the empty choice by using the `emptyText` prop, which can receive a string or a React Element.
+If the input isn't required (using `validate={required()}`), users can select an empty choice. The default value for that empty choice is tyhe empty string (`''`), or `null` if the input is inside a [`<ReferenceInput>`](./ReferenceInput.md).
```
slax57, line 133:
```suggestion
If the input isn't required (using `validate={required()}`), users can select an empty choice. The default value for that empty choice is the empty string (`''`), or `null` if the input is inside a [`<ReferenceInput>`](./ReferenceInput.md).
```

docs/SelectInput.md
```diff
@@ -106,19 +106,56 @@ const choices = [
...
+        { id: 'photography', name: 'Photography' },
+    ]}
+    validate={required()}
+/>
 ```
 
-**Note**: `emptyValue` cannot be set to `undefined` or `null` since the `dataProvider` method will receive an empty string on submit due to the nature of HTML inputs.
+## `emptyValue`
+
+If the input isn't required (using `validate={required()}`), users can select an empty choice. The default value for that empty choice is tyhe empty string (`''`), or `null` if the input is inside a [`<ReferenceInput>`](./ReferenceInput.md).
```
slax57, line 145:
```suggestion
If the input isn't required (using `validate={required()}`), users can select an empty choice. The default value for that empty choice is the empty string (`''`), or `null` if the input is inside a [`<ReferenceInput>`](./ReferenceInput.md).
```

slax57 (changes requested):

slax57 (approved):

PR #8189, opened 2022-09-22
[Doc] Fix FormDataConsumer usage example

### LINKED ISSUE #8179: Unmounting inputs doesn't reset/clear them.
**What you were expecting:**

Expect field inputs to be unmounted and the input value to be reset/clear with this code:
```
<FormDataConsumer>
    {({ formData, ...rest }) => formData.hasEmail &&
        <TextInput source="email" {...rest} />
    }
</FormDataConsumer>
```

**What happened instead:**
Form values are persisted and submitted with the form.

**Steps to reproduce:**
You can copy paste the form from the [official documentation](https://marmelab.com/react-admin/Inputs.html#hiding-inputs-based-on-other-inputs) and test it.

React admin v4.3.1

WiXSL:
Fixes #8179

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -522,7 +522,7 @@ import { FormDataConsumer } from 'react-admin';
  );
 ```
 
-**Important**: By default `react-hook-form` submits values of unmounted components, so if you don't want this behaivour you can set `shouldUnregister` prop to `true` in the form
+**Important**: By default `react-hook-form` submits values of unmounted input components, so if you don't want this behaivour you can set `shouldUnregister` prop to `true` in the form
 
 ```diff
 import { FormDataConsumer } from 'react-admin';
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -213,7 +213,7 @@ form input value (string) ---> parse ---> form state value
     parse={v => parseFloat(v) / 100}
 ```
 
-`parse` often comes in pair with [`format`](#format) to transform the form value value before passing it to the input. See the [Transforming Input Value](#transforming-input-value-tofrom-record) section for more details.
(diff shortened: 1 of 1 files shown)


docs/Inputs.md
```diff
@@ -522,6 +522,26 @@ import { FormDataConsumer } from 'react-admin';
  );
 ```
 
+**Important**: By default `react-hook-form` submits values of unmounted input components, so if you don't want this behaivour you can set `shouldUnregister` prop to `true` in the form
```
fzaninotto, line 525: I'd put the shouldUnregister directly in the first example (L513) instead of duplicating it. then, you can explain why it's necessary.

fzaninotto (changes requested):

docs/Inputs.md
```diff
@@ -522,6 +522,8 @@ import { FormDataConsumer } from 'react-admin';
  );
 ```
 
+**Important**: By default `react-hook-form` submits values of unmounted input components, to prevent this from happen in the above example, `shouldUnregister` prop was set to `true` in the `<SimpleForm>`
```
fzaninotto, line 525:
```suggestion
**Note**: By default, `react-hook-form` submits values of unmounted input components. In the above example, the `shouldUnregister` prop of the `<SimpleForm>` component prevents that from happening. That way, when end users hide an input, its value isn't included in the submitted data.
```

afilp:
I think that this may be an issue when hiding/showing filters too. At least this is what happened with a custom filter of mine when pressing the "clear" ("x") button. The query was still sending the filter value.

fzaninotto:
You'll have to write your own filter form.

https://marmelab.com/react-admin/FilteringTutorial.html#building-a-custom-filter

PR #8196, opened 2022-09-26
[ra-core] use react-query useQuery for permissions fetching

hiaselhans:
I wanted to hook up the builtin permission system from react-admin and I experienced a ton of queries: one per permission check.
the docs have a hidden (pro/paid flagged) hint on how to solve this:
https://marmelab.com/react-admin/AuthRBAC.html#performance

since we are using react-query already all-over why don't we use it for the permission-fetching?
This adds a lot of sane defaults: re-fetch on window focus, cache invalidation on logout, retries with back-off, ...

I think the same could be applied to `useCheckAuth`, `useGetIdentity`,... but I'd like to hear your comments first.

### DIFF
--- a/packages/ra-core/src/auth/usePermissions.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissions.spec.tsx
@@ -6,10 +6,10 @@ import { CoreAdminContext } from '../core/CoreAdminContext';
 import usePermissions from './usePermissions';
 
 const UsePermissions = ({ children }: any) => {
-    const authParams = {
+    const permissionQueryParams = {
         retry: false,
     };
-    const res = usePermissions({}, authParams);
+    const res = usePermissions({}, permissionQueryParams);
     return children(res);
 };
 
--- a/packages/ra-core/src/auth/usePermissions.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissions.spec.tsx
@@ -9,7 +9,7 @@ const UsePermissions = ({ children }: any) => {
     const authParams = {
         retry: false,
     };
-    const res = usePermissions(authParams);
+    const res = usePermissions({}, authParams);
     return children(res);
 };
 
--- a/packages/ra-core/src/auth/usePermissions.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissions.spec.tsx
@@ -5,7 +5,10 @@ import { CoreAdminContext } from '../core/CoreAdminContext';
 
 import usePermissions from './usePermissions';
 
-const UsePermissions = ({ children, authParams }: any) => {
+const UsePermissions = ({ children }: any) => {
+    const authParams = {
+        retry: false,
+    };
     const res = usePermissions(authParams);
     return children(res);
 };
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -1,5 +1,7 @@
+import { useMemo } from 'react';
 import useGetPermissions from './useGetPermissions';
 import { useQuery, UseQueryOptions } from 'react-query';
+import { reactAdminFetchActions } from '../dataProvider';
 
 const emptyParams = {};
 /**
@@ -40,17 +42,19 @@ const usePermissions = <Permissions = any, Error = any>(
 ) => {
     const getPermissions = useGetPermissions();
 
-    const { data, isLoading, error } = useQuery(
+    const result = useQuery(
         ['auth', 'getPermissions', params],
         () => getPermissions(params),
         queryParams
     );
 
-    return {
-        permissions: data,
-        isLoading,
-        error,
-    };
+    return useMemo(() => {
+        return {
+            permissions: result.data,
+            isLoading: result.isLoading,
+            error: result.error,
+        };
+    }, [result]);
 };
 
 export default usePermissions;
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -2,6 +2,7 @@ import { useMemo } from 'react';
 import useGetPermissions from './useGetPermissions';
 import { useQuery, UseQueryOptions } from 'react-query';
 import { reactAdminFetchActions } from '../dataProvider';
+import useAuthProvider from './useAuthProvider';
 
 const emptyParams = {};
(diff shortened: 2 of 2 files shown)


packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +32,23 @@ const emptyParams = {};
...
-                setState({
-                    isLoading: false,
-                    error,
-                });
-            });
-    }, [getPermissions, params, setState]);
-    return state;
+
+    const { data, isLoading, error } = useQuery(
+        ['ra-Permissions'],
```
fzaninotto, line 42: please use `['auth', 'getPermissions', params]` as key

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +32,23 @@ const emptyParams = {};
...
+        getPermissions,
```
fzaninotto, line 43: This is a breaking change, as we used to pass in the params

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +32,23 @@ const emptyParams = {};
...
+        params
+    );
+
+    return {
```
fzaninotto, line 47: You probably need to memo this, or this will trigger unwanted rerenders

fzaninotto (changes requested):
Nice! This is definitely something we wanted to do, so it's nice of you to tackle this.

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: done :)

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: also passed now

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: Are you sure about that?
It's basically just a wrapper for `useQuery`.

fzaninotto: Yes I'm sure: `useQuery` will return an identical object if nothing has changed, but your hook will always return a new object.

fzaninotto: See #8174

packages/ra-core/src/auth/usePermissions.spec.tsx
```diff
@@ -5,8 +5,11 @@ import { CoreAdminContext } from '../core/CoreAdminContext';
 
 import usePermissions from './usePermissions';
 
-const UsePermissions = ({ children, authParams }: any) => {
-    const res = usePermissions(authParams);
+const UsePermissions = ({ children }: any) => {
+    const authParams = {
+        retry: false,
+    };
+    const res = usePermissions({}, authParams);
```
fzaninotto, line 12: please don't name the variable `authParams` as it's misleading

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +33,24 @@ const emptyParams = {};
  *     };
  */
 const usePermissions = <Permissions = any, Error = any>(
-    params = emptyParams
-): State<Permissions, Error> => {
-    const [state, setState] = useSafeSetState<State<Permissions, Error>>({
-        isLoading: true,
-    });
+    params = emptyParams,
+    useQueryParams: UseQueryOptions<Permissions, Error> = {
```
fzaninotto, line 37: I'd call it `queryParams` instead of `useQueryParams`

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +33,24 @@ const emptyParams = {};
...
+        refetchOnMount: false,
+    }
+) => {
     const getPermissions = useGetPermissions();
```
fzaninotto, line 41: Since we don't need to make the effect dependent on `getPermissions` anymore, I'd use `useAuthProvider` instead of `useGetPermissions`.

fzaninotto (changes requested):

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: done.

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: done, thanks!

packages/ra-core/src/auth/usePermissions.ts
hiaselhans: I removed the useGetPermissions and added a conditional if authProvider is null.
does that work for you?

packages/ra-core/src/auth/usePermissions.spec.tsx
hiaselhans: renamed

fzaninotto:
Nice! Waiting for a second review to merge.

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -42,25 +34,28 @@ const emptyParams = {};
...
-                    isLoading: false,
-                    error,
-                });
-            });
-    }, [getPermissions, params, setState]);
-    return state;
+    params = emptyParams,
+    queryParams: UseQueryOptions<Permissions, Error> = {
+        refetchOnMount: false,
+    }
```
djhi, line 40: Can you explain why you chose these defaults?

djhi (commented):
Nice! Thanks

djhi (approved):

packages/ra-core/src/auth/usePermissions.ts
fzaninotto: Shouldn't you use a `staleTime` instead?

hiaselhans: sure, i'm fine with staleTime too.
What do you think is a sane default? 60s? 5min?

fzaninotto: 5 mins is OK

hiaselhans: updated :)

PR #8202, opened 2022-09-27
Fix: <ReferenceField> generates a link even though there is nothing to link to

### LINKED ISSUE #8115: `<ReferenceField>` generates a link even though there is nothing to link to
When I hover over a ReferenceField, it shows the link as `#/rhymes/32`.
But, when clicking on it, the network request made is `/rhymes/list?pagination={"page":1,"perPage":10}&sort={"field":"id","order":"ASC"}&filter={}` (i.e. the default list request).
This results in the page changing to the first page of that resource, which does not even contain the specific entity.

Version: 4.3.0

thibault-barrat:
Fixes #8115


WiXSL (approved):
Looks really good!

slax57 (approved):

PR #8204, opened 2022-09-28
Fix `SimpleFormIterator` defaultValues when adding a record

### LINKED ISSUE #8178: Array input issue when removing and reinserting element
# Bug
The bug happens when I have inside a `SimpleFormIterator` a `React.FC` or a `FormDataConsumer`.
In those cases if I have defaults they are not applied and if there is a record the informations shown does not respect the form state.

# To reproduce
To reproduce the bug use this [CODESANDBOX](https://codesandbox.io/s/heuristic-diffie-gzwqfp?file=/src/comments/CommentEdit.tsx)
Enter into the edit of the only comment present here delete this from the form. Then reinsert it. 
The form seems filled because you can see the numbers but if you press submit the validation error appears informing you that the form is not filled.

Result of bug:
image

# Problem found
The screen below is the `SimpleFormIterator` `addField` function, in this function if I have a form data consumer or any other React.FC that have not source nor the default value on the SimpleFormIterator level the default returned is `{undefined: ""}`
This causes that react admin renders the array value from the record thinking that the form is filling with the record, showing a wrong value from record and not from the form state. 
image

# Drity workaround to partially solve this issue

```tsx
type Props<T> = SimpleFormIteratorProps & {
  defaultValue?: { [TKey in keyof T]: T[TKey] }
}

/**
 * This component is used because react admin cannot handle default values for React.FC children of
 * the standard SimpleFormIterator. For those elements the native one takes the default from the record but it
 * does not update the record with like result having the field filled but with an empty form.
 *
 * @param SimpleFormIteratorProps all the props of the SimpleFormIterator
 * @param defaultValue this one is used when a new element is added
 * @return the standard SimpleFormIterator with the ability to add a default value to each element of the array
 */
const KSimpleFormIterator = <T extends any>({
  children,
  source,
  defaultValue,
  ...rest
}: Props<T>) => {
  // this source is automatically passed by the array input
  if (!source) throw new Error('KSimpleFormIterator: source is required')

  // we are going to perform a watch on the the array input values
  const arrayValue = useWatch({ name: source }) as Array<any>

  // we memorize the array input values
  const [value, setValue] = useState(arrayValue)

  // this hook is used to manipulate the ReactAdmin array input form state
  const { append, remove } = useArrayInput({ ...rest })

  console.log('aaa', {
    children: Children.toArray(children),
  })

  // if an element was added
  if (value.length < arrayValue.length) {
    // remove the last added element with wrong default value
    remove(arrayValue.length - 1)
    // set the default value with the correct one appending it like last element of the array
    append(defaultValue)
    // updates the cached value inside the component
    setValue(arrayValue)
  } else if (value.length > arrayValue.length) {
    // updates the cached value inside the component
    setValue(arrayValue)
  }

  return (
    <SimpleFormIterator {...rest} source={source}>
      {children}
    </SimpleFormIterator>
  )
}

export default KSimpleFormIterator

```

WiXSL:
Fixes #8178

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
@@ -74,8 +74,6 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
                 if (
                     Children.count(children) === 1 &&
                     React.isValidElement(Children.only(children)) &&
-                    (Children.only(children) as ReactElement<any>).type !==
-                        FormDataConsumer &&
                     // @ts-ignore
                     !Children.only(children).props.source
                 ) {
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
@@ -8,6 +8,7 @@ import {
     ReactNode,
     useCallback,
     useMemo,
+    useRef,
 } from 'react';
 import { styled, SxProps } from '@mui/material';
 import clsx from 'clsx';
@@ -30,8 +31,6 @@ import { AddItemButton as DefaultAddItemButton } from './AddItemButton';
 import { RemoveItemButton as DefaultRemoveItemButton } from './RemoveItemButton';
 import { ReOrderButtons as DefaultReOrderButtons } from './ReOrderButtons';
 
-let initialDefaultValue = {};
-
 export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
     const {
         addButton = <DefaultAddItemButton />,
@@ -51,6 +50,7 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
     } = props;
     const { append, fields, move, remove } = useArrayInput(props);
     const record = useRecordContext(props);
+    const initialDefaultValue = useRef({});
 
     const removeField = useCallback(
         (index: number) => {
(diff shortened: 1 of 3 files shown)


thibault-barrat (approved):

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -30,6 +30,8 @@ import { AddItemButton as DefaultAddItemButton } from './AddItemButton';
 import { RemoveItemButton as DefaultRemoveItemButton } from './RemoveItemButton';
 import { ReOrderButtons as DefaultReOrderButtons } from './ReOrderButtons';
 
+let initialDefaultValue = {};
```
fzaninotto, line 33: I'm very scared of that type of global values - it will persist across the entire app. We'll see side effects in tests. 

Can't you find an alternative fix that doesn't rely on a global? Why not a `ref`?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -57,13 +59,22 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
...
+
     const addField = useCallback(
         (item: any = undefined) => {
             let defaultValue = item;
             if (item == null) {
+                defaultValue = initialDefaultValue;
                 if (
                     Children.count(children) === 1 &&
                     React.isValidElement(Children.only(children)) &&
+                    (Children.only(children) as ReactElement<any>).type !==
```
fzaninotto, line 76: Can we avoid inspecting the child type?

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
WiXSL: Actually, I just realized that check is not necessary there at all

slax57 (approved):

fzaninotto:
Actually, this PR misses a unit test...

PR #8212, opened 2022-09-30
Fix null values cause warning in forms

### LINKED ISSUE #7782: Warning when using null value on TextInput

fzaninotto:
Closes #7782

### DIFF
--- a/packages/ra-core/src/form/getFormInitialValues.ts
+++ b/packages/ra-core/src/form/getFormInitialValues.ts
@@ -10,7 +10,7 @@ export default function getFormInitialValues(
         getValues(defaultValues, record),
         record
     );
-    // replace null values by empty string to avoid controllezd/ uncontrolled input warning
+    // replace null values by empty string to avoid controlled/ uncontrolled input warning
     Object.keys(finalInitialValues).forEach(key => {
         if (finalInitialValues[key] === null) {
             finalInitialValues[key] = '';
--- a/packages/ra-core/src/form/getFormInitialValues.ts
+++ b/packages/ra-core/src/form/getFormInitialValues.ts
@@ -10,6 +10,12 @@ export default function getFormInitialValues(
         getValues(defaultValues, record),
         record
     );
+    // replace null values by empty string to avoid controllezd/ uncontrolled input warning
+    Object.keys(finalInitialValues).forEach(key => {
+        if (finalInitialValues[key] === null) {
+            finalInitialValues[key] = '';
+        }
+    });
     return finalInitialValues;
 }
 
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/form/getFormInitialValues.ts
```diff
@@ -10,6 +10,12 @@ export default function getFormInitialValues(
         getValues(defaultValues, record),
         record
     );
+    // replace null values by empty string to avoid controlled/ uncontrolled input warning
+    Object.keys(finalInitialValues).forEach(key => {
+        if (finalInitialValues[key] === null) {
+            finalInitialValues[key] = '';
+        }
+    });
```
slax57, line 18: This doesn't work for nested values, like `author.name` for instance

slax57 (changes requested):

WiXSL (approved):
Apart from @slax57 review seems good!

packages/ra-core/src/form/getFormInitialValues.ts
fzaninotto: So the fix requires a bit of recursion: 

```jsx
export default function getFormInitialValues(
    defaultValues: DefaultValue,
    record?: Partial<RaRecord>
) {
    const finalInitialValues = merge(
        {},
        getValues(defaultValues, record),
        record
    );
    // replace null values by empty string to avoid controlled/ uncontrolled input warning
    return mapValuesDeep(finalInitialValues, value =>
        value === null ? '' : value
    );
}

const mapValuesDeep = (obj, callback) =>
    mapValues(obj, v =>
        isPlainObject(v) ? mapValuesDeep(v, callback) : callback(v)
    );
```

But I'm afraid it could have nasty side effects. Why should we modify the inner properties of embedded objects - some of which my not be modifiable? Besides, the `sanitizeEmptyValues` doesn't work for nested properties, so the empty string would be sent back to the server. 

All in all, this is another corner case, which already exists today, and I don't think this PR makes it worse.

slax57 (approved):

PR #8214, opened 2022-09-30
Feat: Fix radiobuttongroup inse referenceInput

arimet:
https://user-images.githubusercontent.com/18294269/193272249-ed5dc01d-25e3-4be4-840c-5e5a62bc64e4.mp4

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.ts
@@ -125,7 +125,10 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         finalData = possibleValuesData;
         finalTotal = total;
     } else {
-        finalData = [referenceRecord, ...possibleValuesData];
+        finalData = [...possibleValuesData];
+        if (finalData.indexOf(referenceRecord) === -1) {
+            finalData.push(referenceRecord);
+        }
         finalTotal = total == null ? undefined : total + 1;
     }
 


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -125,7 +125,10 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         finalData = possibleValuesData;
         finalTotal = total;
     } else {
-        finalData = [referenceRecord, ...possibleValuesData];
+        finalData = [...possibleValuesData];
```
arimet, line 128: This is the only solution I have found for now. 

During a short time, referenceRecord becomes the selected value in the radio group. 
So we have a finalData with a duplicate. 

I will add it in two steps with a check to avoid the rendering jump

fzaninotto: I suggest you use a JavaScript Set instead

https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Set

fzaninotto (changes requested):

packages/ra-core/src/controller/input/useReferenceInputController.ts
arimet: I had tried this technique, but it creates a "flick" when rendering the component.


https://user-images.githubusercontent.com/18294269/193303452-018a60de-0dbd-4947-b31e-d44c382c28e0.mp4

fzaninotto:
Superseded by #8229

PR #8217, opened 2022-09-30
Feat: move remove filter button and display it on hover

thibault-barrat:
- [x] Display the close button only when hovering the input
- [x] Move the close button to the top right of the input

https://user-images.githubusercontent.com/41492953/193309603-13fc866b-f0b6-4714-a391-10f974638e8a.mov

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterFormInput.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterFormInput.tsx
@@ -64,10 +64,25 @@ const Root = styled('div', {
     display: 'flex',
     alignItems: 'flex-end',
     pointerEvents: 'auto',
+    position: 'relative',
+    '@media (hover: hover)': {
+        '&:hover': {
+            [`& .${FilterFormInputClasses.hideButton}`]: {
+                display: 'flex',
+            },
+        },
+    },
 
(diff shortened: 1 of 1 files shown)


WiXSL (approved):
I really like it! 👍

fzaninotto:
I like it!

Why do you see it as a regression in accessibility?

In my opinion, this feature needs a bit of visual polish on the background and position of the close button, the hover zone, and the tooltip.

slax57:
> I like it!
> 
> Why do you see it as a regression in accessibility?
> 
> In my opinion, this feature needs a bit of visual polish on the background and position of the close button, the hover zone, and the tooltip.

Because you can no longer use the Clear button with keyboard only

septentrion-730n:
I make a reservation on two points:
- filter UI do not display its clearable state without hovering it; I would prefer the button to be displayed by default
- chips are now clickable, without consequent action; they have the UI of a toggler (enabled/disabled) but without being one

fzaninotto:
After thinking about the accessibility regression, we won't move forward with this change. We'll keep the less visually appealing "always on" clear button to remain accessible. 

thanks for your contribution anyway!

fzaninotto:
Superseded by #9224

PR #8218, opened 2022-09-30
[Doc] Filter search by versions

thibault-barrat:
- [x] Add a filter to the algolia request with the current displayed version of the doc
- [ ]  To make it work algolia crawler must be configured accordingly


WiXSL (approved):
Looks great!

PR #8221, opened 2022-09-30
Fix `Datagrid`'s `ExpandAllButton` conflicts with `expandSingle` prop

WiXSL:
If `expandSingle` prop is `true`, the user shouldn't have the possibility to expand all rows.


fzaninotto:
Great! I'd love a story and a test on this one.

slax57 (approved):

PR #8222, opened 2022-10-01
typeof "undefined" shall be undefined

zhujinxuan:
Only `undefined` can match `typeof _ === 'undefined'`, so I change it for clarity.

I also change some && and || statement to `?.` as well.

### DIFF
--- a/examples/crm/src/deals/OnlyMineInput.tsx
+++ b/examples/crm/src/deals/OnlyMineInput.tsx
@@ -12,7 +12,7 @@ export const OnlyMineInput = ({ alwaysOn }: { alwaysOn: boolean }) => {
 
     const handleChange = () => {
         const newFilterValues = { ...filterValues };
-        if (typeof filterValues.sales_id !== 'undefined') {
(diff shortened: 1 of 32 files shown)


fzaninotto:
`typeof x === 'undefined'` and `x === undefined` aren't equivalent in JS: the latter throws a ReferenceError is x isn't declared. 

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined for details

Can you explain why you think your PR is an improvement?

zhujinxuan:
Sorry, I did not know that.

I thought the string compare is slower than `=== undefined`, so removing the string compare would improve the overall efficiency a bit.

fzaninotto:
I prefer to always use typeof, as it always works and prevents us from wondering which technique to use in which case. 

As for performance problems, please profile a real-life web app before and after the fix to prove that its has a noticeable influence. In my opinion, the impact is negligible.

zhujinxuan:
Got it!  I will close this PR until I have evidence of the efficiency impact.

PR #8223, opened 2022-10-02
Remove unused TranslationMessage

bicstone:
I noticed that `ra.configurable.templateError` was added in version 4.4.0 but is unused.
It seems that they deleted `translate(ra.configurable.templateError, ...)` when changed to translate instead of lodash.

https://github.com/marmelab/react-admin/commit/c4a6ac61599fe8dd48d74590a51a36f44205de13#diff-be8dc1889be155c57e45c5b1e0d96f8f05466f57a969429ef5cd766baedcde9aL22-L24
(deleted /packages/ra-core/src/preferences/useRenderTemplate.ts)

### DIFF
--- a/packages/ra-core/src/i18n/TranslationMessages.ts
+++ b/packages/ra-core/src/i18n/TranslationMessages.ts
@@ -177,7 +177,6 @@ export interface TranslationMessages extends StringMap {
         };
         configurable?: {
             customize: string;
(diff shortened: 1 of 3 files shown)


fzaninotto (approved):
Good catch!

fzaninotto:
Thanks!

PR #8225, opened 2022-10-02
ci: add minimum GitHub token permissions for workflows

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using [secure-workflows](https://github.com/step-security/secure-workflows).

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes. Here is an example of the permissions in one of the workflow runs:
https://github.com/marmelab/react-admin/actions/runs/3169425882/jobs/5161329731#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- codeql-analysis.yml
- test.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- [GitHub recommends](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) defining minimum GITHUB_TOKEN permissions.
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,10 +6,17 @@ on:
   schedule:
     - cron: '0 10 * * 3'
 
+permissions:
+  contents: read
+
 jobs:
   CodeQL-Build:
 
     # CodeQL runs on ubuntu-latest and windows-latest
+    permissions:
+      actions: read  # for github/codeql-action/init to get workflow details
+      contents: read  # for actions/checkout to fetch code
+      security-events: write  # for github/codeql-action/autobuild to send a status report
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


slax57:
Duplicate of: https://github.com/marmelab/react-admin/pull/7823
However your PR contains more context info, which allows me to better understand the context and benefits of it.

However, I agree with [danielroe's comment](https://github.com/nuxt/nuxt.js/pull/10734#issuecomment-1265160541), we should rather enforce it globally on our repo than on a per-workflow basis.

I'll keep this PR open until we do, but we won't probably merge it as is.

Still, thanks for the PR and the contextual info.

slax57:
Closing as superseded by https://github.com/marmelab/react-admin/pull/8253

Again, thank you @boahc077 for raising this up!

PR #8226, opened 2022-10-03
Fix StoreContextProvider to be accessible from the dataProvider context

### LINKED ISSUE #8220: Provide store context in the dataprovider
Currently, the dataprovider does not have access to the react admin store. This prevents making use of a user-set variable in the react admin store when making API calls.

A great example is the language switcher, which stores data in the store. It would be practical to be able to send the user's language with each API call, without having to do it per API call using mutationOptions.

1. Move the StoreContextProvider above the DataProviderContext in CoreAdminContext:

https://github.com/marmelab/react-admin/blob/dbd036f8122c1376b65bbc7e5cf6497641bde4de/packages/ra-core/src/core/CoreAdminContext.tsx#L82-L83

2. Provide access to the store as an optional argument to dataprovider functions.

**Additional context**
[I brought up the issue here first
](https://stackoverflow.com/questions/73833840/how-can-i-access-the-react-admin-store-in-the-dataprovider-react-admin-4-x/73839240?noredirect=1#comment130436134_73839240)

WiXSL:
Fixes #8220

### DIFF
--- a/packages/ra-core/src/core/CoreAdminContext.tsx
+++ b/packages/ra-core/src/core/CoreAdminContext.tsx
@@ -80,8 +80,8 @@ React-admin requires a valid dataProvider function to work.`);
 
     return (
         <AuthContext.Provider value={finalAuthProvider}>
-            <DataProviderContext.Provider value={finalDataProvider}>
-                <StoreContextProvider value={store}>
+            <StoreContextProvider value={store}>
+                <DataProviderContext.Provider value={finalDataProvider}>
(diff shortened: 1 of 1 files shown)


fzaninotto:
We need a bit more than that: the `dataProvider` isn't a hook and therefore cannot call the `useStore` hook.

djhi:
It's unnecessary. Just initialize your store in a separate file and import it where needed, no?

WiXSL:
Indeed!

PR #8237, opened 2022-10-05
Fix AutocompleteInput (Format and Parse)

byymster:
## Fix AutocompleteInput - Format and Parse doesn't work

Right now there's no way to use `format` and `parse` in AutocompleteInput (and AutocompleteArrayInput)  because of ordering logical operations.

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -223,9 +223,9 @@ export const AutocompleteInput = <
         formState: formStateOverride,
         onBlur,
         onChange,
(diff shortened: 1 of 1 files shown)


WiXSL:
Nice catch!
Could you fix the typecheck warning?

WiXSL (changes requested):
Nice catch!
Could you fix the typecheck warning?

byymster:
@WiXSL done. should be good

WiXSL (approved):

fzaninotto (approved):

fzaninotto:
Great collaboration, good to go!

fzaninotto:
Thanks for the patch.

PR #8238, opened 2022-10-05
Fix AutocompleteInput input's value not updating correctly

### LINKED ISSUE #8227: `AutocompleteInput` input's value not updating correctly
Manipulating the `AutocompeteInput`'s value by the user may lead to not removable characters until focus is lost

https://user-images.githubusercontent.com/8268610/193594861-53241eed-552a-4f0b-84ed-9ec804c59775.mov

Environment

React-admin version: 4.4.0
Last version that did not exhibit the issue (if applicable):
React version: 18.2.0
Browser: Chrome
Stack trace (in case of a JS error):

WiXSL:
Fixes #8227

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -58,6 +58,37 @@ describe('<AutocompleteInput />', () => {
         expect(screen.queryByDisplayValue('foo')).not.toBeNull();
     });
 
+    it('should not match selection when selected choice id equals the emptyValue while changing the input', async () => {
+        render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <SimpleForm>
+                    <AutocompleteInput
+                        {...defaultProps}
+                        choices={[
+                            { id: 2, foobar: 'foo' },
+                            { id: 3, foobar: 'bar' },
+                        ]}
+                    />
+                </SimpleForm>
+            </AdminContext>
+        );
+        const input = screen.getByLabelText(
+            'resources.users.fields.role'
+        ) as HTMLInputElement;
+
+        fireEvent.focus(input);
+
+        userEvent.type(input, 'f');
+        await waitFor(() => {
+            expect(input.value).toEqual('f');
+        });
+
+        userEvent.type(input, '{backspace}');
+        await waitFor(() => {
+            expect(input.value).toEqual('');
+        });
+    });
+
     describe('emptyText', () => {
         it('should allow to have an empty menu option text by passing a string', () => {
             const emptyText = 'Default';
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -486,16 +486,15 @@ describe('<AutocompleteInput />', () => {
         });
     });
 
-    it('should allow to clear the first character', async () => {
+    it('should not match selection when selected choice id equals the emptyValue while changing the input', async () => {
         render(
             <AdminContext dataProvider={testDataProvider()}>
-                <SimpleForm onSubmit={jest.fn()} defaultValues={{ role: 2 }}>
+                <SimpleForm>
                     <AutocompleteInput
                         {...defaultProps}
-                        optionText="foobar"
                         choices={[
-                            { id: 2, foobar: 'foo' },
-                            { id: 3, foobar: 'bar' },
+                            { id: 2, name: 'foo' },
+                            { id: 3, name: 'bar' },
                         ]}
                     />
                 </SimpleForm>
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -58,34 +58,40 @@ describe('<AutocompleteInput />', () => {
         expect(screen.queryByDisplayValue('foo')).not.toBeNull();
     });
 
-    it('should not match selection when selected choice id equals the emptyValue while changing the input', async () => {
+    it('should allow filter to match the selected choice while removing characters in the input', async () => {
         render(
             <AdminContext dataProvider={testDataProvider()}>
                 <SimpleForm>
                     <AutocompleteInput
                         {...defaultProps}
                         choices={[
-                            { id: 2, foobar: 'foo' },
-                            { id: 3, foobar: 'bar' },
+                            { id: 1, name: 'foo' },
+                            { id: 2, name: 'bar' },
                         ]}
                     />
                 </SimpleForm>
             </AdminContext>
         );
+
         const input = screen.getByLabelText(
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -440,10 +452,12 @@ If you provided a React element for the optionText prop, you must also provide t
             }
 
             return eventType && eventType === 'change'
-                ? selectedItemTexts.includes(filter) && selectedChoice
+                ? selectedItemTexts.includes(filter) &&
+                      selectedChoice &&
+                      selectedChoice[optionValue || 'id'] !== emptyValue
```
slax57, line 457: Not sure this will work in the `multiple` case, because int this case `selectedChoice` seems to be an array (see line 447)

slax57 (changes requested):
Besides, it would be nice if you could add some tests to cover this case in the future :slightly_smiling_face:

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
WiXSL: Doesn't seems to be a problem for multiple, so I'll just make that evaluation when is not multiple

slax57:
I'm afraid the fix is still incomplete.

There are still cases where you cannot use backspace, for example in the `AutocompleteInput/Basic` story:

Peek 07-10-2022 18-47

I think the problem actually comes from this function:

```tsx
    const handleInputChange = (
        event: any,
        newInputValue: string,
        reason: string
    ) => {
        if (!doesQueryMatchSelection(newInputValue, event?.type)) {
            setFilterValue(newInputValue);
            debouncedSetFilter(newInputValue);
        }
    };
```

The `if` makes it so that the input is not updated if the new value is a valid option.
I don't know why this `if` was added. I tried removing it and everything seems to work just fine without it.

Besides, the unit test `should allow to clear the first character` is actually wrong.
Actually it should break from what I see in the storybook.
I believe it's because after focus, `userEvent.type(input, 'f');` is actually **replacing** the whole text content instead of appending to it.
We should fix this test, too.

In the end, I'm not sure we need to change the `doesQueryMatchSelection` callback. Removing the `if` in `handleInputChange` seems to be enough.

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -58,6 +58,43 @@ describe('<AutocompleteInput />', () => {
         expect(screen.queryByDisplayValue('foo')).not.toBeNull();
     });
 
+    it('should allow filter to match the selected choice while removing characters in the input', async () => {
```
slax57, line 61: This new test seems fine, but why did you remove the old one testing the emptyValue case? To me we should keep both.

Besides, if my understanding is correct, this test supersedes the `should allow to clear the first character` test, which is useless anyway, so we should remove it IMO.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -429,21 +441,29 @@ If you provided a React element for the optionText prop, you must also provide t
...
+                    selectedChoice[
+                        typeof optionText === 'string' ? optionText : 'name'
+                    ] === filter;
             }
 
             return eventType && eventType === 'change'
-                ? selectedItemTexts.includes(filter) && selectedChoice
+                ? selectedItemTexts.includes(filter) &&
+                      selectedChoice &&
+                      !isValidChoice
```
slax57, line 463: I am a bit skeptical about having a `doesQueryMatchSelection` function return `false` when the choice is a valid one.

From the tests, your fix is clearly working, I'm not questioning that, but I still have the feeling we put a patch at the wrong place.
What is the purpose of this `doesQueryMatchSelection` function anyway?
To me it is very unclear what it's supposed to do, so it's difficult to tell if its implementation is correct or not.
Did you figure it out?

slax57 (changes requested):

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
WiXSL: Let me research a little bit, at first, removing it makes some tests fails.

packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
WiXSL: You're actually right!

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
slax57: The more I think about it the more I think it's wrong :sweat_smile: 

Actually, putting the `multiple` case aside: 
- `selectedItemTexts` is equal to `[getOptionLabel(selectedChoice)]` so `selectedItemTexts.includes(filter)` is basically equivalent to `getOptionLabel(selectedChoice) === filter`
- `isValidChoice` is, after simplification, equal to `selectedChoice[optionText] === filter`, which is very very similar to `getOptionLabel(selectedChoice) === filter`

So in the end, `selectedItemTexts.includes(filter) && !isValidChoice` will probably always be `false`. We are basically expecting something and its opposite to be `true` at the same time.

WiXSL: When you are right, you are right! 😄

WiXSL: I simplify everything to always run the filter on input change, and call `doesQueryMatchSelection` on anything else

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -422,15 +434,18 @@ If you provided a React element for the optionText prop, you must also provide t
         newInputValue: string,
         reason: string
     ) => {
-        if (!doesQueryMatchSelection(newInputValue, event?.type)) {
+        if (
+            event?.type === 'change' ||
+            !doesQueryMatchSelection(newInputValue, event?.type)
```
slax57, line 439: We no longer need to check the `event` type in `doesQueryMatchSelection()`, because it's already done in `handleInputChange()`.

`handleInputChange()` was the only place where  `doesQueryMatchSelection()` was called with a second parameter, so we can safely remove it IMO.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -444,7 +459,7 @@ If you provided a React element for the optionText prop, you must also provide t
                 ? selectedItemTexts.includes(filter) && selectedChoice
                 : selectedItemTexts.includes(filter);
         },
-        [getOptionLabel, multiple, selectedChoice]
+        [getOptionLabel, multiple, optionText, selectedChoice]
```
slax57, line 462: linter warning here

slax57 (changes requested):

slax57 (approved):

PR #8243, opened 2022-10-06
Fix redirectTo returned by `useEditController`

WiXSL:
`useEditController` should return the redirectTo used to redirect (if any) or the default, just like `useCreateController` does it, not the default

### DIFF
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -220,7 +220,7 @@ export const useEditController = <
         isLoading,
         mutationMode,
         record,
-        redirect: DefaultRedirect,
+        redirect: redirectTo,
         refetch,
         registerMutationMiddleware,
         resource,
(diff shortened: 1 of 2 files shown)


fzaninotto:
I'm not sure what bug this fixes. Could you please add a unit test?

slax57 (approved):

slax57:
Seems good to me! I'll just wait on François' approval before merging

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -220,7 +220,7 @@ export const useEditController = <
         isLoading,
         mutationMode,
         record,
-        redirect: DefaultRedirect,
+        redirect: redirectTo,
```
fzaninotto, line 223: This context variable isn't used anywhere in our code - I think it's a leftover from v3. I'll validate the PR, but I think we should ultimately mark it as deprecated.

fzaninotto (approved):

PR #8244, opened 2022-10-06
[Doc] Add  List's `perPage` and `pagination.rowsPerPages` props relationship note

### LINKED ISSUE #8224: Pagination Don't work as Expected
**What you were expecting:**

- I was expecting the number of rows in a list to be equal to the first value I passed in the array 
  ` <Pagination rowsPerPageOptions={[ 50, 75, 100]} />` As per here, Which was supposed to be 50

**What happened instead:**

-  But, I it seems that the very first index must be 10 rowsPerPage only. If I use something other than 10 (Here I've used 50) , It will still load only 10 rows and the Rows per page selection will remain empty. 

**Steps to reproduce:**

- Set Pagination on a list and write rowsPerPageOptions={[50,75,100]} (Unexpected
- Now, Set Pagination with rowsPerPageOptions={[10,50,75,100]} and Observe. (Expected)

**Related code:**

```JavaScript
const StudentList = () => {
  const StudentPagination = () => (
    <Pagination rowsPerPageOptions={[ 50, 75, 100]} />
  );
  return (
    <List filters={Filters} pagination={<StudentPagination />}>
      <Datagrid rowClick="show"  bulkActionButtons={false}>
        <TextField source="id" />
        <TextField source="name" />
        <ReferenceField  label="SCHOOL" source="school_id" reference="school">
          <TextField  source="name" />
        </ReferenceField>
        <ReferenceField label="UDISE" source="school_id" reference="school">
          <TextField source="udise" />
        </ReferenceField>
        <TextField source="father_name" />
        <TextField source="mother_name" />
        <TextField source="gender" />
        <NumberField source="grade_number" />
        <TextField source="stream_tag" />
        <TextField source="category" />
        <BooleanField source="is_cwsn" label={"CWSN"}/>
        <BooleanField source="is_enabled" label={"Enabled"}/>
        <EditButton/>
      </Datagrid>
    </List>
  );
};
export default StudentList;
```

**Other information:**

**Environment**

* React-admin version: 4.3.0
* React version: 18.2.0
* Browser: Chrome

WiXSL:
Fixes #8224

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -646,15 +646,12 @@ export const PostList = () => (
 
 ```diff
 // in src/MyPagination.js
-import { Pagination, List } from 'react-admin';
-
-const PostPagination = props => <Pagination {...props} />;
+-import { List } from 'react-admin';
++import { List, Pagination } from 'react-admin';
 
 export const PostList = () => (
-     // perPage equals 10 by default, is not incuded in rowsPerPageOptions
--    <List pagination={<PostPagination rowsPerPageOptions={[6, 12, 24, 36]} />}>
-     // perPage equals 6, is incuded in rowsPerPageOptions
-+    <List perPage={6} pagination={<PostPagination rowsPerPageOptions={[6, 12, 24, 36]} />}>
+-    <List perPage={6}>
++    <List perPage={6} pagination={<Pagination rowsPerPageOptions={[6, 12, 24, 36]} />}>
         ...
     </List>
 );
--- a/docs/List.md
+++ b/docs/List.md
@@ -642,7 +642,23 @@ export const PostList = () => (
 );
 ```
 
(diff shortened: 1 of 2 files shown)


docs/List.md
```diff
@@ -642,6 +642,8 @@ export const PostList = () => (
 );
 ```
 
+**Note**: If you've added a [`Pagination`](#pagination-pagination-component) component to your `List` component, make sure to set `perPage` to a value that exists in `rowsPerPageOptions`
```
fzaninotto, line 645: I'd put it differently: I'd explain that the default pagination includes per page options of 5, 10, and 25. If the developer sets the `perPage` to a value not in that set, they must also customize the pagination so that it allows this value, or else there will be an error. And you could show how to do that.

By the wayn I think we should add 50 as a possible perPage option.

fzaninotto (changes requested):

fzaninotto (approved):

PR #8248, opened 2022-10-07
docs(yalc): updated doc for yarn link strategy

lsimone:
yarn link strategy is not working as described in the README (I guess because yarn in react-admin has been updated to v3 without taking into account the breaking changes in its `yarn link` command.

Furthermore, I had a lot of issues in trying to solve dependency inconsistencies (react, @mui, etc.) inside my project.  Finally I solved all my problems just using Use `yalc help` to see available commands. in 5 minutes, and I would like to help people not to waste time as I did :)  This is the strategy I would suggest to  users/contributors. Hope could help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,29 +205,15 @@ Both of these examples work without server - the API is simulated on the client-
 
 ### Testing Your Changes In Your App
 
-Using `yarn link`, you can have your project use a local checkout of the react-admin package instead of npm. This allows you to test react-admin changes in your app: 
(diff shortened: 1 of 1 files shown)


fzaninotto:
Before asking users to add a new tool, could you please describe the errors you get when using the `yarn link` command?

fzaninotto:
bump, please explain the problem you're trying to solve.

slax57:

PR #8259, opened 2022-10-12
Add context to have the ability to access form submit related props from the `SaveButton`

### LINKED ISSUE #7669: Adding `type="button"` to `SaveButton` overrides custom `onSubmit` behaviour on `TabbedForm`
**What you were expecting:**
Adding `type="button"` to the SaveButton simple changes the button's behaviour to not fire submit on enter, as per the [documentation](https://marmelab.com/react-admin/EditTutorial.html#submit-on-enter)

**What happened instead:**
Changing the `type` property seems to have additional side-effects. It doesn't seem to respect the a custom `onSubmit` function passed through to the form component. Instead it seems to revert to the default `onSubmit` behaviour

**Steps to reproduce:**
Access the [repro](https://codesandbox.io/s/optimistic-almeida-dmcfnv?file=/src/posts/PostEdit.tsx)
Navigate to edit post page
Change the text value
Press save

Observe that the default save behaviour is used, instead of our custom `onSubmit`

**Related code:**

  - https://codesandbox.io/s/github/marmelab/react-admin/tree/3.x/examples/simple (v3)

```
```

**Other information:**

**Environment**

* React-admin version:
* React version:
* Browser:

WiXSL:
Fixes #7669

Problems addressed in this PR having `SaveButton` of type "button":

Custom `onSubmit` provided by the user is not being called.
New features `sanitizeEmptyValues` and second event parameter in handleSubmit function are not being used.

- [X] Add submit related context
- [X] Use submit context inside `SaveButton`
- [x] Rename the context to `SubmitContext` (initially `useAugmentedForm` hook was responsible for returning the props, that is why it has that name)
- [x] Add tests


packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -88,13 +93,19 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
-        async values => {
+        async (values, event) => {
             let errors;
-            if (saveContext?.save) {
-                errors = await saveContext.save(values, {
+            const finalValues = sanitizeEmptyValues
+                ? sanitizeValues(values, record)
+                : values;
+
+            if ((mutationOptions || transform) && saveContext?.save) {
```
WiXSL, line 102: If mutationOptions || transform are getting passed, the SaveContext will be used first

slax57: I think there is 1 case not covered by this new logic: if you provide neither `mutationOptions` nor `transform` nor `onSubmit`, then you should still call the `SaveContext` if there is one. There should be a 3rd `else if` IMO.

slax57 (changes requested):
Otherwise this seems to be going in the right direction to me :slightly_smiling_face:

slax57 (changes requested):
All seems good to me! :slightly_smiling_face: 

Just one last change request: you have a linter warning in `packages/ra-ui-materialui/src/button/SaveButton.tsx` line 116

Also, @fzaninotto , I would appreciate having also your review on this one.

slax57 (approved):

fzaninotto:
Nice work, but the added context bothers me. It's a lot of complexity to handle a case that users can deal with otherwise (passing an `onClick` to the SaveButton). 

So I'm going to close this PR, and suggest to solve #7669 the other way around, by updating the documentation. 

Thanks @WiXSL and @slax57 for your work on this one anyway, and sorry for the late reply!

PR #8261, opened 2022-10-13
Fix displaying texts in Loading in case of arguments are not passed

alioguzhan:
Current behavior: 

If we omit the `loadingPrimary` and `loadingSecondary` props, the component still displays the translation keys (`ra.page.loading` and `ra.message.loading`) beneath the spinner. 

This fixes that.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Loading.tsx
+++ b/packages/ra-ui-materialui/src/layout/Loading.tsx
@@ -21,8 +21,8 @@ export const Loading = (props: LoadingProps) => {
                     className={LoadingClasses.icon}
                     color="primary"
                 />
-                <h1>{translate(loadingPrimary)}</h1>
-                <div>{translate(loadingSecondary)}</div>
+                {loadingPrimary ?? <h1>{translate(loadingPrimary)}</h1>}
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

WiXSL:
@alioguzhan we made a mistake here.
Your changes make the component not use the translation function anymore.
Actually, the problem you are trying to solve is the expected behavior.
You need to pass an empty string in `loadingPrimary` and `loadingSecondary` in order to not see the default messages

alioguzhan:
Hmm, I see. I can revert the changes, No problem. But, Isn't it weird that we have to pass empty strings to hide the messages? Maybe we can find another solution? Like passing a flag `hidePrimaryMessage` and `hideSecondaryMessage` ? Or an `options` flag. Examples:

1: 
```jsx
<Loading hidePrimary hideSecondary />
```

2:
```jsx 
<Loading options={{ hide: { primary: true, secondary: false} }} /> 
```

We can find an alternative better than:

```jsx
<Loading loadingPrimary="" loadingSecondary="" />
```

What do you think? @WiXSL

WiXSL:
We've already reverted it.

alioguzhan:
Fair enough. Thanks for your time! @WiXSL

PR #8262, opened 2022-10-13
Fix null value support in inputs

### LINKED ISSUE #8264: Null default value becomes blank string
**What you were expecting:**
I am expecting to get null value from form state when null default value is set.

**What happened instead:**
Blank string from form state returns.

**Steps to reproduce:**
1. Set default field value fo null
2. Get field value

**Related code:**
Set default value:
```
<SimpleForm
  defaultValues={{
    field: null,
  }}
>
```

Use it:
```
<FormDataConsumer>
  {({formData}) => (
    <div>
      field: {JSON.stringify(formData.field)}
    </div>
  )}
</FormDataConsumer>
```

**Result is:**
```
field: ""
```

**Environment**

* React-admin version: 4.4.1
* Last version that did not exhibit the issue (if applicable): 4.3.0
* React version: 18.2.0
* Browser: Chrome 106.0.5249.119

slax57:
**Problem:** 

Because of https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/form/getFormInitialValues.ts#L16, `null` values in records get transformed into empty strings `''`.
We should not change the records value like this, especially if the user does not change them in a form.

This also prevents to submit `null` as value.
Actually, adding a `parse` prop like `parse={value => (value === '' ? null : value)}` actually works, but **only if the value is changed by the user**.

**Solution:** 

- no longer modify the values in `getFormInitialValues`, and instead add a default `format` function to all inputs (via `useInput`) to change the formatted value to an empty string in that case, but not the actual (controlled) value.
- also add a default `parse` function to all inputs to parse all empty strings to `null`, because that's what most backends expect anyway

**Additional info**

- Also fixes #8264
- `sanitizeEmptyValues` is now rendered mostly useless when using only RA input components (such as TextInput, NumberInput, ...) ; it remains useful however if you are using custom inputs that do not use `useInput`
- This PR introduces a (very minor) BC break: custom inputs that do not use `useInput` will need to format their value to `''` when they receive `undefined` or `null` by themselves, otherwise this will raise console warnings about controlled/uncontrolled inputs


packages/ra-core/src/form/useInput.spec.tsx
```diff
@@ -151,28 +151,6 @@ describe('useInput', () => {
...
-                                />
-                            );
-                        }}
-                    </Input>
-                </Form>
-            </CoreAdminContext>
-        );
-        expect(screen.queryByDisplayValue('99')).toBeNull();
-    });
-
```
slax57, line 175: this test was actually a full duplicate

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -13,13 +13,19 @@ export const useApplyInputDefaultValues = (
...
+        formState,
+    } = useFormContext();
     const recordValue = get(record, source);
     const formValue = get(getValues(), source);
+    const { isDirty } = getFieldState(source, formState);
 
     useEffect(() => {
         if (defaultValue == null) return;
-        if (formValue == null && recordValue == null) {
+        if (formValue == null && recordValue == null && !isDirty) {
```
fzaninotto, line 28: Not checking `isDirty` makes the field reinitialize to default value when using a `parse` function to set null for empty values (as now in the `<TextInput>` by default).

packages/ra-ui-materialui/src/input/TextInput.tsx
```diff
@@ -49,7 +49,7 @@ export const TextInput = (props: TextInputProps) => {
     } = useInput({
         defaultValue,
         format,
-        parse,
+        parse: parse ?? defaultParse,
```
fzaninotto, line 52: I believe that a textInput should set the value to `null` when emptied. 

I believe it should be the same for e.g. `<SelectInput>`(which already does that when inside a `<ReferenceInput>`, and several other inputs, too. What do you think? I which case would we want an empty form value *not*  to translate to `null`?

slax57: I agree. Hence I generalized the solution to all inputs.

packages/ra-core/src/form/sanitizeEmptyValues.ts
```diff
@@ -7,7 +7,7 @@
 export const sanitizeEmptyValues = (values: any, record: any = {}): any => {
     const sanitizedValues = {};
     Object.keys(values).forEach(key => {
-        if (values[key] === '') {
+        if (values[key] == null || values[key] === '') {
```
fzaninotto, line 10: I don't understand why you updated this function

fzaninotto (changes requested):

packages/ra-core/src/form/sanitizeEmptyValues.ts
slax57: For react-admin inputs (such as TextInput), when the field is `undefined` from the dataProvider:
- if you touch the field (e.g. change it to `'foo'` then back to `''`)
- if your field has a defaultValue but you clear it
=> In both cases we will send `"field": null`, although the field was absent from what the dataProvider returned.

This is because **sanitizeEmptyValues** only triggers on `''` but not on `null`, however our components now return `null`.

IMO in this case **sanitizeEmptyValues** should still remove the values from what's been sent (hence my change of the implementation is needed).

I'll add a story to play with this to the branch in any case so that you can see for yourself :wink:

fzaninotto (approved):

PR #8272, opened 2022-10-17
Remove support for warnWhenUnsavedChange when react-router doesn't allow it

### LINKED ISSUE #8163: TypeError: navigator.block using the warnWhenUnsavedChanges prop in Form
After upgrading to version 4 I have an issue with using the warnWhenUnsavedChanges prop in Form. 

**What you were expecting:**

Able to use the warnWhenUnsavedChanges prop in the form component and get an alert when leaving a page that a has a form with data so no data is lost but instead saved before leaving the page.

**What happened instead:**

When trying to edit a record I get the Error "navigator.block is not a function".

**Steps to reproduce:**

1- User the Form Component and add the warnWhenUnsavedChanges.
2- Add TextInput fields.
3- Add info in the fields.
4- Get the error. 

**Environment**
Mac

* React-admin version: 4.3.2
* Last version that did not exhibit the issue (if applicable): 3*
* React version: 18.2.0
* Browser: Chrome

Uncaught TypeError: navigator.block is not a function
    at index.js:121476:33
    at commitHookEffectListMount (index.js:159924:26)
    at commitPassiveMountOnFiber (index.js:161700:13)
    at commitPassiveMountEffects_complete (index.js:161665:9)
    at commitPassiveMountEffects_begin (index.js:161652:7)
    at commitPassiveMountEffects (index.js:161640:3)
    at flushPassiveEffectsImpl (index.js:163813:3)
    at flushPassiveEffects (index.js:163758:14)
    at commitRootImpl (index.js:163709:5)
    at commitRoot (index.js:163456:5)

var useWarnWhenUnsavedChanges = function (enable, formRootPathname, control) {
    // react-router v6 does not yet provide a way to block navigation
    // This is planned for a future release
    // See https://github.com/remix-run/react-router/issues/8139
    var navigator = (0,react__WEBPACK_IMPORTED_MODULE_0__.useContext)(react_router_dom__WEBPACK_IMPORTED_MODULE_1__.UNSAFE_NavigationContext).navigator;
    var location = (0,react_router_dom__WEBPACK_IMPORTED_MODULE_1__.useLocation)();
    var translate = (0,_i18n__WEBPACK_IMPORTED_MODULE_2__.useTranslate)();
    var _a = (0,react_hook_form__WEBPACK_IMPORTED_MODULE_3__.useFormState)(control ? { control: control } : undefined), isSubmitSuccessful = _a.isSubmitSuccessful, isSubmitting = _a.isSubmitting, dirtyFields = _a.dirtyFields;
    var isDirty = Object.keys(dirtyFields).length > 0;
    var initialLocation = (0,react__WEBPACK_IMPORTED_MODULE_0__.useRef)(formRootPathname || location.pathname);
    (0,react__WEBPACK_IMPORTED_MODULE_0__.useEffect)(function () {
        if (!enable || !isDirty)
            return;
        var unblock = navigator.block(function (tx) {
            var newLocationIsInsideForm = tx.location.pathname.startsWith(initialLocation.current);
            if (!isSubmitting &&
                (newLocationIsInsideForm ||
                    isSubmitSuccessful ||
                    window.confirm(translate('ra.message.unsaved_changes')))) {
                unblock();
                tx.retry();
            }
        });
        return unblock;
    },

fzaninotto:
## Problem

React-router 6.4.0 removed support for `navigator.block`, which is necessary for the `warnWhenUnsavedChanges` feature (see https://github.com/remix-run/react-router/issues/9262).

This breaks existing applications using react-router >= 6.4 and a custom router. 

## Solution

The `warnWithUnsavedChanges` feature fails silently when using with react-router >= 6.4 

Closes #8163

### DIFF
--- a/docs/EditTutorial.md
+++ b/docs/EditTutorial.md
@@ -672,6 +672,8 @@ const Form = ({ onSubmit }) => {
 
 **Tip**: You can customize the message displayed in the confirm dialog by setting the `ra.message.unsaved_changes` message in your i18nProvider.
 
+**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
+
 ## Submit On Enter
 
 By default, pressing `ENTER` in any of the form inputs submits the form - this is the expected behavior in most cases. To disable the automated form submission on enter, set the `type` prop of the `SaveButton` component to `button`.
--- a/docs/Form.md
+++ b/docs/Form.md
@@ -230,3 +230,4 @@ export const TagEdit = () => (
 );
 ```
 
+**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
--- a/docs/SimpleForm.md
+++ b/docs/SimpleForm.md
@@ -376,6 +376,8 @@ export const TagEdit = () => (
 );
 ```
 
+**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
+
 ## Using Fields As Children
 
 The basic usage of `<SimpleForm>` is to pass [Input components](./Inputs.md) as children. For non-editable fields, you can pass `disabled` inputs, or even [Field components](./Fields.md). But since `<Field>` components have no label by default, you'll have to wrap your inputs in a `<Labeled>` component in that case:
--- a/docs/TabbedForm.md
+++ b/docs/TabbedForm.md
@@ -469,6 +469,8 @@ export const TagEdit = () => (
 );
 ```
 
+**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
+
 ## `<FormTab>`
 
 `<TabbedForm>` expect `<FormTab>` elements as children. `<FormTab>` elements accept four props:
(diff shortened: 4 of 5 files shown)


WiXSL (approved):

crates:
Is it possible to preempt the user's own onSubmit method (or alternatively, fired subsequent to the user's one) with one created by React Admin, so that when the warning prop is in use, it's that interstitial method that triggers first, thereby calling the warning up instead of submitting the form? That would not require any use of navigator.block, but would result in a similar outcome.

fzaninotto:
@crates I'm not sure I understand. Are you suggesting that  `warnWhenUnsavedChange` auto-submits the form when the user leaves the page without submitting? That would be very unexpected for most users.

crates:
No, you are totally right. That suggestion made no sense.

How about a custom router component provided by React Admin? Something like what they are discussing on this thread?

https://stackoverflow.com/questions/69953377/react-router-v6-how-to-use-navigate-redirection-in-axios-interceptor

TrejGun:
that actually still works with 
https://github.com/remix-run/react-router/issues/8139#issuecomment-1023105785

and custom history
```
import { FC } from "react";
import { unstable_HistoryRouter as HistoryRouter } from "react-router-dom";
import { history } from "history";


const App: FC = () => {
  return (
    <HistoryRouter history={history}>
      <YOUR CODE HERE>
    </HistoryRouter>
  );
};

export default App;
```

crates:
@TrejGun - Agreed; that's what we're using on our project also, with React Router 6.4

slax57 (approved):

fzaninotto:
> How about a custom router component provided by React Admin? Something like what they are discussing on this thread?
> 
> https://stackoverflow.com/questions/69953377/react-router-v6-how-to-use-navigate-redirection-in-axios-interceptor

@crates reimplementing a custom router is outside of the scope of react-admin. We have enough work without it ;)

> that actually still works with 
> https://github.com/remix-run/react-router/issues/8139#issuecomment-1023105785
> 
> and custom history

@TrejGun You're right, that's why the PR description says

> This breaks existing applications using react-router >= 6.4 *and a custom router*.

So we have to prevent the app from breaking in that case, and that's why this PR is necessary.

PR #8275, opened 2022-10-18
Fix improve SaveButton enable state

### LINKED ISSUE #8258: `ArrayInput` makes edited form always dirty on input blur
1- Goto to the simple example project: https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple?file=src%2Findex.tsx
2- Copy "authors" `ArrayInput` from `PostEdit` form and paste it inside `CommentEdit` form.
3- Edit any Comment record.
4- Focus any input and then blur it

Expected: 
`SaveButton` state is disabled

Result:
`SaveButton` state is enabled

WiXSL:
Fixes #8258

**PROBLEM**

`SaveButton` enable state may be true, even though no inputs have been changed.

This is because `useFormState().isDirty` might differ from `useFormState().dirtyFields` (https://github.com/react-hook-form/react-hook-form/issues/4740)
This is the expected behavior of `react-hook-form` and it may happen when you have a missing `defaultValue` for one of the inputs

· `dirtyFields` will mark the field dirty or not individually.
· `isDirty` will always do a deep equal against your defaultValues to determined form is dirty or not.
https://github.com/react-hook-form/react-hook-form/discussions/7860#discussioncomment-2208434

This is key, since `defaultValue` prop might be `undefined`, and in some cases you end up with `isDirty` being `true` but `dirtyFields` being an empty object. Like in the test I've added in this PR, plus the ones I've fixed in this PR

**SOLUTION**

Use `useFormState().dirtyFields` to enable/disable `SaveButton` instead of `useFormState().isDirty`

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -12,7 +12,12 @@ import {
 import { SaveButton } from './SaveButton';
 import { SimpleForm, Toolbar } from '../form';
 import { Edit } from '../detail';
-import { TextInput } from '../input';
+import {
+    TextInput,
+    ArrayInput,
+    SimpleFormIterator,
+    NumberInput,
+} from '../input';
 import { AdminContext } from '../AdminContext';
 
 const invalidButtonDomProps = {
@@ -397,6 +402,60 @@ describe('<SaveButton />', () => {
         await waitFor(() =>
             expect(screen.getByLabelText('ra.action.save')['disabled']).toEqual(
                 false
+            )
+        );
+    });
+
+    it('should not be enabled if no inputs have changed', async () => {
+        render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <SimpleForm
+                    resource="myresource"
+                    onSubmit={jest.fn}
+                    defaultValues={{
+                        test: 'test',
+                        arr: [
+                            {
+                                id: 123,
+                                foo: 'bar',
+                                nested: {
+                                    deep: '1',
+                                },
+                            },
+                            {
+                                id: 456,
+                                foo: 'baz',
+                                nested: {
+                                    deep: '2',
+                                },
+                            },
+                        ],
+                    }}
+                >
+                    <TextInput source="test" />
+                    <ArrayInput resource="foo" source="arr">
+                        <SimpleFormIterator>
+                            <NumberInput source="id" />
+                            <ArrayInput resource="bar" source="arr.nested">
+                                <SimpleFormIterator>
+                                    <NumberInput source="deep" />
+                                </SimpleFormIterator>
+                            </ArrayInput>
+                        </SimpleFormIterator>
+                    </ArrayInput>
+                </SimpleForm>
+            </AdminContext>
+        );
+
+        const testInput = screen.getByLabelText(
+            'resources.myresource.fields.test'
+        );
+        fireEvent.focus(testInput);
+        fireEvent.blur(testInput);
+
+        await waitFor(() =>
+            expect(screen.getByLabelText('ra.action.save')['disabled']).toEqual(
+                true
             )
         );
     });
--- a/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
@@ -414,37 +414,12 @@ describe('<SaveButton />', () => {
                     onSubmit={jest.fn}
                     defaultValues={{
                         test: 'test',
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -65,7 +65,8 @@ export const SaveButton = <RecordType extends RaRecord = any>(
     const translate = useTranslate();
     const form = useFormContext();
     const saveContext = useSaveContext();
-    const { isDirty, isValidating, isSubmitting } = useFormState();
+    const { dirtyFields, isValidating, isSubmitting } = useFormState();
+    const isDirty = Object.keys(dirtyFields).length;
```
WiXSL, line 69: `useFormState().isDirty` might differ from `useFormState().dirtyFields` (https://github.com/react-hook-form/react-hook-form/issues/4740)

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -400,4 +405,58 @@ describe('<SaveButton />', () => {
...
+                                },
+                            },
+                        ],
+                    }}
+                >
+                    <TextInput source="test" />
+                    <ArrayInput resource="foo" source="arr">
+                        <SimpleFormIterator>
+                            <NumberInput source="id" />
+                            <ArrayInput resource="bar" source="arr.nested">
```
slax57, line 439: Something seems weird in your test: `arr.nested` is not an array

packages/ra-ui-materialui/src/input/DateTimeInput.spec.tsx
```diff
@@ -223,7 +223,7 @@ describe('<DateTimeInput />', () => {
             const onSubmit = jest.fn();
             render(
                 <AdminContext dataProvider={testDataProvider()}>
-                    <SimpleForm onSubmit={onSubmit}>
+                    <SimpleForm mode="onBlur" onSubmit={onSubmit}>
```
slax57, line 226: Now this worries me a bit. Why did you need to change this test?

slax57 (changes requested):

packages/ra-ui-materialui/src/input/DateTimeInput.spec.tsx
WiXSL: Well, it seems that the save button wasn't being enabled, which, should be right, I think it was wrong before

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
WiXSL: You are right

packages/ra-ui-materialui/src/input/DateTimeInput.spec.tsx
slax57: It's really weird that the test passed before this PR then.

But you convinced me your fix is right. Let's keep it.

WiXSL: Actually, my explanation of why this test should be right wasn't precise, is not relevant that the input is required, but that is being cleared and back to not being dirty, which is what matters to the SaveButton disable state

packages/ra-ui-materialui/src/button/SaveButton.tsx
fzaninotto: Could you add that as a comment above this line? It'd help future us understand why we're not using `isDirty` here. 

also, I'm nipticking, but I'd prefer

```jsx
const isDirty = Object.keys(dirtyFields).length > 0;
```

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
```diff
@@ -400,4 +405,62 @@ describe('<SaveButton />', () => {
...
+                                ],
+                            },
+                        ],
+                    }}
+                >
+                    <TextInput source="test" />
+                    <ArrayInput resource="foo" source="arr">
+                        <SimpleFormIterator>
+                            <NumberInput source="id" />
+                            <ArrayInput resource="bar" source="arr.nested">
```
fzaninotto, line 443: Why do you need two tested arrayinputs to test that case?

fzaninotto (changes requested):
Nice work!

packages/ra-ui-materialui/src/button/SaveButton.spec.tsx
WiXSL: You're right! I can simplify it

fzaninotto (approved):

PR #8291, opened 2022-10-24
Fix `<CheckboxGroupInput options>` prop is ignored

fzaninotto:
- [x] Fix CheckboxGroupInput doc doesn't mention source format
- [x] Improve usage guidelines
- [x] Add screenshots to the documentation
- [x] Add stories to test all props
- [x] Fix `options` prop isn't taken into account


docs/CheckboxGroupInput.md
```diff
@@ -58,17 +122,46 @@ const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
...
+    const record = useRecordContext();
+    return <span>{record.first_name} {record.last_name}</span>;
+}
+
 <CheckboxGroupInput source="gender" choices={choices} optionText={<FullNameField />}/>
 ```
 
+## `optionValue`
+
+You can customize the properties to use for the option value (instead of the default `_id`) thanks to the `optionValue` prop:
```
slax57, line 143:
```suggestion
You can customize the properties to use for the option value (instead of the default `id`) thanks to the `optionValue` prop:
```

docs/CheckboxGroupInput.md
```diff
@@ -58,17 +122,46 @@ const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
...
+const choices = [
+    { _id: 123, name: 'Leo Tolstoi' },
+    { _id: 456, name: 'Jane Austen' },
+];
+<CheckboxGroupInput source="author_id" choices={choices} optionValue="_id" />
+```
+
+## `row`
+
+By default, the checkboxes are displayed in a row. You can change that an let react-admin render one choice per row by setting the `row` prop to `false`:
```
slax57, line 155:
```suggestion
By default, the checkboxes are displayed in a row. You can change that and let react-admin render one choice per row by setting the `row` prop to `false`:
```

slax57 (changes requested):
One last thing, you should probably change the title of the PR since you are doing more than just documenting the source format.

Otherwise, this is a real improvement of the docs, I love it!

fzaninotto:
@slax57 Thanks for your review! I've taken it all into account.

WiXSL (approved):

slax57 (approved):

PR #8296, opened 2022-10-24
Fix `AutocompleteInput` displays empty choice when `validate={required()}`

slax57:
`AutocompleteInput` should not display the empty choice when `validate={required()}`


WiXSL (changes requested):
`useInput` hook already resolves which `isRequired` is going to be used and returns it in `isRequired`.
So I think this can be changed to simply `isRequired={isRequired}`

https://github.com/marmelab/react-admin/blob/6a298cb845c651937117a9eaeb079b22da8e503c/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx#L541

fzaninotto (approved):

PR #8305, opened 2022-10-25
Fix `AutocompleteInput` should only add an empty option when there is an `emptyText`

### LINKED ISSUE #8293: AutocompleteInput doesn't show 'No options' when empty value is allowed
**What you were expecting:**
`AutocompleteInput` should display both 'No options' and 'empty' when the empty value is allowed (and `emptyText="empty"`) and the query returns no result.

**What happened instead:**
`AutocompleteInput` only shows 'empty' among the options.

**Steps to reproduce:**
1. Go to https://stackblitz.com/edit/github-varpfb?file=src/posts/PostCreate.tsx
2. Log in as **admin**
3. Go to PostCreate
4. Click '+' to add an author
5. Select 'Logan Schowalter' as the author
6. In the same AutocompleteInput, type in 'test'

**Environment**

* React-admin version: 4.4.4
* Last version that did not exhibit the issue (if applicable): ???
* React version: 18
* Browser: Chrome

WiXSL:
Fixes #8293

### DIFF
--- a/docs/AutocompleteInput.md
+++ b/docs/AutocompleteInput.md
@@ -100,9 +100,7 @@ See [Using in a `ReferenceInput>`](#using-in-a-referenceinput) below for more in
 
 ## `emptyText`
 
-If the input isn't required (using `validate={required()}`), users can select an empty choice with an empty text `''` as label.
-
-You can override that label with the `emptyText` prop.
+If the input isn't required (using `validate={required()}`), and you need a choice to represent the empty value, set `emptyText` prop and a choice will be added at the top, with its value as label.
 
 ```jsx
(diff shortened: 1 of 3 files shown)


slax57 (changes requested):
Other than that, it seems to work great!! I haven't been able to find a case where it would not work, so for me it's all good! Nice job :slightly_smiling_face:

slax57:
This PR will make react-admin's `AutocompleteInput` deal with empty values the same way MUI's `Autocomplete` does by default, and will allow to fix https://github.com/marmelab/react-admin/issues/8293

slax57 (approved):

PR #8306, opened 2022-10-25
ReferenceOneField accept sort & filter props

### LINKED ISSUE #8304: Allow to apply sorting and filtering on ReferenceOneField
The doc currently says:

> This field fetches a one-to-one relationship, e.g. the details of a book, when using a foreign key on the distant resource.

But actually it could be used to display only one record of many references (eg the latest reference of a record). To choose the right record to display, filtering & sorting might be useful. For example, to get the latest reference, I would like to sort DESC on a createdAt field. But I can't because the ASC order is hard coded [here](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/field/useReferenceOneFieldController.tsx#L49-L50).

Would you accept a PR on this feature?

nicgirault:
Fix #8304

### DIFF
--- a/docs/ReferenceOneField.md
+++ b/docs/ReferenceOneField.md
@@ -47,13 +47,37 @@ const BookShow = () => (
 
 **Tip**: As with `<ReferenceField>`, you can call `<ReferenceOneField>` as many times as you need in the same component, react-admin will only make one call to `dataProvider.getManyReference()` per reference.
 
+**Tip**: `<ReferenceOneField>` can also be used to display one record of a one-to-many relationship. Use `sort` and/or `filter` props to select the appropriate record to display. The first record will be displayed.
+
+```jsx
+const BookShow = () => (
+    <Show>
+        <SimpleShowLayout>
+            <TextField source="title" />
+            <DateField source="published_at" />
+            <ReferenceOneField
+                label="Latest cool review"
+                reference="book_reviews"
+                target="book_id"
+                sort={{ field: "createdAt", order: "DESC" }}
+                filter={{ rating: 5 }}
+            >
+                <TextField source="title" />
+            </ReferenceOneField>
+        </SimpleShowLayout>
+    </Show>
+);
+```
+
 ## Properties
 
 | Prop         | Required | Type               | Default                          | Description                                                                         |
-| ------------ | -------- | ------------------ | -------------------------------- | ----------------------------------------------------------------------------------- |
-| `children`   | Optional | `Element`          | -                                | The Field element used to render the referenced record                              |
-| `link`      | Optional | `string | Function` | `edit`   | Target of the link wrapping the rendered child. Set to `false` to disable the link.                                 |
-| `reference`  | Required | `string`           | -                                | The name of the resource for the referenced records, e.g. 'book_details'                   |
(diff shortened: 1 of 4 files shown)


docs/ReferenceOneField.md
```diff
@@ -5,7 +5,7 @@ title: "The ReferenceOneField Component"
 
 # `<ReferenceOneField>`
 
-This field fetches a one-to-one relationship, e.g. the details of a book, when using a foreign key on the distant resource.
+This field fetches a one-to-one relationship or the first record of a one-to-many relationship, e.g. the details of a book, when using a foreign key on the distant resource.
```
slax57, line 8: I would not mention this use-case so early in the doc, since it's after all a quite specialized use-case, and we want this first paragraph to introduce the most common use-case of the component only (let's keep things simple :wink: ).

Instead I'd mention this ability later, for instance as a **Tip**.

docs/ReferenceOneField.md
```diff
@@ -18,7 +18,7 @@ This field fetches a one-to-one relationship, e.g. the details of a book, when u
                        └──────────────┘
 ```
 
-`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the foreign key (`book_id`). Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` to fetch the related details, and takes the first one.
+`<ReferenceOneField>` behaves like `<ReferenceManyField>`: it uses the current `record` (a book in this example) to build a filter for the book details with the foreign key (`book_id`) and optional additional filters. Then, it uses `dataProvider.getManyReference('book_details', { target: 'book_id', id: book.id })` to fetch the related details, and takes the first one.
```
slax57, line 21: Same, let's not mention the 'optional additional filters' so early, and keep them for later in the doc.

docs/ReferenceOneField.md
slax57: Also, you need to update the `## Properties` section too, to document these new props

packages/ra-ui-materialui/src/field/types.ts
```diff
@@ -5,6 +5,10 @@ import { TableCellProps } from '@mui/material/TableCell';
 
 type TextAlign = TableCellProps['align'];
 type SortOrder = 'ASC' | 'DESC';
+export interface SortPayload {
+    field: string;
+    order: SortOrder;
+}
```
slax57, line 11: I would recommend avoiding duplication of the `SortPayload` interface, and using the one from `ra-core` instead. It does not force the SortOrder to be `'ASC' | 'DESC'`, but actually we shouldn't force it since it's dependent on the dataProvider.

slax57 (changes requested):
Otherwise, everything seems good, thanks for your contribution!

slax57:
Oh and one last thing, since it's a new feature, your PR should target the `next` branch instead of `master`. Can you change it?

nicgirault:
@slax57 thanks for your review. Are you satisfied with the latest changes?

slax57 (approved):
Yes that's great, thank you so much!

PR #8314, opened 2022-10-26
Fix `RichTextInput` doesn't update when record field updates

### LINKED ISSUE #8252: RichTextInput does not update when record field updates
We are migrating the old quill `richTextInput` to have a more "standard" react-admin component but it does not behave as for other RA components.

**What you were expecting:**

When its record field updates, it should change the shown input box content

**What happened instead:**

Nothing happens (`TextInput` updates instead => see codesandbox example)

**Steps to reproduce:**

- Bind a RichTextInput to a field in a form
- Update record's value (I used a `TextInput` to do it but it can be done in any way)

**Related code:**

In [codesandbox PostEdit](https://codesandbox.io/s/dreamy-banzai-x609ts?file=/src/posts/PostEdit.tsx) module I just added 2 new `TextInput` with the same source and I triggered a record update by changing the value of the first one:

#### expected

I would expect to see the value of the second `TextInput` and of the `RichTextInput` updated

#### actual

Just the second `TextInput` updates

```
          {/* this RichTextInput does not update when 'body' record value updates! */}
          <RichTextInput
            source="body"
            label=""
            validate={required()}
            fullWidth
          />
          {/* INSTEAD, this TextInput DOES update when 'body' record value updates! */}
          <TextInput source="body" label="body 1" fullWidth />
          <TextInput source="body" label="body 2" fullWidth />

```

**Other information:**

**Environment**

* React-admin version: ^4.4.1
* React version: ^17.0.0
* Browser: Chrome

WiXSL:
Fixes #8252

- [x] Update `RichTextInput` value on field update
- [x] Add test

### DIFF
--- a/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
+++ b/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
@@ -14,12 +14,12 @@ describe('<RichTextInput />', () => {
             );
         });
 
-        const newRecord = { id: 123, body: '<h1>Goodby world!</h1>' };
+        const newRecord = { id: 123, body: '<h1>Goodbye world!</h1>' };
         rerender(<Basic record={newRecord} />);
 
         await waitFor(() => {
             expect(container.querySelector('#body').innerHTML).toEqual(
-                '<h1>Goodby world!</h1>'
+                '<h1>Goodbye world!</h1>'
             );
         });
     });
--- a/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
+++ b/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
@@ -1,14 +1,6 @@
 import * as React from 'react';
 import expect from 'expect';
-import { screen, render, waitFor } from '@testing-library/react';
-import {
-    RecordContextProvider,
-    Form,
-    testDataProvider,
-    CoreAdminContext,
-} from 'ra-core';
-
-import { RichTextInput } from './RichTextInput';
+import { render, waitFor } from '@testing-library/react';
 import { Basic } from './RichTextInput.stories';
 
 describe('<RichTextInput />', () => {
--- a/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
+++ b/packages/ra-input-rich-text/src/RichTextInput.spec.tsx
@@ -1,0 +1,34 @@
+import * as React from 'react';
+import expect from 'expect';
+import { screen, render, waitFor } from '@testing-library/react';
+import {
+    RecordContextProvider,
+    Form,
+    testDataProvider,
+    CoreAdminContext,
+} from 'ra-core';
+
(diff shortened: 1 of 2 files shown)


packages/ra-input-rich-text/src/RichTextInput.tsx
```diff
@@ -108,10 +108,19 @@ export const RichTextInput = (props: RichTextInputProps) => {
 
     const { error, invalid, isTouched } = fieldState;
 
+    useEffect(() => {
+        if (!editor) return;
+
+        editor.commands.setContent(field.value, false, {
```
fzaninotto, line 114: I'm afraid this leads to an infinite query. Besides, this doesn't solve what the issue is about: changing the field value when the *record* changes.

fzaninotto (changes requested):

packages/ra-input-rich-text/src/RichTextInput.tsx
WiXSL: By passing `false` to the second argument of `editor.commands.setContent` command (emmitUpdate?: boolean), you prevent and infinite query.
I think it is a solution for the reported issue, with this change, the content of the `RichTextInput` updates correctly when the field value changes

WiXSL: I've just added a test for it

packages/ra-input-rich-text/src/RichTextInput.spec.tsx
```diff
@@ -0,0 +1,26 @@
...
+        });
+
+        const newRecord = { id: 123, body: '<h1>Goodby world!</h1>' };
+        rerender(<Basic record={newRecord} />);
+
+        await waitFor(() => {
+            expect(container.querySelector('#body').innerHTML).toEqual(
+                '<h1>Goodby world!</h1>'
+            );
+        });
```
slax57, line 24:
```suggestion
        const newRecord = { id: 123, body: '<h1>Goodbye world!</h1>' };
        rerender(<Basic record={newRecord} />);

        await waitFor(() => {
            expect(container.querySelector('#body').innerHTML).toEqual(
                '<h1>Goodbye world!</h1>'
            );
        });
```

packages/ra-input-rich-text/src/RichTextInput.tsx
```diff
@@ -108,10 +108,19 @@ export const RichTextInput = (props: RichTextInputProps) => {
...
+            preserveWhitespace: true,
+        });
+    }, [editor, field.value]);
+
     useEffect(() => {
         if (!editor) return;
 
         editor.setOptions({
+            content: field.value,
```
slax57, line 123: did you really mean to keep this line? I don't think it's needed (and it causes a linter warning in the useEffect dependency array)

slax57 (changes requested):

fzaninotto (approved):

Issue #8312, opened 2022-10-26
Wrong documentation example for localStorageStore


fzan:
**What you were expecting:**
localStorageStore accepting a numeric value as stated in documentation
**What happened instead:**
Using a numeric value (eg:0) causes store invalidation. The store invalidation calls localStorage.clean that wipe also other data that maybe are unrelated directly with react admin.

**Steps to reproduce:**
1) do not use typescript in your project
2) follow the example from
https://github.com/marmelab/react-admin/blob/master/docs/Store.md

```
import { Admin, Resource, localStorageStore } from 'react-admin';

const STORE_VERSION = 2; //<-- this is wrong!

const App = () => (
    <Admin dataProvider={dataProvider} store={localStorageStore(STORE_VERSION)}>
        <Resource name="posts" />
    </Admin>
);
```
so, because on LocalStorageStore.ts you check
`if (storedVersion && storedVersion !== version) {`

this is inconsistent because you are saving storedVersion on localStorage as a string, but version in memory is numeric.

note that the session will be invalidated if you refresh the page.

as you declared in localStorageStore.d.ts:
`export declare const localStorageStore: (version?: string, appKey?: string) => Store;`
With typescript this should trigger a type mismatch

my conclusions are:
1) the documentation is wrong ( STORE_VERSION should be "2" instead of 2) and this is the original problem
2) i also want to suggest that you should not clean all the localStorage, (by calling localStorage.clean()  ) on version verify, but maybe i can suggest to add a prefix to RaStore variables to clean (for example, "RaStore.myVariable" or "RaStore.2.myVariable") because you cannot be sure that the localStorage doesn't have some custom logic values (as my case) that needs to be preserved 👨‍🔧

Thank you as usual for your great work 🙏

* React-admin version: 4.4.4 
* Last version that did not exhibit the issue (if applicable):
* React version: not related
* Browser: Chrome
* Stack trace (in case of a JS error):

fzaninotto:
Thanks for your ticket!

1. You're right. Would you like to open a PR to fix the doc?
2. You're also right. We are indeed using a prefix, so on version change, we should find all keys with that prefix and delete only these one (kind of like what we do in removeItems). Would you like to work on a second PR for this?

PR #8315, opened 2022-10-26
localStorageStore now delete only items prefixed on version change

fzan:
fix issue on #8312

### DIFF
--- a/packages/ra-core/src/store/localStorageStore.spec.ts
+++ b/packages/ra-core/src/store/localStorageStore.spec.ts
@@ -1,6 +1,6 @@
 import expect from 'expect';
 
-import { localStorageStore } from './localStorageStore';
+import { localStorageStore, getStorage } from './localStorageStore';
 
 describe('localStorageStore', () => {
     it('should allow to store and retrieve a value', () => {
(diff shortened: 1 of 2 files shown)


WiXSL (approved):

packages/ra-core/src/store/localStorageStore.ts
```diff
@@ -83,7 +83,12 @@ export const localStorageStore = (
             if (localStorageAvailable) {
                 const storedVersion = getStorage().getItem(`${prefix}.version`);
                 if (storedVersion && storedVersion !== version) {
-                    getStorage().clear();
+                    const storage = getStorage();
+                    Object.keys(storage).forEach(key => {
+                        if (key.startsWith(`${prefix}`)) {
```
slax57, line 88: suggestion:

```suggestion
                        if (key.startsWith(prefix)) {
```

packages/ra-core/src/store/localStorageStore.spec.ts
```diff
@@ -40,4 +40,14 @@ describe('localStorageStore', () => {
...
+    describe('reset preserving items outside Ra-Store', () => {
+        it('should reset the store', () => {
+            const store = localStorageStore();
+            store.setItem('foo', 'bar');
+            getStorage().setItem('baz', 'baz'); //set custom item in localstorage
+            store.reset();
+            expect(getStorage().getItem('baz')).toEqual('baz'); //expect not to be wiped on store reset
+            expect(store.getItem('foo')).toEqual(undefined);
+        });
+    });
```
slax57, line 52: This test is good (and should be kept), but you are actually not testing the code you added. You should add a test about version change.

slax57 (changes requested):

packages/ra-core/src/store/localStorageStore.ts
fzan: oops, sorry my bad 😅

fzan (commented):
More appropriate test case
Removed use of unneeded literal

fzaninotto (approved):

fzaninotto:
Thanks!

PR #8316, opened 2022-10-26
[Doc] Improve Input docs

fzaninotto:


docs/RichTextInput.md
```diff
@@ -155,3 +113,57 @@ export const MyEditorOptions = {
 	],
 };
 ```
+
+## `toolbar`
+
+The `<RichTextInput>` component has a `toolbar` prop that accepts a `ReactNode`.
+
+You can leverage this to change the buttons [size](#api):
```
WiXSL, line 121: Wrong anchor: ... `[size]`(#api): 
Could be `https://mui.com/material-ui/api/button` ?

WiXSL (changes requested):

WiXSL (changes requested):

docs/RichTextInput.md
fzaninotto: You're right, I removed the anchor.

docs/DateInput.md
```diff
@@ -5,20 +5,41 @@ title: "The DateInput Component"
...
+## Usage
+
 ```jsx
 import { DateInput } from 'react-admin';
 
 <DateInput source="published_at" />
 ```
 
-`<DateInput>` also accepts the [common input props](./Inputs.md#common-input-props).
+The field value must be a string with the pattern `YYYY-MM-DD` (ISO 8601), e.g. `'2022-04-30'`.
```
slax57, line 22: I believe it also accepts JS `Date` objects, doesn't it? (even though in this case it requires a custom `parse` function)

docs/DateTimeInput.md
```diff
@@ -5,16 +5,27 @@ title: "The DateTimeInput Component"
...
+## Usage
+
 ```jsx
 import { DateTimeInput } from 'react-admin';
 
 <DateTimeInput source="published_at" />
 ```
 
-`<DateTimeInput>` also accepts the [common input props](./Inputs.md#common-input-props).
+The input value must be a valid date string, i.e. a string understood by JavasSript's [`Date.parse()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse). Strings with [the ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601) 'yyyy-MM-ddThh:mm' are the most common (e.g. `'2022-04-30T12:30'`). The field value may contain a timezone offset, e.g. `'2022-04-30T12:30+02:00'`. If no timezone is specified, the browser's timezone is used.
```
slax57, line 22: Same, I believe it also accepts `Date` objects

docs/FileInput.md
```diff
@@ -5,56 +5,165 @@ title: "The FileInput Component"
...
+
+```jsx
+import { FileInput, FileField } from 'react-admin';
+
+<FileInput source="attachments">
+    <FileField source="src" title="title" />
+</FileInput>
+```
+
+`<FileInput>` uses its child component to give a preview of the files. `<FileInput>` renders it child once per file, inside a `<RecordContext>`, so the child can be a Field component. The default [`<FileField>`](./FileField.md) renders the name of the file(s), with a hyperlink.
```
slax57, line 24:
```suggestion
`<FileInput>` uses its child component to give a preview of the files. `<FileInput>` renders its child once per file, inside a `<RecordContext>`, so the child can be a Field component. The default [`<FileField>`](./FileField.md) renders the name of the file(s), with a hyperlink.
```

docs/ImageInput.md
```diff
@@ -5,53 +5,149 @@ title: "The ImageInput Component"
 
 # `<ImageInput>`
 
-`<ImageInput>` allows to upload some pictures using [react-dropzone](https://github.com/okonet/react-dropzone).
+`<ImageInput>` allows editing and uploading iumages (png, jpg, gif, etc). It is powered by [react-dropzone](https://github.com/okonet/react-dropzone).
```
slax57, line 8:
```suggestion
`<ImageInput>` allows editing and uploading images (png, jpg, gif, etc). It is powered by [react-dropzone](https://github.com/okonet/react-dropzone).
```

docs/ImageInput.md
```diff
@@ -5,53 +5,149 @@ title: "The ImageInput Component"
...
+
+```jsx
+import { ImageInput, ImageField } from 'react-admin';
+
+<ImageInput source="pictures" label="Related pictures">
+    <ImageField source="src" title="title" />
+</ImageInput>
+```
+
+`<ImageInput>` uses its child component to give a preview of the files. `<ImageInput>` renders it child once per file, inside a `<RecordContext>`, so the child can be a Field component. The default [`<ImageField>`](./ImageField.md) renders the name of the file(s), with a hyperlink.
```
slax57, line 22: don't think there is an hyperlink, only the image is displayed I believe

docs/ImageInput.md
```diff
@@ -5,53 +5,149 @@ title: "The ImageInput Component"
...
+- '.png,.gif,.jpg,.jpeg' 
+- 'image/*'
 
-`<ImageInput>` delegates the preview of currently selected images to its child. `<ImageInput>` clones its child as many times as there are selected images, passing the image as the `record` prop. To preview a simple list of image thumbnails, you can use `<ImageField>` as child, as follows:
+If left empty, all file types are accepted (even non-image types).
+
+## `children`
+
+`<ImageInput>` delegates the preview of currently selected images to its child. `<ImageInput>` clones its child component once per file, inside a [`<RecordContext>`](./useRecordContext.md), so the child can be a [Field](./Fields.md) component. The default [`<ImageField>`](./ImageField.md) renders the name of the file(s), with a hyperlink.
```
slax57, line 86: same

slax57 (changes requested):

docs/DateInput.md
fzaninotto: Yes, but I don't see the point in mentioning it, as the data comes from an API response - and therefore was serialized to string. Unless some data providers return dates as `Date` objects?

docs/ImageInput.md
fzaninotto: Right, fixed

docs/ImageInput.md
fzaninotto: Fixed

docs/DateTimeInput.md
fzaninotto: Here, I think it makes sense to mention it, so I modified the doc.

WiXSL (approved):

docs/DateInput.md
slax57: Good point. I don't know any dataProvider that do that, my comment was based on the code only.

slax57 (approved):

PR #8319, opened 2022-10-27
Fix `List` page display on dataProvider error

### LINKED ISSUE #8241: List show nothing when http error after upgraded to v4
**What you were expecting:**

`<List>` should at least show some normal components when error happens, such as filter, grid (with old data), pagination and others which not cause the error.

**What happened instead:**

After upgraded to v4, if HTTP error happens, all `<List>` show nothing like abnormal.

**Steps to reproduce:**

Something same as in #7415, just return a 500 error from any loading in list.

**Related code:**

https://github.com/marmelab/react-admin/commit/5cc275af36120e4153eafe6a884c5578a0f2ea3b

```
if (error) {
    return null;
}
```

This logic is so simple and rough to fix the old issue.

**Other information:**

**Environment**

* React-admin version: 4.4.0
* Last version that did not exhibit the issue (if applicable): 3.19.0
* React version: 17.0.2
* Browser: Chrome 105.0.5195.125

Console error:

```
react_devtools_backend.js:4026 Error
    at new HttpError (HttpError.ts:15:1)
    at fetch.ts:50:1
```

HTTP response:

```json
{"timestamp":1665035231396,"error":{"key":"error.unknown"}}
```

WiXSL:
Fixes #8241

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -9,7 +9,7 @@ import { ComponentPropType, useListContext, RaRecord } from 'ra-core';
 
 import { Title, TitlePropType } from '../layout/Title';
 import { ListToolbar } from './ListToolbar';
-import { Pagination as DefaultPagination } from './pagination';
+import { Pagination as DefaultPagination, PaginationLimit } from './pagination';
 import { ListActions as DefaultActions } from './ListActions';
 import { Empty } from './Empty';
 
@@ -50,10 +50,6 @@ export const ListView = <RecordType extends RaRecord = any>(
         return null;
     }
 
-    if (error) {
-        return null;
-    }
-
     const renderList = () => (
         <div className={ListClasses.main}>
             {(filters || actions) && (
@@ -70,7 +66,7 @@ export const ListView = <RecordType extends RaRecord = any>(
                       })
                     : children}
             </Content>
-            {pagination !== false && pagination}
+            {error ? <PaginationLimit /> : pagination !== false && pagination}
         </div>
     );
 


packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -70,7 +66,7 @@ export const ListView = <RecordType extends RaRecord = any>(
                       })
                     : children}
             </Content>
-            {pagination !== false && pagination}
+            {error ? <PaginationLimit /> : pagination !== false && pagination}
```
fzaninotto, line 69: I don't think you should say 'no results found' in case of error.

WiXSL: Should I add an equivalent component to display an error message?

WiXSL:
> How does it behave if the error comes after a successful render, e.g. at page 2?

It shows the notification error and gets back to render the previous page

packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -70,7 +66,11 @@ export const ListView = <RecordType extends RaRecord = any>(
                       })
                     : children}
             </Content>
-            {pagination !== false && pagination}
+            {error ? (
+                <PaginationLimit message="ra.notification.data_provider_error" />
```
slax57, line 70: It feels weird to me to render a component named 'pagination' here, since we don't even know if the data is supposed to be paginated or not.

I understand that the code of the component you want to display is very similar to the code of `PaginationLimit`, but to me they have different purposes, so they should be two different components (even if the code is duplicated).

packages/ra-ui-materialui/src/list/List.spec.tsx
```diff
@@ -207,4 +207,28 @@ describe('<List />', () => {
...
+                    <List resource="posts">
+                        <Datagrid />
+                    </List>
+                </ThemeProvider>
+            </CoreAdminContext>
+        );
+        await waitFor(() => {
+            expect(
+                screen.getByText('ra.notification.data_provider_error')
+            ).not.toBeNull();
```
slax57, line 231: You should either have

```suggestion
            expect(
                screen.queryByText('ra.notification.data_provider_error')
            ).not.toBeNull();
```

Or simply

```tsx
            expect(
                screen.getByText('ra.notification.data_provider_error')
            );
```

(which I prefer)

But having both `getByText` and `not.toBeNull()` is redundant

slax57 (changes requested):

packages/ra-ui-materialui/src/list/ListView.tsx
WiXSL: How about adding an `Error` component:

error

slax57: I'm not too found of this solution either, cause we don't do it for other dataProvider errors, so why only do it for the list?

To me, the content of the `PaginationLimit` was enough for our needs (provided we change the message of course), but I just thought we should create a separate component for this use.

However, this `Error` component makes me realize it could be a welcome addition to add a 'Retry' button, that would call `refresh` from `useRefresh`. This could be useful in case it's only a temporary error.

What do you think?

fzaninotto: I like the idea of the Error page. Displaying an empty page with a warning is less user friendly IMHO.

slax57: alright then :sweat_smile:

slax57 (approved):

slax57:
Okay for me!
I'll let @fzaninotto approve it as well before merging

fzaninotto (approved):

PR #8324, opened 2022-10-27
Fix List filter `Remove all filters` item not present on alwaysOn filters

### LINKED ISSUE #8277: Clear filter action doesn't appear when filter has `alwaysOn` props
**Description:**

When any filter elements that has `alwaysOn` as true in filters prop on `<List />` component, `Remove Filter` action menu doesn't appears on `Add Filter` Button.

I assume, user should be able to remove filter by clicking reset(`x` mark on the right side of input) button and by clicking `Remove Filter` menu on `Add Filter` Button.

**What you were expecting:**

`Remove Filter` menu is appeared, even the filter input has `alwaysOn` prop.
So that, a user can remove all filter by clicking the menu.

**What happened instead:**

`Remove Filter` menu doesn't appear on the menu.

code sandbox link : https://codesandbox.io/s/broken-silence-8k4x5u?file=/src/posts/PostList.tsx

**Steps to reproduce:**

1. Choose any id on `<AutocompleteInput />` that is produced or add any value on `<SearchInput />` that is provided
2. Click `Add Filter Button`

**Environment**

* React-admin version: "^4.2.0"
* React version: "^17.0.2",
* Browser: Chrome

WiXSL:
Fixes #8277

- [X]  Fix
- [x] Update tests

Note: The related story already includes an alwaysOn filter

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
@@ -92,7 +92,7 @@ describe('<FilterButton />', () => {
             );
         });
 
-        it('should remove all filters when the "Clear all filters" button is clicked', async () => {
+        it('should remove all non alwaysOn filters when the "Remove all filters" button is clicked', async () => {
             render(<Basic />);
 
             // First, check we don't have a clear filters option yet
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -92,7 +92,7 @@ describe('<FilterButton />', () => {
             );
         });
 
-        it('should remove all filters when the "Clear all filters" button is clicked', async () => {
+        it('should remove all non alwaysOn filters when the "Remove all filters" button is clicked', async () => {
```
slax57, line 95: I would not have changed the title of this test.

Rather, I would have had:
- Test 1: `should remove all filters when the "Remove all filters" button is clicked`
- Test 2 (the one you added): `should also remove all alwaysOn filters when the "Remove all filters" button is clicked`

But you are right that `Clear all filters` was renamed to `Remove all filters` so we should keep this part

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -198,7 +198,7 @@ export const FilterButton = (props: FilterButtonProps): JSX.Element => {
                         })}
                     </MenuItem>
                 )}
-                {hasDisplayedFilters && (
+                {hasFilterValues && (
```
slax57, line 201: You have a new linter warning: `'hasDisplayedFilters' is assigned a value but never used`

slax57 (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
WiXSL: 😩

slax57 (approved):

PR #8340, opened 2022-11-01
Fix FormDataConsumer's values not being up-to-date on mount

### LINKED ISSUE #7603: Input level default values aren't applied to the formState on mount
With code like this:

```
const PostCreate = () => {
  return (
    <Create
      transform={(r) => {
        console.log(r);
        return r;
      }}
    >
      <SimpleForm>
        <BooleanInput source="hi" defaultValue />
        <FormDataConsumer>
          {({ formData }) => {
            if (formData.hi) return null;

            return <TextInput source="boo" />;
          }}
        </FormDataConsumer>
      </SimpleForm>
    </Create>
  );
};
```

I would expect to not see the `TextInput` rendered on page load. This is because we don't display it if the `BooleanInput` is set to true, (which is its default value).

You can see this behaviour in the [PostCreate](https://codesandbox.io/s/ecstatic-pond-nhb7ct?file=/src/posts/PostCreate.tsx) page

It seems that if an input has a `defaultValue` that value isn't applied to the formState immediately. The `formState` is only then updated once we change one of the form's inputs. 

In addition to the formState not being up to date, it also means that the save button is disabled until we change one of the inputs, which shouldn't be the case if we've set default values on one or more of our inputs

A temporary work around is to set these values from within the `Form`'s `defaultValues` object. 

EDIT: 

It seems that even with the temporary workaround, the save button is still disabled until we make a change to one of the inputs. My initial thought is that this not expected behaviour. For a good UX, we might want to set sensible defaults to our form, which our end-user could save without having to make any additional changes themselves.

I assumed that there may be a way to override this behaviour via the `Form`'s props, but I can't see an attribute to override the SaveButton's disabled state. I figure I would need to pass through a custom save button with the `alwaysEnable` prop set to `true`.

**Environment**

* React-admin version:
* React version:
* Browser:

WiXSL:
Fixes #7603

### DIFF
--- a/packages/ra-core/src/form/FormDataConsumer.spec.tsx
+++ b/packages/ra-core/src/form/FormDataConsumer.spec.tsx
@@ -1,7 +1,15 @@
 import * as React from 'react';
-import { render } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
 
-import { FormDataConsumerView } from './FormDataConsumer';
+import FormDataConsumer, { FormDataConsumerView } from './FormDataConsumer';
+import { testDataProvider } from '../dataProvider';
+import {
+    AdminContext,
+    BooleanInput,
+    SimpleForm,
+    TextInput,
+} from 'ra-ui-materialui';
+import expect from 'expect';
 
 describe('FormDataConsumerView', () => {
     it('does not call its children function with scopedFormData and getSource if it did not receive an index prop', () => {
@@ -26,6 +34,7 @@ describe('FormDataConsumerView', () => {
     it('calls its children function with scopedFormData and getSource if it received an index prop', () => {
         const children = jest.fn(({ getSource }) => {
             getSource('id');
+            return null;
         });
         const formData = { id: 123, title: 'A title', authors: [{ id: 0 }] };
 
@@ -45,5 +54,27 @@ describe('FormDataConsumerView', () => {
         expect(children.mock.calls[0][0].getSource('id')).toEqual(
             'authors[0].id'
         );
+    });
+
+    it('calls its children with updated formData on first render', async () => {
+        let globalFormData;
+        render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <SimpleForm>
+                    <BooleanInput source="hi" defaultValue />
+                    <FormDataConsumer>
+                        {({ formData, ...rest }) => {
+                            globalFormData = formData;
+
+                            return <TextInput source="bye" {...rest} />;
+                        }}
+                    </FormDataConsumer>
+                </SimpleForm>
+            </AdminContext>
+        );
+
+        await waitFor(() => {
+            expect(globalFormData).toEqual({ hi: true, bye: undefined });
+        });
     });
 });
--- a/packages/ra-core/src/form/FormDataConsumer.tsx
+++ b/packages/ra-core/src/form/FormDataConsumer.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { ReactNode } from 'react';
-import { useWatch } from 'react-hook-form';
+import { useFormContext } from 'react-hook-form';
 import get from 'lodash/get';
 
 import warning from '../util/warning';
@@ -43,9 +43,9 @@ import warning from '../util/warning';
  * );
  */
 const FormDataConsumer = (props: ConnectedProps) => {
-    const formData = useWatch();
+    const { getValues } = useFormContext();
 
-    return <FormDataConsumerView formData={formData} {...props} />;
+    return <FormDataConsumerView formData={getValues()} {...props} />;
 };
 
 export const FormDataConsumerView = (props: Props) => {


djhi (approved):

packages/ra-core/src/form/FormDataConsumer.spec.tsx
```diff
@@ -26,6 +34,7 @@ describe('FormDataConsumerView', () => {
     it('calls its children function with scopedFormData and getSource if it received an index prop', () => {
         const children = jest.fn(({ getSource }) => {
             getSource('id');
+            return null;
```
slax57, line 37: FMI, why did you need to change this?

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -43,22 +43,20 @@ import warning from '../util/warning';
  * );
  */
 const FormDataConsumer = (props: ConnectedProps) => {
-    const formData = useWatch();
+    const { getValues } = useFormContext();
 
-    return <FormDataConsumerView formData={formData} {...props} />;
+    return <FormDataConsumerView formData={getValues()} {...props} />;
```
slax57, line 48: I'm afraid this is not strictly equivalent, as `formData` is actually (I believe, kinda like) a state, that will trigger re-render when it changes, but `getValues()` is actually a callback, that only works when we call it.

The following `PostEdit` (based on [this doc example](https://marmelab.com/react-admin/Inputs.html#linking-two-inputs)) no longer works with your change:

```tsx
import {
    ArrayInput,
    Edit,
    FormDataConsumer,
    SelectInput,
    SimpleForm,
    SimpleFormIterator,
    TextInput,
} from 'react-admin'; // eslint-disable-line import/no-unresolved

const PostEdit = () => (
    <Edit>
        <SimpleForm>
            <ArrayInput source="authors">
                <SimpleFormIterator>
                    <TextInput source="name" />
                    <FormDataConsumer>
                        {({
                            formData, // The whole form data
                            scopedFormData, // The data for this item of the ArrayInput
                            getSource, // A function to get the valid source inside an ArrayInput
                            ...rest
                        }) =>
                            scopedFormData && scopedFormData.name ? (
                                <SelectInput
                                    source={getSource('role')} // Will translate to "authors[0].role"
                                    choices={[
                                        { id: 1, name: 'Head Writer' },
                                        { id: 2, name: 'Co-Writer' },
                                    ]}
                                    {...rest}
                                />
                            ) : null
                        }
                    </FormDataConsumer>
                </SimpleFormIterator>
            </ArrayInput>
        </SimpleForm>
    </Edit>
);

export default PostEdit;

```

slax57 (changes requested):

packages/ra-core/src/form/FormDataConsumer.spec.tsx
WiXSL: It was throwing a TypeScript error on children usage

packages/ra-core/src/form/FormDataConsumer.tsx
WiXSL: Great review!

slax57 (approved):

slax57:
Let's wait for @fzaninotto 's review before merging this one

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -43,22 +43,25 @@ import warning from '../util/warning';
  * );
  */
 const FormDataConsumer = (props: ConnectedProps) => {
-    const formData = useWatch();
+    const { getValues } = useFormContext();
+    let formData = useWatch();
```
WiXSL, line 47: `useWatch` docs: `Note: the first render will always return defaultValue when it's supplied.`

Initial discussion:
https://github.com/react-hook-form/react-hook-form/issues/3758

So, is not a bug, is a feature of react-hook-form.

fzaninotto:
I understand that useWatch returns the form defaultValues, which is an empty record by default. But when the Form parent passes a record, it resets the form values, and useWatch should return the right values in this case. 

I saw this in [the react-hook-form docs](https://react-hook-form.com/api/usewatch):

image

Isn't that the problem?

slax57:
@fzaninotto I think I have a clear proof that your intuition is correct.
If you invert the two components `<FormDataConsumer>` and `<BooleanInput>` in @andrico1234 's [PostCreate](https://codesandbox.io/s/ecstatic-pond-nhb7ct?file=/src/posts/PostCreate.tsx), then you no longer have the issue.

I think it makes for a very dirty gotcha, that we need to either document or, preferably, fix.

However, I can think of no better solution than what @WiXSL suggests.
Hence I advocate for accepting and merging this PR as is.

FTR I have tried to manually call `form.watch()` in `useAugmentedForm`, but it doesn't work.

fzaninotto:
We need help from @bluebill1049 on this one. I suggest you go on react-hook-form's Discord server and ask for help.

WiXSL:
I've made a codesandbox that roughly reproduces the situation: https://codesandbox.io/s/eager-hill-o4g71q?file=/src/App.tsx

WiXSL:
I've improved the tests to guarantee the component is reactive.
So far, no news from discord

slax57 (approved):

fzaninotto (approved):

PR #8350, opened 2022-11-03
Fix buttons props

WiXSL:
Fixes #8347

PR #8356, opened 2022-11-04
Fix `<Appbar>` buttons don't have consistent spacing

### LINKED ISSUE #8347: IconButton size on AppBar is not very consistent
**What you were expecting:**

I expect all the IconButton on AppBar to have consistent size. (or a feature to allow better override of default size)

**What happened instead:**

Some of IconButtons are large and some are medium, also the spacing, they are not very consistent.

**Steps to reproduce:**

**Related code:**

Medium by default https://github.com/marmelab/react-admin/blob/a408de0db2c6db0281b3d39e27c1f2de3deed44e/packages/ra-ui-materialui/src/button/ToggleThemeButton.tsx#L39

Medium by default https://github.com/marmelab/react-admin/blob/64a9091c49a2a36a8974f1c261727edc64eeb39e/packages/ra-ui-materialui/src/preferences/InspectorButton.tsx#L40

large https://github.com/marmelab/react-admin/blob/a408de0db2c6db0281b3d39e27c1f2de3deed44e/packages/ra-ui-materialui/src/button/RefreshIconButton.tsx#L32

large https://github.com/marmelab/react-admin/blob/a408de0db2c6db0281b3d39e27c1f2de3deed44e/packages/ra-ui-materialui/src/layout/UserMenu.tsx#L107

large https://github.com/marmelab/react-admin/blob/479bc4af86d8a6fa31022ab57a7fe72428c82e36/packages/ra-ui-materialui/src/layout/SidebarToggleButton.tsx#L32

**Other information:**

**Environment**

* React-admin version: 4.5.0
* React version: 18.2.0
* Browser:

fzaninotto:
## Before

image

## After

image

Closes #8347

### DIFF
--- a/packages/ra-ui-materialui/src/button/LocalesMenuButton.tsx
+++ b/packages/ra-ui-materialui/src/button/LocalesMenuButton.tsx
@@ -56,12 +56,10 @@ export const LocalesMenuButton = (props: LocalesMenuButtonProps) => {
                 aria-label=""
                 aria-haspopup="true"
                 onClick={handleLanguageClick}
+                startIcon={<LanguageIcon />}
+                endIcon={<ExpandMoreIcon fontSize="small" />}
             >
-                <LanguageIcon />
-                <div className={LocalesMenuButtonClasses.selectedLanguage}>
-                    {getNameForLocale(locale)}
-                </div>
-                <ExpandMoreIcon fontSize="small" />
+                {getNameForLocale(locale)}
             </Button>
             <Menu
                 id="simple-menu"
@@ -86,18 +84,12 @@ export const LocalesMenuButton = (props: LocalesMenuButtonProps) => {
 
 const PREFIX = 'RaLocalesMenuButton';
 
-export const LocalesMenuButtonClasses = {
-    selectedLanguage: `${PREFIX}-selectedLanguage`,
-};
+export const LocalesMenuButtonClasses = {};
 
 const Root = styled(Box, {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
-})(({ theme }) => ({
-    [`& .${LocalesMenuButtonClasses.selectedLanguage}`]: {
-        marginLeft: theme.spacing(1),
-    },
-}));
+})({});
 
 export interface LocalesMenuButtonProps {
     languages?: { locale: string; name: string }[];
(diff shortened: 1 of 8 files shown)


packages/ra-ui-materialui/src/button/LocalesMenuButton.tsx
```diff
@@ -86,18 +84,12 @@ export const LocalesMenuButton = (props: LocalesMenuButtonProps) => {
...
-export const LocalesMenuButtonClasses = {
-    selectedLanguage: `${PREFIX}-selectedLanguage`,
-};
+export const LocalesMenuButtonClasses = {};
 
 const Root = styled(Box, {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
-})(({ theme }) => ({
-    [`& .${LocalesMenuButtonClasses.selectedLanguage}`]: {
```
slax57, line 97: This change requires a doc update

packages/ra-ui-materialui/src/layout/SidebarToggleButton.tsx
```diff
@@ -29,11 +29,7 @@ export const SidebarToggleButton = (props: SidebarToggleButtonProps) => {
             )}
             enterDelay={500}
         >
-            <StyledIconButton
-                color="inherit"
-                onClick={() => setOpen(!open)}
-                size="large"
-            >
+            <StyledIconButton color="inherit" onClick={() => setOpen(!open)}>
```
slax57, line 32: #8350 also suggested to pass the `...rest` props to the button, don't you think we should do it too?

slax57 (changes requested):

packages/ra-ui-materialui/src/layout/SidebarToggleButton.tsx
WiXSL: > #8350 also suggested to pass the `...rest` props to the button, don't you think we should do it too?

+1 on this

WiXSL (changes requested):

packages/ra-ui-materialui/src/layout/SidebarToggleButton.tsx
fzaninotto: I don't see the need for this

WiXSL (approved):

PR #8359, opened 2022-11-04
[Doc] Improve `SimpleFormIterator` example

WiXSL:


docs/SimpleFormIterator.md
```diff
@@ -150,18 +150,20 @@ const PostEdit = () => (
 **Note**: `<SimpleFormIterator>` only accepts `Input` components as children. If you want to use some `Fields` instead, you have to use a `<FormDataConsumer>` to get the correct source, as follows:
```
WiXSL, line 150: This is actually incorrect, it accepts fields, but without a label, once you add the Labeled component you need the `FormDataConsumer`

fzaninotto: Well, it depends: you can use Label+Fields directly to display the `record` value, but if you want to render a reactive value based on the form inputs, you have to use the formDataConsumer. 

I think this should be clarified in the doc.

fzaninotto (changes requested):

docs/SimpleFormIterator.md
WiXSL: Well, actually, this is how it goes inside a `SimpleFormIterator`:

```jsx
<ArrayInput source="authors">
    <SimpleFormIterator>
        <TextField source="role" />        // => Value is shown , no label
        <Labeled label="Role">              // => Label is shown , no value
            <TextField source="role" />
        </Labeled>
        <FormDataConsumer>               // => Label + value are shown
            {({ getSource }) => {
                return (
                    <Labeled label="Role">
                        <TextField source={getSource('role')} />
                    </Labeled>
                );
            }}
        </FormDataConsumer>
        <FormDataConsumer>                   // => Label is shown, no value
            {({ getSource, scopedFormData }) => {
                return (
                    <Labeled label="Role">
                        <TextField
                            source={getSource('role')}
                            record={scopedFormData}
                        />
                    </Labeled>
                );
            }}
        </FormDataConsumer>
    </SimpleFormIterator>
</ArrayInput>
```

docs/SimpleFormIterator.md
```diff
@@ -150,18 +150,17 @@ const PostEdit = () => (
...
         <FormDataConsumer>
-            {({ getSource, scopedFormData }) => {
+            {({ getSource }) => {
                 return (
-                    <TextField
-                        source={getSource('url')}
-                        record={scopedFormData}
-                    />
+                    <Labeled label="Url">
+                        <TextField source={getSource('url')} />
```
fzaninotto, line 162: This is wrong, this will show the data from the record, not the form. 

If it doesn't work with scopedFormData, it may be because the usewatch() in FormDataConsumer doesn't return data before changes. This is what we must fix, not the docs.

slax57: Might be linked to https://github.com/marmelab/react-admin/pull/8340

slax57 (approved):

PR #8369, opened 2022-11-07
[Doc] Add missing `Admin`'s notification prop

WiXSL:


docs/Admin.md
```diff
@@ -533,6 +533,33 @@ const App = () => (
 );
 ```
 
+## `notification`
```
fzaninotto, line 536: please add it to the list of props at the beginning of this chapter, too

docs/Admin.md
```diff
@@ -533,6 +533,33 @@ const App = () => (
...
+
+You can override the notification component, for instance to change the notification duration. It defaults to 4000, i.e. 4 seconds, and you can override it using the `autoHideDuration` prop. For instance, to create a custom Notification component with a 5 seconds default:
```
fzaninotto, line 538:
```suggestion
You can override the notification component, for instance to change the notification duration. A common use case is to change the `autoHideDuration`, and force the notification to remain on screen longer than the default 4 seconds. For instance, to create a custom Notification component with a 5 seconds default:
```

docs/Admin.md
```diff
@@ -533,6 +533,33 @@ const App = () => (
...
+
+```jsx
+// in src/MyNotification.js
+import { Notification } from 'react-admin';
+
+const MyNotification = props => <Notification {...props} autoHideDuration={5000} />;
```
fzaninotto, line 544:
```suggestion
const MyNotification =() => <Notification autoHideDuration={5000} />;
```

fzaninotto (changes requested):

docs/Admin.md
WiXSL: This was taken from the recipes section, should I change it there too?

fzaninotto (approved):

PR #8372, opened 2022-11-08
Add ability to refresh the user identity in `useGetIdentity`

fzaninotto:
## Problem

If an application contains a form letting the current user update their name and/or avatar, developers may want to refresh the identity after the form is submitted. It is currently cumbersome as it implies adding a context on top of the app (cf https://marmelab.com/blog/2020/12/14/react-admin-v3-userprofile.html) 

## Solution

Use react-query to fetch the authProvider in `useGetIdentity`

## Sample code

```jsx
const IdentityForm = () => {
    const { isLoading, error, identity, refetch } = useGetIdentity();
    const [newIdentity, setNewIdentity] = useState('');

    if (isLoading) return <>Loading</>;
    if (error) return <>Error</>;

    const handleChange = event => {
        setNewIdentity(event.target.value);
    };
    const handleSubmit = (e) => {
        e.preventDefault();
        if (!newIdentity) return;
        fetch('/update_identity', {
            method: 'POST',
            headers: { 'Content-Type': 'application/json' },
            body: JSON.stringify({ identity: newIdentity })
        }).then(() => { 
            // call authProvider.getIdentity() again and notify the listeners of the result,
            // including the UserMenu in the AppBar
            refetch();
         });
    };
    
    return (
        <form onSubmit={handleSubmit}>
            <input defaultValue={identity.fullName} onChange={handleChange} />
            <input type="submit" value="Save" />
        </form>
    );
};
```


packages/ra-core/src/auth/useRefreshIdentity.ts
```diff
@@ -0,0 +1,36 @@
+import { useCallback } from 'react';
```
fzaninotto, line 1: This hook is undocumented on purpose - I'm not sure it's better than documenting the react-query queryKey

fzaninotto:
I considered exposing the `refetch` from react-query in the `useGetPermissions` response instead. I still hesitate, and I'm open to comments.

packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -1,33 +1,36 @@
...
+ * @returns The current user identity. Destructure as { isLoading, data, error, refetch }.
  *
  * @example
- *
  * import { useGetIdentity, useGetOne } from 'react-admin';
  *
  * const PostDetail = ({ id }) => {
  *     const { data: post, isLoading: postLoading } = useGetOne('posts', { id });
- *     const { identity, isLoading: identityLoading } = useGetIdentity();
+ *     const { data: identity, isLoading: identityLoading } = useGetIdentity();
```
fzaninotto, line 33: I made this change to make the API more intuitive now that we use `useQuery`. The change is backwards compatible, as the hook still returns `identity`.

packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -38,42 +41,60 @@ const defaultIdentity = {
...
+            result.isLoading
+                ? { isLoading: true }
+                : result.error
+                ? { error: result.error, isLoading: false }
+                : {
+                      data: result.data,
+                      identity: result.data,
+                      refetch: result.refetch,
+                      isLoading: false,
+                  },
```
djhi, line 68: Why not return react-query result, just extending it to add the `identity` property?

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -49,13 +49,14 @@ const usePermissions = <Permissions = any, Error = any>(
...
-    return useMemo(() => {
-        return {
+    return useMemo(
+        () => ({
             permissions: result.data,
             isLoading: result.isLoading,
             error: result.error,
-        };
-    }, [result]);
+        }),
```
djhi, line 57: Same here with `permissions`?

packages/ra-core/src/auth/useGetIdentity.ts
fzaninotto: I've tried, but TypeScript complains because of the discriminative union.

packages/ra-core/src/auth/usePermissions.ts
fzaninotto: I'm pretty sure TypeScript complains that `permissions` may be null even after checking `isLoading` and `error`.

But improving the return type of `usePermissions` is outside the scope of this PR.

I only changed the syntax slightly for my eye pleasure on this file.

WiXSL (changes requested):

WiXSL (changes requested):

docs/useGetIdentity.md
```diff
@@ -5,7 +5,27 @@ title: "useGetIdentity"
...
+`useGetIdentity()` calls `authProvider.getIdentity()` on mount. It returns an object containing the loading state, the error state, and the identity.
+
+```jsx
+const { data, isLoading, error } = useGetIdentity();
+```
+
+Once loaded, the `data` object contains the following properties:
+
+```jsx
+const { id, fullName, avatar } = identity;
```
slax57, line 23:
```suggestion
const { id, fullName, avatar } = data;
```

packages/ra-core/src/auth/useGetIdentity.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+    it('should return the identity', async () => {
+        render(<Basic />);
+        await screen.findByText('John Doe');
+    });
+    it('should return the authProvider error', async () => {
+        jest.spyOn(console, 'error').mockImplementationOnce(() => {});
+        render(<ErrorCase />);
+        await screen.findByText('Error');
+    });
+    it('should allow to invalidate the query cache', async () => {
```
slax57, line 16:
```suggestion
    it('should update the identity after a change', async () => {
```

packages/ra-core/src/auth/useGetIdentity.stories.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from 'react';
+import { QueryClientProvider, QueryClient, useQueryClient } from 'react-query';
```
slax57, line 2: Linter warning about `useQueryClient` not being used

slax57 (changes requested):
Minor changes requests, otherwise I approve the main changes

packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -38,42 +41,61 @@ const defaultIdentity = {
...
-const useGetIdentity = () => {
-    const [state, setState] = useSafeSetState<State>({
-        isLoading: true,
-    });
+export const useGetIdentity = (
+    queryParams: UseQueryOptions<UserIdentity, Error> = defaultQueryParams
+): UseGetIdentityResult => {
     const authProvider = useAuthProvider();
-    useEffect(() => {
-        if (authProvider && typeof authProvider.getIdentity === 'function') {
```
smeng9, line 47: Removing typeof authProvider.getIdentity === 'function' check causes a tiny BC break.

We found there are console errors after upgrading pass 4.6 because we have never implemented the getIdentity method since it was introduced in https://github.com/marmelab/react-admin/pull/5180 

In the previous pull request it says 
> The change is backward compatible (if the authProvider.getIdentity() method isn't implemented, the admin shows an anonymous user avatar with no name, as before).

In v4 upgrade guide it did not mention this method is now required to be implemented. I believe we should now add a note about that breaking change.

smeng9: It will be very easy for us to catch the issue and implement the method.

Just for other users they should be aware when they upgrade.

slax57: Thanks for pointing that out!
The `getIdentity` function should remain optional, even in v4, so I believe we should reintroduce this check to avoid errors. I'll open a new PR.

slax57: Here it is: https://github.com/marmelab/react-admin/pull/8509

PR #8376, opened 2022-11-09
Fix `<DatagridConfigurable>` ignores children without label or source

### LINKED ISSUE #8366: `<DatagridConfigurable>` ignores children without label
**What you were expecting:**

All children of `<DatagridConfigurable>` are shown by default

**What happened instead:**

Only children with a `label` prop appear. Therefore, actions columns are hidden.

**Steps to reproduce:**

Create a DatagridConfigurable using `<EditButton>` as a child:

```jsx
<DatagridConfigurable>
    <TextField source="id" />
    <TextField source="title" label="Original title" />
    <TextField source="author" />
    <TextField source="year" />
    <EditButton />
</DatagridConfigurable>
```

The last column doesn't appear, and the editor ignores it, too.

image

**Workaround:**

Give the `<EditButton>` a label. 

```diff
<DatagridConfigurable>
    <TextField source="id" />
    <TextField source="title" label="Original title" />
    <TextField source="author" />
    <TextField source="year" />
-   <EditButton />
+   <EditButton label="Actions" />
</DatagridConfigurable>
```

image

**Environment**

* React-admin version: 4.5.0
* React version: 17
* Browser: Chrome

WiXSL:
Fixes #8366

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
@@ -52,12 +52,14 @@ export const DatagridConfigurable = ({
     React.useEffect(() => {
         // first render, or the preference have been cleared
         const columns = React.Children.map(props.children, (child, index) =>
-            React.isValidElement(child) &&
-            (child.props.source || child.props.label)
+            React.isValidElement(child)
                 ? {
                       index: String(index),
                       source: child.props.source,
-                      label: child.props.label,
+                      label:
+                          child.props.source || child.props.label
+                              ? child.props.label
+                              : `Column-${index}`,
                   }
                 : null
         ).filter(column => column != null);
@@ -114,7 +116,7 @@ export interface DatagridConfigurableProps extends DatagridProps {
 
 export interface ConfigurableDatagridColumn {
     index: string;
-    source: string;
+    source?: string;
     label?: string;
 }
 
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
```diff
@@ -52,12 +52,14 @@ export const DatagridConfigurable = ({
...
-            (child.props.source || child.props.label)
+            React.isValidElement(child)
                 ? {
                       index: String(index),
                       source: child.props.source,
-                      label: child.props.label,
+                      label:
+                          child.props.source || child.props.label
+                              ? child.props.label
+                              : `Column-${index}`,
```
WiXSL, line 62: Label suggestions?

slax57: How about `Unlabeled column #${index}`

slax57 (approved):

packages/ra-language-french/src/index.ts
```diff
@@ -190,6 +190,7 @@ const frenchMessages: TranslationMessages = {
                 title: 'Inspecteur',
                 content: 'Sélectionner un composant pour le configurer',
                 reset: 'Réinitialiser',
+                unlabeled: 'Colonne sans libellé #%{column}',
```
slax57, line 193: French translation is good :grin: 

But in this specific case I'd rather keep 'label' instead of 'libellé', to refer to the actual prop name rather than the notion

```suggestion
                unlabeled: 'Colonne sans label #%{column}',
```

slax57 (changes requested):

slax57 (approved):

PR #8382, opened 2022-11-10
Fix authProvider hooks support for redirectTo: absolute URL

fzaninotto:
This was possible in v3 since #6469, but it was broken when migrating to v4. 

It's necessary for OIDC / OAuth workflows.

### DIFF
--- a/packages/ra-core/src/auth/useLogout.ts
+++ b/packages/ra-core/src/auth/useLogout.ts
@@ -70,11 +70,20 @@ const useLogout = (): Logout => {
                     // do not redirect
                     return;
                 }
-                // redirectTo can contain a query string, e.g. '/login?foo=bar'
-                // we must split the redirectTo to pass a structured location to navigate()
-                const redirectToParts = (
-                    redirectToFromProvider || redirectTo
-                ).split('?');
+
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/auth/useLogout.ts
```diff
@@ -70,11 +70,20 @@ const useLogout = (): Logout => {
...
+                    // absolute link (e.g. https://my.oidc.server/login)
+                    resetStore();
+                    queryClient.clear();
+                    window.location.href = finalRedirectTo;
+                    return finalRedirectTo;
+                }
+
+                // redirectTo is an internal location that may contain a query string, e.g. '/login?foo=bar'
+                // we must split it to pass a structured location to navigate()
+                const redirectToParts = finalRedirectTo.split('?');
```
WiXSL, line 86: It throws an error here in simple project with:
```js
//authProvider
{
   logout: () => {
         return Promise.resolve({ redirectTo: 'https://...' }); 
        //return Promise.resolve({ redirectTo: false }); // as well
    },
}
```

WiXSL (changes requested):

WiXSL (approved):

slax57 (approved):

Issue #8353, opened 2022-11-03
feat: addition of generic in `FormDataConsumer`


Gabriel-Malenowitch:
**Is your feature request related to a problem? Please describe.**
 - A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
 - Currently It turns the code more verbose. it's very common for me to write forms with this component and its became a boring work, because typescript don't get me the suggestions

**Describe the solution you'd like**
 - Adding the generic in the `FormDataConsumer` component
 How it is currently:
```typescript
const FormDataConsumer = (props: ConnectedProps) => {
    const formData = useWatch();

    return <FormDataConsumerView formData={formData} {...props} />;
};
```

How could be to turn our life happier :green_heart: :
```typescript
const FormDataConsumer = <FieldValues extends Record<string, unknown> = {}>(props: ConnectedProps) => {
    const formData = useWatch<FieldValues>();

    return <FormDataConsumerView formData={formData} {...props} />;
};
```

**Describe alternatives you've considered**
 - There are a lot of ways to type something with react and typescript, think big or think small, i just want to use typescript with this component : )
**Additional context**
 - I don't know how I could be more specific, but please let me know if you have any questions.

fzaninotto:
Nice idea for an enhancement! Feel free to open a PR to implement it.

Gabriel-Malenowitch:
@fzaninotto my PR is ready for review. Can you review it pls? https://github.com/marmelab/react-admin/pull/8389

viniciusbsneto:
Looking forward to this PR!

lukascivil:
@fzaninotto help us here, pls!

fzaninotto:
Fixed by #8389

PR #8389, opened 2022-11-13
[RFR] enhancement: add generic in form data consumer input

Gabriel-Malenowitch:
## What is it?
 - Following the issue https://github.com/marmelab/react-admin/issues/8353

 ## Code example:
 ```typescript
const countries = ['bar', 'foo'] as const

 interface FieldValues {
     country: typeof countries[number] | undefined
     city: string
 }

 const OrderEdit = (props) => (
     <Edit {...props}>
         <SimpleForm>
             <SelectInput source="country" choices={countries} />
             <FormDataConsumer<FieldValues>>
                 {({ formData, ...rest }) =>
                      <SelectInput
                          source="city"
                          choices={getCitiesFor(formData.country)}
                          {...rest}
                      />
                 }
             </FormDataConsumer>
         </SimpleForm>
     </Edit>
 );
 ```


packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -99,26 +105,24 @@ export const FormDataConsumerView = (props: Props) => {
...
+    children: FormDataConsumerRender<TFieldValues>;
     form?: string;
     record?: any;
     source?: string;
     [key: string]: any;
 }
-
-interface Props extends ConnectedProps {
-    formData: any;
+interface Props<TFieldValues extends FieldValues> extends ConnectedProps {
```
lukascivil, line 125: @Gabriel-Malenowitch we had the line broken in some cases, i think it's worth going back.

Gabriel-Malenowitch: I've reviewed it here and couldn't identify the broken lines you say (only from eslint, which broke a few).
This interface (`Props`) has just been improved. Were you talking about her?

image

viniciusbsneto: @Gabriel-Malenowitch I don't know why but it's not inferring optional fields neither null ones.

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -42,13 +41,20 @@ import warning from '../util/warning';
...
-const FormDataConsumer = (props: ConnectedProps) => {
-    const formData = useWatch();
-
-    return <FormDataConsumerView formData={formData} {...props} />;
+const FormDataConsumer = <TFieldValues extends FieldValues = FieldValues>(
+    props: ConnectedProps<TFieldValues>
+) => {
+    const formData = useWatch<TFieldValues>();
+    return (
+        <FormDataConsumerView<TFieldValues> formData={formData} {...props} />
```
slax57, line 49: I have a TS warning here

packages/ra-core/src/form/FormDataConsumer.tsx
slax57: I'm surprised you don't need to do

```suggestion
interface Props<TFieldValues extends FieldValues> extends ConnectedProps<TFieldValues> {
```

or something similar

slax57: Besides, you removed all the extra lines breaks in the file, which I think hurts its readability. Can you add them back?

slax57 (changes requested):

Gabriel-Malenowitch:
@slax57 @lukascivil @viniciusbsneto About break lines, that was nice point. Thanks! 
https://github.com/marmelab/react-admin/pull/8389/commits/0d771fe7c1fdf075e204921670b31a60dd7cbc46
Is missing broke someline? I checked here and that was all

packages/ra-core/src/form/FormDataConsumer.tsx
Gabriel-Malenowitch: Nice point guys! https://github.com/marmelab/react-admin/pull/8389/commits/0d771fe7c1fdf075e204921670b31a60dd7cbc46

Gabriel-Malenowitch:
@slax57 
 > I'm surprised you don't need to do
 > or something similar[

Here is some points:
 - `= FieldValues` set a default type, so we don't need to pass it here
      - image
- ConnectedProps is beeing used in two places:
     - here
 ```typescript
const FormDataConsumer = <TFieldValues extends FieldValues = FieldValues>(
    props: ConnectedProps<TFieldValues>
) => {
```
     - In first place we don't need to use it because the formData is declared in own interface, and the generic  in `ConnectedProps` is to type his children.
If we delete the generic here
image
image

 ##### Thats don't work
image


##### And if we add again, that works
image
image

##### Because is typing his children
image

##### So here (code piece where you did comment) this generic has no usability in `ConnectecProps`

packages/ra-core/src/form/FormDataConsumer.tsx
Gabriel-Malenowitch: Strange, because for me it didn't happen 🤔 😕 . But in rebase this was shown to me, and i fixed it. Can you try it again?
Fix commit -> https://github.com/marmelab/react-admin/pull/8389/commits/4c15b3793eac58c990db23bb227695960e736b60

fzaninotto (approved):

fzaninotto:
Thanks!

PR #8397, opened 2022-11-16
Fix DatagridConfigurable uses translation key with wrong case

fzaninotto:
Follows #8376


fzaninotto:
My bad, I've seen that we've used the same Uppercase for SimpleList... I'll close this PR.

PR #8403, opened 2022-11-18
Fix `NumberInput` doesn't show `helperText` error `onBlur` if its required

WiXSL:


packages/ra-ui-materialui/src/input/NumberInput.tsx
```diff
@@ -46,6 +47,7 @@ export const NumberInput = ({
         isRequired,
     } = useInput({
         defaultValue,
+        onBlur,
```
WiXSL, line 50: Not really sure why this solves it, but it does

slax57: It changes what you pass to the MUI `TextField` (via `...sanitizeInputRestProps(rest)`)

slax57 (approved):

PR #8422, opened 2022-11-22
`NumberInput` not passing `format` and `parse` props to `useInput` call

### LINKED ISSUE #8405: Linking Two Inputs: `format` function change does not trigger a rerender
I have two inputs in a form: one for the value and another for the unit

**What you were expecting:**

I expect changing the unit from selection input will cause the value also be reformatted based on the new unit

**What happened instead:**

The value does not get reformatted

**Steps to reproduce:**

Go to https://stackblitz.com/edit/github-m7rhmv?file=src%2Fcomments%2FCommentCreate.tsx 
Go to Comments -> Create new comments
In the form change the units to degrees or radians, however in the value input only input adornment gets changed, but not the input value.

**Related code:**

The value may need to be recalculated here https://github.com/marmelab/react-admin/blob/032e1464cd23cf15dbc4d641e9663f74b37b9563/packages/ra-core/src/form/useInput.ts#L94

Maybe missing a dependency in useEffect https://github.com/marmelab/react-admin/blob/032e1464cd23cf15dbc4d641e9663f74b37b9563/packages/ra-ui-materialui/src/input/NumberInput.tsx#L69

  - https://codesandbox.io/s/github/marmelab/react-admin/tree/3.x/examples/simple (v3)
* A StackBlitz forked from 
  - https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple (v4)

```
https://stackblitz.com/edit/github-m7rhmv?file=src%2Fcomments%2FCommentCreate.tsx
```

**Other information:**

**Environment**

* React-admin version: 4.5.1
* React version: 18.0.2
* Browser: Firefox 106

WiXSL:
Fixes #8405

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
@@ -7,8 +7,7 @@ import { AdminContext } from '../AdminContext';
 import { SaveButton } from '../button';
 import { SimpleForm, Toolbar } from '../form';
 import { required } from 'ra-core';
-import { SelectInput } from './SelectInput';
-import userEvent from '@testing-library/user-event';
+import { Format, Parse } from './NumberInput.stories';
 
 describe('<NumberInput />', () => {
     const defaultProps = {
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
```diff
@@ -290,6 +292,47 @@ describe('<NumberInput />', () => {
...
+                    </SimpleForm>
+                </AdminContext>
+            );
+
+            userEvent.click(
+                screen.getByLabelText('resources.undefined.fields.unit')
+            );
+            userEvent.click(screen.getByRole('option', { name: /^radian/i }));
+
+            await waitFor(() => {
```
fzaninotto, line 328: This test is incomplete ; it doesn't show that the value has changed, because it never asserted the value before the change. 

Besides, I don't understand why you have to use a dependent input to test the format prop. Can't you write a simpler test?

packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
```diff
@@ -290,6 +292,47 @@ describe('<NumberInput />', () => {
...
+                    />
+                );
+            };
+
+            const units = ['degree', 'radian'];
+            const toChoices = items =>
+                items.map(item => ({ id: item, name: item }));
+
+            render(
+                <AdminContext>
```
fzaninotto, line 315: Please move that to a Story for easier manual testing.

packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
```diff
@@ -290,6 +292,47 @@ describe('<NumberInput />', () => {
...
+            userEvent.click(screen.getByRole('option', { name: /^radian/i }));
+
+            await waitFor(() => {
+                expect(
+                    (screen.getByLabelText(
+                        'resources.posts.fields.views'
+                    ) as HTMLInputElement).value
+                ).toEqual('1');
+            });
+        });
```
fzaninotto, line 335: We're missing a test for the parse prop.

fzaninotto (changes requested):

fzaninotto (approved):

PR #8433, opened 2022-11-24
Fix `useFormGroup` `isDirty` / `isTouched` state could have non-boolean values

### LINKED ISSUE #8400: isDirty from form groups is broken with array inputs
**What you were expecting:**
When using the `FormGroupContextProvider` around an `ArrayInput`, the `isDirty` property of `useFormGroup` still is a boolean, as the type definitions indicate.

**What happened instead:**
When using the `FormGroupContextProvider` around an `ArrayInput`, the `isDirty` property of `useFormGroup` is broken: When changing the array input, the `isDirty` property is set to some array that containes information about properties of array elements that have changed. But it should be just a `boolean`.

**Steps to reproduce:**
1. Use an `ArrayInput` inside a `FormGroupContextProvider`
2. Inside that `FormGroupContextProvider`, also use a component that uses `isDirty` via `useFormGroup` in some way (e.g., by displaying it)
3. Adjust values managed by the `ArrayInput`, or add/delete items of the `ArrayInput`
4. Observe that the `isDirty` property is set to an array, instead of just a `boolean`.

**Related code:**
https://codesandbox.io/s/goofy-tess-wzpom6?file=/src/posts/PostEdit.tsx:5416-5440

In the `PostsEdit.tsx` in the miscellaneous tab, I added a `FormGroupContextProvider` around the `backlinks` `ArrayInput`, which shows the problem.

**Other information:**
output

WiXSL:
Fixes #8400

### DIFF
--- a/packages/ra-core/src/form/useFormGroup.spec.tsx
+++ b/packages/ra-core/src/form/useFormGroup.spec.tsx
@@ -8,9 +8,9 @@ import {
     TextInput,
 } from 'ra-ui-materialui';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { testDataProvider } from '../../dist/cjs';
 import expect from 'expect';
 import { FormGroupContextProvider } from './FormGroupContextProvider';
+import { testDataProvider } from '../dataProvider';
 
 describe('useFormGroup', () => {
     test.each([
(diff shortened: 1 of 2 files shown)


slax57:
Can you clarify what non-boolean values were obtained, and in which cases?
Also, could you add tests to cover this case i the future?

slax57 (approved):

PR #8435, opened 2022-11-25
Fix simple project UserEdit save usage

WiXSL:


fzaninotto:
What bug does that solve? Wan you add a unit test to prove it?

slax57 (approved):

WiXSL:
> It seems to fix the ability to edit an User in the simple demo, which currently (indeed) doesn't work. (I tested it successfully) But you are correct, we should add this case to the already existing e2e tests to avoid a regression on this in the future.

Yes, this is what I am fixing.
I'l improve the e2e tests on this

PR #8439, opened 2022-11-26
Fix bugs in SearchInput and TextField related to the `label` property

kosmotema:
Starting with MUI 5.4.0, they changed the way of displaying the label for TextField. Now the label will not be rendered only if `null`, `undefined` or `""` value is provided, but react-admin passes `false` as the label (so it will be rendered). This often affects the SearchInput component, which doesn't (and shouldn't) have label: placeholder is shown only when SearchInput is active (because the label replaces the placeholder when the input is not active).

Also, I fixed a bug, where when providing an explicit `undefined` value to the SearchInput's `label` property, the default label (based on `source`) was shown:
```jsx
<SearchInput source="q" label={undefined} />
```

### DIFF
--- a/packages/ra-ui-materialui/src/input/SearchInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/SearchInput.spec.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { testDataProvider } from 'ra-core';
+
+import { AdminContext } from '../AdminContext';
+import { SearchInput } from '.';
+import { FilterForm } from '../list';
+
+describe('<SearchInput />', () => {
+    it('should not render label if passed explicit `undefined` value', async () => {
+        const source = 'test';
+
+        const filters = [<SearchInput source={source} label={undefined} />];
+        const displayedFilters = {
+            [source]: true,
+        };
+
+        const { container } = render(
+            <AdminContext dataProvider={testDataProvider()}>
+                <FilterForm
+                    setFilters={jest.fn()}
+                    filters={filters}
+                    displayedFilters={displayedFilters}
+                />
+            </AdminContext>
+        );
+
+        expect(container.querySelector(`label`)).toBeNull();
+    });
+});
--- a/packages/ra-ui-materialui/src/input/SearchInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SearchInput.tsx
@@ -8,9 +8,11 @@ import { CommonInputProps } from './CommonInputProps';
 import { TextInput, TextInputProps } from './TextInput';
 
 export const SearchInput = (props: SearchInputProps) => {
+    const { label, ...rest } = props;
+
     const translate = useTranslate();
 
-    if (props.label) {
+    if (label) {
         throw new Error(
             "<SearchInput> isn't designed to be used with a label prop. Use <TextInput> if you need a label."
         );
@@ -18,7 +20,6 @@ export const SearchInput = (props: SearchInputProps) => {
 
     return (
         <StyledTextInput
-            hiddenLabel
             label=""
             resettable
             placeholder={translate('ra.action.search')}
@@ -30,7 +31,7 @@ export const SearchInput = (props: SearchInputProps) => {
                 ),
             }}
             size="small"
-            {...props}
+            {...rest}
         />
     );
 };
--- a/packages/ra-ui-materialui/src/input/TextInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/TextInput.spec.tsx
@@ -166,4 +166,36 @@ describe('<TextInput />', () => {
             expect(input.value).toEqual('bar');
         });
     });
+
+    describe('label', () => {
+        it('should not render label when `label` prop is `false`', () => {
+            const { container } = render(
+                <AdminContext dataProvider={testDataProvider()}>
+                    <SimpleForm
+                        defaultValues={{ title: 'hello' }}
+                        onSubmit={jest.fn}
+                    >
+                        <TextInput {...defaultProps} label={false} />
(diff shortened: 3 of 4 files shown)


WiXSL (approved):

WiXSL:
Thank you!

packages/ra-ui-materialui/src/input/SearchInput.tsx
```diff
@@ -8,17 +8,18 @@ import { CommonInputProps } from './CommonInputProps';
 import { TextInput, TextInputProps } from './TextInput';
 
 export const SearchInput = (props: SearchInputProps) => {
+    const { label, ...rest } = props;
```
fzaninotto, line 11: I don't understand why you made this change

packages/ra-ui-materialui/src/input/SearchInput.tsx
```diff
@@ -8,17 +8,18 @@ import { CommonInputProps } from './CommonInputProps';
...
-    if (props.label) {
+    if (label) {
         throw new Error(
             "<SearchInput> isn't designed to be used with a label prop. Use <TextInput> if you need a label."
         );
     }
 
     return (
         <StyledTextInput
-            hiddenLabel
```
fzaninotto, line 21: why did you remove that?

packages/ra-ui-materialui/src/input/TextInput.spec.tsx
```diff
@@ -166,4 +166,36 @@ describe('<TextInput />', () => {
             expect(input.value).toEqual('bar');
         });
     });
+
+    describe('label', () => {
+        it('should not render label when `label` prop is `false`', () => {
```
fzaninotto, line 171: You're missing a test showing that a proper label prop renders a label.

packages/ra-ui-materialui/src/input/SearchInput.spec.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { testDataProvider } from 'ra-core';
+
+import { AdminContext } from '../AdminContext';
+import { SearchInput } from '.';
+import { FilterForm } from '../list';
+
+describe('<SearchInput />', () => {
+    it('should not render label if passed explicit `undefined` value', async () => {
```
fzaninotto, line 10: I don't understand this test.

fzaninotto (changes requested):
Could you show before / after screenshots?

Also, I understand that MUI changed, but I want to be sure that react-admin works with both the old and the new MUI versions. Are you sure this change is backwards compatible?

packages/ra-ui-materialui/src/input/SearchInput.tsx
kosmotema: To handle the case when specifying `undefined` for a SearchInput's label explicitly (like `<SearchInput source="q" alwaysOn label={undefined} />`) will result in the default label is being displayed.
[Reproduction on codesandbox](https://codesandbox.io/s/react-admin-explicit-undefined-label-miiic7?file=/src/posts/PostList.tsx)

packages/ra-ui-materialui/src/input/SearchInput.tsx
kosmotema: My bad, added it back

packages/ra-ui-materialui/src/input/TextInput.spec.tsx
kosmotema: I've added two more tests: when `label` not specified and when `label` is a non-empty string

packages/ra-ui-materialui/src/input/SearchInput.spec.tsx
kosmotema: Added to cover the case of passing explicit `undefined` for the SearchInput's label property (more info [here](https://github.com/marmelab/react-admin/pull/8439#discussion_r1033939044))

kosmotema:
> Could you show before / after screenshots?

I made a reproduction of the bug:
1. [With MUI 5.0](https://codesandbox.io/s/react-admin-mui-5-0-8yp49i)
2. [With MUI 5.4](https://codesandbox.io/s/react-admin-mui-5-4-37ivhr)
3. [With MUI 5.0 and patched react-admin](https://codesandbox.io/s/react-admin-patched-mui-5-0-7hgvc3)
4. [With MUI 5.4 and patched react-admin](https://codesandbox.io/s/react-admin-patched-mui-5-4-6p0ef9)

And with screenshots:
The placeholder is not displayed when using MUI > 5.4 and latest react-admin (because MUI [due to changed logic of displaying a label] renders the `label` element with empty text):

mui > 5.4, stable react-admin, placeholder is hidden

But the placeholder is visible after clicking on the SearchInput:

placeholder is visible

With the proposed changes, the placeholder now displayed correctly (both until the SearchInput is focused, and after that):

placholder is visible, SearchInput not focused

placeholder is also visiblde, SearchInput as focused

packages/ra-ui-materialui/src/input/SearchInput.tsx
kosmotema: Also noticed, that it's possible to use SearchInput with `0` label like:
```js
<SearchInput source="q" alwaysOn label={0} />
```

I think, it would be better to revert all changes in this PR, that related to the SearchInput's label overwriting logic and open a new issue to discuss, what and how to do about this case (IMHO it's better to prevent users from providing a `label` at all instead of throwing an error).

slax57 (approved):

slax57:
Tested the fix in `TextInput` successfully, and the change in `SearchInput` is harmless, so I'll merge this PR as is.

Thanks for this contribution and well done for the analysis!

PR #8447, opened 2022-11-30
Show focus on clear buttons in `ResettableTextfield`

septentrion-730n:
Ripple effects had been disabled on resettable textfields, this removes the feedback during tabulation navigation.


WiXSL (approved):

WiXSL:
I wonder why it was disabled in the first place

slax57 (approved):
I agree with @WiXSL 
@djhi or @fzaninotto do you have any clues?

djhi (approved):

djhi:
> I wonder why it was disabled in the first place

For the same reason we used to disable button ripples by default before: performances

PR #8449, opened 2022-11-30
Fix ArrayInput doesn't allow null as value

### LINKED ISSUE #7754: ArrayInput breaks when value is null
I'm not sure this is a bug or if this was intended with 4.x version of react-admin.

**What you were expecting:**
ArrayInputs render when field value is null.

**What happened instead:**
A `cannot read properties of null (reading 'map')` error is thrown here in react-hook-form: https://github.com/react-hook-form/react-hook-form/blob/v7.27.0/src/useFieldArray.ts#L103. It appears unless a field value is undefined react-hook-form will not default to an empty array. Based on line 101 unless a field value is undefined react-hook-form will not default to an empty array.

**Steps to reproduce:**
Set any field value in the v4 sandbox test [data.tsx](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple?file=/src/data.tsx) file intended to be rendered with ArrayInput to `null` (ie: 'backlinks').

You can see the error here in this sandbox when you navigate to the edit view for post with id 1.  https://codesandbox.io/s/boring-benz-5odpst

**Environment**

* React-admin version: 4.1.0
* React version: 1.17.0
* Browser: Chrome 101.0.4951.64 (Official Build) (x86_64)

WiXSL:
Fixes #7754

Reproduction case: https://stackblitz.com/edit/github-yb7hqu?file=package.json
1. Edit Post of id=1
Result: You get a `cannot read properties of null (reading 'map')` error.

To fix it, change `package.json` line 29 like so:

```diff
- "react-hook-form": "7.34.2",
+ "react-hook-form": "7.38.0",
```
and hard refresh the website

### DIFF
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -26,7 +26,7 @@
         "react": "^17.0.0",
         "react-admin": "^4.5.4",
         "react-dom": "^17.0.0",
-        "react-hook-form": "^7.34.2",
+        "react-hook-form": "^7.38.0",
         "react-query": "^3.32.1",
         "react-router": "^6.1.0",
         "react-router-dom": "^6.1.0"
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -26,7 +26,7 @@
         "react": "^17.0.0",
         "react-admin": "^4.5.4",
         "react-dom": "^17.0.0",
-        "react-hook-form": "^7.38.0",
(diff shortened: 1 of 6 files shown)


djhi (changes requested):
As discussed we should probably update to their latest minor

slax57 (approved):

djhi (approved):

PR #8453, opened 2022-12-01
fix format does not get triggered after function change

smeng9:
fixes https://github.com/marmelab/react-admin/issues/8405


slax57 (approved):

slax57:
Thanks!

PR #8454, opened 2022-12-01
Fix `<NumberInput>` parses 0 as string when used with a custom `parse` prop

### LINKED ISSUE #8451: NumberInput misinterprets zero-values (0 or 0.0) as type of string instead of number
**What you were expecting:**
Number inputs with values equal to ```0``` or ```0.0``` should be regarded/handled as number types.

**What happened instead:**
Whenever you enter a zero value into a NumberInput (no matter if via arrow buttons or by key input) they are interpreted as of type of string.

**Steps to reproduce:**

1. Select _Posts_ from the Menu of the provided [trimmed simple example](https://codesandbox.io/s/optimistic-goldwasser-s6wnu6?file=/src/posts/PostEdit.tsx)
2. Click one of the posts to go to its edit page (which was trimmed down to a single NumberInput)
3. Enter a zero value (```0``` or ```0.0```) into the NumberInput ("Average note")
4. Check the console info output for the actual type of the parsed/formatted NumberInput value

**Environment**

* React-admin version: 4.5.0 (see forked codesandbox of master/example/simple)
* React version: 17.0.2
* Browser: Brave 1.45.133 Chromium: 107.0.5304.141 (64-Bit)

slax57:
Fixes #8451

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.spec.tsx
@@ -290,6 +290,50 @@ describe('<NumberInput />', () => {
             });
             expect(onSubmit.mock.calls[0][0].views).toBeNull();
         });
+
+        it('should cast value to a numeric with a custom parse function', async () => {
+            const onSubmit = jest.fn();
+
+            render(
+                <AdminContext>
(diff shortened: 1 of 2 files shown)


djhi (approved):

pamuche:
:kissing_heart: Wow, thank you very much for this PR.

PR #8457, opened 2022-12-01
Add login-callback hook and authProvider optional new `handleLoginCalback` method

djhi:
## Problem

When using third party services such as Auth0, you currently need to setup a custom route and work around the authProvider methods.

## Solution

Introduce a new route at `/login-callback` that can be set as the callback endpoint for third party authentication services. This route will call the new `handleCallback` optional method of the `AuthProvider` then redirect to the home page by default.

## Tasks

- [x] Implementation
- [x] Documentation


packages/ra-core/src/auth/useHandleLoginCallback.ts
```diff
@@ -0,0 +1,57 @@
...
+import { removeDoubleSlashes } from '../routing/useCreatePath';
+import { AuthRedirectResult } from '../types';
+import useAuthProvider from './useAuthProvider';
+
+/**
+ * This hook calls the `authProvider.handleLoginCallback()` method. This is meant to be used in a route called
+ * by an external authentication service (e.g. Auth0) after the user has logged in.
+ * By default, it redirects to the `redirectTo` location, home page if undefined.
+ *
+ * @returns The result of the `handleLoginCallback` call. Destructure as { isLoading, data, error, refetch }.
```
fzaninotto, line 12: `{ isLoading, data, error, refetch }` is not the result of the `handleLoginCallback` call

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
```diff
@@ -79,7 +79,6 @@ const useRoutesAndResourcesFromChildren = (
     // We need to know right away wether some resources were declared to correctly
     // initialize the status at the next stop
     const doLogout = useLogout();
-    const { authenticated } = useAuthState();
```
fzaninotto, line 82: This is super weird. Could you do some archeology in the git history to understand why it was put here in the first place?

packages/ra-core/src/types.ts
```diff
@@ -67,9 +67,14 @@ export type AuthProvider = {
     checkError: (error: any) => Promise<void>;
     getIdentity?: () => Promise<UserIdentity>;
     getPermissions: (params: any) => Promise<any>;
+    handleLoginCallback?: <
```
fzaninotto, line 70: I find the name too long. Possible alternatives:

- `handleCallback`
- `afterLogin`
- `processToken`

I think we should emphasize the "token" in the usage doc - the only utility of the redirection is actually to pass the token, right?

packages/ra-ui-materialui/src/AdminUI.tsx
```diff
@@ -27,5 +27,6 @@ AdminUI.defaultProps = {
     catchAll: NotFound,
     loading: LoadingPage,
     loginPage: Login,
+    loginCallbackPage: LoginCallback,
```
fzaninotto, line 30: I think you must do the same in the `react-admin` package

fzaninotto (changes requested):
Encouraging!

docs/Admin.md
```diff
@@ -441,6 +442,24 @@ See The [Authentication documentation](./Authentication.md#customizing-the-login
...
+import MyLoginCallbackPage from './MyLoginCallbackPage';
+
+const App = () => (
+    <Admin loginCallbackPage={MyLoginCallbackPage}>
+        ...
+    </Admin>
+);
+```
+
+You can also disable it completely along with the `/login-callback` route by passing `false` to this prop.
```
septentrion-730n, line 459: Shouldn't we say it is the default value?

```suggestion
You can also disable it completely along with the `/login-callback` route by passing `false` to this prop. This corresponds to the default setting.
```

docs/Admin.md
```diff
@@ -441,6 +442,24 @@ See The [Authentication documentation](./Authentication.md#customizing-the-login
 
 **Tip**: Before considering writing your own login page component, please take a look at how to change the default [background image](./Theming.md#using-a-custom-login-page) or the [MUI theme](#theme). See the [Authentication documentation](./Authentication.md#customizing-the-login-component) for more details.
 
+## `loginCallbackPage`
+
+If you want to customize the LoginCallback page, pass a component of your own as the `loginCallbackPage` prop. React-admin will display this component whenever the `/login-callback` route is called.
```
septentrion-730n, line 447: I would introduce this paragraph with a short usage context

```suggestion
Used for external authentication services callbacks, the LoginCallback page can be customized by passing a component of your own as the `loginCallbackPage` prop. React-admin will display this component whenever the `/login-callback` route is called.
```

packages/ra-core/src/types.ts
djhi: As dicussed there might be more than just a token. I think `handleCallback` is probably the best

docs/Admin.md
djhi: I don't get this addition

djhi: It is not the default value

packages/ra-core/src/auth/useHandleLoginCallback.ts
```diff
@@ -0,0 +1,57 @@
...
+            onError: err => {
+                const redirectTo = (err as AuthRedirectResult)?.redirectTo;
+
+                if (redirectTo === false) {
+                    return;
+                }
+
+                redirect(
+                    err == null || redirectTo === true
+                        ? removeDoubleSlashes(`${basename}/`)
```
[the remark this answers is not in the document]
djhi: If the home page requires auth, users will either be redirected back to the third party auth page or to the login page, won't they?

packages/ra-ui-materialui/src/AdminUI.tsx
djhi: Nope

packages/ra-core/src/auth/useHandleLoginCallback.ts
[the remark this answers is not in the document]
fzaninotto: right, but since they are already authenticated, the third-party auth page will redirect them to the callback URL, therefore endless loop

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
djhi: I think I got it. This has been done to ensure we re-evaluate permissions when users sign in with different accounts. But I believe we have better ways to do that in the `usePermissions` hook that should re-execute on user change

packages/ra-core/src/auth/useHandleLoginCallback.ts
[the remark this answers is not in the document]
djhi: Should we logout first then? Something else?

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
djhi: Added a test to validate permissions are refreshed after users log out and back in with different permissions

cypress/e2e/permissions.cy.js
```diff
@@ -132,4 +132,18 @@ describe('Permissions', () => {
             cy.contains('Role');
         });
     });
+
+    it.only('refreshes permissions after logging out and back in with a different user', () => {
```
fzaninotto, line 136: .only detected

docs/Admin.md
```diff
@@ -41,6 +41,7 @@ Here are all the props accepted by the component:
 - [`theme`](#theme)
 - [`layout`](#layout)
 - [`loginPage`](#loginpage)
+- [`loginCallbackPage`](#logincallbackpage)
```
fzaninotto, line 44: bad name/anchor

docs/Authentication.md
```diff
@@ -82,6 +84,15 @@ Now the admin is secured: The user can be authenticated and use their credential
 
 If you have a custom REST client, don't forget to add credentials yourself.
 
+## Handling External Authentication Services Callbacks
+
+When using external authentication services such as those implementing OAuth, you usually need a callback route. React-admin provides a default one at `/login-callback`. It will call the `AuthProvider.handleLoginCallback` method
+that may validate the params received from the URL and redirect users to any page (the home page by default) afterwards.
+
+It's up to you to decide when to redirect users to the third party authentication service, for instance:
```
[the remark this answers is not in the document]
fzaninotto: I don't see why this section occurs here. It has nothing to do with the login callback

docs/Authentication.md
```diff
@@ -82,6 +82,59 @@ Now the admin is secured: The user can be authenticated and use their credential
 
 If you have a custom REST client, don't forget to add credentials yourself.
 
+## Handling External Authentication Services Callbacks
+
+When using external authentication services such as those implementing OAuth, you usually need a callback route. React-admin provides a default one at `/login-callback`. It will call the `AuthProvider.handleLoginCallback` method that may validate the params received from the URL and redirect users to any page (the home page by default) afterwards.
```
fzaninotto, line 87: please add details about the expected return type (for redirections)

packages/ra-core/src/auth/index.ts
```diff
@@ -15,6 +15,7 @@ export * from './types';
 export * from './useAuthenticated';
 export * from './useCheckAuth';
 export * from './useGetIdentity';
+export * from './useHandleAuthCallback';
```
fzaninotto, line 18: I don't understand why is it called `useHandleAuthCallback` and the authProvider method is called `handleLoginCallback` ("Auth" / "Login")

fzaninotto (changes requested):

docs/Authentication.md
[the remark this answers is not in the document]
djhi: Shouls I introduce a new dedicated page to third party authentication services ?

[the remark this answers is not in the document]
fzaninotto: at least a section in the Authentication introduction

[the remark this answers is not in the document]
djhi: This is the authentication introduction page

docs/Authentication.md
djhi: Done in https://github.com/marmelab/react-admin/pull/8457/files#diff-b3d8b3b7ae1bfb881040b85c6ce05bad019fe65b83538fb0c80f3aef787ac8e9

packages/ra-core/src/auth/useHandleAuthCallback.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+                    <QueryClientProvider client={queryClient}>
+                        <TestComponent customError />
+                    </QueryClientProvider>
+                </AuthContext.Provider>
+            </HistoryRouter>
+        );
+        await waitFor(() => {
+            expect(redirect).toHaveBeenCalledWith('/test');
+        });
+    });
```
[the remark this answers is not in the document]
slax57: Besides, the way `useHandleAuthCallback.ts` is written, I think the default `onError` will be overridden altogether with the custom one, hence disabling the logout and the redirection.

packages/ra-core/src/types.ts
```diff
@@ -67,9 +67,17 @@ export type AuthProvider = {
     checkError: (error: any) => Promise<void>;
     getIdentity?: () => Promise<UserIdentity>;
     getPermissions: (params: any) => Promise<any>;
+    handleCallback?: <
+        HandleCallbackResult = AuthRedirectResult
+    >() => Promise<HandleCallbackResult | void>;
```
slax57, line 72: Why do you need a generic type here?

Besides, I believe you should type the error too (error type was `void | { redirectTo?: string, logoutOnFailure?: boolean, message?: string }` if I'm not mistaken).

packages/ra-ui-materialui/src/auth/AuthError.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const translate = useTranslate();
+    return (
+        <Root className={className} {...rest}>
+            <div className={AuthErrorClasses.message}>
+                <h1>{translate(title, { _: title })}</h1>
+                <div>{translate(message, { _: message })}</div>
+                <Button label="ra.auth.sign_in">
+                    <LockIcon />
+                </Button>
```
slax57, line 24: Does this button do anything? :thinking:

packages/ra-ui-materialui/src/auth/LoginCallback.tsx
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @example
+ *     import MyAuthCallbackPage from './MyAuthCallbackPage';
+ *     const App = () => (
+ *         <Admin authCallbackPage={MyAuthCallbackPage} authProvider={authProvider}>
+ *             ...
+ *        </Admin>
+ *     );
+ */
+export const AuthCallback = () => {
```
slax57, line 21: File should be renamed to `AuthCallback.tsx`

packages/ra-ui-materialui/src/auth/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './Login';
+export * from './LoginCallback';
```
slax57, line 2: Aren't you missing the export from `AuthError.tsx` too?

slax57 (changes requested):

packages/ra-core/src/types.ts
djhi: > Why do you need a generic type here?

Do avoid adding `any` as we did on `login`

slax57: Btw, this also means I was wrong about adding `message` to `AuthRedirectResult`, I think we souldn't

slax57 (approved):

slax57 (approved):

packages/ra-core/src/core/CoreAdminUI.tsx
```diff
@@ -70,6 +72,14 @@ export const CoreAdminUI = (props: CoreAdminUIProps) => {
             {LoginPage !== false && LoginPage !== true ? (
                 <Route path="/login" element={createOrGetElement(LoginPage)} />
             ) : null}
+
+            {LoginCallbackPage !== false && LoginCallbackPage !== true ? (
+                <Route
+                    path="/login-callback"
```
djhi, line 78: Should this be `auth-callback` as well?

fzaninotto: yes!

fzaninotto (approved):

fzaninotto:
Awesome!

PR #8459, opened 2022-12-02
Fix `<SavedQueryFilterListItem>` sets wrong value for `displayedFilters` query parameter 

groomain:
Hi,

I get this in my url when I click on a saved filter; 
```
displayedFilters=%5Bobject%20Object%5D
```

I used the same code as in the FilterButton component that work.

Thanks!

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/SavedQueryFilterListItem.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/SavedQueryFilterListItem.tsx
@@ -50,7 +50,9 @@ export const SavedQueryFilterListItem = memo(
                     order: value.sort.order,
                     page: 1,
                     perPage: value.perPage,
-                    displayedFilters: value.displayedFilters,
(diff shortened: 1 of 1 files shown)


fzaninotto:
Bug and fix confirmed, thanks!

packages/ra-ui-materialui/src/list/filter/SavedQueryFilterListItem.tsx
```diff
@@ -50,7 +50,9 @@ export const SavedQueryFilterListItem = memo(
                     order: value.sort.order,
                     page: 1,
                     perPage: value.perPage,
-                    displayedFilters: value.displayedFilters,
+                    displayedFilters: JSON.stringify(
+                        value.displayedFilters
```
fzaninotto, line 54: linter warning

fzaninotto (changes requested):

fzaninotto:
Thanks!

PR #8465, opened 2022-12-04
Add OutlinedInput for correct label handling in SelectArrayInput

sebastianbuechler:
Currently, the label of the `SelectArrayInput` is not handled correctly when pushed upwards:
image
I used the example code from the react-admin docs: https://marmelab.com/react-admin/SelectArrayInput.html#usage

According to this [SO answer](https://stackoverflow.com/questions/53916720/label-of-multiple-select-is-crossed-out-by-the-outline-of-the-input-field/74049905#74049905) this can be fixed directly by MUI: https://mui.com/material-ui/react-select/#MultipleSelectCheckmarks.js

However, I could not test it locally as I was not able to include this fix in my project. I believe I don't properly understand the structure of the react-admin repo in order to consume my fix. I tried it with `yarn add https://gitpkg.now.sh/sebastianbuechler/react-admin/packages/react-admin`, but on startup, my project can't find react-admin.

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
@@ -22,6 +22,7 @@ import {
 } from 'ra-core';
 import { InputHelperText } from './InputHelperText';
 import { FormControlProps } from '@mui/material/FormControl';
+import OutlinedInput from '@mui/material/OutlinedInput';
 
 import { LinearProgress } from '../layout';
 import { CommonInputProps } from './CommonInputProps';
@@ -253,6 +254,7 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                 <Select
                     autoWidth
                     labelId={`${label}-outlined-label`}
+                    input={<OutlinedInput label={label} />}
                     multiple
                     error={
                         !!fetchError || ((isTouched || isSubmitted) && invalid)


sebastianbuechler:
It seems that it does not have the desired effect when opening in the StoryBook:
image

Could it be that react-admin applies some different default styling to the `Select` component of MUI?

sebastianbuechler:
> Just to be sure, could you clear your localStorage and see if the issue is still there? In the past we had reports of themes being broken because of outdated versions in the store. Thanks

I just tested it with a cleared localStorage and cache, but same result.

If I go to this page: https://mui.com/material-ui/react-select/#MultipleSelectCheckmarks.js And remove the line `input={<OutlinedInput label="Tag" />}` I get the same result:
image

slax57:
@sebastianbuechler 

I failed to reproduce the issue by using the `outlined` variant on a `<SelectArrayInput>`.

Here is the code I used (inside the SelectArrayInput storybook):

```tsx
export const Variant = () => (
    <AdminContext i18nProvider={i18nProvider}>
        <Create
            resource="users"
            record={{ roles: ['u001', 'u003'] }}
            sx={{ width: 600 }}
        >
            <SimpleForm>
                <SelectArrayInput
                    source="roles"
                    choices={[
                        { id: 'admin', name: 'Admin' },
                        { id: 'u001', name: 'Editor' },
                        { id: 'u002', name: 'Moderator' },
                        { id: 'u003', name: 'Reviewer' },
                    ]}
                    sx={{ width: 300 }}
                    options={{ variant: 'outlined' }}
                />
            </SimpleForm>
        </Create>
    </AdminContext>
);
```

Could you provide a repro of the issue?

Besides, I'm a bit worried that your fix will only work for `outlined` variants. What about other variants?

Lastly, if your PR aims at fixing a bug, it should rather target the `master` branch.

Thanks.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -253,6 +254,7 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                 <Select
                     autoWidth
                     labelId={`${label}-outlined-label`}
+                    input={<OutlinedInput label={label} />}
```
slax57, line 257: I'm afraid this will only work with the outlined variant. How about other variants?

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -22,6 +22,7 @@ import {
 } from 'ra-core';
 import { InputHelperText } from './InputHelperText';
 import { FormControlProps } from '@mui/material/FormControl';
+import OutlinedInput from '@mui/material/OutlinedInput';
```
slax57, line 25: Please put this import along the other ones from `'@mui/material'`

slax57 (changes requested):

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
sebastianbuechler: Will do in the next iteration.

sebastianbuechler:
> @sebastianbuechler
> 
> I failed to reproduce the issue by using the `outlined` variant on a `<SelectArrayInput>`.
> 
> Here is the code I used (inside the SelectArrayInput storybook):
> 
> ```tsx
> export const Variant = () => (
>     <AdminContext i18nProvider={i18nProvider}>
>         <Create
>             resource="users"
>             record={{ roles: ['u001', 'u003'] }}
>             sx={{ width: 600 }}
>         >
>             <SimpleForm>
>                 <SelectArrayInput
>                     source="roles"
>                     choices={[
>                         { id: 'admin', name: 'Admin' },
>                         { id: 'u001', name: 'Editor' },
>                         { id: 'u002', name: 'Moderator' },
>                         { id: 'u003', name: 'Reviewer' },
>                     ]}
>                     sx={{ width: 300 }}
>                     options={{ variant: 'outlined' }}
>                 />
>             </SimpleForm>
>         </Create>
>     </AdminContext>
> );
> ```
> 
> Could you provide a repro of the issue?
> 
> Besides, I'm a bit worried that your fix will only work for `outlined` variants. What about other variants?
> 
> Lastly, if your PR aims at fixing a bug, it should rather target the `master` branch.
> 
> Thanks.

Thank you for your reply. 

Here's the replication of the issue: https://stackblitz.com/edit/github-rmhseu-mj4jem?file=src/posts/PostList.tsx

I was not able to easily reproduce it on Storybook, but maybe it would be helpful to have the editable code editor for react-admin storybooks: https://storybook.js.org/addons/storybook-addon-code-editor

Also, you're right, my "fix" would break other variants and I was not aware of them, as we only use the outlined variant. 

I found a temporary fix by using:
```typescript
        <SelectArrayInput
          source="roles"
          variant="outlined"
          choices={[
            { id: 'admin', name: 'Admin' },
            { id: 'u001', name: 'Editor' },
            { id: 'u002', name: 'Moderator' },
            { id: 'u003', name: 'Reviewer' },
          ]}
          options={{ input: <OutlinedInput label="Roles" /> }}
        />
```
Obviously, this is suboptimal, as the label has to be set twice and it's not intuitive to find this solution.

Interestingly, this bug does not happen for `SelectArray` with the same outlined variant (I also added it in the Stackblitz).

Do you have any idea how to solve this?

slax57:
@sebastianbuechler

I was able to reproduce the issue, and I agree there is something wrong on MUI's side.

As stated in [their docs](https://mui.com/material-ui/react-select/#props), it should choose the correct input variant automatically:

> It extends the [text field components](https://mui.com/material-ui/react-text-field/) sub-components, either the [OutlinedInput](https://mui.com/material-ui/api/outlined-input/), [Input](https://mui.com/material-ui/api/input/), or [FilledInput](https://mui.com/material-ui/api/filled-input/), depending on the variant selected.

But in their [examples with multiple values](https://mui.com/material-ui/react-select/#default), they needed to add `input={<OutlinedInput label="Name" />}` manually...

IMO this is rather something they should fix on their side, so I'd recommend opening an issue on MUI's repo instead.

If you are willing to provide a workaround in the meantime for RA, here is what I got so far.

In `SelectArrayInput.tsx`, adding the following lines fixes the label for variant "outlined" only.

```diff
                <Select
                    autoWidth
                    labelId={`${label}-outlined-label`}
                    multiple
                    error={
                        !!fetchError || ((isTouched || isSubmitted) && invalid)
                    }
                    renderValue={(selected: any[]) => (
                        <div className={SelectArrayInputClasses.chips}>
                            {selected
                                .map(item =>
                                    (allChoices || []).find(
                                        choice =>
                                            getChoiceValue(choice) === item
                                    )
                                )
                                .filter(item => !!item)
                                .map(item => (
                                    <Chip
                                        key={getChoiceValue(item)}
                                        label={renderMenuItemOption(item)}
                                        className={SelectArrayInputClasses.chip}
                                        size="small"
                                    />
                                ))}
                        </div>
                    )}
                    data-testid="selectArray"
                    size="small"
+                   input={
+                       variant === 'outlined' ? (
+                           <OutlinedInput
+                               label={
+                                   <FieldTitle
+                                       label={label}
+                                       source={source}
+                                       resource={resource}
+                                       isRequired={isRequired}
+                                   />
+                               }
+                           />
+                       ) : undefined
+                   }
                    {...field}
                    {...options}
                    onChange={handleChangeWithCreateSupport}
                    value={field.value || []}
                >
                    {finalChoices.map(renderMenuItem)}
                </Select>
```

But I'd recommend making sure this kind of workaround is not needed for other variants too, for example by adding new stories to the storybook.

slax57:
Related issue: https://github.com/marmelab/react-admin/issues/8498

sebastianbuechler:
> @sebastianbuechler
> 
> I was able to reproduce the issue, and I agree there is something wrong on MUI's side.
> 
> As stated in [their docs](https://mui.com/material-ui/react-select/#props), it should choose the correct input variant automatically:
> 
> > It extends the [text field components](https://mui.com/material-ui/react-text-field/) sub-components, either the [OutlinedInput](https://mui.com/material-ui/api/outlined-input/), [Input](https://mui.com/material-ui/api/input/), or [FilledInput](https://mui.com/material-ui/api/filled-input/), depending on the variant selected.
> 
> But in their [examples with multiple values](https://mui.com/material-ui/react-select/#default), they needed to add `input={<OutlinedInput label="Name" />}` manually...
> 
> IMO this is rather something they should fix on their side, so I'd recommend opening an issue on MUI's repo instead.
> 
> If you are willing to provide a workaround in the meantime for RA, here is what I got so far.
> 
> In `SelectArrayInput.tsx`, adding the following lines fixes the label for variant "outlined" only.
> 
> ```diff
>                 <Select
>                     autoWidth
>                     labelId={`${label}-outlined-label`}
>                     multiple
>                     error={
>                         !!fetchError || ((isTouched || isSubmitted) && invalid)
>                     }
>                     renderValue={(selected: any[]) => (
>                         <div className={SelectArrayInputClasses.chips}>
>                             {selected
>                                 .map(item =>
>                                     (allChoices || []).find(
>                                         choice =>
>                                             getChoiceValue(choice) === item
>                                     )
>                                 )
>                                 .filter(item => !!item)
>                                 .map(item => (
>                                     <Chip
>                                         key={getChoiceValue(item)}
>                                         label={renderMenuItemOption(item)}
>                                         className={SelectArrayInputClasses.chip}
>                                         size="small"
>                                     />
>                                 ))}
>                         </div>
>                     )}
>                     data-testid="selectArray"
>                     size="small"
> +                   input={
> +                       variant === 'outlined' ? (
> +                           <OutlinedInput
> +                               label={
> +                                   <FieldTitle
> +                                       label={label}
> +                                       source={source}
> +                                       resource={resource}
> +                                       isRequired={isRequired}
> +                                   />
> +                               }
> +                           />
> +                       ) : undefined
> +                   }
>                     {...field}
>                     {...options}
>                     onChange={handleChangeWithCreateSupport}
>                     value={field.value || []}
>                 >
>                     {finalChoices.map(renderMenuItem)}
>                 </Select>
> ```
> 
> But I'd recommend making sure this kind of workaround is not needed for other variants too, for example by adding new stories to the storybook.

Opened a ticket here: https://github.com/mui/material-ui/issues/36164

PR #8466, opened 2022-12-05
Change the definition of sort order to ASC|DESC

zhujinxuan:
Currently `order` in `SortPayload` type definition is a general string, but it shall be `ASC` or `DESC` in accordance with the docs and sort clicks in `Datagrid`  This PR changes the definition of `order` to `ASC|DESC`

I also changed some `desc` in spec or demo files to `DESC`, because, if we use `Datagrid` from the react-admin, then our order would be predefined `DESC` or `ASC`, not the small cased `desc` or `asc`.

### DIFF
--- a/examples/demo/src/invoices/InvoiceList.tsx
+++ b/examples/demo/src/invoices/InvoiceList.tsx
@@ -22,7 +22,7 @@ const InvoiceList = () => (
     <List
         filters={listFilters}
         perPage={25}
-        sort={{ field: 'date', order: 'desc' }}
+        sort={{ field: 'date', order: 'DESC' }}
     >
         <Datagrid
             rowClick="expand"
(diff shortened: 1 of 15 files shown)


slax57:
Thanks for your contribution.

You are correct, there are places in the code where we actually expect the value to be one of these, but since it's not currently enforced, I'm afraid that restricting the allowed types may break things. 
We must investigate to make sure that it doesn't break things in some edge cases. 
But it's not top priority for us, so we won't do that in the near future.

fzaninotto:
Hi, and sorry it took us so long to look at your PR. We're ready to move forward. 

I've merged `master` to your branch to resolve conflicts, and now there are new errors. Would you mind taking a look?

Also, as there may be an impact on some codebases, I'd prefer that we publish this change in a minor release. Would you mind rebasing on `next` instead of `master`?

fkowal:
not to be a buzz kill

but this is a pain for ie. [ra-data](https://github.com/hasura/ra-data-hasura)

where there are most sorting options
`desc_null_last`
or multicolumn sort `{field: 'col1,col2', order: 'asc,desc'}`

i'd revise the reasoing behind this change and revert the `ts order: type` back to string

Issue #8191, opened 2022-09-23
cannot pass meta via ReferenceInput/useReferenceInputController


stesie:
**What you were expecting:**

Receiving `meta` in data provider when using `ReferenceInput` and passing in `queryOptions`.

**What happened instead:**

`meta` is undefined.

`ReferenceInput` passes all props towards `useReferenceInputController`.
Then `useReferenceInputController` destructures its props, among others into `queryOptions`.

Those `queryOptions` are then *all* passed to `useGetList` in https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/input/useReferenceInputController.ts#L101
... however `meta` would have to go to the second parameter object.  Not the third.

Contrary `useListController` is separating those here https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/list/useListController.ts#L48 (and passing separately in turn)

**Steps to reproduce:**

**Related code:**

```
                    <ReferenceInput source="pricelistId" reference="pricelists" queryOptions={{ meta }}>
                        <SelectInput fullWidth optionText="name" disabled />
                    </ReferenceInput>
```

**Other information:**

**Environment**

* React-admin version: 4.2.7
* Last version that did not exhibit the issue (if applicable):
* React version: 18.2.0
* Browser: Chrome 105
* Stack trace (in case of a JS error):

WiXSL:
Thanks for the report!
Indeed, `ReferenceInput` has yet to have this feature.

fzaninotto:
Fixed by #8132

PR #8477, opened 2022-12-07
Fix useReferenceInputController to pass meta to useReference

stesie:
This relates to #8191, reporting that meta isn't passed on with `ReferenceInput`.  The issue mentions `useGetList` only, however it also uses `useReference` (which in turn calls `getMany` on the data provider).

So #8192 changed `ReferenceInput` to pass meta to the `getList` call.  However still not using it with the `getMany` call which is triggered to resolve an already assigned value.

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.ts
@@ -116,6 +116,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         reference,
         options: {
             enabled: currentValue != null && currentValue !== '',
+            meta,
         },
     });
     // add current value to possible sources
--- a/packages/ra-core/src/controller/useReference.ts
+++ b/packages/ra-core/src/controller/useReference.ts
@@ -46,11 +46,12 @@ export interface UseReferenceResult<RecordType extends RaRecord = any> {
 export const useReference = <RecordType extends RaRecord = any>({
     reference,
     id,
-    options,
+    options = {},
 }: UseReferenceProps<RecordType>): UseReferenceResult<RecordType> => {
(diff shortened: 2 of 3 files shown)


packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -116,6 +116,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         reference,
         options: {
             enabled: currentValue != null && currentValue !== '',
+            meta,
```
slax57, line 119: Why not pass it the other query options as well?

slax57 (commented):
Otherwise your fix seems good, thanks!

packages/ra-core/src/controller/input/useReferenceInputController.ts
stesie: @slax57 those are not compatible typewise. `queryOptions` is currently typed as `UseQueryOptions<{
        data: RecordType[];
        total?: number;
        pageInfo?: {
            hasNextPage?: boolean;
            hasPreviousPage?: boolean;
        };
    }>` i.e. providing data with pagination info (which matches the `getList` result of the data provider just fine).

Contrary query options of `useReference` are typed as `UseQueryOptions<RecordType[], Error>;`, i.e. just an array of records, but no pagination info.

So if we'd want to provide the full set of react-query options to `useReference` then we might need to add another prop to `useReferenceInputController` that's specific to the item call.  But I don't feel that's feasible

slax57: Alright, let's keep it that way for now then.

slax57 (approved):

PR #8490, opened 2022-12-10
Fix incorrect `<Ready>` page when dynamically load routes with no resources

smeng9:
As a follow up of feature https://github.com/marmelab/react-admin/pull/7609 under the use case of defining an admin without resources but routes,

There seems to be a bug if additional routes are _lazily_ added to the list of child routes when user navigates in the app.

It triggers a rerender https://github.com/marmelab/react-admin/blame/master/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx#L123 and it incorrectly redirects the user to `<Ready />` component

This pull requests adds additional checks similar to https://github.com/marmelab/react-admin/blame/master/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx#L253

### DIFF
--- a/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.spec.tsx
+++ b/packages/ra-core/src/core/useConfigureAdminRouterFromChildren.spec.tsx
@@ -1,11 +1,14 @@
 import * as React from 'react';
 import { render, screen, waitFor } from '@testing-library/react';
 import expect from 'expect';
+import { Route } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
+
 import { useResourceDefinitions } from './useResourceDefinitions';
 import { CoreAdminContext } from './CoreAdminContext';
 import { CoreAdminRoutes } from './CoreAdminRoutes';
 import { Resource } from './Resource';
+import { CustomRoutes } from './CustomRoutes';
 import { CoreLayoutProps } from '../types';
 import { AuthProvider, ResourceProps } from '..';
 
@@ -31,6 +34,7 @@ const MyLayout = ({ children }: CoreLayoutProps) => (
 );
 const CatchAll = () => <div />;
 const Loading = () => <>Loading</>;
+const Ready = () => <>Ready</>;
 
 const TestedComponent = ({ role }) => {
     const history = createMemoryHistory();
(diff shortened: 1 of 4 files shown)


smeng9:
Needs to fix implementation of `<Layout>` for ra-no-code because if there are any `<CustomRoutes>` in https://github.com/marmelab/react-admin/blob/0796893e4989716cb6e11a5cfef767a2ed4c263e/packages/ra-no-code/src/Admin.tsx#L41 the `<Ready>` with import CSV pop up will not be shown. (which now should be the correct behavior, previous behavior is incorrect.) The `<Ready>` needs to be moved to `<Layout>`

packages/ra-no-code/src/ui/Layout.tsx
```diff
@@ -1,8 +1,17 @@
...
 
-export const Layout = (props: LayoutProps) => (
-    <RaLayout {...props} appBar={AppBar} menu={Menu} />
-);
+export const Layout = (props: LayoutProps) => {
+    const [resources] = useResourcesConfiguration();
+    const hasResources = !!resources && Object.keys(resources).length > 0;
+
+    if (!hasResources) {
+        return <Ready />;
```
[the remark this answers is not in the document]
smeng9: The ready is removed from `packages/ra-no-code/src/Admin.tsx`

smeng9:
> But it would be nice to have some tests to prove that it works, and cover this case in the future. Could you add some?

I have added a test case so the initial state of lazyRoutes is empty and later is set to a route by useEffect. It triggers a rerender of the hook useConfigureAdminRouterFromChildren.

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.spec.tsx
```diff
@@ -185,6 +189,30 @@ const TestedComponentWithPermissions = () => {
...
+            <CoreAdminRoutes
+                layout={MyLayout}
+                catchAll={CatchAll}
+                loading={Loading}
+                ready={Ready}
+            >
+                <CustomRoutes noLayout>
+                    <Route path="/foo" element={<div>Foo</div>} />
+                </CustomRoutes>
+                <CustomRoutes>{...lazyRoutes}</CustomRoutes>
```
slax57, line 210: Thanks for adding a test case.
However I don't fully understand it. Why don't you provide only dynamic routes?
Besides, when removing the fix, the test still passes, so I believe it should be written differently.

Can you fix it?
Thanks

slax57 (changes requested):
Also, there is a linter warning in `packages/ra-no-code/src/Admin.tsx` line 15:

> 'Ready' is defined but never used

packages/ra-core/src/core/useConfigureAdminRouterFromChildren.spec.tsx
smeng9: I have improved the test cases so when removing the fix the following case won't pass:
Screenshot from 2022-12-14 14-19-04

slax57 (approved):
Thanks!

PR #8496, opened 2022-12-12
Use react-query for useAuthState and useAuthenticated

djhi:


packages/ra-core/src/core/CustomRoutes.spec.tsx
```diff
@@ -15,7 +15,7 @@ describe('<CustomRoutes>', () => {
         expect(screen.queryByText('Password recovery')).not.toBeNull();
     });
 
-    test('should render custom routes that need authentication only when authenticated', async () => {
+    test.only('should render custom routes that need authentication only when authenticated', async () => {
```
slax57, line 18: test.only spotted

slax57 (changes requested):
otherwise, all good :slightly_smiling_face:

packages/ra-core/src/core/CustomRoutes.spec.tsx
djhi:
```suggestion
    test('should render custom routes that need authentication only when authenticated', async () => {
```

slax57 (approved):

PR #8497, opened 2022-12-12
[TypeScript] Fix return type of `useShowContext`, `useEditContext`, and `useCreateContext`

fzaninotto:
The Partial doesn't exist in useListContext, and is wrong. It leads to TypeScript error when accessing any property from the context without testing its existence first.

### DIFF
--- a/packages/ra-core/src/controller/create/useCreateContext.tsx
+++ b/packages/ra-core/src/controller/create/useCreateContext.tsx
@@ -24,7 +24,7 @@ import { CreateControllerResult } from './useCreateController';
  */
 export const useCreateContext = <RecordType extends RaRecord = RaRecord>(
     props?: Partial<CreateControllerResult<RecordType>>
-): Partial<CreateControllerResult<RecordType>> => {
+): CreateControllerResult<RecordType> => {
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/controller/create/useCreateContext.tsx
```diff
@@ -24,7 +24,7 @@ import { CreateControllerResult } from './useCreateController';
  */
 export const useCreateContext = <RecordType extends RaRecord = RaRecord>(
     props?: Partial<CreateControllerResult<RecordType>>
-): Partial<CreateControllerResult<RecordType>> => {
```
djhi, line 27: record does not need an `id` here though

fzaninotto: `CreateControllerResult` already does that job with `record?: Partial<RecordType>`:

https://github.com/marmelab/react-admin/blob/f99917435abd73ece4b38f75420028e81ff578c9/packages/ra-core/src/controller/create/useCreateController.ts#L195-L206

djhi (approved):

PR #8500, opened 2022-12-13
Referenced items won't be selected when IDs type differ in `<ReferenceArrayInput>` and `<AutoCompleteInput>`

### LINKED ISSUE #8499: Referenced items won't be selected when IDs type differ in `<ReferenceArrayInput>` and `<AutoCompleteInput>`
**What you were expecting:**

Inputs handling arrays of IDs such as `<ReferenceArrayInput>` and `<AutoCompleteInput>` should allow types discrepancies in IDs types. For now, comparing ID "2" and 2 produces unexpected array of behaviours often reported in support or forums.

**Related code:**

[Demo application](https://stackblitz.com/edit/github-b5t57y?file=src%2Findex.tsx)

For the Artists resource, IDs are typed as `number` in the dataProvider `getList` method.
In the Bands resource, members are typed as `(number | string)[]`.

Please note that members defined a string typed ID are not selected as initial display. 

```ts
const fakeData = {
        bands: [{ id: 1, name: 'band_1', members: [1, "2"] }],
        artists: [
            { id: 1, name: 'artist_1' },
            { id: 2, name: 'artist_2' },
            { id: 3, name: 'artist_3' },
        ],
    };
```

Members inputs should display two artists of IDs '#1' and '#2'.

septentrion-730n:
Fix #8499

- [x] Fix behaviour
- [x] Add storybook stories
- [x] Add tests

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
@@ -18,6 +18,7 @@ import {
 import { TextField } from './TextField';
 import { SingleFieldList } from '../list';
 import { AdminContext } from '../AdminContext';
+import { DifferentIdTypes } from './ReferenceArrayField.stories';
 
 const theme = createTheme({});
 
@@ -78,7 +79,7 @@ describe('<ReferenceArrayField />', () => {
             </MemoryRouter>
         );
         expect(queryAllByRole('progressbar')).toHaveLength(0);
-        expect(container.firstChild.textContent).not.toBeUndefined();
+        expect(container.firstChild?.textContent).not.toBeUndefined();
         expect(getByText('hello')).not.toBeNull();
         expect(getByText('world')).not.toBeNull();
     });
@@ -106,7 +107,7 @@ describe('<ReferenceArrayField />', () => {
             </ThemeProvider>
         );
         expect(queryAllByRole('progressbar')).toHaveLength(0);
-        expect(container.firstChild.textContent).toBe('');
+        expect(container.firstChild?.textContent).toBe('');
     });
 
     it('should support record with string identifier', () => {
@@ -138,7 +139,7 @@ describe('<ReferenceArrayField />', () => {
             </MemoryRouter>
         );
         expect(queryAllByRole('progressbar')).toHaveLength(0);
-        expect(container.firstChild.textContent).not.toBeUndefined();
+        expect(container.firstChild?.textContent).not.toBeUndefined();
         expect(getByText('hello')).not.toBeNull();
         expect(getByText('world')).not.toBeNull();
     });
@@ -300,5 +301,13 @@ describe('<ReferenceArrayField />', () => {
         await screen.findByText('tag:programming');
         await screen.findByText('tag:management');
         await screen.findByText('tag:design');
+    });
+
+    it('should handle IDs of different types', async () => {
+        render(<DifferentIdTypes />);
+
+        expect(screen.findByText('artist_1')).not.toBeNull();
+        expect(screen.findByText('artist_2')).not.toBeNull();
+        expect(screen.findByText('artist_3')).not.toBeNull();
     });
 });
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.stories.tsx
@@ -1,0 +1,42 @@
+import * as React from 'react';
+import fakeRestProvider from 'ra-data-fakerest';
+
+import { AdminContext } from '../AdminContext';
+import { Datagrid, List } from '../list';
+import { ReferenceArrayField } from './ReferenceArrayField';
+import { TextField } from './TextField';
+
+const fakeData = {
+    bands: [{ id: 1, name: 'band_1', members: [1, '2', '3'] }],
+    artists: [
+        { id: 1, name: 'artist_1' },
+        { id: 2, name: 'artist_2' },
+        { id: 3, name: 'artist_3' },
+        { id: 4, name: 'artist_4' },
+    ],
+};
+const dataProvider = fakeRestProvider(fakeData, true);
+
+export default { title: 'ra-ui-materialui/fields/ReferenceArrayField' };
+
+export const HandlingIdsDiscrependies = () => {
+    return (
+        <AdminContext dataProvider={dataProvider}>
+            <List resource="bands" sx={{ width: 600 }}>
+                <Datagrid>
+                    <TextField source="name" fullWidth />
+                    <ReferenceArrayField
+                        fullWidth
+                        source="members"
+                        reference="artists"
+                    >
+                        <Datagrid>
+                            <TextField source="id" />
+                            <TextField source="name" />
+                        </Datagrid>
+                    </ReferenceArrayField>
+                </Datagrid>
+            </List>
+        </AdminContext>
+    );
+};
--- a/packages/ra-ui-materialui/src/field/ReferenceArrayField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceArrayField.stories.tsx
@@ -2,9 +2,11 @@ import * as React from 'react';
 import fakeRestProvider from 'ra-data-fakerest';
 
 import { AdminContext } from '../AdminContext';
-import { Datagrid, List } from '../list';
+import { Datagrid } from '../list';
 import { ReferenceArrayField } from './ReferenceArrayField';
 import { TextField } from './TextField';
+import { Show } from '../detail';
+import { CardContent } from '@mui/material';
 
 const fakeData = {
     bands: [{ id: 1, name: 'band_1', members: [1, '2', '3'] }],
@@ -15,28 +17,27 @@ const fakeData = {
         { id: 4, name: 'artist_4' },
     ],
 };
-const dataProvider = fakeRestProvider(fakeData, true);
 
 export default { title: 'ra-ui-materialui/fields/ReferenceArrayField' };
 
-export const HandlingIdsDiscrependies = () => {
+export const DifferentIdTypes = () => {
     return (
-        <AdminContext dataProvider={dataProvider}>
-            <List resource="bands" sx={{ width: 600 }}>
-                <Datagrid>
+        <AdminContext dataProvider={fakeRestProvider(fakeData, false)}>
+            <CardContent>
+                <Show resource="bands" id={1} sx={{ width: 600 }}>
                     <TextField source="name" fullWidth />
                     <ReferenceArrayField
                         fullWidth
                         source="members"
                         reference="artists"
                     >
-                        <Datagrid>
+                        <Datagrid bulkActionButtons={false}>
                             <TextField source="id" />
                             <TextField source="name" />
                         </Datagrid>
                     </ReferenceArrayField>
-                </Datagrid>
-            </List>
+                </Show>
+            </CardContent>
         </AdminContext>
     );
 };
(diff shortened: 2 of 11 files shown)


packages/ra-ui-materialui/src/input/ReferenceArrayInput.spec.tsx
```diff
@@ -244,4 +249,52 @@ describe('<ReferenceArrayInput />', () => {
...
+        ).toBeNull();
+    });
+
+    it('should unselect a value whose id type is inconsistant', async () => {
+        render(<HandlingTypesDiscrepencies />);
+
+        const unselect = (queriedText: string) => () => {
+            const chip = screen.queryByText(queriedText, {
+                selector: '.MuiChip-label',
+            })?.nextSibling;
```
slax57, line 278: if you expect the element to be there, you should use `getByText` instead of `queryByText`

Have a look at [this cheatsheet](https://testing-library.com/docs/react-testing-library/cheatsheet/#queries) to learn more about the differences.

packages/ra-ui-materialui/src/input/ReferenceArrayInput.spec.tsx
```diff
@@ -244,4 +249,52 @@ describe('<ReferenceArrayInput />', () => {
...
+            fireEvent.click(chip);
+        };
+
+        await waitFor(unselect('#1'));
+        await waitFor(unselect('#2'));
```
slax57, line 283: I'm not fond of the way you wrote this. To me you should 
1. wait for the element to be ready with `await screen.findByXXX()`
2. then call `unselect` on it, without the need for `waitFor`

`waitFor` should be reserved for assertions and cases that cannot be covered by `await screen.findByXXX()`. Using it to wait for an action not to fail (which will lead to the action being attempted multiple times) is a bad practice IMHO.

packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx
```diff
@@ -575,4 +576,40 @@ describe('<SelectArrayInput />', () => {
...
+    });
+
+    it('should unselect values when ids type are inconsistant', async () => {
+        render(<HandlingTypesDiscrepencies />);
+
+        await waitFor(() => {
+            expect(
+                screen.queryByText('resources.bands.fields.members')
+            ).not.toBeNull();
+        });
```
slax57, line 596: same

packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
```diff
@@ -99,3 +101,29 @@ export const CreateProp = () => (
         </Create>
     </AdminContext>
 );
+
+export const HandlingTypesDiscrepencies = () => {
```
slax57, line 105:
```suggestion
export const DifferentIdTypes = () => {
```

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -159,6 +159,16 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
...
+                eventOrChoice.target.value = eventOrChoice.target.value.reduce(
+                    (acc, value) => {
+                        // eslint-disable-next-line eqeqeq
+                        const index = acc.findIndex(v => v == value);
+                        return index < 0
+                            ? [...acc, value]
+                            : [...acc.slice(0, index), ...acc.slice(index + 1)];
+                    },
+                    []
+                );
```
slax57, line 171: what the heck is this supposed to do??

At least it requires some comments, but I even fear it's wrong (why do you remove the value if it was found in the accumulator?)

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -193,7 +203,8 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
     const renderMenuItemOption = useCallback(
         choice =>
             !!createItem &&
-            choice?.id === createItem.id &&
+            // eslint-disable-next-line eqeqeq
+            choice?.id == createItem.id &&
```
slax57, line 207: why not compare strings here?

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -209,7 +220,8 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                     disabled={getDisableValue(choice)}
                 >
                     {renderMenuItemOption(
-                        !!createItem && choice?.id === createItem.id
+                        // eslint-disable-next-line eqeqeq
+                        !!createItem && choice?.id == createItem.id
```
slax57, line 224: same,

Besides, some of the code you changed is related to the ability to create a new item. I believe this should be covered by tests too then.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -262,8 +274,8 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                             {selected
                                 .map(item =>
                                     (allChoices || []).find(
-                                        choice =>
-                                            getChoiceValue(choice) === item
+                                        // eslint-disable-next-line eqeqeq
+                                        choice => getChoiceValue(choice) == item
```
slax57, line 278: same

this probably deserves some tests with nullish values as well

slax57 (changes requested):

packages/ra-ui-materialui/src/field/ReferenceArrayField.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        source="members"
+                        reference="artists"
+                    >
+                        <Datagrid>
+                            <TextField source="id" />
+                            <TextField source="name" />
+                        </Datagrid>
+                    </ReferenceArrayField>
+                </Datagrid>
+            </List>
```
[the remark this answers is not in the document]
septentrion-730n: Agreed

packages/ra-ui-materialui/src/input/ReferenceArrayInput.spec.tsx
```diff
@@ -187,32 +192,32 @@ describe('<ReferenceArrayInput />', () => {
                 .querySelector('input');
 
         await waitFor(() => {
-            expect(getCheckbox1().checked).toEqual(true);
-            expect(getCheckbox2().checked).toEqual(false);
+            expect(getCheckbox1()?.checked).toEqual(true);
+            expect(getCheckbox2()?.checked).toEqual(false);
```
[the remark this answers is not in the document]
septentrion-730n: Just a boy scout edition to remove a few typescript errors. Shouldn't I?

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -791,3 +800,74 @@ export const EmptyText = () => (
...
+                fans specify a prefered artist. <code>prefer</code> value is
+                evaluated again artists IDs.
+            </Typography>
+            <List resource="fans" actions={<></>}>
+                <Datagrid
+                    body={<NullishValuesDatagridBody />}
+                    bulkActionButtons={false}
+                >
+                    <></>
+                </Datagrid>
```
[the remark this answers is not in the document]
septentrion-730n: Now it's done, I can say it's the equivalent volume of code.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
septentrion-730n: `eventOrChoice.target.value` can, in this case, be set in this case as `[2, '2', 1, '1']`

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
septentrion-730n: I've been told to wait for the issue 😏

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
septentrion-730n: Left as was

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
septentrion-730n: Left as was.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
septentrion-730n: Added comment.

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
```diff
@@ -301,4 +302,12 @@ describe('<ReferenceArrayField />', () => {
         await screen.findByText('tag:management');
         await screen.findByText('tag:design');
     });
+
+    it('should handle IDs of different types', async () => {
+        render(<DifferentIdTypes />);
+
+        expect(screen.findByText('artist_1')).not.toBeNull();
+        expect(screen.findByText('artist_2')).not.toBeNull();
+        expect(screen.findByText('artist_3')).not.toBeNull();
```
slax57, line 311:
```suggestion
        expect(await screen.findByText('artist_1')).not.toBeNull();
        expect(await screen.findByText('artist_2')).not.toBeNull();
        expect(await screen.findByText('artist_3')).not.toBeNull();
```

Don't forget to make sure your test actually fails when you provide wrong values too :wink:

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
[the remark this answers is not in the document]
slax57: I think it's clearer now :slightly_smiling_face:

packages/ra-ui-materialui/src/input/ReferenceArrayInput.spec.tsx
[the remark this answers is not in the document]
slax57: You did not only change the type, your also changed what the code does.

And since the assertions are `toEqual()` and not `toStrictEqual()` this could change the test assertions too.

You should ignore them or add `ts-ignore` comments if they really bother you IMO

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
slax57: If this case can happen then it should be covered by a test.

I'm not convinced it can happen but event if it did, I tested it manually by changing the 'DifferentIdTypes' story, but this code doesn't work. It doesn't remove duplicates, and worse, it removes the wrong item.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
slax57: For the record, I advise **against** these changes if they are not properly tested.
Ping @fzaninotto

slax57 (changes requested):

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
slax57: Okay, took me a while to actually understand why you need this snippet and how it works.
Now I am convinced that it works and you should keep it.

I'd simply recommend being a little more verbose in the comments because it's really not straightforward:
```js
                // when used with different IDs types, unselection leads to double selection with both types
                // instead of the value being removed from the array
                // e.g. we receive eventOrChoice.target.value = [1, '2', 2] instead of [1] after removing 2
                // this snippet removes a value if it is present twice
```

And it is indeed covered by the test `should unselect values when ids type are different`, so no worries there, my bad!

slax57 (approved):

djhi (approved):

PR #8522, opened 2022-12-19
Fix a docs error in useList

zhujinxuan:
`getGetList` doesn't exist in the lib


fzaninotto:
Thanks! This doc page does indeed have a few more problems, so I opened #8524 that supersedes your PR. 

But don't hesitate to open new PRs in the future!

PR #8523, opened 2022-12-19
Allow user to customize sort function

zhujinxuan:
Sometimes we might want to put some summaries on the top or bottom of the list, for example, 'sum', 'sum of records matching certain criteria'. However, the position of sum rows/records is messed up when we do sorting.   This PR allows user to override the default behavior of `sort` function so user can control the position of such records.

I will make it a PR once I update the docs.


fzaninotto:
I don't understand the problem. Can you give examples?

zhujinxuan:
I will close this PR because I find I need another kind of interface (a decorate function after pagination).  The problem I would like to solve is like the attached photo.  I try to have 'Total' and 'Available' rows on top of the Datagrid, but it would be messed up by sorting and pagination.
Screen Shot 2022-12-19 at 3 20 25 PM

I will submit a new PR because the problem is affected by pagination as well.

PR #8526, opened 2022-12-20
Memoize mutation callbacks

### LINKED ISSUE #8518: Mutation hooks return non-memoized functions
I'm really unsure if it's a bug, a support question or a feature request but I don't see why it should work like that so I categorized it as a bug. The problem is that hooks like `useUpdate` or `useDelete` return a bare function, that is, a function that's not wrapped in `useCallback`. If I use it inside `useEffect` and pass it as a dependency, this effect fires on each render because `useUpdate` returns a new instance of the update function for no obvious reason.

**What you were expecting:**
Mutation hooks return the same memoized function unless their parameters change between renders.

**What happened instead:**
Mutation hooks return new function instances on each render.

**Steps to reproduce:**
Take the update function as `const [update] = useUpdate();`, then use it inside `useCallback` or `useEffect`. React linter would complain that it should be present in the dependency array and if you add it there the callback will update and effect will fire on each render. It's wasteful and might become a problem, for example, if you setup a timer in that effect and cancel it on cleanup. If the user keeps causing rerenders (that shouldn't affect the timer) it might be delayed a lot.

**Other information:**
The offending line is here: https://github.com/marmelab/react-admin/blob/bdff90eeb1e3e7012565c695d6cd7bf5d953d5a6/packages/ra-core/src/dataProvider/useUpdate.ts#L268 IMO, this function should be wrapped in `useCallback` but the list of dependencies might be long (I didn't try it). Same for `useUpdateMany`, `useDelete` and `useCreate`.

A quick and dirty way is to simply ignore the linter and not include the function as a dependency. I'm not sure if it causes any issues but so far it seems to work okay for me despite the warning. Not the cleanest way of course and it's not right: other side-effect hooks such as `useNotify` or `useRefresh` don't have this problem as they're properly wrapped in `useCallback`.

**Environment**

* React-admin version: 4.6.2
* Last version that did not exhibit the issue (if applicable): unsure
* React version: 18.1
* Browser: Brave 1.46.144 Chromium: 108.0.5359.128

rkfg:
Fixes #8518.

### DIFF
--- a/packages/ra-core/src/dataProvider/useCreate.ts
+++ b/packages/ra-core/src/dataProvider/useCreate.ts
@@ -9,6 +9,7 @@ import {
 
 import { useDataProvider } from './useDataProvider';
 import { RaRecord, CreateParams } from '../types';
+import { useEvent } from '../util';
 
 /**
  * Get a callback to call the dataProvider.create() method, the result and the loading state.
@@ -142,7 +143,7 @@ export const useCreate = <
         );
     };
 
-    return [create, mutation];
+    return [useEvent(create), mutation];
 };
 
 export interface UseCreateMutateParams<RecordType extends RaRecord = any> {
--- a/packages/ra-core/src/dataProvider/useDelete.ts
+++ b/packages/ra-core/src/dataProvider/useDelete.ts
@@ -16,6 +16,7 @@ import {
     MutationMode,
     GetListResult as OriginalGetListResult,
 } from '../types';
+import { useEvent } from '../util';
 
 /**
  * Get a callback to call the dataProvider.delete() method, the result and the loading state.
@@ -387,7 +388,7 @@ export const useDelete = <
         }
     };
 
-    return [mutate, mutation];
+    return [useEvent(mutate), mutation];
 };
 
 type Snapshot = [key: QueryKey, value: any][];
--- a/packages/ra-core/src/dataProvider/useDeleteMany.ts
+++ b/packages/ra-core/src/dataProvider/useDeleteMany.ts
@@ -16,6 +16,7 @@ import {
     MutationMode,
     GetListResult as OriginalGetListResult,
 } from '../types';
+import { useEvent } from '../util';
 
 /**
  * Get a callback to call the dataProvider.delete() method, the result and the loading state.
@@ -394,7 +395,7 @@ export const useDeleteMany = <
         }
     };
 
-    return [mutate, mutation];
+    return [useEvent(mutate), mutation];
 };
 
 type Snapshot = [key: QueryKey, value: any][];
--- a/packages/ra-core/src/dataProvider/useUpdate.ts
+++ b/packages/ra-core/src/dataProvider/useUpdate.ts
@@ -16,6 +16,7 @@ import {
     MutationMode,
     GetListResult as OriginalGetListResult,
 } from '../types';
+import { useEvent } from '../util';
 
 /**
  * Get a callback to call the dataProvider.update() method, the result and the loading state.
@@ -418,7 +419,7 @@ export const useUpdate = <
         }
     };
 
-    return [update, mutation];
+    return [useEvent(update), mutation];
 };
 
 type Snapshot = [key: QueryKey, value: any][];
--- a/packages/ra-core/src/dataProvider/useUpdateMany.ts
+++ b/packages/ra-core/src/dataProvider/useUpdateMany.ts
@@ -16,6 +16,7 @@ import {
     MutationMode,
     GetListResult as OriginalGetListResult,
 } from '../types';
+import { useEvent } from '../util';
 import { Identifier } from '..';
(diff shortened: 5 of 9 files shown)


slax57 (approved):
Thanks! It seems good to me, but I'd feel safer having @fzaninotto review that one too :grin:

fzaninotto:
This won't work - because passing an empty array as dependency means the callback will never be updated. The `useLayoutEffect` should always fire, so it should receive no second argument. I lead you on a false track - `useEventCallback` cannot do the trick. You'll need to create a `useEvent` shim:


```jsx

// allow the hook to work in SSR
const useLayoutEffect =
    typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;

/**
 * Alternative to useCallback that doesn't update the callback when dependencies change
 *
 * @see https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback
 * @see https://github.com/facebook/react/issues/14099#issuecomment-440013892
 */
export const useEvent = <Args extends unknown[], Return>(
    fn: (...args: Args) => Return,
): ((...args: Args) => Return) => {
    const ref = React.useRef<(...args: Args) => Return>(() => {
        throw new Error('Cannot call an event handler while rendering.');
    });

    useLayoutEffect(() => {
        ref.current = fn;
        // eslint-disable-next-line react-hooks/exhaustive-deps
    });

    return useCallback((...args: Args) => ref.current(...args), []);
};
```

rkfg:
I think it should always fire: https://github.com/marmelab/react-admin/blob/66e0d78d5008f664e522ac199797d24655ad52a8/packages/ra-core/src/util/useEventCallback.ts#L25
The dependency array here includes the function that we pass and it will be different on each render (that's the problem we're solving after all!), plus all the dependencies we pass (but we pass an empty array so it doesn't matter). Am I missing something?

fzaninotto:
You're right, it will work, but the code isn't clear enough. I think passing a dependency that is always new to force a re-execution of the effect is a useless indirection. We should pass no dependency at all.

rkfg:
Okay, I'll add the shim you've proposed and use it instead, makes sense to signal the intent clearer + it might be useful for other things in the future.

fzaninotto:
wait, in fact you should *replace*  the `useEventCallback` with the shim I suggested above. The current version actually executes at every render because the `fn` parameter is always new. So the dependency array is useless and should be removed. 

Sorry for the back and forth, but this is the part of React I'm the less comfortable with.

rkfg:
No problem, I'm always happy to learn. I'll see what I can do tomorrow.

rkfg:
Just to confirm, I'll also remove the dependencies here while replacing the hook: https://github.com/marmelab/react-admin/blob/ce4f4de3040a73e16cb4c930887d96cd582553ec/packages/ra-core/src/store/useStore.ts#L67
This lambda is recreated on each render so the dependency array is indeed useless. The only situation it'd be useful is if the function that's passed to `useEventCallback` is itself memoized (with `useCallback` for example) but this `useStore` hook is the only one that uses `useEventCallback` so I guess it's safe to make the replacement.

fzaninotto (approved):

fzaninotto:
Seems good to me. I still have to test that it actually reduces re-renders using the React profiler.

fzaninotto:
Happy New Year! I finally took some time to test your PR. 

I didn't find any different while using the simple example with or without your changes. Could you please set up a story that demonstrates the change?

Also, this is technically a new feature, so please rebase to PR against `next`.

rkfg:
Happy New Year! I rebased on next. I suppose it'd be hard to provide a working example, in my case I use a complex component that completely replaces the edit view of a resource, it's also put into its own context so re-renders might be triggered by many things including that context value change.

The code that caused issues was my custom hook that marks the currently opened item as "seen" after 5 seconds if the user doesn't navigate away. So that hook starts a timer in `useEffect` (the effect cleanup function cancels this timer) and when it fires, it updates some data using the callback that `useUpdate` returns. Since the timer works inside `useEffect` and uses that callback I have to pass it as one of the dependencies. If for whatever reason this component re-renders, `useUpdate` returns a different instance of the mutation callback which causes the effect to cleanup (and cancel the timer), then start another timer. I noticed that some re-renders happen when the user scrolls my custom component and that delays the timer. I realize that it's of course not a react-admin's problem, but the only dependency of that `useEffect` that changes is exactly the mutation callback which should retain referential equality between renders, and my patch fixes this.

I'm truly sorry but I can't come up with a natural example of these re-renders to occur during the more or less classic RA usage, the default list/edit/show components don't re-render spontaneously. And my custom component is too big and niche to adapt it for an example. Most of the time the current behavior shouldn't hit users I think but eventually it will.

fzaninotto (approved):

fzaninotto:
Thanks!

Issue #8520, opened 2022-12-19
Infinite year digits when manually entering date inside DateTimeInput/DateInput


rinxma:
This happens on both DateTimeInput and DateInput

To test and reproduce go to demo:
https://marmelab.com/react-admin-demo/#/invoices

Try setting the date manually by entering with keyboard without using the picker:
![](https://i.gyazo.com/f914efcfb981220e40ae7952d8c3d783.png)

The day and month input works fine and accepts 2 digits while the year should accept 4 digits, it accepts infinite number of digits which is a bug that needs to be fixed

This started happening after migrating from v3 to v4.

septentrion-730n:
Thanks for your report!
This is indeed a bug and will be marked as such.

shavidze:
validate={maxValue('12-12-2099')} maybe this should work for DateInput

aniketdevarkar:
I was checking possible solutions on report of this bug and following are the points after analysis,
1. If we check the date picker even that allows user to keep scrolling the year as long as user wants and no upper cap is declared.
2. what if some app/ user  has a use case which need input as year which is more than 2099 or higher.

So, It would be more flexible if we add an optional prop which allows to set maximum date , which can be used to cap the maximum year.

fzaninotto:
This is the way `<input type="date">` works (see https://developer.mozilla.org/fr/docs/Web/HTML/Element/input/date for an example). If you want to limit the max year, add a `max` value. 

So I don't think this is a bug. If it doesn't fit your need, I advise you to use an alternative Date input component.

PR #8534, opened 2022-12-26
 fixed: DateInput years digits length is 4.

shavidze:
https://github.com/marmelab/react-admin/issues/8520 
based on this issue, I made a change into the DateInput component.
As we want to allow users to input only 4 digits to years I just take the last 4 digits user is entering to years.


fzaninotto:
Your PR embarks plenty of useless files, doesn't include a test, and fixes something that I don't think is a bug. So sorry, we won't merge it.

PR #8558, opened 2023-01-09
Add new authProviders (auth0, cognito)

djhi:


docs/AuthProviderList.md
```diff
@@ -7,6 +7,7 @@ title: "Supported Auth Provider Backends"
 
 It's very common that your auth logic is so specific that you'll need to write your own `authProvider`. However, the community has built a few open-source Auth Providers that may fit your need:
 
+- **[Auth0](https://auth0.com/)**: [ra-auth-auth0](https://github.com/marmelab/ra-auth-auth0)
```
fzaninotto, line 10: it is already on the list below (L18), I think you should also remove it from there.

fzaninotto (approved):

Issue #8549, opened 2023-01-05
browser verbose "advice": [DOM] Input elements should have autocomplete attributes (suggested: "username")


jrobbins-LiveData:
**What you were expecting:**
Expecting a clean browser console with no yellow advice.

**What happened instead:**
The `authProvider` feature pops up a `login form` that has a `username` input which fails to provide an `autocomplete=username` attribute.  The absence of this attribute causes Version 108.0.5359.125 (Official Build) (64-bit) of Chrome to log advice in the console that `[DOM] Input elements should have autocomplete attributes (suggested: "username"): (More info: https://goo.gl/9p2vKq) `

**Environment**
Chrome is up to date
Version 108.0.5359.125 (Official Build) (64-bit)

* React-admin version:      "version": "4.6.2",
* Last version that did not exhibit the issue (if applicable):
* React version:   "version": "18.2.0",
* Browser: Chrome Version 108.0.5359.125 (Official Build) (64-bit)

slax57:
I assume you are talking about the simple demo. If not you should provide more info about the reproduction environment.

With Chrome Version 108.0.5359.124 (latest at my disposal) I do not reproduce this issue.

Can you provide more information about how to reproduce it?

Also, please note that you can provide your own Login form using [these instructions](https://marmelab.com/react-admin/Authentication.html#customizing-the-login-component).

jrobbins-LiveData:
Thanks for the quick response!  It must be my slightly newer Chrome version.  When I load your website's [demo](https://marmelab.com/react-admin-demo/#/login) with my Chrome version, I see this in the browser console:

```
[DOM] Input elements should have autocomplete attributes (suggested: "current-password"): (More info: https://goo.gl/9p2vKq) 
```

While this is advice about the password input, and not the username input, it appears to be an instance of the issue I thought I should report.

slax57:
I managed to see the warning by enabling the 'Verbose' log level in the DevTools console.
So yeah, you are right, this is something we could improve. I'll label this issue as possible enhancement and good first issue.
Thanks for raising this up.

Fayaz-B:
Hello, Can I have a go at this issue ?

slax57:
@Fayaz-B sure, would appreciate it!

Fayaz-B:
Its literally my first time in open sourcing, sorry to ask but, I'm finding it difficult to navigate through the project.
```make install``` is giving me an error, I installed make for windows. If possible can I get some guidance.

[...]

PR #8559, opened 2023-01-09
fix autocomplete warning in google console for issue #8549

Fayaz-B:
Set autoComplete attribute to username, which fixed the warning in the developer console.
reference : [Chromium Projects](https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands/)


djhi:
Thanks!

PR #8562, opened 2023-01-09
Modifiable size prop for SelectArrayInput

### LINKED ISSUE #8515: Hardcoded size value for SelectArrayInput
**What you were expecting:**
`SelectArrayInput` to honor defaultProps set for MuiSelect component

**What happened instead:**
`SelectArrayInput` has a hardcoded size prop that is not possible to override

**Steps to reproduce:**
add following to the `components` sections of the theme
```
MuiSelect: {
      defaultProps: {
        variant: 'outlined',
        size: 'medium',
      },
    },
```

**Related code:**
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx#L295

**Other information:**

**Environment**

* React-admin version: latest
* React version: 
* Browser:

amilosmanli:
fixes #8515 

This PR fixes 2 issues with `SelectArrayInput`:

- Removes hardcoded value for `size` prop
- Fixes label display issue when variant is set to `outlined`

`size` prop on `SelectArrayInput` was hardcoded to `small` which was preventing modifying size of the input through MUI's theming system. 

Additionally, when setting `SelectArrayInput` variant to `outlined` label of the input wasn't showing correctly. This is fixed by adding `input` prop with `OutlinedInput` component when variant is set to `outlined`.

Before:
Screenshot 2023-01-09 at 4 33 31 PM

After:
Screenshot 2023-01-09 at 4 34 12 PM

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
@@ -127,3 +127,50 @@ export const DifferentIdTypes = () => {
         </AdminContext>
     );
 };
+
+export const DifferentSizes = () => {
+    const fakeData = {
+        bands: [{ id: 1, name: 'band_1', members: [1, '2'] }],
+        artists: [
+            { id: 1, name: 'artist_1' },
+            { id: 2, name: 'artist_2' },
+            { id: 3, name: 'artist_3' },
+        ],
+    };
+    const dataProvider = fakeRestProvider(fakeData, false);
+    return (
+        <AdminContext dataProvider={dataProvider}>
+            <Edit resource="bands" id={1} sx={{ width: 600 }}>
+                <SimpleForm>
+                    <TextInput source="name" fullWidth />
+                    <SelectArrayInput
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="small"
+                    />
+                    <SelectArrayInput
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="medium"
+                    />
+                    <SelectArrayInput
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="small"
+                        variant="outlined"
+                    />
+                    <SelectArrayInput
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="medium"
+                        variant="outlined"
+                    />
+                </SimpleForm>
+            </Edit>
+        </AdminContext>
+    );
+};
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
@@ -11,6 +11,7 @@ import {
     FormHelperText,
     FormControl,
     Chip,
+    OutlinedInput,
 } from '@mui/material';
 import {
     ChoicesProps,
@@ -108,6 +109,7 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
         optionValue,
         parse,
         resource: resourceProp,
+        size = 'small',
         source: sourceProp,
         translateChoice,
         validate,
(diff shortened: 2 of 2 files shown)


slax57:
This will supersede https://github.com/marmelab/react-admin/pull/8465

amilosmanli:
Hey @slax57 

When do you think you could check and merge this change?

packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
```diff
@@ -127,3 +127,63 @@ export const DifferentIdTypes = () => {
...
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="medium"
+                    />
+                    <SelectArrayInput
+                        fullWidth
+                        source="members"
+                        choices={fakeData.artists}
+                        size="large"
```
slax57, line 162: I'm seing TS errors saying that "Type '"large"' is not assignable to type '"small" | "medium" | undefined'"

Besides, in the storybook I see no difference between `size="medium"` and `size="large"`.

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -108,6 +110,7 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
         optionValue,
         parse,
         resource: resourceProp,
+        size,
```
slax57, line 113: `size` should default to `"false"` to be backwards compatible

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -247,6 +251,24 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
...
+        source,
+    });
+
+    const outlinedInputProps =
+        variant === 'outlined'
+            ? {
+                  input: (
+                      <OutlinedInput
+                          id="select-multiple-chip"
+                          label={labelPlaceholder}
```
slax57, line 266:
```suggestion
                          label={
                              <FieldTitle
                                  label={label}
                                  source={source}
                                  resource={resource}
                                  isRequired={isRequired}
                              />
                          }
```

Otherwise you disable the feature that puts a `'*'` at the end of the label when the field is required

slax57 (changes requested):
Sorry for the late review, our team is a bit short-handed at the moment but hopefully this will get better soon :innocent:

amilosmanli:
@slax57 thanks for the review and the feedback. Addressed all the comments. PTAL

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -116,6 +119,7 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
     } = props;
 
     const inputLabel = useRef(null);
+    const translateLabel = useTranslateLabel();
```
slax57, line 122: `translateLabel` is no longer used so you should remove it

slax57 (changes requested):
Otherwise I'm good with the changes, thanks!

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
amilosmanli: @slax57 removed the ununsed line

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -11,13 +11,15 @@ import {
...
     Chip,
+    OutlinedInput,
 } from '@mui/material';
 import {
     ChoicesProps,
     FieldTitle,
     useInput,
     useChoicesContext,
     useChoices,
+    useTranslateLabel,
```
slax57, line 22: Almost there but you forgot to remove this import too.

I need to request this change because otherwise our CI will raise a warning here that we'll need to fix later, so we might as well do it now.

slax57 (changes requested):

amilosmanli:
@slax57 PTAL
Sorry it took a while

slax57 (approved):

slax57:
Thanks!

davidhenley:
I noticed the label had the same issues in the ReferenceManyInput as well. Does this fix that too?

slax57:
@davidhenley `ReferenceManyInput` does not render a `SelectArrayInput` (unless this is one of the inputs you pass as children of the `SimpleFormIterator` of course), so I think you may have mistaken it with `ReferenceManyToManyInput`.
If you are referring to `ReferenceManyToManyInput` then yes, this PR should fix it as well, provided you upgrade the latest react-admin version of course.

davidhenley:
@slax57 No, this is ReferenceManyInput. The label is off on the outline variant. Just wondered if this fixed that as well, but guess that's another issue.

CleanShot 2023-03-29 at 10 11 23

PR #8565, opened 2023-01-10
[Doc] Add `disableAuthentication` prop description on <Show> component

septentrion-730n:


docs/Show.md
```diff
@@ -236,6 +237,20 @@ const PostShow = props => (
 ```
 {% endraw %}
 
+## `disableAuthentication`
+
+By default, the `<Show>` component will automatically redirect the user to the login page if the user is not authenticated. If you want to disable this behavior and allow anonymous access to a creation page, set the `disableAuthentication` prop to `true`.
```
WiXSL, line 242:
```suggestion
By default, the `<Show>` component will automatically redirect the user to the login page if the user is not authenticated. If you want to disable this behavior and allow anonymous access to a show page, set the `disableAuthentication` prop to `true`.
```

WiXSL (changes requested):

docs/Show.md
WiXSL: 😃

fzaninotto (approved):

Issue #7792, opened 2022-06-07
AutocompleteArrayInput disableCloseOnSelect prop not respected


nerixim:
**What you were expecting:**

`AutocompleteArrayInput` should not close after selection when `disableCloseOnSelect` is `true`.

**What happened instead:**

It closes after each selection.

**Steps to reproduce:**

**Related code:**

[Codesandbox](https://codesandbox.io/s/silly-visvesvaraya-wjg345?file=/src/posts/PostCreate.tsx) (Categories field on create post form)

**Other information:**

**Environment**

* React-admin version: 4.1.2
* Last version that did not exhibit the issue (if applicable):
* React version: 18.1
* Browser:
* Stack trace (in case of a JS error):

septentrion-730n:
Thanks for the report. Bug spotted!

MathiasDierickx:
The reason is because `blurOnSelect` is enabled by default.
Setting both `disableCloseOnSelect` and `blurOnSelect={false}` fixes this.
https://codesandbox.io/s/playground-demo-material-ui-forked-wrjr9v?file=/demo.js

arimet:
Indeed, the mui documentation should at least stipulate that the two props cannot coexist. An [issue](https://github.com/mui/material-ui/issues/34038) has been opened. 
Meanwhile as MathiasDierickx points out, adding `blurOnSelect={false}` to the `AutocompleteArrayInput` fixes the problem

PR #8569, opened 2023-01-11
[Doc] - Add `disableCloseOnSelect` tip to AutoCompleteArrayInput

TurtIeSocks:
Adds an additional tip that may save some future headaches, after having spent a bit of time yesterday trying to troubleshoot this myself. 

Source: #7792


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8574, opened 2023-01-13
Introduce withRefreshAuth

djhi:


fzaninotto:
I don't like the idea of a single decorator for two very different objects. I also see that it forces you to do some borderline inspection to determine in which case you are. How about one for each?

djhi:
> How about one for each?

How would you name them? 
`addRefreshAuthToDataProvider` and `addRefreshAuthToAuthProvider`?

fzaninotto:
yes, fine by me

slax57 (changes requested):

docs/addRefreshAuthToDataProvider.md
```diff
@@ -0,0 +1,83 @@
...
+
+# `addRefreshAuthToDataProvider`
+
+This helper function wraps an existing [`dataProvider`](./DataProviderIntroduction.md) to support authentication token refreshing mechanisms.
+
+## Usage
+
+Use `addRefreshAuthToDataProvider` to decorate an existing data provider. In addition to the base provider, this function takes a function responsible for refreshing the authentication token if needed.
+
+Here is a simple example that refreshes an expired JWT token when needed:
```
slax57, line 14: I wonder if we should add this simple example as its own dedicated paragraph in the Security > Introduction page too. What do you think @fzaninotto ?

slax57 (approved):

docs/addRefreshAuthToDataProvider.md
```diff
@@ -28,52 +28,43 @@ export const refreshAuth = () => {
...
-const myAuthProvider = {
-    // ...Usual AuthProvider methods
-};
-
-export const authProvider = withRefreshAuth(myAuthProvider, refreshAuth);
+export const dataProvider = addRefreshAuthToDataProvider(baseDataProvider, refreshAuth);
 ```
 
-Then, inject the decorated providers in the `<Admin>` component:
+Then, inject the decorated provider in the `<Admin>` component:
```
djhi, line 40:
```suggestion
Then, pass the decorated provider to the `<Admin>` component
```

docs/addRefreshAuthToAuthProvider.md
```diff
@@ -0,0 +1,85 @@
...
+import { refreshAuth } from 'refreshAuth';
+
+const myAuthProvider = {
+    // ...Usual AuthProvider methods
+};
+
+export const authProvider = addRefreshAuthToAuthProvider(myAuthProvider, refreshAuth);
+```
+
+Then, inject the decorated provider in the `<Admin>` component:
```
djhi, line 41:
```suggestion
Then, pass the decorated provider to the `<Admin>` component
```

docs/addRefreshAuthToDataProvider.md
fzaninotto: 👍

djhi: Done

docs/Authentication.md
```diff
@@ -232,6 +232,42 @@ It's up to you to decide when to redirect users to the third party authenticatio
 * Directly in the `AuthProvider.checkAuth()` method as above;
 * When users click a button on a [custom login page](#customizing-the-login-component)
 
+## Handling Refresh Tokens
+
+[Refresh tokens](https://oauth.net/2/refresh-tokens/) are an important security mechanism. In order to leverage them, you should decorate the `dataProvider` and the `authProvider` so that they can check whether the authentication must be refreshed and actually refresh it.
+
+You can use the [`addRefreshAuthToDataProvider`](addRefreshAuthToDataProvider.md) and [`addRefreshAuthToAuthProvider`](./addRefreshAuthToAuthProvider.md) functions for this purpose. They accept a `dataProvider` or `authProvider` respectively and a function that should refresh the authentication token if necessary:
```
slax57, line 239:
```suggestion
You can use the [`addRefreshAuthToDataProvider`](./addRefreshAuthToDataProvider.md) and [`addRefreshAuthToAuthProvider`](./addRefreshAuthToAuthProvider.md) functions for this purpose. They accept a `dataProvider` or `authProvider` respectively and a function that should refresh the authentication token if necessary:
```

slax57 (changes requested):

slax57 (approved):

djhi:
Indeed, we could do a better job on the documentation side. One way to avoid the double refresh would be to store the first promise and return it on subsequent calls until it is resolved. You can check how we did it in `ra-directus` for instance: https://github.com/marmelab/ra-directus/blob/main/packages/ra-directus/src/directusRefreshAuthToken.ts

Maybe we should handle this ourselves so you don't have to worry about it though

fzaninotto:
I'v just changed the repo visibility, you should be able to see the code.

davidhenley:
Thanks @fzaninotto! I see it now

djhi:
```tsx
let refreshPromise = null;

const refreshToken = () => {

  if (refreshPromise != null) {
    return refreshPromise;
  }

  const accessToken = localStorage.getItem('authentication.access_token');
  const refreshToken = localStorage.getItem('authentication.refresh_token');

  if (!accessToken && !refreshToken) {
    return Promise.reject();
  }

  const now = new Date()
  const expired = jwtDecode(accessToken).exp < now.getTime() / 1000

  if (expired) {
    refreshPromise = (async () => {
      const request = new Request(`${process.env.REACT_APP_API_URL}/auth/refresh`, {
        method: 'POST',
        body: JSON.stringify({accessToken, refreshToken}),
        headers: new Headers({'Content-Type': 'application/json'}),
      });
      const response = await fetch(request);
      if (response.status < 200 || response.status >= 300) {
        throw new HttpError(response.statusText, response.status);
      }
      const data = await response.json();
      localStorage.setItem('authentication.access_token', data.token);
      localStorage.setItem('authentication.refresh_token', data.refreshToken);
    })().finally(() => {
      refreshPromise = null;
    });
    return refreshPromise;
  }
}
```
This will work. However, we're going to provide a way to do that automatically in a later minor version.

davidhenley:
@djhi That was it! Thank you

PR #8576, opened 2023-01-14
Optimize subscribe and publish function of memeryStore and localStora…

HEKEH:
Currently, `subscriptions` in `localStorageStore` and `memoryStore` is a two-level object while is a little confusing, and need to traverse each callback element to compare with `key` when use `publish`. 
This PR changes `subscriptions` into a single-level object, and relevant callbacks can be found directly by a key. This makes the code easier to understand, and performance is also a little improved.

### DIFF
--- a/packages/ra-core/src/store/localStorageStore.ts
+++ b/packages/ra-core/src/store/localStorageStore.ts
@@ -1,8 +1,7 @@
 import { Store } from './types';
 
-type Subscription = {
-    key: string;
-    callback: (value: any) => void;
+type Subscriptions = {
+    [key: string]: Set<(value: any) => void>;
 };
 
 const RA_STORE = 'RaStore';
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks! Can you provide before/after performance benchmarks?

HEKEH:
> Thanks! Can you provide before/after performance benchmarks?

Not test it yet, but the previous `publish` function has to traverse every element to find some elements matched with the `key`, but the function modified does not need to, it can find the target callbacks directly, so the performance is a little improved.

HEKEH:
> Thanks! Can you provide before/after performance benchmarks?

Not test it yet, but the previous `publish` function has to traverse every element to find some elements matched with the `key`, but the function modified does not need to, it can find the target callbacks directly, so the performance is a little improved.

HEKEH:
> Thanks! Can you provide before/after performance benchmarks?
The main reason of this modification is that I feel using random number as key is a little weird, and I think it's not really needed. The performance improvement is not the key point.

fzaninotto:
The tests don't pass in your PR, it doesn't "fix" anything, and it hasn't moved in three month. So I'm going to close it. 

If it's not broken, don't fix it™!

PR #8582, opened 2023-01-16
Make handling empty state the responsibility of the `<List>` child

djhi:
## Problem

Currently the `<List>` component is responsible for displaying something when there is no data. However, we introduced the `empty` prop on the `<Datagrid>` component as well. When passing `empty={false}` to the `<List>`, it displays its child and the `empty` provided to the `<Datagrid>` is then used.

However, the `<Pagination>` component also displays something special when there is no data, which is not desirable in this case.

## Solution

Make handling empty state the responsibility of the `<List>` child. Both the `<Datagrid>` and the `<SimpleList>` now accept an `empty` prop. This prop accepts either an element or `false`.

Besides, we deprecated the `limit` prop on the `<Pagination>` component. It now renders `null` when there is no data.

## TODO

- [x] Deprecate the `empty` prop on `<List>`
- [x] Add the `empty` prop on `<SimpleList>`
- [x] Deprecate the `limit` prop on `<Pagination>`
- [x] Make `<Pagination>` renders `null` when there is no data
- [x] Update documentation


packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -207,17 +210,19 @@ export const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
         );
     }
 
+    const shouldRenderEmptyPage =
+        (data == null || data.length === 0 || total === 0) &&
+        filterValues != null && // Can be null when outside a <ListContext>
+        !Object.keys(filterValues).length &&
+        empty !== false;
+
```
djhi, line 218: I added the check for `null` on the `filterValues` to make this works with the `<ArrayField>`. However, I'm not sure about what we should do here. It may be better to display the `empty` in `ArrayField` as well.

djhi:
Should I add an `empty` prop to the `SingleFieldList` component as well?

djhi:
After internal discussion, I'll open another PR with a simpler approach as this has a breaking change and make usage of datagrid in fields awkward

Issue #8572, opened 2023-01-13
Using square brackets in the source description of list filters prevents their hiding/removal


poe42:
We are using DateTimeInputs for aggregating our custom date range filters. Therefore we use Api-Platform's [Date Filter](https://api-platform.com/docs/core/filters/#date-filter), like `before` and `after`, as in:
```
<DateTimeInput source="issuingDate[after]" resettable />
<DateTimeInput source="issuingDate[before]" resettable />
```

The filtering works like a charm (and no additional tweaking of the dataProvider is needed) but the removing/hiding of those filters fails. In this regard, the usage of _square brackets_ leads to some unexpected behaviors. 

**What you were expecting:**
After adding a filter, that uses _square brackets_ in its source description, and successfully filtering a list, this filter should be removed again, when clicking the filter's _hide_ button.

**What happened instead:**

a) Precondition: The filter is dirty (inputs were already made) -> When clicking on the filter's hide button, the according filter does **not** get removed -> Unexpected!
b) Precondition: The filter is dirty (inputs were already made) -> When clicking on the filter's **reset** button on the right, the according filter gets removed -> Unexpected!
c) Precondition: The filter was initiated but no user inputs were made -> When clicking on the filter's hide button, the according filter gets removed -> OK

**Steps to reproduce:**
At Example Simple (v4) go to [PostList](https://codesandbox.io/s/thirsty-newton-mv4n8n?file=/src/posts/PostList.tsx) and add the filter "Published before". Try removing the filter:
a) click _hide_ after setting a date
b) click _reset_ after setting a date
c) click _hide_ right after filter init

**Environment**
* React-admin version: v4
* Browser: firefox, chrome

yanchesky:
It's very likely that I've found the source of the problem. The function `hideFilter` does not properly clear the filter value. That's why when the input is empty hiding works fine. 
I will make a PR soon to solve that issue if nobody minds 😊

djhi:
Thanks for the report. We would gladly accept a pull request!

fzaninotto:
There is something I don't understand. In your codesandbox, the filter that is sent to the API is not 

`{ 'published_at[before]': value }`,

it's

`{ published_at: { before: value } }`

and it is not what your API expects, right?

So you must be some kind of transformation in your API provider to turn the filter to the square bracket syntax. And if this is the case, why can't you use the dot notation in your filter, i.e. `published_at.before`?

fzaninotto:
[...]

PR #8592, opened 2023-01-19
Fix hide filters with brackets in source

yanchesky:
Fixes https://github.com/marmelab/react-admin/issues/8572

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -2,13 +2,12 @@ import get from 'lodash/get';
 import { useCallback, useEffect, useRef } from 'react';
 import isEqual from 'lodash/isEqual';
 
-import { useSafeSetState, removeEmpty } from '../../util';
+import { useSafeSetState, removeEmpty, removeKey } from '../../util';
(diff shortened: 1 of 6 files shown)


yanchesky:
I have thought it through and It may be wrong to change `useList` and `useReferenceManyFieldController` hiding filter implementation. It is `useListParams`'s responsibility to show filters that way. Therefore, only its hide function should work with regard to how the show function works. Changing in other places could break someone's implementation. Am I right or not? 🤔

slax57:
Thanks for providing this PR.
Having read through it rapidly, I believe you are correct, it would rather be `useListParams`'s responsibility imo.

Also, have you seen [@fzaninotto's comment here](https://github.com/marmelab/react-admin/issues/8572#issuecomment-1404949164)?
To me this calls the need for this change into question.
Do you have a use-case that would justify its need?

yanchesky:
I don't really have a use case for this. However, I think that `useListParams`'s **showFilter** and **hideFilter** functions should work complementary. If one function converts dot and bracket notation, then the other should work accordingly.

fzaninotto:
If there is no bug, don't fix it ;)

fzaninotto:
No traction for this, so I'll close this PR. Thanks for your changes anyway, and don't hesitate to open more PRs!

PR #8602, opened 2023-01-27
Fix AutocompleteInput when multiple is true and value is not an array

djhi:


fzaninotto (approved):

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -742,7 +742,7 @@ const getSelectedItems = (
     multiple
 ) => {
     if (multiple) {
-        return (value || [])
+        return (Array.isArray(value || []) ? value : [value])
```
nerixim, line 745: @fzaninotto @djhi 
I'm not sure why I'm getting an empty string here, but I get an error here (`(intermediate value)(intermediate value)(intermediate value).map is not a function`)

```js
const value = ""
Array.isArray(value || []) ? value : [value]
// ''
```

Nullish coalescing seems to fix this particular case, though not sure of any caveats.
```js
const value = ""
Array.isArray(value ?? []) ? value : [value]
// [ '' ]
```

djhi: Good catch

PR #8605, opened 2023-01-28
Fix AutocompleteInput getSelectedItems throwing error on empty string

nerixim:
https://github.com/marmelab/react-admin/pull/8602/files#r1089750166

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -742,7 +742,8 @@ const getSelectedItems = (
     multiple
 ) => {
     if (multiple) {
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -742,7 +742,8 @@ const getSelectedItems = (
     multiple
 ) => {
     if (multiple) {
-        return (Array.isArray(value || []) ? value : [value])
+        const coalesced = value || []
```
djhi, line 745:
```suggestion
        const coalesced = value || [];
```

djhi (changes requested):

djhi:
Thanks! I added some tests in #8608 though and had to change a few things so I'm closing this one

PR #8607, opened 2023-01-30
[Doc] Tutorial: Fix typos, snippets and file extensions

### LINKED ISSUE #8606: Tutorial Posts and authProvider
I think I found two minor bugs with the 4.7.2 Tutorial Page.

1. In the [Handling Relationships](https://marmelab.com/react-admin/Tutorial.html#handling-relationships) section, the "diff" references the existing `users` line, not the new `posts` line as expected.

2. In the [Adding a Login Page](https://marmelab.com/react-admin/Tutorial.html#adding-a-login-page) section, I think that the first code line should include the type of the `const authProvider`. IntelliJ WebStorm (and `npm start`) is flagging the pasted code as not declaring any types for TypeScript.

```diff
< export const authProvider = [
> import {AuthProvider} from "react-admin";
>
> export const authProvider: AuthProvider = {
```

**Steps to reproduce:**
1. image
2. image

**Environment**
Browsing the 4.7.2 Tutorial on Chrome, using the 4.7.2 code in latest IntelliJ WebStorm EAP 231.4840.391.

* React-admin version: 4.7.2
* React version: 18.2.0
* Browser: Chrome

React and TypeScript are not my strong subjects (yet!), sorry if I missed anything.

erikvanderwerf:
Fix #8606

1. Moved the diff marker to the "posts" line, as expected.
2. Minor spell-checks.
3. Updated sample TypeScript in Tutorial to compile with `typescript@4.9.4`

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -459,8 +459,8 @@ import { UserList } from "./users";
 
 const App = () => (
   <Admin dataProvider={dataProvider}>
-+   <Resource name="users" list={UserList} />
-    <Resource name="posts" list={ListGuesser} />
+    <Resource name="users" list={UserList} />
++   <Resource name="posts" list={ListGuesser} />
   </Admin>
 );
 
@@ -755,7 +755,7 @@ export const PostEdit = () => (
 
 [![Post Edit Title](./img/tutorial_post_title.png)](./img/tutorial_post_title.png)
 
-This component uses the same `useRecordContext` hook as the custom `<UrlField>` commponent described earlier.
+This component uses the same `useRecordContext` hook as the custom `<UrlField>` component described earlier.
 
 As users can access the post editing page directly by its url, the `<PostTitle>` component may render *without a record* while the `<Edit>` component is fetching it. That's why you must always check that the `record` returned by `useRecordContext` is defined before using it - as in `PostTitle` above.
 
@@ -845,9 +845,11 @@ For this tutorial, since there is no public authentication API, we can use a fak
 
 The `authProvider` must expose 5 methods, each returning a `Promise`:
 
-```jsx
+```js
 // in src/authProvider.ts
-export const authProvider = {
+import {AuthProvider} from "react-admin";
+
+export const authProvider: AuthProvider = {
   // called when the user attempts to log in
   login: ({ username }) => {
     localStorage.setItem("username", username);
@@ -924,13 +926,13 @@ React-admin calls the Data Provider with one method for each of the actions of t
 The code for a Data Provider for the `my.api.url` API is as follows:
 
 ```js
-import { fetchUtils } from "react-admin";
+import { DataProvider, fetchUtils } from "react-admin";
 import { stringify } from "query-string";
 
 const apiUrl = 'https://my.api.com/';
 const httpClient = fetchUtils.fetchJson;
 
-export const dataProvider= {
+export const dataProvider: DataProvider = {
     getList: (resource, params) => {
         const { page, perPage } = params.pagination;
         const { field, order } = params.sort;
@@ -943,7 +945,7 @@ export const dataProvider= {
 
         return httpClient(url).then(({ headers, json }) => ({
             data: json,
-            total: parseInt(headers.get('content-range').split('/').pop(), 10),
+            total: parseInt((headers.get('content-range') || "0").split('/').pop()!, 10),
         }));
     },
 
@@ -975,7 +977,7 @@ export const dataProvider= {
 
         return httpClient(url).then(({ headers, json }) => ({
             data: json,
-            total: parseInt(headers.get('content-range').split('/').pop(), 10),
+          total: parseInt((headers.get('content-range') || "0").split('/').pop()!, 10),
         }));
     },
 
@@ -1040,4 +1042,4 @@ React-admin was built with customization in mind. You can replace any react-admi
 
 Now that you've completed the tutorial, continue your journey with [the Features chapter](./Features.md), which lists all the features of react-admin.
 
-**Tip**: To help you close the gap between theoritical knowledge and practical experience, take advantage of the react-admin [Demos](./Demos.md). They are great examples of how to use react-admin in a real world application. They also show the best practices for going beyond simple CRUD apps.
+**Tip**: To help you close the gap between theoretical knowledge and practical experience, take advantage of the react-admin [Demos](./Demos.md). They are great examples of how to use react-admin in a real world application. They also show the best practices for going beyond simple CRUD apps.
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -69,7 +69,7 @@ JSONPlaceholder provides endpoints for users, posts, and comments. The admin we'
 Bootstrap the admin app by replacing the `src/App.tsx` by the following code:
 
 ```jsx
-// in src/App.tsx
+// in src/App.jsx
 import { Admin } from "react-admin";
 import jsonServerProvider from "ra-data-json-server";
 
@@ -126,7 +126,7 @@ We'll start by adding a list of users.
 The `<Admin>` component expects one or more `<Resource>` child components. Each resource maps a name to an endpoint in the API. Edit the `App.tsx` file to add a resource named `users`:
 
 ```diff
-// in src/App.tsx
+// in src/App.jsx
 -import { Admin } from "react-admin";
 +import { Admin, Resource, ListGuesser } from "react-admin";
 import jsonServerProvider from "ra-data-json-server";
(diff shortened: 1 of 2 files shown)


docs/Tutorial.md
```diff
@@ -845,9 +845,11 @@ For this tutorial, since there is no public authentication API, we can use a fak
 
 The `authProvider` must expose 5 methods, each returning a `Promise`:
 
-```jsx
+```tsx
```
djhi, line 848: Please revert this one, we want to keep the documentation in JS only for now

docs/Tutorial.md
```diff
@@ -923,14 +925,14 @@ React-admin calls the Data Provider with one method for each of the actions of t
 
 The code for a Data Provider for the `my.api.url` API is as follows:
 
-```js
-import { fetchUtils } from "react-admin";
+```ts
```
djhi, line 928: Same

djhi (changes requested):
Thanks!

docs/Tutorial.md
erikvanderwerf: The top comment in all the code blocks indicates typescript.
```
// in src/authProvider.ts
```
Which portions should I revert? Only this line about the code block type, or are you also including the rest of my changes here?

My concern is only that if this is just JS and not TS, then my change is wholly unnecessary and we should change all the TypeScript files back to JavaScript. Right now the example tutorial is written with `*.ts` files.

examples/tutorial/src/authProvider.ts
```diff
@@ -1,4 +1,6 @@
-export const authProvider = {
+import { AuthProvider } from 'react-admin';
```
erikvanderwerf, line 1: Like this one ^

docs/Tutorial.md
fzaninotto: The tutorial uses the `react-ts` vite template, because it's the best practice even though it doesn't force the user to write TypeScript code. And we want the tutorial to be gentle and inclusive of non-TS developers, so we use only JS.

So all new files should be `*.ts`  or `*.tsx` to follow the template, but there should ideally be no TypeScript-specific code, or it should be explained.

erikvanderwerf: I changed the header back to `` ```jsx `` and IntelliJ raised a concern about JS. Seems like it thinks JS does not support typing, so unless I should proceed with committing anyway, we can't have it both ways.
image

erikvanderwerf: Without the type hints, pasting the sample into a TypeScript file will not compile. So I'll defer to your direction 😄

erikvanderwerf: But it does work with `` ```js `` instead. Curious and baffling to me.

erikvanderwerf:
^ Sorry! I tried to hit the buttons for both but GitHub is being weird.

docs/Tutorial.md
```diff
@@ -924,13 +926,13 @@ React-admin calls the Data Provider with one method for each of the actions of t
 The code for a Data Provider for the `my.api.url` API is as follows:
 
 ```js
-import { fetchUtils } from "react-admin";
+import { DataProvider, fetchUtils } from "react-admin";
```
djhi, line 929: Same here

djhi (changes requested):

djhi:
I'm sorry I should have say so in your initial issue :(

docs/Tutorial.md
```diff
@@ -845,9 +845,11 @@ For this tutorial, since there is no public authentication API, we can use a fak
 
 The `authProvider` must expose 5 methods, each returning a `Promise`:
 
-```jsx
+```js
 // in src/authProvider.ts
-export const authProvider = {
+import {AuthProvider} from "react-admin";
```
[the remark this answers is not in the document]
erikvanderwerf: I don't feel like I can without some more direction or help. The code samples without the types simply do not compile in TypeScript.

Do you mean for me to just remove the import line?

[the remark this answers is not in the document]
djhi: Yes I'm sorry I should have say so in your issue but we'll keep the examples in JS until we find a way to have both without duplicating all of them. So yes, please remove the types even though that won't compile in TS.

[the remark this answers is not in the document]
fzaninotto: I think we can find a middle ground:

- the code snippets are in JS
- When they don't compile in TS, a note following the snippet explains how to modify it to make it compile.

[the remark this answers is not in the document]
erikvanderwerf: This is a fair compromise. I have made a new commit with the above.

fzaninotto:
I'm sorry if I wan't clear. I meant the code should be pure JS, but the files should be `.ts` and `.tsx`. When the TS compilation fails with pure JS code, we can add a note after the JS snippet to explain how to modify the file for make it work.

erikvanderwerf:
I don't feel comfortable doing that. If the files are JavaScript then own it. What's the value in the TypeScript extension if you are not actually using it? JavaScript code belongs in .js files, with notes on how to make it work in .ts.

erikvanderwerf:
The tutorial directory (https://github.com/marmelab/react-admin/tree/master/examples/tutorial/src) is written with .ts files, so anyone reading the example will notice the difference.

docs/Tutorial.md
```diff
@@ -845,8 +845,10 @@ For this tutorial, since there is no public authentication API, we can use a fak
 
 The `authProvider` must expose 5 methods, each returning a `Promise`:
 
-```jsx
-// in src/authProvider.ts
+```js
+// in src/authProvider.js
+
+// TypeScript must reference the type `: AuthProvider`
```
slax57, line 851:
```suggestion
// TypeScript users must reference the type : `AuthProvider`
```

docs/Tutorial.md
```diff
@@ -924,13 +926,15 @@ React-admin calls the Data Provider with one method for each of the actions of t
...
 ```js
+// in src/dataProvider.js
 import { fetchUtils } from "react-admin";
 import { stringify } from "query-string";
 
 const apiUrl = 'https://my.api.com/';
 const httpClient = fetchUtils.fetchJson;
 
-export const dataProvider= {
+// TypeScript must reference the type `: DataProvider`
```
slax57, line 936:
```suggestion
// TypeScript users must reference the type : `DataProvider`
```

slax57 (changes requested):
Also, as asked by @fzaninotto , you should leave all the file extensions as `*.ts` or `*.tsx`.

Indeed the tutorial is written so that typescript is enabled, and also the file paths will match the files from the sample repository.

If ever a code snippet, written in pure JS but saved in a TS(X) file raises a warning in your IDE, we simply need  to mention it beneath or inside the snippet as comment, just like you did.

erikvanderwerf:
I did not change any file extensions in the repo. If you are referring to the snippets' `// in ...` comments, it would be misleading to someone making their own project to leave them as TS.

Walk through the tutorial yourself as a new user, like I did. It is confusing to be make TypeScript files that do not compile. I would rather understand that the example is written in JS, and any upgrades to TS are on my own. I would understand that more is required from me.

erikvanderwerf:
If the examples are written in TS, why not update the tutorial to reflect that? It would be clearer for everyone.

slax57:
@erikvanderwerf Yes I am referring to the snippets' `// in ...` comments

I know it's a bit confusing, because we are currently between two seats. But I'll try my best to explain.

- We cannot (for now) change the snippets to TS, for technical reasons (they are not yet supported by our code highlighter) and also because we do not want to turn off people who are still not ready to use TS and only know JS
- We name our files `*.ts` and `*.tsx` because, on the other hand, we want to encourage the reader to use TS (since it's a good opportunity because they are starting a new project) and also because they match the file names from the tutorial example in the repo
- So we end up in this _compromise_ of a solution that consists in having JS code inside `.ts` files. But as you saw, only a handful of files require adjustments to get rid of the TS warnings. Most of the snippets can be copy-pasted with no issues at all. That's why we consider it to be a good compromise for now.

Hopefully this will make it clearer for you why we ask for these changes.

erikvanderwerf:
Technologic challenges are a valid reason for not upgrading the Tutorial code blocks 😄 

So if I understand correctly:
1. The code block needs to indicate JS for syntax highlighting purposes (`` ```js ``).
2. But the comment `// in *.ts` will indicate TS to encourage people to use TS.
3. But the code as written will be JS, meaning that TS might not compile if a new user copies the code.
4. But with comments in the JS explaining the changes needed so that it would compile in TS.

It certainly is confusing (and I don't exactly agree, but I can live with it). If the above understanding is correct, I'll make the appropriate changes.

slax57:
@erikvanderwerf Yes that's correct!

slax57 (approved):
Thanks!

PR #8611, opened 2023-01-30
add Postgres data provider

mingfang:


docs/DataProviderList.md
```diff
@@ -46,6 +46,7 @@ If you can't find a Data Provider for your backend below, no worries! [Writing a
 * **[OpenCRUD](https://www.opencrud.org/)**: [weakky/ra-data-opencrud](https://github.com/Weakky/ra-data-opencrud)
 * **[Parse](https://parseplatform.org/)**: [almahdi/ra-data-parse](https://github.com/almahdi/ra-data-parse)
 * **[PostGraphile](https://www.graphile.org/postgraphile/)**: [bowlingx/ra-postgraphile](https://github.com/BowlingX/ra-postgraphile)
+* **[Postgres](https://github.com/mingfang/ra-postgres-dataprovider)
```
fzaninotto, line 49: The syntax isn't correct: the list shows first the name of the backend in bold, then the name of the adapter. Could you please fix it?

fzaninotto (changes requested):

fzaninotto:
There is something I don't understand: node-pg is a server-side module. React-admin is supposed to work with an API. Does that restrict the usage of your data provider to server-side rendered apps?

mingfang:
> There is something I don't understand: node-pg is a server-side module. React-admin is supposed to work with an API. Does that restrict the usage of your data provider to server-side rendered apps?

node-pg is server side only.

To work on the client side, then the combination of 
Neon WSProxy https://github.com/neondatabase/wsproxy and
Neon Serverless(pg drop-in replacement) https://github.com/neondatabase/serverless

The flow would be React-Admin(client side) -websocket-> wsproxy(server side) -> Postgres

fzaninotto:
Understood. You probably need to mention that in your Readme though ;)

mingfang:
> 



> Understood. You probably need to mention that in your Readme though ;)

I added this information to my Readme as suggested.

mingfang:
> There is something I don't understand. Does that mean that the credentials to the Posgres database have to be prensent in the dataProvider, i.e. in the JS code sent to the client? If that's the case, this is highly vulnerable to all sorts of attacks and I can't see any real-life scenario.

The sample code hardcoded the credentials to keep the code simple as a demonstration only.
The Postgres credentials should be secured the same way as any other RA providers.

mingfang:
> 



> Then you should update your documentation to show a secure-by-default example.

I moved the URL to an .env.sample file

fzaninotto:
> I moved the URL to an .env.sample file

This doesn't change the fact that the connection string will be bundled in the client JS. 

All in all, I don't see any secure usage for this package, and I prefer not in include it in the list of available data providers.

PR #8626, opened 2023-02-07
allow -0 at NumberInput filling

pamuche:
### Problem

If you want to enter a negative number <1 (e.g. -0.14) `<NumberInput/>`-`onChange` automagically corrects `-0` to `0` while typing. You can add the `-` after a not-0-number (e.g. 0.14) is filled in <NumberInput/>.

### Solution

Adding special `-0` input treatment in `convertStringToNumber`.

### DIFF
--- a/packages/ra-ui-materialui/src/input/NumberInput.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.tsx
@@ -197,7 +197,7 @@ const convertStringToNumber = value => {
         return value;
     }
     
-    const float = value === '-0' ? '-0' : parseFloat(value);
+    const float = parseFloat(value);
 
     return isNaN(float) ? 0 : float;
 };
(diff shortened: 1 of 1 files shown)


fzaninotto:
I can't reproduce the bug you're describing with the latest version (4.7.4):

https://user-images.githubusercontent.com/99944/217244710-6c7225c9-500e-4573-ae8f-73d733247a43.mov

Can you include a way to reproduce the bug, or even better, a unit test that fails in master and passes with your patch?

pamuche:
Yes, I was confused, that it was working in storybooks.

Here is a screencast from the poster galore demo, where it won't work:

Peek 2023-02-07 14-30

I will check again. and look into php unit testing react-admin.

fzaninotto:
The Posters Galore demo hadn't been updated in a few weeks, and the fix is recent (https://github.com/marmelab/react-admin/pull/8610). That explains why you had the bug there. 

I've updated the Posters Galore demo, and I don't see the bug anymore.

pamuche:
:100: :+1:  Yes this will fix my problem. Hrnks, sorry for bothering you. But also thanks for the kind help. :point_up: Will update before my next PR. :point_up: :grin:

PR #8627, opened 2023-02-07
fix: highlight error in red when form is submitted

sumithmano:
If the checkbox is required by the form but is touched then upon submit validation error is displayed, but not highlighted in red color.


fzaninotto:
I'm sorry, I don't understand the problem. Would you mind elaborating a bit, or even better, adding a screencast of what happens and what you'd like instead?

sumithmano:
If CheckboxGroupInput is required in a form and submit form button is triggered without touching the component.
then the error helperText is not highlighted in error(red) color.
Screenshot 2023-02-14 at 11 13 19 AM

fzaninotto:
Thanks for the description, now I understand better. 

Your PR has linter warnings and misses a storybook to be able to test it. To push this fix forward, I opened a new PR #8652 with these changes. That PR supersedes yours.

PR #8644, opened 2023-02-13
Fix `<RichTextField>` XSS vulnerability

fzaninotto:
## Problem

`<RichTextField>` leverages the `dangerouslySetInnerHTML` attribute, expecting the value to be already sanitized server-side. 

If it is not, a malicious user can execute an XSS attack by injecting malicious data. 

Proof-of-concept:

```jsx
    <RecordContextProvider
        value={{
            id: 1,
            body: `
<p>
<strong>War and Peace</strong> is a novel by the Russian author
<a href="https://en.wikipedia.org/wiki/Leo_Tolstoy" onclick="document.getElementById('stolendata').value='credentials';">Leo Tolstoy</a>,
published serially, then in its entirety in 1869.
</p>
<p onmouseover="document.getElementById('stolendata').value='credentials';">
It is regarded as one of Tolstoy's finest literary achievements and remains a classic of world literature.
</p>
<img src="x" onerror="document.getElementById('stolendata').value='credentials';" />
`,
        }}
    >
        <RichTextField source="body" />
        <hr />
        <div>
            <h4>Stolen data:</h4>
            <input id="stolendata" defaultValue="none" />
        </div>
    </RecordContextProvider>
```

## Solution

Sanitize the value by default using [DomPurify](https://github.com/cure53/DOMPurify). This adds 8.6kB gzipped to the final bundle for people using `<RichTextField>`, but it's inevitable.

### DIFF
--- a/packages/ra-ui-materialui/src/field/RichTextField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/RichTextField.spec.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render, waitFor, screen, fireEvent } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 import { RecordContextProvider } from 'ra-core';
 
 import { RichTextField, removeTags } from './RichTextField';
--- a/packages/ra-ui-materialui/src/field/RichTextField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/RichTextField.spec.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render } from '@testing-library/react';
+import { render, waitFor, screen, fireEvent } from '@testing-library/react';
 import { RecordContextProvider } from 'ra-core';
 
 import { RichTextField, removeTags } from './RichTextField';
+import { Secure } from './RichTextField.stories';
 
 describe('stripTags', () => {
     it('should strip HTML tags from input', () => {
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/field/RichTextField.spec.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render } from '@testing-library/react';
+import { render, waitFor, screen, fireEvent } from '@testing-library/react';
```
slax57, line 3: Linter warning here
> 'waitFor' is defined but never used

packages/ra-ui-materialui/src/field/RichTextField.tsx
```diff
@@ -32,7 +48,11 @@ export const RichTextField: FC<RichTextFieldProps> = memo<RichTextFieldProps>(
                 ) : stripTags ? (
                     removeTags(value)
                 ) : (
-                    <span dangerouslySetInnerHTML={{ __html: value }} />
+                    <span
+                        dangerouslySetInnerHTML={{
+                            __html: purify.sanitize(value),
```
slax57, line 53: That's neat :)

But how about `RichTextInput`? Shouldn't we also sanitize the value we fetch from the server before feeding it to `RichTextInput`?

slax57 (changes requested):

packages/ra-ui-materialui/src/field/RichTextField.tsx
fzaninotto: `<RichTextInput>` doesn't use `dangerouslySetInnerHTML`.

slax57: And [tiptap should no longer be vulnerable](https://security.snyk.io/vuln/SNYK-JS-TIPTAP-575143). So that's fine by me!

slax57 (approved):

PR #8646, opened 2023-02-13
Fix useList filter option not working with nested objects

### LINKED ISSUE #8632: useList filter option not working with nested objects
**What you were expecting:**
When I use the `useList` hook with a nested filter option, e.g. `{ "author.name": "Fred" }`
1. the `data` should be filtered correctly

**What happened instead:**
1. the `data` is returning an empty array

**Steps to reproduce:**
1. Open code sandbox
1. Type "fred" in Author name input and Submit
1. The list will be empty

**Related code:**

Codesandbox: https://codesandbox.io/p/sandbox/peaceful-colden-p8qw7n

```tsx
const posts = [{ id: 1, title: "post 1", author: { name: "fred" } }]
const list = useList({
    data: posts,
    filter: { author: { name: "fred" } },
});
// list.data should be equal to posts
```

**Other information:**
The hook `useList` uses `Object.entries` to get the `filterName` and it doesn't work as expected with nested objects.
A possible solution would be to implement a method to flatten the keys into a string path. 
e.g `{ author: { name: "fred" } }`  ->  `{ "author.name": "fred" }`

https://codesandbox.io/p/sandbox/peaceful-colden-p8qw7n?file=%2Fsrc%2Fuse-list.ts

**Environment**

* React-admin version: 4.7.4
* React version: 18.0.15
* Browser: Chrome

WiXSL:
Fixes #8632

### DIFF
--- a/packages/ra-core/src/controller/list/useList.spec.tsx
+++ b/packages/ra-core/src/controller/list/useList.spec.tsx
@@ -263,4 +263,32 @@ describe('<useList />', () => {
             );
         });
     });
+
+    it('should filter data based on a custom filter with nested objects', () => {
+        const callback = jest.fn();
+        const data = [
+            { id: 1, title: { name: 'hello' } },
+            { id: 2, title: { name: 'world' } },
+        ];
+
+        render(
+            <UseList
+                data={data}
+                filter={{ title: { name: 'world' } }}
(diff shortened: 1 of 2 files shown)


WiXSL:
👍

PR #8666, opened 2023-02-20
Fix `Notification` raises a MUI warning when `message` is a string

### LINKED ISSUE #8664: Failed prop type: Invalid prop `children` of type `boolean` supplied to `ForwardRef(Snackbar)`
**What you were expecting:**
No errors in the console.

**What happened instead:**
The following error appears, when I save one of my Edit views (which triggers a default "Element updated" notification):

<pre>
Warning: Failed prop type: Invalid prop `children` of type `boolean` supplied to `ForwardRef(Snackbar)`, expected a single ReactElement.
    at Snackbar (static/js/bundle.js:130817:82)
    at static/js/bundle.js:54696:66
    at Notification (static/js/bundle.js:235317:25)
    at div
    at static/js/bundle.js:54696:66
    at ScopedCssBaseline (static/js/bundle.js:129248:82)
    at AdminUI (static/js/bundle.js:222060:25)
    at App (static/js/bundle.js:128:105)
    at div
    at ThemeRegion (static/js/bundle.js:174155:5)
    at InnerThemeProvider (static/js/bundle.js:141150:70)
    at ThemeProvider (static/js/bundle.js:140058:5)
    at ThemeProvider (static/js/bundle.js:141169:5)
    at ThemeProvider (static/js/bundle.js:235879:21)
    at ResourceDefinitionContextProvider (static/js/bundle.js:211977:15)
    at NotificationContextProvider (static/js/bundle.js:219498:21)
    at I18nContextProvider (static/js/bundle.js:218616:15)
    at Router (static/js/bundle.js:274885:15)
    at HistoryRouter (static/js/bundle.js:220057:21)
    at InternalRouter (static/js/bundle.js:219971:21)
    at BasenameContextProvider (static/js/bundle.js:220026:21)
    at AdminRouter (static/js/bundle.js:219953:20)
    at QueryClientProvider (static/js/bundle.js:272148:21)
    at PreferencesEditorContextProvider (static/js/bundle.js:219706:21)
    at StoreContextProvider (static/js/bundle.js:220424:18)
    at CoreAdminContext (static/js/bundle.js:211505:28)
    at AdminContext (static/js/bundle.js:222002:18)
    at AbilityProvider (static/js/bundle.js:18502:5)
    at ApolloProvider (static/js/bundle.js:349720:19)
    at ThemeProvider (static/js/bundle.js:54726:64)
    at MantineProvider (static/js/bundle.js:100923:5)
    at AppContainer (static/js/bundle.js:430:62)
</pre>

**Steps to reproduce:**
Since it's a non public repo I cannot share the code, but I am not doing anything crazy here. 
I assume it has to do with the newer versions of @mui, where Snackbar does not accept boolean as children anymore.
But here it looks like Notification will pass `false` to the Snackbar sometimes:
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/Notification.tsx#L120
image

**Environment**

* React-admin version:  4.8.0
* Material UI version: 5.11.9
* React version: 18.2.0
* Browser: chrome

slax57:
Fix #8664

### Notes

Didn't add a test cause I'm not sure it would make sense to have a test ensuring MUI does not raise any warning (as it's pretty specific). But I can add one if you guys think otherwise.

The fix can be tested with the `Basic` story of the `Notification` storybook.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Notification.tsx
+++ b/packages/ra-ui-materialui/src/layout/Notification.tsx
@@ -117,7 +117,7 @@ export const Notification = (props: NotificationProps) => {
             {...rest}
             {...options}
         >
-            {message && typeof message !== 'string' && message}
+            {message && typeof message !== 'string' ? message : null}
         </StyledSnackbar>
     );
 };


djhi (approved):

magnattic:
Thank you! This helps us a lot. Minor nitpick, the title of the PR should probably be

> Fix Notification raises a MUI warning when message is NOT a string

right?

slax57:
@magnattic No, it's the other way around. When `message` is not a string, it is a React Element, and in this case the `Snackbar` throws no error because it is supported.
When `message` is a string, then instead of passing no child to `Snackbar` we passed `false`, which caused the warning.
So the fix actually applies when `message` IS a string.

PR #8667, opened 2023-02-20
Fix onError type

djhi:
Fix the `onError` type name to be Pascal cased like the others

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -247,7 +247,11 @@ export type OnSuccess = (
     variables?: any,
     context?: any
 ) => void;
+
+// @deprecated - use OnError instead
 export type onError = (error?: any, variables?: any, context?: any) => void;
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/types.ts
```diff
@@ -247,7 +247,11 @@ export type OnSuccess = (
     variables?: any,
     context?: any
 ) => void;
+
+// @deprecated - use OnError instead
 export type onError = (error?: any, variables?: any, context?: any) => void;
+export type OnError = onError;
```
fzaninotto, line 253: I'd do it the other way around

```jsx
export type OnError = (error?: any, variables?: any, context?: any) => void;;
// @deprecated - use OnError instead
export type onError = OnError
```

fzaninotto:
Can you please add a description?

slax57 (approved):

fzaninotto (approved):

PR #8676, opened 2023-02-22
Allow to customize how `<FilterListItem>` applies filters

djhi:
## Problem
There is now easy way to implement cumulative filters (filtering by multiple tags for instance) without reimplementing  `<FilterListItem>`.

### Solution
Introduce two new props:
- isSelected: accept the item value and the current filters and return a boolean
- toggleFilter: accept the item value and the current filters and return new filters

### TODO
- [x] Implementation
- [x] Storybook
- [x] Tests
- [x] Documentation


fzaninotto (approved):

docs/FilterList.md
```diff
@@ -185,3 +185,65 @@ const CustomerList = props => (
 {% endraw %}
 
 **Tip**: The `<FilterList>` Sidebar is not a good UI for small screens. You can choose to hide it on small screens (as in the previous example). A good tradeoff is to use `<FilterList>` on large screens, and the Filter Button/Form combo on Mobile.
+
+## Customize How Filters Are Applied
```
fzaninotto, line 189: I'd call it "cumulative filters"

docs/FilterList.md
```diff
@@ -185,3 +185,65 @@ const CustomerList = props => (
...
+
+Sometimes, you may want to customize how filters are applied. For instance, by allowing users to select multiple items such as selecting multiple categories. The `<FilterListItem>` component accepts two props for this purpose:
+
```
fzaninotto, line 192: I'd add a screencast of your story to illustrate this

fzaninotto (changes requested):

docs/FilterList.md
djhi: I disagree, that's only one of the possible use cases

docs/FilterList.md
djhi: Done

docs/FilterList.md
```diff
@@ -185,3 +185,67 @@ const CustomerList = props => (
...
+- `isSelected`: accept a function that receive the item value and the currently applied filters. It must return a boolean.
```
slax57, line 193:
```suggestion
- `isSelected`: accepts a function that receives the item value and the currently applied filters. It must return a boolean.
```

docs/FilterList.md
```diff
@@ -185,3 +185,67 @@ const CustomerList = props => (
...
+- `toggleFilter`: accept a function that receive the item value and the currently applied filters. It is called when user toggle a filter and must return the new filters to apply.
```
slax57, line 194:
```suggestion
- `toggleFilter`: accepts a function that receives the item value and the currently applied filters. It is called when user toggles a filter and must return the new filters to apply.
```

packages/ra-ui-materialui/src/list/filter/FilterListItem.spec.tsx
```diff
@@ -140,6 +147,44 @@ describe('<FilterListItem/>', () => {
...
+
+    it('should allow to customize isSelected and toggleFilter', () => {
+        const { container } = render(<Cumulative />);
+
+        expect(
+            Array.from(
+                container.querySelectorAll<HTMLElement>(
+                    '[data-selected="true"]'
+                )
+            ).map(item => item.textContent)
```
slax57, line 163: factorize this method would improve the test readability

packages/ra-ui-materialui/src/list/filter/FilterListItem.spec.tsx
```diff
@@ -140,6 +147,44 @@ describe('<FilterListItem/>', () => {
...
+        ).toEqual(['News', 'Tutorials']);
+
+        screen.getByText(/"category":.*\[.*"tutorials",.*"news".*\]/);
```
slax57, line 166: This regex seems a bit too permissive IMO. For instance it would still match if there are too many categories selected.
Why not simply copy paste the exact expected string?

packages/ra-ui-materialui/src/list/filter/FilterListItem.spec.tsx
```diff
@@ -140,6 +147,44 @@ describe('<FilterListItem/>', () => {
...
+        screen.getByText('Tutorials').click();
+
+        expect(
+            Array.from(
+                container.querySelectorAll<HTMLElement>(
+                    '[data-selected="true"]'
+                )
+            ).map(item => item.textContent)
+        ).toEqual([]);
+        screen.getByText(/{}/);
```
slax57, line 188: Your test only checks for removal cases, you should also add one addition case IMO

slax57 (changes requested):

packages/ra-ui-materialui/src/list/filter/FilterListItem.spec.tsx
djhi: Unfortunately I didn't manage to match the full string yet. I'll try

djhi: Done

packages/ra-ui-materialui/src/list/filter/FilterListItem.spec.tsx
djhi: Added

slax57 (approved):

docs/FilterList.md
slax57: @djhi Can you still mention the words "cumulative filters" in the doc, before the example, so that an Algolia search would still lead to this page?

djhi: Cumulative, additive. I'm not sure what's the best one here

slax57 (approved):

davidhenley:
Thank you so much Gildas! This will help a ton.

davidhenley:
Hey @djhi is there a reason that this is in the 4.9 milestone and not being merged into one of the more recent releases?

fzaninotto:
Yes: we follow semver. 

- bug fix every week (bugfix or x.x.X)
- new backward compatible features every month (minor or x.X.x)
- breaking changes very seldom (major or X.x.x)

This change was merged after the latest minor (4.8.0), so it's scheduled for the next minor (4.9.0).

davidhenley:
Thanks @fzaninotto!

PR #8677, opened 2023-02-22
Fix `<Datagrid>` `rowClick` type and documentation

djhi:
The TS definition didn't accept `false` as a vaild value.
The documentation did not mention you can pass `false`.

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -655,9 +655,10 @@ export const PostList = () => (
 * "show" to redirect to the show vue
 * "expand" to open the `expand` panel
 * "toggleSelection" to trigger the `onToggleItem` function
-* a function `(id, resource, record) => path` to redirect to a custom path
+* `false` to do nothing
+* a function `(id, resource, record) => path` that may return any of the above values or a custom path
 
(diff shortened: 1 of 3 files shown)


docs/Datagrid.md
```diff
@@ -655,9 +655,10 @@ export const PostList = () => (
 * "show" to redirect to the show vue
 * "expand" to open the `expand` panel
 * "toggleSelection" to trigger the `onToggleItem` function
-* a function `(id, resource, record) => path` to redirect to a custom path
+* `false` to do nothing
+* a function `(id, resource, record) => path` that may return any of the above values or a custom path
 
-**Tip**: If you pass a function, it can return `edit`, `show`; or a router path. This allows to redirect to either `edit` or `show` after checking a condition on the record. For example:
+**Tip**: If you pass a function, it can return `edit`, `show`; `false` or a router path. This allows to redirect to either `edit` or `show` after checking a condition on the record. For example:
```
slax57, line 661:
```suggestion
**Tip**: If you pass a function, it can return `'edit'`, `'show'`, `false` or a router path. This allows to redirect to either the Edit or Show view after checking a condition on the record. For example:
```

slax57 (changes requested):

slax57 (approved):

PR #8678, opened 2023-02-22
[Doc] Fix `MenuItem` example usage in Upgrade guide

### LINKED ISSUE #8409: UserMenu example does not link to route
<table>
<tr>
<th>
My Code
</th>
<th>
Code Example from <a href="https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/layout/UserMenu.tsx#L24">UserMenu.tsx</a>
</th>
</tr>
<tr>
<td><pre lang="tsx">
const MyProfile = 
React.forwardRef((props, ref) => {
    const { onClose } = useUserMenu();
    return 
    < MenuItem
       ref={ref} // ignored in "Upgrading to v4" guide
       {...props}
       // used instead in "Upgrading to v4" guide:
    // component={Link}
       to="/profile"
       onClick={onClose}
    >
        < ListItemIcon>
            < SettingsIcon />
        < /ListItemIcon>
        < ListItemText >
            My Profile
        < /ListItemText>
    < /MenuItem>;
});
<br />
const UserMenu = (props: UserMenuProps) => 
< RaUserMenu {...props}>
    < MyProfile />
    < Logout />
< /RaUserMenu >;
</pre>
</td>
<td>
<pre lang="tsx">
const ConfigurationMenu = 
React.forwardRef((props, ref) => {
    const { onClose } = useUserMenu();
    return (
        < MenuItem
            ref={ref}
            {...props}
<br />
<br />
            to="/configuration"
            onClick={onClose}
        >
            < ListItemIcon>
                < SettingsIcon />
            < /ListItemIcon>
            < ListItemText>
                Configuration
            < /ListItemText>
        < /MenuItem>);
});
<br />
const MyUserMenu = () => (
    < UserMenu>
        < ConfigurationMenu />
        < Logout />
    < /UserMenu>
);
</pre>
</td>
</tr>
</table>

Above you can see my code compared to the example given by react-admin. (The additional spaces at the beginning of tags are only there so the markdown will render correctly in a table)
As you can see it is more or less identical and it also renders correctly (apart from inconsistent spacing):

grafik

However, when I click on `My Profile`, nothing happens, except that the menu closes.
I'd expect the navigation to happen, since it is never mentioned in the example you would have to implement it manually.

In the [Upgrading to v4 Guide](https://marmelab.com/react-admin/Upgrade.html#appbar-and-usermenu-no-longer-inject-props) there is another example for this, but using the `component` property with `Link` and ignoring the `ref` property. (You can see it as a comment in my code above).
With this variant the navigation will work, but there are now runtime errors because `ref` is ignored.

Another issue I found is that the menu item will always look like it is active. With `MenuItemLink` this would work and only highlight the item when the corresponding `Route` is active.

- **I can't get this to work without errors and both documentation examples fall short.**
- **What is the recommended way to do this?**
- **The documentation should get updated accordingly**

WiXSL:
Fixes #8409

### DIFF
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -2890,7 +2890,7 @@ import SettingsIcon from '@mui/icons-material/Settings';
 +           ref={ref}
             // It's important to pass the props to allow MUI to manage the keyboard navigation
             {...props}
-            component={Link}
++           component={Link}
             to="/configuration"
 -            onClick={props.onClick}
 +            onClick={onClose}
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -2887,6 +2887,7 @@ import SettingsIcon from '@mui/icons-material/Settings';
     return (
 -       <MenuItemLink
 +       <MenuItem
++           ref={ref}
             // It's important to pass the props to allow MUI to manage the keyboard navigation
             {...props}
             component={Link}
--- a/examples/demo/src/layout/AppBar.tsx
+++ b/examples/demo/src/layout/AppBar.tsx
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { AppBar, Logout, UserMenu, useTranslate } from 'react-admin';
+import {
+    AppBar,
+    Logout,
+    UserMenu,
+    useTranslate,
+    useUserMenu,
+} from 'react-admin';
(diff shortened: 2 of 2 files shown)


djhi (approved):

djhi:
Thanks!

PR #8686, opened 2023-02-28
Type downloadCSV function

zhujinxuan:
Because `downloadCSV` function is exported by react-admin, it is good to make it typed rather than `any`.


djhi (approved):
Thanks!

PR #8690, opened 2023-03-02
Add functions to get typed sources on fields and inputs

djhi:
- [x] Add all fields
- [ ] Add all inputs
- [ ] Add shortcuts from typed views (?)


fzaninotto:
Encouraging! This should be PRed against `next`, though.

PR #8696, opened 2023-03-02
Use Vite for the CRM demo

djhi:
No more CRA. Our `yarn.lock` got slimmer


slax57 (changes requested):
Changes look good :slightly_smiling_face: 

However I'm having a weird issue when running the CRM demo locally.
The 'loading state' spinner in the AppBar keeps spinning for a while, and then I get this error in the console:

```
Unknown method
overrideMethod	@	react_devtools_backend.js:2655
onError	@	query.js:356
reject2	@	retryer.js:67
(anonymous)	@	retryer.js:132
```

Seems to be related to the react devtools, which is weird, but I don't see it in the 'prod' demo, and I tried to refresh the page and all, still same result.

Do you see it on your end too?

slax57:
FI it also happens when running `yarn build` and `yarn preview`

djhi:
This problem already exists on the deployed demo: https://marmelab.com/react-admin-demo/

slax57 (approved):

PR #8700, opened 2023-03-03
Avoid double translations

### LINKED ISSUE #8662: Form validation errors are translated twice
**What you were expecting:**
In our i18nProvider we have code to raise an error if a translation is missing

```node
export const i18nProvider = {
  ...polyglotProvider,
  translate: (key: string, options: any) => {
    const translation = polyglotProvider.translate(key, options);

    if (translation === key) throw new Error(`missing translation "${key}"`);

    return translation;
  },
};
```

We are expecting validation translations to work when we have code to ensure translations aren't missing

```node
<TextInput validate={minLength(10), 'error.minLength'} />
```

**What happened instead:**
It fails with `Error: missing translation "Must be 10 characters at least"`

**Steps to reproduce:**
1. Add the above code
2. Enter text in text input which is less than 10 characters

**Other information:**
If adding logging to the above function it does first get called with `error.minLength` then gets called with return value of that e.g. `Must be 10 characters at least`. I suspect somewhere it calling translate again when it is not needed.

I have found one area where this happens
https://github.com/marmelab/react-admin/blob/a408de0db2c6db0281b3d39e27c1f2de3deed44e/packages/ra-ui-materialui/src/button/SkipNavigationButton.tsx#L13

But cannot find where this happen in the validation code.

**Environment**

* React-admin version: 4.8.0
* React version: 17.0.2
* Browser: any

djhi:
Some components were translating labels twice. Besides the validation messages were translated twice as well. Fixing this requires a potential breaking change: when using the `InputHelperText` directly, the error message isn't translated anymore unless it's an object with a `message` property. I believe this is a very rare usage if it even exists.

FIxes #8662

### DIFF
--- a/packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
+++ b/packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
@@ -6,7 +6,6 @@ import {
     useResourceDefinitions,
     useGetResourceLabel,
     useCreatePath,
-    useTranslate,
 } from 'ra-core';
 
 import { MenuItemLink } from './MenuItemLink';
@@ -14,7 +13,6 @@ import { MenuItemLink } from './MenuItemLink';
 export const ResourceMenuItem = ({ name }: { name: string }) => {
     const resources = useResourceDefinitions();
     const getResourceLabel = useGetResourceLabel();
-    const translate = useTranslate();
     const createPath = useCreatePath();
     if (!resources || !resources[name]) return null;
     return (
--- a/packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
+++ b/packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
@@ -6,6 +6,7 @@ import {
     useResourceDefinitions,
     useGetResourceLabel,
     useCreatePath,
+    useTranslate,
 } from 'ra-core';
 
 import { MenuItemLink } from './MenuItemLink';
@@ -13,6 +14,7 @@ import { MenuItemLink } from './MenuItemLink';
 export const ResourceMenuItem = ({ name }: { name: string }) => {
     const resources = useResourceDefinitions();
     const getResourceLabel = useGetResourceLabel();
+    const translate = useTranslate();
     const createPath = useCreatePath();
     if (!resources || !resources[name]) return null;
     return (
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
```diff
@@ -22,7 +24,14 @@ export const ResourceMenuItem = ({ name }: { name: string }) => {
...
             state={{ _scrollToTop: true }}
-            primaryText={getResourceLabel(name, 2)}
+            primaryText={
+                <>
+                    {translate(`resources.${name}.name`, {
+                        smart_count: 2,
+                        _: getResourceLabel(name, 2),
+                    })}
+                </>
+            }
```
slax57, line 34: I don't understand this change. Doesn't `getResourceLabel()` already do the translation?

djhi: To avoid a double translation (See MenuItemLink)

packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
```diff
@@ -22,7 +24,14 @@ export const ResourceMenuItem = ({ name }: { name: string }) => {
...
             })}
             state={{ _scrollToTop: true }}
-            primaryText={getResourceLabel(name, 2)}
+            primaryText={
+                <>
+                    {translate(`resources.${name}.name`, {
+                        smart_count: 2,
+                        _: getResourceLabel(name, 2),
+                    })}
+                </>
```
djhi, line 33:
```suggestion
                <>{getResourceLabel(name, 2)}</>
```

packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
```diff
@@ -6,13 +6,15 @@ import {
...
     useCreatePath,
+    useTranslate,
 } from 'ra-core';
 
 import { MenuItemLink } from './MenuItemLink';
 
 export const ResourceMenuItem = ({ name }: { name: string }) => {
     const resources = useResourceDefinitions();
     const getResourceLabel = useGetResourceLabel();
+    const translate = useTranslate();
```
slax57, line 17: `useTranslate` is still unused

slax57 (changes requested):

packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
djhi, line 17:
```suggestion
```

packages/ra-ui-materialui/src/layout/ResourceMenuItem.tsx
```diff
@@ -6,13 +6,15 @@ import {
     useResourceDefinitions,
     useGetResourceLabel,
     useCreatePath,
+    useTranslate,
```
djhi, line 9:
```suggestion
```

Issue #8498, opened 2022-12-13
SelectArrayInput label is crossed by topBorder of the input


JasonStrazisar:
**What you were expecting:**
I want the SelectArrayInput design to be the same as the SelectInput: 
![SelectInput](https://user-images.githubusercontent.com/59898438/207277758-fb139ad6-d0f8-4d7d-bed8-21f4a1a43af8.png)

**What happened instead:**
The top border of the SelectArrayInput cross the InputLabel: 
![SelectArrayInput](https://user-images.githubusercontent.com/59898438/207277939-25e4f2c0-56e3-4123-b7a1-e66d6e164fd6.png)

**Related code:**
```
  <RaReferenceArrayInput
      {...props}
      required
      source="customers"
      reference="customers"
      filter={{ ignoreCustomerFilter: true }}
      perPage={2000}
    >
      <RaSelectArrayInput optionText="name" fullWidth />
    </RaReferenceArrayInput>
```

Or simply

```
 <RaSelectArrayInput source="source" choice={[...choices]}/>
```

**Other information:**

* React-admin version: 4.6.1
* Last version that did not exhibit the issue (if applicable): 3.19
* React version: 17.0.2
* Browser: Chrome and Firefox

slax57:
Thanks for submitting this.
This issue is currently being addressed by https://github.com/marmelab/react-admin/pull/8465

slax57:
Fixed by https://github.com/marmelab/react-admin/pull/8705

PR #8705, opened 2023-03-04
Fix/set-correct-label-for-select-array-input

sebastianbuechler:
After receiving an answer from the MUI team (https://github.com/mui/material-ui/issues/36164) I open the fix for the original issue in #8465 here. For better testing, I added the two remaining variants of the `SelectArrayInput` into the storybook [here](https://react-admin-storybook-qrjp3wcs9-marmelab.vercel.app/?path=/story/ra-ui-materialui-input-selectarrayinput--basic).


slax57 (approved):
Thanks!

sebastianbuechler:
@slax57 Thank you, really enjoy working with react-admin!

WiXSL:
This is related to #8498

slax57:
You're 100% right @WiXSL , thanks for pointing that out! :slightly_smiling_face:

Issue #8731, opened 2023-03-15
ReferenceManyCount doe snot support sortBy and sortByOrder properties


jonathan-marmelab:
**What you were expecting:**
The `ReferenceManyCount` component should be able to support the `sortBy` and `sortByOrder` properties. This can be useful when working with Postgrest and Many-to-Many relationship without id on join tables.

**What happened instead:**
The `sortBy` and `sortByOrder` properties are ignored by `ReferenceManyCount`.

**Steps to reproduce:**
First, you will need to setup a postgrest server with a Many-to-Many relationship and a join table without an id field. Then, setup a `react-admin` project with [ra-data-postgrest](https://github.com/raphiniert-com/ra-data-postgrest) and use a `ReferenceManyCount` in alist view.

**Related code:**

```js
<ReferenceManyCount
        label="Games"
        reference="board_user"
        target="user_id"
        sortBy="board_id"
        sortByOrder="ASC"
      />
```

**Environment**

* React-admin version: 4.8.3
* @promitheus/ra-data-postgrest: 1.2.2
* React version: 18
* Browser: Chrome

PR #8732, opened 2023-03-15
Fix `Count` and `ReferenceManyCount` should allow to override the default `sort`

slax57:
Fixes https://github.com/marmelab/react-admin/issues/8731

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceManyCount.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceManyCount.spec.tsx
@@ -1,7 +1,14 @@
 import * as React from 'react';
 import { render, screen } from '@testing-library/react';
 
-import { Basic, ErrorState, WithFilter } from './ReferenceManyCount.stories';
+import {
+    Basic,
+    ErrorState,
+    WithFilter,
+    Wrapper,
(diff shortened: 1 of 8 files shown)


slax57:
In the end, I'm not so sure about setting the `sort` prop to `null`, because dataProvider implementations might contain code expecting non-null `sort` props, like:
```js
const { field, order } = params.sort;
```

I think it's safer to keep the default value, and simply add the ability for the developers to override it.

packages/ra-ui-materialui/src/field/ReferenceManyCount.spec.tsx
```diff
@@ -1,7 +1,14 @@
...
 import { render, screen } from '@testing-library/react';
 
-import { Basic, ErrorState, WithFilter } from './ReferenceManyCount.stories';
+import {
+    Basic,
+    ErrorState,
+    WithFilter,
+    Wrapper,
+} from './ReferenceManyCount.stories';
+import { ResourceContextProvider } from 'ra-core';
```
djhi, line 10:
```suggestion
```

djhi (changes requested):

djhi (approved):

PR #8739, opened 2023-03-17
[Doc] Add AuthProvider and DataProvider for SurrealDB.

djedi23:
Hi,

I write a package for a an AuthProvider and a DataProvider for SurrealDB.


djhi (approved):
Thanks!

PR #8745, opened 2023-03-18
Typed admin

djhi:
## Problem

In any medium or large admin, ensuring new, renamed or deleted properties for each resource are used correctly is difficult. Our current typescript experience is not good enough to give developers the ability to refactor with ease.

The problem comes in part from the fact that although many components and hooks accept a generic type for the record they work on, developers have to specify it every time.

## Solution

Ideally, developers should be able to provide the resources types only once and get it propagated correctly across all admin components and hooks.

This PR aims to solve this by introducing a `createResource` function that takes a record type and returns an object containing all the hooks and components from react-admin with type safety where it was missing.

For instance, we can validate the `source` prop, not only by checking its value is one of the record properties but also that the property value is of the correct type (for instance with a `BooleanField` or `BooleanInput` will only accept a `source` for boolean properties of the record.

## Tradeoffs

Using the `createResource` will make it impossible for most bundlers to treeshake the library correctly.
We must make sure to allow passing the `Record` on every components and hooks in the core and ui packages to ensure users for whom this is a concern can still benefit from the enhanced typings by passing the `Record` generic every time.
They could then have a `postResource.ts` file re-exporting all used hooks and components with the correct types for easier use across the app.

For large apps where most components and hooks are used anyway, the bundle size difference probably won't matter much.

## TODO

- [ ] Add all fields
- [ ] Add all inputs
- [ ] Add all views
- [ ] Add all dataProvider hooks
- [ ] Add all hooks


djhi:
> Thanks for your efforts! Looking forward to typed resources heart
> 
> Considering the tree-shaking problem, could the alternative approach be to use a {...register(field-name)} similar to how it is done in react-hook-forms? https://react-hook-form.com/get-started/

Nope, it will likely looks like:

```tsx
<TextField<Post> source="title" />
```

djhi:
The only other way I can think that wouldn't have any impact on the bundle size would be to generate types just like some other tools already do (prisma, supabase, graphql, etc)

MadsEKirchhoff:
@djhi What happened to this? I've been looking into typing inputs generically the same way fields are currently and this looks quite promising? Was it abandoned or refactored?

djhi:
This solution had many issues but I'm working on another (generating strongly typed components from your types, for instance in a Vite plugin). First results are promising but there's no ETA

PR #8751, opened 2023-03-22
add renderTextField prop to autocomplete

hiaselhans:
This adds a new prop to autocomplete making the inputfield customizable:

``` typescript
interface newProp  {
    renderTextField?: (props: TextFieldProps, selectedValue: any) => React.ReactNode
}
``` 

My use-case is that i want to visually highlight a `created` value using an inputadornment.


djhi:
You can already override the `renderInput` prop. We'd rather not add yet another prop to an already complex component. I leave the PR open to let @fzaninotto and @slax57 comment if needed

hiaselhans:
thanks for your answers.

1. I can not find a renderInput prop
2. textfieldprops is missing the dynamic behavior based on the selected value

I think in general, using `renderX` provides a lot more flexibility over `XProps` with not much verbosity added.
If i had to choose I would be in favor of removing TextFieldProps for RenderTextField

Issue #8750, opened 2023-03-22
`ReferenceManyCount` renders a corrupted link


oleg-semyonov:
**What you were expecting:**
When used with `link` prop, `ReferenceManyCount` should render a clickable link that leads to the filtered resource page

**What happened instead:**
The link is rendered as `span` tag with `to` attribute set to `[object Object]`.

**Steps to reproduce:**
It is actually reproducible on the [simple example](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple) straight out of the box. Just click "inspect" on any "Nb comment" field in Posts grid.

**Other information:**
This is easily fixable by removing a single line: https://github.com/marmelab/react-admin/blob/1a1b80adb3fd402dd8fd2af23692eadb4696eaa0/packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx#L82

I've filed a PR with the proposed change: #8752 

I don't think using `span` was intended, probably just flowed over from `Typography` component props below.

**Environment**

* React-admin version: 4.6.2
* React version: 18.2.0
* Browser: Chrome

slax57:
Reproduced, thanks!

slax57:
Sorry @oleg-semyonov  I did not see you had already opened a PR for this issue.
Mine has the advantage of allowing to remove the `// @ts-ignore` for good so I'd recommend keeping this one over yours.
Still grateful for the report and the PR though! :pray:

oleg-semyonov:
Wow, thanks for such a quick response!

I checked your PR, and indeed that's a much more permanent solution. I closed my PR in favor of yours.

slax57:
@oleg-semyonov  plot twist: https://github.com/marmelab/react-admin/pull/8754#discussion_r1144501853 actually gives arguments in favor of your PR :grin: 
Would you like to reopen it? (seems like I can't do it myself)
Sorry for the time loss :sweat_smile:

slax57:
It looks good to me, thanks!

PR #8752, opened 2023-03-22
Fix `ReferenceManyCount` rendering link incorrectly

oleg-semyonov:
Fixes #8750 

Tested on my local environment, and nothing has changed visually, while the link works properly now.

PR #8754, opened 2023-03-22
Fix `ReferenceManyCount` link

slax57:
Fixes https://github.com/marmelab/react-admin/issues/8750

### DIFF
--- a/packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
@@ -6,7 +6,12 @@ import {
     useCreatePath,
     SortPayload,
 } from 'ra-core';
-import { Typography, TypographyProps, CircularProgress } from '@mui/material';
+import {
+    Typography,
+    TypographyProps,
+    CircularProgress,
+    Box,
+} from '@mui/material';
 import ErrorIcon from '@mui/icons-material/Error';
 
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
```diff
@@ -69,22 +74,21 @@ export const ReferenceManyCount = (props: ReferenceManyCountProps) => {
...
-                })}`,
-            }}
-            variant="body2"
-            component="span"
-            onClick={e => e.stopPropagation()}
-            {...rest}
-        >
-            {body}
-        </Link>
+        <Box component="span" {...rest}>
```
fzaninotto, line 77: We don't need a wrapping span in that case, plus it's a breaking change since the rest parameters no longer apply to the Link.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
slax57: Right. In this case, it seems to me that https://github.com/marmelab/react-admin/pull/8752 was the better solution then! I'll reopen it.

slax57:
Closing in favor of https://github.com/marmelab/react-admin/pull/8752

PR #8760, opened 2023-03-23
Add ability to pass `null` to `TrueIcon` or `FalseIcon` in `<BooleanField>`

slax57:


davidhenley:
Shouldn't this be `false` instead of null to align with other React-Admin examples like passing `false` to `bulkActionButtons`

I've just been passing MUI's `Icon` when I wanted to hide this.

PR #8766, opened 2023-03-24
Provide TS validation for props related to Record type on Inputs

djhi:
This PR mainly focuses on inputs with the exception of the `BooleanField` that was done to test the same logic applied and the `FileField` and `ImageField` so that the story for file and image input is complete.

I will address the remaining fields in a dedicated PR.

image
image

### DIFF
--- a/packages/ra-ui-materialui/package.json
+++ b/packages/ra-ui-materialui/package.json
@@ -61,6 +61,7 @@
         "clsx": "^1.1.1",
         "css-mediaquery": "^0.1.2",
         "dompurify": "^2.4.3",
+        "hotscript": "^1.0.10",
         "inflection": "~1.12.0",
         "jsonexport": "^3.2.0",
         "lodash": "~4.17.5",
--- a/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
@@ -25,7 +25,9 @@ type Post = {
     title: string;
 };
 
-export const Typed = (props: Omit<BooleanFieldProps, 'source' | 'sortBy'>) => {
+export const Typed = (
+    props: Omit<BooleanFieldProps, 'source' | 'sortBy' | 'record'>
+) => {
     const [published, setPublished] = React.useState(true);
     return (
         <Stack direction="row">
@@ -35,7 +37,7 @@ export const Typed = (props: Omit<BooleanFieldProps, 'source' | 'sortBy'>) => {
                 onChange={e => setPublished(e.target.checked)}
             />
             <BooleanField<Post>
-                record={{ published }}
+                record={{ published, reported: false, title: '' }}
                 source="published"
                 sortBy="published"
                 {...props}
--- a/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
@@ -26,16 +26,16 @@ type Post = {
 };
 
 export const Typed = (props: Omit<BooleanFieldProps, 'source' | 'sortBy'>) => {
-    const [value, setValue] = React.useState(true);
+    const [published, setPublished] = React.useState(true);
     return (
         <Stack direction="row">
             <input
                 type="checkbox"
-                checked={value}
-                onChange={e => setValue(e.target.checked)}
+                checked={published}
+                onChange={e => setPublished(e.target.checked)}
             />
             <BooleanField<Post>
-                record={{ value }}
+                record={{ published }}
                 source="published"
                 sortBy="published"
                 {...props}
--- a/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
@@ -5,7 +5,7 @@ import { BooleanField, BooleanFieldProps } from './BooleanField';
 
 export default { title: 'ra-ui-materialui/fields/BooleanField' };
 
-export const Basic = (props: BooleanFieldProps) => {
+export const Basic = (props: Omit<BooleanFieldProps, 'source'>) => {
     const [value, setValue] = React.useState(true);
     return (
         <Stack direction="row">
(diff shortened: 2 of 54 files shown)


packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
```diff
@@ -71,6 +71,46 @@ export const Basic = () => (
     </Admin>
 );
 
+type Book = {
+    id: number;
+    title: string;
+    authors: [];
```
slax57, line 77:
```suggestion
    authors: Author[];
```

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
```diff
@@ -67,30 +66,45 @@ import { AutocompleteInput, AutocompleteInputProps } from './AutocompleteInput';
  */
 
 export const AutocompleteArrayInput = <
-    OptionType extends RaRecord = RaRecord,
+    RecordType extends Record<string, unknown> = never,
+    OptionType extends Record<string, unknown> = never,
```
slax57, line 70: Why `unknown` instead of `any` here?

```suggestion
    RecordType extends Record<string, any> = never,
    OptionType extends Record<string, any> = never,
```

Besides, isn't this a **breaking change** for people using `AutocompleteArrayInput` like this today?

```tsx
<AutocompleteArrayInput<MyOptionType>>...</AutocompleteArrayInput>
```

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -101,6 +101,55 @@ export const Basic = () => (
...
+};
+
+const TypedBookEdit = () => {
+    const choices = [
+        { id: 1, name: 'Leo Tolstoy' },
+        { id: 2, name: 'Victor Hugo' },
+        { id: 3, name: 'William Shakespeare' },
+        { id: 4, name: 'Charles Baudelaire' },
+        { id: 5, name: 'Marcel Proust' },
+    ];
```
slax57, line 123:
```suggestion
    const choices: Author[] = [
        { id: 1, name: 'Leo Tolstoy', disabled: false },
        { id: 2, name: 'Victor Hugo', disabled: false },
        { id: 3, name: 'William Shakespeare', disabled: false },
        { id: 4, name: 'Charles Baudelaire', disabled: false },
        { id: 5, name: 'Marcel Proust', disabled: false },
    ];
```

packages/ra-ui-materialui/package.json
```diff
@@ -61,6 +61,7 @@
         "clsx": "^1.1.1",
         "css-mediaquery": "^0.1.2",
         "dompurify": "^2.4.3",
+        "hotscript": "^1.0.10",
```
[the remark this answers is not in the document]
slax57: @djhi indeed, you are probably right!

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -85,17 +89,25 @@ BooleanField.propTypes = {
...
         InjectedFieldProps,
         Omit<TypographyProps, 'textAlign'> {
     valueLabelTrue?: string;
     valueLabelFalse?: string;
     TrueIcon?: SvgIconComponent | null;
     FalseIcon?: SvgIconComponent | null;
     looseValue?: boolean;
+    source?: [RecordType] extends [never]
+        ? string
+        : Call<Objects.AllPaths, RecordType>;
```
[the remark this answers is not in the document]
djhi: Unfortunately no, I tried

packages/ra-ui-materialui/src/input/AutocompleteArrayInput.tsx
djhi: It's unknown because it is actually unknown so it's better semantically and a reminder that you should provide the type

packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
```diff
@@ -19,6 +19,31 @@ export const Basic = (props: BooleanFieldProps) => {
...
+    const [published, setPublished] = React.useState(true);
+    return (
+        <Stack direction="row">
+            <input
+                type="checkbox"
+                checked={published}
+                onChange={e => setPublished(e.target.checked)}
+            />
+            <BooleanField<Post>
+                record={{ published }}
```
slax57, line 38: I see a TS error at this line: `Type '{ published: boolean; }' is missing the following properties from type 'Post': reported, title`

Unfortunately, when I fix it, I get another error which I have not been able to fix: `Type 'unknown' is not assignable to type 'Post | undefined'`

packages/ra-ui-materialui/src/input/NumberInput.stories.tsx
```diff
@@ -25,6 +25,26 @@ export const Basic = () => (
...
+
+export const Typed = () => (
+    <AdminContext>
+        <Create
+            resource="posts"
+            record={{ id: 123, views: 23 }}
+            sx={{ width: 600 }}
+        >
+            <SimpleForm>
+                <NumberInput<Post> source="views" />
```
slax57, line 42: Since the change to `extends CommonInputProps<number>` I would have expected to only be able to use `"views"` as value for `source`, unfortunately all 3 fields are valid according to TS. Is it intentional?

packages/ra-ui-materialui/src/input/BooleanInput.stories.tsx
```diff
@@ -16,6 +16,18 @@ export const Basic = () => (
...
 
+type Post = {
+    published: boolean;
+    reported: boolean;
+    title: string;
+};
+
+export const Typed = () => (
+    <Wrapper>
+        <BooleanInput<Post> source="published" />
```
slax57, line 27: If I remember correctly, the `source` prop only allowed the value `"published"` or `"reported"` with the previous version. But now all 3 values (`"published"`, `"reported"`, `"title"`) are possible. Is this regression intended.

slax57 (changes requested):

packages/ra-ui-materialui/src/input/NumberInput.stories.tsx
djhi: It is. Because of format/parse we decided to not add more constraints, a at least for now

packages/ra-ui-materialui/src/input/BooleanInput.stories.tsx
djhi: Same as above

packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
```diff
@@ -121,7 +121,7 @@ ReferenceArrayField.propTypes = {
 
 export interface ReferenceArrayFieldProps
     extends PublicFieldProps,
-        InjectedFieldProps {
+        InjectedFieldProps<any> {
```
djhi, line 124: Equivalent to what we had before. Will be addressed when working on the fields. Same for the others

slax57 (approved):

djhi:
Now that the way to go is clear, I will open small PRs to ease the review

PR #8681, opened 2023-02-24
Improve AppBar Customization

fzaninotto:
## Problem

Adding new buttons or a search engine to the AppBar is cumbersome

## Solution

- [x] Add a `<TitlePortal>` component to avoid copy/pasting a large `<Typography>` from the doc
- [x] Add a `toolbar` prop to let users override the default toolbar
- [x] Add an `alwaysOn` prop to facilitate always on toolbars (and deprecate `component` prop)
- [x] Fix styling glitches
- [x] Fix `<Title>` doesn't work without routing
- [x] Add Storybook
- [x] Add a dedicated `AppBar` doc chapter
- [x] Add a dedicated `Title` doc chapter
- [x] Fix `<AppBar>` usage examples in the doc and demos
- [x] Include screenshots in the doc
- [x] Final proofreading

## Resulting API

To add buttons to the app bar, you can use the `<AppBar>` [`children` prop](#children).

For instance, to add `<ToggleThemeButton>`:

```jsx
// in src/MyAppBar.js
import { 
    AppBar,
    TitlePortal,
    ToggleThemeButton,
    defaultTheme
} from 'react-admin';

const darkTheme = { palette: { mode: 'dark' } };

export const MyAppBar = () => (
    <AppBar>
        <TitlePortal />
        <ToggleThemeButton lightTheme={defaultTheme} darkTheme={darkTheme} />
    </AppBar>
);
```

PR #8768, opened 2023-03-24
[Docs] fix index page links

carlos-duran:
Little fix from https://github.com/marmelab/react-admin/pull/8681#discussion_r1147860491

image


djhi (approved):
Thanks!

PR #8769, opened 2023-03-26
[Doc] Fix hook naming (use 'beforeSave')

afilp:


afilp:
Please note that I found this code in the docs, this is why I proposed the PR. But I do not see it in the actual code now, so maybe the other doc is wrong?

```
{
        resource: "comments",
        beforeRead: async (data, dataProvider) => { /* ... */ },
        afterCreate: async (result, dataProvider) => { /* ... */ },
    },
```

afilp:
ok, I added a new commit with the other change.

fzaninotto:
Thanks!

PR #8773, opened 2023-03-27
Update FilterList documentation with coherent example

davidhenley:
Making the filter customization more readable.


djhi (approved):
Thanks!

Issue #7589, opened 2022-04-25
Fix right sidebar scrolling on documentation pages


bobah2000:
**What you were expecting:**
Vertical scrolling in right sidebar when there are many list items and the height of the viewport is not enough to place all of them.

**What happened instead:**
There are no scrolling and it is not possible to click on last list items

**Steps to reproduce:**
1. Visit react admin documentation page [https://marmelab.com/react-admin/Tutorial.html](https://marmelab.com/react-admin/Tutorial.html) 
2. Decrease the height of browser or open this page on display with low resolution

**Environment**

* React-admin version: 4.0.1 
* Browser: Google Chrome

![ra1](https://user-images.githubusercontent.com/51819476/165065728-4aa4effc-50aa-489f-9dbf-34d0ddac8be2.png)
![ra2](https://user-images.githubusercontent.com/51819476/165065746-5506221c-d8fc-48f9-b628-5d64c399f3cf.png)
![ra3](https://user-images.githubusercontent.com/51819476/165065777-d83d798d-ffb5-4607-aad1-320000aad260.png)

djhi:
Thanks for reporting this!

fzaninotto:
I've reduced the font size for the table of contents. In my experience, it renders correctly even on small screens (tested with a height of 750px). 

Perhaps we should simply hide the ToC for screens with smaller height. 

Making it low priority though.

PR #8778, opened 2023-03-29
Fix right sidebar scrolling on documentation pages

gitstart:
### Description 
The right sidebar now hides when the height is reduced to a certain level.

### Fixes
Fixes: #7589

### Screenshots
**Large Screen:** 

**Reduced Screen:**

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


gitstart:
> I don't understand why you hide the TOC in this case. It's not a good fix for #7589 IMHO.

As per the comment [here](https://github.com/marmelab/react-admin/issues/7589#issuecomment-1114241660) It was suggested to hide it.
But if you have another solution in mind, kindly let us know and we'll make changes accordingly.

fzaninotto:
That was a bad idea (I know it was mine ;)) It would hide the TOC even when it's small enough to fit on the screen. 

I don't have a better solution in my head.

gitstart:
> That was a bad idea (I know it was mine ;)) It would hide the TOC even when it's small enough to fit on the screen.
> 
> I don't have a better solution in my head.

What if we make the TOC scrollable?

fzaninotto:
> What if we make the TOC scrollable?

Yes! If you can update your PR to do that, we'll merge it.

gitstart:
> > What if we make the TOC scrollable?
> 
> Yes! If you can update your PR to do that, we'll merge it.

The ToC is now scrollable @fzaninotto

slax57:
Superseded by https://github.com/marmelab/react-admin/pull/8898

PR #8779, opened 2023-03-29
Add ability to pass false to TrueIcon or FalseIcon in <BooleanField>

davidhenley:
https://github.com/marmelab/react-admin/pull/8760#issuecomment-1488515581

### DIFF
--- a/packages/ra-ui-materialui/src/field/BooleanField.tsx
+++ b/packages/ra-ui-materialui/src/field/BooleanField.tsx
@@ -93,8 +93,8 @@ export interface BooleanFieldProps
         Omit<TypographyProps, 'textAlign'> {
     valueLabelTrue?: string;
     valueLabelFalse?: string;
-    TrueIcon?: SvgIconComponent | boolean;
-    FalseIcon?: SvgIconComponent | boolean;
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -93,8 +93,8 @@ export interface BooleanFieldProps
         Omit<TypographyProps, 'textAlign'> {
     valueLabelTrue?: string;
     valueLabelFalse?: string;
-    TrueIcon?: SvgIconComponent | null;
-    FalseIcon?: SvgIconComponent | null;
+    TrueIcon?: SvgIconComponent | boolean;
+    FalseIcon?: SvgIconComponent | boolean;
```
slax57, line 97: I think it would be better to only support `SvgIconComponent | false`, because it wouldn't make much sense to set one of these props to `true`, would it?

fzaninotto:
Hi, could you update your PR following @slax57's review? Also, as it is a new feature, please target the `next` branch.

davidhenley:
@fzaninotto Sorry, I missed those.

PR #8780, opened 2023-03-29
[RFC] Stabilize the useInput return properties

Nickman87:
I was writing some more complex custom components inside of react-admin that use `useInput` to update fields.
Somewhere in my code I use an effect, based on the results of this hook, to trigger updates like so:

```typescript
const [someState] = useState("someValue");
const testField = useInput({source: "test"});

useEffect(() => {
  if (testField.field.value !== someState) {
    testField.field.onChange(someState);
  }
}, [someState, testField.field.value, testField.field.onChange]);
```
(I know this might look like stupid code, but this is a very trimmed down example of what I want to achieve in my component:) )

This resulted in an infinite loop because the `onChange` that is returned from `useInput` is not stable, it is always a new function, even though nothing significant changes.

I made a quick fix for this in this PR, will add tests or more fixes if this is deemed to be a good fix or not. Not sure if there are further implications or reasonings why this is done the way it is now, so looking for feedback.

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -88,28 +88,39 @@ export const useInput = (props: InputProps): UseInputValue => {
     // This hook ensures that the input default value is applied when a new record is loaded but has
     // no value for the input.
     useApplyInputDefaultValues(props);
-
-    const field = {
-        ...controllerField,
-        value: format ? format(controllerField.value) : controllerField.value,
-        onBlur: (...event: any[]) => {
+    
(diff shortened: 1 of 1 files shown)


slax57:
I believe that's because we never had a use-case where this would be a problem, but currently I see no good reason not to memoize the `onChange` and `onBlur` callbacks.
I'm just not a huge fan of the name `stableOnChange`. You should just call them `onChange` and `onBlur` IMO.
But that's definitely a PR I would approve if it doesn't break anything!
Thanks for submitting it!

PR #8781, opened 2023-03-29
Infinite list

fzaninotto:
`<InfiniteList>` is a drop-in replacement for `<List>` that leverages `useInfiniteGetList()` to allow infinite pagination with `<SimpleList>` and `<Datagrid>`

- [x] Add `useInfiniteListController` hook
- [x] Add `<InfiniteListBase>` component
- [x] Add `<InfiniteList>` component
- [x] Add `<InfinitePagination>` component
- [x] Fix data provider mutation hooks don't update the infinite query cache
- [x] Add story to demonstrate it
- [x] Use it in simple example
- [x] Find a way to display the total number of results
- [x] Add tests (must be e2e because jsdom doesn't have an IntersectionObserver)
- [x] Add documentation

## Sample usage

```jsx
const BookList = () => (
    <InfiniteList>
        <SimpleList primaryText={record => record.title} />
    </InfiniteList>
);
```

https://user-images.githubusercontent.com/99944/228621309-ea13b096-c5a2-4e39-99fd-582678c3eb91.mov


packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -80,7 +79,7 @@ export const useInfiniteGetList = <RecordType extends RaRecord = any>(
         Error,
         GetInfiniteListResult<RecordType>
     >(
-        [resource, 'getList', { pagination, sort, filter, meta }],
+        [resource, 'getInfiniteList', { pagination, sort, filter, meta }],
```
fzaninotto, line 82: This is necessary because otherwise the normal and infinite list queries share the same cache, which leads to errors because they don't have the same structure. 

This means that all the dataProvider hooks that optimistically update the `getList` query cache must also do it for the `getInfiniteList`.

packages/ra-ui-materialui/src/layout/useSidebarState.ts
```diff
@@ -13,7 +16,12 @@ import { useStore } from 'ra-core';
  *     );
  * };
  */
-export const useSidebarState = (): useSidebarStateResult =>
-    useStore<boolean>('sidebar.open', true);
+export const useSidebarState = (): useSidebarStateResult => {
+    const isXSmall = useMediaQuery<Theme>(
```
fzaninotto, line 20: The sidebar was open by default on Mobile. It was not user-friendly, but most importantly it prevented e2e testing on mobile, as every test had to start by closing the menu.

djhi (commented):
Awesome!

slax57 (changes requested):
:clap:

docs/InfiniteList.md
```diff
@@ -0,0 +1,123 @@
...
+                        elevation={2}
+                        sx={{ px: 2, py: 1, mb: 1, display: 'inline-block' }}
+                    >
+                        <Typography variant="body2">{total} results</Typography>
+                    </Card>
+                </Box>
+            )}
+        </>
+    );
+};
```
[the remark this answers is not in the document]
fzaninotto: Because I don't think this is a general solution, it's just an example. People will want to put the number of results on top of the list, in a datagrid footer, in the action bar, etc... There is no standard for that UI.

packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,338 @@
...
+                        primaryText="%{title}"
+                        secondaryText="%{author}"
+                    />
+                </InfiniteList>
+            )}
+        />
+    </Admin>
+);
+
+export const PaginationClassic = () => (
```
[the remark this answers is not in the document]
fzaninotto: I wrote this story to show that the default pagination callbacks still work even for the infinite list. I think we should keep it. 

For the "Load more" pagination, it's a good idea. I created a story and a doc entry, but just like the component that displays the count, I'm not sure it's a standard solution, so I prefer not to publish a component for now.

slax57 (approved):

slax57:
Not sure if this should qualify as a bug or not, but when your screen is already tall enough to display more than 2 pages of results, then the 3rd page is never fetched (only the second page is fetched).

See attached screen capture.

[Peek 04-04-2023 16-13.webm](https://user-images.githubusercontent.com/14542336/229821334-dd5db466-5d84-463a-b9f2-89d8107e1010.webm)

fzaninotto:
> Not sure if this should qualify as a bug or not, but when your screen is already tall enough to display more than 2 pages of results, then the 3rd page is never fetched (only the second page is fetched).
> 
> See attached screen capture.

That's absolutely a bug. Good catch!

fzaninotto:
Should be OK now!

packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,398 @@
...
+        return new Promise(resolve =>
+            setTimeout(
+                () => resolve(baseDataProvider[name](resource, params)),
+                500
+            )
+        );
+    },
+});
+
+const Admin = ({ children, dataProvider, layout }: any) => (
```
djhi, line 103: Can you wrap provide a new queryClient, memoryHistory and store to each story? Otherwise, visiting filtered stories make the others (pagination related for instance) weird as the filters are already stored and applied

djhi (changes requested):

packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
fzaninotto: Done

djhi: Nope :)
> queryClient, memoryHistory and store

fzaninotto: Not necessary in my tests: the stories don't share a state anymore

djhi: Checked and confirmed

djhi (approved):

djhi:
It needs a rebase though

fzaninotto:
rebased

Issue #8785, opened 2023-03-31
ArrayInput element parse might not work as intended


aximov:
**What you were expecting:**
I expected that the `parse` functionality of the element inside the `SimpleFormIterator` and `ArrayInput` would handle empty values, converting them to `null` as intended according to the [December 2022 update for better support for empty values](https://marmelab.com/blog/2022/12/08/react-admin-december-2022-update.html#better-support-for-empty-values-in-inputs).

**What happened instead:**
It seems that the `parse` functionality is not working correctly, resulting in empty values being sent as `""` instead of `null`.

**Steps to reproduce:**
1. Visit the [blog sample](https://iuxnw.sse.codesandbox.io/#/posts/create).
2. Add a new backlink without filling in any information.
3. Click "Save and Edit".
4. Observe the console, noting that the saved values for `date` and `url` are `""` rather than `null`. 

The same issue occurs in my production code, where empty strings are sent to API.

**Related code:**

You can find the related code in the [React-Admin repository](https://github.com/marmelab/react-admin/blob/d5ada6858b71cfc8b14ad4f2184a6d0a82f5ec26/examples/simple/src/posts/PostCreate.tsx#L146). 

**Other information:**
Please investigate this issue and confirm whether the current behavior of the `parse` functionality inside `ArrayInput` is by design or if it needs to be fixed to handle empty values correctly.

**Environment**

* React-admin version: 4.9.0 and 4.4.2
* Last version that did not exhibit the issue (if applicable): 4.0.1
* React version: 17.0.2
* Browser: Chrome 111.0.5563.110 (Official Build) (arm64)
* Stack trace (in case of a JS error): N/A

Thanks to @uzimith for their contribution in discovering this issue.

slax57:
Fixed by https://github.com/marmelab/react-admin/pull/8792

PR #8787, opened 2023-03-31
[SimpleFormIterator] default to null instead of empty string

kriskw1999:
The `SimpleFormIterator` actually is defaulting every input inside a `FormDataConsumer` to an empty string. So if I have some arrays, objects, or numbers inside, they default to the empty strings when I add new elements to the array.

I propose to make the default equal to null instead. I know this is a kind of breaking change but I think is more consistent, in case we don't have enough information to determine the right return type null should be better.

This should also solve this issue: https://github.com/marmelab/react-admin/issues/8785 because it will default to null when no default value will be provided.

Let me know what you think about it!


slax57:
Thanks for submitting this PR.
See my comment [here](https://github.com/marmelab/react-admin/issues/8785#issuecomment-1491963605) if you don't have already.
We'll be in touch.

kriskw1999:
I have done the rebase on the next branch, but comparing it with the next branch I see some extra stuff in the diff. I will open a clean PR with a cherry-pick and close this one. https://github.com/marmelab/react-admin/pull/8792

Issue #8182, opened 2022-09-20
"MUI: The value provided to Autocomplete is invalid." error on deselection 


PawelSuwinski:
**What you were expecting:**

Case of autocomplete on reference input with paging data provider. Initial list shows only limited, first page of fetched data. Autocomplete query filter reload the list. Deselecting selected value should restore initial state (empty value and initial first page list) without any warnings. 

Example of usage:
```
      <ReferenceInput {...props} defaultValue="">
        <AutocompleteInput
          filterToQuery={(searchText) => ({ title: searchText })}
          optionText="title"
        />
      </ReferenceInput>
```

**What happened instead:**

If selected value is from out of first page deselecting of it will restore the initial state but also raise a warning rised on the console, for example:

```js
MUI: The value provided to Autocomplete is invalid.
None of the options match with `{"id":12,"title":"Qui tempore rerum et v(...)
```

**Environment**

* React-admin version: 4.4.3
* React version: ^17.0.0 | ^18.0.0
* Browser: Firefox

WiXSL:
Thanks for reporting this. Please provide a sample application showing the issue by forking the following CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple).

PawelSuwinski:
Here it is:
https://codesandbox.io/s/broken-leftpad-knxqnl

Steps:
- click "+" to create new comment 
- click on post autocomplete dropdown
- type "Sed" and select last listed item or filter and select any post out site from initial list
- deselect selected post

See the console:
![obraz](https://user-images.githubusercontent.com/1256986/192502100-f4835ad1-5c5d-4077-adc4-5b481866767b.png)

WiXSL:
Reproduced.
Thanks

ns-vpanfilov:
I am experiencing the same issue with react-admin 4.8.2

mohammad-23:
is there any update on this? Facing this on 4.5.0.

slax57:
@mohammad-23 you're right this has been open for too long. We'll try to take some time to address this issue within the next 2 weeks.

PR #8788, opened 2023-03-31
Fix MUI Autocomplete warning on deselection with ReferenceInput

slax57:
Fixes https://github.com/marmelab/react-admin/issues/8182

## Todo

- [x] Fix
- [ ] Unit test -> did not manage to do it :slightly_frowning_face:

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.spec.tsx
@@ -165,7 +165,7 @@ describe('useReferenceInputController', () => {
         );
 
         await waitFor(() => {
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

dpyryev:
What about `packages/ra-core/src/controller/input/useReferenceArrayInputController.ts` ?

`ReferenceArrayInput` has the same issue. RA version 4.16.

I mean. Let say you have options AAA, BBB, CCC
Select  any at first eg. `AAA`
Then try to type `B` or any else, that is absolutely differs from selected value.
You will get same error and flickering value, text you just typed will be cleared.

What I have seen at `useAutocomplete.js` that current value contained [AAA] while list of options was [BBB]

PR #8797, opened 2023-04-03
feat(input): full width on low screen resolution

### LINKED ISSUE #8795: Make the filter inputs full width by default on low screen resolution (mobile)
It make more sense to me to have all the input taking a full width when we are on mobile view.

So instead of this being the default:

<details><summary>Before</summary>

Screen Shot 2023-04-03 at 21 00 22

</details> 

I suggest to have this:

<details><summary>After</summary>

Screen Shot 2023-04-03 at 21 04 26

</details> 

This is what I am willing to on my project. I may work on that but I would like to be sure about the idea first.

Thanks!

soullivaneuh:
# Summary

closes #8795

### DIFF
--- a/packages/ra-ui-materialui/src/defaultTheme.ts
+++ b/packages/ra-ui-materialui/src/defaultTheme.ts
@@ -22,6 +22,19 @@ export const defaultTheme = {
         closedWidth: 50,
     },
     components: {
+        MuiAutocomplete: {
+            variants: [
+                {
+                    props: {},
+                    style: ({ theme }) => ({
+                        [theme.breakpoints.down('sm')]: {
+                            // @see MuiTextField.styleOverrides.root
+                            width: '100%',
+                        },
+                    }),
+                },
+            ],
+        },
         MuiFilledInput: {
             variants: [
                 {
@@ -41,6 +54,17 @@ export const defaultTheme = {
                 margin: 'dense' as const,
                 size: 'small' as const,
             },
+            variants: [
+                {
+                    props: {},
+                    style: ({ theme }) => ({
+                        [theme.breakpoints.down('sm')]: {
+                            // @see MuiTextField.styleOverrides.root
+                            width: '100%',
+                        },
+                    }),
+                },
+            ],
         },
         MuiFormControl: {
             defaultProps: {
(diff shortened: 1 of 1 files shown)


soullivaneuh:
I am not sure about what fix to provide for the [typecheck issue](https://github.com/marmelab/react-admin/actions/runs/4602030405/jobs/8130569948?pr=8797).

May you give me a hint?

slax57:
> I am not sure about what fix to provide for the [typecheck issue](https://github.com/marmelab/react-admin/actions/runs/4602030405/jobs/8130569948?pr=8797).
> 
> May you give me a hint?
> 
> I followed that documentation: https://mui.com/material-ui/customization/theme-components/#theme-style-overrides

It's odd, but it seems that, although this syntax is documented, MUI is not willing to support it in the long term and hence not willing to support it in the TS types (source: https://github.com/mui/material-ui/issues/29319)

You might have to find another way to do this...

soullivaneuh:
@slax57 Thanks!

That is quite a bad news because the other found I started to get into was to add the `fullWidth` property logic in each input component, which will lead to a maintenance nightmare in my opinion.

According to [that comment](https://github.com/mui/material-ui/issues/29319#issuecomment-952587677), we may be able to achieve our needs using `theme.components[Component].variants`. I'll dig onto that.

soullivaneuh:
I got that fixed using variants! :tada: 

As the usage of `styleOverrides`  looks not recommended for the future, I also took the initiative to migrate the existing on a separated commit: b217bb621

packages/ra-ui-materialui/src/defaultTheme.ts
```diff
@@ -22,22 +22,49 @@ export const defaultTheme = {
...
+                    props: {},
+                    style: {
                         backgroundColor: 'rgba(0, 0, 0, 0.04)',
+                        '&$disabled': {
+                            backgroundColor: 'rgba(0, 0, 0, 0.04)',
+                        },
                     },
                 },
-            },
+            ],
```
slax57, line 49: Can you explain why you needed to change these lines too?

soullivaneuh: Please see: https://github.com/marmelab/react-admin/pull/8797#issuecomment-1496479298

> As the usage of styleOverrides looks not recommended for the future, I also took the initiative to migrate the existing on a separated commit: https://github.com/marmelab/react-admin/commit/b217bb621e94097bcfe38a58ee75c26585092353

If you prefer to keep the change to the target, I can simply drop this commit. :+1:

fzaninotto:
This PR doesn't fix the aforementioned problem, see the deployed storybook https://react-admin-storybook-jxzza3gyv-marmelab.vercel.app/?path=/story/ra-ui-materialui-list-list--filters

image

fzaninotto:

fzaninotto:
Superseded by #9169

PR #8798, opened 2023-04-03
fix(input): width consistency

### LINKED ISSUE #8796: Inconsistent width between text and autocomplete inputs
**What you were expecting:**

The text and autocomplete should have the same base width.

**What happened instead:**

The base width of the autocomplete input is smaller than the text input.

**Steps to reproduce:**

https://react-admin-storybook-mpmncf8xi-marmelab.vercel.app/?path=/story/ra-ui-materialui-input-arrayinput--autocomplete-first

<details><summary>Details</summary>
<p>

image

</p>
</details> 

**Other information:**

**Environment**

* React-admin version: latest
* React version: from storybook
* Browser: Firefox

soullivaneuh:
# Summary

closes #8796

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -651,11 +651,7 @@ export const AutocompleteInputClasses = {
 const StyledAutocomplete = styled(Autocomplete, {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
-})(({ theme }) => ({
-    [`& .${AutocompleteInputClasses.textField}`]: {
-        minWidth: theme.spacing(20),
-    },
-}));
+})(() => ({}));
 
 // @ts-ignore
 export interface AutocompleteInputProps<
--- a/packages/ra-ui-materialui/src/defaultTheme.ts
+++ b/packages/ra-ui-materialui/src/defaultTheme.ts
@@ -1,5 +1,7 @@
 import { ThemeOptions } from '@mui/material';
 
+const inputMinWidth = 222;
+
 export const defaultTheme = {
     palette: {
         background: {
@@ -22,6 +24,13 @@ export const defaultTheme = {
         closedWidth: 50,
     },
     components: {
+        MuiAutocomplete: {
+            styleOverrides: {
+                inputRoot: {
+                    minWidth: inputMinWidth,
+                },
+            },
+        },
         MuiFilledInput: {
             styleOverrides: {
                 root: {
@@ -38,12 +47,22 @@ export const defaultTheme = {
                 margin: 'dense' as const,
                 size: 'small' as const,
             },
+            styleOverrides: {
+                root: {
+                    minWidth: inputMinWidth,
+                },
+            },
         },
         MuiFormControl: {
             defaultProps: {
                 variant: 'filled' as const,
                 margin: 'dense' as const,
                 size: 'small' as const,
+            },
+            styleOverrides: {
+                root: {
+                    minWidth: inputMinWidth,
+                },
             },
         },
     },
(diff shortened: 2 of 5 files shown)


fzaninotto:
Why did you change the default text input width? 

## Before

image


## After

image

This will break many existing layouts, so that would be considered a breaking change.

soullivaneuh:
Because I didn't find the reference of the original size so I created a new one that looks acceptable visually.

However, it can be easily changed if you have a referenced size to give to me! :+1:

soullivaneuh:
I just figured out you are already using a custom `minWidth` style property on some component, including `AutocompleteInput`:

- https://github.com/marmelab/react-admin/blob/88c777c033cacbeb086983cce83392849e43eadc/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx#L656
- https://github.com/marmelab/react-admin/blob/989a9ccd34583a1b372984eca95262f608c7564f/packages/ra-ui-materialui/src/input/SelectInput.tsx#L424
- https://github.com/marmelab/react-admin/blob/989a9ccd34583a1b372984eca95262f608c7564f/packages/ra-ui-materialui/src/input/SelectInput.tsx#L424
- https://github.com/marmelab/react-admin/blob/69d1cd1af77a432e2d1a3c2215ee93a2dedcd4be/packages/ra-ui-materialui/src/layout/LinearProgress.tsx#L56

Problems:

- It is a per component customization that leads to maintenance weakness. I suggest to replace them by a global configuration like I started
- I did not find any width setting for the basic `TextField` neither here or on the `material-ui` project. I surely did a wrong search and I am accepting any hint!

I updated the merge request to change the default width to `222`, matching what I have by inspecting the default `TextField` width with the browser inspector. Yeah, not that great.

I also remove the `minWidth: theme.spacing(20)` for the benefit of global theming.

Some input stories was cleanup from manually set width because it does not show the default display. I may revert it or move the custom width demo to a separated story if needed (I prefer the second option to keep the default display).

Please let me know what do you think of the proposition and how I can improve this hard-coded width value. :upside_down_face:

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -651,11 +651,7 @@ export const AutocompleteInputClasses = {
 const StyledAutocomplete = styled(Autocomplete, {
     name: PREFIX,
     overridesResolver: (props, styles) => styles.root,
-})(({ theme }) => ({
-    [`& .${AutocompleteInputClasses.textField}`]: {
-        minWidth: theme.spacing(20),
-    },
-}));
+})(() => ({}));
```
soullivaneuh, line 654: I had to keep this empty function in order to not have the unit test breaking and kept the previous config line of the `styled` function call.

I am not very comfortable with the usage of the `styled` function. Is the `StyledAutocomplete` component still useful on that case?

slax57:
I fear this could be a BC in the layout for many apps. We should probably keep this change for RA v5.

fzaninotto:
I have mixed feelings about this PR. 

Pros:
- It's good to look for a nice alignment by default

Cons: 
- You still change the default TextInput width, which will break existing layouts (compare [the storybook for your PR](https://react-admin-storybook-41yqze5ik-marmelab.vercel.app/?path=/story/ra-ui-materialui-forms-simpleform--basic) with [the storybook for the master branch](https://react-admin-storybook.vercel.app/?path=/story/ra-ui-materialui-forms-simpleform--basic))
- You fix the TextInput width, while MUI leaves it to the default browser width (which I think is a better choice)
- You introduce a constant in the theme that can't be changed

So I don't think this is a good solution to the problem.

fzaninotto:
As seen in #8798, we won't override MUI's default on inputs, so I'm closing this PR. Thanks for your work!

PR #8800, opened 2023-04-04
Fix FilterForm infinite loop when used in a ReferenceManyField

### LINKED ISSUE #8642: Make ReferenceManyField compatible with filters

Today there is not guidance and support for filters within a ReferenceManyField. I am having difficulties trying to integrate react-admin filter system within a ReferenceManyField component.

Digging into the source code of the `List` component, it looks like the react-admin way to setup filters in ReferenceManyField would be to write something like:

```jsx
<ReferenceManyField
  reference="comments"
  target="postId"
  pagination={<Pagination />}
>
  <ListToolbar
    filters={commentFilters}
    actions={
      <TopToolbar>
        <FilterButton />
      </TopToolbar>
    }
  />
  <Datagrid bulkActionButtons={false}>
  ...
  </Datagrid>
</ReferenceManyField>
```

But this generates infinite loop when writing within a filter input (probably due to the fact that reference many field context is not exactly similar to the list context, especially in terms of URL sync settings).

Note 1: Today, ReferenceManyField pagination state is not storable in the URL and this feature request might be related to this subject: could a solution be to be able to specify a "urlKey" props to the ReferenceManyField component so that all the state of this field can be stored in URL without conflicting with possible other components?

Note 2: I am not quite sure to understand why react-admin needs to implement a `useReferenceManyFieldController`: why ReferenceManyField component can't directly use a `useListController` hook instead?

fzaninotto:
Closes #8642

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
@@ -20,10 +20,11 @@ import {
     useForm,
     useFormContext,
 } from 'react-hook-form';
-import lodashSet from 'lodash/set';
-import lodashUnset from 'lodash/unset';
-import lodashGet from 'lodash/get';
+import set from 'lodash/set';
+import unset from 'lodash/unset';
+import get from 'lodash/get';
 import cloneDeep from 'lodash/cloneDeep';
+import isEqual from 'lodash/isEqual';
 
 import { FilterFormInput } from './FilterFormInput';
 import { FilterContext } from '../FilterContext';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -48,7 +49,9 @@ export const FilterForm = (props: FilterFormProps) => {
     // Reapply filterValues when the URL changes or a user removes a filter
     useEffect(() => {
         const newValues = getFilterFormValues(form.getValues(), filterValues);
-        form.reset(newValues);
+        if (!isEqual(newValues, form.getValues())) {
+            form.reset(newValues);
+        }
```
fzaninotto, line 54: This is the fix - the rest of the changes in this file are just cosmetic

PR #8802, opened 2023-04-04
Add ability to change the sort, filter and selection of ArrayField

### LINKED ISSUE #8761: ArrayField does not support onUnselectItems prop in ListContextProvider, causing console error
I encountered an issue with the ArrayField component in react-admin. When using the bulk actions menu and attempting to unselect items, it does not unselect the selected items, and an error is logged in the developer console: "onUnselectItems is not a function". This is because the ListContextProvider created by ArrayField sets the onUnselectItems prop to null.

### Steps to reproduce:

1. Create a react-admin application with a resource that has an ArrayField.
2. Use the Datagrid component inside the ArrayField to display nested fields.
3. Enable bulk actions in the resource's list view.
4. Select some items in the nested list and click the "unselect items" button in the bulk actions menu.

### Expected behavior:
The selected items in the nested list should be unselected, and no error should appear in the console.

### Actual behavior:
The selected items remain selected, and an error is logged in the developer console: "onUnselectItems is not a function".

### Proposed solution:
Update the ArrayField component's implementation of the ListContextProvider to handle the onUnselectItems prop correctly. This could be done by providing a default implementation for the onUnselectItems prop or by allowing it to be passed as a prop to the ArrayField component.

### Additional context:
I have created a workaround by creating a custom ArrayField component that wraps the original ArrayField and implements the ListContextProvider with the necessary properties. However, it would be better if this functionality were included in react-admin's core implementation.

fzaninotto:
## Problem

`<ArrayField>` creates a very limited `ListContext` that doesn't allow manipulating the list

## Solution

Use `useList` to prepare the List context. It already does all we need. 

Closes #8761

## To Do

- [x] use `useList` in `<ArrayField>` to create a complete `ListContext``
- [x] add support for `sort`, `filter`, and `perPage` props
- [x] Add story
- [x] Add tests
- [x] Add documentation

## Example

Leveraging `perPage` and `<Pagination>` in the `<ArrayField>`:

```jsx
import { 
    ArrayField,
    Datagrid,
    Pagination,
    Show,
    SimpleShowLayout,
    TextField
} from 'react-admin';

const PostShow = () => (
    <Show>
        <SimpleShowLayout>
            <TextField source="title" />
            <ArrayField source="backlinks" perPage={5}>
                <Datagrid>
                    <TextField source="uuid" />
                    <TextField source="date" />
                    <TextField source="url" />
                </Datagrid>
                <Pagination />
            </ArrayField>
        </SimpleShowLayout>
    </Show>
);
```


packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -162,8 +162,6 @@ const Root = styled('div', {
 })(({ theme }) => ({
     display: 'flex',
     flexWrap: 'wrap',
-    marginTop: theme.spacing(-1),
```
fzaninotto, line 165: This looked bad in a SimpleShowLayout. It was designed to make Datagrid rows thinner, but we can do it in user land via the `sx` prop (see modified PostList).

djhi (approved):

packages/ra-ui-materialui/src/field/ArrayField.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+    <MemoryRouter>
+        <ArrayField record={{ id: 123, books }} source="books">
+            <SingleFieldList>
+                <ChipField source="title" />
+            </SingleFieldList>
+        </ArrayField>
+    </MemoryRouter>
+);
+
+export const PerPage = () => (
```
djhi, line 37: This one would benefit from setting up the I18nProvider

djhi (changes requested):

packages/ra-ui-materialui/src/field/ArrayField.stories.tsx
fzaninotto: You're right, fixed!

djhi (approved):

PR #8803, opened 2023-04-04
Fix `useReferenceArrayInputController` does not pass meta to `getMany`

### LINKED ISSUE #8772: Missing meta in getMany dataprovider function when using ReferenceArrayInput
**What you were expecting:**
access meta in getMany dataProvider function when it's called by ReferenceArrayInput.

**What happened instead:**
 meta in getMany dataProvider function is empty

**Steps to reproduce:**
set something like `queryOptions={{ meta: { pretty: true } }}` in ReferenceArrayInput.
look at getMany fn in dataprovider -> params.meta is empty
```

```

**Related code:**

https://codesandbox.io/p/sandbox/wispy-water-155fuq?welcome=true&file=%2FREADME.md

`PostEdit.tsx`
```
...
<ReferenceArrayInput
        {...props}
        perPage={5}
        filter={{ published }}
        queryOptions={{ meta: { pretty: true } }}
>
        <AutocompleteArrayInput
          create={<CreateTag />}
          optionText={`name.${locale}`}
        />
</ReferenceArrayInput>
...
```

`dataProvider.tsx`
```
...
   getMany: (resource, params) => {
    console.log("missing meta", params);
    return dataProvider.getMany(resource, params);
  },
  getList: (resource, params) => {
    console.log("not missing meta", params);
    return dataProvider.getList(resource, params);
  }
...
```

**Environment**

* React-admin version: 4.9.0
* React version: 17.0.2
* Browser:

djhi:
Fix #8772

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceArrayInputController.spec.tsx
+++ b/packages/ra-core/src/controller/input/useReferenceArrayInputController.spec.tsx
@@ -3,13 +3,20 @@ import { ReactElement } from 'react';
 import expect from 'expect';
 import { render, screen, waitFor, fireEvent } from '@testing-library/react';
 
-import { useReferenceArrayInputController } from './useReferenceArrayInputController';
+import {
+    useReferenceArrayInputController,
+    UseReferenceArrayInputParams,
+} from './useReferenceArrayInputController';
 import { CoreAdminContext } from '../../core';
 import { testDataProvider } from '../../dataProvider';
-import { Form } from '../../form';
+import { ChoicesContextValue, Form } from '../../form';
 import { SORT_ASC } from '../list/queryReducer';
 
-const ReferenceArrayInputController = props => {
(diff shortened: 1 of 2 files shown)


fzaninotto:
So we pass the same `meta` prop to `useGetMany` and `useGetList`? Isn't it going to break things?

slax57 (approved):

slax57:
> So we pass the same `meta` prop to `useGetMany` and `useGetList`? Isn't it going to break things?

In theory it indeed could be a BC for some (pretty specific) use-cases.

However, if my recollection is correct I believe we had the same discussion with the (EE) TreeController, and we figured it was best to pass the `meta` parameter everywhere and let the `dataProvider` decide where to use it or not. So in this regard it would improve the consistency.

Hence I'm in favor of this change.

fzaninotto (approved):
Fair enough

PR #8804, opened 2023-04-04
[Minor BC] Fix `<Datagrid>` does not apply `className` to its root element

slax57:
## Problem

`<Datagrid>` does not apply `className` to its root element, but rather to its nested `Table` element.

This creates a discrepancy between styles applied via `sx` and styles applied via `styled`.

This is among others visible with the (EE) `<EditableDatagrid>` component.

## Solution

`sx` and `className` should be applied to the same component, i.e. `<DatagridRoot>`.

:warning: :warning: :warning: **Breaking change** :warning: :warning: :warning: 

This change is technically a BC. But it should be really minor as it would affect only people that use `styled` to apply styles to the `Datagrid` component. People applying styles via `sx` shouldn't be affected.

## Screenshots (of the added story)

Before

Capture d’écran du 2023-04-04 16-56-52

After

Capture d’écran du 2023-04-04 16-56-26

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
@@ -9,7 +9,7 @@ import {
     useGetList,
     useList,
 } from 'ra-core';
-import { Box } from '@mui/material';
+import { Box, styled } from '@mui/material';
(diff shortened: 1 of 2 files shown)


fzaninotto:
To be moved to next

djhi (approved):

PR #8811, opened 2023-04-05
Better types for useRedirect and useCreatePath

djhi:
Provide hints for shortcuts values but still allow to provide a string, false or a function:

image

### DIFF
--- a/packages/ra-core/src/routing/useRedirect.ts
+++ b/packages/ra-core/src/routing/useRedirect.ts
@@ -41,9 +41,9 @@ export const useRedirect = () => {
         (
             redirectTo: RedirectionSideEffect,
             resource: string = '',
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/routing/useRedirect.ts
```diff
@@ -41,9 +41,9 @@ export const useRedirect = () => {
         (
             redirectTo: RedirectionSideEffect,
             resource: string = '',
-            id?: Identifier,
-            data?: Partial<RaRecord>,
-            state: object = {}
+            id?: Identifier | null,
+            data?: Partial<RaRecord> | null,
+            state: object | null = {}
```
djhi, line 46: I can either keep the null or fix the tests

slax57 (approved):

PR #8814, opened 2023-04-06
Provide TS validation for props related to Record type on Fields

djhi:
Follow #8766


djhi:
Now that the way to go is clear, I will open small PRs to ease the review

PR #8815, opened 2023-04-06
Add some utility generics for better typings

IAmVisco:
## Summary

Just a couple of generics that should be there.


djhi (changes requested):
Thanks! This is technically a new feature though. Can you target the `next` branch?

djhi (approved):
Thanks!

PR #8821, opened 2023-04-09
[Doc] Improve usage examples for `<ReferenceInput>` and `<ReferenceArrayInput>`

fzaninotto:
## Problem

For `<SelectInput>` and other selection inputs, we document fetching related records by advertising the solution ("using in `<ReferenceInput>` rather than the problem ("Fetching choices"). 

As a consequence, newcomers struggle to understand that `ReferenceInput` thing. 

## Solution

Write the doc the other way around.


WiXSL (commented):
There are several files that make reference to the anchor `#using-in-a-referenceinput`, but there is no such anchor

WiXSL (changes requested):

slax57 (approved):

Issue #8799, opened 2023-04-04
form validation function runs on on change for onSubmit forms


ns-vpanfilov:
**What you were expecting:**

I have an async function that validates that email belongs to an active user. Although mode is set to `onSubmit`, the function is executed on change for. I expect validation function to only run on change 
```js
const emailValidators= async (v) => {
  console.log(v)
  return undefined
}

return (
    <SimpleForm mode='onSubmit' reValidateMode='onSubmit'>
      <TextInput label='Email' source='email' validate={emailValidators} />
    </SimpleForm>
  )
```

**What happened instead:**
Validation function runs on key press (console.log is triggered in the example above)

**Environment**

* React-admin version: 4.9.0
* React version: 18.2.0

slax57:
Can you please provide a reproduction sandbox?

ns-vpanfilov:
@slax57 , if you go to "Create Post"  and type in any fields, you will notice that counter at the top is refreshing via "validate function

[SANDBOX LINK](https://codesandbox.io/p/sandbox/hopeful-sid-plcl4w?file=%2Fsrc%2Fposts%2FPostCreate.tsx&selection=%5B%7B%22endColumn%22%3A22%2C%22endLineNumber%22%3A10%2C%22startColumn%22%3A22%2C%22startLineNumber%22%3A10%7D%5D)

![image](https://user-images.githubusercontent.com/105231423/229800402-9fbd5b26-9702-4979-85b3-a876d42ba781.png)

ns-vpanfilov:
It is worth noting that field will not be marked as invalid until submit is pressed; however validate function still runs on each field change, increasing load on REST api server. Also note that changing a different field (`Teaser`) will also trigger validate function on `Title`

ns-vpanfilov:
I think react-hook-form works fine. Here is an example - only runs validate onSubmit

[LINK to hook form](https://codesandbox.io/s/sample-react-hook-form-validation-onsubmit-8uxqek?file=/src/App.tsx:0-1154)

ns-vpanfilov:
@slax57, I also noticed that that the entire form re-renders on each event in any field (onChange and onBlur). See link below that displays validation function run counts and re-render counts. Notice that react-hook-form does not re-render oChange and onBlur.

[...]

PR #8826, opened 2023-04-11
Fix Form should only trigger field validation on submit if not told otherwise

slax57:
Fix https://github.com/marmelab/react-admin/issues/8799

## Todo

- [x] Fix
- [x] Test
- [x] Non-regression tests with EE components
- [x] Mention the related RHF rule in OSS docs

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -891,6 +891,13 @@ const PersonEdit = () => (
 );
 ```
 
+**Reminder:** [react-hook-form's `formState` is wrapped with a Proxy](https://react-hook-form.com/api/useformstate/#rules) to improve render performance and skip extra computation if specific state is not subscribed. So, make sure you deconstruct or read the `formState` before render in order to enable the subscription.
+
+```js
+const { isDirty } = useFormState(); // ✅
+const formState = useFormState(); // ❌ should deconstruct the formState      
+```
+
 ## Third-Party Components
 
 You can find components for react-admin in third-party repositories.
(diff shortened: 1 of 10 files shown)


docs/Inputs.md
```diff
@@ -891,6 +891,13 @@ const PersonEdit = () => (
 );
 ```
 
+**Reminder:** [react-hook-form's `formState` is wrapped with a Proxy](https://react-hook-form.com/api/useformstate/#rules) to improve render performance and skip extra computation if specific state is not subscribed. So, make sure you deconstruct or read the `formState` before render in order to enable the subscription.
```
djhi, line 894: This one is weird as we don't use `useFormState` anywhere before or did I miss something?

docs/useInput.md
```diff
@@ -100,6 +100,13 @@ const LatLngInput = props => {
 };
 ```
 
+**Reminder:** [react-hook-form's `formState` is wrapped with a Proxy](https://react-hook-form.com/api/useformstate/#rules) to improve render performance and skip extra computation if specific state is not subscribed. So, make sure you deconstruct or read the `formState` before render in order to enable the subscription.
```
djhi, line 103: Same

docs/useInput.md
```diff
@@ -146,3 +153,10 @@ const PersonEdit = () => (
     </Edit>
 );
 ```
+
```
djhi, line 156: Same

djhi (changes requested):

docs/Inputs.md
slax57: We use the `formState` (`formState: { isSubmitted }`) so I figured I'd rather add the Reminder here anyway. But I'm open to discussion :grin:

djhi (approved):

PR #8830, opened 2023-04-12
fix: `<BooleanInput helperText={false} />` displays `FormHelperText` anyway

### LINKED ISSUE #8829: Inputs should not display helper text if `helperText` is set to `false` without validation errors
**What you were expecting:**
Setting `helperText={false}` should hide helperText element in input components.

**What happened instead:**
An empty paragraph is generated introducing a small vertical offset - 3px by default 😱.

Most of the input components share the same behaviour, but usually RA uses false to prevent a ReactNode parameter to be displayed. Am I wrong on this point?
 
**Steps to reproduce:**

- Open simple example below
- Edit any post
- Inspect inputs, helperText elements are outlined for easier tracking

**Related code:**
https://stackblitz.com/edit/github-hnmzx9?file=src%2Fposts%2FPostEdit.tsx

#### For example `<BooleanInput>` generates the following HTML

```html
<div class="MuiFormGroup-root ra-input ra-input-commentable css-dmmspl-MuiFormGroup-root">
  {...}
  <p class="MuiFormHelperText-root css-1d1r5q-MuiFormHelperText-root"></p>
</div>
```

septentrion-730n:
Fix #8829

### DIFF
--- a/packages/ra-ui-materialui/src/input/BooleanInput.tsx
+++ b/packages/ra-ui-materialui/src/input/BooleanInput.tsx
@@ -91,13 +91,15 @@ export const BooleanInput = (props: BooleanInputProps) => {
                     />
                 }
             />
-            <FormHelperText error={(isTouched || isSubmitted) && invalid}>
-                <InputHelperText
-                    touched={isTouched || isSubmitted}
-                    error={error?.message}
-                    helperText={helperText}
-                />
-            </FormHelperText>
+            {helperText !== false && (
+                <FormHelperText error={(isTouched || isSubmitted) && invalid}>
+                    <InputHelperText
+                        touched={isTouched || isSubmitted}
+                        error={error?.message}
+                        helperText={helperText}
+                    />
+                </FormHelperText>
+            )}
         </FormGroup>
     );
 };


djhi (changes requested):
This also prevent validation messages to be displayed

djhi (changes requested):

packages/ra-ui-materialui/src/input/BooleanInput.tsx
```diff
@@ -91,13 +91,15 @@ export const BooleanInput = (props: BooleanInputProps) => {
...
                 }
             />
-            <FormHelperText error={(isTouched || isSubmitted) && invalid}>
-                <InputHelperText
-                    touched={isTouched || isSubmitted}
-                    error={error?.message}
-                    helperText={helperText}
-                />
-            </FormHelperText>
+            {helperText !== false && (
```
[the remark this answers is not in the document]
septentrion-730n: I misunderstood default behaviour, thanks for your attention @djhi !

djhi (changes requested):
If we want to do this in all inputs then maybe we should move the `FormHelperText` inside `InputHelperText` and fix it in one place?

slax57:
> If we want to do this in all inputs then maybe we should move the `FormHelperText` inside `InputHelperText` and fix it in one place?

I agree. Here are all the components I found that are affected by this issue:
- ArrayInput
- BooleanInput
- CheckboxGroupInput
- DatagridInput
- FileInput
- RadioButtonGroupInput
- SelectArrayInput

Interesting thing to note, is that `ArrayInput` already had a similar implementation in place (but it's the only one):

```tsx
            {!!((isDirty || isSubmitted) && !!error) || helperText ? (
                <FormHelperText error={(isDirty || isSubmitted) && !!error}>
                    <InputHelperText
                        touched={isDirty || isSubmitted}
                        error={error?.root?.message ?? error?.message}
                        helperText={helperText}
                    />
                </FormHelperText>
            ) : null}
```

djhi (approved):

packages/ra-ui-materialui/src/input/BooleanInput.tsx
```diff
@@ -91,13 +91,16 @@ export const BooleanInput = (props: BooleanInputProps) => {
...
                 }
             />
-            <FormHelperText error={(isTouched || isSubmitted) && invalid}>
-                <InputHelperText
-                    touched={isTouched || isSubmitted}
-                    error={error?.message}
-                    helperText={helperText}
-                />
-            </FormHelperText>
+            {(helperText !== false ||
```
djhi, line 94: Can you make this a ternary ?
```helperText !== false ? .... : null```

**Review by djhi — CHANGES_REQUESTED** [review]

File: packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx [hunk 3]
```diff
@@ -235,16 +235,20 @@ export const CheckboxGroupInput: FunctionComponent<CheckboxGroupInputProps> = pr
...
-                error={fetchError || ((isTouched || isSubmitted) && !!error)}
-                className={CheckboxGroupInputClasses.helperText}
-            >
-                <InputHelperText
-                    touched={isTouched || isSubmitted || fetchError}
-                    error={error?.message || fetchError?.message}
-                    helperText={helperText}
-                />
-            </FormHelperText>
+            {helperText !== false ? (
```
slax57, line 238: I disagree with this implementation. To me we should display the `<FormHelperText>` when there is a validation error, even when `helperText === false`

slax57 (changes requested):

adguernier:
If I understand the subject of this PR correctly: we want to remove the space after an input when `helperText={false}` is set on an input component. 
image

So, and after investigation, I think this PR is useless because, `helperText={false}` does the job already, thanks to `InputHelperText` implementation that return `null` if `helperText === false`, but return error if exist.

PR #8838, opened 2023-04-18
docs: repeated information

ms10596:
Adding meta to dataProvider call is already mentioned in `queryOptions` section.


slax57:
I believe this duplicate is intended.
Indeed we believe that putting key information at multiple places in the docs is necessary to increase their visibility and discoverability.
Hence I'd rather keep this section, although it's a duplicate.
Thank you for your contribution in any case, and feel free to open new PRs if you find other issues with the docs!

Issue #8819, opened 2023-04-09
redirect and warnWhenUnsavedChanges don't work together


Dreamsorcerer:
**What you were expecting:**
redirect to work when warnWhenUnsavedChanges is set.

Both seem to work individually.
If no redirect is specified, then Create will correctly redirect to the Edit view.
If warnWhenUnsavedChanges is not given, then Create will redirect to the view specified by the redirect attribute.
But, when both are used, the redirect is broken and does not redirect at all.

**What happened instead:**
No redirect happened. It just gives a notification that the element was created, but the form doesn't change.

**Steps to reproduce:**
Set `redirect="show"` in the Create element and `warnWhenUnsavedChanges` in the SimpleForm.

**Related code:**
```
const AiohttpCreate = (resource, name, permissions) => (
    <Create redirect="show">
        <SimpleForm warnWhenUnsavedChanges>
            {createInputs(resource, name, "add", permissions)}
        </SimpleForm>
    </Create>
);
```

**Other information:**
Documentation suggests a couple of other props can't be used with the redirect, but nothing I saw seemed to suggest it couldn't be used with warnWhenUnsavedChanges.

**Environment**

* React-admin version: 4.9.1
* React version: 18.2
* Browser: Firefox

djhi:
Thanks for reporting this. Please provide a sample application showing the issue by forking the following CodeSandbox (https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple).

Dreamsorcerer:
It just says 'Initializing Sandbox Container' forever while consuming all the CPU...

Not sure how easy it is for you to run Python, but this is the actual code:
https://github.com/aio-libs/aiohttp-admin/blob/redirect/admin-js/src/App.js#L228-L234

It should be possible to run with something like:
```
git clone https://github.com/aio-libs/aiohttp-admin.git
git checkout redirect
cd aiohttp-admin/
pip3 install -r requirements.txt  # Or maybe pip (without the 3)
cd admin-js/
yarn install
yarn build
cd ../examples/
PYTHONPATH='..' python3 simple.py  # Or maybe python (without the 3).
```
Then go to http://localhost:8080/admin
(login and password is 'admin')
Then just click create, enter a random number and value and click save. It fails to redirect.

slax57:
@Dreamsorcerer same as https://github.com/marmelab/react-admin/issues/8817#issuecomment-1504887822

[...]

PR #8839, opened 2023-04-18
Fix custom redirect in `Create` and `Edit` with `warnWhenUnsavedChanges`

slax57:
Fix https://github.com/marmelab/react-admin/issues/8819

### DIFF
--- a/packages/ra-core/src/controller/create/useCreateController.ts
+++ b/packages/ra-core/src/controller/create/useCreateController.ts
@@ -97,8 +97,9 @@ export const useCreateController = <
                     : data
             ).then(async (data: Partial<RecordType>) => {
                 const mutate = getMutateWithMiddlewares(create);
+                let shouldRedirect = false;
                 try {
-                    await mutate(
+                    const savedData = await mutate(
                         resource,
                         { data, meta },
                         {
@@ -118,12 +119,7 @@ export const useCreateController = <
                                     type: 'info',
                                     messageArgs: { smart_count: 1 },
                                 });
-                                redirect(
-                                    finalRedirectTo,
-                                    resource,
-                                    data.id,
-                                    data
-                                );
+                                shouldRedirect = true;
                             },
                             onError: onErrorFromSave
                                 ? onErrorFromSave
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -203,6 +212,9 @@ export const useEditController = <
                                   },
                         }
                     );
+                    if (shouldRedirectAfterUpdate) {
```
fzaninotto, line 215: You should add the same comment as for the useCreateController here, otherwise it's unclear why you need to do that.

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -151,6 +147,18 @@ export const useCreateController = <
                                   },
                         }
                     );
+                    if (shouldRedirect) {
+                        // We need to wait for the 'isSubmitting' for state to be updated
```
fzaninotto, line 151: I don't understand that comment.

fzaninotto (changes requested):

slax57:
What fixed the issue in my PR was probably luck, because I am avoiding the unfavorable case of a race condition, but not fixing the race condition itself.

Basically the issue is the following:

`useAugmentedForm` defines the form's `handleSubmit` fn as:
```ts
        async (values, event) => {
            // ...
            let errors = await saveContext.save(values, event);
            // ...
        }
```

The `saveContext.save` is defined by the Edit/Create controller as:
```ts
const [update] = useUpdate(resource, recordCached, {
        ...otherMutationOptions,
        mutationMode,
        returnPromise: mutationMode === 'pessimistic',
});

const save = useCallback(
    data => 
        async data => {
            const mutate = getMutateWithMiddlewares(update);
            await mutate(
                resource,
                { id, data, meta: mutationMeta },
                {
                    onSuccess: async data => {
                        redirect(redirectTo, resource, data.id, data);
                    }
                }
        }
);
```

Which means `save` will call `redirect()` before resolving.

However, in `useWarnWhenUnsavedChanges`, we define:

```ts
        let unblock = navigator.block((tx: Transition) => {
            if (
                !isSubmitting &&
                (isSubmitSuccessful ||
                    window.confirm(translate('ra.message.unsaved_changes')))
            ) {
                unblock();
                tx.retry();
            }
        });
```

Which means the navigation **will be blocked** when `isSubmitting==true`.

This is why the redirect fails silently, it is actually blocked by `useWarnWhenUnsavedChanges`, which thinks the form is still submitting, because the `handleSubmit` fn RA provides to RHF has not resolved yet.

My fix was basically delaying the call to `redirect()` a bit later, but it still does not depend on the RHF's form submitting status, so it is probably not robust enough...

So far I can't think of a good way to fix this. I to put more thinking into this...

slax57:
Superseded by https://github.com/marmelab/react-admin/pull/8859

PR #8844, opened 2023-04-18
Clarify DateField types

elstgav:
Clarifies the types for `DateField`. Both `locales` and `options` are expected to be used in `.toLocaleDateString`, which has more specific (and built-in/native) types we can leverage. Setting to more explicit types provides useful auto-correct and documentation.


slax57 (changes requested):
Indeed, thanks for pointing this out! :slightly_smiling_face: 

Could you also please update the type of `options` in the test `DateField.spec.tsx` line 110?

elstgav:
@slax57 Done ✅. I made the changes through GitHub’s UI vs a full editor (hoping this is a small/simple edit 😄). Please let me know if I need to download locally to verify anything 👍🏻 (in case CI has some blindspots).

PR #8847, opened 2023-04-20
fix: credentials and headers belong to HttpLink

rlucia:
We need to pass credentials and headers to HttpLink to correctly build a fetch instance

CoAuthored with @dgopsq


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8850, opened 2023-04-22
fix: valid value for prop `sx` on <div> tag using ImageField

RoBYCoNTe:
If you have many menu items, navigating between them can produce temporary invalid records with empty sourceValue, in this case sometimes we received this error **valid value for prop `sx` on <div> tag** related to the usage of `ImageField` with specification of `sx` prop. 

To avoid this we replaced standard `div` with `Typography` that can accept sx without problems.


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8862, opened 2023-04-27
Update RaRecord type & fix some other types

djhi:
Initial review of RaRecord type to allow future work on TS validation & completion of props such as `source`, `optionText`, etc


examples/crm/src/deals/DealShow.tsx
```diff
@@ -48,7 +49,7 @@ export const DealShow = ({ open, id }: { open: boolean; id?: string }) => {
 };
 
 const DealShowContent = () => {
-    const record = useRecordContext();
+    const record = useRecordContext<Deal>();
```
djhi, line 52: This is a BC, back to WIP

fzaninotto (approved):

PR #8863, opened 2023-04-27
Add TS support to fields

djhi:
Follow #8862 

If users provide types **extending** `RaRecord`, nothing changes, they won't have TS validation of the `source` prop but they won't have any error either. Same when they don't provide types.

If users provide types that don't extend `RaRecord`, they'll get validation and completion on the `source` prop


fzaninotto (changes requested):

examples/crm/src/types.ts
```diff
@@ -1,12 +1,12 @@
-import { RaRecord, Identifier } from 'react-admin';
-
-export interface Sale extends RaRecord {
+export type Sale = {
```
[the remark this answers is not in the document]
djhi: Reverted

docs/Fields.md
```diff
@@ -597,4 +597,34 @@ You can find components for react-admin in third-party repositories.
...
+}
+
+export const PostShow = () => (
+    <Show>
+        <SimpleShowLayout>
+            <TextField<Post> source="title" />
+            <TextField<Post> source="teaser" />
+            <RichTextField<Post> source="body" />
+            <DateField<Post> label="Publication date" source="published_at" />
+        </SimpleShowLayout>
```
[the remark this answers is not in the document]
djhi: Added

examples/demo/src/types.ts
```diff
@@ -1,13 +1,13 @@
-import { RaRecord, Identifier } from 'react-admin';
-
 export type ThemeName = 'light' | 'dark';
 
-export interface Category extends RaRecord {
+export type Category = {
```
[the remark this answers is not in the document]
djhi: Reverted

examples/demo/src/visitors/ColoredNumberField.tsx
```diff
@@ -13,6 +13,8 @@ const ColoredNumberField = (props: NumberFieldProps) => {
     );
 };
 
-ColoredNumberField.defaultProps = NumberField.defaultProps;
+ColoredNumberField.defaultProps = {
```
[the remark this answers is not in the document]
djhi: Reverted

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -74,30 +75,37 @@ import { PublicFieldProps, InjectedFieldProps, fieldPropTypes } from './types';
...
-export const ArrayField = memo<ArrayFieldProps>(props => {
+const ArrayFieldImpl = <
+    RecordType extends Record<string, unknown> = Record<string, any>
+>(
+    props: ArrayFieldProps<RecordType>
+) => {
     const { children, resource, source, perPage, sort, filter } = props;
     const record = useRecordContext(props);
-    const data = get(record, source, emptyArray) || emptyArray;
+    const data = (get(record, source, emptyArray) as any[]) || emptyArray;
```
fzaninotto, line 85: `Array<Record<string, any>>` would be better than `any[]`

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -74,30 +75,37 @@ import { PublicFieldProps, InjectedFieldProps, fieldPropTypes } from './types';
...
-
-// @ts-ignore
-ArrayField.propTypes = {
-    ...fieldPropTypes,
 };
 
-export interface ArrayFieldProps extends PublicFieldProps, InjectedFieldProps {
+export const ArrayField = genericMemo(ArrayFieldImpl);
+
+ArrayFieldImpl.propTypes = fieldPropTypes;
```
fzaninotto, line 96: I'd expect this line to come before the memo

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -74,30 +75,37 @@ import { PublicFieldProps, InjectedFieldProps, fieldPropTypes } from './types';
...
+export interface ArrayFieldProps<
+    RecordType extends Record<string, unknown> = Record<string, any>
+> extends FieldProps<RecordType> {
     children: ReactNode;
     perPage?: number;
     sort?: SortPayload;
     filter?: FilterPayload;
 }
 
+ArrayFieldImpl.displayName = 'ArrayFieldImpl';
```
fzaninotto, line 108: same

packages/ra-ui-materialui/src/field/BooleanField.tsx
```diff
@@ -85,11 +88,14 @@ BooleanField.propTypes = {
     looseValue: PropTypes.bool,
 };
 
+BooleanField.propTypes = BooleanFieldImpl.propTypes;
```
fzaninotto, line 91: Same, please define the proptypes of the implementation first then copy it to the memoized component. By the way, can't you bake in this proptype copy in the `genericMemo` component? Same for the `displayName` copy.

packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
```diff
@@ -22,3 +22,32 @@ export const Basic = (props: BooleanFieldProps) => {
 export const NoFalseIcon = () => <Basic FalseIcon={null} />;
 
 export const NoTrueIcon = () => <Basic TrueIcon={null} />;
+
+type Post = {
```
fzaninotto, line 26: That's not a record, it requires an `id`. I'm surprised that TypeScript doesn't warn you about this one.

packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
```diff
@@ -22,3 +22,32 @@ export const Basic = (props: BooleanFieldProps) => {
...
+            <input
+                type="checkbox"
+                checked={published}
+                onChange={e => setPublished(e.target.checked)}
+            />
+            <BooleanField<Post>
+                record={{ published, deep: { reported: false }, title: '' }}
+                source="deep.reported"
+                sortBy="published"
+                {...props}
```
fzaninotto, line 49: Why do you need that? It shows a practice that shouldn't be encouraged.

packages/ra-ui-materialui/src/field/DateField.tsx
```diff
@@ -70,7 +74,13 @@ export const DateField: FC<DateFieldProps> = memo(props => {
         ) : null;
     }
 
-    const date = value instanceof Date ? value : new Date(value);
+    const date =
+        value instanceof Date
+            ? value
+            : typeof value === 'string' || typeof value === 'number'
+            ? new Date(value)
+            : undefined;
```
fzaninotto, line 82: 👍 TypeScript is useful after all!

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -103,54 +109,48 @@ ReferenceField.defaultProps = {
...
-    const createPath = useCreatePath();
-    const resourceDefinition = useResourceDefinition({ resource: reference });
-
-    const resourceLinkPath =
-        link === false ||
-        (link === 'edit' && !resourceDefinition.hasEdit) ||
-        (link === 'show' && !resourceDefinition.hasShow)
-            ? false
-            : createPath({
-                  resource: reference,
```
fzaninotto, line 135: Any particular reason you moved that logic down to PureReferenceFieldView? By the way that's a BC, as ReferenceFieldView is public.

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
```diff
@@ -84,7 +87,7 @@ export const ReferenceOneField = <RecordType extends RaRecord = any>(
                 isLoading={isLoading}
                 isFetching={isFetching}
                 referenceRecord={referenceRecord}
-                resourceLinkPath={resourceLinkPath}
+                link={resourceLinkPath}
```
fzaninotto, line 90: If you had to change it here, it will break other components (e.g. Enterprise components), so you should revert this change.

packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -37,7 +41,7 @@ export const FileField = (props: FileFieldProps) => {
         ...rest
     } = props;
     const record = useRecordContext(props);
-    const sourceValue = get(record, source);
+    const sourceValue = get(record, source) as string;
```
fzaninotto, line 44: Why don't you use `toString()` as in `RichTextField` and `TextField`? Seems a better way to make sure that what you get is a string. 

Same for other `as string` forced casting.

fzaninotto (changes requested):
Awesome!

packages/ra-core/src/types.ts
```diff
@@ -14,8 +13,8 @@ export interface RaRecord<IdentifierType extends Identifier = Identifier>
     id: IdentifierType;
 }
 
-export interface SortPayload {
-    field: string;
+export interface SortPayload<FieldType extends string = string> {
+    field: FieldType;
```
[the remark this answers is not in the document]
djhi: I'll remove that but later, users could provide their type and by default, this will match the resource properties as well, just like `source`

packages/ra-ui-materialui/package.json
```diff
@@ -62,6 +62,7 @@
         "clsx": "^1.1.1",
         "css-mediaquery": "^0.1.2",
         "dompurify": "^2.4.3",
+        "hotscript": "^1.0.12",
```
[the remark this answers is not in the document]
djhi: Because **you** asked me to put it in dependencies ^^
image

packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -151,7 +151,7 @@ interface Props<
     mutationOptions?: UseMutationOptions<
         RecordType,
         MutationOptionsError,
-        CreateParams<RecordType> | UpdateParams<RecordType>
+        CreateParams | UpdateParams<RecordType>
```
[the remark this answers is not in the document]
djhi: Can revert for now but related to our discussion about the RecordType and the FormValuesType

packages/ra-ui-materialui/src/field/types.ts
```diff
@@ -1,19 +1,48 @@
...
+    emptyText?: string;
+    fullWidth?: boolean;
+    record?: RecordType;
+    resource?: string;
+}
 
-export interface PublicFieldProps {
-    sortBy?: string;
+/**
+ * @deprecated use FieldProps instead
```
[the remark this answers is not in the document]
djhi: Because it's a duplicate of `Fields` that has no purpose. Why keep it?

packages/ra-ui-materialui/src/field/BooleanField.stories.tsx
djhi: Because our fields don't require a RaRecord anymore

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -74,30 +75,35 @@ import { PublicFieldProps, InjectedFieldProps, fieldPropTypes } from './types';
...
             {children}
         </ListContextProvider>
     );
-});
-
-// @ts-ignore
-ArrayField.propTypes = {
-    ...fieldPropTypes,
 };
+ArrayFieldImpl.propTypes = fieldPropTypes;
```
fzaninotto, line 95: We used to clone the prop types before. There was probably a reason, I think you shouldn't change that.

examples/crm/src/contacts/TagsListEdit.tsx
```diff
@@ -61,18 +61,16 @@ export const TagsListEdit = () => {
...
         update('contacts', {
             id: record.id,
             data: { tags },
             previousData: record,
         });
     };
 
-    const handleAddTag = (id: Identifier) => {
-        const tags: Identifier[] = [...record.tags, id];
+    const handleAddTag = (id: number) => {
```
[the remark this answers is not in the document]
fzaninotto: Bump

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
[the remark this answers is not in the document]
fzaninotto: bump

packages/ra-ui-materialui/package.json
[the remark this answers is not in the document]
fzaninotto: bump

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/ArrayField.tsx
djhi: I can't think of a valid reason. We do the same in EmailField.

I think someone did not simplify the syntax when removing an old propType specific to the ArrayField that did not make sense in v4. See https://github.com/marmelab/react-admin/commit/f7d9f18505996a3af92f29b31d3bc1d104052dfe

fzaninotto: I vaguely remember having to clone proptypes for an obscure reason. Let's revert it if we have no reason to do it.

djhi: What about the EmailField, UrlField and WrapperField then?

fzaninotto: Let's keep them as in master. I know, in master there is no consistency. Let's keep it like that. 

My opinion is that this PR changes quite a lot of stuff, and we should avoid all unnecessary changes to reduce any potential BC break.

fzaninotto (approved):

akillingbeck89:
Just upgraded from 4.5.1 to 4.11.1 and having TS issues all over the place where I set my "source" parameter, is there a way to disable this?

slax57:
@akillingbeck89 This change should be backwards compatible. If it is not, this is to be considered as a bug.
Can you please open a new issue with more information about the errors you are getting, and ideally a reproduction sandbox?
This way we can better grasp what's going on and offer adequate help.
For instance, it would help to know if you are passing custom types to the fields or not.

grevolution:
> Dont have time right now, will do it next week, but no custom types are being passed and an example error I'm getting is:
> 
> Type '"channelId"' is not assignable to type '`[${key}]` | `[${key}].${string}` | null'.ts(2322)

How did you fix this issue?

PR #8874, opened 2023-05-02
Add ability to default to dark mode when users prefer it

### LINKED ISSUE #8670: Changing the theme using ToggleThemeButton does not persist function overrides
**What you were expecting:**

Toggling theme would persist all overrides.

**What happened instead:**

Overrides defined with a function get lost after theme toggle

**Steps to reproduce:**

After loading the app with clean localStorage, the theme loads correctly, but after clicking the `ToggleThemeButton`, the value gets saved to the `localStorage` and the function overrides get lost:

image
image

This happens to both Mui and RA components, with the only exception being RA components not having typings, but that's probably a separate issue and not as critical: 

image

**Other information:**

I understand you can't serialize functions reliably, so maybe store only the theme mode in the store? Just thoughts.

**Environment**

* React-admin version: 4.8.0
* Last version that did not exhibit the issue (if applicable): Not sure
* React version: 18
* Browser: Chrome 109

fzaninotto:
## Problems

- When using an app with two themes, it defaults to the light theme even though the user has a system preference for dark theme
- When developers change the theme, if a user has used the theme toggle in the past, they won't get the theme change
- Themes must be serializable to JSON to use the ToggleThemeButton, which forbids themes with functions

##  Solution

- If the developer specifies an `<Admin darkTheme>` prop, and a user prefers dark mode, the app defaults to dark mode
- The `useTheme` hook now saves a theme *mode* ('light' or 'dark') in the preferences rather than a theme *object*
- `<ToggleThemeButton>` no longer needs theme props

The solution is made backwards compatible by keeping the old `<ToggleThemeButton>`.

## Example

```jsx
// in src/MyAppBar.js
import { AppBar, TitlePortal, ToggleThemeButton } from 'react-admin';

export const MyAppBar = () => (
    <AppBar>
        <TitlePortal />
        <ToggleThemeButton />
    </AppBar>>
);

// in src/App.js
import { Admin, Layout } from 'react-admin';
import { MyAppBar } from './MyAppBar';
import { darkTheme, lightTheme } from './themes';

const MyLayout = (props) => <Layout {...props} appBar={MyAppBar} />;

const App = () => (
    <Admin
        dataProvider={dataProvider}
        layout={MyLayout} 
        theme={lightTheme}
        darkTheme={darkTheme}
    >
        ...
    </Admin>
);
```

Closes #8670

- [x] Add `<Admin darkTheme>` prop
- [x] Add `<Admin defaultToLighTheme>` prop
- [x] Create `ThemesContext`
- [x] Update `<ToggleThemeButton>` to use it
- [x] Update the AppBar to display the `<ToggleThemeButton>` by default when a dark mode is defined
- [x] Use a media query to default to the user preferred theme
- [x] Make it backwards compatible
- [x] Write documentation
- [x] Add tests
- [x] Use the `<ToggleThemeButton>` in one of our demos (ecommerce)


docs/Admin.md
fzaninotto, line 1: I moved and rewrote a lot of things in there. Best to read the new version rather than read the diff.

packages/ra-ui-materialui/src/AdminContext.tsx
```diff
@@ -1,16 +1,37 @@
...
+        lightTheme = defaultLightTheme,
+        darkTheme,
+        defaultToLightTheme,
+        children,
+        ...rest
+    } = props;
     return (
         <CoreAdminContext {...rest}>
-            <ThemeProvider theme={theme}>{children}</ThemeProvider>
+            <ThemesContext.Provider
```
fzaninotto, line 18: I'm not fond of this syntax (ThemesContextProvider wraps ThemeProvider), but it's a good way to make the whole change backwards compatible.

packages/ra-ui-materialui/src/button/Button.stories.tsx
```diff
@@ -85,10 +85,10 @@ const theme = createTheme({
 
 declare module '@mui/material/styles' {
     interface Palette {
-        userDefined: PaletteColor;
+        userDefined?: PaletteColor;
```
fzaninotto, line 88: This was making TS compilation fail in other tests

fzaninotto:
Switching to RFR

docs/Admin.md
```diff
@@ -27,31 +33,156 @@ export default App;
...
+| `dataProvider`         | Required | `DataProvider` | -              | The data provider for fetching resources                 |
+| `children`             | Required | `ReactNode`    | -              | The routes to render                                     |
+| `authCallbackPage`     | Optional | `Component`    | `AuthCallback` | The content of the authentication callback page          |
+| `authProvider`         | Optional | `AuthProvider` | -              | The authentication provider for security and permissions |
+| `basename`             | Optional | `string`       | -              | The base path for all URLs                               |
+| `catchAll`             | Optional | `Component`    | `NotFound`     | The fallback component for unknown routes                |
+| `dashboard`            | Optional | `Component`    | -              | The content of the dashboard page                        |
+| `darkTheme`            | Optional | `object`       | -              | The dark theme configuration                             |
+| `defaultTo LightTheme` | Optional | `boolean`      | `false`        | Flag to default to the light theme                       |
```
djhi, line 149:
```suggestion
| `defaultToLightTheme` | Optional | `boolean`      | `false`        | Flag to default to the light theme                       |
```

packages/ra-ui-materialui/src/layout/Theme/ThemesContext.ts
```diff
@@ -0,0 +1,10 @@
+import { createContext } from 'react';
+import { RaThemeOptions } from './types';
+
+export const ThemesContext = createContext<ThemesContextValue>({});
+
+export interface ThemesContextValue {
+    darkTheme?: RaThemeOptions;
+    lightTheme?: RaThemeOptions;
+    defaultToLightTheme?: boolean;
```
djhi, line 9: Why no a `defaultTheme` instead, allowing you to default to the dark theme if you want?

djhi (changes requested):
Looks good!

docs/Admin.md
fzaninotto: This was made on purpose to add a line break and make the table look okayish despite the large content

packages/ra-ui-materialui/src/AdminContext.tsx
fzaninotto: I think the existing FIXMEs introduced in this PR will suffice

packages/ra-ui-materialui/src/layout/Theme/ThemesContext.ts
fzaninotto: 👍 You're right, it's better. Added in the next commit.

PR #8882, opened 2023-05-04
Fix custom redirect in pessimistic `Edit` or `Create` when using `warnWhenUnsavedChanges` (new approach)

slax57:
Fix https://github.com/marmelab/react-admin/issues/8819

Supersedes https://github.com/marmelab/react-admin/pull/8859

The idea is that changing the way the save side-effects should be called is a BC.
Instead, this PR aims at fixing the `useWarnWhenUnsavedChanges` hook.

The fix is working well (tested manually and with Jest).
However it does seem a bit hacky, so let's discuss!


djhi (approved):
:+1:

PR #8884, opened 2023-05-05
fix: synched prop types definition with exported props interface

RoBYCoNTe:
If you use `ImageField` inside a `FileInput`, you will probably receive warning about invalid prop `file` because, when parsing with PropTypes, this is defined as an object but in the interface is defined as any.

I've alligned these two props. Is it correct?


slax57:
Hi,
Thank you for your contribution.
The fix would seem appropriate, however I'm not able to reproduce the issue.
Actually we have this use-case in our [storybook](https://react-admin-storybook.vercel.app/?path=/story/ra-ui-materialui-input-imageinput--basic), since `ImageInput` is basically a wrapper around `FileInput` with specific labels, but I don't reproduce the issue there.
Can you provide a reproduction (e.g. in [stackblitz](https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple))?
Thanks

RoBYCoNTe:
Sorry, I'm late. 
Right now I'm not able to recreate this behavior with standard stackblitz config :(

I can provide you an example screenshot of the error, but I've to declare few things:
- I'm working on an application that uses a custom theme
- I'm using custom dataProvider and authProvider
- I've completed override the AdminContext with my own.

I don't know if this can be useful but that is:

Screenshot 2023-05-19 alle 12 52 20

fzaninotto:
Nope, it's not useful. We need a way to reproduce it and look at the code at the same time.

slax57:
@RoBYCoNTe did you manage to reproduce the issue in Stackblitz?

RoBYCoNTe:
Sorry but right now I'm not able to reproduce it outside my project, this is a really strange situation, give me more time to understand how to replicate it 😅

fzaninotto:
I'm closing this PR because it's been inactive for almost 2 months. Please comment with new information if you want to reopen it.

PR #8887, opened 2023-05-06
[typescript] Fix sidebar props

smeng9:
fix types for sidebar in layout so we can pass custom components that uses sidebar props


djhi (changes requested):
Can you fix the linter warning? Thanks!

smeng9 (commented):
fix linter warning

djhi (approved):

djhi:
Thanks!

PR #8903, opened 2023-05-11
Fix custom input's onChange handlers should have access to updated context value

WiXSL:
I'm updating a project from ra-v3 to ra-v4 where the value of change inputs is taken from the Form state within the onChange / onBlur event handler (instead of its `e.target.value` argument), and it doesn't have the updated value.

I found out that :
In react-admin **v3** inputs run custom event handlers **after** react-final-form ones.
In react-admin **v4** inputs run custom event handlers **before** react-hook-form ones.

So, changing this order of execution to match ra-v3 allows `useFormContext().getValues()` to have the latest value inside the custom event handlers.

### DIFF
--- a/packages/ra-core/src/form/useInput.spec.tsx
+++ b/packages/ra-core/src/form/useInput.spec.tsx
@@ -17,6 +17,25 @@ const Input: FunctionComponent<
     return children(inputProps);
 };
 
+const InputWithCustomOnChange: FunctionComponent<
+    {
+        children: (props: ReturnType<typeof useInput>) => ReactElement;
+    } & InputProps
+> = ({ children, ...props }) => {
+    const { getValues } = useFormContext();
+
+    return (
+        <Input
+            {...props}
+            onChange={e => {
+                props.onChange(e, getValues()[props.source]);
+            }}
+        >
+            {children}
+        </Input>
+    );
+};
+
 describe('useInput', () => {
     it('returns the props needed for an input', () => {
         let inputProps;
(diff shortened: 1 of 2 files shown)


slax57:
Hi! Thanks for submitting this! :slightly_smiling_face: 

However, I'm not sure I understand, though. You can still access the newer value of the input via `e.target.value`, right? So why would you need to access it via `useFormContext().getValues()`? What is the benefit? :thinking:

WiXSL:
The benefit is to grab all the values of the form at once to give it to a process, including the one that just have been changed, without having to differentiate that one by grab it from another place.
I have a lot of code that takes all the values from the event handlers like that, and I have to go handler by handler taking the value that has changed from e.target.value plus the rest of the form values.

slax57:
I see. Thanks for the clarifications. :slightly_smiling_face: 
I was wondering, because this change is technically a BC.
However I can't think of any scenario where this would be problematic, so I would be OK to accept the change.
@djhi  @fzaninotto  what do you think? Can you think of a scenario where this might break things?

slax57 (approved):

packages/ra-core/src/form/useInput.spec.tsx
```diff
@@ -106,6 +126,46 @@ describe('useInput', () => {
...
     });
 
+    it('custom onChange handler should have access to updated context input value', () => {
+        let targetValue, contextValue;
+        const handleChange = e => {
+            targetValue = e.target.value;
+        };
+        const getContextValue = value => {
+            contextValue = value;
+        };
```
fzaninotto, line 136: This one should probably be named `setXXX`, it's not a getter

fzaninotto (changes requested):
I agree that this is a welcome change, however I'd prefer that you PR against `next`. As it may break things in corner cases, we need to prevent that from happening in a bug fix release. 

Also, the order change was introduced by @djhi in 9dfaf4e6357864e9435c2a8eb369655e89b9869a. I wonder it you have any recollection of why you didi so?

PR #8905, opened 2023-05-12
Add `purifyOptions` prop to `<RichTextField>`

slax57:
Also document how to open all links in a new tab in a secure way, by [adding a hook to DOMPurify](https://github.com/cure53/DOMPurify/issues/317#issuecomment-698800327).

### DIFF
--- a/packages/ra-ui-materialui/src/field/RichTextField.tsx
+++ b/packages/ra-ui-materialui/src/field/RichTextField.tsx
@@ -31,6 +31,7 @@ export const RichTextField: FC<RichTextFieldProps> = memo<RichTextFieldProps>(
             emptyText,
             source,
             stripTags = false,
+            purifyOptions,
             ...rest
         } = props;
         const record = useRecordContext(props);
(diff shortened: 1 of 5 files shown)


djhi (commented):
Added questions

packages/ra-ui-materialui/src/field/RichTextField.tsx
```diff
@@ -64,13 +65,20 @@ RichTextField.propTypes = {
     ...Typography.propTypes,
     ...fieldPropTypes,
     stripTags: PropTypes.bool,
+    purifyOptions: PropTypes.any,
+};
+
+export type PurifyOptions = purify.Config & {
+    RETURN_DOM_FRAGMENT?: false | undefined;
```
djhi, line 72: Why? What are those?

slax57: I'll add a comment to explain

djhi (approved):

PR #8907, opened 2023-05-12
Bump RHF version and Simplify `useInput` validation

djhi:


slax57 (approved):

slax57 (changes requested):
I realize, since we are bumping the RHF version, we should probably keep this for the next minor version, don't you think?

PR #8910, opened 2023-05-15
Fix custom input's onChange handlers should have access to updated context value

WiXSL:
I'm updating a project from ra-v3 to ra-v4 where the value of change inputs is taken from the Form state within the onChange / onBlur event handler (instead of its `e.target.value` argument), and it doesn't have the updated value.

I found out that :
In react-admin **v3** inputs run custom event handlers **after** react-final-form ones.
In react-admin **v4** inputs run custom event handlers **before** react-hook-form ones.

So, changing this order of execution to match ra-v3 allows `useFormContext().getValues()` to have the latest value inside the custom event handlers.

Supersedes #8903

### DIFF
--- a/packages/ra-core/src/form/useInput.spec.tsx
+++ b/packages/ra-core/src/form/useInput.spec.tsx
@@ -17,6 +17,26 @@ const Input: FunctionComponent<
     return children(inputProps);
 };
 
+const InputWithCustomOnChange: FunctionComponent<
+    {
+        children: (props: ReturnType<typeof useInput>) => ReactElement;
+    } & InputProps & { getContextValue?: (value: string) => void }
+> = ({ children, getContextValue, ...props }) => {
+    const { getValues } = useFormContext();
+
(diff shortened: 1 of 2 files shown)


slax57:
@djhi May I let you check [this comment](https://github.com/marmelab/react-admin/pull/8903#pullrequestreview-1426314893) and then merge if you approve this PR? :slightly_smiling_face:

WiXSL:
@slax57, @djhi, you didn't resolve what to do about this PR. Could you re-open it?

djhi:
Sorry, the branch was deleted when I closed the merge master to next PR. Thanks for reporting!

djhi (approved):

WiXSL:
I believe this wasn't mentioned in the latest ChangeLog.
Just a heads-up

slax57:
Right, thanks for the heads-up @WiXSL :slightly_smiling_face:

PR #8911, opened 2023-05-15
Simplify form reset on record change, and support custom `resetOptions`

slax57:
- [x] Leverage [new feature in RHF](https://github.com/react-hook-form/react-hook-form/releases/tag/v7.41.0) allowing to call `form.reset()` automatically when the `record` (or `defaultValues`) change
- [x] Forward the `resetOptions` prop from `Form` to RHF's `useForm` to allow for customization of the reset options
- [x] Remove unused and undocumented hook `useInitializeFormWithRecord`

### Additional info

Had to change the tests because this comes with a minor BC, in case the `defaultValues` change when there is no `record`:
- Before: the form was not reset
- Now: the form is reset
=> But the way I see it it's rather a bug fix than a BC, because I would expect changes in the `defaultValues` to be reflected in the form, regardless of the presence of a `record`.


packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -1119,7 +1119,7 @@ describe('<AutocompleteInput />', () => {
                 <SimpleForm
                     mode="onBlur"
                     onSubmit={jest.fn()}
-                    defaultValues={{ role: 2 }}
+                    defaultValues={{ language: 'ang' }}
```
slax57, line 1122: To me these tests have nothing to do with changing the `defaultValues` (I think it's a copy paste error), hence I changed that to keep the same `defaultValues` across rerenders.

djhi (approved):

PR #8915, opened 2023-05-16
Fix `SimpleForm` and `TabbedForm` do not sanitize the `resetOptions` prop

slax57:
Necessary since https://github.com/marmelab/react-admin/pull/8911 which allows the new `resetOptions` prop


djhi (approved):

djhi:
Would you mind extracting a `sanitizeFormProps` function?

PR #8917, opened 2023-05-16
Add WithListContext component

fzaninotto:
## Problem

Reading data from a `ListContext` requires creating another component, as it's only available via a hook.

## Solution 

A render prop version of `useListContext` called `WithListContext`

```jsx
import { List, Datagrid, TextField, ReferenceArrayField, WithListContext } from 'react-admin';
import { Chip, Stack } from '@mui/material';

const BookList = () => (
    <List>
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <ReferenceArrayField label="Tags" reference="tags" source="tag_ids">
                <WithListContext render={({ isLoading, data }) => (
                    !isLoading && (
                        <Stack direction="row" spacing={1}>
                            {data.map(tag => (
                                <Chip key={tag.id} label={tag.name} />
                            ))}
                        </Stack>
                    )
                )} />
            </ReferenceArrayField>
        </Datagrid>
    </List>
);
```

This also facilitates the use of charts in react-admin:

```jsx
import { ListBase, WithListContext } from 'react-admin';
import { LineChart, Line, XAxis, YAxis, CartesianGrid } from 'recharts';

const FruitChart = () => (
    <ListBase resource="fruits" disableSyncWithLocation perPage={100}>
        <WithListContext
            render={({ data }) => (
                <LineChart width={700} height={300} data={data}>
                    <XAxis dataKey="date" />
                    <YAxis />
                    <CartesianGrid stroke="#eee" strokeDasharray="5 5" />
                    <Line dataKey="apples" type="monotone" stroke="#8884d8" />
                    <Line dataKey="blueberries" type="monotone" stroke="#82ca9d" />
                    <Line dataKey="carrots" type="monotone" stroke="#ffc658" />
                </LineChart>
            )}
        />
    </ListBase>
);
```

WithListContext-chart

- [x] Create the component
- [x] Create a storybook
- [x] Document it
- [x] Add mentions of it in other parts of the documentation


docs/WithListContext.md
```diff
@@ -0,0 +1,184 @@
...
+
+`<WithListContext>` executes its `render` function using the current `ListContext` as parameter. It's the render prop version of [the `useListContext` hook](./useListContext.md).
+
+Use it to render a list of records already fetched.
+
+## Usage
+
+The most common use case for `<WithListContext>` is to build a custom list view on-the-fly, without creating a new component, in a place where records are available inside a `ListContext`. 
+
+For instance, an list of book tags fetched via [`<ReferenceArrayField>`](./ReferenceArrayField.md):
```
djhi, line 16:
```suggestion
For instance, a list of book tags fetched via [`<ReferenceArrayField>`](./ReferenceArrayField.md): 
```

docs/WithListContext.md
```diff
@@ -0,0 +1,184 @@
...
+
+Whether you use `<WithListContext>` or `useListContext` is a matter of coding style.
+
+## Props
+
+`<WithListContext>` accepts a single `render` prop, which should be a function.
+
+## `render`
+
+A function which will be called with the current [`ListContext`](./useListContext.md) as argument. It should return a React element.
```
djhi, line 84:
```suggestion
A function which will be called with the current [`ListContext`](./useListContext.md) as argument. It should return a React node.
```

docs/WithListContext.md
```diff
@@ -0,0 +1,184 @@
...
+    resource, // the resource name, deduced from the location. e.g. 'posts'
+    refetch, // callback for fetching the list data again
+}) => ( ... )}>
+```
+
+## Availability
+
+Whenever you use a react-admin component fetching a list of records, react-admin stores it in a [`ListContext`](./useListContext.md). This means that `<WithListContext>` works out of the box:
+
+- in descendants of the [`<List>`](./ListBase.md), [`<InfiniteList>`](./InfiniteList.md), and [`<ListBase>`](./ListBase.md) components
```
djhi, line 139:
```suggestion
- as a child of the [`<List>`](./ListBase.md), [`<InfiniteList>`](./InfiniteList.md), and [`<ListBase>`](./ListBase.md) components
```

djhi: I find the existing sentence confusing

djhi (changes requested):

docs/WithListContext.md
fzaninotto: Your suggestion leads to another confusion: that the component would only work when used as child and not as grandchild. 

I think that "descendants" is the right term.

djhi (approved):

PR #8923, opened 2023-05-17
Fix Create generic record type

### LINKED ISSUE #8921: The "id" field for the RaRecord type at Create makes no sense
**What you were expecting:**
I do not expect an error because my request in the body does not require an id

**What happened instead:**
I'm getting a "TS2344: Type 'MyTypeDto' does not satisfy the constraint 'RaRecord'.   Property 'id' is missing in type 'MyTypeDto' but required in type 'RaRecord'" error

**Steps to reproduce:**
Add a type without an "id" field and pass in the generic type for the Create component

**Related code:**
```
import { Create } from 'react-admin';

type MyTypeDto = {
  name: string;
  someKey: string;
};

export default function CreateDetails() {
  return (
    <Create<MyTypeDto> title="Add Details" resource="send-to">
      ...
    </Create>
  );
}
```

**Environment**

* React-admin version: 4.9.0
* React version: 18.2.0
* Stack trace (in case of a JS error): TS types.d.ts(9, 5): 'id' is declared here.

djhi:
Fixes #8921

### DIFF
--- a/packages/ra-core/src/controller/create/CreateBase.tsx
+++ b/packages/ra-core/src/controller/create/CreateBase.tsx
@@ -37,7 +37,7 @@ import { ResourceContextProvider } from '../../core';
  *     </CreateBase>
  * );
  */
-export const CreateBase = <RecordType extends RaRecord = any>({
+export const CreateBase = <RecordType extends Omit<RaRecord, 'id'> = any>({
     children,
     ...props
 }: CreateControllerProps<RecordType> & { children: ReactNode }) => {
--- a/packages/ra-core/src/controller/create/CreateBase.tsx
+++ b/packages/ra-core/src/controller/create/CreateBase.tsx
@@ -5,7 +5,7 @@ import {
     CreateControllerProps,
 } from './useCreateController';
 import { CreateContextProvider } from './CreateContextProvider';
-import { RaRecord } from '../../types';
+import { Identifier, RaRecord } from '../../types';
 import { ResourceContextProvider } from '../../core';
 
 /**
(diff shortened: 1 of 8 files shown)


packages/ra-core/src/types.ts
```diff
@@ -209,7 +209,7 @@ export interface CreateParams<T = any> {
     data: T;
     meta?: any;
 }
-export interface CreateResult<RecordType extends RaRecord = any> {
+export interface CreateResult<RecordType extends Omit<RaRecord, 'id'> = any> {
```
slax57, line 212: I believe that `CreateResult` should still require an `id`

packages/ra-core/src/controller/create/CreateBase.tsx
```diff
@@ -37,11 +37,20 @@ import { ResourceContextProvider } from '../../core';
...
     children,
     ...props
-}: CreateControllerProps<RecordType> & { children: ReactNode }) => {
-    const controllerProps = useCreateController<RecordType>(props);
+}: CreateControllerProps<RecordType, Error, ResultRecordType> & {
+    children: ReactNode;
+}) => {
+    const controllerProps = useCreateController<
+        RecordType,
+        Error,
```
slax57, line 51: why not set this to `unknown` instead of `Error`?

djhi: Ideally we should accept an Error generic parameter in CreateBase too. I figured it would be more convenient to have a default Error

slax57 (approved):

PR #8630, opened 2023-02-07
Use rowSx instead of rowStyle to style selected cell based on row record

zhujinxuan:
This PR intends to add `rowSx`instead of `rowStyle` for styling the Datagrid rows for two reasons:
1. `Sx` is more common in mui instead of `style` to style.
2. `rowStyle` will apply to the whole row. If one wants to highlight specific cells in the datagrid body, for example highlight duplicate phone numbers, they have to write their own `DatagridBody` and `DatagridBodyRow`. Meanwhile, `rowSx` can fix the problem.

`rowStyle` has been kept in this PR for compatibility.

I have not yet updated the spec and docs according to `rowSx'. If you think this change is good, I will update docs and spec in the same PR.

PR #8925, opened 2023-05-17
Replace rowStyle with rowSx

zhujinxuan:
This PR intends to substitute `rowStyle` by `rowSx` for styling the `Datagrid` and `SimpleList` rows for two reasons:
1. `Sx` is more common in mui instead of `style` to style.
2. `rowStyle` will apply to the whole row. If one wants to highlight specific cells in the datagrid body, for example highlight duplicate phone numbers, they have to write their own `DatagridBody` and `DatagridBodyRow`. Meanwhile, `rowSx` can fix the problem.

Follow up of https://github.com/marmelab/react-admin/pull/8630

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -644,7 +644,7 @@ export const PostList = () => (
 
 ## `rowSx`
 
-You can customize the styles of rows and cells in `<Datagrid>` (applied to the `<DatagridRow>` element) based on the record, thanks to the `rowSx` prop, which expects a function. React-admin calls this function for each row, passing the current record and index as arguments. The function should return a Material UI `sx`, which react-admin uses as a `<TableRow sx>` prop. 
+You can customize the styles of rows and cells in `<Datagrid>` (applied to the `<DatagridRow>` element) based on the record, thanks to the `rowSx` prop, which expects a function. React-admin calls this function for each row, passing the current record and index as arguments. The function should return a Material UI [`sx`](https://mui.com/system/getting-started/the-sx-prop/), which react-admin uses as a `<TableRow sx>` prop. 
 
 For instance, this allows to apply a custom background to the entire row if one value of the record - like its number of views - passes a certain threshold.
 
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -51,6 +51,7 @@ Here are all the props accepted by the component:
 * [`isRowExpandable`](#isrowexpandable)
 * [`isRowSelectable`](#isrowselectable)
 * [`optimized`](#optimized-better-performance-for-large-tables)
+* [`rowStyle`](#rowstyle)
 * [`rowSx`](#rowsx)
 * [`rowClick`](#rowclick)
 * [`size`](#size)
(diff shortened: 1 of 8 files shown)


djhi (changes requested):
Thanks! This is nice but please make it an alternative to `rowStyle`, not a replacement.

zhujinxuan:
Sorry, I was busy this week. I will make the change next Tuesday.

zhujinxuan:
Sorry for the delay, I have had a back pain in last few weeks that kept me away from computer. I have add `rowStyle` back to the document, to the `ra-ui-materialui` and to `Datagrid.stories.tsx`. But I keep the `rowSx` change for the demo because I think it would be better if the code style of demo is more consistent with mui. 

Please let me know if there is any thing needed to be updated about this PR.

docs/Datagrid.md
```diff
@@ -641,6 +642,27 @@ export const PostList = () => (
 );
 ```
 
+## `rowSx`
+
+You can customize the styles of rows and cells in `<Datagrid>` (applied to the `<DatagridRow>` element) based on the record, thanks to the `rowSx` prop, which expects a function. React-admin calls this function for each row, passing the current record and index as arguments. The function should return a Material UI `sx`, which react-admin uses as a `<TableRow sx>` prop.
```
djhi, line 647:
```suggestion
You can customize the styles of rows and cells in `<Datagrid>` (applied to the `<DatagridRow>` element) based on the record, thanks to the `rowSx` prop, which expects a function. React-admin calls this function for each row, passing the current record and index as arguments. The function should return a Material UI [`sx`](https://mui.com/system/getting-started/the-sx-prop/), which react-admin uses as a `<TableRow sx>` prop. 
```

docs/SimpleList.md
```diff
@@ -175,6 +176,24 @@ export const PostList = () => (
 );
 ```
 
+## `rowSx`
+
+This optional prop should be a function, which gets called for each row. It receives the current record and index as arguments, and should return a style object. The style object is applied to the `<ListItem>` styles prop.
```
djhi, line 181:
```suggestion
This optional prop should be a function, which gets called for each row. It receives the current record and index as arguments, and should return a Material UI [`sx`](https://mui.com/system/getting-started/the-sx-prop/). The style object is applied to the `<ListItem>` `sx` prop.
```

djhi (approved):
Great! Thanks!

mshd (approved):

zhujinxuan:
Hi, @fzaninotto . I made this change because there is no way to style a single cells. We can add `sx` to `Datagrid`, but it cannot style single cell because cell classes in each row are the same. Currently, the only way to style a single cell is to use `nth` two times, which is not intuitive.

djhi:
I actually think `rowSx` to be pretty intuitive as you define the styles according to the record at the same place whereas `rowClass` would require you to check two different props to ensure you applied the correct className for the conditions you check on the record.

slax57:
Personally, I'm also in favor of adding the `rowSx` prop.
Besides, this feature allows for things like setting the style of the row directly depending on the record, for instance using a numeric value (like a percentage) from the record directly to compute a bg or fg color.
This would be way harder to do with `rowClass`.

fzaninotto:
Seems I'm in minority, so we'll go for it! I'll just deprecate the `rowStyle` property in another Pr to avoid confusion. 

Thanks for your contribution!

PR #8929, opened 2023-05-19
Fix `<DatagridConfigurable>` inspector hides the wrong column when using empty children

### LINKED ISSUE #8927: conditionally visible column disappears from DatagridConfigurable and column visibility switcher is bugged
**What you were expecting:**

the column is shown, visibility switched should work changing the visibility of a correct column. Typescript error might appear that the `DatagridConfigurable` does not expect null as children. This is hard to spot that we miss one column there in the app with 80 lists.

**What happened instead:**

I used the condition to show one column only in dev mode, and it happened that on PROD the first column was missing from the list
```
          <DatagridConfigurable bulkActionButtons={false}>
            {isLocalhost() && (
              <TextField
                emptyText={NUL}
                label="id"
                sortable={isVehiclePropSortable("id")}
                source="id"
              />
            )}
```
and the column switcher was not switching the visibility of the correct columns, it switched some other columns

Screenshot 2023-05-19 at 12 39 01

Probably the solution is to use `{isLocalhost() ? ... : <Fragment/>` instead - then the first column of a table is visible again and the visibility switcher works. (but switcher shows "hidden" column as "unlabelled column" which does nothing)

...

**Steps to reproduce:**

**Related code:**
https://stackblitz.com/edit/github-tpakk4?file=src%2Fposts%2FPostList.tsx,src%2Findex.tsx,src%2Fposts%2Findex.tsx
- in `PostList.tsx` you can see the first column as conditionally hidden, it does not appear on the page - then switching the visibility of any column hides a different one

**Other information:**

**Environment**

* React-admin version:
*     "react-admin": "4.8.4",
* Last version that did not exhibit the issue (if applicable): not sure
* React version:
*     "react": "18.2.0",
* Browser: chrome

fzaninotto:
Closes #8927

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
@@ -58,29 +58,23 @@ export const DatagridConfigurable = ({
 
     React.useEffect(() => {
         // first render, or the preference have been cleared
-        const columns = React.Children.map(props.children, (child, index) =>
-            React.isValidElement(child)
-                ? {
-                      index: String(index),
-                      source: child.props.source,
-                      label:
-                          child.props.label &&
-                          typeof child.props.label === 'string' // this list is serializable, so we can't store ReactElement in it
-                              ? child.props.label
-                              : child.props.source
-                              ? //  force the label to be the source
-                                undefined
-                              : // no source or label, generate a label
-                                translate(
-                                    'ra.configurable.Datagrid.unlabeled',
-                                    {
-                                        column: index,
-                                        _: `Unlabeled column #%{column}`,
-                                    }
-                                ),
-                  }
-                : null
-        ).filter(column => column != null);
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
```diff
@@ -58,29 +58,23 @@ export const DatagridConfigurable = ({
...
-                                    {
-                                        column: index,
-                                        _: `Unlabeled column #%{column}`,
-                                    }
-                                ),
-                  }
-                : null
-        ).filter(column => column != null);
+        const columns = React.Children.toArray(props.children)
+            .filter(child => React.isValidElement(child))
```
fzaninotto, line 62: I basically made the filter before the map, so that we don't remove columns (which would break the indexing), but instead never create the columns for invalid children.

djhi (approved):

PR #8934, opened 2023-05-22
[Doc] Fix tutorial with create-react-admin

fzaninotto:
Now that create-react-admin is the default installation mode, the "Getting Started" tutorial need a few adjustments.


docs/css/prism.css
```diff
@@ -6,7 +6,6 @@ pre[class*="language-"] {
   background: 0 0;
   text-shadow: 0 1px #fff;
   font-family: Consolas, Monaco, "Andale Mono", "Ubuntu Mono", monospace;
-  font-size: 1em;
```
fzaninotto, line 9: This made code in paragraph too big:

image

PR #8936, opened 2023-05-22
Update Storybook & Cypress to use Vite

djhi:


fzaninotto:
It seems the new storybook code panel rewrites the code and doesn't allow to see the surrounding code. For many of our stories, it means the code isn't explicit...

slax57 (commented):
I noticed that the Code pane no longer displays the full storybook code, but only the (merged) code of the currently displayed component:



Do you think we should keep it this way? Is there an option to go back the way it was before?

Other than that, it seems to work just fine! Well done! :clap:

djhi:
It only does that the first time right? If you start it again, it shouldn't display them

fzaninotto:
That's right. But still, can we avoid it?

djhi:
> That's right. But still, can we avoid it?

As you probably noticed, the message includes a recommanded fix that is already applied in https://github.com/marmelab/react-admin/pull/8936/files#diff-cafe2123a72c4ce3a9f7e9ee4b0e188256eb02d7ec1e54fa2cadfac99b22f92bR38 

Not sure what else I can do

.storybook/main.js
```diff
@@ -19,20 +23,25 @@ module.exports = {
     },
     babel: async options => {
         const { plugins = [] } = options;
+        console.log(plugins);
```
djhi, line 26:
```suggestion
```

fzaninotto:
You've added `loose: true` in another rule than the one advised in the logged message I pasted. 

@babel/plugin-proposal-private-property-in-object

vs

@babel/plugin-proposal-class-properties

.storybook/main.js
slax57: the `console.log` is still there

slax57 (changes requested):

slax57 (approved):

PR #8937, opened 2023-05-23
Upgrade to TS 5

djhi:


.eslintrc
```diff
@@ -1,8 +1,6 @@
 {
-    "parser": "@typescript-eslint/parser",
     "extends": ["react-app", "plugin:prettier/recommended"],
     "plugins": [
-        "@typescript-eslint",
```
slax57, line 5: FMI, why did you disable this plugin?

package.json
```diff
@@ -34,33 +34,36 @@
         "@storybook/react": "^7.0.12",
         "@storybook/react-webpack5": "^7.0.12",
         "@storybook/testing-react": "^2.0.0",
-        "@types/jest": "^26.0.19",
+        "@types/jest": "^29.5.2",
         "@types/react": "^17.0.20",
-        "@typescript-eslint/eslint-plugin": "^5.9.1",
-        "@typescript-eslint/parser": "^5.9.1",
+        "@typescript-eslint/eslint-plugin": "^5.59.7",
+        "@typescript-eslint/parser": "^5.59.7",
```
slax57, line 40: Are you still using them?

slax57 (commented):
I still have this warning when running `make lint`:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.1.0

YOUR TYPESCRIPT VERSION: 5.1.3

Please only submit bug reports when using the officially supported version.

=============
```

Do you think we could try to get rid of it as well?

.eslintrc
```diff
@@ -1,5 +1,9 @@
+
 {
     "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "warnOnUnsupportedTypeScriptVersion": false
```
djhi, line 5: We use version `5.1.3` when latest supported is `5.1.0`. It will be supported soon but we're safe anyway

package.json
```diff
@@ -75,6 +78,9 @@
         }
     },
     "resolutions": {
+        "@typescript-eslint/typescript-estree": "^5.60.0",
+        "@typescript-eslint/parser": "^5.60.0",
+        "@typescript-eslint/eslint-plugin": "^5.60.0",
```
djhi, line 83: This is required because of `eslint-config-react-app` which has not been updated in 2 years. We would probably be better of with our own config

slax57 (approved):

PR #8939, opened 2023-05-23
Fix ra-language-french package.json to avoid including tsconfig

djhi:
Some tools might use it even when not needed. Besides, that's the only misconfigured package

### DIFF
--- a/packages/ra-language-french/package.json
+++ b/packages/ra-language-french/package.json
@@ -6,6 +6,11 @@
         "type": "git",
         "url": "git+https://github.com/marmelab/react-admin.git"
     },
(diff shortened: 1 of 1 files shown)


slax57 (approved):
Could you please put in the PR description the typical error message we get? So that people having this issue can find it more easily.

joanllodrariera:
@slax57 This is the error: [commonjs--resolver] Failed to resolve entry for package "ra-language-french". The package may have incorrect main/module/exports specified in its package.json.

joanllodrariera:
Thanks for the fix @djhi

PR #8945, opened 2023-05-24
Fix show custom empty component on <ListView>

yanchesky:
The logic that checks whether the custom empty component should be shown relies on the `total` value. Since react-admin supports partial pagination, it should handle that case too. This PR fixes that.

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -41,7 +41,6 @@ export const ListView = <RecordType extends RaRecord = any>(
         defaultTitle,
         data,
         error,
-        total,
         isLoading,
         filterValues,
         resource,
@@ -83,7 +82,7 @@ export const ListView = <RecordType extends RaRecord = any>(
 
     const shouldRenderEmptyPage =
         !isLoading &&
-        total === 0 &&
+        data?.length === 0 &&
         !Object.keys(filterValues).length &&
         empty !== false;
 
@@ -144,7 +143,6 @@ ListView.propTypes = {
     setSort: PropTypes.func,
     showFilter: PropTypes.func,
     title: TitlePropType,
-    total: PropTypes.number,
 };
 
 export interface ListViewProps {
--- a/packages/ra-ui-materialui/src/list/List.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/List.spec.tsx
@@ -128,6 +128,33 @@ describe('<List />', () => {
         });
     });
 
+    it('should render custom empty component when data is empty', async () => {
+        const Dummy = () => null;
+        const CustomEmpty = () => <div>Custom Empty</div>;
+
+        const dataProvider = {
+            getList: jest.fn(() =>
+                Promise.resolve({
+                    data: [],
+                    pageInfo: { hasNextPage: false, hasPreviousPage: false },
+                })
+            ),
+        } as any;
+        render(
+            <CoreAdminContext dataProvider={dataProvider}>
+                <ThemeProvider theme={theme}>
+                    <List resource="posts" empty={<CustomEmpty />}>
+                        <Dummy />
+                    </List>
+                </ThemeProvider>
+            </CoreAdminContext>
+        );
+        await waitFor(() => {
+            expect(screen.queryByText('resources.posts.empty')).toBeNull();
+            screen.getByText('Custom Empty');
+        });
+    });
+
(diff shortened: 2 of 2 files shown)


packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -83,7 +82,7 @@ export const ListView = <RecordType extends RaRecord = any>(
 
     const shouldRenderEmptyPage =
         !isLoading &&
-        total === 0 &&
+        data?.length === 0 &&
```
slax57, line 85:
```suggestion
        (!data || data.length === 0) &&
```

packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -144,7 +143,6 @@ ListView.propTypes = {
     setSort: PropTypes.func,
     showFilter: PropTypes.func,
     title: TitlePropType,
-    total: PropTypes.number,
```
slax57, line 147: Nice catch!

Can you also remove the other propTypes that shouldn't be there?

Thanks

slax57 (changes requested):
Nice catch, thanks!

yanchesky:
Removed propTypes as you asked. I'm not sure though if I did not remove too many of them. Some of them are passed to the `useListContext`. I left only those which are directly consumed by <ListView>

Also, I fixed the condition you proposed as it broke the test. Btw what's wrong with using optional chaining?

slax57:
Thanks for cleaning out the propTypes :slightly_smiling_face: 

Regarding the condition, it was not about the conditional chaining which is perfectly fine to use. 
It's just that I would expect the empty page to be displayed also in case `data = undefined`.

With the original condition, when `data = undefined`, `data?.length === 0` will return `false`, and so the empty page will not be displayed, which is not what I'd expect.

Hence I suggested `(!data || data.length === 0)`, which will return `true` in this case.

Note that your latest version `!!data && data.length === 0` also won't work as it returns `false`.

Can you fix it?

yanchesky:
In that case, your suggestion indeed makes sense. However, as I said, returning a custom component in case `data = undefined` or `null` breaks the test. Here is [failed test run](https://github.com/marmelab/react-admin/actions/runs/5092017933) and below is the implementation of the test that fails. We can see that the returned response contains only `error` and the assertion is to render the error component.
https://github.com/marmelab/react-admin/blob/83aa21459faa6692c2aec9feee612380448f98fb/packages/ra-ui-materialui/src/list/List.spec.tsx#L228-L248
The error component is rendered by the `renderList` function rather than `renderEmpty`, thus we intend `shouldRenderEmptyPage` to be false when `data = undefined`. 
https://github.com/marmelab/react-admin/blob/83aa21459faa6692c2aec9feee612380448f98fb/packages/ra-ui-materialui/src/list/ListView.tsx#L97
https://github.com/marmelab/react-admin/blob/83aa21459faa6692c2aec9feee612380448f98fb/packages/ra-ui-materialui/src/list/ListView.tsx#L54-L79

IMO this is the correct approach. The case when `isLoading = false` and `data = undefined` seems like an error to me. Also, dataProvider expects `data` to be somehow defined, which `validateResponseFormat.ts` confirms. I would stay then with my first solution as it did the job. Any other solution would require more refactoring (Which I'm willing to do if you would like so 😊).  

Also speaking of `validateResponseFormat.ts` a while ago I noticed that the error produced by this function is not accurate. This may be misleading for new users, because the actual response may be `{ data: [...], total: 123 }` or `{ data: [...], pageInfo: {...} } `. Would you like me to fix this in this PR, create a new one or leave it as is?
https://github.com/marmelab/react-admin/blob/83aa21459faa6692c2aec9feee612380448f98fb/packages/ra-core/src/dataProvider/validateResponseFormat.ts#L52-L61

slax57:
Okay. I think you're right. Thanks for the detailed answer :slightly_smiling_face: 
We should keep your initial implementation then. I'll revert my suggested change.

Regarding the issue with `validateResponseFormat.ts`, this is a good point.
This would deserve its own PR since it's not linked to this fix. Can you open a new one?
Thank you so much :slightly_smiling_face:

slax57 (approved):

yanchesky:
> Regarding the issue with `validateResponseFormat.ts`, this is a good point. This would deserve its own PR since it's not linked to this fix. Can you open a new one? Thank you so much 🙂

Sure. I'll do that today evening.

PR #8947, opened 2023-05-25
Missing css class in ListView, cannot style List actions

mathzol:
As mentioned by the documentation:

https://marmelab.com/react-admin/List.html#sx-css-api

there have to be a class for actions: _& .RaList-actions_ 
but ListToolbar component never got it. 

I don't know this was a mistake or deliberate.

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListView.tsx
+++ b/packages/ra-ui-materialui/src/list/ListView.tsx
@@ -55,6 +55,7 @@ export const ListView = <RecordType extends RaRecord = any>(
         <div className={ListClasses.main}>
             {(filters || actions) && (
                 <ListToolbar
+                    className={ListClasses.actions}         
                     filters={filters}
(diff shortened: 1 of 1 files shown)


djhi (approved):
Good catch! Thanks :)

djhi (changes requested):
Can you fix the linter error?

mathzol:
@djhi Ofc, sorry for that, I used webeditor for quick

djhi (approved):
:+1:

mathzol:
@fzaninotto, those styles defined by others (RA developers), but didint leave any comments about to "do not use them". Why did RA developers make a documentations about something which is not finished properly in the code? I trusted in the code and thought it was only a missed line in the code. Please use comments in code to avoid much amount of useless debugging time.

djhi:
Superseded by #8995

PR #8950, opened 2023-05-26
Added filterDefaultValues to context to enable detecting change from defaults

timbrandin:


slax57 (changes requested):
Can you provide more information about your use-case, and ideally provide an example leveraging this feature?

Also, since this technically is a new feature, your PR should target the `next` branch instead of `master`.

Lastly, you need to update the `ListControllerResult` type, and probably also the `ListContext` default values in `ListContext.tsx`.

fzaninotto:
Closing due to inactivity.

PR #8955, opened 2023-05-29
Add ability to refetch permissions on demand

mmateja:


docs/usePermissions.md
```diff
@@ -61,6 +61,32 @@ const MyPage = () => {
 }
 ```
 
+## Refreshing permissions
+
+Permissions are loaded on page load and cached. If your application requires permissions to be refreshed, for example after a change modifying user permissions, you can use `refetch` function to trigger reload.
```
slax57, line 66:
```suggestion
Permissions are loaded when the app loads and then cached. If your application requires permissions to be refreshed, for example after a change modifying user permissions, you can use `refetch` function to trigger reload.
```

docs/usePermissions.md
```diff
@@ -61,6 +61,32 @@ const MyPage = () => {
...
+    const record = useRecordContext();
+    const [ update ] = useUpdate();
+    const { refetch } = usePermissions();
+
+    const handleClick = () => {
+        update(
+            "users",
+            { id: record.id, data: { admin: true }, previousData: record },
+        );
+        refetch();
```
slax57, line 79:
```suggestion
        update(
            "users",
            { id: record.id, data: { admin: true }, previousData: record },
            { onSuccess: refetch },
        );
```

docs/usePermissions.md
```diff
@@ -61,6 +61,32 @@ const MyPage = () => {
...
+    }
+
+    return (
+        <Button onClick={handleClick}>
+            Make user an admin
+        </Button>
+    )
+}
+```
```
slax57, line 88: Also, this whole code block should be wrapped with `{% raw %}` `{% endraw %}` to avoid issues with jekyll. Can you wrap it?

slax57 (changes requested):

PR #8956, opened 2023-05-30
Fix `<AutocompleteInput>` should not use `matchSuggestion` when in a `<ReferenceInput>`

slax57:
**Problem**

When provided with a `matchSuggestion` and used in a `<ReferenceInput>`, `<AutocompleteInput>` does actually do the filtering twice: once immediately client-side, and then once server-side.
The client-side filtering does not make much sense because it's done only on a subset of the data, and besides it leads to UI flickering when the server-side results arrive.

**Solution**

Actually this is a bug, we shouldn't call `getSuggestions()` when used inside a reference, at all.
We already do this to disable MUI's Autocomplete default filtering fn, so I just extended this behavior to `getSuggestions()`.

I also had to adapt the error message warning the user that they can't use `optionText` without `matchSuggestion`, to only trigger when **not** used in a reference. Indeed, in a reference, `matchSuggestion` should never be called, as it is used for client-side filtering only.

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -17,6 +17,7 @@ import {
     DifferentShapeInGetMany,
     InsideReferenceInput,
     InsideReferenceInputDefaultValue,
+    InsideReferenceInputWithCustomizedItemRendering,
     Nullable,
     NullishValuesSupport,
     VeryLargeOptionsNumber,
@@ -1392,6 +1393,39 @@ describe('<AutocompleteInput />', () => {
             }
             expect(testFailed).toBe(false);
             expect(input.value).toBe('Leo Tolstoy test');
+        });
+
+        it('should not use getSuggestions to do client-side filtering', async () => {
+            // filtering should be done server-side only, and hence matchSuggestion should never be called
+            const matchSuggestion = jest.fn().mockReturnValue(true);
+            render(
+                <InsideReferenceInputWithCustomizedItemRendering
+                    matchSuggestion={matchSuggestion}
+                />
+            );
+            await waitFor(
+                () => {
+                    expect(
+                        (screen.getByRole('textbox') as HTMLInputElement).value
+                    ).toBe('Leo Tolstoy - Russian');
+                },
+                { timeout: 2000 }
+            );
+            screen.getByRole('textbox').focus();
+            fireEvent.click(screen.getByLabelText('Clear value'));
+            await waitFor(() => {
+                expect(screen.getByRole('listbox').children).toHaveLength(5);
+            });
+            fireEvent.change(screen.getByRole('textbox'), {
+                target: { value: 'French' },
+            });
+            await waitFor(
+                () => {
+                    screen.getByText('No options');
+                },
+                { timeout: 2000 }
+            );
+            expect(matchSuggestion).not.toHaveBeenCalled();
         });
     });
 
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
@@ -24,7 +24,7 @@ import fakeRestProvider from 'ra-data-fakerest';
 
 import { Edit } from '../detail';
 import { SimpleForm } from '../form';
-import { AutocompleteInput } from './AutocompleteInput';
+import { AutocompleteInput, AutocompleteInputProps } from './AutocompleteInput';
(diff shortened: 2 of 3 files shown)


fzaninotto (approved):

moulie415:
This was something I was utilising, I'm using firestore as my db and fetching the data then filtering client side and it was very useful, is there some other way I should be achieving this using the reference input?

moulie415:
@slax57 here's my code if it helps to know what I was trying to achieve

```js
import React from 'react';
import {AutocompleteInput} from 'react-admin';

const MyAutoCompleteInput = props => {
  return (
    <AutocompleteInput
      {...props}
      filterToQuery={() => ''}
      matchSuggestion={(filter, suggestion) => {
        return (
          suggestion[props.optionText].toLowerCase().indexOf(filter) !== -1
        );
      }}
    />
  );
};

export default MyAutoCompleteInput;
```

slax57:
@moulie415 Yes, if you'd like to enable client-side filtering, you need to provide `<AutocompleteInput>` directly with a list of [`choices`](https://marmelab.com/react-admin/AutocompleteInput.html#choices). You can fetch these choices from your API, e.g. by using [`useGetList`](https://marmelab.com/react-admin/useGetList.html).

wcmatheusoliveira:
Hey team, I've a lot of ReferenceInput across my application using  the <AutocompleteInput /> component. Every single autocomplete stop work except for the ones that uses data={CHOICES}. I notice that using the package `ra-ui-materialui: 4.9.2` it works but I got some types errors. We've a medium/large application today, move it to use "choices" instead of ReferenceInput would cost us a lot. We're a enterprise customer, let me know what is the best solution to take from here. 

This is our dependecies list:

"dependencies": {
    "@emotion/react": "11.10.6",
    "@emotion/styled": "11.10.6",
    "@material-ui/core": "4.12.4",
    "@material-ui/icons": "4.11.3",
    "@mui/lab": "5.0.0-alpha.152",
    "@mui/material": "5.12.0",
    "@mui/styles": "5.14.17",
    "@mui/x-data-grid": "6.9.2",
    "@mui/x-data-grid-generator": "6.11.1",
    "@react-admin/ra-calendar": "4.0.1",
    "@react-admin/ra-editable-datagrid": "4.2.1",
    "@react-admin/ra-enterprise": "6.0.4",
    "@react-admin/ra-form-layout": "4.6.2",
    "@react-admin/ra-navigation": "4.4.2",
    "@react-admin/ra-relationships": "4.2.2",
    "@react-admin/ra-tree": "4.2.0",
    "@react-keycloak/web": "3.4.0",
    "@testing-library/jest-dom": "5.16.5",
    "@testing-library/react": "13.4.0",
    "@testing-library/user-event": "14.4.3",
    "classnames": "2.3.2",
    "dayjs": "1.11.9",
    "graphql": "16.6.0",
    "graphql-tag": "2.12.6",
    "jwt-decode": "3.1.2",
    "keycloak-js": "18.0.1",
    "leaflet": "1.9.3",
    "lodash": "4.17.21",
    "mui-tel-input": "3.2.2",
    "papaparse": "5.4.1",
    "pluralize": "8.0.0",
    "powerbi-client": "2.22.2",
    "powerbi-client-react": "1.4.0",
    "prop-types": "15.8.1",
    "ra-data-graphql-simple": "4.8.4",
    "ra-data-json-server": "4.9.2",
    "react": "18.2.0",
    "react-admin": "4.15.5",
    "react-dom": "18.2.0",
    "react-geosuggest": "2.14.1",
    "react-image-file-resizer": "0.4.8",
    "react-leaflet": "4.2.1",
    "react-medium-image-zoom": "5.1.5",
    "react-scripts": "5.0.1",
    "react-sortable-tree": "2.8.0",
    "serve": "14.2.0",
    "typescript": "4.9.5",
    "use-debounce": "9.0.4",
    "web-vitals": "2.1.4"
  },
   "resolutions": {
    "react-hook-form": "7.40.0",
    "@types/react": "18.0.34",
    "@types/react-dom": "18.0.0",
    "@types/react-geosuggest": "2.7.13"
  }

moulie415:
@wcmatheusoliveira it would probably be more helpful to see a code sample of how you're using ReferenceInput with AutoComplete input

wcmatheusoliveira:
```
<ReferenceInput
      alwaysOn
      label="Customer"
      source="customerId"
      reference="customers"
    >
      <AutocompleteInput  label="Customer" />
</ReferenceInput>
```

slax57:
@wcmatheusoliveira Since you are an Enterprise customer you have access to our dedicated [support desk](https://marmelab.com/ra-enterprise#support).
I believe you have already opened a ticket there (which we answered), so feel free to continue the discussion there if you need further help, we'll do our best to assist you.

PR #8957, opened 2023-05-30
Fix `getList` and `getManyReference` error to take partial pagination into account.

yanchesky:
The error displayed when `total` and `pageInfo` properties are missing in the response does not accurately say correct solution. This PR fixes that


fzaninotto (approved):

fzaninotto:
Thanks!

PR #8963, opened 2023-05-31
Remove redundant null checks from FunctionField render type

elstgav:
When using Typescript, the `FunctionField` render prop currently requires redundant null checks, as it claims `record` can be `undefined`. However it’s guaranteed not to, thanks to the null check on line 30:

https://github.com/marmelab/react-admin/blob/bbec8d9d5dd1b6b5045679f3c1494499d8f12260/packages/ra-ui-materialui/src/field/FunctionField.tsx#L30-L39

This same guarantee applies to `<WithRecord>`, which is documented here:

https://github.com/marmelab/react-admin/blob/964310ee2992724892be909eaef43ca17a3441c8/docs/WithRecord.md#L26

This PR updates the `render` prop type to clarify that `record` will be defined, eliminating the need for redundant null checks:

```diff
  <FunctionField
    source='user'
-   render={(record) => record && record.first_name}
+   render={(record) => record.first_name}
  />
```

The PR also:

- creates a `FunctionFieldRenderer` method, that can be used when creating custom render methods
- Clarifies that the `render` method must return a valid `ReactNode`, not `any`.


examples/crm/src/dashboard/LatestNotes.tsx
```diff
@@ -133,9 +133,7 @@ const Contact = ({ note }: any) => (
         >
             <FunctionField<ContactType>
                 variant="body2"
-                render={(contact?: ContactType) =>
```
elstgav, line 136: `ContactType` is already set by line 134

packages/ra-ui-materialui/src/field/FunctionField.tsx
```diff
@@ -48,9 +48,13 @@ FunctionField.propTypes = {
     render: PropTypes.func.isRequired,
 };
 
+export type FunctionFieldRenderer<
+    RecordType extends Record<string, unknown> = Record<string, any>
+> = (record: RecordType, source?: string) => React.ReactNode;
```
elstgav, line 53: I was unsure on what to name this. It may be helpful to further abstract this into a type that’s shared by `<WithRecord>`

packages/ra-core/src/controller/record/WithRecord.tsx
```diff
@@ -14,14 +13,16 @@ import { useRecordContext } from './useRecordContext';
...
 }: WithRecordProps<RecordType>) => {
     const record = useRecordContext<RecordType>();
     return record ? render(record) : null;
 };
 
-export interface WithRecordProps<RecordType extends RaRecord> {
-    render: (record: RecordType) => ReactElement;
+export interface WithRecordProps<
+    RecordType extends Record<string, unknown> = any
+> {
```
elstgav, line 25: Updated these to match #8964

packages/ra-core/src/types.ts
```diff
@@ -361,6 +361,10 @@ export interface ResourceProps {
     children?: ReactNode;
 }
 
+export type RenderRecordFunction<
+    RecordType extends Record<string, unknown> = any
+> = (record: RecordType, source?: string) => ReactNode;
+
```
elstgav, line 367: Let me know if there’s a more appropriate place to put this. Named to match the existing `RenderResourcesFunction`:

https://github.com/marmelab/react-admin/blob/b007f35f98a4046a146cd51799f2829b9020cb89/packages/ra-core/src/types.ts#L303

packages/ra-core/src/controller/record/WithRecord.tsx
elstgav: @fzaninotto @slax57 it might be helpful to create a `StrictRaRecord` type that we can share/extend from?

e.g.

```tsx
// ra-core/src/types.ts

export interface StrictRaRecord<IdentifierType extends Identifier = Identifier> {
    id: IdentifierType;
}

export interface RaRecord<IdentifierType extends Identifier = Identifier> extends StrictRaRecord {
    [key: string]: any;
}

// in other files…

export interface WithRecordProps<RecordType extends StrictRaRecord = RaRecord> { … }
export interface FunctionFieldProps<RecordType extends StrictRaRecord = RaRecord> { … }

elstgav:
@fzaninotto rebased, and created a shared `RenderRecordFunction` type that is shared with the `WithRecord` component.

packages/ra-core/src/controller/record/WithRecord.tsx
fzaninotto: I'm not sure of the added value (extending a StrictRaRecord vs adding an Identifier by hand is about the same). Besides, it's a more general reflexion about struct null checks that we're working on in the long term.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #8965, opened 2023-06-01
[Doc] Add Next.js app router install instructions

fzaninotto:
## Problem

The Next.js tutorial only cover the Pages router. Next.js is putting the App router front and foremost. 

## Solution

Document the App router integration, too.


packages/ra-core/src/core/CoreAdminRoutes.tsx
```diff
@@ -119,10 +119,6 @@ export const CoreAdminRoutes = (props: CoreAdminRoutesProps) => {
     );
 };
 
-CoreAdminRoutes.defaultProps = {
```
fzaninotto, line 122: This one was logging a warning AND was unused

slax57 (approved):

PR #8971, opened 2023-06-05
fix(array-input): update form state w/ default values

henryhobhouse:
## Problem

Currently the Array Input tries to update individual child values using react-hook-forms `resetField` function from their `useFormContext` hook. The issue is this bypasses react-hook-forms own control functions for updating array fields that we use to instantiate the fields initially. Although this appears to work fine the issue is that it seems to stop the form control checking if the fields are updated and as such does not make the form touched or dirty on update and subsequently the submit button remains disabled.

**Example** (bottom input is nested in array input):
 
Screen Recording 2023-06-05 at 11 27 45 32 AM

## Solution

Update the Array Input to use the return control APIs from `useFieldArray` (https://www.react-hook-form.com/api/usefieldarray/) to replace the field values and then reset the form with the same values to remove context that the fields have been updated. 

In simple terms it looks like: 
```
// BEFORE
resetField(source, { defaultValue });

// AFTER
fieldArrayInputControl.replace(defaultValue);
reset({}, { keepValues: true });
```


packages/ra-core/src/form/useApplyArrayInputDefaultValues.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type Props = Partial<InputProps> & {
+    source: string;
+};
+
+/*
+ * This hook updates the array input with the default value if default value is present
+ * and field input is not already populated or dirty
+ */
+export const useApplyArrayInputDefaultValues = (
```
henryhobhouse, line 19: I felt there is sufficient separation of logic for array inputs to use an new hook over further branching within the existing `useApplyInputDefaultValues` hook.  Happy to refactor into one and use additional props to branch if preferred.

packages/ra-core/src/form/useApplyArrayInputDefaultValues.ts
```diff
@@ -0,0 +1,65 @@
...
+            .split('.')
+            .some(pathPart => numericRegex.test(pathPart));
+        if (pathContainsIndex) {
+            const parentPath = source.split('.').slice(0, -1).join('.');
+            const parentValue = get(getValues(), parentPath);
+            if (parentValue == null) {
+                // the parent is undefined, so we don't want to set the default value
+                return;
+            }
+        }
```
henryhobhouse, line 53: Brought across from the `useApplyInputDefaultValues` hook.

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -24,26 +24,15 @@ export const useApplyInputDefaultValues = (
...
-                .split('.')
-                .some(pathPart => numericRegex.test(pathPart));
-            if (pathContainsIndex) {
-                const parentPath = source.split('.').slice(0, -1).join('.');
-                const parentValue = get(getValues(), parentPath);
-                if (parentValue == null) {
-                    // the parent is undefined, so we don't want to set the default value
-                    return;
-                }
-            }
```
henryhobhouse, line 43: Moved to dedicated hook for array input

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
```diff
@@ -264,6 +264,53 @@ describe('<ArrayInput />', () => {
...
 
+    it('should update the form state to dirty, and allow submit, on updating an array input with default value', async () => {
+        const TestArrayInputs = () => {
+            const { reset } = useFormContext();
+
+            // mimic's the scaffolding elsewhere in the app but does not disable the form
+            // without this (before array input side effects)
+            React.useEffect(() => {
+                reset();
+            }, [reset]);
```
henryhobhouse, line 275: Some help here would be useful. The only way I can scaffold the test so the form instantiates like it does in the DOM. Without this the form isn't disabled on first load even though the logic is not changed in events after this. I presume there is some composition that normally does something similar even if it via breaking memorisation somewhere? 🤷

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
```diff
@@ -264,6 +264,53 @@ describe('<ArrayInput />', () => {
...
+                </SimpleForm>
+            </AdminContext>
+        );
+        const submitButton = screen
+            .getByLabelText('ra.action.save')
+            .closest('button');
+
+        await waitFor(() => {
+            expect(submitButton?.disabled).toBe(true);
+        });
```
henryhobhouse, line 303: Another weirdness is that getting the submit button using the preferred (as recommended by the react testing team) `getByRole` followed by `expect(...button).isDisabled()` does not work even when the logged out element does show the disabled property as truthy?

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
```diff
@@ -264,6 +264,53 @@ describe('<ArrayInput />', () => {
         expect(screen.queryByText('test helper text')).not.toBeNull();
     });
 
+    it('should update the form state to dirty, and allow submit, on updating an array input with default value', async () => {
```
henryhobhouse, line 267: Although the test is not as clean as I would like it does protect against the bug identified successfully.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
slax57: I don't know why this is not working, but your solution is good enough IMO

slax57 (commented):
I'm really impressed by how deep you managed to dive into react-admin to come up with this fix, well done and thank you so much.

You may want to remove the manual call to `reset()` in your test and replace it with a `RecordContextProvider` as I suggested, but other than that, I really could find no issues with this change (although I tried hard, since this is a sensitive part of the framework :grin: )

So yeah, thanks, and well done! :clap: 

@fzaninotto if you have time for a second review I would appreciate it, since we can't afford any regressions on such a key component. As I said I could not find any, but another pair of eyes would be welcome to be extra cautious :slightly_smiling_face:

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
henryhobhouse: Refactored and works as you suggested.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -71,9 +71,7 @@ import { ArrayInputContext } from './ArrayInputContext';
 export const ArrayInput = (props: ArrayInputProps) => {
     const {
         className,
-        defaultValue,
         label,
-        isFetching,
```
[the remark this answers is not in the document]
henryhobhouse: Sorry I should have commented. The prop wasn't used. However just realised that by destructuring it will mean its inadvertingly passed into any component that uses the "rest" of the props. Will revert.

slax57 (changes requested):
Thanks for taking my review into account :slightly_smiling_face: 

Actually, after more thinking, we figured it might be best to actually keep the array logic in `useApplyInputDefaultValues.ts` instead of moving it in a dedicated hook, not because we are not happy with the separation of logic (it's the opposite actually), but because it might create regressions in custom input components.

Indeed the `useApplyInputDefaultValues` is used by `useInput`, so other custom inputs may rely on it to also support array inputs.

Hence, could you please move the logic back inside `useApplyInputDefaultValues.ts`?
Thank you so much.

slax57 (approved):
Thanks! :slightly_smiling_face: 

> Assume check for index and if parent prefixing that in path should happen for all to your point on regressions?

It was the case before, so we should keep it that way

> Are you happy to declarative on both stating it's for an array and seperately requiring the array input controls. I felt it would be a more readable solution than assume if has array input control it therefore must be for an array input?

Your solution looks good to me :slightly_smiling_face: 

Again, I'll request one more approval from @fzaninotto and we should be good to go.

djhi (approved):

slax57:
Alright we have @fzaninotto 's blessing, let's merge! :grin: 

Thanks again @henryhobhouse for your contribution :slightly_smiling_face:

PR #8974, opened 2023-06-05
[Doc] Add documentation for `<TreeInput>` and `<ReferenceNodeInput>`

fzaninotto:
Capture d’écran 2023-06-05 à 14 03 35

```tsx
import { Edit, SimpleForm, TextInput } from 'react-admin';
import { TreeInput } from '@react-admin/ra-tree';

export const ProductEdit = () => (
    <Edit>
        <SimpleForm>
            <TextInput source="id" disabled />
            <TextInput source="name" />
            <TreeInput source="category" treeData={[
                { id: 1, title: 'Clothing', isRoot: true, children: [2, 6] },
                { id: 2, title: 'Men', children: [3] },
                { id: 3, title: 'Suits', children: [4, 5] },
                { id: 4, title: 'Slacks', children: [] },
                { id: 5, title: 'Jackets', children: [] },
                { id: 6, title: 'Women', children: [7, 10, 11] },
                { id: 7, title: 'Dresses', children: [8, 9] },
                { id: 8, title: 'Evening Gowns', children: [] },
                { id: 9, title: 'Sun Dresses', children: [] },
                { id: 10, title: 'Skirts', children: [] },
                { id: 11, title: 'Blouses', children: [] },
            ]} />
        </SimpleForm>
    </Edit>
);
```

- [x] Add TreeInput Chapter
- [x] Add ReferenceNodeInput Chapter
- [x] Mention new components in Reference and sidebar navigation
- [x] Mention new components in related chapters


docs/ReferenceInput.md
```diff
@@ -341,6 +341,37 @@ const filterToQuery = searchText => ({ name_ilike: `%${searchText}%` });
 </ReferenceInput>
 ```
 
+## Tree Structure
+
+If the reference resource is a tree, use `<ReferenceNodeInput>` instead of `<ReferenceInput>`.
```
slax57, line 346:
```suggestion
If the reference resource is a tree, use [`<ReferenceNodeInput>`](./ReferenceNodeInput.md) instead of `<ReferenceInput>`.
```

docs/TreeInput.md
```diff
@@ -0,0 +1,189 @@
...
+};
+```
+
+The `isLoading` prop is used to display a loading indicator while the data is being fetched.
+
+However, most of the time, if you need to populate a `<TreeInput>` with choices fetched from another resource, it's because you are trying to set a foreign key. In that case, you should use [`<ReferenceNodeInput>`](./ReferenceNodeInput.md) to fetch the choices instead (see next section). 
+
+## Selecting a Foreign Key
+
+If you use `<TreeInput>` to set a foreign key for a many-to-one or a one-to-one relationship, you’ll have to [fetch choices](#fetching-choices), as explained in the previous section. You’ll also have to fetch the record corresponding to the current value of the foreign key, as it may not be in the list of choices.
```
slax57, line 170: > You’ll also have to fetch the record corresponding to the current value of the foreign key, as it may not be in the list of choices.

This may be misleading as this is not implemented in `ReferenceNodeInput` yet

slax57 (changes requested):

slax57 (approved):

PR #8984, opened 2023-06-07
[Doc] Fix links to react-hook-form's doc

slax57:
Dunno if it's temporary (and intentional) or not, but the RHF doc is now under www.react-hook-form.com instead of react-hook-form.com, so this PR changes the links accordingly

EDIT

According to https://github.com/react-hook-form/documentation/issues/971, the path has to be changed to `/docs`, so this is what this PR does.


slax57:
DNM until https://github.com/react-hook-form/react-hook-form/issues/10503 is answered to

slax57:
Back to RFR

fzaninotto (approved):

PR #8986, opened 2023-06-07
Add rowSx

mshd:
Follow up from #8925, due to inactivity. I kept rowStyle as requested by @djhi

### DIFF
--- a/docs/SimpleList.md
+++ b/docs/SimpleList.md
@@ -48,6 +48,7 @@ It accepts the following props:
 * [`rightAvatar`](#rightavatar)
 * [`rightIcon`](#righticon)
 * [`rowStyle`](#rowstyle)
+* [`rowSx`](#rowsx)
 * [`empty`](#empty)
 
 ## `leftAvatar`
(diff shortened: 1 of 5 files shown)


docs/SimpleList.md
```diff
@@ -176,6 +177,22 @@ export const PostList = () => (
 );
 ```
 
+## `rowSx`
+
+This optional prop should be a function, which gets called for each row. It receives the current record and index as arguments, and should return a style object. The style object is applied to the `<ListItem>` styles prop.
```
djhi, line 182:
```suggestion
This optional prop should be a function, which gets called for each row. It receives the current record and index as arguments, and should return an object accepted by Material UI [`sx` prop](https://mui.com/system/getting-started/the-sx-prop/). The object is applied to the `<ListItem>` `sx` prop.
```

djhi (changes requested):
Thanks! As this is a new feature, can you target the `next` branch?

mshd:
> Thanks! As this is a new feature, can you target the `next` branch?

Fixed

djhi:
Unfortunately, it seems you'll to rebase your branch on `next`

PR #8995, opened 2023-06-12
Fix ListView does not apply CSS classes to its child components

djhi:
Supersedes #8947


slax57 (approved):

fzaninotto:
FTY, the ghost `actions` class was the result of a refactoring achieved in version 2.3 (see #2175). It's safe to assume that the class was never in use in version 3, so this shouldn't have any impact on existing apps.

PR #8999, opened 2023-06-12
Add unique validator

djhi:
Validating the uniqueness of a field is a common requirement so we'd like to provide the `useUnique` hook that returns a validator for this use case.

It will call the `dataProvider.getList` method with a filter to check whether a record exists with the current value of the input for the field matching the input source.

Basic usage (more options in the documentation):

```js
const UserCreateForm = () => {
    const unique = useUnique();
    return (
        <SimpleForm>
            <TextInput source="username" validate={unique()} />
        </SimpleForm>
    );
};
```


docs/Validation.md
```diff
@@ -259,6 +259,85 @@ export const ProductEdit = () => (
 
 **Tip**: The custom validator function can return a promise, e.g. to use server-side validation. See next section for details.
 
+## Per Input Validation: useUnique
```
fzaninotto, line 262: I think this hook deserves its own page (and link in the Reference documentation). It's a good idea to link to it from the Validation doc, though (in [the list of built-in validators](https://marmelab.com/react-admin/Validation.html#per-input-validation-built-in-field-validators))

docs/Validation.md
```diff
@@ -259,6 +259,85 @@ export const ProductEdit = () => (
...
+        </SimpleForm>
+    );
+};
+```
+
+### Options
+
+#### `message`
+
+A custom message to display when the validation fails. Defaults to `ra.validation.unique`.
```
fzaninotto, line 283: I'd document the default English message first, the the message id.

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+import { useResourceContext } from '../core';
+import { useDataProvider } from '../dataProvider';
+import { useTranslate, useTranslateLabel } from '../i18n';
+import { InputProps } from './useInput';
+
+/**
+ * A hook that returns a validation function checking for a record field uniqueness
+ * by calling the dataProvider getList function with a filter.
+ *
+ * @example Passing options at declaration time
```
fzaninotto, line 11:
```suggestion
 * @example // Passing options at declaration time
```
Otherwise the syntax highlighting in the IDE fails. 

Same for the other examples

docs/Validation.md
```diff
@@ -259,6 +259,85 @@ export const ProductEdit = () => (
...
+Validating the uniqueness of a field is a common requirement so React-admin provides the `useUnique` hook that returns a validator for this use case.
+
+It will call the `dataProvider.getList` method with a filter to check whether a record exists with the current value of the input for the field matching the input source.
+
+```js
+const UserCreateForm = () => {
+    const unique = useUnique();
+    return (
+        <SimpleForm>
+            <TextInput source="username" validate={unique()} />
```
fzaninotto, line 273: What if the source is a path (e.g. `user.name`)? Does this work? In any case, you should document it.

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+        try {
+            const { data } = await dataProvider.getList(resource, {
+                filter: { ...filter, [props.source]: value },
+                pagination: { page: 1, perPage: 1 },
+                sort: { field: props.source, order: 'ASC' },
+            });
+
+            if (data.length > 0) {
+                return translate(message, {
+                    _: message,
```
fzaninotto, line 90: The default message should be an English text

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+        );
+
+        try {
+            const { data } = await dataProvider.getList(resource, {
+                filter: { ...filter, [props.source]: value },
+                pagination: { page: 1, perPage: 1 },
+                sort: { field: props.source, order: 'ASC' },
+            });
+
+            if (data.length > 0) {
```
fzaninotto, line 88: why not use total instead? See the `Count` component for reference.

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+            {
+                filter: {},
+                message: 'ra.validation.unique',
+            },
+            options,
+            callTimeOptions
+        );
+
+        try {
+            const { data } = await dataProvider.getList(resource, {
```
fzaninotto, line 82: Open question: should we use react-query to do that?

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+                filter: { ...filter, [props.source]: value },
+                pagination: { page: 1, perPage: 1 },
+                sort: { field: props.source, order: 'ASC' },
```
fzaninotto, line 85: This custom sort is useless and may be slow. Why not use the default `id` sort?

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export const useUnique = (options?: UseUniqueOptions) => {
+    const dataProvider = useDataProvider();
+    const translateLabel = useTranslateLabel();
+    const resource = useResourceContext(
+        typeof options === 'string' ? undefined : options
+    );
+    const translate = useTranslate();
+
+    return (callTimeOptions?: UseUniqueOptions) => async (
```
fzaninotto, line 67: You should name that callback, for easier debugging

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,112 @@
...
+ *             <TextInput source="username" validate={unique({ message: 'Username is already used')} />
+ *         </SimpleForm>
+ *     );
+ * }
+ */
+export const useUnique = (options?: UseUniqueOptions) => {
+    const dataProvider = useDataProvider();
+    const translateLabel = useTranslateLabel();
+    const resource = useResourceContext(
+        typeof options === 'string' ? undefined : options
```
fzaninotto, line 63: ho can options be a string? The options Type says otherwise.

packages/ra-core/src/form/useUnique.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { fireEvent, render, screen } from '@testing-library/react';
+import {
+    Basic,
+    DataProviderErrorOnValidation,
+    WithAdditionalFilters,
+    WithCustomMessage,
+} from './useUnique.stories';
+
+describe('useUnique', () => {
+    it('should show the default error when a field is not unique', async () => {
```
fzaninotto, line 11:
```suggestion
    it('should show the default error when the field value already exists', async () => {
```

packages/ra-core/src/form/useUnique.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+        render(<Basic />);
+
+        await screen.findByDisplayValue('John Doe');
+
+        fireEvent.click(screen.getByText('Submit'));
+
+        await screen.findByText('Must be unique');
+    });
+
+    it('should not show the default error when a field is unique', async () => {
```
fzaninotto, line 21:
```suggestion
    it('should not show the default error when the field value does not exist', async () => {
```

packages/ra-core/src/form/useUnique.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+        render(<Basic />);
+
+        await screen.findByDisplayValue('John Doe');
+        fireEvent.change(screen.getByDisplayValue('John Doe'), {
+            target: { value: 'Jordan Doe' },
+        });
+
+        fireEvent.click(screen.getByText('Submit'));
+
+        expect(screen.queryByText('Must be unique')).toBeNull();
```
fzaninotto, line 32: Not sure this test fails for the right reasons, as the notifications are async

packages/ra-core/src/form/useUnique.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+                    message: 'Someone is already registered with this name',
+                })}
+            />
+
+            <button type="submit">Submit</button>
+        </Form>
+    );
+};
+
+export const WithCustomMessage = () => {
```
fzaninotto, line 143: I'd name this story after the prop, so `Message`. Same for other stories.

fzaninotto (changes requested):

packages/ra-core/src/form/useUnique.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+            dataProvider={dataProvider}
+            queryClient={new QueryClient()}
+            i18nProvider={i18nProvider}
+        >
+            <EditBase
+                id="1"
+                resource="users"
+                mutationOptions={{
+                    onSuccess: () => {
+                        alert('success');
```
fzaninotto, line 89: This is not testable

djhi: We don't care it is

packages/ra-core/src/form/useUnique.ts
djhi: forgot to remove that when I realized we didn't need to pass the `source`

packages/ra-core/src/form/useUnique.ts
djhi: No because it would prevent passing custom messages that are not translation keys

packages/ra-core/src/form/useUnique.spec.tsx
djhi: Done

fzaninotto:
You should update the doc sidebar to add an entry for the new `useUnique` hook.

packages/ra-core/src/form/useUnique.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+        // The dataProvider for this story fails one over two times
+        // Here's the first time, it should show an error
+        fireEvent.click(screen.getByText('Submit'));
+        expect(screen.queryByText('Server communication error')).toBeNull();
+
+        // Here's the second time, it should show the validation message
+        fireEvent.click(screen.getByText('Submit'));
+        expect(screen.queryByText('Must be unique')).toBeNull();
+    });
+});
```
fzaninotto, line 187: We only have unit tests - I miss an integration test. Could you maybe update the simple example to use the new validator?

packages/ra-core/src/form/useUnique.ts
```diff
@@ -0,0 +1,118 @@
...
+ *     );
+ * }
+ */
+export const useUnique = (options?: UseUniqueOptions) => {
+    const dataProvider = useDataProvider();
+    const translateLabel = useTranslateLabel();
+    const resource = useResourceContext(options);
+    const translate = useTranslate();
+
+    const validateUnique = useCallback(
```
fzaninotto, line 67: Don't we need a debounce? I'm afraid this will DDoS the data providers

fzaninotto (changes requested):

docs/useUnique.md
```diff
@@ -0,0 +1,85 @@
...
+            <TextInput source="username" validate={unique({ message: 'myapp.validation.unique' })} />
+        </SimpleForm>
+    );
+};
+```
+
+### `resource`
+
+The resource targeted by the `dataProvider.getList` call. Defaults to the resource from the nearest `ResourceContext`.
+
```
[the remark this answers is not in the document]
djhi: That's just what we do everywhere we need the resource?

packages/ra-core/src/form/useUnique.ts
djhi: Added

packages/ra-core/src/form/useUnique.spec.tsx
djhi: Added

slax57 (changes requested):

docs/useUnique.md
```diff
@@ -0,0 +1,85 @@
...
+# `useUnique`
+
+Validating the uniqueness of a field is a common requirement so React-admin provides the `useUnique` hook that returns a validator for this use case.
+
+It will call the `dataProvider.getList` method with a filter to check whether a record exists with the current value of the input for the field matching the input source.
+
+## Usage
+
+```js
+const UserCreateForm = () => {
```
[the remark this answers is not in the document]
djhi: Done

docs/useUnique.md
```diff
@@ -0,0 +1,85 @@
...
+### `message`
+
+A custom message to display when the validation fails. Defaults to `Must be unique` (translation key: `ra.validation.unique`).
+It accepts a translation key. The [`i18nProvider.translate` function] will be called with the following parameters:
+- `source`: the input name
+- `label`: the translated input label
+- `value`: the current input value
+
+```js
+const i18nProvider = polyglotI18nProvider(() =>
```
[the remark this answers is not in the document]
djhi: Done

packages/ra-core/src/util/asyncDebounce.ts
```diff
@@ -0,0 +1,33 @@
...
+    const debounced = debounce((args: Parameters<FunctionType>) => {
+        func(...args)
+            .then((...res) => {
+                resolveSet.forEach(resolve => resolve(...res));
+                resolveSet.clear();
+            })
+            .catch((...res) => {
+                rejectSet.forEach(reject => reject(...res));
+                rejectSet.clear();
+            });
```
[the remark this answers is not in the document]
djhi: good catch

docs/useUnique.md
```diff
@@ -0,0 +1,85 @@
...
+layout: default
+title: "useUnique"
+---
+
+# `useUnique`
+
+Validating the uniqueness of a field is a common requirement so React-admin provides the `useUnique` hook that returns a validator for this use case.
+
+It will call the `dataProvider.getList` method with a filter to check whether a record exists with the current value of the input for the field matching the input source.
+
```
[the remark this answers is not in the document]
slax57: still relevant IMO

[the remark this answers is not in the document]
slax57: EDIT my bad, didn't look at the latest version. Should be OK.

slax57 (approved):

fzaninotto (approved):

PR #9000, opened 2023-06-13
Doc: IDE error when copy paste snippet

adguernier:
Fix error message `JSX element 'Box' has no corresponding closing tag`


slax57 (approved):
Nice catch!

PR #9001, opened 2023-06-13
Fix: query permissions at startup only when a child function exist

adguernier:
When creating a React Admin App with `create-react-admin`, and choosing to add an `authProvider`, the provider is created with this method: `getPermissions: () => Promise.reject("Unknown method"),`.
This leads to an error log in browser devtools console (e.g: `Unknown method`), even the app doesn't need permission anywhere. 

Reproduce:
- Create a React Admin App with [`create-react-admin`](https://marmelab.com/react-admin/CreateReactAdmin.html)
- Run the app, go on the app with a browser and open browser devtools
- Wait some second and see the error throws in the console

~~Fix: Call `authProvider.getPermissions` only when a [child function exist](https://marmelab.com/react-admin/Permissions.html#restricting-access-to-resources-or-views) and only when the hook `usePermissions` is used in the application.~~

Fix: change `authProvider` template in `create-react-admin` package

### DIFF
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -35,10 +35,11 @@ const emptyParams = {};
  */
 const usePermissions = <Permissions = any, Error = any>(
     params = emptyParams,
-    queryParams: UseQueryOptions<Permissions, Error> = {
-        staleTime: 5 * 60 * 1000,
-    }
+    queryParams: UseQueryOptions<Permissions, Error>
 ) => {
+    if (!queryParams.staleTime) {
+        queryParams = { ...queryParams, staleTime: 5 * 60 * 1000 };
+    }
     const authProvider = useAuthProvider();
 
     const result = useQuery(
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -3,6 +3,9 @@ import { useQuery, UseQueryOptions } from 'react-query';
 import useAuthProvider from './useAuthProvider';
 
 const emptyParams = {};
+const defaultQueryParams: UseQueryOptions = {
+    staleTime: 5 * 60 * 1000,
+};
 /**
  * Hook for getting user permissions
  *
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/core/useConfigureAdminRouterFromChildren.tsx
```diff
@@ -43,7 +43,9 @@ import { useResourceDefinitionContext } from './useResourceDefinitionContext';
 export const useConfigureAdminRouterFromChildren = (
     children: AdminChildren
 ): RoutesAndResources & { status: AdminRouterStatus } => {
-    const { permissions, isLoading } = usePermissions();
+    const { permissions, isLoading } = usePermissions(undefined, {
+        enabled: !!getSingleChildFunction(children),
```
djhi, line 47: We should also ensure we don't loose the default query options defined in `usePermissions` (`staleTime`).

djhi: It would be better to update the `usePermissions` so that it merges the provided options with the default options

djhi (changes requested):

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -35,10 +35,11 @@ const emptyParams = {};
  */
 const usePermissions = <Permissions = any, Error = any>(
     params = emptyParams,
-    queryParams: UseQueryOptions<Permissions, Error> = {
-        staleTime: 5 * 60 * 1000,
-    }
+    queryParams: UseQueryOptions<Permissions, Error>
```
djhi, line 38: This was optional before as we provided a default. We must keep it that way
```suggestion
    queryParams?: UseQueryOptions<Permissions, Error>
```

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -35,10 +35,11 @@ const emptyParams = {};
...
 ) => {
+    if (!queryParams.staleTime) {
+        queryParams = { ...queryParams, staleTime: 5 * 60 * 1000 };
+    }
```
djhi, line 42: You can remove this and replace the current line 50 with:

```diff
- queryParams
+ { staleTime: 5 * 60 * 1000, ...queryParams }
```

djhi (changes requested):

djhi (approved):

PR #9019, opened 2023-06-16
Allow disabling store persistence of the list parameters

djhi:
## Problem

For one-off lists (e.g. in dashboards), `<ListBase>` does the fetching and puts the data in a  `<ListContext>`, so it’s fine.

But as it uses the store by default, a list for a given resource will be affected by the actions of the user in other lists of the same resource.

It’s possible to specify a different `storeKey` to avoid that, but it’s silly to have to define a store key when what I want is to disable the use of the store altogether.

It’s also possible to use `disableSyncWithLocation`, but the name of that prop isn’t clearly related to the store.

## Solution

Allow to set `storeKey` to `false` to disable the store altogether

### DIFF
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -88,8 +88,8 @@ export const useListParams = ({
     const location = useLocation();
     const navigate = useNavigate();
     const [localParams, setLocalParams] = useState(defaultParams);
-    // As we can't conditionally call a hook, ff the storeKey is false,
-    // we'll ignore the params variable later on and won't call  setParams either.
+    // As we can't conditionally call a hook, if the storeKey is false,
+    // we'll ignore the params variable later on and won't call setParams either.
     const [params, setParams] = useStore(
         storeKey || `${resource}.listParams`,
         defaultParams
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -88,6 +88,8 @@ export const useListParams = ({
     const location = useLocation();
     const navigate = useNavigate();
     const [localParams, setLocalParams] = useState(defaultParams);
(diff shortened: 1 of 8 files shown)


packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -88,15 +88,23 @@ export const useListParams = ({
     const location = useLocation();
     const navigate = useNavigate();
     const [localParams, setLocalParams] = useState(defaultParams);
-    const [params, setParams] = useStore(storeKey, defaultParams);
+    const [params, setParams] = useStore(
+        storeKey || `${resource}.listParams`,
```
fzaninotto, line 92: if `storeKey` is false, this will still use the store. Can you add a comment to explain what's the rationale?

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -111,7 +119,10 @@ export const useListParams = ({
         () =>
             getQuery({
                 queryFromLocation,
-                params: disableSyncWithLocation ? localParams : params,
+                params:
+                    disableSyncWithLocation || disableSyncWithStore
```
fzaninotto, line 123: I think you must also dd the check for `disableSyncWithStore` in `changeParams` L146, and anywhere else the `setParams` may be called.

fzaninotto (changes requested):

packages/ra-core/src/controller/list/useListParams.ts
djhi: We can't disable a hook, so although we still read the store, its values are not used later in the code

packages/ra-core/src/controller/list/useListParams.ts
djhi: `changeParams` does not call `setParams`

fzaninotto (approved):

PR #9032, opened 2023-06-20
[Doc] Add missing constant in code snippets

adguernier:
`const sort` was not defined but used in `<Datagrid data={data} sort={sort}>`

### DIFF
--- a/docs/ListTutorial.md
+++ b/docs/ListTutorial.md
@@ -43,10 +43,11 @@ const BookList = () => {
     const [filter, setFilter] = useState('');
     const [page, setPage] = useState(1);
     const perPage = 10;
+    const sort = { field: 'id', order: 'ASC' };
     const { data, total, isLoading } = useGetList('books', {
         filter: { q: filter },
         pagination: { page, perPage },
-        sort: { field: 'id', order: 'ASC' }
+        sort
     });
     if (isLoading) {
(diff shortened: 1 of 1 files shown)


docs/ListTutorial.md
```diff
@@ -43,10 +43,11 @@ const BookList = () => {
     const [filter, setFilter] = useState('');
     const [page, setPage] = useState(1);
     const perPage = 10;
+    const sort = { field: 'id', order: 'ASC' };
```
fzaninotto, line 46: It doesn't make sens to introduce it here IMO. Why `sort` and not `pagination`?

fzaninotto (changes requested):

docs/ListTutorial.md
adguernier: Sorry, wrong section

docs/ListTutorial.md
```diff
@@ -46,7 +46,7 @@ const BookList = () => {
     const { data, total, isLoading } = useGetList('books', {
         filter: { q: filter },
         pagination: { page, perPage },
-        sort: { field: 'id', order: 'ASC' }
```
fzaninotto, line 49: not good

fzaninotto (changes requested):

fzaninotto (approved):

Issue #9029, opened 2023-06-20
BulkUpdateButtons do not support custom mutation meta field


KonkretneKosteczki:
**Is your feature request related to a problem? Please describe.**
One can pass query options to the list, but not mutation options to the BulkUpdateButton
```
export const PostList = () => (
    <List queryOptions={{meta: {foo: "bar"}}}>
        <Datagrid bulkActionButtons={
          <>
            <BulkUpdateButton label="Make commentable" data={{commentable: true}} />
            <BulkDeleteButton />
          </>
        }>
            <TextField source="id" />
            <TextField source="title" />
            <DateField source="published_at" />
            <TextField source="category" />
            <BooleanField source="commentable" />
        </Datagrid>
    </List>
);
```

**Describe the solution you'd like**
It would be nice if the buttons accepted not onlhy mutationMode, but also mutationOptions prop the same way that Edit component does.
```
            <BulkUpdateButton mutationOptions={{meta: {foo: "bar"}}} label="Make commentable" data={{commentable: true}} />
            <BulkDeleteButton mutationOptions={{meta: {foo: "bar"}}} />
```

**Describe alternatives you've considered**
Maybe it doesn't have to be buttons, but list prop could accept the parameter and button scould somehow get it from the context instead.

slax57:
Thanks for this suggestion.

I think we should definitely add `mutationOptions` to `BulkUpdateButton`, especially provided that other similar buttons (`BulkDeleteButton`, `DeleteButton`) already support it.

I think it's best to set the `meta` prop at the button level, as it allows for more flexibility in case you need to pass a specific value for this case, besides I don't see what context could hold this value.

Would you like to open a PR to add this feature?

KonkretneKosteczki:
I added support for the mutationOptions in this pull https://github.com/marmelab/react-admin/pull/9034 I tried to follow implementation of BulkDeleteButton so that they are the same. Please let me know if the code is not up to standard I will update it.

slax57:
Implemented in https://github.com/marmelab/react-admin/pull/9035

PR #9034, opened 2023-06-21
feat(bulk-update-buttons): added mutation options to bulk update buttons

KonkretneKosteczki:
#9029 
1. Extended BulkUpdateWithUndoButtonProps and BulkUpdateWithConfirmButtonProps to inlude mutation options
2. Passing the props to the underlying useUpdateMany calls


slax57 (changes requested):
Looks good to me, thanks! :slightly_smiling_face: 

However, since it's a new feature, it should target the `next` branch instead of `master`.
Can you please rebase?

PR #9037, opened 2023-06-21
Prevent render InputHelperText if helperText props is false in Input components

adguernier:
related to #8830 and #8829

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
@@ -148,6 +148,8 @@ export const ArrayInput = (props: ArrayInputProps) => {
             </Labeled>
         );
     }
+    const renderHelperText =
+        helperText !== false || ((isDirty || isSubmitted) && !!error);
 
     return (
         <Root
@@ -186,7 +188,7 @@ export const ArrayInput = (props: ArrayInputProps) => {
                     disabled,
                 })}
             </ArrayInputContext.Provider>
-            {!!((isDirty || isSubmitted) && !!error) || helperText ? (
+            {renderHelperText ? (
                 <FormHelperText error={(isDirty || isSubmitted) && !!error}>
                     <InputHelperText
                         touched={isDirty || isSubmitted}
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -537,6 +537,9 @@ If you provided a React element for the optionText prop, you must also provide t
         return String(getChoiceValue(option)) === String(getChoiceValue(value));
     };
 
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
+
     return (
         <>
             <StyledAutocomplete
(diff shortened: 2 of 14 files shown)


packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
```diff
@@ -72,11 +73,13 @@ export const NullableBooleanInput = (props: NullableBooleanInputProps) => {
             }
             error={(isTouched || isSubmitted) && invalid}
             helperText={
-                <InputHelperText
-                    touched={isTouched || isSubmitted}
-                    error={error?.message}
-                    helperText={helperText}
-                />
+                helperText ? (
```
djhi, line 76:
```suggestion
                renderHelperText ? (
```

djhi (approved):

packages/ra-ui-materialui/src/input/NullableBooleanInput.tsx
adguernier: Wrong copy / paste :see_no_evil:

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -537,6 +537,9 @@ If you provided a React element for the optionText prop, you must also provide t
         return String(getChoiceValue(option)) === String(getChoiceValue(value));
     };
 
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
```
slax57, line 541: You're missing `fetchError` in this condition

packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
```diff
@@ -199,6 +199,9 @@ export const CheckboxGroupInput: FunctionComponent<CheckboxGroupInputProps> = pr
         );
     }
 
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
+
```
slax57, line 204: You're missing `fetchError` in the condition

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.tsx
```diff
@@ -158,6 +158,10 @@ export const RadioButtonGroupInput = (props: RadioButtonGroupInputProps) => {
             </Labeled>
         );
     }
+
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
```
slax57, line 163: You're missing `fetchError` in the condition

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -274,6 +274,8 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                   ),
               }
             : {};
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
```
slax57, line 278: You're missing `fetchError` in the condition

packages/ra-ui-materialui/src/input/SelectInput.tsx
```diff
@@ -296,6 +296,9 @@ export const SelectInput = (props: SelectInputProps) => {
         );
     }
 
+    const renderHelperText =
+        helperText !== false || ((isTouched || isSubmitted) && invalid);
```
slax57, line 300: You're missing `fetchError` in the condition

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -148,6 +148,8 @@ export const ArrayInput = (props: ArrayInputProps) => {
             </Labeled>
         );
     }
+    const renderHelperText =
+        helperText !== false || ((isDirty || isSubmitted) && !!error);
```
slax57, line 152: This condition is different from the one before.

If `helperText = undefined` (the default):
- Before we would have rendered `null`
- Now we render `<FormHelperText ... />`

This creates an unnecessary additional gap.
You should keep the previous condition IMO.

slax57 (changes requested):

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
djhi: It was wrong before. The additionnal gap isn't unnecessary as it prevents the UI from jumping when error messages are displayed

slax57 (approved):

djhi (approved):

PR #9039, opened 2023-06-22
Update useAuthenticated to fix disableAuthentication use

user1m:
fix useAuthenticated  so that disableAuthentication can be used as intended

Resolving this issue: https://stackoverflow.com/questions/75072440/how-to-create-an-unauthenticated-route-in-react-admin#comment132585925_75073037

### DIFF
--- a/packages/ra-core/src/auth/useAuthenticated.ts
+++ b/packages/ra-core/src/auth/useAuthenticated.ts
@@ -30,7 +30,8 @@ export const useAuthenticated = <ParamsType = any>({
     params,
     ...options
 }: UseAuthenticatedOptions<ParamsType> = {}) => {
-    useAuthState(params ?? emptyParams, true, options);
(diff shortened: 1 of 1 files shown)


slax57:
Thanks for your contribution.

However, to me, the issue you are trying to fix is actually not an issue.
I have just tested the case in our simple example and it works just fine.

Hence I don't think this fix is necessary.

If you are still convinced there is an issue in react-admin, please open an issue first so that we can reproduce and confirm the problem.

slax57:
Here is the working [Stackblitz](https://stackblitz.com/edit/github-qo4dbm?file=src%2FcustomRouteLayout.tsx) for reference. You can logout and browse to `/custom3` and it will display the list of posts without requiring to log in.

user1m:
@slax57 the problem arises when you use `disableAuthentication` on a `<List>, Show, Create, Edit` etc with an anonymous route.  The expectation is that this is suppose to work: https://marmelab.com/react-admin/List.html#disableauthentication

PR #9043, opened 2023-06-25
Add SelectAllButton to BulkActionsToolbar

fzaninotto:
## Problem

When a user selects all the records of a given page, they often want to select all the records regardless of pagination. Doing so is cumbersome as it implies browsing every page and clicking on "select all". 

## Solution

Add a "select all" button allowing to expand the selection to the same query (with the same filters), but without pagination. 

https://github.com/marmelab/react-admin/assets/99944/9697a1c8-df1b-40ef-8b49-63ab601357e0

## Todo

- [x] Make it work
- [ ] Add tests
- [ ] Add stories
- [ ] Document it

### DIFF
--- a/packages/ra-core/src/controller/list/useRecordSelection.spec.tsx
+++ b/packages/ra-core/src/controller/list/useRecordSelection.spec.tsx
@@ -33,47 +33,102 @@ describe('useRecordSelection', () => {
         expect(selected).toEqual([123, 456]);
     });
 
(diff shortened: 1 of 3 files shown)


fzaninotto:
The implementation is wrong: it won't work for a datagrid inside a ReferenceManyField. The `selectAll` callback must be defined at the controller level, and stored in the MistContext.

Issue #8545, opened 2023-01-04
Extended AutocompleteInput with useChoicesContext and optionText clears itself and flickers


harastaivan:
**What you were expecting:**

Input should behave like normal `AutocompleteInput`. Same as without using `useChoicesContext`, options don't flicker, input doesn't clear itself after network call.

**What happened instead:**
After network call is made to `ReferenceInput`'s `reference`, the input clears, the dropdown show previous options and then flickers the filtered options.

**Steps to reproduce:**
Go to any `PostEdit`, first input is `Test users`, type in something, then the flickering and input text clear happen.

**Related code:**
https://stackblitz.com/edit/github-jnpt6d?file=src%2Fposts%2FUserReferenceInput.tsx,src%2Fposts%2FPostEdit.tsx,src%2Fposts%2FUserAutocompleteInput.tsx

Go to any `src/posts/PostEdit.tsx`, first input is

```jsx
<ArrayInput source="test_users" fullWidth>
  <SimpleFormIterator inline>
    <UserReferenceInput />
  </SimpleFormIterator>
</ArrayInput>
```

inside `UserReferenceInput` there is `UserAutocompleteInput` which reproduces the issue.

The bug is reproducible when `UserAutocompleteInput` uses both:
- `useChoicesContext` to somehow read the choices
- `optionText`, `inputText` and `matchSuggestion` functionality.

I tried passing all the props or forwarding the ref which didn't help.

**Environment:**

- React-admin version: 4.6.2
- React version: 17.0.2
- Browser: Chrome 108

djhi:
Can you share a bit more about what you're trying to achieve? I fail to understand why you would call `useChoicesContext` without using any of the values it returns

fzaninotto:
No news for some time, closing.

PR #9046, opened 2023-06-26
Update documentation about AutocompleteInput and AutocompleteArrayInput

djhi:


djhi:
Fixes #8545

docs/AutocompleteArrayInput.md
```diff
@@ -337,25 +337,37 @@ const choices = [
...
 
 <AutocompleteArrayInput source="authors" choices={choices} optionText={optionRenderer} />
 ```
 
-`optionText` also accepts a React Element, that will be rendered inside a [`<RecordContext>`](./useRecordContext.md) using the related choice as the `record` prop. You can use Field components there.
+**Tip**: Make sure you provide a stable reference to the function passed as `optionText`. Either declare it outside the component render function or wrap it inside a [`useCallback`](https://react.dev/reference/react/useCallback).
+
+`optionText` also accepts a React Element, that will be rendered inside a [`<RecordContext>`](./useRecordContext.md) using the related choice as the `record` prop. You can use Field components there. However, using an element as `optionText` implies that you also set two more props, `inputText` and `matchSuggestion`. See [Using A Custom Element For Options](#using-a-custom-element-for-options) for more details.
+
+`optionText` is also useful when the choices are records [fetched from another resource](#fetching-choices), and `<AutocompleteArrayInput>` is a child of a [`<ReferenceArrayInput>`](./ReferenceArrayInput.md).
```
slax57, line 351: To me, this line is redundant with line 325

docs/AutocompleteArrayInput.md
```diff
@@ -337,25 +337,37 @@ const choices = [
...
+
+But if you set the `optionText` prop, react-admin uses it instead of relying on `recordRepresentation`.
+
+```jsx
+import { AutocompleteArrayInput, ReferenceArrayInput } from 'react-admin';
 
-<AutocompleteArrayInput source="authors" choices={choices} optionText={<FullNameField />}/>
+<ReferenceArrayInput label="Author" source="author_id" reference="authors">
+    <AutocompleteArrayInput optionText="last_name" />
+</ReferenceArrayInput>
```
slax57, line 370: This example is redundant with example given lines 328-330.

Besides, you removed the example where we actually pass a React element to `optionText`. We should keep at least one such example IMO.

slax57 (changes requested):
Also, the title of the PR is misleading, because you also changed something in the code, not only the docs

slax57 (approved):

PR #9055, opened 2023-06-27
Add eslint, prettier and default vite gitignore to create-react-admin

djhi:


packages/create-react-admin/templates/common/.eslintrc
```diff
@@ -0,0 +1,26 @@
...
+        "browser": true,
+        "es2021": true
+    },
+    "settings": {
+        "react": {
+            "version": "detect"
+        }
+    },
+    "rules": {
+        "no-unused-vars": [
```
fzaninotto, line 21: why do we need custom rules?

smeng9:
Shall we consider eslint.config.js as the .eslintrc format will be deprecated.

Source: https://eslint.org/docs/latest/use/configure/configuration-files

packages/create-react-admin/templates/common/.eslintrc
fzaninotto: I think we'd better fix the code (or ignore the rule) in our authProvider instead

djhi: Did you look at it? I'd rather not

djhi: Disabled the rule locally

fzaninotto (approved):

slax57 (changes requested):
I'm seing the new config files alright, but I'm having issues when running `npm run lint` and `npm run format` in the generated project:

```
❯ npm run lint

> lint
> eslint --fix --ext .js,.jsx,.ts,.tsx ./src


Oops! Something went wrong! :(

ESLint: 8.43.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config

ESLint looked for configuration files in /home/slax57/workspaces/my-admin/src and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint Discord server: https://eslint.org/chat
```

and

```
❯ npm run format

> format
> prettier --write ./src

[error] Invalid configuration file `src/App.tsx`: Unexpected token 'export'
```

Do you have them too?

packages/create-react-admin/src/cli.tsx
```diff
@@ -17,5 +17,7 @@ const cli = meow(
 if (cli.flags.h) {
     cli.showHelp();
 } else {
-    render(<App name={cli.input.length > 0 ? cli.input[0] : 'my-admin'} />);
+    render(<App name={cli.input.length > 0 ? cli.input[0] : 'my-admin'} />, {
+        debug: true,
+    });
```
slax57, line 22: is this change intended?

slax57 (changes requested):
Formatting seems to work now.

However I see visual regressions:
- When choosing an option in the CLI, the whole section re-renders when the option changes
- Once the script has run, the message "Your application my-admin was successfully generated. blahblahblah" appears twice

slax57 (approved):
Alright, everything looks good this time! :muscle:

PR #9061, opened 2023-06-28
feat: unix boolean prop for `DateField`

TurtIeSocks:
This adds a `unix` boolean prop to the `DateField` component that will multiple a field value by 1000, assuming it is a number.


fzaninotto:
Thanks!

As it's a new feature, please rebase your PR against the `next` branch rather than `master`, which is reserved for bug fixes.

djhi:
I don't get this. It seems to me it already works. In this sandbox (https://stackblitz.com/edit/github-v5fmzn?file=src%2Fdata.tsx), I changed the dates in the `data.ts` file to be unix timestamps (using the date [valueOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf) method). Did I miss something?

fzaninotto:
This is already working since react-admin 4.11 (see #8863)

Issue #8776, opened 2023-03-29
Add the ability to disable title configuration (while still make the rest of the page configurable)


smeng9:
**Is your feature request related to a problem? Please describe.**
The documentation of AppBar says https://marmelab.com/react-admin/AppBar.html#configurable

> By default, users can override the page title [in configurable mode](https://marmelab.com/react-admin/Features.html#configurable-ui).

But did not suggest a way to turn it off while keeping the rest of the page configurable.

**Describe the solution you'd like**

I would like the `<Title />` component to accept a boolean prop called nonconfigurable, so user can optionally disable it

**Describe alternatives you've considered**

The alternative is developer has to manually change list, show, create and import additional react functions to render PageTitle into a portal.

slax57:
I agree, we should allow to disable this feature while still allowing other parts of the app to be configurable.
Would you like to open a PR?

ujjwalpandey922:
Hi,
Can I work on this?

slax57:
@ujjwalpandey922 Of course! :slightly_smiling_face: 
Just remember to read the [Contribution Guidelines](https://github.com/marmelab/react-admin#contributing) first, and to target your PR against the `next` branch as it's a new feature.
Thanks!

Aditay2002:
@smeng9 is this issue still open ? i will like to contribute towards it

slax57:
@Aditay2002 yes the issue is still open, you can go ahead :slightly_smiling_face: 
Thanks

adguernier:
Hi @Mex7180, it appears no one is working on this issue yet. So yes, you can give a try!
Don't forget to read our [Contribution Guidelines](https://github.com/marmelab/react-admin#contributing) first, and to target your PR against the `next` branch.
Thanks for your contribution :pray:

markyao6275:
I have a pull request: https://github.com/marmelab/react-admin/pull/9836 that attempts to tackle this issue. Would appreciate a review, thanks!

slax57:
Implemented in https://github.com/marmelab/react-admin/pull/9836

PR #9062, opened 2023-06-28
Update Title.tsx

Aditay2002:
added a prop in title i.e nonconfigurable and updated the check that renders title component as null .

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Title.tsx
+++ b/packages/ra-ui-materialui/src/layout/Title.tsx
@@ -8,39 +8,41 @@ import { RaRecord, warning } from 'ra-core';
 import { PageTitleConfigurable } from './PageTitleConfigurable';
 
 export const Title = (props: TitleProps) => {
-    const { defaultTitle, title, preferenceKey, ...rest } = props;
-    const [container, setContainer] = useState(() =>
-        typeof document !== 'undefined'
-            ? document.getElementById('react-admin-title')
-            : null
-    );
+  const { defaultTitle, title, preferenceKey, nonconfigurable, ...rest } = props;
+  const [container, setContainer] = useState(() =>
+    typeof document !== 'undefined'
+      ? document.getElementById('react-admin-title')
+      : null
+  );
 
-    // on first mount, we don't have the container yet, so we wait for it
-    useEffect(() => {
-        setContainer(container => {
-            const isInTheDom =
-                typeof document !== 'undefined' &&
-                document.body.contains(container);
-            if (container && isInTheDom) return container;
-            return typeof document !== 'undefined'
-                ? document.getElementById('react-admin-title')
-                : null;
-        });
-    }, []);
+  // on first mount, we don't have the container yet, so we wait for it
+  useEffect(() => {
+    setContainer((container) => {
+      const isInTheDom =
+        typeof document !== 'undefined' && document.body.contains(container);
+      if (container && isInTheDom) return container;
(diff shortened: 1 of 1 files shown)


djhi:
Can you please explain the problem you want to solve with this?

Aditay2002:
@djhi it is not  a problem we are just providing user an additional feature which help user to optionally disable the  feature  of changing the <AppBar> title   in configurable mode.

smeng9:
To clarify, here is the linked issue: https://github.com/marmelab/react-admin/issues/8776

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -8,39 +8,41 @@ import { RaRecord, warning } from 'ra-core';
 import { PageTitleConfigurable } from './PageTitleConfigurable';
 
 export const Title = (props: TitleProps) => {
-    const { defaultTitle, title, preferenceKey, ...rest } = props;
-    const [container, setContainer] = useState(() =>
-        typeof document !== 'undefined'
-            ? document.getElementById('react-admin-title')
-            : null
-    );
+  const { defaultTitle, title, preferenceKey, nonconfigurable, ...rest } = props;
```
slax57, line 11:
```suggestion
  const { defaultTitle, title, preferenceKey, nonConfigurable, ...rest } = props;
```

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -8,39 +8,41 @@ import { RaRecord, warning } from 'ra-core';
...
+  const [container, setContainer] = useState(() =>
+    typeof document !== 'undefined'
+      ? document.getElementById('react-admin-title')
+      : null
+  );
```
slax57, line 16: Please configure your IDE to use our current Prettier configuration, to avoid breaking the indentation of the file

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -8,39 +8,41 @@ import { RaRecord, warning } from 'ra-core';
...
+        typeof document !== 'undefined' && document.body.contains(container);
+      if (container && isInTheDom) return container;
+      return typeof document !== 'undefined'
+        ? document.getElementById('react-admin-title')
+        : null;
+    });
+  }, []);
 
-    if (!container) return null;
+  if (!container || nonconfigurable) return null; // Return null if nonconfigurable prop is true
```
slax57, line 30: I don't understand why you would return `null` in this case, because then you would simply have no title at all. To me you should rather use `nonConfigurable` to choose between returning `<PageTitleConfigurable>` and `<PageTitle>`.

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -53,6 +55,7 @@ Title.propTypes = {
     className: PropTypes.string,
     record: PropTypes.any,
     title: TitlePropType,
+    nonconfigurable: PropTypes.bool,
```
slax57, line 58: Please also update the TS type, not only the propTypes

slax57 (changes requested):
Thanks for this contribution :slightly_smiling_face: 

There are some things you need to fix in your PR, see below.

Also, since this is a new feature, please target your PR to the `next` branch instead of `master`.

slax57:
Superseded by https://github.com/marmelab/react-admin/pull/9068

PR #9064, opened 2023-06-30
Rich Text Input - working as JSON

jhrncar:
I have added a optional prop useJson to RichTextInput. It is very common for these kinds of inputs to be stored in JSON format rather than raw HTML and TipTap has full supoort for this approach - the getJSON() function. IT automatically parses the output back in to the editor and is compatible with everything i found so far.


jhrncar:
oopsie, gonna re-open this, sorry for unneccessary spam

PR #9065, opened 2023-06-30
Rich Text Input - working as JSON

jhrncar:
I have added a optional prop useJson to RichTextInput. It is very common for these kinds of inputs to be stored in JSON format rather than raw HTML and TipTap has full supoort for this approach - the getJSON() function. IT automatically parses the output back in to the editor and is compatible with everything i found so far.

### DIFF
--- a/packages/ra-input-rich-text/README.md
+++ b/packages/ra-input-rich-text/README.md
@@ -28,6 +28,52 @@ export const PostEdit = (props) => (
 );
 ```
 
+## Using JSON as the data format
+Under the hood, the `<RichTextInput>` defaultly uses HTML as the data format - which when saved, is then exported as raw HTML in string. If you want to use JSON as the data format, you can pass the `useJson` prop to the `<RichTextInput>` component. This will then save the data as JSON in raw string, using the TipTap's JSON scheme. 
+
+```jsx
+import { Edit, SimpleForm, TextInput } from 'react-admin';
+import { RichTextInput } from 'ra-input-rich-text';
+
+export const PostEdit = (props) => (
+	<Edit {...props}>
+		<SimpleForm>
+			<TextInput source="title" />
+			<RichTextInput source="body" useJson />
+		</SimpleForm>
+	</Edit>
+);
+```
+
+Example of the JSON data format:
+
+```json
+{
+  "type": "doc",
+  "content": [
+	{
+	  "type": "heading",
+	  "attrs": { "textAlign": "left", "level": 1 },
+	  "content": [
+		{
+		  "type": "text",
+		  "text": "Hello world!"
+		}
+	  ]
+	}
+  ]
+}
+```
+Renders as:
+```html
+<h1 style="text-align: left;">Hello world!</h1>
+```
+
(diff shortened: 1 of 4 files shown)


slax57 (changes requested):
Thanks for the contribution :slightly_smiling_face: 

It looks like a quick feature to implement indeed, so I'm in favor of adding it. :slightly_smiling_face: 

Could you please also:
- Add a test to prove that it works
- Add a story to make it easier to test this new prop
- Document this new prop in the component's documentation file

Thank you so much

slax57:
No worries, it's a promising start so far :slightly_smiling_face: 
Feel free to ask us if you need help running the tests or the storybook.

jhrncar:
Do you guys have any guide on how you do tests and stories? In the Contribution part of the README it doesnt say it :D

adguernier:
Hi ! 

The stories live near the component, so you have a `RichTextInput.stories.tsx` for `RichTextInput.tsx` component. You can add a  story to test the behavior of the component with your new prop.

Then, to see the result, you can launch _Storybook_ with `make storybook`.

Same for test: you have a file `RichTextInput.spec.tsx` near `RichTextInput.tsx`. You can run test with `yarn test-unit RichTextInput`.

Hope this help :slightly_smiling_face:

jhrncar (commented):
Story, test implemented and passed. Doc implemented

jhrncar:
I have stated it in the documentation, but I will say it here as well:

using this prop (useJSON) will **destroy** the compatibility with react-admin's <RichTextField>. A user will have to create his own renderer, unless one is provided - this would require major changes to the <RichTextField>. Currently, it uses dangerouslySetInnerHTML with external sanitizers - the renderer based on JSON should be safer from the start. I have provided a link to a renderer made by 3rd party that served as a basis for my own <RichTextField> - however I have not tested it in production enogh (unlike this prop) so that i will be comfortable making it public. 

Let me know your thoughts on this guys

slax57 (changes requested):
@jhrncar Thank you so much for the work you've started! :slightly_smiling_face: 

Unfortunately, I kind of agree with you that it can be a bit disturbing not to have a Field component matching this Input.
Hence, I'm not in favor of merging this PR in its current state.

Since building the matching Field component would require more work, and a 3rd party lib (which means yet another dependency to maintain), I believe it would be more adequate to publish this component (the RichTextInput with JSON support) along with the corresponding Field component in an external package.

If you are willing to publish them as such, we could then add a link to your package in the official doc.

WDYT?

jhrncar:
> @jhrncar Thank you so much for the work you've started! 🙂
> 
> Unfortunately, I kind of agree with you that it can be a bit disturbing not to have a Field component matching this Input. Hence, I'm not in favor of merging this PR in its current state.
> 
> Since building the matching Field component would require more work, and a 3rd party lib (which means yet another dependency to maintain), I believe it would be more adequate to publish this component (the RichTextInput with JSON support) along with the corresponding Field component in an external package.
> 
> If you are willing to publish them as such, we could then add a link to your package in the official doc.
> 
> WDYT?

Yes, I see your point, but how as I am quite unexperienced with making packages etc, how would this work? Because, if a user would install this package, it would need to overwrite the React-Admin's RichTextInput and RichTextField, no?

PR #9066, opened 2023-06-30
Translations in Rich Text Input Toolbar Buttons

jhrncar:
Most of React Admin uses the ability to specify translations, I think its nice to have the option to have the translations available in these buttons too.


slax57 (commented):
Thank you for this contribution :slightly_smiling_face: 

However, I don't really understand the benefit of being able to pass translations as label props, versus setting the desired labels as translation keys directly in the i18nProvider.

Can you elaborate on your use-case?

jhrncar:
Hello, you are right, I completely forgot about you being able to use  the defined keys in your own provider. This method is redundant. Maybe I wil look into the customization features in the button later <3

PR #9072, opened 2023-07-04
Alert user if he close the window before the end of undoable notification

### LINKED ISSUE #8505: Undoable mutations not executed if user closes browser tab too soon
**What you were expecting:**
I was expecting that undoable mutations are executed unless the user clicks "UNDO" on the "Element updated" notification.

**What happened instead:**
If user closes the browser tab before the "Element updated" notification disappears (that being the time when the mutation **would** be executed), the mutation is not executed. The user has no indication that the mutation didn't go through, and would assume that it did.

**Steps to reproduce:**
1. Make an edit to a resource that does undoable edits
2. Close the browser tab immediately after that (before the "Element updated" notification disappears)
3. Open another browser tab and view the resource that you just edited
4. Observe that the edit was never applied

**Related code:**

n/a

**Other information:**

none

**Environment**

* React-admin version: 4.5.2
* React version: 18.2.0
* Browser: Chrome (or any)

adguernier:
Fix #8505

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Notification.tsx
+++ b/packages/ra-ui-materialui/src/layout/Notification.tsx
@@ -43,6 +43,14 @@ export const Notification = (props: NotificationProps) => {
     const translate = useTranslate();
 
     useEffect(() => {
+        const beforeunload = (e: BeforeUnloadEvent) => {
+            if (messageInfo?.notificationOptions?.undoable) {
+                e.preventDefault();
+            }
+        };
+
+        window.addEventListener('beforeunload', beforeunload);
+
         if (notifications.length && !messageInfo) {
             // Set a new snack when we don't have an active one
             setMessageInfo(takeNotification());
@@ -51,6 +59,10 @@ export const Notification = (props: NotificationProps) => {
             // Close an active snack when a new one is added
             setOpen(false);
         }
+
+        return () => {
+            window.removeEventListener('beforeunload', beforeunload);
+        };
     }, [notifications, messageInfo, open, takeNotification]);
 
     const handleRequestClose = useCallback(() => {


packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -43,6 +43,14 @@ export const Notification = (props: NotificationProps) => {
     const translate = useTranslate();
 
     useEffect(() => {
+        const beforeunload = (e: BeforeUnloadEvent) => {
```
fzaninotto, line 46: This will always prevent the default, no matter if there is an undoable operation ongoing or not.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/layout/Notification.tsx
adguernier: I just did a test and it seems to work thanks to the condition below

fzaninotto: I suggest you only add an event listener when you're in an undoable notification, instead of always adding it and verifying the condition in the listener. Apparently, beforeunload event listeners affect performance (see https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes) and should be used with caution. 

Also, the advised syntax for a beforeunload event listener is to return a string (see https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#examples). Some browsers don't follow the spec, and the preventDefault won't be enough.

adguernier: You are right, add the event listener in the condition is working as expected. I may have made a careless mistake.
I also updated the return type to maximize browsers compatibility according to [this examples](https://developer.mozilla.org/en-US/docs/Web/API/BeforeUnloadEvent#examples)

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -51,6 +62,10 @@ export const Notification = (props: NotificationProps) => {
             // Close an active snack when a new one is added
             setOpen(false);
         }
+
+        return () => {
+            window.removeEventListener('beforeunload', beforeunload);
```
fzaninotto, line 67: remove the event listener only if you added it! You must enclose this in the same condition as line 53

fzaninotto (changes requested):

fzaninotto:
Works like a charm!

PR #9074, opened 2023-07-04
[Doc] Update Tutorial to use TypeScript by default

adguernier:
Update `js` and `jsx` snippets to `ts` and `tsx` in __Getting Started / Tutorial__

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -47,7 +47,7 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 import { App } from './App';
 
-ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
+ReactDOM.createRoot(document.getElementById('root')!).render(
     <React.StrictMode>
         <App />
     </React.StrictMode>
@@ -164,18 +164,18 @@ Let's copy this code, and create a new `UserList` component, in a new file named
 import { List, Datagrid, TextField, EmailField } from "react-admin";
 
 export const UserList = () => (
-  <List>
-    <Datagrid rowClick="edit">
-      <TextField source="id" />
-      <TextField source="name" />
-      <TextField source="username" />
-      <EmailField source="email" />
-      <TextField source="address.street" />
-      <TextField source="phone" />
-      <TextField source="website" />
-      <TextField source="company.name" />
-    </Datagrid>
-  </List>
+    <List>
+        <Datagrid rowClick="edit">
+            <TextField source="id" />
+            <TextField source="name" />
+            <TextField source="username" />
+            <EmailField source="email" />
+            <TextField source="address.street" />
+            <TextField source="phone" />
+            <TextField source="website" />
+            <TextField source="company.name" />
+        </Datagrid>
+    </List>
 );
 ```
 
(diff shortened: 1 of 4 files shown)


fzaninotto:
I think you should change the code type to typescript in all snippets, even if you don't add annotations. Otherwise, users will wonder why some snippets have a JS/TS switcher and some don't.

docs/Tutorial.md
```diff
@@ -877,13 +877,13 @@ For this tutorial, since there is no public authentication API, we can use a fak
...
+```ts
 // in src/authProvider.ts
+import { AuthProvider } from "react-admin";
 
-// TypeScript users must reference the type: `AuthProvider`
-export const authProvider = {
+export const authProvider: AuthProvider = {
   // called when the user attempts to log in
-  login: ({ username }) => {
+  login: ({ username }: { username: string }) => {
```
djhi, line 886: As you already declared `authProvider` as an `AuthProvider`, you don't need to provide more types. Same for all methods

docs/Tutorial.md
```diff
@@ -372,11 +372,11 @@ In react-admin, fields are just React components. When rendered, they grab the `
 
 That means that you can do the same to write a custom Field. For instance, here is a simplified version of the `<UrlField>`:
 
-```jsx
+```tsx
 // in src/MyUrlField.tsx
 import { useRecordContext } from "react-admin";
 
-const MyUrlField = ({ source }) => {
+const MyUrlField = ({ source }: { source: string }) => {
```
djhi, line 379: You can import `FieldProps` from `react-admin` and use it here:
```suggestion
const MyUrlField = ({ source }: FieldProps) => {
```

docs/Tutorial.md
```diff
@@ -415,13 +415,13 @@ The `<MyUrlField>` component is a perfect opportunity to illustrate how to custo
...
 {% raw %}
-```jsx
+```tsx
 // in src/MyUrlField.tsx
 import { useRecordContext } from "react-admin";
 import { Link } from "@mui/material";
 import LaunchIcon from "@mui/icons-material/Launch";
 
-const MyUrlField = ({ source }) => {
+const MyUrlField = ({ source }: { source: string }) => {
```
djhi, line 424: Same

djhi (changes requested):

docs/Tutorial.md
fzaninotto: You marked this as fixed but it isn't.

fzaninotto (changes requested):

docs/Tutorial.md
adguernier: As discussed with @djhi, typed params with `FieldProps` generates Typescript errors, so we decided to let it as is

djhi: I'll work on a fix when I can

docs/Tutorial.md
```diff
@@ -47,7 +47,7 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 import { App } from './App';
 
-ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
+ReactDOM.createRoot(document.getElementById('root')!).render(
```
slax57, line 50: Don't know how others feel about this change, but I liked to original code better personally

fzaninotto: Yet we're based on Vite, and they did the same in https://github.com/vitejs/vite/commit/6fe14911deac34f5874687705a46deb9caf1cb10

fzaninotto (approved):

docs/Tutorial.md
slax57: alright, valid point :grin:

slax57 (approved):

PR #9077, opened 2023-07-05
[Doc] Add doc for `<BulkUpdateButton>`

slax57:


docs/Buttons.md
```diff
@@ -246,11 +246,58 @@ export const PostList = () => (
...
+| `confirmContent`     | Optional | React node                              | -                  | Lets you customize the content of the confirm dialog. Only used in `'pessimistic'` or `'optimistic'` mutation modes |
+| `confirmTitle`       | Optional | `string`                                | -                  | Lets you customize the title of the confirm dialog. Only used in `'pessimistic'` or `'optimistic'` mutation modes   |
+| `label`              | Optional | `string`                                | 'ra.action.delete' | label or translation message to use                                                                                 |
+| `icon`               | Optional | `ReactElement`                          | `<DeleteIcon>`     | iconElement, e.g. `<CommentIcon />`                                                                                 |
+| `mutationMode`       | Optional | `'undoable'|'pessimistic'|'optimistic'` | `'undoable'`       | Mutation mode                                                                                                       |
```
djhi, line 255: This line break the table format

docs/Buttons.md
```diff
@@ -246,11 +246,58 @@ export const PostList = () => (
...
+| `confirmContent`     | Optional | React node                              | -                  | Lets you customize the content of the confirm dialog. Only used in `'pessimistic'` or `'optimistic'` mutation modes |
+| `confirmTitle`       | Optional | `string`                                | -                  | Lets you customize the title of the confirm dialog. Only used in `'pessimistic'` or `'optimistic'` mutation modes   |
+| `icon`               | Optional | `ReactElement`                          | `<ActionUpdate>`   | An icon element                                                                                                     |
+| `label`              | Optional | `string`                                | 'ra.action.update' | Label or translation message to use                                                                                 |
+| `mutationMode`       | Optional | `'undoable'|'pessimistic'|'optimistic'` | `'undoable'`       | Mutation mode                                                                                                       |
```
djhi, line 297: Same

djhi (changes requested):

docs/Buttons.md
```diff
@@ -246,11 +246,58 @@ export const PostList = () => (
...
+    </List>
+);
+```
+{% endraw %}
+
+![Bulk Update button](./img/bulk-update-button.png)
+
+| Prop                 | Required | Type                                    | Default            | Description                                                                                                         |
+| -------------------- | -------- | --------------------------------------- | ------------------ | ------------------------------------------------------------------------------------------------------------------- |
+| `data`               | Required | `object`                                | -                  | An object with the fields that need to be updated on the selected records                                           |
```
adguernier, line 292: Should not we provide an example of the shape of the `data` object ?

slax57: You have one example in the snippet just above the table. Am I missing something?

djhi (commented):
We should probably extract this in a dedicated documentation page and reference it there eventually

slax57:
> We should probably extract this in a dedicated documentation page and reference it there eventually

Eventually yes, along with the other buttons referenced in `Buttons.md`. But for now I think simply documenting this (existing) button next to and in the same style as its siblings (`BulkDeleteButton` and `BulkExportButton`) is good enough.
Wdyt?

PR #9085, opened 2023-07-07
Include newsletter form

djhi:
image


slax57:
Don't know if it's only on my computer or not, but the email input text is white on a white background, so basically invisible :shrug: 

image

PR #9088, opened 2023-07-10
Add UpdateButton

djhi:
[updatebutton.webm](https://github.com/marmelab/react-admin/assets/1122076/6e63b0b8-97db-43b6-8458-a1a0b6272b94)


docs/UpdateButton.md
```diff
@@ -0,0 +1,354 @@
...
+const PostEditActions = () => (
+    <TopToolbar>
+        <UpdateButton label="Reset views" data={{ views: 0 }} mutationMode="pessimistic" />
+    </TopToolbar>
+);
+```
+{% endraw %}
+
+
+**Tip**: When using any other mode than `undoable`, the `<UpdateButton>` displays a confirmation dialog before calling the dataProvider.
```
slax57, line 109: I wonder if we should provide a way to avoid having the confirmation dialog in pessimistic and optimistic modes too. Indeed, it makes sense for a Delete button, but not necessarily for an Update button... WDYT?

slax57 (changes requested):

docs/UpdateButton.md
```diff
@@ -0,0 +1,302 @@
...
+layout: default
+title: "The UpdateButton Component"
+---
+
+# `<UpdateButton>`
+
+This component allows to create a button that updates a record by calling the [`useUpdate hook`](./useUpdate.md).
+
+<video controls playsinline muted loop poster="./img/updatebutton.png" >
+  <source src="./img/updatebutton.webm" type="video/mp4" />
```
slax57, line 11: You're mixing up webm and mp4 on this line. Btw you should have both or only mp4.

slax57 (changes requested):

PR #9090, opened 2023-07-10
Fix InputHelperText error message is not translated

### LINKED ISSUE #9084: String error messages are not translated when using global validation
**What you were expecting:**

When rendering InputHelperText with props: 
```
{
error: 'my.translation.key'
touched: true
}
```
Then error message should be translated

**What happened instead:**

They are not translated (because translate() is not called in the component InputHelperText)

**Steps to reproduce:**

Use yupResolver(<schema>) as resolver in a Form component. Error message displayed in the InputHelperText should be translated, but they are not. Instead the translation key is displayed.

In the sandbox below: 
1 - In the form fill the field 'body' and let field 'post' empty. 
2 - Validate the form
3 - The error displayed in the helper text is not translated (the same key is translated and displayed juste above the form)

**Related code:**

https://stackblitz.com/edit/github-jywgwp?file=src%2Findex.tsx

**Other information:**
**Solution**
change the following in 'packages/ra-ui-materialui/src/input/InputHelperText.tsx' line 14

Change
```
    if (touched && error) {
        if ((error as ValidationErrorMessageWithArgs).message) {
            return <ValidationError error={error} />;
        }
        return <>{error}</>;
    }
```

Into
```
    if (touched && error) {
        if ((error as ValidationErrorMessageWithArgs).message) {
            return <ValidationError error={error} />;
        }
        return <>{translate(error)}</>;
    }
```

**Environment**

* React-admin version: 4.12
* Last version that did not exhibit the issue (if applicable): was working on 4.4 
* React version: 18.2
* Browser: all

WiXSL:
Fixes #9084

### DIFF
--- a/packages/ra-ui-materialui/src/form/SimpleForm.spec.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.spec.tsx
@@ -90,12 +90,6 @@ describe('<SimpleForm />', () => {
             expect(mock).toHaveBeenCalledWith(
                 "Missing translation for key 'The title is required'"
             );
-            expect(mock).not.toHaveBeenCalledWith(
-                "Missing translation for key 'The author is required'"
-            );
-            expect(mock).not.toHaveBeenCalledWith(
-                "Missing translation for key 'The year must be less than 2000'"
-            );
             mock.mockRestore();
         });
     });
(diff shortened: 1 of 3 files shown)


WiXSL (commented):
The failing test in `SimpleForm.spec.ts` is because the way it is written, now the `InputHelperText` translates the messages first.

WiXSL:
> Thanks but that's not good. The story shows that we now indeed have translation warnings when we shouldn't. I have no idea why yet though

Isn't it because now the `InputHelperText` translates the error and then the `translate` function of the story receives the already translated label, generating a warning?

slax57:
Potential duplicate of https://github.com/marmelab/react-admin/pull/9109

fzaninotto:
As seen in #9084, this isn't the right fix, so I'm closing this PR.

PR #9092, opened 2023-07-11
Fix Fields Record Type Parameter

djhi:
Fixes https://github.com/marmelab/react-admin/issues/9102


slax57:
This should also fix https://github.com/marmelab/react-admin/issues/9102

fzaninotto (approved):

PR #9099, opened 2023-07-13
Feat: Improve lighthouse for landing page

arimet:
FROM
image

TO
image

**These results were obtained locally.** 

In production mode, the "Best practice" section drops to 92. 
This is due to the import of the Video player, which raises an error in the console. 

To achieve 100% accessibility, the contrast between the pink color and the background color needs to be modified.

To go to 100% performance, we'd have to do without the video player and see if it's possible via github-page to put assets with more cache (currently only 10min).


landing-page/index.html
```diff
@@ -787,6 +816,9 @@
             <img
               src="./img/your-api.svg"
               class="h-[90px] w-[320px] md:h-[110px] md:w-[410px] lg:h-[135px] lg:w-[500px] xl:h-[150px] xl:w-[560px]"
+              alt="You api"
```
slax57, line 819:
```suggestion
              alt="Your api"
```

slax57 (approved):
Amazing work, thank you so much!

PR #8758, opened 2023-03-22
remove unsupported propTypes

hiaselhans:
these props weren't included in the interface and setting them resulted in a:

```
Warning: React does not recognize the `hasShow` prop on a DOM element.
```

PR #9101, opened 2023-07-14
Fix remove unsupported propTypes

WiXSL:
Supersedes #8758


slax57 (approved):
Nice catch, thank you so much!

PR #9104, opened 2023-07-17
Add support for fullWitdh and sx prop to TranslatableInputs

djhi:


slax57 (commented):
Can you add a 'Stacked' story, demonstrating how one can use the `sx` prop to change the layout of the inputs, since this is probably one of the most common use for it?

djhi:
> Can you add a 'Stacked' story, demonstrating how one can use the `sx` prop to change the layout of the inputs, since this is probably one of the most common use for it?

So, the `sx` prop is actually not enough to allow creating a vertical stack due to the way tabs works. Unfortunately, we'll have to wait for the next minor release and wrap the tab content children in a Stack

slax57:
Okay so let's merge this already, as this fix is still relevant

slax57 (approved):

PR #9114, opened 2023-07-20
[Demo] Improve CRM Demo Kanban code

slax57:
- [x] Improve code readability by removing useless variables and renaming
- [x] extract the update logic to dedicated functions
- [x] fix index computation logic for some corner cases (while persisting the changes)

Changes inspired by the work on https://github.com/marmelab/ra-example-kanban and the (upcoming) related blog post.

### DIFF
--- a/examples/crm/src/deals/DealColumn.tsx
+++ b/examples/crm/src/deals/DealColumn.tsx
@@ -1,20 +1,16 @@
-import * as React from 'react';
-import { Box, Typography } from '@mui/material';
 import { Droppable } from '@hello-pangea/dnd';
-import { Identifier } from 'react-admin';
+import { Box, Typography } from '@mui/material';
 
+import { Deal } from '../types';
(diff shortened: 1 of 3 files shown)


djhi (approved):
:+1:

quentin-decre:
Great ! Thanks I will have a look.
Do you think PostColumn could use InfiniteList to render it's content ? Will all side-to-side infinite lists load next page when I scroll to the bottom of the body ?

djhi:
> Great ! Thanks I will have a look. Do you think PostColumn could use InfiniteList to render it's content ? Will all side-to-side infinite lists load next page when I scroll to the bottom of the body ?

Probably tricky as each column has its own scroll container.

quentin-decre:
:'(
How should we do that ?

djhi:
I have no idea, yet. Could be a nice challenge for a hackday :)

Issue #9113, opened 2023-07-19
Type mismatch for `confirmContent` on <DeleteWithConfirmButton>


elstgav:
**What you were expecting:**

The TypeScript definition for `DeleteWithConfirmButtonProps` allows `confirmContent` to be any `ReactNode`:

https://github.com/marmelab/react-admin/blob/b59f529c6d7876f82b8e5ca69fc95ec46d6ca023/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx#L97-L102

Based on this, we should expect to be able to do something like this:

```tsx
<DeleteWithConfirmButton resource="foo" confirmContent={<strong>test</strong>} />
```

**What happened instead:**

We get a prop types error, as prop types is set to only allow string:

https://github.com/marmelab/react-admin/blob/b59f529c6d7876f82b8e5ca69fc95ec46d6ca023/packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx#L118-L121

**Steps to reproduce:**

```tsx
<DeleteWithConfirmButton resource="foo" confirmContent={<strong>test</strong>} />
```

slax57:
Confirmed, thanks!

And thanks @yurassic for the fix :slightly_smiling_face:

slax57:
Fixed via https://github.com/marmelab/react-admin/pull/9115

Issue #9105, opened 2023-07-18
DeleteWithConfirmButton for ConfirmTitle accept only string


yurassic:
**Is your feature request related to a problem? Please describe.**
We need to add data-attribute for title and use existing field as title value (we want to show full name value in the title), but it only accepts string.

**Describe the solution you'd like**
Add ability to accept react components as well. So it's also brings ability to customize title.

**Describe alternatives you've considered**
-

**Additional context**
<img width="396" alt="Screenshot 2023-07-18 at 15 44 52" src="https://github.com/marmelab/react-admin/assets/10263616/e4a37f6f-1f44-4ed5-a5c8-38ce7c78fdad">

djhi:
Sounds reasonable. Can you open a PR for this ? Make sure to target the `next` branch and to add a story :)

yurassic:
@djhi Hey, I prepared PR but I couldn't push it, can you provide access for me?
Or provide instructions on how to do it.

slax57:
Fixed via https://github.com/marmelab/react-admin/pull/9115

PR #9115, opened 2023-07-20
Improve confirmTitle type

yurassic:
All Buttons with confirmation Dialog for `confirmTitle` prop accepts only `string` type, which make not possible for customization.
In this PR I changed type to `React.ReactNode`, also I fixed PropTypes value for `confirmContent` as it was conflicted with TS types and errored in App console.

This PR resolves these issues: https://github.com/marmelab/react-admin/issues/9113, https://github.com/marmelab/react-admin/issues/9105


djhi (approved):
Nice! Thanks!

packages/ra-ui-materialui/src/layout/Confirm.tsx
```diff
@@ -126,7 +128,7 @@ export interface ConfirmProps extends Omit<DialogProps, 'open' | 'onClose'> {
     loading?: boolean;
     onClose: MouseEventHandler;
     onConfirm: MouseEventHandler;
-    title: string;
+    title: React.ReactNode;
```
slax57, line 131: You need to omit `'title'` from `DialogProps` too I believe (cause  `'title'` in `DialogProps`  only supports `string`)

slax57 (changes requested):
Other than that it looks good, thanks :+1:

packages/ra-ui-materialui/src/layout/Confirm.tsx
yurassic: `DialogProps` not have `title` prop, because it uses component for it `<DialogTitle />`. 
As well I don't see it in types.

yurassic: Hm, I don't see `title` type there and in Modal types too 🤷🏻‍♂️ 
By the way I going to try omit it and will see if it helps.

slax57 (approved):
Thanks

PR #9117, opened 2023-07-21
Call checkError when getPermissions fails

djhi:
## Problem

When declaring resources at runtime, we don't check for authorization related errors and as no resources were returned, RA displays an empty page.

## Solution

When `getPermissions` rejects, pass the error to `checkError` to allow redirecting to the login page

### DIFF
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -1,6 +1,7 @@
 import { useMemo } from 'react';
 import { useQuery, UseQueryOptions } from 'react-query';
 import useAuthProvider from './useAuthProvider';
+import useLogoutIfAccessDenied from './useLogoutIfAccessDenied';
 
 const emptyParams = {};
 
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -41,13 +42,19 @@ const usePermissions = <Permissions = any, Error = any>(
...
         ['auth', 'getPermissions', params],
         authProvider
             ? () => authProvider.getPermissions(params)
             : async () => [],
-        queryParams
+        {
+            ...queryParams,
+            onError: error => {
+                logoutIfAccessDenied(error);
+            },
```
slax57, line 56: You should call original `onError` in this provided `onError`.

Besides, I recommend logging the error in dev mode, to help the developer understand what is happening, just like we do here: https://github.com/marmelab/react-admin/blob/next/packages/ra-core/src/dataProvider/useDataProvider.ts#L114

slax57 (changes requested):

packages/ra-core/src/auth/usePermissions.ts
djhi: Actually it would be a breaking change to not allow overriding the `onError`

slax57 (approved):

PR #9119, opened 2023-07-21
Handle error for values of invalid data types in DateField

qbantek:
This PR proposes changes to handle the scenario when the DateField value does not match the type criteria to be handled as a Date.

Current behavior causes an error when executing `date.toLocaleString();` on an undefined date value.

### DIFF
--- a/packages/ra-ui-materialui/src/field/DateField.tsx
+++ b/packages/ra-ui-materialui/src/field/DateField.tsx
@@ -61,7 +61,18 @@ const DateFieldImpl = <
     }
 
     const value = get(record, source);
-    if (value == null || value === '') {
+    const parsedDate = (value: any) => {
+        if (value == null || value === '') return undefined;
+
+        return value instanceof Date
+            ? value
+            : typeof value === 'string' || typeof value === 'number'
+            ? new Date(value)
+            : undefined;
+    };
+    const date = parsedDate(value);
+
+    if (typeof date === 'undefined') {
         return emptyText ? (
             <Typography
                 component="span"
@@ -73,13 +84,6 @@ const DateFieldImpl = <
             </Typography>
         ) : null;
     }
-
-    const date =
-        value instanceof Date
-            ? value
-            : typeof value === 'string' || typeof value === 'number'
-            ? new Date(value)
-            : undefined;
 
     let dateOptions = options;
     if (
--- a/packages/ra-ui-materialui/src/field/DateField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/DateField.spec.tsx
@@ -37,6 +37,13 @@ describe('<DateField />', () => {
         expect(container.firstChild).toBeNull();
     });
 
+    it('should return null when the record has an invalid type for the source', () => {
+        const { container } = render(
+            <DateField record={{ id: 123, foo: {} }} source="foo" />
+        );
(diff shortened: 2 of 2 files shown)


packages/ra-ui-materialui/src/field/DateField.tsx
```diff
@@ -61,7 +61,18 @@ const DateFieldImpl = <
     }
 
     const value = get(record, source);
-    if (value == null || value === '') {
+    const parsedDate = (value: any) => {
```
fzaninotto, line 64: I doin't understand why you extracted to a function just to call it afterwards.

packages/ra-ui-materialui/src/field/DateField.tsx
```diff
@@ -61,7 +61,18 @@ const DateFieldImpl = <
...
+
+        return value instanceof Date
+            ? value
+            : typeof value === 'string' || typeof value === 'number'
+            ? new Date(value)
+            : undefined;
+    };
+    const date = parsedDate(value);
+
+    if (typeof date === 'undefined') {
```
fzaninotto, line 75: Please keep the `value == null` test instead.

fzaninotto (changes requested):
Thanks for the PR. 

This PR raises an important question: when the value is invalid, should we display the empty text, or a new "invalid value" message? Because the problem also occurs in other fields, and react-admin needs to be consistent.

@djhi @slax57 any feedback on this question?

packages/ra-ui-materialui/src/field/DateField.tsx
qbantek: I did it to encapsulate the logic for validating whether the value can represent a Data or not. Probably a matter of style and personal preference, I can change it to inline code obviously.

packages/ra-ui-materialui/src/field/DateField.tsx
qbantek: Not sure if I understood this request properly: Are you asking to have BOTH `value == null` and `typeof date === 'undefined'` or to get rid if the `undefined` criteria?

The way I see it the null check is handled before (in `parsedDate` function) and the undefined check is needed to avoid the error, kinda the whole point of this PR. Please correct me if I am wrong.

slax57:
@fzaninotto That's a good question indeed. I guess it can't hurt to introduce a new "invalid value" message, which will help both the developer (to see that there is an issue with the data) and the user (to avoid having a runtime error and a blank value).

fzaninotto:
In all the react-admin fields, only two render a different output in case of error: ReferenceField and ReferenceManyCount. 

I think it's a good thing to show to the user that the value is invalid, so I'd recommend we do the same for invalid dates in DateField. In that case, the field could render the same ErrorIcon as ReferenceField.

packages/ra-ui-materialui/src/field/DateField.tsx
fzaninotto: yes, please inline this code

packages/ra-ui-materialui/src/field/DateField.tsx
fzaninotto: No you're right, the check is already made in parsedDate, you can keep your code as is for this part.

djhi:
@qbantek Are you still working on this?

Issue #9125, opened 2023-07-24
FileField doesn't stop click propagation to the Datagrid


SimonMulquin:
**What you were expecting:**
I added a FileField in a Datagrid component, the field display a list of files links (using the format files: [{src, title,...}]. I expect I can click a link and come back to the Datagrid afterward.

**What happened instead:**
Everything works as expected until I click the link. My file shows up in a new tab but when I come back to the previous tab, I see the Edit component instead (The click probably reached the Datagrid rowClick function which is unexpected in the case of a clickable child.

**Steps to reproduce:**

1. Add this field in a Datagrid:

`<FileField
        source="files"
        src="src"
        title="title"
        target="_blank"
      />`
      

2.  Add a property "files" to the resource:

 `"files": [
        {
          "src": "https://.../file",
          "title": "Whatever"
        },
        {
          "src": "https://.../file",
          "title": "Whatever"
        }
      ]`
 

3. Click one of the links in the Datagrid

**Related code:**
See above

**Environment**

* React-admin version: 4.12.2
* Last version that did not exhibit the issue (if applicable): n/a
* React version: 18.2.0
* Browser: Chrome Version 114.0.5735.198 (Official Build) (arm64)
* Stack trace (in case of a JS error): n/a

SimonMulquin:
I believe you can fix this by editing line 70 of FileField 

`<li key={index} onClick={e => e.stopPropagation()}>`

slax57:
Reproduced, thanks!

I also agree with the fix. I'll open a PR. Thanks.

PR #9133, opened 2023-07-25
Fix `<FileField>` should not propagate click

slax57:
Fix https://github.com/marmelab/react-admin/issues/9125

### DIFF
--- a/packages/ra-ui-materialui/src/field/FileField.tsx
+++ b/packages/ra-ui-materialui/src/field/FileField.tsx
@@ -67,7 +67,7 @@ export const FileField = <
                     const srcValue = get(file, src) || title;
 
                     return (
-                        <li key={index}>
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -67,7 +67,7 @@ export const FileField = <
                     const srcValue = get(file, src) || title;
 
                     return (
-                        <li key={index}>
+                        <li key={index} onClick={e => e.stopPropagation()}>
```
fzaninotto, line 70: Why did you put the event handler on the `li` rather than the `Link` element?

slax57: Indeed putting it on the `Link` element works just as well and makes more sense, nice catch.

fzaninotto (approved):

PR #9135, opened 2023-07-26
[Doc] Fix deprecated bulkActionButtons prop of List component

smeng9:
The bulkActionButtons prop is deprecated on both List and ListView https://github.com/marmelab/react-admin/blob/0155a34f7d70535b21a385e5c0b18370ed071e64/packages/ra-ui-materialui/src/list/ListView.tsx#L124

Update the doc to pass the prop to Datagrid component.


fzaninotto (approved):

fzaninotto:
Thanks!

Issue #9129, opened 2023-07-24
allow customization of resource paths


ns-vpanfilov:
We use dynamic resource registration using `useResourceDefinitionContext` (see https://github.com/marmelab/react-admin/issues/7623), and often register resources under custom routes to be able to display multiple tables on the same page, or render lists inside tabbed pages.

Any customization of routes, however, requires heavy customization of existing RA components, since native features utilize `useCreatePath` function to compute routes.

**_Question: Will you accept a pull request that that adds ability for users to customize useCreatePath()  so that native RA features work properly when resources are registered on non-standard routes? Something like the code below_**

_Another alternative is to store callback function inside a context, and return the value from the context if exists; otherwise, just return the one that is defined by useMemo. I can create a pull request that uses context, if you prefer that approach._

```ts
export interface ResourceOptions {
    label?: string;
    listPathOverride?: string; // added
    createPathOverride?: string;  // added
    editPathOverride?: string;  // added
    showPathOverride?: string;  // added
}


export const useCreatePath = () => {
    const basename = useBasename();
    const definitions = useResourceDefinitions();  // added
    return useCallback(
        ({ resource, id, type }: CreatePathParams): string => {
            let pathOverride: string | undefined;  // added
            switch (type) {
                case 'list':
                    pathOverride = definitions?.[resource]?.options?.listPathOverride; // added; used in the next line
                    return removeDoubleSlashes(
                        `${basename}/${pathOverride ?? resource}` 
                    );
                case 'create':
                    pathOverride = definitions?.[resource]?.options?.createPathOverride;  // added; used in the next line
                    return removeDoubleSlashes(
                        `${basename}/${pathOverride ?? resource}/create` 
                    );
                case 'edit': {
                    if (id == null) {
                        // maybe the id isn't defined yet
                        // instead of throwing an error, fallback to list link
                        pathOverride =definitions?.[resource]?.options?.listPathOverride;  // added; used in the next line
                        return removeDoubleSlashes(
                            `${basename}/${pathOverride ?? resource}`
                        );
                    }
                    pathOverride = definitions?.[resource]?.options?.editPathOverride;  // added; used in the next line
                    return removeDoubleSlashes(
                        `${basename}/${
                            pathOverride ?? resource
                        }/${encodeURIComponent(id)}`
                    );
                }
                case 'show': {
                    if (id == null) {
                        // maybe the id isn't defined yet
                        // instead of throwing an error, fallback to list link
[...]

PR #9136, opened 2023-07-27
Add ability to override useCreatePath (DO NOT MERGE)

ns-vpanfilov:
As discussed at https://github.com/marmelab/react-admin/issues/9129, I've implemented proposed changes that allow customizing useCreatePath return values
- packages/ra-core/src/types.ts
- packages/ra-core/src/routing/useCreatePath.ts

For demo and testing purposes, I've updated a simple demo application to add a `ResourceWithReg` component; was based on `Resource` component in React-Admin with additional features:
- It can register a resource during runtime (when accessed from a custom route)
- It will update registration if a resource is already registered from a different path
- It provides different modes of route registration, including:
  - `default` - similar to react-admin; it will register `create/*`,  `:id/show/*`, `:id/*` and `/*` routes inside a route if was called from;
  - `resourceWithParent` - if will register  `<resource>/create/*`,  `<resource>/:id/show/*`, `<resource>/:id/*` and `<resource>/*`  routes inside a route if was called from, plus will register list component to  `/*`; this is useful for displaying multiple lists from the same page
  
  Examples in simple demo:
  
  # custom route /tabs/1
Will register a `posts` resource in `default` mode inside `/tabs/1` route:
- `http://localhost:8080/#/tabs/1` will show list component; notice that "create" button will redirect to proper `create` route
- `http://localhost:8080/#/tabs/1/create` will show 'create' component;

  # custom route /tabs/2
  Will register `posts` and `comments` resources inside `/tabs/2` route:
  - both lists will be visible in page root `http://localhost:8080/#/tabs/2`
  - create routes are `http://localhost:8080/#/tabs/2/posts/create` and `http://localhost:8080/#/tabs/2/comments/create`
  - list component will also be visible from `http://localhost:8080/#/tabs/2/posts` and `http://localhost:8080/#/tabs/2/comments` URLs; however, there pages will only show resources that they own
  - when a resource is updated, the page is redirected to `/tabs/2` that shows both lists
  - limitation: notice that for multi-list pages, we need to disable sync with location; otherwise both resources will be impacted
  
My ask is, can I prepare a pull request that includes changes to the following files? I will remove updates to "simple demo" application, and will also remove `routeType` property - it is not required in ra-core, and only will be used in user land (e.g. `resourceWithParent` component)

- packages/ra-core/src/types.ts
- packages/ra-core/src/routing/useCreatePath.ts

### DIFF
--- a/examples/simple/src/ResourceWithReg.tsx
+++ b/examples/simple/src/ResourceWithReg.tsx
@@ -1,0 +1,270 @@
+import React from 'react';
+
+import { useEffect, ComponentType, ReactElement, isValidElement } from 'react';
+import { isValidElementType } from 'react-is';
+
+import { Route, Routes, useResolvedPath } from 'react-router-dom';
+import {
+    useResourceDefinitionContext,
+    ResourceDefinition,
+    ResourceProps,
+    useBasename,
+    ResourceContextProvider,
+    removeDoubleSlashes,
+    CreatePathParams,
+} from 'react-admin';
+
+const registerResource = ({
+    create,
+    edit,
+    icon,
+    list,
+    name,
+    options,
+    show,
+    recordRepresentation,
+    hasCreate,
+    hasEdit,
+    hasShow,
+    createPath,
+    routeType,
+}: ResourceProps) => ({
+    name,
+    options: options ?? {},
+    hasList: !!list,
+    hasCreate: !!create || !!hasCreate,
+    hasEdit: !!edit || !!hasEdit,
+    hasShow: !!show || !!hasShow,
+    icon,
+    recordRepresentation,
+    createPath,
+    routeType,
+});
+
+export const getElement = (
+    ElementOrComponent: ComponentType<any> | ReactElement
+) => {
+    if (isValidElement(ElementOrComponent)) {
+        return ElementOrComponent;
+    }
+
+    if (isValidElementType(ElementOrComponent)) {
+        return <ElementOrComponent />;
+    }
+
+    return <>{null}</>;
+};
+
+export function ResourceWithReg(props: ResourceProps) {
+    const basename = useBasename();
+
+    const registrationProps = registerResource(props);
+
+    const { create, edit, list, show } = props;
+
+    const resolved = useResolvedPath('');
+
+    const {
+        name,
+        icon,
+        options,
+        hasCreate,
+        hasEdit,
+        hasShow,
+        hasList,
+        recordRepresentation,
+        createPath,
+        routeType,
+    } = registrationProps;
+
+    const {
+        register,
+        unregister,
+        definitions,
+    } = useResourceDefinitionContext();
+
+    const def: ResourceDefinition = definitions[name];
+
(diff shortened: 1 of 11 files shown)


fzaninotto:
On thing I don't understand, is how you'll get the `resource` and `id` route parameters from CRUD pages if the route path isn't the standard one. I don't see it in your example - only `disableSyncWithLocation`, which seems like a regression in that case. Can you elaborate?

ns-vpanfilov:
> On thing I don't understand, is how you'll get the `resource` and `id` route parameters from CRUD pages if the route path isn't the standard one. I don't see it in your example - only `disableSyncWithLocation`, which seems like a regression in that case. Can you elaborate?

@fzaninotto,

When I define custom paths, I always wrap components in `<ResourceContextProvider>` so that resource name is always available. See `ResourceWithReg.tsx`

As to reading IDs, I've have not experienced any issues since I kept paths consistent with RA, so react components read IDs properly. I have not experienced any issues with reading IDs before using, for example, `useRecordContext`. I just updated `PostEdit` page to output a record into console.log and it works just fine. (e.g. http://localhost:8080/#/tabs/2/posts/13)

There are 2 use cases that I have for modifying paths, all were consistent with RA's approach to paths.

**Use case 1  - resource is directly added to a tab**
- Create: `<basename>/<resource>/create/*` --> `<tab_name>/<tab_id>/create/*`
- Show: `<basename>/<resource>/:id/show/*` --> `<tab_name>/<tab_id>/:id/show/*`
- Edit: `<basename>/<resource>/:id/*` --> `<tab_name>/<tab_id>/:id/*`
- List: `<basename>/<resource>/*` --> `<tab_name>/<tab_id>/*`

**Use case 2 - a page with multiple lists on the same page**
- Create: `<basename>/<resource>/create/*` --> `<path_to_page>/<resource>/create/*`
- Show: `<basename>/<resource>/:id/show/*` --> `<path_to_page>/<resource>>/:id/show/*`
- Edit: `<basename>/<resource>/:id/*` --> `<path_to_page>/<resource>/:id/*`
- List (only show list for a resource in question): `<basename>/<resource>/*` --> `<path_to_page>/<resource>/*`
- List (shows all lists): `<basename>/<resource>/*` --> `<path_to_page>/*`

ns-vpanfilov:
@fzaninotto , I see that react-admin-storybook fails. When I click "Details", I see a 404 Error message page; sounds like I do not have access and cannot see error messages.

fzaninotto:
> @fzaninotto , I see that react-admin-storybook fails. When I click "Details", I see a 404 Error message page; sounds like I do not have access and cannot see error messages.

The error says

> The lockfile would have been modified by this install, which is explicitly forbidden.

fzaninotto:
I spent the last 15 minutes looking at your code and I can't make any sense of it. I'm sorry, I'm probably dumb, but I won't comment more on this PR which is too messy.

If you feel like adding support for a custom createPath function to react-admin, please do a PR that actually does that, with a clear implementation, unit tests and documentation. We'll see if we merge it based on the result.

PR #9139, opened 2023-07-28
[Doc] Fix default perPage value of useList

smeng9:
https://github.com/marmelab/react-admin/blob/f568a52d2b20f8c7e4a1c91155d197cb28bb5df3/packages/ra-core/src/controller/list/useList.ts#L64


fzaninotto (approved):

fzaninotto:
Well spotted, thanks!

PR #9140, opened 2023-07-29
[Doc] Fix confirm and cancel icon type

smeng9:
The declared interface here 

https://github.com/marmelab/react-admin/blob/f568a52d2b20f8c7e4a1c91155d197cb28bb5df3/packages/ra-ui-materialui/src/layout/Confirm.tsx#L122

expects component for these prop


fzaninotto (approved):

fzaninotto:
Thanks!

PR #9141, opened 2023-07-29
feat(datagrid): add support to select non selectable rows using bulk actions

felipejh:
## Context
We have received a request to implement row selection in the Datagrid is allowed only in ascending order. Users can select rows starting from the first one and proceed sequentially to the last one. Specifically, users are permitted to select only the first row initially. Once the user selects the first row, the subsequent rows become eligible for selection, following the same sequential pattern.
Only the first row is allowed to select. When it is selected by the user, then the next row is allowed to select and so on.
When the select all button in a bulk toolbar is pressed by the user, it is expected all rows on the list are selected.
The `isRowSelectable` prop of the `Datagrid` was used to satisfy that requirement.

image

## Problem
When the select all button in a bulk toolbar is pressed, only the selectable rows are selected. In this case, only the first row. Then, the second row is allowed to be selectable and the select all button in the bulk toolbar is still not marked.

- CodeSandbox: https://codesandbox.io/p/sandbox/github/felipejh/ra-datagrid-bulk-example/tree/main

demo

## Solution
Add a support to the `Datagrid` component to allow select non selectable rows using the select all bulk toolbar.
By adding an optional prop called `bulkIgnoreNonSelectableRows` to the Datagrid, we can allow users to disable the filtering of non-selectable rows.
I am unsure if that prop name is the best. Your input and ideas are welcome.

While I understand that this is a very specific use case, and the aim is to maintain the react-admin codebase small and manageable, I strongly believe that implementing this feature would be highly beneficial. It's worth noting that this feature will be optional, providing flexibility to users who may not need it while enhancing the experience for those who do.

This would be the result:
ignoreUnselectable


viniciusbsneto:
Nice PR! I think it would be beneficial to have this optional feature available.

PR #9149, opened 2023-07-31
Fix FilterLiveSearch reset button does not reset the value

### LINKED ISSUE #9106: FilterLiveSearch does not clear its value
**What you were expecting:**
Clearing the FilterLiveSearch input using its clear button actually clears the value

**What happened instead:**
It clears but immediately reapply the value.

**Steps to reproduce:**
1. Enter `Fusce` in the live search input
2. Click the only item of the list
3. Go back to the list (using sidebar menu but it would be the same using browser history)
4. Try clearing the live search input using the clear icon

**Related code:**
https://stackblitz.com/edit/github-zb5dpt

**Environment**

* React-admin version: 4.12.1
* React version: 17
* Browser:

fzaninotto:
Closes #9106

FilterLiveSearch used react-admin's `<Form>`, which did too many things and caused the bug. Switching to a simple form powered by `useForm` fixes the issue. 

I had to augment `useList` to make a real-life filterable list and unit test `<FilterLiveSearch>` .

### DIFF
--- a/packages/ra-core/src/controller/list/useList.spec.tsx
+++ b/packages/ra-core/src/controller/list/useList.spec.tsx
@@ -21,70 +21,6 @@ const UseList = ({
 };
 
 describe('<useList />', () => {
-    it('should filter string data based on the filter props', () => {
-        const callback = jest.fn();
-        const data = [
-            { id: 1, title: 'hello' },
-            { id: 2, title: 'world' },
-        ];
-
-        render(
-            <UseList
-                data={data}
-                filter={{ title: 'world' }}
-                sort={{ field: 'id', order: 'ASC' }}
-                callback={callback}
-            />
-        );
-
-        expect(callback).toHaveBeenCalledWith(
-            expect.objectContaining({
-                sort: { field: 'id', order: 'ASC' },
-                isFetching: false,
-                isLoading: false,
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/controller/list/useList.spec.tsx
```diff
@@ -21,70 +21,6 @@ const UseList = ({
 };
 
 describe('<useList />', () => {
-    it('should filter string data based on the filter props', () => {
```
fzaninotto, line 24: I moved these tests further down to group them in the `describe('filter')` group, I didn't remove or modify them

packages/ra-core/src/controller/list/useList.spec.tsx
```diff
@@ -229,66 +165,154 @@ describe('<useList />', () => {
...
+                    isFetching: false,
+                    isLoading: false,
+                    data: [{ id: 2, title: { name: 'world' } }],
+                    error: undefined,
+                    total: 1,
+                })
+            );
+        });
+
+        it('should apply the q filter as a full-text filter', () => {
```
fzaninotto, line 296: This is the only new test

adguernier (approved):

PR #9159, opened 2023-08-02
fix NumberInput setFocus (react-hook-form)

XStepien:
React-hook-form setFocus doesn't work with NumberInput because the ref didn't go through the right props.


fzaninotto:
Thanks!

Can you explain how to reproduce the bug. Even better: can you write a unit test that doesn't pass on master, but does in your branch?

Issue #9155, opened 2023-08-02
Some Inputs with choices in their props do not accept `ReadonlyArray<T>`.


na2na-p:
**What you were expecting:**

`SelectInput` choices does accept `ReadonlyArray<T>`

**What happened instead:**

`SelectInput` choices does not accept `ReadonlyArray<T>`

![image](https://github.com/marmelab/react-admin/assets/49822810/bf8e2e53-6020-455e-a6a1-0de2379aef7e)

**Steps to reproduce:**

- use `"react-admin": "^4.12.3"`

**Related code:**

* Preferably, a sandbox forked from 

- L129
- L151

https://stackblitz.com/edit/github-h13gjc?file=src%2Fposts%2FPostEdit.tsx

* A link to a GitHub repo with the minimal codebase to reproduce the issue

```
import { SelectInput } from 'react-admin';

type Choice = {
  id: number;
  name: string;
};

const mutableArray = [
  {
    id: 1,
    name: 'foo',
  },
  {
    id: 2,
    name: 'bar',
  },
  {
    id: 3,
    name: 'baz',
  },
] satisfies Array<Choice>;

const readonlyArray = [
  {
    id: 1,
    name: 'foo',
  },
  {
    id: 2,
    name: 'bar',
  },
  {
    id: 3,
    name: 'baz',
  },
] as const satisfies ReadonlyArray<Choice>;

export const SelectInputGroup = () => {
  return (
    <>
      <SelectInput source="mutableChoices" choices={mutableArray} />
      <SelectInput source="mutableChoices" choices={readonlyArray} /> // type error
    </>
  );
};

```

**Other information:**

suggestion(Of course, it is necessary to inspect more than just SelectInput.)
```diff
-     choices?: any[];
+     choices?: readonly any[]; // or ReadonlyArray<any>
```
https://github.com/marmelab/react-admin/blob/de8fbaa3285714aab914e0e76cc369c09a817d3c/packages/ra-core/src/form/useChoices.tsx#L16

**Environment**

* React-admin version: 4.12.3
* Last version that did not exhibit the issue (if applicable):
* React version: 18.2.0
* Browser: Google Chrome 115.0.5790.114 (Official Build) （arm64）  
* Stack trace (in case of a JS error):

fzaninotto:
Thanks for the report. 

[...]

PR #9162, opened 2023-08-03
RA-OSS-20 - Some Inputs with choices in their props do not accept ReadonlyArray<T>

gitstart:
[ISSUE](https://github.com/marmelab/react-admin/issues/9155)

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


slax57:
Thanks for this updated version.

I must admit I'm a bit reluctant to accept this solution, because of all the `as any[]` it requires.

An alternative solution, that would be cleaner IMO, would be to use generic types in this manner:

```ts
interface ChoicesProps <T = any[]> {
    choices?: T extends Array<any> ? T : T extends ReadonlyArray<any> ? T : "must be Array or ReadonlyArray";
}
```

However this would also require propagating the type to inner layers of the code. So it may lead to many changed files...

You can give it a try if you will, but I must remind that, for the reasons explained [here](https://github.com/marmelab/react-admin/issues/9155#issuecomment-1663955011), and because it is so easy to fix in user-land, we are not willing to invest too much time on resolving this issue in case it starts becoming too complex.

gitstart:
@slax57 we tried another fix can you check it out.

djhi:
Sorry it took us so long to review this. This looks good to me. Could you merge the latest changes from `master`? Thanks!

slax57 (changes requested):
This PR does not seem to fix the issue it's supposed to address.

slax57 (approved):
UPDATE: my bad, I forgot to build ra-core first

PR #9165, opened 2023-08-07
Feature: prev next button for detail and edit views

adguernier:
## Problem

Some users want to navigate to the previous and next item in a list without going back to the list.

## Solution

Create a ListIdsContextProvider, which fetches all the list and stores only an array of ids in it. It can take optional filters and sort params. If not given, it uses the params from the store (${resource}.listParams), or the default ones.

Create a prev/next component. It creates a ListIdsContextProvider. Using the RecordContext, it looks up the current record id in the list of ids, determines the previous and next id, and forges the link to the next page (either a show or an edit page, depending on the current route)

## Todo
- [x] ~~add context and contextProvider to fetch ids according to the filters and sort parameters~~
- [x] ~~add a PrevNextButton component that consume the above context and display links to navigate between resources~~
- [x] move code above into the `PrevNextButton` component and make it works
- [x] write some stories  for `PrevNextButton`
- [x] write JS DOC for `PrevNextButton`
- [x] write unit tests for `PrevNextButton`
- [x] Extract the logic in a dedicated hook called `useNextPrevController`
- [x] write `PrevNextButton` documentation
- [ ] ~~write `useNextPrevController` documentation~~


packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+            type,
+            resource,
+            id: nextId,
+        });
+        navigate(link);
+    };
+
+    return (
+        <>
+            <IconButton
```
fzaninotto, line 46: you should use react-router-dom's `Link` component as a `component` prop of this button instead of relying on `onClick`. The difference is that the target URL is visible on hover, which is something we definitely want.

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+                disabled={!previousId}
+                aria-label="previous"
```
fzaninotto, line 48: use translation

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const PrevNextButton = (props: PrevNextButtonProps) => {
+    const { linkType = 'edit' } = props;
+    const translate = useTranslate();
+    const recordId = useGetRecordId();
+    const resource = useResourceContext();
+    const createPath = useCreatePath();
+    const [query] = useListParams({
+        resource,
+    });
```
fzaninotto, line 22: `useListParams` does more than you need: it merges params from the URL, the store, and the props. You only need the last two, so you must do the merge manually.

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const list = useGetList(resource, {
+        pagination: { page: query.page, perPage: query.perPage },
+        sort: { field: query.sort, order: query.order },
+        filter: { ...query.filter },
+    });
```
fzaninotto, line 27: you probably need to add a specific cache for this query (configurable via props, of course), to avoid that the `getList` call repeats on every page.

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const { linkType = 'edit' } = props;
+    const translate = useTranslate();
+    const recordId = useGetRecordId();
+    const resource = useResourceContext();
+    const createPath = useCreatePath();
+    const [query] = useListParams({
+        resource,
+    });
+    const list = useGetList(resource, {
+        pagination: { page: query.page, perPage: query.perPage },
```
fzaninotto, line 24: The pagination shouldn't depend on the params - you always want to fetch * all* records. You may want to have a `perPage` prop fixed at an arbitrary level though (e.g. 1000) to avoid exploding the app when there are too many records.

fzaninotto (changes requested):
Yep, that's the right direction!

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
adguernier: So I guess I need to remove `useListParams` in favor of `useStore` and probably another hook?

fzaninotto: exactly

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    const { isLoading, data, isError, error } = useGetList(
+        resource,
+        {
+            sort: {
+                ...{ field: storedParams.sort, order: storedParams.order },
+                ...{ field: listParams.sort, order: listParams.order },
+            },
+            filter: { ...storedParams.filter, ...listParams.filter },
+            ...(limit ? { pagination: { page: 1, perPage: limit } } : {}),
```
fzaninotto, line 43: I think the pagination is compulsory. I wouldn't support the case where there is no limit. If people want an unlimited request, they should set an arbitrary large limit.

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
```diff
@@ -0,0 +1,161 @@
...
+        if (nextId !== null) {
+            nextProps.disabled = false;
+            nextProps.to = createPath({
+                type: linkType,
+                resource,
+                id: nextId,
+            });
+            nextProps.component = Link;
+        }
+    }
```
fzaninotto, line 100: I find it a bit hard to read. Can't you do that in JSX? You may need sub components.

packages/ra-ui-materialui/src/button/PrevNextButton.spec.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { Basic } from './PrevNextButton.stories';
+describe('<PrevNextButton />', () => {
+    beforeEach(() => {
+        window.scrollTo = jest.fn();
+    });
+    it('should render next prev button at position 4 with a total of 905', async () => {
```
fzaninotto, line 8: You should describe the feature you test regardless of the test data. Something like: "It should render the position of the current record in the list"

packages/ra-ui-materialui/src/button/PrevNextButton.tsx
fzaninotto, line 1: As there are 2 buttons, I'd call the component "PrevNextButtons", plural.

packages/ra-ui-materialui/src/button/PrevNextButton.stories.tsx
```diff
@@ -0,0 +1,320 @@
...
+    <TopToolbar
+        sx={{
+            justifyContent: 'space-between',
+        }}
+    >
+        {children}
+    </TopToolbar>
+);
+
+const defaultFiedls = [
```
fzaninotto, line 97: typo

packages/ra-ui-materialui/src/button/PrevNextButton.stories.tsx
```diff
@@ -0,0 +1,320 @@
...
+                    edit={
+                        <Edit
+                            actions={
+                                <MyTopToolbar>
+                                    <PrevNextButton />
+                                    <ShowButton />
+                                </MyTopToolbar>
+                            }
+                        >
+                            <SimpleForm>{defaultInputs}</SimpleForm>
```
fzaninotto, line 132: instead of extracting an array of inputs, extract a custom form component. That's more React-y.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,276 @@
...
+                fontSize="small"
+                titleAccess="error"
+                aria-errormessage={error.message}
+            />
+        );
+    }
+
+    const ids = data ? data.map(record => record.id) : [];
+
+    const index = ids.indexOf(record.id);
```
djhi, line 150: Unfortunately, this would not allow users to implement SaveAndGoToNext and SaveAndGoToPrev buttons.

I would extract the logic inside a `useNextPrevController` that would return an object with the following props:
- `hasNext`
- `hasPrev`
- `navigateToNext`
- `navigateToPrev`

Even better would be to also provide a `PrevNextContext` and a `PrevNextBase` component

djhi (changes requested):

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
adguernier: I followed your recommendations and extract the logic inside `useNextPrevController`. Hope it's better now. 

Next step a `PrevNextContext` and a `PrevNextBase`

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+title: "usePrevNextController"
+---
+
+# `usePrevNextController`
+
+`useShowController` fetche previous and next record IDs for a given record and resource.
+
+It fetches the list of records from the REST API according to the filters and the sort order of listsfrom the store and also merges the filters and the sorting order passed into props.
+
+It used by `<PrevNextButtons>` to render a navigation to move to the next or previous record of a resource.
```
djhi, line 12:
```suggestion
It is used by the `<PrevNextButtons>` component to render navigation buttons that allow users to navigate to the next or previous record of a resource.
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+                    >
+                        <ArrowForwardIcon />
+                    </IconButton>
+                </li>
+            </ul>
+        </nav>
+    );
+};
+
+interface MyPrevNextButtonProps<RecordType extends RaRecord = any> extends UsePrevNextControllerProps<RecordType>
```
djhi, line 100:
```suggestion
interface MyPrevNextButtonProps<RecordType extends RaRecord = any> extends UsePrevNextControllerProps<RecordType> {}
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+| `linkType`     | Optional | `string`         | 'edit'                              | Specifies the view to redirect to when navigating.                                          |
+| `queryOptions` | Optional | `object`         | `{ staleTime: 5 * 60 * 1000 }`      | The options to pass to the useQuery hook.                                                   |
+| `resource`     | Optional | `string`         | -                                   | The resource name, e.g. `customers`.                                                        |
+| `sort`         | Optional | `object`         | `{ field: 'id', order: SORT_ASC } ` | The initial sort parameters.                                                                |
+| `storeKey`     | Optional | `string | false` | -                                   | The key to use to match a filter & sort configuration of a `<List>`. Pass false to disable. |
+
+## `filter`
+
+Just like [Permanent `filter` in `<List>`](./List.md#filter-permanent-filter), you can specify what filter to apply to fetches the list of records.
```
djhi, line 118:
```suggestion
Just like [Permanent `filter` in `<List>`](./List.md#filter-permanent-filter), you can specify what filters to apply when fetching the list of records.
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+
+## `limit`
+
+You can set the maximum number of records to fetch with `limit` prop. By default, `usePrevNextController` fetches a maximum of `1000` records.
+
+## `linkType`
+
+By default `<usePrevNextController>` generates links to `<Edit>` view. You can also set the `linkType` prop to `show` to link to the `<Show>` view instead.
```
djhi, line 126:
```suggestion
By default `<usePrevNextController>` generates links to the `<Edit>` view. You can also set the `linkType` prop to `show` to link to the `<Show>` view instead.
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+
+## `queryOptions`
+
+`usePrevNextController` accepts a `queryOptions` prop to pass options to the react-query client. 
+
+This can be useful e.g. to pass [a custom `meta`](./Actions.md#meta-parameter) to the `dataProvider.getList()` call.
+
+## `resource`
+
+By default, `usePrevNextController` operates on the current `ResourceContext` (defined at the routing level), so under the `/customers` path, the `resource` prop will be `customers`. You may want to force a different resource to fetche. In this case, pass a custom `resource` prop, and it will override the `ResourceContext` value.
```
djhi, line 141:
```suggestion
By default, `usePrevNextController` operates on the current `ResourceContext` (defined at the routing level), so under the `/customers` path, the `resource` prop will be `customers`. Pass a custom `resource` prop to override the `ResourceContext` value.
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+
+## `sort`
+
+Pass an object literal as the `sort` prop to determine `field` and `order` used for sorting:
```
djhi, line 145:
```suggestion
Pass an object literal as the `sort` prop to set the `field` and `order` used for sorting:
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+
+## `storeKey`
+
+`usePrevNextController` grabs list parameters (sort and filters) from the store.
```
djhi, line 149:
```suggestion
`usePrevNextController` gets the list parameters (sort and filters) from the store.
```

docs/usePrevNextController.md
```diff
@@ -0,0 +1,151 @@
...
+This prop is useful if you specified a custom `storeKey` for a `<List>` and you want `usePrevNextController` grabs the same stored parmaters.
```
djhi, line 150:
```suggestion
This prop is useful if you specified a custom `storeKey` for a `<List>` and you want `usePrevNextController` to use the same stored parameters.
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+import { useResourceContext } from '../core';
+import { useGetList } from '../dataProvider';
+import { useStore } from '../store';
+import { FilterPayload, RaRecord, SortPayload } from '../types';
+import { ListParams, SORT_ASC } from './list';
+import { useRecordContext } from './record';
+import { useCreatePath } from '../routing';
+
+/**
+ * A hook used to fetche previous and next record IDs for a given record and resource.
```
djhi, line 11:
```suggestion
 * A hook used to fetch the previous and next record identifiers for a given record and resource.
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * It fetches the list of records from the REST API according to the filters
```
djhi, line 13:
```suggestion
 * It fetches the list of records according to the filters
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+ * and the sort order configured in the lists by the users and
```
djhi, line 14:
```suggestion
 * and the sort order configured in the list and
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+ * also merges the filters and the sorting order passed into props.
```
djhi, line 15:
```suggestion
 * merges the filters and the sorting order passed into props.
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * `usePrevNextController` should be used anywhere a record context is provided
+ * (eg: often inside a `<Show>` or `<Edit>` component).
+ *
+ * @example <captoin>Simple usage</caption>
```
djhi, line 20:
```suggestion
 * @example <caption>Simple usage</caption>
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+            order: sort.order,
+            sort: sort.field,
+        }
+    );
+
+    const { data, error, isLoading } = useGetList<RecordType>(
+        resource,
+        {
+            sort: {
+                ...{ field: storedParams.sort, order: storedParams.order },
```
djhi, line 145:
```suggestion
                field: storedParams.sort,
                order: storedParams.order,
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,220 @@
...
+            hasPreviousPage?: boolean;
+        };
+    }> & { meta?: any };
+}
+
+export interface UsePrevNextControllerResult {
+    hasPrev: boolean;
+    hasNext: boolean;
+    navigateToPrev: string | undefined;
+    navigateToNext: string | undefined;
```
djhi, line 213: Those are not functions so you should avoid using a verb in their names. I suggest `prevPath` and `nextPath`

packages/ra-ui-materialui/src/button/PrevNextButtons.spec.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
```
djhi, line 2: Eslint warning here

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+    UsePrevNextControllerProps,
+} from 'ra-core';
+import { NavigateBefore, NavigateNext } from '@mui/icons-material';
+import ErrorIcon from '@mui/icons-material/Error';
+import { Link } from 'react-router-dom';
+import { CircularProgress, IconButton, SxProps, styled } from '@mui/material';
+import clsx from 'clsx';
+
+/**
+ * A component used to render the previous and next buttons in a Show or Edit view.
```
djhi, line 15:
```suggestion
```

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * The `<PrevNextButtons>` component render a navigation to move to the next or previous record
+ * of a resource, with the current index and the total records, in an edit or show view.
+ *
+ * It uses `usePrevNextController` to fetches the list of records.
```
djhi, line 20:
```suggestion
 * It uses `usePrevNextController` to fetch the list of records.
```

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * `<PrevNextButtons>` can be used anywhere a record context is provided
+ * (eg: often inside a `<Show>` or `<Edit>` component).
+ *
+ * @example <caption>move to edit view by default</caption>
```
djhi, line 25:
```suggestion
 * @example <caption>navigate to edit view by default</caption>
```

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+ * <ShowButton />
+ *
+ * @example <caption>move to show view</caption>
+ * <PrevNextButtons linkType="show" />
+ *
+ * @example <caption>share custom storeKey with other components</caption>
```
djhi, line 31:
```suggestion
 * @example <caption>custom storeKey</caption>
```

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+ * <PrevNextButtons
+ *     linkType="show"
+ *     sort={{
+ *         field: 'first_name',
+ *         order: 'DESC',
+ *     }}
+ *     filter={{ q: 'East a' }}
+ * />
+ *
+ * @example <caption>customize its style</caption>
```
djhi, line 47:
```suggestion
 * @example <caption>customize style</caption>
```

djhi (changes requested):
Good job :+1:

packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -0,0 +1,176 @@
...
+        index,
+        total,
+        error,
+        isLoading,
+    } = usePrevNextController<RecordType>(props);
+
+    const translate = useTranslate();
+
+    if (isLoading) {
+        return <CircularProgress size={14} />;
```
fzaninotto, line 107: I'd use react-admin's `LinearProgress` instead, as it only shows the progress indicator after a 1 second delay.

packages/ra-ui-materialui/src/button/PrevNextButtons.stories.tsx
```diff
@@ -0,0 +1,366 @@
...
+                            color: 'red',
+                        },
+                    }}
+                />
+            </RecordContextProvider>
+        </ResourceContext.Provider>
+    </AdminContext>
+);
+
+export const ErrorState = () => (
```
fzaninotto, line 351: We also need a story to test the loading state, with a getList returning a Promise that never resolves.

fzaninotto (changes requested):
We're getting close!

packages/ra-ui-materialui/src/button/PrevNextButtons.stories.tsx
```diff
@@ -0,0 +1,366 @@
...
+export const Basic = () => (
+    <MemoryRouter>
+        <AdminContext dataProvider={dataProvider} i18nProvider={i18nProvider}>
+            <AdminUI>
+                <Resource
+                    name="customers"
+                    list={
+                        <List>
+                            <DefaultDataGrid />
+                        </List>
```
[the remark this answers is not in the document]
adguernier: Maybe I should use EditGuesser and ShowGuesser too ?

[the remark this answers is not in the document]
fzaninotto: No, because you want to customize the actions, and I don't believe it's possible with the guessers.

docs/PrevNextButtons.md
```diff
@@ -0,0 +1,340 @@
...
+    </Show>
+);
+```
+{% endraw %}
+
+## Navigating Through Records In`<Edit>` Views After Submit
+
+Let's says users want to edit customer records and to navigate between records in the `<Edit>` view. The default react-admin behaviors causes two problems: 
+- when their save a record the user is redirected to the `<List>` view,
+- when their navigate accross records, the form is not saved.
```
djhi, line 317:
```suggestion
- when they save a record the user is redirected to the `<List>` view,
- when they navigate to another record, the form is not saved.
```

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    const { data, error, isLoading } = useGetList<RecordType>(
+        resource,
+        {
+            sort: {
+                ...sort,
+                ...{
+                    field: storedParams.sort,
+                    order: storedParams.order,
+                },
```
djhi, line 144:
```suggestion
                field: storedParams.sort,
                order: storedParams.order,
```

djhi (changes requested):

PR #9172, opened 2023-08-10
Add ability to use SimpleList without `primaryText`

fzaninotto:
## Problem

In most cases, the `primaryText` of a SimpleList is the resource's `resordRepresentation`. If feels repetitive to have to pass an explicit value to this prop. 

## Solution

Default to `recordRepresentation` so we can do the following:

```jsx
Const PostList = () => (
  <List>
    <SimpleList />
  </List>
)
```

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -38,24 +38,24 @@ The `<Datagrid>` is an **iterator** component: it gets an array of records from 
 
 ## Props
 
-Here are all the props accepted by the component:
-
-* [`body`](#body)
-* [`bulkActionButtons`](#bulkactionbuttons)
-* [`children`](#children)
-* [`empty`](#empty)
-* [`expand`](#expand)
-* [`expandSingle`](#expandsingle) 
-* [`header`](#header)
-* [`hover`](#hover)
-* [`isRowExpandable`](#isrowexpandable)
-* [`isRowSelectable`](#isrowselectable)
-* [`optimized`](#optimized-better-performance-for-large-tables)
-* [`rowStyle`](#rowstyle)
-* [`rowSx`](#rowsx)
-* [`rowClick`](#rowclick)
-* [`size`](#size)
-* [`sx`](#sx-css-api)
+| Prop | Required | Type | Default | Description |
+| --- | --- | --- | --- | --- |
+| `children` | Required | Element | n/a | The list of `<Field>` components to render as columns. |
+| `body` | Optional | Element | `<Datagrid Body>` | The component used to render the body of the table. |
+| `bulkActionButtons` | Optional | Element | `<BulkDelete Button>` | The component used to render the bulk action buttons. |
+| `empty` | Optional | Element | `<Empty>` | The component used to render the empty table. |
+| `expand` | Optional | Element |  | The component used to render the expand panel for each row. |
+| `expandSingle` | Optional | Boolean | `false` | Whether to allow only one expanded row at a time. |
+| `header` | Optional | Element | `<Datagrid Header>` | The component used to render the table header. |
+| `hover` | Optional | Boolean | `true` | Whether to highlight the row under the mouse. |
+| `isRowExpandable` | Optional | Function | `() => true` | A function that returns whether a row is expandable. |
+| `isRowSelectable` | Optional | Function | `() => true` | A function that returns whether a row is selectable. |
+| `optimized` | Optional | Boolean | `false` | Whether to optimize the rendering of the table. |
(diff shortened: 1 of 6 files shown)


docs/SimpleList.md
```diff
@@ -37,19 +37,40 @@ export const PostList = () => (
...
-* [`rightAvatar`](#rightavatar)
-* [`rightIcon`](#righticon)
-* [`rowStyle`](#rowstyle)
-* [`rowSx`](#rowsx)
-* [`empty`](#empty)
+## Props
+
+| Prop | Required | Type | Default | Description |
+| --- | --- | --- | --- | --- |
+| `primaryText` | Optional | mixed | record representation | The primary text to display. |
```
adguernier, line 44:
```suggestion
| `primaryText` | Optional | mixed | [record representation](./Resource.html#recordrepresentation) | The primary text to display. |
```

docs/SimpleList.md
```diff
@@ -37,19 +37,40 @@ export const PostList = () => (
...
+| `secondaryText` | Optional | mixed | | The secondary text to display. |
+| `tertiaryText` | Optional | mixed | | The tertiary text to display. |
+| `linkType` | Optional |mixed | `"edit"` | The target of each item click |
```
adguernier, line 47:
```suggestion
| `linkType` | Optional |mixed | `"edit"` | The target of each item click. |
```

docs/SimpleList.md
```diff
@@ -37,19 +37,40 @@ export const PostList = () => (
...
+| `leftIcon` | Optional | function | | A function returning an `<Icon>` component to display before the primary text. |
+| `rightAvatar` | Optional | function | | A function returning an `<Avatar>` component to display after the primary text. |
+| `rightIcon` | Optional | function | | A function returning an `<Icon>` component to display after the primary text. |
+| `rowStyle` | Optional | function | | A function returning a style object to apply to each row. |
+| `rowSx` | Optional | function | | A function returning a sx object to apply to each row. |
+| `empty` | Optional | ReactElement | | A ReactElement to display instead of the list when the data is empty. |
+
+## `empty`
+
+It's possible that a SimpleList will have no records to display. If the SimpleList's parent component does not handle the empty state, the SimpleList will display a message indicating there are no results. This message is translatable and its key is `ra.navigation.no_results`.
```
adguernier, line 58:
```suggestion
It's possible that a `<SimpleList>` will have no records to display. If the `<SimpleList>`'s parent component does not handle the empty state, the `<SimpleList>` will display a message indicating there are no results. This message is translatable with the key `ra.navigation.no_results`.
```

docs/Datagrid.md
```diff
@@ -377,6 +344,39 @@ const CustomResetViewsButton = () => {
...
+## `children`
+
+`<Datagrid>` accepts a list of Field components as children. It inspects each child's `source` and/or `label` props to determine the name of the column.
+
+What's a Field component? Simply a component that reads the record (via `useRecordContext`) and renders a value. React-admin includes many Field components that you can use as children of `<Datagrid>` (`<TextField>`, `<NumberField>`, `<DateField>`, `<ReferenceField>`, and many more). Check [the Fields documentation](./Fields.md) for more information. 
+
+You can even create your own field components.
+
+```jsx
+// in src/posts.js
```
adguernier, line 356: We are in `posts.js` but we build a list for users resource ? 🤔

docs/Datagrid.md
```diff
@@ -628,90 +628,92 @@ const PostList = () => (
...
         </Datagrid>
     </List>
 );
 ```
 
-## `rowSx`
+`rowClick` accepts the following values:
 
-You can customize the styles of rows and cells in `<Datagrid>` (applied to the `<DatagridRow>` element) based on the record, thanks to the `rowSx` prop, which expects a function. React-admin calls this function for each row, passing the current record and index as arguments. The function should return a Material UI [`sx`](https://mui.com/system/getting-started/the-sx-prop/), which react-admin uses as a `<TableRow sx>` prop. 
+* "edit" to redirect to the edition vue
```
adguernier, line 649:
```suggestion
* "edit" to redirect to the edition view
```

docs/Datagrid.md
```diff
@@ -628,90 +628,92 @@ const PostList = () => (
...
+* "show" to redirect to the show vue
```
adguernier, line 650:
```suggestion
* "show" to redirect to the show view
```

docs/SimpleList.md
fzaninotto: We usually don't include links in the props table, which is already hard to read. There should be a link ion the `primaryText` section below, though.

adguernier (approved):

PR #9174, opened 2023-08-11
Fix SingleFieldList children don't use default link color

### LINKED ISSUE #9008: Chipfields that are part of a Reference array have black text, while Chipfields that are singular inherit from primary palette
**What you were expecting:**
I expected Chipfields to look the same regardless of whether they are part of a Reference array or single Reference, perhaps there could be a prop that allows the developer to define whether they want link highlighting or not

**What happened instead:**
Single Chipfield text take on the color of what is defined in theme as primary palette, while Chipfields as part of a reference array take on black text.

What I mean by a single Chipfield:
```
        <ReferenceField
          label="Employee"
          source="user.id"
          reference="employees"
          sortBy="user->name"
          link="show"
        >
          <ChipField
            source="username"
            size="small"
            clickable
            sx={{
              '&.RaChipField-chip': {
                color: theme.palette?.chipFieldText?.main ?? 'rgba(0, 0, 0, 0.87)',
              },
            }}
          />
        </ReferenceField>
```

What I mean by a Chipfield as part of an array:

```
        <ReferenceArrayField label="Members" source="member_ids" reference="employees">
          <SingleFieldList linkType="show">
            <ChipField source="username" size="small" clickable />
          </SingleFieldList>
        </ReferenceArrayField>
```

Note: I have included my workaround to ensure that they are the same. Maybe there is something simpler? 'theme' is constructed by `const [theme, setTheme] = useTheme();`

And theme is created like this:
```
declare module '@mui/material/styles' {
  interface PaletteOptions {
    chipFieldText?: {
      main: string;
    };
  }
}

const chipFieldTextColor = 'rgba(0, 0, 0, 0.87)';
const theme = {
  ...defaultTheme,
  palette: {
    primary: {
      main: import.meta.env.VITE_PALETTE_PRIMARY || '#330808',
    },
    secondary: {
      main: import.meta.env.VITE_PALETTE_SECONDARY || '#A32122',
    },
    chipFieldText: {
      main: chipFieldTextColor,
    },
  },
  components: {
    ...defaultTheme.components,
    RaChipField: {
      styleOverrides: {
        root: {
          '&.RaChipField-chip': {
            color: chipFieldTextColor,
          },
        },
      },
    },
  },
};
```
I had initially hoped to be able to remove the individual sx props from each singular theme by this workaround, but it seems that the component theme definition over here also only applies to array chipfields (maybe this is the root of the problem?)

**Environment**

* React-admin version: 4.9.0
* Last version that did not exhibit the issue (if applicable): NA
* React version: 18.2.0
* Browser: All
* Stack trace (in case of a JS error): NA

fzaninotto:
Closes #9008

### DIFF
--- a/packages/ra-ui-materialui/src/list/SingleFieldList.tsx
+++ b/packages/ra-ui-materialui/src/list/SingleFieldList.tsx
@@ -35,8 +35,8 @@ import { Link } from '../Link';
  * Set the linkType prop to "show" to link to the <Show> page instead.
  *
  * @example
- * <ReferenceManyField reference="books" target="author_id" linkType="show">
- *     <SingleFieldList>
+ * <ReferenceManyField reference="books" target="author_id">
+ *     <SingleFieldList linkType="show">
  *         <ChipField source="title" />
  *     </SingleFieldList>
  * </ReferenceManyField>
@@ -45,8 +45,8 @@ import { Link } from '../Link';
  * `linkType` to false.
  *
  * @example
- * <ReferenceManyField reference="books" target="author_id" linkType={false}>
- *     <SingleFieldList>
+ * <ReferenceManyField reference="books" target="author_id">
+ *     <SingleFieldList linkType={false}>
  *         <ChipField source="title" />
  *     </SingleFieldList>
  * </ReferenceManyField>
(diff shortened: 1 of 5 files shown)


packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -159,14 +159,17 @@ export const SingleFieldListClasses = {
...
     overridesResolver: (props, styles) => styles.root,
-})({
+})(({ theme }) => ({
     display: 'flex',
     flexWrap: 'wrap',
 
     [`& .${SingleFieldListClasses.link}`]: {
         textDecoration: 'none',
+        '& > *': {
+            color: theme.palette.primary.main,
```
fzaninotto, line 169: This is the fix - and the indication of how to revert this change in user land with a custom theme, if your requirements lead you in the opposite direction.

PR #9182, opened 2023-08-14
Fix `<FilterForm>` applies but hides default value on clear

fzaninotto:


packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -351,7 +351,7 @@ export const getQuery = ({
             ? queryFromLocation
             : hasCustomParams(params)
             ? { ...params }
-            : { filter: filterDefaultValues || {} };
```
fzaninotto, line 354: here, we applied filter default values only if there were no saved params. This prevented default values to be applied when a user resets all filters (because if they saved the filters once, they have some params in the store containing the sort and pagination)

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -364,6 +364,10 @@ export const getQuery = ({
         query.page = 1;
     }
 
+    if (!query.filter || Object.keys(query.filter).length === 0) {
+        query.filter = filterDefaultValues || {};
```
fzaninotto, line 368: now we use filterdefaultValues even if there are some params - but only if the filter params is empty.

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -161,6 +161,7 @@ const sanitizeRestProps = ({
     displayedFilters,
     filterValues,
     hasCreate,
+    showFilter,
```
fzaninotto, line 164: This was causing warnings in tests

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -250,7 +251,7 @@ const getInputValue = (
     filterValues: Record<string, any>
 ) => {
     if (formValues[key] === undefined || formValues[key] === null) {
-        return '';
+        return (filterValues && filterValues[key]) || '';
```
fzaninotto, line 254: default to the filterValues from the URL when they change

fzaninotto:
Confirmed, this would be a breaking change, so I'm closing this PR.

PR #9185, opened 2023-08-14
Fix `<FilterLiveSearch>` uses label as placeholder by default

### LINKED ISSUE #9175: FilterLiveSearch label does not show for default variant when values are present
**What you were expecting:**

The component should behave identical to the preview image in https://marmelab.com/react-admin/FilterLiveSearch.html allowing us to show a label for the **filled** variant.

When a value is inserted on the non outline variant the label should move to the top of the input field. With multiple FilterLiveSearch fields it is not obvious for the user what field he provided a value for without adding a separate label or header element.

If hiding the label is expected by **filled** it would be nice if there is a different prop to set to allow showing it to not break current versions.

**What happened instead:**

No label is shown and can not be overwritten with the label property. Currently the variant behaves as if the placeholder is set.

image

**Steps to reproduce:**

Add a FilterLiveSearch without variant outline
````typescript
<FilterLiveSearch source="companyId" label="Company Id" />
````

**Related code:**

The label is extracted from the properties and overwritten for the default variant. As it is extracted we can not overwrite the label via the rest properties. 

```typescript
export const FilterLiveSearch = memo((props: FilterLiveSearchProps) => {
    const { filterValues, setFilters } = useListFilterContext();

    const {
        source = 'q',
        variant,
        label = translate('ra.action.search'),
        ...rest
    } = props;

  ....

  {...(variant === 'outlined'
                    ? { variant: 'outlined', label }
                    : {
                          placeholder: label,
                          label: false,
                          hiddenLabel: true,
                      })}
   {...rest}

```

**Other information:**

**Environment**

* React-admin version: 4.12.3

fzaninotto:
Closes #9175

This could be considered as a minor change, but since the doc was never up to date, and since it's easy to override, I think we can accept it as a bugfix.

### DIFF
--- a/docs/FilterLiveSearch.md
+++ b/docs/FilterLiveSearch.md
@@ -50,7 +50,7 @@ export const CustomerList = () => (
 
 | Prop | Required | Type | Default | Description |
 |------|----------|------|---------|-------------|
-| `hiddenLabel` | Optional | `boolean` | `false` | If true, use the label as a placeholder |
+| `hiddenLabel` | Optional | `boolean` | `false` | If true, use the label as a placeholder. |
 | `label` | Optional | `string` | 'ra.action.search' | The label of the search input. |
 | `source` | Optional | `string` | 'q' | The field to filter on. |
 | `variant` | Optional | `string` | 'standard' | The variant of the search input. Can be one of 'standard', 'outlined', or 'filled'. |
--- a/docs/FilterLiveSearch.md
+++ b/docs/FilterLiveSearch.md
@@ -12,19 +12,24 @@ title: "The FilterLiveSearch Component"
 </video>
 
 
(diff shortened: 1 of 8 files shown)


docs/FilterLiveSearch.md
```diff
@@ -12,24 +12,47 @@ title: "The FilterLiveSearch Component"
...
+    </List>
+);
 ```
+{% endraw %}
+
+## Props
+
+|------|----------|------|---------|-------------|
+| `hiddenLabel` | Optional | `boolean` | `false` | If true, use the label as a placeholder |
```
adguernier, line 53:
```suggestion
| `hiddenLabel` | Optional | `boolean` | `false` | If true, use the label as a placeholder.
```

PR #9191, opened 2023-08-17
Fix validation errors from resolvers are not translated

### LINKED ISSUE #9084: String error messages are not translated when using global validation

djhi:
Fixes #9084

### DIFF
--- a/packages/ra-core/src/form/Form.spec.tsx
+++ b/packages/ra-core/src/form/Form.spec.tsx
@@ -10,8 +10,13 @@ import { Form } from './Form';
 import { useNotificationContext } from '../notification';
 import { useInput } from './useInput';
 import { required } from './validate';
-import { SanitizeEmptyValues } from './Form.stories';
-import { NullValue } from './Form.stories';
+import {
+    FormLevelValidation,
+    InputLevelValidation,
+    ZodResolver,
+    SanitizeEmptyValues,
+    NullValue,
+} from './Form.stories';
 
 describe('Form', () => {
     const Input = props => {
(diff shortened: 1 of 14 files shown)


packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -11,13 +11,26 @@ export interface ValidationErrorProps {
...
-    if ((error as ValidationErrorMessageWithArgs).message) {
-        const { message, args } = error as ValidationErrorMessageWithArgs;
+    // react-hook-form expects errors to be plain strings but our validators can return objects
+    // that have message and args.
+    // To avoid double translation for users that validate with a schema instead of our validators
+    // we use a special format for our validators errors.
+    // The ValidationError component will check for this format and extract the message and args
+    // to translate.
+    if (typeof error === 'string' && error.startsWith('@@react-admin@@')) {
+        errorMessage = JSON.parse(error.replace('@@react-admin@@', ''));
```
fzaninotto, line 23: the error message may contain the string '@@react-admin@@' (because, why not?). Instead of a replace, I'd use a substring for improved robustness.

packages/ra-core/src/form/Form.stories.tsx
```diff
@@ -164,3 +166,25 @@ export const UndefinedValue = () => {
...
+    const [result, setResult] = React.useState<any>();
+    return (
+        <CoreAdminContext>
+            <Form
+                record={{}}
+                onSubmit={data => setResult(data)}
+                resolver={zodResolver(zodSchema)}
+            >
+                <Input source="preTranslated" />
+                <Input source="translationKey" />
```
fzaninotto, line 184: it's not entirely clear what to expect here. Would you mind adding a description?

fzaninotto (changes requested):
This PR is missing some additional tests. Also, as this changes the behavior of the resolvers validation, I'd be more comfortable if this was released in a minor version. Would you mind targeting the `next` branch instead?

packages/ra-core/src/form/useInput.ts
```diff
@@ -43,7 +42,6 @@ export const useInput = <ValueType = any>(
     const formGroupName = useFormGroupContext();
     const formGroups = useFormGroups();
     const record = useRecordContext();
-    const getValidationErrorMessage = useGetValidationErrorMessage();
```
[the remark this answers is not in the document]
djhi: Added

fzaninotto (commented):
Cool, but we really need extensive tests showing that error message translation occurs once and only once in the 3 following cases:

- per input validation
- form-wide validation with `validate`
- form-wide validation with `resolvers`

These tests may be scattered around several components, I'm not sure. A central test for this matter would be welcome.

djhi:
> 

Added to `Form.spec.tsx`

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -32,10 +32,10 @@ const ValidationError = (props: ValidationErrorProps) => {
             message,
             args,
         } = errorMessage as ValidationErrorMessageWithArgs;
-        return <>{translate(message, { _: message, ...args })}</>;
+        return <>{translate(message, args)}</>;
     }
 
-    return <>{translate(errorMessage as string, { _: errorMessage })}</>;
+    return <>{translate(errorMessage as string)}</>;
```
fzaninotto, line 38: Why did you remove that?

fzaninotto (changes requested):

packages/ra-core/src/form/ValidationError.tsx
djhi: To allow warnings on missing translations

packages/ra-core/src/form/ValidationError.tsx
```diff
@@ -32,6 +32,7 @@ const ValidationError = (props: ValidationErrorProps) => {
             message,
             args,
         } = errorMessage as ValidationErrorMessageWithArgs;
+        console.log({ message, args });
```
fzaninotto, line 35: console.log spotted

fzaninotto (changes requested):

packages/ra-core/src/form/ValidationError.tsx
fzaninotto: But then it's a regression, as missing translation warnings will now get logged in the console while they  were not in `master`. 

If this is only for the purpose of tests, you can enable warning on missing message in the parameters of the polyglotI18nProvider options (https://airbnb.io/polyglot.js/#options-overview). And if it doesn't work, you'll have to use a spy i18nProvider.

djhi: The fact that it silenced the missing translations warnings was actually a reggresion and was contradictory to the expected and documented behavior.

packages/ra-core/src/form/Form.spec.tsx
```diff
@@ -17,6 +19,8 @@ import {
     SanitizeEmptyValues,
     NullValue,
 } from './Form.stories';
+import { mergeTranslations } from '../i18n';
+import { de } from 'date-fns/locale';
```
fzaninotto, line 23: linter warning

packages/ra-core/src/form/Form.spec.tsx
```diff
@@ -4,14 +4,23 @@ import { useFormState, useFormContext } from 'react-hook-form';
...
-import { NullValue } from './Form.stories';
+import {
+    FormLevelValidation,
+    InputLevelValidation,
+    ZodResolver,
+    SanitizeEmptyValues,
+    NullValue,
+} from './Form.stories';
+import { mergeTranslations } from '../i18n';
+import { de } from 'date-fns/locale';
```
djhi, line 23:
```suggestion
```

fzaninotto (approved):

PR #9196, opened 2023-08-18
Fix UpdateButton with custom notification doesn't close the confirmation dialog

### LINKED ISSUE #9194: UpdateButton Confirm Dialogue Close Cannot Be Triggered with a Custom onSuccess Mutation Option (no redirect)
**What you were expecting:**
I expect that I would be able to close the update button confirm dialogue even with a custom onSuccess mutation option.

**What happened instead:**
onSuccess method is run (i.e. custom notification, page refresh, etc.) but the confirm dialogue remains open without a page redirect.

**Steps to reproduce:**
Pass a custom onSuccess mutation option to a confirmable UpdateButton that does not trigger a redirect. 

**Related code:**
```
<UpdateButton
          label='Send Password Reset'
          mutationMode="pessimistic"
          confirmTitle="Password Reset"
          confirmContent="Are you sure you want to send a password reset email?"
          data={{ ... }}
          mutationOptions={{
            onSuccess: () => {
              notify('Password reset email sent.', { type: 'success' })
              refresh()
            }
          }} 
        />
```

**Other information:**
Another idea, although somewhat unrelated, would be providing the ability to override the onSuccess / onError notification messages without having to pass custom mutation options. That would be a workaround for my case, but doesn't solve the underlying issue. 

**Environment**

* React-admin version: 4.13.0
* React version: 18.2.0
* Browser: Chrome

fzaninotto:
Closes #9194

### DIFF
--- a/docs/UpdateButton.md
+++ b/docs/UpdateButton.md
@@ -182,13 +182,11 @@ import { useNotify, useRefresh, useRedirect, TopToolbar, UpdateButton } from 're
 
 const PostEditActions = () => {
     const notify = useNotify();
-    const refresh = useRefresh();
     const redirect = useRedirect();
 
     const onSuccess = () => {
         notify(`Changes saved`);
         redirect('/posts');
-        refresh();
     };
 
     return (
@@ -212,7 +210,6 @@ The default `onSuccess` function is:
         messageArgs: { smart_count: 1 },
         undoable: mutationMode === 'undoable'
     });
-    refresh();
 }
 ```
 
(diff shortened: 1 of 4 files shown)


djhi (approved):

slax57:
FTR I don't agree with this fix.
To me this is just moving the issue to another place.

The root issue is not being able to close the dialog from within a side-effect fn, like `onSuccess`. We should instead provide a callback for that as an extra argument, that can be called in user land.

With this fix, the problem is no longer there if we provide a custom `onSuccess`, but we will have the same issue if we override the `onSettled`.
Besides, this is also technically a BC as, from now on, the dialog will also close in case of an error, while it would have remained open in the previous version (which makes it easier to retry the update).

fzaninotto:
You're absolutely right about the BC, I didn't see it. I'll reopen the issue.

PR #9156, opened 2023-08-02
Fix return type of beforeSave in withLifecycleCallbacks.ts

awesome-manuel:
Fixes #9152

PR #9199, opened 2023-08-18
Fix withLifecycleCallbacks types

### LINKED ISSUE #9152: Return type of ResourceCallbacks.beforeSave should be Promise<Partial<T>>
The return type of `ResourceCallbacks.beforeSave` should be the same as the type of the parameter `data`. The function cannot create a `RaRecord` from a `Partial<RaRecord>`.

https://github.com/marmelab/react-admin/blob/de8fbaa3285714aab914e0e76cc369c09a817d3c/packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts#L577C73-L577C74

djhi:
Supersedes #9156
Fixes #9152

### DIFF
--- a/packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
+++ b/packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
@@ -332,7 +332,7 @@ export const withLifecycleCallbacks = (
 
         create: async function <RecordType extends RaRecord = any>(
             resource: string,
-            params: CreateParams<Partial<RecordType>>
+            params: CreateParams<RecordType>
         ) {
             let newParams = params;
             const beforeCreateHandlers = handlers.filter(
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -332,7 +332,7 @@ export const withLifecycleCallbacks = (
 
         create: async function <RecordType extends RaRecord = any>(
             resource: string,
-            params: CreateParams<RecordType>
+            params: CreateParams<Partial<RecordType>>
```
fzaninotto, line 335: I don't think this should be changed - to me, it's the CreateParams and UpdatePArams that should be changed. 

```diff
export interface CreateParams<T = any> {
-   data: T;
+   data: Partial<T>;
    meta?: any;
}
```

djhi: Agreed and done

slax57 (approved):

PR #9201, opened 2023-08-21
Update package.json

kamiyo:
https://github.com/marmelab/react-admin/blob/83dc1d61f0d7efaafebb57885d6b55f78509b581/packages/ra-ui-materialui/src/form/FormTabHeader.tsx has an import from `react-is`, but it is not declared in `package.json`

This was causing errors in my rollup build. I cannot be sure this is the right fix - for my project, I had to add this in my `.yarmrc.yml`:

```
packageExtensions:
  "ra-ui-materialui@*":
    dependencies:
      "react-is": "*"
```
Here, `peerDependencies` doesn't work, but for `package.json` I think it makes sense to put it there.


djhi (changes requested):
Thanks! Can you please commit the updated `yarn.lock` as well?

kamiyo:
Hrm, the yarn update updated some other dependencies at all. Is that okay, or do I somehow have to cherry pick just the `react-is` addition under `ra-ui-materialui`?

slax57:
I took the liberty to update your branch, and fix the `yarn.lock` file (by running `make install`).
Should be good now.

PR #9208, opened 2023-08-21
Revert "Fix UpdateButton with custom notification doesn't close the confirmation dialog"

fzaninotto:
Reverts marmelab/react-admin#9196


slax57 (changes requested):
Actually I think we should not merge this.
See https://github.com/marmelab/react-admin/pull/9196#issuecomment-1686641047.
Sorry about the mistake I made. :sweat:

fzaninotto:
You're right! We don't need to revert.

PR #9209, opened 2023-08-21
Update `<PrevNextButtons>` to fetch less aggressively

fzaninotto:
## Problem

`<PrevNextButtons>` has to fetch the entire list to determine the previous and next id. This may be costly in network and server time (on the client-side, iven with a list of 1000 items, the time is negligible). 

## Solution

Check if the previous and next id can't be determined from the cache first, so that users coming from a list page don't need to fetch the entire list of records

## Out of scope

When a user reaches the end of the list cache, update the list params in the store to increment or decrement the page. 

## To Do

- [x] Use `<PrevNextButtons />` in the demo
- [x] Allow to specify `filterDefaultValues`
- [x] Use the cache if possible

### DIFF
--- a/packages/ra-core/src/controller/usePrevNextController.ts
+++ b/packages/ra-core/src/controller/usePrevNextController.ts
@@ -131,6 +131,8 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
             filter: filterDefaultValues,
             order: initialSort.order,
             sort: initialSort.field,
+            page: 1,
+            perPage: 10,
         }
     );
 
@@ -189,7 +191,7 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
 
     let finalData = canUseCacheData ? queryData.data : data?.data || [];
 
-    if (!record) return null;
+    if (!record || isLoading) return { isLoading: true };
 
     const ids = finalData.map(record => record.id);
     const index = ids.indexOf(record.id);
@@ -242,13 +244,17 @@ export interface UsePrevNextControllerProps<RecordType extends RaRecord = any> {
     }> & { meta?: any };
 }
 
-export interface UsePrevNextControllerResult {
-    hasPrev: boolean;
-    hasNext: boolean;
-    prevPath: string | undefined;
-    nextPath: string | undefined;
-    index: number | undefined;
-    total: number | undefined;
-    error?: any;
-    isLoading: boolean;
-}
+export type UsePrevNextControllerResult =
+    | {
+          isLoading: true;
+      }
+    | {
+          hasPrev: boolean;
+          hasNext: boolean;
+          prevPath: string | undefined;
+          nextPath: string | undefined;
+          index: number | undefined;
+          total: number | undefined;
+          error?: any;
+          isLoading: false;
+      };
--- a/packages/ra-core/src/controller/usePrevNextController.ts
+++ b/packages/ra-core/src/controller/usePrevNextController.ts
@@ -1,6 +1,6 @@
-import { UseQueryOptions } from 'react-query';
+import { UseQueryOptions, useQuery } from 'react-query';
 import { useResourceContext } from '../core';
-import { useGetList } from '../dataProvider';
+import { useDataProvider, useGetList } from '../dataProvider';
 import { useStore } from '../store';
 import { FilterPayload, RaRecord, SortPayload } from '../types';
 import { ListParams, SORT_ASC } from './list';
@@ -107,8 +107,9 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
         linkType = 'edit',
         storeKey,
         limit = 1000,
-        sort = { field: 'id', order: SORT_ASC },
-        filter = {},
+        sort: initialSort = { field: 'id', order: SORT_ASC },
+        filter: permanentFilter = {},
+        filterDefaultValues = {},
         queryOptions = {
             staleTime: 5 * 60 * 1000,
         },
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -124,37 +125,76 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
...
+        queryData?.data &&
+        recordIndexInQueryData !== -1 &&
+        !isRecordIndexFirstInNonFirstPage &&
+        !isRecordIndexLastInNonLastPage;
 
-    if (!record) return null;
+    // If the previous and next ids are not in the cache, fetch the entire list.
+    // This is necessary e.g. when coming directly to a detail page,
+    // without displaying the list first
+    const { data, error, isLoading } = useQuery(
```
fzaninotto, line 181: I used useQuery instead of useGetList to avoid paying the tax of prepopulating the `getOne` cache, which can be costly for large lists.

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -124,37 +125,76 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
...
         resource,
+        'getList',
         {
-            sort: {
-                field: storedParams.sort,
-                order: storedParams.order,
+            ...params,
+            pagination: {
+                page: storedParams.page,
+                perPage: storedParams.perPage,
```
adguernier, line 159: I'm not sure we should use `storedParams.perPage` here.
This parameter does not serve the same purpose than `limit`: 
- `storedParams.perPage` is coming from `<List>`, it use to display `n` items per page in a `<List>`.
- `limit` is made to limit the maximum result to fetch.

For example: if `storedParams.perPage` is set to 25, the `<List>` will show 25 items per page. But `<PrevNextButton>` will fetch only 25 items and the user will not be able to navigate beyond the twenty-fifth item although the item total is greater.

The story `WithLimit` shows the behavior described above.

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -107,8 +107,9 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
         linkType = 'edit',
         storeKey,
         limit = 1000,
-        sort = { field: 'id', order: SORT_ASC },
-        filter = {},
+        sort: initialSort = { field: 'id', order: SORT_ASC },
+        filter: permanentFilter = {},
+        filterDefaultValues = {},
```
adguernier, line 112: I don't undertand why we use two filters params as they seems to serve the same purpose. It's a bit confusing to use in my opinion

packages/ra-core/src/controller/usePrevNextController.ts
fzaninotto: I'm using the pagination params from the store to look for cached data from the list view, to avoid fecthing if possible. It may happen that the current record id isn't in this list. In this case, the component does fetch data using the `limit` (see the useQuery further down)

adguernier (approved):

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -124,37 +125,78 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
...
 
-    const { data, error, isLoading } = useGetList<RecordType>(
+    const dataProvider = useDataProvider();
+    const queryClient = useQueryClient();
+    const pagination = { page: 1, perPage: limit };
+    const sort = {
+        field: storedParams.sort,
+        order: storedParams.order,
+    };
+    const filter = { ...storedParams.filter, ...permanentFilter };
```
slax57, line 146: Shouldn't we rather use lodash merge here, in case the filters are defined on nested fields?

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -124,37 +125,78 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
...
+        [resource, 'getList', params],
+        () => dataProvider.getList(resource, params),
+        {
+            enabled: !canUseCacheData,
+            ...otherQueryOptions,
+        }
     );
 
-    if (!record) return null;
+    let finalData = canUseCacheData ? queryData.data : data?.data || [];
```
slax57, line 192: why not `const`?

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -124,37 +125,78 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
...
         resource,
+        'getList',
         {
-            sort: {
-                field: storedParams.sort,
-                order: storedParams.order,
+            ...params,
+            pagination: {
+                page: storedParams.page,
+                perPage: storedParams.perPage,
```
slax57, line 161: Maybe you shouldn't include `meta` here, while interrogating the cache, because `meta` from `<PrevNextButton>` may not be the same as `meta` from `<List>`.

We may need to provide a different `meta` prop for interrogating the cache and for actually calling `getList` in the context of this controller.

slax57 (changes requested):

packages/ra-core/src/controller/usePrevNextController.ts
fzaninotto: Maybe, but since we do a simple merge as well in `useListController`, this would lead to different results than in the list, so we'd better keep it like that IMO. 

https://github.com/marmelab/react-admin/blob/923b63d85ad4deb20ad07aee440fe9160d63baf8/packages/ra-core/src/controller/list/useListController.ts#L93

packages/ra-core/src/controller/usePrevNextController.ts
fzaninotto: I can't determine if it's better to include it or not. Let's wait and see if the problem arises before making the API more complex.

slax57 (approved):

PR #9218, opened 2023-08-25
Fix Guessers should not log in CI by default

fzaninotto:
## Problem

Guessers are convenient for tests... except they log their output to the console

## Solution

- Disable guessers log outside development by default
- expose a prop to set the logging behavior


djhi (approved):
:tada:

PR #9223, opened 2023-08-25
Add `<SortButton sx>` prop to customize button style

### LINKED ISSUE #9161: Unable to override SortButton styling
**What you were expecting:**
Sort button to use styling props as similar Tool bar action buttons do

**What happened instead:**
styling props have no effect

**Related code:**

With the following code the Sort Button is not styled in the same way as the Export and Create Button

```
        <TopToolbar>
            <SortButton color="inherit" fields={['lastName', 'firstName', 'lastSeen', 'lastBooking', 'createdDate']} />
            <ExportButton color="inherit" />
            <CreateButton color="inherit"/>
        </TopToolbar>

```

fzaninotto:
Closes #9161

image


djhi (changes requested):
You should also add a root css class and assign it to the Root element (with cslx to add className). It's not done automatically by MUI and prevents targetting this element from its parent styles

fzaninotto:
Review applied

slax57 (approved):

slax57 (changes requested):
My bad, this PR is missing one more thing: updating the documentation to mention we support the `sx` prop now.

djhi (approved):

PR #9232, opened 2023-08-29
Reduce `<TopToolbar>` min height to move content up when there is no filter

fzaninotto:
## Problem

In List views, if there is no filter, there is a lot of wasted space between the app bar and the list. This is especially visible when using alternative layouts. 

## Solution

Remove the min height of the `TopToolbar`(which applies to all pages) and add it to the `FilterForm`, so that the top toolbar remains tall enough when there are filters. 

As a consequence:

- list with filters see no change
- list without filters have a reduced top margin

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListToolbar.tsx
+++ b/packages/ra-ui-materialui/src/list/ListToolbar.tsx
@@ -3,7 +3,7 @@ import { FC, memo } from 'react';
 import { styled } from '@mui/material/styles';
 import { ReactElement } from 'react';
 import PropTypes from 'prop-types';
-import { Toolbar, ToolbarProps } from '@mui/material';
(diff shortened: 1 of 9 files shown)


packages/ra-ui-materialui/src/list/ListToolbar.tsx
```diff
@@ -63,18 +63,16 @@ export interface ListToolbarProps
 
 const PREFIX = 'RaListToolbar';
 
-const Root = styled(Toolbar, {
+const Root = styled('div', {
```
fzaninotto, line 66: The ListToolbar renders ListActions, which itself renders TopToolbar, which also renders a Toolbar. So there was a toolbar inside a toolbar, with added margin and padding... not really necessary.

djhi (approved):
:+1:

PR #9236, opened 2023-08-31
Fix useReferenceManyFieldController fails with error when the record is not yet present

djhi:
### Problem:
`<ReferenceManyField>` throws the error: `target and id are required` when it’s used in an `aside` of an `<Edit>` component.

Checking if record does not exist with `useRecordContext` solves the problem.

### Solution
`useReferenceManyFieldController` should check if the record exist before calling `useGetManyReference`

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -162,7 +162,7 @@ export const useReferenceManyFieldController = <
             filter: filterValues,
         },
         {
-            // enabled: !!get(record, source),
+            enabled: !!get(record, source),
             keepPreviousData: true,
             onError: error =>
                 notify(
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -162,6 +162,7 @@ export const useReferenceManyFieldController = <
             filter: filterValues,
         },
         {
+            enabled: !!get(record, source),
```
slax57, line 165:
```suggestion
            enabled: get(record, source) != null,
```

Remember that `0` is a valid id

slax57 (changes requested):
Is it possible to describe the case where you ran into this issue, and/or paste the error you got? This would make it easier for people having this same error to find this issue.

packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
djhi: Good catch :)

slax57 (approved):

Issue #9239, opened 2023-09-01
AutocompleteInput's onInputChange prop is never called


tdnl:
**What you were expecting:**

Looking at the documentation, we can provide a [onInputChange](https://marmelab.com/react-admin/AutocompleteInput.html#selecting-a-foreign-key) to an AutocompleteInput to retrieve the typed value.

**What happened instead:**

`onInputChange` is never called.

**Steps to reproduce:**
- Open stackblitz sandbox : https://stackblitz.com/edit/github-eefhoj-nqszre?file=src%2Fposts%2FPostEdit.tsx
- Click on a post to view edit page
- Add an author
- Type "value" (or anything else) in AutocompleteInput
- "value" should be displayed above the form

**Related code:**

There is currently no use of the AutocompleteInput's `onInputChange` prop, but we could extract it from the props and call it the [handleInputChange](https://github.com/marmelab/react-admin/blob/86670919a15aa213622ee2a5d9807267bc2f1eec/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx#L463) method

**Environment**

* React-admin version: 4.13.3

djhi:
Thanks! Would you mind opening a PR to fix it?

PR #9240, opened 2023-09-01
fix AutocompleteInput's onInputChange is never called

tdnl:
Closes https://github.com/marmelab/react-admin/issues/9239


djhi (approved):
Thanks!

djhi:
Hey @tndl, I've merged a little fast actually. First, we're missing documentation about this prop. Second, it appears there's an issue with this implementation: an effect is ran after the custom `onInputChange` is called and it will be called again with the previous value. 
Can you please make a new PR with a story for this prop and adding documentation?

tdnl:
Hey @djhi, that works, looking into it as soon as I can.

PR #9242, opened 2023-09-04
Add `<Menu.Item>title</Menu.item>` syntax

fzaninotto:
## Problem

The `<Menu.Item primaryText>` syntax isn't intuitive, especially for someone who already knows material-ui, where the menu label is expressed via `children`.

## Solution

Support menu label is expressed via `children` *in addition to* `primaryText` for backwards compatibility. 

```jsx
const CustomMenu = () => (
    <Menu>
        <Menu.Item to="/" leftIcon={<Dashboard />}>
            Dashboard
        </Menu.Item>
        <Menu.Item to="/sales" leftIcon={<PieChartOutlined />}>
            Sales
        </Menu.Item>
        <Menu.Item to="/customers" leftIcon={<PeopleOutlined />}>
            Customers
        </Menu.Item>
        <Menu.Item to="/products" leftIcon={<Inventory />}>
            Catalog
        </Menu.Item>
    </Menu>
);
```

### DIFF
--- a/docs/Menu.md
+++ b/docs/Menu.md
@@ -24,7 +24,7 @@ export const MyMenu = () => (
         <Menu.ResourceItem name="posts" />
         <Menu.ResourceItem name="comments" />
         <Menu.ResourceItem name="users" />
-        <Menu.Item to="/custom-route" primaryText="Miscellaneous" leftIcon={<LabelIcon />}/>
+        <Menu.Item to="/custom-route" leftIcon={<LabelIcon />}>Miscellaneous</Menu.Item>
     </Menu>
 );
 ```
@@ -117,7 +117,7 @@ export const MyMenu = () => (
         <Menu.ResourceItem name="posts" />
         <Menu.ResourceItem name="comments" />
         <Menu.ResourceItem name="users" />
-        <Menu.Item to="/custom-route" primaryText="Miscellaneous" leftIcon={<LabelIcon />}/>
+        <Menu.Item to="/custom-route" leftIcon={<LabelIcon />}>Miscellaneous</Menu.Item>
     </Menu>
 );
 ```
@@ -166,12 +166,12 @@ import LabelIcon from '@mui/icons-material/Label';
 export const MyMenu = () => (
     <Menu>
         ...
-        <Menu.Item to="/custom-route" primaryText="Miscellaneous" leftIcon={<LabelIcon />}/>
+        <Menu.Item to="/custom-route" leftIcon={<LabelIcon />}>Miscellaneous</Menu.Item>
     </Menu>
 );
 ```
 
-The `primaryText` prop accepts a string or a React node. You can use it e.g. to display a badge on top of the menu item:
+The `children` prop accepts a string or a React node. You can use it e.g. to display a badge on top of the menu item:
 
 ```jsx
 import Badge from '@mui/material/Badge';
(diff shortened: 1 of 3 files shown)


djhi (commented):
We should deprecate `primaryText` to remove it in v5 IMO

docs/Menu.md
```diff
@@ -24,7 +24,7 @@ export const MyMenu = () => (
         <Menu.ResourceItem name="posts" />
         <Menu.ResourceItem name="comments" />
         <Menu.ResourceItem name="users" />
-        <Menu.Item to="/custom-route" primaryText="Miscellaneous" leftIcon={<LabelIcon />}/>
+        <Menu.Item to="/custom-route" leftIcon={<LabelIcon />}>Miscellaneous</Menu.Item>
```
fzaninotto, line 27: Not sure this is better, especially since the menu label can be a translation key... Back to the drawing board.

djhi: Just rename it to `label`?

slax57 (changes requested):
Looks very nice !

Just found a little issue: the Tooltip is still rendered on hover when providing children, and it's empty so it's a bit weird.

Capture d’écran du 2023-09-07 11-15-02

fzaninotto:
Because we have to show the menu label on hover on Mobile, we must always support `primaryText`, even when passing `children`. But then we end up with duplicated label, because we have to render both in the expanded menu for backwards compatibility reasons. 

image

The only BC solution I can think of is to deprecate `primaryText` and replace it with `label`, and only render `label` in expanded menu if there is no `children`. But this is a much larger change:

image

So we should probably not move forward with this. Anyone willing to use custom menu content can still use their own component instead of `<Menu.Item>`.

WDYT?

slax57:
@fzaninotto You're right.

However, I don't understand why you are saying "we have to render both in the expanded menu for backwards compatibility reasons".

Unless I'm mistaken, `<Menu.Item>` did not accept `children` in the first place (or rather, ignored them). So why couldn't we implement a logic that says: render the `children` under `<StyledMenuItem>` when there are `children` provided, otherwise render `primaryText` ; and in any case, render `primaryText` as the `<Tooltip title>`?

The following seems to work according to my tests:

```tsx
    const renderMenuItem = () => {
        return (
            <StyledMenuItem
                className={clsx(className, {
                    [MenuItemLinkClasses.active]: !!match,
                })}
                // @ts-ignore
                component={LinkRef}
                ref={ref}
                tabIndex={0}
                {...rest}
                onClick={handleMenuTap}
            >
                {leftIcon && (
                    <ListItemIcon className={MenuItemLinkClasses.icon}>
                        {leftIcon}
                    </ListItemIcon>
                )}
                {children
                    ? children
                    : typeof primaryText === 'string'
                    ? translate(primaryText, { _: primaryText })
                    : primaryText}
            </StyledMenuItem>
        );
    };

    return open ? (
        renderMenuItem()
    ) : (
        <Tooltip
            title={
                typeof primaryText === 'string'
                    ? translate(primaryText, { _: primaryText })
                    : null
            }
            placement="right"
            {...tooltipProps}
        >
            {renderMenuItem()}
        </Tooltip>
    );
```

fzaninotto:
You're right, it works! I've pushed the changes. Switching back to RFR.

slax57 (changes requested):

packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -127,7 +130,7 @@ export const MenuItemLink = forwardRef<any, MenuItemLinkProps>((props, ref) => {
             title={
                 typeof primaryText === 'string'
                     ? translate(primaryText, { _: primaryText })
-                    : primaryText
+                    : null
```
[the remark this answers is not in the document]
fzaninotto: You're right, reverted

djhi (approved):

slax57 (approved):

PR #9243, opened 2023-09-05
fix autocomplete sets previous value using onInputChange

tdnl:
## Problem

When providing `onInputChange` to an AutocompleteInput, a `useEffect` is run and resets the Textbox's value to the previous value. 

As a result, users can not manually type a value when `onInputChange` provided.

## Solution

It turns out the `useEffect` is not necessary if React-Admin's custom `onInputChange` properly sets the [Textbox's state](https://mui.com/material-ui/react-autocomplete/#controlled-states) whenever it is called.

### DIFF
--- a/docs/AutocompleteInput.md
+++ b/docs/AutocompleteInput.md
@@ -652,7 +652,7 @@ const CompanyInput = () => {
             choices={choicesWithCurrentCompany}
             optionText="name"
             disabled={isLoading}
-            onInputChange={e => setFilter({ q: e.target.value })}
+            onInputChange={(_, newInputValue) => setFilter({ q: newInputValue })}
         />
     );
 }
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -111,6 +111,13 @@ const defaultFilterOptions = createFilterOptions();
  *
  * @example
  * <AutocompleteInput source="author_id" options={{ color: 'secondary', InputLabelProps: { shrink: true } }} />
+ *
+ * Retrieve the value displayed in the textbox using the `onInputChange` prop:
+ *
+ * @example
+ * const [state, setState] = useState('')
+ *
+ * <AutocompleteInput source="gender" choices={choices} onInputChange={(_, newInputValue) => setState(newInputValue)} />
  */
 export const AutocompleteInput = <
     OptionType extends RaRecord = RaRecord,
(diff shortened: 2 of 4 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -440,33 +447,14 @@ If you provided a React element for the optionText prop, you must also provide t
     useEffect(() => {
         if (!multiple) {
             const optionLabel = getOptionLabel(selectedChoice);
-            if (typeof optionLabel === 'string') {
-                setFilterValue(optionLabel);
```
tdnl, line 444: I believe the issue happening because of this effect. 

Removing the call to `setFilterValue` solves the problem as long as `handleInputChange` (called by MUI's `onInputChange`) does it. 

I'm keeping the `useEffect` to throw an error if the user does not provide `inputText` along with `optionText`, but it could probably be moved in `handleInputChange` or `finalOnBlur`.

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -515,13 +503,39 @@ If you provided a React element for the optionText prop, you must also provide t
...
+    const handleInputChange = useCallback<
+        AutocompleteProps<
+            OptionType,
+            Multiple,
+            DisableClearable,
+            SupportCreate
+        >['onInputChange']
+    >(
+        (event, newInputValue, reason) => {
+            setFilterValue(newInputValue);
```
tdnl, line 531: MUI fires `onInputChange` when the component is mounted. In the previous iteration, `event` could be null resulting in `setFilterValue` not being called properly.

docs/AutocompleteInput.md
```diff
@@ -583,7 +583,7 @@ const CompanyInput = () => {
             choices={choicesWithCurrentCompany}
             optionText="name"
             disabled={isLoading}
-            onInputChange={e => setFilter({ q: e.target.value })}
+            onInputChange={(_, newInputValue) => setFilter({ q: newInputValue })}
```
slax57, line 586: Could you also add this prop to the list of props in the 'Props' section, and create a new dedicated section for this prop with an example (similar to the other props)?

slax57 (changes requested):
Other than that it looks good to me, thanks and well done!

djhi:
@tdni any news on this one?

djhi:
No news for some time, closing

PR #9253, opened 2023-09-05
fix: typo double text

oxwazz:
i just delete some doubled text like **the the**, **to to**, and **typo typo**


fzaninotto (approved):

fzaninotto:
Thank you thank you.

PR #9255, opened 2023-09-06
[Demo] Add lazy loading to CRM demo to illustrate code splitting

fzaninotto:
## Problem

Large apps need to do code splitting. It turns out it's not as simple as using `React.lazy()`, as the layouts don't use `Suspense` by default. 

## Solution

Update the CRM demo example to illustrate the way to do it.

### DIFF
--- a/examples/crm/src/Layout.tsx
+++ b/examples/crm/src/Layout.tsx
@@ -1,9 +1,9 @@
-import React, { HtmlHTMLAttributes } from 'react';
+import React, { Suspense, HtmlHTMLAttributes } from 'react';
 import { CssBaseline, Container } from '@mui/material';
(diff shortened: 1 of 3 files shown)


examples/crm/src/Layout.tsx
```diff
@@ -14,7 +14,7 @@ const Layout = ({ children }: LayoutProps) => (
             <main id="main-content">
                 {/* @ts-ignore */}
                 <ErrorBoundary FallbackComponent={Error}>
-                    {children}
+                    <Suspense fallback={<Loading />}>{children}</Suspense>
```
fzaninotto, line 17: I wonder if we shouldn't do that by default on all our layouts.

djhi: Probably

djhi (approved):

PR #9258, opened 2023-09-07
add japanese translate default message

Khant-Nyar:


slax57:
Thank you so much for your precious contribution.

Unfortunately, we cannot accept to include this package into the main repo, as we won't have the knowledge to maintain it in the future.

Instead, we encourage you to publish this package as your own, just like all the other [translation packages](https://marmelab.com/react-admin/TranslationLocales.html) we reference.

By the way, there seem to already exist another package for the Japanse language, so you may want to offer your contribution to their repository instead.

Should you decide to publish a new package, if it proves to be superior to the current one, we will gladly accept a PR to reference it in the official docs.

Thank you for your understanding, and, once more, for your contribution.

Khant-Nyar:
Thanks,I don't it was already have.it took me 1 hours from my OT 😔.

PR #9259, opened 2023-09-07
[Doc] Fix typos on Admin, Architecture, EditTutorial, Features, Fields, Inputs, and Tutorial pages

mchaffotte:
Fix various typos

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -658,7 +658,7 @@ React-admin uses [react-query](https://react-query-v3.tanstack.com/) to fetch, c
 * Query results that are no longer used in the current page are labeled as "inactive" and remain in the cache in case they are used again at a later time.
 * By default, "inactive" queries are garbage collected after 5 minutes.
 * Queries that fail are silently retried 3 times, with exponential backoff delay before capturing and displaying an error to the UI.
-* Query results by default are structurally shared to detect if data has actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to `useMemo` and `useCallback`. 
+* Query results by default are structurally shared to detect if data have actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to `useMemo` and `useCallback`. 
 
 If you want to override the react-query default query and mutation default options, or use a specific client or mutation cache, you can create your own `QueryClient` instance and pass it to the `<Admin queryClient>` prop:
 
--- a/docs/Features.md
+++ b/docs/Features.md
@@ -11,7 +11,7 @@ React-admin is a **rich framework** that covers most of the needs of typical adm
 
 With react-admin, developers assemble application components without having to worry about low-level details. They need less code for the same result, and they can **focus on the business logic** of their app.
 
-[![List view without and with react-admin](./img/list-from-react-to-react-admin.webp)](./ListTutorial.md#from-pure-react-to-react-admin)
+[![List view without and with react-admin](./img/list-from-react-to-react-admin.webp)](./img/list-from-react-to-react-admin.webp)
 
 We've crafted the API of react-admin's components and hooks to be as **intuitive** as possible. The react-admin core team uses react-admin every day, and we're always looking for ways to improve the developer experience.
 
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -610,7 +610,7 @@ You can find components for react-admin in third-party repositories.
 
 ## TypeScript
 
-All field components accept a generic type that describes the record. This lets TypeScript validate that the `source` prop targets an actual field of the record:
+All field components accept a generic type that describes the record. This lets TypeScript validates that the `source` prop targets an actual field of the record:
 
 ```tsx
 import * as React from "react";
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -610,7 +610,7 @@ You can find components for react-admin in third-party repositories.
 
 ## TypeScript
 
-All field components accept a generic type that describes the record. This lets TypeScript validates that the `source` prop targets an actual field of the record:
+All field components accept a generic type that describes the record. This lets TypeScript validate that the `source` prop targets an actual field of the record:
 
 ```tsx
 import * as React from "react";
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -1015,7 +1015,7 @@ Once the app reloads, it's now behind a login form that accepts everyone:
 
 ## Connecting To A Real API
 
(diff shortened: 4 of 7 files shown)


docs/Admin.md
```diff
@@ -658,7 +658,7 @@ React-admin uses [react-query](https://react-query-v3.tanstack.com/) to fetch, c
 * Query results that are no longer used in the current page are labeled as "inactive" and remain in the cache in case they are used again at a later time.
 * By default, "inactive" queries are garbage collected after 5 minutes.
 * Queries that fail are silently retried 3 times, with exponential backoff delay before capturing and displaying an error to the UI.
-* Query results by default are structurally shared to detect if data has actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to `useMemo` and `useCallback`. 
+* Query results by default are structurally shared to detect if data have actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to `useMemo` and `useCallback`.
```
slax57, line 661: Not sure about this one. To me, here we are talking about the JS Object named 'data' (which reference has to be stable across rerenders), and hence I'd keep it a singular.

docs/Fields.md
```diff
@@ -610,7 +610,7 @@ You can find components for react-admin in third-party repositories.
 
 ## TypeScript
 
-All field components accept a generic type that describes the record. This lets TypeScript validate that the `source` prop targets an actual field of the record:
+All field components accept a generic type that describes the record. This lets TypeScript validates that the `source` prop targets an actual field of the record:
```
slax57, line 613: Nope, grammar was good on this one, 'lets' is already conjugated :grin:

docs/Tutorial.md
```diff
@@ -1013,7 +1015,7 @@ Once the app reloads, it's now behind a login form that accepts everyone:
 
 ## Connecting To A Real API
 
-Here is the elephant in the room of this tutorial. In real world projects, the dialect of your API (REST? GraphQL? Something else?) won't match the JSONPlaceholder dialect. Writing a Data Provider is probably the first thing you'll have to do to make react-admin work. Depending on your API, this can require a few hours of additional work.
+Here is the elephant in the room of this tutorial. In real world projects, the dialect of your API (REST? GraphQL? Something else?) won't match the JSONPlaceholder dialect. Writing a Data Provider is probably the first thing you'll have to do to make react-admin works. Depending on your API, this can require a few hours of additional work.
```
slax57, line 1018: Here as well, grammar was good ('make react-admin work' and not 'make react-admin works') :grin:

slax57 (changes requested):
Otherwise, these are nice catches, thanks!

docs/Tutorial.md
```diff
@@ -21,6 +21,8 @@ Here is an overview of the result:
 React-admin uses React. We'll use [create-react-admin](https://github.com/marmelab/react-admin/tree/master/packages/create-react-admin) to bootstrap a new admin:
 
 ```sh
+npm create react-admin test-admin
```
fzaninotto, line 24: no, this won't work.

slax57: I have tested it and it does actually

docs/Features.md
```diff
@@ -11,7 +11,7 @@ React-admin is a **rich framework** that covers most of the needs of typical adm
 
 With react-admin, developers assemble application components without having to worry about low-level details. They need less code for the same result, and they can **focus on the business logic** of their app.
 
-[![List view without and with react-admin](./img/list-from-react-to-react-admin.webp)](./ListTutorial.md#from-pure-react-to-react-admin)
```
djhi, line 14: Not sure about this, shouldn't this be an image with a link?

docs/Tutorial.md
fzaninotto: It must be a recent addition to npm

docs/Admin.md
mchaffotte: what about if the data object has ?

docs/Fields.md
mchaffotte: my bad

docs/Tutorial.md
mchaffotte: my bad

docs/Features.md
mchaffotte: The current behaviour of this image link is to go to a 404 page. This is the current url: .../react-admin/ListTutorial.md#from-pure-react-to-react-admin. It points to a md file instead of a html one. SO the quickest fix was to link to the image.

docs/Admin.md
slax57: On 2nd thought, let's go with your initial change. In the first part of the sentence, 'data' really means the data fetched from the API, and we are implicitly talking about the data object only in the second half of the sentence.

docs/Features.md
slax57: Besides, with the previous version it took 2 clicks to open the image fullscreen, whereas now it only takes one, which provides a better UX IMO

slax57 (approved):

docs/Tutorial.md
fzaninotto: `npm create` [has been introduced in node 16](https://docs.npmjs.com/cli/v8/commands/npm-init?v=true) (npm 8), two years ago. By using it, we risk confusing people wondering width Node 14 or 12, wondering qhy it doesn't work. 

I suggest using `npm init` instead - this is supposed to work since way longer (https://docs.npmjs.com/cli/v6/commands/npm-init?v=true). `npm create` is actually an alias of `npm init`.

mchaffotte: I tested: both works. So no problem to use init instead of create but should we support, for the tutorial, node versions (12, 14 16 soon) which are end of life?

fzaninotto: > should we support, for the tutorial, node versions (12, 14 16 soon) which are end of life?

yes: nowhere in the doc we mention that we have minimal Node js requirements.

docs/Features.md
djhi: Then we should remove the link altogether no?
```
![List view without and with react-admin](./img/list-from-react-to-react-admin.webp)
```

mchaffotte: I do not know. My first feeling was to click the image in order to see better the source codes.

adguernier: It seems we do not use this behaviour elsewhere, but I think it's a good idea to provide a link to open images in fullscreen

fzaninotto: The link to the full-size image is better, so let's keep this modification.

fzaninotto (approved):

PR #9260, opened 2023-09-07
Add support for lazy-loaded React components

fzaninotto:
## Problem

To use lazy-loaded commponents as Resource routes, developers must override the default layout to add `<Suspense>`.

## Solution

- include `<Suspense>` in the default layout
- move the rendering delay to `<Loading>` so that it can be used directly as Suspense fallback
- tweak Loading UI to make it less ugly

Capture d’écran 2023-09-07 à 17 57 17


packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
```diff
@@ -247,61 +250,27 @@ describe('<CoreAdminRoutes>', () => {
...
+                </MemoryRouter>
             );
             expect(screen.queryByText('PostList')).toBeNull();
-            expect(screen.queryByText('Loading')).toBeNull();
-            resolve();
-            await waitFor(() =>
-                expect(screen.queryByText('PostList')).not.toBeNull()
-            );
-        });
-        it('should show a loader when requireAuth is true and dataProvider.checkAuth() takes more than 1s to reply', async () => {
```
fzaninotto, line 274: This was testing that CoreAdminRouter displays the loader only after 1s. Now, this feature has moved to the Loader itself.

packages/ra-core/src/core/CoreAdminRoutes.spec.tsx
```diff
@@ -340,7 +309,7 @@ describe('<CoreAdminRoutes>', () => {
                 </CoreAdminContext>
             );
             expect(screen.queryByText('PostList')).toBeNull();
-            expect(screen.queryByText('Loading')).toBeNull();
+            expect(screen.queryByText('Loading')).not.toBeNull();
```
fzaninotto, line 312: As there is no delay, the test loader (which is NOT our `<Loading>` component) is shown right away

packages/ra-ui-materialui/src/layout/Loading.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+            height: 250,
+            width: 400,
+            border: '1px solid gray',
+        }}
+    >
+        <Loading />
+    </div>
+);
+
+export const FullApp = () => (
```
fzaninotto, line 41: This one doesn't work as expected: when clicking on a menu item, the entire app reloads... I have to investigate

fzaninotto: Fixed: the bug appeared in the story because it didn't have an auth provider, which shouldn't happen in reality when using a resource function.

djhi (approved):

PR #9262, opened 2023-09-08
Support for disabling error logging

### LINKED ISSUE #8738: Support for disabling error logging

The data provider will currently [log errors with console.error](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/dataProvider/useDataProvider.ts#L114-L116) if `NODE_ENV !== 'production'`.

If you have a handful of tests for your error states which use e.g. 

```ts
useDataProvider(testDataProvider({
  getList: jest.fn().mockRejectedValue(new Error('Test Error'))
}))
```

The console can quickly get very spam-y as every single (intentional) error gets logged.

I have a few suggestions (any of which I would be happy to submit a PR for)

1. (Simpler solution) Have another `process.env` flag that allows logs to be disabled (e.g. `process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test'` or similar)
2. Add an option on either `useDataProvider` or the `DataProvider` itself to override logging (e.g. `useDataProvider({ logging: false })` or `<AdminContext dataProvider={{ getList: jest.fn(), logging: false }}>
3. Add support for a `logger` at some level (perhaps on `AdminContext`) which defaults to `console` for backwards compatibility and have the `useDataProvider` do a call to `useLogger()` or similar

Running my tests with `NODE_ENV='production'` but that seems like bad practice.

NidhiSharma63:
close #8738

### DIFF
--- a/packages/ra-core/src/auth/usePermissions.ts
+++ b/packages/ra-core/src/auth/usePermissions.ts
@@ -51,7 +51,7 @@ const usePermissions = <Permissions = any, Error = any>(
             : async () => [],
         {
             onError: error => {
-                if (process.env.NODE_ENV !== 'development') {
+                if (process.env.NODE_ENV === 'development') {
                     console.error(error);
                 }
                 logoutIfAccessDenied(error);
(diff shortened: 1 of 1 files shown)


djhi (approved):

djhi:
Thanks!

PR #9264, opened 2023-09-08
[Demo] Fix preferences are shared between demos

mchaffotte:
### Problem
If this application is installed on the same domain as other demo applications. The store is overridden. The data are not synchronized anymore.

### Solution
Add a custom store with a dedicated application key.

### DIFF
--- a/examples/demo/src/App.tsx
+++ b/examples/demo/src/App.tsx
@@ -1,22 +1,20 @@
-import * as React from 'react';
-import { Admin, CustomRoutes, Resource } from 'react-admin';
 import polyglotI18nProvider from 'ra-i18n-polyglot';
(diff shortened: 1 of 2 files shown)


examples/demo/src/App.tsx
```diff
@@ -40,6 +38,7 @@ const App = () => (
         dataProvider={dataProviderFactory(
             process.env.REACT_APP_DATA_PROVIDER || ''
         )}
+        store={localStorageStore(undefined, 'Demo')}
```
djhi, line 41:
```suggestion
        store={localStorageStore(undefined, 'ECommerce')}
```

djhi (changes requested):

djhi (approved):

Issue #9108, opened 2023-07-19
React does not recognize the `hasCreate` prop on a DOM element.


sebastianbuechler:
**What happened:**
When using a `List` element I get this in the console:

```
VM1174:1 Warning: React does not recognize the `hasCreate` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hascreate` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    at p
    at div
    at http://localhost:3000/node_modules/.vite/deps/chunk-BTXKACF2.js?v=50442c42:1404:49
    at ListView (http://localhost:3000/node_modules/.vite/deps/react-admin.js?v=8ddc62c8:21892:20)
    at ListContextProvider (http://localhost:3000/node_modules/.vite/deps/chunk-S7TE75VY.js?v=50442c42:10039:18)
    at ResourceContextProvider (http://localhost:3000/node_modules/.vite/deps/chunk-S7TE75VY.js?v=50442c42:13723:21)
    at ListBase (http://localhost:3000/node_modules/.vite/deps/chunk-S7TE75VY.js?v=50442c42:10286:21)
    at List (http://localhost:3000/node_modules/.vite/deps/react-admin.js?v=8ddc62c8:22065:24)
```

**Steps to reproduce:**
Use `List` element like this:
```
    <List
      resource="client/orders"
      sort={{ field: 'creationTimestamp', order: 'DESC' }}
      exporter={false}
    >
      <Datagrid bulkActionButtons={false}>
        <DateField source="creationTimestamp" label="Creation Timestamp" showTime />
      </Datagrid>
    </List>
```

**Environment**

* React-admin version: 4.12.1
* Last version that did not exhibit the issue (if applicable): -
* React version: 18
* Browser: Chrome and Firefox

slax57:
Thanks for this report.

I don't really understand how you can run into this issue with the code you provided, since it doesn't set `hasCreate` at all.
Can you provide a reproduction?

In any case, there's a good chance this issue will be fixed by https://github.com/marmelab/react-admin/pull/9101 once it is released.

slax57:
EDIT: my bad, actually https://github.com/marmelab/react-admin/pull/9101 cleans other props but not this one which is still needed for some advanced cases including an EE component.

But still, I would require a reproduction sandbox, as I did not reproduce in our simple example.

fzaninotto:
No news for some time, closing.

[...]

PR #9267, opened 2023-09-12
Fix `<TopToolbar>`: React does not recognize the hasCreate prop on a DOM element

slax57:
Fix https://github.com/marmelab/react-admin/issues/9108

### DIFF
--- a/packages/ra-ui-materialui/src/layout/TopToolbar.tsx
+++ b/packages/ra-ui-materialui/src/layout/TopToolbar.tsx
@@ -12,7 +12,7 @@ export const TopToolbar = (props: ToolbarProps) => {
         <StyledToolbar
             disableGutters
             variant={isXSmall ? 'regular' : 'dense'}
-            {...props}
+            {...sanitizeToolbarRestProps(props)}
         />
     );
 };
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/TopToolbar.tsx
```diff
@@ -46,3 +46,5 @@ const StyledToolbar = styled(Toolbar, {
         paddingBottom: 0,
     },
 }));
+
+const sanitizeToolbarRestProps = ({ hasCreate, ...props }: any) => props;
```
djhi, line 50: You should probably add the others as well, `hasList`, etc

djhi (changes requested):

packages/ra-ui-materialui/src/layout/TopToolbar.tsx
slax57: `hasCreate` is the only superfluous prop injected by [`ListView`](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/ListView.tsx#L60) (through [`ListToolbar`](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/list/ListToolbar.tsx#L21)).

fzaninotto (approved):

PR #9270, opened 2023-09-12
[Doc] Improve and reorder ReferenceManyToManyInput documentation

adguernier:
## Problem
The ReferenceManyToManyInput documentation is almost empty.

## Solution
Improve and reorder ReferenceManyToManyInput documentation in the way of the other documentation pages

### DIFF
--- a/docs/ReferenceManyToManyInput.md
+++ b/docs/ReferenceManyToManyInput.md
@@ -5,7 +5,7 @@ title: "The ReferenceManyToManyInput Component"
 
 # `<ReferenceManyToManyInput>`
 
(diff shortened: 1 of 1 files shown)


fzaninotto (approved):

fzaninotto:
Much better!

adguernier:
I notice some props in the table no longer listed alphabetically like other pages. Is this intentional?

fzaninotto:
> I notice some props in the table no longer listed alphabetically like other pages. Is this intentional?

Yes, we put the required props first, then the optional props (only in the table)

Issue #9265, opened 2023-09-09
RichTextInput onBlur validation doesn't trigger


botjaeger:
**What you were expecting:**

It should trigger a validation error when touched

**What happened instead:**

It does nothing

**Steps to reproduce:**

```
<SimpleForm mode='all' reValidateMode='onChange'>
    ...
    <CustomRichTextInput />
</SimpleForm >
```

```
function CustomRichTextInput({ ...props }) {
    return (
        <RichTextInput
            fullWidth
            toolbar={(
                <RichTextInputToolbar size='small'>
                    <FormatButtons size='small' />
                    <ListButtons size='small' />
                </RichTextInputToolbar>
            )}
            validate={required()}
            {...props}
        />
    );
}
```

**Other information:**

**Environment**

* React-admin version: 3.14.3
* Last version that did not exhibit the issue (if applicable):
* React version:
* Browser:
* Stack trace (in case of a JS error):

slax57:
Reproduced, thanks

DxVxDOD:
Hi could I work on this ?

fzaninotto:
@DxVxDOD There is already someone working on this (see the PR mentioned above your message).

DxVxDOD:
Oh yeah... skipped my eye :)

PR #9272, opened 2023-09-12
Fix `<RichTextInput>` does not trigger `onBlur`

slax57:
Fix https://github.com/marmelab/react-admin/issues/9265

### DIFF
--- a/packages/ra-input-rich-text/src/RichTextInput.tsx
+++ b/packages/ra-input-rich-text/src/RichTextInput.tsx
@@ -1,3 +1,5 @@
+import * as React from 'react';
+import { ReactElement, ReactNode, useEffect } from 'react';
 import { FormHelperText } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import { Color } from '@tiptap/extension-color';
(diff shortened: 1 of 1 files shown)


mchaffotte (approved):

packages/ra-input-rich-text/src/RichTextInput.tsx
```diff
@@ -1,26 +1,26 @@
...
+import clsx from 'clsx';
 import { useInput, useResourceContext } from 'ra-core';
 import {
     CommonInputProps,
     InputHelperText,
     Labeled,
     LabeledProps,
 } from 'ra-ui-materialui';
-import { TiptapEditorProvider } from './TiptapEditorProvider';
+import * as React from 'react';
```
fzaninotto, line 20: I don't know why you did that, but every React file in react-admin has this line as first and I don't think we should change it.

fzaninotto (changes requested):

adguernier (approved):

fzaninotto (approved):

PR #9279, opened 2023-09-15
Additional exports for customization in ra-data-graphql-simple

### LINKED ISSUE #9278: Export additional functions from ra-data-graphql-simple for customization

I'm using react-admin with a GraphQL backend with `ra-data-graphql-simple`. Noticed an unexpected behavior in the following case:

- Have an edit page where there is no redirection on save.
- Have an input with a value already set. (e.g. `DateInput` but can be `TextField` as well)
- Clear the value.
- Save the form, will trigger the update call.
- API returns the updated record where the cleared property would be `null` now, since instead of an empty string a `null` will be saved to the database. Especially in case of dates.
- Edit form updates with the new data received from the data provider and cache.
- Affected input field is now displaying the previous value that was there before clearing.

I found out that this is due to `ra-data-graphql-simple`'s the default response parser from this line:

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/getResponseParser.ts#L35

If the response object has a property with a `null` value, it won't be given back, so when the cache is updated with the response, the cleared property will be omitted during merge, thus preserving the old value.

Not sure if the above is a bug or could be a valid scenario in some cases. I would remove the `null` check, while preserving the `undefined` check.
If that change would be too much (could be considered a breaking change depending on scenario?) my idea would be to export additional functions from `ra-data-graphql-simple`. Currently the default export is the one building the data provider, while the only named export is `buildQuery`, which is the `defaultBuildQuery` without any customization:

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/index.ts#L10

I would propose to export most other functions so they can be accessed from outside the library:

`buildQueryFactory`

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/buildQuery.ts#L6

`buildVariables`

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/buildVariables.ts#L23

`buildGqlQuery`

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/buildGqlQuery.ts#L24

`getResponseParser`

https://github.com/marmelab/react-admin/blob/7eaef5e3d1b66a154563492a7d10aa28ba637fe6/packages/ra-data-graphql-simple/src/getResponseParser.ts#L6

This would provide more customization by calling `buildQueryFactory(buildVariables, buildGqlQuery, myCustomResponseParser)` for example.

In short I could imagine removing the `null` check in the response parser as a potential bug fix.

Adding additional exports wouldn't have any drawbacks from what I see.

**Additional context**

No additional context. Please let me know your thoughts about this. Thank you!

dricholm:
Resolves #9278 

## Problem

Without writing a complete build query factory it's not possible to customize `ra-data-graphql-simple`'s default build query. If we only want to modify the response parser for example, we would need to write build GraphQL query and variables logic as well.

## Solution

Export `buildQueryFactory`, `buildGqlQuery`, `buildVariables` and `getResponseParser` from the package to allow for more granular customization.

### DIFF
--- a/packages/ra-data-graphql-simple/src/index.ts
+++ b/packages/ra-data-graphql-simple/src/index.ts
@@ -3,11 +3,16 @@ import buildDataProvider, { BuildQueryFactory, Options } from 'ra-data-graphql';
 import { DataProvider, Identifier } from 'ra-core';
 
 import defaultBuildQuery from './buildQuery';
+
+export const buildQuery = defaultBuildQuery;
+export { buildQueryFactory } from './buildQuery';
+export { default as buildGqlQuery } from './buildGqlQuery';
+export { default as buildVariables } from './buildVariables';
+export { default as getResponseParser } from './getResponseParser';
+
 const defaultOptions = {
     buildQuery: defaultBuildQuery,
 };
-
-export const buildQuery = defaultBuildQuery;
 
 export default (
     options: Omit<Options, 'buildQuery'> & { buildQuery?: BuildQueryFactory }


djhi (changes requested):
Thanks! As we are exposing more things publicly, would you mind targeting the `next` branch?

Issue #9225, opened 2023-08-28
FilterButton adds filter in arbitrary order


fzaninotto:
**What you were expecting:**
 
Adding a new filter using the FilterButton shows the filter at the end of the current filter list

**What happened instead:**

The new filter appears at an arbitrary position

**Steps to reproduce:**

1. Open the e-commerce demo https://marmelab.com/react-admin-demo/#/
2. Click on the Orders menu
3. Add the "Min amount" filter
4. Add the "customer" filter
5. The Customer filter appears in the middle instead of the end of the form

https://github.com/marmelab/react-admin/assets/99944/214b01e2-7ff3-45d2-a462-773717fbd32c

**Related code:**

This comes from the fact that the `displayedFilters` uses an object instead of an array:

https://github.com/marmelab/react-admin/blob/4089ad726b9ae699173790a6f97685a22426e34a/packages/ra-core/src/controller/list/queryReducer.ts#L95-L120

Why is it an object? This choice was made 7 years ago, when first implementing this feature:

https://github.com/marmelab/react-admin/commit/2e2604d0d93a420dc61e1b30d5f6dada501e4f12

**Possible Solution**

Store displayed filters as an array instead. But this breaks backwards compatibility for those with existing filters in localStorage, so the change must allow the code to work with an old data structure. 

**Environment**

* React-admin version: all
* Last version that did not exhibit the issue (if applicable): none

azizChebbi:
@fzaninotto can i get assigned to this one ?

fzaninotto:
No need to assign. Just PR when you're ready.

PR #9281, opened 2023-09-15
[TypeScript] Fix order of filters

mchaffotte:
linked issue: https://github.com/marmelab/react-admin/issues/9225

- [ ] add new param shownFIlters (work in parallel of displayedFilters)
- [ ] stop saving displayedFilters in the local storage
- [ ] keep the displayedFilters works in URI but when changing the filters, the URI changes to the new param
- [ ] add tests


fzaninotto:
The estimated remaining effort to properly address the underlying issue is currently too high for the expected benefit. We're putting this PR on hold to deal with more important issues.

PR #9285, opened 2023-09-15
[Doc] Fix a typo in `Features.md`

portwatcher:

PR #9286, opened 2023-09-16
[Doc] Fix typo in Features example

fzaninotto:
Supersedes #9285

### DIFF
--- a/docs/Features.md
+++ b/docs/Features.md
@@ -70,6 +70,7 @@ import { useDataProvider } from 'react-admin';
 const PostList = () => {
     const [posts, setPosts] = useState();
     const [error, setError] = useState([]);
+    const [isLoading, setIsLoading] = useState(true);
     const dataProvider = useDataProvider();
     useEffect(() => {
         dataProvider.getList('posts', { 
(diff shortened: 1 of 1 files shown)


docs/Features.md
```diff
@@ -70,6 +70,7 @@ import { useDataProvider } from 'react-admin';
 const PostList = () => {
     const [posts, setPosts] = useState();
     const [error, setError] = useState([]);
+    const [isLoading, setIsLoading] = useState(true);
```
slax57, line 73: #9285 also changed the default values for the `posts` and `error` states, and they made more sense there IMO.

I think we should also include this change.

```suggestion
    const [posts, setPosts] = useState([]);
    const [error, setError] = useState();
    const [isLoading, setIsLoading] = useState(true);
```

docs/Features.md
```diff
@@ -78,9 +79,10 @@ const PostList = () => {
             filter: { status: 'published' } 
         })
             .then(({ data }) => setPosts(data))
-            .catch(error => setError(error));
+            .catch(error => setError(error))
+            .finally(() => setIsLoading(false));
     }, []);
-    if (posts) { return <Loading />; }
+    if (isLoading) { return <Loading />; }
```
slax57, line 85: `<Loading />` is undefined I believe.

```suggestion
    if (isLoading) { return <p>Loading</p>; }
```

slax57 (changes requested):

slax57 (approved):

PR #9290, opened 2023-09-18
Fix `<PrevNextButton>` default style

fzaninotto:
## Problems

- `<PrevNextButton>` takes the entire line by default. It should be inline instead of block
- `<PrevNextButton>` doesn't use the base Typography
- `<PrevNextButton>` adds useless markup

## Solution

Use `<Stack>` with custom styling

The change of inner class name is technically a breaking change, but I suggest to accept it because the feature is very recent and wasn't properly styled before. 

## Before
Capture d’écran 2023-09-19 à 00 12 50

## After
Capture d’écran 2023-09-19 à 00 12 27 1

### DIFF
--- a/packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
+++ b/packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
@@ -8,7 +8,14 @@ import {
 import { NavigateBefore, NavigateNext } from '@mui/icons-material';
 import ErrorIcon from '@mui/icons-material/Error';
 import { Link } from 'react-router-dom';
-import { Box, IconButton, SxProps, styled } from '@mui/material';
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/button/PrevNextButtons.tsx
```diff
@@ -108,7 +115,7 @@ export const PrevNextButtons = <RecordType extends RaRecord = any>(
 
     if (isLoading) {
         return (
-            <Box minHeight={theme => theme.spacing(4)}>
```
fzaninotto, line 111: I can't keep consistent height between the empty / loading / loaded versions unless I switch to pixels

djhi (approved):

PR #9300, opened 2023-09-21
[Doc] Add InputHelperText component to examples

ZachSelindh:
In the current docs, a plain `helperText` argument is supplied to the Material-UI input in various input examples. This will result in improper/missing translation of the helperText and/or fieldState.error.

This PR adds the use of the InputHelperText component to said examples.

Thanks!


fzaninotto:
Thanks for your contribution. 

I think it makes the examples too complex, too soon. I suggest that you introduce this component later, in a section specifically dedicated to i18n.

ZachSelindh:
> I think it makes the examples too complex, too soon. I suggest that you introduce this component later, in a section specifically dedicated to i18n.

I'm inclined to agree about the complexity, but I'll point out that without my changes, the copy-pasted previous example code will render untranslated RA errors in plain string; i.e. '@@react-admin@@validate.required'.

That said, I'd be happy to add a new section that explains the use of InputHelperText if that's your preferred solution!

fzaninotto:
> the copy-pasted previous example code will render untranslated RA errors in plain string; i.e. '@https://github.com/react-admin@@validate.required'.

You're right, but I prefer that we make the complexity progressive. So please doc the component in a follow-up section.

ZachSelindh:
Reopening targeting master with new revisions

PR #9303, opened 2023-09-22
[TypeScript] add state props in (Create)Button

mchaffotte:
- [x] add state props in button component
- [x] update e-commerce demo to show how to use it 
- [ ] add tests
- [x] update documentation

### DIFF
--- a/examples/demo/src/products/CreateRelatedReviewButton.tsx
+++ b/examples/demo/src/products/CreateRelatedReviewButton.tsx
@@ -1,0 +1,17 @@
+import * as React from 'react';
+import { CreateButton, useRecordContext } from 'react-admin';
+
+const CreateRelatedReviewButton = () => {
+    const record = useRecordContext();
+
+    return (
+        <CreateButton
+            to={{
+                pathname: '/reviews/create',
+            }}
+            state={{ record: { product_id: record.id } }}
+        />
+    );
+};
+
+export default CreateRelatedReviewButton;
--- a/examples/demo/src/products/CreateRelatedReviewButton.tsx
+++ b/examples/demo/src/products/CreateRelatedReviewButton.tsx
@@ -6,9 +6,7 @@ const CreateRelatedReviewButton = () => {
 
     return (
         <CreateButton
-            to={{
-                pathname: '/reviews/create',
-            }}
+            resource="reviews"
             state={{ record: { product_id: record.id } }}
         />
     );
--- a/examples/demo/src/reviews/StarRatingInput.tsx
+++ b/examples/demo/src/reviews/StarRatingInput.tsx
@@ -1,0 +1,37 @@
+import { Box, Rating, Typography, styled } from '@mui/material';
+import Icon from '@mui/icons-material/Stars';
+import { InputProps, useInput, useTranslate } from 'react-admin';
+
+const StyledRating = styled(Rating)({
+    '& .MuiRating-iconFilled': {
+        color: '#000',
+    },
+    '& .MuiRating-iconHover': {
+        color: '#000',
+    },
+});
+
+const StarRatingInput = (props: InputProps) => {
+    const { name = 'resources.reviews.fields.rating' } = props;
+    const { field } = useInput(props);
+    const translate = useTranslate();
+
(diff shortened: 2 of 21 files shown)


packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -46,7 +48,7 @@
         <StyledFab
             component={Link}
             to={createPath({ resource, type: 'create' })}
-            state={scrollStates[String(scrollToTop)]}
+            state={merge(scrollStates[String(scrollToTop)], state)}
```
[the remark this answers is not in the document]
fzaninotto: IMO, false alert - lodash merge does escape the `__proto__` key.

https://github.com/lodash/lodash/blob/a79c5c434cdecb36f4763cba30b4f1af2b7c392c/test/__proto__-property-bugs.spec.js#L45-L54

examples/demo/src/products/CreateRelatedReviewButton.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { CreateButton, useRecordContext } from 'react-admin';
+
+const CreateRelatedReviewButton = () => {
+    const record = useRecordContext();
+
+    return (
+        <CreateButton
+            to={{
+                pathname: '/reviews/create',
+            }}
```
fzaninotto, line 11: Rather than setting the `to` prop, you should set the `resource` prop and let the button manage the `to` itself.

examples/demo/src/reviews/StarRatingInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const { name = 'resources.reviews.fields.rating' } = props;
+    const { field } = useInput(props);
+    const translate = useTranslate();
+
+    return (
+        <Box
+            sx={{
+                flexDirection: 'column',
+                marginBottom: '1rem',
+            }}
```
fzaninotto, line 24: `Box` supports inline style props, I find it more legible. Besides, you're already using that capability for `display`.

```jsx
<Box  flexDirection="column" marginBottom="1rem" display="flex">
```

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -36,11 +38,12 @@ export const Button = (props: ButtonProps) => {
         color = 'primary',
         size = 'small',
         to: locationDescriptor,
+        state = {},
         ...rest
     } = props;
     const translate = useTranslate();
     const translatedLabel = label ? translate(label, { _: label }) : undefined;
-    const linkParams = getLinkParams(locationDescriptor);
+    const linkParams = merge(getLinkParams(locationDescriptor), { state });
```
fzaninotto, line 46: I find it weird (dangerous?) to add this capability to `Button` while we only need it in `CreateButton`.

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -59,7 +61,7 @@
         <StyledButton
             component={Link}
             to={createPath({ resource, type: 'create' })}
-            state={scrollStates[String(scrollToTop)]}
+            state={merge(scrollStates[String(scrollToTop)], state)}
```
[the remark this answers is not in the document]
fzaninotto: I suggest you precompute the state and extract it into a variable to avoid this copy/paste

fzaninotto (changes requested):

packages/ra-ui-materialui/src/button/Button.tsx
mchaffotte: What do you think @djhi, @slax57 about it? This is the behaviour I understood, isn't it?

djhi: As our Button already supports and already uses the `getLinkParams`, I think it makes sense to do it here

fzaninotto: I think we should only add it if we need it.

djhi:
Superseded by #9319

PR #9305, opened 2023-09-22
[Doc] Fix anchors

WiXSL:


docs/List.md
```diff
@@ -60,7 +60,7 @@ You can find more advanced examples of `<List>` usage in the [demos](./Demos.md)
 | `disable Authentication`  | Optional | `boolean`      | `false`        | Set to `true` to disable the authentication check.                                           |
 | `disable SyncWithLocation`| Optional | `boolean`      | `false`        | Set to `true` to disable the synchronization of the list parameters with the URL.            |
 | `empty`                   | Optional | `ReactElement` | -              | The component to display when the list is empty.                                             |
-| `empty WhileLoading`      | Optional | `boolean`      | `false`        | Set to `true` to return `null` while the list is loading.                                    |
```
fzaninotto, line 63: This was intended: it makes the rendered table more balanced. We accept the fact that the prop name is cut in two lines in that case.

fzaninotto (changes requested):
Nice!

docs/List.md
WiXSL: Ok

fzaninotto (approved):

PR #9309, opened 2023-09-25
Fix package exports

djhi:
## Problem

When using react-admin in some applications that render it on the server using node (such as Remix), our packages triggers errors.

## Solution

Add an `exports` field to our `package.json` files to ensure they can imported in all scenarios

### DIFF
--- a/examples/data-generator/package.json
+++ b/examples/data-generator/package.json
@@ -17,6 +17,7 @@
         },
         "./package.json": "./package.json"
     },
+    "sideEffects": false,
     "scripts": {
         "build": "yarn run build-cjs && yarn run build-esm",
         "build-cjs": "rimraf ./dist/cjs && tsc --outDir dist/cjs",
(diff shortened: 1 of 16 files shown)


mchaffotte (approved):

smeng9:
Hi @djhi 

The `package.json` exports field fix looks good to me.

This may fix our vitest issue with react-hook-form https://stackblitz.com/edit/vitejs-vite-2etbdy?file=src%2Fposts.test.tsx

Do you know why this is closed instead of merged?

djhi:
I opened this one instead: https://github.com/marmelab/react-admin/pull/9310. Making this change on some packages (those that export react components) broke our remix example for some reason, so we're not comfortable to do it for all packages.

smeng9:
I see the updated merge request. I will try if the next version of react-admin fixes out vitest issue with react-hook-form.

For now anything uses form context in a react-admin form breaks in vitest but works in vite.

PR #9310, opened 2023-09-25
Fix package exports

djhi:
This fixes imports in frameworks such as Remix


smeng9:
Continuing the discussion from https://github.com/marmelab/react-admin/pull/9309#issuecomment-1735998501 here. 

After reviewing changed files if we want to solve the module import/export test case for vitest in previous discussion we at least have to touch the `package.json` of `ra-core` to include an `exports` field.

If we are only targeting solving remix in this pull request it's another story, and I can open a different issue for vitest on react-admin issue list.

djhi:
Actually not needed with Remix

Issue #9261, opened 2023-09-08
Setting 'disabled' to false in input components crashes the app


marit970:
This issue appears when updating the react-hook-form package (^7.46.0)

**What you were expecting:**
The input field is not disabled when setting the 'disabled'-property to false 

**What happened instead:**
It works correctly when setting the 'disabled'-property to 'true' or 'undefined', but it crashes the app when using 'false'

**Related code:**

* A link to a GitHub repo with the minimal codebase to reproduce the issue
https://github.com/marit970/react-disabled-bug

**Environment**

* React-admin version: 4.13.3
* React-hook-form version: 7.46.1 
* Last version that did not exhibit the issue (if applicable): 7.45.4 (react-hook-form)
* React version: ^18.2.0
* Browser: Mozilla Firefox
* Stack trace (in case of a JS error):
![Bildschirmfoto von 2023-09-08 11-16-55](https://github.com/marmelab/react-admin/assets/43189340/4f807363-5bc3-47ec-843a-aa9bb84fcf3f)

fzaninotto:
This seems like a react-hook-form regression. Can you try and install https://github.com/react-hook-form/react-hook-form/releases/tag/v7.46.1, which fixes a similar regression?

marit970:
@fzaninotto I've tried it out. This issue appears starting from the v7.46.0 (in 7.46.0 and 7.46.1 as well).

fzaninotto:
Then it's a react-hook-form regression. Could you open an issue in their repository?

marit970:
@fzaninotto sure, thank you!

marit970:
@fzaninotto here is my issue by react-hook-form: https://github.com/react-hook-form/react-hook-form/issues/10907. I have not defined any resource in my example repo, but nothing changes when I add one. Also, I have not found anything in the documentation. Maybe you have an idea of what is wrong/missing here?

qpointsystems:
Boy this killed me today, lol.  Finally tried checking here and violà, someone has logged it, thank god. 
Never thought it'd be in RA or hook form. 
Phew.

slax57:
I have tried to reproduce the issue inside our [simple example project](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple), after upgrading to RHF v7.46.1, but didn't get the error.

@marit970 I see that your reproduction code is somewhat incomplete, for instance it does not provide any `ResourceContext`.
[...]

PR #9313, opened 2023-09-25
Fix inputs with `disabled={false}` throw an error (workaround)

slax57:
Fix https://github.com/marmelab/react-admin/issues/9261

Actually this is a workaround, since the root issue has to be fixed on RHF's side (https://github.com/react-hook-form/react-hook-form/issues/10907)

Cannot be tested automatically.

To test:
- Upgrade to RHF >= 7.46.0
- Upgrade to React 18
- Set `disabled={false}` somewhere in the demo


djhi (approved):

fzaninotto:
TypeScript complains

djhi (approved):

cyonder:
I'm curious to know when this will be released. How often do we have releases?

djhi:
We usually release patch version every week. This one is scheduled for today

PR #9315, opened 2023-09-25
Add i18n InputHelpterText section to Inputs/useInputs docs

ZachSelindh:
This PR adds an i18n section to the Inputs.md doc instructing users on the use of the `InputHelperText` component.

A **tip** referencing the use of said component has also been added to useInput.md.

I also removed a redundant `formState` reminder from useInput.md

Thanks!

### DIFF
--- a/docs/useInput.md
+++ b/docs/useInput.md
@@ -59,10 +59,10 @@ Additional props are passed to [react-hook-form's `useController` hook](https://
 ```jsx
 // in LatLongInput.js
 import TextField from '@mui/material/TextField';
-import { useInput, required } from 'react-admin';
+import { useInput, required, InputHelperText } from 'react-admin';
 
 const BoundedTextField = (props) => {
-    const { onChange, onBlur, ...rest } = props;
+    const { onChange, onBlur, label, helperText, ...rest } = props;
     const {
         field,
         fieldState: { isTouched, invalid, error },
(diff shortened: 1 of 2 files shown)


docs/useInput.md
```diff
@@ -99,14 +99,6 @@ const LatLngInput = props => {
...
 };
 ```
-
-**Reminder:** [react-hook-form's `formState` is wrapped with a Proxy](https://react-hook-form.com/docs/useformstate/#rules) to improve render performance and skip extra computation if specific state is not subscribed. So, make sure you deconstruct or read the `formState` before render in order to enable the subscription.
-
-```js
-const { isDirty } = useFormState(); // ✅
-const formState = useFormState(); // ❌ should deconstruct the formState      
-```
-
```
djhi, line 109: I believe we should keep this one. @fzaninotto?

djhi (commented):
Thanks!

docs/useInput.md
fzaninotto: It's true that the reminder is mentioned twice in this chapter. I'd create a dedicated section explaining how to read the form state, and removing the reminder from the two sections avout TextField and Select.

ZachSelindh: @djhi Do you want to do it or should I? Rock paper scissors?

djhi: :rock: That's not really related to your PR so if you don't, we'll open a new PR for that.

ZachSelindh: I'm on it then, after all, I broached it by removing the repeated section.

slax57:
@ZachSelindh  I've fixed it for you. Should be good now.

slax57 (approved):

djhi (approved):

djhi:
Thanks!

PR #9316, opened 2023-09-26
Add more themes

adguernier:
## Problem

React-admin bundles only one light and one dark theme. Most users need to adapt the app theme to their brand, and the default themes aren't always a good starting point.

## Solution

Build 3 more themes showing variations in fonts, spacing, border-radius, colors, etc.

## Todo
- [x] build Nano theme
- [x] build Radiant theme
- [x] build House theme
- [x] provide these themes in ra-ui-materialui
- [x] move all theme objects, hooks and components into a `theme` folder
- [x] build a theme switcher in the e-commerce demo
- [x] add documentation to introduce these new themes

https://github.com/marmelab/react-admin/assets/99944/956739de-9ada-4639-aaaa-8088584b86e9


djhi:
image

This is bit too dark, can you try to ease it a little?

examples/demo/src/themes/ThemeSwipper.tsx
```diff
@@ -0,0 +1,51 @@
...
+    MenuItem,
+    Select,
+    SelectChangeEvent,
+    Stack,
+} from '@mui/material';
+import { useContext } from 'react';
+import { ToggleThemeButton, useStore, useTranslate } from 'react-admin';
+import { ThemeContext, ThemeType, themes } from './themeContext';
+
+export const ThemeSwipper = () => {
```
djhi, line 13:
```suggestion
export const ThemeSwapper = () => {
```

Don't forget to rename the file

examples/demo/src/themes/synthwaveTheme.ts
```diff
@@ -0,0 +1,263 @@
...
+    ...Button(),
+    ...Datagrid(theme),
+    ...FormControl(),
+    ...Input(),
+    ...ItemLink(theme),
+    ...Layout(),
+    ...Paper(theme),
+    ...Table(),
+    ...TextField(),
+    ...Toolbar(theme),
```
djhi, line 235: I'm not sure those independent functions bring much clarity. Can you merge their content here like in the other themes?

djhi (changes requested):
:heart_eyes:

slax57 (commented):
Shouldn't you target `next` instead of `master`?

fzaninotto:
Loading the 3 new themes adds about 15kB to the demo, which is acceptable and doesn't justify lazy-loading. 

image

packages/ra-language-english/src/index.ts
```diff
@@ -38,7 +38,7 @@ const englishMessages: TranslationMessages = {
             move_up: 'Move up',
             move_down: 'Move down',
             open: 'Open',
-            toggle_theme: 'Toggle Theme',
+            toggle_theme: 'Toggle light/dark mode',
```
djhi, line 41: Did you check the french translations?

packages/ra-ui-materialui/src/layout/LoadingIndicator.tsx
```diff
@@ -16,7 +16,12 @@ export const LoadingIndicator = (props: LoadingIndicatorProps) => {
     const theme = useTheme();
     return (
         <Root className={className} sx={sx}>
-            {loading ? (
+            <RefreshIconButton
```
djhi, line 19: Why?

packages/ra-ui-materialui/src/theme/houseTheme.ts
```diff
@@ -0,0 +1,184 @@
...
+            scrollableX: {
+                overflow: 'visible !important',
+            },
+        },
+    },
+    MuiTextField: {
+        defaultProps: {
+            variant: 'outlined' as const,
+        },
+    },
```
adguernier, line 104: Add margin between glued input

```suggestion
    },
    MuiFormControl: {
        defaultProps: {
            margin: 'normal' as const,
        },
    },
```

packages/ra-language-english/src/index.ts
fzaninotto: Yes, the French translation is already "Thème clair/sombre", which is explicit. 

I changed the English translation because it kind of conflicted with the "change theme" tooltip of the ThemeSwapper button in the demo. Now it is more specific.

packages/ra-ui-materialui/src/layout/LoadingIndicator.tsx
fzaninotto: LoadingIndicator needs to replace the circular arrow icon by a circular loader, and it needs to do it precisely for the loading effect to be nice. The current way of doing it didn't work with a different `spacing` preference in the theme, and couldn't be adjusted in the theme using component `styleOverrides`. 

The new way of showing the circular loader is more robust, can be overridden, and adapts well to smaller font sizes (see the Nano theme).

packages/ra-ui-materialui/src/theme/houseTheme.ts
fzaninotto: Fixed in 0bc94a4a2f126a057677e5c312eb3f97abc02bb3, thanks

packages/ra-ui-materialui/src/button/LocalesMenuButton.tsx
```diff
@@ -48,6 +48,7 @@ export const LocalesMenuButton = (props: LocalesMenuButtonProps) => {
         <Root component="span">
             <Button
                 color="inherit"
+                variant="text"
```
fzaninotto, line 51: This is to avoid an outline when the theme sets the button variant to `outlined`

djhi (approved):

PR #9319, opened 2023-09-26
Add ability to pass url state in `<CreateButton>`

mchaffotte:
Link to: https://github.com/marmelab/react-admin/pull/9303 due to too many modifications the PR should be based on next version.

- [x] add state props in button component
- [x] update e-commerce demo to show how to use it
- [x] update documentation

### DIFF
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -467,7 +467,7 @@ const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
         <CreateButton
-            resource="reviews"
+            resource="comments"
             to={{
                 search: `?source=${JSON.stringify({ post_id: record.id })}`,
             }}
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -424,7 +424,7 @@ You can do the same for error notifications, by passing a custom `onError`  call
 
 ## Prefilling the Form
 
-You sometimes need to pre-populate a record based on a *related* record. For instance, to create a comment related to an existing post. 
+You sometimes need to pre-populate a record based on a *related* record. For instance, to create a comment related to an existing post.
 
 By default, the `<Create>` view starts with an empty `record`. However, if the `location` object (injected by [react-router-dom](https://reacttraining.com/react-router/web/api/location)) contains a `record` in its `state`, the `<Create>` view uses that `record` instead of the empty object. That's how the `<CloneButton>` works under the hood.
 
@@ -433,22 +433,17 @@ That means that if you want to create a link to a creation form, presetting *som
 {% raw %}
 ```jsx
 import * as React from 'react';
-import { Datagrid, useRecordContext } from 'react-admin';
-import { Button } from '@mui/material';
-import { Link } from 'react-router-dom';
+import { CreateButton, Datagrid, useRecordContext } from 'react-admin';
 
 const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
-        <Button
-            component={Link}
+        <CreateButton
             to={{
                 pathname: '/comments/create',
             }}
             state={{ record: { post_id: record.id } }}
-        >
-            Write a comment for that post
-        </Button>
+        />
     );
 };
 
@@ -463,29 +458,22 @@ export default PostList = () => (
 ```
 {% endraw %}
 
-**Tip**: To style the button with the main color from the Material UI theme, use the `Link` component from the `react-admin` package rather than the one from `react-router-dom`.
-
 **Tip**: The `<Create>` component also watches the "source" parameter of `location.search` (the query string in the URL) in addition to `location.state` (a cross-page message hidden in the router memory). So the `CreateRelatedCommentButton` could also be written as:
 
 {% raw %}
 ```jsx
 import * as React from 'react';
-import { useRecordContext } from 'react-admin';
-import Button from '@mui/material/Button';
-import { Link } from 'react-router-dom';
+import { CreateButton, useRecordContext } from 'react-admin';
 
 const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
-        <Button
-            component={Link}
+        <CreateButton
             to={{
                 pathname: '/comments/create',
                 search: `?source=${JSON.stringify({ post_id: record.id })}`,
             }}
-        >
-            Write a comment for that post
-        </Button>
+        />
     );
 };
 ```
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -428,7 +428,7 @@ You sometimes need to pre-populate a record based on a *related* record. For ins
 
 By default, the `<Create>` view starts with an empty `record`. However, if the `location` object (injected by [react-router-dom](https://reacttraining.com/react-router/web/api/location)) contains a `record` in its `state`, the `<Create>` view uses that `record` instead of the empty object. That's how the `<CloneButton>` works under the hood.
 
-That means that if you want to create a link to a creation form, presetting *some* values, all you have to do is to set the location `state`. `react-router-dom` provides the `<Link>` component for that:
+That means that if you want to create a link to a creation form, presetting *some* values, all you have to do is to set the prop `state` of the `<CreateButton>`:
 
 {% raw %}
 ```jsx
@@ -439,7 +439,7 @@ const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
         <CreateButton
-            resource="reviews"
+            resource="comments"
             state={{ record: { post_id: record.id } }}
         />
     );
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -428,12 +428,12 @@ You sometimes need to pre-populate a record based on a *related* record. For ins
 
 By default, the `<Create>` view starts with an empty `record`. However, if the `location` object (injected by [react-router-dom](https://reacttraining.com/react-router/web/api/location)) contains a `record` in its `state`, the `<Create>` view uses that `record` instead of the empty object. That's how the `<CloneButton>` works under the hood.
 
-That means that if you want to create a link to a creation form, presetting *some* values, all you have to do is to set the prop `state` of the `<CreateButton>`:
+That means that if you want to create a link to a creation form, presetting *some* values, all you have to do is to set the `state` prop of the `<CreateButton>`:
 
 {% raw %}
 ```jsx
 import * as React from 'react';
-import { CreateButton, Datagrid, useRecordContext } from 'react-admin';
+import { CreateButton, Datagrid, List, useRecordContext } from 'react-admin';
 
 const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -439,9 +439,7 @@ const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
         <CreateButton
-            to={{
-                pathname: '/comments/create',
-            }}
+            resource="reviews"
             state={{ record: { post_id: record.id } }}
         />
     );
@@ -469,8 +467,8 @@ const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
         <CreateButton
+            resource="reviews"
             to={{
-                pathname: '/comments/create',
                 search: `?source=${JSON.stringify({ post_id: record.id })}`,
             }}
         />
--- a/examples/demo/src/reviews/ReviewCreate.tsx
+++ b/examples/demo/src/reviews/ReviewCreate.tsx
@@ -15,11 +15,6 @@ import {
 import StarRatingInput from './StarRatingInput';
 import { useLocation } from 'react-router';
 
-const customerOptionRenderer = (choice: any) =>
-    `${choice.first_name} ${choice.last_name}`;
-
-const productOptionRenderer = (choice: any) => `${choice.reference}`;
-
 const ReviewCreate = () => {
     const notify = useNotify();
     const redirect = useRedirect();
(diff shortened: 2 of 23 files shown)


packages/ra-ui-materialui/src/button/CloneButton.tsx
```diff
@@ -61,7 +61,7 @@ interface Props {
     scrollToTop?: boolean;
 }
 
-export type CloneButtonProps = Props & ButtonProps;
+export type CloneButtonProps = Props & ButtonProps<typeof Link>;
```
djhi, line 64: You should exclude the `to` prop here or at least make it partial

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const productOptionRenderer = (choice: any) => `${choice.reference}`;
+
+const ReviewCreate = () => {
+    const notify = useNotify();
+    const redirect = useRedirect();
+    const location = useLocation();
+
+    const onSuccess = (_: any) => {
+        const record = getRecordFromLocation(location);
```
djhi, line 29: You shouldn't have to do that as `onSuccess` is called with the result of `dataProvider.create`. If our dataProvider for this example does not return it, please fix it

djhi (changes requested):
Almost there

docs/Create.md
```diff
@@ -433,22 +433,15 @@ That means that if you want to create a link to a creation form, presetting *som
 {% raw %}
 ```jsx
-import { Datagrid, useRecordContext } from 'react-admin';
-import { Button } from '@mui/material';
-import { Link } from 'react-router-dom';
+import { CreateButton, Datagrid, useRecordContext } from 'react-admin';
```
djhi, line 436: You should change the text of the previous paragraph at L431

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,71 @@
...
+    getRecordFromLocation,
+} from 'react-admin';
+
+import StarRatingInput from './StarRatingInput';
+import { useLocation } from 'react-router';
+
+const customerOptionRenderer = (choice: any) =>
+    `${choice.first_name} ${choice.last_name}`;
+
+const productOptionRenderer = (choice: any) => `${choice.reference}`;
```
djhi, line 21:
```suggestion
```

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <SimpleForm>
+                <ReferenceInput source="customer_id" reference="customers">
+                    <AutocompleteInput
+                        optionText={customerOptionRenderer}
+                        validate={required()}
+                    />
+                </ReferenceInput>
+                <ReferenceInput source="product_id" reference="products">
+                    <AutocompleteInput
+                        optionText={productOptionRenderer}
```
djhi, line 49:
```suggestion
                        optionText="reference"
```

You can provide a property name here for simple cases

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,71 @@
...
+    useRedirect,
+    getRecordFromLocation,
+} from 'react-admin';
+
+import StarRatingInput from './StarRatingInput';
+import { useLocation } from 'react-router';
+
+const customerOptionRenderer = (choice: any) =>
+    `${choice.first_name} ${choice.last_name}`;
+
```
djhi, line 20: We usually put utilities like this at the end of the file. Important things first

examples/demo/src/reviews/StarRatingInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const StyledRating = styled(Rating)({
+    '& .MuiRating-iconFilled': {
+        color: '#000',
+    },
+    '& .MuiRating-iconHover': {
+        color: '#000',
+    },
+});
+
```
djhi, line 13: Same, please move this at the end of the file

packages/ra-ui-materialui/src/button/CloneButton.tsx
```diff
@@ -29,10 +29,10 @@ export const CloneButton = (props: CloneButtonProps) => {
                           search: stringify({
                               source: JSON.stringify(omitId(record)),
                           }),
-                          state: { _scrollToTop: scrollToTop },
                       }
                     : pathname
             }
+            state={{ _scrollToTop: scrollToTop }}
```
djhi, line 35: This probably means we introduced a breaking change even though passing the previous `to` to the `Button` without a `component={Link}` didn't work before. I'm not sure what we should do here. @fzaninotto @slax57 @adguernier?

djhi (changes requested):

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { useLocation } from 'react-router';
+
+const ReviewCreate = () => {
+    const notify = useNotify();
+    const redirect = useRedirect();
+    const location = useLocation();
+
+    const onSuccess = (_: any) => {
+        const record = getRecordFromLocation(location);
+        notify(`Changes saved`);
```
fzaninotto, line 25: use an existing translation message instead - the EE demo is multilingual.

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,69 @@
...
+import StarRatingInput from './StarRatingInput';
+import { useLocation } from 'react-router';
+
+const ReviewCreate = () => {
+    const notify = useNotify();
+    const redirect = useRedirect();
+    const location = useLocation();
+
+    const onSuccess = (_: any) => {
+        const record = getRecordFromLocation(location);
```
fzaninotto, line 24: Why do you get the record from the location and not the record from the onSuccess callback parameter?

packages/ra-ui-materialui/src/button/CloneButton.tsx
fzaninotto: I agree that this is a breaking change and we should not do it. Instead, we must deprecate the support for the to prop in button, but keep the logic there.

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -41,12 +44,17 @@ const CreateButton = (props: CreateButtonProps) => {
     const isSmall = useMediaQuery((theme: Theme) =>
         theme.breakpoints.down('md')
     );
+    const state = merge(
+        scrollStates[String(scrollToTop)],
```
fzaninotto, line 48: security warning

fzaninotto (changes requested):

examples/demo/src/reviews/ReviewCreate.tsx
mchaffotte: I used the location in order to keep the possibility to create straightly reviews. If I use the parameter of the callback I do not know from where I come from and the redirection works in all cases.

djhi: In short: to handle cases where you don't come from a product edition view

fzaninotto: ah, I get it: the record isn't the created `review`, but a `product`. To clarify, you should perhaps name the variable `product`.

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,69 @@
...
+            redirect(`/reviews`);
+        }
+    };
+
+    return (
+        <Create mutationOptions={{ onSuccess }}>
+            <SimpleForm>
+                <ReferenceInput source="customer_id" reference="customers">
+                    <AutocompleteInput
+                        optionText={customerOptionRenderer}
```
fzaninotto, line 38: you should rely on the `recordRepresentation` instead.

packages/ra-ui-materialui/src/auth/AuthError.tsx
```diff
@@ -19,7 +20,7 @@ export const AuthError = (props: AuthErrorProps) => {
             <div className={AuthErrorClasses.message}>
                 <h1>{translate(title, { _: title })}</h1>
                 <div>{translate(message, { _: message })}</div>
-                <Button to="/login" label="ra.auth.sign_in">
+                <Button component={Link} to="/login" label="ra.auth.sign_in">
```
fzaninotto, line 23: I believe this change is no longer necessary?

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -54,8 +56,8 @@ export const Button = (props: ButtonProps) => {
                     className={className}
                     color={color}
                     size="large"
-                    {...rest}
```
fzaninotto, line 57: Why did you switch the order?

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -137,7 +141,9 @@ const StyledButton = styled(MuiButton, {
     },
 });
 
-const getLinkParams = (locationDescriptor?: LocationDescriptor | string) => {
+export const getLinkParams = (
```
fzaninotto, line 144: Why do you need to export it? I don't see it used anywhere else.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/button/Button.tsx
mchaffotte: In order to not overlap the value of 'state' prop coming from CreateButton (for example)

packages/ra-ui-materialui/src/auth/AuthError.tsx
mchaffotte: This writing is the new behaviour so it seems to be good to use it.

djhi: I believe this didn't work before this PR anyway as the `Button` did not set the `component` prop automatically

packages/ra-ui-materialui/src/button/Button.tsx
mchaffotte: Thanks I missed to add the behaviour in CreateButton

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,63 @@
...
+import StarRatingInput from './StarRatingInput';
+import { useLocation } from 'react-router';
+
+const ReviewCreate = () => {
+    const notify = useNotify();
+    const redirect = useRedirect();
+    const location = useLocation();
+
+    const onSuccess = (_: any) => {
+        const record = getRecordFromLocation(location);
```
djhi, line 24: It's not a product, it's an object containing default values for the new review. I would leave it as it is or rename it to `formInitialValues`

packages/ra-ui-materialui/src/button/Button.tsx
djhi: Can you remove the `export` now?

djhi (changes requested):

examples/demo/src/reviews/ReviewCreate.tsx
mchaffotte: If the parameter name is _,  it is due to a lint warning unused parameter. I prefer to let it like that. It is the review but from this parameter I have no clue to know if the review comes from a simple creation or a creation from a product.

djhi: @fzaninotto and I were referring to the `record` variable ;)

docs/Create.md
```diff
@@ -463,29 +456,22 @@ export default PostList = () => (
...
-import { Link } from 'react-router-dom';
+import { CreateButton, useRecordContext } from 'react-admin';
 
 const CreateRelatedCommentButton = () => {
     const record = useRecordContext();
     return (
-        <Button
-            component={Link}
+        <CreateButton
+            resource="reviews"
```
slax57, line 470: Why reviews and not comments? :thinking: 

```suggestion
            resource="comments"
```

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -41,29 +47,34 @@
     const isSmall = useMediaQuery((theme: Theme) =>
         theme.breakpoints.down('md')
     );
+    const state = merge(scrollStates[String(scrollToTop)], initialState);
```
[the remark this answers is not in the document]
slax57: You can probably replace the JS base object with a `Map` instead to mitigate this risk

slax57 (commented):
Otherwise, LGTM. :muscle: 

Also: please rename the PR title to make it clearer this PR actually implements a new feature, and does not only update the docs (and the demo) :wink:

examples/demo/src/reviews/ReviewCreate.tsx
```diff
@@ -0,0 +1,63 @@
...
+            redirect(`/reviews`);
+        }
+    };
+
+    return (
+        <Create mutationOptions={{ onSuccess }}>
+            <SimpleForm>
+                <ReferenceInput source="customer_id" reference="customers">
+                    <AutocompleteInput validate={required()} />
+                </ReferenceInput>
```
slax57, line 38: One more thing: the created reviews miss a status. You can probably use `defaultValues` to set the default status to be `'pending'`.

slax57 (changes requested):

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -41,50 +42,56 @@ const CreateButton = (props: CreateButtonProps) => {
...
```
slax57, line 45: The way to access values inside a `Map` is using `get()`

```suggestion
    const state = merge(scrollStates.get(String(scrollToTop)), initialState);
```

slax57 (changes requested):

slax57 (approved):

Issue #8609, opened 2023-01-30
Possibility to add label to rowsPerPageOptions prop of TablePagination component


murattishkul:
**Is your feature request related to a problem? Please describe.**
I need to have customizable label in rowsPerPageOptions prop of TablePagination component like in mui

**Describe the solution you'd like**
I want to change prop type of rowsPerPageOptions prop from number[] to { label: string; value: number; } | number[] like it is in mui

**Describe alternatives you've considered**
-

**Additional context**
<img width="1230" alt="image" src="https://user-images.githubusercontent.com/38204149/215457722-1fb7c48b-d5e2-49e3-971e-333a12823ae0.png">
<img width="668" alt="image" src="https://user-images.githubusercontent.com/38204149/215457769-402cf25d-ea58-4562-9708-db4fe0c39848.png">

slax57:
This seems to be a TS issue only.
It should already be supported if you add a `// @ts-ignore` comment. Can you confirm?
Would you like to open a PR to update the TS type?

murattishkul:
hi @slax57 , it is working with @ts-ignore indeed, however I would not like to use it 
sure, would like to make a pr

7twik:
I would like to work on this issue, pls assign me this issue

fzaninotto:
We don't have a formal assignment process. Just open a PR referencing that issue, and we'll do the review ;)

PR #9324, opened 2023-10-02
Added rowsPerPageOptions

7twik:
Fix of issue #8609 

Solution asked was to add label to rowsPerPageOptions
Referred to docs : https://mui.com/material-ui/api/table-pagination/

Fix SS: image

### DIFF
--- a/packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
+++ b/packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
@@ -111,7 +111,7 @@ export const Pagination: FC<PaginationProps> = memo(props => {
                 rowsPerPage={perPage}
                 page={page - 1}
                 onPageChange={handlePageChange}
-                rowsPerPageOptions={rowsPerPageOptions}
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
```diff
@@ -153,7 +153,12 @@ Pagination.propTypes = {
     rowsPerPageOptions: PropTypes.arrayOf(PropTypes.number),
 };
 
-const DefaultRowsPerPageOptions = [5, 10, 25, 50];
+const DefaultRowsPerPageOptions = [
+    { label: '5 per page', value: 5 },
+    { label: '10 per page', value: 10 },
+    { label: '25 per page', value: 25 },
+    { label: '50 per page', value: 50 },
```
djhi, line 160: Unfortunately, this is not translated. Can you introduce a new translation key that take `rowCount` parameter? And as we have to introduce a new translation, we must target the `next` branch.

djhi (changes requested):
Thanks!

PR #9301, opened 2023-09-21
[Docs] Update third-party component reference to new dateinput library

ZachSelindh:
The Third-Party Components section of the docs previously directed users seeking DateInput components to a now-deprecated library which had not been maintained since react-admin v3, and was not compatible with MUI v5.

This PR updates the docs to direct users to a package that I have deployed to npm, containing the pickers which I have used in my current project for the last couple of years. These pickers are fully MUI v5-compatible and use the latest react-admin standards (InputHelperText component, current patterns of useInput, etc.).

Thanks!

PR #9326, opened 2023-10-02
[Docs] Update third-party component section; new date/time inputs

ZachSelindh:
Clone of #9301 , which seems to have been closed because I initially targeted 'next'

The Third-Party Components section of the docs previously directed users seeking DateInput components to a now-deprecated library which had not been maintained since react-admin v3, and was not compatible with MUI v5.

This PR updates the docs to direct users to a package that I have deployed to npm, containing the pickers which I have used in my current project for the last couple of years. These pickers are fully MUI v5-compatible and use the latest react-admin standards (InputHelperText component, current patterns of useInput, etc.).

Thanks!


djhi (approved):
Thanks!

PR #9328, opened 2023-10-03
added rowsPerPage label and its translation

7twik:
For issue #8609 

As requested I have added label to rowsPerPageOptions
Referred to docs : https://mui.com/material-ui/api/table-pagination/
Screenshot: image

Also at PR #9324 , it was asked by @djhi sir that I should add the translation feature which I did
Screenshot: image

But unfortunately I was not able to fork 'next' branch of react-admin so I was unable to create the pull request to next branch as there was a lot of conflict if I tried to merge from master branch since the next branch is behind master branch by quiet some commits...

* [ Also a newbie mistake, I tried to fork the next branch in doing so I thought I need to delete my current forked repo as a result my last PR #9324 got closed, really sorry for that ]

### DIFF
--- a/packages/ra-core/src/i18n/TranslationMessages.ts
+++ b/packages/ra-core/src/i18n/TranslationMessages.ts
@@ -181,6 +181,12 @@ export interface TranslationMessages extends StringMap {
             remove_message: string;
             help: string;
         };
+        rowCount: {
+            per_page_5: string;
+            per_page_10: string;
+            per_page_25: string;
+            per_page_50: string;
+        };
         configurable?: {
             customize: string;
             configureMode: string;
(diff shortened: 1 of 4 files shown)


slax57 (changes requested):
Thank you for your contribution.

However, I'm afraid you are going in the wrong direction.

The [original issue](https://github.com/marmelab/react-admin/issues/8609) is about adding the ability to pass custom labels to the Pagination component via the `rowsPerPageOptions` prop (in which you can call `translate()` if needed).

As I said [here](https://github.com/marmelab/react-admin/issues/8609#issuecomment-1408668665), this is already supported!
The only issue is that we use an incorrect TS type, which makes the developer believe it is not supported, wrongly.

TLDR: the fix should be to fix the TS type (at [this line](https://github.com/marmelab/react-admin/blob/more-themes/packages/ra-ui-materialui/src/list/pagination/Pagination.tsx#L162)).
I don't think it's necessary to add labels to the `DefaultRowsPerPageOptions`.

PS: I also believe your translations keys are wrong, in such a use case you should use [Interpolation](https://marmelab.com/react-admin/TranslationTranslating.html#interpolation-pluralization-and-default-translation).

fzaninotto:

PR #9334, opened 2023-10-05
Fix useUnique send request even if the field value is empty

adguernier:
## Problem

While using `useUnique` in an non required input, and when sending the form with empty value within this input, `useUnique` search for data with `null` value and returns results that indicates the record is not unique and indeed, cause errors on form.

This issue happens with some DBMS and DataProvider

## Solution

Do not send request to the API if value tested by useUnique is empty.

## Todo
- [x]  use `isEmpty` function to check if the tested input is empty

### DIFF
--- a/packages/ra-core/src/form/useUnique.ts
+++ b/packages/ra-core/src/form/useUnique.ts
@@ -87,6 +87,9 @@ export const useUnique = (options?: UseUniqueOptions) => {
             );
 
             return async (value: any, allValues: any, props: InputProps) => {
+                if (!value) {
+                    return undefined;
+                }
                 try {
                     const finalFilter = set(
                         merge({}, filter),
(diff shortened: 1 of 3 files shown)


djhi (approved):

packages/ra-core/src/form/useUnique.ts
```diff
@@ -87,6 +87,9 @@ export const useUnique = (options?: UseUniqueOptions) => {
             );
 
             return async (value: any, allValues: any, props: InputProps) => {
+                if (!value) {
```
djhi, line 90: Can you extract and use 'isEmpty` from `validate.ts` instead?

djhi (changes requested):

djhi (approved):

Issue #9318, opened 2023-09-26
Changing any parameters of the ListGuesser empties the page 


djhi:
With the `ListGuesser`, when changing the sorting or the pagination parameters using the UI, we see a empty page (only the layout remains) for a time.

https://stackblitz.com/edit/github-qad8w5?file=src%2Fposts%2Findex.tsx

dinaco:
Hey @djhi could you assign this issue to me? I'd like to work on it

djhi:
Hi! We don't assign issues. Just open a PR :)

Issue #7599, opened 2022-04-29
ListGuesser only works for one resource


fzaninotto:
**What you were expecting:**
 Being able to use guessers on several resources, e.g.

```tsx
const App = () => (
    <Admin dataProvider={dataProvider}>
      <Resource name="posts" list={ListGuesser} edit={EditGuesser} />
      <Resource name="comments" list={ListGuesser} edit={EditGuesser} />
      <Resource name="tags" list={ListGuesser} edit={EditGuesser} />
      <Resource name="users" list={ListGuesser} edit={EditGuesser} />
    </Admin>
);
```

**What happened instead:**

When navigating from resource to resource using the left menu (showing the posts list, then comments list, etc), the columns of the datagrid remain the same, even though the data changes. 

See https://codesandbox.io/s/pensive-euclid-pfotfq

https://user-images.githubusercontent.com/99944/165921504-5f01d6f7-93af-401b-a217-499235440279.mov

**Other information:**

**Environment**

* React-admin version: 4.0
* Last version that did not exhibit the issue (if applicable): 3.19
* React version: 18
* Browser: Chrome

PR #9335, opened 2023-10-05
[WIP] removed query options from ListGuesser

dinaco:
relates to issue [#9318](https://github.com/marmelab/react-admin/issues/9318)


fzaninotto:
You're reverting the change introduced in https://github.com/marmelab/react-admin/pull/7605, which fixed a bug (https://github.com/marmelab/react-admin/issues/7599), so you're effectively reintroducing the bug.

dinaco:
> You're reverting the change introduced in #7605, which fixed a bug (#7599), so you're effectively reintroducing the bug.

Ok, I'll look into it again

Issue #4189, opened 2019-12-19
Calling setSort and setFilters after one another filckers and doesn't setSort


estephan:
**What you were expecting:**

I've built a button and added it as part of the `actions` props on a `List` component. This button is intended to clear the filters applied to the `List` as well as return the sort to its default value. The button calls the `setSort` and `setFilters` functions that are passed in as props to `actions`. I expected the filters to be cleared and the sort to return to its default value.

**What happened instead:**

Instead, what I'm seeing is that the filters are cleared (as expected) however the sort "flicks" to the default sort options and then flicks back to the previously selected sort.
If I call `setSort` alone or `setFilters` alone, they behave as expected and it seems that this issue only occurs when both are called after one another.

**Steps to reproduce:**

1. Go to the Posts list in the following [sandbox](https://codesandbox.io/s/elastic-raman-kvypc).
2. Click on the "Title" column to sort posts by title.
3. In the search box, type "Mai" to add a filter.
4. Then click the "Clear sort and filters" button in the top right.
5. You'll see the sort goes back to "Published At" as expected but then flicks back to "Title".

https://codesandbox.io/s/elastic-raman-kvypc

**Environment**

* React-admin version: 3.1.0
* Last version that did not exhibit the issue (if applicable): 2.x
* React version: 16.12.0
* Browser: Chrome (v. 78.0.3904.108)

fzaninotto:
You call `setSort()` with no argument, it's not something expected by react-admin (it should not be accepted at all, in fact). Pass the default sort you want as parameter, or else react uses the *last selected sort* (we persist user choices in the Redux store).

The flickr comes from the fact that we use the URL to store the sort and filters. `setSort` and `setFilter` each change the URL once, so that triggers two fetches of the dataProvider. Nothing anormal.

In your case, your button should set the query parameter directly to do all in once. That would avoid the flickr.

Anyway, this is not a bug IMO, so I'm closing the issue.

estephan:
No arguments passed to `setSort` was just an example - if you try it with `setSort("published_at")` the same flicker result happens.

Try with https://codesandbox.io/s/kind-wave-qjxb9

fzaninotto:
You're right - I believe that's an issue with a bad dependency in a `useEffect`. Repoening and classifying as a bug. Thanks for the CodeSandbox links.

ZachSelindh:
[...]

PR #9340, opened 2023-10-06
[TypeScript] add setParams in ListContext

mchaffotte:
## Problem
In some situations you might want to change multiple parameters on the ListContext (ex: sort & filter). Currently, you have to call two methods: setFilter and setSort. However, this sometimes fail to apply both results and you loose one of them ([Calling setSort and setFilters after one another filckers and doesn't setSort · Issue #4189 · marmelab/react-admin](https://github.com/marmelab/react-admin/issues/4189) ). There are workaround but they are not always available (you can’t always use the navigation API, e.g ReferenceManyField)

## Solution
Add a setParams function that accepts the getList payload pagination, sort and filter).

Make sure to update all components that implements ListContext (in enterprise too)

To be added in:

use ListController
useList
useReferenceManyFieldController
useReferenceArrayInputController
useReferenceManyInputController
useReferenceManyToManyInputController
…


fzaninotto:
Superseded by #9638

Issue #9331, opened 2023-10-04
Documentation issues with yarn link


smeng9:
Hi @djhi, I am still trying to troubleshoot the vitest issue in https://github.com/marmelab/react-admin/pull/9309 by changing exports field in package.json 

However I found part of the tutorial in https://github.com/marmelab/react-admin#testing-your-changes-in-your-app is outdated since this commit. https://github.com/marmelab/react-admin/commit/8fe89469d1ad44056215e66ff2cd84aa72b2c595

**What you were expecting:**

Yarn link should not produce any problem regardless of yarn version

**What happened instead:**

Unable to link to a local copy with error:
Unknown Syntax Error: Not enough positional arguments.

**Steps to reproduce:**

Follow the steps in Testing Your Changes in Your App

**Other information:**

We should split this part of tutorial.
If client app is using yarn classic, they should follow https://classic.yarnpkg.com/lang/en/docs/cli/link/ to use

> $ YARN_IGNORE_PATH=1 yarn link

If client app is using yarn berry, they only need to link to the workspace

`yarn link /code/path/to/react-admin/packages`

**Environment**

* React-admin version: 4.14.4
* Last version that did not exhibit the issue (if applicable): 4.0.0-alpha.1 before https://github.com/marmelab/react-admin/commit/8fe89469d1ad44056215e66ff2cd84aa72b2c595 is merged
* React version: 18.2.0
* Browser: Safari 16.5.1
* Stack trace (in case of a JS error):

djhi:
Good catch! Do you mind opening a PR?

djhi:
Fixed by #9341

PR #9341, opened 2023-10-06
update yarn linking instruction

smeng9:
@djhi, Even though it's still possible to link a yarn 3 project to a yarn 1 project by setting the `YARN_IGNORE_PATH` environment variable as mentioned in https://github.com/marmelab/react-admin/issues/9331 , I don't think it is a recommended way to do it.

Let's consider dropping the yarn v1 version support. 

Dropping yarn v1 support in the documentation is not a breaking change IMO because the yarn version bump https://github.com/marmelab/react-admin/commit/8fe89469d1ad44056215e66ff2cd84aa72b2c595 commit already happened in react-admin v4 alpha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,22 +207,24 @@ Both of these examples work without server—the API is simulated on the client-
 
 ### Testing Your Changes In Your App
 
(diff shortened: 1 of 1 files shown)


djhi (approved):
Thanks!

PR #9351, opened 2023-10-11
Fix DatagridConfigurable editor allows to drag fields outside its list

djhi:
# Problem

When configuring the `DatagridConfigurable` using its editor, one can drag and drop a field outside the list of fields (for instance below the button)

image

# Solution

Wrap the items list in a `ul` element and double check users drop the column inside it

### DIFF
--- a/packages/ra-ui-materialui/src/preferences/FieldToggle.tsx
+++ b/packages/ra-ui-materialui/src/preferences/FieldToggle.tsx
@@ -70,7 +70,6 @@ export const FieldToggle = props => {
             }
         }
 
-        console.log({ dropItem });
         if (dropItem && list === dropItem.closest('ul')) {
             onMove(selectedItem.dataset.index, dropIndex.current);
         } else {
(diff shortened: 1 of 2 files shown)


mchaffotte (approved):

slax57 (changes requested):
Seems mostly OK, but I ran across I weird issue.
If you still drop the item below the buttons, then visually the item is moved to the latest position in the list, but the column is actually not moved in the Datagrid.


https://github.com/marmelab/react-admin/assets/14542336/4a8e7f6b-7c6a-407d-9aa9-025b5cf5b968



And it you repeat that multiple times, then afterwards the component is completely broken and columns no longer match the order of the inspector.

packages/ra-ui-materialui/src/preferences/FieldToggle.tsx
```diff
@@ -48,8 +52,31 @@ export const FieldToggle = props => {
...
+                y.current >
+                selectedItem.closest('ul').getBoundingClientRect().bottom
+            ) {
+                dropItem = list.lastChild as HTMLElement;
+            } else {
+                dropItem = list.firstChild as HTMLElement;
+            }
+        }
+
+        console.log({ dropItem });
```
djhi, line 73:
```suggestion
```

slax57 (approved):
All good to me, WD!

PR #9353, opened 2023-10-12
[RFR] add disabling invalid form notification

### LINKED ISSUE #9350: Allow to disable notification that is shown on invalid form submits
When a form is invalid a notification is shown on the bottom of the screen. I want to be able to disable the notification.

There could be a prop or similar to disable the notification.

I checked the code and couldn't find an option to disable the notification. Only if I create a custom Form component, which wouldn't make much sense, since it would be basically copy paste.

**Additional context**
image

tim-hoffmann:
This should close #9350

### DIFF
--- a/packages/ra-core/src/form/Form.spec.tsx
+++ b/packages/ra-core/src/form/Form.spec.tsx
@@ -167,6 +167,33 @@ describe('Form', () => {
         });
     });
 
+    it('should not display a notification on submit when invalid', async () => {
+        const Notification = () => {
+            const { notifications } = useNotificationContext();
+            return notifications.length > 0 ? (
+                <div>{notifications[0].message}</div>
+            ) : null;
+        };
+
+        render(
+            <CoreAdminContext>
+                <>
+                    <Form onSubmit={jest.fn()} disableInvalidFormNotification>
+                        <Input source="name" validate={required()} />
+                        <button type="submit">Submit</button>
+                    </Form>
+                    <Notification />
+                </>
+            </CoreAdminContext>
+        );
+
+        fireEvent.click(screen.getByText('Submit'));
+        await waitFor(() => {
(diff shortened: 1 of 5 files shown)


fzaninotto:
Could you please add documentation and unit tests?

tim-hoffmann:
> Could you please add documentation and unit tests?

added docs and a test to verify my changes  ✌️

packages/ra-core/src/form/Form.spec.tsx
```diff
@@ -167,6 +167,33 @@ describe('Form', () => {
...
+                    <Notification />
+                </>
+            </CoreAdminContext>
+        );
+
+        fireEvent.click(screen.getByText('Submit'));
+        await waitFor(() => {
+            const notification = screen.queryByText('ra.message.invalid_form');
+            expect(notification).toBeNull();
+        });
```
slax57, line 194: This test passes even without `disableInvalidFormNotification`. 

It is not always easy to check for the absence of something with Jest/RTL. You need to find an event to await for, that proves the form submission is complete, before attempting to check if the notification is there or not. You can try providing a `jest.fn()` as `onSuccess` to see if it helps, or try rendering the record and have RTL look for the changed value in the record.

slax57 (changes requested):
Otherwise it should be good. I'll just request @fzaninotto 's review to validate the new prop's name.
In any case, thanks!

packages/ra-core/src/form/Form.spec.tsx
tim-hoffmann: Thanks for feedback! 🙏 
I updated the unit test so that I'm waiting for the validation message to be shown before checking the absence of the notification.

slax57 (approved):
LGTM, thanks!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9354, opened 2023-10-12
[Doc] Explain how `<PasswordInput>` can be used to update a password

fzaninotto:
## Problem

The doc doesn't explain how to use `<PasswordInput>` in Edit view. This may lead developers to expose passwords in clear.

## Solution

Be very clear about never exposing password, even in the Edition view.

### DIFF
--- a/docs/PasswordInput.md
+++ b/docs/PasswordInput.md
@@ -58,6 +58,45 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
 ```
 {% endraw %}
 
+## Validating Identical Passwords
+
+If you want to validate that the user has entered the same password in two different password inputs, you must use [global validation](./Validation.md#global-validation) at the form level:
+
+```jsx
+import { Create, SimpleForm, TextInput, PasswordInput } from 'react-admin';
+
+const validate = values => {
+    const errors = {} as any;
+    if (!values.name) {
+        errors.name = 'Name is required';
+    }
+    if (!values.email) {
+        errors.email = 'Email is required';
+    }
+    if (!values.password) {
+        errors.password = 'Password is required';
+    }
+    if (values.password && values.password !== values.confirm_password) {
+        errors.confirm_password = 'The two password fields must match';
+    }
+    return errors;
+}
+
+const UserCreate = () => (
+    <Create validate={validate}>
+        <SimpleForm>
+            <TextInput source="name" isRequired />
+            <TextInput source="email" isRequired />
+            <PasswordInput source="password" isRequired />
+            <PasswordInput source="confirm_password" />
+        </SimpleForm>
+    </Create>
+);
+```
+
+You can't use the `validate` prop in the Input components, as global and field-level validation are mutually exclusive. To show that an input is required, use the `isRequired` prop instead of `validate`.
+
+
 ## Usage in Edit Views
 
 You may want to allow users to *update* a password on an existing record. The usual solution to this is to include a `new_password` input in the Edition form. The API will then check if this field is present in the payload, and update the password accordingly.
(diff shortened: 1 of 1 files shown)


docs/PasswordInput.md
```diff
@@ -47,3 +57,22 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
 <PasswordInput source="password" inputProps={{ autocomplete: 'current-password' }} />
 ```
 {% endraw %}
+
```
fzaninotto, line 60: We probably need to explain how to have 2 password inputs, and how to validate that they are equal.

fzaninotto: explained.

docs/PasswordInput.md
```diff
@@ -47,3 +57,61 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
...
+## Validating Identical Passwords
+
+If you want to validate that the user has entered the same password in two different password inputs, you must use [global validation](./Validation.md#global-validation) at the form level:
```
djhi, line 63: Not really, validators have access to the other form values, so you could write an `equalToFieldValidator`:

```tsx
const equalToFieldValidator = (source: string) => (value: any, allValues: FormValues) => ...
```

djhi (changes requested):

docs/PasswordInput.md
fzaninotto: TIL!

docs/PasswordInput.md
```diff
@@ -14,11 +14,21 @@ title: "The PasswordInput Component"
 Use it like a [`<TextInput>`](./TextInput.md):
 
 ```jsx
-import { PasswordInput } from 'react-admin';
-
-<PasswordInput source="password" />
+import { Create, SimpleForm, TextInput, PasswordInput } from 'react-admin';
+
+const UserCreate = () => (
```
adguernier, line 19:
```suggestion
export const UserCreate = () => (
```

docs/PasswordInput.md
```diff
@@ -47,3 +57,61 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
...
+    if (!values.password) {
+        errors.password = 'Password is required';
+    }
+    if (values.password && values.password !== values.confirm_password) {
+        errors.confirm_password = 'The two password fields must match';
+    }
+    return errors;
+}
+
+const UserCreate = () => (
```
adguernier, line 85:
```suggestion
export const UserCreate = () => (
```

docs/PasswordInput.md
```diff
@@ -47,3 +57,61 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
...
+
+
+## Usage in Edit Views
+
+You may want to allow users to *update* a password on an existing record. The usual solution to this is to include a `new_password` input in the Edition form. The API will then check if this field is present in the payload, and update the password accordingly.
+
+```jsx
+import { Edit, SimpleForm, TextInput, PasswordInput } from 'react-admin';
+
+const UserEdit = () => (
```
adguernier, line 107:
```suggestion
export const UserEdit = () => (
```

docs/PasswordInput.md
```diff
@@ -47,3 +57,61 @@ Set the [`autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/
...
+    </Create>
+);
+```
+
+You can't use the `validate` prop in the Input components, as global and field-level validation are mutually exclusive. To show that an input is required, use the `isRequired` prop instead of `validate`.
+
+
+## Usage in Edit Views
+
+You may want to allow users to *update* a password on an existing record. The usual solution to this is to include a `new_password` input in the Edition form. The API will then check if this field is present in the payload, and update the password accordingly.
```
adguernier, line 102:
```suggestion
You may want to allow users to *update* a password on an existing record. The usual solution to this is to include a `new_password` input in the Edition form. Your API should then check if this field is present in the payload, and update the password accordingly.
```

adguernier (approved):

djhi (approved):

PR #9356, opened 2023-10-12
[Demo] Improve customer aside

fzaninotto:
## Problem

The current customer aside doesn't allow to list all orders or reviews, and the timeline is clunky

## Solution

Reorganize the information in the aside

Before:

image

After:

image


djhi:
Wouldn't it be better to fix the timeline height and make it scrollable? It looks weird to scroll the whole page just to look at the history

fzaninotto:
It may be better, but we don't have to do it in this PR.

djhi (approved):

PR #9357, opened 2023-10-12
Added type for `mutationMode` to `mutationOptions` prop for <BulkDeleteWithUndoButton>

yanchesky:
\<BulkDeleteWithUndoButton> prop `mutationOptions` does not have sufficient types.
 It lacks MutationMode. It is in fact hardcoded but it can be also overwritten by `otherMutationOptions`. 

https://github.com/marmelab/react-admin/blob/6afaa0a64b4fa9d9fc11c499395ce23a43efc42f/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx#L30-L32

https://github.com/marmelab/react-admin/blob/6afaa0a64b4fa9d9fc11c499395ce23a43efc42f/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx#L69-L74

The use case is utilization of existing bulk delete button without confirmation but with pessimistic mode.

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx
@@ -12,6 +12,7 @@ import {
     useListContext,
     RaRecord,
     DeleteManyParams,
+    MutationMode,
 } from 'ra-core';
 
 import { Button, ButtonProps } from './Button';
@@ -109,7 +110,7 @@ export interface BulkDeleteWithUndoButtonProps<
         RecordType,
         MutationOptionsError,
         DeleteManyParams<RecordType>
-    > & { meta?: any };
+    > & { meta?: any; mutationMode?: MutationMode };
 }
 
 const PREFIX = 'RaBulkDeleteWithUndoButton';


djhi (approved):
Thanks!

slax57:
I disagree with this change. As the name implies, `BulkDeleteWithUndoButton` is to be used with `'undoable'` mutation mode only. That's why you cannot override it via props.
The notification will still contain an 'UNDO' button, since you did not change this line: https://github.com/yanchesky/react-admin/blob/98a7f98291f0a07be5e75c5c629c8045930d5045/packages/ra-ui-materialui/src/button/BulkDeleteWithUndoButton.tsx#L49

To me, if you do not want any kind of confirmation (neither confirm dialog nor undoable notification) then you need to provide your own `BulkDeleteButton`.
Alternatively, if you believe this is a use case we should support, this could be debated, but we would only accept a PR targeting `next`, offering this as a new feature.

yanchesky:
This change is more a reflection of the current implementation not a new one. You can overwrite it, and it will work. You just have to dig through the implementation to see it and add `ts-ignore`. 

On the other hand the type of `mutationOptions` provides `onSuccess` prop which you can also overwrite which IMO is more dangerous because if you don't include notification with `undoable: true` the mutation will not be executed.

This component is **not documented** but available to use (which is really great!). Therefore I assume that it's for more advanced users who found it in the implementation. My aim with that PR is to give small convenience of what's possible.

slax57:
I think I get your point, and you are correct about `onSuccess`, good catch.

However I still don't agree on the how. Sure, `BulkDeleteWithUndoButton` is currently almost capable of doing it, but still, I think it's not the purpose of this specific component. And again, this is a new feature and not a simple type fix. It would require some thinking about the DX, and proper documentation and tests.

I'm still open to the idea of adding this feature though, as I think it's a good one. I simply don't agree on the fact that this is a fix, and with the proposed implementation.

PR #9359, opened 2023-10-13
Fix `useGetList` optimistic cache update leads to ui freeze when too many records are returned

slax57:
**Solution**: only pre-populate the `getOne` query cache when there is a reasonable number of records returned, say 100.

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetList.ts
+++ b/packages/ra-core/src/dataProvider/useGetList.ts
@@ -91,9 +91,9 @@ export const useGetList = <RecordType extends RaRecord = any>(
                 // optimistically populate the getOne cache
                 if (
                     value?.data &&
-                    value?.data?.length <= MAX_DATA_LENGTH_TO_CACHE
+                    value.data.length <= MAX_DATA_LENGTH_TO_CACHE
                 ) {
-                    value?.data?.forEach(record => {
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -87,12 +89,21 @@ export const useGetList = <RecordType extends RaRecord = any>(
...
                 // optimistically populate the getOne cache
-                value?.data?.forEach(record => {
-                    queryClient.setQueryData(
-                        [resource, 'getOne', { id: String(record.id), meta }],
-                        oldRecord => oldRecord ?? record
-                    );
-                });
+                if (
+                    value?.data &&
+                    value?.data?.length <= MAX_DATA_LENGTH_TO_CACHE
```
fzaninotto, line 94: the `?` here and line 96 is normally not necessary.

fzaninotto (changes requested):
👍

packages/ra-core/src/dataProvider/useGetList.ts
slax57: Fixed

mchaffotte (approved):

fzaninotto (approved):

PR #9364, opened 2023-10-16
doc: improve StackedFilters documentation

mchaffotte:
Improve StackedFilters documentation


slax57 (approved):
There were some `{% raw %}` / `{% endraw %}` missing (jekyll requires to add them around code snippets that includes `{{ ... }}` otherwise it thinks this is a template variable to replace), so I added them.
Otherwise it's good for me, thanks!

PR #9371, opened 2023-10-18
fix(ra-data-graphql-simple): Handle nested types

### LINKED ISSUE #9370: Incorrect handling of lists and non-nullability in arguments by ra-data-graphql-simple
**What you were expecting:**

I have a GraphQL endpoint, which expects a non-nullable list with non-nullable items. Using `ra-data-graphql-simple` it should build the query matching the introspection result, for the above example both list and item should be non-nullable: `[ID!]!`.

**What happened instead:**

`ra-data-graphql-simple` creates the argument as `[ID!]`, meaning only the items are non-nullable, while the list is nullable. This will throw an error as the expected types are different.

**Steps to reproduce:**

- Have a GraphQL endpoint where an argument should be of type list with both list and items non-nullable.
- Let `ra-data-graphql-simple` generate the query automatically.
- Confirm generated query is not according to the introspection result and API throws an error due to incorrect types.

**Related code:**

Since it requires a GraphQL endpoint couldn't create a reproducible sandbox. I will open a PR with updated unit tests. Running the unit tests against `master` would fail in `buildGqlQuery.test.ts`:

```ts
it('returns the arg type for NON_NULL LIST with nullable item type', () => {
    expect(
        print(
            getArgType({
                name: 'arg',
                type: {
                    kind: TypeKind.NON_NULL,
                    ofType: {
                        kind: TypeKind.LIST,
                        ofType: {
                            kind: TypeKind.SCALAR,
                            name: 'ID',
                        },
                    },
                },
            })
        )
    ).toEqual('[ID]!');
});
```

**Other information:**

[Article about nullability and lists on Apollo blog](https://www.apollographql.com/blog/graphql/basics/using-nullability-in-graphql/#nullability-and-lists). As stated lists can be nested and non-nullable can apply to both list and items independently.

**Environment**

* React-admin version: 4.15.0
* React version: 18
* Browser: Chrome

dricholm:
Fixes #9370, since I couldn't provide a sandbox environment due to GraphQL endpoint requirement. Unit tests should help illustrate the case.

Nullability and lists can be nested, which was incorrectly handled previously. A scenario:

- Have an argument with both list and item non-nullable with any scalar, e.g. `ID`
- `ra-data-graphql-simple` generates: `[ID!]` (item is non-nullable, while list is nullable), however it should be `[ID!]!`
- Above result causes an error on server due to incorrect type.

Moved logic to a new file and added additional unit tests.


packages/ra-data-graphql-simple/src/buildGqlQuery.ts
```diff
@@ -241,27 +240,5 @@ export const buildApolloArgs = (
...
-    if (required) {
-        return gqlTypes.nonNullType(
-            gqlTypes.namedType(gqlTypes.name(type.name))
-        );
-    }
-
-    return gqlTypes.namedType(gqlTypes.name(type.name));
-};
+export const getArgType = (arg: IntrospectionInputValue): TypeNode =>
+    getGqlType(arg.type);
```
mchaffotte, line 244: Why do you extract the content of the function in another one? Now getArgType seems to be an empty shell. You should move back the content of etGqlType there.

mchaffotte (changes requested):
Nice catch

packages/ra-data-graphql-simple/src/buildGqlQuery.ts
dricholm: It's because `getGqlType` is a recursive function. `getArgType` has a different parameter, `IntrospectionInputValue`, while the recursive one is called with the type, `IntrospectionType | IntrospectionListTypeRef | IntrospectionTypeRef`.

Although after checking again I see `getArgType` is only called in one place. Should I move the call inline above? https://github.com/marmelab/react-admin/blob/fccba79c118082be6d8c40ba5322ff82f02aaa56/packages/ra-data-graphql-simple/src/buildGqlQuery.ts#L235
So it would be:

```ts
gqlTypes.variableDefinition(
    gqlTypes.variable(gqlTypes.name(arg.name)),
    getGqlType(arg.type)
),
```

slax57 (changes requested):
Your fix looks great, thanks!

You simply have 2 linter warnings left in `buildGqlQuery.ts`, due to unused imports.
Can you fix them?

dricholm:
> Your fix looks great, thanks!
> 
> You simply have 2 linter warnings left in `buildGqlQuery.ts`, due to unused imports. Can you fix them?

Sure, thanks for the notice. Removed them.

slax57 (approved):
Thanks!

PR #9374, opened 2023-10-20
Fix security vulnerability in dev dependencies due to outdated storybook

fzaninotto:
babel/traverse has issued [CVE-2023-45133](https://nvd.nist.gov/vuln/detail/CVE-2023-45133), and Github has warned us about a possible vulneraility. 

We only use this module via Storybook, which is onny for development. 

We're upgrading storybook to dismiss that alert.


yarn.lock
```diff
@@ -179,6 +196,29 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/core@npm:^7.22.0, @babel/core@npm:^7.22.9":
```
slax57, line 199: There seems to be some duplicate deps (e.g. `@babel/core` here). I'll try to run `yarn dedupe` to see if it helps.

slax57 (approved):

PR #9381, opened 2023-10-23
[Doc] add illustration and stories for SelectField

erwanMarmelab:
## Problem
We actually don't show any illustration for the `<SelectField />` component and we don't have stories on it.

## Solution
Add an illustrations and stories for this component.

## Doc before
image

## Doc after
image


packages/ra-ui-materialui/src/field/SelectField.stories.tsx
```diff
@@ -0,0 +1,74 @@
+import dompurify from 'dompurify';
```
fzaninotto, line 1: linter warnings

docs/img/SelectField.png
fzaninotto, line 1: I'd use a red frame to highlight what's to look at in this screenshot, as in https://marmelab.com/react-admin/ChipField.html

Also, can you add more rows and columns to the screenshot, to make it look more real?

packages/ra-ui-materialui/src/field/SelectField.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+const countries = [{ name: 'Arabic', code: 'ar' }];
+export const OptionValue = () => (
+    <RecordContextProvider value={record}>
+        <SelectField source="language" choices={countries} optionValue="code" />
+    </RecordContextProvider>
+);
+
+const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
+const authors = [{ id: 1, first_name: 'John', last_name: 'Doe' }];
+export const Record = () => (
```
fzaninotto, line 53: The name of this story isn't representative of what it shows. How about OptionTextfunction?

packages/ra-ui-materialui/src/field/SelectField.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <RecordContextProvider value={record}>
+        <SelectField
+            source="id"
+            choices={authors}
+            optionText={optionRenderer}
+        />
+    </RecordContextProvider>
+);
+
+export const TranslateChoice = () => (
```
fzaninotto, line 63: this one is misleading: as there is no i18NProvider in the story, the choices won't be translated anyway.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/SelectField.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+export const OptionValue = () => (
+    <RecordContextProvider value={record}>
+        <SelectField source="language" choices={countries} optionValue="code" />
+    </RecordContextProvider>
+);
+
+const optionRenderer = choice => `${choice.first_name} ${choice.last_name}`;
+const authors = [{ id: 1, first_name: 'John', last_name: 'Doe' }];
+export const OptionTextFunction = () => (
+    <RecordContextProvider value={record}>
```
adguernier, line 67: Since this example has no relation with the record above, it may be confusing. I suggest replacing the `RecordContextProvider` value with:

```suggestion
    <RecordContextProvider value={{ id: 1 }}>
```

fzaninotto (approved):

PR #9382, opened 2023-10-23
[Doc] Add illustration for `<Labeled>` component

erwanMarmelab:
## Problem
We actually don't show any illustration for `<Labelled>` component.

## Solution
Add an illustration.

# Doc before
image

# Doc after
image


docs/img/Labeled.png
fzaninotto, line 1: Since you have a large width available, why not make the image a bit wider, to avoid the label of the first column from wrapping to two lines?

fzaninotto (changes requested):

slax57 (approved):

fzaninotto (approved):

PR #9391, opened 2023-10-25
Fix update button cannot be used in a list

erwanMarmelab:
## Problem 
UpdateButton cannot be used in a List because it does not allow to stop the propagation of the click event, hence it redirects to the Edit view. We should add a way to prevent this.

## Solution 
Stop the propagation and add test and story

### DIFF
--- a/packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
+++ b/packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
@@ -94,7 +94,7 @@ const PostList = () => {
     );
 };
 
-export const WithAList = () => (
+export const InsideAList = () => (
     <AdminContext dataProvider={getDataProvider()} i18nProvider={i18nProvider}>
         <AdminUI>
             {JSON.stringify(getDataProvider())}
--- a/packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
+++ b/packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
@@ -1,17 +1,18 @@
-import * as React from 'react';
+import { createMemoryHistory } from 'history';
+import { Resource, useNotify, withLifecycleCallbacks } from 'ra-core';
+import fakeRestDataProvider from 'ra-data-fakerest';
 import polyglotI18nProvider from 'ra-i18n-polyglot';
 import englishMessages from 'ra-language-english';
 import frenchMessages from 'ra-language-french';
-import { Resource, useNotify, withLifecycleCallbacks } from 'ra-core';
-import fakeRestDataProvider from 'ra-data-fakerest';
-import { createMemoryHistory } from 'history';
+import * as React from 'react';
 
-import { UpdateButton } from './UpdateButton';
 import { AdminContext } from '../AdminContext';
 import { AdminUI } from '../AdminUI';
-import { NumberField, TextField } from '../field';
 import { Show, SimpleShowLayout } from '../detail';
+import { NumberField, TextField } from '../field';
 import { TopToolbar } from '../layout';
+import { Datagrid, List } from '../list';
+import { UpdateButton } from './UpdateButton';
 
 export default { title: 'ra-ui-materialui/button/UpdateButton' };
 
@@ -44,7 +45,7 @@ const getDataProvider = () =>
                     id: 1,
                     title: 'Lorem Ipsum',
                     body: 'Lorem ipsum dolor sit amet',
-                    views: 1000,
+                    views: 500,
                 },
             ],
             authors: [],
@@ -78,6 +79,29 @@ const PostShow = () => {
         </Show>
     );
 };
+
+const PostList = () => {
+    return (
+        <List>
+            <Datagrid rowClick="show">
+                <TextField source="id" />
+                <TextField source="title" />
+                <TextField source="body" />
+                <NumberField source="views" />
+                <UpdateButton label="Reset views" data={{ views: 0 }} />
+            </Datagrid>
+        </List>
+    );
+};
+
+export const WithAList = () => (
+    <AdminContext dataProvider={getDataProvider()} i18nProvider={i18nProvider}>
+        <AdminUI>
+            {JSON.stringify(getDataProvider())}
+            <Resource name="posts" list={<PostList />} show={<PostShow />} />
+        </AdminUI>
+    </AdminContext>
+);
 
 export const Undoable = () => (
     <AdminContext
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
```diff
@@ -44,7 +45,7 @@ const getDataProvider = () =>
                     id: 1,
                     title: 'Lorem Ipsum',
                     body: 'Lorem ipsum dolor sit amet',
-                    views: 1000,
+                    views: 500,
```
slax57, line 48: why change this line?

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
```diff
@@ -79,6 +80,29 @@ const PostShow = () => {
...
+                <TextField source="title" />
+                <TextField source="body" />
+                <NumberField source="views" />
+                <UpdateButton label="Reset views" data={{ views: 0 }} />
+            </Datagrid>
+        </List>
+    );
+};
+
+export const WithAList = () => (
```
slax57, line 97:
```suggestion
export const InsideAList = () => (
```

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
```diff
@@ -79,6 +80,29 @@ const PostShow = () => {
...
+    <AdminContext dataProvider={getDataProvider()} i18nProvider={i18nProvider}>
+        <AdminUI>
+            {JSON.stringify(getDataProvider())}
```
slax57, line 100: this looks like debug code. please remove

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
```diff
@@ -95,11 +96,23 @@ describe('<UpdateWithUndoButton />', () => {
...
-                    previousData: { id: 123, title: 'lorem', views: 1000 },
+                    previousData: { id: 123, title: 'lorem', views: 500 },
                     resource: 'posts',
                 },
                 { snapshot: expect.any(Array) }
             );
         });
     });
+
+    it("should'nt open the show page caused by click propagation", async () => {
```
slax57, line 107:
```suggestion
    it("should prevent click propagation", async () => {
```

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
```diff
@@ -95,11 +96,23 @@ describe('<UpdateWithUndoButton />', () => {
...
+        render(<WithAList />);
+        const startedUrl = global.window.location.pathname;
```
slax57, line 109: I'm not 100% sure but I'd say this is bad practice to use `global.window.location.pathname` in tests. Please use text that only appear in the Show view for your tests instead.

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
```diff
@@ -95,11 +96,23 @@ describe('<UpdateWithUndoButton />', () => {
...
+        const resetButton = await screen.findByRole('button', {
+            name: 'Reset views',
+        });
+        screen.getByText('500');
+        fireEvent.click(resetButton);
+        expect(screen.queryByText('500')).not.toBeNull();
```
slax57, line 115: I would have expected this to be null, on the contrary, since, now, the counter should be 0 and no longer 500

slax57 (changes requested):

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
```diff
@@ -95,11 +96,23 @@ describe('<UpdateWithUndoButton />', () => {
...
         });
     });
+
+    it("should'nt open the show page caused by click propagation", async () => {
+        render(<WithAList />);
+        const startedUrl = global.window.location.pathname;
+        const resetButton = await screen.findByRole('button', {
+            name: 'Reset views',
+        });
+        screen.getByText('500');
```
adguernier, line 113: What this line is for?

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
erwanMarmelab: Cause it is easier to find the string '500' than find '1 000' with space on tests

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
erwanMarmelab: OH YES THX !

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
erwanMarmelab: Yes but 500 contains 0 two times. Instead of this line I could test something like this         
`expect(screen.getAllByText('0').length).equal(1);`

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
erwanMarmelab: It is here to test that the click on the button "RESET VIEWS" works. If I have 500 views before click and 0 after the button works AND if the button works, the test to prevent click propagation has a real purpose.

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
```diff
@@ -79,6 +80,28 @@ const PostShow = () => {
     );
 };
 
+const PostList = () => {
+    return (
```
fzaninotto, line 84: no hook, so make it a return-less function as the following one.

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
adguernier: I was speaking about the previous line. But it's OK. Forget that :smile: My bad

packages/ra-ui-materialui/src/button/UpdateWithUndoButton.spec.tsx
slax57: No, that's not how `getByText` works. It looks for a node in the DOM containing this exact text. Partial matches won't be returned (unless you specifically tell the RTL to do so).

You can safely use `await screen.findByText('0');` here.

slax57 (changes requested):
See https://github.com/marmelab/react-admin/pull/9391#discussion_r1373364113

fzaninotto (approved):

PR #9393, opened 2023-10-25
GraphQL Simple Data Provider DELETE_MANY and UPDATE_MANY Support

maxschridde1494:
The goal here is to add an alternative to the virtual `DELETE_MANY` and `UPDATE_MANY` actions that trigger multiple `DELETE` and `UPDATE` requests under the hood. If the GraphQL Schema supports bulk deletes / updates, the Data Provider can be initialized with the `bulkActionsEnabled` option (turned off by default) and bulk `DELETE_MANY` and `UPDATE_MANY` will light up.

### DIFF
--- a/packages/ra-data-graphql-simple/README.md
+++ b/packages/ra-data-graphql-simple/README.md
@@ -82,7 +82,12 @@ type Mutation {
     views: Int!
     user_id: ID!
   ): Post
+  updatePosts(
+    ids: [ID!]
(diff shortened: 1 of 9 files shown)


slax57:
Hi,
Thank you so much for submitting this contribution.
It's gonna take us some time to review, as it's a quite substantial PR, but we are definitely willing to have a look.
Thank you for your patience.

PR #9395, opened 2023-10-26
Fix total in `useInfiniteGetList` for empty results

### LINKED ISSUE #9388: useInfiniteGetList returns total as undefined, when real value equal to 0
I have my own dataProvider for getList and I return object like `{ data: [], total: 0 };`

**What you were expecting:**
I am expecting the total value in useListContext is equal to 0.

**What happened instead:**
When total in dataProvider is equal to 0, I receive in useListContext total value as undefined.

There is such code in useInfiniteGetList:
```
return (result.data
	? {
		  ...result,
		  data: result.data,
		  total: result.data?.pages[0]?.total || undefined,
	  }
	: result) as UseInfiniteQueryResult<
	GetInfiniteListResult<RecordType>,
	Error
> & {
	total?: number;
};
```

**Environment**

* React-admin version: 4.15.1
* React version: 18.2.0
* Browser: Chrome

KislyakovDS:
- total should return 0 if the data is empty;

Closes #9388

### DIFF
--- a/packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
+++ b/packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
@@ -515,4 +515,36 @@ describe('useInfiniteListController', () => {
             });
         });
     });
+
+    describe('checkTotalValue', () => {
+        it('should return correct total value for empty data', async () => {
+            const getList = jest.fn().mockImplementation(() =>
+                Promise.resolve({
+                    data: [],
+                    total: 0,
+                })
+            );
+            const dataProvider = testDataProvider({ getList });
+
+            render(
+                <CoreAdminContext dataProvider={dataProvider}>
+                    <InfiniteListController resource="posts">
+                        {({ total }) => (
+                            <div aria-label="Total value">{String(total)}</div>
+                        )}
+                    </InfiniteListController>
+                </CoreAdminContext>
+            );
+
+            await waitFor(() => {
+                const totalDivNode = screen.getByLabelText('Total value');
+                const totalInnerHTML = totalDivNode.innerHTML;
+                const totalValue = Number(totalInnerHTML);
+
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
```diff
@@ -515,4 +515,36 @@ describe('useInfiniteListController', () => {
             });
         });
     });
+
+    describe('checkTotalValue', () => {
```
djhi, line 519:
```suggestion
```

Not needed

packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
```diff
@@ -515,4 +515,36 @@ describe('useInfiniteListController', () => {
...
+                const totalDivNode = screen.getByLabelText('Total value');
+                const totalInnerHTML = totalDivNode.innerHTML;
+                const totalValue = Number(totalInnerHTML);
+
+                expect(totalInnerHTML).not.toEqual('undefined');
+                expect(totalValue).not.toBeNaN();
+                expect(totalValue).toEqual(0);
+            });
+        });
+    });
```
djhi, line 549:
```suggestion
```

djhi (changes requested):
Thanks!

packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
KislyakovDS: done

packages/ra-core/src/controller/list/useInfiniteListController.spec.tsx
KislyakovDS: done

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9396, opened 2023-10-26
[Doc] Update canAccess doc

adguernier:
Update the [`canAccess` action](https://marmelab.com/react-admin/canAccess.html#action) doc to match the example in the Enterprise Edition doc

### DIFF
--- a/docs/canAccess.md
+++ b/docs/canAccess.md
@@ -52,7 +52,7 @@ The `PostEditButton` component will render the `<EditButton>`.
 
 ## `action`
 
-If you want to check a specific action, for instance if the user can `delete` a `post` , you can specify this action in the dedicated prop: 
+If you want to check a specific action, for instance if the user can `delete` a `post`, you can specify this action in the dedicated prop: 
 
 ```tsx
 const permissions = [
--- a/docs/canAccess.md
+++ b/docs/canAccess.md
@@ -68,7 +68,7 @@ canAccess({ permissions, resource: 'posts', action: 'export' }); // false
 If you just want to know whether users can access any of the resources, you don't have to provide an `action`. For instance, here's how you may display different components depending on resources access rights in the dashboard:
 
 ```tsx
-import { Admin } from 'react-admin';
+import { Admin, usePermissions } from 'react-admin';
 import { canAccess } from '@react-admin/ra-rbac';
 import { dataProvider } from './dataProvider';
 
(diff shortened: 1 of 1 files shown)


erwanMarmelab (approved):

docs/canAccess.md
```diff
@@ -52,23 +52,23 @@ The `PostEditButton` component will render the `<EditButton>`.
 
 ## `action`
 
-Specify the action you want to check. 
+If you want to check a specific action, for instance if the user can `delete` a `post` , you can specify this action in the dedicated prop:
```
slax57, line 55:
```suggestion
If you want to check a specific action, for instance if the user can `delete` a `post`, you can specify this action in the dedicated prop: 
```

docs/canAccess.md
```diff
@@ -52,23 +52,23 @@ The `PostEditButton` component will render the `<EditButton>`.
...
+canAccess({ permissions, resource: 'posts', action: 'delete' }); // true
 canAccess({ permissions, resource: 'posts', action: 'export' }); // false
 ```
 
-You don't have to provide an `action` if you just want to know whether users can access the CRUD pages of a resource. This is useful to leverage `canAccess` in an `<Admin>` component children function:
+If you just want to know whether users can access any of the resources, you don't have to provide an `action`. For instance, here's how you may display different components depending on resources access rights in the dashboard:
 
 ```tsx
-import { Admin, Resource, ListGuesser, EditGuesser } from 'react-admin';
+import { Admin } from 'react-admin';
```
slax57, line 71:
```suggestion
import { Admin, usePermissions } from 'react-admin';
```

docs/canAccess.md
```diff
@@ -84,28 +84,44 @@ const authProvider = {
...
     </Admin>
 );
 ```
 
-In this example, users will see the products list and will be able to click on its category link to edit the category. However, they won't see the categories list nor the commands list.
+In this example, users will see the list of last products and the list of last categories but they won't be able to see the list of last orders.
+
 
-Note that [ra-rbac's `<Resource>` component](./AuthRBAC.md#resource) does this check automatically, so you don't actually need to use `canAccess` in this case.
+**Note**: that [ra-rbac's `<Resource>` component](./AuthRBAC.md#resource) check automatically the `list`, `show `, `create` and `edit` actions, so you don't actually need to use `canAccess` if you want to restrict a whole resource.
```
slax57, line 124:
```suggestion
**Note**: [ra-rbac's `<Resource>` component](./AuthRBAC.md#resource) automatically checks for the `list`, `show`, `create` and `edit` actions, so you don't actually need to use `canAccess` if you want to restrict a whole resource.
```

slax57 (changes requested):

fzaninotto (approved):

PR #9400, opened 2023-10-26
Fix i18next translation provider fails to use smart count in BulkActionsToolbar

erwanMarmelab:
## Problem
The bulk actions toolbar label doesn’t replace the token with the value

## Solution
Change the variable name from `count` to `smart_count` on the i18nextProbider

## Before
image

## After
image


fzaninotto (approved):

slax57 (commented):
Can you also include the fix in the story, to avoid calling `convertRaTranslationsToI18next` twice?

fzaninotto (approved):

PR #9402, opened 2023-10-26
[Doc] explain how to upload file with with FormData

adguernier:
Add documentation to explain how to send file with FormData


docs/ImageInput.md
```diff
@@ -170,3 +170,214 @@ The `<ImageInput>` component accepts the usual `className` prop. You can also ov
...
+import simpleRestProvider from 'ra-data-simple-rest';
+
+const dataProvider = withLifecycleCallbacks(simpleRestProvider('http://path.to.my.api/'), [
+  {
+    resource: "posts",
+    beforeSave: async (params: any, dataProvider: DataProvider) => {
+      const formData = new FormData();
+      formData.append("file", params.picture.rawFile);
+  
+      const imageResponse = await fetch(`http://path.to.my.api/posts`, {
```
fzaninotto, line 259: Isn't it unlikely that the URL for image upload is the same as the post update URL?

docs/ImageInput.md
```diff
@@ -170,3 +170,214 @@ The `<ImageInput>` component accepts the usual `className` prop. You can also ov
...
+    });
+
+export default myDataProvider;
+```
+
+**Tip**: Use `beforeSave` instead of `beforeUpdate` to do the same for both create and update calls.
+
+### How To Send File Using FormData
+
+In case you need to upload files to your API, as you would with an HTML form, you can use [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API that uses the same format a form would use if the encoding type were set to `multipart/form-data`.
```
fzaninotto, line 240: I was confused by this section, as I expected that you'd explain how to send both the form data and the image in one call (which is what people expect when using multipart).

fzaninotto (changes requested):

docs/ImageInput.md
adguernier: I thought the API could detect if a file is present on the request, but yes, it will be clearer if the upload URL is different.

docs/FileInput.md
```diff
@@ -237,3 +237,6 @@ const MyEdit = (props) => {
 }
 ```
 
+## Uploading Files
+
+Handling files uploads in react-admin is the dataProvider's job. Learn how to do that in the [dedicated documentation](./DataProviders.md/#handling-file-uploads).
```
fzaninotto, line 242: the link text should show where the user will land. "dedicated documentation" isn't descriptive enough. Prefer something like "the data provider documentation"

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
 };
 ```
 
-The [Querying the API](./Actions.md) documentation lists all the hooks available for querying the API, as well as the options and return values for each of them.
\ No newline at end of file
+The [Querying the API](./Actions.md) documentation lists all the hooks available for querying the API, as well as the options and return values for each of them.
+
+## Handling File Uploads
+
+When a user submits a form with a file input, the dataProvider method (`create` or `delete`) receives [a `File` object](https://developer.mozilla.org/en-US/docs/Web/API/File). You can use that `File` object to send the file the format your server expect:
```
fzaninotto, line 593:
```suggestion
When a user submits a form with a file input, the dataProvider method (`create` or `delete`) receives [a `File` object](https://developer.mozilla.org/en-US/docs/Web/API/File). You can use that `File` object to send the file in the format your server expects:
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+- you can [send files as Base64 string](#sending-files-in-base64), using the [`FileReader`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) API;
+- you can [send file using multipart/form-data](#sending-files-in-multipartform-data) that allow you to send a whole resource and their files;
```
fzaninotto, line 595:
```suggestion
- you can [send file using multipart/form-data](#sending-files-in-multipartform-data) (i.e. send the record data and their files in one query);
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+- or you might want [to send files to a third party service](#sending-files-to-a-third-party-service) such as CDN;
+- etc.
+
+### Sending Files In Base64
+
+The following `dataprovider` extends an existing `dataprovider` to convert images passed to `dataProvider.update('posts')` into Base64 strings. The example leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.update()` method for the `posts` resource only.
```
fzaninotto, line 601:
```suggestion
The following `dataProvider` extends an existing `dataProvider` to convert images passed to `dataProvider.update('posts')` into Base64 strings. The example leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.update()` method for the `posts` resource only.
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+    });
+
+export default myDataProvider;
+```
+
+**Tip**: Use `beforeSave` instead of `beforeUpdate` to do the same for both create and update calls.
+
+### Sending Files In `multipart/form-data`
+
+In case you need to upload files, as you would with an HTML form but to your API, you can use [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API that uses the same format a form would use if the encoding type were set to `multipart/form-data`.
```
fzaninotto, line 661:
```suggestion
In case you need to upload files, as you would with an HTML form but to your API, you can use the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API . It uses the same format as a form would use if the encoding type were set to `multipart/form-data`.
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+The following `dataProvider` extends an existing one and tweaks its `create` and `update` methods for the `posts` resource only. It is the role of your API to negotiate the request and process the image. 
+
+Its job is to send the `post` data with the attached file and works this way:
```
fzaninotto, line 665:
```suggestion
The data provider sends the `post` data with the attached file and works this way:
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+- it checks if the treated resource is `posts`;
```
fzaninotto, line 666:
```suggestion
- it checks if the resource is `posts`;
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+- if not, it simply uses the `baseDataProvider` base methods.
+
+Lets have a look:
+
+```ts
+import simpleRestDataProvider from "ra-data-simple-rest";
+import {
+  type CreateParams,
+  type UpdateParams,
+  type DataProvider,
```
fzaninotto, line 677: We don't use that syntax in any other place of the doc, so please remove the `type`

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+    }
+
+    return baseDataProvider.update(resource, params);
+  },
+};
+```
+
+### Sending Files To A Third-Party Service
+
+If the third-party service you use allow this, you can upload files via [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) like the previous way.
```
fzaninotto, line 738: This is confusing. Why mention formData here? It's not the focus of that section. 

I'd write the following introduction instead:

> A common way to handle file uploads in single-page-apps is to first upload the file to a CDN like [Cloudinary](https://cloudinary.com/) or [CloudImage](https://www.cloudimage.io/en/home), then to use the file URL in the CDN in the record creation / modification payload.

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+Lets see an example with [https://cloudinary.com/](Cloudinary) service, by adapting the `dataprovider` according to [their "Authenticated requests" example](hhttps://cloudinary.com/documentation/upload_images#authenticated_requests).
```
fzaninotto, line 740:
```suggestion
Lets see an example with the [https://cloudinary.com/](Cloudinary) service, by adapting the `dataProvider` according to [their "Authenticated requests" example](hhttps://cloudinary.com/documentation/upload_images#authenticated_requests).
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+To do that, you need an API that serves a [`signature`](https://cloudinary.com/documentation/upload_images#generating_authentication_signatures) the format that Cloudinary expect. To make it easier, you can install the [`cloudinary package`](https://cloudinary.com/documentation/node_integration#installation_and_setup).
+
+The following code example live in a Remix application. It generates and serves the signature needed to send files to Cloudinary:
```
fzaninotto, line 744:
```suggestion
The followin example generates and serves the signature needed to send files to Cloudinary. It uses Remix:
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+```ts
+// handles the "get-cloudinary-signature" route and should be secured
+import { type LoaderFunctionArgs, json } from "@remix-run/node";
```
fzaninotto, line 748: loose the `type`

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+  return json({
+    timestamp,
+    signature,
+    cloud_name: process.env.CLOUDINARY_CLOUD_NAME,
+    api_key: process.env.CLOUDINARY_API_KEY,
+  });
+};
+```
+
+The following `dataprovider` extends an existing one and leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.create()` and `dataProvider.update()` methods for the `posts` resource only with the [`beforeSave`](./withLifecycleCallbacks.md#beforesave) methods.
```
fzaninotto, line 777:
```suggestion
The following `dataProvider` extends an existing one and leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.create()` and `dataProvider.update()` methods for the `posts` resource only with the [`beforeSave`](./withLifecycleCallbacks.md#beforesave) methods. 
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+    signature,
+    cloud_name: process.env.CLOUDINARY_CLOUD_NAME,
+    api_key: process.env.CLOUDINARY_API_KEY,
+  });
+};
+```
+
+The following `dataprovider` extends an existing one and leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.create()` and `dataProvider.update()` methods for the `posts` resource only with the [`beforeSave`](./withLifecycleCallbacks.md#beforesave) methods. 
+
+Its only job is to send files to Cloudinary it works as follow:
```
fzaninotto, line 779:
```suggestion
It works as follow:
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+- it grabs the Cloudinary signature to allow the autentication in the request;
+- it creates a new `FormData` object with the file received from the form;
+- it sends this file to the Cloudinary API; 
+- it hydrates the `params.picture` object with the data sent by Cloudinary;
+- finally it returns the `params` object to allow the parent `dataProvider` to carry out its processes.
```
fzaninotto, line 784:
```suggestion
- it returns the `params` object to allow the parent `dataProvider` to carry out its processes.
```

fzaninotto (changes requested):

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+                ...base64Pictures.map((dataUrl, index) => ({
+                    src: dataUrl,
+                    title: newPictures[index].name,
+                })),
+                ...formerPictures,
+            ];
+            return dataProvider.update(
+                resource,
+                { data: { ...params.data, pictures } }
+            );
```
slax57, line 636: This example is missing passing other required params, such as `previousData` and `id`

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+    });
+
+export default myDataProvider;
+```
+
+**Tip**: Use `beforeSave` instead of `beforeUpdate` to do the same for both create and update calls.
+
+### Sending Files In `multipart/form-data`
+
+In case you need to upload files, as you would with an HTML form but to your API, you can use the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API . It uses the same format as a form would use if the encoding type were set to `multipart/form-data`.
```
slax57, line 661:
```suggestion
In case you need to upload files, as you would with an HTML form but to your API, you can use the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API. It uses the same format as a form would use if the encoding type was set to `multipart/form-data`.
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+The following `dataProvider` extends an existing one and tweaks its `create` and `update` methods for the `posts` resource only. It is the role of your API to negotiate the request and process the image.
```
slax57, line 663:
```suggestion
The following `dataProvider` extends an existing one and tweaks its `create` and `update` methods for the `posts` resource only. It is the responsibility of your API to negotiate the request and process the image. 
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+### Sending Files In `multipart/form-data`
+
+In case you need to upload files, as you would with an HTML form but to your API, you can use the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API . It uses the same format as a form would use if the encoding type were set to `multipart/form-data`.
+
+The following `dataProvider` extends an existing one and tweaks its `create` and `update` methods for the `posts` resource only. It is the role of your API to negotiate the request and process the image. 
+
+The data provider sends the `post` data with the attached file and works this way:
+- it checks if the resource is `posts`;
+- if so, it creates a new `FormData` object with the file received from the form and it sends this `FormData` to the API with the react-admin `fetchUtils.fetchJson()` function (a `fetch().then(r => r.json())` shortcut); 
+- if not, it simply uses the `baseDataProvider` base methods.
```
slax57, line 668: These explanations are a bit confusing to me because:

- A lot of it is dedicated to explaining when to trigger the FormData implementation, and what to do when it is not triggered ; but the interesting part is actually the rest, not this
- you can probably skip all these details, and focus on the interesting part, by rewriting the example below to use `withLifecycleCallbacks`, and simply linking to its documentation for people not yet knowing how it works
- it is not clear enough (IMHO) that the whole form values will be sent inside the `FormData` object, not only the image file
- The use of `fetchUtils` is an implementation detail, you could as well do it with a simple `fetch`. I'd either not mention it at all, or simply mention it and provide a link to its doc, without explaining further what it does, as it adds unnecessary complexity IMHO

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+
+Lets have a look:
```
slax57, line 670:
```suggestion
Let's have a look:
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+  };
+};
+
+const createPostFormData = (
+  params: CreateParams<PostParams> | UpdateParams<PostParams>
+) => {
+  const formData = new FormData();
+  formData.append("file", params.data.picture?.rawFile ?? "");
+  formData.append("title", params.data.title ?? "");
+  formData.append("content", params.data.content ?? "");
```
slax57, line 701: Not sure it's the most relevant implementation. Id' rather do stg like this

```suggestion
  params.data.picture?.rawFile != null && formData.append("file", params.data.picture.rawFile);
  params.data.title != null && formData.append("title", params.data.title);
  params.data.content != null && ("content", params.data.content);
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+  return formData;
+};
+
+export const dataProvider: DataProvider = {
+  ...baseDataProvider,
+  create: (resource, params) => {
+    if (resource !== "posts") {
```
slax57, line 708:
```suggestion
    if (resource === "posts") {
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+          method: "POST",
+          body: formData,
+        })
+        .then(({ json }) => ({ data: json }));
+    }
+
+    return baseDataProvider.create(resource, params);
+  },
+  update: (resource, params) => {
+    if (resource !== "posts") {
```
slax57, line 721:
```suggestion
    if (resource === "posts") {
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+};
+```
+
+### Sending Files To A Third-Party Service
+
+A common way to handle file uploads in single-page-apps is to first upload the file to a CDN such as [Cloudinary](https://cloudinary.com/) or [CloudImage](https://www.cloudimage.io/en/home), then use the the file URL generated by the CDN in the record creation / modification payload.
+
+Lets see an example with the [https://cloudinary.com/](Cloudinary) service, by adapting the `dataProvider` according to [their "Authenticated requests" example](hhtps://cloudinary.com/documentation/upload_images#authenticated_requests).
+
+To do that, you need an API that serves a [`signature`](https://cloudinary.com/documentation/upload_images#generating_authentication_signatures) the format that Cloudinary expect. To make it easier, you can install the [`cloudinary package`](https://cloudinary.com/documentation/node_integration#installation_and_setup).
```
slax57, line 742:
```suggestion
To do that, you need an API that serves a [`signature`](https://cloudinary.com/documentation/upload_images#generating_authentication_signatures) in the format that Cloudinary expect. To make it easier, you can install the [`cloudinary` package](https://cloudinary.com/documentation/node_integration#installation_and_setup).
```

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+};
+```
+
+The following `dataProvider` extends an existing one and leverages [`withLifecycleCallbacks`](#adding-lifecycle-callbacks) to modify the `dataProvider.create()` and `dataProvider.update()` methods for the `posts` resource only with the [`beforeSave`](./withLifecycleCallbacks.md#beforesave) methods. 
+
+It works as follow:
+- it grabs the Cloudinary signature to allow the autentication in the request;
+- it creates a new `FormData` object with the file received from the form;
+- it sends this file to the Cloudinary API; 
+- it hydrates the `params.picture` object with the data sent by Cloudinary;
```
slax57, line 783: I'm not sure you can really use the term hydration here, as it holds a very [specific meaning](https://en.wikipedia.org/wiki/Hydration_(web_development)) in the web world. I'm worried it may be misleading here. Is there a reason I'm missing why you chose to use this word?

docs/DataProviders.md
```diff
@@ -650,4 +586,263 @@ const UserProfile = ({ userId }) => {
...
+          method: "POST",
+          body: formData,
+        });
+
+        const image: CloudinaryFile = await imageResponse.json();
+
+        params.picture = {
+          src: image.secure_url,
+          title: image.asset_id,
+        };
```
slax57, line 839: Avoid mutating the function parameter directly. Prefer creating a new object instead.

slax57 (changes requested):

docs/DataProviders.md
adguernier: To me, hydrating an object means taking an empty object, and filling it with data from a database, network, file system, or whatever. As `params.picture` was not empty, 'hydrate' might not fit in the end.

"Populates" could be OK for you?

docs/DataProviders.md
```diff
@@ -650,4 +586,268 @@ const UserProfile = ({ userId }) => {
...
+  update: (resource, params) => {
+    if (resource === "posts") {
+      const formData = createPostFormData(params);
+      return fetchUtils
+        .fetchJson(`${endpoint}/${resource}`, {
+          method: "PUT",
+          body: formData,
+        })
+        .then(({ json }) => ({ data: json }));
+    }
```
slax57, line 733: This example seems somewhat incomplete, as you never actually send the record id for instance. Don't know if that's a huge deal though, maybe we can still keep it that way since the 'essence' of the example (the `FormData`) is there.

docs/DataProviders.md
```diff
@@ -650,4 +586,268 @@ const UserProfile = ({ userId }) => {
...
+                    src: dataUrl,
+                    title: newPictures[index].name,
+                })),
+                ...formerPictures,
+            ];
+
+            return {
+                ...params,
+                pictures,
+            };
```
slax57, line 638: This kind of return value is valid for a `beforeSave`, but not a `beforeUpdate`.

```suggestion
            return {
                ...params,
                data: {
                    ...params.data,
                    pictures,
                }
            };
```

slax57 (changes requested):

slax57 (approved):

PR #9392, opened 2023-10-25
GraphQL Simple Data Provider Sparse Field Support

maxschridde1494:
### Sparse Field Support for Queries and Mutations

By default, most data provider requests return 

1. All of the resource's top level fields as defined in the GraphQL schema
2. Association objects containing only the association's ID. 

This feature enables sparse field support for all data provider requests. Sparse fields are passed through via the existing [meta param](https://marmelab.com/react-admin/Actions.html#meta-parameter). This can increase efficiency, optimize client performance, improve security and reduce over-fetching. Also, it allows for the request of nested association fields beyond just their ID. It is available for all dataprovider actions.

An example `getOne` request would look like:

```js
dataProvider.getOne(
    'posts',
    { 
        id, 
        meta: { 
            sparseFields: [
                'id', 
                'title', 
                { 
                    comments: [
                        'description', 
                        { 
                            author : [
                                'name', 
                                'email'
                            ]
                        }
                    ]
                }
            ]
        }
    },
);
```

PR #9403, opened 2023-10-26
GraphQL Simple Data Provider Extensions Support

maxschridde1494:
### Data Provider Extensions

A GraphQL backend may support functionality that extends beyond the default Data Provider methods. One such example of this would be implementing GraphQL subscriptions and integrating [ra-realtime](https://marmelab.com/ra-enterprise/modules/ra-realtime) to power realtime updates in your React-Admin application. The extensions pattern allows you to easily expand the Data Provider methods to power this additional functionality. A Data Provider Extension is defined by the following type:

```js
type DataProviderExtension = {
  methodFactory: (
    dataProvider: DataProvider,
    ...args: any[]
    ) => { [k: string]: DataProviderMethod };
  factoryArgs?: any[];
  introspectionOperationNames?: IntrospectionOptions['operationNames'];
}
```

The `methodFactory` is a required function attribute that generates the additional Data Provider methods. It always receives the dataProvider as it's first argument. Arguments defined in the factoryArgs optional attribute will also be passed into `methodFactory`. `introspectionOperationNames` is an optional object attribute that allows you to inform React-Admin hooks and UI components of how these methods map to the GraphQL schema.

## Realtime Extension

I've added an example Realtime Data Provider Extension that can be opted into. If your app uses [ra-realtime](https://marmelab.com/ra-enterprise/modules/ra-realtime), you can drop in the Realtime Extension and light up realtime events. Here is an example integration:

```js
// in App.js
import React from 'react';
import { Component } from 'react';
import buildGraphQLProvider, { defaultOptions, DataProviderExtensions } from 'ra-data-graphql-simple';
import { Admin, Resource } from 'react-admin';

import { PostCreate, PostEdit, PostList } from './posts';

const dPOptions = {
    clientOptions: { uri: 'http://localhost:4000' },
    extensions: [DataProviderExtensions.Realtime]
}

const App = () => {

    const [dataProvider, setDataProvider] = React.useState(null);
    React.useEffect(() => {
        buildGraphQLProvider(dPOptions)
            .then(graphQlDataProvider => setDataProvider(() => graphQlDataProvider));
    }, []);

    if (!dataProvider) {
        return <div>Loading < /div>;
    }

    return (
        <Admin dataProvider= { dataProvider } >
            <Resource name="Post" list = { PostList } edit = { PostEdit } create = { PostCreate } />
        </Admin>
    );
}

export default App;
```


slax57:
Hi,
Thank you so much for submitting this contribution.
It's gonna take us some time to review, as it's a quite substantial PR, but we are definitely willing to have a look.
Thank you for your patience.

packages/ra-data-graphql/src/index.ts
```diff
@@ -153,6 +153,8 @@ export default async (options: Options): Promise<DataProvider> => {
             if (typeof name === 'symbol' || name === 'then') {
                 return;
             }
+            if (name == 'client') return client; // make client accessible to the proxy
```
djhi, line 156:
```suggestion
            if (name === 'client') return client; // make client accessible to the proxy
```

djhi (changes requested):
This looks really cool. I wonder whether the extension code should be in `ra-data-graphql` directly

maxschridde1494:
@djhi That's a great point. I think moving it down a level with be a good idea. That was definitely on my wishlist when integrating with `ra-data-graphql-simple` dp--a more accessible and customizable interface with `ra-data-graphql`. Plus, the extensions pattern would then be accessible to other `ra-data-graphql` wrappers. Should I close this PR and re-open against `ra-data-graphql`?

djhi (commented):
Can you please explain what is currently preventing you from implementing such custom methods? Is it because we use a Proxy? Because we don't allow you to specify the DataProvider type? Something else?

maxschridde1494:
All of our `ra-data-graphql-simple` enhancements are feature driven. Some of these features include 
- single request bulk actions
- live dashboard updates
- default client accessibility for non-REST requests
- default client accessibility to open up more of the apollo-client api (currently restricted to query & mutation).

As mentioned in https://github.com/marmelab/react-admin/pull/9392#issuecomment-1832454182, `ra-data-graphql-simple` was a great starting point for us, but did not handle some of these features out of the box. Initially, we forked and then extended. After we saw how little we needed to change, and thinking about how common these use cases are, it seemed sensible to come up with a pattern that would allow for easy extensibility without the requirement of building a new provider.

maxschridde1494:
@djhi Sorry I missed your change requests. Just got those pushed up. Do you have any final thoughts on this proposal?

djhi:
After careful consideration, we believe this abstraction layer is unnecessary. You can already add those methods by extending the resulting dataProvider:

```ts
const dataProvider = await buildGraphQLProvider({ clientOptions: { uri: 'http://localhost:4000' } });
return {
    ...dataProvider,
    subscribe: async (topic: string, subscriptionCallback: any) => {
        // The code you added for it
    },
    // etc
};
```

Besides, should you need to add custom queries generated from the schema, you would need to provide your own `buildQuery` implementation anyway. The issue is that you don't have access to the `introspectionResults` nor to [the new internal `callApollo`](https://github.com/marmelab/react-admin/pull/9487/files#diff-f31485136671519259100483dab8bf875fc60818d010451f00aec2d0e7c4f5cdR156) at this point:

```ts
import { DataProvider } from 'ra-core';
import { BuildQueryFactory } from 'ra-data-graphql';
import buildDataProvider, { buildQuery } from 'ra-data-graphql-simple';

interface CustomDataProvider extends DataProvider {
    subscribe: (resource: string, params: any) => void;
}

export const buildCustomDataProvider: Promise<CustomDataProvider> = async (
    options: Parameters<typeof buildDataProvider>[0]
) => {
    const dataProvider = await buildDataProvider({
        ...options,
        buildQuery: buildCustomQuery,
    });

    return {
        ...dataProvider,
        subscribe: (resource, params) => {
            // takes introspectionResults but we don't have access to it here
            buildCustomQuery()
        },
    };
};

const buildCustomQuery: BuildQueryFactory = introspectionResults => (
    name: string,
    resource: string,
    params: any
) => {
    if (name === 'subscribe') {
        // Build the subscription query and return it
        return null;
    }

    return buildQuery(introspectionResults)(name, resource, params);
};
```

I believe the solution is to add the `callApollo` method to the returned `dataProvider`, maybe called differently to emphasize it's internal? Besides, should we expose it, we should probably avoid including `Apollo` in its name as it's an implementation detail.

```ts
import { DataProvider } from 'ra-core';
import { BuildQueryFactory } from 'ra-data-graphql';
import buildDataProvider, { buildQuery } from 'ra-data-graphql-simple';

interface CustomDataProvider extends DataProvider {
    subscribe: (resource: string, params: any) => void;
}

export const buildCustomDataProvider: Promise<CustomDataProvider> = async (
    options: Parameters<typeof buildDataProvider>[0]
) => {
    const dataProvider = await buildDataProvider({
        ...options,
        buildQuery: buildCustomQuery,
    });

    return {
        ...dataProvider,
        subscribe: (resource, params) => {
            return dataProvider.callApollo('subscribe', resource, params);
        },
    };
};

const buildCustomQuery: BuildQueryFactory = introspectionResults => (
    name: string,
    resource: string,
    params: any
) => {
    if (name === 'subscribe') {
        // Build the subscription query and return it
        return null;
    }

    return buildQuery(introspectionResults)(name, resource, params);
};
```

PR #9407, opened 2023-10-27
Fix `<CreateButton>` keeps state between pages

### LINKED ISSUE #9394: Using state on CreateButton will polute/break other Create components
**What you were expecting:**
according to https://marmelab.com/react-admin/Create.html#prefilling-the-form if in any context (for example show page of an item) i render a
```
<CreateButton
            resource="comments"
            state={{ record: { post_id: record.id } }}
 />
```
This will only apply to the create form state if and only if i click this particular button.

**What happened instead:**
If i just visit the `show` page where this button is rendered, then go back to the main list of posts and click the add button (so the create form/page unrelated to comments), when clicking save, the `data` sent to the dataprovider (and ultimately to the api ) will contain the `post_id` field (invalid data which certain backends complain about with 400 bad request)

**sample code**
This code is not really tested (encountered the error in my own setup with custom components) but i believe it will trigger the error.

If you just click the standard "Create" button (not the custom one), The submitted form will contain `post_id` filed

```
import { List, Datagrid, TextField, DateField, BooleanField } from 'react-admin';
import { Create, SimpleForm, TextInput, DateInput, required, CreateButton, TopToolbar } from 'react-admin';
import RichTextInput from 'ra-input-rich-text';

export const PostCreate = () => (
    <Create>
        <SimpleForm>
            <TextInput source="title" validate={[required()]} fullWidth />
            <TextInput source="teaser" multiline={true} label="Short description" />
            <RichTextInput source="body" />
            <DateInput label="Publication date" source="published_at" defaultValue={new Date()} />
        </SimpleForm>
    </Create>
);

export const PostList = () => (
    <List
        actions= {
            < TopToolbar >
                <CreateButton />
                <CreateButton label="Create Comment" resource="comments" state={{ record: { post_id: 123 } }} />
            < /TopToolbar >}
    >
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <TextField source="category" />
        </Datagrid>
    </List>
);

// in src/App.jsx
import { Admin, Resource } from 'react-admin';

import { PostList } from './posts';

const App = () => (
    <Admin dataProvider={jsonServerProvider('https://jsonplaceholder.typicode.com')}>
    </Admin>
);

export default App;
```

In essence, render a `CreateButton` with the state parameter and record set to something then go to any other create form (even unrelated) and the data submitted should contain the state from the custom create button.
If it is not immediately obvious and easy to reproduce, let me know and i'll look into creating a example

fzaninotto:
Closes #9394

### DIFF
--- a/packages/ra-ui-materialui/src/button/CreateButton.tsx
+++ b/packages/ra-ui-materialui/src/button/CreateButton.tsx
@@ -43,7 +43,11 @@ const CreateButton = (props: CreateButtonProps) => {
     const isSmall = useMediaQuery((theme: Theme) =>
         theme.breakpoints.down('md')
     );
-    const state = merge(scrollStates.get(String(scrollToTop)), initialState);
+    const state = merge(
+        {},
+        scrollStates.get(String(scrollToTop)),
+        initialState
+    );
     // Duplicated behaviour of Button component (legacy use) which will be removed in v5.
     const linkParams = getLinkParams(locationDescriptor);
 
@@ -135,7 +139,8 @@ export default React.memo(CreateButton, (prevProps, nextProps) => {
         prevProps.label === nextProps.label &&
         prevProps.translate === nextProps.translate &&
         prevProps.disabled === nextProps.disabled &&
-        isEqual(prevProps.to, nextProps.to)
+        isEqual(prevProps.to, nextProps.to) &&
+        isEqual(prevProps.state, nextProps.state)
     );
 });
 


packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -43,7 +43,11 @@ const CreateButton = (props: CreateButtonProps) => {
     const isSmall = useMediaQuery((theme: Theme) =>
         theme.breakpoints.down('md')
     );
-    const state = merge(scrollStates.get(String(scrollToTop)), initialState);
+    const state = merge(
+        {},
```
fzaninotto, line 47: rookie mistake

PR #9410, opened 2023-10-30
[Doc] add scneencasts for StackedFiltersForm and update code sample

adguernier:
## Problem

The StackedFiltersForm includes a screenshot but it was missing.

## Solution

Add a screencast to demonstrate StackedFiltersForm and update code sample based on this screencast


docs/StackedFilters.md
```diff
@@ -346,37 +351,39 @@ import {
 } from '@mui/material';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 
-const postListFilters: FiltersConfig = {
+const PostListFilters: FiltersConfig = {
```
slax57, line 354: You should keep the `PostListFilters` starting with a lowercase, as it is a configuration object and not a React Component

```suggestion
const postListFilters: FiltersConfig = {
```

slax57 (changes requested):
Other than that, LGTM!

You will probably need to port this change back to the EE doc as well :grin:

slax57 (approved):

docs/StackedFilters.md
```diff
@@ -347,36 +352,38 @@ import {
...
+                            {Object.keys(filterValues).length} filter(s) applied
+                        </>
+                    ) : (
+                        'Filters'
+                    )}
+                </Typography>
             </AccordionSummary>
             <AccordionDetails id="filters-content">
-                <StackedFiltersForm config={postListFilters} />
+                <StackedFiltersForm config={PostListFilters} />
```
fzaninotto, line 382: no, this should be a lowercase

docs/StackedFilters.md
```diff
@@ -347,36 +352,38 @@ import {
...
                 expandIcon={<ExpandMoreIcon />}
                 aria-controls="filters-content"
                 id="filters-header"
             >
-                <Typography>Filters</Typography>
+                <Typography>
+                    {Object.keys(filterValues).length ? (
+                        <>
+                            {Object.keys(filterValues).length} filter(s) applied
+                        </>
```
fzaninotto, line 375: use a template instead to make it fit in one line

fzaninotto (changes requested):

docs/StackedFilters.md
adguernier: This snippet comes from Storybook, I did a naive copy/paste 😅

fzaninotto (approved):

Issue #9408, opened 2023-10-27
Chip remove button of AutocompleteInput filter are shown on top of bulk actions toolbar


lvernaillen:
**What you were expecting:**
When the bulk actions toolbar appears, the underlying filter components are not visible through that toolbar.

**What happened instead:**
When the bulk actions toolbar appears, the remove button of the chips of an `AutocompleteInput` filter are shown on top of the bulk actions toolbar.

![z-index](https://github.com/marmelab/react-admin/assets/4180374/cef44c63-40f2-4765-9eb9-6e3548f8c94d)

**Steps to reproduce:**

- Add an `AutocompleteInput` filter to the PostList.tsx of the simple demo application.

```javascript
<AutocompleteArrayInput 
  label="Post" 
  source="id" 
  choices={[{ id: '1', name: 'post1' }, { id: '2', name: 'post2' } ]} 
  alwaysOn
  multiple
/>,
```
- Select an item of that `AutocompleteInput`
- Select some rows so that the bulk actions toolbar appears on top of the datagrid
- Notice that the remove icon 'x' is shown over the toolbar instead of being hidden behind it.

I already updated react-admin's simple demo app to show this [here](https://stackblitz.com/edit/github-dwpb7y-hs2hq6?file=src%2Fposts%2FPostList.tsx)

**Related code:**

It is probably related to the fact that the  '.MuiSvgIcon-root' has a z-index of 100.
It is added in this [commit](https://github.com/marmelab/react-admin/commit/7de027dce7c3dff1fa11ca1210dfc7f7083b8858#diff-6b9bae6c911006c471eb260b0ed9ef98559e579e980ea14cdf91d864072e6303R416-R421) and has a FIXME comment.

```
  '.MuiSvgIcon-root': {
      // FIXME: Workaround to allow choices deletion
      // Maybe related to storybook and mui using different versions of emotion
      zIndex: 100,
  },
```

**Other information:**

**Environment**

* React-admin version: verified the issue in latest master 4.15.2 and in 4.8.4
* Browser: Chrome Version 118.0.5993.117 (Official Build) (64-bit)

fzaninotto:
reproduced, thanks!

PR #9413, opened 2023-10-31
Fix AutoCompletInput Clear button is over BulkActionsToolbar

erwanMarmelab:
## Problem
When the bulk actions toolbar appears, the remove button of the chips of an AutocompleteInput filter are shown on top of the bulk actions toolbar.
image

## Solution
Fix index of this button

## Issue
#9408


slax57 (changes requested):
Otherwise the fix looks good to me (don't know why we needed this z-index in the first place :shrug: )

slax57 (approved):

fzaninotto (approved):

Issue #9418, opened 2023-11-03
Expose onClick on LoadingIndicator


david-bezero:
**Is your feature request related to a problem? Please describe.**
The suggested way to trigger cache invalidation when the user presses the built-in "refresh" icon in the `AppBar` is to replace the button, as there is no way to know that a request came from the refresh button from the data-provider layer. The default component in the `AppBar` is the `LoadingIndicator`, which provides the nice behaviour of showing a loading state, but otherwise just wraps `RefreshIconButton`.

`RefreshIconButton` has an `onClick` property, which can be used to trigger cache invalidation. But this property is not exposed when using `LoadingIndicator`.

**Describe the solution you'd like**
Add `onClick` to `LoadingIndicator` and forward this property to `RefreshIconButton`.

The usage would be:

```jsx
function doMyCacheInvalidation() {
  // store some global state which can be picked up by the data-provider to force a cache invalidation
}

const MyAppBar = () => {
  const toolbar = (
    <>
      <LoadingIndicator onClick={doMyCacheInvalidation} />
    </>
  );
  return <AppBar toolbar={toolbar} />;
};
```

**Describe alternatives you've considered**
Automatically passing the "refresh" information to the data provider as a `meta` property when the button is clicked would also work, but it seems that has been suggested before and rejected.

Another option is to forego the `LoadingIndicator` and use `RefreshIconButton` directly, but this loses the built-in loading state. It can be added back in manually, but then a lot of code needs to be copied over from react-admin into the client app to get consistent styling.

fzaninotto:
I fail to understand your use case. Can you please rephrase your pain point? What data are you trying to refresh?

david-bezero:
I basically have the same use-case as this person: https://github.com/marmelab/react-admin/issues/3617

I'm implementing the approach you suggested here: https://github.com/marmelab/react-admin/issues/3617#issuecomment-525812935

> In react-admin, the path between the click on Refresh and the actual queries in the data provider takes many, many steps, and passing parameters through these steps isn't possible. So I suggest you solve this in userland, by implementing your own RefreshButton that stores a boolean in sessionStorage to force refresh, and the dataProvider can read this boolean before making requests.

So the (working) code I have right now looks like this:

```tsx
import { FC } from 'react';
import { AppBar, RefreshIconButton, useLoading } from 'react-admin';
import CircularProgress from '@mui/material/CircularProgress';
import { useTheme } from '@mui/material/styles';

[...]

Issue #3617, opened 2019-08-28
Pass through `refresh` property to data provider?


yelworc:
**Is your feature request related to a problem? Please describe.**
When working with API resources in react-admin (e.g. displaying objects in a `List` view), changes made through another client are not visible. Clicking the "refresh" icon in the title bar does not help because of HTTP caching. The only way to see recent changes (beyond waiting for the cached responses to expire) is clearing the browser cache.

**Describe the solution you'd like**
I noticed there is already a `refresh` property in some CRUD actions (e.g. `crudGetOne`), so I was wondering if it would make sense to add that to _all_ those actions, and pass it through to the data provider (e.g. in `params`). The provider could then set HTTP headers to bypass the cache (`Cache-Control: must-revalidate`, I suppose).

**Describe alternatives you've considered**
Haven't really thought about alternative solutions yet, but totally open to suggestions.

If the above idea sounds reasonable, and would be a suitable first contribution (not touching too many parts of the codebase, etc), I'd be up for working on a PR.

fzaninotto:
The `refresh` property in actions tells react-admin to refresh the view, not to pass a refresh to the dataProvider. It's different semantics, and besides, it's been removed in 3.0.

I don't see how react-admin could pass that information to the dataProvider. React-admin can't guess your caching policy. In my opinion, it's your job to write a dataProvider that sends a `must-revalidate` after a `PUT`, a `POST` or a `DELETE` on a resource.  

So I'm closing it as it can be solved in userland (in your `dataProvider`).

yelworc:
Hmm… your approach (revalidate after previous non-idempotent requests) only applies to the scenario where something was changed in "this" instance of react-admin. The more general case, where data has been modified through _any other_ means, can't be solved that way. 

You're totally right that react-admin can't guess caching policies, but when the _user_ explicitly expresses they want to see _up-to-date_ data (by clicking the refresh icon), I think it would make sense to pass this intent to the data provider somehow.

I agree that the `refresh` prop in CRUD actions should not be used for this purpose, as it has a different meaning.

fzaninotto:
In react-admin, the path between the click on Refresh and the actual queries in the data provider takes many, many steps, and passing parameters through these steps isn't possible. So I suggets you solve this in userland, by implementing your own RefreshButton that stores a boolean in sessionStorage to force refresh, and the dataProvider can read this boolean before making requests.

yelworc:
Alright – thanks for taking the time to explain!

PR #9420, opened 2023-11-03
Add `onClick` prop to `<LoadingIndicator>`, allowing to trigger actions (like a refresh) on click

david-bezero:
As requested in #9418

This makes it possible to invoke a callback when a user clicks the `LoadingIndicator` to trigger a refresh, primarily to address this use-case: https://github.com/marmelab/react-admin/issues/3617#issuecomment-525812935

I didn't find any existing unit tests for this part of the code, but I'm happy to add a unit test if somebody can point me to where it should go.

### DIFF
--- a/packages/ra-ui-materialui/src/layout/LoadingIndicator.tsx
+++ b/packages/ra-ui-materialui/src/layout/LoadingIndicator.tsx
@@ -6,7 +6,7 @@ import { useTheme } from '@mui/material/styles';
 import CircularProgress from '@mui/material/CircularProgress';
 import { useLoading } from 'ra-core';
 
-import { RefreshIconButton } from '../button';
+import { RefreshIconButton, RefreshIconButtonProps } from '../button';
 import { SxProps } from '@mui/system';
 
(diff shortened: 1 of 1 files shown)


fzaninotto:
That's indeed a very minor change that we can merge. Yet it's a new feature, so you should PR against `next` instead of `master`.

Also, TypeScript complains.

david-bezero:
I have fixed the types. To avoid going out-of-sync in the future, I'm just forwarding the type from `RefreshIconButtonProps` (technically the type it expects is `((e: MouseEvent) => void) & MouseEventHandler<HTMLButtonElement>`, which looks like a mistake — it should probably be `MouseEventHandler<HTMLButtonElement>` — but that's a separate issue in `RefreshIconButton`)

I followed the existing pattern for merging props from `RefreshIconButton`.

slax57 (approved):
LGTM, thanks!

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9426, opened 2023-11-07
[Doc] Improve the Create redirect section

situplastik:
### Problem

The chapters of the Create redirect section don't list function option as it show redirect Edit section while react-admin also admit that option.

### Solution

Add that options as Edit redirect chapters


slax57 (approved):
Nice catch, thanks!

PR #9430, opened 2023-11-08
Revamp demo page and add EE demo

erwanMarmelab:
## Problem
The page in https://marmelab.com/react-admin/Demos.html doesn’t mention the EE demo, but it’s a great one!

The section should highlight the most important features.

It’s a good opportunity to revamp the demos page, e.g. following https://mui.com/toolpad/examples/ design

## Solution
Revamp the demo page, add EE demo, and highlight important features

## Todo
- [x] Revamp demo page
- [x] Add EE demo
- [x] Highlight important features
- [x] Adapt mobile version
- [x] Fix video
- [x] Change iframe to video for the 1st demo

### Before
image

### After
image


docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+    @media (max-width: 600px) {
+        .demos-list {
+            display: flex;
+            flex-wrap: wrap;
+            gap: 10px;
+        }
+    }
+
+    .card {
+        box-shadow: 0 4px 8px 0 rgba(0,0,0,0.2);
```
adguernier, line 31: The shadow is a bit strong in my opinion
```suggestion
        box-shadow: 0 2px 4px 0 rgba(0,0,0,0.2);
```

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
 
+## Overview
+
+Discover our demos that show you the react-admin potential and help you familiarize with it.
+
+<style>
+    .demos-list {
+        display: grid;
+        grid-template-columns: repeat(2, 1fr);
+        grid-template-rows: repeat(3, 1fr);
```
slax57, line 18: I'm not a css grid specialist, but I'm a bit surprised you have to set both `grid-template-columns` and `grid-template-rows`. To me, you could keep only `grid-template-columns`.

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+        grid-gap: 10px;
+    }
+
+    @media (max-width: 600px) {
+        .demos-list {
+            display: flex;
+            flex-wrap: wrap;
+            gap: 10px;
+        }
+    }
```
slax57, line 28: Why switch to `flex` on mobile mode? Why not keep `grid` and only change `grid-template-columns` to be `repeat(1, 1fr);`?

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+
+    .card {
+        box-shadow: 0 4px 8px 0 rgba(0,0,0,0.2);
+        transition: 0.3s;
+        background-color: rgba(243, 246, 249, 0.4);
+        border: 1px solid rgba(0,0,0,0.25);
+        border-radius: 30px;
```
slax57, line 35: `border-radius` is a bit too high IMHO. It doesn't fit well with the rest of the UI, and crops out part of the videos in a weird way, especially on mobile.

image


But this is a taste issue, you may want to gather more feedback from the team before changing anything.

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+    }
+
+    .card:hover {
+        box-shadow: 0 8px 16px 0 rgba(0,0,0,0.2);
+    }
+
+    .card > img, .card > video {
+        border-radius: 15px;
+        width: 100%;
+    }
```
slax57, line 48: It's a bit of a shame that the card's video/img/iframe do not all have the same height, as it creates inconsistencies in the layout...

image


Ideally they should all be the same height, but that would involve re-making some of the videos, which can be time consuming...

Alternatively, maybe it is enough to only reserve the same height for each item?

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+    .no-decoration, .link {
+        text-decoration: none !important;
+    }
+</style>
+
+<div class="demos-list">
+    <div class="card">
+        <iframe src="https://player.vimeo.com/video/474999017?autoplay=1&loop=1&muted=1" frameborder="0" allowfullsreen></iframe>
+        <div class="content-card">
+            <a href="#e-commerce" class="no-decoration">
```
adguernier, line 111: I'm a bit lost when I click the links anchor. I would have imagined dedicated pages for each sections instead, that explain further what they demonstrate. 
What do you think about it?

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+                <p><b><a href="https://marmelab.com/react-admin-demo/" class="demo link">Demo</a></b></p>
+                <p><b><a href="https://github.com/marmelab/react-admin/tree/master/examples/demo" class="source-code link">Source code</a></b></p>
+            </div>
+        </div>
+    </div>
+    <div class="card">
+        <video src="https://github.com/marmelab/react-admin/assets/131013150/93ec41f5-81bf-45e1-940b-d5aeccfe628f" autoplay loop></video>
+        <div class="content-card">
+            <a href="#e-commerce-enterprise-version" class="no-decoration">
+                <p class="title-card"><b>E-commerce enterprise version</b></p>
```
slax57, line 127:
```suggestion
                <p class="title-card"><b>E-commerce Enterprise</b></p>
```

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+                <p  class="description-card">Based on the E-commerce demo and upgraded with some ReactAdmin-Enterprise modules.</p>
```
slax57, line 128:
```suggestion
                <p  class="description-card">Based on the E-commerce demo and upgraded with some React Admin Enterprise features. Built by the core team.</p>
```

docs/Demos.md
```diff
@@ -29,6 +219,30 @@ The source shows how to implement the following features:
 - [Custom page with static list](https://github.com/marmelab/react-admin/blob/master/examples/demo/src/routes.tsx)
 - [Edit view in a sidebar](https://github.com/marmelab/react-admin/blob/master/examples/demo/src/reviews/ReviewList.tsx)
 
+## E-commerce enterprise version
+
+Based on the E-commerce demo and upgraded with some ReactAdmin-Enterprise modules.
```
slax57, line 224:
```suggestion
Based on the E-commerce demo and upgraded with some React Admin Enterprise features.
```

docs/Demos.md
```diff
@@ -29,6 +219,30 @@ The source shows how to implement the following features:
...
+
+<video src="https://github.com/marmelab/react-admin/assets/131013150/93ec41f5-81bf-45e1-940b-d5aeccfe628f" controls="controls" style="max-width: 100%"></video>
```
slax57, line 226: Why not use the same tags (and style) as other videos in the docs?

In particular, it seems the video player is a but cropped to the right.

image

Also, can you not include the browser frame in the video?

Lastly, apart from a few exceptions, it seems to me you are not focusing enough on the Enterprise features included in this demo, and you are instead demonstrating mainly the same features as the OSS version of this demo. We can have a look together of the Enterprise features we should highlight if you need.

docs/Demos.md
```diff
@@ -7,6 +7,196 @@ title: "React-admin demos"
...
+        flex-direction: column;
+    }
+
+    .card:hover {
+        box-shadow: 0 8px 16px 0 rgba(0,0,0,0.2);
+    }
+
+    .card > img, .card > video {
+        border-radius: 15px;
+        width: 100%;
```
adguernier, line 47: Videos are not centered, there is more right margin because of the `.markdown-section img, .markdown-section video` rules.
image
This code will fix this:
```suggestion
        max-width: 100%;
```

docs/Demos.md
```diff
@@ -29,6 +219,30 @@ The source shows how to implement the following features:
...
+- [SolarLayout](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/layout/Layout.tsx)
+- [Locks](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/products/GridList.tsx)
+- [ListLive](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/orders/OrderList.tsx)
+- [RealtimeDataProvider](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/dataProvider/addRealtimeMethodsWithFakeTransport.ts)
+- [CompleteCalendar](.https://github.com/marmelab/ra-enterprise-demo/blob/master/src/visits/VisitList.tsx)
+- [Audit Log](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/i18n/en.ts)
+- [SearchWithResult](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/layout/SearchSubMenu.tsx)
+- [TreeWithDetails](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/categories/CategoryList.tsx)
+- [MarkdownField](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/products/ProductPreview.tsx)
+- [MarkdownInput](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/products/ProductCreate.tsx)
```
slax57, line 244: This list is missing:

- [Tours](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/tours/TourList.tsx)
- [EditableDatagrid](https://github.com/marmelab/ra-enterprise-demo/blob/master/src/stores/StoreList.tsx)

docs/img/enterprise-demo.png
slax57, line 1: Where is this file used?

slax57 (changes requested):

adguernier (changes requested):

docs/Demos.md
erwanMarmelab: I'm a bit worried about the number of pages this will create. won't it overload the navbar?

slax57: For the record, I disagree with @adguernier , IMO the anchors are fine and we should keep them that way.
To be discussed with @fzaninotto then.

fzaninotto (changes requested):
The new layout is rad!

I think the anchor experience is fine and we shouldn't add more pages. 

However, there are two adjustments you need to make:

1. The cards shouldn't show an autoplay video of each demo, because these videos are too small, and it makes the page change too much. Instead, the cards should only show a still image extracted from each demo, and the section below would show the video.
2. The video for the Ee edition is too large to be leaded as a mp4 file. We need to host it in a video provider (Vimeo). I'll host it with the Marmelab account and we'll adjust the embed.

fzaninotto (approved):

Issue #9432, opened 2023-11-09
Feature Request: Enhancing `<CheckForApplicationUpdate>` for module federation apps


smeng9:
**Is your feature request related to a problem? Please describe.**

1. Our application uses the module federation design but the `<CheckForApplicationUpdate>` can only check for a single url. It cannot check multiple urls for each module

2. Additionally the fetch function in getHashForUrl https://github.com/marmelab/react-admin/blob/f030350259ab8a5eb006b82a2a800568c6ee9230/packages/ra-core/src/util/useCheckForApplicationUpdate.ts#L61 does not support passing additional options like cache control, so even after the interval has passed, it may check for a stale copy of data from cache which breaks the purpose of CheckForApplicationUpdate. This may be a bug.

3. We would also like an additional callback function handleApplicationUpdate passed into the `<CheckForApplicationUpdate>` so we can customize the button click behavior https://github.com/marmelab/react-admin/blob/f030350259ab8a5eb006b82a2a800568c6ee9230/packages/ra-ui-materialui/src/layout/ApplicationUpdatedNotification.tsx#L17 like making backup of localstorage, give user warning about potential data loss, etc.

**Describe the solution you'd like**

Enhance the `<CheckForApplicationUpdate>` interface to make url support an array of strings, add the fetch options object, and add the callback.

**Describe alternatives you've considered**

Stack a bunch of CheckForApplicationUpdate in the layout, each checking for a single url. However that would not solve problem 2 and 3.
Hope these enhancement can benefit other people so they don't have to reimplement the react hooks.

**Additional context**

fzaninotto:
Hi, and thanks for your feedback. 

Regarding your points:

1. The `CheckForApplicationUpdate` component won't support module federation. You will have to build your own component for that. 
2. We're open to a PR adding the support for an `options` prop passed as second argument to `fetch`
3. It's already possible: the parameter is called `onNewVersionAvailable`. We do need to document it, though, so I'm marking this as a documentation issue.

PR #9437, opened 2023-11-09
Add documentation for useCheckForApplicationUpdate

smeng9:
Closes https://github.com/marmelab/react-admin/issues/9432


fzaninotto:
Thanks for the PR. 

I think the hook should remain private - we just can't document all hooks, or the doc will become too large. As the component supports all the props of the hook, you should just document the `onNewVersionAvailable` prop in [the `<CheckForApplicationUpdate>` component chapter](https://marmelab.com/react-admin/CheckForApplicationUpdate.html).

fzaninotto (approved):

fzaninotto:
Thanks!

slax57:
@smeng9 Thanks for your reply.

You are correct about the TS type, I'll open a quick PR.

Regarding the space in the prop's name in the documentation, this is intentional, to allow the table to break the word for a better layout.

PR #9439, opened 2023-11-10
Update `<SingleFieldList>` to make it more powerful

### LINKED ISSUE #7634: SingleFieldList hasn't empty props
When using ReferenceArrayField, the emptyText props isn't available. Instead the doc say that we should use the child empty props. The Datagrid component has an empty props but the SingleFieldList hasn't

fzaninotto:
## Problems

`<SingleFieldList>` has many issues:

- It still clones its children (we don't do this since 3.x)
- It forces users to pass children, while most of the time it's a `<ChipField>`
- It renders kind of a `Stack`, except it's not a `Stack`
- It doesn't support the `empty` prop

## Solution

- `<SingleFieldList>` now renders a real `<Stack>`, and supports all its props
- `<SingleFieldList>` used without children renders a `<ChipField>` with the `recordRepresentation`
- `<SingleFieldList>` supports an `empty` prop

```jsx
const PostList = () => (
    <List>
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <NumberField source="views" />
            <ReferenceArrayField label="Tags" reference="tags" source="tags">
                <SingleFieldList />
            </ReferenceArrayField>
        </Datagrid>
    </List>
)
```

Closes #7634

## Minor BC Break

If you used `<SingleFieldList>` with a `<ChipField>` child and enabled the links, the chips background will no longer change on hover. To reenable this visual effect, add the `clickable` prop to the `<ChipField>`:

```diff
const PostList = () => (
    <List>
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <DateField source="published_at" />
            <BooleanField source="commentable" />
            <NumberField source="views" />
            <ReferenceArrayField label="Tags" reference="tags" source="tags">
                <SingleFieldList>
-                  <ChipField source="name" />
+                  <ChipField source="name" clickable />
               </SingleFieldList>
            </ReferenceArrayField>
        </Datagrid>
    </List>
)
```

Alternately, if the `<ChipField source>` is the same as the resource's `recordRepresentation`, you can omit the child altogether. 

```diff
const PostList = () => (
    <List>
        <Datagrid>
            <TextField source="id" />
            <TextField source="title" />
            <DateField source="published_at" />
            <BooleanField source="commentable" />
            <NumberField source="views" />
            <ReferenceArrayField label="Tags" reference="tags" source="tags">
-               <SingleFieldList>
-                  <ChipField source="name" />
-              </SingleFieldList>
+              <SingleFieldList />
            </ReferenceArrayField>
        </Datagrid>
    </List>
)
```

### DIFF
--- a/docs/SingleFieldList.md
+++ b/docs/SingleFieldList.md
@@ -13,7 +13,9 @@ Use `<SingleFieldList>` when you want to display only one property for each reco
 
 ## Usage
 
-`<SingleFieldList>` grabs the current `ListContext`, and renders a Material UI `<Stack>` with one `<ChipField>` for each record in the list, using the `recordRepresentation`. It is especially useful as child of `<ReferenceManyField>` and `<ReferenceArrayField>` components. 
+`<SingleFieldList>` grabs the current `ListContext`, and renders a Material UI `<Stack>` with one `<ChipField>` for each record in the list, using the `recordRepresentation`. It is especially useful as child of `<ReferenceManyField>` and `<ReferenceArrayField>` components.
+
+Here is an example of a Post show page showing the list of tags for the current post:
 
 ```jsx
 import {
@@ -36,7 +38,7 @@ const PostShow = () => (
 );
 ```
 
-You can customize how each record is displayed by passing a Field component as child. For example, here is how to use `<SingleFieldList>` to display a list of tags for each post in a Datagrid:
+You can also use  `<SingleFieldList>` in a list view, e.g. to display the tags for each post in a `<Datagrid>`:
 
 ```jsx
 import { 
@@ -68,6 +70,14 @@ const PostList = () => (
 ```
 
 ![SingleFieldList in Datagrid](./img/singlefieldlist-datagrid.png)
+
+You can customize how each record is displayed by passing a Field component as child. For example, you can change the field name used by the `<ChipField>`:
+
+```jsx
+<SingleFieldList>
+    <ChipField source="tag" clickable />
+</SingleFieldList>
+```
 
 ## Props
 
--- a/docs/SingleFieldList.md
+++ b/docs/SingleFieldList.md
@@ -13,15 +13,30 @@ Use `<SingleFieldList>` when you want to display only one property for each reco
 
 ## Usage
 
-Use `<SingleFieldList>` wherever there is a `ListContext`. It is especially useful as child of `<ReferenceManyField>` and `<ReferenceArrayField>` components. `<SingleFieldList>` expects a single `<Field>` as child.
+`<SingleFieldList>` grabs the current `ListContext`, and renders a Material UI `<Stack>` with one `<ChipField>` for each record in the list, using the `recordRepresentation`. It is especially useful as child of `<ReferenceManyField>` and `<ReferenceArrayField>` components. 
 
 ```jsx
-<SingleFieldList>
-    <ChipField source="name" />
-</SingleFieldList>
+import {
+    Show,
+    SimpleShowLayout,
+    TextField,
+    ReferenceArrayField,
+    SingleFieldList
+} from 'react-admin';
+
+const PostShow = () => (
+    <Show>
+        <SimpleShowLayout>
+            <TextField source="title" />
+            <ReferenceArrayField label="Tags" reference="tags" source="tags">
+                <SingleFieldList />
+            </ReferenceArrayField>
+        </SimpleShowLayout>
+    </Show>
+);
 ```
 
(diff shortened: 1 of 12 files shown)


packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -89,12 +102,9 @@ export const SingleFieldList = (props: SingleFieldListProps) => {
                                 to={resourceLinkPath}
                                 onClick={stopPropagation}
                             >
-                                {cloneElement(Children.only(children), {
-                                    record,
-                                    resource,
-                                    // Workaround to force ChipField to be clickable
-                                    onClick: handleClick,
-                                })}
```
fzaninotto, line 97: This is technically a BC break, as people may use SingleFieldList with children expecting `record` as prop. However, it's not supposed to be the case since 4.0.

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -174,7 +181,11 @@ const Root = styled('div', {
 // useful to prevent click bubbling in a datagrid with rowClick
 const stopPropagation = e => e.stopPropagation();
 
-// Our handleClick does nothing as we wrap the children inside a Link but it is
```
fzaninotto, line 177: We no longer need this: the default child uses the `clickable` prop only when necessary. And when developers pass a custom child, they'll have to set the `clickable` themselves if they want. In that case, this PR is a slight BC break.

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -55,20 +51,35 @@ export const SingleFieldList = (props: SingleFieldListProps) => {
     const {
         className,
         children,
+        empty,
         linkType = 'edit',
-        component: Component = Root,
```
fzaninotto, line 59: `component` is now passed down to `<Stack>`, which knows how to handle it.

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
```diff
@@ -1,22 +1,18 @@
...
     RecordContextProvider,
+    RecordRepresentation,
     ComponentPropType,
     useCreatePath,
 } from 'ra-core';
 
+import { LinearProgress } from '../layout/LinearProgress';
 import { Link } from '../Link';
 
 /**
```
adguernier, line 18: Since the component works without children, you might add an example to demonstrate it.

adguernier (approved):

docs/SingleFieldList.md
```diff
@@ -13,15 +13,30 @@ Use `<SingleFieldList>` when you want to display only one property for each reco
...
+            <ReferenceArrayField label="Tags" reference="tags" source="tags">
+                <SingleFieldList />
+            </ReferenceArrayField>
+        </SimpleShowLayout>
+    </Show>
+);
 ```
 
-The following example shows how to use `<SingleFieldList>` to display a list of tags for each post in a Datagrid:
+You can customize how each record is displayed by passing a Field component as child. For example, here is how to use `<SingleFieldList>` to display a list of tags for each post in a Datagrid:
```
slax57, line 39: I'm confused by this paragraph and the code snippet below. I would have expected an example showing how to pass a custom children to `<SingleFieldList>`, like a `<ChipField>`. Instead, this example is in the end very similar to the previous one, except it's in a `<Datagrid>` instead of a `<SimpleShowLayout>`. 
Unless this is actually the purpose of this example, in which case I would have removed the sentence "You can customize how each record is displayed by passing a Field component as child".

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
slax57: Unless I'm mistaken, until now developers always **had** to pass a child. So it's a BC for everyone using `<SingleFieldList>` right now unless they have specifically disabled the link.

I agree it's a very minor BC, but it will still impact almost everyone using `<SingleFieldList>` right now. Maybe we should insist on it more in the docs and/or the changelog?

slax57 (changes requested):
Other than that, it works great! :tada: 
And the stories are very helpful.

packages/ra-ui-materialui/src/list/SingleFieldList.tsx
fzaninotto: You're right. I updated the PR description, and we should include the "Minor BC break" section in the release changelog.

slax57 (approved):

PR #9448, opened 2023-11-15
[Doc] Fix api implementation in Remix tutorial throws a decoding error when less than 10 records are returned

### LINKED ISSUE #9445: Remix & Supabase Example possible correction
I was following the remix example and experienced the following.

When requesting a list of less than 10 items the fetchUtils was giving a net::ERR_CONTENT_DECODING_FAILED 200 (OK) response.
When requesting a list of 11 items or more then fetchUtils was working perfectly !

I cloned the marmelab/ra-supabase example and adding a Remix app to the project it works perfectly with the example.

There appears to be a difference between a local supabase setup and using the hosted online version.

It appears the hosted Supabase adds Content-Encoding of "br" when the list is less than 10. This causes the resolution of the response.text promise within the fetchUtil to fail.

Changing the headers in the admin.api.$.tsx to specify Accept-Encoding: '' appears to resolve the issue and supabase returns the content not encoded (compressed) and the fetchUtil works.

    headers: {
      prefer: request.headers.get("prefer") ?? "",
      accept: request.headers.get("accept") ?? "application/json",
      'Accept-Encoding': '',
      apiKey: `${process.env.SUPABASE_SERVICE_ROLE}`,
      Authorization: `Bearer ${process.env.SUPABASE_SERVICE_ROLE}`,
    },

If there is a better alternative or fix which is usually used and the above is incorrect I'd appreciate knowing.

erwanMarmelab:
## Problem
In the Remix tuto, RA had a decoding error (net::ERR_CONTENT_DECODING_FAILED) on query that have less than 10 results.

##  Solution
Add accept `Accept-Encoding` prop to header of the query

## Issue
Closes #9445

### DIFF
--- a/docs/Remix.md
+++ b/docs/Remix.md
@@ -5,7 +5,7 @@ title: "Remix Integration"
 
 # Remix Integration
 
-[Remix](https://remix.run/) is a Node.js framework for server-side-rendered React apps. But even if react-admin is designed to build Single-Page Applications, Remix and react-admin integrate seamlessly. 
+[Remix](https://remix.run/) is a Node.js framework for server-side-rendered React apps. But even if react-admin is designed to build Single-Page Applications, Remix and react-admin integrate seamlessly.
 
 These instructions are targeting Remix v2.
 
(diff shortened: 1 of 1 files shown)


slax57 (approved):
I must admit I fail to understand why this fix works, but I could confirm it does, so I'll accept it.
Kudos @ShaneC112 for finding it!

Issue #9451, opened 2023-11-16
useShowContext returns record from list endpoint


sebashwa:
I'm using a REST-API and I'm returning a minimal record from the list endpoint (e.g. `/posts`) and a full version record from the detail endpoint (e.g. `/posts/2`). I'm also using `useShowContext` for the show page.

**What you were expecting:**
I was expecting `useShowContext` to always return the record fetched via `getOne`, as stated in the documentation:

```JS
const {
  [...]
  record, // record fetched via dataProvider.getOne() based on the id from the location
  [...]
} = useShowContext();
```

**What happened instead:**
The record from the list endpoint (probably cached via react query) is returned from `useShowContext`.

**Steps to reproduce:**
Return different records from your REST-API for list / detail endpoints (see edited `getList` for posts in the example `dataProvider` below). Navigate from your list page (posts) to the show page.

**Related code:**
[See this example](https://stackblitz.com/edit/github-msupcd?file=src%2FdataProvider.tsx,src%2Fposts%2FPostShow.tsx)

**Environment**

* React-admin version: 4.15.4
* Last version that did not exhibit the issue (if applicable): -
* React version: 18.2.0
* Browser: all

erwanMarmelab:
Hi @sebashwa ,

The bug you are reporting is a feature. As you can read in [optimistic rendering](https://marmelab.com/react-admin/Caching.html#optimistic-rendering) and [features](https://marmelab.com/react-admin/Features.html#fast) documentations, we make an "optimistic rendering".

Cheers,

Erwan

sebashwa:
Hi @erwanMarmelab,

thanks for your fast answer! Sorry I didn't find the Caching docs, it explains the behaviour very well. I think it would be good to link this page from some other places in the documentation (especially where it states that the response from `.getOne` is returned for `record`). I'd like to open a PR for this, if you think that's useful?

fzaninotto:
Great idea, please open a PR!

PR #9453, opened 2023-11-16
Link to caching documentation page for hooks returning records

sebashwa:
Records returned from a context / controller hook are sometimes `undefined` or a cached version of the record. This is now stated more explicitly in the hook documentations for hooks that return the record from a context.

Things changed:

* Had to move some documentation out of code blocks to add those links to the caching page. 
* Tried to make the documentation for the list / edit / create / show hooks more consistent (added headings / using the same documentation format for Arguments / Return Value). 
* Added "Props"-Headings for some more components to pop up in the navigation

Related to #9451


docs/useCreateContext.md
```diff
@@ -40,14 +40,10 @@ const PostCreateForm = () => {
...
 
-```jsx
-const {
-    defaultTitle, // the translated title based on the resource, e.g. 'Create New Post'
-    redirect, // the default redirection route. Defaults to 'list'
-    resource, // the resource name, deduced from the location. e.g. 'posts'
-    save, // the update callback, to be passed to the underlying form as submit handler
-    saving, // boolean that becomes true when the dataProvider is called to create the record
-} = useCreateContext();
-```
```
fzaninotto, line 53: I prefer the previous version

docs/useCreateController.md
```diff
@@ -43,24 +43,24 @@ export const BookCreate = () => {
...
 
-```jsx
-const {
-    defaultTitle, // the translated title based on the resource, e.g. 'Create New Post'
-    redirect, // the default redirection route. Defaults to 'list'
-    resource, // the resource name, deduced from the location. e.g. 'posts'
-    save, // the update callback, to be passed to the underlying form as submit handler
-    saving, // boolean that becomes true when the dataProvider is called to create the record
-} = useCreateController();
-```
```
fzaninotto, line 65: I prefer this version

docs/useEditContext.md
```diff
@@ -45,23 +45,19 @@ const Aside = () => {
...
-    isLoading, // boolean that is true until the record is available for the first time
-    mutationMode, // mutation mode argument passed as parameter, or 'undoable' if not defined
-    record, // record fetched via dataProvider.getOne() based on the id from the location
-    redirect, // the default redirection route. Defaults to 'list'
-    refetch, // a function that allows you to refetch the record 
-    resource, // the resource name, deduced from the location. e.g. 'posts'
-    save, // the update callback, to be passed to the underlying form as submit handler
-    saving, // boolean that becomes true when the dataProvider is called to update the record
-} = useEditContext();
-```
```
fzaninotto, line 64: Same

docs/useEditController.md
```diff
@@ -46,31 +46,31 @@ export const BookEdit = () => {
...
-    resource, // the resource name, deduced from the location. e.g. 'posts'
-    save, // the update callback, to be passed to the underlying form as submit handler
-    saving, // boolean that becomes true when the dataProvider is called to update the record
-} = useEditController();
-```
+## Return Value
+
+`useEditController` returns an object with the following keys:
+
+* `defaultTitle`: Translated title based on the resource, e.g. 'Post #123'
```
fzaninotto, line 66: keep the object format

docs/useList.md
```diff
@@ -255,39 +256,46 @@ const { data } = useList({
 
 ## Return Value
 
-`useList` returns an object matching the shape of the `ListContext`: 
+`useList` returns an object with keys matching the shape of the `ListContext`: 
 
-```jsx
```
fzaninotto, line 260: and the object syntax

docs/useListContext.md
```diff
@@ -56,42 +56,49 @@ export const PostList = () => (
 
 ## Return Value
 
-The `useListContext` hook returns an object with the following keys:
+`useListContext` returns an object with the following keys:
 
-```jsx
```
fzaninotto, line 61: Please keep the object value

docs/useListController.md
```diff
@@ -141,39 +141,46 @@ You can disable this feature by setting the `storeKey` prop to `false`. When dis
 
 ## Return Value
 
-The return value of `useListController` has the following shape:
```
fzaninotto, line 144: PLease keep the object format

docs/useSaveContext.md
```diff
@@ -41,3 +37,11 @@ const MyComponent = () => {
 }
 ```
 {% endraw %}
+
+## Return Value
+
+`useSaveContext` returns an object with the following keys:
+
+* `save`: Create or update callback which receives form data and calls `dataProvider`
```
fzaninotto, line 45: MAke it a code snippet

docs/useShowContext.md
```diff
@@ -49,6 +43,19 @@ const PostShow = () => (
 )
 ```
 
+## Return Value
+
+`useShowContext` returns an object with the same keys as [`useShowController`](./useShowController.md):
+
+* `defaultTitle`: Translated title based on the resource, e.g. 'Post #123'
```
fzaninotto, line 50: same

docs/useShowController.md
```diff
@@ -176,6 +166,18 @@ const PostShow = () => {
 };
 ```
 
+## Return Value
+
+`useShowController` returns an object with the following keys:
+
+* `defaultTitle`: Translated title based on the resource, e.g. 'Post #123'
```
fzaninotto, line 173: Please use an object in a code snippet

docs/useCreateController.md
```diff
@@ -43,24 +43,24 @@ export const BookCreate = () => {
 
 **Tip**: If you just use the return value of `useCreateController` to put it in an `CreateContext`, use [the `<CreateBase>` component](./CreateBase.md) instead for simpler markup.
 
-`useCreateController` accepts an options argument, with the following fields, all optional:
+## Arguments
```
fzaninotto, line 46: We use Parameters in most hooks. Why is Arguments better? And if you change it here, you must change it everywhere.

fzaninotto (changes requested):
Nice! Although I don't agree with all your proposed changes.

docs/useCreateController.md
sebashwa: Wasn't sure what's the preference, both work for me. Will change it back soon!

sebashwa:
Hey @fzaninotto,

thanks for your feedback. My intention to use the listing instead of the object syntax in a codeblock were:

* It's more consistent with the documentation for parameters (which are listings in most of the places, although sometimes also tables in case types / default values are documented) and I don't see why the return value should be documented differently from the parameters.
* It's not necessary to scroll to the right to read the descriptions of the object keys (at least I have to, maybe it's a resolution thing)
* I thought it's not possible to put links inside codeblocks and I wanted to link to the "Caching" page for the returned `record` key. This was the main reason why I abandoned the codeblocks, although I think now, that's actually not true.

If you still think I should switch back to codeblocks, I will of course do so but imho I think it's neither more readable nor consistent with the rest of the documentation. What do you think?

fzaninotto:
I hear your arguments, yet I prefer the code snippet for the return values.

fzaninotto:
Any updates?

sebashwa:
> Any updates?

Wasn't able to add those changes earlier.
I reverted to the code block documentation style, although it's now no longer possible to link to the Caching page from within the code block. So it just remains a hint to the Caching page for now.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9454, opened 2023-11-17
Fix: reorder Discord and Github icons to match the website order

adguernier:
Buttons in toolbar of RA website and RA docs are not in the same order!

This PR attempts to reorganize them and fight against the chaos


slax57 (approved):
:crossed_swords:  :octopus:

Issue #7580, opened 2022-04-21
New bundling strategy makes it impossible to debug and apply patches


andrico1234:
Hi folks!

Congrats on the 4.0.0 release 🥳 🎈 

We've been upgrading our codebase, and we're coming across a handful of issues. We've managed to get to the bottom of most of them, but we're no longer able to debug in the same way that we were able to previously.

I believe that this is result of two significant changes:

## Issue 1
First, the React Admin output is minified. 

In v3, the code was transpiled down from TS to JS,with little or no other code transformation being applied. Since v4 the code is also being minified, which causes a lot of problems with debugging. 

No longer can we jump into our node_modules and run debuggers and logs to observe what's happening with our code. It also means that it's impossible to use tools like `patch-package` to make any small bug fixes that may not be addressed until future versions of RA.

I believe that minification is an application-level concern, and library authors shouldn't need to perform this step. It's up to developers to decide on how they want to bundle their code for deployment. It makes a lot of sense to revert this change, though I'm also interested to hear what the benefits are of this new approach.

## Issue 2
The other issue, which may have just been a little oversight is that we're no longer able to click on an export to get directed to the source code, instead, it's reverted to taking us to the types again. 

I don't think that I'll be able to replicate this via a CodeSandbox, but I'll whip up a before/after video

## Before
<img width="1417" alt="image" src="https://user-images.githubusercontent.com/17087167/164444704-f98bfbd8-566f-4e2a-9c52-b1e538e61c18.png">

## After

Shows what the output of the ESM bundle looks like
<img width="1641" alt="image" src="https://user-images.githubusercontent.com/17087167/164444157-d4d3f76e-0fc6-4380-879d-6b13d0343966.png">

Edit:

Removing `--minify` from the build script prevents the output from being minified, which is good, but keeps the entire output contained in a single `index.js` file. Are there code splitting concerns to be had with compiling the entire output into a single file?

**Other information:**

**Environment**

* React-admin version:
* Last version that did not exhibit the issue (if applicable):
* React version:
* Browser:
* Stack trace (in case of a JS error):

djhi:
Those are good points, thank you for reporting them. I'll discuss with the team to see what we can do

djhi:
FYI, the `src` directory is still included in the packages. Can't you configure your bundler and `patch-package` to use it?

andrico1234:
I figure that's not too difficult to do. I haven't tried it, but it seems feasible. I'll let you know how I get on!

[...]

PR #9457, opened 2023-11-18
Add exports to package.json in all packages

### LINKED ISSUE #9452: Add exports entry point to support node's module resolution.
React admin provides cjs and esm bundles. However, it relies on `main` and `module` in the package.json file to decide which version to use. It will work with most bundlers but fails when using node's module resolution as it doesn't support the `module` field. I was hit with this problem when trying to test a component using vitest.

Example:
```js
function ComponentA() {
   return <SimpleForm><ComponentB /></SimpleForm>
}

function ComponentB() {
   const { watch } = useFormContext()
   return ...
}
```

SimpleForm is loaded using cjs and uses FormProvider (cjs) from react-hook-form.
ComponentB uses useFormContext from react-hook-form esm bundle and context will be null.

Add `exports` to package.json specifying the entry points for `import` and `require`.

**Additional context**
https://github.com/vitest-dev/vitest/discussions/4233
https://nodejs.org/api/packages.html#package-entry-points

otaviosoares:
fixes #9452

### DIFF
--- a/examples/data-generator/package.json
+++ b/examples/data-generator/package.json
@@ -9,7 +9,7 @@
     "types": "dist/index.d.ts",
     "scripts": {
         "build": "yarn tsup",
-        "watch": "tsc --outDir dist/esm --module es2015 --watch"
+        "watch": "yarn tsup --watch"
     },
     "dependencies": {
         "date-fns": "^2.19.0",
--- a/examples/data-generator/package.json
+++ b/examples/data-generator/package.json
@@ -4,13 +4,11 @@
     "homepage": "https://github.com/marmelab/react-admin/tree/master/examples/data-generator",
     "bugs": "https://github.com/marmelab/react-admin/issues",
     "license": "MIT",
-    "main": "dist/cjs/index.js",
-    "module": "dist/esm/index.js",
-    "types": "dist/cjs/index.d.ts",
+    "main": "dist/index.js",
+    "module": "dist/index.mjs",
+    "types": "dist/index.d.ts",
     "scripts": {
-        "build": "yarn run build-cjs && yarn run build-esm",
-        "build-cjs": "rimraf ./dist/cjs && tsc --outDir dist/cjs",
-        "build-esm": "rimraf ./dist/esm && tsc --outDir dist/esm --module es2015",
+        "build": "yarn tsup",
         "watch": "tsc --outDir dist/esm --module es2015 --watch"
     },
     "dependencies": {
@@ -21,6 +19,7 @@
         "cross-env": "^5.2.0",
         "ra-core": "^4.15.5",
         "rimraf": "^3.0.2",
+        "tsup": "^8.0.0",
         "typescript": "^5.1.3"
     },
     "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
         "react-dom": "^17.0.0",
         "storybook": "^7.5.1",
         "ts-jest": "^29.1.0",
+        "tsup": "^7.3.0",
         "typescript": "^5.1.3",
         "wait-on": "^3.2.0",
         "whatwg-fetch": "^3.0.0"
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -10,6 +10,13 @@
     "main": "dist/cjs/index.js",
     "module": "dist/esm/index.js",
     "types": "dist/cjs/index.d.ts",
+    "exports": {
+        ".": {
+            "types": "./dist/cjs/index.d.ts",
+            "import": "./dist/esm/index.js",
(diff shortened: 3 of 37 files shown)


djhi:
Thanks! Isn't it a duplicate of #9332?

smeng9:
I think it's a duplicate.

#9332 fixes vitest but creates an another problem for remix after we add the exports field in package.json to force remix to resolve to an esm package instead of cjs. The reason of failure is remix's file resolution algorithm for esm packages requires specific file extension and package type during compile. If you go to https://www.npmjs.com/package/react-admin?activeTab=code react-admin/dist/esm/index.js and you can find it is `export * from './Admin';` without .js extension. While vite is smart enough to iterate through all possible extensions and find the file, remix is not clever.

Now we may have two options:
1. Add exports field, then add `{"type": "module"}` to `dist/esm/package.json`, then add *.js to all files in the source code during exports, then add *.js to all files regarding `lodash` and `@mui/icon-material` imports, then fix the import of `@mui/material/styles`, then fix react-router by importing the whole package and then unpack it since remix cannot correctly resolve default exports and we are done. Both remix and vitest can be happy.

2. Wait another year for remix to release v3 as they announced in their blog post https://remix.run/blog/remix-heart-vite they don't want to "be building a worse version of Vite", and "In fact, with Vite, Remix is no longer a compiler. Remix itself is just a Vite plugin:"

djhi:
> Wait an undisclosed amount of time or another year for remix to release v3 as they announced in their blog post https://remix.run/blog/remix-heart-vite they don't want to "be building a worse version of Vite", and "In fact, with Vite, Remix is no longer a compiler. Remix itself is just a Vite plugin:"

Vite is already available in Remix v2, although flagged as unstable. I'm using it on another project without any issues. Haven't tested react-admin with it though

smeng9:
Oh I don't no the already support it in unstable.

However shall we update the remix tutorial to recommend users to use unstable remix plugin for vite?

otaviosoares:
> Now we may have two options:
> 
> 1. Add exports field, then add `{"type": "module"}` to `dist/esm/package.json`, then add *.js to all files in the source code during exports, then add *.js to all files regarding `lodash` and `@mui/icon-material` imports, then fix the import of `@mui/material/styles`, then fix react-router by importing the whole package and then unpack it since remix cannot correctly resolve default exports and we are done. Both remix and vitest can be happy.
> 2. Wait an undisclosed amount of time or another year for remix to release v3 as they announced in their blog post https://remix.run/blog/remix-heart-vite they don't want to "be building a worse version of Vite", and "In fact, with Vite, Remix is no longer a compiler. Remix itself is just a Vite plugin:"

what about using `moduleResolution: bundler` in tsconfig.json and bundle the code using vite or esbuild to avoid having to add file extensions to imports?

smeng9:
Do you mean we have to change how we build and release react-admin, switching from tsc to vite or esbuild?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

otaviosoares:
> Do you mean we have to change how we build and release react-admin, switching from tsc to vite or esbuild?

Build yes, release can stay the same. Changes are minimal, though.

I gave it a try with tsup (esbuild) and it works nicely. But I hit another wall with `@mui/icons-material` and remix.

Welp

smeng9:
As long as we do not bring back the regression here https://github.com/marmelab/react-admin/issues/7580 that bundles everything using tsup which makes it hard to debug and do module federation it should be fine.

smeng9:
Hi @otaviosoares 

I have checked how the team did using tsup in 4.0.x, Seems they have tried to mark react-hook-form as external https://github.com/marmelab/react-admin/blob/59f8e3dfd20b3ee9c6958068d091894a73feef63/packages/ra-core/package.json#L23C19-L23C50 so it can avoid dual package hazard.

djhi:
tsup won't do. We can't have TS source maps with it

otaviosoares:
Added a few changes to address the points you raised:



> As long as we do not bring back the regression here #7580 that bundles everything using tsup which makes it hard to debug and do module federation it should be fine.

Disabled minifying and bundling in tsup. Using a glob as entrypoint.



> tsup won't do. We can't have TS source maps with it

Using tsup's onSuccess + tsc emitDeclarationOnly to generate source maps.


@smen9 react-hook-form should be picked up as external automatically since it's listed as a dependency / peerDependency. I'm going to try to reproduce the scenario where it's being bundled.

djhi:
Is this still relevant/needed? When trying to run the tests locally, I get the following error:
> error TS5095: Option 'bundler' can only be used when 'module' is set to 'es2015' or later.

fzaninotto:
No news for some time, plus the PR is outdated, so I'm closing it. Feel free to comment further if you think this is still relevant.

PR #9460, opened 2023-11-20
[TypeScript] Fix useRecordContext may return undefined

groomain:
Follow https://github.com/marmelab/react-admin/pull/7498

Because of the type in useContext, the return type of useRecordContext will never be undefined.

But the doc in https://marmelab.com/react-admin/useRecordContext.html#typescript
mention that useRecordContext can be undefined.


djhi (approved):
Indeed, thanks!

PR #9466, opened 2023-11-22
Use resource definition to set default `<Datagrid rowClick>` value

slax57:
`<Datagrid>` should be smarter than having `rowClick` default to `false`.
It can use the resource definition to look for an existing Show or Edit view, and default to those instead.
This PR changes the default value to be a bit smarter.

### Minor Breaking Change

`<Datagrid>` will now make the rows clickable as soon as a Show or Edit view is declared on the resource.

If you previously relied on the fact that the rows were not clickable by default, you now need to explicitly disable the `rowClick` feature:

```diff
-<Datagrid>
+<Datagrid rowClick={false}>
   ...
</Datagrid>
```

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -166,7 +166,7 @@ import { List, Datagrid, TextField, EmailField } from "react-admin";
 
 export const UserList = () => (
     <List>
-        <Datagrid rowClick="edit">
+        <Datagrid>
             <TextField source="id" />
             <TextField source="name" />
             <TextField source="username" />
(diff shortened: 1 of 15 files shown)


docs/Tutorial.md
```diff
@@ -604,7 +592,7 @@ But now that the `users` resource has a `show` view, you can also link to it fro
 // in src/posts.tsx
 export const PostList = () => (
     <List>
-        <Datagrid rowClick="edit">
+        <Datagrid>
 -           <ReferenceField source="userId" reference="users" />
 +           <ReferenceField source="userId" reference="users" link="show" />
```
djhi, line 597: Not for this PR but this should probably be automatic as well (detection of the available views with the same priorities)

packages/ra-ui-materialui/src/list/ListGuesser.tsx
```diff
@@ -93,7 +91,7 @@ const ListViewGuesser = (
             );
             const inferredChild = new InferredElement(
                 listFieldTypes.table,
-                { hasEdit, hasShow },
```
djhi, line 96: Why ?

PR #9473, opened 2023-11-24
Upgrade react-query to v5, react 18 and yarn v4

djhi:
## Problems

- React-query v3 isn't compatible with React 18
- We are locked on react-query v3, because the library was renamed to tanstack query after that, and upgrading would be a breaking change
- No new features are pushed to react-query v3.

## Solution

- [x] Upgrade to tanstack query v5.
- [x] Upgrade React to v18
- [x] Upgrade Yarn (for internal use) to v4
- [x] use isPending for mutations
- [x] Change signature of auth provider hooks so that they always return `data`
- [x] Stabilize tests for react 18
- [x] Update demos

### React 18

React-admin v5 uses React 18. If you use react-admin as a library in your own application, you'll have to upgrade to React 18 as well.

The React team has published a [migration guide](https://react.dev/blog/2022/03/08/react-18-upgrade-guide) to help you upgrade. On most projects, this should be a matter of updating the root file of your application:

```diff
-import { render } from 'react-dom';
-const container = document.getElementById('app');
-render(<App tab="home" />, container);
+import { createRoot } from 'react-dom/client';
+const container = document.getElementById('app');
+const root = createRoot(container); // createRoot(container!) if you use TypeScript
+root.render(<App tab="home" />);
```

React 18 adds out-of-the-box performance improvements by doing more batching by default. 

### Drop support for IE11

React-admin v5 uses React 18, which dropped support for Internet Explorer. If you need to support IE11, you'll have to stay on react-admin v4.

### Rename `isLoading` to `isPending` in mutations

In the return state of data provider mutation hooks, `isLoading` has been renamed to `isPending` to be consistent with the same change in react-query.

The following hooks are impacted by this change:

- `useCreate`
- `useDelete`
- `useDeleteMany`
- `useUpdate`
- `useUpdateMany`
- `useDeleteWithUndoController`
- `useDeleteWithConfirmController`

If you use these hooks, replace `isLoading` by `isPending`:

```diff
import { useUpdate, useRecordContext, Button } from 'react-admin';

const ApproveButton = () => {
    const record = useRecordContext();
-    const [approve, { isLoading }] = useUpdate('comments', { id: record.id, data: { isApproved: true }, previousData: record });
+    const [approve, { isPending }] = useUpdate('comments', { id: record.id, data: { isApproved: true }, previousData: record });
-    return <Button label="Approve" onClick={() => approve()} disabled={isLoading} />;
+    return <Button label="Approve" onClick={() => approve()} disabled={isPending} />;
};
```

### Use `data` in Authentication hooks return value

Hooks that call the `authProvider` now return the `data` property of the result instead of a variable named after the hook.

The following hooks are impacted by this change:

- `useAuthState`
- `useIdentity`
- `usePermissions`

If you use these hooks, replace the variable name by `data`:

```diff
-const { isLoading, authenticated }       = useAuthState();
+const { isLoading, data: authenticated } = useAuthState();

-const { isLoading, identity }       = useIdentity();
+const { isLoading, data: identity } = useIdentity();

-const { isLoading, permissions }       = usePermissions();
+const { isLoading, data: permissions } = usePermissions();
```

### Use `@tanstack/react-query` instead of `react-query`

React-admin now uses `react-query` v5 instead of v3. The library name has changed to `@tanstack/react-query` (but it's almost the same API).
 
If you used `react-query` directly in your code, you'll have to update it, following their migration guides:

- [From react-query v3 to @transtack/query v4](https://tanstack.com/query/v5/docs/react/guides/migrating-to-react-query-4) 
- [From @transtack/query v4 to @transtack/query v5](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5).

Here is a focus of the most important changes. 

The package has been renamed to `@tanstack/react-query` so you'll have to change your imports:

```diff
-import { useQuery } from 'react-query';
+import { useQuery } from '@tanstack/react-query';
```

All react-query hooks, `queryClient` and `queryCache` methods now accept a single object argument:

```diff
- useQuery(key, fn, options)
+ useQuery({ queryKey, queryFn, ...options })
- useInfiniteQuery(key, fn, options)
+ useInfiniteQuery({ queryKey, queryFn, ...options })
- useMutation(fn, options)
+ useMutation({ mutationFn, ...options })
- useIsFetching(key, filters)
+ useIsFetching({ queryKey, ...filters })
- useIsMutating(key, filters)
+ useIsMutating({ mutationKey, ...filters })

- queryClient.isFetching(key, filters)
+ queryClient.isFetching({ queryKey, ...filters })
- queryClient.ensureQueryData(key, filters)
+ queryClient.ensureQueryData({ queryKey, ...filters })
- queryClient.getQueriesData(key, filters)
+ queryClient.getQueriesData({ queryKey, ...filters })
- queryClient.setQueriesData(key, updater, filters, options)
+ queryClient.setQueriesData({ queryKey, ...filters }, updater, options)
- queryClient.removeQueries(key, filters)
+ queryClient.removeQueries({ queryKey, ...filters })
- queryClient.resetQueries(key, filters, options)
+ queryClient.resetQueries({ queryKey, ...filters }, options)
- queryClient.cancelQueries(key, filters, options)
+ queryClient.cancelQueries({ queryKey, ...filters }, options)
- queryClient.invalidateQueries(key, filters, options)
+ queryClient.invalidateQueries({ queryKey, ...filters }, options)
- queryClient.refetchQueries(key, filters, options)
+ queryClient.refetchQueries({ queryKey, ...filters }, options)
- queryClient.fetchQuery(key, fn, options)
+ queryClient.fetchQuery({ queryKey, queryFn, ...options })
- queryClient.prefetchQuery(key, fn, options)
+ queryClient.prefetchQuery({ queryKey, queryFn, ...options })
- queryClient.fetchInfiniteQuery(key, fn, options)
+ queryClient.fetchInfiniteQuery({ queryKey, queryFn, ...options })
- queryClient.prefetchInfiniteQuery(key, fn, options)
+ queryClient.prefetchInfiniteQuery({ queryKey, queryFn, ...options })

- queryCache.find(key, filters)
+ queryCache.find({ queryKey, ...filters })
- queryCache.findAll(key, filters)
+ queryCache.findAll({ queryKey, ...filters })
```

### Removed deprecated hooks

The following deprecated hooks have been removed

- `usePermissionsOptimized`. Use `usePermissions` instead.


packages/ra-core/src/auth/useAuthState.ts
```diff
@@ -52,43 +52,64 @@ const emptyParams = {};
...
+                    result.error && result.error.message === false;
+                !shouldSkipNotify &&
+                    notify(
+                        getErrorMessage(
+                            result.error,
+                            'ra.auth.auth_check_error'
+                        ),
+                        {
+                            type: 'error',
+                        }
```
fzaninotto, line 108: Not sure if it's prettier fault, but the previous code was more compact (and therefore easier to read).

packages/ra-core/src/auth/useAuthenticated.spec.tsx
```diff
@@ -146,9 +146,14 @@ describe('useAuthenticated', () => {
...
-        await waitFor(() => {
-            expect(authProvider.logout).toHaveBeenCalledTimes(1);
-        });
+        await waitFor(
+            () => {
+                expect(authProvider.logout).toHaveBeenCalledTimes(1);
+            },
+            {
+                timeout: 4000,
+            }
```
fzaninotto, line 155: please compactify on 1 line

packages/ra-core/src/auth/useHandleAuthCallback.spec.tsx
```diff
@@ -49,34 +40,41 @@ const authProvider: AuthProvider = {
...
 describe('useHandleAuthCallback', () => {
     afterEach(() => {
-        redirect.mockClear();
+        jest.clearAllMocks();
     });
 
     it('should redirect to the home route by default when the callback was successfully handled', async () => {
-        const history = createMemoryHistory({ initialEntries: ['/'] });
+        const history = createMemoryHistory({
+            initialEntries: ['/auth-callback'],
```
fzaninotto, line 53: why did you change that?

packages/ra-core/src/auth/useHandleAuthCallback.spec.tsx
```diff
@@ -49,34 +40,41 @@ const authProvider: AuthProvider = {
...
-                        <TestComponent />
+                    <QueryClientProvider client={new QueryClient()}>
+                        <Routes>
+                            <Route path="/" element={<div>Home</div>} />
+                            <Route path="/test" element={<div>Test</div>} />
+                            <Route
+                                path="/auth-callback"
+                                element={<TestComponent />}
+                            />
+                        </Routes>
```
fzaninotto, line 66: I'm confused, this isn't the same test. Why do you need to put the tested component in routes?

packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -163,7 +163,7 @@ export const useReferenceManyFieldController = <
         },
         {
             enabled: get(record, source) != null,
-            keepPreviousData: true,
```
fzaninotto, line 166: I searched the reason for this change, so I put it here for reference: https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function

Note that our docs mention `keepPreviousData` here and there, so its needs to be updated, too.

fzaninotto (changes requested):
Review status: 25%

fzaninotto:
There are mentions of `'react-query'` in the doc that you must change, too:
image

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     - name: Use Node.js LTS
       uses: actions/setup-node@v1
       with:
-        node-version: '16.x'
+        node-version: '18.x'
```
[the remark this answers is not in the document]
djhi: Not for react-admin users

cypress/e2e/mobile.cy.js
```diff
@@ -9,12 +9,12 @@ describe('Mobile UI', () => {
     });
 
     describe('Infinite Scroll', () => {
-        it.only('should load more items when scrolling to the bottom of the page', () => {
+        it('should load more items when scrolling to the bottom of the page', () => {
             ListPagePosts.navigate();
+            cy.contains('Sint dignissimos in architecto aut');
             cy.contains('Sed quo et et fugiat modi').should('not.exist');
-            cy.scrollTo('bottom');
```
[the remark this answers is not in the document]
djhi: See L17. scrollToBottom was unreliable so instead I scroll to the last element

packages/ra-core/src/auth/Authenticated.spec.tsx
```diff
@@ -38,7 +38,7 @@ describe('<Authenticated>', () => {
         const authProvider = {
             login: jest.fn().mockResolvedValue(''),
             logout: jest.fn().mockResolvedValue(''),
-            checkAuth: jest.fn().mockRejectedValue(undefined),
+            checkAuth: jest.fn().mockRejectedValue({}),
```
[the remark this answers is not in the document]
djhi: Kind of. See https://tanstack.com/query/latest/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries

examples/demo/src/index.tsx
```diff
@@ -1,8 +1,11 @@
-import 'proxy-polyfill';
```
fzaninotto, line 1: this one should still be required?

examples/demo/src/authProvider.ts
```diff
@@ -13,7 +13,7 @@ const authProvider: AuthProvider = {
     checkError: () => Promise.resolve(),
     checkAuth: () =>
         localStorage.getItem('username') ? Promise.resolve() : Promise.reject(),
-    getPermissions: () => Promise.resolve(),
+    getPermissions: () => Promise.resolve([]),
```
fzaninotto, line 16: Why do we need it? getPermissions should be able to return undefined

examples/crm/vite.config.ts
```diff
@@ -46,7 +46,7 @@ export default defineConfig({
         preserveSymlinks: true,
         alias: [
             // allow profiling in production
-            { find: 'react-dom', replacement: 'react-dom/profiling' },
+            // { find: 'react-dom', replacement: 'react-dom/profiling' },
```
fzaninotto, line 49: commented out code

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -35,40 +35,51 @@ const emptyParams = {};
  *         }
  *     };
  */
-const usePermissions = <Permissions = any, Error = any>(
```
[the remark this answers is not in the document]
djhi: Are you reffering to the `Error` type parameter?

packages/ra-core/src/auth/addRefreshAuthToAuthProvider.spec.ts
```diff
@@ -8,8 +8,8 @@ describe('addRefreshAuthToAuthProvider', () => {
         logout: jest.fn(),
         checkAuth: jest.fn(),
         checkError: jest.fn(),
-        getIdentity: jest.fn(),
-        getPermissions: jest.fn(),
+        getIdentity: jest.fn().mockResolvedValue({}),
```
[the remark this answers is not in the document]
djhi: See https://tanstack.com/query/latest/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries

packages/ra-core/src/auth/useAuthState.ts
djhi: prettier

packages/ra-core/src/auth/useHandleAuthCallback.spec.tsx
```diff
@@ -49,34 +40,41 @@ const authProvider: AuthProvider = {
         return Promise.resolve();
     },
     getPermissions: () => Promise.reject('not authenticated'),
-    handleCallback: () => Promise.resolve(),
+    handleCallback: () => Promise.resolve({}),
```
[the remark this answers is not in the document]
djhi: Same as the others

examples/demo/src/index.tsx
```diff
@@ -1,8 +1,11 @@
...
-
-import * as React from 'react';
```
fzaninotto, line 3: I'm surprised that you don't need this anymore

examples/demo/vite.config.ts
```diff
@@ -46,7 +46,7 @@ export default defineConfig({
         preserveSymlinks: true,
         alias: [
             // allow profiling in production
-            { find: 'react-dom', replacement: 'react-dom/profiling' },
+            // { find: 'react-dom', replacement: 'react-dom/profiling' },
```
fzaninotto, line 49: commented out code

examples/simple/src/Layout.tsx
```diff
@@ -12,9 +12,11 @@ const MyAppBar = () => (
 export default props => (
     <>
         <Layout {...props} appBar={MyAppBar} />
-        <ReactQueryDevtools
-            initialIsOpen={false}
-            toggleButtonProps={{ style: { width: 20, height: 30 } }}
-        />
+        {process.env.NODE_ENV === 'development' ? (
```
fzaninotto, line 15: > By default, React Query Devtools are only included in bundles when process.env.NODE_ENV === 'development', so you don't need to worry about excluding them during a production build.

https://tanstack.com/query/v5/docs/react/devtools

examples/simple/src/authProvider.tsx
```diff
@@ -47,17 +47,17 @@ export default {
     },
     checkError: ({ status }) => {
         return status === 401 || status === 403
-            ? Promise.reject()
+            ? Promise.reject('')
```
fzaninotto, line 50: This change shouldn't be necessary

examples/simple/src/dataProvider.tsx
```diff
@@ -17,6 +18,8 @@ const dataProvider = withLifecycleCallbacks(fakeRestProvider(data, true), [
             await dp.deleteMany('comments', {
                 ids: comments.map(comment => comment.id),
             });
+            // The queryClient would be unaware of the deleted comments without this.
+            queryClient.invalidateQueries({ queryKey: ['comments'] });
```
fzaninotto, line 22: the `useDeleteMany` should take care of this, otherwise that's a bug

examples/simple/src/index.tsx
```diff
@@ -14,13 +14,18 @@ import Layout from './Layout';
...
+const container = document.getElementById('root');
+const root = createRoot(container);
+
+root.render(
     <React.StrictMode>
         <Admin
             authProvider={authProvider}
             dataProvider={dataProvider}
             i18nProvider={i18nProvider}
+            queryClient={queryClient}
```
fzaninotto, line 28: As it's the standard query client, I don't understand why you need to specify it.

examples/simple/src/queryClient.ts
```diff
@@ -0,0 +1,3 @@
+import { QueryClient } from '@tanstack/react-query';
```
fzaninotto, line 1: useless custom file IMHO

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -100,36 +100,36 @@ export const useCreate = <
...
-        }
-    );
+            if (options.onSuccess) {
+                options.onSuccess(data, variables, context);
+            }
+            // call-time success callback is executed by react-query
+        },
+    });
 
     const create = (
```
fzaninotto, line 134: you need to add calltime/runtime side effect discrimination

packages/ra-core/src/dataProvider/useDeleteMany.ts
```diff
@@ -210,101 +215,85 @@ export const useDeleteMany = <
...
+                    data,
+                    error,
+                    variables,
+                    context
+                );
+            }
+        },
+    });
 
     const mutate = async (
```
fzaninotto, line 298: you need to add calltime/runtime side effect discrimination

fzaninotto (changes requested):
Review status: 50%

packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -42,63 +46,42 @@ const defaultQueryParams = {
  * }
  */
 export const useGetIdentity = (
```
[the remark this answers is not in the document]
djhi: Re added to avoid an unnecessary BC

packages/ra-core/src/auth/usePermissions.ts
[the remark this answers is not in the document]
djhi: Re added to avoid an unnecessary BC

packages/ra-core/src/auth/useAuthenticated.spec.tsx
djhi: prettier. I could add an `prettier-ignore` but that will be worse IMO

examples/demo/src/index.tsx
djhi: https://caniuse.com/proxy

examples/demo/src/authProvider.ts
djhi: Then we must handle the undefined case in the hook

djhi:
> There are mentions of `'react-query'` in the doc that you must change, too: image

As discussed, will be done in a dedicated PR

examples/crm/package.json
```diff
@@ -4,32 +4,32 @@
     "private": true,
     "dependencies": {
         "@hello-pangea/dnd": "^16.3.0",
-        "@mui/icons-material": "^5.0.1",
-        "@mui/material": "^5.0.2",
+        "@mui/icons-material": "^5.14.18",
```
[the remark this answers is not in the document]
djhi: It isn't, I was trying to get rid of some warnings in the console. I can revert if you prefer

packages/ra-core/src/auth/Authenticated.spec.tsx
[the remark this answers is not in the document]
djhi: Or we can handle it at the hook level

packages/ra-core/src/auth/useHandleAuthCallback.spec.tsx
djhi: Jest mocks were unreliable. The new test actually tests the same behavior as before

packages/ra-core/src/dataProvider/useGetMany.spec.tsx
```diff
@@ -325,14 +349,6 @@ describe('useGetMany', () => {
...
         );
         expect(hookValue.mock.calls[3][0]).toStrictEqual(
-            expect.objectContaining({
-                data: undefined,
-                isError: false,
-                isFetching: true,
-                isLoading: true,
-            })
-        );
-        expect(hookValue.mock.calls[4][0]).toStrictEqual(
```
fzaninotto, line 335: Why do we have one less state?

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -77,39 +79,61 @@ export const useGetMany = <RecordType extends RaRecord = any>(
...
+            } else {
+                return records as RecordType[];
+            }
+        },
+        retry: false,
+        ...queryOptions,
+    });
+
+    useEffect(() => {
+        if (result.data != null) {
```
fzaninotto, line 106: why two imbricated ifs instead of just one with `&`? Same for other hooks.

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -97,50 +107,58 @@ export const useInfiniteGetList = <RecordType extends RaRecord = any>(
...
+            return lastLoadedPage.pageParam === 1
+                ? undefined
+                : lastLoadedPage.pageParam - 1;
+        },
+    });
+
+    useEffect(() => {
+        if (result.data != null) {
+            // optimistically populate the getOne cache
+            result.data.pages.forEach(page => {
```
fzaninotto, line 142: I'm thinking we should put the same limit as in `useGetList` to avoid slowing down the UI too much with long lists.

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -132,22 +133,26 @@ export const useUpdateMany = <
             data?: RecordType[];
         };
 
-        ids.forEach(id =>
+        ids.forEach(id => {
             queryClient.setQueryData(
                 [resource, 'getOne', { id: String(id), meta }],
```
fzaninotto, line 138: Must be converted to object with queryKey

packages/ra-core/src/form/Form.tsx
```diff
@@ -42,7 +46,9 @@ import { useAugmentedForm } from './useAugmentedForm';
  *
  * @link https://react-hook-form.com/docs/useformcontext
  */
-export const Form = (props: FormProps) => {
+export const Form = <RecordType extends RaRecord = any>(
```
fzaninotto, line 49: Form can be used in a Create page, where there is no `id`. Therefore it's not necessarily operating on a `RaRecord`

packages/ra-core/src/form/useAugmentedForm.ts
```diff
@@ -30,7 +30,9 @@ import { sanitizeEmptyValues as sanitizeValues } from './sanitizeEmptyValues';
  * - notification on invalid form
  * - stop form submission event propagation
  */
-export const useAugmentedForm = (props: UseAugmentedFormProps) => {
+export const useAugmentedForm = <RecordType extends RaRecord = any>(
```
fzaninotto, line 33: Same, can be an object without `id`

packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
```diff
@@ -20,7 +20,7 @@ import {
 import { Confirm } from '../layout';
 import { Button, ButtonProps } from './Button';
 import { BulkActionProps } from '../types';
-import { UseMutationOptions } from 'react-query';
+import { UseMutationOptions } from '@tanstack/react-query';
```
fzaninotto, line 23: I'm not even sure we still need this type

packages/ra-ui-materialui/src/button/UpdateWithConfirmButton.spec.tsx
```diff
@@ -69,15 +69,15 @@ describe('<UpdateWithConfirmButton />', () => {
             </Toolbar>
         );
         render(
-            <ThemeProvider theme={theme}>
-                <CoreAdminContext dataProvider={dataProvider}>
+            <CoreAdminContext dataProvider={dataProvider}>
+                <ThemeProvider theme={theme}>
```
fzaninotto, line 73: CoreAdminContext+ThemeProvider = AdminContext

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
```diff
@@ -47,7 +47,7 @@ describe('<ReferenceArrayField />', () => {
         );
 
         await new Promise(resolve => setTimeout(resolve, 1001));
-        expect(queryAllByRole('progressbar')).toHaveLength(1);
```
fzaninotto, line 50: linter warning

packages/ra-ui-materialui/src/input/ReferenceInput.spec.tsx
```diff
@@ -181,23 +181,32 @@ describe('<ReferenceInput />', () => {
...
         };
         render(<Basic dataProvider={dataProvider} />);
-        await screen.findByDisplayValue('Leo Tolstoy');
-        const input = screen.getByLabelText('Author') as HTMLInputElement;
+        const input = (await screen.findByDisplayValue(
+            'Leo Tolstoy'
+        )) as HTMLInputElement;
         input.focus();
         screen.getByLabelText('Clear value').click();
+        await screen.findByDisplayValue('');
```
fzaninotto, line 189: what's that?

packages/ra-ui-materialui/src/input/ReferenceInput.spec.tsx
```diff
@@ -181,23 +181,32 @@ describe('<ReferenceInput />', () => {
...
+                    expect.objectContaining({
+                        data: expect.objectContaining({
+                            author: null,
+                        }),
+                    })
+                );
+            },
+            {
+                timeout: 4000,
+            }
```
fzaninotto, line 209: please compactify

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -210,11 +209,14 @@ describe('<FilterButton />', () => {
...
-                    screen.getAllByTestId('CheckBoxOutlineBlankIcon')
-                ).toHaveLength(2);
-            });
+            await waitFor(
+                () => {
+                    expect(screen.getAllByRole('checkbox')).toHaveLength(2);
+                },
+                {
+                    timeout: 4000,
+                }
```
fzaninotto, line 218: please compactify

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
```diff
@@ -23,6 +23,8 @@ export const Pagination: FC<PaginationProps> = memo(props => {
         rowsPerPageOptions = DefaultRowsPerPageOptions,
         actions,
         limit = null,
+        hasNextPage: _hasNextPage,
+        hasPreviousPage: _hasPreviousPage,
```
fzaninotto, line 27: We should make the change in `useListController`'s `injectedProps` instead, the bug is earlier in the chain.

packages/ra-ui-materialui/src/list/pagination/PaginationActions.tsx
```diff
@@ -13,6 +13,8 @@ export const PaginationActions: FC<PaginationActionsProps> = memo(props => {
         onPageChange,
         size = 'small',
         className,
+        // @ts-ignore
+        slotProps,
```
fzaninotto, line 17: No, add it to the rest props sanitizer instead (cf https://github.com/marmelab/react-admin/pull/9474/files)

fzaninotto (changes requested):
Wow, this is impressive! Kudos for this PR.

examples/demo/src/index.tsx
fzaninotto: Now I understand that upgrading to React 18 means dropping support for IE

https://react.dev/blog/2022/03/08/react-18-upgrade-guide#dropping-support-for-internet-explorer

We should make it clear in the upgrade guide. Also, there are many ie11 polyfills in our codebase, this PR should remove them.

packages/ra-core/src/auth/Authenticated.spec.tsx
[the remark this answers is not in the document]
djhi: handled at the hook level

packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
djhi: Documentation will be updated in a dedicated PR

examples/demo/src/index.tsx
djhi: Will do in a dedicated PR

examples/crm/vite.config.ts
djhi: Yes, haven't yet investigated how we can add profiling back. I'll address this issue in another PR

examples/simple/src/dataProvider.tsx
djhi: This is a lifecycle callback. The `useDelete` hook was called on the `post` resource. It doesn't even know about `comments`

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -133,8 +134,14 @@ export const useDelete = <
             { updatedAt }
         );
         queryClient.setQueriesData(
-            [resource, 'getInfiniteList'],
-            (res: UseInfiniteQueryResult<GetInfiniteListResult>['data']) => {
+            {
+                queryKey: [resource, 'getInfiniteList'],
+            },
```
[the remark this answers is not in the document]
djhi: prettier

packages/ra-core/src/dataProvider/useGetMany.spec.tsx
djhi: Misteries of the new react-query or react

packages/ra-core/src/dataProvider/useGetMany.ts
djhi: What imbricated if?

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
djhi: Another PR

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -79,7 +79,42 @@ const useDeleteWithConfirmController = <RecordType extends RaRecord = any>(
     const notify = useNotify();
     const unselect = useUnselect(resource);
     const redirect = useRedirect();
-    const [deleteOne, { isLoading }] = useDelete<RecordType>();
+    const [deleteOne, { isPending }] = useDelete<RecordType>(
+        resource,
+        { id: record?.id },
+        {
+            onSuccess: () => {
```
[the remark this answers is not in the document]
djhi: Handled in the dataProvider hooks instead

packages/ra-core/src/controller/button/useDeleteWithUndoController.tsx
```diff
@@ -57,7 +57,39 @@ const useDeleteWithUndoController = <RecordType extends RaRecord = any>(
     const notify = useNotify();
     const unselect = useUnselect(resource);
     const redirect = useRedirect();
-    const [deleteOne, { isLoading }] = useDelete<RecordType>();
+    const [deleteOne, { isPending }] = useDelete<RecordType>(
+        resource,
+        { id: record?.id },
+        {
+            onSuccess: () => {
```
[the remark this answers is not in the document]
djhi: Handled in the dataProvider hooks instead

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -94,7 +98,7 @@ export const useCreateController = <
                 onError: onErrorFromSave,
                 transform: transformFromSave,
                 meta: metaFromSave,
-            } = {}
+            } = {} as SaveHandlerCallbacks
```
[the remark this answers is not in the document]
djhi: Handled in the dataProvider hook. Moved the side effects at declaration time

packages/ra-core/src/dataProvider/useCreate.ts
djhi: Added

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -198,101 +205,85 @@ export const useDelete = <
...
+                    data,
+                    error,
+                    variables,
+                    context
+                );
+            }
+        },
+    });
 
     const mutate = async (
```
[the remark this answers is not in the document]
djhi: Added

packages/ra-core/src/dataProvider/useDeleteMany.ts
djhi: Added

examples/demo/src/authProvider.ts
fzaninotto: Then you need to revert this change

examples/demo/src/index.tsx
fzaninotto: Bump. The default root for a React project in Vite imports `React` for JSX to work. Why is it not necessary here?

packages/ra-core/src/form/Form.tsx
fzaninotto: Bump

packages/ra-core/src/form/useAugmentedForm.ts
fzaninotto: Bump

packages/ra-ui-materialui/src/button/UpdateWithConfirmButton.spec.tsx
fzaninotto: Bump

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
fzaninotto: Bump

examples/demo/src/index.tsx
djhi: Well, just run the demo, you'll see it works both in development and production

packages/ra-core/src/auth/usePermissions.spec.tsx
```diff
@@ -15,8 +15,8 @@ const UsePermissions = ({ children }: any) => {
 
 const stateInpector = state => (
     <div>
-        <span>{state.isLoading && 'LOADING'}</span>
-        {state.permissions && <span>PERMISSIONS: {state.permissions}</span>}
+        <span>{state.isPending && 'LOADING'}</span>
+        {state.data && <span>PERMISSIONS: {state.data}</span>}
```
fzaninotto, line 19:
```suggestion
        {state.permissions && <span>PERMISSIONS: {state.permissions}</span>}
```

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -13,62 +17,89 @@ const emptyParams = {};
...
+        },
+        ...queryOptions,
+    });
+
+    useEffect(() => {
+        if (result.data != null && onSuccess) {
+            onSuccess(result.data);
         }
-    );
+    }, [onSuccess, result.data]);
```
fzaninotto, line 66: memoize onSuccess with useEvent

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -13,62 +17,89 @@ const emptyParams = {};
...
+        if (result.error != null) {
+            if (onError) {
+                return onError(result.error);
+            }
+            if (process.env.NODE_ENV === 'development') {
+                console.error(result.error);
+            }
+            logoutIfAccessDenied(result.error);
+        }
+    }, [logoutIfAccessDenied, onError, result.error]);
```
fzaninotto, line 78: memoize onError with useEvent

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -13,62 +17,89 @@ const emptyParams = {};
...
+
+    useEffect(() => {
+        if (result.data != null && onSuccess) {
+            onSuccess(result.data);
         }
-    );
+    }, [onSuccess, result.data]);
+
+    useEffect(() => {
+        if (result.error != null) {
```
fzaninotto, line 69: leave early

fzaninotto (changes requested):
Second review 13% done (reviewed ra-core/auth)

packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -66,13 +67,17 @@ export const useGetList = <RecordType extends RaRecord = any>(
     } = params;
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
+    const { onError = noop, onSuccess = noop, ...queryOptions } = options;
```
fzaninotto, line 70: fun fact: useQuery v3 used to support an `onSettled` function, too. So we need to reimplement it here, too, using effects.

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -52,23 +54,26 @@ import { useDataProvider } from './useDataProvider';
...
     resource: string,
     params: Partial<GetManyParams> = {},
-    options?: UseQueryOptions<RecordType[], Error>
+    options: UseGetManyOptions<RecordType> = {}
 ): UseGetManyHookValue<RecordType> => {
     const { ids, meta } = params;
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
     const queryCache = queryClient.getQueryCache();
+    const { onError = noop, onSuccess = noop, ...queryOptions } = options;
```
fzaninotto, line 63: don't forget onSettled

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -77,39 +82,63 @@ export const useGetMany = <RecordType extends RaRecord = any>(
...
+            } else {
+                return records as RecordType[];
+            }
+        },
+        retry: false,
+        ...queryOptions,
+    });
+
+    useEffect(() => {
+        if (result.data != null) {
```
fzaninotto, line 109: leave early

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -77,39 +82,63 @@ export const useGetMany = <RecordType extends RaRecord = any>(
...
+            // optimistically populate the getOne cache
+            result.data.forEach(record => {
+                queryClient.setQueryData(
+                    [resource, 'getOne', { id: String(record.id), meta }],
+                    oldRecord => oldRecord ?? record
+                );
+            });
+
+            // execute call-time onSuccess if provided
+            if (onSuccessEvent) {
```
fzaninotto, line 119: this is an `if (true) `

packages/ra-core/src/dataProvider/useGetMany.ts
```diff
@@ -77,39 +82,63 @@ export const useGetMany = <RecordType extends RaRecord = any>(
...
+                onSuccessEvent(result.data);
+            }
         }
-    );
+    }, [queryClient, meta, onSuccessEvent, resource, result.data]);
+
+    useEffect(() => {
+        if (result.error != null && onErrorEvent) {
```
fzaninotto, line 126: same

fzaninotto (changes requested):
Second review 21% done (half of ra-core/dataProvider)

.yarnrc.yml
```diff
@@ -1,11 +1,8 @@
 compressionLevel: 0
 
 enableGlobalCache: true
+enableHardenedMode: false
```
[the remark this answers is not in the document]
djhi: Issues with jest without it

packages/ra-core/src/auth/useAuthState.spec.tsx
```diff
@@ -9,23 +9,13 @@ const UseAuth = (authParams: any) => {
...
         <div>
-            <span>{state.isLoading && 'LOADING'}</span>
+            <span>{state.isPending && 'LOADING'}</span>
             <span>{state.authenticated && 'AUTHENTICATED'}</span>
         </div>
     );
 };
 
 describe('useAuthState', () => {
-    it('should return a loading state on mount', () => {
```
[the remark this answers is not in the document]
djhi: Because by using `placeholderData` the query is not in pending state anymore so this test isn't relevant. We could test the `isLoading` though but I fail to see the point

packages/ra-core/src/auth/useGetIdentity.spec.tsx
```diff
@@ -54,6 +54,6 @@ describe('useGetIdentity', () => {
             expect(screen.queryByText('Loading')).toBeNull();
         });
         expect(screen.queryByText(/Error/)).toBeNull();
-        expect(screen.queryByText('undefined')).not.toBeNull();
+        expect(screen.queryByText('null')).not.toBeNull();
```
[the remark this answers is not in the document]
djhi: That's because `useIdentity` never returns `undefined` (react-query requires queries to return something else that `undefined`)

packages/ra-core/src/auth/useGetIdentity.ts
```diff
@@ -41,64 +45,59 @@ const defaultQueryParams = {
...
     const authProvider = useAuthProvider();
+    const { onSuccess, onError, ...queryOptions } = options;
 
-    const result = useQuery(
-        ['auth', 'getIdentity'],
-        authProvider
-            ? () => authProvider.getIdentity()
-            : async () => defaultIdentity,
-        {
-            enabled: typeof authProvider?.getIdentity === 'function',
```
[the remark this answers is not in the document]
djhi: No, if `getIdentity` isn't defined, we return `null`

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -183,7 +184,7 @@ export interface UseDeleteWithConfirmControllerParams<
 
 export interface UseDeleteWithConfirmControllerReturn {
     open: boolean;
-    isLoading: boolean;
```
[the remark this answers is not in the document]
djhi: Added back but GH don't see it :man_shrugging:

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -145,19 +150,15 @@ const useDeleteWithConfirmController = <RecordType extends RaRecord = any>(
...
-            redirectTo,
             resource,
-            unselect,
         ]
     );
 
     return {
         open,
-        isLoading,
+        isPending,
```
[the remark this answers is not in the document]
djhi: Added back but GH don't see it 🤷‍♂️

packages/ra-core/src/controller/button/useDeleteWithUndoController.tsx
```diff
@@ -1,5 +1,5 @@
 import { useCallback, ReactEventHandler } from 'react';
-import { UseMutationOptions } from 'react-query';
+import { UseMutationOptions } from '@tanstack/react-query';
```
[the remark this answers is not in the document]
djhi: Done

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -409,7 +414,10 @@ export const useDelete = <
...
                     mutation.mutate(
                         { resource: callTimeResource, ...callTimeParams },
-                        { onSettled, onError }
+                        {
+                            onSettled: callTimeOnSettled,
+                            onError: callTimeOnError,
+                        }
                     );
                 }
             });
```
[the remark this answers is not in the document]
djhi: Added

packages/ra-core/src/dataProvider/useDeleteMany.ts
```diff
@@ -418,7 +423,10 @@ export const useDeleteMany = <
...
                     mutation.mutate(
                         { resource: callTimeResource, ...callTimeParams },
-                        { onSettled, onError }
+                        {
+                            onSettled: callTimeOnSettled,
+                            onError: callTimeOnError,
+                        }
                     );
                 }
             });
```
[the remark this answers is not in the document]
djhi: Added

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -79,7 +79,42 @@ const useDeleteWithConfirmController = <RecordType extends RaRecord = any>(
     const notify = useNotify();
     const unselect = useUnselect(resource);
     const redirect = useRedirect();
-    const [deleteOne, { isLoading }] = useDelete<RecordType>();
+    const [deleteOne, { isPending }] = useDelete<RecordType>(
+        resource,
+        { id: record?.id },
```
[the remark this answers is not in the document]
djhi: That was because it didn't matter as we specify them at call time, thus overriding them. I did that just to make TS happy as I needed to pass the options only

packages/ra-core/src/auth/useAuthState.ts
```diff
@@ -49,57 +49,102 @@ const emptyParams = {};
...
     );
 
+    useEffect(() => {
+        if (result.data == null) return;
+        onSuccessEvent(result.data);
+    }, [onSuccessEvent, result.data]);
+
+    useEffect(() => {
+        if (result.error == null) return;
+        onErrorEvent(result.error);
```
fzaninotto, line 122: You're passing the error here, but the callback isn't using it.

packages/ra-core/src/auth/useAuthState.ts
```diff
@@ -49,57 +49,102 @@ const emptyParams = {};
...
     return useMemo(() => {
         return {
+            ...result,
             // If the data is undefined and the query isn't loading anymore, it means the query failed.
             // In that case, we set authenticated to false unless there's no authProvider.
-            authenticated:
-                result.data ?? result.isLoading ? true : authProvider == null, // Optimistic
-            isLoading: result.isLoading,
-            error: result.error,
+            authenticated: result.data,
```
fzaninotto, line 130: This is wring: `authenticated` will always be `true`:

- while loading: the placeholder data is `true``
- on success, the value is `true`
- on error, you still return the `data`, which is the placeholder data, `true`

I think we definitely need a unit test for that case.

packages/ra-core/src/auth/useAuthState.spec.tsx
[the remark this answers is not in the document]
fzaninotto: useAuthState is designed to render anonymous content if not authenticated (cf https://marmelab.com/react-admin/useAuthState.html), so I think we still need this test, using `isLoading` and not `isPending`.

packages/ra-core/src/auth/useHandleAuthCallback.ts
```diff
@@ -21,36 +23,62 @@ export const useHandleAuthCallback = (
...
+    const onErrorEvent = useEvent(onError ?? noop);
+
+    useEffect(() => {
+        if (queryResult.error == null) return;
+        onErrorEvent(queryResult.error);
+    }, [onErrorEvent, queryResult.error]);
+
+    useEffect(() => {
+        if (queryResult.data === undefined) return;
+        onSuccessEvent(queryResult.data);
```
fzaninotto, line 64: you're passing ythe data to the callback, but the default success callback doesn't use it.

packages/ra-core/src/auth/useHandleAuthCallback.ts
```diff
@@ -21,36 +23,62 @@ export const useHandleAuthCallback = (
...
     );
+    const onErrorEvent = useEvent(onError ?? noop);
+
+    useEffect(() => {
+        if (queryResult.error == null) return;
+        onErrorEvent(queryResult.error);
+    }, [onErrorEvent, queryResult.error]);
+
+    useEffect(() => {
+        if (queryResult.data === undefined) return;
```
fzaninotto, line 63: I'm confused: sometimes you test success with `!== undefined`, and sometimes with `!= null`. It seems to me that the first one is correct, but why isn't it the same test everywhere?

packages/ra-core/src/auth/usePermissions.ts
```diff
@@ -13,62 +18,94 @@ const emptyParams = {};
...
+        ...queryOptions,
+    });
+
+    const onSuccessEvent = useEvent(onSuccess ?? noop);
+    const onErrorEvent = useEvent(
+        onError ??
+            (() => {
                 if (process.env.NODE_ENV === 'development') {
-                    console.error(error);
+                    console.error(result.error);
```
fzaninotto, line 68: use the error param instead of result.error

fzaninotto (changes requested):
A few more remarks on the ra-core/auth part

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -318,7 +316,7 @@ export const useDelete = <
         if (mode.current === 'pessimistic') {
             return mutation.mutate(
                 { resource: callTimeResource, ...callTimeParams },
-                { onSuccess, onSettled, onError }
+                callTimeOptions
```
fzaninotto, line 319: you're passing too many options to the underlying mutation (dataProvider.delete()  doesn't understand `mutationMode`)

packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -116,21 +121,35 @@ export const useGetManyAggregate = <RecordType extends RaRecord = any>(
...
+            // optimistically populate the getOne cache
+            (result.data ?? []).forEach(record => {
+                queryClient.setQueryData(
+                    [resource, 'getOne', { id: String(record.id), meta }],
+                    oldRecord => oldRecord ?? record
+                );
+            });
+
+            // execute call-time onSuccess if provided
+            if (onSuccessEvent) {
```
fzaninotto, line 140: if(true)

packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -116,21 +121,35 @@ export const useGetManyAggregate = <RecordType extends RaRecord = any>(
...
+                onSuccessEvent(result.data);
+            }
+        }
+    }, [queryClient, meta, onSuccessEvent, resource, result.data]);
+
+    useEffect(() => {
+        if (result.error != null && onErrorEvent) {
```
fzaninotto, line 147: if(true)

packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -68,15 +69,19 @@ import { useDataProvider } from './useDataProvider';
 export const useGetManyAggregate = <RecordType extends RaRecord = any>(
     resource: string,
     params: GetManyParams,
-    options: UseQueryOptions<RecordType[], Error> = {}
+    options: UseGetManyAggregateOptions<RecordType> = {}
 ): UseGetManyHookValue<RecordType> => {
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
     const queryCache = queryClient.getQueryCache();
+    const { onError = noop, onSuccess = noop, ...queryOptions } = options;
```
fzaninotto, line 77: missing onSettled

packages/ra-core/src/dataProvider/useGetManyReference.ts
```diff
@@ -104,19 +106,31 @@ export const useGetManyReference = <RecordType extends RaRecord = any>(
...
+                    [resource, 'getOne', { id: String(record.id), meta }],
+                    oldRecord => oldRecord ?? record
+                );
+            });
         }
-    );
+
+        // execute call-time onSuccess if provided
+        if (onSuccess) {
+            onSuccess(result.data);
```
fzaninotto, line 125: needs memoization

packages/ra-core/src/dataProvider/useGetManyReference.ts
```diff
@@ -104,19 +106,31 @@ export const useGetManyReference = <RecordType extends RaRecord = any>(
...
+        }
+    }, [queryClient, meta, onSuccess, resource, result.data]);
+
+    useEffect(() => {
+        if (result.error != null && onError) {
+            onError(result.error);
```
fzaninotto, line 131: needs memoization

packages/ra-core/src/dataProvider/useGetManyReference.ts
```diff
@@ -74,17 +74,19 @@ export const useGetManyReference = <RecordType extends RaRecord = any>(
     } = params;
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
+    const { onError, onSuccess, ...queryOptions } = options;
```
fzaninotto, line 77: missing onSettled

fzaninotto (changes requested):
A few missed point in the core/dataProvider files

packages/ra-core/src/auth/useAuthState.ts
djhi: The existing tests already cover that:

> * while loading: the placeholder data is `true`
> * on success, the value is true

`should return authenticated by default after a tick`: this is the default optimistic case where we consider users authenticated 

> * on error, you still return the `data`, which is the placeholder data, `true`

`should return an error after a tick if the auth fails`: in case of error, we check if `AUTHENTICATED` is `null` so `data` and `authenticated` do not contain the placeholder data anymore or this test would fail

What's missing?

packages/ra-core/src/auth/usePermissions.ts
djhi: Why?

packages/ra-core/src/controller/create/useCreateController.ts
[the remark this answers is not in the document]
fzaninotto: Just for my understanding, the side effects were previously set at call time, now they're set at definition time. Is there a reason for this change?

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
fzaninotto, line 50: linter warning

packages/ra-core/src/controller/create/useCreateController.ts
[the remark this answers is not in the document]
djhi: To avoid side effects not being executed on optimistic modes. Indeed, if the component is unmounted, call time side effects are not executed, only declaration time are. This was the case in react-query v3 and I don't understand how it could work before.

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
djhi: Fixed

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
fzaninotto: Bump

packages/ra-core/src/auth/useAuthState.spec.tsx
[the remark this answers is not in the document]
fzaninotto: We probably need a test for the `false` return value

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
djhi: Actually I can't reproduce the bug anymore. Reverting

packages/ra-core/src/dataProvider/useGetManyAggregate.ts
```diff
@@ -116,21 +127,36 @@ export const useGetManyAggregate = <RecordType extends RaRecord = any>(
...
+
+        // optimistically populate the getOne cache
+        (result.data ?? []).forEach(record => {
+            queryClient.setQueryData(
+                [resource, 'getOne', { id: String(record.id), meta }],
+                oldRecord => oldRecord ?? record
+            );
+        });
+
+        onSuccessEvent(result.data);
```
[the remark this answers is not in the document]
djhi: That was already the case with react-query v3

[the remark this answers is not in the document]
djhi: Indeed but that's actually more consistent with the other hooks such as `useGetList`. @fzaninotto any opinion?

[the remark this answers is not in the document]
fzaninotto: I agree with @djhi. There are a few uses cases for not prepopulating the cache (e.g. when the `getList()` and `getOne` records don't have the same shape), but we don't want to support these use cases and there are other ways to get around them (e.g. by removing the cache prepopulation in the custom onSuccess)

packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
```diff
@@ -184,7 +186,7 @@ export const CheckboxGroupInput: FunctionComponent<CheckboxGroupInputProps> = pr
         [allChoices, formOnChange, formOnBlur, optionValue, value]
     );
 
-    if (isLoading && (!allChoices || allChoices.length === 0)) {
+    if (isPending && (!allChoices || allChoices.length === 0)) {
```
slax57, line 189: Shouldn't we use `isPending || isLoading` instead of `isPending` for backwards compatibility?

(same question applies to other similar inputs)

slax57 (changes requested):
Review status: 70%

packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
djhi: `isLoading` implies `isPending`

djhi: And no, we want to follow the react query practices anyway

slax57: Right, I got confused by spending too much time on this, sorry about that :sweat_smile: 
Resuming the review...

packages/react-admin/README.md
```diff
@@ -9,7 +9,7 @@ A frontend Framework for building data-driven applications running in the browse
 ## Features
 
 * Adapts to any backend (REST, GraphQL, SOAP, etc.)
-* Powered by [Material UI](https://mui.com/material-ui/getting-started/), [react-query](https://react-query-v3.tanstack.com/), [react-hook-form](https://react-hook-form.com), [react-router](https://reacttraining.com/react-router/) and a few more
+* Powered by [Material UI](https://mui.com/material-ui/getting-started/), [react-query](https://react-query-v5.tanstack.com/), [react-hook-form](https://react-hook-form.com), [react-router](https://reacttraining.com/react-router/) and a few more
```
slax57, line 12:
```suggestion
* Powered by [Material UI](https://mui.com/material-ui/getting-started/), [react-query](https://tanstack.com/query/v5), [react-hook-form](https://react-hook-form.com), [react-router](https://reacttraining.com/react-router/) and a few more
```
(not a valid URL)

examples/simple/package.json
```diff
@@ -5,28 +5,30 @@
     "scripts": {
         "dev": "vite",
         "build": "vite build",
+        "typecheck": "tsc --noEmit",
```
slax57, line 8: Duplicate with 2 lines below. Is this intentional? :thinking:

slax57 (changes requested):
Review complete! Now to do some testing...

slax57 (approved):
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

Issue #9468, opened 2023-11-22
TabbedForm and TabbedShowLayout tab navigation is broken


silversonicaxel:
**What you were expecting:**
I've an `<Edit />` component View page to edit a record and I use it in combination with `<TabbedForm />` component.
I expect to switch among `<TabbedForm.Tab />`s and edit the fields of the record.

**What happened instead:**
The `<TabbedForm.Tab />`s beyond the first one are not reachable, instead first `<TabbedForm.Tab />` keeps being visible. What also happens in the url is that, every time I click a `<TabbedForm.Tab />`, the url keeps adding the index of the clicked `<TabbedForm.Tab />` to the end of the url. Application is not broken though, so no javascript error occurs.

**Steps to reproduce:**
The application is based on Nextjs and ReactAdmin, but it is private, so unsharable.

What I can say is that I've a CRUD scenario, with an Edit view based on Tabs and since the upgrade from version 4.14.3 to 4.16.0 happened, the Application has this behaviour.

**Related code:**
I can't really show examples, but an official react-admin demo example https://github.com/marmelab/react-admin/tree/master/examples/simple visible at this link
https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple shows also that current implementation has the same issue. To reproduce it in the example app do the following steps:
- Click on the left sidebar the menu item Tags
- Edit one tag, for example Sport
- Edit one item in the bottom grid.
- Click one of the tabs in the Edit view
- Tabs are not switching

**Other information:**
/

**Environment**

* React-admin version: 4.16.0
* Last version that did not exhibit the issue (if applicable): 4.14.3
* React version: 18.2.0
* Browser: all
* Stack trace (in case of a JS error): /

fzaninotto:
Yep, I saw that bug, too, only a bit later. This is a regression, we'll work on fixing it soon.

silversonicaxel:
As far as I understand, the `simple` example is used for e2e tests. Could something be improved there?

slax57:
Turns out this is due to 2 regressions introduced in **react-router v6.19.0** and **v6.20.0**.

I have opened an issue on their repo to request for a fix: https://github.com/remix-run/react-router/issues/11053

Until this is fixed, you will need to use version **v6.18.0** at most, which does not show the issue.

Thanks for the report!

fzaninotto:
[...]

PR #9476, opened 2023-11-28
[DNM] Fix tab navigation with react-router >= 6.19

slax57:
Fix https://github.com/marmelab/react-admin/issues/9468

### DIFF
--- a/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
+++ b/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
@@ -36,7 +36,7 @@ export const TabbedFormView = (props: TabbedFormViewProps): ReactElement => {
         ...rest
     } = props;
     const location = useLocation();
-    const resolvedPath = useResolvedPath('');
+    const resolvedPath = useResolvedPath('..');
     const resource = useResourceContext(props);
     const [tabValue, setTabValue] = useState(0);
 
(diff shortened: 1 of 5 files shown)


packages/ra-ui-materialui/src/form/TabbedFormView.tsx
```diff
@@ -36,7 +36,7 @@ export const TabbedFormView = (props: TabbedFormViewProps): ReactElement => {
         ...rest
     } = props;
     const location = useLocation();
-    const resolvedPath = useResolvedPath('');
+    const resolvedPath = useResolvedPath('.');
```
slax57, line 39: Unfortunately this part doesn't work:
- React-router 6.18 requires `useResolvedPath('.')`
- While React-router 6.20 requires `useResolvedPath('..')`

I can't find a way to make both versions happy with the same code...

I'm open to any suggestions if any, or, you may want to upvote https://github.com/remix-run/react-router/issues/11048#issuecomment-1829349584 and https://github.com/remix-run/react-router/issues/11052#issuecomment-1829467624 to request for a fix from react-router.

fzaninotto: I suggest we change the package.json to forbid recent react-router versions.

fzaninotto:
Superseded by #9480

PR #9478, opened 2023-11-28
[Doc] Fix snippets fails to render in JS

fzaninotto:
## Before

image

## After

image

### DIFF
--- a/docs/DataProviderWriting.md
+++ b/docs/DataProviderWriting.md
@@ -305,7 +305,7 @@ interface DeleteParams {
 interface DeleteResult {
     data: Record;
 }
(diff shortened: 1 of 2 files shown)


docs/DataProviderWriting.md
```diff
@@ -305,7 +305,7 @@ interface DeleteParams {
 interface DeleteResult {
     data: Record;
 }
-function delete(resource: string, params: DeleteParams): Promise<DeleteResult>
```
fzaninotto, line 308: `delete` is a reserved word, so there is no other way to make it compile

adguernier (approved):

slax57 (approved):

PR #9480, opened 2023-11-29
Fix TabbedForm and TabbedShowLayout tab navigation: limit react-router version to 6.18.0 as a workaround

slax57:
Workaround for https://github.com/marmelab/react-admin/issues/9468

Since it is still unclear whether or not the regressions brought in https://github.com/remix-run/react-router/issues/11052 and https://github.com/remix-run/react-router/issues/11048 will be reverted or not, we need to force the react-router version to be 6.18.0 at most.

Tested manually on an external project:
- By default npm resolves react-router and react-router-dom to the latest version compatible with the peer dependencies: 6.18.0
- Running `npm add --save react-router react-router-dom` still yields the 6.20.0 version, but displays a warning about unmet peer deps


slax57:
Back to WIP

fzaninotto (approved):

PR #9494, opened 2023-12-06
doc: improve EditableDatagrid doc

adguernier:
The OSS version of the EditableDatagrid is very poor.

This PR attemps to make it rich!


fzaninotto (approved):
Two minor remarks:

- This article could use more screenshots / screencasts
- the first example doesn't match the initial screencast

But that's already way better than the previous version!

slax57 (changes requested):

PR #9503, opened 2023-12-08
Fix `useTheme` may return `undefined`

### LINKED ISSUE #9502: useTheme returns undefined as value before using the ToggleThemeButton
**What you were expecting:**
Getting the theme value with the react-admin hook: useTheme().

**What happened instead:**
useTheme() hook returns undefined as value.

**Steps to reproduce:**
Go to this sandbox https://stackblitz.com/edit/github-m4as1b?file=src%2Fposts%2FPostList.tsx and see how in PostList that theme is undefined on first load of the application before using the ToggleThemeButton.

**Related code:**
```
import { useTheme } from 'react-admin;
//...
const [theme] = useTheme();
console.log('theme', theme);
```

**Other information:**

This was reported before https://github.com/marmelab/react-admin/issues/7805.

And fixed by @djhi on Jul 18, 2022: https://github.com/marmelab/react-admin/pull/7960/files#diff-5b4990e166ffe43cc265179d72851c8246ac299966ca41c97996202c540db43b

The fix got kind of reverted on May 3, 2023 with this commit https://github.com/marmelab/react-admin/commit/566fae414540d7936cc48494609fbbc6d5a9a248#diff-5b4990e166ffe43cc265179d72851c8246ac299966ca41c97996202c540db43b

**Environment**

* React-admin version: 4.16.2
* Last version that did not exhibit the issue (if applicable): see dates/versions of earlier fix and fix-revert
* React version:
* Browser: Chrome

djhi:
Fixes #9502

### DIFF
--- a/packages/ra-ui-materialui/src/theme/useTheme.ts
+++ b/packages/ra-ui-materialui/src/theme/useTheme.ts
@@ -34,5 +34,7 @@ export const useTheme = (
         'theme',
         type ?? (prefersDarkMode && darkTheme ? 'dark' : 'light')
     );
-    return [theme, setter];
+
+    // Ensure that even though the store has its value set to 'dark', we still use the light theme when no dark theme is available
+    return [darkTheme != null ? theme : 'light', setter];
 };
--- a/packages/ra-ui-materialui/src/theme/useTheme.ts
+++ b/packages/ra-ui-materialui/src/theme/useTheme.ts
@@ -1,5 +1,6 @@
 import { useStore } from 'ra-core';
 import { RaThemeOptions, ThemeType } from './types';
+import { useMediaQuery } from '@mui/material';
 
 export type ThemeSetter = (theme: ThemeType | RaThemeOptions) => void;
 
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/theme/useTheme.ts
```diff
@@ -23,7 +24,13 @@ export type ThemeSetter = (theme: ThemeType | RaThemeOptions) => void;
 export const useTheme = (
     type?: ThemeType | RaThemeOptions
 ): [ThemeType | RaThemeOptions, ThemeSetter] => {
+    const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)', {
+        noSsr: true,
+    });
     // FIXME: remove legacy mode in v5, and remove the RaThemeOptions type
-    const [theme, setter] = useStore<ThemeType | RaThemeOptions>('theme', type);
+    const [theme, setter] = useStore<ThemeType | RaThemeOptions>(
```
adguernier, line 31: `useTheme` returns the theme in the store even if it's not available in the application. Shouldn't we use `ThemesContext` to check if a `darkTheme` is defined?

djhi: Good catch! FIxed

fzaninotto (approved):

PR #9523, opened 2023-12-13
Fix `useReferenceManyFieldController` does not debounce `setFilters`

### LINKED ISSUE #9517: ReferenceManyField ignores filter debounce
**What you were expecting:**

ReferenceManyField has same filter debounce behavoir as normal list context.

[Documentation](https://marmelab.com/react-admin/FieldsForRelationships.html#referencemanyfield):
> \<ReferenceManyField> creates a ListContext with the related records, so you can use any component relying on this context (\<Datagrid>, \<SimpleList>, etc.).

**What happened instead:**

No debounce effect when using ReferenceManyField with filters.

**Steps to reproduce:**

* Preferably, a sandbox: https://stackblitz.com/edit/github-8jmtgu?file=src%2Fposts%2FPostShow.tsx (not sure if that link works...)

Screenshot 2023-12-12 at 19 58 48

To reproduce in the sandbox I basically added FilterLiveSearch to [PostShow.tsx](https://github.com/marmelab/react-admin/blob/master/examples/simple/src/posts/PostShow.tsx#L110):
```diff
            <ReferenceManyField
              reference="comments"
              target="post_id"
              sort={{ field: 'created_at', order: 'DESC' }}
            >
+             <FilterLiveSearch source="body" />
              <Datagrid>
                <DateField source="created_at" />
                <TextField source="author.name" />
                <TextField source="body" />
                <EditButton />
              </Datagrid>
            </ReferenceManyField>
```

**Related code:**

To me it looks like the implementation of `useReferenceManyFieldController` ignores the debounce parameter of `setFilters`:
```tsx
// Example generic code to set filters:
const { setFilters } = useListContext();

// ListControllerResult defines:
    setFilters: (
        filters: any,
        displayedFilters: any,
        debounce?: boolean
    ) => void;

// Inside <List> this correctly returns:
    const setFilters = useCallback(
        (filter, displayedFilters, debounce = true) =>
            debounce
                ? debouncedSetFilters(filter, displayedFilters)
                : changeParams({
                      type: SET_FILTER,
                      payload: {
                          filter: removeEmpty(filter),
                          displayedFilters,
                      },
                  }),
        [changeParams] // eslint-disable-line react-hooks/exhaustive-deps
    );

// Inside <ReferenceManyField> it instead returns:
    const setFilters = useCallback(
        (filters, displayedFilters) => {
            setFilterValues(removeEmpty(filters));
            setDisplayedFilters(displayedFilters);
            setPage(1);
        },
        [setDisplayedFilters, setFilterValues, setPage]
    );
// NOTE: inside useReferenceManyFieldController the third parameter `debounce` is ignored

```

**Other information:**

- https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/list/useListController.ts#L419
- https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/list/useListParams.ts#L226
- https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts#L131

**Environment**

* React-admin version: 4.16.2
* React version: 18.2.0
* Browser: Chrome

djhi:
Fixes #9517

### DIFF
--- a/docs/ReferenceManyField.md
+++ b/docs/ReferenceManyField.md
@@ -180,7 +180,7 @@ By default, `<ReferenceManyField>` does not refresh the data as soon as the user
 You can customize the debounce duration in milliseconds - or disable it completely - by passing a `debounce` prop to the `<ReferenceManyField>` component:
 
 ```jsx
-// wait 1 seconds instead of 500 milliseconds befoce calling the dataProvider
+// wait 1 seconds instead of 500 milliseconds before calling the dataProvider
 const PostCommentsField = () => (
     <ReferenceManyField debounce={1000}>
         ...
--- a/docs/ReferenceManyField.md
+++ b/docs/ReferenceManyField.md
@@ -111,6 +111,7 @@ This example leverages [`<SingleFieldList>`](./SingleFieldList.md) to display an
 | `pagination` | Optional | `Element` | -       | Pagination element to display pagination controls. empty by default (no pagination) |
 | `perPage`    | Optional | `number`  | 25      | Maximum number of referenced records to fetch                                       |
 | `sort`       | Optional | `{ field, order }` | `{ field: 'id', order: 'DESC' }` | Sort order to use when fetching the related records, passed to `getManyReference()` |
+| `debounce`   | Optional | `number`  | 500     | debounce time in ms for the `setFilters` callbacks                                  |
 
 `<ReferenceManyField>` also accepts the [common field props](./Fields.md#common-field-props), except `emptyText` (use the child `empty` prop instead).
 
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/controller/field/useReferenceManyFieldController.spec.tsx
```diff
@@ -273,4 +273,51 @@ describe('useReferenceManyFieldController', () => {
             );
         });
     });
+
+    it('should take only last change in case of a burst of setFilters calls (case of inputs being currently edited)', async () => {
+        let childFunction = ({ setFilters, filterValues }) => (
+            // TODO: we shouldn't import mui components in ra-core
```
fzaninotto, line 279: I think this comment should be removed

fzaninotto (changes requested):
Tou probably also need to update `ReferenceManyField` and its documentation

packages/ra-core/src/controller/field/useReferenceManyFieldController.spec.tsx
djhi, line 279:
```suggestion
```

slax57 (changes requested):
Code seems good but I agree with @fzaninotto

djhi:
> You probably also need to update `ReferenceManyField` and its documentation

I don't see anything to update

docs/ReferenceManyField.md
```diff
@@ -172,6 +173,21 @@ export const AuthorShow = () => (
...
 ```
 
+## `debounce`
+
+By default, `<ReferenceManyField>` does not refresh the data as soon as the user enters data in the filter form. Instead, it waits for half a second of user inactivity (via `lodash.debounce`) before calling the `dataProvider` on filter change. This is to prevent repeated (and useless) calls to the API.
+
+You can customize the debounce duration in milliseconds - or disable it completely - by passing a `debounce` prop to the `<ReferenceManyField>` component:
+
+```jsx
+// wait 1 seconds instead of 500 milliseconds befoce calling the dataProvider
```
slax57, line 183:
```suggestion
// wait 1 seconds instead of 500 milliseconds before calling the dataProvider
```

slax57 (changes requested):
Better :grin:

slax57 (approved):

fzaninotto (approved):

PR #9527, opened 2023-12-14
fix: AutocompleteInput TextFieldProps are not applied

adguernier:
## Problem
Some props such as `startAdornment`, `disabled` etc are not applied to the renderer input of the `<AutocompleInput>` component, because `TextFieldProps` are not applied.

## Solution 
Re-order props to avoid `TextFieldProps` to be overloaded

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -591,8 +591,8 @@ If you provided a React element for the optionText prop, you must also provide t
                         margin={margin}
                         variant={variant}
                         className={AutocompleteInputClasses.textField}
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -591,8 +591,8 @@ If you provided a React element for the optionText prop, you must also provide t
                         margin={margin}
                         variant={variant}
                         className={AutocompleteInputClasses.textField}
-                        {...TextFieldProps}
                         {...params}
+                        {...TextFieldProps}
```
adguernier, line 595: This is the fix

slax57 (approved):

slax57 (approved):

PR #9533, opened 2023-12-18
Introduce SourcePrefixContext

djhi:
## Problem

We need to prefix sources in multiple cases: ArrayInput + SimpleFormIterator, and many EE components.

To do that we currently clone children.

## Solution

This is avoidable by introducing a context that useInput would use to set the source.

It’s a breaking change though for custom inputs that don’t use `useInput`.

### DIFF
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -163,6 +163,21 @@ const CompanyField = () => (
 ```
 {% endraw %}
 
+## `<SimpleFormIterator>` no longer clone its children
+
+One consequence is that defining the `disabled` prop on the `<ArrayInput>` component does not disable its children inputs anymore. If you relied on this behavior, you now have to specify the `disabled` prop on each input:
+
+```diff
+<ArrayInput disabled={someCondition}>
+   <SimpleFormIterator>
+-      <TextInput source="lastName" />
+-      <TextInput source="firstName" />
++      <TextInput source="lastName" disabled={someCondition} />
++      <TextInput source="firstName" disabled={someCondition} />
+   </SimpleFormIterator>
+</ArrayInput>
+```
+
 ## Upgrading to v4
 
 If you are on react-admin v3, follow the [Upgrading to v4](https://marmelab.com/react-admin/doc/4.16/Upgrade.html) guide before upgrading to v5. 
--- a/packages/ra-core/src/core/SourcePrefixContext.tsx
+++ b/packages/ra-core/src/core/SourcePrefixContext.tsx
@@ -1,0 +1,19 @@
+import { createContext, useContext } from 'react';
+
+export type SourcePrefixContextValue = string;
+
+/**
+ * Context to that provides a possible prefix for the source prop of fields and inputs.
+ */
+export const SourcePrefixContext = createContext<SourcePrefixContextValue>('');
+
+export const SourcePrefixContextProvider = SourcePrefixContext.Provider;
+
+/**
+ * Hook to get the source prefix that a field or input should add to its source prop.
+ * @returns The source prefix that a field or input should add to its source prop.
+ */
+export const useSourcePrefix = (): string => {
+    const prefix = useContext(SourcePrefixContext);
+    return prefix ?? '';
+};
--- a/packages/ra-core/src/form/FormDataConsumer.spec.tsx
+++ b/packages/ra-core/src/form/FormDataConsumer.spec.tsx
@@ -34,31 +34,6 @@ describe('FormDataConsumerView', () => {
         });
     });
 
-    it('calls its children function with scopedFormData and getSource if it received an index prop', () => {
-        const children = jest.fn(({ getSource }) => {
-            getSource('id');
-            return null;
-        });
-        const formData = { id: 123, title: 'A title', authors: [{ id: 0 }] };
-
-        render(
-            <FormDataConsumerView
-                form="a-form"
-                source="authors[0]"
-                index={0}
-                formData={formData}
-            >
-                {children}
-            </FormDataConsumerView>
-        );
-
-        expect(children.mock.calls[0][0].formData).toEqual(formData);
-        expect(children.mock.calls[0][0].scopedFormData).toEqual({ id: 0 });
-        expect(children.mock.calls[0][0].getSource('id')).toEqual(
-            'authors[0].id'
-        );
-    });
-
     it('calls its children with updated formData on first render', async () => {
         let globalFormData;
         render(
@@ -130,10 +105,7 @@ describe('FormDataConsumerView', () => {
                                     globalScopedFormData = scopedFormData;
                                     return scopedFormData &&
                                         scopedFormData.name ? (
-                                        <TextInput
-                                            source={getSource('role')}
-                                            {...rest}
-                                        />
+                                        <TextInput source="role" {...rest} />
                                     ) : null;
                                 }}
                             </FormDataConsumer>
--- a/packages/ra-core/src/form/FormDataConsumer.tsx
+++ b/packages/ra-core/src/form/FormDataConsumer.tsx
@@ -3,6 +3,7 @@ import { ReactNode } from 'react';
 import { useFormContext, FieldValues } from 'react-hook-form';
 import get from 'lodash/get';
 import { useFormValues } from './useFormValues';
+import { useSourcePrefix } from '../core';
 
 /**
  * Get the current (edited) value of the record from the form and pass it
@@ -64,13 +65,17 @@ export const FormDataConsumerView = <
 >(
     props: Props<TFieldValues>
 ) => {
-    const { children, form, formData, source, index, ...rest } = props;
+    const { children, form, formData, source, ...rest } = props;
     let ret;
 
+    const prefix = useSourcePrefix();
+    const matches = ArraySourceRegex.exec(prefix);
+
     // If we have an index, we are in an iterator like component (such as the SimpleFormIterator)
-    if (typeof index !== 'undefined' && source) {
-        const scopedFormData = get(formData, source);
-        const getSource = (scopedSource: string) => `${source}.${scopedSource}`;
+    if (matches && prefix) {
+        const scopedFormData = get(formData, prefix);
+        // Not needed anymore. Kept to avoid breaking existing code
+        const getSource = (scopedSource: string) => scopedSource;
         ret = children({ formData, scopedFormData, getSource, ...rest });
     } else {
         ret = children({
@@ -84,6 +89,8 @@ export const FormDataConsumerView = <
 };
 
 export default FormDataConsumer;
+
+const ArraySourceRegex = new RegExp(/.+\.\d+$/);
 
 export interface FormDataConsumerRenderParams<
     TFieldValues extends FieldValues = FieldValues,
--- a/packages/ra-core/src/form/useApplyInputDefaultValues.ts
+++ b/packages/ra-core/src/form/useApplyInputDefaultValues.ts
@@ -7,6 +7,7 @@ import {
 import get from 'lodash/get';
 import { useRecordContext } from '../controller';
 import { InputProps } from './useInput';
+import { useSourcePrefix } from '../core';
 
 interface StandardInput {
     inputProps: Partial<InputProps> & { source: string };
@@ -32,6 +33,9 @@ export const useApplyInputDefaultValues = ({
     fieldArrayInputControl,
 }: Props) => {
     const { defaultValue, source } = inputProps;
+    const prefix = useSourcePrefix();
+    const finalSource = prefix ? `${prefix}.${source}` : source;
+
     const record = useRecordContext(inputProps);
     const {
         getValues,
@@ -41,8 +45,8 @@ export const useApplyInputDefaultValues = ({
         reset,
     } = useFormContext();
     const recordValue = get(record, source);
-    const formValue = get(getValues(), source);
-    const { isDirty } = getFieldState(source, formState);
+    const formValue = get(getValues(), finalSource);
+    const { isDirty } = getFieldState(finalSource, formState);
 
     useEffect(() => {
         if (
@@ -58,11 +62,11 @@ export const useApplyInputDefaultValues = ({
         // Since we use get(record, source), if source is like foo.23.bar,
         // this effect will run. However we only want to set the default value
         // for the subfield bar if the record actually has a value for foo.23
-        const pathContainsIndex = source
+        const pathContainsIndex = finalSource
             .split('.')
             .some(pathPart => numericRegex.test(pathPart));
         if (pathContainsIndex) {
-            const parentPath = source.split('.').slice(0, -1).join('.');
+            const parentPath = finalSource.split('.').slice(0, -1).join('.');
             const parentValue = get(getValues(), parentPath);
             if (parentValue == null) {
                 // the parent is undefined, so we don't want to set the default value
@@ -88,7 +92,7 @@ export const useApplyInputDefaultValues = ({
             return;
         }
 
-        resetField(source, { defaultValue });
+        resetField(finalSource, { defaultValue });
     });
 };
 
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -16,6 +16,7 @@ import { useFormGroupContext } from './useFormGroupContext';
 import { useFormGroups } from './useFormGroups';
 import { useApplyInputDefaultValues } from './useApplyInputDefaultValues';
 import { useEvent } from '../util';
+import { useSourcePrefix } from '../core';
 
 // replace null or undefined values by empty string to avoid controlled/uncontrolled input warning
 const defaultFormat = (value: any) => (value == null ? '' : value);
@@ -38,7 +39,9 @@ export const useInput = <ValueType = any>(
         validate,
         ...options
     } = props;
-    const finalName = name || source;
+    const prefix = useSourcePrefix();
+    const finalSource = prefix ? `${prefix}.${source}` : source;
+    const finalName = name || finalSource;
     const formGroupName = useFormGroupContext();
     const formGroups = useFormGroups();
     const record = useRecordContext();
@@ -48,12 +51,12 @@ export const useInput = <ValueType = any>(
             return;
         }
 
-        formGroups.registerField(source, formGroupName);
+        formGroups.registerField(finalSource, formGroupName);
 
         return () => {
-            formGroups.unregisterField(source, formGroupName);
+            formGroups.unregisterField(finalSource, formGroupName);
         };
-    }, [formGroups, formGroupName, source]);
+    }, [formGroups, formGroupName, finalSource]);
 
     const sanitizedValidate = Array.isArray(validate)
         ? composeValidators(validate)
@@ -120,7 +123,7 @@ export const useInput = <ValueType = any>(
     };
 
     return {
-        id: id || source,
+        id: id || finalSource,
         field,
         fieldState,
         formState,
--- a/packages/ra-core/src/util/FieldTitle.tsx
+++ b/packages/ra-core/src/util/FieldTitle.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { ReactElement, memo } from 'react';
 
 import { useTranslateLabel } from '../i18n';
+import { useSourcePrefix } from '../core';
 
 export interface FieldTitleProps {
     isRequired?: boolean;
@@ -12,6 +13,8 @@ export interface FieldTitleProps {
 
 export const FieldTitle = (props: FieldTitleProps) => {
     const { source, label, resource, isRequired } = props;
+    const prefix = useSourcePrefix();
+    const finalSource = prefix ? `${prefix}.${source}` : source;
     const translateLabel = useTranslateLabel();
 
     if (label === true) {
@@ -33,7 +36,7 @@ export const FieldTitle = (props: FieldTitleProps) => {
             {translateLabel({
                 label,
                 resource,
-                source,
+                source: finalSource,
             })}
             {isRequired && <span aria-hidden="true">&thinsp;*</span>}
         </span>
--- a/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
@@ -3,6 +3,7 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {
     RecordContextProvider,
+    ResourceContextProvider,
     minLength,
     required,
     testDataProvider,
@@ -104,22 +105,24 @@ describe('<ArrayInput />', () => {
     it('should clone each input once per value in the array', () => {
         render(
             <AdminContext dataProvider={testDataProvider()}>
-                <SimpleForm
-                    onSubmit={jest.fn}
-                    defaultValues={{
-                        arr: [
-                            { id: 123, foo: 'bar' },
-                            { id: 456, foo: 'baz' },
-                        ],
-                    }}
-                >
-                    <ArrayInput resource="bar" source="arr">
-                        <SimpleFormIterator>
-                            <NumberInput source="id" />
-                            <TextInput source="foo" />
-                        </SimpleFormIterator>
-                    </ArrayInput>
-                </SimpleForm>
+                <ResourceContextProvider value="bar">
+                    <SimpleForm
+                        onSubmit={jest.fn}
+                        defaultValues={{
+                            arr: [
+                                { id: 123, foo: 'bar' },
+                                { id: 456, foo: 'baz' },
+                            ],
+                        }}
+                    >
+                        <ArrayInput source="arr">
+                            <SimpleFormIterator>
+                                <NumberInput source="id" />
+                                <TextInput source="foo" />
+                            </SimpleFormIterator>
+                        </ArrayInput>
+                    </SimpleForm>
+                </ResourceContextProvider>
             </AdminContext>
         );
         expect(
@@ -143,29 +146,30 @@ describe('<ArrayInput />', () => {
     it('should apply validation to both itself and its inner inputs', async () => {
         render(
             <AdminContext dataProvider={testDataProvider()}>
-                <SimpleForm
-                    onSubmit={jest.fn}
-                    defaultValues={{
-                        arr: [],
-                    }}
-                >
-                    <ArrayInput
-                        resource="bar"
-                        source="arr"
-                        validate={[minLength(2, 'array_min_length')]}
+                <ResourceContextProvider value="bar">
+                    <SimpleForm
+                        onSubmit={jest.fn}
+                        defaultValues={{
+                            arr: [],
+                        }}
                     >
-                        <SimpleFormIterator>
-                            <TextInput
-                                source="id"
-                                validate={[required('id_required')]}
-                            />
-                            <TextInput
-                                source="foo"
-                                validate={[required('foo_required')]}
-                            />
-                        </SimpleFormIterator>
-                    </ArrayInput>
-                </SimpleForm>
+                        <ArrayInput
+                            source="arr"
+                            validate={[minLength(2, 'array_min_length')]}
+                        >
+                            <SimpleFormIterator>
+                                <TextInput
+                                    source="id"
+                                    validate={[required('id_required')]}
+                                />
+                                <TextInput
+                                    source="foo"
+                                    validate={[required('foo_required')]}
+                                />
+                            </SimpleFormIterator>
+                        </ArrayInput>
+                    </SimpleForm>
+                </ResourceContextProvider>
             </AdminContext>
         );
 
--- a/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
@@ -10,6 +10,7 @@ import {
     useGetValidationErrorMessage,
     useFormGroupContext,
     useFormGroups,
+    useSourcePrefix,
 } from 'ra-core';
 import { useFieldArray, useFormContext } from 'react-hook-form';
 import {
@@ -90,6 +91,8 @@ export const ArrayInput = (props: ArrayInputProps) => {
 
     const formGroupName = useFormGroupContext();
     const formGroups = useFormGroups();
+    const prefix = useSourcePrefix();
+    const finalSource = prefix ? `${prefix}.${source}` : source;
 
     const sanitizedValidate = Array.isArray(validate)
         ? composeSyncValidators(validate)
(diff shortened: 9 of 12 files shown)


packages/ra-core/src/util/FieldTitle.tsx
```diff
@@ -33,7 +36,7 @@ export const FieldTitle = (props: FieldTitleProps) => {
             {translateLabel({
                 label,
                 resource,
-                source,
+                source: finalSource,
```
fzaninotto, line 39: No, otherwise the inputs in an ArrayInput will all have an (untranslatable) name.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -158,15 +161,15 @@ export const ArrayInput = (props: ArrayInputProps) => {
             margin={margin}
             className={clsx(
                 'ra-input',
-                `ra-input-${source}`,
+                `ra-input-${finalSource}`,
```
fzaninotto, line 164: No, I think we should keep the original source here.

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
```diff
@@ -501,17 +474,14 @@ describe('<SimpleFormIterator />', () => {
                     <ArrayInput source="emails">
                         <SimpleFormIterator>
                             <FormDataConsumer>
-                                {({ getSource }) => (
+                                {() => (
```
fzaninotto, line 477: We don't need a FormDataConsumer then

fzaninotto (changes requested):

docs/Upgrade.md
```diff
@@ -163,6 +163,21 @@ const CompanyField = () => (
 ```
 {% endraw %}
 
+## `<SimpleFormIterator>` no longer clone its children
+
+One consequence is that defining the `disabled` prop on the `<ArrayInput>` component does not disable its children inputs anymore. If you relied on this behavior, you now have to specify the `disabled` prop on each input:
+
+```diff
+<ArrayInput disabled={someCondition}>
```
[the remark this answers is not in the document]
djhi: Nope because we're also going to pass the disabled in the `ArrayInputContext` so that the default buttons for adding, removing, etc are disabled too

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -85,6 +90,8 @@ export const FormDataConsumerView = <
 
 export default FormDataConsumer;
 
+const ArraySourceRegex = new RegExp(/.+\.\d+$/);
```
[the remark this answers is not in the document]
djhi: That will not because 'formData` doesn't change. It will still return the full form data

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -64,13 +65,17 @@ export const FormDataConsumerView = <
...
+    const matches = ArraySourceRegex.exec(prefix);
+
     // If we have an index, we are in an iterator like component (such as the SimpleFormIterator)
-    if (typeof index !== 'undefined' && source) {
-        const scopedFormData = get(formData, source);
-        const getSource = (scopedSource: string) => `${source}.${scopedSource}`;
+    if (matches && prefix) {
+        const scopedFormData = get(formData, prefix);
+        // Not needed anymore. Kept to avoid breaking existing code
+        const getSource = (scopedSource: string) => scopedSource;
```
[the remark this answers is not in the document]
djhi: I thought we wanted to avoid BC?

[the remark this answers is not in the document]
fzaninotto: Yes, but what's the use case for this `getSource` function now that we have this context?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
djhi: True but the test validates that everything works even when you have a `FormDataConsumer`

packages/ra-core/src/util/FieldTitle.tsx
djhi: True, but we probably don't want the input name alone. What happen if you have two fields with the same source, one at the record root and another in the ArrayInput?

djhi: Actually, it seems we had this before: `resources.RESOURCE.fields.ARRAY_INPUT_SOURCE.INPUT_SOURCE`

djhi: Well this change allows the FieldTitle to behave as before. That can be verified with the `calls its children with updated scopedFormData when inside an ArrayInput` test of the `FormDataConsumer`

adguernier (approved):

adguernier (changes requested):
Just two minor omissions in the documentation otherwise it looks good to me

docs/SimpleFormIterator.md
```diff
@@ -114,7 +114,7 @@ A list of Input elements, that will be rendered on each row.
 
 By default, `<SimpleFormIterator>` renders one input per line, but they can be displayed inline with the `inline` prop.
 
-`<SimpleFormIterator>` also accepts `<FormDataConsumer>` as child. When used inside a form iterator, `<FormDataConsumer>` provides two additional properties to its children function:
+`<SimpleFormIterator>` also accepts `<FormDataConsumer>` as child. When used inside a form iterator, `<FormDataConsumer>` provides one additional propertie to its children function:
```
adguernier, line 117:
```suggestion
`<SimpleFormIterator>` also accepts `<FormDataConsumer>` as child. When used inside a form iterator, `<FormDataConsumer>` provides one additional property to its children function:
```

docs/Inputs.md
```diff
@@ -534,7 +534,7 @@ const OrderEdit = () => (
 );
 ```
 
-**Tip**: When using a `FormDataConsumer` inside an `ArrayInput`, the `FormDataConsumer` will provide two additional properties to its children function:
+**Tip**: When using a `FormDataConsumer` inside an `ArrayInput`, the `FormDataConsumer` will provide one additional propertie to its children function:
```
adguernier, line 537: My bad:
```suggestion
**Tip**: When using a `FormDataConsumer` inside an `ArrayInput`, the `FormDataConsumer` will provide one additional property to its children function:
```

adguernier (changes requested):

packages/ra-core/src/core/SourcePrefixContext.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react';
+
+export type SourceContextValue = (source: string) => string;
+
+/**
+ * Context that provides a function that accept a source and return a modified source (prefixed, suffixed, etc.) for fields and inputs.
+ */
```
fzaninotto, line 7: could you please add an example in jsDoc?

packages/ra-core/src/core/SourcePrefixContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const SourceContext = createContext<SourceContextValue>(null);
+
+export const SourceContextProvider = SourceContext.Provider;
+
+/**
+ * Hook to get a source that may be prefixed, suffixed, etc. by a parent component.
+ * @param source The original field or input source
+ * @returns The modified source
+ */
```
fzaninotto, line 16: Could you please add an example in jsDoc?

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
```diff
@@ -104,68 +105,71 @@ describe('<ArrayInput />', () => {
...
+                                <TextInput source="foo" />
+                            </SimpleFormIterator>
+                        </ArrayInput>
+                    </SimpleForm>
+                </ResourceContextProvider>
             </AdminContext>
         );
         expect(
-            screen.queryAllByLabelText('resources.bar.fields.arr.id')
+            screen.queryAllByLabelText('resources.bar.fields.id')
```
fzaninotto, line 129: This doesn't feel right and it's a BC. I know that I asked you to revert the change that causes this, but it was my mistake. 

translation messages of inputs in arrayinputs should have the arrayinput source in it - just not the index. 

So with this setup:

```
<ArrayInput source="arr">
   <SimpleFormIterator>
      <NumberInput source="id" />
   </SimpleFormIterator>
</ArrayInput>
```

The default label of all the `id` number inputs should be `resources.bar.fields.arr.id` (but not `resources.bar.fields.arr.0.id` or `resources.bar.fields.id`)

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.spec.tsx
```diff
@@ -48,7 +48,7 @@ describe('<SimpleFormIterator />', () => {
             </Wrapper>
         );
         const inputElements = screen.queryAllByLabelText(
-            'resources.undefined.fields.emails.email'
```
fzaninotto, line 51: Same, this feels wrong

packages/ra-core/src/form/FormDataConsumer.spec.tsx
```diff
@@ -150,7 +116,7 @@ describe('FormDataConsumerView', () => {
         expect(globalScopedFormData).toEqual({ name: null });
 
         fireEvent.change(
-            screen.getByLabelText('resources.undefined.fields.authors.name'),
+            screen.getByLabelText('resources.undefined.fields.name'),
```
fzaninotto, line 119: same

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
```diff
@@ -94,24 +96,9 @@ export const SimpleFormIteratorItem = React.forwardRef(
...
-                                    index: source ? undefined : index2,
-                                    resource,
-                                    disabled,
-                                    ...inputProps,
-                                });
-                            }
-                        )}
+                        <SourceContextProvider value={sourceContext}>
+                            {children}
+                        </SourceContextProvider>
```
fzaninotto, line 101: Why do you add the SourceContext in each SimpleFormIteratorItem instead of once in the wrapper SimpleFormIterator?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
```diff
@@ -94,24 +96,9 @@ export const SimpleFormIteratorItem = React.forwardRef(
...
-                            (input: ReactElement, index2) => {
-                                if (!isValidElement<any>(input)) {
-                                    return null;
-                                }
-                                const { source, ...inputProps } = input.props;
-                                return cloneElement(input, {
-                                    source: source
-                                        ? `${member}.${source}`
-                                        : member,
-                                    index: source ? undefined : index2,
```
fzaninotto, line 108: in packages/ra-core/src/form/FormDataConsumer.spec.tsx, first test, I see a reference to the `index` prop being injected. Maybe there are other.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
djhi: Because `member` actually contains the index

docs/Upgrade.md
```diff
@@ -163,6 +163,21 @@ const CompanyField = () => (
...
+   <SimpleFormIterator>
+-      <TextInput source="lastName" />
+-      <TextInput source="firstName" />
++      <TextInput source="lastName" disabled={someCondition} />
++      <TextInput source="firstName" disabled={someCondition} />
+   </SimpleFormIterator>
+</ArrayInput>
+```
+
```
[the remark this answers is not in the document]
djhi: Added

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
djhi: Nope

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -41,8 +44,8 @@ export const useApplyInputDefaultValues = ({
         reset,
     } = useFormContext();
     const recordValue = get(record, source);
```
slax57, line 46: Why not `get(record, finalSource)`? Am I missing stg here? :thinking:

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
```diff
@@ -23,7 +18,11 @@ export const TranslatableInputsTabContent = (
     props: TranslatableInputsTabContentProps
 ): ReactElement => {
     const { children, groupKey = '', locale, ...other } = props;
-    const { selectedLocale, getLabel, getSource } = useTranslatableContext();
+    const { selectedLocale, getSource } = useTranslatableContext();
```
slax57, line 21: Unless I'm mistaken, we should mark `getLabel` as deprecated in `TranslatableContext`

packages/ra-ui-materialui/src/input/TranslatableInputs.spec.tsx
```diff
@@ -126,17 +126,17 @@ describe('<TranslatableInputs />', () => {
         );
 
         fireEvent.change(
-            screen.getAllByLabelText('resources.undefined.fields.name')[0],
+            screen.getByLabelText('resources.undefined.fields.name.en'),
```
slax57, line 129: Isn't that a BC? I'm worried the labels for TranslatableInputs will no longer be translated correctly, because the translation key has changed

slax57 (changes requested):

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
djhi: I haven't changed the `TranslatableFields` yet which still uses it. Should I update it in this PR as well?

packages/ra-ui-materialui/src/input/TranslatableInputs.spec.tsx
djhi: Indeed, good catch. I don't have a good solution yet for this.

docs/Upgrade.md
```diff
@@ -163,6 +163,58 @@ const CompanyField = () => (
...
+## `<FormDataConsumer>` no longer pass a `getSource` function
```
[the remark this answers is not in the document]
slax57: Did you see this comment?

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
slax57: Good question. I'd advocate for a second PR since this one is already big enough. Can you add a card in our internal tracker (including the need to mark getLabel as deprecated)?

slax57 (changes requested):

packages/ra-core/src/i18n/useTranslateLabel.ts
```diff
@@ -30,13 +33,14 @@ export const useTranslateLabel = () => {
             return translate(
                 ...getFieldLabelTranslationArgs({
                     label: label as string,
+                    labelFromSourceContext: sourceContext?.getLabel(source),
```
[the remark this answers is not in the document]
djhi: The issue is that I would have to also pass the unmodified source. You can see I pass the `finalSource` at L40. Besides, as the `labelFromSourceContext` would be `undefined` if `sourceContext` isn't present, I fail to see what it would change

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
```diff
@@ -24,6 +19,13 @@ export const TranslatableInputsTabContent = (
 ): ReactElement => {
     const { children, groupKey = '', locale, ...other } = props;
     const { selectedLocale, getLabel, getSource } = useTranslatableContext();
+    const sourceContext = useMemo(
+        () => ({
+            getSource: (source: string) => getSource(source, locale),
+            getLabel,
```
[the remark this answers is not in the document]
djhi: Yes but `useTranslatable` call `translateLabel` with an unmodified source (without the `locale`). The alternative would be to rebuild the inferred translation key instead in `useTranslatable` like we do in `ArrayInput`. We can either do that or replicate what we do in `useTranslatable` inside `ArrayInput` to have only one place where we build the translation key (`resources.RESOURCE.fields.FIELD`)

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -32,6 +33,9 @@ export const useApplyInputDefaultValues = ({
     fieldArrayInputControl,
 }: Props) => {
     const { defaultValue, source } = inputProps;
+    const sourceContext = useSourceContext();
+    const finalSource = sourceContext?.getSource(source) ?? source;
```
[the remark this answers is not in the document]
djhi: Not convinced it's useful.

packages/ra-core/src/core/SourceContext.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export type SourceContextValue = {
+    /*
+     * Returns the source for a field or input, modified according to the context.
+     */
+    getSource: (source: string) => string;
+    /*
+     * Returns the label for a field or input, modified according to the context. Returns a translation key.
+     */
+    getLabel: (source: string) => string;
```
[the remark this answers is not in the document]
djhi: In the only two cases where we have a SourceContext, we already have access to everything we need so I don't think we need that

packages/ra-core/src/util/getFieldLabelTranslationArgs.spec.ts
```diff
@@ -5,16 +5,6 @@ describe('getFieldLabelTranslationArgs', () => {
...
 
-    it('should return the label when given', () => {
-        expect(
-            getFieldLabelTranslationArgs({
-                label: 'foo',
-                resource: 'posts',
-                source: 'title',
-            })
-        ).toEqual(['foo', { _: 'foo' }]);
-    });
```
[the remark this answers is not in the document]
djhi: Added back

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
djhi: It's not deprecated though

packages/ra-core/src/form/FormDataConsumer.tsx
```diff
@@ -64,34 +65,36 @@ export const FormDataConsumerView = <
...
+    );
+
     // If we have an index, we are in an iterator like component (such as the SimpleFormIterator)
-    if (typeof index !== 'undefined' && source) {
-        const scopedFormData = get(formData, source);
-        const getSource = (scopedSource: string) => `${source}.${scopedSource}`;
-        ret = children({ formData, scopedFormData, getSource, ...rest });
+    if (matches) {
+        const scopedFormData = get(formData, matches[0]);
+        ret = children({ formData, scopedFormData, ...rest });
```
fzaninotto, line 80: I'm thinking that we could extract the logic for getting scopedFormData in a custom hook depending only on `useSourceContext`. That way, we could encourage the good practice of using this hook instead of expecting the scoped form data to be injected (which is very react-admin v3).

packages/ra-core/src/form/useApplyInputDefaultValues.ts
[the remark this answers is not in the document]
fzaninotto: I'm +1 with @slax57

packages/ra-core/src/form/useApplyInputDefaultValues.ts
```diff
@@ -58,11 +62,11 @@ export const useApplyInputDefaultValues = ({
         // Since we use get(record, source), if source is like foo.23.bar,
         // this effect will run. However we only want to set the default value
         // for the subfield bar if the record actually has a value for foo.23
-        const pathContainsIndex = source
+        const pathContainsIndex = finalSource
             .split('.')
             .some(pathPart => numericRegex.test(pathPart));
         if (pathContainsIndex) {
-            const parentPath = source.split('.').slice(0, -1).join('.');
+            const parentPath = finalSource.split('.').slice(0, -1).join('.');
```
fzaninotto, line 69: can't we get the parent path simply by calling the SourceContext?

packages/ra-core/src/i18n/useTranslateLabel.ts
```diff
@@ -2,12 +2,13 @@ import { useCallback, ReactElement } from 'react';
 
 import { useTranslate } from './useTranslate';
 import { useLabelPrefix, getFieldLabelTranslationArgs } from '../util';
-import { useResourceContext } from '../core';
+import { useResourceContext, useSourceContext } from '../core';
 
 export const useTranslateLabel = () => {
     const translate = useTranslate();
     const prefix = useLabelPrefix();
```
fzaninotto, line 9: I think this is no longer needed.

packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    value={{
+                        getSource: source => source,
+                        getLabel: source => `test.${source}`,
+                    }}
+                >
+                    <TranslateLabel source="title" resource="posts" />
+                </SourceContextProvider>
+            </TestTranslationProvider>
+        );
+        screen.getByText('Title');
```
fzaninotto, line 131: This I don't understand, as the labelFrom Source Context will be `test.title`

packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        getLabel: source => `test.${source}`,
+                    }}
+                >
+                    <TranslateLabel source="title" resource="posts" />
+                </SourceContextProvider>
+            </TestTranslationProvider>
+        );
+        screen.getByText('Label for title');
+    });
+});
```
fzaninotto, line 155: I think we're missing tests cases for sources with dots and / or inside ArrayInputs.

packages/ra-core/src/i18n/useTranslateLabel.ts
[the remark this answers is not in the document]
slax57: I disagree, `sourceContext?.getLabel(source)` will return `undefined` if there is no `sourceContext` but it will throw an error if `sourceContext.getLabel` is `undefined`

packages/ra-core/src/util/getFieldLabelTranslationArgs.ts
```diff
@@ -37,6 +45,10 @@ export default (options?: Args): TranslationArguments => {
         ['underscore', 'humanize']
     );
 
+    if (labelFromSourceContext) {
+        return [labelFromSourceContext, { _: defaultLabel }];
+    }
+
```
fzaninotto, line 51: Because of early return on L37, I'm afraid this won't work.

packages/ra-core/src/i18n/useTranslateLabel.ts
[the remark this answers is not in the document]
fzaninotto: I don't understand why you need to pass both labels. It shouldn't matter to `getFieldLabelTranslationArgs` where the label comes from.

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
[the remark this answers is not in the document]
fzaninotto: I agree with @slax57 : the `useTranslatable hook rebuilds part of the SourceContext. We should leave the `source` and `label` responsibilities to the `SourceContext`.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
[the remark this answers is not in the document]
djhi: So rebuilding the key here like in SimpleFormIteratorContext ?

packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
djhi: Because the translation does not exist so we fall back to inferred label.

packages/ra-core/src/i18n/useTranslateLabel.ts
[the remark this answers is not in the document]
djhi: Because in the first label case, we don't want the inferred label fallback (users might pass something which is not a translation key) whereas in the second label case (the one from SourceContext) we do want to fallback to the inferred label

packages/ra-core/src/util/getFieldLabelTranslationArgs.ts
djhi: It does, if you didn't provide your own label

packages/ra-core/src/form/FormDataConsumer.tsx
djhi: a render prop does not inject anything (that's actually one of its strength) and saves you the trouble to declare a component just to call a hook. IMO that's way more pratical. We could do both though to accommodate all use cases.

packages/ra-core/src/form/useApplyInputDefaultValues.ts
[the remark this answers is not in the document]
djhi: Added

packages/ra-core/src/form/useApplyInputDefaultValues.ts
djhi: The SourceContext does not give the parent path, it returns the modified source for its children inputs/fields. Did I miss something?

packages/ra-core/src/form/FormDataConsumer.tsx
djhi: Can I address this in a dedicated PR?

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
```diff
@@ -77,6 +80,25 @@ export const SimpleFormIteratorItem = React.forwardRef(
...
+        const sourceContext = useMemo(
+            () => ({
+                getSource: (source: string) =>
+                    source ? `${member}.${source}` : member,
+                getLabel: (source: string) => {
+                    // remove digits, e.g. 'book.authors.2.categories.3.identifier.name' => 'book.authors.categories.identifier.name'
+                    const itemSource = `${member.replace(
+                        /\.\d+/g,
+                        ''
+                    )}.${source}`;
```
slax57, line 93: Not sure you need this, since, if I'm not mistaken, `getLabel` will be called with the original `source` (i.e. `"name"`) and not the `finalSource` (i.e. `'book.authors.2.categories.3.identifier.name'`)

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
```diff
@@ -23,7 +21,21 @@ export const TranslatableInputsTabContent = (
...
     const { children, groupKey = '', locale, ...other } = props;
-    const { selectedLocale, getLabel, getSource } = useTranslatableContext();
+    const resource = useResourceContext(props);
+    const { selectedLocale } = useTranslatableContext();
+    const parentSourceContext = useSourceContext();
+    const sourceContext = useMemo(
+        () => ({
+            getSource: (source: string) => `${source}.${locale}`,
+            getLabel: (source: string) => {
+                const itemSource = source.replace(`.${locale}`, '');
```
slax57, line 31: same

slax57 (changes requested):
Other than that it looks good! Well done :sweat_smile:

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
djhi: Nope because for instance with an `<ArrayInput source="authors">` containing a `<TextInput source="name">`  we want the label key to `authors.name` not `name`

packages/ra-ui-materialui/src/input/TranslatableInputsTabContent.tsx
djhi: True for this one though

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
slax57: Okay I get it, I missed the fact that you were applying the `replace` on `member` and not `source`.

However this is still missing something to me: just like for `getSource`, you need to check if `source` is defined before concatenating with `.${source}`.

slax57 (changes requested):

docs/Upgrade.md
```diff
@@ -204,6 +204,58 @@ const CompanyField = () => (
...
+        <SimpleForm>
+            <ArrayInput source="authors">
+                <SimpleFormIterator>
+                    <TextInput source="name" />
+                    <FormDataConsumer>
+                        {({
+                            formData, // The whole form data
+                            scopedFormData, // The data for this item of the ArrayInput
+-                           getSource,
+                            ...rest
```
fzaninotto, line 240: rest params don't serve any purpose in this example, we should remove them here (and line 247).

packages/ra-core/src/form/useInput.ts
```diff
@@ -38,22 +39,33 @@ export const useInput = <ValueType = any>(
...
     const formGroups = useFormGroups();
     const record = useRecordContext();
 
+    if (
+        !source &&
+        props.label == null &&
+        process.env.NODE_ENV === 'development'
+    ) {
+        console.warn(
+            'If your input source is empty, you must provide a label prop.'
```
fzaninotto, line 54:
```suggestion
            'Input components require either a source or a label prop.'
```

fzaninotto (changes requested):
minor remarks

fzaninotto (approved):

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
```diff
@@ -167,6 +167,47 @@ export const Scalar = () => (
...
+                        onSuccess: data => {
+                            console.log(data);
+                        },
+                    }}
+                >
+                    <SimpleForm>
+                        <TextInput source="title" />
+                        <ArrayInput source="tags" fullWidth>
+                            <SimpleFormIterator disableReordering>
+                                <TextInput source="" helperText={false} />
```
slax57, line 201: This story is great for testing the corner case where we have no `label`, but it should demonstrate the recommended way to do a scalar array input, and have a `label` by default (besides, otherwise, it raises the newly added warning in `useInput`)

slax57 (changes requested):
Otherwise it's good to me, we're so close!! :grin:

slax57 (approved):
:tada: :tada: :tada:

PR #9537, opened 2023-12-20
Feat(usePreference): Throw an error if used outisde right context

### LINKED ISSUE #9526: usePreferences behavior is unexpected when used outside of Configurable
**What you were expecting:**

I would have expected the hook `usePreferences` to save a value inside the RaStore.preferences path.

If `usePreferences` is not to be used outside of a Configurable (which in my opinion it should be based on the name), at least a a warning should be printed to the console if the `PreferenceKeyContext` is not provided or empty.

**What happened instead:**

The key value forwarded to the `setItem` function in the store is empty, resulting in a value being set for `RaStore.` no matter what the actual key is.

image

**Steps to reproduce:**

- Use the  following code snipped outside of a Configurable component

````tsx
   //Does not work but should be identical to the hook below
   const [showExpertView, setShowExpertView] = usePreferences(
    "expertView",
    false,
  );
  
 //Works
  const [showExpertView, setShowExpertView] = useStore(
    "preferences.expertView",
    false,
  );
````

### PR DESCRIPTION
**Pull request by arimet**
Resolve #9526

### DIFF
--- a/docs/Configurable.md
+++ b/docs/Configurable.md
@@ -37,7 +37,7 @@ const ConfigurableTextBlock = ({ preferenceKey = "textBlock", ...props }) => (
 
 `<Configurable>` creates a context for the `preferenceKey`, so that both the child component and the editor can access it.
 
-The editor commponent lets users edit the preferences for the configurable compoonent. It does so using the `usePreference` hook, which is a namespaced version of [the `useStore` hook](./useStore.md) for the current `preferenceKey`:
+The editor commponent lets users edit the preferences for the configurable component. It does so using the `usePreference` hook, which is a namespaced version of [the `useStore` hook](./useStore.md) for the current `preferenceKey`:
 
 ```jsx
 import { usePreference } from 'react-admin';
@@ -58,7 +58,6 @@ const TextBlockEditor = () => {
     );
 };
 ```
- 
 The inner component reads the preferences using the same `usePreference` hook:
 
 ```jsx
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/Configurable.md [hunk 1]
```diff
@@ -58,7 +58,7 @@ const TextBlockEditor = () => {
     );
 };
 ```
- 
+This hook is only used within a Configurable component.
```
slax57, line 61: This line is redundant with the information given above. IMO it's not necessary to add it.

packages/ra-ui-materialui/src/preferences/Configurable.stories.tsx
```diff
@@ -218,3 +218,8 @@ export const I18n = () => {
 };
 
 export const NotInContext = () => <TextBlock />;
+
+export const usePreferenceNotInConfigurable = () => {
```
adguernier, line 222:
```suggestion
export const UsePreferenceOutsideConfigurable = () => {
```

adguernier (changes requested):

packages/ra-ui-materialui/src/preferences/Configurable.stories.tsx
slax57, line 222:
```suggestion
export const UsePreferenceNotInConfigurable = () => {
```

Not sure it's necessary to create a story for this corner-case though...

djhi (approved):

PR #9543, opened 2023-12-22
Fix(input): Set DateInput and DateTimeInput to uncontrolled input

### LINKED ISSUE #9330: DateTimeInput or DateInput reinitialise when typing a date starting with '0' - only on Firefox
Note: Only noticed on firefox.

**What you were expecting:**

I have a DateTimeInput field with a date `10/01/1985 10:20`:
image

I select the field and type "01". The date is now `01/01/1985 10:20`

**What happened instead:**

The field is reinitialized as soon as I type the '0' 
image

**Steps to reproduce:**

Go to https://react-admin-storybook.vercel.app/?path=/story/ra-ui-materialui-input-datetimeinput--basic
Type a valid date.
Now focus the beginning of the field, and type '0' (as if you want to type '01' for exemple) 

**Related code:**

The basic stackblitz is enough to reproduce. 
https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple
On firefox : Edit a Post > Miscellanous > Edit a date input by starting by a '0'

```
<DateTimeInput source='foo'/>
```

**Other information:**

I first though of a behaviour of the html input on firefox. But trying it directly : 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local 
I did not reproduce the same behaviour.

**Environment**

* React-admin version: "4.13.2"
* React version: "18.2.0"
* Browser:  Firefox 118.0.1 (64 bits)

arimet:
Fix #9330

### DIFF
--- a/packages/ra-ui-materialui/src/input/DateInput.tsx
+++ b/packages/ra-ui-materialui/src/input/DateInput.tsx
@@ -42,35 +42,101 @@ export const DateInput = ({
     margin,
     onBlur,
     onChange,
+    onFocus,
     parse,
     validate,
     variant,
     ...rest
 }: DateInputProps) => {
-    const { field, fieldState, formState, id, isRequired } = useInput({
+    const {
+        field,
+        fieldState: { error, invalid, isTouched },
+        formState: { isSubmitted },
+        id,
+        isRequired,
+    } = useInput({
         defaultValue,
         name,
-        format,
-        parse,
         onBlur,
-        onChange,
         resource,
         source,
         validate,
         ...rest,
     });
+    const [renderCount, setRenderCount] = React.useState(1);
 
-    const { error, invalid, isTouched } = fieldState;
-    const { isSubmitted } = formState;
+    const initialDefaultValueRef = React.useRef(defaultValue);
+
+    React.useEffect(() => {
+        if (initialDefaultValueRef.current !== defaultValue) {
+            setRenderCount(r => r + 1);
+            parse
+                ? field.onChange(parse(defaultValue))
+                : field.onChange(defaultValue);
+            initialDefaultValueRef.current = defaultValue;
+        }
+    }, [defaultValue, setRenderCount, parse, field]);
+
+    const { onBlur: onBlurFromField } = field;
+    const hasFocus = React.useRef(false);
+
+    // update the input text when the user types in the input
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        if (onChange) {
+            onChange(event);
+        }
+        if (
+            typeof event.target === 'undefined' ||
+            typeof event.target.value === 'undefined'
+        ) {
+            return;
+        }
+        const target = event.target;
+
+        const newValue =
+            target.valueAsDate !== undefined &&
+            target.valueAsDate !== null &&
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -35,42 +35,120 @@ export const DateInput = ({
...
+        if (initialDateValue !== fieldDateValue) {
+            setRenderCount(r => r + 1);
+            parse
+                ? field.onChange(parse(field.value))
+                : field.onChange(field.value);
+            initialDefaultValueRef.current = field.value;
+        }
+    }, [setRenderCount, parse, field]);
+
+    const { onBlur: onBlurFromField } = field;
```
slax57, line 83: Why have a different logic for `onBlur` and `onFocus`?

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -35,42 +35,120 @@ export const DateInput = ({
...
+    };
 
-    const { error, invalid, isTouched } = fieldState;
-    const { isSubmitted } = formState;
     const renderHelperText =
         helperText !== false || ((isTouched || isSubmitted) && invalid);
 
+    const { ref, name } = field;
+
+    console.log('field', field);
```
slax57, line 138: `console.log` spotted

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -133,7 +211,8 @@ const getStringFromDate = (value: string | Date) => {
     // null, undefined and empty string values should not go through dateFormatter
     // otherwise, it returns undefined and will make the input an uncontrolled one.
     if (value == null || value === '') {
-        return '';
+        console.log('toto');
```
slax57: `console.log` spotted

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -35,42 +35,120 @@ export const DateInput = ({
...
         helperText !== false || ((isTouched || isSubmitted) && invalid);
 
+    const { ref, name } = field;
+
+    console.log('field', field);
+
     return (
         <TextField
             id={id}
-            {...field}
```
slax57, line 71: Why remove `{...field}` ?

djhi (changes requested):

slax57 (changes requested):

slax57:
I'm afraid you brought a regression to the `DateInput` component:


https://github.com/marmelab/react-admin/assets/14542336/73693c31-34de-4eab-84df-2f32b729542e

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: Because the input is fully controlled locally now

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: react-hook-form only handle `onBlur` not `onFocus` and it will call any custom `onBlur` provided

PR #9544, opened 2023-12-22
Make create-react-admin usable in non-interactive mode

djhi:
## Problem 

The `create-react-admin` currently requires user to make choices interactively, preventing automated usages

## Solution

Add flags for all parameters


fzaninotto:
LGTM. Would you mind pulling against `next` rather than `master`?

djhi:
> LGTM. Would you mind pulling against `next` rather than `master`?

Done

fzaninotto (approved):

PR #9553, opened 2024-01-03
Fix ResettableTextField size prop can't be overridden

### LINKED ISSUE #9546: TextField "size" prop is not overrideable by theme
**What you were expecting:**
A theme override should be able to set the `size` prop on the TextInputs, given that it's a style prop that doesn't depend on anything else.

**What happened instead:**

Setting the `size` prop's default on the theme to "medium" doesn't actually change any inputs; they need to be changed on a one-by-one basis by passing the prop `size="medium"` to ALL inputs if we want a global override.

**Steps to reproduce:**

1. Creat a theme with the following definition:
```ts
const theme = {
  components: {
    MuiTextField: {
      defaultProps: {
        size: 'medium',
      },
  },
};
```
2. Apply the theme
3. Create a `TextInput` somewhere

**Related code:**

Relevant code: 
```ts
// ra-ui-materialui/src/input/ResettableTextField lines 155-174
<StyledTextField
    value={value}
    InputProps={{
        classes:
            props.select && variant === 'filled'
                ? { adornedEnd: inputAdornedEnd }
                : {},
        endAdornment: getEndAdornment(),
        ...InputPropsWithoutEndAdornment,
    }}
    disabled={disabled}
    variant={variant}
    margin={margin}
    className={className}
    // HERE
    size="small"
    // ^
    {...rest}
    onFocus={handleFocus}
    onBlur={handleBlur}
    inputRef={ref}
/>
```

**Other information:**

This can be easily fixed by removing the prop from here and applying it to RA's default theme as a defaultProp instead; this way, we can override it ourselves if needed.

**Environment**

* React-admin version: 4.16.4

WiXSL:
Fixes #9546

### DIFF
--- a/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
+++ b/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
@@ -26,7 +26,6 @@ export const ResettableTextField = forwardRef(
             variant,
             margin,
             className,
-            size = 'small',
             ...rest
         } = props;
 
--- a/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
+++ b/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
@@ -26,6 +26,7 @@ export const ResettableTextField = forwardRef(
             variant,
             margin,
             className,
+            size = 'small',
             ...rest
         } = props;
 
@@ -166,7 +167,7 @@ export const ResettableTextField = forwardRef(
                 variant={variant}
                 margin={margin}
                 className={className}
-                size="small"
+                size={size}
                 {...rest}
                 onFocus={handleFocus}
                 onBlur={handleBlur}
--- a/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
+++ b/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
@@ -166,7 +166,6 @@ export const ResettableTextField = forwardRef(
                 variant={variant}
                 margin={margin}
                 className={className}
-                size={size}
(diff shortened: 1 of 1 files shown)


PedroPerpetua:
This doesn't fix #9546 - the Textfield size should be moved to the themeing, not props; in fact, you can already override it with props - this PR effectively does nothing.

packages/ra-ui-materialui/src/input/ResettableTextField.tsx
```diff
@@ -26,6 +26,7 @@ export const ResettableTextField = forwardRef(
             variant,
             margin,
             className,
+            size = 'small',
```
fzaninotto, line 29: I think this doesn't fix the issue, as the underlying MUI TextField will still receive `size="small"`, while we want it to receive nothing so it ca use the theme override.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/input/ResettableTextField.tsx
WiXSL: You are right.
I'll change it

fzaninotto:
@WiXSL The author of the issue has also opened a PR to fix it (see #9546). Both your PRs are identical, so it's hard to choose the keeper. For this time, we'll keep the other one and close yours, but it's nothing personal, we still love your contributions!

WiXSL:
That's great!
The important ting is to maintain the project stable for everyone, I don't mind If are PR's made by me or not, haha

PR #9554, opened 2024-01-03
Fix hardcoded TextField size prop

### LINKED ISSUE #9546: TextField "size" prop is not overrideable by theme

PedroPerpetua:
This fixes the issue the issue with the TextField size prop not being override-able by theme. RA's default theme actually already sets the MuiTextField size prop default value as "small", so it's redundant to have it here.

However, in a possibly unrelated note, it may be benefitial to extend RA's documentation and themeing options with a re-export of [MUI's `createTheme`](https://mui.com/material-ui/customization/theming/#createtheme-options-args-theme) function, that allows the deep-merging of themes, or maybe recommend [`lodash.merge`](https://www.npmjs.com/package/lodash.merge) to set up themes.

With this PR, overriding the theme's TextField variant as per [RA's docs](https://marmelab.com/react-admin/AppTheme.html#overriding-default-props) will result in the `size="small"` property to be lost.

Closes #9546


packages/ra-ui-materialui/src/input/ResettableTextField.tsx
```diff
@@ -166,7 +166,6 @@ export const ResettableTextField = forwardRef(
                 variant={variant}
                 margin={margin}
                 className={className}
-                size="small"
```
djhi, line 169: This is a BC. Can you extract the small prop and set its default value to `"small"` then pass it here? That should solve the issue as well

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/input/ResettableTextField.tsx
PedroPerpetua: "BC"? Sorry I don't follow.

But this wouldn't resolve the issue.
1. The prop is already passed with the "rest" props that follow immediately
2. Setting the default to "small" in the component would prevent it from being overrideable by the theme (RA's default theme already sets this default to "small", so we don't need to define it at a component level)

djhi: I forgot about the theme default. Thanks for the reminder :) LGTM then but I'll wait for other core members opinion

djhi (approved):

adguernier (approved):
I agree too since RA's `defaultTheme` sets the `MuiTextField` `defaultProps` to "small

fzaninotto (changes requested):
This is the right code, thanks! However we need to document this change in the upgrade guide. People with a custom theme will see their inputs size change from small to medium, and this is a regression. We also need to document the custom size somewhere in the Inputs documentation.

We usually put the doc and the code in the same PR, would you mind doing these changes in your PR?

PedroPerpetua:
Okay, I've updated the documentation, and took the liberty of changing the theme configurations to use `deepmerge` and `createTheme` from MUI instead of object deconstruction, as suggested by [MUI's docs](https://mui.com/material-ui/customization/theming/#createtheme-options-args-theme). I find this to be a much better way for users to set up their own themes when basing themselves on the default themes, and this easily solves the issue of overriding a couple deep properties of the theme.

For example, using deepmerge, you can easily override _just_ the `variant` prop in a `TextField` while keeping the `size` prop the same.

It's a "big breaking change" (old way would still work). I hope this is fine.

packages/ra-ui-materialui/src/theme/ThemeProvider.tsx
```diff
@@ -46,11 +46,7 @@ export const ThemeProvider = ({
     const themeValue = useMemo(() => {
         try {
             return createTheme(
-                typeof mode === 'object'
-                    ? mode // FIXME: legacy useTheme, to be removed in v5
```
PedroPerpetua, line 50: I also went ahead and removed this legacy code, as suggested by the comment left here

packages/ra-ui-materialui/src/theme/defaultTheme.ts
```diff
@@ -46,45 +48,46 @@ const defaultThemeInvariants = {
...
-    palette: {
-        background: {
-            default: '#fafafb',
-        },
-        secondary: {
-            light: '#6ec6ff',
-            main: '#2196f3',
-            dark: '#0069c0',
-            contrastText: '#fff',
+export const defaultLightTheme: RaThemeOptions = createTheme(
```
fzaninotto, line 51: you shouldn't call `createTheme` here as our theme provider does it, too.

packages/ra-ui-materialui/src/theme/defaultTheme.ts
```diff
@@ -46,45 +48,46 @@ const defaultThemeInvariants = {
...
+    deepmerge(defaultThemeInvariants, {
```
fzaninotto, line 52: We don't need the deepMerge here, as we know what we're merging. I prefer the old version.

packages/ra-ui-materialui/src/theme/ThemeProvider.tsx
fzaninotto: Then you have to add a section in the Upgrade guide to explain what to do for apps using the old useTheme value.

fzaninotto (changes requested):
You're still missing a section in the v5 upgrade guide for people with a custom theme in v4.

PedroPerpetua:
Alright, I'll give the review a spin, hopefully monday.

packages/ra-ui-materialui/src/theme/defaultTheme.ts
PedroPerpetua: deepMerge here allows new changes to be made more easily in the future, in my opinion. Imagine, as a crude example, you want the darkTheme to have their textfields with a width of 50% by default.

To do this change without using deepmerge:
```diff
const defaultDarkTheme: RaThemeOptions = {
    ...defaultThemeInvariants,
    palette: {
        mode: 'dark',
        primary: {
            main: '#90caf9',
        },
        background: {
            default: '#313131',
        },
    },
+    components: {
+        ...defaultThemeInvariants.components,
+        MuiTextField: {
+            ...defaultThemeInvariants.components.MuiTextField,
+            defaultProps: {
+                ...defaultThemeInvariants.components.MuiTextField.defaultProps,
+                sx: {
                    /* 
                    If you later add a default sx to defaultThemeInvariants,
                    you need to remember to come here and add the following line:

+                    ...defaultThemeInvariants.components.MuiTextField.defaultProps.sx,
                    
                    */
+                    width: "50%"
+                }
+            },
+        },
+    }
}
```
With deepmerge:
```diff
export const defaultDarkTheme: RaThemeOptions = deepmerge(defaultThemeInvariants, {
        palette: {
            mode: 'dark',
            primary: {
                main: '#90caf9',
            },
            background: {
                default: '#313131',
            },
        },
+        components: {
+            MuiTextField: {
+                defaultProps: {
+                    sx: {
+                        width: "50%"
+                    }
+                }
+            }
+        }
    })
);
```

The retro-active update mentioned in the comment being specially worrying in my opinion.

PedroPerpetua:
Okay, I've added the useTheme changes to the migration guide; I also went ahead and removed the legacy code on `useTheme` and the `toggleThemeButton` fully; it's a bit out of the original scope of this PR, but given that the original issue was a 1 line bugfix, I figured they can go along and the PR can be renamed as you find appropriate.

Regarding this:
> We don't need the deepMerge here, as we know what we're merging. I prefer the old version.

@fzaninotto 

I left a response to your comment; waiting on a reply to know if I should revert the changes there or keep them.

PedroPerpetua:
BUMPing this since I'm still awaiting on a reply from @fzaninotto to know if it needs more changes or is ready 😉

docs/AppTheme.md
```diff
@@ -425,15 +423,14 @@ const App = () => (
 );
 ```
 
-You can write a custom theme from scratch, or start from the [default theme](#default-theme) and override some values:
+You can write a custom theme from scratch, or start from the [default theme](#default-theme) and override some values, using the Material UI's utility function deepmerge:
```
fzaninotto, line 426:
```suggestion
You can write a custom theme from scratch, or start from the [default theme](#default-theme) and override some values, using Material UI's utility function deepmerge:
```

docs/AppTheme.md
```diff
@@ -499,46 +515,48 @@ const defaultThemeInvariants = {
...
-    palette: {
-        background: {
-            default: '#fafafb',
-        },
-        secondary: {
-            light: '#6ec6ff',
-            main: '#2196f3',
-            dark: '#0069c0',
-            contrastText: '#fff',
+export const defaultLightTheme: RaThemeOptions = createTheme(
```
fzaninotto, line 518: you should not leave createTheme here

docs/AppTheme.md
```diff
@@ -499,46 +515,48 @@ const defaultThemeInvariants = {
...
 
-export const defaultDarkTheme: RaThemeOptions = {
-    palette: {
-        mode: 'dark',
-        primary: {
-            main: '#90caf9',
-        },
-        background: {
-            default: '#313131',
+export const defaultDarkTheme: RaThemeOptions = createTheme(
```
fzaninotto, line 546: same

packages/ra-ui-materialui/src/button/ToggleThemeButton.tsx
```diff
@@ -31,16 +30,6 @@ export const ToggleThemeButton = (props: ToggleThemeButtonProps) => {
         defaultTheme || (prefersDarkMode && darkTheme ? 'dark' : 'light')
     );
 
-    // FIXME: remove in v5
```
fzaninotto, line 34: It's a breaking change: You need to mention this in the upgrdae guide.

fzaninotto (changes requested):
We're getting close to the merge!

PedroPerpetua:
Fixed the latest comments and the merge conflits

fzaninotto (approved):

fzaninotto:
Thanks a lot!

PR #9555, opened 2024-01-04
Fix `<MenuItemLink>` type by omitting `placeholder` prop

### LINKED ISSUE #9507: [v4 typescript] `<MenuItemLink>` seems now have missing placeholder prop
**What you were expecting:**
I expect when I follow the tutorial there would be no type error https://marmelab.com/react-admin/Buttons.html#menuitemlink

**What happened instead:**
There is a type related error

```
Property 'placeholder' is missing in type '{ to: string; primaryText: string; }' but required in type 'Pick<MenuItemLinkProps, "replace" | "slot" | "style" | "title" | "children" | "role" | "action" | "about" | "security" | "primaryText" | "id" | "type" | "key" | ... 275 more ... | "tooltipProps">'.(2741)
```

**Steps to reproduce:**
Go to the example below
Run $ yarn type-check and look for errors in the src/Layout.tsx file

**Related code:**
https://stackblitz.com/edit/github-cubnvw?file=src%2FLayout.tsx

```
```

**Other information:**

**Environment**

* React-admin version: 4.16.2
* React version: 18.2.0
* Browser: Safari

smeng9:
With this merge request https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67170 introduced into @types/react 18.2.43, `placeholder` prop is no longer in `HTMLAttributes`.

Somehow since 18.2.43 the `<MenuItemLink>` requires us to pass the `placeholder` prop to the `<a>` tag below, where there are no such prop defined in [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)

Closes #9507

### DIFF
--- a/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
+++ b/packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
@@ -140,19 +140,16 @@ export const MenuItemLink = forwardRef<any, MenuItemLinkProps>((props, ref) => {
     );
 });
 
-export type MenuItemLinkProps = Omit<
-    LinkProps &
-        MenuItemProps<'li'> & {
-            leftIcon?: ReactElement;
-            primaryText?: ReactNode;
-            /**
-             * @deprecated
-             */
-            sidebarIsOpen?: boolean;
-            tooltipProps?: TooltipProps;
-        },
-    'placeholder'
->;
+export type MenuItemLinkProps = LinkProps &
+    Omit<MenuItemProps<'li'>, 'placeholder'> & {
+        leftIcon?: ReactElement;
+        primaryText?: ReactNode;
+        /**
+         * @deprecated
+         */
+        sidebarIsOpen?: boolean;
+        tooltipProps?: TooltipProps;
+    };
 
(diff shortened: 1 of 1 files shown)


packages/ra-ui-materialui/src/layout/MenuItemLink.tsx
```diff
@@ -140,16 +140,19 @@ export const MenuItemLink = forwardRef<any, MenuItemLinkProps>((props, ref) => {
...
-    MenuItemProps<'li'> & {
-        leftIcon?: ReactElement;
-        primaryText?: ReactNode;
-        /**
-         * @deprecated
-         */
-        sidebarIsOpen?: boolean;
-        tooltipProps?: TooltipProps;
-    };
+export type MenuItemLinkProps = Omit<
```
fzaninotto, line 143: I think you should put the Omit on the MenuItemProps, not the larger union.

fzaninotto (changes requested):

fzaninotto (approved):

djhi (approved):

djhi:
Thanks!

PR #9556, opened 2024-01-04
Explain how to customize notification message in form validation

adguernier:
## Problem

The documentation does not explain how to customize the notification message with form server-side validation.

## Solution

Explain how to customize this message in form server-side validation documentation.

### DIFF
--- a/docs/Validation.md
+++ b/docs/Validation.md
@@ -370,6 +370,7 @@ Server-side validation is supported out of the box for `pessimistic` mode only. 
 ```
 {
     body: {
+        message: 'Error: my custom message in notification popover',
         errors: {
             title: 'An article with this title already exists. The title must be unique.',
             date: 'The date is required',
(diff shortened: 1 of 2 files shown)


fzaninotto (approved):

docs/Validation.md
```diff
@@ -370,6 +370,7 @@ Server-side validation is supported out of the box for `pessimistic` mode only.
 ```
 {
     body: {
+        message: 'Error: my custom message in notification popover',
```
slax57, line 373: I'm confused. Shouldn't the message be outside the body? :thinking: 
See https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/edit/useEditController.ts#L189 for ref

docs/Validation.md
```diff
@@ -383,7 +384,9 @@ Server-side validation is supported out of the box for `pessimistic` mode only.
 
 **Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key.
 
-**Tip**: If your data provider leverages React Admin's [`httpClient`](https://marmelab.com/react-admin/DataProviderWriting.html#example-rest-implementation), all error response bodies are wrapped and thrown as `HttpError`. This means your API only needs to return an invalid response with a json body containing the `errors` key.
+Note: you can use the `message` key to customize the message displayed in the notification popover but it is not mandatory. By default the notification message is the translatable key `ra.notification.http_error`.
```
slax57, line 387: I thought this only worked with React 17. Should we really mention this in the docs then?

packages/ra-ui-materialui/src/form/SimpleForm.stories.tsx
```diff
@@ -145,6 +147,36 @@ export const GlobalValidation = () => (
     </Wrapper>
 );
 
+export const CustomNotificationMessage = () => (
```
slax57, line 150: IMO this story is misplaced. The logic responsible for this feature is actually implemented in `useEditController`, not in `SimpleForm`.

slax57 (changes requested):

slax57:
The solution documented in this PR doesn't work, and what's more it relies on a race condition between the [http_error notification](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/controller/edit/useEditController.ts#L186) and the [invalid_form notification](https://github.com/marmelab/react-admin/blob/master/packages/ra-core/src/form/useNotifyIsFormInvalid.ts#L29), so it's not robust.

Solving this problem can only be addressed by introducing a new field in the server's response, and supporting the ability to return `message: false` to disable the http_error notification, so this will be done in the upcoming v5.

PR #9558, opened 2024-01-04
style: narrowing ReactQueryDevtools button

adguernier:
## Problem
The ReactQueryDevtools button is too big

image


examples/simple/assets/app.css
```diff
@@ -0,0 +1,4 @@
+.tsqd-transitions-container>div {
```
fzaninotto, line 1: Could you add a comment explaining why you add this rule?

fzaninotto (changes requested):

djhi (approved):

PR #9561, opened 2024-01-05
Follow up fix of `<MenuItemLink>` `placeholder` prop type

smeng9:
Hi @fzaninotto,

After applying your review from https://github.com/marmelab/react-admin/pull/9555 , I built and tested it locally with `yarn link` to my own project. Omitting from just `MenuItemProps` won't fix the problem. 

It seems we have to omit the `placeholder` from the union of `LinkProps` and `MenuItemProps`. Both of them seems has some issue require a non-existent `placeholder`.


fzaninotto (approved):

slax57 (approved):
Thanks!

PR #9564, opened 2024-01-05
[Doc] add a section demonstrating the `<Datagrid>` headers pinning feature 

adguernier:
## Problem
`<Datagrid>` support headers pinning by default but it is not mentioned anywhere.

## Solution

Mention headers pinning support in the <Datagrid> documentation and add a screencast


slax57 (approved):

docs/Datagrid.md
```diff
@@ -832,6 +832,15 @@ const PostList = () => (
 
 **Tip**: The `Datagrid` component `classes` can also be customized for all instances of the component with its global css name `"RaDatagrid"` as [describe here](https://marmelab.com/blog/2019/12/18/react-admin-3-1.html#theme-overrides)
 
+## Headers pinning
```
fzaninotto, line 835:
```suggestion
## Header Pinning
```

docs/Datagrid.md
```diff
@@ -832,6 +832,15 @@ const PostList = () => (
...
+
+`<Datagrid>` pins headers by default and you don't need to do anything extra for it to work:
```
fzaninotto, line 837:
```suggestion
`<Datagrid>` has sticky headers, which means that the header row will remain visible even when scrolling down the page.
```

fzaninotto: Here I'm using several names for the same feature (sticky headers, header pinning) to make it more discoverable

docs/Datagrid.md
```diff
@@ -832,6 +832,15 @@ const PostList = () => (
...
+
+<video controls autoplay playsinline muted loop>
+  <source src="./img/datagrid-headers-pinning.mp4" type="video/mp4"/>
+  Your browser does not support the video tag.
+</video>
+
```
fzaninotto, line 843:
```suggestion

You don't need to do anything for this to work, as it's enabled by default. 
```

fzaninotto (changes requested):

docs/Datagrid.md
```diff
@@ -832,15 +832,16 @@ const PostList = () => (
...
 
-`<Datagrid>` pins headers by default and you don't need to do anything extra for it to work:
+`<Datagrid>` has sticky headers, which means that the header row will remain visible even when scrolling down the page.
 
 <video controls autoplay playsinline muted loop>
   <source src="./img/datagrid-headers-pinning.mp4" type="video/mp4"/>
   Your browser does not support the video tag.
 </video>
 
+You don't need to do anything for this to work, as it's enabled by default.
```
djhi, line 844:
```suggestion
You don't need to do anything for this to work, as it's enabled by default. 

```

djhi (changes requested):

fzaninotto (approved):

docs/Datagrid.md
adguernier: I don't understand this suggestion because it seems to do the same thing as before :thinking:

djhi (approved):

Issue #9516, opened 2023-12-12
[v5] Stop conflating IDs with 'id' columns


Dreamsorcerer:
There are 2 concepts that react-admin is mixing up (I'll use these terms below to differentiate between them):

- ID: A unique ID that can be used to identify a record within react-admin and the API.
- 'id' column: A column/attribute of a row/record in the database/storage with the name 'id'.

This design decision causes all sorts of problems when trying to handle edge cases reliably. I would hugely appreciate, if there is a v5 in the future (I do see a milestone exists), that these 2 concepts are separated. Throughout the development of the aiohttp-admin library, these are some of the issues I've encountered that are caused by this (the ones I can recall today):

- A resource can't have an 'id' column if it's not the _sole_ primary key. It would be incredibly hacky to try and make this work, by internally renaming the column and remembering to change the label etc. anytime it's displayed. It's already hacky enough trying to get it to work with all the other issues, so I'm not even going to attempt this. The library currently just emits a warning and fails badly if a user provides such a model.
- An ID should be uniform across the library, so I use strings to represent IDs. This then means we need a bunch of hacks, as when an 'id' column is not a string, I need to remember to convert between string and the native type throughout the library at numerous locations. It has been very fiddly and required a lot of regression tests to ensure that we are getting the correct type at various points in the library.
- When a resource does _not_ have an 'id' column, we then again need a bunch of hacks to remember to remove the column from data at various points when submitted to the API by react-admin.
- Even with these hacks, there are still more edge cases that I keep finding where the behaviour is wrong. For example, the export button grabs the data from getMany() so it ends up exporting rows that include the ID. The ID is meant to be internal to react-admin and doesn't exist in the data, therefore it should not appear in an export. The API doesn't know if the call is for export or not, so it can't drop the ID there.

I'm sure you can come up with solutions, but as a 1 possible approach, a resource could be represented something like:
```
{
  'id': ...  // The ID of the react-admin resource.
  'data': {
    'id': ... // The 'id' column of the data, which may or may not be the same, and may not exist at all.
    ...
  }
}
```

If such a change were made, it would make react-admin a lot more reliable when trying to use it with arbitrary data models that haven't been pre-designed for react-admin.

fzaninotto:
Thanks for your feedback.

Configurable ID for resources is indeed a common request, but it adds a tremendous amount of complexity everywhere in react-admin. We don't want this added complexity, and since there are workarounds, we don't want to add support for it. 

[...]

PR #9566, opened 2024-01-06
Add lookupSource to ReferenceInput

Dreamsorcerer:
This is a proposal from #9516. @fzaninotto

Let me know if this looks reasonable, and then I'll take a look at tests/docs.

The basic idea of lookupSource (or maybe idSource?) is to use a different source for the ID used in the getMany call.
ReferenceInput will continue to use source when actually updating the value, but will use lookupSource to fetch the record for repr.
lookupSource defaults to source, so should be fully backwards-compatible.

--------------

This will allow using a resource structure that can support composite keys, PKs not named 'id' etc.

Taking the example from that conversation:
```
Book = {
  id: str
  data: {
    ...
    authorId: int  // FK in DB
  }
  fk_authorId: str
}
```

We would then use `<ReferenceInput source="data.authorId" lookupSource="fk_authorId" {...}>`.
This would then lookup the current record using the normalised fk_authorId, but when the selection is changed in the input it is the data.authorId attribute that would be updated (from the int value).

### DIFF
--- a/packages/ra-core/src/controller/input/useReferenceInputController.ts
+++ b/packages/ra-core/src/controller/input/useReferenceInputController.ts
@@ -58,6 +58,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         queryOptions = {},
         reference,
         source,
+        lookupSource = source
     } = props;
     const { meta, ...otherQueryOptions } = queryOptions;
 
@@ -72,6 +73,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
 
     // selection logic
     const currentValue = useWatch({ name: source });
+    const lookupCurrentValue = useWatch({ name: lookupSource });
 
     const isGetMatchingEnabled = enableGetChoices
         ? enableGetChoices(params.filterValues)
@@ -112,7 +114,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         isLoading: referenceLoading,
         isFetching: referenceFetching,
     } = useReference<RecordType>({
-        id: currentValue,
+        id: lookupCurrentValue,
         reference,
         // @ts-ignore the types of the queryOptions for the getMAny and getList are not compatible
         options: {
@@ -210,5 +212,6 @@ export interface UseReferenceInputControllerParams<
     resource?: string;
     sort?: SortPayload;
     source: string;
+    lookupSource?: string;
     enableGetChoices?: (filters: any) => boolean;
 }


fzaninotto:
In my opinion, this is a highly unusual use case: you want an input that *reads* from a field, but *writes* into another field. I don't want to clutter the react-admin API to support such a corner case.

In addition, you've found ho to build your own alternative `ReferenceInput`, so it is already possible in user land.  

So I'm -1 for this change. I'll let other core team members voice their opinion.

Dreamsorcerer:
> In my opinion, this is a highly unusual use case: you want an input that _reads_ from a field, but _writes_ into another field. I don't want to clutter the react-admin API to support such a corner case.

Well, that is what I described in https://github.com/marmelab/react-admin/issues/9516#issuecomment-1874748023. If we can get ReferenceInput working like this, then my solution can be generally applied to any kind of data layout. Essentially, the actual DB record is under `record.data`, anything outside of that is basically just metadata for react-admin to work with, but will be ignored by the API on input and generated automatically on output.

I'm just trying to figure out a general solution that should work in almost all use cases, which I'd have expected react-admin to want to achieve too.

> In addition, you've found ho to build your own alternative `ReferenceInput`, so it is already possible in user land.

Only by copying the entirety of ReferenceInput and useReferenceInputController(). That's not exactly a maintainable solution, unless I'm missing something?


I'm still looking at some other details, so I may have a different proposal next week anyway.

fzaninotto:
> Essentially, the actual DB record is under `record.data`

Then why don't you use `<ReferenceInput source="data.authorId" />`?

> a general solution that should work in almost all use cases, which I'd have expected react-admin to want to achieve too

No, we only target the most common use cases, and get out of the way to let users do the remaining

Dreamsorcerer:
> > Essentially, the actual DB record is under `record.data`
> 
> Then why don't you use `<ReferenceInput source="data.authorId" />`?

Because that's not the ID for the API. To make things uniform and avoid lots of complexity in the APIs, the ID (i.e. `record.id`) is a string, whereas the data attributes could be arbitrary types. The string ID may also be something that identifies a record from a composite key (where obviously there are multiple separate attributes in `record.data`).

Dreamsorcerer:
Going back to my first idea of having a callback that can set the other attributes on modification, I've managed to get it working with no modifications to react-admin now:
```
const _ReferenceInput = (props) => {
    const {referenceKeys, ...innerProps} = props;
    const {setValue} = useFormContext();
    const change = (value, record) => {
        for (let [this_k, foreign_k] of referenceKeys)
            setValue(`data.${this_k}`, record ? record.data[foreign_k] : null);
    };

    return (
        <ReferenceInput {...innerProps}>
            <AutocompleteInput onChange={change} />
        </ReferenceInput>
    );
};
```
referenceKeys is then used to specify the extra keys that should be updated when this input is changed. So, using the same example from above, it'd be used like (setting book.data.authorId to author.data.Id on change):
```
<_ReferenceInput source="fk_authorId" referenceKeys={[["authorId", "id"]]} />
```

This seems like a good solution to me, but let me know if it looks like something could be better.

PR #9567, opened 2024-01-07
Fix click on `<Datagrid>` Select All button bubbles up to parent Datagrid

Dreamsorcerer:
This adds stopPropagation() to the entire DataGrid, to avoid propagation when clicking anywhere in the grid.

Currently, clicking the select all checkbox or an empty space  on the header still triggers the click on the parent element (e.g. nest 2 DataGrids and the select all button will trigger the rowClick of the parent DataGrid).

If this approach seems too much, then we could move this line to the checkbox in DataGridHeader.


fzaninotto:
This approach is too much, especially since it means to support a feature (nested datagrids) that we don't want to support.

fzaninotto:
Yes, the checkbox can stop propagation

Dreamsorcerer:
Moved to Select all checkbox.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9577, opened 2024-01-11
Feat : withLifecycleCallbacks support for wilcard * and array of callbacks + refacto

quentin-decre:
# Problem

Using withLifecycleCallbacks when you have many callbacks can be very verbose and hard to read / maintain. This can be improved a lot with two features :

## 1. Support for a wildcard resource : *
Imagine for example your API returns "identifier" and you want to rename it to "id" to make it work with react-admin. You will need to create a callback for each resource. But now you can do that :
```typescript
const dataProvider = withLifecycleCallbacks(
  jsonServerProvider('http://localhost:3000'), [
  {
    resource: '*',
    afterRead: async (data, { dataProvider, resource }) => {
      // rename field to the record
      data.id = data.identifier
      return data
    },
  },
])
```

## 2. Support for arrays of callbacks

When you have a lot of callbacks / hooks (which is often the case when having a graphql API), you may want to have one callback per purpose, and name it after that purpose. So you could have many callbacks for the same resource / hook. This is already possible but quite verbose / difficult to read.
The idea is to allow to pass an array of callbacks, while still supporting passing directly one callback.

Currently :
```typescript
      {
        resource: '*',
        beforeGetList: defaultSortById,
      },
      {
        resource: '*',
        beforeGetList: addNullsLastToSortOrders,
      },
      {
        resource: '*',
        beforeGetList: convertFilters,
      },
```

After :
```typescript
      {
        resource: '*',
        beforeGetList: [defaultSortById, addNullsLastToSortOrders, convertFilters],
      },
```


djhi:
This looks great :+1:  I like that it encourages good practices and better readability. Would you mind adding some tests and updating the documentation?

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -67,13 +67,13 @@ import {
  *   [
  *     {
  *       resource: "posts",
- *       afterRead: async (data, dataProvider) => {
+ *       afterRead: async (data, {dataProvider, resource}) => {
```
fzaninotto, line 70: This is a breaking change. You can add the same feature in a backward compatible way, by adding a third parameter instead.

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -290,42 +268,23 @@ export const withLifecycleCallbacks = (
...
-            let result = await dataProvider.update<RecordType>(
+            newParams = await applyCallbacks(
                 resource,
+                'beforeUpdate',
                 newParams
             );
-
-            const afterUpdateHandlers = handlers.filter(
-                h => h.resource === resource && h.afterUpdate
+            newParams = await applyCallbacks(resource, 'beforeSave', newParams);
```
fzaninotto, line 277: This is another breaking change: beforeSave used to be applied on params.data, not on params

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -335,42 +294,19 @@ export const withLifecycleCallbacks = (
...
-
-            const afterCreateHandlers = handlers.filter(
-                h => h.resource === resource && h.afterCreate
-            );
-            for (let handler of afterCreateHandlers) {
-                result = await handler.afterCreate(result, dataProvider);
-            }
-            const afterSaveHandlers = handlers.filter(
-                h => h.resource === resource && h.afterSave
+            newParams = await applyCallbacks(resource, 'beforeSave', newParams);
```
fzaninotto, line 303: same

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -462,107 +373,88 @@ export const withLifecycleCallbacks = (
...
-        params: UpdateParams<T>,
-        dataProvider: DataProvider
-    ) => Promise<UpdateParams<T>>;
-    beforeUpdateMany?: (
-        params: UpdateManyParams<T>,
-        dataProvider: DataProvider
-    ) => Promise<UpdateManyParams<T>>;
+    afterCreate?:
+        | ResourceCallback<CreateResult<T>>[]
+        | ResourceCallback<CreateResult<T>>;
```
fzaninotto, line 406: You should create a type for this union, to avoid repetition

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -123,6 +123,40 @@ export const withLifecycleCallbacks = (
...
     handlers: ResourceCallbacks[]
 ): DataProvider => {
+    /**
+     * Apply callbacks to the params for the given resource and hook
+     * @param {string} resource The resource name
+     * @param {string} hook The hook name (beforeGetList, afterGetOne, etc.)
+     * @param {U} params The params / result to pass to the callbacks
+     * @returns {Promise<U>} The params / result after the callbacks have been applied
+     */
+    const applyCallbacks = async function <U>(
```
fzaninotto, line 133: I suggest you extract that function, and unit test it.

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -574,20 +466,48 @@ export type ResourceCallbacks<T extends RaRecord = any> = {
...
      */
-    afterSave?: (record: T, dataProvider: DataProvider) => Promise<T>;
+    afterSave?: ((record: T, utils: ResourceCallbackUtils) => Promise<T>)[];
 };
+
+/**
+ * The list of all possible data provider hooks
+ * @see https://marmelab.com/react-admin/DataProviders.html#data-provider-hooks
+ */
+export const dataProviderHooks = [
```
fzaninotto, line 491: Why do you export it?

fzaninotto (changes requested):
As this is a new feature, it must be PRed against `next` rather than `master`

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: changed

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: changed

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: I use it in my own code to apply a callback for every possible hook (for login purpose essentially).
I understand this could be in my own code though.

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: changed

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: changed

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
quentin-decre: If I extract this function, I see two options :
1. either we have to pass the handlers + dataProvider each time, the drawback is that it's quite verbose and the code is less clear :
```javascript
newParams = await applyCallbacks(
                dataProvider,
                handlers,
                resource,
                'beforeGetList',
                newParams
            );
```
2. Or I could bind the the repeated parameters :
```typescript
    const applyCallbacksWithContext = applyCallbacks.bind(
        null,
        dataProvider,
        handlers
    );

....

newParams = await applyCallbacksWithContext(
                resource,
                'beforeGetList',
                newParams
            );

```

I tend to prefer the first option because the code is less complex to understand even though it's very verbose.

What is your opinion on that ?

quentin-decre:
Regarding the documentation, I hesitated between added the new features in the main part of the documentation, or add an "advanced use" section, and explain the new possibilities, and how they are very useful, in which cases, best practices, .... 
Also add some details about execution order and the possibility to add the same resource multiple times (it was already possible, but not very clear for me in the documentation).
What is your opinion on that ?

Can you tell me if the implementation is OK for you ? If yes, I will move on unit testing :)

packages/ra-core/src/dataProvider/withLifecycleCallbacks.ts
```diff
@@ -574,20 +542,45 @@ export type ResourceCallbacks<T extends RaRecord = any> = {
...
+    beforeGetList = 'beforeGetList',
+    beforeGetMany = 'beforeGetMany',
+    beforeGetManyReference = 'beforeGetManyReference',
+    beforeGetOne = 'beforeGetOne',
+    beforeUpdate = 'beforeUpdate',
+    beforeUpdateMany = 'beforeUpdateMany',
+    beforeSave = 'beforeSave',
+    afterRead = 'afterRead',
+    afterSave = 'afterSave',
+}
```
djhi, line 586: As this is not used anywhere, I believe we should remove it

djhi (commented):
Thanks!

> Regarding the documentation, I hesitated between added the new features in the main part of the documentation, or add an "advanced use" section, and explain the new possibilities, and how they are very useful, in which cases, best practices, ....

I would probably add this in the **Code Organization** section.
 
>  Also add some details about execution order and the possibility to add the same resource multiple times (it was already possible, but not very clear for me in the documentation).

This is definitely worth adding in the **Usage** section

quentin-decre:
I have fixed + added some unit tests for the new features

quentin-decre:
> Thanks!
> 
> > Regarding the documentation, I hesitated between added the new features in the main part of the documentation, or add an "advanced use" section, and explain the new possibilities, and how they are very useful, in which cases, best practices, ....
> 
> I would probably add this in the **Code Organization** section.
> 
> > Also add some details about execution order and the possibility to add the same resource multiple times (it was already possible, but not very clear for me in the documentation).
> 
> This is definitely worth adding in the **Usage** section

agreed, did that :)

djhi (changes requested):
Can you please remove the `package-lock.json` file? Besides, I don't get why there are changes in the `yarn.lock` file.

quentin-decre:
> Can you please remove the `package-lock.json` file? Besides, I don't get why there are changes in the `yarn.lock` file.

deleted + reverted :)

fzaninotto (approved):

fzaninotto:
Thanks!

PR #9578, opened 2024-01-11
Add tests to the app generated by create-react-admin with ra-data-fakerest

djhi:
# How To Test

`make build-create-react-admin install`

Then from outside the project monorepo, call:

`[PATH TO MONOREPO]/node_modules/.bin/create-react-admin myadmin --data-provider ra-data-fakerest --auth-provider none --install npm`
like:
`./react-admin-next/node_modules/.bin/create-react-admin myadmin2 --data-provider ra-data-fakerest --auth-provider none --install npm`

and run the tests:
```
cd myadmin
npm run test
```


packages/create-react-admin/templates/common/gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
djhi, line 1: I wondered whether to name this file `.gitignore.template` to make it more explicit

slax57 (changes requested):
I also have a few more important remarks.

I tested the 'published' version of the CLI (following the instructions in the Readme) and found out that:

1. Running `npm install` to install the build version of the CLI locally fails, because of the following error: `Could not resolve dependency: peer react@"^16.5.2 || ^17.0.0" from ink-select-input@4.2.2` (since we now require react 18) => Do we need to update to a newer version of `ink-select-input`?
2. in the generated app, we have both a `.gitignore` and a `gitignore` file
3. the generated Readme does not explain how to run the tests
4. The test is failing in the generated project, because of a timeout (`Test timed out in 5000ms`) although the app is running properly locally :man_shrugging: 
5. The tests seem to assume we chose an `authProvider`, but it's not mandatory. We should only include the test if an `authProvider` was selected IMO

packages/create-react-admin/templates/ra-data-fakerest/src/App.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+	fireEvent.click(await screen.findByText("Save"));
+	await screen.findByText("Post 1 edited");
+
+	// Test comments
+	fireEvent.click(await screen.findByText("Comments"));
+	fireEvent.click(await screen.findByText("Comment 1"));
+	fireEvent.click(await screen.findByDisplayValue("#0"));
+	fireEvent.click(await screen.findByText("#11"));
+	fireEvent.click(await screen.findByText("Save"));
+	await screen.findByText("#11", { selector: "a *" });
```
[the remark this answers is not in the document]
djhi: Added comments

slax57 (changes requested):
Code looks good, but:

- There is still a `gitignore` file generated in addition to `.gitignore`
- Tests are still failing when run in the generated project (timeout after 5000 ms)

djhi:
> Code looks good, but:
> 
> * There is still a `gitignore` file generated in addition to `.gitignore`

FIxed

> * Tests are still failing when run in the generated project (timeout after 5000 ms)

Not sure what happen on your machine. It works fine on mine and on the CI (https://github.com/marmelab/react-admin/pull/9580)

packages/create-react-admin/src/generateAppTestFile.ts
```diff
@@ -0,0 +1,56 @@
...
+	fireEvent.click(await screen.findByText("Comments"));
+	// Open the first comment
+	fireEvent.click(await screen.findByText("Comment 1"));
+	// Edit the comment selected post
+	fireEvent.click(await screen.findByDisplayValue("#0"));
+	fireEvent.click(await screen.findByText("#11"));
+	fireEvent.click(await screen.findByText("Save"));
+	// Check the comment has been updated by finding the post link in the comments list page
+	await screen.findByText("#11", { selector: "a *" });
+});
```
djhi, line 52:
```suggestion
}, 6000);
```

slax57:
> Tests are still failing when run in the generated project (timeout after 5000 ms)

Okay so I pushed the investigation further
- Tests run fine with the parameter `--auth-provider none`
- Tests fail (on my machine) with the parameter `--auth-provider local-auth-provider` (timeout after 5000 ms)
- If I increase the test timeout to 6000 ms then it works

=> Hence I recommend increasing the timeout a bit by default, otherwise we risk having the test fail on the CI too :shrug:

slax57 (approved):

PR #9580, opened 2024-01-11
create-react-admin gh action to test new apps

djhi:
## Problem 

When working on react-admin, we have our dependencies pinned and we may not see errors that may happen with more recent versions of those dependencies like what happened with react-router.

## Solution

Introduce a github action running on new tagged releases that will create a new app using `create-react-admin` and run some basic tests on it. 

Follow #9578


.github/workflows/test.yml
```diff
@@ -133,4 +133,28 @@ jobs:
...
-          GREENFRAME_SECRET_TOKEN: ${{secrets.GREENFRAME_SECRET_TOKEN}}
\ No newline at end of file
+          GREENFRAME_SECRET_TOKEN: ${{secrets.GREENFRAME_SECRET_TOKEN}}
+
+  create-react-admin:
+    runs-on: ubuntu-latest
+    name: create-react-admin
+    needs: []
+    # TODO: uncomment once ready
+    # if: success() && github.ref_type == 'tag'
```
slax57, line 143: Remaining TODO

packages/create-react-admin/templates/ra-data-fakerest/src/App.spec.tsx
```diff
@@ -0,0 +1,36 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
```
slax57, line 1: Forgot to mention that on #9578 , but this file should probably be deleted since we now generate the tests dynamically

slax57 (changes requested):

packages/create-react-admin/templates/ra-data-fakerest/src/App.spec.tsx
slax57: Did you see this comment @djhi ?

.github/workflows/test.yml
```diff
@@ -133,4 +133,27 @@ jobs:
...
+          GREENFRAME_SECRET_TOKEN: ${{secrets.GREENFRAME_SECRET_TOKEN}}
+
+  create-react-admin:
+    runs-on: ubuntu-latest
+    name: create-react-admin
+    needs: []
+    if: success() && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/next')
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
adguernier, line 145: **question (non-blocking)**: I'm just wondering why not using the latest version`actions/checkout@v4`

adguernier (approved):

.github/workflows/test.yml
djhi: I used the same we already have

slax57 (approved):

PR #9586, opened 2024-01-15
[Demo] Remove useless defaultProps usage

fzaninotto:
React discourages the use of defaultProps and may one day remove support for it. They aren't even documented in the latest version of the React docs. 

Our demos contain custom components that don't really need defaultProps, so this PR replaces it with real props.

Some custom components with defaultProps remain in the demo (StarRatingField, ColoredNumberField, etc) because removing defaultProps for them would lead to a lot of repetition. These currently need defaultProps and there is no better alternative for now.

### DIFF
--- a/examples/demo/src/reviews/ReviewListMobile.tsx
+++ b/examples/demo/src/reviews/ReviewListMobile.tsx
@@ -22,12 +22,4 @@ const ReviewListMobile = () => {
     );
 };
 
-ReviewListMobile.propTypes = {
-    data: PropTypes.any,
-    hasBulkActions: PropTypes.bool.isRequired,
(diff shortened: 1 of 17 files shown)


examples/demo/src/reviews/ReviewListMobile.tsx
```diff
@@ -22,17 +22,4 @@ const ReviewListMobile = () => {
     );
 };
 
-ReviewListMobile.propTypes = {
```
djhi, line 25: Linter warning in this file.

djhi (changes requested):

djhi (approved):

Issue #9587, opened 2024-01-15
Unable to omit specific fields which share the same source


Dreamsorcerer:
**Is your feature request related to a problem? Please describe.**
If there are multiple fields which use the same source (e.g. to render different views of the same data), then it's not possible to omit specific fields from DatagridConfigurable as the omit property works with sources only. Given that they can be individually toggled on/off in the columns selection, it seems reasonable that they should also be configurable individually in omit.

**Describe the solution you'd like**
Haven't thought of one yet.

slax57:
Thank you for your suggestion.
This seems like a very specific use-case, I'm not sure it's that common.
I suggest we wait for more feedback from the community before taking actions.

Dreamsorcerer:
Just to give an example, you could have a NumberField to display the value of a foreign key, a ReferenceField linking to the foreign record, and a field that displays a link to some external resource, all using the same source attribute. Another example could be a TextField showing a file name and a custom field that displays an audio player for that file. I suspect there are many examples of these around, they probably just haven't tried to omit one individually.

I think another issue is that you can't omit columns without a source. For example, putting an Edit button in the list view, it then can't be omitted by default, even though the user can hide it with the columns button.

slax57:
Okay I get your point, thanks for providing additional examples.
I'll mark this as enhancement then, so anyone can open a PR implementing this feature (remember to target the `next` branch if your do).
Note that we probably won't work on this ourselves though, since we don't have that need on our own projects.

Dreamsorcerer:
So, the Columns button appears to use the column index. I'm not sure that's going to be a reliable way to implement this feature. Could we maybe extend omit so it matches the label _or_ the source?

fzaninotto:
The columns preference is a mix of `availableColumns` (used to display the list of possible columns in SelectColumnsButtons) and `columns` (which stores the user's choice regarding the orer and visibility of the columns). The second setting is related to the first one.

You have to consider that the feature must still work after the developer changes the number of orer of Fields in the datagrid, given users have already saved preferences.

Dreamsorcerer:
Right, so columns is actually the index listed in availableColumns. I'm not sure how that index is kept in sync with developer changes (presumably by checking for columns with the same source or label?).

But, availableColumns contains both source and label, which suggests to me that allowing omit to match the label would be a resonable approach?

fzaninotto:
No news for some time, closing.

Dreamsorcerer:
[...]

PR #9588, opened 2024-01-16
Use elem.key from filter elements

Dreamsorcerer:
Tangentially related to #9587, if 2 inputs are used in filters with the same source then they end up with duplicate keys even when those components already provide their own unique keys.

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
@@ -59,8 +59,8 @@ describe('<FilterForm />', () => {
     it('should retain key values in the form inputs', () => {
         // As key is not rendered, we just test that the React warning doesn't occur.
         const origWarn = console.warn;
-        console.warn = (message) => {
-            throw new Error(message)
+        console.warn = message => {
+            throw new Error(message);
         };
 
         const setFilters = jest.fn();
--- a/packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterForm.spec.tsx
@@ -78,7 +78,7 @@ describe('<FilterForm />', () => {
             </AdminContext>
         );
         expect(screen.queryAllByLabelText('Title')).toHaveLength(1);
-        expect(screen.queryAllByLabelText('Name')).toHaveLength(1);
+        expect(screen.queryAllByLabelText('Title2')).toHaveLength(1);
     });
 
     it('should change the filter when the user updates an input', async () => {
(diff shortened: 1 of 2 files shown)


fzaninotto:
Can you please add a unit test showing the bug, and explain the use case for this? #9587 gives a good use case for datagrid fields, but I fail to understand in which case you need the same for filters.

Dreamsorcerer:
> explain the use case for this

If you have a FK that links to 2 different resources, you could have 2 ReferenceInputs in order to filter based on one linked resource or the other (you might even want a plain NumberInput or similar to set the value directly). Another case might be something like a combobox, radio buttons or similar to select from a number of favourite/recent options, while also having a text or slider or similar to be able to filter on other values.

Dreamsorcerer:
> Can you please add a unit test showing the bug

I put a test in, but realised the key is not rendered, so I don't think there's anything I can actually assert with the test.

Not sure how you handle warnings, but I have `jest-fail-on-console` on my own test suite to fail on warnings. This code should produce a warning from React without the fix.

fzaninotto:
I'm still not convinced by the use cases you list. It may be something you want to do in your app, but I don't think it's something we want to support with the react-admin basic filters. I'll wait for the feedback of other contributors. 

Your test doesn't seem to do any useful assertion. Also, I understand that the filter already works in your use case, but simply logs a warning in the console. Am I right?

Dreamsorcerer:
> It may be something you want to do in your app, but I don't think it's something we want to support with the react-admin basic filters.

I mean, it's not like I'm asking you to explicitly support these complex features. I just want to be able to specify the key on components explicitly, as I've done everywhere else in the app. It seems odd to me that it throws away the key that already exists on the component, and the fix for that is 2 words.

> Your test doesn't seem to do any useful assertion.

That is what I said. The only thing I can test, as far as I can tell, is that the warning doesn't occur.

Dreamsorcerer:
> Also, I understand that the filter already works in your use case, but simply logs a warning in the console.

Just to expand on that, this is the warning. It's pretty clear that the behaviour is bad, may cause issues today, and may become a error in future:

> Warning: Encountered two children with the same key, `fk_child_id__ref_num`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

Dreamsorcerer:
I've updated the test so it's explicitly testing something now.

> Another case might be something like a combobox, radio buttons or similar to select from a number of favourite/recent options, while also having a text or slider or similar to be able to filter on other values.

A specific example of this is filtering on age range. You might have one input that provides a 2-point slider to select a range. Then a second input that provides a drop-down of pre-selected age ranges for convenience ("Adult (18+)", "Child (3-17)", "Infant (<=2)").

fzaninotto:
Your test passes even without the fix in FilterForm... It's not testing the right code.

Dreamsorcerer:
Sorry, I've been super busy. If the test is still passing, then I can only imagine the tests are already doing something with warnings and managing to ignore them?

fzaninotto:
Not that I know of.

Dreamsorcerer:
> Not that I know of.

OK, it's an error, not a warning. Test works properly now. FYI, there are plenty of other errors and warning getting logged in the other tests.

slax57 (approved):
Thanks!

PR #9591, opened 2024-01-17
Remove injected props in Admin and Layout components

fzaninotto:
## Problems

In order to build a custom layout for `<Admin layout>`, developers must clone props:

```jsx
// in src/MyLayout.js
import { Layout } from 'react-admin';
import { MyAppBar } from './MyAppBar';

export const MyLayout = props => (
   <Layout {...props} appBar={MyAppBar} />
);
```

Developers wonder: what is `props`? What's its type? Why do I have to pass it down? It’s not so usual in modern React.

The same happens for other components that can be set on `<Layout>`: `appBar`, `menu`, `error`, `catchAll`.

## Solution

The only useful data injected is `title` and `dashboard`. The rest is useless as the data can be obtained in another way. 

Use a context instead of props injection for these values. That way, a custom layout just needs to pass the children down (which requires no explaining).

```jsx
// in src/MyLayout.js
import { Layout } from 'react-admin';
import { MyAppBar } from './MyAppBar';

export const MyLayout = ({ children }) => (
  <Layout appBar={MyAppBar}>
     {children}
  </Layout>
);
```

Also, remove deprecated prop injection as there are alternatives. 

## Tasks

- [x] Remove prop injection in `layout`
- [x] Remove prop injection in `menu`
- [x] Remove prop injection in `appBar`
- [x] Remove prop injection in `error`
- [x] Remove prop injection in `catchAll`
- [x] Change types to reflect that
- [x] Update documentation and jsDoc
- [x] Fix outdated props injection in jsDoc snippets
- [x] Write Upgrade guide


docs/Upgrade.md
```diff
@@ -133,6 +133,144 @@ Here are the available codemods you may need to run on your codebase:
 
 Check out React Query [codemod documentation](https://tanstack.com/query/latest/docs/react/guides/migrating-to-v5#codemod) for more information.
 
+## `<Admin menu>` Is No Longer Supported
+
+The `<Admin menu>` prop was deprecated since 4.0. It's no longer supported. If you want to customize the application menu, you'll have to do it un a custom Layout instead:
```
djhi, line 138:
```suggestion
The `<Admin menu>` prop was deprecated since 4.0. It's no longer supported. If you want to customize the application menu, you'll have to do it in a custom Layout instead:
```

djhi (changes requested):
LGTM

djhi (approved):

adguernier (approved):

PR #9594, opened 2024-01-19
Fix label inference regression

djhi:
# Problem

When introducing the `SourceContext`, we also introduced a regression on the label inference (when no translations are provided). We pass the modified `source` to the inference function (eg: `title.en`) instead of the input `source` (eg: `title`).
This results in wrong label like `Title en` instead of `Title`


fzaninotto:
Can you please add a description and a unit test?

fzaninotto (approved):

slax57 (approved):

PR #9600, opened 2024-01-22
Fix `useUpdateMany` doesn't properly update cache in pessimistic mode

fzaninotto:
## Problem

When writing a bulk action button based on `useUpdateMany` with the default mutation mode (pessimistic), the list data doesn't get refreshed after the mutation ends. 

You can reproduce it in the following Stackblitz: https://stackblitz.com/edit/github-4g364g?file=src%2Fposts%2FResetViewsButton.tsx

1. Select a few posts in the list
2. Click on "reset views"
3. When the notification confirms the mutation success, the views of the selected rows don't reset. 

## Solution

There is a bug in `useUpdateMany`because we use the variable name `data` at several places.

### DIFF
--- a/packages/ra-ui-materialui/src/button/BulkUpdateWithUndoButton.tsx
+++ b/packages/ra-ui-materialui/src/button/BulkUpdateWithUndoButton.tsx
@@ -41,7 +41,6 @@ export const BulkUpdateWithUndoButton = (
                 undoable: true,
             });
             unselectAll();
-            refresh();
         },
         onError = (error: Error | string) => {
             notify(
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -146,7 +146,7 @@ export default PostList;
 
 ## `bulkActionButtons`
 
-Bulk action buttons appear when users select one or several rows, and affect all the selected records. This is useful for actions like mass deletion or mass edition.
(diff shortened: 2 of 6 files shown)


fzaninotto:
Back to WIP: the behavior in optimistic mode is now wrong (the values blink)

packages/ra-ui-materialui/src/button/BulkUpdateWithUndoButton.tsx
```diff
@@ -41,7 +41,6 @@ export const BulkUpdateWithUndoButton = (
                 undoable: true,
             });
             unselectAll();
-            refresh();
```
fzaninotto, line 44: The success refresh was necessary because the cache update was wrong. After the fix, the refresh is not only unnecessary, it also poses a problem in optimistic mode.

In this mode, the button updates the cache, then refreshes (and shows the old version), then proceeds to the mutation, which properly updates the value. This leads to a misleading blink:

old value => new value => old value => new value

We also remove the refresh in the BulkUpdateWithConfirmButton as it can be used in optimistic mode.

fzaninotto:
Back to RFR

djhi (commented):
You can remove the `useRefresh` hook in `BulkUpdateWithConfirmButton` now

slax57 (approved):

djhi (approved):

PR #9602, opened 2024-01-23
[Doc] Fix link to documentation root

fzaninotto:
The page https://marmelab.com/react-admin/Readme.html is no longer maintained an isn't the documentation root anymore. Yet it is visited by users who find links elsewhere in the documentation. 

We can't remove this page though, as existing installations of react-admin will still point to it in the Error page.

### DIFF
--- a/docs/404.html
+++ b/docs/404.html
@@ -58,7 +58,11 @@ d88   888  888    888 d88   888
     function replaceNavigationWithVersion(version) {
         var location = document.location.href.split('#')[0];
         var locationSlices = location.split('/');
-        var href = [...locationSlices.slice(0, locationSlices.length - 2), version, 'Readme.html'].join('/');
+        var href = [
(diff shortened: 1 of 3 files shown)


djhi (approved):

docs/404.html
```diff
@@ -58,7 +58,11 @@
     function replaceNavigationWithVersion(version) {
         var location = document.location.href.split('#')[0];
         var locationSlices = location.split('/');
-        var href = [...locationSlices.slice(0, locationSlices.length - 2), version, 'Readme.html'].join('/');
+        var href = [
+            ...locationSlices.slice(0, locationSlices.length - 2),
+            version,
+            'navigation.html',
```
slax57, line 64: I don't understand why you chose `'navigation.html'` here an not `'documentation.html'`. Is it intentional?

fzaninotto: No, it's a mistake!

PR #9613, opened 2024-01-25
Sending json Content-Type header for empty body breaks some parsers 

chrisDeFouRire:
I'm using Fastify on the backend, and it refuses `DELETE` requests sent from react-admin's `ra-data-json-server`.
It is because `fetchUtils.fetchJson` is setting the `Content-Type` header to `application/json`, even with an empty body like in the delete requests. The function that is setting the header is `createHeadersFromOptions` in `fetch.ts`.

It is not against the RFCs to do so, but it just breaks Fastify, and it makes little sense to send a Content-Type with a body that doesn't adhere to the content type: an empty body isn't valid json.

This pull request fixes the issue and adds new unit tests to ensure there is no regression.

`yarn test-unit -- ./packages/ra-core/src/dataProvider/fetch.spec.ts` passes.

I guess the alternatives are:
- sending a body if there isn't one already (with `'null'` in it)
- modifying the `ra-data-json-server` package (and other data providers which are using DELETE with no body) to specify a body.

I thought fixing the `createHeadersFromOptions` option was the cleanest

### DIFF
--- a/packages/ra-core/src/dataProvider/fetch.spec.ts
+++ b/packages/ra-core/src/dataProvider/fetch.spec.ts
@@ -62,13 +62,20 @@ describe('flattenObject', () => {
 });
 
 describe('createHeadersFromOptions', () => {
-    it('should add a Content-Type header for POST requests', () => {
-        const options = {
+    it('should add a Content-Type header for POST requests if there is a body', () => {
+        const optionsWithBody = {
+            method: 'POST',
+            body: JSON.stringify(null),
+        };
+
+        const headers = createHeadersFromOptions(optionsWithBody);
+        expect(headers.get('Content-Type')).toStrictEqual('application/json');
+        const optionsWithoutBody = {
             method: 'POST',
         };
 
-        const headers = createHeadersFromOptions(options);
-        expect(headers.get('Content-Type')).toStrictEqual('application/json');
+        const headersWithoutBody = createHeadersFromOptions(optionsWithoutBody);
+        expect(headersWithoutBody.get('Content-Type')).toBeNull();
     });
 
     it('should not add a Content-Type header for GET requests', () => {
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/dataProvider/fetch.ts
```diff
@@ -14,9 +14,12 @@ export const createHeadersFromOptions = (options: Options): Headers => {
...
         })) as Headers;
     if (
+        // An application/json Content-Type header in requests without a body breaks some parsers, like fastify
+        options &&
+        options.body &&
         !requestHeaders.has('Content-Type') &&
-        !(options && (!options.method || options.method === 'GET')) &&
-        !(options && options.body && options.body instanceof FormData)
+        !(!options.method || options.method === 'GET') &&
+        !(options.body instanceof FormData)
```
fzaninotto, line 22: The accumulation of conditions and the double negations make it hard to read. Could you please make the code clearer? Something like:

```jsx
const shouldSetContentType = () => {
        const hasBody = options && options.body;
        const isContentTypeSet = requestHeaders.has('Content-Type');
        const isGetMethod = !options?.method || options?.method === 'GET';
        const isFormData = options?.body instanceof FormData;

        return hasBody && !isContentTypeSet && !isGetMethod && !isFormData;
    };
```

packages/ra-core/src/dataProvider/fetch.spec.ts
```diff
@@ -62,13 +62,20 @@ describe('flattenObject', () => {
...
 
-        const headers = createHeadersFromOptions(options);
+        const headers = createHeadersFromOptions(optionsWithBody);
         expect(headers.get('Content-Type')).toStrictEqual('application/json');
+        const optionsWithoutBody = {
+            method: 'POST',
+        };
+
+        const headersWithoutBody = createHeadersFromOptions(optionsWithoutBody);
+        expect(headersWithoutBody.get('Content-Type')).toBeNull();
```
fzaninotto, line 78: Please extract this into a new test

packages/ra-core/src/dataProvider/fetch.spec.ts
```diff
@@ -85,4 +92,40 @@ describe('createHeadersFromOptions', () => {
         );
         expect(headersWithoutMethod.get('Content-Type')).toBeNull();
     });
+
+    it('should not add a Content-Type header if there is no body', () => {
```
fzaninotto, line 96: This title reads like the one in L65. You should probably explicit the fact that it's for DELETE here

packages/ra-core/src/dataProvider/fetch.spec.ts
```diff
@@ -85,4 +92,40 @@ describe('createHeadersFromOptions', () => {
...
+            optionsWithDeleteAndBody
+        );
+        expect(headersWithDeleteAndBody.get('Content-Type')).toStrictEqual(
+            'application/json'
+        );
+    });
+
+    it('should not add a Content-Type header if there already is a Content-Type header', () => {
+        const optionsWithContentType = {
+            headers: new Headers({
```
fzaninotto, line 118: And you should probably use another method than the default, otherwise you're falling back on the test on GET.

fzaninotto (changes requested):
Nice!

packages/ra-core/src/dataProvider/fetch.ts
chrisDeFouRire: right! I did just that

packages/ra-core/src/dataProvider/fetch.spec.ts
chrisDeFouRire: you're correct!

fzaninotto (approved):

chrisDeFouRire:
Thank you for giving us React Native! it's a great tool ! and seeing how easily I can contribute makes me think it's the perfect tool for us even more! 🙏🙏

PR #9620, opened 2024-01-29
Leverage SourceContext in fields

djhi:
- [x] Use `SourceContext` for `TranslatableFields`
- [x] Introduce `useFieldValue` to get a record value
- [x] Document `useFieldValue`

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -508,7 +508,7 @@ import * as React from 'react';
 -import { useRecordContext } from 'react-admin';
 +import { useFieldValue } from 'react-admin';
 
-const TextField = ({ source }) => {
+const TextField = (props) => {
 -    const record = useRecordContext();
 +   const value = useFieldValue(props);
 -   return record ? <span>{record[source]}</span> : null;
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -512,11 +512,13 @@ const TextField = (props) => {
 -    const record = useRecordContext();
 +   const value = useFieldValue(props);
 -   return record ? <span>{record[source]}</span> : null;
-+   return record ? <span>{value}</span> : null;
++   return <span>{value}</span> : null;
 }
 
 export default TextField;
 ```
+
+**Tip**: Note that when using `useFieldValue`, you don't need to check that `record` is defined.
 
 ## Hiding A Field Based On The Value Of Another
 
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -501,17 +501,18 @@ export const UserList = () => (
 
 **Tip**: In such custom fields, the `source` is optional. React-admin uses it to determine which column to use for sorting when the column header is clicked. In case you use the `source` property for additional purposes, the sorting can be overridden by the `sortBy` property on any `Field` component.
 
-If you build a reusable field accepting a `source` props, you will probably want to support deep field sources (e.g. source values like `author.name`). Use [lodash/get](https://www.npmjs.com/package/lodash.get) to replace the simple object lookup. For instance, for a Text field:
+If you build a reusable field accepting a `source` props, you will probably want to support deep field sources (e.g. source values like `author.name`). Use the [`useFieldValue` hook](/useFieldValue.md) to replace the simple object lookup. For instance, for a Text field:
 
 ```diff
 import * as React from 'react';
-+import get from 'lodash/get';
-import { useRecordContext } from 'react-admin';
+-import { useRecordContext } from 'react-admin';
++import { useFieldValue } from 'react-admin';
 
 const TextField = ({ source }) => {
-    const record = useRecordContext();
+-    const record = useRecordContext();
++   const value = useFieldValue(props);
 -   return record ? <span>{record[source]}</span> : null;
-+   return record ? <span>{get(record, source)}</span> : null;
++   return record ? <span>{value}</span> : null;
 }
 
 export default TextField;
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -511,7 +511,7 @@ import * as React from 'react';
 const TextField = (props) => {
 -    const record = useRecordContext();
 +   const value = useFieldValue(props);
--   return record ? <span>{record[source]}</span> : null;
+-   return record ? <span>{record[props.source]}</span> : null;
 +   return <span>{value}</span> : null;
 }
 
--- a/docs/useFieldValue.md
+++ b/docs/useFieldValue.md
@@ -42,7 +42,7 @@ The name of the property on the record object that contains the value to display
 
 ```tsx
 import * as React from 'react';
-+import { useFieldValue } from 'react-admin';
+import { useFieldValue } from 'react-admin';
 
 export const CustomerCard = () => {
     const firstName = useFieldValue({ source: 'firstName' });
--- a/docs/useFieldValue.md
+++ b/docs/useFieldValue.md
@@ -11,14 +11,27 @@ A hook that gets the value of a field of the current record. It gets the current
 
 Here is an example `TextField` component:
 
-```jsx
+```tsx
+// In TextField.tsx
 import * as React from 'react';
-+import { useFieldValue } from 'react-admin';
++import { useFieldValue, type FieldProps } from 'react-admin';
 
-const TextField = (props) => {
+export const TextField = (props: FieldProps) => {
     const value = useFieldValue(props);
-    return record ? <span>{value}</span> : null;
+    return <span>{value}</span>;
 }
+
+// In PostShow.tsx
+import { Show, SimpleShowLayout } from 'react-admin';
+import { TextField } from './TextField.tsx';
+
+export const PostShow = () => (
+    <Show>
+        <SimpleShowLayout>
+            <TextField source="author.name" label="Author" />
+        </SimpleShowLayout>
+    </Show>
+);
 ```
 
 ## Params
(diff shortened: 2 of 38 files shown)


packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -87,8 +90,6 @@ export const FileField = <
         );
     }
 
-    const titleValue = get(record, title)?.toString() || title;
```
fzaninotto, line 90: you've lost the `toString()` in the process.

packages/ra-ui-materialui/src/field/ImageField.tsx
```diff
@@ -62,8 +66,6 @@ export const ImageField = <
         );
     }
 
-    const titleValue = get(record, title)?.toString() || title;
```
fzaninotto, line 65: same

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -63,9 +63,10 @@ export const ReferenceField = <
 >(
     props: ReferenceFieldProps<RecordType, ReferenceRecordType>
 ) => {
-    const { source, emptyText, link = 'edit', ...rest } = props;
```
fzaninotto, line 66: You still pass down the rest props, which means you'll pass down teh `source`, that is explicitly excluded in NonEmptyReferenceField

packages/ra-core/src/util/useFieldValue.spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useFieldValue, UseFieldValueOptions } from './useFieldValue';
+import { RecordContextProvider } from '../controller';
+import { SourceContextProvider } from '..';
+
+describe('useFieldValue', () => {
+    const Component = (props: UseFieldValueOptions) => {
+        return <div>{useFieldValue(props) ?? 'None'}</div>;
+    };
+
+    it('should return undefined if no record is available', async () => {
```
fzaninotto, line 12: you're missing a test about deep sources

packages/ra-core/src/i18n/TranslatableContext.ts
```diff
@@ -5,8 +5,6 @@ export const TranslatableContext = createContext<
 >(undefined);
 
 export interface TranslatableContextValue {
-    getLabel: GetTranslatableLabel;
```
fzaninotto, line 8: The two types lines 13 & 14 also need removing

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -41,34 +55,34 @@ export const TranslatableFieldsTabContent = (
...
-                        ) : (
-                            cloneElement(field, {
-                                ...field.props,
-                                label: getLabel(field.props.source),
-                                record,
-                                source: getSource(field.props.source, locale),
-                            })
-                        )}
-                    </div>
+                    <ResourceContextProvider
```
fzaninotto, line 58: I don't understand why you need this. We already are in a ResourceContext

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -41,34 +55,34 @@ export const TranslatableFieldsTabContent = (
...
+                        value={resource}
+                        key={field.props.source}
+                    >
+                        <RecordContextProvider value={record}>
```
fzaninotto, line 62: I don't understand why you need this. We already are in a RecordContext

fzaninotto (changes requested):
I see a regression in the TranslatableFields storybook: labels are no longer translated

docs/useFieldValue.md
```diff
@@ -0,0 +1,32 @@
...
+```jsx
+import * as React from 'react';
++import { useFieldValue } from 'react-admin';
+
+const TextField = ({ source }) => {
+    const value = useFieldValue(props);
+    return record ? <span>{value}</span> : null;
+}
+```
+
```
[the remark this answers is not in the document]
djhi: Explained in the introduction:
> It gets the current record from the context or use the one provided as a prop. It supports deep sources such as `name.fr`.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
djhi: What about when you pass the `resource` prop?

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
djhi: Same for `record`

packages/ra-core/src/util/useFieldValue.ts
```diff
@@ -0,0 +1,32 @@
...
+) => {
+    const { source } = options;
+    const sourceContext = useSourceContext();
+    const record = useRecordContext<RecordType>(options);
+
+    return get(record, sourceContext?.getSource(source) ?? source);
+};
+
+export interface UseFieldValueOptions<RecordType = RaRecord> {
+    source?: string;
```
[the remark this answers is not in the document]
djhi: No, just like in all our fields

packages/ra-core/src/i18n/TranslatableContext.ts
djhi: Removed

packages/ra-core/src/util/useFieldValue.spec.tsx
djhi: Added

packages/ra-ui-materialui/src/field/FileField.tsx
djhi: Added back

packages/ra-ui-materialui/src/field/ImageField.tsx
djhi: Added back

packages/ra-ui-materialui/src/field/ReferenceField.tsx
djhi: Added back

docs/Fields.md
```diff
@@ -501,17 +501,18 @@ export const UserList = () => (
...
++import { useFieldValue } from 'react-admin';
 
-const TextField = ({ source }) => {
-    const record = useRecordContext();
+const TextField = (props) => {
+-    const record = useRecordContext();
++   const value = useFieldValue(props);
 -   return record ? <span>{record[source]}</span> : null;
-+   return record ? <span>{get(record, source)}</span> : null;
++   return record ? <span>{value}</span> : null;
```
fzaninotto, line 515: `record` is not defined

docs/useFieldValue.md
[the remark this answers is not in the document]
fzaninotto: You could add an example usage after this snippet where you use this new `TextField` with a compound source, like

```
const PostShow = () => (
    <SimpleShowLayout>
        <TextField source="author.name" label="Author" />
```
```

File: docs/useFieldValue.md [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+## Params
+
+### `source`
+
+The name of the property on the record object that contains the value to display. Can be a deep path.
```
fzaninotto, line 28: Please include an example.

docs/useFieldValue.md
```diff
@@ -0,0 +1,32 @@
...
+
+### `record`
+
+The record from which to read the value. Read from the `RecordContext` by default.
```
fzaninotto, line 32: please include an example

docs/Fields.md
```diff
@@ -501,17 +501,18 @@ export const UserList = () => (
...
 }
```
fzaninotto, line 517: Also, you should add that you don't need to check the existence of the `record` before (contrary to what's explained earlier) when using `useFieldValue`.

packages/ra-core/src/util/useFieldValue.ts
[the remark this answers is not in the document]
fzaninotto: I don't understand the rationale: in which case can we call this hook without `source`? But that's probably for another PR.

packages/ra-core/src/util/useFieldValue.ts
```diff
@@ -0,0 +1,32 @@
...
+    const sourceContext = useSourceContext();
+    const record = useRecordContext<RecordType>(params);
+
+    return get(record, sourceContext?.getSource(source) ?? source);
+};
+
+export interface UseFieldValueOptions<RecordType = RaRecord> {
+    source?: string;
+    record?: RecordType;
+}
```
fzaninotto, line 32: while we are at it, I suggest adding a `defaultValue` option to make this hook more useful

fzaninotto (changes requested):
Almost there!

packages/ra-core/src/util/useFieldValue.ts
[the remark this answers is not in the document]
djhi: You're right but the issue is that our fields don't require `source` so we wouldn't be able to call this hook if it required `source`.

packages/ra-core/src/util/useFieldValue.ts
djhi: Added

docs/Fields.md
djhi: Added

docs/useFieldValue.md
djhi: Added

docs/useFieldValue.md
djhi: Added

fzaninotto (approved):

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -41,34 +58,32 @@ export const TranslatableFieldsTabContent = (
...
+                            <SourceContextProvider value={sourceContext}>
+                                <div>
+                                    {addLabel ? (
+                                        <Labeled
+                                            resource={resource}
+                                            label={field.props.source}
```
slax57, line 71: **issue:** You changed `field.props.label‏` to `field.props.source` which I think is a bug because it introduces a regression in the default label key.

Following screenshots were taken in the `Basic` story, after removing the `i18nProvider`.

BEFORE




AFTER

slax57 (changes requested):
One more thing: I see TS errors about `source` in both stories and tests files in my IDE. Do you see them too?

docs/Fields.md
```diff
@@ -501,22 +501,25 @@ export const UserList = () => (
...
-import { useRecordContext } from 'react-admin';
+-import { useRecordContext } from 'react-admin';
++import { useFieldValue } from 'react-admin';
 
-const TextField = ({ source }) => {
-    const record = useRecordContext();
+const TextField = (props) => {
+-    const record = useRecordContext();
++   const value = useFieldValue(props);
 -   return record ? <span>{record[source]}</span> : null;
```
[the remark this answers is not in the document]
djhi: Nope it's a removed line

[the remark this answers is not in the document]
slax57: sure, but you did change a line which is not changed by the diff (neither `-` nor `+`) so you need to reflect changes on the state of the code _before_ the diff (i.e. lines starting with `-` as well).

If you consider the code _before_ the diff is applied, then this code does not compile.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
djhi: It's actually a bugfix as this is how all fields are handled elsewhere unless I missed something?

fzaninotto:
Still missing a section in the upgrade guide for custom fields

slax57 (approved):

PR #9627, opened 2024-01-31
[Doc] Add providers logos

erwanMarmelab:
## Problem

The doc pages for dataProviders and authProviders are austere

## Solution

Add a small logo in front of each backend name.

0 0 0 0_4000_AuthProviderList html

0 0 0 0_4000_DataProviderList html


docs/css/style-v19.css
```diff
@@ -719,3 +719,27 @@ body.no-sidebar .sidenav-trigger {
     text-decoration: none !important;
 }
 
+/* Auth and Data Provider icons */
+
+img {
```
fzaninotto, line 724: Not good: we have other images than just the ones you've added. Now all the images are VERY small! Only style under a subclass.

fzaninotto (changes requested):

djhi (approved):

fzaninotto (approved):

PR #9634, opened 2024-02-01
Fix FieldToggle cannot be styled using the application theme

megantaylor:
so custom styles can be applied via theme


fzaninotto:
Thanks!

PR #9637, opened 2024-02-02
Fix `<ReferenceInput>` accepts a validate prop while it should not

### LINKED ISSUE #9568: Using validate on ReferenceInput
**What you were expecting:**
Setting `validate=required()` on ReferenceInput to make the label appear as required.

**What happened instead:**
I needed to set validate on AutocompleteInput instead.

Maybe it would be good to mention validate in the ReferenceInput documentation to avoid this confusion? Unless there's a code fix to ensure that validate gets passed through?

Code I used looks something like:
```
<ReferenceInput {...props}>
    <AutocompleteInput label={props["label"]} />
</ReferenceInput>
```

fzaninotto:
Closes #9568

### DIFF
--- a/packages/ra-ui-materialui/src/input/ReferenceInput.tsx
+++ b/packages/ra-ui-materialui/src/input/ReferenceInput.tsx
@@ -84,6 +84,11 @@ export const ReferenceInput = (props: ReferenceInputProps) => {
         filter,
     });
 
+    if (props.validate) {
+        throw new Error(
+            '<ReferenceInput> does not accept a validate prop. Set the validate prop on the child instead.'
+        );
+    }
     if (Children.count(children) !== 1) {
         throw new Error('<ReferenceInput> only accepts a single child');
     }
--- a/packages/ra-ui-materialui/src/input/ReferenceInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/ReferenceInput.spec.tsx
@@ -247,4 +247,23 @@ describe('<ReferenceInput />', () => {
         screen.getByLabelText('Save').click();
         await screen.findByText('Proust', undefined, { timeout: 5000 });
     });
+
+    it('should throw an error when using the validate prop', () => {
+        jest.spyOn(console, 'error').mockImplementation(jest.fn());
+        const dataProvider = testDataProvider({
+            getList: async () => ({ data: [], total: 25 }),
+        });
+        expect(() =>
+            render(
+                <CoreAdminContext dataProvider={dataProvider}>
+                    <Form>
+                        <ReferenceInput
+                            {...defaultProps}
+                            validate={() => undefined}
+                        />
+                    </Form>
+                </CoreAdminContext>
+            )
+        ).toThrowError();
+    });
 });


packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -84,6 +84,11 @@ export const ReferenceInput = (props: ReferenceInputProps) => {
         filter,
     });
 
+    if (props.validate) {
```
djhi, line 87: We could also modify the `ReferenceInputProps` type:
```
extends Omit<InputProps, 'validate'>
```

djhi (changes requested):

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
fzaninotto: I don't think that would be a better DX: Developers would see a failed compilation because of `validate`, without knowing what to do instead.

slax57 (approved):

djhi (approved):

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
adguernier: If we change the type it will be a BC, right? 
This two changes break something, so IMO we should update the types in V5 and revert this PR to put it in V5 too.

fzaninotto: The bug in your code wasn't caused by this change, it was revealed by this change. 

As explained, a type adjustments isn't a better DX IMO. Unless we add a "never" type and explain it with JSDoc?

digilist: Well, yes and no. You could say there was a bug, but defining the validation prop on the `ReferenceInput` did not have any effect and it was also defined on the `<AutocompleteInput>`. The validation was working and one validation prop was unnecessary, but that's not really a bug, is it?  But now the application/page is crashing, which imho is more severe and I would call this a bug that was introduced by this change.

fzaninotto: You're right. We shouldn't throw an error in production-only in development. I'll open a new PR to fix it. We won't introduce a type check until v5 though, as it is a breaking change.

digilist: Thank you :tada:

strarsis:
I just noticed this error after updating `react-admin` `4.x`.
The documentation appears to be incorrect, including for `react-admin` `5.x`, as it still uses the `validate` prop in `<ReferenceInput>`:

- [for `4.16`](https://marmelab.com/react-admin/doc/4.16/Inputs.html#:~:text=%3CReferenceInput%20label%3D%22User%22%20source%3D%22userId%22%20reference%3D%22users%22%20validate%3D%7B%5Brequired()%5D%7D%3E)
- [for `5.1.1`](https://marmelab.com/react-admin/Inputs.html#:~:text=%3CReferenceInput%20label%3D%22User%22%20source%3D%22userId%22%20reference%3D%22users%22%20validate%3D%7B%5Brequired()%5D%7D%3E)

And what should be used instead to make a `<ReferenceInput` having a required value, is this needed at all for it then/now?

fzaninotto:
@strarsis Thanks for the heads up, you're right! We'll fix the documentation shortly. 

As the error says: "Set the validate prop on the child instead.". See #10134 for the correct usage.

PR #9638, opened 2024-02-03
Add setParams callback to ListContext

### LINKED ISSUE #4189: Calling setSort and setFilters after one another filckers and doesn't setSort

fzaninotto:
## Problem

In a `<List>`, it's not possible to set both the sort and filter simultaneously. 

## Solution

Introduce a new calback in ListContext, `setParams`, which lets users set all the params they want in one shot. 

Closes #4189
Supersedes #9340

## Todo

- [x] Basic implementation for List
- [x] Unit tests for List
- [x] Basic implementation for InfiniteList
- [ ] Unit tests for InfiniteList
- [x] Basic implementation for useList
- [x] Unit tests for useList
- [x] Basic implementation for ReferenceManyField
- [ ] Unit tests for ReferenceManyField
- [x] Basic implementation for ReferenceInput
- [ ] Unit tests for ReferenceInput
- [x] Basic implementation for ReferenceArrayInput
- [ ] Unit tests for ReferenceArrayInput
- [ ] Add stories
- [ ] Add documentation

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -6,7 +6,7 @@ import { useSafeSetState, removeEmpty } from '../../util';
 import { useGetManyReference } from '../../dataProvider';
 import { useNotify } from '../../notification';
 import { Identifier, RaRecord, SortPayload } from '../../types';
-import { ListControllerResult } from '../list';
+import { ListControllerResult, ListParams } from '../list';
 import usePaginationState from '../usePaginationState';
 import { useRecordSelection } from '../list/useRecordSelection';
 import useSortState from '../useSortState';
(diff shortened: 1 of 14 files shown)


fzaninotto:
As discussed in #4189, I now believe this change isn't necessary, as there is a simple workaround for the initial problem. We won't move forward with this change.

PR #9643, opened 2024-02-05
Make useGetRecordRepresentation smarter

erwanMarmelab:
## Problem
useGetRecordRepresentation could be smarter when there is no record representation on the resource.

## Solution
It could contain a list of fields that can be used as record representation by default, such as:

1. name
2. title
3. label
4. reference
5. id

### DIFF
--- a/docs/useGetRecordRepresentation.md
+++ b/docs/useGetRecordRepresentation.md
@@ -9,6 +9,13 @@ Get a function that returns the record representation, leveraging the [`<Record 
 
 You can also use the component version: [`<RecordRepresentation>`](./RecordRepresentation.md).
 
+If you don't specify anything, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
+1. `name`
+2. `title`
+3. `label`
+4. `reference`
+5. `id`
+
 ## Usage
 
 ```tsx
--- a/docs/useGetRecordRepresentation.md
+++ b/docs/useGetRecordRepresentation.md
@@ -9,13 +9,6 @@ Get a function that returns the record representation, leveraging the [`<Record 
 
 You can also use the component version: [`<RecordRepresentation>`](./RecordRepresentation.md).
 
-when `<Resource recordRepresentation>` is not defined, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
-1. `name`
-2. `title`
-3. `label`
-4. `reference`
-5. `id`
-
 ## Usage
 
 ```tsx
@@ -59,6 +52,17 @@ const PostEdit = () => (
     </EditBase>
 )
 ```
+
+## Preferences
+
+When [`<Resource recordRepresentation>`](./Resource.md#recordrepresentation) is not defined, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
+1. `name`
+2. `title`
+3. `label`
+4. `reference`
+5. `id`
+
+
 
 ## Options
 
--- a/docs/useGetRecordRepresentation.md
+++ b/docs/useGetRecordRepresentation.md
@@ -9,7 +9,7 @@ Get a function that returns the record representation, leveraging the [`<Record 
 
 You can also use the component version: [`<RecordRepresentation>`](./RecordRepresentation.md).
 
-If you don't specify anything, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
+when `<Resource recordRepresentation>` is not defined, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
 1. `name`
 2. `title`
 3. `label`
--- a/docs/useGetRecordRepresentation.md
+++ b/docs/useGetRecordRepresentation.md
@@ -53,9 +53,9 @@ const PostEdit = () => (
 )
 ```
 
-## Preferences
+## Default Representation
 
-When [`<Resource recordRepresentation>`](./Resource.md#recordrepresentation) is not defined, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
+When [`<Resource recordRepresentation>`](./Resource.md#recordrepresentation) is not defined, `useGetRecordRepresentation` will return the first non-empty field from this list:  
 1. `name`
 2. `title`
 3. `label`
--- a/packages/ra-core/src/controller/record/RecordRepresentation.spec.tsx
+++ b/packages/ra-core/src/controller/record/RecordRepresentation.spec.tsx
@@ -10,7 +10,7 @@ import {
 describe('RecordRepresentation', () => {
     it('should render the record id when not provided on its parent <Resource>', async () => {
         render(<NoRecordRepresentation />);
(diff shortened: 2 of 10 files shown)


djhi (changes requested):

docs/useGetRecordRepresentation.md
```diff
@@ -9,6 +9,13 @@ Get a function that returns the record representation, leveraging the [`<Record
 
 You can also use the component version: [`<RecordRepresentation>`](./RecordRepresentation.md).
 
+If you don't specify anything, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
```
adguernier, line 12: **suggestion:** From my point of view this is not clear:: "If you don't specify anything", but where?
We could explain that this behavior occurs when `<Resource recordRepresentation>` is not defined. It makes more in my opinion.
Additionally, I suggest adding a mention in the `<RecordRepresentation>` doc, to explain that this component uses the `useRecordRepresentation` hook and that the same rules are therefore applied.
‏
‏

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -831,7 +831,7 @@ describe('<Edit />', () => {
                     </Edit>
                 </AdminContext>
             );
-            await screen.findByText('Foo #123');
+            await screen.findByText('Foo lorem');
```
adguernier, line 834: **issue:** The test expect `defaultTitle` to use the record id by default. But that is no longer the case.. We need to change the test title or remove the `title` in the `getOne` result.
What do you think @djhi ?‏

packages/ra-ui-materialui/src/detail/Show.spec.tsx
```diff
@@ -117,7 +117,7 @@ describe('<Show />', () => {
 
     it('should display a default title based on resource and id', async () => {
         render(<Basic />);
-        await screen.findByText('Book #1');
+        await screen.findByText('Book War and Peace');
```
adguernier, line 120: **issue:** The title no longer describes what the test does. We should rename the test IMO‏

adguernier (changes requested):

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
djhi: Indeed

docs/useGetRecordRepresentation.md
djhi: https://marmelab.com/react-admin/useGetRecordRepresentation.html#usage

packages/ra-core/src/core/useGetRecordRepresentation.ts
```diff
@@ -31,6 +31,18 @@ export const useGetRecordRepresentation = (
...
+                return record.name;
+            }
+            if ('title' in record) {
+                return record.title;
+            }
+            if ('label' in record) {
+                return record.label;
+            }
+            if ('reference' in record) {
+                return `#${record.reference}`;
```
fzaninotto, line 44: I'd lose the # decoration in that case

docs/useGetRecordRepresentation.md
```diff
@@ -53,6 +53,17 @@ const PostEdit = () => (
 )
 ```
 
+## Preferences
```
slax57, line 56: **question:** ‏Why "Preferences"?

docs/useGetRecordRepresentation.md
```diff
@@ -53,6 +53,17 @@ const PostEdit = () => (
...
+
+When [`<Resource recordRepresentation>`](./Resource.md#recordrepresentation) is not defined, the `useGetRecordRepresentation` function will choose the column to return according to this preference order:
```
slax57, line 58:
```suggestion
When [`<Resource recordRepresentation>`](./Resource.md#recordrepresentation) is not defined, `useGetRecordRepresentation` will return the first non-empty field from this list:

```

packages/ra-core/src/controller/record/RecordRepresentation.spec.tsx
```diff
@@ -10,7 +10,7 @@ import {
 describe('RecordRepresentation', () => {
     it('should render the record id when not provided on its parent <Resource>', async () => {
         render(<NoRecordRepresentation />);
-        await screen.findByText('#1');
+        await screen.findByText("The Hitchhiker's Guide to the Galaxy");
```
slax57, line 13: **polish:** ‏The test description should be updated accordingly

packages/ra-core/src/core/useGetRecordRepresentation.spec.tsx
```diff
@@ -93,4 +93,67 @@ describe('useRecordRepresentation', () => {
...
+                    label: 'lorem-ipsum',
+                    reference: '456',
+                    id: '123',
+                    author: 'John Doe',
+                }}
+            />
+        );
+        screen.getByText('Lorem');
+    });
+    it('should return the record label at second', () => {
```
slax57, line 127: **question:** ‏Why are there two "second"? :joy:

packages/ra-core/src/core/useGetRecordRepresentation.spec.tsx
```diff
@@ -93,4 +93,67 @@ describe('useRecordRepresentation', () => {
...
+    it('should return the record reference at third', () => {
+        render(
+            <UseRecordRepresentation
+                resource="users"
+                record={{ reference: '456', id: '123', author: 'John Doe' }}
+            />
+        );
+        screen.getByText('#456');
+    });
+    it('should return the record id by default', () => {
```
slax57, line 150: **praise:** ‏Otherwise nice tests :slightly_smiling_face:

packages/ra-core/src/core/useGetRecordRepresentation.ts
```diff
@@ -31,6 +31,18 @@ export const useGetRecordRepresentation = (
             if (React.isValidElement(recordRepresentation)) {
                 return recordRepresentation;
             }
+            if ('name' in record) {
+                return record.name;
+            }
```
slax57, line 36: **suggestion:** ‏Instead of checking of the field exists in the record, you should check that it is not null nor undefined nor an empty string. 

```suggestion
            if (record?.name != null && record?.name !=== '') {
                return record.name;
            }
```

Same applies below.

slax57 (changes requested):
Also, I'm surprise nobody raised that before, but since this is a **feature**, this PR should target `next` instead of `master` :grimacing:

erwanMarmelab:
> There are some warnings left in `useGetRecordRepresentation`

@djhi Yes, I just corrected the problem with the previous commit.

djhi (approved):

adguernier (changes requested):
Should target `next`

erwanMarmelab:
Closing this PR because there is a new one who can be merged on `next` branch #9650

PR #9647, opened 2024-02-06
Update DataProviders.md

davidhenley:
The name is actually either on the rawFile or the title prop. The File itself does not have a name property.

This can be `newPictures[index].title` or `newPictures[index].rawFile.name` but not what is there currently


slax57 (approved):
Right, nice catch! Thanks!

PR #9648, opened 2024-02-06
Fix field label translation concatenates previous prefix

adguernier:
## Problem
`<LabelPrefixContextProvider>` concatenates previous prefix if it is inside a `LabelPrefixContext`.

It's lead to return translation key such as `resources.posts.fields.title.resources.comments.fields.body` but it make more sense to simply return `resources.comments.fields.body` even `LabelPrefixContext` is nested inside another.

`<LabelPrefixContextProvider>` was introduce used in this #7710 but keeping previous prefix seems to be useful. 

## Solution
Do not keep previous prefix

### DIFF
--- a/packages/ra-core/src/form/Form.tsx
+++ b/packages/ra-core/src/form/Form.tsx
@@ -50,7 +50,10 @@ export const Form = (props: FormProps) => {
 
     return (
         <OptionalRecordContextProvider value={record}>
-            <LabelPrefixContextProvider prefix={`resources.${resource}.fields`}>
+            <LabelPrefixContextProvider
+                prefix={`resources.${resource}.fields`}
+                concatenate={false}
+            >
                 <FormProvider {...form}>
                     <FormGroupsProvider>
                         <form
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/util/LabelPrefixContextProvider.tsx
```diff
@@ -1,17 +1,9 @@
 import * as React from 'react';
 import { LabelPrefixContext } from './LabelPrefixContext';
-import { useLabelPrefix } from './useLabelPrefix';
 
-export const LabelPrefixContextProvider = ({
-    prefix,
-    concatenate = true,
```
djhi, line 7: Looks good but I wonder why we didn't set this to `false` on the `Form`

adguernier: Yes we could 😅. But `LabelPrefixContext` is not used elsewhere, so there is no point in keeping this logic IMO.

packages/ra-core/src/form/Form.tsx
```diff
@@ -50,7 +50,10 @@ export const Form = (props: FormProps) => {
 
     return (
         <OptionalRecordContextProvider value={record}>
-            <LabelPrefixContextProvider prefix={`resources.${resource}.fields`}>
+            <LabelPrefixContextProvider
+                prefix={`resources.${resource}.fields`}
+                concatenate={false}
+            >
```
slax57, line 56: **issue:** ‏We probably need to do the same thing in `FilterForm.tsx`

slax57 (changes requested):

erwanMarmelab (approved):

slax57 (approved):

PR #9649, opened 2024-02-07
Update Datagrid documention regarding preferenceKey

erwanMarmelab:
## Problem
In [React-admin - The Datagrid Component](https://marmelab.com/react-admin/Datagrid.html#configurable) , we don’t mention that you should pass the same preferenceKey to the SelectColumnsButton. That leads to unexpected behaviors when users forget it.

## Solution
Update the documentation and add an example

### DIFF
--- a/docs/Datagrid.md
+++ b/docs/Datagrid.md
@@ -881,9 +881,22 @@ const PostList = () => (
 );
 ```
 
-If you render more than one `<DatagridConfigurable>` in the same page, you must pass a unique `preferenceKey` prop to each one.
+If you render more than one `<DatagridConfigurable>` in the same page, you must pass a unique `preferenceKey` prop to each one:
 
-And if you include a `<SelectColumnsButton>` to your `<DatagridConfigurable>`, you have to link the two components by giving them the same preferenceKey:
+```tsx
+const PostList = () => (
+    <List>
+        <DatagridConfigurable preferenceKey="posts.datagrid">
+            <TextField source="id" />
+            <TextField source="title" />
+            <TextField source="author" />
+            <TextField source="year" />
+        </DatagridConfigurable>
+    </List>
+);
+```
+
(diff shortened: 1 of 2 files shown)


docs/Datagrid.md
```diff
@@ -896,6 +896,22 @@ const PostList = () => (
...
+const PostList = () => (
+    <List>
+        <SelectColumnsButton preferenceKey="posts.datagrid" />
+        <DatagridConfigurable preferenceKey="posts.datagrid">
+            <TextField source="id" />
+            <TextField source="title" />
+            <TextField source="author" />
+            <TextField source="year" />
+        </DatagridConfigurable>
+    </List>
```
slax57, line 911: **suggestion:** ‏This code snippet does not show a good example of how you should use `SelectColumnsButton` (it should be used inside the List `actions`). Please create an example based on [this one](https://marmelab.com/react-admin/SelectColumnsButton.html#usage) instead.

slax57 (changes requested):

docs/SelectColumnsButton.md
```diff
@@ -65,30 +65,22 @@ const ListActions = () => (
...
-};
+const BookListActions = () => (
+    <TopToolbar>
+        <SelectColumnsButton preferenceKey="postList1" />
+    </TopToolbar>
+);
+
+const BookList = () => (
+    <List actions={<BookListActions />}>
+        <DatagridConfigurable preferenceKey="postList1">
```
adguernier, line 76: **issue:** ‏Strange that the  `preferenceKey` is `postList1` than component is called `<BookList>`. Based on the other example (`<PostList>`), we could set `preferenceKey` to `books.datagrid`

adguernier (changes requested):

slax57 (approved):

PR #9657, opened 2024-02-09
Upgrade `react-router` to 6.22.0, use data router, stabilize `useWarnWhenUnsavedChanges` and remove `<Admin history>` prop

slax57:
## Todo

- [x] Upgrade `react-router` to 6.22.0
- [x] Rewrite `useWarnWhenUnsavedChanges` to use `useBlocker` instead of `UNSAFE_NavigationContext`
- [x] Use data router in the simple demo
- [x] Test and compare with previous version of `useWarnWhenUnsavedChanges`
- [x] Implement blocking when closing the browser tab
- [x] Use data router in ra-core
- [x] Remove (deprecated) `history` prop
- [x] Update all tests that use `history`
- [x] Provide a codemod
- [x] Clean up test code
- [x] Remove dependency on `history`
- [x] Write the Upgrade guide

### Technical notes

- This new implementation no longer waits until the submission is complete before redirecting. Instead, it blocks navigation and opens the confirmation dialog if the form is currently submitting. While the dialog is open, submission continues in the background.
- Implementing blocking when closing the browser tab now requires listening to the `'beforeunload'` event. [Modern browsers no longer allow to customize the message](https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup) displayed in the confirmation dialog, so we can only display the default one.


packages/ra-core/src/form/useWarnWhenUnsavedChanges.tsx
```diff
@@ -14,66 +13,82 @@ export const useWarnWhenUnsavedChanges = (
...
+            if (shouldNotBlock) {
+                blocker.proceed();
+                return;
             }
-            return;
+
+            setShouldNotify(true);
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [blocker.state]);
```
fzaninotto, line 57: shouldNotBlock is missing here. It seems you know why but the reader of that code won't. Would you mind explaining it in a comment?

fzaninotto:
needs rebase

packages/ra-core/src/controller/list/useListParams.spec.tsx
```diff
@@ -414,21 +441,37 @@ describe('useListParams', () => {
...
+                <TestMemoryRouter
+                    locationCallback={l => {
+                        location = l;
+                    }}
                 >
-                    <Component />
-                    <StoreReader />
-                </CoreAdminContext>
+                    <CoreAdminContext dataProvider={testDataProvider()}>
+                        <Component disableSyncWithLocation />
```
slax57, line 450: I suspect this test was buggy in the first place, as it did not set `disableSyncWithLocation` although it is necessary to disable sync. Hopefully by adding `await screen.findByText('page: 10');` I have fixed it.

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
```diff
@@ -102,16 +102,13 @@ describe('useDataProvider', () => {
...
-        expect(r).toThrow(
-            new Error(
-                'The dataProvider threw an error. It should return a rejected Promise instead.'
-            )
+        render(
+            <CoreAdminContext dataProvider={dataProvider}>
+                <UseGetOne />
+            </CoreAdminContext>
+        );
+        await screen.findByText(
```
slax57, line 110: I suspect these tests no longer work because the new data router takes one extra tick to actually render the child route. I fixed this test (an similar others) by leveraging the fact that react-admin includes an `ErrorBoundary` in the default layout, which will render any rendering-phase error message on screen.

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -207,17 +209,18 @@ export const OptionTextElement = () => (
...
         <AutocompleteInput
             source="author"
-            choices={[
-                { id: 1, name: 'Leo Tolstoy' },
-                { id: 2, name: 'Victor Hugo' },
-                { id: 3, name: 'William Shakespeare' },
-                { id: 4, name: 'Charles Baudelaire' },
-                { id: 5, name: 'Marcel Proust' },
-            ]}
+            choices={choicesForCreationSupport}
```
slax57, line 223: This fixes a regression introduced by https://github.com/marmelab/react-admin/pull/9678

packages/ra-ui-materialui/src/input/TextInput.stories.tsx
```diff
@@ -12,7 +12,7 @@ import { FormInspector } from './common';
 
 export default { title: 'ra-ui-materialui/input/TextInput' };
 
-export const Wrapper = ({ children }) => (
+const Wrapper = ({ children }) => (
```
slax57, line 15: I don't think exporting this component as a story was intended

packages/ra-core/codemods/replace-Admin-history.ts
```diff
@@ -0,0 +1,154 @@
+import j from 'jscodeshift';
```
fzaninotto, line 1: 👍  I'm so glad you commented this file thoroughly!

docs/Upgrade.md
```diff
@@ -482,6 +482,109 @@ import { FieldProps, useRecordContext } from 'react-admin';
...
+> **DISCLAIMER**
+>
+> This codemod was used to migrate the react-admin test suite, but it was never designed to cover all cases, and was not tested against other code bases. You can try using it as basis to see if it helps migrating your code base, but please review the generated changes thoroughly!
+>
+> Applying the codemod might break your code formatting, so please don't forget to run `prettier` and/or `eslint` after you've applied the codemod!
+
+For `.js` or `.jsx` files:
+
+```sh
+npx jscodeshift ./path/to/src/ \
```
fzaninotto, line 574: **Question**: jscodeshift knows how to execute TypeScript files? Users don't need to use a transpiled JS version (that we currently don't publish)?

packages/ra-core/codemods/__testfixtures__/replace-Admin-history-useEditController.input.tsx
```diff
@@ -0,0 +1,1057 @@
+import {
```
fzaninotto, line 1: **Nitpick**: maybe a bit too long for just a test...

docs/Upgrade.md
```diff
@@ -482,6 +482,109 @@ import { FieldProps, useRecordContext } from 'react-admin';
...
+    --extensions=js,jsx \
+    --transform=./node_modules/ra-core/codemods/replace-Admin-history.ts
```
fzaninotto, line 576: There are 2 codemods, you only document one. By the way, shouldn't we present only one codemode doing everything?

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -1,8 +1,7 @@
 import { useCallback } from 'react';
 // @ts-ignore
 import { parse } from 'query-string';
-import { useLocation } from 'react-router-dom';
-import { Location } from 'history';
+import { useLocation, Location } from 'react-router-dom';
```
fzaninotto, line 4: 👍 I believe the codemod didn't catch this one, am I right?

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
fzaninotto: **Nitpick**: You should probably change the test description then

fzaninotto (changes requested):
46/114 files reviewed. So far, it's excellent, kudos!

packages/ra-core/src/routing/useCreatePath.spec.tsx
```diff
@@ -4,6 +4,7 @@ import { render, screen } from '@testing-library/react';
 import { CreatePathType, useCreatePath } from './useCreatePath';
 import { AtRoot, SubPath } from './useCreatePath.stories';
 import { Identifier } from '../types';
+import { TestMemoryRouter } from '.';
```
fzaninotto, line 7:
```suggestion
import { TestMemoryRouter } from './TestMemoryRouter';
```

packages/ra-core/src/routing/useCreatePath.spec.tsx
```diff
@@ -50,7 +51,11 @@ describe('useCreatePath', () => {
     });
 
     it('creates valid links when used without a basename', async () => {
-        render(<AtRoot />);
+        render(
+            <TestMemoryRouter>
```
fzaninotto, line 55: I wonder why you removed the memory router wrapping in the story to add it back here. Would you care to explain?

packages/ra-core/src/storybook/FakeBrowser.tsx
```diff
@@ -56,33 +36,26 @@ const Browser = ({
...
-        const unsubscribe = history.listen(update => {
-            setLocation(update.location);
-        });
+const BrowserBar = () => {
+    const location = useLocation();
+    const navigate = useNavigate();
 
-        return unsubscribe;
-    }, [history]);
+    const fullLocation = `${location.pathname}${location.search}`;
```
fzaninotto, line 50: shouldn't you also include `location.hash`?

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
```diff
@@ -91,23 +95,33 @@ const PostList = () => (
 );
 
 export const InsideAList = () => (
-    <AdminContext dataProvider={getDataProvider()} i18nProvider={i18nProvider}>
-        <AdminUI>
-            <Resource name="posts" list={<PostList />} show={<PostShow />} />
-        </AdminUI>
-    </AdminContext>
+    <TestMemoryRouter>
```
fzaninotto, line 98: **Question**: this test didn't use a custom history before. Did it break if you didn't wrap it with TestMemoryRouter (which would be a sign of BC break), or did you just wrap it for consistency and isolation with the other tests?

fzaninotto (changes requested):
Finished my review. I only have a few questions and nitpicks. This PR is great!

By the way, you forgot one remaining MemoryRouter in `BulkUpdateButton.stories.tsx`

docs/Upgrade.md
slax57: It is my understanding that is does: https://github.com/facebook/jscodeshift?tab=readme-ov-file#transform-module:~:text=As%20of%20v0.6.1%2C%20this%20module%20can%20also%20be%20written%20in%20TypeScript.

docs/Upgrade.md
slax57: Indeed.

- the `replace-Admin-history.ts` codemod helps replacing the `history` prop by a `<TestMemoryRouter>`, which is what I document in the Upgrade guide anyway
- the `replace-MemoryRouter.ts` codemod is required to replace instances of `<MemoryRouter>`, which is **not** a data router, by `<TestMemoryRouter>`, which is a data router. Technically this change is only required for tests (and stories) including a `<Form>`.

I figured I could not include this last codemod in the Upgrade guide without getting too much into technical details about why this is needed. So for the sake of simplicity I chose to omit it. But I'm open to discussion on the matter.

And, to me, both codemods serve a different purpose, so I thought it would be clearer (and simpler to maintain and to test) to keep them separate. Again, I'm open to discussion here.

packages/ra-core/src/controller/create/useCreateController.ts
slax57: Actually the codemod removes all imports from `history`, not only `createMemoryHistory`, so I saw it thanks to a combination of the codemod and TypeScript complaining about the type from `Location` being incorrect.

packages/ra-core/src/form/useWarnWhenUnsavedChanges.spec.tsx
```diff
@@ -72,18 +73,23 @@ const FormUnderTest = () => {
...
-    <MemoryRouter initialEntries={initialEntries} initialIndex={0}>
-        <Routes>
-            <Route path="/form" element={<FormUnderTest />} />
-            <Route path="/form/show" element={<span>Show</span>} />
-            <Route path="/form/:part" element={<FormUnderTest />} />
-            <Route path="/submitted" element={<span>Submitted</span>} />
-            <Route path="/somewhere" element={<span>Somewhere</span>} />
-        </Routes>
-    </MemoryRouter>
+const Root = () => (
```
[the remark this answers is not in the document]
slax57: The reason is simply because I updated these tests prior to introducing the `TestMemoryRouter` :sweat_smile: . Fixed.

packages/ra-core/src/routing/useCreatePath.spec.tsx
slax57: It's because the story previously relied on `FakeBrowserDecorator` to provide a memory/hash history, and was autonomous to create a new router if  `FakeBrowserDecorator` was absent. Now, `FakeBrowserDecorator` provides its own router, but the story in standalone use no longer does it, so we need to provide the router in the tests.

Hopefully my explanations are clear enough :sweat_smile:

packages/ra-ui-materialui/src/button/UpdateButton.stories.tsx
slax57: Yes it is for isolation with the other tests. The story works fine without it but you may land in to a non existing route if you are coming from another story (also not using TestMemoryRouter)

fzaninotto (approved):

Issue #9655, opened 2024-02-09
Test failure caused by ra-material-ui@4.16.8+


Dreamsorcerer:
This is going to be tricky for me to create a reproducer here, so I'm hoping the issue will be obvious to one of you.

When using resolutions, if I set ra-material-ui to 4.16.7 then my tests pass. At 4.16.8+ they start failing.
Looking through the small number of changes in that release, this seems like the only obvious change that could have caused it:
https://github.com/marmelab/react-admin/pull/9600/files#diff-5f9cd2ee8b802d89b35ab1d0f4df69abfd29f37dc0208f4a836c4d6dbb629266

The test code looks like:
```javascript
        const container = await screen.findByRole("columnheader", {"name": "Select all"});
        const selectAll = within(container).getByRole("checkbox");
        await userEvent.click(selectAll);
        expect(selectAll).toBeChecked();
        // This is a bulk update button.
        await userEvent.click(await screen.findByRole("button", {"name": "Set to 7"}));
        // This is the confirmation dialog (as a result of mutationMode=pessimistic).
        expect(await screen.findByText("Update 6 simples")).toBeInTheDocument();
        await userEvent.click(screen.getByRole("button", {"name": "Confirm"}));
        await waitFor(() => screen.getAllByText("7"));
```

The test now fails with (https://github.com/aio-libs/aiohttp-admin/actions/runs/7705984738/job/21389490839?pr=854):
```
Unable to find an element with the text: 7
...

      50 |         expect(await screen.findByText("Update 6 simples")).toBeInTheDocument();
      51 |         await userEvent.click(screen.getByRole("button", {"name": "Confirm"}));
    > 52 |         await waitFor(() => screen.getAllByText("7"));
         |                      ^
      53 |
      54 |         const table = await screen.findByRole("table");
      55 |         const rows = within(table).getAllByRole("row");
```

Using screen.debug(), I can see that the confirmation dialog is still visible in the HTML. It hasn't closed since clicking the button.

My guess is that without the refresh() call, the confirmation dialog doesn't get removed in jest-dom.
The dialog seems to work correctly when I use the app in Firefox though, so this seems to only affect tests.

Dreamsorcerer:
Also, this would've been a little easier to detect and avoid breaking things if react-admin actually depended on pinned versions of the ra-* packages. Given that all packages are released at the same time, I don't see any reason it would be desirable to install newer versions of a ra-* package than the react-admin version.

[...]

PR #9659, opened 2024-02-11
Test for #9655

Dreamsorcerer:
#9655


fzaninotto:
If you're not able to build a unit test demonstrating your problem, perhaps you could fork the [simple example codesandbox](https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple) and tweak it until it show the problem? Or does it only show in unit tests?

Dreamsorcerer:
> Or does it only show in unit tests?

Yes, I mentioned it's only causing a test failure. It seems the dialog doesn't get closed in jest-dom, whereas it does get closed in Firefox when using it as a normal user.

fzaninotto:
In that case, we'll wait until you can provide a unit test that fails because of the issue you mentioned.

Dreamsorcerer:
Any tips on how to make it display some results? I'm copying from other, similar tests in this project, so can't see why it's not working here if it's working in those other tests..

slax57:
> Any tips on how to make it display some results? I'm copying from other, similar tests in this project, so can't see why it's not working here if it's working in those other tests..

Tests results are visible in the logs:
https://github.com/marmelab/react-admin/actions/runs/7863886168/job/21455003047?pr=9659#step:5:3073

If you cloned the project locally and already ran `make install` successfully, you should be able to run the tests and iterate on it by running `make test-unit-watch`.

Dreamsorcerer:
> > Any tips on how to make it display some results? I'm copying from other, similar tests in this project, so can't see why it's not working here if it's working in those other tests..
> 
> Tests results are visible in the logs: https://github.com/marmelab/react-admin/actions/runs/7863886168/job/21455003047?pr=9659#step:5:3073

Sorry, I don't think I was clear. I meant, if you look at the test output it renders `ra.navigation.no_results`. I don't understand why the list view has no results when I've added a dataProvider with getOne, getList and getMany. It looks similar to the other tests in this project...

slax57:
Thanks for the additional explanations. 

I believe the list is empty because you are providing a `ListContextProvider` with no data. If you expect the list to call `getList`, you should use the `useListController` via either `List` or `ListBase`.

fzaninotto:
May I suggest you test your changes locally before pushing them? We get notified of each of your tries.

Dreamsorcerer:
I wasn't able to run the tests locally using the instructions. Currently, the record data seems to be failing to update from the bulk update button entirely. Not sure why this isn't working...

fzaninotto:
We alredy have a story that shows

fzaninotto:
Sorry, I closed this PR by mistake. 

I have created a test for the BulkUpdateButton in #9826, and it seems to work fine. Unless you can tweak this test to make it fail, we'll consider that the problem is on your side.

fzaninotto:

PR #9666, opened 2024-02-15
[Doc] Add video tutorial to the Datagrid doc

fzaninotto:


erwanMarmelab:
**Suggestion:** Shouldn’t we add it to the `Feature` doc page too?
And maybe remove `datagrid.mp4`

fzaninotto:
I think the features page should be easily scannable. Imbedding a video there would break the reading flow. 

datagrid.mp4 is still used in the Features page once we fix the typo in the file path :)

slax57 (approved):

PR #9671, opened 2024-02-16
Fix  typo in jsonexport

kav:
jsonexport is a JSON to CSV converter.  Not a CSV to JSON converter. Minor note but it is confusing as written.


slax57 (approved):
Thanks!

PR #9676, opened 2024-02-20
[WIP] add use of rest parameters in combineDataProviders

### LINKED ISSUE #9675: combineDataProviders does not support more than 2 parameters in custom dataProvider methods
**What you were expecting:**
According to the documentation examples, custom dataProvider methods should not follow the same signature as default methods, in example it's `banUser: (userId) => Promise`, hence I expect to be able to have any signature that is required for my needs, including functions with more than 2 parameters

**What happened instead:**
combineDataProviders calls dataProvider methods limiting it to 2 params: 
```dataProviderMatcher(resource)[name](resource, params)```

**Steps to reproduce:**
- create a custom method in dataProvider
- use combineDataProviders when you attach dataProvider to the admin context
- call a custom method and you will get undefined on 3rd and following parameters.

**Related code:**
[Example on StackBlitz](https://stackblitz.com/edit/github-n69bqi?file=src%2FdataProvider.tsx)
Here in `src/dataProvider.tsx` I've created a new method called customMethod with 3rd parameter.
In `src/Layout.tsx` I call this method and I expect different console.log outputs but always get the falsy one.

Example of this custom method:
```
// Always falsy
  customMethod: (resource, params, optionalFlag) => {
    console.log(resource, params, optionalFlag);
    if (optionalFlag === true) {
      console.log('data provider customMethod optionalFlag is true');
    } else {
      console.log('data provider customMethod optionalFlag is false');
    }
  },
```

**Environment**

* React-admin version: 4.16.11
* React version: 17
* Browser: Firefox Developer Edition

karpushchenko:
Closes #9675
The current implementation doesn't work with dataProvider methods with more than 2 parameters.  According to documentation, the method parameter format is not strictly typed, so it is possible to add as many params as needed and then guess why combineDataProviders doesn't work

### DIFF
--- a/packages/ra-core/src/dataProvider/combineDataProviders.spec.ts
+++ b/packages/ra-core/src/dataProvider/combineDataProviders.spec.ts
@@ -27,4 +27,26 @@ describe('combineDataProviders', () => {
         expect(dataProvider1.getOne).not.toHaveBeenCalled();
         expect(dataProvider2.getOne).toHaveBeenCalled();
     });
+    it('calls custom dataProvider methods with any number of parameters', async () => {
+        const dataProviderWithCustomMethod = testDataProvider({
+            customMethod: jest
+                .fn()
+                .mockResolvedValue({ data: { id: 1, foo: 'bar' } }),
+        });
+        const dataProvider = combineDataProviders(resource => {
+            switch (resource) {
+                case 'custom':
+                    return dataProviderWithCustomMethod;
+                default:
+                    throw new Error('Unknown resource');
+            }
+        });
+        dataProvider.customMethod('custom', 1, 2, 3);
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks. Could you please add unit tests?

karpushchenko:
> Thanks. Could you please add unit tests?

Yep, why not, I will update it probably tomorrow

karpushchenko:
@fzaninotto updated with a corresponding test

fzaninotto (changes requested):
Great! But this is technically a new feature, so you should PR against `next` rather than `master`.

karpushchenko:
Ok, I will update it later today, not just a simple retarget of PR

karpushchenko:
@fzaninotto, done! Sorry for the delay

fzaninotto:
Thanks!

PR #9677, opened 2024-02-23
Keep errors and other state on ArrayInput unmount

sweco-sedalh:
We have a rather large form, in which we for optimization reasons unmount parts of the form when not visible. This doesn't work to well with `<ArrayInput />` however, which clears all state (errors, touched, etc.) when it is unmounted, which both means that the general valid/invalid state of the form is no longer correct.

To reproduce:

1. Setup a react-admin project with an `<ArrayInput />` that can be unmounted and some field that can have errors as child
2. Open, enter an invalid value, try to save, field gets marked as error
3. Trigger unmounting and remounting of `<ArrayInput />`
4. The error is gone, despite the value still being invalid

Here is a reproducible example: https://stackblitz.com/edit/github-ntzyud-hb8uu4?file=src%2Fposts%2FPostEdit.tsx (note that tabbed forms have a similar issue, however that can more easily be worked around in user-space code)

This fix uses the `react-hook-form` options to keep the state of the `<ArrayInput />` when it is unmounted.

I'm not 100% sure if this might break BC (I think it would be possible to build a form that depends on the current behavior, but I don't think it would make much sense to do so), if that is the case let me now and I'll retarget this PR to the `next` branch.

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
@@ -256,6 +256,60 @@ describe('<ArrayInput />', () => {
         });
     });
 
+    it('should not clear errors of children when unmounted', async () => {
+        let setArrayInputVisible;
+
+        const MyArrayInput = () => {
+            const [visible, setVisible] = React.useState(true);
+
+            setArrayInputVisible = setVisible;
+
+            return visible ? (
+                <ArrayInput resource="bar" source="arr">
+                    <SimpleFormIterator>
+                        <TextInput source="id" />
+                        <TextInput source="foo" />
+                    </SimpleFormIterator>
+                </ArrayInput>
+            ) : null;
+        };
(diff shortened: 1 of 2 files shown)


slax57 (commented):
Thanks for your contribution, and for the detailed explanation!

AFAICT this goes in the right direction, and fixes the form's behavior to make it more coherent.

I can't think of a use-case where this would be a breaking change, since you already cannot validate the form if the ArrayInput is invalid, even unmounted. This fix simply allows to see the validation error at the correct time.

However I'll request the review of @djhi  or @fzaninotto , just to be on the safe side.

Also, it would be nice if you could add a unit test covering this case, so that we don't run into this issue in the future. Can you try adding a new test?

Thanks

fzaninotto:
Seems fair to me, too. I agree this needs unit tests to prove that it fixes something, and that we won't break it in the future.

sweco-sedalh:
Done!

slax57 (approved):
Thanks!

PR #9679, opened 2024-02-24
Added queryOptions to ReferenceManyField (#9669)

### LINKED ISSUE #9669: Add a meta prop to react-admin reference components
I need to pass a boolean flag to the backend when using ReferenceManyField.

A meta prop that can be used to send data in the request using the optional meta parameter in the dataProvider.
i.e. `<ReferenceManyField ... meta={orm: true}>` would result in a referenceMany() call with `meta={orm: true}`.

As a simple workaround, I've instead done `filter={'__meta__': {orm: true}}` and then popped it out of the filters in the backend. Though this is a bit of a hack and makes the code less maintainable.

Based on the documentation, it makes it sound like the recommendation is to write a custom component. For such a simple thing, this appears to require copying hundreds of lines of code out of react-admin and maintaining it, which doesn't seem like a viable solution.

Nela62:
Closes #9669 

- Added `queryOptions` to `ReferenceManyField.tsx` and `useReferenceManyFieldController.ts`
- Added `<RecordType>` to  `const record = useRecordContext(props)` in `ReferenceManyField.tsx`

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -26,6 +26,7 @@ export interface UseReferenceManyFieldControllerParams<
     sort?: SortPayload;
     source?: string;
     target: string;
+    queryOptions?: { meta?: any };
 }
 
 const defaultFilter = {};
@@ -72,8 +73,10 @@ export const useReferenceManyFieldController = <
         page: initialPage,
         perPage: initialPerPage,
         sort: initialSort = { field: 'id', order: 'DESC' },
+        queryOptions = {},
     } = props;
     const notify = useNotify();
+    const { meta } = queryOptions;
     const resource = useResourceContext(props);
 
     // pagination logic
@@ -178,6 +181,7 @@ export const useReferenceManyFieldController = <
             pagination: { page, perPage },
             sort,
             filter: filterValues,
+            meta,
         },
         {
             enabled: get(record, source) != null,
--- a/packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
+++ b/packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
@@ -77,8 +77,9 @@ export const ReferenceManyField = <
         sort = defaultSort,
         source = 'id',
         target,
+        queryOptions,
     } = props;
-    const record = useRecordContext(props);
+    const record = useRecordContext<RecordType>(props);
 
     const controllerProps = useReferenceManyFieldController<
         RecordType,
@@ -94,6 +95,7 @@ export const ReferenceManyField = <
         sort,
         source,
         target,
+        queryOptions,
     });
 
     return (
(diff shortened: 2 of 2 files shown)


packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -178,6 +181,7 @@ export const useReferenceManyFieldController = <
             pagination: { page, perPage },
             sort,
             filter: filterValues,
+            meta,
         },
```
fzaninotto, line 185: You should also pass the remaining quetry options to the query, see e.g. https://github.com/marmelab/react-admin/blob/5e2ed7d3f43c0a849ae33e18737ef83429dad639/packages/ra-core/src/controller/edit/useEditController.ts#L99

packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -26,6 +26,7 @@ export interface UseReferenceManyFieldControllerParams<
     sort?: SortPayload;
     source?: string;
     target: string;
+    queryOptions?: { meta?: any };
```
fzaninotto, line 29: Bad type. This should be ` UseGetManyReferenceHookOptions<RecordType>`

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
```diff
@@ -118,6 +120,7 @@ export interface ReferenceManyFieldProps<
     reference: string;
     sort?: SortPayload;
     target: string;
+    queryOptions?: { meta?: any };
```
fzaninotto, line 123: same

fzaninotto (changes requested):
This is a new feature, so it has to target `next` rather than `master`. Also, please add a unit test to make sure the feature works (e.g. https://github.com/marmelab/react-admin/blob/5e2ed7d3f43c0a849ae33e18737ef83429dad639/packages/ra-core/src/controller/edit/useEditController.spec.tsx#L176-L201)

packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
Nela62: Do you mean that I should change it like this?
```javascript
export interface UseGetManyReferenceHookOptions<RecordType> {
  meta?: any
}
...
  queryOptions?: UseGetManyReferenceHookOptions<RecordType>
```

This is the type of `queryOptions` in `ReferenceOneField.tsx`, which I used as a reference
```javascript
queryOptions?: UseQueryOptions<{
        data: ReferenceRecordType[];
        total: number;
    }> & { meta?: any };
```

slax57: @Nela62 

Indeed there is no `UseGetManyReferenceHookOptions`. But I believe what @fzaninotto means is that you should use the same type as what `useGetManyReference()` expects, plus `meta`, i.e.:

```ts
queryOptions?: UseQueryOptions<{ data: RecordType[]; total: number }, Error> & { meta?: any };
```

fzaninotto:

PR #9686, opened 2024-02-27
Make input width consistent by default

fzaninotto:
## Problem

MUI doesn't defaine a consistent input width, and neither do we. As a consequence, forms mixing various input types don't show a proper alignment by default.

See [this story](https://react-admin-storybook.vercel.app/?path=/story/ra-ui-materialui-input--all-inputs) for an illustration. Refs #8796.

## Solution

Define a minimum input width in each themes.

| Before | After | 
| ----- | ----- |
|  image  | image |

The story has been updated to let the user test all existing themes, as well as an empty theme (default MUI) and a compat theme (to get the same input widths as in react-admin v4)

### DIFF
--- a/docs/AppTheme.md
+++ b/docs/AppTheme.md
@@ -514,6 +514,12 @@ const defaultThemeInvariants = {
                 margin: 'dense' as const,
                 size: 'small' as const,
             },
+            variants: [
+                {
+                    props: {},
+                    style: () => ({ minWidth: 210 }),
+                },
+            ],
         },
     },
 };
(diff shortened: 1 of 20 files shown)


fzaninotto:
After thinking it through, I think the hassle isn't worth the benefit. Users will wonder why setting a custom `width` doesn't shrink an input. And if they worry about the alignment problem, it's easy enough to solve. 

I think we should only document the 2 techniques for sizing inputs:

1. make all inputs fullWidth, and lay them out using a layout component (`Grid` or `Stack`)
2. Alter the theme to add a `minWidth`

fzaninotto:
So after discussion with the rest of the core team, we won't move forward with a default `minWidth` for inputs as it would cause too much problems with custom (short) widths. 

So I'm closing this PR.

PR #9690, opened 2024-02-29
Fix `<ReferenceInput>` throws a recoverable error in production

fzaninotto:
## Problem

Since #9637, users see an error in production while the app isn't broken

## Solution

Only throw in development.

### DIFF
--- a/packages/ra-ui-materialui/src/input/ReferenceInput.tsx
+++ b/packages/ra-ui-materialui/src/input/ReferenceInput.tsx
@@ -84,13 +84,10 @@ export const ReferenceInput = (props: ReferenceInputProps) => {
         filter,
     });
 
-    if (props.validate) {
+    if (props.validate && process.env.NODE_ENV !== 'production') {
         throw new Error(
             '<ReferenceInput> does not accept a validate prop. Set the validate prop on the child instead.'
         );
-    }
-    if (Children.count(children) !== 1) {
-        throw new Error('<ReferenceInput> only accepts a single child');
     }
 
     return (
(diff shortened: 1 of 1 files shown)


adguernier (approved):

digilist (approved):

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -84,14 +84,11 @@ export const ReferenceInput = (props: ReferenceInputProps) => {
...
 
-    if (props.validate) {
+    if (props.validate && process.env.NODE_ENV !== 'production') {
         throw new Error(
             '<ReferenceInput> does not accept a validate prop. Set the validate prop on the child instead.'
         );
     }
-    if (Children.count(children) !== 1) {
-        throw new Error('<ReferenceInput> only accepts a single child');
-    }
```
slax57, line 94: **Question:** Why did you remove this test?

Also, this creates a linter warning at line 1.

slax57 (changes requested):

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
fzaninotto: I removed it because it was only required when we cloned the child. We no longer do that, so there is no need to force a single child.

fzaninotto: Also, I fixed the warning

slax57 (approved):

PR #9697, opened 2024-03-04
[Doc] Fix not parsed mutationOptions value in useNotify/undoable

ValentinnDimitroff:
Fixed not properly parsed jsx in the useNotify docs page for the `mutationOptions`

image


djhi (approved):
Thanks!

PR #9700, opened 2024-03-04
[Doc] Update doc for Search and SearchWithResult

erwanMarmelab:
## Problem
Search and SearchWithResult components now ignore the `historySize` and `options` props

## Solution
Update the doc


fzaninotto:
I don't see a correletion between the PR description and its changeset...

djhi (approved):

PR #9712, opened 2024-03-12
Fix `<AutocompleteInput createLabel>`

### LINKED ISSUE #9616: AutocompleteInput createLabel prop does not work.
Either the docs need to be updated or the functionality needs to be added.

We still REALLY need this functionality for our office applications as it is not apparent that you can start typing to create a new item. 🤞

https://github.com/marmelab/react-admin/issues/7822

https://marmelab.com/react-admin/AutocompleteInput.html#create

CleanShot 2024-01-25 at 21 20 31@2x

fzaninotto:
Closes #9616

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteArrayInput.spec.tsx
@@ -809,7 +809,7 @@ describe('<AutocompleteArrayInput />', () => {
                 selector: 'input',
             })
         );
-        expect(screen.queryAllByRole('option')).toHaveLength(2);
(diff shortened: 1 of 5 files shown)


erwanMarmelab (approved):

davidhenley:
Just so you guys know, after updating, the create is now passing the entire "Create xxx" to the CreateSuggestionContext filter. So if I type "Test" it will pass to "Create Test" to the filter when before this it would just pass "Test"

erwanMarmelab:
> useCreateSuggestionContext

Thanks for the report, do not hesitate to create an issue with a CodeSandbox :+1:

davidhenley:
@erwanMarmelab Sorry I wish I had the time. I ended up just having to revert back to the last version to fix it.

PR #9723, opened 2024-03-14
Fix react-admin requires custom routers to be data routers

fzaninotto:
The reimplentation of the `warnWhenUnsavedChanges` feature with react-router's `useBlocker` in #9657 created a breaking change. As `useBlocker` throws an error when not called in a DataRouter, and since we cannot call hooks conditionally, this meant that react-admin MUST be used inside a data router.

By moving the `warnWhenUnsavedChanges` logic from the `useAugmentedForm` hook to a custom component, we can now call the component conditionally - and avoid calling it when not inside a data router.

### DIFF
--- a/docs/AccordionForm.md
+++ b/docs/AccordionForm.md
@@ -351,7 +351,7 @@ export const TagEdit = () => (
 );
 ```
 
-**Note**: Due to limitations in react-router, this feature only works if you use the default router provided by react-admin, or if you use a Data Router.
+**Note**: Due to limitations in react-router, this feature only works if you use the default router provided by react-admin, or if you use a [Data Router](https://reactrouter.com/en/6.22.3/routers/picking-a-router).
 
 ## `sx`: CSS API
 
--- a/docs/AccordionForm.md
+++ b/docs/AccordionForm.md
@@ -351,7 +351,7 @@ export const TagEdit = () => (
 );
 ```
 
-**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
+**Note**: Due to limitations in react-router, this feature only works if you use the default router provided by react-admin, or if you use a Data Router.
 
 ## `sx`: CSS API
 
--- a/docs/Upgrade.md
+++ b/docs/Upgrade.md
@@ -622,7 +622,7 @@ The `warnWhenUnsavedChanges` feature is a little more restrictive than before:
 
 This behavior allows to prevent unwanted data loss in more situations. No changes are required in the code.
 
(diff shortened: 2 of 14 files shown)


packages/ra-core/src/form/Form.tsx
```diff
@@ -1,11 +1,15 @@
...
-import { ReactNode } from 'react';
+import { ReactNode, useContext } from 'react';
 import {
     FormProvider,
     FieldValues,
     UseFormProps,
     SubmitHandler,
 } from 'react-hook-form';
+import {
+    UNSAFE_DataRouterContext,
```
fzaninotto, line 10: Unfortunately react-router doesn't expose the DataRouterContext, see https://github.com/remix-run/react-router/discussions/11347

docs/AccordionForm.md
```diff
@@ -351,7 +351,7 @@ export const TagEdit = () => (
 );
 ```
 
-**Warning**: This feature only works if you have a dependency on react-router 6.3.0 **at most**. The react-router team disabled this possibility in react-router 6.4, so `warnWhenUnsavedChanges` will silently fail with react-router 6.4 or later.
+**Note**: Due to limitations in react-router, this feature only works if you use the default router provided by react-admin, or if you use a Data Router.
```
slax57, line 354: nitpick: I'd add a link to react-router's documentation for those not knowing what a Data Router is

docs/Upgrade.md
```diff
@@ -613,11 +613,20 @@ import { FieldProps, useRecordContext } from 'react-admin';
...
 
-For instance, if you were using `react-router`'s `BrowserRouter`, you will need to migrate to `createBrowserRouter` and wrap your app in a `RouterProvider`:
+- It will open a confirmation dialog (and block the navigation) if a navigation is fired when the form is currently submitting (submission will continue in the background).
+- [Due to browser constraints](https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup), the message displayed in the confirmation dialog when closing the browser's tab cannot be customized (it is managed by the browser).
+
+This behavior allows to prevent unwanted data loss in more situations. No changes are required in the code.
+
+The `warnWhenUnsavedChanges` requires a Data Router (a new type of router from react-router) to work. React-admin uses such a data router by default, so the feature works out of the box in v5. 
+
+However, if you use a [custom router](./Routing.md#using-a-custom-router) and the `warnWhenUnsavedChanges` prop, the "warn when unsaved change"s feature will be disabled.
```
slax57, line 627:
```suggestion
However, if you use a [custom router](./Routing.md#using-a-custom-router) and the `warnWhenUnsavedChanges` prop, the "warn when unsaved changes" feature will be disabled.
```

?

slax57 (commented):
Other than that, LGTM! :slightly_smiling_face:

docs/AccordionForm.md
fzaninotto: you're right, added.

djhi (approved):

slax57 (approved):

PR #9724, opened 2024-03-15
Allow overriding the error props in Input components

klassm:
The basic idea is to be able to override when a helper text becomes an error. In our case we want to override the behaviour of react-admin to only consider a field as errored when the field is touched or submitted. We would _always_ like to show the error message. Unfortunately, react-hook-form does not provide a way to set the internal touched state. However, when being able to override the error prop we can customize the behaviour ourselves.

The CheckboxGroupInput.tsx is also changed here, as the initial omitting of the error prop was introduced to fix a bug in this component. That's also why I moved the omitting of the prop to the local omit within the component.

### DIFF
--- a/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
+++ b/packages/ra-ui-materialui/src/input/CheckboxGroupInput.tsx
@@ -269,6 +269,7 @@ const sanitizeRestProps = ({
     setSort,
     loaded,
     touched,
+    error,
     ...rest
 }: any) => sanitizeInputRestProps(rest);
 
--- a/packages/ra-ui-materialui/src/input/NumberInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/NumberInput.stories.tsx
@@ -62,13 +62,18 @@ export const DefaultValue = () => (
 
 export const HelperText = () => (
     <Wrapper>
-        <NumberInput source="views" />
-        <NumberInput source="views" helperText={false} />
-        <NumberInput
-            source="views"
-            helperText="Number of times the post was read"
-        />
-    </Wrapper>
+                <NumberInput source="views" />
+                <NumberInput source="views" helperText={false} />
+                <NumberInput
+                    source="views"
+                    helperText="Number of times the post was read"
+                />
+                <NumberInput
+                    source="views"
+                    helperText="Number of times the post was read"
+                    error={true}
+                />
+            </Wrapper>
 );
 
 export const Label = () => (
--- a/packages/ra-ui-materialui/src/input/TextInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/TextInput.stories.tsx
@@ -50,13 +50,18 @@ export const DefaultValue = () => (
 
 export const HelperText = () => (
     <Wrapper>
-        <TextInput source="title" />
-        <TextInput source="title" helperText={false} />
-        <TextInput
-            source="title"
-            helperText="Number of times the post was read"
-        />
-    </Wrapper>
+                <TextInput source="title" />
+                <TextInput source="title" helperText={false} />
+                <TextInput
+                    source="title"
+                    helperText="Number of times the post was read"
+                />
+                <TextInput
+                    source="title"
+                    helperText="Number of times the post was read"
+                    error={true}
+                />
+            </Wrapper>
 );
 
 export const Label = () => (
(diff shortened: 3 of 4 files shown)


vlzuiev:
This would be so helpful.

vlzuiev:
This PR is ready for review 
So, if you have time please review that :)

klassm:
@fzaninotto I seem to not be able to add a #rfr label - happy to add any more changes you see here :-)

klassm:
Any chance to get this merged @fzaninotto ?

fzaninotto:
Thanks for your contribution. 

We'll only merge PRs that cover a use case we consider common enough to support. I don't understand the use case for this one, and the story you have included doesn't illustrate it, either. Can you please explain / document it further?

Also, as `master` is feature-freeze, you'll have to PR against `next` to have a chance to get this merged.

klassm:
Yes sure. In the end this is all about flexibility. By default, react-hook-form will only allow to set as errored in very very specific conditions. Unfortunately, those conditions do not always apply - or at least not for every form that can be built, e.g. via react-admin. The idea is to give this flexibility back to the developer, so they can decide themselves on whether a field should be shown as errored or not, not relying purely on the react-hook-form functionality.
React-admin completely abstracts away react-hook-forms, so even modifying react-hook-forms would be of no help here, as then the behaviour would also change for everyone. 
Why was this sanitization and limitation introduced in the very first place? Is there a reason to limit the developer ability to override this behaviour? The default would not change in this case, it would just provide more flexibility.

fzaninotto:
React-admin Input components were designed as an integration between react-hook-form and material UI's Fields. If you don't need the integration with react-hook-form, I suggest you don't use the Input components but the MUI Field components directly. This gives you the ability to bind with other form frameworks, or to bind with react-hook-form in another way.

As for integrating your changes to react-admin's Input components, I'm -1. For such advanced cases as yours, the correct solution is to use your own Input components. 

I'll let other contributors voice their opinion on this one.

slax57:
My 2c: I have never come across a case where I need to mark a field as erroneous without doing it via RHF.

Also, I'm worried this could lead to developer confusion, as passing `error={true}` to a field would make it visually render as erroneous, but RHF would have no idea it's in error so it would not prevent form submission for instance.

Still, this explains why I would not approve documenting this as a feature. But it seems it can still be supported in the code with no effort, so in the end, why not... :shrug:

klassm:
If it were possible to have it, it would make our life so much easier. Otherwise we now would need to basically duplicate every form component from react-admin, which is quite some effort for both maintenance and implementation. Actually we try to stick to react-admin as much as possible, that's the only place where we currently need to deviate 😕

quentin-decre:
I made a quick workaround to show errors on mount. That's far from ideal, but better for us than overriding all react-admin form components.

This component, when added to your form, will try a submit which will show errors if any. You may also remove the redirect on your form to stay on the form even if no errors. As I use the react-admin Auto-Save feature, this has not been an issue for us.

```typescript
export default function SaveOnMount() {
  const form = useFormContext()
  const saveContext = useSaveContext()

  const handleSubmit = useCallback(
    async (values) => {
      let errors
      if (saveContext?.save) {
        errors = await saveContext.save(values)
      }
      if (errors != null) {
        setSubmissionErrors(errors, form.setError)
      }
    },
    [form.setError, saveContext],
  )
  // validate the form at the beginning
  useEffect(() => {
    // validate after 500ms and 2000ms
    const timeout500 = setTimeout(() => {
      console.debug('ValidateOnMount re-validating 500')
      form.handleSubmit(handleSubmit)()
    }, 500)
    return () => {
      clearTimeout(timeout500)
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])

  return null
}
```

klassm:
Any chance to still get this merged @fzaninotto @slax57 ? 😅

slax57:
Hi there!

We've discussed this again with the team, and have come to the conclusion that we are not willing to support this way of overriding the `error` prop. Indeed this may lead to inconsistencies between the displayed errors and the form state (since it allows to bypass `react-hook-form` to display validation errors), which would probably hurt the user experience.

Note that, however, we have just opened https://github.com/marmelab/react-admin/pull/9781 , which should help with manually managing validation errors, as it removes the restriction that required inputs to be touched before displaying error messages.
This should cover most needs, even advanced ones, while still maintaining consistency with the `react-hook-form` form state.

If this still doesn't suit your needs, then as @fzaninotto suggested you will have to provide your own input components.

I'll close this PR for the reasons explained above. But we are still grateful for your contribution and your feedback!

PR #9728, opened 2024-03-16
Update authProvider.ts

hbendev:
Ignore @typescript-eslint scoped rule for local auth provider template file.


hbendev:
Hi! Thanks for the framework! When using the `create-react-admin` CLI, and selecting `Hard coded local username/password` option this generated line is marked as an error by ESLint if the configuration extends `"plugin:@typescript-eslint/recommended"`. 

If you see any value in these, I'm happy to
1. Address this extra additional disabled rule in other template files.
2. Change the default generated ESLint configuration to extend upon the `@typescript-eslint/recommended` values.

djhi:
Thanks!

PR #9730, opened 2024-03-18
Fixed onBlur callback on NumberInput and AutocompleteInput

### LINKED ISSUE #9726: NumberInput's `onBlur` prop does not provide an event object
**What you were expecting:**
I expected that the `onBlur` event handler on NumberInput component would receive an event as an argument.

**What happened instead:**
The function provided to the `onBlur` prop receives `undefined` instead of the `SyntheticBaseEvent`.

**Steps to reproduce:**
Simply add the following prop to the NumberInput component.
```
onBlur={event => console.log("event:", event)} //  "event: undefined"
```

**Related code:**
Here is the implementation that causes such behavior 
https://github.com/marmelab/react-admin/blob/10cc3218319e8642ad3cd1990c5cc52065e7555c/packages/ra-ui-materialui/src/input/NumberInput.tsx#L116-L119
Where I would expect it to be just like the `handleFocus` function.
https://github.com/marmelab/react-admin/blob/10cc3218319e8642ad3cd1990c5cc52065e7555c/packages/ra-ui-materialui/src/input/NumberInput.tsx#L109-L112
However, it's not straightforward to fix. Simple solution as below **works** but raises typescript error `TS2554: Expected 0 arguments, but got 1`
```
const handleBlur = (event) => {
    if (onBlurFromField) {
        onBlurFromField(event);
    }
}
```
That's because the type of the `field`'s `onBlur` function is `() => void`. This is reasonable because in `react-hook-form` that function is used to programmatically blur an input, not to hook into an event. However below you overwrite `react-hook-form` `useController`'s `onBlur` function to support passing events:
https://github.com/marmelab/react-admin/blob/54defedf6aad0ba95c562c01f887858c2f045eea/packages/ra-core/src/form/useInput.ts#L97-L102
Then you pass it to the return object of `useInput` hook
https://github.com/marmelab/react-admin/blob/54defedf6aad0ba95c562c01f887858c2f045eea/packages/ra-core/src/form/useInput.ts#L115-L128
The problem is that return type of `field` property is `ControllerRenderProps` whose `onBlur` function is `() => void` but should be something like `(...event: any[]) => void`

The issue lies in the return type of the `field` property, which is `ControllerRenderProps`. The `onBlur` function within it is specified as `() => void`, yet it ought to be defined as `(...event: any[]) => void`.
https://github.com/marmelab/react-admin/blob/54defedf6aad0ba95c562c01f887858c2f045eea/packages/ra-core/src/form/useInput.ts#L155-L161

**Other information:**
That behavior was also observed in the AutocompleteInput component.

**Environment**

* React-admin version: 4.16.13

yanchesky:
Fixes #9726

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -168,7 +168,9 @@ export type InputProps<ValueType = any> = Omit<
 export type UseInputValue = {
     id: string;
     isRequired: boolean;
-    field: ControllerRenderProps;
+    field: Omit<ControllerRenderProps, 'onBlur'> & {
+        onBlur: (...event: any[]) => void;
+    };
     formState: UseFormStateReturn<Record<string, string>>;
     fieldState: ControllerFieldState;
 };
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -73,7 +73,11 @@ export const useInput = <ValueType = any>(
     // This ensures dynamically added inputs have their value set correctly (ArrayInput for example).
     // We don't do this for the form level defaultValues so that it works as it should in react-hook-form
     // (i.e. field level defaultValue override form level defaultValues for this field).
-    const { field: controllerField, fieldState, formState } = useController({
+    const {
+        field: controllerField,
+        fieldState,
+        formState,
+    } = useController({
         name: finalName,
         defaultValue: get(record, finalSource, defaultValue),
         rules: {
@@ -111,10 +115,11 @@ export const useInput = <ValueType = any>(
     });
 
     const onChange = useEvent((...event: any[]) => {
-        const eventOrValue = (props.type === 'checkbox' &&
-        event[0]?.target?.value === 'on'
-            ? event[0].target.checked
-            : event[0]?.target?.value ?? event[0]) as any;
+        const eventOrValue = (
+            props.type === 'checkbox' && event[0]?.target?.value === 'on'
+                ? event[0].target.checked
+                : event[0]?.target?.value ?? event[0]
+        ) as any;
         controllerField.onChange(parse ? parse(eventOrValue) : eventOrValue);
         if (initialOnChange) {
             initialOnChange(...event);
@@ -167,7 +172,9 @@ export type UseInputValue = {
     field: Omit<ControllerRenderProps, 'onBlur'> & {
         onBlur: (
             event?:
-                | FocusEvent<HTMLInputElement | HTMLTextAreaElement>
+                | FocusEvent<
+                      HTMLInputElement | HTMLTextAreaElement | HTMLDivElement
+                  >
                 | EditorEvents['blur']
         ) => void;
     };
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/form/useInput.ts
```diff
@@ -153,7 +153,9 @@ export type InputProps<ValueType = any> = Omit<
 export type UseInputValue = {
     id: string;
     isRequired: boolean;
-    field: ControllerRenderProps;
+    field: Omit<ControllerRenderProps, 'onBlur'> & {
+        onBlur: (...event: any[]) => void;
```
fzaninotto, line 157: Why don't you use React's built-in FocusEventHandler?

fzaninotto (changes requested):
I don't know why the mobile e2e tests are failing, you'll have to investigate.

packages/ra-core/src/form/useInput.ts
yanchesky: I relied on implementation of onBlur function in `useInput`
https://github.com/marmelab/react-admin/blob/e87f8f491b6e340084585fe0d6585aa8f710bc70/packages/ra-core/src/form/useInput.ts#L97-L102
and onBlur of InputProps where type is written in such way.

There are also 2 cases where the FocusEventHandler will fail the type contract.

1. A component needs to blur the input (BooleanInput, RichTextInput, FileInput, CheckboxGroupInput) because FocusEventHandler requires event as an argument.
```javascript
        event => {
            field.onChange(event);
            // Ensure field is considered as touched
            field.onBlur();
        },
        [field]
    );
```
2. RichTextInput provides EditorEvents['blur'] instead, which doesn't match with the FocusEvent.

Here is the solution I came up that matches type contracts. If you like it i'll commit it
```javascript
 field: Omit<ControllerRenderProps, 'onBlur'> & {
        onBlur: (
            event?:
                | FocusEvent<HTMLInputElement | HTMLTextAreaElement>
                | EditorEvents['blur']
        ) => void;
    };
```

yanchesky:
> I don't know why the mobile e2e tests are failing, you'll have to investigate.

https://github.com/yanchesky/react-admin/pull/1
On my fork, all tests pass, leading me to suspect that this failure is not related to committed changes.

packages/ra-core/src/form/useInput.ts
djhi: Sorry it took so long to answer. Yes, this new type looks fine

djhi:
> > I don't know why the mobile e2e tests are failing, you'll have to investigate.
> 
> [yanchesky#1](https://github.com/yanchesky/react-admin/pull/1) On my fork, all tests pass, leading me to suspect that this failure is not related to committed changes.

All e2e tests pass locally. Probably a random CI failure

yanchesky:
Are there any other steps I should take in order to finalise this PR?

djhi:
> Are there any other steps I should take in order to finalise this PR?

Can you implement the `onBlur` type as you suggested in https://github.com/marmelab/react-admin/pull/9730#discussion_r1530142688? Hopefully that should trigger the CI again. For some Github reason, I can't restart the CI job. Probably because we took to long to react

yanchesky:
Instead of adding a new `HTMLSpanElement` type, I changed the `FocusEvent` to accept a generic `HTMLElement` type.

slax57 (approved):
Thanks!

PR #9731, opened 2024-03-19
[Feat] Expose RichTextIinput editor with a ref

erwanMarmelab:
## Problem
Except in the <RichTextInput> toolbar, there is no way to access the TipTap editor. 

## Solution
Expose the TipTap editor with a ref

## Todo

- [x] Add the "ref" prop
- [x] Add a story
- [x] Test it
- [x] Document it

### DIFF
--- a/docs/RichTextInput.md
+++ b/docs/RichTextInput.md
@@ -175,14 +175,16 @@ const MyRichTextInput = ({ size, ...props }) => (
 
 You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can assign a `ref` in the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component which returns the editor:
 
+{% raw %}
 ```tsx
+import React from 'react';
 import { Edit, SaveButton, SimpleForm, TextInput, Toolbar } from 'react-admin';
 import { DefaultEditorOptions, RichTextInput } from 'ra-input-rich-text';
-import { Buuton } from 'ra-ui-materialui';
+import { Button } from 'ra-ui-materialui';
 import { Editor } from '@tiptap/react';
 
 export const PostEdit = () => {
-    const editorRef = React.useRef<Editor>(null);
+    const editorRef = React.useRef<Editor | null>(null);
 
     return (
         <Edit>
@@ -191,16 +193,22 @@ export const PostEdit = () => {
                     <Toolbar>
                         <SaveButton />
                         <Button
-                            onClick={() => editorRef.current.commands.setContent('<h3>Here is my template</h3>')}
+                            onClick={() =>
+                                editorRef.current
+                                    ? editorRef.current.commands.setContent(
+                                          '<h3>Here is my template</h3>'
+                                      )
+                                    : null
+                            }
                         >
-                            Use template
+                            <>Use template</>
                         </Button>
                     </Toolbar>
                 }
             >
                 <TextInput source="title" />
-                <RichTextInput 
-                    source="body" 
+                <RichTextInput
+                    source="body"
                     editorOptions={{
                         ...DefaultEditorOptions,
                         onCreate: ({ editor }: { editor: Editor }) => {
@@ -210,9 +218,10 @@ export const PostEdit = () => {
                 />
             </SimpleForm>
         </Edit>
-    )
+    );
 };
 ```
+{% endraw %}  
 
 ## AI Writing Assistant
 
--- a/docs/RichTextInput.md
+++ b/docs/RichTextInput.md
@@ -173,11 +173,11 @@ const MyRichTextInput = ({ size, ...props }) => (
 
 ## Reference the editor
 
-You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can pass a `ref` to the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component:
+You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can assign a `ref` in the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component which returns the editor:
 
 ```tsx
 import { Edit, SaveButton, SimpleForm, TextInput, Toolbar } from 'react-admin';
-import { RichTextInput } from 'ra-input-rich-text';
+import { DefaultEditorOptions, RichTextInput } from 'ra-input-rich-text';
 import { Buuton } from 'ra-ui-materialui';
 import { Editor } from '@tiptap/react';
 
--- a/docs/RichTextInput.md
+++ b/docs/RichTextInput.md
@@ -44,7 +44,8 @@ export const PostEdit = () => (
 | Prop   | Required | Type     | Default | Description |
 | ------ | -------- | -------- | ------- | ----------- |
 | `editorOptions` | Optional | `Object` | - | Options object to pass to the underlying TipTap editor. |
-| `toolbar` | Optional| ReactNode | - | The toolbar to use. If not set, the default toolbar is used. |
+| `toolbar` | Optional| `ReactNode` | - | The toolbar to use. If not set, the default toolbar is used. |
+| `ref` | Optional| `MutableRefObject<Editor>` | - | To reference the editor object  |
 
 `<RichTextInput>` also accepts the [common input props](./Inputs.md#common-input-props).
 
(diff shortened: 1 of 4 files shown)


djhi (changes requested):
I think you should pass the default `ref` to the root HTML element and introduce a `tiptapRef` for the tiptap API. Indeed, some users might need a ref for other purposes.

Simply add it to the props interface like any other prop

docs/RichTextInput.md
```diff
@@ -171,6 +172,41 @@ const MyRichTextInput = ({ size, ...props }) => (
...
+You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can pass a `ref` to your `<RichTextInput>` component and use it:
+
+```tsx
+import { Edit, SaveButton, SimpleForm, TextInput, Toolbar } from 'react-admin';
+import { RichTextInput } from 'ra-input-rich-text';
+import { Buuton } from 'ra-ui-materialui';
+import { Editor } from '@tiptap/react';
+
+export const PostEdit = () => {
+    const myRef = React.useRef<Editor>(null);
```
fzaninotto, line 186:
```suggestion
    const editorRef = React.useRef<Editor>(null);
```

packages/ra-input-rich-text/src/RichTextInput.spec.tsx
```diff
@@ -23,4 +23,22 @@ describe('<RichTextInput />', () => {
             );
         });
     });
+    it("should update its content with the ref editor's commands", async () => {
```
fzaninotto, line 26: Wrap a `describe('ref')` around the tests for that prop.

packages/ra-input-rich-text/src/RichTextInput.stories.tsx
```diff
@@ -187,6 +191,61 @@ export const Validation = (props: Partial<SimpleFormProps>) => (
...
+            onSubmit={() => {}}
+            {...props}
+        >
+            <RichTextInput source="body" toolbar={<MyRichTextInputToolbar />} />
+            <FormInspector />
+        </SimpleForm>
+    </AdminContext>
+);
+
+export const Reference = (props: Partial<SimpleFormProps>) => {
```
fzaninotto, line 215: Please rename that story to the name of the prop you're testing (in this case, `Ref`)

packages/ra-input-rich-text/src/RichTextInput.stories.tsx
```diff
@@ -187,6 +191,61 @@ export const Validation = (props: Partial<SimpleFormProps>) => (
...
 
+const MyRichTextInputToolbar = ({ ...props }) => {
+    return (
+        <RichTextInputToolbar {...props}>
+            <FormatButtons />
+        </RichTextInputToolbar>
+    );
+};
+
+export const EditedToolbar = (props: Partial<SimpleFormProps>) => (
```
fzaninotto, line 202: Please rename that story to the name of the prop you're testing (in this case, `Toolbar`)

docs/RichTextInput.md
```diff
@@ -171,6 +172,41 @@ const MyRichTextInput = ({ size, ...props }) => (
...
+                        <Button
+                            onClick={() => myRef.current.commands.setContent('<h3>Here is my template</h3>')}
+                        >
+                            Use template
+                        </Button>
+                    </Toolbar>
+                }
+            >
+                <TextInput source="title" />
+                <RichTextInput source="body" ref={myRef} />
```
fzaninotto, line 203:
```suggestion
                <RichTextInput source="body" ref={editorRef} />
```

packages/ra-input-rich-text/src/RichTextInput.tsx
```diff
@@ -73,117 +73,123 @@ import { TiptapEditorProvider } from './TiptapEditorProvider';
...
-    );
+            [disabled, editorOptions, readOnly, id]
+        );
 
-    const { error, invalid, isTouched } = fieldState;
+        const { error, invalid, isTouched } = fieldState;
 
-    useEffect(() => {
-        if (!editor) return;
+        if (ref) {
```
fzaninotto, line 119: You should not set ref during rendering, see https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs

fzaninotto (changes requested):
I wonder if this is necessary, as it's possible to get the editor instance using the `onCreate` editor option. Something like this should work:

```js
const editorRef = useRef();

return <RichTextInput source="foo" editorOptions={{ onCreate: ({ editor }) => { editorRef.current = editor; } }} />;
```

If it does work, the please revert your changes and just document the above trick.

erwanMarmelab:
> ```js
> ```js
> return <RichTextInput source="foo" editorOptions={{ onCreate: ({ editor }) => { editorRef.current = editor; } }} />;
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

Unfortunately, it doesn't work

image

image

packages/ra-input-rich-text/src/RichTextInput.tsx
erwanMarmelab: done

packages/ra-input-rich-text/src/RichTextInput.stories.tsx
erwanMarmelab: done

packages/ra-input-rich-text/src/RichTextInput.spec.tsx
erwanMarmelab: done

packages/ra-input-rich-text/src/RichTextInput.stories.tsx
erwanMarmelab: done

docs/RichTextInput.md
```diff
@@ -172,9 +171,9 @@ const MyRichTextInput = ({ size, ...props }) => (
 );
 ```
 
-## `ref`
+## Reference the editor
 
-You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can pass a `ref` to your `<RichTextInput>` component and use it:
+You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can pass a `ref` to the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component:
```
djhi, line 176: This doesn't look right. You don't pass a ref to the function. 
```suggestion
You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, use the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component which returns the editor:
```

djhi (changes requested):

docs/RichTextInput.md
fzaninotto, line 176:
```suggestion
You might need access to the editor to use [the dedicated API](https://tiptap.dev/docs/editor/api/editor). To do it, you can assign a `ref` in the `onCreate` function in the `editorOptions` prop of your `<RichTextInput>` component:
```

docs/RichTextInput.md
```diff
@@ -200,7 +199,15 @@ export const PostEdit = () => {
                 }
             >
                 <TextInput source="title" />
-                <RichTextInput source="body" ref={editorRef} />
+                <RichTextInput 
+                    source="body" 
+                    editorOptions={{
+                        ...DefaultEditorOptions,
```
fzaninotto, line 205: Missing import

docs/RichTextInput.md
erwanMarmelab: I didn't change `you can pass a 'href'` to `use` because there was a conflit with [this comment](#discussion_r1530631189)

erwanMarmelab: But `which returns the editor` added

erwanMarmelab: done

docs/RichTextInput.md
erwanMarmelab: done

docs/RichTextInput.md
```diff
@@ -171,6 +171,49 @@ const MyRichTextInput = ({ size, ...props }) => (
...
+                        </Button>
+                    </Toolbar>
+                }
+            >
+                <TextInput source="title" />
+                <RichTextInput 
+                    source="body" 
+                    editorOptions={{
+                        ...DefaultEditorOptions,
+                        onCreate: ({ editor }: { editor: Editor }) => {
```
adguernier, line 206: **polish:** ‏Typescript complaints here that I haven't been able to resolve

adguernier (changes requested):

adguernier (approved):

fzaninotto (approved):

PR #9733, opened 2024-03-19
[Fix] Fix CRM demo error

erwanMarmelab:
## Problem
In the CRM demo, there is an error on viewing a newly created deal.:

When you create a new deal. You didn't set se `start_at` value.
On the show page, we try to format this date.
When there is any value in the `start_at` column, we have an error.

## Solution
Format this value conditionally

### DIFF
--- a/examples/crm/src/deals/DealShow.tsx
+++ b/examples/crm/src/deals/DealShow.tsx
@@ -87,7 +87,8 @@ const DealShowContent = () => {
                                 Start
                             </Typography>
                             <Typography variant="subtitle1">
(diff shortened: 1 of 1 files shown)


fzaninotto:
This is not the right fix. A new deal should have a date, so you'd better add a `start_date` input in the `DealCreate` form, under the `company_id` input.

```
                    <DateInput
                        source="start_at"
                        defaultValue={new Date()}
                        fullWidth
                    />
```

fzaninotto (approved):

Issue #9740, opened 2024-03-22
storeKey prop is not passed to useRecordSelection hook inside useListController


nbalaguer:
English is not my native language, my apologies in advance for any weird grammar or implicit tone of the writing. I'll try my best 😃

**What you were expecting:**

I would expect the storeKey prop to work on both useListParams and useRecordSelction hooks.

**What happened instead:**

Right now the storeKey prop only gets passed to the useListParams hook.

**Steps to reproduce:**

-

**Related code:**

Prop is received
https://github.com/marmelab/react-admin/blob/ec44ac5d8b86726036ec984991be3aa46edd410b/packages/ra-core/src/controller/list/useListController.ts#L35-L46

Prop is forwared to useListParams
https://github.com/marmelab/react-admin/blob/ec44ac5d8b86726036ec984991be3aa46edd410b/packages/ra-core/src/controller/list/useListController.ts#L65-L73

Prop is NOT forwarded to useRecordSelection
https://github.com/marmelab/react-admin/blob/ec44ac5d8b86726036ec984991be3aa46edd410b/packages/ra-core/src/controller/list/useListController.ts#L75

storeKey is hardcoded based on the resource inside useRecordSelection
https://github.com/marmelab/react-admin/blob/ec44ac5d8b86726036ec984991be3aa46edd410b/packages/ra-core/src/controller/list/useRecordSelection.ts#L24

useListParams uses state to keep localParams when storeKey is false and the custom storeKey if provided, defaulting to a storeKey derived from the resource. useRecordSelection should do something similar.

**Other information:**

In this example, using 2 lists to display the same resource, one with storeKey set to false and the other with storeKey set to a custom string, you can see that when selecting a row in one list, the same row gets selected in the other list.

https://stackblitz.com/edit/github-asfjgd?file=src%2Fposts%2FPostList.tsx

**Environment**

* React-admin version: 4.16.12
* Last version that did not exhibit the issue (if applicable): -
* React version: 18.2.0
* Browser:
* Stack trace (in case of a JS error):

djhi:
Good catch! Would you mind opening a PR to fix it?

djhi:
Actually, we already had a few similar issues about this and the answer is not so simple: see [#8536](https://github.com/marmelab/react-admin/issues/8536#issuecomment-1369911990) and [#7401](https://github.com/marmelab/react-admin/issues/7401#issuecomment-1073922817)

nbalaguer:
@djhi 

[...]

PR #9742, opened 2024-03-22
Disable all store interactions for a list with storeKey set to false (sort, pagination, filters and now also selection state).

nbalaguer:
Proposed update in #9740

### DIFF
--- a/packages/ra-core/src/controller/list/useRecordSelection.ts
+++ b/packages/ra-core/src/controller/list/useRecordSelection.ts
@@ -1,17 +1,30 @@
-import { useMemo } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import { useStore, useRemoveFromStore } from '../../store';
 import { RaRecord } from '../../types';
 
+type UseRecordSelectionWithResourceArgs = {
+    resource: string;
+    disableSyncWithStore?: false;
+};
+type UseRecordSelectionWithNoStoreArgs = {
+    resource?: string;
+    disableSyncWithStore: true;
+};
+type UseRecordSelectionArgs =
+    | UseRecordSelectionWithResourceArgs
+    | UseRecordSelectionWithNoStoreArgs;
+
 /**
  * Get the list of selected items for a resource, and callbacks to change the selection
  *
- * @param resource The resource name, e.g. 'posts'
+ * @param args.resource The resource name, e.g. 'posts'
+ * @param args.disableSyncWithStore Controls the selection syncronization with the store
  *
  * @returns {Object} Destructure as [selectedIds, { select, toggle, clearSelection }].
  */
 export const useRecordSelection = <RecordType extends RaRecord = any>(
-    resource: string
+    args: UseRecordSelectionArgs
 ): [
     RecordType['id'][],
     {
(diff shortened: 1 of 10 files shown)


fzaninotto:
Nice! Would you mind adding a unit tests for the local version?

packages/ra-core/src/controller/list/useRecordSelection.ts
```diff
@@ -7,11 +7,13 @@ import { RaRecord } from '../../types';
  * Get the list of selected items for a resource, and callbacks to change the selection
  *
  * @param resource The resource name, e.g. 'posts'
+ * @param disableSyncWithStore Disable the synchronization with the store
  *
  * @returns {Object} Destructure as [selectedIds, { select, toggle, clearSelection }].
  */
 export const useRecordSelection = <RecordType extends RaRecord = any>(
-    resource: string
+    resource: string,
```
fzaninotto, line 15: if disableSyncWithStore is true, `resource` isn't required.

fzaninotto:
Also, this must be PRed against `next` as the `master` branch is feature-freeze.

packages/ra-core/src/controller/list/useRecordSelection.ts
nbalaguer: I've added the unit tests based on the already implemented ones, let me know if anything needs to change!

Regarding the typing of the hook I have some doubts:

Would it be enough to just set resource to optional? I'm not really sure since it may not error when calling this hook without any arguments.

After searching around to find information about making function parameters optional based on other parameters and messing around with the typescript playground, I've arrived to [this type definition](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDGB7ATgEwMoQDYOAS0QDsBBZAcwGcoBeKAbQuGT0LIBooAzAQywogD8AXSgAfeo2asxUQgFcsWDkzkQhAKFCRYCFBmy4ChOBQXAaUAG6I8qKAHp7sxObwBbMCmDdC5ziihgAAs8KggAD24PHHV1JEJGKDl+OCQ0TBx4fCIALigACgA6Yu5yCjzUvQzDIlJKAEoaAD4dNP1M7OMIUyxzWnyAazzJFjIZeUUOACMBPKnERBwfRuoWgG91KC2oZAhgOWRCdQBfWOSISvSDLKN8+vVzy-aawnyAIm4KVE43+8fdK4dW4fL4-Dg8PgQP4pAHPG5Ed6fb5vZTIVTQi6w6rw14TLAYp7Yzr5PHg3j8AlY67E0mBNFQ9RAA). I didn't want to change the hook from an arrow function to a normal function and this is the best I've been able to come up with (after a fair bit of googling) to do some kind of signature overloading.

In the react-admin project, since strictNullChecks is not enabled, lines 12 and 13 of the playground code do not error.

What would be the best way to approach this?

fzaninotto: I'd change the argument to an object:

```
type UseRecordSelectionWithResourceArgs = {
 resource: string;
}
type UseRecordSelectionWithNoStoreArgs = {
  disableSyncWithStore: false;
}
type UseRecordSelectionArgs = UseRecordSelectionWithResourceArgs | UseRecordSelectionWithNoStoreArgs;
export const useRecordSelection = (args: UseRecordSelectionArgs) => {
  // ...
```

nbalaguer: And then change all instances of calling the hook to use an object right? If that's ok I'll get to it today :D

fzaninotto: Yes! But again this must be done in the `next` branch.

nbalaguer: Hey! I hope this is not too much of a bother but to avoid any issues when changing the destination branch to `next` (there were a couple of commits that weren't mine) I updated my branch by starting directly from the latest state of `next` and applying my changes in a single commit.

The finishing changes are:

### Changing `useRecordSelection`'s signature

Following your suggestion, I changed from `(resource, disableSyncWithStore)` to `(args)`. I had a couple of issues with only having

```typescript
type UseRecordSelectionWithResourceArgs = {
 resource: string;
}
type UseRecordSelectionWithNoStoreArgs = {
  disableSyncWithStore: true;
}
```

as the type. When using the hook like

```typescript
const [selectedIds, selectionModifiers] = useRecordSelection({
    resource,
    disableSyncWithStore: storeKey === false,
});
```

typescript was complaining that `boolean` is not assignable to `true`, so I modified it a bit and arrived to

```typescript
type UseRecordSelectionWithResourceArgs = {
    resource: string;
    disableSyncWithStore?: false;
};
type UseRecordSelectionWithNoStoreArgs = {
    resource?: string;
    disableSyncWithStore: true;
};
```

which seems to work well. The error disappears and it still complains if you don't specify a `resource` and don't set `disableSyncWithStore` to `true`.

### Updating the documentation

I also tweaked a bit the documentation when talking about `storeKey` stuff. I hope it reflects the changes made to the use of storeKey apropriately. If any improvements need to be made plese feel free to tell me.

nbalaguer: @fzaninotto I realized my email address was wrong inside my linux subsystem and was authoring commits from a different account to this one. I updated it that's why the commit is so recent. Sorry for any inconvenience :(

fzaninotto (approved):

Issue #9622, opened 2024-01-29
TypeScript types are generated with unsafe types because strictNullChecks is false


gauthierm:
**What you were expecting:**

With this code the type of `record` should be `MyRecord | undefined`:

```ts
const record = useRecordContext<MyRecord>();
```

**What happened instead:**

The type of `record` is `MyRecord`.

**Steps to reproduce:**

1. Install react-admin in a TypeScript project
2. Write some code that uses useRecordContext
3. Use the record without checking for undefined because the return type says it cannot be undefined
4. Get a runtime error.

**Other information:**

This issue is caused by the `tsconfig.json` in the base react-admin project not using strict mode, specifically `strictNullChecks`. This means TypeScript will coalesce a type like `RaRecord | undefined` into `RaRecord`. See https://www.typescriptlang.org/tsconfig#strictNullChecks

A previous bug was recorded https://github.com/marmelab/react-admin/issues/7588 and fixed in PR https://github.com/marmelab/react-admin/pull/7498. The fix is correct if strictNullChecks is used, but because TypeScript is not configured to consider undefined, the fix doesn't actually do anything.

To confirm this, you can check the build output in either the esm or cjs folders. The `| undefined` is omitted from the generated types.

**Environment**

* React-admin version: 4.16.7
* Last version that did not exhibit the issue (if applicable): n/a
* React version: 18.x
* Browser: n/a
* Stack trace (in case of a JS error): n/a

fzaninotto:
Hi, and thanks for your report. 

We're well aware of the limitations caused by `structNullChecks=false` in the quality of types. We have it on our radar, but the benefit / cost ratio of this task is low, so it's not a priority for us. Fixing it is non-trivial, as we have to make `ra-core` compilable with `strict?NullChecks=true`. Previous attempts (#8141) lead to a dead end.

You're welcome to contribute small PRs fixing types of individual hooks & components for strict null checks. 

I'm closing this issue as it's a long-term goal.

gauthierm:
Thanks for letting us know it's a long term goal. It's not possible to fix the issue with the hooks until strictNullChecks is enabled. It's currently not possible to use the generated types safely in projects.

gauthierm:
@fzaninotto What sort of approach would be required to get strictNullChecks enabled? It looks like someone tried to help you with it in #8141, but because the scope of the change was too big the PR was closed.

I'm not sure how you could partially add strict null checks to make for a smaller PR that would be accepted.

[...]

PR #9743, opened 2024-03-25
[TypeScript] Make types more strict in ra-core, part II

fzaninotto:
With strictNullChecks turned on, this PR reduces TS compilation errors on ra-core from 180 to 60 (-67%)

Refs #9622

## Page Contexts Are Now Types Instead of Interfaces

The return type of page controllers is now a type. If you were using an interface extending one of:

- `ListControllerResult`,
- `InfiniteListControllerResult`,
- `EditControllerResult`,
- `ShowControllerResult`, or
- `CreateControllerResult`,

you'll have to change it to a type:

```diff
import { ListControllerResult } from 'react-admin';

-interface MyListControllerResult extends ListControllerResult {
+type MyListControllerResult = ListControllerResult & {
    customProp: string;
};
```

## TypeScript: Stronger Types For Page Contexts

The return type of page context hooks is now smarter. This concerns the following hooks:

- `useListContext`,
- `useEditContext`,
- `useShowContext`, and
- `useCreateContext`

Depending on the fetch status of the data, the type of the `data`, `error`, and `isPending` properties will be more precise:

- Loading: `{ data: undefined, error: undefined, isPending: true }`
- Success: `{ data: <Data>, error: undefined, isPending: false }`
- Error: `{ data: undefined, error: <Error>, isPending: false }`
- Error After Refetch: `{ data: <Data>, error: <Error>, isPending: false }`

This means that TypeScript may complain if you use the `data` property without checking if it's defined first. You'll have to update your code to handle the different states:

```diff
const MyCustomList = () => {
    const { data, error, isPending } = useListContext();
    if (isPending) return <Loading />;
+   if (error) return <Error />;
    return (
        <ul>
            {data.map(record => (
                <li key={record.id}>{record.name}</li>
            ))}
        </ul>
    );
};
```

Besides, these hooks will now throw an error when called outside of a page context. This means that you can't use them in a custom component that is not a child of a `<List>`, `<ListBase>`, `<Edit>`, `<EditBase>`, `<Show>`, `<ShowBase>`, `<Create>`, or `<CreateBase>` component.

### DIFF
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -240,7 +240,6 @@ export const useEditController = <
         ]
     );
 
-    // @ts-ignore FIXME cannot find another way to fix this error: "Types of property 'isPending' are incompatible: Type 'boolean' is not assignable to type 'false'."
     return {
         defaultTitle,
         error,
@@ -256,7 +255,7 @@ export const useEditController = <
         save,
         saving,
         unregisterMutationMiddleware,
-    };
+    } as EditControllerResult<RecordType>;
 };
 
 const DefaultRedirect = 'list';
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -240,6 +240,7 @@ export const useEditController = <
         ]
     );
 
+    // @ts-ignore FIXME cannot find another way to fix this error: "Types of property 'isPending' are incompatible: Type 'boolean' is not assignable to type 'false'."
     return {
         defaultTitle,
         error,
@@ -258,6 +259,8 @@ export const useEditController = <
     };
 };
 
+const DefaultRedirect = 'list';
+
 export interface EditControllerProps<
     RecordType extends RaRecord = any,
     ErrorType = Error
@@ -273,20 +276,47 @@ export interface EditControllerProps<
     [key: string]: any;
 }
 
-export interface EditControllerResult<RecordType extends RaRecord = any>
+export interface EditControllerBaseResult<RecordType extends RaRecord = any>
     extends SaveContextValue<RecordType> {
-    // Necessary for actions (EditActions) which expect a data prop containing the record
-    // @deprecated - to be removed in 4.0d
-    data?: RecordType;
-    error?: any;
     defaultTitle?: string;
     isFetching: boolean;
     isLoading: boolean;
-    isPending: boolean;
-    record?: RecordType;
     refetch: UseGetOneHookValue<RecordType>['refetch'];
     redirect: RedirectionSideEffect;
     resource: string;
 }
 
-const DefaultRedirect = 'list';
+interface EditControllerLoadingResult<RecordType extends RaRecord = any>
+    extends EditControllerBaseResult<RecordType> {
+    record: undefined;
+    error: null;
+    isPending: true;
+}
+interface EditControllerLoadingErrorResult<
+    RecordType extends RaRecord = any,
+    TError = Error
+> extends EditControllerBaseResult<RecordType> {
+    record: undefined;
+    error: TError;
+    isPending: false;
+}
+interface EditControllerRefetchErrorResult<
+    RecordType extends RaRecord = any,
+    TError = Error
+> extends EditControllerBaseResult<RecordType> {
+    record: RecordType;
+    error: TError;
+    isPending: false;
+}
+interface EditControllerSuccessResult<RecordType extends RaRecord = any>
+    extends EditControllerBaseResult<RecordType> {
+    record: RecordType;
+    error: null;
+    isPending: false;
+}
+
+export type EditControllerResult<RecordType extends RaRecord = any> =
+    | EditControllerLoadingResult<RecordType>
+    | EditControllerLoadingErrorResult<RecordType>
+    | EditControllerRefetchErrorResult<RecordType>
+    | EditControllerSuccessResult<RecordType>;
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -285,13 +285,13 @@ export interface EditControllerBaseResult<RecordType extends RaRecord = any>
     resource: string;
 }
 
-interface EditControllerLoadingResult<RecordType extends RaRecord = any>
+export interface EditControllerLoadingResult<RecordType extends RaRecord = any>
     extends EditControllerBaseResult<RecordType> {
     record: undefined;
     error: null;
     isPending: true;
 }
-interface EditControllerLoadingErrorResult<
+export interface EditControllerLoadingErrorResult<
     RecordType extends RaRecord = any,
     TError = Error
 > extends EditControllerBaseResult<RecordType> {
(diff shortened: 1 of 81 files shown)


packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -259,20 +276,47 @@ export interface EditControllerProps<
     [key: string]: any;
 }
 
-export interface EditControllerResult<RecordType extends RaRecord = any>
+export interface EditControllerBaseResult<RecordType extends RaRecord = any>
     extends SaveContextValue<RecordType> {
-    // Necessary for actions (EditActions) which expect a data prop containing the record
-    // @deprecated - to be removed in 4.0d
-    data?: RecordType;
```
fzaninotto, line 266: breaking change

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -259,20 +276,47 @@ export interface EditControllerProps<
...
+    isPending: false;
+}
+interface EditControllerSuccessResult<RecordType extends RaRecord = any>
+    extends EditControllerBaseResult<RecordType> {
+    record: RecordType;
+    error: null;
+    isPending: false;
+}
+
+export type EditControllerResult<RecordType extends RaRecord = any> =
```
fzaninotto, line 318: breaking change - this is now a type and not an interface

packages/ra-core/src/controller/list/useInfiniteListController.ts
```diff
@@ -222,8 +223,9 @@ export interface InfiniteListControllerProps<
     storeKey?: string | false;
 }
 
-export interface InfiniteListControllerResult<RecordType extends RaRecord = any>
-    extends ListControllerResult<RecordType> {
+export type InfiniteListControllerResult<
```
fzaninotto, line 226: breaking change - this is now a type instead of an interface

packages/ra-core/src/controller/list/useListController.ts
```diff
@@ -478,3 +444,73 @@ export const sanitizeListRestProps = props =>
...
+}
+interface ListControllerSuccessResult<RecordType extends RaRecord = any>
+    extends ListControllerBaseResult<RecordType> {
+    data: RecordType[];
+    total: number;
+    error: null;
+    isPending: false;
+}
+
+export type ListControllerResult<RecordType extends RaRecord = any> =
```
fzaninotto, line 512: breaking change - this is now a type instead of an interface

packages/ra-core/src/controller/show/useShowController.ts
```diff
@@ -125,16 +136,46 @@ export interface ShowControllerProps<RecordType extends RaRecord = any> {
     resource?: string;
 }
 
-export interface ShowControllerResult<RecordType extends RaRecord = any> {
-    defaultTitle: string;
-    // Necessary for actions (EditActions) which expect a data prop containing the record
-    // @deprecated - to be removed in 4.0d
-    data?: RecordType;
```
fzaninotto, line 132: breaking change

packages/ra-core/src/controller/show/useShowController.ts
```diff
@@ -125,16 +136,46 @@ export interface ShowControllerProps<RecordType extends RaRecord = any> {
...
+    isPending: false;
+}
+interface ShowControllerSuccessResult<RecordType extends RaRecord = any>
+    extends ShowControllerBaseResult<RecordType> {
+    record: RecordType;
+    error: null;
+    isPending: false;
+}
+
+export type ShowControllerResult<RecordType extends RaRecord = any> =
```
fzaninotto, line 177: breaking change - this is now a type instead of an interface

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
```diff
@@ -156,14 +156,13 @@ export const ReferenceManyFieldView: FC<ReferenceManyFieldViewProps> = props =>
...
 
-export interface ReferenceManyFieldViewProps
-    extends Omit<
-            ReferenceManyFieldProps,
-            'resource' | 'page' | 'perPage' | 'sort'
-        >,
-        ListControllerResult {
-    children: ReactElement;
-}
+export type ReferenceManyFieldViewProps = Omit<
```
fzaninotto, line 159: breaking change - this is now a type instead of an interface

packages/ra-core/src/controller/list/ListContext.tsx
```diff
@@ -52,33 +52,6 @@ import { ListControllerResult } from './useListController';
...
-    resource: null,
-    selectedIds: undefined,
-    setFilters: null,
-    setPage: null,
-    setPerPage: null,
-    setSort: null,
-    showFilter: null,
-    total: null,
-});
+export const ListContext = createContext<ListControllerResult | null>(null);
```
fzaninotto, line 55: Having a default value would force choosing between the pending/error/success case, which doesn't make sense e.g. when using `<Datagrid>` outside of  a `ListContext` but with a `data` prop. 

On the other hand, using a `null` default value should mean that `useListContext` returns `ListControllerResult | null`, but for some reason the typeScript inference sees the return value as `ListControllerResult `, even without return type annotation. 

This is convenient, because the alternative is to turn the return type of `useListContext` as `Partial< ListControllerResult>`, which loses in the process the union between loading/error/success types. Or we could have two hooks:

- one that throws if there is no context, but always returns a `ListControllerResult`
- the other that doesn't throw but returns a `Partial< ListControllerResult>`, forcing a lot more runtime checks. 

Having two hooks seems too complex, so let's keep the code as is for now.

fzaninotto: There aren't many components that actually need the prop fallback: Datagrid, SingleFieldList, and SimpleList. I ended up making a second hook especially for these ones. This allows me to throw an error whenever a `useXXXController` hook is invoked outside of a `XXXContext`, and ensure that the types are correct when reading the context with these hooks.

packages/ra-ui-materialui/src/list/BulkActionsToolbar.tsx
```diff
@@ -24,7 +19,7 @@ export const BulkActionsToolbar = (props: BulkActionsToolbarProps) => {
         className,
         ...rest
     } = props;
-    const { selectedIds = [], onUnselectItems } = useListContext(props);
+    const { selectedIds = [], onUnselectItems } = useListContext();
```
fzaninotto, line 22: Note that a standalone `<Datagrid>` with `bulkActionToolbars` not set to `false` will now fail loudly, wit han exception from `useListContext`. But it's better than before, because `<Datagrid>` only used to pass the `selectedIds` to the `BulkActionToolbar`, which didn't let the user select or unselect rows. I modified the doc to make it obvious that if you want bulk actions with a standalone Datagrid, you'll have to provide a context using `useList`.

fzaninotto:
Switching to RFR

packages/ra-ui-materialui/src/button/BulkDeleteWithConfirmButton.tsx
```diff
@@ -37,7 +37,7 @@ export const BulkDeleteWithConfirmButton = (
         ...rest
     } = props;
     const { meta: mutationMeta, ...otherMutationOptions } = mutationOptions;
-    const { selectedIds, onUnselectItems } = useListContext(props);
+    const { selectedIds, onUnselectItems } = useListContext();
```
djhi, line 40: Shouldn't it call `useListContextWithProps`? It seems we loose the ability to override through props as before. Same for all the other UI components

fzaninotto: you're right, but I don't see any use case for overriding the selectedIds on this one.

djhi (approved):
:muscle:

PR #9748, opened 2024-03-28
Add queryOptions support for ReferenceManyField

### LINKED ISSUE #9669: Add a meta prop to react-admin reference components

erwanMarmelab:
## Problem
`<ReferenceManyField>` does not support `queryOptions`

## Solution
Add `queryOptions` support for `<ReferenceManyField>` and  `useReferenceArrayFieldController`

Closes #9669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## v4.16.13
+
+* Fix `<AutocompleteInput createLabel>` ([#9712](https://github.com/marmelab/react-admin/pull/9712)) ([fzaninotto](https://github.com/fzaninotto))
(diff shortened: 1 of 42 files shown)


fzaninotto:
This should be PRed against `next` as it's a new feature.

PR #9751, opened 2024-03-29
Add support of variant prop for FilterButton

adguernier:
This PR attempts to add MUI `variant` support for `<FilterButton>`

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
@@ -372,3 +372,45 @@ export const WithAutoCompleteArrayInput = (args: {
         </MemoryRouter>
     );
 };
+
+export const WithMuiButtonProps = () => {
+    const postFilters: React.ReactElement[] = [
+        <TextInput
+            label="Title"
+            source="title"
+            defaultValue="Accusantium qui nihil voluptatum quia voluptas maxime ab similique"
+        />,
+    ];
+    return (
+        <MemoryRouter>
+            <Admin
+                dataProvider={fakerestDataProvider(data)}
+                store={memoryStore()}
+            >
+                <Resource
+                    name="posts"
+                    list={
+                        <ListBase>
+                            <TopToolbar>
+                                <>
+                                    <FilterForm filters={postFilters} />
+                                    <FilterButton
+                                        filters={postFilters}
+                                        variant="outlined"
+                                        size="small"
+                                    />
+                                </>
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
```diff
@@ -372,3 +372,45 @@ export const WithAutoCompleteArrayInput = (args: {
         </MemoryRouter>
     );
 };
+
+export const WithMuiButtonProps = () => {
```
erwanMarmelab, line 376: **suggestion:** ‏Could  it be interesting to have 2 stories to show it: `size` and `variant`?

adguernier: Done

packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
```diff
@@ -179,17 +182,25 @@ const ListToolbar = (props: {
...
-            <TextField source="id" />
-            <TextField source="title" />
-            <TextField source="body" />
-        </Datagrid>
-        <Pagination />
-    </ListBase>
-);
+    buttonProps?: FilterButtonProps;
+}) => {
+    console.log('buttonPropssss', { ...props.buttonProps });
```
erwanMarmelab, line 187: console.log spotted

packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
```diff
@@ -372,3 +383,65 @@ export const WithAutoCompleteArrayInput = (args: {
         </MemoryRouter>
     );
 };
+
+export const WithButtonVariant = () => {
```
erwanMarmelab, line 387: **polish:** ‏May we name it as the prop name

```suggestion
export const Variant = () => {
```

packages/ra-ui-materialui/src/list/filter/FilterButton.stories.tsx
```diff
@@ -372,3 +383,65 @@ export const WithAutoCompleteArrayInput = (args: {
...
+                            }}
+                        />
+                    }
+                />
+            </Admin>
+        </MemoryRouter>
+    );
+};
+
+export const WithButtonSize = () => {
```
erwanMarmelab, line 418: **polish:** ‏May we name it as the prop name

```suggestion
export const Size = () => {
```

fzaninotto (approved):

PR #9752, opened 2024-04-02
Fix JS docs (@typedef)

erwanMarmelab:
We have many components that have @typedef and @prop in their documentation.

They should only use param

Example

```
  * @typedef Props
  * @property {Object} record: The current resource record
```

To:

```
  * @param {Object} props
  * @param {Object} props.record The current resource record
```

### DIFF
--- a/packages/ra-core/src/controller/create/useCreateContext.tsx
+++ b/packages/ra-core/src/controller/create/useCreateContext.tsx
@@ -15,7 +15,7 @@ import { CreateControllerResult } from './useCreateController';
  *
  * The given props will take precedence over context values.
  *
- * @typedef {Object} CreateControllerProps
+ * @param {Object} CreateControllerProps
  *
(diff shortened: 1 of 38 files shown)


fzaninotto:
I don't agree. The second definition means the function has 2 params, which is wrong.  What needs to be done is to set the param to the defined type, e.g. 

```
  * @typedef Props
  * @property {Object} record: The current resource record
  *
  * @param {Props} 
```

fzaninotto:
What we actually need to do is to move the `@param` JSDoc annotations to the TypeScript Prop types definitions, as in 

https://github.com/marmelab/react-admin/blob/6b78433893e0f7fc48db69c63a529b3c2e3d5c60/packages/ra-core/src/controller/list/useListController.ts#L181-L392

fzaninotto:
Closing as we need to go in another direction

PR #9760, opened 2024-04-08
[TypeScript] Make types more strict in ra-core, part III

fzaninotto:
This PR finally turns `strictNullChecks` to `true` in `ra-core`, and solves the remaining 63 TS compilation errors on `ra-core`.

Refs #9622
Follows #9644, #9741 , #9743

## `useRecordContext` Returns `undefined` When No Record Is Available

The `useRecordContext` hook reads the current record from the `RecordContext`. This context may be empty (e.g. while the record is being fetched). The return type for `useRecordContext` has been modified to `Record | undefined` instead of `Record` to denote this possibility.

As a consequence, the TypeScript compilation of your project may fail if you don't check the existence of the record before reading it.

To fix this error, your code should handle the case where `useRecordContext` returns `undefined`:

```diff
const MyComponent = () => {
    const record = useRecordContext();
+   if (!record) return null;
    return (
        <div>
            <h1>{record.title}</h1>
            <p>{record.body}</p>
        </div>
    );
};
```

### DIFF
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -157,6 +157,11 @@ export const useListParams = ({
                 tempParams.current = queryReducer(query, action);
                 // schedule side effects for next tick
                 setTimeout(() => {
+                    if (!tempParams.current) {
+                        throw new Error(
+                            'Race condition in changeParams detected'
+                        );
+                    }
                     if (disableSyncWithLocation) {
                         setLocalParams(tempParams.current);
                     } else {
@@ -262,10 +267,10 @@ export const useListParams = ({
 
     return [
         {
-            displayedFilters: displayedFilterValues,
             filterValues,
             requestSignature,
             ...query,
+            displayedFilters: displayedFilterValues,
         },
         {
             changeParams,
@@ -375,6 +380,9 @@ export const getNumberOrDefault = (
     possibleNumber: string | number | undefined,
     defaultValue: number
 ) => {
+    if (typeof possibleNumber === 'undefined') {
+        return defaultValue;
+    }
     const parsedNumber =
         typeof possibleNumber === 'string'
             ? parseInt(possibleNumber, 10)
--- a/packages/ra-core/src/controller/list/useListParams.ts
+++ b/packages/ra-core/src/controller/list/useListParams.ts
@@ -158,9 +158,8 @@ export const useListParams = ({
                 // schedule side effects for next tick
                 setTimeout(() => {
                     if (!tempParams.current) {
-                        throw new Error(
-                            'Race condition in changeParams detected'
-                        );
+                        // the side effects were already processed by another changeParams
+                        return;
                     }
                     if (disableSyncWithLocation) {
                         setLocalParams(tempParams.current);
--- a/packages/ra-core/src/core/CoreAdminContext.tsx
+++ b/packages/ra-core/src/core/CoreAdminContext.tsx
@@ -3,7 +3,11 @@ import { useMemo } from 'react';
 import { QueryClientProvider, QueryClient } from '@tanstack/react-query';
 
 import { AdminRouter } from '../routing';
-import { AuthContext, convertLegacyAuthProvider } from '../auth';
+import {
+    AuthContext,
+    defaultAuthProvider,
+    convertLegacyAuthProvider,
+} from '../auth';
 import {
     DataProviderContext,
     convertLegacyDataProvider,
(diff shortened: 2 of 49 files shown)


packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -229,60 +230,62 @@ export const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
...
-                                rowStyle,
-                                selectedIds,
-                                isRowSelectable,
-                            },
-                            children
-                        )}
-                    </Table>
-                </div>
-            </DatagridRoot>
+            <OptionalResourceContextProvider value={resource}>
```
fzaninotto, line 233: Stricter types revealed a bug: Using a Datagrid in controlled mode with an EditButton as a child used to fail silently (the EditButton link didn't work), because the resource set on the Datagrid was never propagated to the child. Now the failure is loud ans forces to add this ResourceContext.

packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -109,11 +110,13 @@ export const useFormGroup = (name: string): FormGroupState => {
...
 
     useEffect(() => {
+        if (!formGroups) return;
         // Whenever the group content changes (input are added or removed)
         // we must update its state
-        return formGroups.subscribe(name, () => {
+        const unsubscribe = formGroups.subscribe(name, () => {
             updateGroupState();
         });
+        return unsubscribe;
```
djhi, line 119: I don't think we have an `unsubscribe` method. We should add one

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
```diff
@@ -70,7 +70,7 @@ describe('<ReferenceArrayField />', () => {
                             record={{ id: 123, barIds: [1, 2] }}
                             reference="bar"
                         >
-                            <SingleFieldList>
+                            <SingleFieldList linkType={false}>
```
djhi, line 73: Why is this necessary? `linkType` is optional in `SingleFieldList` props

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -157,6 +157,11 @@ export const useListParams = ({
                 tempParams.current = queryReducer(query, action);
                 // schedule side effects for next tick
                 setTimeout(() => {
+                    if (!tempParams.current) {
+                        throw new Error(
+                            'Race condition in changeParams detected'
```
djhi, line 162: Users might need more information to solve this

djhi (changes requested):

adguernier (approved):

packages/ra-ui-materialui/src/field/ReferenceArrayField.spec.tsx
fzaninotto: Because the test doesn't define a `ResourceContext`, and `<SingleFieldList>` defaults to creating a link. It's not possible to create a link without a Resource. Before this PR, this caused an empty/invalid link. Now, it throws an error. To avoid this error, we must disable the ink or define a `ResourceContext`.

packages/ra-core/src/controller/list/useListParams.ts
fzaninotto: I agree, but this error should never occur (check the source code). To be honest, I added it just to please TypeScript.

packages/ra-core/src/form/useFormGroup.ts
fzaninotto: Yes we do:

https://github.com/marmelab/react-admin/blob/131619a8bfca889968daab748f1993e25c763d54/packages/ra-core/src/form/FormGroupsProvider.tsx#L34-L38

packages/ra-core/src/controller/list/useListParams.ts
fzaninotto: Now that you mention it, throwing an error may break the app in legitimate cases (e.g. two successive changeParams calls). I've converted it to an early return.

djhi (approved):

packages/ra-core/src/core/CoreAdminContext.tsx
```diff
@@ -187,9 +191,11 @@ React-admin requires a valid dataProvider function to work.`);
...
     const finalAuthProvider = useMemo(
         () =>
-            authProvider instanceof Function
-                ? convertLegacyAuthProvider(authProvider)
-                : authProvider,
+            authProvider
+                ? authProvider instanceof Function
+                    ? convertLegacyAuthProvider(authProvider)
+                    : authProvider
+                : defaultAuthProvider,
```
slax57, line 198: Just found out this is a BC, because now all admins have a default non-null authProvider.
So this breaks code like the following:

```tsx
        <Root
            label={
                authProvider != null
                    ? 'ra.auth.user_menu'
                    : 'ra.configurable.customize'
            }
            icon={
                authProvider ? (
                    <Avatar src={identity.avatar} />
                ) : (
                    <SettingsIcon />
                )
            }
            {...props}
        />
```

Is there a way to avoid it?

slax57: https://github.com/marmelab/react-admin/pull/9861 attempts to fix this BC

PR #9761, opened 2024-04-08
[TypeScript] Make types more strict in ra-ui-materialui

fzaninotto:
Decrease TS compilation errors in strictNullChecks mode from 216 to 164.  Small steps.

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
@@ -167,6 +167,7 @@ export const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
     // we manage row selection at the datagrid level to allow shift+click to select an array of rows
     const handleToggleItem = useCallback(
         (id, event) => {
+            if (!data) return;
             const ids = data.map(record => record.id);
             const lastSelectedIndex = ids.indexOf(lastSelected.current);
             lastSelected.current = event.target.checked ? id : null;
@@ -182,7 +183,7 @@ export const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
                     ? union(selectedIds, idsBetweenSelections)
                     : difference(selectedIds, idsBetweenSelections);
 
-                onSelect(
+                onSelect?.(
                     isRowSelectable
                         ? newSelectedIds.filter((id: Identifier) =>
                               isRowSelectable(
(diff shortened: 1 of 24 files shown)


packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -259,7 +260,6 @@ export const Datagrid: FC<DatagridProps> = React.forwardRef((props, ref) => {
                                     hasBulkActions,
                                     isRowSelectable,
                                     onSelect,
-                                    resource,
```
fzaninotto, line 262: We no longer need to pass down resource as there is now always a ResourceContext, even in controlled mode.

fzaninotto:
Switching to RFR

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
```diff
@@ -2,19 +2,20 @@ import * as React from 'react';
 import { memo } from 'react';
 import IconButton from '@mui/material/IconButton';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
-import { useTranslate, useExpandAll } from 'ra-core';
+import { useTranslate, useExpandAll, useResourceContext } from 'ra-core';
 import { DatagridClasses } from './useDatagridStyles';
 import clsx from 'clsx';
 
 interface ExpandAllButtonProps {
-    resource: string;
```
slax57, line 10: same

packages/ra-ui-materialui/src/list/datagrid/DatagridHeader.tsx
```diff
@@ -189,7 +186,6 @@ export interface DatagridHeaderProps<RecordType extends RaRecord = any> {
     data?: RecordType[];
     onSelect?: (ids: Identifier[]) => void;
     onToggleItem?: (id: Identifier) => void;
-    resource?: string;
```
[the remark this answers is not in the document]
fzaninotto: i don't see in which case users would inject the `resource` themselves, outside of the context of the Datagrid itself (which used to override the resource anyway). So I'm -1 for adding it to the Upgrade guide.

packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.tsx
slax57: Actually this is an internal component only used by our own DatagridHeader, so this doesn't count as a BC

slax57 (approved):

PR #9764, opened 2024-04-10
Add better types to data-generator

djhi:
## Problem

When using the `data-generator-retail` package, the generated is not typed, making consumers declare the generated types themselves.

## Solution

Provide types

Wait for next version of `json-graphql-server` or the GQL demo won't work

### DIFF
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -37,7 +37,7 @@
         "recharts": "^2.1.15"
     },
     "scripts": {
-        "dev": "vite --force",
+        "dev": "vite",
         "build": "tsc && vite build",
         "preview": "vite preview"
     },
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -18,7 +18,7 @@
         "graphql": "^15.6.0",
         "graphql-tag": "^2.12.6",
         "inflection": "~1.12.0",
-        "json-graphql-server": "~2.3.0",
+        "json-graphql-server": "~3.0.1",
         "prop-types": "^15.8.1",
         "query-string": "^7.1.1",
         "ra-data-fakerest": "^5.0.0-alpha.0",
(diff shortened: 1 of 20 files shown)


fzaninotto (commented):
Would you mind rebasing on `next`? Also, the e-commerce demo redefines those types, it should now import them instead.

https://github.com/marmelab/react-admin/blob/ae7c8ef1f2b0637dc98c7b0a6e15389107779e25/examples/demo/src/types.ts#L5-L73

fzaninotto (approved):
LGTM. But I'm thinking: since we're targeting v5, maybe we can get rid of the serialized/unserialized date switch? I think it makes more sense to have it always serialized.

djhi:
> LGTM. But I'm thinking: since we're targeting v5, maybe we can get rid of the serialized/unserialized date switch? I think it makes more sense to have it always serialized.

Sounds good to me.

djhi:
> LGTM. But I'm thinking: since we're targeting v5, maybe we can get rid of the serialized/unserialized date switch? I think it makes more sense to have it always serialized.

Actually I don't think we should. Many libraries

examples/demo/package.json
```diff
@@ -37,7 +37,7 @@
         "recharts": "^2.1.15"
     },
     "scripts": {
-        "dev": "vite",
+        "dev": "vite --force",
```
fzaninotto, line 40: why this change?

examples/demo/package.json
fzaninotto, line 1: you're missing a bump on the   "json-graphql-server" version, and also an update to the yarn.lock file.

fzaninotto (changes requested):

examples/demo/package.json
djhi: There are actually some issues with `json-graphql-server`. Still investigating

examples/demo/package.json
djhi: Reverted

fzaninotto (approved):

Issue #9695, opened 2024-03-01
Change title property type in core display components <Edit>, <Show>, <View> 


antoinefricker:
#### Is your feature request related to a problem?
In the `<Edit>`, `<Create>` and `<Show>` components, the title property is typed as `title?: string | ReactElement;`
In our project, we would like to customize title content using the `<Title>` component within the aforementioned components children. It is now uncomfortable to implement as RA stacks default title and the userland defined title.  

#### Describe the solution you'd like
I would find it very _react-admin styled_ to allow a `false` value for the `title` property of the core display components.
It would require to update those, adding a render condition to the `<Title>` component - [here for example](https://github.com/marmelab/react-admin/blob/466f49756ff368bb987b115d3e30d7762ee8d85b/packages/ra-ui-materialui/src/detail/EditView.tsx#L47)

> By the way, using an empty string does not do the job as there is a [falsy evaluation](https://github.com/marmelab/react-admin/blob/466f49756ff368bb987b115d3e30d7762ee8d85b/packages/ra-ui-materialui/src/layout/PageTitle.tsx#L18C14-L18C20) of the string in `<PageTitle>`.

#### Describe alternatives you've considered
None

#### Additional context
No relevant

fzaninotto:
Thanks for your suggestion. 

Supporting a falsy value for `<Edit title>` props does add a bit of flexibility, and we'd welcome a PR for that. 
`<Edit>` doesn't do much more than rendering the title, actions, and a `Card` around the page content. If you don't use the Title feature of `<Edit>`, I suggest you use `<EditBase>` instead.

PR #9768, opened 2024-04-10
edit / show title prop: add 'false' and callback choices

hiaselhans:
This PR adds two more choices to the 'title' prop of edit / show views:

- false to display no title (#9695)
- callback with signature `(arg0?: RaRecord) => string`

this proved to be a very flexible and simple solution for me.


fzaninotto:
`master` is feature-freeze, please PR against `next` instead. Also, please add unit tests and document this new feature.

PR #9774, opened 2024-04-11
Restore List Scroll Position on Edit and Create Views side effects

djhi:
## Problem

When users have a long list page, navigate from it to an edit or create page, the `onSuccess` redirection of those pages redirect them to the list but loose their previous scroll position. This is bad UX.

## Solution

- Introduce the `useRestoreScrollPosition` hook that tracks the page scroll position and restore it when the component mounts.
- Use in the `ra-ui-materialui` `List` component
- Remove the `_scrollToTop` state when redirecting to a list view

### DIFF
--- a/packages/ra-core/src/routing/useRedirect.ts
+++ b/packages/ra-core/src/routing/useRedirect.ts
@@ -71,7 +71,7 @@ export const useRedirect = () => {
             } else {
                 // redirection to an internal link
                 navigate(createPath({ resource, id, type: redirectTo }), {
-                    state: { _scrollToTop: true, ...state },
+                    state: { _scrollToTop: redirectTo !== 'list', ...state },
                 });
                 return;
             }
--- a/packages/ra-core/src/routing/useRestoreScrollPosition.ts
+++ b/packages/ra-core/src/routing/useRestoreScrollPosition.ts
@@ -1,0 +1,49 @@
+import { useEffect } from 'react';
+import { useStore } from '../store';
+import { debounce } from 'lodash';
+
+/**
+ * A hook that tracks the scroll position and restores it when the component mounts.
+ * @param key The key under which to store the scroll position in the store
+ * @param debounceMs The debounce time in milliseconds
+ *
+ * @example
+ * import { ListBase, useRestoreScrollPosition } from 'ra-core';
+ *
+ * const MyCustomList = (props) => {
+ *    useRestoreScrollPosition('my-list');
+ *   return <ListBase {...props} />;
+ * };
+ */
+export const useRestoreScrollPosition = (key: string, debounceMs = 250) => {
+    const position = useTrackScrollPosition(key, debounceMs);
+
+    useEffect(() => {
+        if (position != null) {
+            window.scrollTo(0, position);
+        }
+        // We only want to run this effect on mount
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+};
+
+export const useTrackScrollPosition = (key: string, debounceMs = 250) => {
+    const [position, setPosition] = useStore(key);
+
+    useEffect(() => {
+        if (typeof window === 'undefined') {
+            return;
+        }
+        const handleScroll = debounce(() => {
+            setPosition(window.scrollY);
+        }, debounceMs);
+
+        window.addEventListener('scroll', handleScroll);
+
+        return () => {
+            window.removeEventListener('scroll', handleScroll);
+        };
+    }, [debounceMs, setPosition]);
+
+    return position;
+};
--- a/packages/ra-core/src/routing/index.ts
+++ b/packages/ra-core/src/routing/index.ts
@@ -6,5 +6,6 @@ export * from './useBasename';
 export * from './useCreatePath';
 export * from './useRedirect';
 export * from './useScrollToTop';
+export * from './useRestoreScrollPosition';
 export * from './types';
 export * from './TestMemoryRouter';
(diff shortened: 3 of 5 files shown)


packages/ra-core/src/routing/useRestoreScrollPosition.ts
```diff
@@ -0,0 +1,52 @@
...
+ *   return <ListBase {...props} />;
+ * };
+ */
+export const useRestoreScrollPosition = (key: string, debounceMs = 250) => {
+    const position = useTrackScrollPosition(key, debounceMs);
+    const location = useLocation();
+
+    useEffect(() => {
+        if (position != null && location.state?._scrollToTop !== true) {
+            console.log('Restoring scroll position', position);
```
adguernier, line 25: console.log spotted

adguernier (changes requested):

packages/ra-core/src/routing/useRestoreScrollPosition.ts
djhi, line 25:
```suggestion
```

adguernier (approved):

fzaninotto:
Now I'm thinking: what if there is more than one `<List>` per page? Or if the List is not the main page component (e.g. in the TagEdit page in the simple example)? We don't have any way to disable the scroll restoration in List. I think that's a problem.

djhi:
Should we add a `disableScrollRestoration` prop?

fzaninotto:
Other suggestion: scroll restoration should be dealt with earlier, in the CoreAdminRoutes, instead of in the List component.

packages/ra-core/src/routing/RestoreScrollPosition.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * @example
+ * import { ListBase, RestoreScrollPosition } from 'ra-core';
+ *
+ * const MyCustomList = (props) => (
+ *   <RestoreScrollPosition key="my-list>
+ *     <ListBase {...props} />
+ *   </RestoreScrollPosition>
+ * };
+ */
+export const RestoreScrollPosition = ({
```
fzaninotto, line 19: the component isn't exported

packages/ra-core/src/routing/RestoreScrollPosition.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * @param debounceMs The debounce time in milliseconds
+ *
+ * @example
+ * import { ListBase, RestoreScrollPosition } from 'ra-core';
+ *
+ * const MyCustomList = (props) => (
+ *   <RestoreScrollPosition key="my-list>
+ *     <ListBase {...props} />
+ *   </RestoreScrollPosition>
+ * };
```
fzaninotto, line 17: I think you should use another example than a list, as it's useless in that case

fzaninotto (changes requested):

packages/ra-core/src/routing/RestoreScrollPosition.tsx
djhi: Done but somehow git does not show this file as outdated

packages/ra-core/src/routing/RestoreScrollPosition.tsx
djhi: Fixed

packages/ra-core/src/routing/useRedirect.ts
```diff
@@ -71,7 +71,10 @@ export const useRedirect = () => {
             } else {
                 // redirection to an internal link
                 navigate(createPath({ resource, id, type: redirectTo }), {
-                    state: { _scrollToTop: true, ...state },
+                    state:
+                        redirectTo === 'list'
+                            ? state
```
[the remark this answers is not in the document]
djhi: Added but somehow git does not flag the file as outdated

packages/ra-core/src/routing/RestoreScrollPosition.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * A component that tracks the scroll position and restores it when the component mounts.
+ * @param children The content to render
+ * @param key The key under which to store the scroll position in the store
+ * @param debounceMs The debounce time in milliseconds
+ *
+ * @example
+ * import { RestoreScrollPosition } from 'ra-core';
+ *
+ * const MyCustomPage = () => {
+ *   <RestoreScrollPosition key="my-list>
```
fzaninotto, line 14: syntax error, missing quote

packages/ra-core/src/routing/useRedirect.ts
```diff
@@ -71,7 +71,12 @@ export const useRedirect = () => {
             } else {
                 // redirection to an internal link
                 navigate(createPath({ resource, id, type: redirectTo }), {
-                    state: { _scrollToTop: true, ...state },
+                    state:
+                        // We want to restore the scroll position only if we are navigating to a list
+                        // This is done with <RestoreScrollPosition> in <Resource>
```
fzaninotto, line 76:
```suggestion
                        // We force the scrollToTop except when navigating to a list
                        // where this is already done by <RestoreScrollPosition> in <Resource>
```

packages/ra-core/src/routing/RestoreScrollPosition.tsx
djhi, line 14:
```suggestion
 *   <RestoreScrollPosition key="my-list">
```

fzaninotto (approved):

PR #9781, opened 2024-04-18
[v5] Remove condition restricting error display to touched inputs

slax57:
### Problem

All RA inputs have this [condition](https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/TextInput.tsx#L79):

```js
error={(isTouched || isSubmitted) && invalid}
```

which only allows to display validation errors if the input is touched or the form was submitted.

This prevents displaying errors preventively on inputs that have not been touched. Besides, it could be argued that it’s the form lib’s responsibility to decide what inputs are in error, and RA should not be opinionated about this.

### Solution

Remove this check from all inputs.

### Additional info

This will probably render https://github.com/marmelab/react-admin/pull/9724 obsolete.


packages/ra-ui-materialui/src/input/InputHelperText.tsx
```diff
@@ -3,7 +3,7 @@ import { isValidElement, ReactElement } from 'react';
 import { useTranslate, ValidationError, ValidationErrorMessage } from 'ra-core';
 
 export const InputHelperText = (props: InputHelperTextProps) => {
-    const { helperText, touched, error } = props;
+    const { helperText, touched = true, error } = props;
```
fzaninotto, line 6: Why do you keep the `touched` prop?

slax57: To avoid a breaking change (since `InputHelperText` is mentioned in the docs). But if you think that is unnecessary I can remove it.

packages/ra-ui-materialui/src/input/InputErrors.stories.tsx
```diff
@@ -0,0 +1,364 @@
+import * as React from 'react';
+import { FormGroupContextProvider, required, useFormGroup } from 'ra-core';
+import { useFormContext } from 'react-hook-form';
+import { Button } from '@mui/material';
```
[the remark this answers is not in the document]
slax57: nice catches! thanks

adguernier (approved):

packages/ra-ui-materialui/src/input/InputHelperText.tsx
fzaninotto: Yes, it's the right moment to clean this up

docs/Upgrade.md
```diff
@@ -906,6 +906,16 @@ The following components are not affected and can still be used in standalone mo
...
+## Inputs No Longer Require To Be Touched To Display A Validation Error
+
+In previous versions, validation errors were only displayed after the input was touched or the form was submitted. In v5, validation errors are fully entrusted to the form library (`react-hook-form`), which is responsible to decide when to display them.
+
+For most use-cases this will have no impact, because `react-hook-form` works the same way (it will wait for an input to be touched before triggering its validation).
+
+But this should help with some advanced cases, for instance if some validation errors need to be displayed on untouched fields.
+
+It will also improve the user experience, as the form `isValid` state will be consistent with error messages displayed on inputs, regardless of whether they have been touched or not.
+
```
djhi, line 918: Might be useful to remind users of the `mode` prop with a link to RHF documentation

packages/ra-ui-materialui/src/input/InputHelperText.tsx
slax57: done

docs/Upgrade.md
slax57: done

djhi (approved):

fzaninotto (approved):

PR #9788, opened 2024-04-22
Add a default unique id for `useInput`

### LINKED ISSUE #9534: useInput defaults to non-unique IDs
This is mostly relevant when using the enterprise package `@react-admin/ra-form-layout`

use-case: using a modal "create" form while on an "edit" page to create a sub-resource of the same type (but could also apply to different types which share property names)

**What you were expecting:**
All form elements (i.e. `useInput`) should default to having a unique ID which does not conflict with other forms (i.e. via React's `useId` hook)

**What happened instead:**
Form elements default to using the `source` as an ID, which is not unique if multiple forms are on the page.

**Steps to reproduce:**

```tsx
const MyEdit = () => (
  <Edit resource="person">
    <SimpleForm>
      <TextInput name="name" source="name" label="Name" /> {/* id is implicitly 'name' */}
      <CreateInDialogButton label="Add child" resource="person">
        <SimpleForm>
          <TextInput name="name" source="name" label="Name" /> {/* id is also implicitly 'name' */}
        </SimpleForm>
      </CreateInDialogButton>
    </SimpleForm>
  </Edit>
);
```

The browser will complain about multiple fields sharing the same ID, and in some situations this will cause the fields to be impossible to select (e.g. playwright browser testing fails to find the fields)

It is possible to work around this by explicitly providing an `id` to all fields, but this is tedious, easy to forget, and difficult to diagnose failures. Changing the default to a unique ID would make it guaranteed to be correct without extra work for the developer.

**Environment**

* React-admin version: 4.16.2
* React version: 18.2.0

erwanMarmelab:
## To do

- [x] change default id
- [x] adapt tests
- [x] Add a section in the upgrade guide

Closes #9534


packages/ra-core/src/form/useInput.ts
```diff
@@ -132,7 +132,7 @@ export const useInput = <ValueType = any>(
     };
 
     return {
-        id: id || finalSource,
+        id: id || `use-input-${finalSource}`,
```
djhi, line 135: We should leverage `useId` if `id` is not provided

djhi (changes requested):

packages/ra-core/src/form/useInput.ts
```diff
@@ -132,7 +133,9 @@ export const useInput = <ValueType = any>(
     };
 
     return {
-        id: id || finalSource,
+        id:
+            id ||
+            `${defaultId.substring(1, defaultId.length - 1)}-${finalSource}`,
```
djhi, line 138: Why not just `defaultId`?

erwanMarmelab: It was to be more readable. 
Applied

fzaninotto (changes requested):

fzaninotto:
You should probably add a section in the upgrade guide as this is a breaking change

packages/ra-core/src/form/useInput.ts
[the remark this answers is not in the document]
erwanMarmelab: removed

docs/useInput.md
```diff
@@ -41,16 +41,16 @@ const TitleInput = ({ source, label }) => {
...
+|----------------|----------|--------------------------------|-------------------- |-------------------------------------------------------------------|
+| `source`       | Required | `string`                       | -                   | The name of the field in the record                               |
+| `defaultValue` | Optional | `any`                          | -                   | The default value of the input                                    |
+| `format`       | Optional | `Function`                     | -                   | A function to format the value from the record to the input value |
+| `parse`        | Optional | `Function`                     | -                   | A function to parse the value from the input to the record value  |
+| `validate`     | Optional | `Function` &#124; `Function[]` | -                   | A function or an array of functions to validate the input value   |
+| `id`           | Optional | `string`                       | `:r[input number]:` | The id of the input                                               |
```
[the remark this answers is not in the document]
erwanMarmelab: done

docs/Upgrade.md
```diff
@@ -928,6 +928,12 @@ If you were using this prop, you can safely remove it.
 
 The `data-generator-retail` package has been updated to provide types for all its records. In the process, we renamed the `commands` resource to `orders`. Accordingly, the `nb_commands` property of the `customers` resource has been renamed to `nb_orders` and the `command_id` property of the `invoices` and `reviews` resources has been renamed to `order_id`.
 
+## Inputs default ids are auto-generated
+
+In previous versions, the input default id was the source of the input. In v5, inputs defaults ids are auto-generated with [React useId()](https://react.dev/reference/react/useId).
+
+**Tip:** You still can pass an id as prop of any [react-admin input](./Inputs.md) or use a [reference](https://fr.react.dev/reference/react/useRef).
+
```
djhi, line 936: What's the impact for users? You should mention that if they relied on this for tests, they should pass the `id`

erwanMarmelab: something like this ?

djhi (approved):

PR #9790, opened 2024-04-23
[Typescript] Make types more strict in ra-ui-materialui, part II

fzaninotto:
This PR reduces TypeScript compilation errors with StrictNullCheck from 162 to 109 (-32%)

Refs #9622
Follows #9761

### DIFF
--- a/packages/ra-ui-materialui/src/button/ExportButton.tsx
+++ b/packages/ra-ui-materialui/src/button/ExportButton.tsx
@@ -113,10 +113,6 @@ ExportButton.propTypes = {
     label: PropTypes.string,
     maxResults: PropTypes.number,
     resource: PropTypes.string,
-    sort: PropTypes.exact({
-        field: PropTypes.string,
-        order: PropTypes.oneOf(['ASC', 'DESC'] as const),
-    }),
     icon: PropTypes.element,
     meta: PropTypes.any,
 };
--- a/packages/ra-ui-materialui/src/button/ExportButton.tsx
+++ b/packages/ra-ui-materialui/src/button/ExportButton.tsx
@@ -92,10 +92,8 @@ const defaultIcon = <DownloadIcon />;
 const sanitizeRestProps = ({
     resource,
     ...rest
-}: Omit<
-    ExportButtonProps,
-    'sort' | 'maxResults' | 'label' | 'exporter' | 'meta'
->) => rest;
+}: Omit<ExportButtonProps, 'maxResults' | 'label' | 'exporter' | 'meta'>) =>
+    rest;
 
 interface Props {
     exporter?: Exporter;
--- a/packages/ra-ui-materialui/src/list/ListActions.tsx
+++ b/packages/ra-ui-materialui/src/list/ListActions.tsx
@@ -48,7 +48,6 @@ export const ListActions = (props: ListActionsProps) => {
     const { className, filters: filtersProp, hasCreate: _, ...rest } = props;
 
     const {
-        sort,
         displayedFilters,
         filterValues,
         exporter,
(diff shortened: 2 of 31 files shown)


packages/ra-ui-materialui/src/list/ListActions.tsx
```diff
@@ -72,12 +71,7 @@ export const ListActions = (props: ListActionsProps) => {
                     : filters && <FilterButton />}
                 {hasCreate && <CreateButton />}
                 {exporter !== false && (
-                    <ExportButton
-                        disabled={total === 0}
-                        resource={resource}
-                        sort={sort}
-                        filterValues={filterValues}
-                    />
+                    <ExportButton disabled={total === 0} resource={resource} />
```
adguernier, line 74: **question**: we no longer want `<ExportButton>` export the same data as the list?
Also and if I'm not mistaken, `sort` param seems to be not used, don't we should remove it from `ListActionsProps`?

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
```diff
@@ -94,7 +94,7 @@ export const Pagination: FC<PaginationProps> = memo(props => {
     }
 
     // Avoid rendering TablePagination if "page" value is invalid
-    if (total === 0 || page < 1 || (total != null && page > totalPages)) {
+    if (total === 0 || page < 1 || (total != null && page > totalPages!)) {
```
adguernier, line 97: **question**: `totalPages` could be `undefined` isn't it?

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -62,15 +62,28 @@ const DatagridRow: FC<DatagridRowProps> = React.forwardRef((props, ref) => {
...
 
+    if (typeof id === 'undefined') {
+        throw new Error('DatagridRow expects an id prop');
+    }
     const context = useDatagridContext();
     const translate = useTranslate();
     const record = useRecordContext(props);
+    if (!record) {
+        throw new Error(
+            'DatagridRow can only be used within a RecordContext or be passed a recoerd prop'
```
adguernier, line 73:
```suggestion
            'DatagridRow can only be used within a RecordContext or be passed a record prop'
```

packages/ra-ui-materialui/src/button/ExportButton.tsx
```diff
@@ -92,10 +92,8 @@ const defaultIcon = <DownloadIcon />;
 const sanitizeRestProps = ({
     resource,
     ...rest
-}: Omit<
-    ExportButtonProps,
-    'sort' | 'maxResults' | 'label' | 'exporter' | 'meta'
->) => rest;
+}: Omit<ExportButtonProps, 'maxResults' | 'label' | 'exporter' | 'meta'>) =>
```
adguernier, line 95: **suggestion**: Since `ExportButtonProps` does not contains the `sort` prop, we could remove this from `ExportButton.propTypes` as well

fzaninotto: Right, fixed

packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
fzaninotto: not if total is not null

adguernier (approved):

packages/ra-ui-materialui/src/list/ListActions.tsx
```diff
@@ -115,7 +104,6 @@ ListActions.propTypes = {
 };
 
 export interface ListActionsProps extends ToolbarProps {
-    sort?: SortPayload;
```
erwanMarmelab, line 118: **chore:** 'SortPayload' is imported but no more used‏, we may could remove it
(SortPayload was imported line 6 but I can't add a comment on)

erwanMarmelab (approved):

packages/ra-ui-materialui/src/list/ListActions.tsx
fzaninotto: Right, fixed

PR #9797, opened 2024-04-29
[TypeScript] Fix remaining strictNullCheck errors in ra-ui-materialui

djhi:


packages/ra-ui-materialui/src/field/ImageField.tsx
```diff
@@ -48,8 +49,10 @@ export const ImageField = <
             <Root className={className} {...sanitizeFieldRestProps(rest)}>
                 <ul className={ImageFieldClasses.list}>
                     {sourceValue.map((file, index) => {
-                        const fileTitleValue = get(file, title) || title;
-                        const srcValue = get(file, src) || title;
+                        const fileTitleValue = title
+                            ? get(file, title, title)
+                            : title;
+                        const srcValue = src ? get(file, src, src) : src;
```
fzaninotto, line 55: you've changed the meaning there, shouldn't this be 
```suggestion
                        const srcValue = src ? get(file, src, title) : title;
```

packages/ra-ui-materialui/src/field/TranslatableFields.tsx
```diff
@@ -78,7 +78,17 @@ export const TranslatableFields = (
         className,
     } = props;
     const record = useRecordContext(props);
+    if (!record) {
+        throw new Error(
+            `<TranslatableFields> was called outside of a ResourceContext and without a resource prop. You must set the resource prop.`
```
fzaninotto, line 83:
```suggestion
            `<TranslatableFields> was called outside of a RecordContext and without a record prop. You must set the record prop.`
```

packages/ra-ui-materialui/src/field/TranslatableFields.tsx
```diff
@@ -78,7 +78,17 @@ export const TranslatableFields = (
...
+        );
+    }
     const resource = useResourceContext(props);
+    if (!resource) {
+        throw new Error(
+            `<TranslatableFields> was called outside of a ResourceContext and without a resource prop. You must set the resource prop.`
```
fzaninotto, line 89:
```suggestion
            `<TranslatableFields> was called outside of a RecordContext and without a record prop. You must set the record prop.`
```

packages/ra-core/src/preferences/PreferencesEditorContext.tsx
```diff
@@ -2,22 +2,20 @@ import * as React from 'react';
...
-          enable: () => void;
-          disable: () => void;
-          path: string | null;
-          setPath: (path: string) => void;
-      }
-    | undefined;
+export type PreferencesEditorContextValue = {
+    editor: ReactElement | null;
+    setEditor: React.Dispatch<React.SetStateAction<ReactElement | null>>;
+    preferenceKey: string | null;
```
fzaninotto, line 11: I wonder why you prefer null to undefined here? Is this change necessary?

packages/ra-core/src/preferences/PreferencesEditorContextProvider.tsx
```diff
@@ -8,7 +8,7 @@ import {
 export const PreferencesEditorContextProvider = ({ children }) => {
     const [isEnabled, setIsEnabled] = useState(false);
     const [editor, setEditor] = useState<ReactElement | null>(null);
-    const [preferenceKey, setPreferenceKey] = useState<string>();
+    const [preferenceKey, setPreferenceKey] = useState<string | null>(null);
```
fzaninotto, line 11: you could default to undefined

packages/ra-ui-materialui/src/form/SimpleFormConfigurable.tsx
```diff
@@ -31,27 +31,28 @@ export const SimpleFormConfigurable = ({
...
+                                  : translate(
+                                        'ra.configurable.SimpleForm.unlabeled',
+                                        {
+                                            input: index,
+                                            _: `Unlabeled input #%{input}`,
+                                        }
+                                    ),
+                      }
+                    : null
+            )?.filter(column => column != null) ?? [];
```
fzaninotto, line 52: I think you should use an emptyArray const defined outside of the element to avoid rerenderings when the effect returns a different empty array object (modified 3 times)

packages/ra-ui-materialui/src/preferences/Configurable.tsx
```diff
@@ -73,6 +73,7 @@ export const Configurable = (props: ConfigurableProps) => {
     }
 
     const handleOpenEditor = () => {
+        if (!setEditor) return;
```
fzaninotto, line 76: I think we should rather throw an error in this case

packages/ra-ui-materialui/src/detail/CreateView.tsx
```diff
@@ -22,10 +22,7 @@ export const CreateView = (props: CreateViewProps) => {
     const { resource, defaultTitle } = useCreateContext();
 
     return (
-        <Root
-            className={clsx('create-page', className)}
-            {...sanitizeRestProps(rest)}
```
fzaninotto, line 27: I'm surprised this works, since CreateViewProps is declared as CreateProps (which is a mistake, as there are some Create props that CreateView will never get injected)

fzaninotto (changes requested):
We're getting close!

packages/ra-ui-materialui/src/detail/CreateView.tsx
djhi: And yet it does without any warning at runtime in all demos

packages/ra-core/src/preferences/PreferencesEditorContext.tsx
djhi: We could but it looked more coherent with the other props

packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -67,8 +68,10 @@ export const FileField = <
         return (
             <StyledList className={className} {...sanitizeFieldRestProps(rest)}>
                 {sourceValue.map((file, index) => {
-                    const fileTitleValue = get(file, title) || title;
-                    const srcValue = get(file, src) || title;
+                    const fileTitleValue = title
+                        ? get(file, title, title)
+                        : title;
+                    const srcValue = src ? get(file, src, src) : src;
```
djhi, line 74:
```suggestion
                    const srcValue = src ? get(file, src, title) : title;
```

packages/ra-ui-materialui/src/preferences/Configurable.tsx
```diff
@@ -73,6 +73,12 @@ export const Configurable = (props: ConfigurableProps) => {
     }
 
     const handleOpenEditor = () => {
+        if (!setEditor) {
+            throw new Error(
+                'PreferencesEditorContext is not available. Did you forget to put a PreferencesEditorContextProvider at a higher level?'
```
adguernier, line 78: **suggestion**(niptick): I like the way you present this error but other errors of this type throw this message: `Configurable must be used inside a PreferencesEditorContextProvider`. Could use the same shape to stay consistent?

```suggestion
                'Configurable must be used inside a PreferencesEditorContextProvider'
```

packages/ra-ui-materialui/src/preferences/Configurable.tsx
```diff
@@ -83,6 +89,11 @@ export const Configurable = (props: ConfigurableProps) => {
                 key: prefixedPreferenceKey,
             })
         );
+        if (!setPreferenceKey) {
+            throw new Error(
+                'PreferencesEditorContext is not available. Did you forget to put a PreferencesEditorContextProvider at a higher level?'
```
adguernier, line 94: (niptick)
```suggestion
                'Configurable must be used inside a PreferencesEditorContextProvider'
```

adguernier (approved):

packages/ra-ui-materialui/src/detail/CreateView.tsx
```diff
@@ -41,8 +40,15 @@ export const CreateView = (props: CreateViewProps) => {
     );
 };
 
-export type CreateViewProps = CreateProps;
-
+export interface CreateViewProps {
```
fzaninotto, line 43: I think it should extend the base div props (e.g. `style`), otherwise valid HTML attributes, that are passed down to the root components, con't be accepted. This also allows you to get rid of `className`. Same for the other 2.

fzaninotto (approved):

fzaninotto:
\o/

PR #9804, opened 2024-04-30
Upgrade inflection to v3

MohammedFaragallah:


examples/demo/src/categories/CategoryList.tsx
```diff
@@ -5,7 +5,7 @@ import {
     RecordContextProvider,
     useListContext,
 } from 'react-admin';
-import inflection from 'inflection';
+import * as inflection from 'inflection';
```
fzaninotto, line 8: I think you should only import the utility you need for each script, to enable tree-shaking.

MohammedFaragallah: I was thinking the same thing I'll work on it 🤝

fzaninotto (approved):

fzaninotto:
Excellent, thanks!

PR #9828, opened 2024-05-03
[Dev] Speed up unit tests by not using MUI Icons barrel files

fzaninotto:
Following the advice at https://mui.com/material-ui/guides/minimizing-bundle-size/#development-environment, I fixed the few named imports from `@mui/material-icons` and replaced them with default imports for individual icons.

The results are spectacular. The `npx jest` command now takes a fraction of the time. 

On my laptop:

* Before: 118s
* After: 41s

On the CI:

* Before: 4m58s
* After: 3m28s


djhi (approved):
Now we should try avoiding barrel files import in all our code and measure it too :fire:

PR #9830, opened 2024-05-04
docs: update italian locale reference

christianascone:
I created a new repository for Italian translation because the previous one (https://github.com/stefsava/ra-italian) is obsolete and it seems to be no longer maintained, so I updated the reference in `docs/TranslationLocales.md`.
Nothing more than this.


djhi (approved):
Thanks!

PR #9833, opened 2024-05-06
Fix strictNullCheck errors in examples

djhi:


examples/simple/src/authProvider.tsx
```diff
@@ -60,10 +62,16 @@ export default {
         return Promise.resolve(role ?? '');
     },
     getIdentity: () => {
+        const id = localStorage.getItem('login');
```
fzaninotto, line 65: No, we want the user to be authenticated by default in the simple example (see `checkAuth`), this will break it. 
Return the default identity if there is no login in localStorage.

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -113,7 +113,7 @@ const PostCreate = () => {
                 <FileInput
                     source="pdffile"
                     label="PDF-Template"
-                    accept="application/pdf"
+                    accept={{ 'application/pdf': ['.pdf'] }}
```
fzaninotto, line 116: Shouldn't we update the doc, too?

examples/simple/src/users/UserEdit.tsx
```diff
@@ -45,6 +45,7 @@ const EditActions = () => (
 const UserEditForm = () => {
     const { permissions } = usePermissions();
     const { save } = useSaveContext();
+    if (!save) return null;
```
fzaninotto, line 48: I'm surprised that `save` may be undefined, isn't this an error in the context definition?

fzaninotto (changes requested):

fzaninotto:
I think we should add the typescheck of the simple example to the typecheck CI step

examples/simple/src/users/UserEdit.tsx
djhi: Because the `SaveContext` is optional. You can use a `Form` without it by providing your own `onSubmit`

examples/simple/src/authProvider.tsx
djhi: As users are authenticated by default, this is just to make TS happy but ok

examples/simple/src/posts/PostCreate.tsx
djhi: Updated

djhi:
> I think we should add the typescheck of the simple example to the typecheck CI step

Added

docs/FileInput.md
```diff
@@ -72,7 +72,7 @@ Files are accepted or rejected based on the `accept`, `multiple`, `minSize` and
 Equivalent of [the `accept` attribute of an `<input type="file">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept). `accept` must be a valid [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml), according to [input element specification](https://www.w3.org/wiki/HTML/Elements/input/file) or a valid file extension.
 
 ```jsx
-<FileInput source="attachments" accept="application/pdf">
+<FileInput source="attachments" accept={{ 'application/pdf': ['.pdf'] }}>
```
fzaninotto, line 75: Can you also add a mention about this change in the Upgrade guide?

.github/workflows/test.yml
```diff
@@ -67,6 +67,27 @@ jobs:
...
+      uses: actions/checkout@v4
+    - name: Use Node.js LTS
+      uses: actions/setup-node@v4
+      with:
+        node-version: '18.x'
+    - uses: bahmutov/npm-install@v1
+      with:
+        install-command: yarn --immutable
+    - name: Build
+      run: make build
```
fzaninotto, line 83: Why do you need to build here? It shouldn't be necessary

.github/workflows/test.yml
```diff
@@ -67,6 +67,27 @@ jobs:
       env:
         CI: true
 
+simple-example:
```
fzaninotto, line 70: Can't you just add a step in the typecheck workflow?

docs/FileInput.md
djhi: Added

fzaninotto (approved):

PR #9836, opened 2024-05-08
Allow <Title /> to be nonConfigurable

markyao6275:
This addresses https://github.com/marmelab/react-admin/issues/8776

Testing:

Added storybook entry

Screenshot 2024-05-08 at 9 51 04 AM

### DIFF
--- a/packages/ra-ui-materialui/src/layout/PageTitle.tsx
+++ b/packages/ra-ui-materialui/src/layout/PageTitle.tsx
@@ -1,13 +1,7 @@
 import * as React from 'react';
 import { useTranslate } from 'ra-core';
 
-export const PageTitle = ({
-    title,
-    defaultTitle,
-    className,
-    nonConfigurable,
-    ...rest
-}: any) => {
+export const PageTitle = ({ title, defaultTitle, className, ...rest }: any) => {
     const translate = useTranslate();
 
     return (
--- a/packages/ra-ui-materialui/src/layout/PageTitle.tsx
+++ b/packages/ra-ui-materialui/src/layout/PageTitle.tsx
@@ -1,19 +1,16 @@
 import * as React from 'react';
-import { useRecordContext, useTranslate, usePreference } from 'ra-core';
+import { useTranslate } from 'ra-core';
 
-export const PageTitle = ({ title, defaultTitle, className, ...rest }: any) => {
-    const [titleFromPreferences] = usePreference();
+export const PageTitle = ({
+    title,
+    defaultTitle,
+    className,
+    nonConfigurable,
+    ...rest
+}: any) => {
     const translate = useTranslate();
-    const record = useRecordContext();
 
-    return titleFromPreferences ? (
-        <span className={className} {...rest}>
-            {translate(titleFromPreferences, {
-                ...record,
-                _: titleFromPreferences,
-            })}
-        </span>
-    ) : (
+    return (
         <span className={className}>
             {!title ? (
                 <span {...rest}>{defaultTitle}</span>
--- a/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import { useLocation } from 'react-router-dom';
-import { usePreferenceInput } from 'ra-core';
+import {
+    usePreferenceInput,
+    usePreference,
+    useRecordContext,
+    useTranslate,
+} from 'ra-core';
 import { TextField } from '@mui/material';
 
 import { Configurable } from '../preferences';
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
                 },
             }}
         >
-            <PageTitle {...props} />
+            <PageTitleConfigurableInner {...props} />
         </Configurable>
+    );
+};
+
+const PageTitleConfigurableInner = ({ className, ...rest }: any) => {
+    const [titleFromPreferences] = usePreference();
+    const translate = useTranslate();
+    const record = useRecordContext();
+
+    return titleFromPreferences ? (
+        <span className={className} {...rest}>
+            {translate(titleFromPreferences, {
+                ...record,
+                _: titleFromPreferences,
+            })}
+        </span>
+    ) : (
+        <PageTitle className={className} {...rest} />
     );
 };
--- a/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
@@ -27,7 +27,12 @@ export const PageTitleEditor = () => {
     );
 };
 
-export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
+export const PageTitleConfigurable = ({
+    preferenceKey,
+    title,
+    defaultTitle,
+    ...props
+}) => {
     const { pathname } = useLocation();
     return (
         <Configurable
@@ -39,24 +44,30 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
                 },
             }}
         >
-            <PageTitleConfigurableInner {...props} />
+            <PageTitleConfigurableInner
+                title={title}
+                defaultTitle={defaultTitle}
+                {...props}
+            />
         </Configurable>
     );
 };
 
-const PageTitleConfigurableInner = props => {
+const PageTitleConfigurableInner = ({ title, defaultTitle, ...props }) => {
     const [titleFromPreferences] = usePreference();
     const translate = useTranslate();
     const record = useRecordContext();
 
     return titleFromPreferences ? (
-        <span className={props.className}>
+        <span className={props.className} {...props}>
             {translate(titleFromPreferences, {
                 ...record,
                 _: titleFromPreferences,
             })}
         </span>
     ) : (
-        <PageTitle {...props} />
+        <>
+            <PageTitle title={title} defaultTitle={defaultTitle} {...props} />
+        </>
     );
 };
--- a/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
@@ -44,19 +44,19 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
     );
 };
 
-const PageTitleConfigurableInner = ({ className, ...rest }: any) => {
+const PageTitleConfigurableInner = props => {
     const [titleFromPreferences] = usePreference();
     const translate = useTranslate();
     const record = useRecordContext();
 
     return titleFromPreferences ? (
-        <span className={className} {...rest}>
+        <span className={props.className}>
             {translate(titleFromPreferences, {
                 ...record,
                 _: titleFromPreferences,
             })}
         </span>
     ) : (
-        <PageTitle className={className} {...rest} />
+        <PageTitle {...props} />
     );
 };
--- a/packages/ra-ui-materialui/src/layout/Title.tsx
+++ b/packages/ra-ui-materialui/src/layout/Title.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { useEffect, useState } from 'react';
 import { ReactElement } from 'react';
 import { createPortal } from 'react-dom';
+import { useLocation } from 'react-router-dom';
 import { RaRecord, TitleComponent, warning } from 'ra-core';
 
 import { PageTitle } from './PageTitle';
@@ -9,6 +10,7 @@ import { PageTitleConfigurable } from './PageTitleConfigurable';
 
 export const Title = (props: TitleProps) => {
     const { defaultTitle, title, preferenceKey, ...rest } = props;
+    const { pathname } = useLocation();
     const [container, setContainer] = useState<HTMLElement | null>(() =>
         typeof document !== 'undefined'
             ? document.getElementById('react-admin-title')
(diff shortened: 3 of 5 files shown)


packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -28,3 +28,10 @@ export const Sx = () => (
         <Title title="Hello, world" />
     </TestMemoryRouter>
 );
+
+export const NonConfigurable = () => (
```
adguernier, line 32: **suggestion:** This behavior should be tested in the context a `<DatagridConfigurable>` component since that's the goal of this PR.

packages/ra-ui-materialui/src/layout/PageTitle.tsx
```diff
@@ -1,19 +1,16 @@
...
-import { useRecordContext, useTranslate, usePreference } from 'ra-core';
+import { useTranslate } from 'ra-core';
 
-export const PageTitle = ({ title, defaultTitle, className, ...rest }: any) => {
-    const [titleFromPreferences] = usePreference();
+export const PageTitle = ({
+    title,
+    defaultTitle,
+    className,
+    nonConfigurable,
```
adguernier, line 8: **suggestion:** `nonConfigurable` seems to be useless here

```suggestion
    
```

packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
```diff
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
                 },
             }}
         >
-            <PageTitle {...props} />
+            <PageTitleConfigurableInner {...props} />
         </Configurable>
     );
 };
+
+const PageTitleConfigurableInner = ({ className, ...rest }: any) => {
```
adguernier, line 47: **suggestion:** We should avoid `any` to type props. 

`className` seems to be not passed to this component (see `<PageTitleConfigurable>`). Yet, `className` is present in the `...rest` props. So you can remove `className` and avoid `any`.

```suggestion
const PageTitleConfigurableInner = ({ ...rest }) => {
```

packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
```diff
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
...
+    const [titleFromPreferences] = usePreference();
+    const translate = useTranslate();
+    const record = useRecordContext();
+
+    return titleFromPreferences ? (
+        <span className={className} {...rest}>
```
adguernier, line 53:
```suggestion
        <span {...rest}>
```

packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
```diff
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
...
+            {translate(titleFromPreferences, {
+                ...record,
+                _: titleFromPreferences,
+            })}
+        </span>
+    ) : (
+        <PageTitle className={className} {...rest} />
```
adguernier, line 60:
```suggestion
        <PageTitle {...rest} />
```

adguernier (changes requested):

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
adguernier: Sorry, it was not clear. The goal is to test that the title is no longer configurable in "configurable mode" (not necessarily in a `<DatagridConfigurable />` context) with the prop you added. 
To achieve this, you can take inspiration from `react-admin/packages/ra-ui-materialui/src/preferences/Configurable.stories.tsx`. You can create a new story in this file and run storybook with `make storybook` to see the result 🙂

packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
```diff
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
                 },
             }}
         >
-            <PageTitle {...props} />
+            <PageTitleConfigurableInner {...props} />
         </Configurable>
     );
 };
+
+const PageTitleConfigurableInner = props => {
```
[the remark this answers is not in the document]
markyao6275: The reason is actually more to do with the `<PageTitle />` component.

Previously, the `<PageTitle />` component always checks preferences and is only used in `<PageTitleConfigurable />`.

Now, `<PageTitle />` also needs to be used in a `nonConfigurable` context, so it cannot check preferences, which is why I removed that logic from `<PageTitle />` (I thought `PageTitle` was a good name for a non configurable title, so I repurposed it instead of creating a new component).

Now, the logic to check preferences (`const [titleFromPreferences] = usePreference();`) needs to live somewhere. It can't live directly in `PageTitleConfigurable` since I think it needs to live in a child of `Configurable`. Thus, I needed a child component that checks preferences and renders `<PageTitle />` and that's what `<PageTitleConfigurableInner />` is for.

Please let me know if that's clear or if there's a better way to do this.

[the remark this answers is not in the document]
fzaninotto: Super clear, thanks for the explanation.

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
fzaninotto: Agreed, the story must be in a Configurable context, and show that when switching to configurable mode, the title isn't editable.

markyao6275: Just pushed this change. Lmk if you see any issues :)

markyao6275:
@fzaninotto @adguernier I believe this is ready for review again. Thanks!

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -28,3 +32,21 @@ export const Sx = () => (
         <Title title="Hello, world" />
     </TestMemoryRouter>
 );
+
+export const NonConfigurable = ({ preferenceKey = 'title' }) => (
+    <TestMemoryRouter>
+        <PreferencesEditorContextProvider>
+            <Inspector />
+            <InspectorButton />
+            <Configurable
```
adguernier, line 41: **suggestion**: This story is misleading because we see that we can edit the title but we cannot save it. I think there is no need to wrap `<TitlePortal>` and `<Title>` in a `<Configurable>` component, since `<Title>` already does that.

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -4,10 +4,17 @@ import { ReactElement } from 'react';
...
+import { PageTitle } from './PageTitle';
 import { PageTitleConfigurable } from './PageTitleConfigurable';
 
 export const Title = (props: TitleProps) => {
-    const { defaultTitle, title, preferenceKey, ...rest } = props;
+    const {
+        defaultTitle,
+        title,
+        preferenceKey,
+        nonConfigurable,
```
adguernier, line 15: **suggestion**: I realize... instead of adding a new prop, we could check if `preferenceKey` is false and render `<PageTitle>` or `<PageTitleConfigurable>` accordingly

adguernier (changes requested):

adguernier:
Reviewed. Thanks for your work :+1:

markyao6275:
Thanks for the comments @adguernier! Should be ready for another review :)

adguernier (approved):

adguernier:
Aside from the last two comments, LGTM. Thanks for your work :muscle:

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -35,3 +38,16 @@ export const Sx = () => (
...
+
+export const NonConfigurable = () => (
+    <TestMemoryRouter>
+        <PreferencesEditorContextProvider>
+            <Inspector />
+            <InspectorButton />
+            <>
+                <TitlePortal variant="body1" />
+                <Title title="Hello, world" preferenceKey={undefined} />
+            </>
```
djhi, line 50:
```suggestion
            <TitlePortal variant="body1" />
            <Title title="Hello, world" preferenceKey={undefined} />
```

djhi (changes requested):
Thanks! I think the default `preferenceKey` in `Title` should be the same as in `PageTitleConfigurable` and you should check if `preferenceKey` is strictly equal to `false`.

markyao6275:
@djhi you mean I should check if `preferenceKey` strictly equal to `undefined` right?

Anyway I tried to address your feedback in the most recent commit so please let me know if that's what you had in mind.

djhi:
> @djhi you mean I should check if `preferenceKey` strictly equal to `undefined` right?

I mean that to have it work like it did before, we should have a default `preferenceKey` set and users have to explicitly disable it by setting it to `false`. Hence, you should check if it is strictly equal to `false` to render the non configurable version. This is coherent with other part of react-admin.

markyao6275:
@djhi just acted on your feedback

For 

> I mean that to have it work like it did before, we should have a default preferenceKey

I just passed through the `preferenceKey` from `<Title />` to `<PageTitleConfigurable />` so that I don't have to specify the default `${pathname}.title` in more than one place.

Let me know if there's anything else you'd like me to change.

markyao6275:
@djhi friendly bump here

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -2,12 +2,15 @@ import * as React from 'react';
...
 import { createPortal } from 'react-dom';
+import { useLocation } from 'react-router-dom';
 import { RaRecord, TitleComponent, warning } from 'ra-core';
 
+import { PageTitle } from './PageTitle';
 import { PageTitleConfigurable } from './PageTitleConfigurable';
 
 export const Title = (props: TitleProps) => {
     const { defaultTitle, title, preferenceKey, ...rest } = props;
+    const { pathname } = useLocation();
```
djhi, line 13: Do we still need this?

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
-import { TestMemoryRouter } from 'ra-core';
-import { PreferencesEditorContextProvider } from 'ra-core';
+import { PreferencesEditorContextProvider, TestMemoryRouter } from 'ra-core';
 
 import { TitlePortal } from './TitlePortal';
 import { Title } from './Title';
+import { InspectorButton } from '../preferences/InspectorButton';
+import { Inspector } from '../preferences/Inspector';
+import { Configurable } from '../preferences/Configurable';
```
djhi, line 8: This is not used, can you remove it?

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -1,9 +1,12 @@
...
+import { PageTitleEditor } from './PageTitleConfigurable';
```
djhi, line 9: Same

djhi (changes requested):

markyao6275:
@djhi addressed, thanks!

markyao6275:
@djhi friendly bump here

packages/ra-ui-materialui/src/layout/Title.tsx
```diff
@@ -31,25 +32,25 @@ export const Title = (props: TitleProps) => {
...
+    return <>{createPortal(pageTitle, container)}</>;
 };
 
 export interface TitleProps {
     className?: string;
     defaultTitle?: TitleComponent;
     record?: Partial<RaRecord>;
     title?: string | ReactElement;
-    preferenceKey?: string;
+    preferenceKey?: string | false;
```
slax57, line 55: It would be nice to update the [documentation](https://github.com/marmelab/react-admin/blob/next/docs/Title.md#preferencekey) to indicate that this feature is now possible.

slax57 (changes requested):
Other than that, it works fine, thanks! :slightly_smiling_face:

packages/ra-ui-materialui/src/layout/PageTitleConfigurable.tsx
```diff
@@ -34,7 +39,24 @@ export const PageTitleConfigurable = ({ preferenceKey, ...props }) => {
...
+    const [titleFromPreferences] = usePreference();
+    const translate = useTranslate();
+    const record = useRecordContext();
+
+    return titleFromPreferences ? (
+        <span className={props.className}>
```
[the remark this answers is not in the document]
markyao6275: Done, thanks for catching

markyao6275:
@slax57 addressed your comments, thanks!

packages/ra-ui-materialui/src/layout/TitlePortal.stories.tsx
```diff
@@ -35,3 +36,14 @@ export const Sx = () => (
...
+export const NonConfigurable = () => (
+    <TestMemoryRouter>
+        <PreferencesEditorContextProvider>
+            <Inspector />
+            <InspectorButton />
+            <TitlePortal variant="body1" />
+            <Title title="Hello, world" preferenceKey={false} />
+        </PreferencesEditorContextProvider>
+    </TestMemoryRouter>
+);
```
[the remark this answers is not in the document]
slax57: It was useful to test the template replacement abilities (e.g. with titles like `Hello %{foo}`), but we can live without it so don't bother with that for now.

slax57 (approved):

PR #9840, opened 2024-05-10
[Doc] Explain how to change page with `useNavigation`

erwanMarmelab:
## Problem

Sometimes, you’d like to pass a state when redirecting. However the `redirect` only accept a string.

## Solution

~Change the `redirect` and `RedirectToFunction` functions signature to accept a [`Location`](https://github.com/remix-run/history/blob/main/docs/api-reference.md#location) from `react-router`~
Document useNavigate

## Todo

~- [x] Accept the right props~
~- [x] Add a story~
- [x] Document ~it~ useNavigate more

### DIFF
--- a/docs/useRedirect.md
+++ b/docs/useRedirect.md
@@ -47,9 +47,7 @@ redirect(false);
 
 Note that `useRedirect` allows redirection to an absolute URL outside the current React app.
 
-## `useNavigate`
-
-For even more specific navigation, you can use the Hook [`useNavigate`](https://reactrouter.com/en/main/hooks/use-navigate) from `react-router-dom` as follows
+**Tip:** For even more specific navigation, you can use the [`useNavigate`](https://reactrouter.com/en/main/hooks/use-navigate) hook from `react-router-dom` as follows:
 
 ```jsx
 import { useNavigate } from 'react-router-dom';
--- a/docs/useRedirect.md
+++ b/docs/useRedirect.md
@@ -42,6 +42,7 @@ redirect((resource, id, data) => {
 // redirect to edit view with state data
 redirect('edit', 'posts', 1, {}, { record: { post_id: record.id } });
 // redirect to a history Location object
+// TODO: change here
 redirect({ pathname: '/some/path', search: '?query=string', hash: '#hash', state: null, key: 'my_key' });
 // do not redirect (resets the record form)
 redirect(false);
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/routing/useRedirect.ts
```diff
@@ -1,5 +1,6 @@
 import { useCallback } from 'react';
 import { useNavigate, To } from 'react-router-dom';
+import { Location } from 'history';
```
fzaninotto, line 3: Why don't you import the `To` type imported from 'react-router-dom' above? It would avoid the added dependency

packages/ra-core/src/routing/useRedirect.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const Home = () => {
+    const redirect = useRedirectRA();
+    return (
+        <>
+            <h1>Home</h1>
+            <ul>
+                <li>
+                    <button onClick={() => redirect('/dashboard')}>
+                        Dashboard
```
fzaninotto, line 20: I'd use types as labels for your buttons: Relative url, View name, View name with details, Location, Function

fzaninotto (changes requested):

packages/ra-core/src/routing/useRedirect.stories.tsx
erwanMarmelab: Nice idea, applied

fzaninotto:
No, I'm thinking: since we offer exactly the same API as usenavigate, why would isers choose to use useredirect ? we should better document usenavigate rather than modify our API to mimic it

packages/ra-core/src/routing/useRedirect.ts
```diff
@@ -70,14 +70,21 @@ export const useRedirect = () => {
...
+                    }),
+                    {
+                        state:
+                            // We force the scrollToTop except when navigating to a list
+                            // where this is already done by <RestoreScrollPosition> in <Resource>
+                            redirectTo === 'list'
+                                ? state
+                                : { _scrollToTop: true, ...state },
+                    }
+                );
```
fzaninotto, line 87: I think we shouldn't add this feature to useRedirect

fzaninotto (changes requested):

adguernier (changes requested):

docs/useRedirect.md
```diff
@@ -46,3 +46,28 @@ redirect(false);
 ```
 
 Note that `useRedirect` allows redirection to an absolute URL outside the current React app.
+
+## `useNavigate`
```
adguernier, line 50: **polish**: I'd put this section in a tip but it's a personal opinion.

adguernier (approved):

docs/useRedirect.md
erwanMarmelab: done

fzaninotto (approved):

PR #9842, opened 2024-05-12
[Demo] Add task management to CRM app

fzaninotto:
image

### DIFF
--- a/examples/crm/src/companies/CompanyShow.tsx
+++ b/examples/crm/src/companies/CompanyShow.tsx
@@ -152,7 +152,7 @@ const ContactsIterator = () => {
     const now = Date.now();
     return (
         <Box>
-            <List>
+            <List dense>
                 {contacts.map(contact => (
                     <RecordContextProvider key={contact.id} value={contact}>
                         <ListItem
@@ -217,7 +217,7 @@ const DealsIterator = () => {
     const now = Date.now();
     return (
         <Box>
-            <List>
+            <List dense>
                 {deals.map(deal => (
                     <ListItem
                         button
--- a/examples/crm/src/contacts/ContactAside.tsx
+++ b/examples/crm/src/contacts/ContactAside.tsx
@@ -6,13 +6,13 @@ import {
     ReferenceManyField,
     EditButton,
     ShowButton,
-    useListContext,
     ReferenceField,
     FunctionField,
     useRecordContext,
 } from 'react-admin';
-import { Box, Typography, Divider, List, ListItem } from '@mui/material';
+import { Box, Typography, Divider } from '@mui/material';
 import { TagsListEdit } from './TagsListEdit';
+import { TasksIterator } from '../tasks/TasksIterator';
 
 import { Contact, Sale } from '../types';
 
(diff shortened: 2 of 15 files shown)


packages/ra-data-fakerest/src/index.ts
```diff
@@ -134,7 +134,8 @@ export default (data, loggingEnabled = false): DataProvider => {
             return Promise.reject(error);
         }
         if (loggingEnabled) {
-            log(type, resource, params, response);
+            const { signal, ...paramsWithoutSignal } = params;
```
fzaninotto, line 137: I also fixed the log for fakerest - it used to show the `signal` parameter in every query, which is useless in this case.

examples/crm/src/dashboard/Dashboard.tsx
```diff
@@ -19,10 +20,10 @@ export const Dashboard = () => (
             <LatestNotes />
         </Grid>
         <Grid item xs={12} md={3}>
-            <HotContacts />
+            <TasksList />
         </Grid>
         <Grid item xs={12} md={3}>
-            <DealsPipeline />
```
fzaninotto, line 25: I removed the DealsPipeline from the home page as there is already a chart for that. I kept the component for when we offer a way to customize the home page.

examples/crm/src/companies/CompanyShow.tsx
```diff
@@ -152,7 +152,7 @@ const ContactsIterator = () => {
     const now = Date.now();
     return (
         <Box>
-            <List>
+            <List dense>
```
fzaninotto, line 155: I densified the interface a bit - this is a desktop app.

examples/crm/src/contacts/ContactAside.tsx
```diff
@@ -107,34 +108,14 @@ export const ContactAside = ({ link = 'edit' }: { link?: 'edit' | 'show' }) => {
...
                 <TagsListEdit />
             </Box>
-            <ReferenceManyField target="contact_id" reference="tasks">
-                <TasksIterator />
-            </ReferenceManyField>
-        </Box>
-    );
-};
-
-const TasksIterator = () => {
```
fzaninotto, line 117: Moved to its own file

examples/crm/src/dashboard/LatestNotes.tsx
```diff
@@ -72,7 +72,11 @@ export const LatestNotes = () => {
                             key={`${note.type}_${note.id}`}
                             sx={{ marginBottom: 2 }}
                         >
-                            <Typography variant="body2" color="textSecondary">
+                            <Typography
+                                variant="body2"
+                                color="textSecondary"
+                                component="div"
```
fzaninotto, line 78: this is to avoid a warning on ReferenceField, because the Loader uses a div, and this is forbidden inside a `<p>`

examples/crm/src/dashboard/TasksList.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Contact } from '../types';
+
+export const TasksList = () => {
+    const { identity } = useGetIdentity();
+
+    // get all the contacts for this sales
+    const { data: contacts, isPending: contactsLoading } = useGetList<Contact>(
+        'contacts',
+        {
+            pagination: { page: 1, perPage: 500 },
```
fzaninotto, line 23: Yes, that's a lot of contacts, but since we can't do a join there is no lighter way of fetching the tasks of the contacts followed by the current sales.

examples/crm/src/dataGenerator/tasks.ts
```diff
@@ -39,8 +39,11 @@ export const generateTasks = (db: Db) => {
             contact_id: contact.id,
             type: random.arrayElement(type),
             text: lorem.sentence(),
-            due_date: randomDate(new Date(contact.first_seen)),
-            sales_id: contact.sales_id,
```
fzaninotto, line 43: tasks are linked to a contact, this is enough

examples/crm/src/dataGenerator/tasks.ts
```diff
@@ -39,8 +39,11 @@ export const generateTasks = (db: Db) => {
...
+            due_date: randomDate(
+                random.boolean() ? new Date() : new Date(contact.first_seen),
+                new Date(Date.now() + 100 * 24 * 60 * 60 * 1000)
```
fzaninotto, line 44: We want some of the tasks in the fixtures to be in the future

examples/crm/src/tasks/AddTask.tsx
```diff
@@ -0,0 +1,140 @@
...
+                />
+            </Box>
+            <Dialog
+                open={open}
+                onClose={() => setOpen(false)}
+                aria-labelledby="form-dialog-title"
+                fullWidth
+                maxWidth="sm"
+            >
+                <form onSubmit={handleCreateTask}>
```
djhi, line 71: Why don't you use a `Form` here and our validation? And why not `CreateBase` instead of the hook?

fzaninotto: Seemed overkill. What more would it bring?

djhi: Consistency regarding how errors are displayed in the app?

fzaninotto: We're using the browser native validation, so we're consistent with other sites. Thats Jakob's Law of UX ;)

https://lawsofux.com/jakobs-law/

djhi: Don't we use react-admin validation for the other forms in this app ?

fzaninotto: You won, I fixed the code

djhi (approved):
Lovely :)

PR #9844, opened 2024-05-14
Explain additional steps needed to use useGetIdentity()

JesperWe:
Add some required usage details that would have saved me a couple of hours if I had known.

### DIFF
--- a/docs/useGetIdentity.md
+++ b/docs/useGetIdentity.md
@@ -46,7 +46,7 @@ const PostDetail = ({ id }) => {
 }
 ```
 
-**Tip**: This hook requires the component where it is used to be wrapped in a react-query `<QueryClientProvider>` with a default `QueryClient`. 
+**Tip**: When using this hook in your <Admin> component, it requires that component to be wrapped in a react-query `<QueryClientProvider>` with a default `QueryClient` in its parent component. 
 
 ## Refreshing The Identity
 
--- a/docs/useGetIdentity.md
+++ b/docs/useGetIdentity.md
@@ -23,7 +23,7 @@ Once loaded, the `data` object contains the following properties:
 const { id, fullName, avatar } = data;
 ```
 
-`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`.
+`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`. This means you will need to wrap the component where you want to use it in a react-query `<QueryClientProvider>` with a default `QueryClient`. To use `useGetIdentity()` in your `App.tsx` you wrap it in `index.tsx`.
 
 ## Usage
 
--- a/docs/useGetIdentity.md
+++ b/docs/useGetIdentity.md
@@ -23,7 +23,7 @@ Once loaded, the `data` object contains the following properties:
 const { id, fullName, avatar } = data;
 ```
 
-`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`. This means you will need to wrap the component where you want to use it in a react-query `<QueryClientProvider>` with a default `QueryClient`. To use `useGetIdentity()` in your `App.tsx` you wrap it in `index.tsx`.
+`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`.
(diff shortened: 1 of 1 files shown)


docs/useGetIdentity.md
```diff
@@ -23,7 +23,7 @@ Once loaded, the `data` object contains the following properties:
 const { id, fullName, avatar } = data;
 ```
 
-`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`.
+`useGetIdentity` uses [react-query's `useQuery` hook](https://tanstack.com/query/v3/docs/react/reference/useQuery) to call the `authProvider`. This means you will need to wrap the component where you want to use it in a react-query `<QueryClientProvider>` with a default `QueryClient`. To use `useGetIdentity()` in your `App.tsx` you wrap it in `index.tsx`.
```
fzaninotto, line 26: I don't understand this precision. Perhaps you could share an example of the code you had and the error it produced?

slax57, line 26: Thanks for your contribution.

I get your point, and I agree it's a good idea to document this gotcha.

However I disagree with the place and way you chose to do it:
- Using `useGetIdentity` outside the `<QueryClientProvider>` is rather uncommon. Most of the time, calling this hook will work with no extra step => This additional piece of information should rather be added as a **Tip**. Besides, it should be moved to the **Usage** section and not the **Syntax** section.
- ``To use `useGetIdentity()` in your `App.tsx` you wrap it in `index.tsx`.`` This sentence makes little sense to me, and relies on conventions regarding the naming of the files that are not standard. I'd recommend removing it. Mentioning that `useGetIdentity` needs to be used inside a `<QueryClientProvider>` is probably enough.

=> Can you please move the first part of your tip to the **Usage** section, and make it a **Tip**?

slax57 (changes requested):

docs/useGetIdentity.md
JesperWe: Adding this hook to an app that does not use QueryClientProvider makes the app crash. If you are not familiar with react-query the original text about the hook using is not helpful in resolving the problem.

docs/useGetIdentity.md
```diff
@@ -46,6 +46,8 @@ const PostDetail = ({ id }) => {
 }
 ```
 
+**Tip**: This hook requires the component where it is used to be wrapped in a react-query `<QueryClientProvider>` with a default `QueryClient`.
```
fzaninotto, line 49: I think this tip adds confusion more than it helps. In 90% of the cases, adding a QueryClientProvider isn't necessary. It's only useful when using the hook directly in `<Admin>`. Your tip should clarify that.

JesperWe: Better? I am just trying to help people (like myself) getting started with you system to not get stuck on this like I did.

slax57: > Better? I am just trying to help people (like myself) getting started with you system to not get stuck on this like I did.

We totally understand this and highly appreciate your will to contribute! :slightly_smiling_face: 

But on the other hand, we need to keep a good balance in our documentation between general usage advice and use-case specific tips. It's no easy task and requires many iterations, that's why we are here to provide guidance with the help of the experience we have acquired by writing docs for several years now.

Now, I still find your latest iteration a bit unclear. So I tried to rewrite it myself to suggest a version closer to what I have in mind. 

Here's what I've got so far (to be added to the **Usage** section):

> ### Usage Outside Of The `<Admin>` Component
> 
> `useGetIdentity` needs to be used inside a `<QueryClientProvider>`. It will be the case if you call it inside the `<Admin>` Component. However if you need to call it outside of it, you will have to wrap your your component in a `<QueryClientProvider>` and provide the same `queryClient` to it:
> 
> ```tsx
> import { Admin } from 'react-admin';
> import { QueryClient } from 'react-query';
> import { dataProvider } from './dataProvider';
> 
> const queryClient = new QueryClient();
> 
> const MyIdentity = () => {
>     const { data: identity, isLoading } = useGetIdentity();
>     if (isLoading) return <>Loading...</>;
>     return <div>{identity.fullName}</div>;
> }
> 
> const App = () => (
>     <>
>         <QueryClientProvider client={queryClient}>
>             <MyIdentity />
>         </QueryClientProvider>
>         <Admin queryClient={queryClient} dataProvider={dataProvider}>
>             ...
>         </Admin>
>     </>
> );
> ```

However, this raised a question: calling `useGetIdentity` outside `<Admin>` makes it impossible for the hook to have access to the `authProvider`. So this snippet can't work. How did you solve that issue?

Can you share the code you have so far?

Thanks.

fzaninotto:

PR #9848, opened 2024-05-14
Fix race condition between http error notification and server-side validation error notification

slax57:
### Problem

When validating server side, you can’t provide a global error message and react-admin displays a default non customizable one.

In some cases (depending on the React version) users could use the error's `message` field to render a custom message in the notification, but due to a race condition between the 2 notifications, it would not always work.

### Solution

In `useCreateController` and `useEditController`, check if there are validation errors and if we are in pessimistic mode, and don't notify in this case (as it will be done by `useNotifyIsFormInvalid`).

In `useNotifyIsFormInvalid` notify with the `root.serverError` message if available (otherwise fall back to 'ra.message.invalid_form' like currently).

### DIFF
--- a/docs/Validation.md
+++ b/docs/Validation.md
@@ -384,7 +384,7 @@ Server-side validation is supported out of the box for `pessimistic` mode only. 
 
 **Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop. The only exception is the `root.serverError` key, which can be used to define a global error message for the form.
 
-**Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key. However `root.serverError` must be a plain string.
+**Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key. However `root.serverError` does not accept translation objects.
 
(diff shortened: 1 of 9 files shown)


docs/Validation.md
```diff
@@ -367,21 +367,24 @@ const CustomerCreate = () => (
...
         }
     }
 }
 ```
 
-**Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop.
+**Tip**: The shape of the returned validation errors must match the form shape: each key needs to match a `source` prop. The only exception is the `root.serverError` key, which can be used to define a global error message for the form.
 
-**Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key.
+**Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key. However `root.serverError` must be a plain string.
```
djhi, line 387: This is confusing. is a translation string a plain string? As we pass the error to `notify`, translation is supported, isn't it?
```suggestion
**Tip**: The returned validation errors might have any validation format we support (simple strings, translation strings or translation objects with a `message` attribute) for each key. However `root.serverError` does not accept translation objects.
```

djhi (changes requested):
:muscle:

djhi (approved):

PR #9853, opened 2024-05-16
[Typescript] Fix compilation error in `MenuItemLink`, `ResettableTextField` and `InpectorButton` with latest `@types/react`

### LINKED ISSUE #9737: react admin error with Menu.Item missing properties
I got an error when using the component Menu.Item of react admin. It seems that all the properties are missing.
Therefore I can't build my app without adding all the 250+ properties of Menu.Item. I noticed that it works when using the component MenuItemLink.
You can reproduce it by using the documentation example of Menu.Item. I am using node version v21.6.1 and npm version 10.2.4

it does not build with the code:
```
// in src/MyMenu.js
import { Menu } from 'react-admin';

export const MyMenu = () => (
    <Menu>
        ...
        <Menu.Item to="/custom-route" primaryText="Miscellaneous" />
    </Menu>
);
```

* React-admin version: 4.16.13
* React version:18.2.0
* Stack trace: 
```
error TS2739: Type '{ to: string; primaryText: string; }' is missing the following properties from type 'Pick<MenuItemLinkProps, "children" | "replace" | "value" | "id" | "className" | "dense" | "title" | "defaultChecked" | "defaultValue" | "suppressContentEditableWarning" | ... 277 more ... | "tooltipProps">': onPointerEnterCapture, onPointerLeaveCapture
```
it seems related to this issue: https://github.com/marmelab/react-admin/issues/1196 that has already been resolved but adding @types/react-router and @types/react-router-dom did not solve the problem.

ilia-os:
Closes #9737

### Problem 
@types/react v18.2.66 update makes some of our components require extra props.

1. `MenuItemLink` - requires `onPointerEnterCapture`, `onPointerLeaveCapture`
2. `ResettableTextField` - requires `onPointerEnterCapture`, `onPointerLeaveCapture`
3. `InspectorButton` - requires `onPointerEnterCapture`, `onPointerLeaveCapture` and `placeholder`

Probably InspectorButton.placeholder issue was introduced even before v18.2.66

This breaks new apps bootstrapped from `create-react-admin` - you'll get typescript error if using these components

### Solution
Explicitly omit these otherwise junk props and allow to use the components without compilation error

### DIFF
--- a/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
+++ b/packages/ra-ui-materialui/src/input/ResettableTextField.tsx
@@ -203,7 +203,10 @@ interface Props {
 }
 
 export type ResettableTextFieldProps = Props &
-    Omit<TextFieldProps, 'onChange'> & {
+    Omit<
+        TextFieldProps,
+        'onChange' | 'onPointerEnterCapture' | 'onPointerLeaveCapture'
+    > & {
         onChange?: (eventOrValue: any) => void;
     };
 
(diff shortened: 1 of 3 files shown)


fzaninotto (approved):

slax57 (approved):
Thanks!

PR #9856, opened 2024-05-17
Upgrade to React 19

djhi:
We want to estimate the things that break in react-admin when using React 19 (currently in Release Candidate). Depending on the React 19 release date, we may even make it a requirement for react-admin v5. 

- [x] Make it build
- [x] Make it pass the tests
- [x] Fix storybook
~~- [ ] Fix storybook source addon~~


smeng9:
Can you clarify a bit about 
> Depending on the React 19 release date, we may even make it a requirement for react-admin v5.

Does that mean if react 19 is released before 5.0 stable version, then react 18 will be dropped from 5.0 support?

fzaninotto:
> Does that mean if react 19 is released before 5.0 stable version, then react 18 will be dropped from 5.0 support?

Yes. This allows us to use React 19 features without waiting for the next react-admin major (probably in 2 years from now).

fzaninotto:
We need to update this PR to allow react-admin to work both with React 18 and React 19.

PR #9857, opened 2024-05-17
Optimize CI

djhi:
Parallelize the CI steps to avoid rebuilding the project several times. This should reduce the total number of CI minutes consumed for each build, at the price of a slightly longer build (because we need to reinstall yarn dependencies).

Build time

| Before | After |
| -- | -- |
| 4m 35s | 5m 38s |

Total build time

| Before | After |
| -- | -- |
| 22m 08s | 16m 17s |

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,14 @@ jobs:
       run: make build
     - name: Type check simple example
       run: cd examples/simple && yarn type-check
+    - name: Zip packages build artifact
+      run: zip packages-build.zip -r packages/*/esm packages/*/lib node_modules/@react-admin/*
+    - name: Upload packages build artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: packages-build
+        path: packages-build.zip
+        retention-days: 1
 
   doc-check:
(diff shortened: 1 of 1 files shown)


erwanMarmelab (approved):

fzaninotto:
To be rebased now that https://github.com/marmelab/react-admin/pull/9829 is merged. Also, could you move the simple example build out of the typecheck step to reduce the build time?

.github/workflows/test.yml
```diff
@@ -71,6 +94,7 @@ jobs:
 
   e-commerce:
     runs-on: ubuntu-latest
+    needs: [typecheck]
```
djhi, line 97: Now depends on the `typecheck` job. The `build` step also run a typecheck on the e-commerce demo

.github/workflows/test.yml
```diff
@@ -54,6 +82,7 @@ jobs:
 
   e2e-test:
     runs-on: ubuntu-latest
+    needs: [typecheck]
```
djhi, line 85:
```suggestion
```

fzaninotto (approved):

PR #9858, opened 2024-05-18
[Doc] Improve doc for `<Autocomplete onCreate>` and similar props

fzaninotto:
## Problem

I had to enable choice creation in an AutocompleteInput. I followed the docs and I didn't manage to do it. I had to dig down the code.

## Solution

Improve choice creation documentation

Note: I know that there is still work to do for ReferenceArrayInput & co, but Rome wasn't built in a day and this is enough for a PR.


docs/ReferenceInput.md
```diff
@@ -166,32 +166,6 @@ You can filter the query used to populate the possible values. Use the `filter`
 
 **Note**: When users type a search term in the `<AutocompleteInput>`, this doesn't affect the `filter` prop. Check the [Customizing the filter query](#customizing-the-filter-query) section below for details on how that filter works.
 
-## `format`
```
fzaninotto, line 169: Moved below as " Transforming The Input Value" because the `format` prop doesn't actually exist in this component.

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -241,17 +241,40 @@ export const OptionTextElement = () => (
...
 
+const choicesForCreationSupport = [
+    { id: 1, name: 'Leo Tolstoy' },
+    { id: 2, name: 'Victor Hugo' },
+    { id: 3, name: 'William Shakespeare' },
+    { id: 4, name: 'Charles Baudelaire' },
+    { id: 5, name: 'Marcel Proust' },
+];
+
 export const OnCreate = () => (
```
fzaninotto, line 252: This story was broken since b6554db5b9e1516bd3b83cf4a26d2d1259c4c25c (my bad)

docs/img/SelectInput-onCreate.mp4
erwanMarmelab, line 1: **nitpick (non-blocking):** ‏
I'm not sure that keeping the black browser strip is useful for this screencast

erwanMarmelab (approved):

docs/img/SelectInput-onCreate.mp4
fzaninotto: If I don't, the alert is cut (as it is anchored to the URL input rather than the window top)

PR #9861, opened 2024-05-21
Fix `useAuthProvider` may return `undefined` when no `authProvider` is available

slax57:
Fixes BC introduced by https://github.com/marmelab/react-admin/pull/9760#discussion_r1605238420

### DIFF
--- a/packages/ra-core/src/auth/useLogout.ts
+++ b/packages/ra-core/src/auth/useLogout.ts
@@ -62,77 +62,81 @@ const useLogout = (): Logout => {
             params = {},
             redirectTo = loginUrl,
             redirectToCurrentLocationAfterLogin = true
-        ) =>
-            authProvider.logout(params).then(redirectToFromProvider => {
-                if (redirectToFromProvider === false || redirectTo === false) {
-                    resetStore();
-                    queryClient.clear();
-                    // do not redirect
-                    return;
-                }
+        ) => {
+            if (authProvider) {
+                return authProvider
+                    .logout(params)
+                    .then(redirectToFromProvider => {
+                        if (
+                            redirectToFromProvider === false ||
+                            redirectTo === false
+                        ) {
+                            resetStore();
+                            queryClient.clear();
+                            // do not redirect
+                            return;
+                        }
 
-                const finalRedirectTo = redirectToFromProvider || redirectTo;
+                        const finalRedirectTo =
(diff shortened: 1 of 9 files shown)


packages/ra-core/src/auth/useLogout.ts
```diff
@@ -62,77 +62,81 @@ const useLogout = (): Logout => {
...
-                }
-            );
-            resetStore();
-            queryClient.clear();
-            return Promise.resolve();
+                return Promise.resolve();
+            }
         },
-        [resetStore, location, navigate, loginUrl, queryClient]
+        [authProvider, resetStore, loginUrl, queryClient, location, navigate]
```
fzaninotto, line 136: This is problematic: `location` changes on every navigation, so the callback memoization is weak. You're using a locationRef for the sarch and hash, why not for the pathName?

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -46,76 +46,76 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
...
-                    const logoutUser = e?.logoutUser ?? true;
-                    //manual debounce
-                    if (timer) {
-                        // side effects already triggered in this tick, exit
-                        return true;
-                    }
-                    timer = setTimeout(() => {
-                        timer = undefined;
-                    }, 0);
+                ? authProvider
```
fzaninotto, line 49: I think this method is a bit too long for a legible ternary. I'd prefer a *leave early*.

fzaninotto (changes requested):
A couple nitpicks, but apart from that, 👍

fzaninotto (approved):

packages/ra-core/src/auth/useLogout.ts
slax57: I don't know why it was not done for `pathName` (I just moved the content of the second `useCallback`). But I assume we probably simply forgot to change this second callback to use refs while we changed the first callback, 2 years ago. Besides I can't think of any negative consequences this might have, so I used refs in this second part, too.
Nice catch!

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
slax57: changed

PR #9863, opened 2024-05-21
replaced "hotscript" types with typescript types and removed hotscript dependency

anmol-fzr:


packages/ra-core/src/util/useFieldValue.ts
```diff
@@ -18,30 +17,29 @@ import { useSourceContext } from '../core';
...
-    // FIXME: Find a way to throw a type error when defaultValue is not of RecordType[Source] type
-    defaultValue?: any;
-    source: Call<Objects.AllPaths, RecordType> extends never
-        ? AnyString
-        : Call<Objects.AllPaths, RecordType>;
-    record?: RecordType;
+  // FIXME: Find a way to throw a type error when defaultValue is not of RecordType[Source] type
+  defaultValue?: any;
+
+  source?: keyof RecordType extends never ? AnyString : keyof RecordType;
```
djhi, line 41: keyof is not enough to handle deep path fields and arrays. The issue suggested using `FieldPath` from react-hook-form

djhi (changes requested):

djhi:
This has been addressed in #10279. Thanks for the report and the PR :)

PR #9864, opened 2024-05-22
[Doc] Add Forms dedicated chapter

fzaninotto:
## Problem

I always struggle to find important doc about forms. Are there in the EditTutorial? In the SimpleForm chapter? In the Inputs doc? 

## Solution

Gather all the form-related doc into a dedicated chapter.


djhi (changes requested):
Nice :+1:

djhi (approved):

PR #9868, opened 2024-05-22
customising success message in delete with confirm

### LINKED ISSUE #9837: Customising the message that appears after the delete operation in DeleteWithConfirmButton
I want to customise the notification message that appears to the DeleteWithConfirmButton component after the delete operation is successful. (Only for some pages.) I can give the message using notify in the onSuccess method of the MutationOptions prop, but this time confirm is not turned off.

Is it possible to add a new prop to customise the notify message that appears after the delete operation is successful in the DeleteWithConfirmButton component?

```
{ onSuccessMessage =  'ra.notification.deleted' } = props;

onSuccess: () => {
  setOpen(false);
  notify(onSuccessMessage, {
    type: 'info',
    messageArgs: { smart_count: 1 },
    undoable: mutationMode === 'undoable',
  });
  unselect([record.id]);
  redirect(redirectTo, resource);
}
```

Or you can get props named handleSuccess, handleError and call them in the default onSuccess and onError methods in useDeleteWithConfirmController.

```
{ handleSuccess, handleError } = props;

onSuccess: () => {
  ...
 if (typeof handleSuccess === 'function') {
  handleSuccess();
 }
},
onError: () => {
  ...
 if (typeof handleError === 'function') {
  handleError();
 }
}

```

I can rewrite DeleteWithConfirmButton and useDeleteWithConfirmController components but it will be very costly for me.

**Additional context**
image

ethembynkr:
Closes #9837


slax57:
As is is a new feature, it needs to target the `next` branch. React Admin v4 is feature-freeze as we are in the verge of releasing React Admin v5. This feature will hence have to be released in v5.
Can you rebase onto `next`?

ethembynkr:
Hi @slax57. I changed target branch to next. Can you review?

ethembynkr:
I've added unit tests.

packages/ra-core/src/controller/button/useDeleteWithUndoController.spec.tsx
```diff
@@ -47,4 +48,54 @@ describe('useDeleteWithUndoController', () => {
...
+        fireEvent.click(button);
+
+        await waitFor(() => {
+            expect(notificationsSpy).toEqual([
+                {
+                    message: successMessage,
+                    type: 'info',
+                    notificationOptions: {
+                        messageArgs: { smart_count: 1 },
+                        undoable: false,
```
slax57, line 95: I was surprised this test was passing with `undoable: false` while I would have expected `undoable: true` since we are in undoable mode.

Then I realized there is a (pre-existing) bug in this file: we are not importing `useDeleteWithUndoController` from the right place!

```diff
 import useDeleteWithUndoController, {
-    UseDeleteWithConfirmControllerParams,
-} from './useDeleteWithConfirmController';
+    UseDeleteWithUndoControllerParams,
+} from './useDeleteWithUndoController';
```

Can you fix this import (and also the test) please? :pray: 

```suggestion
                        undoable: true,
```

slax57 (changes requested):
Other than that, the code looks good, thanks for the PR!

packages/ra-core/src/controller/button/useDeleteWithUndoController.spec.tsx
ethembynkr: I fixed it, thanks for the review.

slax57 (approved):
All good, thanks!
Sorry for taking so long to review... :grimacing:

PR #9875, opened 2024-05-23
Refactor middlewares to handle optimistic cases

djhi:
## Problem

By the time optimistic and undoable mutations dataProvider calls are really executed, the calling component might have been unmounted leading to the middlewares being unregistered and hence, not applied.

## Solution

When the mutation is called, keep the current callbacks in a closure

### DIFF
--- a/packages/ra-core/src/controller/saveContext/useMutationMiddlewares.ts
+++ b/packages/ra-core/src/controller/saveContext/useMutationMiddlewares.ts
@@ -51,12 +51,11 @@ export const useMutationMiddlewares = <
         []
     );
 
-    const mutateWithMiddlewares = useCallback(
-        (
-            fn: MutateFunc,
-            ...args: Parameters<MutateFunc>
-        ): ReturnType<MutateFunc> => {
-            let index = callbacks.current.length - 1;
+    const getMutateWithMiddlewares = useCallback((fn: MutateFunc) => {
+        // Stores the current callbacks in a closure to avoid losing them if the calling component is unmounted
+        const currentCallbacks = [...callbacks.current];
+        return (...args: Parameters<MutateFunc>): ReturnType<MutateFunc> => {
+            let index = currentCallbacks.length - 1;
 
             // Called by middlewares to call the next middleware function
             // Should take the same arguments as the original mutation function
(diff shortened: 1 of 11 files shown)


packages/ra-core/src/controller/saveContext/useMutationMiddlewares.ts
```diff
@@ -51,12 +51,11 @@ export const useMutationMiddlewares = <
...
 
-    const mutateWithMiddlewares = useCallback(
-        (
-            fn: MutateFunc,
-            ...args: Parameters<MutateFunc>
-        ): ReturnType<MutateFunc> => {
-            let index = callbacks.current.length - 1;
+    const getMutateWithMiddlewares = useCallback((fn: MutateFunc) => {
+        // Stores the current callbacks in a closure to avoid losing them if the calling component is unmounted
+        const currentCallbacks = [...callbacks.current];
```
slax57, line 56: Code looks good. However I'm wondering if we are still at risk of a regression regarding optimistic cases, since this part of the code (storing the callbacks in a closure) is not covered by automated tests.
Do you think we could cover that part too?

packages/ra-core/src/dataProvider/useCreate.stories.tsx
```diff
@@ -0,0 +1,307 @@
...
+        {
+            data: { title: 'Hello World' },
+        },
+        {
+            getMutateWithMiddlewares: mutate => async (resource, params) => {
+                return mutate(resource, {
+                    ...params,
+                    data: { title: `${params.data.title} from middleware` },
+                });
+            },
```
slax57, line 194: Also, these lines where you pass `getMutateWithMiddlewares` in stories throw TS errors (both `useCreate` and `useUpdate` stories).

`Type 'Promise<CreateResult<RaRecord<Identifier>>>' is not assignable to type 'ReturnType<CreateFunctionType>'.`

Can you fix the types or at least silence the warning if it is intended?

slax57 (changes requested):

packages/ra-core/src/controller/saveContext/useMutationMiddlewares.ts
slax57: Nice! :muscle:

slax57 (approved):
:muscle:

PR #9880, opened 2024-05-28
[v5] [Storybook] Fix some stories appear broken on devices which prefer dark mode

slax57:
The following stories appeared broken when rendered on a device which prefers dark mode:

- http://localhost:9010/?path=/story/ra-ui-materialui-auth-logout--user-authenticated : icon button appears missing
- http://localhost:9010/?path=/story/ra-ui-materialui-button-savebutton--basic : button is not visible at all
- http://localhost:9010/?path=/story/ra-ui-materialui-button-togglethemebutton--legacy : toggle theme button is rendered twice => I guess it's because we no longer support BC in v5, so I simply removed the story

Setting `defaultTheme="light"` fixes them.

### DIFF
--- a/packages/ra-ui-materialui/src/auth/Logout.stories.tsx
+++ b/packages/ra-ui-materialui/src/auth/Logout.stories.tsx
@@ -15,7 +15,7 @@ const MinimalAdmin = (props: { authenticated: boolean }) => {
         getPermissions: () => Promise.resolve(),
     };
     return (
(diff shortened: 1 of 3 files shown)


slax57:
> The problem is that, in dark mode, MUI components expect to be rendered on a dark background. Instead of forcing the theme, we should render these stories on a `Paper` to force a dark background in dark mode.

Fixed

PR #9881, opened 2024-05-29
[Doc] Update ra-audit-log documentation for v5

erwanMarmelab:
## Problem

`ra-audit-log` was updated to v5.0.0

## Solution

Update OSS doc


slax57 (approved):
nice catch!

PR #9883, opened 2024-05-29
docs: remove trailing slash on url

PaulieScanlon:
Remove trailing slash after `v1`

- ❌ `${process.env.SUPABASE_URL}/rest/v1/${requestUrl}`
- ✅ `${process.env.SUPABASE_URL}/rest/v1${requestUrl}`

If this is not removed the resulting URL contains a double slash before the endpoint, e.g

`https://abc123.supabase.co/rest/v1//users?offset=0&limit=10&order=id.asc`


fzaninotto:
Thanks!

PR #9891, opened 2024-06-01
fix useUnique.ts

alirezaCodeHub:
replace useRecordContext with useGetRecordId.
it's always has value

### DIFF
--- a/packages/ra-core/src/form/useUnique.ts
+++ b/packages/ra-core/src/form/useUnique.ts
@@ -1,12 +1,11 @@
 import merge from 'lodash/merge';
 import { useResourceContext } from '../core';
-import { useDataProvider } from '../dataProvider';
+import { useDataProvider, useGetRecordId } from '../dataProvider';
 import { useTranslate, useTranslateLabel } from '../i18n';
 import { InputProps } from './useInput';
 import { useCallback, useRef } from 'react';
 import set from 'lodash/set';
 import { asyncDebounce } from '../util';
-import { useRecordContext } from '../controller';
 import { isEmpty } from './validate';
 
 /**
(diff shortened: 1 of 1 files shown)


fzaninotto:
Your issue says "fix". What exactly are you fixing? Can you provide a repro and detailed steps for the bug?

alirezaCodeHub:
Assume that the _users_ resource has three fields username, name, email.

The username field cannot be changed and the email field is unique.

If creating an Edit page with current `useUnique()` for _users_ resource and try to edit name, validator of email raises error "Must be unique", while the email has not changed.

after debuging `useUnique()` source code, I found that the `useRecordContext()` return undefined. in the docs I found another way to get the current page id.

altough this version of `useUnique()` solve my problem in Edit page, but I think there is a problem in Create page, because Create page hasn't id.


https://github.com/marmelab/react-admin/assets/76510826/2e4b3e19-cf43-448d-9c3a-6bb145453e9a

fzaninotto:
I cannot reproduce the bug you describe. Please add a link to a CodeSandbox showing the issue.

alirezaCodeHub:
https://codesandbox.io/p/devbox/2jtvy2?migrateFrom=6s3fj4

slax57:
Thanks for providing the codesandbox.

Indeed, `useGetRecordId` improves the usability of `useUnique` for Edit pages, but it will throw an error if you try to use it in Create pages.
This is what makes the unit tests fail by the way.

This is probably the reason why the [documentation](https://marmelab.com/react-admin/useUnique.html#usage) requires to wrap the `<Form>` (but not the `<Edit>`) with `useUnique`:

```tsx
import { SimpleForm, TextInput, useUnique } from 'react-admin';

const UserCreateForm = () => {
    const unique = useUnique();
    return (
        <SimpleForm>
            <TextInput source="username" validate={unique()} />
        </SimpleForm>
    );
};
```

Hence, we will not merge this PR.
Thank you anyway for your contribution.

PR #9894, opened 2024-06-03
Fix `useDelete` doesn't delete record if its id is zero

adguernier:
## Problem

The `useDelete` hook does not delete the record if its id is zero. However, a record with 0 as identifier could exist.
Introduced by #9741 [here](https://github.com/marmelab/react-admin/pull/9741/files#diff-cd53d268c8805400093877effffadd54241b534266a80e1ab0bed1646cf51c59R208)

## Solution

Allow deletion of record with id 0

### DIFF
--- a/packages/ra-core/src/dataProvider/useDelete.ts
+++ b/packages/ra-core/src/dataProvider/useDelete.ts
@@ -204,7 +204,7 @@ export const useDelete = <
                     'useDelete mutation requires a non-empty resource'
                 );
             }
-            if (!callTimeId) {
+            if (callTimeId === undefined || callTimeId === null) {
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -204,7 +204,7 @@ export const useDelete = <
                     'useDelete mutation requires a non-empty resource'
                 );
             }
-            if (!callTimeId) {
+            if (callTimeId === undefined || callTimeId === null) {
```
fzaninotto, line 207: We use

```suggestion
            if (callTimeId != null) {
```

in other parts of the core (more concise).

Issue #9893, opened 2024-06-02
Filter with an object as defaultValue is still visible after user click the remove filter button


hmatthieu:
**What you were expecting:**
When a user clicks the remove filter button, the filter should not be active anymore and should be hidden from the UI.

**What happened instead:**
The filter is not active but is still visible in the UI.

**Steps to reproduce:**
Pass an object to the `defaultValue` prop of a `filters` List prop. In the react-admin simple example, you can replace in `examples/simple/src/posts/PostList.tsx:l56`
```jsx
<QuickFilter
    label="resources.posts.fields.commentable"
    source="commentable"
    defaultValue
/>
```
with
```jsx
<QuickFilter
    label="resources.posts.fields.commentable"
    source="commentable"
    defaultValue={{ commentEnabled: true, readOnly: false }}
/>
```
See this live example:
https://stackblitz.com/edit/github-kh8m7q?file=src%2Fposts%2FPostList.tsx

- Open the **Posts** page
- Press "Add a filter" to open the filters
- Press "Commentable" (The filter becomes activated and visible)
- Remove the "Commentable" filter (The filter becomes deactivated but is still visible)

**Related code:**
https://stackblitz.com/edit/github-kh8m7q?file=src%2Fposts%2FPostList.tsx

**Other information:**
At first, I thought it could be a design choice from React-Admin to not handle complex objects, but it's clear in the codebase that some logic is implemented to cover this use case. I would be happy to submit a merge request, but I want to first check if you agree that it's indeed a bug.

A quick note regarding why a complex object is beneficial: With a server that exposes database queries as an object, it's super easy to write filters that are passed from the UI to the database without any processing in between. [express-restify-mongoose](https://www.npmjs.com/package/express-restify-mongoose) is a good example, as well as GQL.

**How to fix**
The issue comes from `packages/ra-ui-materialui/src/list/filter/FilterForm.tsx:getFilterFormValues()`, and how react-admin rebuilds the filter (See codebase comments from l256 to l260).
When the user hides the filter, react-admin tries to reset the filter value, but this blank value is not compatible with how `getShownFilters:l119` chooses to display or not display a filter. This results in a misalignment between the filter state and the filter visibility.

To maintain the same behavior but allow a better reset behavior, the developer could provide a reset value, or react-admin might use the defaultFilterValue. There are different options to avoid breaking anything, and I'm happy to have your feedback on this.

**How to bypass**
[...]

PR #9898, opened 2024-06-04
Fix filters with complex object are not removed from the UI

hmatthieu:
When filters with a complex object as defaultValue the "empty value" was not considered empty by the UI. This commit ensure we don't display empty filters on the UI.

Refs: #9893

### DIFF
--- a/cypress/e2e/mobile.cy.js
+++ b/cypress/e2e/mobile.cy.js
@@ -9,10 +9,10 @@ describe('Mobile UI', () => {
     });
 
     describe('Infinite Scroll', () => {
-        it.only('should load more items when scrolling to the bottom of the page', () => {
+        it('should load more items when scrolling to the bottom of the page', () => {
             ListPagePosts.navigate();
+            cy.contains('Fusce massa lorem').should('exist');
             cy.contains('Sed quo et et fugiat modi').should('not.exist');
-            cy.scrollTo('bottom');
             cy.wait(500);
             cy.scrollTo('bottom');
             cy.contains('Sed quo et et fugiat modi');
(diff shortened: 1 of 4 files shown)


hmatthieu:
I reviewed the logic around the reset form behavior a bit. In my opinion, it could be simplified, but I don't want to risk removing code that seems overkill. The changes I made in this PR aim to reduce breaking changes and simply address the UI issue that was originally reported.

djhi:
Thanks for the PR! Would you mind adding a test for it?

hmatthieu:
Tests added

adguernier:
Thanks for your work :muscle: Waiting for another review

djhi (approved):

hmatthieu:
I'll review the failing test

hmatthieu:
I added a fix (to fix the test :x). I tested on my local setup, with a real chrome & firefox (managed by cypress). Was working properly. From my XP, those failing tests comes from CPU perfs and bad starting conditions for the test. I added a condition where we wait for an element to be present (which is safe than waiting for an element that is not present).

I also removed a `.only` that was forgotten there.

adguernier (approved):

cypress/e2e/mobile.cy.js
```diff
@@ -9,10 +9,10 @@ describe('Mobile UI', () => {
     });
 
     describe('Infinite Scroll', () => {
-        it.only('should load more items when scrolling to the bottom of the page', () => {
+        it('should load more items when scrolling to the bottom of the page', () => {
```
djhi, line 12: :scream:

PR #9901, opened 2024-06-05
Avoid double queries strict mode

### LINKED ISSUE #9896: V5 beta: duplicated queries on mount
**What you were expecting:**
`useGetOne` and `useGetList` only trigger one call to the API. 

**What happened instead:**
They call the API twice

**Steps to reproduce:**
1. Launch the simple example
2. Go to a comment edit page
3. The `getOne` and `getList` calls are doubled. 

Note that the duplication disappears when I remove StrictMode. However, the example is using React 18, so this isn't caused by [changes in strict mode in React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#strict-mode-improvements). When using react-admin v4 with React 18, the bug doesn't appear, so this a regression introduced by v5.

image

**Related code:**

https://stackblitz.com/github/marmelab/react-admin/tree/next/examples/simple

**Environment**

* React-admin version: 5.0.0-beta.3
* Last version that did not exhibit the issue (if applicable): 4.16
* React version: 18
* Browser: Chrome

djhi:
Fix #9896 

### Not enabled in development

- Run the demo
- Add a breakpoint in simple-rest data provider queries functions and check that `signal` is `undefined`

### Enabled in production

- Build the demo
- Run the demo `yarn preview` command
- Add a breakpoint in simple-rest data provider queries functions and check that `signal` is not `undefined`

### DIFF
--- a/docs/DataProviderWriting.md
+++ b/docs/DataProviderWriting.md
@@ -475,6 +475,15 @@ It's up to you to use this `meta` parameter in your data provider.
 
 All data provider queries can be called with an extra `signal` parameter. This parameter will receive an [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) that can be used to abort the request.
 
+To enable this feature, your data provider must have a `supportAbortSignal` property set to `true`. This is necessary to avoid queries to be sent twice in `development` mode when rendering your application inside [`<React.StrictMode>`](https://react.dev/reference/react/StrictMode).
+
+```tsx
+const dataProvider = simpleRestProvider('https://myapi.com');
+dataProvider.supportAbortSignal = true;
+// You can set this property depending on the production mode, e.g in Vite
+dataProvider.supportAbortSignal = import.meta.env.MODE === 'production';
+```
+
 When React Admin calls a data provider query method, it wraps it using [React Query](https://tanstack.com/query/v5/docs/react/overview), which supports automatic [Query Cancellation](https://tanstack.com/query/latest/docs/framework/react/guides/query-cancellation) thanks to the `signal` parameter.
 
 You can also benefit from this feature if you wrap your calls to the dataProvider with `useQuery`, and pass the `signal` parameter to the dataProvider:
--- a/examples/demo/src/dataProvider/index.ts
+++ b/examples/demo/src/dataProvider/index.ts
@@ -14,7 +14,7 @@ export default (type: string) => {
     // the login page. By the time they come back to the admin as a signed-in user,
     // the fake server will be initialized.
     const dataProviderWithGeneratedData = new Proxy(defaultDataProvider, {
-        get(target, name) {
+        get(_, name) {
             if (name === 'supportAbortSignal') {
                 return supportAbortSignal;
             }
--- a/examples/demo/src/dataProvider/index.ts
+++ b/examples/demo/src/dataProvider/index.ts
@@ -5,6 +5,7 @@ export default (type: string) => {
     // The fake servers require to generate data, which can take some time.
     // Here we start the server initialization but we don't wait for it to finish
     let dataProviderPromise = getDataProvider(type);
+    let supportAbortSignal: boolean | undefined = false;
 
     // Instead we return this proxy which may be called immediately by react-admin if the
     // user is already signed-in. In this case, we simply wait for the dataProvider promise
@@ -13,9 +14,13 @@ export default (type: string) => {
     // the login page. By the time they come back to the admin as a signed-in user,
     // the fake server will be initialized.
     const dataProviderWithGeneratedData = new Proxy(defaultDataProvider, {
-        get(_, name) {
+        get(target, name) {
+            if (name === 'supportAbortSignal') {
+                return supportAbortSignal;
+            }
             return (resource: string, params: any) => {
                 return dataProviderPromise.then(dataProvider => {
+                    supportAbortSignal = dataProvider.supportAbortSignal;
                     return dataProvider[name.toString()](resource, params);
                 });
             };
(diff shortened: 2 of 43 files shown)


docs/DataProviderWriting.md
```diff
@@ -475,6 +475,15 @@ It's up to you to use this `meta` parameter in your data provider.
...
+```tsx
+const dataProvider = simpleRestProvider('https://myapi.com');
+dataProvider.supportAbortSignal = true;
+// You can set this property depending on the production mode, e.g in Vite
+dataProvider.supportAbortSignal = import.meta.env.MODE === 'production';
+```
+
 When React Admin calls a data provider query method, it wraps it using [React Query](https://tanstack.com/query/v5/docs/react/overview), which supports automatic [Query Cancellation](https://tanstack.com/query/latest/docs/framework/react/guides/query-cancellation) thanks to the `signal` parameter.
 
 You can also benefit from this feature if you wrap your calls to the dataProvider with `useQuery`, and pass the `signal` parameter to the dataProvider:
```
djhi, line 489: Should we document how to leverage `supportAbortSignal` in custom query hooks?

packages/ra-core/src/controller/usePrevNextController.ts
```diff
@@ -190,8 +190,17 @@ export const usePrevNextController = <RecordType extends RaRecord = any>(
     // without displaying the list first
     const { data, error, isFetching, isLoading, isPending } = useQuery({
         queryKey: [resource, 'getList', params],
-        queryFn: ({ signal }) =>
-            dataProvider.getList(resource, { ...params, signal }),
+        queryFn: queryParams => {
+            const dataProviderParams: GetListParams = {
+                ...params,
+                signal: undefined,
```
fzaninotto, line 196: Why do you need that line? TypeScript?

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -84,11 +84,14 @@ export const useDataProvider = <
     const logoutIfAccessDenied = useLogoutIfAccessDenied();
 
     const dataProviderProxy = useMemo(() => {
-        return new Proxy(dataProvider, {
+        const proxy = new Proxy(dataProvider, {
```
fzaninotto, line 87: Not sure you need this change

packages/ra-core/src/dataProvider/useDataProvider.ts
```diff
@@ -84,11 +84,14 @@ export const useDataProvider = <
...
             get: (target, name) => {
                 if (typeof name === 'symbol' || name === 'then') {
                     return;
                 }
+                if (name === 'supportAbortSignal') {
+                    return target.supportAbortSignal;
```
fzaninotto, line 93: We use `'dataProvider` everywhere, why use `target` here?

packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -83,20 +83,26 @@ export const useGetList = <RecordType extends RaRecord = any>(
...
-                    meta,
-                    signal,
-                })
+        queryFn: queryParams => {
+            const dataProviderParams: GetListParams = {
+                pagination,
+                sort,
+                filter,
+                meta,
+                signal: undefined,
```
fzaninotto, line 92: **Suggestion**:  `signal: dataProvider.supportAbortSignal === true ? queryParams.signal : undefined`

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -101,24 +101,31 @@ export const useInfiniteGetList = <RecordType extends RaRecord = any>(
             'getInfiniteList',
             { pagination, sort, filter, meta },
         ],
-        queryFn: ({ pageParam = pagination.page, signal }) =>
```
fzaninotto, line 104: Not sure the changes are necessary here, cf https://github.com/TanStack/query/issues/3633#issuecomment-1133581032:

> with InfiniteQueries, we currently always abort fetches even if the AbortSignal is not consumed, which is likely a bug

packages/ra-data-simple-rest/src/index.ts
```diff
@@ -37,156 +37,169 @@ export default (
...
+                            'Content-Type': 'text/plain',
+                        }),
+                    })
+                )
+            ).then(responses => ({
+                data: responses.map(({ json }) => json.id),
+            })),
+    };
+
+    dataProvider.supportAbortSignal = process.env.NODE_ENV === 'production';
```
fzaninotto, line 202: I wouldn't enable it by default in production, otherwise there is  different behavior in dev and prod. Instead, I would document how to enable it.

fzaninotto (changes requested):
That was fast!

examples/demo/src/dataProvider/index.ts
```diff
@@ -13,9 +14,13 @@ export default (type: string) => {
     // the login page. By the time they come back to the admin as a signed-in user,
     // the fake server will be initialized.
     const dataProviderWithGeneratedData = new Proxy(defaultDataProvider, {
-        get(_, name) {
+        get(target, name) {
+            if (name === 'supportAbortSignal') {
+                return supportAbortSignal;
```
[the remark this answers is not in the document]
djhi: Yes but as we load the dataProvider asynchronously, I don't see any other way. Do you?

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
djhi: The *which is likely a bug* makes me think we should do it here just in case

examples/demo/src/dataProvider/index.ts
```diff
@@ -13,9 +14,13 @@ export default (type: string) => {
     // the login page. By the time they come back to the admin as a signed-in user,
     // the fake server will be initialized.
     const dataProviderWithGeneratedData = new Proxy(defaultDataProvider, {
-        get(_, name) {
+        get(target, name) {
```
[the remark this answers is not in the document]
djhi: Because we have `"no-unused-vars": "off"` in our config

fzaninotto (approved):

PR #9903, opened 2024-06-05
Fix types of Field components

slax57:
This PR fixes the following issues

1. `<ReferenceOneField>` throws a TS error if the `source` prop is absent (although it is not mandatory)
2. `<ReferenceOneField>` throws a TS error if the `queryOptions` prop does not contain a `queryKey` (although it should not be overridable)
3. Field components memoized with `genericMemo` do not have the correct type, and hence will throw a TS error when passing props (such as `source`) to them
4. Some stories and test files contained TS errors

### DIFF
--- a/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
@@ -37,7 +37,11 @@ describe('<ArrayField />', () => {
     it('should not fail for empty records', () => {
         render(
             <Wrapper>
-                <ArrayField source="arr" record={{ id: 123 }}>
+                <ArrayField
+                    // @ts-expect-error source prop does not have a valid value
+                    source="arr"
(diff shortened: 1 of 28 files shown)


fzaninotto:
Alternative: rewrite `genericMemo` as:

```jsx
export const genericMemo = <T>(component: T): T => {
    const result = memo(component as FunctionComponent);

    // We have to set the displayName on both the field implementation and the memoized version.
    // On the implementation so that the memoized version can pick them up and users may reference the defaultProps in their components.
    // On the memoized version so that components that inspect their children props may read them.
    // @ts-ignore
    result.displayName = component.displayName?.replace('Impl', '');
    return result as unknown as T;
};
```

slax57:
@fzaninotto Fixed

adguernier (approved):

fzaninotto (approved):

djhi (approved):

PR #9904, opened 2024-06-05
Make ErrorBoundary take fullHeight of the page

erwanMarmelab:
## Problem

As we can see in the story `react-admin`/`Admin`/`DefaultError`, the background of the error boundary doesn't take full height of the page.

### Before
___

image

---
### After
---

image

### DIFF
--- a/packages/ra-ui-materialui/src/AdminUI.tsx
+++ b/packages/ra-ui-materialui/src/AdminUI.tsx
@@ -22,7 +22,7 @@ export const AdminUI = ({
     error = Error,
     ...props
 }: AdminUIProps) => (
-    <ScopedCssBaseline enableColorScheme sx={{ height: '100vh' }}>
+    <ScopedCssBaseline enableColorScheme>
         <CoreAdminUI
             layout={layout}
             catchAll={catchAll}
--- a/packages/ra-ui-materialui/src/AdminUI.tsx
+++ b/packages/ra-ui-materialui/src/AdminUI.tsx
@@ -22,7 +22,7 @@ export const AdminUI = ({
     error = Error,
     ...props
 }: AdminUIProps) => (
-    <ScopedCssBaseline enableColorScheme>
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/AdminUI.tsx
```diff
@@ -22,7 +22,7 @@ export const AdminUI = ({
     error = Error,
     ...props
 }: AdminUIProps) => (
-    <ScopedCssBaseline enableColorScheme>
+    <ScopedCssBaseline enableColorScheme sx={{ height: '100vh' }}>
```
fzaninotto, line 25: This will break things. Can't you do it in the Error component instead?

erwanMarmelab: We can do it, yes. But in this case, the admin will be bigger than the view height as we can see with the scrollbar: 
image

adguernier, line 25: Given that `DefaultLayout` is already `100vh`, you could set `margin: 'auto` to the `Root` error component instead of changing the whole app height

In `Error.tsx`:
```ts
const Root = styled('div', {
    name: PREFIX,
})(({ theme }) => ({
    margin: 'auto', // <- this
    display: 'flex',
```
This won't work for the `Default Error` story because it doesn't have a parent with a full height. But it should work for others.

adguernier (changes requested):

adguernier (approved):

djhi (approved):

fzaninotto:
I don't see any change in Chrome

image

adguernier:
> I don't see any change in Chrome
> 
> image

This is because the `Default Error` story doesn't have a parent with a full height. But it should work for stories inside react-admin `<Layout>`.

Please see [my comment about it](https://github.com/marmelab/react-admin/pull/9904#discussion_r1629460441)

fzaninotto:
Reverted by #9912

PR #9908, opened 2024-06-07
Upgrade FakeRest to 4.0

fzaninotto:


test-setup.js
```diff
@@ -13,7 +13,11 @@ jest.spyOn(console, 'error').mockImplementation((...args) => {
  * Mock fetch objects Response, Request and Headers
  */
 const { Response, Headers, Request } = require('whatwg-fetch');
+// TextEncoder isn't bundled with JsDom, see https://stackoverflow.com/a/68468204/1333479
+const { TextEncoder, TextDecoder } = require('util');
```
fzaninotto, line 17: this is cumbersome - it will also bite our developers. We need to fix FakeRest for this

fzaninotto: The error comes from MSW, but ra-data-fakerest doesn't use it.

```
FAIL packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.spec.tsx
  ● Test suite failed to run

    ReferenceError: TextEncoder is not defined

    > 1 | import { Database } from 'fakerest';
        | ^
      2 | import { DataProvider } from 'ra-core';
      3 |
      4 | /* eslint-disable no-console */

      at node_modules/fakerest/node_modules/@mswjs/interceptors/lib/browser/chunk-UJZOJSMP.mjs:2:15
      at node_modules/fakerest/dist/fakerest.umd.cjs:2:66
      at Object.<anonymous> (node_modules/fakerest/dist/fakerest.umd.cjs:3:3)
      at Object.<anonymous> (packages/ra-data-fakerest/src/index.ts:1:1)
      at Object.<anonymous> (packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx:15:1)
      at Object.<anonymous> (packages/ra-ui-materialui/src/list/datagrid/ExpandAllButton.spec.tsx:4:1)
```

The fakerest lib doesn't allow tree-shaking.

fzaninotto: OK, fakerest does allow tree-shaking, but Jest doesn't know about tree-shaking. So we need FakeRest to extract adapters from the main package.

packages/ra-data-fakerest/src/index.ts
```diff
@@ -33,12 +43,11 @@ function log(type, resource, params, response) {
  *   ],
  * })
  */
-export default (data, loggingEnabled = false): DataProvider => {
-    const restServer = new FakeRest.Server();
-    restServer.init(data);
+export default (data, loggingEnabled = false, delay?: number): DataProvider => {
```
djhi, line 46: Is it time to accept an object of options?

djhi (approved):
:+1:

PR #9911, opened 2024-06-10
`<ArrayInput>` uses `SourceContext` instead of cloning children

adguernier:
## Problem

`<ArrayInput>` and `<SimpleFormIterator>` are still cloning their children. Besides, `<SimpleFormIterator>` throws an Error when used without a source prop.

## Solution

Migrate both components to support the `SourceContext` and no longer require cloning their children.

This is a BC so this needs to be done before the v5 release.

### DIFF
--- a/packages/ra-core/src/i18n/useTranslatable.spec.ts
+++ b/packages/ra-core/src/i18n/useTranslatable.spec.ts
@@ -1,0 +1,40 @@
+import { getRecordForLocale } from './useTranslatable';
+describe('useTranslatable', () => {
+    describe('getRecordForLocale', () => {
+        it('should return a record where translatable fields have their values set to the value of the given locale', () => {
+            // Given the record { title: { en: 'title_en', fr: 'title_fr' } } and the locale 'fr',
+            // the record for the locale 'fr' will be { title: 'title_fr' }
+            const record = {
+                fractal: true,
+                title: { en: 'title_en', fr: 'title_fr' },
+                items: [
+                    { description: { en: 'item1_en', fr: 'item1_fr' } },
+                    { description: { en: 'item2_en', fr: 'item2_fr' } },
+                ],
+            };
+
+            const recordForLocale = getRecordForLocale(record, 'fr');
+
+            expect(recordForLocale).toEqual({
+                fractal: true,
+                title: 'title_fr',
+                items: [
+                    { description: 'item1_fr' },
+                    { description: 'item2_fr' },
+                ],
+            });
+        });
+
+        it('should return the record as is if it is empty', () => {
+            const record = {};
+            const recordForLocale = getRecordForLocale(record, 'fr');
+            expect(recordForLocale).toEqual({});
+        });
+
+        it('should return the record as is if it is undefined', () => {
+            const record = undefined;
+            const recordForLocale = getRecordForLocale(record, 'fr');
+            expect(recordForLocale).toEqual(undefined);
+        });
+    });
+});
--- a/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
+++ b/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
@@ -75,10 +75,10 @@ export const TranslatableFieldsTabContent = (
             className={className}
             {...other}
         >
-            {Children.map(children, field =>
-                field && isValidElement<any>(field) ? (
-                    <RecordContextProvider value={recordForLocale}>
-                        <SourceContextProvider value={sourceContext}>
+            <RecordContextProvider value={recordForLocale}>
+                <SourceContextProvider value={sourceContext}>
+                    {Children.map(children, field =>
+                        field && isValidElement<any>(field) ? (
                             <div>
                                 {addLabel ? (
                                     <Labeled
@@ -94,10 +94,10 @@ export const TranslatableFieldsTabContent = (
                                     field
                                 )}
                             </div>
-                        </SourceContextProvider>
-                    </RecordContextProvider>
-                ) : null
-            )}
+                        ) : null
+                    )}
+                </SourceContextProvider>
+            </RecordContextProvider>
         </Root>
     );
 };
--- a/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
+++ b/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
@@ -53,6 +53,10 @@ export const TranslatableFieldsTabContent = (
     );
     const addLabel = Children.count(children) > 1;
 
+    // As fields rely on the RecordContext to get their values and have no knowledge of the locale,
+    // we need to create a new record with the values for the current locale only
+    // Given the record { title: { en: 'title_en', fr: 'title_fr' } } and the locale 'fr',
+    // the record for the locale 'fr' will be { title: 'title_fr' }
     const [recordForLocale] = useState(() =>
         getRecordForLocale(record, locale)
     );
--- a/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
+++ b/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
@@ -5,7 +5,7 @@ import {
     useTranslatableContext,
     RaRecord,
     SourceContextProvider,
-    useSourceContext,
+    useOptionalSourceContext,
     getResourceFieldLabelKey,
     RecordContextProvider,
     ResourceContextProvider,
@@ -30,7 +30,7 @@ export const TranslatableFieldsTabContent = (
     } = props;
     const { selectedLocale } = useTranslatableContext();
 
-    const parentSourceContext = useSourceContext();
+    const parentSourceContext = useOptionalSourceContext();
     const sourceContext = React.useMemo(
         () => ({
             getSource: (source: string) =>
--- a/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
+++ b/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
@@ -1,6 +1,14 @@
 import * as React from 'react';
+import {
+    Children,
+    isValidElement,
+    ReactElement,
+    ReactNode,
+    useState,
+} from 'react';
 import { styled } from '@mui/material/styles';
-import { Children, isValidElement, ReactElement, ReactNode } from 'react';
+import set from 'lodash/set';
+import get from 'lodash/get';
 import {
     useTranslatableContext,
     RaRecord,
@@ -8,7 +16,6 @@ import {
     useOptionalSourceContext,
     getResourceFieldLabelKey,
     RecordContextProvider,
-    ResourceContextProvider,
 } from 'ra-core';
 import { Labeled } from '../Labeled';
 
@@ -46,6 +53,10 @@ export const TranslatableFieldsTabContent = (
     );
     const addLabel = Children.count(children) > 1;
 
+    const [recordForLocale] = useState(() =>
+        getRecordForLocale(record, locale)
+    );
+
     return (
         <Root
             role="tabpanel"
@@ -57,28 +68,23 @@ export const TranslatableFieldsTabContent = (
         >
             {Children.map(children, field =>
                 field && isValidElement<any>(field) ? (
-                    <ResourceContextProvider
-                        value={resource}
-                        key={field.props.source}
-                    >
-                        <RecordContextProvider value={record}>
-                            <SourceContextProvider value={sourceContext}>
-                                <div>
-                                    {addLabel ? (
-                                        <Labeled
-                                            resource={resource}
-                                            label={field.props.label}
-                                            source={field.props.source}
-                                        >
-                                            {field}
-                                        </Labeled>
-                                    ) : (
-                                        field
-                                    )}
-                                </div>
-                            </SourceContextProvider>
-                        </RecordContextProvider>
-                    </ResourceContextProvider>
+                    <RecordContextProvider value={recordForLocale}>
+                        <SourceContextProvider value={sourceContext}>
+                            <div>
+                                {addLabel ? (
+                                    <Labeled
+                                        resource={resource}
+                                        label={field.props.label}
+                                        source={field.props.source}
+                                    >
+                                        {field}
+                                    </Labeled>
+                                ) : (
+                                    field
+                                )}
+                            </div>
+                        </SourceContextProvider>
+                    </RecordContextProvider>
                 ) : null
             )}
         </Root>
@@ -109,3 +115,31 @@ const Root = styled('div', {
     border: `1px solid ${theme.palette.divider}`,
     borderTop: 0,
 }));
+
+const getRecordForLocale = (record: RaRecord, locale: string) => {
+    // Get all paths of the record
+    const paths = getRecordPaths(record);
+
+    // For each path, if a path ends with the locale, set the value of the path without the locale
+    // to the value of the path with the locale
+    const recordForLocale = paths.reduce((acc, path) => {
+        if (path.includes(locale)) {
+            const pathWithoutLocale = path.slice(0, -1);
+            const value = get(record, path);
+            return set(acc, pathWithoutLocale, value);
+        }
+        return acc;
+    }, structuredClone(record));
+
+    return recordForLocale;
+};
+
+const getRecordPaths = (
+    record: any = {},
+    path: Array<string> = []
+): Array<Array<string>> =>
+    Array.isArray(record) || Object(record) !== record
+        ? [path]
+        : Object.entries(record).flatMap(([k, v]) =>
+              getRecordPaths(v, [...path, k])
+          );
--- a/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
+++ b/packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
@@ -50,13 +50,10 @@ export const TranslatableFieldsTabContent = (
                 parentSourceContext
                     ? parentSourceContext.getSource(`${source}.${locale}`)
                     : `${source}.${locale}`,
-            getLabel: (source: string) => {
-                const label = parentSourceContext
+            getLabel: (source: string) =>
+                parentSourceContext
                     ? parentSourceContext.getLabel(source)
-                    : getResourceFieldLabelKey(resource, source);
-                console.log(source, label);
-                return label;
-            },
+                    : getResourceFieldLabelKey(resource, source),
         }),
         [locale, parentSourceContext, resource]
     );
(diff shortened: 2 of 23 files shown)


packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -177,15 +198,13 @@ export const ArrayInput = (props: ArrayInputProps) => {
...
-                    record,
-                    resource: resourceFromProps,
-                    source,
-                    variant,
-                    margin,
-                    disabled,
-                })}
+                <ResourceContextProvider value={resourceFromProps}>
+                    <SourceContextProvider value={sourceContext}>
+                        <RecordContextProvider value={record}>
```
fzaninotto, line 203: Use `OptionalRecordContextProvider` instead

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -141,6 +147,21 @@ export const ArrayInput = (props: ArrayInputProps) => {
...
 
+    const sourceContext = React.useMemo<SourceContextValue>(
+        () => ({
+            getSource: (source: string) => {
+                return source ? `${finalSource}.${source}` : finalSource;
+            },
+            getLabel: (source: string) =>
+                parentSourceContext
+                    ? parentSourceContext.getLabel(
+                          `resources.${resourceFromProps}.fields.${source}`
```
fzaninotto, line 158: This is mysterious to me. Would you mind adding a comment to give an example expected label? 

Because to me, inside another ArrayInput, this will lead to labels like `resources.A.fields.B.0.resources.A.fields.C` instead of `resources.A.fields.B.0.fields.C`

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -92,6 +97,7 @@ export const ArrayInput = (props: ArrayInputProps) => {
     const formGroupName = useFormGroupContext();
     const formGroups = useFormGroups();
     const finalSource = useWrappedSource(source);
+    const parentSourceContext = useSourceContext();
```
fzaninotto, line 100: I find it strange to call both `useWrappedSource` and `useSourceContext` in the same file, as the first one calls the second one.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -141,6 +147,21 @@ export const ArrayInput = (props: ArrayInputProps) => {
...
+        () => ({
+            getSource: (source: string) => {
+                return source ? `${finalSource}.${source}` : finalSource;
+            },
+            getLabel: (source: string) =>
+                parentSourceContext
+                    ? parentSourceContext.getLabel(
+                          `resources.${resourceFromProps}.fields.${source}`
+                      )
+                    : `resources.${resourceFromProps}.fields.${source}`,
```
fzaninotto, line 160: use the `resource` from `ResourceContext` instead of `resourceFromProps`

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -141,6 +147,21 @@ export const ArrayInput = (props: ArrayInputProps) => {
 
     const { error } = getFieldState(finalSource, formState);
 
+    const sourceContext = React.useMemo<SourceContextValue>(
+        () => ({
+            getSource: (source: string) => {
+                return source ? `${finalSource}.${source}` : finalSource;
```
fzaninotto, line 153: This I don't understand. In other instances, we check if there is a parent source context, then call that parent's getSource, otherwise return the source. Here you're doing the exact opposite...

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -165,14 +172,14 @@ export const SimpleFormIterator = (inProps: SimpleFormIteratorProps) => {
...
-                            member={`${source}.${index}`}
+                            member={`${index}`}
                             onRemoveField={removeField}
                             onReorder={handleReorder}
                             record={(records && records[index]) || {}}
                             removeButton={removeButton}
                             reOrderButtons={reOrderButtons}
                             resource={resource}
-                            source={source}
+                            source={finalSource}
```
fzaninotto, line 182: We should stop passing the source altogether, and let the item read the sourceContext instead

fzaninotto (changes requested):
I'm confused by this PR. It probably requires more tests and more comments to make the code robust and maintainable.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -46,6 +53,14 @@ export const TranslatableFieldsTabContent = (
     );
     const addLabel = Children.count(children) > 1;
 
+    // As fields rely on the RecordContext to get their values and have no knowledge of the locale,
+    // we need to create a new record with the values for the current locale only
+    // Given the record { title: { en: 'title_en', fr: 'title_fr' } } and the locale 'fr',
+    // the record for the locale 'fr' will be { title: 'title_fr' }
+    const [recordForLocale] = useState(() =>
```
fzaninotto, line 60: I don't understand why you use useState here.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -109,3 +119,31 @@ const Root = styled('div', {
...
+            const value = get(record, path);
+            return set(acc, pathWithoutLocale, value);
+        }
+        return acc;
+    }, structuredClone(record));
+
+    return recordForLocale;
+};
+
+const getRecordPaths = (
```
fzaninotto, line 141: Please add a comment to explain the output for a given input

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -30,7 +37,7 @@ export const TranslatableFieldsTabContent = (
     } = props;
     const { selectedLocale } = useTranslatableContext();
 
-    const parentSourceContext = useSourceContext();
+    const parentSourceContext = useOptionalSourceContext();
     const sourceContext = React.useMemo(
         () => ({
             getSource: (source: string) =>
```
fzaninotto, line 43: We shouldn't do anything special here. The sourceContext is only useful for the label.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -109,3 +119,31 @@ const Root = styled('div', {
...
 }));
+
+const getRecordForLocale = (record: RaRecord, locale: string) => {
+    // Get all paths of the record
+    const paths = getRecordPaths(record);
+
+    // For each path, if a path ends with the locale, set the value of the path without the locale
+    // to the value of the path with the locale
+    const recordForLocale = paths.reduce((acc, path) => {
+        if (path.includes(locale)) {
```
fzaninotto, line 130: This is too weak IMO, as it will match 'francois' for the 'fr' locale. I'd use a RegExp, testing for entries like 'XXX.[locale]' only.

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -56,9 +56,15 @@ export const SimpleFormIterator = (inProps: SimpleFormIteratorProps) => {
         fullWidth,
         sx,
     } = props;
-    if (!source) {
-        throw new Error('SimpleFormIterator requires a source prop');
+
+    const sourceContext = useSourceContext();
+    const finalSource = sourceContext?.getSource('');
```
fzaninotto, line 61:
```suggestion
    const finalSource = sourceContext.getSource('');
```

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
```diff
@@ -17,36 +17,13 @@ import { ArrayInput } from './ArrayInput';
 import { SimpleFormIterator } from './SimpleFormIterator';
 import { useFormContext } from 'react-hook-form';
 import { GlobalValidation, ValidationInFormTab } from './ArrayInput.stories';
+import { useArrayInput } from './useArrayInput';
 
 describe('<ArrayInput />', () => {
-    it('should pass its record props to its child', async () => {
```
fzaninotto, line 22: That's a problem that it doesn't work. Without a RecordContext, we can't use a field as child of ArrayInput.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
```diff
@@ -156,36 +199,32 @@ export const ArrayInput = (props: ArrayInputProps) => {
...
-                `ra-input-${source}`,
+                `ra-input-${finalSource}`,
                 ArrayInputClasses.root,
                 className
             )}
             error={!!error}
             {...sanitizeInputRestProps(rest)}
         >
             <InputLabel
-                htmlFor={finalSource}
```
fzaninotto, line 167: why did you remove that?

test-setup.js
```diff
@@ -17,3 +17,8 @@ const { Response, Headers, Request } = require('whatwg-fetch');
 global.Response = Response;
 global.Headers = Headers;
 global.Request = Request;
+
+// jsdom does not support structuredClone
```
fzaninotto, line 21: All our users will have to do that. Please use a shim instead.

fzaninotto (changes requested):

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
djhi: `path` is an array, not a string

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.spec.tsx
djhi: ArrayInput does not set the RecordContext, SimpleFormIteratorItem does

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.tsx
djhi: Because you can't label a non focusable element

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -57,28 +53,21 @@ export const TranslatableFieldsTabContent = (
         >
             {Children.map(children, field =>
                 field && isValidElement<any>(field) ? (
-                    <ResourceContextProvider
-                        value={resource}
-                        key={field.props.source}
-                    >
-                        <RecordContextProvider value={record}>
-                            <SourceContextProvider value={sourceContext}>
```
fzaninotto, line 65: We still need a SourceContextProvider to have the label translated

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -28,23 +30,17 @@ export const TranslatableFieldsTabContent = (
...
+    const addLabel = Children.count(children) > 1;
 
-    const parentSourceContext = useSourceContext();
-    const sourceContext = React.useMemo(
-        () => ({
-            getSource: (source: string) =>
-                parentSourceContext
-                    ? parentSourceContext.getSource(`${source}.${locale}`)
-                    : `${source}.${locale}`,
-            getLabel: (source: string) =>
```
fzaninotto, line 40: I think this is still required

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -56,9 +56,14 @@ export const SimpleFormIterator = (inProps: SimpleFormIteratorProps) => {
         fullWidth,
         sx,
     } = props;
-    if (!source) {
-        throw new Error('SimpleFormIterator requires a source prop');
+
+    const finalSource = useWrappedSource('');
+    if (!finalSource) {
+        throw new Error(
+            'SimpleFormIterator should be wrapped in a SourceContext'
```
fzaninotto, line 63:
```suggestion
            'SimpleFormIterator can only be called within an iterator input like ArrayInput'
```

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
```diff
@@ -72,27 +71,38 @@ export const SimpleFormIteratorItem = React.forwardRef(
...
+                        // </ArrayInput>
                         return parentSourceContext.getSource(
-                            source ? `${member}.${source}` : member
+                            `${index}.${source}`
                         );
                     }
-                    return source ? `${member}.${source}` : member;
                 },
                 getLabel: (source: string) => {
-                    // remove digits, e.g. 'book.authors.2.categories.3.identifier.name' => 'book.authors.categories.identifier.name'
```
fzaninotto, line 83: I think we should keep this comment

fzaninotto (changes requested):
I think we miss a story where we can see non-humanized input labels, to check that they make sense.

test-setup.js
fzaninotto: We should not use structuredClone.

packages/ra-core/src/i18n/useTranslatable.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+                items: [
+                    { description: { en: 'item1_en', fr: 'item1_fr' } },
+                    { description: { en: 'item2_en', fr: 'item2_fr' } },
+                ],
+            };
+
+            const recordForLocale = getRecordForLocale(record, 'fr');
+
+            expect(recordForLocale).toEqual({
+                fractal: true,
```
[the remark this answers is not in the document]
djhi: It includes the `fr` letter to make sure we don't mistake it for the `fr` locale. See https://github.com/marmelab/react-admin/pull/9911#discussion_r1634873152

packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIteratorItem.tsx
djhi: We have a way better one now and you wrote it ;)

packages/ra-core/src/i18n/useTranslatable.spec.ts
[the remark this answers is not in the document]
fzaninotto: Then your test proves that the algorith is wrong: `fractal` shouldn't be included in the record for the `fr` locale.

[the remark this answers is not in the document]
djhi: No, that's actually what I expect, have the record returned with its translatable properties fixed for the locale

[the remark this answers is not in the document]
djhi: If for some reason, you have a `FormDataConsumer`, you'll still get the full record

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
```diff
@@ -392,11 +452,113 @@ export const NestedInline = () => (
...
+                                                sx={{ width: 200 }}
+                                            />
+                                        </Labeled>
+                                    </TranslatableInputs>
+                                    <TranslatableFields locales={['en', 'fr']}>
+                                        <TextField
+                                            source="name"
+                                            sx={{ width: 200 }}
+                                        />
+                                        <TextField
```
fzaninotto, line 523: why is it doubled?

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
```diff
@@ -392,11 +452,113 @@ export const NestedInline = () => (
...
+                                console.log(data);
+                            },
+                        }}
+                    >
+                        <SimpleForm>
+                            <TextInput source="customer" helperText={false} />
+                            <DateInput source="date" helperText={false} />
+                            <ArrayInput source="items">
+                                <SimpleFormIterator
+                                    inline
```
fzaninotto, line 503: Please avoid inline, this makes inputs smaller and truncates the labels

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -57,28 +77,25 @@ export const TranslatableFieldsTabContent = (
...
-                                            {field}
-                                        </Labeled>
-                                    ) : (
-                                        field
-                                    )}
-                                </div>
-                            </SourceContextProvider>
-                        </RecordContextProvider>
-                    </ResourceContextProvider>
+                    <RecordContextProvider value={recordForLocale}>
```
fzaninotto, line 80: missing key. Also, it's the same record for all the fields, why do we create the RecordContext inside the Children.map? It should be outside.

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
```diff
@@ -24,14 +30,21 @@ export const TranslatableFieldsTabContent = (
...
         ...other
     } = props;
-    const { selectedLocale } = useTranslatableContext();
+    const { selectedLocale, getRecordForLocale } = useTranslatableContext();
+    const addLabel = Children.count(children) > 1;
 
-    const parentSourceContext = useSourceContext();
-    const sourceContext = React.useMemo(
+    const parentSourceContext = useOptionalSourceContext();
+    const resource = useResourceContext(props);
```
fzaninotto, line 41: unused?

fzaninotto (changes requested):
Getting close! 🔥

packages/ra-core/src/i18n/useTranslatable.spec.ts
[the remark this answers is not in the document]
fzaninotto: OK, gotcha

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
djhi: Because when there's only one child, the `TranslatableFields` does not add labels

packages/ra-ui-materialui/src/field/TranslatableFieldsTabContent.tsx
djhi: Nope, it's used in `getLabel`

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
fzaninotto: A comment in the code will make future you aware of that peculiarity

fzaninotto (approved):

PR #9918, opened 2024-06-12
Fix ArrayInput ghost error after removing scalar item

### LINKED ISSUE #9905: Removing an item from an ArrayInput with primitive values and validation removes more than one input
**What you were expecting:**
Removing an item should only remove that one item

**What happened instead:**
When empty inputs are present removing an item removes other items as well

**Steps to reproduce:**
Have an ArrayInput with string values. The array and values are validated as required.
Sandbox: https://codesandbox.io/p/devbox/arrayinput-remove-item-issue-sx4zky?file=%2Fsrc%2FApp.tsx%3A25%2C26&workspaceId=03aee9bf-4bf8-480d-97ef-9f80cfaf967b

Steps 1:

1. Open edit view
2. Add a new item, leave it empty
3. Try to save the form, validation errors will appear
4. Click on the remove button of the first item

Expected result:
5. Only the first item should be removed
Actual result:
5. The first and the last, invalid and empty item was removed, the validation error from the whole array still remained

https://github.com/marmelab/react-admin/assets/4885562/624fb9fb-4b24-4d56-a7ac-3fabbefb9676

Steps 2:
1. Open edit view
2. Add a new item, leave it empty
3. Remove the value of the second item
4. Click on the remove button of the first item

Expected result:
5. Only the first item should be removed

Actual result:
5. One empty input the third item remained, the other empty input and fourth item was also removed 

https://github.com/marmelab/react-admin/assets/4885562/1b0e9877-124d-4c36-a805-1b0078b0a9ca

**Related code:**

```
<SimpleForm>
      <ArrayInput source="names" validate={required()}>
        <SimpleFormIterator>
          <TextInput source="" validate={required()} />
        </SimpleFormIterator>
      </ArrayInput>
</SimpleForm>
```

**Environment**

* React-admin version: 4.16.9
* React version: 18.2.0
* Browser: Chrome

djhi:
Fixes #9905

This PR only fixes the ghost error. The removal of the last empty item is done by react-hook-form

### DIFF
--- a/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
+++ b/packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
@@ -61,7 +61,7 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
     } = props;
     const [confirmIsOpen, setConfirmIsOpen] = useState<boolean>(false);
     const { append, fields, move, remove, replace } = useArrayInput(props);
-    const { resetField } = useFormContext();
+    const { resetField, trigger } = useFormContext();
     const translate = useTranslate();
     const record = useRecordContext(props);
     const initialDefaultValue = useRef({});
@@ -69,8 +69,11 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
     const removeField = useCallback(
         (index: number) => {
             remove(index);
+            // Trigger validation on the Array to avoid ghost errors.
+            // Otherwise, validation errors on removed fields might still be displayed
+            trigger(source);
         },
-        [remove]
+        [remove, trigger, source]
     );
 
     if (fields.length > 0) {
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/input/ArrayInput/SimpleFormIterator.tsx
```diff
@@ -61,16 +61,19 @@ export const SimpleFormIterator = (props: SimpleFormIteratorProps) => {
...
     const translate = useTranslate();
     const record = useRecordContext(props);
     const initialDefaultValue = useRef({});
 
     const removeField = useCallback(
         (index: number) => {
             remove(index);
+            // Trigger validation on the Array to avoid ghost errors.
+            // Otherwise, validation errors on removed fields might still be displayed
+            trigger(source);
```
fzaninotto, line 74: Triggering validation on remove is unexpected. Can we do that only if we're using scalar values?

adguernier (approved):

Issue #9462, opened 2023-11-21
[RFC] React-admin v5


fzaninotto:
## Rationale

It's been [a year and a half](https://github.com/marmelab/react-admin/releases/tag/v4.0.0) since version 4.0. The React ecosystem has evolved a lot since then, and we must accept a few breaking changes in the react-admin codebase to keep up-to-date. So we're starting the work on the next major version, react-admin v5.

The main constraint of this version is to be a **painless migration from v4**. The 3.0 to 4.0 migration was costly for many developers, we don't want to reiterate that. The migration should take at most a day on a moderate size admin app. 

## Breaking Changes

- [x] Bump React version to 18
- [x] Underline links by default (#9483)
- [x] Infer datagrid rowClick based on resource definition (#9466)
- [x] Enable dark theme by default, so that new admins have a theme switcher (#9479)
- [x] Remove injected props in `Admin` and `Layout` components (`layout`, `appBar`, `menu`, `error`, `catchAll`) (#9591)
- [x] Make all input widths consistent (#9704)
- [x] Replace `defaultProps` with default params values (#9586)
- [x] Remove `<Admin history>` prop and use react-router's Data Router (#9657)
- [x] Remove list's setFilters default debounce (#9682)
- [x] Drop support for IE (#9530)
- [x] Allow to set button size via theme (#9735)
- [x] Remove instances of `cloneElement` that we can replace with context
  - #9533
  - #9707
  - #9719
  - #9805
- [x] Update dependencies to their last major 
  - #9473
  - #9657
  - #9763
  - #9924
  - #9801
  - #9804
  - #9811
  - #9819
  - #9818
  - #9822
  - #9823
  - #9874
  - #9908
  - ~~#9812~~ (reverted in #9871 because it causes compilation problems)
- [x] Stricter TypeScript types 
  - #9644
  - #9741
  - #9743
  - #9760
  - #9761
  - #9790
  - #9795
  - #9797
- [x] Remove PropTypes (#9851)
- [x] Apply new prettier rules (#9874)

See the progress in the following milestones:
- [x] [5.0.0-alpha.0](https://github.com/marmelab/react-admin/milestone/290?closed=1)
- [x] [5.0.0-beta.0](https://github.com/marmelab/react-admin/milestone/304?closed=1)
- [x] [5.0.0-beta.1](https://github.com/marmelab/react-admin/milestone/309?closed=1)
[...]

PR #9926, opened 2024-06-18
Fix create react admin

djhi:
## Problem

`create-react-admin` has not been updated for v5:
https://github.com/marmelab/react-admin/issues/9462#issuecomment-2174559652

## Solution

- Upgrade ink so that we use react 18
- Upgrade the templates for latest rc

### DIFF
--- a/packages/create-react-admin/src/generateProject.ts
+++ b/packages/create-react-admin/src/generateProject.ts
@@ -1,14 +1,17 @@
 import path from 'path';
 import fs from 'fs';
 import fsExtra from 'fs-extra';
-import merge from 'lodash/merge';
+import url from 'url';
+import merge from 'lodash/merge.js';
 import { ProjectConfiguration } from './ProjectState.js';
 import { generateAppFile } from './generateAppFile.js';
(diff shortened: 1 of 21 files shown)


packages/create-react-admin/src/generateProject.ts
```diff
@@ -1,14 +1,17 @@
...
 import fsExtra from 'fs-extra';
-import merge from 'lodash/merge';
+import url from 'url';
+import merge from 'lodash/merge.js';
 import { ProjectConfiguration } from './ProjectState.js';
 import { generateAppFile } from './generateAppFile.js';
 import { generateAppTestFile } from './generateAppTestFile.js';
 
+const __filename = url.fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
djhi, line 11: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#what-do-i-use-instead-of-__dirname-and-__filename

fzaninotto (approved):

PR #9939, opened 2024-06-20
Fix useFormGroup does not reflect its fields state

djhi:
# Problem

`useFormGroup` does not reflect its field state: when typing in a `TextInput` for instance, the `isDirty` state stays false until the input looses focus.

# Solution

Fix the effect dependency so that it is correctly triggered when the form state changes.

### DIFF
--- a/packages/ra-core/src/form/useFormGroup.ts
+++ b/packages/ra-core/src/form/useFormGroup.ts
@@ -1,8 +1,9 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import get from 'lodash/get';
 import isEqual from 'lodash/isEqual';
 import { useFormState } from 'react-hook-form';
 import { useFormGroups } from './useFormGroups';
+import { useEvent } from '../util';
 
 type FieldState = {
     name: string;
@@ -10,6 +11,7 @@ type FieldState = {
     isDirty: boolean;
     isTouched: boolean;
     isValid: boolean;
+    isValidating: boolean;
 };
 
 type FormGroupState = {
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/form/useFormGroup.ts
```diff
@@ -63,16 +66,27 @@ type FormGroupState = {
...
+    const { dirtyFields, touchedFields, validatingFields, errors } =
+        useFormState();
+
+    // dirtyFields, touchedFields and validatingFields are objects with keys being the field names
+    // Ex: { title: true }
+    // However, they are not correctly serialized when using JSON.stringify
+    // To avoid our effects to not be triggered when they should, we extract the keys and use that as a dependency
+    const dirtyFieldsNames = Object.keys(dirtyFields);
+    const touchedFieldsNames = Object.keys(touchedFields);
+    const validatingFieldsNames = Object.keys(validatingFields);
```
slax57, line 78: will this work with nested inputs? :thinking:

packages/ra-ui-materialui/src/input/ResettableTextField.tsx
```diff
@@ -39,20 +39,6 @@ export const ResettableTextField = forwardRef(
...
-            },
-            [onFocus]
-        );
-
-        const handleBlur = useCallback(
-            event => {
-                onBlur && onBlur(event);
-            },
-            [onBlur]
-        );
```
slax57, line 54: why did you remove this? :thinking:

djhi: Because they don't do anything ?

packages/ra-core/src/form/useFormGroup.ts
djhi: Yes, react-hook-form field name for deep fields are `one.two.three`

PR #9948, opened 2024-06-24
Fix DatagridHeader Tooltip when using React element as a field label

djhi:
## Problem

When using a React element as `label` for a field, the datagrid header tooltip displays `Sort by [object Oject] descending`.

## Solution

Fallback to the default inferred label

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/List.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/List.stories.tsx
@@ -504,32 +504,3 @@ export const ErrorInFetch = () => (
         </Admin>
     </TestMemoryRouter>
 );
-
-export const LabelElements = () => (
-    <TestMemoryRouter initialEntries={['/books']}>
-        <Admin dataProvider={dataProvider}>
-            <Resource
-                name="books"
-                list={
-                    <List>
-                        <Datagrid>
-                            <TextField source="id" label={<span>ID</span>} />
-                            <TextField
-                                source="title"
-                                label={<span>TITLE</span>}
-                            />
-                            <TextField
-                                source="author"
-                                label={<span>AUTHOR</span>}
-                            />
-                            <TextField
-                                source="year"
-                                label={<span>YEAR</span>}
-                            />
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridHeaderCell.tsx
```diff
@@ -23,7 +23,9 @@ export const DatagridHeaderCell = (
     const sortLabel = translate('ra.sort.sort_by', {
         field: field
             ? translateLabel({
-                  label: field.props.label,
+                  label: isValidElement(field.props.label)
+                      ? undefined
+                      : field.props.label,
```
fzaninotto, line 28: I'd invert the logic by only using the label if it's a string

packages/ra-ui-materialui/src/list/List.stories.tsx
```diff
@@ -504,3 +504,32 @@ export const ErrorInFetch = () => (
         </Admin>
     </TestMemoryRouter>
 );
+
+export const LabelElements = () => (
```
fzaninotto, line 508: his should be a Datagrid story

fzaninotto (changes requested):

adguernier (approved):

packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
```diff
@@ -29,6 +29,7 @@ import { List } from '../List';
 import { EditGuesser } from '../../detail';
 import { DatagridRowProps } from './DatagridRow';
 import DatagridBody, { DatagridBodyProps } from './DatagridBody';
+import { Admin } from 'react-admin';
```
fzaninotto, line 32: Can't you use AdminUI and AdminContext from the current package instead?

fzaninotto (changes requested):

fzaninotto (approved):

PR #9951, opened 2024-06-24
Update `<ReferenceField>` to render a link to the show view when relevant

fzaninotto:
## Problem

`<ReferenceField>` renders a link to the edit view by default. When a resource has a show view but no edit view, and the developer hasn't specified the `link` type, a `<ReferenceField>` targeting that resource doesn't show any link. 

Also, when there is a conflict between the `<ReferenceField link>` and the associated `<Resource>` (e.g. when the developer specified that the `ReferenceField` should link to the `edit` view even though the `Resource` doesn't defined such a view), react-admin ignores the developer choice and removes the link. Yet the developer may want to link to a nested `CustomRoute` that matches the `link`. 

## Solution

Make `<ReferenceField>` default to `show` when `link` is undefined and there are both an edit and a show view. 

Make `<ReferenceField>` smarter so that it can link to the show view if it exists. 

Make `<ReferenceField>` always honor the `link` prop.

**Before**

|    link                  | default  | edit    | show    | false   |
|--------------------------|--------- |---------|---------|---------|
| Resource hasShow hasEdit | edit     | edit    | show    | no link |
| Resource  hasEdit        | edit     | edit    | no link | no link |
| Resource hasShow         | no link  | no link | show    | no link |
| Resource                 | no link  | no link | no link | no link |

**This PR**

|    link                  | default  | edit    | show    | false   |
|--------------------------|--------- |---------|---------|---------|
| Resource hasShow hasEdit | **show**   | edit  | show  | no link |
| Resource  hasEdit        | edit   | edit  | **show** | no link |
| Resource hasShow         | **show**   | **edit**  | show  | no link |
| Resource                 | no link | **edit**  | **show**  | no link |

The link behavior of `<Datagrid rowClick>`, `<ReferenceField link>` and `<ReferenceOneField link>` is now consistent.

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceFieldController.ts
@@ -1,9 +1,9 @@
 import { useMemo } from 'react';
 import { UseQueryOptions } from '@tanstack/react-query';
+
 import { RaRecord } from '../../types';
-import { LinkToType, useCreatePath } from '../../routing';
+import { LinkToType, useGetRouteForRecord } from '../../routing';
 import { UseReferenceResult, useReference } from '../useReference';
-import { useResourceDefinition } from '../../core';
 import { useFieldValue } from '../../util';
 
 export const useReferenceFieldController = <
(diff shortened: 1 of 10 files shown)


packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -146,25 +145,21 @@ export const ReferenceFieldView = <
     if (link) {
         return (
             <Root className={className} sx={sx}>
-                <RecordContextProvider value={referenceRecord}>
```
fzaninotto, line 149: The RecordContextProvider was added twice, as `ReferenceFieldBase` already does it.

packages/ra-core/src/controller/field/useReferenceFieldController.ts
```diff
@@ -33,31 +33,34 @@ export const useReferenceFieldController = <
...
+                                    referenceRecordQuery.referenceRecord,
+                                    reference
+                                )
+                              : link ?? defaultLink,
+                  });
+        return {
+            ...referenceRecordQuery,
+            link: linkString,
+        } as const;
+    }, [createPath, link, reference, referenceRecordQuery, resourceDefinition]);
```
djhi, line 63: We could probably extract this in a custom hook to use here and for the Datagrid rowClick

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (!resource) {
+        throw new Error(
+            'Cannot generate a link for a record without a resource. You must use useGetRouteForRecord within a ResourceContextProvider, or pass a resource prop.'
+        );
+    }
+    const createPath = useCreatePath();
+    const resourceDefinition = useResourceDefinition({ resource });
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    const linkFunc = useCallback(
```
fzaninotto, line 56: We cannot use `useEvent` here because we need to call the function right away, and `useEvent` throws an Error in that case.

djhi: That means that if the function did change, it will never be updated here though right?

fzaninotto: yes indeed, that's a limitation I'm willing to accept

packages/ra-ui-materialui/src/list/datagrid/DatagridRow.tsx
```diff
@@ -123,42 +114,35 @@ const DatagridRow: React.ForwardRefExoticComponent<
...
                 handleToggleSelection(event);
                 return;
             }
-            navigate(type);
+            if (path === false || path == null) {
+                return;
+            }
+            navigate(path, {
+                state: { _scrollToTop: true },
+            });
```
fzaninotto, line 143: This is a breaking change: if the user passed a `path` or a function that returns a `path`, it will have `scrollToTop: true` while it used not to.

PR #9952, opened 2024-06-25
Fix flaky tests

djhi:
- [x] Fix useDelete tests
- [x] Fix create-react-admin tests

### DIFF
--- a/packages/create-react-admin/src/generateAppTestFile.ts
+++ b/packages/create-react-admin/src/generateAppTestFile.ts
@@ -20,6 +20,9 @@ test("should pass", async () => {
             ? `
 
 	// Sign in
+  	// Ensure the form is fully loaded before interacting with it
+  	await new Promise((resolve) => setTimeout(resolve, 1000));
+
 	fireEvent.change(await screen.findByLabelText("Username *"), {
 		target: { value: "janedoe" },
 	});
@@ -33,6 +36,9 @@ test("should pass", async () => {
 	// Open the first post
 	fireEvent.click(await screen.findByText("Post 1"));
 	fireEvent.click(await screen.findByText("Edit"));
+	await screen.findByDisplayValue("Post 1");
+	// Ensure the form is fully loaded before interacting with it
+	await new Promise((resolve) => setTimeout(resolve, 1000));
 	// Update its title
 	fireEvent.change(await screen.findByDisplayValue("Post 1"), {
 		target: { value: "Post 1 edited" },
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/dataProvider/useDelete.spec.tsx
```diff
@@ -267,22 +267,22 @@ describe('useDelete', () => {
     });
 
     describe('mutationMode', () => {
-        it('when pessimistic, displays result and success side effects when dataProvider promise resolves', async () => {
+        it.only('when pessimistic, displays result and success side effects when dataProvider promise resolves', async () => {
```
slax57, line 270: `it.only` spotted

```suggestion
        it('when pessimistic, displays result and success side effects when dataProvider promise resolves', async () => {
```

slax57 (changes requested):

packages/ra-core/src/dataProvider/useDelete.spec.tsx
```diff
@@ -273,16 +273,16 @@ describe('useDelete', () => {
             screen.getByText('Delete first post').click();
             await waitFor(() => {
                 expect(screen.queryByText('success')).toBeNull();
-                expect(screen.queryByText('Hello')).not.toBeNull();
-                expect(screen.queryByText('World')).not.toBeNull();
                 expect(screen.queryByText('mutating')).not.toBeNull();
             });
+            expect(screen.queryByText('Hello')).not.toBeNull();
+            expect(screen.queryByText('World')).not.toBeNull();
             await waitFor(() => {
```
fzaninotto, line 280: the default timeout for waitFor is 1000ms. The tested story adds a delay of 1000ms to the delete. It's logical that this may fail from time to time. 

I advise to reduce the delay to 500 in the story to make it more robust

packages/create-react-admin/src/generateAppTestFile.ts
```diff
@@ -20,6 +20,9 @@ test("should pass", async () => {
             ? `
 
 	// Sign in
+  	// Ensure the form is fully loaded before interacting with it
+  	await new Promise((resolve) => setTimeout(resolve, 1000));
```
fzaninotto, line 24: You should use an assertion in a waitfor instead of a timeout. Same for the other 2 instances

djhi: We already do. I still had to add this

fzaninotto: Instead, add a timeout option to the initial `findByLabelText`, as it proxies its options to `waitFor`.

PR #9960, opened 2024-06-26
Fix logging of queries in ra-data-fakerest

### LINKED ISSUE #9959: ra-data-fakerest doesn't log response to the console
**What you were expecting:**
When opening the log line of a fakerest query, see the response

**What happened instead:**
it shows an empty Promise instead

image

**Steps to reproduce:**

1. Open the [simple example stackblitz](https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple)
2. Open the console
3. Develop one of the lines logged by the data provider

**Environment**

* React-admin version: 5.0.1
* Last version that did not exhibit the issue (if applicable): 4.x
* React version: 18
* Browser: Chrome

fzaninotto:
Closes #9959

### DIFF
--- a/packages/ra-data-fakerest/src/index.ts
+++ b/packages/ra-data-fakerest/src/index.ts
@@ -153,27 +153,27 @@ export default (data, loggingEnabled = false, delay?: number): DataProvider => {
      * @param {Object} params The data request params, depending on the type
      * @returns {Promise} The response
      */
-    const handle = (type, resource, params): Promise<any> => {
+    const handle = async (type, resource, params): Promise<any> => {
         const collection = database.getCollection(resource);
         if (!collection && type !== 'create') {
             const error = new UndefinedResourceError(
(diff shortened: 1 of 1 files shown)


packages/ra-data-fakerest/src/index.ts
```diff
@@ -153,27 +153,27 @@ export default (data, loggingEnabled = false, delay?: number): DataProvider => {
...
                 `Undefined collection "${resource}"`
             );
             error.code = 1; // make that error detectable
-            return Promise.reject(error);
+            throw error;
         }
         let response;
         try {
-            response = getResponse(type, resource, params);
+            response = await getResponse(type, resource, params);
```
fzaninotto, line 167: This was the root cause: since the introduction of the delay in #9908, `getResponse` may return a Promise.

djhi (approved):

PR #9966, opened 2024-06-28
Fix `<SimpleShowLayout>` uses a wrong translation key for field labels

### LINKED ISSUE #9965: SimpleShowLayout does not pass resource to Labeled Component, translation fails
**What you were expecting:**
When nesting Field Components into a SimpleShowLayout the Label should be correctly translated.
**What happened instead:**
After migration to react-admin v5 the Fields in a SimpleShow Layout have not the correct translation
**Steps to reproduce:**
Just nest Fields into a SimpleShowLayout with a translation of fields to a resource in place
**Related code:**

SimpleShowLayout is not passing resource to Labeled
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/detail/SimpleShowLayout.tsx
Labeled ist only using the passed resource prop and not also the ResourceContext
https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/Labeled.tsx

**Environment**

* React-admin version: 5.0.2
* Last version that did not exhibit the issue (if applicable): 4.x.x
* React version: 18
* Browser: Chrome

fzaninotto:
The default SourceContext introduced by 4d9820d5931c3eb992010035c23c00935f194047 breaks default field labels.

Closes #9965

### DIFF
--- a/packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
+++ b/packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
@@ -1,177 +1,95 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
-import { useTranslateLabel } from './useTranslateLabel';
-import { TestTranslationProvider } from './TestTranslationProvider';
-import { SourceContextProvider } from '..';
+
+import {
+    Basic,
+    I18nLabelAsKey,
+    I18nNoTranslation,
+    I18nTranslation,
+    InSourceContext,
+    InSourceContextI18nKey,
+    InSourceContextNoTranslation,
+    InSourceContextWithResource,
+    LabelElement,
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/i18n/useTranslateLabel.spec.tsx
```diff
@@ -1,177 +1,95 @@
...
+    LabelElement,
+    LabelEmpty,
+    LabelFalse,
+    LabelText,
+    Resource,
+    Source,
+} from './useTranslateLabel.stories';
 
 describe('useTranslateLabel', () => {
-    const TranslateLabel = ({
```
fzaninotto, line 8: The components to test have moved to the stories file

PR #9968, opened 2024-07-01
[TypeScript] Improve List exporter type

fzaninotto:
## Problem

Writing a custom `<List exporter>` by hand is cumbersome, as TypeScript complains a lot. 

## Solution

Avoid `any` in the type definition of the `fetchRelatedRecords` function parameter. 

Give an example in the documentation.


packages/ra-core/src/export/fetchRelatedRecords.ts
```diff
@@ -13,7 +18,8 @@ import { RaRecord, Identifier, DataProvider } from '../types';
  *     );
  */
 const fetchRelatedRecords =
```
djhi, line 20: Good time to make it a named export?

fzaninotto: Done

docs/List.md
```diff
@@ -607,31 +607,32 @@ In many cases, you'll need more than simple object manipulation. You'll need to
...
+    }));
+    return jsonExport(commentsWithPostTitle, {
+        headers: ['id', 'post_id', 'post_title', 'body'],
+    }, (err, csv) => {
+        downloadCSV(csv, 'comments');
     });
 };
 
 const CommentList = () => (
     <List exporter={exporter}>
```
djhi, line 632: We can probably go one step further and make `exporter` typed when passing a type to `List`:
```tsx
<List<Comment> exporter={exporter}>
```

It would be nice

packages/ra-core/src/types.ts
```diff
@@ -368,15 +368,17 @@ export interface ResourceProps {
 
 export type Exporter = (
```
djhi, line 369: Exporter should accept a generic type parameter for `data`

djhi (approved):

PR #9969, opened 2024-07-01
Add ability to disable title in page components

### LINKED ISSUE #9695: Change title property type in core display components <Edit>, <Show>, <View>

fzaninotto:
Closes #9695

### DIFF
--- a/docs/Create.md
+++ b/docs/Create.md
@@ -318,7 +318,7 @@ export const PostCreate = () => (
 );
 ```
 
-The `title` value can be a string or a React element.
+The title can be either a string, a React element, or `false` to disable the title.
 
 ## `transform`
 
--- a/docs/Edit.md
+++ b/docs/Edit.md
@@ -519,9 +519,19 @@ To override the style of all instances of `<Edit>` components using the [applica
 
 ## `title`
 
-By default, the title for the Edit view is “Edit [resource_name] #[record_id]”.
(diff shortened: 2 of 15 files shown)


djhi (commented):
Looks good :+1: I just noticed we don't have any stories for `Create` either

djhi (approved):

PR #9970, opened 2024-07-02
[TypeScript] Fix data provider packages export non-strict types

fzaninotto:
Since third-party data providers stem from ours, we should provide data providers that compile with strictNullChecks

### DIFF
--- a/packages/ra-data-json-server/src/index.ts
+++ b/packages/ra-data-json-server/src/index.ts
@@ -35,14 +35,17 @@ import { fetchUtils, DataProvider } from 'ra-core';
  */
 export default (apiUrl, httpClient = fetchUtils.fetchJson): DataProvider => ({
     getList: (resource, params) => {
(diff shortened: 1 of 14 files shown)


packages/ra-data-json-server/src/index.ts
```diff
@@ -35,14 +35,17 @@ import { fetchUtils, DataProvider } from 'ra-core';
  */
 export default (apiUrl, httpClient = fetchUtils.fetchJson): DataProvider => ({
     getList: (resource, params) => {
-        const { page, perPage } = params.pagination;
-        const { field, order } = params.sort;
+        const { page, perPage } = params.pagination || {};
```
fzaninotto, line 38: We should have to do the same for `getManyReference`, except the Params types for the two methods don't match... params are optional in `getList` and required for `getManyReference`.

djhi (approved):

PR #9971, opened 2024-07-02
[TypeScript] Fix useGetOne and useGetMany params type when id param is undefined

fzaninotto:
## Problem

TypeScript complains when `useGetOne` is called with an undefined `id`. This is fine, except when using the `enabled` option:

image

This is a consequence of the strictNullCheck flag in ra-core. 

## Solutions

1. Use type overloads to allow optional id when enabled is specified. This forces the developer to set teh enabled flag when id may be undefined - not ideal.
2. Change input type for the data provider hooks to allow an undefined id, and set the enabled flag in the hook accordingly. this offers a better DX as TypeScript won't complain but user land code won't break either. 

We choose solution 2.

### DIFF
--- a/packages/ra-core/src/dataProvider/useGetMany.spec.tsx
+++ b/packages/ra-core/src/dataProvider/useGetMany.spec.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react';
 import expect from 'expect';
 import { render, waitFor } from '@testing-library/react';
+import { QueryClient } from '@tanstack/react-query';
 
 import { CoreAdminContext } from '../core';
 import { useGetMany } from './useGetMany';
+import { useGetOne } from './useGetOne';
 import { testDataProvider } from '../dataProvider';
-import { useState } from 'react';
-import { QueryClient } from '@tanstack/react-query';
 
 const UseGetMany = ({
     resource,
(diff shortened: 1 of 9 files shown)


fzaninotto:
Now I wonder if that's the right fix. Instead of forcing users to set the enabled flag when the main param may be undefined, shouldn't we relax the params type to accept undefined values, and set the enabled flag ourselves if the main param is undefined?

djhi:
> Now I wonder if that's the right fix. Instead of forcing users to set the enabled flag when the main param may be undefined, shouldn't we relax the params type to accept undefined values, and set the enabled flag ourselves if the main param is undefined?

I like this idea :+1:

packages/ra-core/src/dataProvider/useGetOne.spec.tsx
```diff
@@ -299,4 +299,38 @@ describe('useGetOne', () => {
...
+                };
+                const { data, error, isPending } = useGetOne<Post>('posts', {
+                    id: comment?.post_id,
+                });
+                if (isPending || error) return null;
+                return <div>{data.title}</div>;
+            };
+            // no render needed, only checking types
+        });
+    });
```
djhi, line 335: Unfortunately, as we don't compile our tests, we won't ever see this fail in CI

packages/ra-core/src/dataProvider/useGetMany.spec.tsx
```diff
@@ -398,4 +398,45 @@ describe('useGetMany', () => {
             expect(abort).toHaveBeenCalled();
         });
     });
+
+    describe('TypeScript', () => {
```
djhi, line 402: same

PR #9976, opened 2024-07-05
fix(ra-data-graphql): fix an issue with custom context being overwritten on queries

### LINKED ISSUE #9973: [ra-data-graphql] custom query context overridden in v5
**What you were expecting:**

After upgrading to `^v5.0.0` some of our GraphQL queries have broken. This is the result of [this commit](https://github.com/marmelab/react-admin/commit/afab2c40ae47856d888d0cb113986a0162ee7525#diff-f31485136671519259100483dab8bf875fc60818d010451f00aec2d0e7c4f5cdR188) where a `context` was added _after_ `getOptions`.

**What happened instead:**

We're unable to populate custom context on a GraphQL query.

**Steps to reproduce:**

1. Add a custom context object to the return statement of the `query` option.
2. Try to access that context variable in something like `ApolloLink.split(...)`.

**Related code:**

Here is [the commit where this change was introduced](https://github.com/marmelab/react-admin/commit/afab2c40ae47856d888d0cb113986a0162ee7525#diff-f31485136671519259100483dab8bf875fc60818d010451f00aec2d0e7c4f5cdR188) where a `context` was added _after_ `getOptions`:

```ts
// ...
...getOptions(otherOptions.query, raFetchMethod, resource),
context: {
    fetchOptions: {
        signal: params?.signal,
    },
},
// ...
```

The reason for this is that we've passed a custom `query` option to the `buildGraphQLProvider` function. This `query` function populates some custom context like so:

```ts
buildGraphQLProvider({
  // ...
  query: (resource: string) => ({
      context: { foo: 'bar' },
  }),
  // ...
});
```

We then use this context in `ApolloLink.split()` to determine what endpoint to route a given request to.

**Other information:**

I realize that this might be a niche use-case. However, I think the problem would be easily avoidable via one of the following:

1. checking of any `context` was generated from the `query` function and, if so merging it with the new `fetchOptions` context.
2. specify the custom context _above_ the `...getOptions(otherOptions.query, raFetchMethod, resource),` call and leave it to the end user to deal with (e.g. force them to pass that option in their context.

I'll also add that I'm happy to submit a PR if the maintainers approve.

**Environment**

* React-admin version: `5.0.3`
* Last version that did not exhibit the issue (if applicable): `4.16.19`.
* React version: `^18.3.0`
* Browser: Brave

jspizziri:
fix #9973

### DIFF
--- a/packages/ra-data-graphql/src/index.ts
+++ b/packages/ra-data-graphql/src/index.ts
@@ -186,12 +186,16 @@ const buildGraphQLProvider = (options: Options): DataProvider => {
                 ...query,
                 fetchPolicy: 'network-only',
                 ...getOptions(otherOptions.query, raFetchMethod, resource),
-                context: {
+            };
+
+            apolloQuery.context = merge(
+                {
                     fetchOptions: {
                         signal: params?.signal,
                     },
                 },
-            };
+                apolloQuery.context
+            );
 
             return (
                 client


djhi (approved):
Thanks!

packages/ra-data-graphql/src/index.ts
```diff
@@ -186,12 +186,16 @@ const buildGraphQLProvider = (options: Options): DataProvider => {
...
+            };
+
+            apolloQuery.context = merge(
+                {
                     fetchOptions: {
                         signal: params?.signal,
                     },
                 },
-            };
+                apolloQuery.context,
```
djhi, line 197:
```suggestion
                apolloQuery.context
```

djhi (changes requested):

jspizziri:
@djhi , updated per your feedback.

jspizziri:
@djhi , friendly ping.

PR #9983, opened 2024-07-09
Fix(WrapperField): source props should not be required

jonathan-marmelab:
`WrapperField` [documentation](https://marmelab.com/react-admin/WrapperField.html) marks `source` props as not required, however changes in #9620 makes it required. This PR reverts the change for `WrapperField` only


djhi (approved):

slax57 (approved):
Thanks

PR #9994, opened 2024-07-12
Create objectfield component

rnsnceman:
## Problem

No simple solution for nested objects

## Solution

I have created an `<ObjectField>` component which can be used in any situation that can provide the record data of an individual item such as within a map() function, or in a `<Show>` component (the record prop is optional in a `<Show>` component).

Grid example
Show view example

## How To Test

Using the `simple` template in `examples`, add a nested object with entries in one of the comment data's items in the `data.tsx` file. Then import `<ObjectField>` in the `CommentList.tsx` file and add the `<ObjectField>` component in the `<CommentGrid>` component and wrap it around `<TextField>` components representing the entries of the nested object. Here's an example:

```jsx
<CardContent>
        <ObjectField source="shipping" record={record} >
            <TextField source="address" />
            <TextField source="zipcode" />
        </ObjectField>
    </CardContent>
```

## Additional Checks

- [ ] The PR includes **unit tests** (if not possible, describe  #why)

#


fzaninotto:
Thanks for your PR. 

You say your problem is:

> No simple solution for nested objects

But all react-admin fields can use a path as source, e.g. 

```
<TextField source="shipping.address" />
```

So there is a simple solution for nested objects. 

Can you elaborate on why this solution doesn't fit in your case?

Also, your PR is embarking a lot of commits it shouldn't - you have probably messed up a rebase.

rnsnceman:
Oh wow, I didn't realize you could do it that way. My apologies. Is there any specific documentation i can reference that from?

PR #9997, opened 2024-07-12
Feat(crm): Add CTA for empty contact

arimet:
## How to test

In contact page, open console. Set `window._database.collections.contacts.items = []`
Refresh via the refresh button.

image


jonathan-marmelab (approved):

examples/crm/src/contacts/ContactList.tsx
```diff
@@ -4,34 +4,58 @@ import {
...
-        </RaList>
+            <ContactListLayout />
+        </ListBase>
+    );
+};
+
+const ContactListLayout = () => {
+    const { data, isPending } = useListContext();
+    if (isPending) return <LinearProgress />;
+    if (!data?.length) return <ContactEmpty />;
```
fzaninotto, line 45: You must also test if there is another filter than the current sales, otherwise the create invite page will also show up when the user applies a filter to the contact list that gives no result.

PR #10000, opened 2024-07-12
Feat(contacts): Add import CSV feature

jonathan-marmelab:
[Capture vidéo du 19-07-2024 11:36:59.webm](https://github.com/user-attachments/assets/79b60aaa-4983-46b3-9de7-f64801e4ec6a)


examples/crm/src/contacts/ContactList.tsx
```diff
@@ -38,6 +40,9 @@ export const ContactList = () => {
 const ContactListActions = () => (
     <TopToolbar>
         <SortButton fields={['last_name', 'first_name', 'last_seen']} />
+        <ImportButton
+            sampleUrl={toDataUrl(sampleCsv.default, 'text/csv;charset=utf-8')}
```
fzaninotto, line 44: I wouldn't put it here as an argument, and import the sampleCSV directly in ImportButton.

examples/crm/src/contacts/contacts_export.csv
```diff
@@ -0,0 +1,2 @@
+first_name,last_name,gender,title,company,email,phone_number1,phone_number2,background,acquisition,avatar,first_seen,last_seen,has_newsletter,status,tags,sales
```
fzaninotto, line 1: It would be better if the example showed the syntax for multiple tags

examples/crm/src/contacts/ContactImportButton.tsx
```diff
@@ -0,0 +1,264 @@
...
+    };
+
+    const startImport = async () => {
+        if (!file) {
+            return;
+        }
+
+        setImportState('running');
+
+        Papa.parse<ContentImportSchema>(file, {
```
arimet, line 63: I'm just wondering if this part shouldn't be in a separate file.

arimet (approved):

examples/crm/src/contacts/ContactImportButton.tsx
jonathan-marmelab: I think this part can be factorized when implementing company import if it's good for you ?

PR #10001, opened 2024-07-13
separate test and default data providers

Nilegfx:
## Problem
defaultDataProvider behaviour changed in V5. the change was intended to just support the tests. this affect the end users if they use the exported `defaultDataProvider` in their implementation which means one extra migration step for them and one extra migration documentation entry point react-admin team has to add.

## Solution
Implement testing-related changes for the `defaultDataProvider` in the right place `testDataProvider` and revert the `defaultDataProvider` to its original behaviour (which is dummy responses but at least it doesn't throw)

## Struggles/Challenges
I implemented the refactoring in 30 seconds, then I spent almost half an hour trying to make the `defaultDataProvider` work with the `DataProvider` type but I failed. I would appreciate if one of the core members can explain to me what I miss here or how to correctly type the `defaultDataProvider`.
you can reproduce the issue by just changing `export const defaultDataProvider = {` to `export const defaultDataProvider:DataProvider = {` (and of course import the `DataProvider` type.

I used a workaround `export const defaultDataProvider: DataProvider = {} as DataProvider` but I don't like this. I am still interested to understand this TypeError (I used AI but it couldn't give me logical answer)

## How To Test

giving that this change was intended to keep everything as the same, I assumed that the current tests should be enough for this simple and small refactoring

## Additional Checks

- [ ] The PR includes **unit tests**  (check the above How To Test section)
- [ ] The PR includes one or several **stories** (check the above How To Test section)

### DIFF
--- a/packages/ra-core/src/dataProvider/defaultDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/defaultDataProvider.ts
@@ -1,32 +1,12 @@
-import { DataProvider } from '../types';
-
 // avoids adding a context in tests
-export const defaultDataProvider: DataProvider = {
-    create: async () => {
-        throw new Error('create is not implemented');
-    },
-    delete: async () => {
-        throw new Error('delete not implemented');
-    },
-    deleteMany: async () => {
-        throw new Error('deleteMany is not implemented');
-    },
-    getList: async () => {
-        throw new Error('getList is not implemented');
-    },
-    getMany: async () => {
-        throw new Error('getMany is not implemented');
-    },
-    getManyReference: async () => {
-        throw new Error('getManyReference is not implemented');
-    },
-    getOne: async () => {
-        throw new Error('getOne is not implemented');
-    },
-    update: async () => {
-        throw new Error('update not implemented');
-    },
-    updateMany: async () => {
-        throw new Error('updateMany not implemented');
-    },
+export const defaultDataProvider = {
+    create: () => Promise.resolve({ data: null }),
+    delete: () => Promise.resolve({ data: null }),
+    deleteMany: () => Promise.resolve({ data: [] }),
+    getList: () => Promise.resolve({ data: [], total: 0 }),
+    getMany: () => Promise.resolve({ data: [] }),
+    getManyReference: () => Promise.resolve({ data: [], total: 0 }),
+    getOne: () => Promise.resolve({ data: { id: 'o' } }),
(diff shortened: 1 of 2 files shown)


adguernier:
Hi @Nilegfx and thanks for your contribution :+1: 

We used to type result as `any` in our tests and stories (e.g: `getOne: () => Promise.resolve({ data: null } as any),`. But to avoid using `any`, I'd type the `defaultDataProvider` as follows:

```ts
import {
    CreateResult,
    DataProvider,
    DeleteResult,
    GetOneResult,
    UpdateResult,
} from '../types';

export const defaultDataProvider: DataProvider = {
    create: () => Promise.resolve<CreateResult>({ data: null }),
    delete: () => Promise.resolve<DeleteResult>({ data: null }),
    deleteMany: () => Promise.resolve({ data: [] }),
    getList: () => Promise.resolve({ data: [], total: 0 }),
    getMany: () => Promise.resolve({ data: [] }),
    getManyReference: () => Promise.resolve({ data: [], total: 0 }),
    getOne: () => Promise.resolve<GetOneResult>({ data: null }),
    update: () => Promise.resolve<UpdateResult>({ data: null }),
    updateMany: () => Promise.resolve({ data: [] }),
};
```
What do you think about that @djhi ?

packages/ra-core/src/dataProvider/defaultDataProvider.ts
```diff
@@ -2,31 +2,13 @@ import { DataProvider } from '../types';
...
-        throw new Error('updateMany not implemented');
-    },
-};
+    create: () => Promise.resolve({ data: null }),
+    delete: () => Promise.resolve({ data: null }),
+    deleteMany: () => Promise.resolve({ data: [] }),
+    getList: () => Promise.resolve({ data: [], total: 0 }),
+    getMany: () => Promise.resolve({ data: [] }),
+    getManyReference: () => Promise.resolve({ data: [], total: 0 }),
+    getOne: () => Promise.resolve({ data: { id: 'o' } }),
```
adguernier, line 11: **question**: why  `{ data: { id: 'o' } }` instead of `{ data: null }` like the others?

Nilegfx: oh, that was committed by mistake, I will remove it.

adguernier (approved):

adguernier:
Awaiting further approval 
Thanks again :+1:

djhi (approved):

PR #10007, opened 2024-07-14
Feat(crm): Add auto import avatar for contact

arimet:


examples/crm/src/misc/getContactAvatar.ts
```diff
@@ -0,0 +1,75 @@
...
+    const hashBuffer = await crypto.subtle.digest('SHA-256', utf8);
+    const hashArray = Array.from(new Uint8Array(hashBuffer));
+    const hashHex = hashArray
+        .map(bytes => bytes.toString(16).padStart(2, '0'))
+        .join('');
+    return hashHex;
+}
+
+const DOMAINS_NOT_SUPPORTING_FAVICON = [
+    'google.com',
```
jonathan-marmelab, line 14: I'm not sure that the `google.com` is public mail, IMHO it is only used by googlers.

jonathan-marmelab (approved):

examples/crm/src/dataProvider.ts
```diff
@@ -11,6 +12,21 @@ const TASK_DONE_NOT_CHANGED = 'TASK_DONE_NOT_CHANGED';
...
+        beforeCreate: async params => {
+            const { data } = params;
+            const avatarUrl = await getAvatarUrl(data);
+            data.avatar = avatarUrl || null;
+            return params;
+        },
+        beforeUpdate: async params => {
+            const { data } = params;
+            const avatarUrl = await getAvatarUrl(data);
+            data.avatar = avatarUrl || null;
```
fzaninotto, line 26: This will delete existing avatar URLs on update, am I right?

examples/crm/src/misc/getContactAvatar.spec.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * @jest-environment node
+ */
```
fzaninotto, line 3: Why do you need that?

examples/crm/src/misc/getContactAvatar.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { Contact } from '../types';
+import { getAvatarUrl, hash } from './getContactAvatar';
+// eslint-disable-next-line no-global-assign
+import { webcrypto } from 'node:crypto';
+
+Object.defineProperty(globalThis, 'crypto', {
```
fzaninotto, line 10: Can't you use  jest.mock instead ?

examples/crm/src/misc/getContactAvatar.spec.ts
arimet: For hashing the email for Gravatar, we use TextEncoder. 
However, it seems that this API is not embedded with the version of jsdom used by jest on your pipeline. If you have a solution, I'm interested

PR #10017, opened 2024-07-16
Feat(crm): Add signup page when no user

arimet:


djhi (commented):
You should probably have a custom route for the signup. Take a look at `ra-supabase`. It supports invites among other things, password resets, etc.

examples/crm/src/SignupPage.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const {
+        register,
+        handleSubmit,
+        formState: { isValid },
+    } = useForm<UserInput>({
+        mode: 'onChange',
+    });
+
+    // For the moment, we only allow one user to sign up. Other users must be created by the administrator.
```
jonathan-marmelab, line 30: Maybe you should also handle the `isPending` state here, to avoid showing the form if response takes ages to be retrieved ?

examples/crm/src/dataProvider.ts
```diff
@@ -57,168 +57,209 @@ const beforeContactUpsert = async (
...
-            await dataProvider.update('contacts', {
-                id: contact_id,
-                data: {
-                    nb_notes: (contact.nb_notes ?? 0) - 1,
-                },
-                previousData: contact,
-            });
-            return result;
-        },
+const dataProviderWithCustomMethode = {
```
jonathan-marmelab, line 60:
```suggestion
const dataProviderWithCustomMethod = {
```

examples/crm/src/dataProvider.ts
```diff
@@ -57,168 +57,209 @@ const beforeContactUpsert = async (
...
-                    ? TASK_MARKED_AS_DONE
-                    : TASK_MARKED_AS_UNDONE;
-            } else {
-                taskUpdateType = TASK_DONE_NOT_CHANGED;
-            }
-            return params;
+};
+
+export const dataProvider = withLifecycleCallbacks(
+    dataProviderWithCustomMethode,
```
jonathan-marmelab, line 81:
```suggestion
    dataProviderWithCustomMethod,
```

jonathan-marmelab (changes requested):

PR #10037, opened 2024-07-22
Feat(crm): Add sales admin support

jonathan-marmelab:
## Solution

Implements full support for managing Sales and administrator permissions.

## How To Test

Log on as Jane Doe (default on the demo) or clear all data and start from scratch.
`Object.keys(window._database.collections).forEach(key => {window._database.collections[key].items = []})`

## Additional Checks

- [x] Account manager list
- [x] Account manager delete
- [x] Account manager create
- [x] Account manager create

[Capture vidéo du 22-07-2024 16:03:05.webm](https://github.com/user-attachments/assets/181f9cf9-8087-40d9-8f54-8f7c5dfd9c5c)


examples/crm/src/App.tsx
```diff
@@ -45,6 +46,7 @@ const App = () => (
         <Resource name="deals" {...deals} />
         <Resource name="contacts" {...contacts} />
         <Resource name="companies" {...companies} />
+        <Resource name="sales" {...sales} />
```
arimet, line 49: **polish:** Shouldn't we prevent access to the resource directly? rather than displaying an empty page?

In any case, if we don't block, we'd need a message rather than an empty page.‏

examples/crm/src/dataProvider.ts
```diff
@@ -82,11 +83,99 @@ const dataProviderWithCustomMethod = {
...
+                            pagination: {
+                                page: 1,
+                                perPage: 10_000,
+                            },
+                            sort: { field: 'id', order: 'ASC' },
+                        }),
+                    ]);
+
+                await Promise.all([
+                    dataProvider.deleteMany('companies', {
```
arimet, line 163: **issue:** 
We don't want to delete the linked resources, but update them so that they are linked to the administrator.
‏

examples/crm/src/sales/SalesCreate.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function SalesCreate() {
+    const isAdmin = useIsAdmin();
+    if (!isAdmin) {
+        return null;
+    }
+
+    return (
+        <Create>
+            <SimpleForm>
+                <SalesForm />
```
arimet, line 14: **note (non-blocking):** 
When we create a new sale, its administrator status is not displayed in the list‏

examples/crm/src/sales/SalesEdit.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (!isAdmin) {
+        return null;
+    }
+
+    return (
+        <Edit>
+            <SimpleForm toolbar={<EditToolbar />}>
+                <SalesForm />
+
+                <BooleanInput
```
arimet, line 42: **note (non-blocking):** I don't think this readOnly props work. When i'm logged as Jane Doe, i can update my status from this page...‏

examples/crm/src/sales/SalesList.tsx
```diff
@@ -0,0 +1,66 @@
...
+function SalesListActions() {
+    return (
+        <TopToolbar>
+            <ExportButton />
+            <CreateButton variant="contained" label="New Sale" />
+        </TopToolbar>
+    );
+}
+
+const filters = [<SearchInput source="q" alwaysOn />];
```
arimet, line 29: **nitpick (non-blocking):** ‏I don't know if it's possible but I can't search via prenon + name. Could we do better or are we limited by fakerest?

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const { identity } = useGetIdentity();
+    const dataProvider = useDataProvider<CustomDataProvider>();
+    const refresh = useRefresh();
+    const navigate = useNavigate();
+    const {
+        data: sales,
+        error,
+        isPending,
+    } = useGetList('sales', {
+        filter: { administrator: false },
```
arimet, line 40: **issue:** Starting from scratch, if I create my admin and then a new sale, I can't transfer my role, the selection is empty. I think it's linked to the fact that the administrator status is not well defined at creation.‏

examples/crm/src/sales/SalesCreate.tsx
arimet: You should add an default value for adminstrator

fzaninotto:
I'd remove the buttons from the list and use `rowClick="edit"`. Administrator can always delete a user from the Edit page. 
Also, I'd rename the "Account Managers" menu item to "Sales" for brevity.

examples/crm/src/dataProvider.ts
```diff
@@ -82,11 +83,99 @@ const dataProviderWithCustomMethod = {
         }
         return sale;
     },
+    transferAdministrator: async (from: Identifier, to: Identifier) => {
```
fzaninotto, line 86: `transferAdministratorRole`?

examples/crm/src/sales/TransferAdminRole.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+                        <Typography component="p">
+                            You can{' '}
+                            <Link
+                                onClick={handleShowModal}
+                                sx={{
+                                    cursor: 'pointer',
+                                }}
+                            >
+                                transfer administrator role
```
fzaninotto, line 49:
```suggestion
                                transfer the administrator role
```

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const [to, setTo] = useState<Sale | null>(null);
+    const [loading, setLoading] = useState(false);
+
+    const handleTransfer = () => {
+        if (!identity || !to) {
+            return;
+        }
+
+        setLoading(true);
+        void dataProvider
```
fzaninotto, line 56: no need for `void`

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+            .transferAdministrator(identity.id, to.id)
```
fzaninotto, line 57: I'd call it with useMutation and use side effects instead of `then`/ `finally`

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+            .then(updatedUser => {
+                localStorage.setItem(
+                    USER_STORAGE_KEY,
+                    JSON.stringify(updatedUser?.data)
+                );
+            })
+            .finally(() => {
+                onClose();
+                setLoading(false);
+                navigate('/');
```
fzaninotto, line 67: Why change page?

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+    } = useGetList('sales', {
+        filter: { administrator: false },
+        pagination: {
+            page: 1,
+            perPage: 1000,
+        },
+        sort: { field: 'first_name', order: 'ASC' },
+    });
+    const [to, setTo] = useState<Sale | null>(null);
+    const [loading, setLoading] = useState(false);
```
fzaninotto, line 48: by using `useMutation` you won't have to handle that manually

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+    onClose(): void;
+};
+
+export function TransferAdminRoleModal({
+    open,
+    onClose,
+}: TransferAdminRoleModalProps) {
+    const { identity } = useGetIdentity();
+    const dataProvider = useDataProvider<CustomDataProvider>();
+    const refresh = useRefresh();
```
fzaninotto, line 33: no, use the `refetch` from `useGetList` instead

examples/crm/src/sales/TransferAdminRoleModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        onChange={(event, newValue) => setTo(newValue)}
+                    />
+                )}
+            </DialogContent>
+            <DialogActions>
+                <Button onClick={onClose}>Cancel</Button>
+                <Button
+                    color="primary"
+                    variant="contained"
+                    onClick={handleTransfer}
```
fzaninotto, line 106: MAke it a sumbit button instead (you'll need a Form for that) so that the form can be validated with the keyboard.

examples/crm/src/sales/useIsAdmin.tsx
```diff
@@ -0,0 +1,12 @@
+import { useGetIdentity, useGetOne } from 'react-admin';
+import { Sale } from '../types';
+
+export function useIsAdmin() {
+    const { identity } = useGetIdentity();
+    const user = useGetOne<Sale>('sales', { id: identity?.id });
```
fzaninotto, line 6:
```suggestion
    const{ data, isPending, error } = useGetOne<Sale>('sales', { id: identity?.id }, { enabled: !!identity);
```

examples/crm/src/sales/useIsAdmin.tsx
```diff
@@ -0,0 +1,12 @@
...
+    if (!user?.data) {
```
fzaninotto, line 7:
```suggestion
    if (isPending || error) {
```

fzaninotto (changes requested):

examples/crm/src/sales/TransferAdminRoleModal.tsx
jonathan-marmelab: This is requested in the card, as the current user is no longer an administrator once they transfered the role

examples/crm/src/sales/TransferAdminRoleModal.tsx
jonathan-marmelab: Since the current user is no longer an administrator once they transfered the role, I am not sure that I can only refresh the page, I need to refresh its permissions as well

examples/crm/src/sales/useIsAdmin.tsx
jonathan-marmelab: This was removed by using permissions instead

examples/crm/src/sales/SalesCreate.tsx
```diff
@@ -0,0 +1,13 @@
+import { Create, PasswordInput, required, SimpleForm } from 'react-admin';
+import { SalesForm } from './SalesForm';
+
+export function SalesCreate() {
+    return (
+        <Create>
```
arimet, line 6: **question:** ‏We redirect on edit page after creation, it's on purpose ?

PR #10038, opened 2024-07-22
Update SelectInput to support an array of strings as choices

fzaninotto:
## Problem

For `SelectInput` and other selection inputs, sometimes the list of options is simply an array of strings. Doing the conversion to an array of option objects isn't intuitive. 

## Solution

Support string choices by default. 

```jsx
<SelectInput source="gender" choices={['Male', 'Female']} />
```

This concerns the following components:

- `SelectInput`
- `SelectArrayInput`
- `AutocompleteInput`
- `AutocompleteArrayInput`
- `RadioButtonGroupInput`
- `CheckboxGroupInput`

### DIFF
--- a/packages/ra-core/src/form/choices/useChoicesContext.ts
+++ b/packages/ra-core/src/form/choices/useChoicesContext.ts
@@ -4,14 +4,20 @@ import { useList } from '../../controller';
 import { ChoicesContext, ChoicesContextValue } from './ChoicesContext';
 
 export const useChoicesContext = <ChoicesType extends RaRecord = RaRecord>(
-    options: Partial<ChoicesContextValue> & { choices?: ChoicesType[] } = {}
+    options: Partial<ChoicesContextValue> & {
+        choices?: ChoicesType[];
+    } = {}
 ): ChoicesContextValue<ChoicesType> => {
     const context = useContext(
         ChoicesContext
     ) as ChoicesContextValue<ChoicesType>;
+    const choices =
+        options.choices && isArrayOfStrings(options.choices)
+            ? convertOptionsToChoices(options.choices)
+            : options.choices;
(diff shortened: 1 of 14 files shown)


packages/ra-ui-materialui/src/input/SelectInput.spec.tsx
```diff
@@ -130,6 +131,14 @@ describe('<SelectInput />', () => {
             expect(options.length).toEqual(6);
             expect(options[1].textContent).toEqual('Created');
         });
+
+        it('should accept strings as choices', () => {
```
fzaninotto, line 135: I only tested SelectInput because the logic is shared by all selection inputs via `useChoicesContext`.

packages/ra-core/src/form/choices/useChoicesContext.ts
```diff
@@ -4,14 +4,20 @@ import { useList } from '../../controller';
 import { ChoicesContext, ChoicesContextValue } from './ChoicesContext';
 
 export const useChoicesContext = <ChoicesType extends RaRecord = RaRecord>(
```
djhi, line 6: I'm surprised this doesn't trigger a TS error as `RaRecord` can't be a `string`. I believe this is because `ChoicesProps` in `useChoices` accepts `any[]` for `choices`. This might be an issue later

fzaninotto: We can't type the choices as users may pass anything and set the `optionText` and `optionValue`. It's never been an issue ;)

djhi: We should not restrict the `ChoicesType` then but let's do that in another PR

djhi (approved):

PR #10047, opened 2024-07-23
Feat(crm): Add ability to get back to show from deal edit and on save

jonathan-marmelab:
[Capture vidéo du 23-07-2024 16:46:07.webm](https://github.com/user-attachments/assets/917c97f0-edd2-48b9-86eb-c0cefae688a6)


arimet (changes requested):
**issue:** Dialog for creating Deal not working anymore.  When we want to add, it's also trying to display the DealEdit with an id = 'create' :thinking: 
‏

arimet (approved):

PR #10054, opened 2024-07-23
[Doc] Mention `CreateInDialog` in `ReferenceManyField` doc

erwanMarmelab:
## Problem

When rendering a list of related records in a ReferenceManyField, editing and creating a related element isn’t straightforward. One can use CreateButton, but the user experience is clunky.

## Solution

Add a “Adding a related record” section, showing the use of CreateInDialog button.

cf [https://react-admin.github.io/ra-enterprise/?path=/story/ra-form-layout-dialogform-createindialogbutton--in-reference-field](https://react-admin.github.io/ra-enterprise/?path=/story/ra-form-layout-dialogform-createindialogbutton--in-reference-field "smartCard-inline")

## Screenshot

image

### DIFF
--- a/docs/ReferenceManyField.md
+++ b/docs/ReferenceManyField.md
@@ -346,3 +346,45 @@ In these cases, use [the `<ReferenceOneField>` component](./ReferenceOneField.md
 </ReferenceOneField>
 ```
 {% endraw %}
+
+## Adding a related record
+
+To allow users to create a record without leaving the current view, use the [`<CreateInDialogButton>`](./CreateInDialogButton.md) component.
+
+{% raw %}
+```jsx
+import { Edit, SimpleForm, TextInput, ReferenceManyField, WithRecord, Datagrid } from 'react-admin';
+import { CreateInDialogButton } from "@react-admin/ra-form-layout";
+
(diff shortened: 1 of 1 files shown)


docs/ReferenceManyField.md
erwanMarmelab, line 1: Do we want an other section with an example of `<EditInDialogButton>` too ?

djhi (approved):

docs/ReferenceManyField.md
```diff
@@ -346,3 +346,53 @@ In these cases, use [the `<ReferenceOneField>` component](./ReferenceOneField.md
...
+                          </SimpleForm>
+                      </CreateInDialogButton>
+                  )}
+              />
+              <Datagrid>
+                  <TextField source="first_name" />
+                  <TextField source="last_name" />
+                  <EditInDialogButton>
+                        <SimpleForm
+                          record={{ employer_id: record.id }}
```
djhi, line 386: I think you're missing a `WithRecord` here

djhi (changes requested):

djhi (approved):

PR #10057, opened 2024-07-23
Feat(crm): Display total of each columns in DealList

arimet:
[Screencast from 2024-07-23 22-28-01.webm](https://github.com/user-attachments/assets/53f67fc3-69bb-45ac-af3d-e65985b4e37b)


examples/crm/src/deals/DealColumn.tsx
```diff
@@ -11,49 +11,72 @@ export const DealColumn = ({
...
-                        <DealCard key={deal.id} deal={deal} index={index} />
-                    ))}
-                    {droppableProvided.placeholder}
-                </Box>
-            )}
-        </Droppable>
-    </Box>
-);
+                    (
+                    {totalAmount.toLocaleString('en-US', {
```
jonathan-marmelab, line 45: **non-blocking(nit-pick)**: Maybe this should be configurable in a config file or based on user's brower preference

jonathan-marmelab (approved):

PR #10062, opened 2024-07-24
Feat(crm): Adds a redirection to home from the application logo

arimet:
## Problem

Some users don't know how to get back to the reception area

## Solution

Ajoute une redirection vers la home depuis le logo de l'application

[Screencast from 2024-07-24 21-04-15.webm](https://github.com/user-attachments/assets/c1c39acf-ad57-40ef-8190-0eab79e8a938)

### DIFF
--- a/examples/crm/src/Header.tsx
+++ b/examples/crm/src/Header.tsx
@@ -15,6 +15,7 @@ import {
     LoadingIndicator,
     Logout,
     usePermissions,
+    useRedirect,
     UserMenu,
     useUserMenu,
 } from 'react-admin';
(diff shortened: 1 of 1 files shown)


examples/crm/src/Header.tsx
```diff
@@ -45,10 +46,18 @@ const Header = () => {
             <AppBar position="static" color="primary">
                 <Toolbar variant="dense">
                     <Box flex={1} display="flex" justifyContent="space-between">
-                        <Box display="flex" alignItems="center">
+                        <Box
+                            display="flex"
+                            alignItems="center"
+                            onClick={() => redirect('/')}
```
fzaninotto, line 52: If you use an `onClick` handler, the target doesn't appear when hovering the link. Onstead, use react-router's `Link` component. It will also change the mouse pointer automatically. 

Reserve navigation with click handlers for conditional navigation.

fzaninotto (changes requested):

jonathan-marmelab (approved):

PR #10070, opened 2024-07-25
Feat(crm): Hamonize forms

arimet:
## Problem

Sometimes “Save” and “Delete” buttons are reverse.

## Solution

Harmonize all forms with primary “Save” button on the left and secondary “Delete” on the right
Add a cross to close all popup

## Additional Checks

- [x] Harmonize Deal Forms.

[Screencast from 2024-07-25 14-35-19.webm](https://github.com/user-attachments/assets/6608d488-2ffb-4bd9-a74d-33bfb281cc51)

### DIFF
--- a/examples/crm/src/deals/DealEdit.tsx
+++ b/examples/crm/src/deals/DealEdit.tsx
@@ -13,7 +13,7 @@ import {
 import { Link } from 'react-router-dom';
 import { Deal } from '../types';
 import { DealInputs } from './DealInputs';
-import { DialogContent } from '@mui/material';
+import { DialogContent, Stack, Button } from '@mui/material';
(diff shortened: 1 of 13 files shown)


examples/crm/src/deals/DealEdit.tsx
```diff
@@ -46,31 +50,17 @@ function EditHeader() {
...
-            }}
-        >
-            <Stack
-                direction="row"
-                alignItems="center"
-                justifyContent="space-between"
-                spacing={1}
+        <DialogTitle>
+            Edit{' '}
+            <Typography
```
jonathan-marmelab, line 55: **thought:** I'm not sure that this is a great UX to hide back to show in a link without any other information, but I do not have a better idea for this problem

jonathan-marmelab (approved):

PR #10071, opened 2024-07-25
Fix/activity log

jonathan-marmelab:
## Problem

Company activity log was not loaded as fake rest does not use hooks initially

## Solution

Removed hooks and loaded activity loag from rest endpoints

## How To Test

Go to any company > activity log

[Capture vidéo du 25-07-2024 14:38:56.webm](https://github.com/user-attachments/assets/d2e5aadc-2675-4b3f-b421-feaf749b717f)


examples/crm/src/activity/ActivityLogIterator.tsx
```diff
@@ -0,0 +1,83 @@
...
+    activities: Activity[];
+};
+
+export function ActivityLogIterator({ activities }: ActivityLogIteratorProps) {
+    const [activitiesDisplayed, setActivityDisplayed] = useState(PAGE_SIZE);
+
+    const filteredActivities = activities.slice(0, activitiesDisplayed);
+
+    return (
+        <List
```
arimet, line 29: **note (non-blocking):** ‏I Think we need more spacing between activities

examples/crm/tsconfig.json
```diff
@@ -1,28 +1,22 @@
...
-    "noEmit": true,
-    "jsx": "react-jsx",
-    "strictNullChecks": true
-  },
-  "include": [
-    "src"
-  ],
-  "references": [{ "path": "./tsconfig.node.json" }]
+    "compilerOptions": {
+        "target": "ES2019",
```
arimet, line 3: **question:** ‏Is that okay with the RA team?

examples/crm/src/dataProvider/activity.ts
```diff
@@ -0,0 +1,192 @@
...
+        getDealsLog(dataProvider, sales, companies),
+    ]);
+
+    return companies.companiesLog.concat(contactsLog, dealsLog).sort(
+        (a, b) => a.date.localeCompare(b.date) * -1 // sort by date desc
+    );
+}
+
+async function getSales(dataProvider: DataProvider) {
+    const salesDict = await dataProvider
```
arimet, line 37: **nitpick (non-blocking):** More understandable variable name?‏

arimet (commented):
**note (non-blocking):** I have the impression that certain things could be factorized. But how can this be done without losing comprehension? :thinking: ‏

examples/crm/tsconfig.json
jonathan-marmelab: cc @djhi @adguernier, is it ok if we up the ES target for the CRM ?

examples/crm/src/activity/ActivityLogIterator.tsx
jonathan-marmelab: Validated by RA team

arimet (approved):

PR #10079, opened 2024-07-25
Feat(crm): Display Contact inputs by category in columns and improve UX form

arimet:
## Problem

 As a user, I want to reduce the number of clicks and understand the parts of the contact form at a glance

## Solution

- [x] Display inputs by category in columns and give the category a title.
- [x] Replace the “Gender” selector into radio buttons to reduce the number of clicks + remove “Gender” label
- [x] Add a "Type" selector next to each "Phone number" inputs with the values "Work", "Home" and "Other". "Work" is default value.
- [x] When the form is submitted, update the sidebar by displaying "Work", "Home" or nothing if the value is "Other" next to the telephone number.

## Screenshots

Before
image

After
image

### DIFF
--- a/examples/crm/src/contacts/ContactAside.tsx
+++ b/examples/crm/src/contacts/ContactAside.tsx
@@ -66,33 +66,31 @@ export const ContactAside = ({ link = 'edit' }: { link?: 'edit' | 'show' }) => {
                     />
                 </Stack>
             )}
-            {record.phone_number1 && (
+            {record.phone_number1.number && (
                 <Stack direction="row" alignItems="center" gap={1}>
                     <PhoneIcon color="disabled" fontSize="small" />
                     <Box>
-                        <TextField source="phone_number1" />{' '}
-                        <Typography
-                            variant="body2"
-                            color="textSecondary"
-                            component="span"
-                        >
-                            Work
-                        </Typography>
+                        <TextField source="phone_number1.number" />{' '}
+                        {record.phone_number1.type !== 'Other' && (
+                            <TextField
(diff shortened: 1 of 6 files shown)


examples/crm/src/contacts/ContactInputs.tsx
```diff
@@ -29,7 +38,64 @@ const isLinkedinUrl = (url: string) => {
 };
 
 export const ContactInputs = () => {
+    const theme = useTheme();
+    const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
```
jonathan-marmelab, line 42: **nit-pick:** I would have gone for `md` break point here to detect mobile, otherwise the phone number field is too tight

examples/crm/src/contacts/ContactAside.tsx
```diff
@@ -66,33 +66,31 @@ export const ContactAside = ({ link = 'edit' }: { link?: 'edit' | 'show' }) => {
                     />
                 </Stack>
             )}
-            {record.phone_number1 && (
+            {record.phone_number1.number && (
```
jonathan-marmelab, line 69: **question:** Will this be OK when we will use supabase to have a nested field ?

jonathan-marmelab (changes requested):

examples/crm/src/contacts/ContactAside.tsx
arimet: Supabase normally supports unstructured data. So I don't think there will be any problems. At worst, we'll change it then

PR #10083, opened 2024-07-26
Fix(crm): Replace see more button with a link in activity log

jonathan-marmelab:
## Problem

Uses can make a mistake between see more and load more buttons on activity log

## Solution

Replace see more button to a see more link

## How To Test

Go to activity log on dashboard or on any company

### DIFF
--- a/examples/crm/src/activity/ActivityLogNote.tsx
+++ b/examples/crm/src/activity/ActivityLogNote.tsx
@@ -1,8 +1,8 @@
-import Button from '@mui/material/Button';
 import ListItem from '@mui/material/ListItem';
 import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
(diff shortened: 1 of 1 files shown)


examples/crm/src/activity/ActivityLogNote.tsx
```diff
@@ -1,8 +1,8 @@
-import Button from '@mui/material/Button';
 import ListItem from '@mui/material/ListItem';
 import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
-import { ReactNode, useState } from 'react';
+import { MouseEventHandler, ReactNode, useState } from 'react';
+import { Link } from 'react-admin';
```
arimet, line 5: **note (non-blocking):** ‏It shouldn't be a Mui Link ?

arimet (approved):

PR #10096, opened 2024-07-26
 Fix(crm): Add create contact button to company even if there is no contacts attached to it

jonathan-marmelab:
## Problem

It is hard to create a contact to the company when it has been added since the contact tab is hidden

## Solution

Show the contact tab even if no contact is in the company

## How To Test

Create a new company and and add a contact from the company page

Capture d’écran du 2024-07-26 15-52-43


examples/crm/src/dataProvider.ts
```diff
@@ -450,6 +484,23 @@ export const dataProvider = withLifecycleCallbacks(
                     },
                 };
             },
+            afterDelete: async (result, dataProvider) => {
+                const { data: company } = await dataProvider.getOne<Company>(
```
arimet, line 488: **nitpick (non-blocking):** ‏Could be a shared function ?

arimet (approved):

arimet (approved):

PR #10098, opened 2024-07-28
Feat(crm): Implement a component for cropping image

arimet:
[Screencast from 2024-07-28 23-14-42.webm](https://github.com/user-attachments/assets/7273ca6d-6428-45b8-8840-9f78b29a58b4)


examples/crm/src/settings/SettingsPage.tsx
```diff
@@ -80,13 +81,46 @@ const SettingsForm = ({
...
+            'sales',
+            {
+                id: identity.id,
+                data: values,
+                previousData: identity,
+            },
+            {
+                onSuccess: data => {
+                    // Update local user
+                    localStorage.setItem(
```
jonathan-marmelab, line 107: **suggestion:** The local user update is (almost) the same has `SettingsPage#handleOnSubmit`, can you factorize them

examples/crm/src/misc/ImageEditorField.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { FieldProps, Toolbar } from 'react-admin';
+import { Cropper, ReactCropperElement } from 'react-cropper';
+import { useDropzone } from 'react-dropzone';
+import { useFormContext } from 'react-hook-form';
+import { DialogCloseButton } from './DialogCloseButton';
+
+const ImageEditorField = (props: ImageEditorFieldProps) => {
+    const { getValues } = useFormContext();
+    const imageUrl = getValues()[props.source];
+    const [isHovered, setIsHovered] = useState(false);
```
jonathan-marmelab, line 24: **question: ** I think that you want to manage the `change` link on hover using state, isn't it ? If it is the case, maybe you can use a css `:hover` instead ?

examples/crm/src/settings/SettingsPage.tsx
arimet: I think it's overkill, I'll see if it's used somewhere else.

jonathan-marmelab (approved):

Issue #10100, opened 2024-07-29
rowClick is called automatically


7system7:
**What you were expecting:**
When I click on a `Datagrid`'s row, then the `rowClick` callback should be called. So, it should be called only on an `onClick` event.

**What happened instead:**
After upgrading to v5, and if I define the the `rowClick` attribute explicitly as a callback fn, then it is automatically called on `Datagrid`.

**Steps to reproduce:**
1. Install a react-admin. ` $ npx create-react-admin@latest your-admin-name`
1. Create a list w/ the attribute `rowClick` i.e.
    ```javascript
    ...
    <List>
        <Datagrid rowClick={(id, resource, record) => console.log(id)}>
    ...
    ```
1. It will be called as much as row the `Datagrid` has.

**Environment**

* React-admin version: >= 5.0.0
* Last version that did not exhibit the issue (if applicable): 4.16.20
* React version: 18.3.1
* Browser: Chromium, Firefox, Safari etc.
* Stack trace (in case of a JS error): -

djhi:
Reproduced. That's indeed a regression

fzaninotto:
The regression was caused by #9951

fzaninotto:
Fixed by #10102

PR #10102, opened 2024-07-29
Fix `<Datagrid rowClick>` regression

djhi:
## Problem

#10100

## Additional Checks

### DIFF
--- a/packages/ra-core/src/routing/useGetPathForRecord.ts
+++ b/packages/ra-core/src/routing/useGetPathForRecord.ts
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback } from 'react';
-import { useResourceContext, useResourceDefinition } from '../core';
+import { useResourceContext, useResourceDefinitions } from '../core';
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -49,82 +49,112 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
+        }).then(resolvedLink => {
+            // update the path when the promise resolves
+            setPath(resolvedLink);
+        });
+    }, [getPathForRecord, link, record, resource]);
 
     return path;
 };
 
+export const useGetRouteForRecordCallback = <
```
fzaninotto, line 87: I'd call this function useGetRouteForRecordCallback and extract it to another file (and export it)

fzaninotto (changes requested):
Thanks! The code is easier to read with this change 👍

erwanMarmelab (approved):

fzaninotto (approved):

PR #10103, opened 2024-07-29
Feat(crm): Redirect on company page when deleting an user from company

arimet:
## Problem

As a user, I want to delete a contact from the company page easily

## Solution

Add a Delete button at the end of the sidebar of a contact

When the user deletes the contact, and that it comes from the company page, redirect it on the company page else redirect it on the contacts list

Remove “Delete” button on the contact form which is now no longer so necessary

[Screencast from 2024-07-29 17-58-48.webm](https://github.com/user-attachments/assets/bc0d8071-ed96-4401-b52a-5bbc5c757677)

### DIFF
--- a/examples/crm/src/contacts/ContactAside.tsx
+++ b/examples/crm/src/contacts/ContactAside.tsx
@@ -4,6 +4,7 @@ import PhoneIcon from '@mui/icons-material/Phone';
 import { Box, Divider, Stack, Typography } from '@mui/material';
 import {
     DateField,
+    DeleteButton,
     EditButton,
     EmailField,
     FunctionField,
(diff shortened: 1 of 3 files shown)


examples/crm/src/contacts/ContactAside.tsx
```diff
@@ -167,6 +170,7 @@ export const ContactAside = ({ link = 'edit' }: { link?: 'edit' | 'show' }) => {
                 </ReferenceManyField>
                 <AddTask />
             </Box>
+            <DeleteButton redirect={location.state?.from || undefined} />
```
jonathan-marmelab, line 173: **suggestion:** Use a `DeleteWithConfirmButton` instead to confirm debore deletion

jonathan-marmelab (changes requested):

examples/crm/src/contacts/ContactAside.tsx
arimet: The undoable is used everywhere else, so for me it's important to remain consistent.

PR #10104, opened 2024-07-29
Feat(crm): Allow quick task create from Dashboard

arimet:
## Problem

As a user, I don’t know how create a task

## Solution

Add a button secondary “New task” on the homepage that opens the task creation popup with a contacts selector.

Contacts selector is required.

Contacts selector shouldn’t appears on the other popup (edit, creation from the contacts page)

[Screencast from 2024-07-29 19-11-23.webm](https://github.com/user-attachments/assets/ad892a20-fe26-4995-b80d-f61546e32a5e)

### DIFF
--- a/examples/crm/src/tasks/AddTask.tsx
+++ b/examples/crm/src/tasks/AddTask.tsx
@@ -11,10 +11,12 @@ import {
 import * as React from 'react';
 import { useState } from 'react';
 import {
+    AutocompleteInput,
     CreateBase,
     DateInput,
     Form,
     RecordRepresentation,
+    ReferenceInput,
     SaveButton,
     SelectInput,
     TextInput,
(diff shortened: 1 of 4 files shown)


examples/crm/src/tasks/AddTask.tsx
```diff
@@ -79,6 +86,21 @@ export const AddTask = () => {
                                 multiline
                                 helperText={false}
                             />
+                            {selectContact && (
+                                <ReferenceInput
+                                    source="contact_id"
+                                    reference="contacts"
+                                >
+                                    <AutocompleteInput
+                                        label="Contacts"
```
jonathan-marmelab, line 95:
```suggestion
                                        label="Contact"
```

jonathan-marmelab (changes requested):

jonathan-marmelab (approved):

PR #10112, opened 2024-07-31
Feat(crm): Remove transfor role + new password for sales

arimet:
## Problem

Transfer admin role isn't required anymore and an adminstrator should not update sale password.

Screenshot from 2024-07-31 12-08-28


jonathan-marmelab (changes requested):
**issue:** You forgot to remove the data provider method as it is not used anymore

jonathan-marmelab (approved):

PR #10113, opened 2024-07-31
Feat(crm): Keep company tab when changing  page

arimet:
## Problem

As a user, when I navigate through the company tabs, change pages and return to the previous page, I can't get to the previous tab

## Solution

Set tabValue in Localstorage
If you have a better solution, I'll take it. 
I wanted to do it via react-router, but couldn't find a clean way. 

[Screencast from 2024-08-01 10-06-55.webm](https://github.com/user-attachments/assets/be81faf6-2345-4e8e-a925-2313cb4a97aa)


jonathan-marmelab (commented):
**thought:** I'm not sure that this implementation matches what was expected in the card. If you do multiple history back, I'm not sure that this would work.

Can you implement the same feature using routes instead to provide full history compatibility ?

fzaninotto:
I made a quick fix to preserve the margins. But the PR isn't mergeable - there is a conflict.

fzaninotto (approved):

PR #10122, opened 2024-08-01
Feat(crm): Soft delete for Contact and Sale

arimet:


jonathan-marmelab (commented):
You can access the contact's page from the activity log once it has been deleted, but the page is empty. Can you add a message for this case (same for sales)

When you delete a contact or a sale, it is still present in the list for a short delay, can you fix that ?

Do we need to be able to restore a user or contact ?

PR #10123, opened 2024-08-01
Feat(crm): Add possibility to disable Sale

arimet:
## Problem

As a user, I lose deleted user activity

## Solution

Add a disabled boolean for Sale users. 

image

### DIFF
--- a/examples/crm/src/dataProvider.ts
+++ b/examples/crm/src/dataProvider.ts
@@ -89,7 +89,7 @@ const dataProviderWithCustomMethod = {
         }
 
         const sale = sales.data.find(sale => sale.email === email);
-        if (!sale) {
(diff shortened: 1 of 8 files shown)


examples/crm/src/dataProvider.ts
```diff
@@ -89,7 +89,7 @@ const dataProviderWithCustomMethod = {
         }
 
         const sale = sales.data.find(sale => sale.email === email);
-        if (!sale) {
+        if (!sale || sale.disabled) {
```
jonathan-marmelab, line 92: **note:** We will need to add an error message here if the account is disbaled in the supabase version

jonathan-marmelab (approved):

PR #10126, opened 2024-08-05
Feat(crm): Set isImage into dataProvider

arimet:
Since the merge, the images for the attachments are no longer necessarily visible.

## Solution

Fix the function and place it in the dataProvider for easy rewriting

### DIFF
--- a/examples/crm/src/dataProvider.ts
+++ b/examples/crm/src/dataProvider.ts
@@ -152,19 +152,6 @@ const dataProviderWithCustomMethod = {
     getActivityLog: async (companyId?: Identifier) => {
         return getActivityLog(baseDataProvider, companyId);
     },
-    isImage(attachment: AttachmentNote): Promise<boolean> {
-        return new Promise((resolve, reject) => {
-            try {
-                if (!attachment || !attachment.rawFile) {
-                    resolve(false);
-                } else {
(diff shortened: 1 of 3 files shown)


examples/crm/src/dataProvider.ts
```diff
@@ -143,6 +152,19 @@ const dataProviderWithCustomMethod = {
...
+                if (!attachment || !attachment.rawFile) {
+                    resolve(false);
+                } else {
+                    resolve(attachment.rawFile.type.startsWith('image/'));
+                }
+            } catch (error) {
+                reject(error);
+            }
+        });
+    },
```
slax57, line 167: This fn is no longer necessary I believe

slax57 (changes requested):

examples/crm/src/dataProvider.ts
```diff
@@ -12,7 +12,16 @@ import generateData from './dataGenerator';
 import { getActivityLog } from './dataProvider/activity';
 import { getCompanyAvatar } from './misc/getCompanyAvatar';
 import { getContactAvatar } from './misc/getContactAvatar';
-import { Company, Contact, Deal, Sale, Task } from './types';
+import {
+    AttachmentNote,
```
slax57, line 16: Linter warning left :grimacing:

PR #10148, opened 2024-08-19
Fix: `useInput` should call a custom validator with the final source in respect to the `<SourceContext>`

slax57:
## Problem

When passing a [custom validator function](https://marmelab.com/react-admin/Validation.html#per-input-validation-custom-function-validator) to an input, the `source` prop it recieves is the original `source` prop declared on the input, instead of the 'final' source computed using the `<SourceContext>`.

This is an issue when the input is inside a component wrapping it with a custom `<SourceContext>`, such as `<SimpleFormIterator>`.

Example:

```tsx
<ArrayInput source="posts">
  <SimpleFormIterator>
    <TextInput source="title" validate={validator} />
  </SimpleFormIterator>
</ArrayInput>
```

In that case, `validator` will be called like so: `validator("value", { posts: [{ title: "value" }] }, { source: "title", /* ... */ })`
Although it should have been: `validator("value", { posts: [{ title: "value" }] }, { source: "posts.0.title", /* ... */ })`

## Solution

Use the `<SourceContext>` and pass a prop called `finalSource` when calling the custom validator in addition to the original `source` prop.

`validator("value", { posts: [{ title: "value" }] }, { source: "title", finalSource: "posts.0.title", /* ... */ })`

## How To Test

Run unit tests.
Alternatively, use a custom validator inside an `<ArrayInput>` in one of the demos.

## Additional Checks

- [ ] The PR includes one or several **stories** (if not possible, describe why) <- did not find it useful for such a technical case

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -90,7 +90,7 @@ export const useInput = <ValueType = any>(
                 if (!sanitizedValidate) return true;
                 const error = await sanitizedValidate(value, values, {
                     ...props,
-                    source: finalSource,
+                    finalSource,
                 });
 
                 if (!error) return true;
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -88,7 +88,10 @@ export const useInput = <ValueType = any>(
         rules: {
             validate: async (value, values) => {
                 if (!sanitizedValidate) return true;
-                const error = await sanitizedValidate(value, values, props);
+                const error = await sanitizedValidate(value, values, {
+                    ...props,
+                    source: finalSource,
+                });
 
                 if (!error) return true;
                 // react-hook-form expects errors to be plain strings but our validators can return objects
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/form/useInput.ts
```diff
@@ -88,7 +88,10 @@ export const useInput = <ValueType = any>(
         rules: {
             validate: async (value, values) => {
                 if (!sanitizedValidate) return true;
-                const error = await sanitizedValidate(value, values, props);
+                const error = await sanitizedValidate(value, values, {
+                    ...props,
+                    source: finalSource,
+                });
```
slax57, line 94: **Open question:** I wonder if it's best to _override_ the `source` prop, or to pass an additional `finalSource` prop. The [docs](https://marmelab.com/react-admin/Validation.html#per-input-validation-custom-function-validator) simply say that the signature should be `(value, allValues, props)`, so technically we could choose either solution.
What do you guys think?

fzaninotto: I'm +1 for passing both `source` and `finalSource`

slax57: done

fzaninotto (approved):

PR #10153, opened 2024-08-20
Avoid "no source" warning when theres a finalSource in useInput

GuilhermeCarra:
## Problem

Currently `useInput` is generating erroneous warnings when using `ArrayInput` with `TextInput` without source. The source is provided by `useWrappedSource`

```
        <ArrayInput source="users">
              <SimpleFormIterator>
                  <TextInput source="" />
              </SimpleFormIterator>
        </ArrayInput>
```

## Solution

Change conditional check to finalSource

## How To Test

Create an ArrayInput as indicated

## Additional Checks

- [ ] The PR includes unit tests (if not possible, describe why)
- [ ] The PR includes one or several stories (if not possible, describe why) (N/A)
- [ ] The documentation is up to date

### DIFF
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -53,7 +53,7 @@ export const useInput = <ValueType = any>(
         process.env.NODE_ENV === 'development'
     ) {
         console.warn(
-            'Input components require either a source or a label prop.'
+            'Input components require either a source or a name prop.'
         );
     }
 
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -49,7 +49,7 @@ export const useInput = <ValueType = any>(
     const defaultId = useId();
 
     if (
-        !source &&
+        !finalSource &&
         props.label == null &&
         process.env.NODE_ENV === 'development'
     ) {
--- a/packages/ra-core/src/form/useInput.ts
+++ b/packages/ra-core/src/form/useInput.ts
@@ -48,10 +48,7 @@ export const useInput = <ValueType = any>(
     const parse = useEvent(parseProp);
     const defaultId = useId();
 
-    if (
-        !finalName &&
-        process.env.NODE_ENV === 'development'
-    ) {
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/form/useInput.ts
```diff
@@ -49,7 +49,7 @@ export const useInput = <ValueType = any>(
     const defaultId = useId();
 
     if (
-        !source &&
+        !finalSource &&
         props.label == null &&
         process.env.NODE_ENV === 'development'
     ) {
```
slax57, line 55: I agree we should change this condition, thanks for your report!

Actually I wonder if we should go even further. AFAICT this check was mainly added to make sure we won't call [useController](https://react-hook-form.com/docs/usecontroller) with an empty `name` (and also, to make TS happy).

So we should maybe replace it altogether by:

```js
    if (
        !finalName &&
        process.env.NODE_ENV === 'development'
    ) { /* ... */ }
```

@fzaninotto wdyt?

fzaninotto: You're right, an input with no source but with a name should not raise any warning!

GuilhermeCarra: @slax57 @fzaninotto thanks for your feedback! Are you guys going to work on this? I saw the changes las week in [useInput](https://github.com/marmelab/react-admin/blame/master/packages/ra-core/src/form/useInput.ts) but I'm still getting the warnings which is making my application very slow when editing/creating
Thanks!

slax57: @GuilhermeCarra the changes you refer to are unrelated (you can have a look at the originating [PR](https://github.com/marmelab/react-admin/pull/10148) description to understand the context). 

Regarding this PR, if you apply the changes I requested [above](https://github.com/marmelab/react-admin/pull/10153#discussion_r1723486408) we will be happy to merge it. Thanks!

GuilhermeCarra: @slax57 thanks! I've made the change 🙌

fzaninotto (approved):

fzaninotto:
Thanks!

PR #10157, opened 2024-08-26
fix(DatagridConfigurable): better detect and handle columns changes

quentin-decre:
## Problem

When you change the columns order of a ConfigurableDatagrid, or just if you remove a column and add another one, this is not managed and the column management is broken : aka the displayed columns in the configure list does not control the corresponding columns.

Here is a quick illustration of the issue : https://www.loom.com/share/c5a6afaac36a4111bedf08ece818da17 (in french, sorry)

That's because currently, only a change in the number of columns trigger and update. And even in this case, you may have unexpected results (like the column enabled to be displayed are not the same as before).

## Solution

This changes bring :
* **better detection** : now, if you change the source of a column, add or remove a column (or both), it triggers the update
* **keeps your previous settings** : it keeps the columns you displayed before
* **keeps your order** : it keeps the columns displayed in the order you have set
* **enable non omited new columns** : if you added new columns, and not omited them, they will be displayed at the end

## How To Test

I have tested it multiple times on the following scenarios :
* change the source of a column
* add a column and remove another one at the same time
* change the order of the columns passed to the datagrid
* add a new non omited columns

I have ran the tests for non regression
image

Also, this PR is live on prod for us now. And this is a very used feature among our users :)

## Additional Checks

- [X] The PR targets `master` for a bugfix, or `next` for a feature
- [ ] The PR includes **unit tests** (if not possible, describe why) : not new feature. tests still pass
- [ ] The PR includes one or several **stories** (if not possible, describe why) : not a new feature. stories are still relevant
- [ ] The **documentation** is up to date : no change needed to the documentation

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.tsx
@@ -1,10 +1,10 @@
-import * as React from 'react';
 import {
-    useResourceContext,
     usePreference,
+    useResourceContext,
     useStore,
     useTranslate,
 } from 'ra-core';
+import * as React from 'react';
 
 import { Configurable } from '../../preferences';
 import { Datagrid, DatagridProps } from './Datagrid';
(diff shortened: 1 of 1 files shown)


slax57 (changes requested):
Thanks for the PR!
It looks promising :slightly_smiling_face: 
Could you add some unit tests to prove that it works, including in corner cases, and does not break compatibility?
Thanks!

slax57:
I have added some example unit tests, showing how you can check what happens when the code is changed, and how to check what's in the store.
Hope this helps!

fzaninotto:
@slax57's unit tests have disappeared in the force push... And this PR has gone stale. Besides, we now recommend using DataTable instead of DatagridConfigurable.

PR #10162, opened 2024-08-26
Add support for computed fields in `ra-data-json-server`'s `dataProvider.create()`

### LINKED ISSUE #9979: ra-data-json-server: Create request ignores response content
https://github.com/marmelab/react-admin/blob/45151e0b844a83dc8309a56ad9ca8ea355ce3aee/packages/ra-data-json-server/src/index.ts#L154

What is the reasoning for this line? Any server-side logic to the data is ignored and cannot be accounted for as part of the response.

I would expect this to be:
`            data: { ...json },`
or 
`            data: { ...params.data, ...json } as any,`

fzaninotto:
Closes #9979

### DIFF
--- a/packages/ra-data-json-server/src/index.ts
+++ b/packages/ra-data-json-server/src/index.ts
@@ -151,7 +151,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson): DataProvider => ({
             method: 'POST',
             body: JSON.stringify(params.data),
         }).then(({ json }) => ({
-            data: { ...params.data, id: json.id } as any,
+            data: { ...params.data, ...json } as any,
         })),
 
     delete: (resource, params) =>


packages/ra-data-json-server/src/index.ts
```diff
@@ -151,7 +151,7 @@ export default (apiUrl, httpClient = fetchUtils.fetchJson): DataProvider => ({
             method: 'POST',
             body: JSON.stringify(params.data),
         }).then(({ json }) => ({
-            data: { ...params.data, id: json.id } as any,
+            data: { ...params.data, ...json } as any,
```
ThieryMichel, line 154: Should'nt we add a test?

ThieryMichel (changes requested):

packages/ra-data-json-server/src/index.ts
slax57: IMO it's not necessary for something so trivial

slax57 (approved):

PR #10163, opened 2024-08-26
Fix input validation and dirty state after re-enabling disabled fields

### LINKED ISSUE #9870: Initially disabled inputs are not validated after enabled when form is submitted
**What you were expecting:**

I have a form with 2 inputs, both inputs are required but the second is disabled if there is no value for the first input, after I provide a value for the first input the second input becomes enabled but if I submit the form without providing a value for second input the form submission should failed because the required() should fail,
**What happened instead:**

The form was submitted and ignored the fact that there is no value on a required() validation input.
**Steps to reproduce:**

1.Go to create post,
Fill in the required fields but leave the average_note blank, it is required() and minimum 1 but it can submit the form without providing value for it

**Related code:**

* Preferably, a sandbox forked from 
https://stackblitz.com/edit/github-xm5t9q?file=src%2Fposts%2FPostCreate.tsx
```
**Environment**

* React-admin version:4.16.18
* React version:17.0
* Browser:Chrome

### LINKED ISSUE #10156: Disabling and enabling form Inputs breaks their isDirty state
**What you were expecting:**
When using a form input that is conditionally enabled/disabled by a switch:
- Changing the input's value triggers validation and marks the field as dirty (i.e., `isDirty` becomes `true`).
- Disabling the input should reset `isDirty` to `false` since validation shouldn't run on disabled inputs.
- Re-enabling the input should allow it to become dirty again when its value is changed, triggering validation as expected.

**What happened instead:**
After re-enabling a previously disabled input, changing its value does not set `isDirty` to `true`, and validation is not triggered for that field.

**Steps to reproduce:**
1. Create a form with an input field that has validation and can be disabled/enabled based on a switch or another condition.
2. Change the value of the input and observe that validation runs as expected.
3. Disable the input, then re-enable it using the switch.
4. Change the input value again. Notice that validation does not run, and the field does not become dirty.

**Reproduction Link:**
- [Example on StackBlitz](https://stackblitz.com/edit/github-v5xpqg?file=src%2Fposts%2FPostCreate.tsx)

In this example, after toggling the switch, the "title" field can be submitted with an empty value, even though it is marked as required.

**Other information:**
After investigating, I believe this issue is caused by [this PR](https://github.com/marmelab/react-admin/pull/9313/files). When the input is re-enabled, `disabled = undefined` is passed to `react-hook-form`'s controller, but it seems the controller doesn't fully clear the disabled state internally, preventing the field from becoming dirty again.

Does the original issue with `react-hook-form` that the workaround was intended to fix still persist?
It seems like it was fixed [here](https://github.com/react-hook-form/react-hook-form/commit/d7d45c80d50f32b735c9d7574498e0242e130349) maybe we can remove that workaround.

michel-paiva:
## Problem

- Disabled and re-enabled fields did not update their dirty state, causing validation to run after they have been re-enabled.

Related issues:
Closes #10156
Closes #9870 

## Solution

- This fix addresses the issue remove the workaround and making sure that a boolean value is passed to `react-hook-form` controller.

## How To Test

- Verify that input fields with validation are correctly marked as dirty and trigger validation when they are re-enabled after being disabled.

### DIFF
--- a/examples/simple/package.json
+++ b/examples/simple/package.json
@@ -26,7 +26,7 @@
         "react": "^18.3.1",
         "react-admin": "^5.1.2",
         "react-dom": "^18.3.1",
-        "react-hook-form": "^7.52.0",
+        "react-hook-form": "^7.53.0",
         "react-router": "^6.22.0",
         "react-router-dom": "^6.22.0"
     },
--- a/packages/ra-core/package.json
+++ b/packages/ra-core/package.json
@@ -40,7 +40,7 @@
         "jscodeshift": "^0.15.2",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
-        "react-hook-form": "^7.52.0",
+        "react-hook-form": "^7.53.0",
         "react-router": "^6.25.1",
         "react-router-dom": "^6.25.1",
         "react-test-renderer": "^18.2.0",
@@ -53,7 +53,7 @@
     "peerDependencies": {
         "react": "^18.0.0 || ^19.0.0",
         "react-dom": "^18.0.0 || ^19.0.0",
-        "react-hook-form": "^7.52.0",
+        "react-hook-form": "^7.53.0",
         "react-router": "^6.22.0",
         "react-router-dom": "^6.22.0"
     },
--- a/packages/ra-core/src/form/useFormGroup.ts
+++ b/packages/ra-core/src/form/useFormGroup.ts
@@ -69,13 +69,14 @@ export const useFormGroup = (name: string): FormGroupState => {
     const { dirtyFields, touchedFields, validatingFields, errors } =
         useFormState();
 
-    // dirtyFields, touchedFields and validatingFields are objects with keys being the field names
+    // dirtyFields, touchedFields, validatingFields and errors are objects with keys being the field names
     // Ex: { title: true }
     // However, they are not correctly serialized when using JSON.stringify
     // To avoid our effects to not be triggered when they should, we extract the keys and use that as a dependency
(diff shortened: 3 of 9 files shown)


slax57 (approved):
Thank you, Boss. :saluting_face:

slax57:
It looks like one of the unit tests is failing, and this seems to be consequent to the react-hook-form version upgrade. This doesn't seem to be related to the code changed by this PR.
I'm currently investigating it...

fzaninotto:
IMHO, this PR should be against the `next` branch as it upgrades a dependency to a minor version. There *is* a possibility that it breaks existing apps, which would be unexpected in a minor and even more in a bugfix release.

slax57:
Okay, I managed to track down the issue, and filed a new issue on react-hook-form's repo: https://github.com/react-hook-form/react-hook-form/issues/12217

Depending on their answer, we may need to apply the workaround I mentioned to the `useFormGroup.ts` file.

Let's wait for their reply before taking actions.

michel-paiva:
@slax57 seems like they tried to optimize renders by only setting isValidating for inputs that has async validation
https://github.com/react-hook-form/react-hook-form/pull/12192/files

slax57:
@michel-paiva  Okay, so this will probably not get fixed on their side.
That's why we opened https://github.com/marmelab/react-admin/pull/10168 to fix this issue on our side on `master`.

Now, can you please rebase your branch and PR to target the `next` branch?
You may also need to cherry-pick the fix I made to have your PR pass the CI checks.

Thank you.

michel-paiva:
@slax57 done as you said, let's see!

slax57 (approved):

PR #10166, opened 2024-08-27
Add Access Control 

ThieryMichel:
## Problem

Implementing access control in react-admin components (e.g. showing/hiding datagrid columns of buttons depending on the user permissions) is hard because it's not built-in. It's doable by wrapping components in another component that checks permissions before rendering the actual component. But it's not the same component anymore. 

```tsx
// pseudo code
const DatagridWithAccessControl = ({ children, ...rest }) => {
   const authorizedChildren = useAuthorizedChildren(children);
   return <Datagrid {...rest}>{authorizedChildren}</Datagrid>;
}
```

React-admin Enterprise Edition uses this approach to provide alternative components with built-in access control (see https://react-admin-ee.marmelab.com/documentation/ra-rbac). But these components aren't compatible with other overrides (e.g. adding realtime updates, alternative layouts, etc). 

## Solution

Have react-admin components call an access control hook before rendering, leaving the actual implementation to the `authProvider`.

```tsx
// pseudo code
const Datagrid = ({ children, ...rest }) => {
   const authorizedChildren = useAuthorizedChildren(children);
   // ...continue with authorizedChildren
}
```

## TODO

- [x] add useCanAccess hook with tests
- [x] add useCanAccessCallback hook with tests
- [x] add useCanAccessRecordSources hook with tests
- [x] add accessControl in useListController for exports with tests
- [x] add accessConrol in DataGrid with tests
- [ ] List contexts #10188 #10195
- [ ] Resource
- [ ] Edit view
- [ ] Show view
- [ ] Create view
- [ ] Buttons (Create, etc)
- [ ] Optimize useCanAccess hooks to not use react-query at all when no authProvider (return a fake react-query result)

## How To Test

There are stories for each hook and component.

## Additional Checks


packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
```diff
@@ -585,3 +587,71 @@ export const LabelElements = () => (
         </AdminContext>
     </TestMemoryRouter>
 );
+
+export const AccessControl = ({
+    authorizedResources = {
```
ThieryMichel, line 592: Tried to have dynamic authorizedResources, like in useListController.accessControl, but it seems that queryClient.clear does not retrigger query created with useQueries.

djhi: Fixed

docs/useCanAccessCallback.md
```diff
@@ -0,0 +1,61 @@
...
+title: "useCanAccessCallback"
+---
+
+# `useCanAccessCallback`
+
+This hook calls the authProvider.canAccess() method using react-query for a provided resource and action (and optionally a record).
+
+## Usage
+
+`useCanAccessCallback` returns a function that takes an object `{ action, resource, record }` as argument. This function returns an object with the following properties:
```
fzaninotto, line 12:
```suggestion
`useCanAccessCallback` returns a async function that takes an object `{ action, resource, record }` as argument. This function returns an object with the following properties:
```

docs/useCanAccessCallback.md
```diff
@@ -0,0 +1,61 @@
...
+
+```jsx
+import { useCanAccessCallback, useRecordContext, useDataProvider, Button } from 'react-admin';
+
+const ResetViewsButton = () => {
+    const record = useRecordContext();
+    const dataProvider = useDataProvider();
+    const checkAccess = useCanAccessCallback();
+    const resetViews = () => {
+        const { canAccess, error } = checkAccess({ action: 'resetViews', resource: 'users', record });
```
fzaninotto, line 25: Wrong example: this is an async function

docs/useCanAccessRecordSources.md
```diff
@@ -0,0 +1,47 @@
...
+title: "useCanAccessRecordSources"
+---
+
+# `useCanAccessRecordSources`
+
+This hook calls the authProvider.canAccess() method using react-query for a provided resource and action (and optionally a record).
+
+## Usage
+
+`useCanAccessRecordSources` returns takes an object `{ action, resource, record, sources }` as argument. In addition to react-query result properties, it returns a `canAccess` object that has a property for each provided source determining whether the user has access to it.
```
fzaninotto, line 12: "return takes"? Also, what are the "sources" for?

fzaninotto (changes requested):
Just reviewed the doc, which requires some re-reading

docs/useCanAccessRecordSources.md
djhi, line 12:
```suggestion
`useCanAccessRecordSources` takes an object `{ action, resource, record, sources }` as argument. In addition to react-query result properties, it returns a `canAccess` object that has a property for each provided source determining whether the user has access to it.
```

docs/useCanAccessCallback.md
djhi, line 25:
```suggestion
    const resetViews = async () => {
        const { canAccess, error } = await checkAccess({ action: 'resetViews', resource: 'users', record });
```

docs/useCanAccessRecordSources.md
djhi: I'll add a prop table instead

djhi: Actually there's already a prop table

packages/ra-core/src/auth/useCanAccessRecordSources.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <CoreAdminContext
+        authProvider={authProvider != null ? authProvider : undefined}
+        queryClient={queryClient}
+    >
+        <UseCanAccessRecordSources
+            action="read"
+            resource="posts"
+            sources={['id', 'title', 'author']}
+        >
+            {StateInpector}
```
fzaninotto, line 60: Same remark

packages/ra-core/src/auth/useCanAccessRecordSources.ts
```diff
@@ -0,0 +1,152 @@
...
+    UseQueryOptions,
+    UseQueryResult,
+} from '@tanstack/react-query';
+import useAuthProvider from './useAuthProvider';
+import useLogoutIfAccessDenied from './useLogoutIfAccessDenied';
+import { useEvent } from '../util';
+import { useEffect } from 'react';
+import { useResourceContext } from '../core';
+
+export const combine = <ErrorType>(
```
fzaninotto, line 12: please move this function down to make the exported function the first in the file - it's more expected. 

Also, the name "combine" is too broad IMHO. What do we combine? Into what?

packages/ra-core/src/auth/useCanAccessRecordSources.ts
```diff
@@ -0,0 +1,152 @@
...
+              )
+            : undefined,
+        isPending: results.some(result => result.isPending),
+        isError: results.some(result => result.isError),
+        error: results.find(result => result.error)?.error || undefined,
+    };
+};
+
+/**
+ * Checks if the user can access a resource keys.
```
fzaninotto, line 49: I don't know what it means

packages/ra-core/src/auth/useCanAccessRecordSources.ts
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * `useCanAccessRecordSources` returns an object describing the state of the RBAC request.
```
fzaninotto, line 51: We shouldn't mention RBAC here.

packages/ra-core/src/auth/useCanAccessRecordSources.ts
```diff
@@ -0,0 +1,152 @@
...
+ * As calls to the `authProvider` are asynchronous, the hook returns
+ * a `loading` state in addition to the `canAccess` key.
+ *
```
fzaninotto, line 54: missing jsDoc to explain that the `canAccess` return is a Map.

packages/ra-core/src/auth/useCanAccessRecordSources.ts
```diff
@@ -0,0 +1,152 @@
...
+ *     return (
+ *         <SimpleList
+ *              primaryText={record => canAccess.name ? record.name : ''}
+ *              secondaryText={record => canAccess.email ? record.email : ''}
+ *              tertiaryText={record => canAccess.id ? record.id : ''}
+ *          />
+ *     );
+ * };
+ */
+export const useCanAccessRecordSources = <ErrorType extends Error = Error>(
```
fzaninotto, line 73: I think we should extract the generic parallel `canAccess` fetcher based on `useQueries`, keeping only the logic of the ${resource}.${source}` combination here.

fzaninotto (changes requested):
I reviewed the `auth` part so far.

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,128 @@
...
+        result.error,
+        result.status,
+        result.isFetching,
+    ]);
+
+    return useMemo(() => {
+        return {
+            ...result,
+            canAccess: result.data,
+        };
```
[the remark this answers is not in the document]
djhi: All our other hooks returns the react-query result and add a property such as `canAccess` (`useAuthState` for instance)

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -60,8 +76,63 @@ export interface EditViewProps
...
+    return (
+        <Box
+            sx={theme => ({
+                display: 'flex',
+                flexDirection: 'column',
+                alignItems: 'center',
+                justifyContent: 'center',
+                [theme.breakpoints.down('md')]: {
+                    padding: '1em',
+                },
```
fzaninotto, line 99: Prefer the version not depending on theme: `padding: { sm: '1em', md: 0 }`

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -17,18 +22,29 @@ export const EditView = (props: EditViewProps) => {
...
     const finalActions =
         typeof actions === 'undefined' && hasShow ? defaultActions : actions;
-    if (!children) {
+    if (!children || isPending) {
         return null;
     }
 
+    if (!canAccess) {
+        return unauthorizedView;
+    }
```
fzaninotto, line 46: We must do the same in List

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -60,8 +76,63 @@ export interface EditViewProps
...
+        >
+            <h1 role="alert">
+                <LockIcon
+                    sx={{
+                        width: '2em',
+                        height: '2em',
+                        marginRight: '0.5em',
+                    }}
+                />
+                {translate('ra-rbac.page.unauthorized')}
```
fzaninotto, line 116: wrong prefix: we're not in ra-rbac. Same below. 

Also, we must provide a default translation and populate the en and fr packages.

packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
```diff
@@ -585,3 +587,198 @@ export const LabelElements = () => (
...
+        'books.author': boolean;
+        'books.year': boolean;
+    };
+    queryClient: QueryClient;
+}) => {
+    return (
+        <div>
+            <div>
+                <label>
+                    <input
```
fzaninotto, line 612: unchecking this checkbox has no effect. Is it normal?

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -150,6 +153,23 @@ export const Datagrid: React.ForwardRefExoticComponent<
...
     } = useListContextWithProps(props);
 
+    const resourceKeys =
+        React.Children.map(children, field => {
+            if (!isValidElement(field)) {
+                return null;
+            }
+
+            return field.props.source;
+        })?.filter(value => typeof value === 'string') || [];
```
fzaninotto, line 163: I understand that fields without source won't be checked, am I right? This deserves a warning in the doc.

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -150,6 +153,23 @@ export const Datagrid: React.ForwardRefExoticComponent<
         total,
     } = useListContextWithProps(props);
 
+    const resourceKeys =
```
fzaninotto, line 156: sourceKeys?

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -200,7 +220,24 @@ export const Datagrid: React.ForwardRefExoticComponent<
         [data, isRowSelectable, onSelect, onToggleItem, selectedIds]
     );
 
-    if (isPending === true) {
+    const accessibleChildren = useMemo(() => {
```
fzaninotto, line 223: This may break the `optimized` variant

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
fzaninotto, line 1: The new access control features in Datagrid need documentation

fzaninotto (changes requested):

packages/ra-core/src/export/useExporter.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const exporterWithAccessControl = async (
+        records: Record<string, unknown>[],
+        _,
+        __,
+        resource
+    ) => {
+        const keys = getAllKeys(records);
+
+        const accessRecord = await keys.reduce(async (acc, key) => {
```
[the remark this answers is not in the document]
djhi: Nope or we'd have to introduce a callback version of the batch version too

packages/ra-ui-materialui/src/list/datagrid/Datagrid.stories.tsx
djhi: Removed the checkbox as it made no sense here

docs/AuthProviderWriting.md
```diff
@@ -402,6 +404,10 @@ React-admin doesn't use permissions by default, but it provides [the `usePermiss
 
 [The Role-Based Access Control (RBAC) module](./AuthRBAC.md) allows fined-grained permissions in react-admin apps, and specifies a custom return format for `authProvider.getPermissions()`. Check [the RBAC documentation](./AuthRBAC.md#authprovider-methods) for more information.
 
+### `canAccess`
+
+This method should return a boolean indicating whether users can perform the provided action on the provided resource (in the [Role-Based Access Control (RBAC)](https://en.wikipedia.org/wiki/Role-based_access_control) sense).
+
```
fzaninotto, line 410: This will need a rewrite to list the calls made by react-admin in various places. Also, we shouldn't mention only RBAC as this syntax also allows ABAC, ACL, and more. 

Also, it's not clear by reading the doc what the authProvider should return when the user doesn't have the rights for a resource (a rejected Promise or a resolved Promise to false?)

docs/Datagrid.md
```diff
@@ -1368,3 +1368,30 @@ export const PostList = () => (
     </List>
 );
 ```
+
+### Controlling Access To The Columns
+
+Should your `authProvider` implements the [`canAccess` method](./AuthProviderWriting.md#canaccess), the `<Datagrid>` will call it for each record property to ensure it only display the columns users have access to. For instance, given the following `<Datagrid>`:
```
fzaninotto, line 1374:
```suggestion
Should your `authProvider` implement the [`canAccess` method](./AuthProviderWriting.md#canaccess), the `<Datagrid>` will call to display only the columns users have access to. For instance, given the following `<Datagrid>`:
```

docs/Datagrid.md
```diff
@@ -1368,3 +1368,30 @@ export const PostList = () => (
...
+        <Datagrid>
+            <TextField source="title" />
+            <TextField source="author" />
+            <TextField source="published_at" />
+        </Datagrid>
+    </List>
+);
+```
+
+The `authProvider.canAccess` method will be called 3 times with the following parameters:
```
fzaninotto, line 1391:
```suggestion
The `<Datagrid>` will call `authProvider.canAccess()` 3 times, using the following parameters:
```

docs/Datagrid.md
```diff
@@ -1368,3 +1368,30 @@ export const PostList = () => (
...
+
+1. `{ action: "read", resource: "posts.title" }`
+2. `{ action: "read", resource: "posts.author" }`
+3. `{ action: "read", resource: "posts.published_at" }`
+
+**Note**: `<Datagrid>` children that don't have a `source` prop will always be displayed.
```
fzaninotto, line 1397:
```suggestion
**Note**: `<Datagrid>` children that don't have a `source` prop will always be displayed regardless of access control.
```

docs/Datagrid.md
```diff
@@ -1368,3 +1368,30 @@ export const PostList = () => (
     </List>
 );
 ```
+
+### Controlling Access To The Columns
```
fzaninotto, line 1372:
```suggestion
## Access Control
```

docs/List.md
```diff
@@ -1266,3 +1266,23 @@ const ProductList = () => {
 ```
 
 `useListController` returns callbacks to sort, filter, and paginate the list, so you can build a complete List page. Check [the `useListController`hook documentation](./useListController.md) for details.
+
+## Controlling Access To The List
+
+### Controlling Access In Exports
```
fzaninotto, line 1272:
```suggestion
## Access Control

### Exported Fields
```

docs/List.md
```diff
@@ -1266,3 +1266,23 @@ const ProductList = () => {
...
+
+Should your authProvider implements the [`canAccess` method](./AuthProviderWriting.md#canaccess), the [`exporter`](#exporter) will call it for each record property to ensure it only exports fields to which users have access to. For instance, given the following record shape:
```
fzaninotto, line 1274:
```suggestion
Should your authProvider implement the [`canAccess` method](./AuthProviderWriting.md#canaccess), the [`exporter`](#exporter) will call it to export only fields to which the current user has access to. For instance, given the following record shape:
```

docs/List.md
```diff
@@ -1266,3 +1266,23 @@ const ProductList = () => {
...
+{
+    "id": 0,
+    "title": "Never Gonna Give You Up",
+    "author": "Rick Astley",
+}
+```
+
+The `authProvider.canAccess` method will be called 3 times with the following parameters:
+
+1. `{ action: "read", resource: "posts.id" }`
```
fzaninotto, line 1286: I think we should use the `export` action for that. This allows to have a distinction between fields that are shown and the ones that are exported.

docs/useCanAccess.md
```diff
@@ -5,15 +5,14 @@ title: "useCanAccess"
 
 # `useCanAccess`
 
-This hook, part of [the ra-rbac module](https://react-admin-ee.marmelab.com/documentation/ra-rbac)<img class="icon" src="./img/premium.svg" />, calls the `authProvider.getPermissions()` to get the role definitions, then checks whether the requested action and resource are allowed for the current user. 
+This hook calls the authProvider.canAccess() method using react-query for a provided resource and action (and optionally a record).
```
fzaninotto, line 8:
```suggestion
This hook calls the `authProvider.canAccess()` method on mount for a provided resource and action (and optionally a record). It returns an object containing a `canAccess` boolean set to `true` is the current user can access the resource and action. 
```

The fact that it uses react-query is an implementation detail IMHO

docs/Authentication.md
```diff
@@ -49,6 +49,8 @@ const authProvider = {
     getIdentity: () => Promise.resolve(),
     // get the user permissions (optional)
     getPermissions: () => Promise.resolve(),
+    // get the user authorization to perform an action on a resource (optional)
```
fzaninotto, line 52: "get the user authorization" is confusing. Are we asking the user for his authorization?

docs/useCanAccessCallback.md
```diff
@@ -0,0 +1,61 @@
...
+
+```jsx
+import { useCanAccessCallback, useRecordContext, useDataProvider, Button } from 'react-admin';
+
+const ResetViewsButton = () => {
+    const record = useRecordContext();
+    const dataProvider = useDataProvider();
+    const checkAccess = useCanAccessCallback();
+    const resetViews = async () => {
+        const { canAccess, error } = await checkAccess({ action: 'resetViews', resource: 'users', record });
```
fzaninotto, line 25: I think you should catch errors using a try/catch block

packages/ra-core/src/auth/useCanAccessCallback.ts
```diff
@@ -0,0 +1,94 @@
...
+                UseCanAccessCallbackResult<ErrorType>
+            >(['auth', 'canAccess', JSON.stringify(params)]);
+
+            if (cachedResult && cachedResult.canAccess !== undefined) {
+                return cachedResult;
+            }
+            return authProvider
+                .canAccess(params)
+                .then(data => ({ canAccess: data, error: undefined }))
+                .catch(error => ({ canAccess: undefined, error }));
```
fzaninotto, line 60: I don't understand why you do this. Why don't you just return `canAccess` and let the user try/catch for errors?

docs/useCanAccessCallback.md
```diff
@@ -0,0 +1,61 @@
...
+# `useCanAccessCallback`
+
+This hook returns a callback to call the `authProvider.canAccess()` method for a provided resource and action (and optionally a record).
+
+## Usage
+
+`useCanAccessCallback` returns an async function that takes an object `{ action, resource, record }` as argument. This function returns an object with the following properties:
+
+- `canAccess`: a boolean determining whether the user has access to the resource.
+- `error`: the error that might have occurred.
```
fzaninotto, line 15: This API is strange. I'd expect it to return `canAccess` directly and throw in case of error. Why return an object instead?

docs/useCanAccessCallback.md
```diff
@@ -0,0 +1,61 @@
...
+
+```jsx
+import { useCanAccessCallback, useRecordContext, useDataProvider, Button } from 'react-admin';
+
+const ResetViewsButton = () => {
```
fzaninotto, line 20: This example isn't great: if the user cannot reset the views, they shouldn't see the button in the first place.

docs/useCanAccessRecordSources.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: default
+title: "useCanAccessRecordSources"
```
fzaninotto, line 3: outdated doc

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,124 @@
...
+    extends Omit<UseQueryOptions<boolean, ErrorType>, 'queryKey' | 'queryFn'> {
+    resource: string;
+    action: string;
+    record?: unknown;
+    onSuccess?: (data: boolean) => void;
+    onError?: (err: ErrorType) => void;
+    onSettled?: (data?: boolean, error?: ErrorType | null) => void;
+}
+
+export type UseCanAccessResult<ErrorType = Error> = QueryObserverResult<
```
fzaninotto, line 117: Can we be a bit smarter and use a union to help TypeScript know when `canAccess` is actually non-null, just as we do in `useGetList`?

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const result = useQuery({
+        queryKey: ['auth', 'canAccess', JSON.stringify(params)],
+        queryFn: async ({ signal }) => {
+            if (!authProvider || !authProvider.canAccess) {
+                return true;
+            }
+            return authProvider.canAccess({
+                ...params,
+                signal,
```
fzaninotto, line 58: we've had to introduce `supportAbortSignal` in #9901 to avoid problems in strict mode. I fear we may have to do the same here.

packages/ra-core/src/auth/useCanAccess.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+            checkAuth: () => Promise.reject('bad method'),
+            checkError: () => Promise.reject('bad method'),
+            getPermissions: () => Promise.reject('bad method'),
+            canAccess: () => Promise.resolve(false),
+        };
+        render(<Basic authProvider={authProvider} />);
+
+        await waitFor(() => {
+            expect(screen.queryByText('LOADING')).toBeNull();
+            expect(screen.queryByText('canAccess: NO')).not.toBeNull();
```
fzaninotto, line 68: can you also check that the ERROR doesn't display in this case?

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @example
+ *     import { useCanAccess } from 'react-admin';
+ *
+ *     const PostDetail = () => {
+ *         const { isPending, canAccess } = useCanAccess({
+ *             resource: 'posts',
+ *             action: 'read',
+ *         });
+ *         if (isPending || canAccess) {
```
fzaninotto, line 37:
```suggestion
 *         if (isPending || !canAccess) {
```

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,124 @@
...
+import useLogoutIfAccessDenied from './useLogoutIfAccessDenied';
+import { useEvent } from '../util';
+
+/**
+ * A hook that calls the authProvider.canAccess() method using react-query for a provided resource and action (and optionally a record).
+ *
+ * The return value updates according to the request state:
+ *
+ * - start: { isPending: true }
+ * - success: { canAccess: true, isPending: false }
```
fzaninotto, line 17: can access can also be false in this state

packages/ra-core/src/auth/useCanAccessResources.ts
```diff
@@ -0,0 +1,148 @@
...
+import useLogoutIfAccessDenied from './useLogoutIfAccessDenied';
+import { useEvent } from '../util';
+
+/**
+ * Checks whether users can access the provided resources (resource in the RBAC sense).
+ *
+ * `useCanAccessResources` returns an object describing the state of the request:
+ *
+ * - start: { isPending: true }
+ * - success: { canAccess: true, isPending: false }
```
fzaninotto, line 17: no, the value of `canAccess` in this case is a map

packages/ra-core/src/auth/useCanAccessResources.ts
```diff
@@ -0,0 +1,148 @@
...
+    useQueries,
+    UseQueryOptions,
+    UseQueryResult,
+} from '@tanstack/react-query';
+import useAuthProvider from './useAuthProvider';
+import useLogoutIfAccessDenied from './useLogoutIfAccessDenied';
+import { useEvent } from '../util';
+
+/**
+ * Checks whether users can access the provided resources (resource in the RBAC sense).
```
fzaninotto, line 12: Remove the RBAC mention

packages/ra-core/src/auth/useCanAccessResources.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * @example
+ * import { useCanAccessResources } from 'react-admin';
+ *
+ * const DeleteUserButton = ({ record }) => {
+ *     const { isPending, canAccess } = useCanAccessResources({
+ *         action: 'delete',
+ *         resources: ['users.id', 'users.name', 'users.email'],
+ *         record
+ *     });
```
fzaninotto, line 35: this example doesn't make sense

fzaninotto (changes requested):

packages/ra-core/src/controller/list/useListController.accessControl.stories.tsx
```diff
@@ -0,0 +1,233 @@
...
+                    )
+                }
+            >
+                Export
+            </button>
+            <ul style={styles.ul}>
+                {!params.isPending &&
+                    params.data!.map(post => (
+                        <li key={`post_${post.id}`}>
+                            {`${post.title} - ${post.votes} votes`}
```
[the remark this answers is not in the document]
djhi: I added some text to clarify

packages/ra-core/src/controller/list/useListController.ts
```diff
@@ -154,13 +162,17 @@ export const useListController = <RecordType extends RaRecord = any>(
...
         sort: currentSort,
         data,
         defaultTitle,
         displayedFilters: query.displayedFilters,
         error,
-        exporter,
+        exporter: exporterWithAccessControl,
         filter,
         filterValues: query.filterValues,
         hideFilter: queryModifiers.hideFilter,
```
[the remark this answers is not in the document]
djhi: Done

packages/ra-core/src/types.ts
```diff
@@ -69,6 +69,13 @@ export type AuthProvider = {
     handleCallback?: (
         params?: QueryFunctionContext
     ) => Promise<AuthRedirectResult | void | any>;
+    canAccess?: (
```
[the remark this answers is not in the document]
djhi: Done

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -807,6 +813,9 @@ describe('<Edit />', () => {
                     </Edit>
                 </AdminContext>
             );
+            await waitFor(() => {
+                expect(screen.getAllByText('Hello')).toHaveLength(1);
+            });
```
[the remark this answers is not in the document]
djhi: Removed the other one

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
djhi: Done

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
djhi: Added

docs/useCanAccess.md
djhi: That was so users expect react-query properties in the result but ok

docs/useCanAccessCallback.md
djhi: Any suggestion?

docs/useCanAccessCallback.md
djhi: Didn't you suggest to throw in https://github.com/marmelab/react-admin/pull/10166/files#r1747220540?

packages/ra-core/src/auth/useCanAccessResources.ts
djhi: How would you suggest to make the distinction with React-Admin resources

packages/ra-core/src/auth/useCanAccessResources.ts
djhi: Only the component name, right?

packages/ra-core/src/export/useExporter.ts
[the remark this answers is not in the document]
djhi: Should we introduce it?

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
djhi: What optimized variant?

packages/ra-core/src/export/useExporter.ts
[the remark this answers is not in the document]
fzaninotto: not until we need it a second time

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
fzaninotto: `<Datagrid optimized>`, which memoizes the render regardless of changes in the children

docs/useCanAccessCallback.md
fzaninotto: Sorry, I don't see the target of that link.

docs/useCanAccessCallback.md
fzaninotto: - A click handler that links to the record edit view if the user has the permission, or to the show view otherwise
- a custom exporter

packages/ra-core/src/auth/useCanAccessResources.ts
fzaninotto: "resource in the access control sense"?

packages/ra-core/src/auth/useCanAccessResources.ts
fzaninotto: not only: you check for the delete action and then use it to determine whether you can hows it...

fzaninotto (changes requested):
We're nearly there 👍

packages/ra-core/src/auth/useCanAccess.ts
```diff
@@ -0,0 +1,137 @@
...
+    action: string;
+    record?: unknown;
+    onSuccess?: (data: boolean) => void;
+    onError?: (err: ErrorType) => void;
+    onSettled?: (data?: boolean, error?: ErrorType | null) => void;
+}
+
+export type UseCanAccessResult<ErrorType = Error> =
+    | UseCanAccessLoadingResult<ErrorType>
+    | UseCanAccessRefetchErrorResult<ErrorType>
```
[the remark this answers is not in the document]
djhi: Added

PR #10179, opened 2024-09-02
Add support for response metadata in dataProvider.getList()

### LINKED ISSUE #9569: Possibility to pass additional data in the server response for getList
An API might want to send additional data since all the data isn't send when using pagination

The `useListContext()` hook, as documented [here](https://marmelab.com/react-admin/useListContext.html), provides only `data` and `total`, not additional values sent back by `getList()`

The solutions that could be implemented to allow this mechanism are not practical at all (feel free to explain it here if you were to find one)

I also found this [stackoverflow question](https://stackoverflow.com/questions/75715770/add-extra-data-to-getlist-in-dataprovider) that raises the same issue

--------------------------------------------------------------------------------------

Here is an example using chips to show a summary of all the data (not working since the feature doesn't exist) :
Screenshot 2024-01-07 223943

fzaninotto:
Closes #9569

## Problem

Even though some API backends return additional data in the `getList` results (e.g. facets and other aggregations), there is currently no way to access them.

## Solution

Update `useGetList` and `ListContext` to return any `meta` property present in the API response.

### DIFF
--- a/docs/DataProviderWriting.md
+++ b/docs/DataProviderWriting.md
@@ -55,7 +55,7 @@ interface GetListParams {
     pagination: { page: number, perPage: number };
     sort: { field: string, order: 'ASC' | 'DESC' };
     filter: any;
-    meta?: any;
+    meta?: any; // request metadata
     signal?: AbortSignal;
 }
 interface GetListResult {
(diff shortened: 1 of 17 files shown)


djhi (changes requested):
That's a great addition :+1:

djhi (approved):

PR #10183, opened 2024-09-04
feat(ra-ui-materialui): Improve FilterButton UX

erwanMarmelab:
## Problem

There is no way to differentiate between filter choices and filter actions (such as save current query) in the FilterButton

when you select a filter, it disappears from the list, which changes the place of each choice item and makes it hard to remember where is the filter you want to enable ; same for the actions (Remove all filters)

## To Do

- [x] Add a divider to separate between the filter choices and filter actions
- [x] Keep all choices in the list at all times, and simply gray out the items when they are not available
- [x] Test icons in the Filter Button Menu items
- [x] Add checkbox before filter names 

## Additional Checks

image

image

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
@@ -12,7 +12,6 @@ import {
     MenuItem,
     styled,
     ButtonProps as MuiButtonProps,
-    Box,
     Divider,
 } from '@mui/material';
 import ContentFilter from '@mui/icons-material/FilterList';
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -212,6 +213,16 @@ export const FilterButton = (props: FilterButtonProps) => {
                         </MenuItem>
                     )
                 )}
+                {hasFilterValues && (
+                    <Box
```
fzaninotto, line 217: use MUI's `Divider` instead

fzaninotto (changes requested):

erwanMarmelab:
PR based from the wrong branch. New PR available : https://github.com/marmelab/react-admin/pull/10186

PR #10184, opened 2024-09-05
Update `<Datagrid>` and `<SimpleList>` empty message when a filter is active

erwanMarmelab:
## Problem

When coming back to a list page with active filters and no results, the default message (“no results found”) is misleading when the active filters aren’t visible enough.

## Solution

If the list is empty **AND** there is an active filter,

- Replace “No [data] found” with “No [data] found using the current filters. Clear filters.”
- Add a link on “Clear filter” which removes all active filters

## To Do

- [x] Change dictionaries messages
- [x] Add the "Clear filter" button
- [x] Display this button if there is an active filter

## Additional Checks

- [x] The PR includes **unit tests** 
- [x] The PR includes one or several **stories** 

image

### DIFF
--- a/packages/ra-language-french/src/index.ts
+++ b/packages/ra-language-french/src/index.ts
@@ -118,7 +118,9 @@ const frenchMessages: TranslationMessages = {
         },
         navigation: {
             clear_filters: 'Effacer les filtres.',
-            no_results: 'Aucun résultat trouvé avec les filtres actuels.',
+            no_filtred_results:
+                'Aucun résultat trouvé avec les filtres actuels.',
+            no_results: 'Aucun résultat',
             no_more_results:
                 'La page numéro %{page} est en dehors des limites. Essayez la page précédente.',
             page_out_of_boundaries: 'La page %{page} est en dehors des limites',
--- a/packages/ra-language-french/src/index.ts
+++ b/packages/ra-language-french/src/index.ts
@@ -117,7 +117,8 @@ const frenchMessages: TranslationMessages = {
                 "Certains changements n'ont pas été enregistrés. Êtes-vous sûr(e) de vouloir quitter cette page ?",
         },
         navigation: {
-            no_results: 'Aucun résultat',
+            clear_filters: 'Effacer les filtres.',
+            no_results: 'Aucun %{resource} trouvé avec les filtres actuels.',
             no_more_results:
                 'La page numéro %{page} est en dehors des limites. Essayez la page précédente.',
             page_out_of_boundaries: 'La page %{page} est en dehors des limites',
--- a/packages/ra-language-french/src/index.ts
+++ b/packages/ra-language-french/src/index.ts
@@ -118,7 +118,7 @@ const frenchMessages: TranslationMessages = {
         },
         navigation: {
             clear_filters: 'Effacer les filtres.',
-            no_results: 'Aucun %{resource} trouvé avec les filtres actuels.',
+            no_results: 'Aucun résultat trouvé avec les filtres actuels.',
             no_more_results:
                 'La page numéro %{page} est en dehors des limites. Essayez la page précédente.',
             page_out_of_boundaries: 'La page %{page} est en dehors des limites',
--- a/packages/ra-language-french/src/index.ts
+++ b/packages/ra-language-french/src/index.ts
@@ -117,8 +117,8 @@ const frenchMessages: TranslationMessages = {
                 "Certains changements n'ont pas été enregistrés. Êtes-vous sûr(e) de vouloir quitter cette page ?",
         },
         navigation: {
-            clear_filters: 'Effacer les filtres.',
-            no_filtred_results:
+            clear_filters: 'Effacer les filtres',
+            no_filtered_results:
                 'Aucun résultat trouvé avec les filtres actuels.',
             no_results: 'Aucun résultat',
             no_more_results:
--- a/packages/ra-ui-materialui/src/list/ListNoResults.tsx
+++ b/packages/ra-ui-materialui/src/list/ListNoResults.tsx
@@ -8,14 +8,16 @@ import { Link } from '@mui/material';
 export const ListNoResults = memo(() => {
     const translate = useTranslate();
     const resource = useResourceContext();
-    const { setFilters } = useListController({ resource });
+    const { filterValues, setFilters } = useListController({ resource });
     return (
         <CardContent>
             <Typography variant="body2">
                 {translate('ra.navigation.no_results', { resource })}{' '}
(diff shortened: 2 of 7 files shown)


packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
```diff
@@ -1,11 +1,25 @@
 import * as React from 'react';
-import { render, screen, waitFor, within } from '@testing-library/react';
-import { ListContext, ResourceContextProvider } from 'ra-core';
+import {
+    fireEvent,
```
djhi, line 3:
```suggestion
```

packages/ra-language-french/src/index.ts
```diff
@@ -117,7 +117,8 @@ const frenchMessages: TranslationMessages = {
                 "Certains changements n'ont pas été enregistrés. Êtes-vous sûr(e) de vouloir quitter cette page ?",
         },
         navigation: {
-            no_results: 'Aucun résultat',
+            clear_filters: 'Effacer les filtres.',
+            no_results: 'Aucun %{resource} trouvé avec les filtres actuels.',
```
djhi, line 121: This won't work well with feminine resources in french

djhi (changes requested):

packages/ra-language-french/src/index.ts
fzaninotto:
```suggestion
            no_results: 'Aucun résultat trouvé avec les filtres actuels.',
```

packages/ra-ui-materialui/src/list/ListNoResults.tsx
```diff
@@ -2,15 +2,22 @@ import * as React from 'react';
...
 
 export const ListNoResults = memo(() => {
     const translate = useTranslate();
     const resource = useResourceContext();
+    const { filterValues, setFilters } = useListController({ resource });
     return (
         <CardContent>
             <Typography variant="body2">
-                {translate('ra.navigation.no_results', { resource })}
+                {translate('ra.navigation.no_results', { resource })}{' '}
```
fzaninotto, line 15: you need two different messages for when there is a filter and when there isn't

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
```diff
@@ -134,18 +142,59 @@ describe('<SimpleList />', () => {
...
-                value={{
-                    isLoading: false,
-                    data: [],
-                    total: 0,
-                    resource: 'posts',
-                }}
-            >
-                <SimpleList />
-            </ListContext.Provider>
+            <TestMemoryRouter>
```
fzaninotto, line 145: Why is one necessary?

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
```diff
@@ -134,18 +142,59 @@ describe('<SimpleList />', () => {
...
+
+    it('should display a message when there is no result but filters applied', async () => {
+        const MyListView = () => {
+            const { setFilters } = useListFilterContext();
+            setFilters({ title: 'foo' }, ['title']);
+            return <SimpleList />;
+        };
+
+        render(
+            <TestMemoryRouter>
```
fzaninotto, line 178: same

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
```diff
@@ -143,11 +150,29 @@ describe('<SimpleList />', () => {
...
+        );
+
+        fireEvent.change(screen.getByLabelText('Search'), {
+            target: { value: 'w' },
+        });
+
+        expect(
+            await screen.findByText('No posts found using the current filters.')
+        ).not.toBeNull();
+        expect(screen.getByText('Clear filters.')).not.toBeNull();
```
djhi, line 173: You should also test that the button actually clears the filters

djhi (changes requested):

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
```diff
@@ -143,11 +150,45 @@ describe('<SimpleList />', () => {
...
             >
                 <SimpleList />
-            </ListContext.Provider>
+            </ListContext.Provider>,
+            { wrapper: Wrapper }
         );
         expect(screen.queryByText('ra.navigation.no_results')).not.toBeNull();
     });
 
+    it('should display a message when there is no result but filters applied', async () => {
```
djhi, line 159: Why isn't this test in the FilterButton tests file?

erwanMarmelab: Because we test the `<SimpleList>` logic here. I use the `FilterButton`'s story to be able to test it with writing a new component. Do you prefer I write a new component to render that test ?

djhi: nvm

djhi (approved):

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.spec.tsx
djhi: But then, why the SimpleList and not the List? Should we do both? @fzaninotto

fzaninotto: In fact, the List?oResults is used by `SimpleList` and `Datagrid`, so in theory we should test both. Instead, I added a test for `ListNoResults` itself.

fzaninotto (approved):

packages/ra-ui-materialui/src/list/ListNoResults.tsx
```diff
@@ -2,15 +2,20 @@ import * as React from 'react';
...
 import CardContent from '@mui/material/CardContent';
 import Typography from '@mui/material/Typography';
-import { useResourceContext, useTranslate } from 'ra-core';
+import { useListController, useResourceContext, useTranslate } from 'ra-core';
+import { Link } from '@mui/material';
 
 export const ListNoResults = memo(() => {
     const translate = useTranslate();
     const resource = useResourceContext();
+    const { setFilters } = useListController({ resource });
```
fzaninotto: `useListController`, not `useListContext`! 

I fixed the mistake in 199fbaec041f6c5ded6f78bbc0eb4d3d2b4c8069

djhi (approved):

PR #10186, opened 2024-09-06
feat(ra-ui-materialui): Improve FilterButton UX

erwanMarmelab:
## Problem

There is no way to differentiate between filter choices and filter actions (such as save current query) in the FilterButton

when you select a filter, it disappears from the list, which changes the place of each choice item and makes it hard to remember where is the filter you want to enable ; same for the actions (Remove all filters)

## To Do

- [x] Add a divider to separate between the filter choices and filter actions
- [x] Keep all choices in the list at all times, and simply gray out the items when they are not available
- [x] Test icons in the Filter Button Menu items
- [x] Add checkbox before filter names 

## Additional Checks

image

image

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
@@ -44,7 +44,7 @@ describe('<FilterButton />', () => {
     });
 
     describe('filter selection menu', () => {
-        it('should check applied filters adn display decdicated inputs', async () => {
+        it('should control filters display by checking/unchecking it in the menu', async () => {
             render(<Basic />);
 
             fireEvent.click(await screen.findByLabelText('Add filter'));
@@ -89,7 +89,7 @@ describe('<FilterButton />', () => {
             expect(checkboxs[2].checked).toBe(false);
         }, 7000);
 
-        it('should uncheck removed filters by removing dedicated inputs', async () => {
+        it('should remove the checked state of the menu item when removing its matching filter', async () => {
             render(<Basic />);
 
             fireEvent.click(await screen.findByLabelText('Add filter'));
--- a/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
@@ -43,28 +43,27 @@ describe('<FilterButton />', () => {
     });
 
     describe('filter selection menu', () => {
-        it('should display only hidden filters', () => {
-            const hiddenFilter = (
-                <TextInput source="Returned" label="Returned" />
-            );
-            const { getByLabelText, queryByText } = render(
-                <AdminContext theme={theme}>
-                    <ResourceContextProvider value="posts">
-                        <ListContextProvider value={defaultListContext}>
-                            <FilterButton
-                                filters={defaultProps.filters.concat(
-                                    hiddenFilter
-                                )}
-                            />
-                        </ListContextProvider>
-                    </ResourceContextProvider>
-                </AdminContext>
-            );
+        it('should disable applied filters', async () => {
+            render(<Basic />);
 
-            fireEvent.click(getByLabelText('ra.action.add_filter'));
+            fireEvent.click(await screen.findByLabelText('Add filter'));
 
-            expect(queryByText('Returned')).not.toBeNull();
-            expect(queryByText('Name')).toBeNull();
+            let filters = screen.getAllByRole('menuitem');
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -1,6 +1,12 @@
 import * as React from 'react';
 import expect from 'expect';
-import { render, fireEvent, screen, waitFor } from '@testing-library/react';
+import {
+    render,
+    fireEvent,
+    screen,
+    waitFor,
+    within,
```
djhi, line 8:
```suggestion
```

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -74,13 +80,18 @@ export const FilterButton = (props: FilterButtonProps) => {
     }
 
     const hiddenFilters = filters.filter(
```
djhi, line 82: This should be renamed. `allTogglableFilters`?

packages/ra-ui-materialui/src/list/filter/FilterButton.tsx
```diff
@@ -113,6 +124,24 @@ export const FilterButton = (props: FilterButtonProps) => {
...
+            // We have to fallback to imperative code because the new FilterFormInput
+            // has no way of knowing it has just been displayed (and thus that it should focus its input)
+            setTimeout(() => {
+                const inputElement = document.querySelector(
+                    `input[name='${source}']`
+                ) as HTMLInputElement;
+                if (inputElement) {
+                    inputElement.focus();
+                }
+            }, 50);
```
djhi, line 139: Why do we have to focus the filter input that we just removed?

djhi (changes requested):

djhi:
Besides, the tests are failing

djhi (changes requested):
We're missing tests about removing filters:
- Removing from the filter button we already had should uncheck its matching item in the menu
- Unchecking an applied filter from the menu should remove it (both the input and the filter value)

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -36,35 +36,90 @@ describe('<FilterButton />', () => {
...
-                        <ListContextProvider value={defaultListContext}>
-                            <FilterButton
-                                filters={defaultProps.filters.concat(
-                                    hiddenFilter
-                                )}
-                            />
-                        </ListContextProvider>
-                    </ResourceContextProvider>
-                </AdminContext>
+        it('should check applied filters adn display decdicated inputs', async () => {
```
djhi, line 47:
```suggestion
        it('should control filters display by checking/unchecking it in the menu', async () => {
```

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -36,35 +36,90 @@ describe('<FilterButton />', () => {
...
-            fireEvent.click(getByLabelText('ra.action.add_filter'));
+            fireEvent.click(screen.getByLabelText('Add filter'));
+            checkboxs = screen.getAllByRole('checkbox');
+            expect(checkboxs).toHaveLength(3);
+            expect(checkboxs[0].checked).toBe(false);
+            expect(checkboxs[1].checked).toBe(false);
+            expect(checkboxs[2].checked).toBe(false);
+        }, 7000);
+
+        it('should uncheck removed filters by removing dedicated inputs', async () => {
```
djhi, line 92:
```suggestion
        it('should remove the checked state of the menu item when removing its matching filter', async () => {
```

djhi (changes requested):
Almost there!

packages/ra-ui-materialui/src/list/filter/FilterButton.spec.tsx
```diff
@@ -36,35 +36,90 @@ describe('<FilterButton />', () => {
...
-                        <ListContextProvider value={defaultListContext}>
-                            <FilterButton
-                                filters={defaultProps.filters.concat(
-                                    hiddenFilter
-                                )}
-                            />
-                        </ListContextProvider>
-                    </ResourceContextProvider>
-                </AdminContext>
+        it('should control filters display by checking/unchecking it in the menu', async () => {
```
djhi, line 47:
```suggestion
        it('should control filters display by checking/unchecking them in the menu', async () => {
```

djhi (approved):

Issue #9100, opened 2023-07-13
Add ability to disable sync with location in List without disabling saving params in the Store


slax57:
Currently, in a List, we can:
- :heavy_check_mark: enable sync with location and :heavy_check_mark: enable storing the list params, with the default props
- ❌ disable sync with location and ❌ disable storing the list params, with `disableSyncWithLocation`
- :heavy_check_mark: enable sync with location and ❌ disable storing the list params, with `storeKey={false}`

But we can't ❌ disable sync with location and and :heavy_check_mark: enable storing the list params.

This is what https://github.com/marmelab/react-admin/pull/8530 is suggesting.

I think RA should offer that ability, either via https://github.com/marmelab/react-admin/pull/8530 or another PR if this one does not reach completion.

btbenedi:
+1 I opened that original PR and real work has gotten in the way of me finishing that PR

fzaninotto:
Fixed by #10187

PR #10187, opened 2024-09-06
Enable Store sync when `disableSyncWithLocation` is true

WiXSL:
## Problem

#9100

## Solution

Implements #9100 and continues over #8530

## Additional Checks

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -415,7 +415,7 @@ const BoolkList = () => (
 
 By default, react-admin synchronizes the `<List>` parameters (sort, pagination, filters) with the query string in the URL (using `react-router` location) and the [Store](./Store.md).
 
-When you use a `<List>` component anywhere else than as `<Resource list>`, you may want to disable this synchronization to keep the parameters in a local state, independent for each `<List>` instance. This allows to have multiple lists on a single page. The drawback is that a hit on the "back" button doesn't restore the previous list parameters. To do so, pass the `disableSyncWithLocation` prop.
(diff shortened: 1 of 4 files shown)


docs/List.md
```diff
@@ -445,7 +445,35 @@ const Dashboard = () => (
 ```
 {% endraw %}
 
-**Tip**: As `disableSyncWithLocation` also disables the persistence of the list parameters in the Store, any custom string specified in the `storeKey` prop is ignored when `disableSyncWithLocation` is set to `true`.
+**Tip**: As `disableSyncWithLocation` also disables the persistence of the list parameters in the Store, any custom string specified in the `storeKey` prop is ignored when `disableSyncWithLocation` is set to `true`. To solve this, you can pass a custom `storeKey` prop. react-admin will store and sync the parameters using the passed storeKey.
```
djhi, line 448:
```suggestion
**Tip**: `disableSyncWithLocation` also disables the persistence of the list parameters in the Store by default. To enable the persistence of the list parameters in the Store, you can pass a custom `storeKey` prop.
```

djhi:
Thanks!

djhi (approved):

packages/ra-core/src/controller/list/useListParams.ts
```diff
@@ -82,7 +82,7 @@ export const useListParams = ({
     perPage = 10,
     resource,
     sort = defaultSort,
-    storeKey = `${resource}.listParams`,
+    storeKey = disableSyncWithLocation ? '' : `${resource}.listParams`,
```
djhi, line 85: Shouldn't the default value be `false` by default when `disableSyncWithLocation === true`?

djhi (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #10200, opened 2024-09-12
Avoid unnecessary rerenders with canAccess hooks when there is no authProvider

djhi:
## Problem

As we use react-query for the canAccess hooks, users may see at least one rerender because it run asynchronously.

## Solution

Detect there is no support for canAccess and return a fake react-query result if needed.

## How To Test

The `useCanAccess` and `useCanAccessResources` have stories with and without an AuthProvider that shows the behavior.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/auth/useCanAccess.spec.tsx
+++ b/packages/ra-core/src/auth/useCanAccess.spec.tsx
@@ -6,17 +6,10 @@ import { QueryClient } from '@tanstack/react-query';
 import { Basic } from './useCanAccess.stories';
 
 describe('useCanAccess', () => {
-    it('should return a loading state on mount', () => {
+    it('should allow access on mount when there is no authProvider', () => {
         render(<Basic authProvider={null} />);
-        expect(screen.queryByText('LOADING')).not.toBeNull();
-    });
-
-    it('should return isPending: true by default after a tick', async () => {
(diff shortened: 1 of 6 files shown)


packages/ra-core/src/auth/useCanAccess.spec.tsx
```diff
@@ -6,17 +6,10 @@ import { QueryClient } from '@tanstack/react-query';
...
-    });
-
-    it('should return isPending: true by default after a tick', async () => {
-        render(<Basic authProvider={null} />);
-        expect(screen.queryByText('LOADING')).not.toBeNull();
-        await waitFor(() => {
-            expect(screen.queryByText('LOADING')).toBeNull();
-        });
+        expect(screen.queryByText('LOADING')).toBeNull();
+        screen.getByText('canAccess: YES');
```
fzaninotto, line 12: You shouldn't replace these two tests, you should add one more

fzaninotto (changes requested):

packages/ra-core/src/auth/useCanAccess.spec.tsx
djhi: Are we testing react-query?

fzaninotto (approved):

PR #10201, opened 2024-09-12
Update create/update/delete notification messages to include resource name

fzaninotto:
## Problem

The snackbar that confirms the creation/update/deletion use a generic "element" name. It's impersonal and sometimes confusing. 

Capture d’écran 2024-09-12 à 15 54 20
Capture d’écran 2024-09-12 à 15 54 12
Capture d’écran 2024-09-12 à 15 53 54
Capture d’écran 2024-09-12 à 15 54 01

## Solution

Use the resource name 

Capture d’écran 2024-09-12 à 15 52 27
Capture d’écran 2024-09-12 à 15 52 12
Capture d’écran 2024-09-12 à 15 47 48
Capture d’écran 2024-09-12 à 15 48 08

French:

Capture d’écran 2024-09-12 à 15 53 18
Capture d’écran 2024-09-12 à 15 53 09
Capture d’écran 2024-09-12 à 15 52 48
Capture d’écran 2024-09-12 à 15 52 56

## To be fixed

In languages with gender, there is no way to alter the translation according to the resource gender. 

e.g. If e use the same French translation for `ra.notification.created`, it will only work in one case. 

- "Camion créé"
- "Voiture créé" => missing final "e"

## Breaking change

This may be a breaking change for applications that notify the following messages and use the default `en` and `fr` translations:

- `ra.notification.created`
- `ra.notification.updated`
- `ra.notification.deleted`

These are internal messages, and they shouldn't be used in user land. We should however document the upgrade path. 

## How To Test

1. Launch the simple example
2. Edit a post, and save => see update one notification
3. Edit a post, and delete => delete one notification
4. Select several posts, and delete => delete many notification
5. Create a new post and save => create notification

## Additional Checks

### DIFF
--- a/examples/demo/src/reviews/ReviewCreate.tsx
+++ b/examples/demo/src/reviews/ReviewCreate.tsx
@@ -10,6 +10,7 @@ import {
     useNotify,
     useRedirect,
     getRecordFromLocation,
+    useGetResourceLabel,
 } from 'react-admin';
 import { useLocation } from 'react-router';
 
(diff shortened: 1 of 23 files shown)


fzaninotto:
This is a dead end, I'll look for an alternative implementation.

bartbrinkman:
@fzaninotto It seems that the `successMessage` prop is available for this, but doesn't actually get passed. Is this a solution direction?

fzaninotto:
@bartbrinkman I have an alternative solution coming, stay tuned

fzaninotto:
@bartbrinkman Check out https://github.com/marmelab/react-admin/pull/10203

PR #10206, opened 2024-09-16
Fix for ids not being escaped in paths in the Simple Rest Data Provider

### LINKED ISSUE #10207: ids not being escaped in paths in the Simple Rest Data Provider
**What you were expecting:**
Records may have ids with characters that break the Simple REST Data Provider paths if not escaped. 
For example, a record with id "Post#123" has a getOne path of '/posts/Post#123' which is not desired.
IDs like this can occur when using single table inheritance with DynamoDB or with other document data stores.
The solution is to escape IDs in paths for getOne, update, updateMany, delete and deleteMany. An example path with escaped id is '/posts/Post%23123'.

**What happened instead:**
A record with id "Post#123" has a getOne path of '/posts/Post#123' which is not desired.

**Steps to reproduce:**
Use a record with id 'Post#123' and perform anything that does a getOne operation.

**Related code:**
Here is a PR with a fix:
[10206](https://github.com/marmelab/react-admin/pull/10206)

ckhsponge:
## Problem

Records may have ids with characters that break the Simple REST Data Provider paths if not escaped. 
For example, a record with id "Post#123" has a getOne path of '/posts/Post#123' which is not desired.
IDs like this can occur when using single table inheritance with DynamoDB or with other document data stores.

## Solution

The solution is to escape IDs in paths for getOne, update, updateMany, delete and deleteMany. An example path with escaped id is '/posts/Post%23123'.

Closes #10207

## How To Test

Perform getOne, update, updateMany, delete and deleteMany with records that have escape-requiring ids e.g. 'Post#123'. 

## Additional Checks

- [ x ] The PR targets `master` for a bugfix, or `next` for a feature
- [ x ] The PR includes **unit tests** (if not possible, describe why)
- [ x ] The **documentation** is up to date

### DIFF
--- a/packages/ra-data-simple-rest/src/index.spec.ts
+++ b/packages/ra-data-simple-rest/src/index.spec.ts
@@ -70,7 +70,7 @@ describe('Data Simple REST Client', () => {
     });
     describe('getOne', () => {
         it('should allow numeric id in path', async () => {
-            const httpClient = jest.fn().mockResolvedValue({ id: 'Post#123' });
+            const httpClient = jest.fn().mockResolvedValue({ id: 123 });
             const client = simpleClient('http://localhost:3000', httpClient);
 
             await client.getOne('posts', { id: 123 });
@@ -113,12 +113,12 @@ describe('Data Simple REST Client', () => {
         it('should escape id in path', async () => {
             const httpClient = jest
                 .fn()
-                .mockResolvedValue({ json: { id: 'Post#123' } });
+                .mockResolvedValue({ json: ['Post#123'] });
             const client = simpleClient('http://localhost:3000', httpClient);
 
             await client.updateMany('posts', {
-                data: undefined,
-                ids: [{ id: 'Post#123' }],
+                data: { body: '' },
+                ids: ['Post#123'],
             });
 
             expect(httpClient).toHaveBeenCalledWith(
@@ -165,7 +165,7 @@ describe('Data Simple REST Client', () => {
     });
     describe('deleteMany', () => {
         it('should set the `Content-Type` header to `text/plain`', async () => {
-            const httpClient = jest.fn().mockResolvedValue({ json: { id: 1 } });
+            const httpClient = jest.fn().mockResolvedValue({ json: [1] });
 
             const client = simpleClient('http://localhost:3000', httpClient);
 
@@ -196,11 +196,11 @@ describe('Data Simple REST Client', () => {
         it('should escape id in path', async () => {
             const httpClient = jest
                 .fn()
-                .mockResolvedValue({ json: { id: 'Post#123' } });
+                .mockResolvedValue({ json: ['Post#123'] });
             const client = simpleClient('http://localhost:3000', httpClient);
 
             await client.deleteMany('posts', {
-                ids: [{ id: 'Post#123' }],
+                ids: ['Post#123'],
             });
 
             expect(httpClient).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


packages/ra-data-simple-rest/src/index.ts
```diff
@@ -156,10 +156,13 @@ export default (
     updateMany: (resource, params) =>
         Promise.all(
             params.ids.map(id =>
-                httpClient(`${apiUrl}/${resource}/${id}`, {
-                    method: 'PUT',
-                    body: JSON.stringify(params.data),
-                })
+                httpClient(
+                    `${apiUrl}/${resource}/${encodeURIComponent(typeof id === 'object' ? id.id : String(id))}`,
```
fzaninotto, line 160: I don't understand why you test that `id` may be an object. The interface doesn't allow for that. Same for `deleteMany`.

packages/ra-data-simple-rest/src/index.spec.ts
```diff
@@ -68,6 +68,65 @@ describe('Data Simple REST Client', () => {
             expect(result.total).toEqual(42);
         });
     });
+    describe('getOne', () => {
+        it('should allow numeric id in path', async () => {
+            const httpClient = jest.fn().mockResolvedValue({ id: 'Post#123' });
```
fzaninotto, line 73: This may cause a warning as the requested `id` and the returned `id` aren't the same.

packages/ra-data-simple-rest/src/index.spec.ts
```diff
@@ -68,6 +68,65 @@ describe('Data Simple REST Client', () => {
...
+                'http://localhost:3000/posts/Post%23123',
+                expect.any(Object)
+            );
+        });
+    });
+    describe('updateMany', () => {
+        it('should escape id in path', async () => {
+            const httpClient = jest
+                .fn()
+                .mockResolvedValue({ json: { id: 'Post#123' } });
```
fzaninotto, line 116: This isn't a valid `updateMany` response and it will cause warnings.

packages/ra-data-simple-rest/src/index.spec.ts
```diff
@@ -120,5 +193,20 @@ describe('Data Simple REST Client', () => {
                 }
             );
         });
+        it('should escape id in path', async () => {
+            const httpClient = jest
+                .fn()
+                .mockResolvedValue({ json: { id: 'Post#123' } });
```
fzaninotto, line 199: This isn't a valid `updateMany` response

fzaninotto (changes requested):
Thanks for your PR! I found a few problems to be fixed before we can merge it.

packages/ra-data-simple-rest/src/index.spec.ts
ckhsponge: Fixing.

packages/ra-data-simple-rest/src/index.spec.ts
ckhsponge: Fixing.

packages/ra-data-simple-rest/src/index.spec.ts
ckhsponge: Fixing.

packages/ra-data-simple-rest/src/index.ts
ckhsponge: An existing deleteMany test had that as input so I assumed that was the format. Fixing.

ckhsponge:
Thanks for reviewing my PR!

I think an existing deleteMany test had an incorrect input which made me think objects were allowed as inputs there and for updateMany. I fixed that test and mine and removed that check from the data provider.

fzaninotto (approved):

fzaninotto:
Thanks!

PR #10209, opened 2024-09-17
[Doc] Doc needs to be consistent with itself regarding isPending vs isLoading

smeng9:
The doc states:

> As a consequence, you should always use isPending to determine if you need to show a loading indicator.

It should be consistent with itself.

## Probably there are some additional things needs to be discussed and fixed

1. Some of the stories still uses isLoading
2. useList still uses isLoading here https://github.com/marmelab/react-admin/blob/f2fc70f42ec8f7209e5fa24f12330276a7ddd788/packages/ra-core/src/controller/list/useList.ts#L173
3. In some enterprise components we pass isPending to the isLoading prop which is confusing https://github.com/marmelab/react-admin/blob/f2fc70f42ec8f7209e5fa24f12330276a7ddd788/docs/Features.md?plain=1#L988

## Additional Checks

### DIFF
--- a/docs/DataProviderWriting.md
+++ b/docs/DataProviderWriting.md
@@ -503,12 +503,12 @@ import { useDataProvider, Loading, Error } from 'react-admin';
 
 const UserProfile = ({ userId }) => {
     const dataProvider = useDataProvider();
-    const { data, isLoading, error } = useQuery({
+    const { data, isPending, error } = useQuery({
         queryKey: ['users', 'getOne', { id: userId }], 
         queryFn: ({ signal }) => dataProvider.getOne('users', { id: userId, signal })
     });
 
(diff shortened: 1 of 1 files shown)


djhi (approved):
Thanks!

fzaninotto (approved):
Thanks for your PR! And you're right, we have some work left on the documentation. 

In useList, checking `isLoading` is a mistake. Would you like to open a PR to fix it? 

Enterprise Edition components have recently been updated to react-admin v5 and should use `isPending`. The doc about EE components is probably not up to date (it is though at https://react-admin-ee.marmelab.com/whats-new). We'll do a pass to fix the remaining `isLoading`, but you're welcome to give us a hand!

PR #10213, opened 2024-09-17
[Doc] Update `<Filterbutton>`'s screencast

erwanMarmelab:
## Problem

`<Filterbutton>`'s doc was updated but the screencast was not.

## Solution

- [x] Record and update screencast
- [x] Remove `webm` and `gif` files
- [x] Update imports

## Additional Checks


docs/img/list_filter.mp4
fzaninotto, line 1: Something is wrong with this video. It's excrutiatingly slow. Also, the example doesn't make sense. A filter "bar" on the "nested" field?? I suggest you follow the exact same scenario as the original.

fzaninotto (changes requested):

djhi (approved):
The video looks good now

fzaninotto (approved):

PR #10218, opened 2024-09-17
Fix more cases that should use isPending

smeng9:
Additional follow ups of https://github.com/marmelab/react-admin/pull/9473

Update most of the docs to use isPending (except EE components, wait until the docs over [there](https://react-admin-ee.marmelab.com/documentation) to be updated)

Update stories and examples to use isPending

Update useList to use isPending


fzaninotto (approved):

fzaninotto:
Thanks!

PR #10220, opened 2024-09-18
[Doc] Update docs related to router to be consistent with migration guide

smeng9:
## Problem

Current docs is outdated, does not work [data APIs](https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis)

React-router recommends user to update to use new API.

## Solution

We should let user use [`warnWhenUnsavedChanges`](https://marmelab.com/react-admin/Upgrade.html#warnwhenunsavedchanges-changes) when they perform their setup with latest best practice

## How To Test

## Additional Checks

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -1042,9 +1042,9 @@ React-admin uses [the react-router library](https://reactrouter.com/) to handle 
 But you may want to use another routing strategy, e.g. to allow server-side rendering of individual pages. React-router offers various Router components to implement such routing strategies. If you want to use a different router, simply put your app in a create router function. React-admin will detect that it's already inside a router, and skip its own router. 
 
 ```tsx
-import { RouterProvider, createBrowserRouter } from "react-router-dom";
-import { Admin, Resource } from "react-admin";
-import { dataProvider } from "./dataProvider";
+import { RouterProvider, createBrowserRouter } from 'react-router-dom';
+import { Admin, Resource } from 'react-admin';
+import { dataProvider } from './dataProvider';
 
 const App = () => {
     const router = createBrowserRouter([
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -1035,44 +1035,59 @@ const App = () => (
 export default App;
 ```
 
-## Using A Custom Router 
+## Using A Custom Router
 
 React-admin uses [the react-router library](https://reactrouter.com/) to handle routing, with a [HashRouter](https://reactrouter.com/en/6/router-components/hash-router#hashrouter). This means that the hash portion of the URL (i.e. `#/posts/123` in the example) contains the main application route. This strategy has the benefit of working without a server, and with legacy web browsers. 
 
-But you may want to use another routing strategy, e.g. to allow server-side rendering of individual pages. React-router offers various Router components to implement such routing strategies. If you want to use a different router, simply wrap it around your app. React-admin will detect that it's already inside a router, and skip its own router. 
+But you may want to use another routing strategy, e.g. to allow server-side rendering of individual pages. React-router offers various Router components to implement such routing strategies. If you want to use a different router, simply put your app in a create router function. React-admin will detect that it's already inside a router, and skip its own router. 
 
 ```tsx
-import { BrowserRouter } from 'react-router-dom';
-import { Admin, Resource } from 'react-admin';
-import { dataProvider } from './dataProvider';
+import { RouterProvider, createBrowserRouter } from "react-router-dom";
+import { Admin, Resource } from "react-admin";
+import { dataProvider } from "./dataProvider";
 
-const App = () => (
-    <BrowserRouter>
-        <Admin dataProvider={dataProvider}>
-            <Resource name="posts" />
-        </Admin>
-    </BrowserRouter>
-);
+const App = () => {
+    const router = createBrowserRouter([
+        {
+            path: "*",
+            element: (
+                <Admin dataProvider={dataProvider}>
+                    <Resource name="posts" />
+                </Admin>
+            ),
+        },
+    ]);
+    return <RouterProvider router={router} />;
+};
 ```
 
 ## Using React-Admin In A Sub Path
 
 React-admin links are absolute (e.g. `/posts/123/show`). If you serve your admin from a sub path (e.g. `/admin`), react-admin works seamlessly as it only appends a hash (URLs will look like `/admin#/posts/123/show`).
 
-However, if you serve your admin from a sub path AND use another Router (like [`<BrowserRouter>`](https://reactrouter.com/en/main/router-components/browser-router) for instance), you need to set the `<Admin basename>` prop, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
+However, if you serve your admin from a sub path AND use another Router (like [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) for instance), you need to set the `<Admin basename>` prop and the [`opts.basename`](https://reactrouter.com/en/main/routers/create-browser-router#optsbasename) of `createBrowserRouter` function, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
 
 ```tsx
 import { Admin, Resource } from 'react-admin';
-import { BrowserRouter } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks for your PR.

The current doc mention a way to use a regular router, and it works. It is true that if you want to use an data router, the doc needs to be updated. But I'm not sure that *replacing*  the doc with the data router version is the best thing to do, as the setup is a bit more complicate din this case. 

Any second opinion on that?

slax57:
@fzaninotto Using a regular router works, sure, except regarding the `warnWhenUnsavedChanges` feature. This may come as a (bad) suprise to users willing to use that feature in the future.

Hence, I would agree with @smeng9 that it's best to encourage using a data router as soon as possible in the docs, to minimize the chances of users running into the issue of the non-working `warnWhenUnsavedChanges` feature.

fzaninotto:
You've convinced me! Let's merge this.

docs/Routing.md
```diff
@@ -125,74 +125,103 @@ export default App;
...
+                path: "*",
+                element: (
+                    <Routes>
+                        <Route path="/" element={<StoreFront />} />
+                        <Route path="/admin/*" element={<StoreAdmin />} />
+                    </Routes>
+                ),
+            },
+        ],
+        { basename: "/admin" },
```
fzaninotto, line 209: Are you sure about this one?

docs/Routing.md
```diff
@@ -125,74 +125,103 @@ export default App;
...
+import { RouterProvider, createBrowserRouter } from 'react-router-dom';
+import { dataProvider } from './dataProvider';
+
+const App = () => {
+    const router = createBrowserRouter(
+        [
+            {
+                path: "*",
+                element: (
+                    <Admin basename="/admin" dataProvider={dataProvider}>
```
fzaninotto, line 169: Do we need to declare the basename in Admin if it's already declared in the router?

fzaninotto (changes requested):

docs/Routing.md
```diff
@@ -125,74 +125,103 @@ export default App;
 
 ## Using A Custom Router
 
-By default, react-admin creates a [HashRouter](https://reactrouter.com/en/6/router-components/hash-router#hashrouter). The hash portion of the URL (i.e. `#/posts/123` in the example) contains the main application route. This strategy has the benefit of working without a server, and with legacy web browsers. 
+React-admin uses [the react-router library](https://reactrouter.com/) to handle routing, with a [HashRouter](https://reactrouter.com/en/6/router-components/hash-router#hashrouter). This means that the hash portion of the URL (i.e. `#/posts/123` in the example) contains the main application route. This strategy has the benefit of working without a server, and with legacy web browsers.
```
slax57, line 128:
```suggestion
React-admin uses [the react-router library](https://reactrouter.com/) to handle routing, with a [HashRouter](https://reactrouter.com/en/routers/create-hash-router). This means that the hash portion of the URL (i.e. `#/posts/123` in the example) contains the main application route. This strategy has the benefit of working without a server, and with legacy web browsers. 
```

docs/Routing.md
slax57: Yes we do.
`<Admin basename>` is used to do both:
- Set the basename in the default router (being createHashRouter)
- Add the basename as prefix to all redirections happening in react-admin (e.g. clicking on a CreateButton)

When we provide our own router, we need to set the basename to our own router, but the `<Admin basename>` is still required to compute the redirections properly.

docs/Admin.md
```diff
@@ -1086,18 +1101,27 @@ If you want to use react-admin as a sub path of a larger React application, chec
...
+                path: "*",
+                element: (
+                    <Routes>
+                        <Route path="/" element={<StoreFront />} />
+                        <Route path="/admin/*" element={<StoreAdmin />} />
+                    </Routes>
+                ),
+            },
+        ],
+        { basename: "/admin" },
```
slax57, line 1121: same here

smeng9: Removed the misadded basename

docs/Routing.md
smeng9: updated link

slax57 (approved):
Thanks!

docs/Routing.md
slax57: Okay so I actually tested this case locally, and it turns out @fzaninotto was right! `<Admin basename>` must be left empty, otherwise the path is added twice (`/admin/admin/posts/1/show`).

```suggestion
                    <Admin dataProvider={dataProvider}>
```

docs/Admin.md
```diff
@@ -1035,44 +1035,59 @@ const App = () => (
...
-        </Admin>
-    </BrowserRouter>
-);
+const App = () => {
+    const router = createBrowserRouter(
+        [
+            {
+                path: "*",
+                element: (
+                    <Admin basename="/admin" dataProvider={dataProvider}>
```
slax57, line 1081: same here

```suggestion
                    <Admin dataProvider={dataProvider}>
```

slax57 (changes requested):

docs/Routing.md
```diff
@@ -125,74 +125,102 @@ export default App;
...
+    return <RouterProvider router={router} />;
+};
 ```
 
 ## Using React-Admin In A Sub Path
 
 React-admin links are absolute (e.g. `/posts/123/show`). If you serve your admin from a sub path (e.g. `/admin`), react-admin works seamlessly as it only appends a hash (URLs will look like `/admin#/posts/123/show`).
 
-However, if you serve your admin from a sub path AND use another Router (like `BrowserRouter` for instance), you need to set the `<BrowserRouter basename>` prop, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
+However, if you serve your admin from a sub path AND use another Router (like [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) for instance), you need to set the `<Admin basename>` prop and the [`opts.basename`](https://reactrouter.com/en/main/routers/create-browser-router#optsbasename) of `createBrowserRouter` function, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
```
slax57, line 156: more changes I forgot (sorry for the spam!)

```suggestion
However, if you serve your admin from a sub path AND use another Router (like [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) for instance), you need to set the [`opts.basename`](https://reactrouter.com/en/main/routers/create-browser-router#optsbasename) of `createBrowserRouter` function, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
```

docs/Admin.md
```diff
@@ -1035,44 +1035,59 @@ const App = () => (
...
+    return <RouterProvider router={router} />;
+};
 ```
 
 ## Using React-Admin In A Sub Path
 
 React-admin links are absolute (e.g. `/posts/123/show`). If you serve your admin from a sub path (e.g. `/admin`), react-admin works seamlessly as it only appends a hash (URLs will look like `/admin#/posts/123/show`).
 
-However, if you serve your admin from a sub path AND use another Router (like [`<BrowserRouter>`](https://reactrouter.com/en/main/router-components/browser-router) for instance), you need to set the `<Admin basename>` prop, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
+However, if you serve your admin from a sub path AND use another Router (like [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) for instance), you need to set the `<Admin basename>` prop and the [`opts.basename`](https://reactrouter.com/en/main/routers/create-browser-router#optsbasename) of `createBrowserRouter` function, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
```
slax57, line 1068: more changes I forgot (sorry for the spam!)

```suggestion
However, if you serve your admin from a sub path AND use another Router (like [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) for instance), you need to set the [`opts.basename`](https://reactrouter.com/en/main/routers/create-browser-router#optsbasename) of `createBrowserRouter` function, so that react-admin routes include the basename in all links (e.g. `/admin/posts/123/show`).
```

slax57 (approved):

PR #10225, opened 2024-09-20
Add access control to views and action buttons

djhi:
## Problem

Giving access to views depending on users rights requires a lot of code.

## Solution

Do it automatically when the authProvider supports `canAccess`

## How To Test

Added stories for each button

## Additional Checks


packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.stories.tsx
```diff
@@ -46,17 +47,19 @@ const AuthorField = () => <TextField label="Author" source="author" />;
...
         <ThemeProvider theme={theme}>
             <PreferencesEditorContextProvider>
-                <TestMemoryRouter>
-                    <Inspector />
-                    <Box display="flex" justifyContent="flex-end">
-                        <InspectorButton />
-                    </Box>
-                    <Box p={2}>{children}</Box>
-                </TestMemoryRouter>
+                <QueryClientProvider client={queryClient}>
```
djhi, line 54: This is because the buttons (EditButton here) requires react-query to check access rights

packages/ra-ui-materialui/src/list/datagrid/SelectColumnsButton.stories.tsx
```diff
@@ -41,50 +42,54 @@ const theme = createTheme();
 export const Basic = () => (
     <ThemeProvider theme={theme}>
         <PreferencesEditorContextProvider>
-            <TestMemoryRouter>
-                <Box p={2}>
-                    <Box textAlign="right">
-                        <SelectColumnsButton resource="books" />
+            <QueryClientProvider client={new QueryClient()}>
```
djhi, line 45: Same

packages/ra-core/src/types.ts
```diff
@@ -69,11 +69,11 @@ export type AuthProvider = {
     handleCallback?: (
         params?: QueryFunctionContext
     ) => Promise<AuthRedirectResult | void | any>;
-    canAccess?: (
+    canAccess?: <RecordType extends RaRecord | Omit<RaRecord, 'id'> = RaRecord>(
```
fzaninotto, line 72: I wouldn't use RaRecord here, as it's possible to use an empty object.

fzaninotto (changes requested):

packages/ra-core/src/types.ts
djhi: `Omit<RaRecord, 'id'>` is basically any object

fzaninotto (approved):

PR #10227, opened 2024-09-20
Add access control to `<Datagrid rowClick>`

djhi:
## Problem

In a `<Datagrid>`, the default `rowClick` action depends on the availability of an `edit` and a `show` view. But if developers enabled Access Control, this default action may conflict with permissions. 

For instance, the default row click may be `edit`, but the current user may not have access to the `{ action: "edit", resource}` permission. In this case, the row click is enabled, but when users click on a row, they see the access denied page. 

This isn't a good user experience. In this particular case, they shouldn't be able to click on the row in the first place. 

## Solution

The `<Datagrid>` should check access control before setting the default `rowClick` action.

Part of https://github.com/marmelab/react-admin/pull/10222

## How To Test

## Additional Checks

Part of #10222


packages/ra-ui-materialui/src/list/datagrid/DatagridConfigurable.stories.tsx
```diff
@@ -47,19 +48,21 @@ const AuthorField = () => <TextField label="Author" source="author" />;
...
-                    <Inspector />
-                    <Box display="flex" justifyContent="flex-end">
-                        <InspectorButton />
-                    </Box>
-                    <Box p={2}>{children}</Box>
-                </TestMemoryRouter>
-            </PreferencesEditorContextProvider>
-        </ThemeProvider>
-    </StoreContextProvider>
+    <QueryClientProvider client={new QueryClient()}>
```
djhi, line 51: `QueryClientProvider` is now necessary as `rowClick` uses it

packages/ra-ui-materialui/src/list/datagrid/SelectColumnsButton.stories.tsx
```diff
@@ -39,52 +40,56 @@ const data = [
 const theme = createTheme();
 
 export const Basic = () => (
-    <ThemeProvider theme={theme}>
-        <PreferencesEditorContextProvider>
-            <TestMemoryRouter>
-                <Box p={2}>
-                    <Box textAlign="right">
-                        <SelectColumnsButton resource="books" />
+    <QueryClientProvider client={new QueryClient()}>
```
djhi, line 43: Same

docs/useCanAccessCallback.md
fzaninotto, line 1: needs an entry in the navigation and the Reference

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -50,36 +51,36 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
+        const updatePath = async () => {
+            const resolvedLink = await getPathForRecord({
+                record,
+                resource,
+                link,
+            });
+            if (resolvedLink && ['edit', 'show'].includes(resolvedLink)) {
+                if (
+                    !(await canAccess({
+                        action: resolvedLink,
```
fzaninotto, line 70: `resolvedLink` is like `/posts/123`, it's not `edit` or `show`, am I right? If so, it's not what you want to use for the `canAccess` action.

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -50,36 +51,36 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
-        }).then(resolvedLink => {
+        const updatePath = async () => {
+            const resolvedLink = await getPathForRecord({
+                record,
+                resource,
+                link,
+            });
+            if (resolvedLink && ['edit', 'show'].includes(resolvedLink)) {
+                if (
+                    !(await canAccess({
```
fzaninotto, line 69: I don't understand why you check for access here as it's already done in the callback hook

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
```diff
@@ -41,19 +44,41 @@ export const useGetPathForRecordCallback = <
...
-                        return;
+                const resolvedLink = await linkResult;
+                if (resolvedLink === false) {
+                    // already set to false by default
+                    return;
+                }
+                if (['edit', 'show'].includes(resolvedLink)) {
+                    if (
+                        !(await canAccess({
+                            action: resolvedLink,
```
fzaninotto, line 55: `resolvedLink` is like `/posts/123`, it's not `edit` or `show`, am I right? If so, it's not what you want to use for the `canAccess` action.

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
```diff
@@ -41,19 +44,41 @@ export const useGetPathForRecordCallback = <
...
+                    resource: finalResource,
+                    id: record.id,
+                    type: resolvedLink,
                 });
             }
 
+            if (linkResult !== false && ['edit', 'show'].includes(linkResult)) {
+                if (
+                    !(await canAccess({
+                        action: linkResult,
```
fzaninotto, line 73: same

fzaninotto (changes requested):
This one requires a bit more work.

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
djhi: Actually no, not here. We use `resolvedLink` as the `type` for `createPath` below

djhi: Besides, I think we should actually check for both `edit` and `show` when both are available as you may have access to one but not the other and that may change the `resolvedLink`

packages/ra-core/src/routing/useGetPathForRecord.ts
djhi: Removed

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
djhi: As discussed, this is very likely not a case

docs/useCanAccessCallback.md
djhi: Added (for others as well)

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -52,33 +52,21 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
-                // already set to false by default
-                return;
-            }
-            // update the path when the promise resolves
-            setPath(resolvedLink);
-        });
-
-        return false;
-    });
+    const [path, setPath] = useState<string | false | undefined>();
```
fzaninotto, line 55: The path was initialized with `false`, now it's `undefined`. If it's on purpose, can you explain why?

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -52,33 +52,21 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
+                record,
+                resource,
+                link,
+            });
             // update the path when the promise resolves
             setPath(resolvedLink);
-        });
+        };
+
+        updatePath();
```
fzaninotto, line 69: My understanding is that you changed nothing here (except Promise => await). I don't think the result is better and it brings some noise in this PR, would you mind reverting this change?

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -52,33 +52,21 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
     const getPathForRecord = useGetPathForRecordCallback<RecordType>(options);
 
     // we initialize the path with the link value
```
fzaninotto, line 54: This is no longer true

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
fzaninotto: I don't think we should do the access control check here, as we don't know if the link is a default one or one specified by the developer. If it's the default, it's the responsibility of this hook to determine a default compatible with access control. If it's one specified by the developer, it's their responsibility and we should not do any access control. 

Instead, you should check whether the user has access to show and edit views preemptively, and use the result when computing defaultLink above.

fzaninotto (changes requested):

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -51,23 +51,7 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
...
-    const [path, setPath] = useState<string | false | undefined>(() => {
-        getPathForRecord({
-            record,
-            resource,
-            link,
-        }).then(resolvedLink => {
-            if (resolvedLink === false) {
-                // already set to false by default
-                return;
-            }
```
fzaninotto, line 64: We should probably keep this escape hatch in the effect below

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
```diff
@@ -23,46 +26,74 @@ export const useGetPathForRecordCallback = <
...
+                        resource: finalResource,
+                        record,
+                    }))
+                ) {
                     return createPath({
                         resource: finalResource,
                         id: record.id,
-                        type: resolvedLink,
+                        type: 'edit',
                     });
```
fzaninotto, line 63: This is nice, but this code will do two subsequent calls to `canAccess` in many cases. Can we do the two checks (`edit` and `show`) in parallel, then set the default value?

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -612,6 +615,23 @@ describe('<ReferenceField />', () => {
             });
         });
     });
+    describe('Security', () => {
+        it('should render a link only when users have access to the requested action for the referenced resource', async () => {
+            render(<AccessControl />);
```
fzaninotto, line 620: This test is slow because of the default auth provider delay. Instead, build a custom auth provider in this test that uses a suspended Promise, pass this auth provider to the story, and resolve the Promise manually later in the test.

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -612,6 +615,23 @@ describe('<ReferenceField />', () => {
             });
         });
     });
+    describe('Security', () => {
+        it('should render a link only when users have access to the requested action for the referenced resource', async () => {
```
fzaninotto, line 619: You're not actually testing that. The test should check:

1. the default link is show
2. disabling show access turns the default link to edit
3. Disabling both show and edit should disable the default link

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -612,6 +615,23 @@ describe('<ReferenceField />', () => {
...
+            render(<AccessControl />);
+            await waitFor(
+                () => {
+                    expect(
+                        screen
+                            .getByText('Lewis Carroll', {
+                                selector: 'a > span',
+                            })
+                            .parentElement?.getAttribute('href')
+                    ).toEqual('/authors/5/show');
```
fzaninotto, line 629: Instead of testing the link value, click on it and check that you're on the right page.

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -612,6 +615,23 @@ describe('<ReferenceField />', () => {
...
+                },
+                { timeout: 8000 }
+            );
+        }, 10000);
```
fzaninotto, line 633: Test smell(TM)

packages/ra-core/src/routing/useGetPathForRecord.ts
```diff
@@ -51,23 +51,7 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
     }
     const getPathForRecord = useGetPathForRecordCallback<RecordType>(options);
 
-    // we initialize the path with the link value
-    const [path, setPath] = useState<string | false | undefined>(() => {
```
fzaninotto, line 55: I wonder: Why did we add this callback in the first place since the effect below does the same thing? Some `git blame` may reveal a hidden reason.

packages/ra-core/src/routing/useGetPathForRecord.stories.tsx
fzaninotto, line 1: I would also love to see a story on `<Datagrid>` showing real usage.

packages/ra-core/src/routing/useGetPathForRecord.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+export const AccessControlWithLinkTypeProvided = ({
+    authProvider = {
+        login: () => Promise.resolve(),
+        logout: () => Promise.resolve(),
+        checkAuth: () => Promise.resolve(),
+        checkError: () => Promise.resolve(),
+        getPermissions: () => Promise.resolve(),
+        canAccess: ({ action }) =>
+            new Promise(resolve => setTimeout(resolve, 300, action === 'edit')),
```
fzaninotto, line 96: tests will be slow because of that. Can you use a suspended Promise in tests instead?

fzaninotto (changes requested):
You also need to update the `<Datagrid>` and `<ReferenceField>` docs to mention access control.

packages/ra-core/src/routing/useGetPathForRecord.stories.tsx
djhi: I based the test on another story

packages/ra-core/src/routing/useGetPathForRecord.ts
djhi: Done in https://github.com/marmelab/react-admin/commit/d57d112bcca00db4ea499537d663b4496c8f7982 when we added support for async link functions. I believe it should not have been merged as is

packages/ra-core/src/routing/useGetPathForRecord.ts
djhi: Not anymore. As we use `useQuery`, the result may change over time.

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
```diff
@@ -23,46 +26,76 @@ export const useGetPathForRecordCallback = <
...
-                return linkResult.then(resolvedLink => {
-                    if (resolvedLink === false) {
-                        // already set to false by default
-                        return;
-                    }
+            // When the link prop is not provided, we infer a default value and check whether users
+            // can access it
+            if (link == null) {
+                const [canAccessShow, canAccessEdit] = await Promise.all([
+                    resourceDefinition.hasShow
```
fzaninotto, line 38: Why do you test `resourceDefinition.hasShow` here and also L54?

fzaninotto (approved):

packages/ra-core/src/routing/useGetPathForRecordCallback.ts
djhi: Because you many not have an authProvider and in this case, useCanAccess will always return true

PR #10228, opened 2024-09-22
[ui-materialui]: Hide `<AutoCompleteInput>` "Create" option for empty input

### LINKED ISSUE #10043: `<Autocomplete onCreate>` briefly displays wrong option name
**What you were expecting:**
When creating a new option on-the-fly using `<Autocomplete onCreate>`, the selected option should be the new option. 

**What happened instead:**
The selected option switches to "Create XXX" before showing the correct "XXX"

https://github.com/user-attachments/assets/94bf4e77-03b3-4e2b-a9f1-b678e32b9f49

**Steps to reproduce:**

1. Open the CRM demo and create a new contact (https://marmelab.com/react-admin-crm/#/contacts/create)
2. Enter "Acme" in the company name
3. Select "Create Acme" in the list of options

**Related code:**

https://github.com/marmelab/react-admin/blob/a9f46489453a29f0e2fa8c3c201c928c2825aef8/examples/crm/src/contacts/ContactInputs.tsx#L31-L52

**Environment**

* React-admin version: v5
* Last version that did not exhibit the issue (if applicable): n/a
* React version: 18
* Browser: Chrome

erwanMarmelab:
## Problem

The “create option on-the-fly” UI of Autocomplete inputs is confusing. I can create an option with "" as value.
Fixes #10043

## Solution

- [x] Don’t display “Create” on the input unless the user types something
- [x] Fix the brief flash of `Create xxx` -> `xxx`

## How To Test

- run `ONLY=ra-ui-materialui make storybook`
- go to `http://localhost:9010/?path=/story/ra-ui-materialui-input-autocompleteinput--on-create`
- clear the input and see that you cannot create an empty option

## Additional Checks

~- [ ] The PR includes one or several **stories** (if not possible, describe why)~

## Screencasts

Before flash fix:
[Capture vidéo du 2024-09-25 19-42-25.webm](https://github.com/user-attachments/assets/824e9575-9473-4dff-8663-f06d8d4d6f1d)
After:
[Capture vidéo du 2024-09-25 19-43-16.webm](https://github.com/user-attachments/assets/12f1c759-60f6-4f12-90a9-2d7aa8fcaf7e)


packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -465,7 +465,13 @@ If you provided a React element for the optionText prop, you must also provide t
             event?.type === 'change' ||
             !doesQueryMatchSelection(newInputValue)
         ) {
-            setFilterValue(newInputValue);
+            const createOption = translate(createItemLabel, {
```
djhi, line 468:
```suggestion
            const createOptionLabel = translate(createItemLabel, {
```

packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
```diff
@@ -465,7 +465,13 @@ If you provided a React element for the optionText prop, you must also provide t
...
+                item: filterValue,
+                _: createItemLabel,
+            });
+            const isCreate = newInputValue === createOption;
```
djhi, line 472:
```suggestion
            const isCreate = newInputValue === createOptionLabel;
```

djhi (changes requested):
Looks good otherwise :muscle:

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -357,33 +357,42 @@ export const CreateLabel = () => (
...
-                        name: newAuthorName,
-                    };
-                    choicesForCreationSupport.push(newAuthor);
-                    return newAuthor;
-                }
+                const newOption = {
+                    id: choicesForCreationSupport.length + 1,
+                    name: filter,
+                };
+                choicesForCreationSupport.push(newOption);
```
slax57, line 368: I've had issues while testing stories like this one, or OnCreate, which I think are due to the fact that running `choicesForCreationSupport.push(newOption)` does not always trigger a rerender.
I believe `choicesForCreationSupport` should be turned into a React State to avoid such problems.

In the following recording, the first item I create ('new') does not appear immediately, nor does it appear in the list. But when I create a second item ('new2'), then, suddenly, both items appear in the list.


https://github.com/user-attachments/assets/558c9aae-6786-4f1d-a67c-3a7ed4114feb

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -357,33 +357,42 @@ export const CreateLabel = () => (
...
-                { id: 3, name: 'William Shakespeare' },
-                { id: 4, name: 'Charles Baudelaire' },
-                { id: 5, name: 'Marcel Proust' },
-            ]}
+            choices={choicesForCreationSupport}
             onCreate={filter => {
-                const newAuthorName = window.prompt(
-                    'Enter a new author',
-                    filter
-                );
```
slax57, line 371: Removing the prompt from this story (which is a good thing IMO because it was redundant with the ability to fill in the filter), reveals a pre-existing odd behavior: if I click on the "Start typing to create a new item" label, then the value is filled with "Start typing to create a new item", which is probably not what I want. This special entry in the list should not be clickable IMO.
But this can be addressed in another PR since it was already existing.


https://github.com/user-attachments/assets/2a2f2e60-4275-4dad-97a7-7f3c2ff19258

slax57 (changes requested):
Other than that it seems to work great! Well done!

PR #10230, opened 2024-09-23
ui-materialui: Add support for `<Edit emptyWhileLoading>`

erwanMarmelab:
## Problem

The Show and List components support that prop, not the Edit.

## Solution

Add support for `<Edit emptyWhileLoading>`

## Additional Checks

~- [ ] The PR includes **unit tests** (if not possible, describe why)~

### DIFF
--- a/docs/Edit.md
+++ b/docs/Edit.md
@@ -263,7 +263,7 @@ const AsideComponent = () => {
 }
 ```
 
-The `<Edit emptyWhileLoading>` prop provides a convenient shortcut for that use case. When enabled, `<Edit>` won't render its child until `data` is defined.
+The `<Edit emptyWhileLoading>` prop provides a convenient shortcut for that use case. When enabled, `<Edit>` won't render its `aside`, `actions` and `title` components until `record` is defined.
 
 ```diff
 const BookEdit = () => (
--- a/docs/Edit.md
+++ b/docs/Edit.md
@@ -222,6 +222,57 @@ const PostEdit = () => (
 );
 ```
 
+## `emptyWhileLoading`
+
+By default, `<Edit>` renders its child component even before the `dataProvider.getOne()` call returns. And default layout components (`<Datagrid>` and `<SimpleList>`) return null when the data is loading. If you use a custom layout component instead, you'll have to handle the case where the `data` is not yet defined.
+
+But if you use a custom child component that expects the record context to be defined, your component will throw an error. For instance, the following will fail on load with a "ReferenceError: data is not defined" error:
+
+```jsx
+import { Edit, useEditContext } from 'react-admin';
+import { Stack, Typography } from '@mui/icons-material/Star';
+
+const SimpleBookEdit = () => {
+    const { record } = useEditContext();
+    return (
+        <Typography>
+            <i>{record.title}</i>, by {record.author} ({record.year})
+        </Typography>
+    );
+}
+
+const BookEdit = () => (
+    <Edit>
+        <SimpleBookEdit />
+    </Edit>
+);
+```
+
+You can handle this case by getting the `isPending` variable from the [`useEditContext`](./useEditContext.md) hook:
+
+```jsx
+const SimpleBookEdit = () => {
+    const { record, isPending } = useEditContext();
+    if (isPending) return null;
+    return (
+        <Typography>
+            <i>{record.title}</i>, by {record.author} ({record.year})
+        </Typography>
+    );
+}
+```
+
+The `<Edit emptyWhileLoading>` prop provides a convenient shortcut for that use case. When enabled, `<Edit>` won't render its child until `data` is defined.
+
+```diff
+const BookEdit = () => (
+-   <Edit>
++   <Edit emptyWhileLoading>
+        <SimpleBookEdit />
+    </Edit>
+);
+```
+
 ## `id`
 
 Components based on `<Edit>` are often used as `<Resource edit>` props, and therefore rendered when the URL matches `/[resource]/[id]`. The `<Edit>` component generates a call to `dataProvider.update()` using the id from the URL by default.
--- a/docs/Edit.md
+++ b/docs/Edit.md
@@ -66,6 +66,7 @@ You can customize the `<Edit>` component using the following props:
 * `className`: passed to the root component
 * [`component`](#component): override the root component
 * [`disableAuthentication`](#disableauthentication): disable the authentication check
+* [`emptyWhileLoading`](#emptywhileloading): Set to `true` to return `null` while the edit is loading.
 * [`id`](#id): the id of the record to edit
 * [`mutationMode`](#mutationmode): switch to optimistic or pessimistic mutations (undoable by default)
 * [`mutationOptions`](#mutationoptions): options for the `dataProvider.update()` call
@@ -224,7 +225,7 @@ const PostEdit = () => (
 
 ## `emptyWhileLoading`
 
-By default, `<Edit>` renders its child component even before the `dataProvider.getOne()` call returns. And default layout components (`<Datagrid>` and `<SimpleList>`) return null when the data is loading. If you use a custom layout component instead, you'll have to handle the case where the `data` is not yet defined.
+By default, `<Edit>` renders its child component even before the `dataProvider.getOne()` call returns. And default layout components return null when the data is loading. If you use a custom layout component instead, you'll have to handle the case where the `data` is not yet defined.
 
 But if you use a custom child component that expects the record context to be defined, your component will throw an error. For instance, the following will fail on load with a "ReferenceError: data is not defined" error:
 
--- a/docs/Edit.md
+++ b/docs/Edit.md
@@ -225,15 +225,15 @@ const PostEdit = () => (
 
 ## `emptyWhileLoading`
 
-By default, `<Edit>` renders its child component even before the `dataProvider.getOne()` call returns. And default layout components return null when the data is loading. If you use a custom layout component instead, you'll have to handle the case where the `data` is not yet defined.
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/detail/Edit.stories.tsx
```diff
@@ -300,3 +300,45 @@ export const Default = () => (
         </Admin>
     </TestMemoryRouter>
 );
+
+export const emptyWhileLoading = () => {
```
djhi, line 304:
```suggestion
export const EmptyWhileLoading = () => {
```

djhi (changes requested):
Can you also add a test?

docs/Edit.md
```diff
@@ -222,6 +222,57 @@ const PostEdit = () => (
 );
 ```
 
+## `emptyWhileLoading`
+
+By default, `<Edit>` renders its child component even before the `dataProvider.getOne()` call returns. And default layout components (`<Datagrid>` and `<SimpleList>`) return null when the data is loading. If you use a custom layout component instead, you'll have to handle the case where the `data` is not yet defined.
```
fzaninotto, line 227: Why mention `Datagrid` and `SimpleList` here? We're not in a list page.

docs/Edit.md
```diff
@@ -222,6 +222,57 @@ const PostEdit = () => (
...
+
+But if you use a custom child component that expects the record context to be defined, your component will throw an error. For instance, the following will fail on load with a "ReferenceError: data is not defined" error:
```
fzaninotto, line 229: The first sentence says more or less the same as the previous one... not clear. The second mentions an error about undefined data but there is no data in the snippet.

docs/Edit.md
```diff
@@ -222,6 +222,57 @@ const PostEdit = () => (
 );
 ```
 
+## `emptyWhileLoading`
```
fzaninotto, line 225: don't forget to add it to the list of props earlier in the doc

docs/EditTutorial.md
```diff
@@ -479,7 +479,7 @@ import * as React from "react";
 import { Edit, SimpleForm, TextInput, SelectInput } from "react-admin";
 
 export const BookEdit = () => (
-  <Edit>
+  <Edit emptyWhileLoading>
```
fzaninotto, line 482: I think you should revert the changes in this file. We don't need to introduce this prop, which is for corner case, in a general presentation of the Edit logic.

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -106,6 +107,13 @@ describe('<Edit />', () => {
         });
     });
 
+    it("shoudln't display the record while loading", async () => {
```
fzaninotto, line 110:
```suggestion
    it("shoudln't display the Edit child while loading when emptyWhileLoading is true", async () => {
```

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -106,6 +107,13 @@ describe('<Edit />', () => {
...
+        render(<EmptyWhileLoading />);
+        expect(screen.queryByText('Book War and Peace')).toBeNull();
+        await new Promise(resolve => setTimeout(resolve, 3000));
+        expect(screen.queryByText('Book War and Peace')).not.toBeNull();
```
fzaninotto, line 114: This test doesn't make sense. Of course the record won't show up during loading, because the data isn't there. The purpose of the `emptyWhileLoading` prop is to avoid rendering the child altogether. 
So you should add a static string to the Edit child and make sure it's not rendered while the data provider is loading. 

Also, don't use timeouts. Check https://github.com/marmelab/react-admin/pull/10238/files#diff-1f5141788db3c747028c3cb2da9eaa6f9244e63d77addb0ea7eb70688fcb9ba9 for an example of test that uses a suspended Promise resolved bu the test itself.

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -25,7 +26,7 @@ export const EditView = (props: EditViewProps) => {
 
     const finalActions =
         typeof actions === 'undefined' && hasShow ? defaultActions : actions;
-    if (!children) {
+    if (!children || (!record && emptyWhileLoading)) {
```
fzaninotto, line 29: you must also check the `pending` state, as in the other controllers.

fzaninotto (changes requested):

docs/Edit.md
```diff
@@ -222,6 +223,57 @@ const PostEdit = () => (
...
+    if (isPending) return null;
+    return (
+        <Typography>
+            <i>{record.title}</i>, by {record.author} ({record.year})
+        </Typography>
+    );
+}
+```
+
+The `<Edit emptyWhileLoading>` prop provides a convenient shortcut for that use case. When enabled, `<Edit>` won't render its child until `data` is defined.
```
erwanMarmelab, line 266:
```suggestion
The `<Edit emptyWhileLoading>` prop provides a convenient shortcut for that use case. When enabled, `<Edit>` won't render its `aside`, `actions` and `title` components until `record` is defined.
```

docs/EditTutorial.md
```diff
@@ -495,4 +495,7 @@ export const BookEdit = () => (
 
 React-admin components are not magic, they are React components designed to let you focus on the business logic and avoid repetitive tasks. 
 
-Tip: Actually, `<Edit>` does more than the code it replaces in the previous example: it handles notification and redirection upon submission, it sets the page title, and handles the error logic.
+**Tip:** Actually, `<Edit>` does more than the code it replaces in the previous example: it handles notification and redirection upon submission, it sets the page title, and handles the error logic.
+
+**Tip**: With `emptyWhileLoading` turned on, the `<Edit>` component doesn't render its child component until the record is available. Without this flag, the Field components would render even during the loading phase, and may break if they aren't planned to work with an empty record context.
```
erwanMarmelab, line 500:
```suggestion
**Tip**: With `emptyWhileLoading` turned on, the `<Edit>` component doesn't render its `aside`, `actions` and `title` components until the record is available. Without this flag, the Field components would render even during the loading phase, and may break if they aren't planned to work with an empty record context.
```

docs/Edit.md
```diff
@@ -222,6 +223,57 @@ const PostEdit = () => (
 );
 ```
 
+## `emptyWhileLoading`
+
+By default, `<Edit>` doesn't render its child component even before the `dataProvider.getOne()` call returns, but it returns its [`aside`](#aside), [`actions`](#actions) and [`title`](#title) components.
+
+If you use one of those components customized with a record, your component will throw an error. For instance, the following will fail on load with a "ReferenceError: record is not defined" error:
```
djhi, line 230:
```suggestion
If a component you use in those props relies on the record, it will throw an error. For instance, the following will fail on load with a "ReferenceError: record is not defined" error:
```

djhi (changes requested):

docs/EditTutorial.md
```diff
@@ -495,4 +495,7 @@ export const BookEdit = () => (
 
 React-admin components are not magic, they are React components designed to let you focus on the business logic and avoid repetitive tasks. 
 
-Tip: Actually, `<Edit>` does more than the code it replaces in the previous example: it handles notification and redirection upon submission, it sets the page title, and handles the error logic.
+**Tip:** Actually, `<Edit>` does more than the code it replaces in the previous example: it handles notification and redirection upon submission, it sets the page title, and handles the error logic.
+
+**Tip**: With `emptyWhileLoading` turned on, the `<Edit>` component doesn't render its `aside`, `actions` and `title` components until the record is available. Without this flag, those components would render even during the loading phase, and may break if they aren't planned to work with an empty record context.
```
fzaninotto, line 500: This is advanced configuration and we shouldn't mention it in the EditTutorial, which is a general introduction.

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -106,6 +107,27 @@ describe('<Edit />', () => {
...
+        expect(screen.queryByText('Something went wrong')).toBeNull();
+
+        await waitFor(
+            () => {
+                expect(
+                    screen.queryByText('War and Peace, by Leo Tolstoy (1869)')
+                ).not.toBeNull();
+            },
+            { timeout: 2500 }
+        );
```
fzaninotto, line 126: This will slow down the unit tests. Instead of relying on a timeout, inject a data provider with a suspended Promise that you resolve manually.

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -885,7 +907,12 @@ describe('<Edit />', () => {
                     i18nProvider={i18nProvider}
                 >
                     <ResourceDefinitionContextProvider
-                        definitions={{ foo: { recordRepresentation: 'title' } }}
+                        definitions={{
+                            foo: {
+                                recordRepresentation: 'title',
+                                name: '',
```
fzaninotto, line 913: does not make sense

packages/ra-ui-materialui/src/detail/Edit.spec.tsx
```diff
@@ -885,7 +919,9 @@ describe('<Edit />', () => {
                     i18nProvider={i18nProvider}
                 >
                     <ResourceDefinitionContextProvider
-                        definitions={{ foo: { recordRepresentation: 'title' } }}
+                        definitions={{
+                            foo: { recordRepresentation: 'title' },
```
djhi, line 923:
```suggestion
                            foo: { name: 'foo', recordRepresentation: 'title' },
```

I believe TS is angry about this

djhi (commented):
Would you mind fixing the Show documentation while you're at it? The `emptyWhileLoading` description in the props table is wrong:
> Set to true to return null while the list is loading

fzaninotto (approved):

PR #10242, opened 2024-09-27
[Doc] Mention warning about disabled inputs

fzaninotto:
I was bitten by that pesky effect, so we'd better document it loud and clear.

### DIFF
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -171,9 +171,7 @@ The `disabled` prop set to true makes the element not mutable, focusable, or eve
 
 Contrary to read-only controls, disabled controls can not receive focus and are not submitted with the form.
 
-**Warning:** Note that `disabled` inputs are **not** included in the form values, and hence may trigger `warnWhenUnsavedChanges` if the input previously had a value in the record.
-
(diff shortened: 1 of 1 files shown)


docs/Inputs.md
```diff
@@ -171,9 +171,7 @@ The `disabled` prop set to true makes the element not mutable, focusable, or eve
 
 Contrary to read-only controls, disabled controls can not receive focus and are not submitted with the form.
 
-**Warning:** Note that `disabled` inputs are **not** included in the form values, and hence may trigger `warnWhenUnsavedChanges` if the input previously had a value in the record.
-
-**Tip:** To include the input in the form values, you can use `readOnly` instead of `disabled`.
+**Warning:** **React-hook-form sets the value of `disabled` inputs to `undefined`**, which is probably not what you want. For instance, when used on the `id` source, the `disabled` prop will *erase* the `id`. Another example is that `disabled` inputs trigger the `warnWhenUnsavedChanges` warning if the input previously had a value in the record. To avoid this, use the `readOnly` prop instead.
```
fzaninotto, line 174: an update with the value `undefined` for a field should not update the field - this is a bug of FakeRest. The warning needs to be adjusted.

PR #10247, opened 2024-09-30
Add access control to controllers

djhi:
## Problem

Follow #10222 

## How To Test

Stories for each controller

## Additional Checks


fzaninotto:
Conditional menu items and buttons are out of the scope of this PR, right?

djhi:
> Conditional menu items and buttons are out of the scope of this PR, right?

Yes, next PR

docs/useRequireAccess.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: default
+title: "useRequireAccess"
+---
+
+# `useRequireAccess`
+
+This hook calls the `authProvider.canAccess()` method on mount for a provided resource and action (and optionally a record). If users don't have access, it will redirect them to the [`/unauthorized`](./Admin.md#unauthorized) page. If `authProvider.canAccess()` throws an error, it will redirect them to the [`/authentication-error`](./Admin.md#authenticationerror) page.
```
fzaninotto, line 8: Now I think the names of the routes are wrong. 

The first one should be `/access-denied`.

The second could be `/authorization-error`, but we may keep `/authentication-error` as a generic page for all errors related to the `authProvider`.

docs/useRequireAccess.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Usage
+
+`useRequireAccess` takes an object `{ action, resource, record }` as argument. It returns an object describing the state of the request. As calls to the `authProvider` are asynchronous, the hook returns a `isPending` state in addition to the `canAccess` key.
```
fzaninotto, line 12:
```suggestion
`useRequireAccess` takes an object `{ action, resource, record }` as argument. It returns an object describing the state of the request. As calls to the `authProvider` are asynchronous, the hook returns a `isPending` state.
```

We never care about the actual `canAccess` as it can't be false.

docs/useRequireAccess.md
```diff
@@ -0,0 +1,52 @@
...
+
+For instance, here's how you can protect a [custom route](./CustomRoutes.md) for editing users settings:
+
+```jsx
+import { useRequireAccess, useRecordContext, DeleteButton } from 'react-admin';
+
+export export const SettingsPage = () => {
+    const record = useRecordContext();
+    const { isPending } = useRequireAccess({ action: 'edit', resource: 'settings', record });
```
fzaninotto, line 21: Don't forget the error case.

docs/useRequireAccess.md
```diff
@@ -0,0 +1,52 @@
...
+## Parameters
+
+`useRequireAccess` expects a single parameter object with the following properties:
+
+| Name | Required | Type | Default | Description |
+| --- | --- | --- | --- | --- |
+| `resource` | Required | `string` | - | The resource to check, e.g. 'users', 'comments', 'posts', etc. |
+| `action` | Required | `string` | - | The action to check, e.g. 'read', 'list', 'export', 'delete', etc. |
+| `record` | Optional | `object` | - | The record to check. If passed, the child only renders if the user has permissions for that record, e.g. `{ id: 123, firstName: "John", lastName: "Doe" }` |
+
```
fzaninotto, line 52: I think we need to add a section about customization of error pages.

packages/ra-core/src/auth/CanAccess.tsx
```diff
@@ -1,50 +1,53 @@
...
 /**
  * A component that only displays its children after checking whether users are authorized to access the provided resource and action.
  * @param options
  * @param options.action The action to check. One of 'list', 'create', 'edit', 'show', 'delete', or a custom action.
  * @param options.resource The resource to check. e.g. 'posts', 'comments', 'users'
  * @param options.children The component to render if users are authorized.
- * @param options.loading An optional element to render while the authorization is being checked. Defaults to the loading component provided on `Admin`.
- * @param options.unauthorized An optional element to render if users are not authorized. Defaults to the unauthorized component provided on `Admin`.
+ * @param options.loading An optional element to render while the authorization is being checked. Defaults to null.
+ * @param options.unauthorized An optional element to render if users are not authorized. Defaults to null.
```
fzaninotto, line 13: I'd rename it to `accessDenied`. I know it's a pain to do, but since we named the functionality Access Control it makes more sense.

docs/navigation.html
```diff
@@ -64,6 +64,7 @@
     <li {% if page.path == 'useLogout.md' %} class="active" {% endif %}><a class="nav-link" href="./useLogout.html"><code>useLogout</code></a></li>
     <li {% if page.path == 'usePermissions.md' %} class="active" {% endif %}><a class="nav-link" href="./usePermissions.html"><code>usePermissions</code></a></li>
     <li {% if page.path == 'useCanAccess.md' %} class="active" {% endif %}><a class="nav-link" href="./useCanAccess.html"><code>useCanAccess</code><img class="premium" src="./img/premium.svg" /></a></li>
+    <li {% if page.path == 'useRequireAccess.md' %} class="active" {% endif %}><a class="nav-link" href="./useRequireAccess.html"><code>useRequireAccess</code><img class="premium" src="./img/premium.svg" /></a></li>
```
fzaninotto, line 67: it's not a premium feature. And by the way, `useCanAccess` isn't either.

packages/ra-core/src/auth/useCanAccess.spec.tsx
```diff
@@ -76,23 +76,18 @@ describe('useCanAccess', () => {
         });
     });
 
-    it('should return an error after a tick if the auth.canAccess call fails and checkError resolves', async () => {
+    it('should returns the error when auth.canAccess call fails', async () => {
```
fzaninotto, line 79:
```suggestion
    it('should return the error when auth.canAccess call fails', async () => {
```

packages/ra-core/src/auth/useRequireAccess.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+        });
+    });
+
+    it('should allow access on mount when there is no authProvider', () => {
+        render(<Basic authProvider={null} />);
+        expect(screen.queryByText('Loading')).toBeNull();
+        screen.getByText('Protected Content');
+    });
+
+    it('should allow its caller to render when canAccess return true', async () => {
```
fzaninotto, line 23:
```suggestion
    it('should allow access when canAccess return true', async () => {
```

packages/ra-core/src/auth/useRequireAccess.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+            checkAuth: () => Promise.reject('bad method'),
+            checkError: () => Promise.reject('bad method'),
+            getPermissions: () => Promise.reject('bad method'),
+            canAccess: () => Promise.resolve(true),
+        };
+        render(<Basic authProvider={authProvider} />);
+        await screen.findByText('Protected Content');
+    });
+
+    it('should allow its caller to render when auth provider does not have an canAccess method', async () => {
```
fzaninotto, line 36:
```suggestion
    it('should allow access when auth provider does not have an canAccess method', async () => {
```

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -60,13 +60,18 @@ export const useCreateController = <
...
     } = props;
 
-    useAuthenticated({ enabled: !disableAuthentication });
     const resource = useResourceContext(props);
     if (!resource) {
         throw new Error(
             'useCreateController requires a non-empty resource prop or context'
         );
     }
+    useAuthenticated({ enabled: !disableAuthentication });
```
fzaninotto, line 69: The authentication check should be done before the authorization check. It does not make sense to verify the user rights if they are not authenticated. Therefore, you must read the `isPending` from `useAuthenticated`, and use it in the `enabled` flag of the `useRequireAccess`.

Same for other controllers.

packages/ra-core/src/controller/create/useCreateController.ts
```diff
@@ -182,7 +187,7 @@ export const useCreateController = <
     return {
         isFetching: false,
         isLoading: false,
-        isPending: saving,
+        isPending: isPendingCanAccess,
```
fzaninotto, line 190: Should be something like `isPendingAuthentication || isPendingAccessControl`

packages/ra-core/src/controller/create/useCreateController.security.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+            { id: 4, title: 'Post #4', votes: 40 },
+            { id: 5, title: 'Post #5', votes: 50 },
+            { id: 6, title: 'Post #6', votes: 60 },
+            { id: 7, title: 'Post #7', votes: 70 },
+        ],
+    },
+    process.env.NODE_ENV === 'development'
+);
+
+const Post = (props: Partial<CreateControllerProps>) => {
```
fzaninotto, line 42: rename the component to `CreatePost`

packages/ra-core/src/controller/create/useCreateController.security.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+}: {
+    authProviderDelay?: number;
+    queryClient: QueryClient;
+}) => {
+    const [authorizedResources, setAuthorizedResources] = React.useState({
+        'posts.list': true,
+        'posts.create': true,
+        'posts.edit': false,
+        'posts.show': false,
+    });
```
fzaninotto, line 134: this seems overkill as you only ever modify the `create` permission

fzaninotto (changes requested):
Awesome!

docs/useRequireAccess.md
djhi: So `accessDenied` for the prop?

docs/useRequireAccess.md
djhi: Nope, `useRequireAccess` already handle the error case by redirecting users to `/unauthorized`. That's the difference with `useCanAccess`.

docs/useRequireAccess.md
djhi: It's in the main hook description already

packages/ra-core/src/controller/create/useCreateController.ts
djhi: As discussed, it's actually not needed

packages/ra-core/src/controller/create/useCreateController.security.stories.tsx
djhi: Keeping at least the list makes testing manually and automatically by navigating easier

fzaninotto (approved):

PR #10248, opened 2024-09-30
Fix List empty component wrongly appears when using partial pagination

### LINKED ISSUE #10240: Empty List when partial paging and items in list less than pageSize
**What you were expecting:**

Typical List with Datagrid work as described in documentation with partial pagination even if there are small amount of items in list (i.e. 1,2,3)

**What happened instead:**

After upgrade to 5.2.1 listview displays empty widget when amount of items in list is less or equal to current pageSize.
In that case getList result total field is null due to partial pagination, also hasPreviousPage and hasNextPage are both false, leading following code to show empty widget even when data is not empty.

Temporary workaround is to artificially set total: in dataprovider when hasPrevious === false && hasNext === false

**Related code:**

Code in react-admin/packages/ra-ui-materialui/src/list/ListView.tsx
```
const shouldRenderEmptyPage =
        // the list is not loading data for the first time
        !isPending &&
        // the API returned no data (using either normal or partial pagination)
        (total === 0 ||
            (total == null &&
                hasPreviousPage === false &&
                hasNextPage === false)) &&
        // the user didn't set any filters
        !Object.keys(filterValues).length &&
        // there is an empty page component
        empty !== false;
```

**Environment**

* React-admin version: 5.2.1

fzaninotto:
Closes #10240

### DIFF
--- a/packages/ra-ui-materialui/src/list/List.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/List.stories.tsx
@@ -342,7 +342,7 @@ export const PartialPagination = () => (
         <Admin
             dataProvider={
                 {
-                    getList: async (resource, params) => ({
+                    getList: async (_resource, _params) => ({
                         data: [
                             {
                                 id: 1,
@@ -391,7 +391,7 @@ export const EmptyPartialPagination = () => (
         <Admin
             dataProvider={
                 {
-                    getList: async (resource, params) => ({
+                    getList: async (_resource, _params) => ({
                         data: [],
                         pageInfo: {
                             hasNextPage: false,
--- a/packages/ra-ui-materialui/src/list/List.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/List.stories.tsx
@@ -5,6 +5,7 @@ import {
     Resource,
     useListContext,
     TestMemoryRouter,
+    DataProvider,
 } from 'ra-core';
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/List.stories.tsx
```diff
@@ -336,6 +337,39 @@ export const Component = () => (
     </TestMemoryRouter>
 );
 
+export const PartialPagination = () => (
+    <TestMemoryRouter initialEntries={['/authors']}>
+        <Admin
+            dataProvider={
+                {
+                    getList: async (resource, params) => ({
```
djhi, line 345:
```suggestion
                    getList: async (_resource, _params) => ({
```

packages/ra-ui-materialui/src/list/List.stories.tsx
```diff
@@ -352,6 +386,34 @@ export const Empty = () => (
     </TestMemoryRouter>
 );
 
+export const EmptyPartialPagination = () => (
+    <TestMemoryRouter initialEntries={['/authors']}>
+        <Admin
+            dataProvider={
+                {
+                    getList: async (resource, params) => ({
```
djhi, line 394:
```suggestion
                    getList: async (_resource, _params) => ({
```

djhi (changes requested):

djhi (approved):

PR #10249, opened 2024-09-30
Fix `disabled` inputs temporarily erase values

fzaninotto:
## Problem

React-hook-form sets the value of `disabled` inputs to `undefined`. For most data providers, this isn't a problem (except the disabled field won't be sent in the update).

But FakeRest is local, so it receives the `undefined` value, and uses it to override the existing value. In effect, this erases the value.

```jsx
dataProvider.update(123, data: { id: undefined }); // erases the id with FakeRest
```

I also noticed that we made the same mistake in optimistic updates.

This can be seen e.g. in the simple example, by changing the first input of PostEdit to `<TexuInput source="id" disabled" />`

## Solution

Remove `undefined` values from the edit payload:

- [x] In `ra-data-fakerest`
- [x] in `useUpdate` optimistic update
- [x] in `useUpdateMany` optimistic update

Supersedes #10242

### DIFF
--- a/packages/ra-core/src/dataProvider/useUpdate.ts
+++ b/packages/ra-core/src/dataProvider/useUpdate.ts
@@ -132,7 +132,13 @@ export const useUpdate = <RecordType extends RaRecord = any, ErrorType = Error>(
             }
             return [
                 ...old.slice(0, index),
-                { ...old[index], ...data } as RecordType,
+                {
+                    ...old[index],
+                    // Stringify and parse the data to remove undefined values.
+                    // If we don't do this, an update with { id: undefined } as payload
+                    // would remove the id from the record, which no real data provider does.
+                    ...JSON.parse(JSON.stringify(data)),
+                } as RecordType,
                 ...old.slice(index + 1),
             ];
         };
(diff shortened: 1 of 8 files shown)


packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -132,7 +132,13 @@ export const useUpdate = <RecordType extends RaRecord = any, ErrorType = Error>(
...
             return [
                 ...old.slice(0, index),
-                { ...old[index], ...data } as RecordType,
+                {
+                    ...old[index],
+                    // Stringify and parse the data to remove undefined values.
+                    // If we don't do this, an update with { id: undefined } as payload
+                    // would remove the id from the record, which no real data provider does.
+                    ...JSON.parse(JSON.stringify(data)),
+                } as RecordType,
```
djhi, line 141: Shouldn't we have tests for this?

packages/ra-core/src/dataProvider/useUpdateMany.ts
```diff
@@ -124,7 +124,13 @@ export const useUpdateMany = <
...
                 newCollection = [
                     ...newCollection.slice(0, index),
-                    { ...newCollection[index], ...data },
+                    {
+                        ...newCollection[index],
+                        // Stringify and parse the data to remove undefined values.
+                        // If we don't do this, an update with { id: undefined } as payload
+                        // would remove the id from the record, which no real data provider does.
+                        ...JSON.parse(JSON.stringify(data)),
+                    },
```
djhi, line 133: Same

djhi (approved):

PR #10250, opened 2024-09-30
[Doc] Rewrite access control documentation

fzaninotto:
Follows #10222

### DIFF
--- a/docs/Permissions.md
+++ b/docs/Permissions.md
@@ -1,19 +1,172 @@
 ---
 layout: default
-title: "Permissions"
+title: "Authorization"
 ---
 
-# Permissions
+# Authorization
 
-By default, react-admin apps don't check user permissions. They only require users to be logged in for the list, create, edit, and show pages if there is an `authProvider`.
+Once a user is authenticated, your application may need to check if the user has the right to access a specific resource or perform a specific action. React-admin provides two ways to do so:
 
-However, some applications may require fine-grained permissions to enable or disable access to certain features. Since there are many possible strategies (single role, multiple roles or rights, ACLs, etc.), react-admin delegates the permission logic to the `authProvider`.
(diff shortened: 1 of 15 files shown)


docs/Permissions.md
erwanMarmelab, line 1: Could you please rename the file to update the doc url ?

docs/Permissions.md
```diff
@@ -1,44 +1,179 @@
...
-        { action: ["write"], resource: "tasks.completed" },
-    ])
+    login: async ({ username, password }) => {
+        // ...
+        // permissions look like 
+        // [
+        //   { action: 'read', resource: 'posts' },
+        //   { action: 'write', resource: 'posts' }
+        //]
+        const permissions = await fetchPermissions();
```
erwanMarmelab, line 52: The `...` comment makes the rest of the comments confusing
```suggestion
        // ...
        const permissions = await fetchPermissions();
        // permissions look like 
        // [
        //   { action: 'read', resource: 'posts' },
        //   { action: 'write', resource: 'posts' }
        //]
```

docs/Permissions.md
```diff
@@ -85,71 +220,34 @@ export default {
 ```
 {% endraw %}
 
-## Getting User Permissions
+### `usePermissions`
```
erwanMarmelab, line 223: **note (non-blocking):** ‏This reference may could confuse users in the Algolia search between this and the `userPermissions` page

erwanMarmelab (changes requested):

docs/Permissions.md
fzaninotto: You're right, but we may reintegrate the doc for this hook here and delete the standalone chapter, to avoid overwhelming uses with a long list of hooks. In other words, this is a work in progress.

docs/Permissions.md
fzaninotto: I don't want to do it to avoid having 404s on (internal and external) links to the Permissions page

PR #10255, opened 2024-10-02
Introduce `<NavigateToFirstResource>`

djhi:
## Problem

With the introduction of `canAccess`, we must now detect the first resource users have access to.

## Additional Checks

### DIFF
--- a/docs/Admin.md
+++ b/docs/Admin.md
@@ -488,6 +488,38 @@ const authProvider = {
 }
 ```
 
+**Tip**: If your authProvider implements [the `canAccess` method](./AuthProviderWriting.md#canaccess) and you don't provide a dashboard, React-Admin will use the first resource for which users have access to the list page as the home page for your admin. Make sure you order them to suit your needs.
+
+**Tip**: The detection of the first resource implies checking users are authenticated. Should your first resource be accessible without authentication or access right checks, you must provide a dashboard that redirects to it:
+
+```tsx
+// in src/Dashboard.js
+import * as React from "react";
+import { Navigate } from 'react-router';
+import { Title } from 'react-admin';
+
+export const Dashboard = () => (
+    <Navigate to="/unprotected" />
+);
+```
+
+```tsx
+// in src/App.js
+import * as React from "react";
+import { Admin, Resource } from 'react-admin';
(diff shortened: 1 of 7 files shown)


docs/Admin.md
```diff
@@ -488,6 +488,38 @@ const authProvider = {
 }
 ```
 
+**Tip**: If your authProvider implements [the `canAccess` method](./AuthProviderWriting.md#canaccess) and you don't provide a dashboard, React-Admin will use the first resource for which users have access to the list page as the home page for your admin. Make sure you order them to suit your needs.
+
+**Tip**: The detection of the first resource implies checking users are authenticated. Should your first resource be accessible without authentication or access right checks, you must provide a dashboard that redirects to it:
```
fzaninotto, line 493:
```suggestion
**Tip**: The detection of the first resource implies checking users are authenticated. Should your first resource be accessible without authentication, you must provide a `dashboard` component that redirects to it:
```

packages/ra-core/src/core/NavigateToFirstResource.tsx
fzaninotto, line 1: Must be exported

packages/ra-core/src/core/NavigateToFirstResource.tsx
```diff
@@ -0,0 +1,35 @@
...
+    if (resource) {
+        return (
+            <Navigate
+                to={createPath({
+                    resource,
+                    type: 'list',
+                })}
+            />
+        );
+    }
```
fzaninotto, line 30: I think we should throw an error if there is no accessible resource.

fzaninotto (changes requested):
Excellent!

packages/ra-core/src/core/NavigateToFirstResource.tsx
djhi: Should we redirect to `/authentication-error` or `/access-denied`?

fzaninotto: We don't know if the problem comes from a lack of resource or a lack of accessible resource, so I would just throw an error and let the error boundary to its job.

fzaninotto (approved):

PR #10256, opened 2024-10-03
Remove redundant optimisation of placeholder data query cache lookup

wattroll:
This is essentially a revert of commit e4e0f2c. See #7028 for why it was necessary. TanStack/query#6918 included equivalent optimisation ([14aa69f]), making it redundant here.

The dependency is bumped to be at least [@tanstack/react-query@5.21.7].

[14aa69f]: https://github.com/TanStack/query/commit/14aa69f
[@tanstack/react-query@5.21.7]: https://github.com/TanStack/query/releases/tag/v5.21.7


wattroll:
The typechecking errors appear to be caused by bumping react-query version and refer to code that haven't been changed here. I'll have a quick look at them, but should I rather:

a. Fix them in this PR (separate commit), or
b. Make another PR just fixing the type errors and have it merged first

I am averse to removing optimisation without bumping the dependency bound since that would be an unnecessarily breaking change in a very subtle performance-impacting way.

wattroll:
I also have confirmed that all 3 [errors when building against latest react-query version](https://github.com/marmelab/react-admin/actions/runs/11162215932/job/31026381980?pr=10256) appear in places where there is some degree of typechecker dodging. I assume that changes in react-query broke these coercions and they now need to be refined or resolved soundly. Specifically:

1. https://github.com/marmelab/react-admin/blob/0538e798434aa83e2d9da49bd7ac2d98e70cbc56/packages/ra-core/src/controller/input/useReferenceInputController.ts#L119
2. https://github.com/marmelab/react-admin/blob/0538e798434aa83e2d9da49bd7ac2d98e70cbc56/packages/ra-core/src/dataProvider/useGetList.ts#L176
3. https://github.com/marmelab/react-admin/blob/0538e798434aa83e2d9da49bd7ac2d98e70cbc56/packages/ra-core/src/dataProvider/useGetManyReference.ts#L159

wattroll:
[Rebased](https://github.com/marmelab/react-admin/pull/10256#event-14555935924) onto master in order to resolve the conflict in yarn.lock

fzaninotto:
Thanks!

Since you're bumping the required version for a dependency, this change should land in a minor version (e.g., 5.4) rather than a bug fix (e.g., 5.3.2). So please PR against the `next` branch instead of `master`.

wattroll:
It is now rebased onto the `next` branch. There were no conflicts.

slax57 (approved):
Thanks !

PR #10257, opened 2024-10-04
Make authProvider.getPermissions optional

djhi:
## Problem

Even though using permissions in React-Admin is optional, the AuthProvider type required users to implement the `getPermissions` method.

## Solution

Make the `getPermissions` method optional.
This also allows us to optimize the `usePermissions` hook.

## How To Test

Stories and tests

## Additional Checks

### DIFF
--- a/packages/ra-core/src/auth/usePermissions.spec.tsx
+++ b/packages/ra-core/src/auth/usePermissions.spec.tsx
@@ -1,47 +1,47 @@
 import * as React from 'react';
 import expect from 'expect';
 import { waitFor, render, screen } from '@testing-library/react';
-import { CoreAdminContext } from '../core/CoreAdminContext';
-
-import usePermissions from './usePermissions';
 import { QueryClient } from '@tanstack/react-query';
-
-const UsePermissions = ({ children }: any) => {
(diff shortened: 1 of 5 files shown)


packages/ra-core/src/auth/usePermissions.spec.tsx
```diff
@@ -53,14 +53,10 @@ describe('usePermissions', () => {
             getPermissions: () => Promise.resolve('admin'),
         };
         render(
-            <CoreAdminContext authProvider={authProvider}>
-                <UsePermissions>{stateInpector}</UsePermissions>
-            </CoreAdminContext>
+            <WithAuthProviderAndGetPermissions authProvider={authProvider} />
```
fzaninotto, line 56: this story has a 300ms delay. Make it configurable so the the test is faster than that.

fzaninotto (changes requested):

packages/ra-core/src/auth/usePermissions.spec.tsx
djhi: That's why I pass an authProvider ;)

fzaninotto (approved):

PR #10258, opened 2024-10-04
Access control: pessimistic rendering in CRUD views

djhi:
## Problem

All CRUD views currently starts to render optimistically even though authentication and optionally authorization checks are pending. This is a security issue.

## Solution

Make sure the base components for all CRUD views verify that no auth checks are pending before rendering.
Introduce a loading prop accepting a ReactNode on those components that is displayed while auth checks are pending.

## How To Test

## Additional Checks

### DIFF
--- a/packages/ra-core/src/controller/create/CreateBase.tsx
+++ b/packages/ra-core/src/controller/create/CreateBase.tsx
@@ -63,7 +63,8 @@ export const CreateBase = <
     }
 
     return (
-        <OptionalResourceContextProvider value={controllerProps.resource}>
+        // We pass props.resource here as we don't need to create a new ResourceContext if the props is not provided
+        <OptionalResourceContextProvider value={props.resource}>
             <CreateContextProvider value={controllerProps}>
                 {children}
             </CreateContextProvider>
--- a/packages/ra-core/src/controller/create/CreateBase.tsx
+++ b/packages/ra-core/src/controller/create/CreateBase.tsx
@@ -6,7 +6,8 @@ import {
 } from './useCreateController';
 import { CreateContextProvider } from './CreateContextProvider';
 import { Identifier, RaRecord } from '../../types';
-import { ResourceContextProvider } from '../../core';
+import { OptionalResourceContextProvider } from '../../core';
+import { useIsAuthPending } from '../../auth';
 
 /**
  * Call useCreateController and put the value in a CreateContext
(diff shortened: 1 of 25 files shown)


packages/ra-core/src/controller/create/CreateBase.tsx
```diff
@@ -40,28 +41,45 @@ import { ResourceContextProvider } from '../../core';
...
+        resource: controllerProps.resource,
+        action: 'create',
+    });
+
+    if (isAuthPending && !props.disableAuthentication) {
+        return loading;
+    }
+
+    return (
+        <OptionalResourceContextProvider value={controllerProps.resource}>
```
fzaninotto, line 66:
```suggestion
        <OptionalResourceContextProvider value={props.resource}>
```

Otherwise it's a BC. Same for other controllers.

packages/ra-ui-materialui/src/detail/Create.tsx
```diff
@@ -81,6 +82,7 @@ export const Create = <
             disableAuthentication={disableAuthentication}
             hasEdit={hasEdit}
             hasShow={hasShow}
+            loading={defaultLoading}
```
fzaninotto, line 85: `<Create>` should allow overriding the `loading` component. Same for the others.

fzaninotto (changes requested):
That's great! Now apps with a slow `checkAuth` won't reveal anything in CRUD pages. 

The UX is a bit different for the Dashboard, as users don't see the Loading page while auth is checked. Is there a way to make this view consistent with the others?

packages/ra-core/src/controller/create/CreateBase.tsx
djhi: No because the controller already handle it

packages/ra-ui-materialui/src/detail/Create.tsx
djhi: Stupid me

djhi:
> That's great! Now apps with a slow `checkAuth` won't reveal anything in CRUD pages.
> 
> The UX is a bit different for the Dashboard, as users don't see the Loading page while auth is checked. Is there a way to make this view consistent with the others?

Yes, by wrapping user provided dashboard in another component

fzaninotto (approved):

PR #10259, opened 2024-10-04
Support 'labelRowsPerPage' prop on Pagination

afilp:
I have a use case where I do not want to use the translated text but a conditionally different one.

## Problem

I have a calendar and want to restrict the number of events per month on each call, so I need pagination on that "List" (that just uses a calendar). So, I do not want the pagination label to be called **"Rows per page"** there but rather "Events Shown on Calendar".

I cannot do this as the code is now.

image

## Solution

Support the 'labelRowsPerPage' as a prop that we can use it.

## How To Test

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
+++ b/packages/ra-ui-materialui/src/list/pagination/Pagination.tsx
@@ -136,7 +136,10 @@ export const Pagination: FC<PaginationProps> = memo(props => {
                 disabled: !hasNextPage,
             }}
             component="span"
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks for you PR. 

There are more labels in this component than just the "rows per page". Providing a way to override one and not the others leads to a half-baked functionality. So I'm reluctant to merge your contribution as is. Also, it lacks documentation. 

Besides, what you ask is achievable in user land without too much hassle, so I wonder if we should move forward with this PR or not.

djhi:
I agree with @fzaninotto. Thanks for the PR anyway :)

PR #10263, opened 2024-10-07
[Doc] Update default port of remix and use latest yarnpkg.com link

smeng9:
Follow up of https://github.com/marmelab/react-admin/pull/9982

Due to switching to vite plugin,
  Remix seems uses a different default port now.
  The path of /admin without the trailing slash will result in 404 not found.
  The files in a newly created remix project gets slightly changed.


slax57 (approved):
Reviewed, tested and approved! Thanks!

PR #10270, opened 2024-10-10
Add support for embedding and prefetching

fzaninotto:
## Problem

Pages aggregating data from more than one resource always have to fetch more than one API endpoint in cascade, and with a slow backend this leads to an incomplete and blinking UI.

For example, the Comment list page in the simple example shows the post for each comment. So it has to fetch `dataProvider.getList('comments')`, then `dataProvider.getMany('posts', { ids: [comment postIds] })` (via `<ReferenceField>`).

image

Some backends are capable of embedding related records (e.g. the post of a comment). However, populating the cache based on these embeds is cumbersome.

## Solution

- Allow developers to embed related data in the main data
- Allow developers to prefetch related records and populate the cache with it

## TODO

- [x] Allow every data provider method to return a response `meta`
- [x] Update `useDataProvider` to populate query cache when a response contains a `meta: { prefetched }`
- [x] Add support for embedding and prefetching to `ra-data-fakerest`
- [x] Add support for embedding to `ra-data-json-server`
- ~~[ ] Add support for embedding and prefetching to `ra-data-simplerest`~~(out of scope)
- [x] Update the simple example to use `prefetch` in the comment list and show views.
- [x] add tests
- [x] Add doc

```jsx
const CommentList = () => (
    <ListBase queryOptions={{ meta: { prefetch: ['post'] } }}>
        <ListView />
    </ListBase>
);
```

image

### DIFF
--- a/docs/DataProviderWriting.md
+++ b/docs/DataProviderWriting.md
@@ -393,111 +393,6 @@ dataProvider.getList('posts', {
 
 React-admin's `<Pagination>` component will automatically handle the `pageInfo` object and display the appropriate pagination controls.
 
-## Embedded Data
-
-The response for `getList`, `getOne`, `getMany`, and `getManyReference` can include embedded data. This is useful when you want to avoid additional requests to fetch related records (e.g., when using [`<ReferenceField>`](./ReferenceField.md#prefetching)). The embedded data should be included in the `meta._embed` property of the response.
-
-```jsx
-const { data, meta } = dataProvider.getOne('posts', { id: 123 }, { meta: { _embed: ['author'] } });
-console.log(data); // { id: 123, title: "Hello, world", author_id: 456 }
-console.log(meta._embed); // { authors: [{ id: 456, name: "John Doe" }] }
-```
-
-By convention, the `meta._embed` response key must be an object where each key is the name of the embedded resource, and each value is an array of records.
-
-```jsx
-// example _embed value
-{
-  authors: [
-    { id: 1, name: 'John Doe' }
-  ],
-  comments: [
-    { id: 1, post_id: 1, body: 'Nice post!' },
-    { id: 2, post_id: 1, body: 'I agree!' }
-],
-}
-```
-
-It's the Data Provider's job to build the `meta._embed` object based on the API response.
-
-For example, the [JSON server](https://github.com/typicode/json-server?tab=readme-ov-file#embed) backend supports embedded data using the `_embed` query parameter:
-
-```txt
-GET /posts/123?_embed=author
-```
-
-```json
-{
-    "id": 123,
-    "title": "Hello, world",
-    "author_id": 456,
-    "author": {
-        "id": 456,
-        "name": "John Doe"
-    }
-}
-```
-
-To add support for embedded records in the response, the JSON Server Data Provider extracts the embedded data from the response, and puts them in the `meta._embed` property:
-
-```jsx
-const dataProvider = {
-    getOne: async (resource, params) => {
-        let query = `${apiUrl}/${resource}/${params.id}`;
-        if (params.embed) {
-            query += `?_embed=${params._embed.join(',')}`;
-        }
-        const { json: record } = await httpClient(query);
-        const embeds = {};
-        if (params._embed) {
-            params._embed.forEach(embed => {
-                if (record[embed]) {
-                    const embedKey = embed.endsWith('s') ? embed : `${embed}s`;
-                    if (!embeds[embedKey]) {
-                        embeds[embedKey] = [];
-                    }
-                    if (!embeds[embedKey].find(r => r.id === record[embed].id)) {
-                        embeds[embedKey].push(record[embed]);
-                    }
-                    delete record[embed];
-                }
-            });
-        }
-        return { data: record, meta: { _embed: embeds } };
-    },
-    getList: async (resource, params) => {
-        let query = `${apiUrl}/${resource}`;
-        if (params._embed) {
-            query += `?_embed=${params._embed}`;
-        }
-        const { json: records, headers } = await httpClient(query);
-        const embeds = {};
-        if (params._embed) {
-            records.forEach(record => {
-                params._embed.forEach(embed => {
-                    if (record[embed]) {
-                        const embedKey = embed.endsWith('s') ? embed : `${embed}s`;
-                        if (!embeds[embedKey]) {
-                            embeds[embedKey] = [];
-                        }
-                        if (!embeds[embedKey].find(r => r.id === record[embed].id)) {
-                            embeds[embedKey].push(record[embed]);
-                        }
-                        delete record[embed];
-                    }
-                });
-            });
-        }
-        return {
-            data: records,
-            total: parseInt(headers.get('content-range').split('/').pop(), 10),
-            meta: { _embed: embeds }
-        };
-    },
-    // ...
-}
-```
-
 ## Error Format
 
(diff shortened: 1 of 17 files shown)


djhi:
*Note*: [json-server](https://github.com/typicode/json-server?tab=readme-ov-file#embed) supports embeds

docs/ReferenceField.md
```diff
@@ -237,6 +237,30 @@ React-admin accumulates and deduplicates the ids of the referenced records to ma
...
+            <ReferenceField source="author_id" />
+        </Datagrid>
+    </List>
+);
+```
+{% endraw %}
+
+**Note**: For prefetching to function correctly, your data provider must support [Relationships Embedding](./DataProviders.md#embedding-relationships). Refer to your data provider's documentation to verify if this feature is supported.
+
+**Note**: Prefetching is a frontend performance feature, designed to avoid flickers and repaints. It doesn't always prevent `<ReferenceField>` to fetch the data. For instance, when coming to a show view from a list view, the main record is already in the cache, so the page renders immediately, and both the page controller and the `<ReferenceField>` controller fetch the data in parallel. The embedded data from the page controller arrives after the first render of the `<ReferenceField>`, so the data provider fetches the related data anyway. But from a user perspective, the page displays immediately, including the `<ReferenceField>`.
```
djhi, line 262: Might be worth mentioning that playing with the queryClient staleTime option can improve this

docs/DataProviderWriting.md
```diff
@@ -548,7 +672,15 @@ const { data } = dataProvider.getOne('posts', { id: 123 })
 // }
 ```
 
-This will cause the Edit view to blink on load. If you have this problem, modify your Data Provider to return the same shape for all methods. 
+This will cause the Edit view to blink on load. If you have this problem, modify your Data Provider to return the same shape for all methods.
+
+**Note**: If the `getList` and `getOne` methods user different `meta` parameters, they won't share the cache. You can use this as an escape hatch to avoid flickering in the Edit view.
```
djhi, line 677:
```suggestion
**Note**: If the `getList` and `getOne` methods use different `meta` parameters, they won't share the cache. You can use this as an escape hatch to avoid flickering in the Edit view.
```

packages/ra-core/src/dataProvider/populateQueryCache.ts
```diff
@@ -0,0 +1,48 @@
...
+ * // setQueryData(['comments', 'getOne', { id: '1' }], { id: 1, post_id: 1, body: 'Nice post!' });
+ * // setQueryData(['comments', 'getMany', { ids: ['1'] }], [{ id: 1, post_id: 1, body: 'Nice post!' });
+ */
+export const populateQueryCache = ({
+    data,
+    queryClient,
+    staleTime = 500, // ms
+}: PopulateQueryCacheOptions) => {
+    // setQueryData doesn't accept a stale time option
+    // So we set an updatedAt in the future to make sure the data is considered stale
```
djhi, line 30:
```suggestion
    // So we set an updatedAt in the future to make sure the data isn't considered stale
```

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
```diff
@@ -335,4 +337,60 @@ describe('useDataProvider', () => {
...
+                    </ReferenceFieldBase>
+                </RecordContextProvider>
+            );
+        };
+
+        render(
+            <CoreAdminContext dataProvider={dataProvider}>
+                <FetchPost />
+            </CoreAdminContext>
+        );
```
djhi, line 390: This should be a story

djhi (changes requested):

packages/ra-data-json-server/README.md
```diff
@@ -100,6 +100,38 @@ const httpClient = (url, options = {}) => {
...
+`ra-data-json-server` supports [Embedded Relationships](https://marmelab.com/react-admin/DataProviders.html#embedding-relationships). Use the `meta.embed` query parameter to specify the relationships that you want to embed. 
+
+```jsx
+dataProvider.getOne('posts', { id: 1, meta: { embed: 'author' } });
+// { 
+//    data: { id: 1, title: 'FooBar', author: { id: 1, name: 'John Doe' } },
+// }
+```
+
+The name of the embedded resource must be singular for a many-to-one relationship, and plural for a one-to-many relationship.
```
slax57, line 114:
```suggestion
The name of the embedded resource must be singular for a one-to-one relationship, and plural for a one-to-many relationship.
```

?

packages/ra-core/src/dataProvider/populateQueryCache.ts
```diff
@@ -0,0 +1,48 @@
...
+ * // setQueryData(['posts', 'getOne', { id: '2' }], { id: 2, title: 'FooBar' });
+ * // setQueryData(['posts', 'getMany', { ids: ['1', '2'] }], [{ id: 1, title: 'Hello, world' }, { id: 2, title: 'FooBar' }]);
+ * // setQueryData(['comments', 'getOne', { id: '1' }], { id: 1, post_id: 1, body: 'Nice post!' });
+ * // setQueryData(['comments', 'getMany', { ids: ['1'] }], [{ id: 1, post_id: 1, body: 'Nice post!' });
+ */
+export const populateQueryCache = ({
+    data,
+    queryClient,
+    staleTime = 500, // ms
+}: PopulateQueryCacheOptions) => {
```
slax57, line 28: `staleTime = 500` might be a bit too low as a default.

On my laptop (8th get intel core i7), it takes more than 500ms to process the `CommentList` data and render the Reference fields, so the call to `getMany posts {"ids":[6,9,3,1]}` is still fired (in the demo).
If I increase the `staleTime` to 1000 ms for instance, then the prefetch works and the call to `getMany` is not fired.

More generally, this setting is a bit deep in the component tree, so it might be cumbersome to customize. I'm wondering if it would be that hard to add an extra parameter to the returned `meta` allowing to customize this `staleTime` on a per-query basis. Wdyt?

slax57 (changes requested):
Incredible feature! And the actual implementation is actually not as hard as I would have imagined.
Love it!

packages/ra-data-fakerest/README.md
```diff
@@ -129,6 +129,70 @@ This data provider uses [FakeRest](https://github.com/marmelab/FakeRest) under t
...
+`ra-data-fakerest` supports [Embedded Relationships](https://marmelab.com/react-admin/DataProviders.html#embedding-relationships). Use the `meta.embed` query parameter to specify the relationships that you want to embed. 
+
+```jsx
+dataProvider.getOne('posts', { id: 1, meta: { embed: ['author'] } });
+// { 
+//    data: { id: 1, title: 'FooBar', author: { id: 1, name: 'John Doe' } },
+// }
+```
+
+You can embed more than one related record, so the `embed` value must be an array. The name of the embedded resource must be singular for a many-to-one relationship, and plural for a one-to-many relationship.
```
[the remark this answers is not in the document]
fzaninotto: I disagree with this proposed change. book=>author is a many-to-one relationship and is concerned by that doc. A one-to-one relationship is a particular case of a many-to-one relationship.

packages/ra-core/src/dataProvider/populateQueryCache.ts
fzaninotto: Changed to 1000 stale time by default. Making it configurable is a nice to have, and should be done after making the *main* query stale time configurable (e.g. getList).

djhi: Isn't it already configurable through `queryOptions`?

fzaninotto: The main staleTime is configurable through QueryOptions, so we may use it. But is it correct to use it for  prefetched data? Not sure. 

The idea was to allow the server to set the cache preferences, and so to use a response meta to update the cache `updated_at`. To be done sometime in the future.

djhi (approved):

slax57 (approved):

PR #10271, opened 2024-10-11
[Doc] Document react-hook-form limitation preventing using dynamically added inputs in `<ArrayInput>` with `shouldUnregister`

### LINKED ISSUE #10194: Values of InputFields are not displayed in edit form if shouldUnregister is set
**What you were expecting:**
In an EditForm with an ArrayInput all input fields show the given record values. This also holds for conditionally rendered field values if the condition is fullfilled and the `shouldUnregister` property is set to `true`.

**What happened instead:**
Having the following complex EditForm.
There is an ArrayInput. 
One field of the array elements is only conditionally rendered depending on the value of another. Then the value of the conditionally rendered input field is not displayed (even though the field itself is rendered). This only happens if the `shouldUnregister` property is set to true for the conditionally rendered input field.

We just upgraded from react-admin@4.16.20 (where everything worked) to react-admin@5.1.3 where the bug occurs. In both cases we use react-hook-form@7.53.0 

**Steps to reproduce:**
Open [this code sandbox](https://stackblitz.com/edit/github-ijv3mw?file=src%2Fcomments%2FCommentEdit.tsx). 
Go to the comments records and open one of the comments. The value in the field "Bar" is not shown even though we would expect to see "world" there. If the "shouldUnregister" property is removed in the `CommentEdit.tsx` (line 166) the value is displayed correctly. However, then we would lose the unregister functionality of the field.

Hint: Starting from the provided example code sandbox for v5 we only changed the `data.tsx` and `CommendEdit.tsx` to provide the example.

**Environment**

* React-admin version: 5.1.3
* Last version that did not exhibit the issue (if applicable): 4.16.20
* React version: 18.3.1
* Browser: Chrome, Firefox

djhi:
## Problem

Fixes #10194

## Solution

~~We need to provide both `values` and `defaultValues` to `useForm`, because~~
~~- we need `values` to be able to reactively reset the form when the record changes~~
~~- we need `defaultValues` because `useController` requires them to function properly~~
~~(omitting them can cause issues with reset or with shouldUnregister)~~
~~See https://react-hook-form.com/docs/usecontroller#props~~

UPDATE: Since this appears to be a react-hook-form issue/limitation, we should not attempt to fix this issue.
Instead we will update the documentation, saying the correct way to do this is using the `transform` prop.

## How To Test

~~Open the http://localhost:9010/?path=/story/ra-ui-materialui-input-arrayinput--with-should-unregister story~~
N/A

## Additional Checks

- [x] The PR includes **unit tests** (if not possible, describe why) -> won't do, because officially the docs recommend against this use case


djhi (commented):
:+1:

slax57 (approved):

slax57:
**EDIT**: On 2nd thought, we shouldn't fix this issue in react-admin, as it appears to be a react-hook-form issue/limitation.
We will however keep the documentation changes, saying the correct way to do this is using the `transform` prop.

packages/ra-ui-materialui/src/input/ArrayInput/ArrayInput.stories.tsx
```diff
@@ -1,24 +1,24 @@
...
+import { AutocompleteInput } from '../AutocompleteInput';
 import { DateInput } from '../DateInput';
 import { NumberInput } from '../NumberInput';
-import { AutocompleteInput } from '../AutocompleteInput';
+import { TextInput } from '../TextInput';
 import { TranslatableInputs } from '../TranslatableInputs';
-import { ReferenceField, TextField, TranslatableFields } from '../../field';
-import { Labeled } from '../../Labeled';
+import { ArrayInput } from './ArrayInput';
+import { SimpleFormIterator } from './SimpleFormIterator';
```
djhi, line 21: Can you really revert?

slax57: done

slax57 (approved):

Issue #10252, opened 2024-10-02
Stackblitz v4 issue template does not launch sample application 


KilianB:
**What you were expecting:**

Opening the v4 issue template I would expect a sample application to start up

**What happened instead:**

A blank white screen.

![image](https://github.com/user-attachments/assets/43e299ee-4b55-4f86-9fca-c9c56688ab56)

**Steps to reproduce:**

Navigate to https://stackblitz.com/github/marmelab/react-admin/tree/4.x/examples/simple

**Other information:**

Stopping the dev server in the terminal, calling `npm run build` and `npm run preview` works fine

**Environment**

* Browser: Chrome

fzaninotto:
Are you sure it's not just slow?

fzaninotto:
Edit: no, it's not slow, it's broken. Marking this as a bug.

fzaninotto:
I have opened an issue on Stackblitz: https://github.com/stackblitz/webcontainer-core/issues/1515

fzaninotto:
I pushed a lock file, and the v4 stackblitz seems to launch OK (even though it's suuuper slow to start, at least 30s on my laptop). Can you confirm it's fixed for you too?

KilianB:
It is fixed but not really useable at work.

I opened it on a very old MacBook Air 2015 there it took 5+ minutes to show the screen.
On my work laptop chrome does not allow Stackblitz due to 3rd party cookies and Firefox errored out due to a DataCloneError which is a Stackblitz internal issue I believe. 

I guess it would somehow work when I use my beefy work station back at home, I'll report back later today.

fzaninotto:
Second issue opened on StackBlitz: https://github.com/stackblitz/webcontainer-core/issues/1530

Please add a +1 there if you suffer from slow stackblitz, too!

PR #10273, opened 2024-10-11
Add CI action to sync published versions simple example to sandbox repository

djhi:
## Problem

As explained in #10252, Stackblitz is currently very slow. Besides, we can't use codesandbox to open a subpath of our monorepo anymore.

## Solution

We now have a dedicated repository for the simple example: https://github.com/marmelab/react-admin-sandbox
This PR adds a CI action that sync the monorepo simple example with the sandbox repository.
It does so only when a new tag is released, except for `next`, `beta` and `alpha` tags

## Additional Checks


.github/workflows/test.yml
```diff
@@ -157,7 +162,7 @@ jobs:
     runs-on: ubuntu-latest
     name: GreenFrame
     needs: [e-commerce]
-    if: success() && github.ref == 'refs/heads/master'
+    if: success() && github.ref == 'refs/heads/master' && github.ref_type != 'tag'
```
slax57, line 165: **thought**: on the contrary, we could run the Greenframe analysis only when a new tag/version is released, as opposed to any commit on `master` which might be a little too often. Wdyt?

.github/workflows/test.yml
```diff
@@ -208,3 +213,21 @@ jobs:
...
+    # Only run on new tags that target a release (not latest nor next) and avoid alpha and beta tags
+    if: github.event_name == 'push' && github.ref_type == 'tag' && github.ref_name == 'refs/tags/v*' && !contains('beta', github.ref_name) && !contains('alpha', github.ref_name)
+    needs: [typecheck, simple-example-typecheck, unit-test, e2e-test]
+    steps:
+        - name: Checkout
+          uses: actions/checkout@v3
+        - name: Use Node.js LTS
+          uses: actions/setup-node@v3
+          with:
+              node-version: '16.x'
```
slax57, line 228:
```suggestion
              node-version: '18.x'
```

To be consistent with other jobs of the workflow

Makefile
```diff
@@ -172,3 +172,6 @@ storybook: ## Launch the storybook
 build-storybook: ## Build the storybook
 	@echo "Building storybook..."
 	@yarn build-storybook
+
+update-sandbox: ## Push the local version of the simple example to the sandbox repository
+	./update-demo.sh
```
slax57, line 177:
```suggestion
	./update-sandbox.sh
```

(there is no `update-demo.sh`)

.github/workflows/test.yml
```diff
@@ -208,3 +213,21 @@ jobs:
...
+        - name: Update Sandbox Repository
+          env:
+              SSH_DEPLOY_KEY: ${{ secrets.SSH_SANDBOX_DEPLOY_KEY }}
```
slax57, line 231:
```suggestion
              SSH_SANDBOX_DEPLOY_KEY: ${{ secrets.SSH_SANDBOX_DEPLOY_KEY }}
```

To be consistent with the actual script

slax57 (changes requested):

scripts/update-sandbox.sh
```diff
@@ -0,0 +1,34 @@
...
+(rm -rf $TEMPD/src)
+(rm -rf $TEMPD/assets)
+
+# Copy the demo files into the temporary directory
+(cp -r ./examples/simple/. $TEMPD)
+
+# Update the demo repository
+# Allow empty commits because even though we may have published new version of the enterprise packages,
+# the demo code may not have changed. However we still want it to be redeployed and, as we don't keep its
+# yarn.lock, it should use the new packages
```
slax57, line 33: I understand the rationale, but I believe @fzaninotto was in favor of actually computing and committing the `yarn.lock` file, in order to reduce the sandbox startup time

.github/workflows/test.yml
djhi: @fzaninotto any opinion?

fzaninotto: I agree that we should only test greenframe when releasing a new version

.github/workflows/test.yml
```diff
@@ -208,3 +213,21 @@ jobs:
...
+          uses: actions/checkout@v3
+        - name: Use Node.js LTS
+          uses: actions/setup-node@v3
+          with:
+              node-version: '18.x'
+        - name: Update Sandbox Repository
+          env:
+              SSH_SANDBOX_DEPLOY_KEY: ${{ secrets.SSH_SANDBOX_DEPLOY_KEY }}
+              SANDBOX_REPOSITORY: ${{ secrets.SANDBOX_REPOSITORY }}
+          run: make update-sandbox
```
erwanMarmelab, line 233: **nitpick (non-blocking):** ‏Not sure that we will use this Makefile command manually. We may could directly run the script here and remove the new Makefile command.

scripts/update-sandbox.sh
```diff
@@ -0,0 +1,33 @@
...
+(rm -rf $TEMPD/src)
+(rm -rf $TEMPD/assets)
+
+# Copy the demo files into the temporary directory
+(cp -r ./examples/simple/. $TEMPD)
+# Install dependencies to that the sandbox has a lock file
+yarn install --cwd $TEMPD
+
+# Update the demo repository
+(cd $TEMPD && git add -A && git commit --allow-empty -m "Update sandbox" && git push)
```
erwanMarmelab, line 33: **question:** ‏Why concatenate all those commands while the rest of the script has one command per line?

erwanMarmelab:
```suggestion
(cd $TEMPD)
(git add -A)
(git commit --allow-empty -m "Update sandbox")
(git push)
```

.github/workflows/test.yml
djhi: Done

scripts/update-sandbox.sh
djhi: No reason and not important

.github/workflows/test.yml
djhi: Useful for manual testing

erwanMarmelab (approved):

slax57 (approved):

PR #10278, opened 2024-10-15
Fix `useCanAccess` causes extra rerender + use `useCanAccess` in the simple demo

slax57:
## Problems

1. The React Admin [simple blog demo](https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple) uses `usePermissions`. But React Admin v5.3 will bring a better replacement feature called `useCanAccess`, which we should demonstrate in this demo.
2. `useCanAccess` causes an unnecessary extra rerender to components using it (e.g. `EditButton` or `ShowButton`) when there is no `authProvider` or no `authProvider.canAccess` implementation.

## Solution

1. Replace `usePermissions` by `useCanAccess` / `<CanAccess>` in the simple demo.
2. Enable the react-query query in `useCanAccess`  only when there is an actual `authProvider.canAccess` implementation (otherwise it creates an extra rerender because of the query status update).

## How To Test

1. Run the simple demo (`make run`).
Try signing in as:
- login
- user
- admin

2. Run the `useCanAccess` unit tests

## Todo

- [x] Disable the `useCanAccess` query when there is no `authProvider.canAccess` as it causes unnecessary rerenders!
- [x] In the simple example, change the way `authProvider.canAccess` is implemented to return `true` by default, and call a fn stored in memory that will actually implement the `canAccess` for a given role

## Additional Checks

### Current use of the permissions in the demo

role
- login
  - none
- user
  - `<Resource name="users" {...users} />`
- admin
  - `<Resource name="users" {...users} />`
  - PostCreate: `<ArrayInput source="authors">`
  - PostEdit: `<ArrayInput source="authors">`
  - UserCreate: `<SaveButton label="user.action.save_and_add" />`
  - UserCreate: `<TabbedForm.Tab label="user.form.security" path="security">`
  - UserEdit: `<TextInput source="id" InputProps={{ disabled: true }} />`
  - UserEdit: `<TabbedForm.Tab label="user.form.security" path="security">`
  - UserList: `filters.push(<TextInput source="role" />)`
  - UserList: `rowClick = 'edit'`
  - UserList: `<SimpleList secondaryText={record.role}>`
  - UserList: Datagrid: `<TextField source="role" />`
  - UserShow: `<TabbedShowLayout.Tab label="user.form.security">`

### Issues found while testing

- UserCreate: `<TabbedForm.Tab label="user.form.security" path="security">`
  - error: `the value prop is required at runtime`
  - TabbedForm does not support `<CanAccess>`, because it still clones children
  - => No choice but to use `useCanAccess` instead
- UserCreate: Toolbar is misplaced
  - Already exists on `master`


examples/simple/src/users/UserCreate.tsx
```diff
@@ -70,7 +76,7 @@ const UserCreate = () => {
                         validate={[required(), isValidName, unique()]}
                     />
                 </TabbedForm.Tab>
-                {permissions === 'admin' && (
+                {canManageUsers ? (
```
slax57, line 79: Could not use `<CanAccess>` here because `<TabbedForm>` still scans its children, so having `<CanAccess>` as one of its child couldn't work.

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -153,7 +152,7 @@ const PostCreate = () => {
                         <TextInput source="url" defaultValue="" />
                     </SimpleFormIterator>
                 </ArrayInput>
-                {permissions === 'admin' && (
+                <CanAccess action="manage_post_authors">
```
fzaninotto, line 155: This action name is actually an action and a resource... Why not use `action="edit" resource="users"`?

slax57: In my first implementation I did this, but then I realized that for such simple access rules we don't need that much complexity, and it allows me to omit the `resource` prop completely, making the code easier to read.

slax57:
Seems I need to fix the E2E tests. Back to WIP.

examples/simple/src/posts/PostCreate.tsx
slax57: I ended up implementing a mixed approach: keeping a 'custom' `action` for things like 'batch_create', and using `resource` for fields like 'posts.authors' and 'users.role'.

slax57:
All green, back to RFR!

djhi (approved):

djhi (approved):

Issue #9849, opened 2024-05-15
Consider replacing "hotscript" types with "react-hook-form" types


endoval:
Hello there,
we stumbled upon the fact that your typing of **source** property on `*Field` components uses `hotscript` types like so: `Call<Objects.AllPaths, Resource>` (added here: https://github.com/marmelab/react-admin/pull/8863)

We were wondering why that is the case because you are afaik using `react-hook-form` underneath, which provides a nice type to check paths as well with `FieldPath<Resource>`. Sadly, those two types require a different syntax for "array paths", which both functionally work. Consider the following example to illustrate this:

```
import { FC } from 'react';

import { Call, Objects } from 'hotscript';
import { TextField } from 'react-admin';
import { FieldPath } from 'react-hook-form';

interface MyRecord {
  tags: string[];
}

const MyComponent: FC = () => {
  const source1: FieldPath<MyRecord> = 'tags.0';
  const source2: Call<Objects.AllPaths, MyRecord> = 'tags[0]';

  return (<>
    <TextField label="Tag (source1)" source={source1} />
    <TextField label="Tag (source2)" source={source2} />
  </>);
}
```

Both versions work just fine (because the used `lodash/get` understands both), but of course the types are not matching. This might not seem like a problem right now, but once you start typing the `*Input` components all the way down to react-hook-form's `useController()` (within your `useInput()`) the types will clash. This already happens in our code base with some custom-written Inputs that use `react-hook-form` directly.

What we are wondering: Do you intend to stick to the `hotscript` types or consider switching to `react-hook-form` types?

Thank you! :)

slax57:
That's a very good question, thank you for raising this up!
Indeed this seems to indicate we will run into issues when extending this feature to inputs.
Your suggestion seems valid to me. 
@djhi @fzaninotto wdyt?

djhi:
That would indeed be better and avoid an extra dependency! Thanks for the suggestion.

endoval:
Great, thanks for the positive response. Looking forward to this change! 😌🙏

anmol-fzr:
Hi 👋, I would like to work on this issue. Can you assign this issue to me.

slax57:
@anmol-fzr No need to assign the issue, you can open a PR mentioning this issue right away and we will review it carefully. Please remember to target the `next` branch. Thanks for offering to work on this :slightly_smiling_face:

fzaninotto:
[...]

PR #10279, opened 2024-10-15
Make records paths inferred from type compatible with react-hook-form

djhi:
## Problem

Fixes https://github.com/marmelab/react-admin/issues/9849

Records paths inferred from type using `hotscript` aren't compatible with those inferred by `react-hook-form`

## Solution

Declare our own types for path inference that currently just re-export those from `react-hook-form`, making us more resilient to future changes.

## How To Test

- Open the `InvoiceShow` of the CRM demo and add a type parameter to the `TextField` of the Order reference. The `TextField` `source` should be valid and trying to replace it should trigger the IDE autocompletion, showing all `Order` fields

## Additional Checks

### DIFF
--- a/packages/ra-core/src/types.ts
+++ b/packages/ra-core/src/types.ts
@@ -1,4 +1,5 @@
 import { ComponentType, ReactElement, ReactNode } from 'react';
+import { FieldPath } from 'react-hook-form';
 import { WithPermissionsChildrenParams } from './auth/WithPermissions';
 import { AuthActionType } from './auth/types';
 
@@ -13,9 +14,11 @@ export interface RaRecord<IdentifierType extends Identifier = Identifier>
     id: IdentifierType;
 }
 
+export type SortOrder = 'ASC' | 'DESC';
+
 export interface SortPayload {
     field: string;
-    order: 'ASC' | 'DESC';
+    order: SortOrder;
 }
 export interface FilterPayload {
     [k: string]: any;
@@ -402,3 +405,20 @@ export type FormFunctions = {
 export type HintedString<KnownValues extends string> =
     | (string & {})
     | KnownValues;
+
+// Re-export react-hook-form implementation of FieldPath that returns all possible paths of an object
+// This will allow us to either include the FieldPath implementation from react-hook-form or replace it with our own
+// should we move away from react-hook-form
+// type Post = { title: string; author: { name: string; }; tags: { id: string; name: string} };
+// => Valid paths are "title" | "author" | "author.name" | "tags.id" | "tags.name"
+export type RecordValues = Record<string, any>;
+export type RecordPath<TRecordValues extends RecordValues> =
+    FieldPath<TRecordValues>;
+
+// Returns the union of all possible paths of a type if it is provided, otherwise returns a string
+// Useful for props such as "source" in react-admin components
+export type ExtractRecordPaths<T extends RecordValues> =
+    // Trick that allows to check whether T was provided
+    [T] extends [never] ? string : RecordPath<T>;
+
+export type AnyString = string & {};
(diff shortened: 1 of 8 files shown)


packages/ra-ui-materialui/src/field/types.ts
```diff
@@ -25,7 +23,7 @@ export interface FieldProps<
      *     </List>
      * );
      */
-    sortBy?: Call<Objects.AllPaths, RecordType> | AnyString;
+    sortBy?: HintedString<ExtractRecordPaths<RecordType>>;
```
fzaninotto, line 26: This is a regression: we could order by anystring

packages/ra-core/src/types.ts
```diff
@@ -13,9 +14,11 @@ export interface RaRecord<IdentifierType extends Identifier = Identifier>
     id: IdentifierType;
 }
 
+export type SortOrder = 'ASC' | 'DESC';
```
fzaninotto, line 17: This type is already defined in ra-ui-materialui (packages/ra-ui-materialui/src/field/types.ts), I'm afraid of a conflic. As this change seems to be unrelated to the subject of the PR, could you revert it?

packages/ra-core/src/types.ts
```diff
@@ -402,3 +405,20 @@ export type FormFunctions = {
...
+// => Valid paths are "title" | "author" | "author.name" | "tags.id" | "tags.name"
+export type RecordValues = Record<string, any>;
+export type RecordPath<TRecordValues extends RecordValues> =
+    FieldPath<TRecordValues>;
+
+// Returns the union of all possible paths of a type if it is provided, otherwise returns a string
+// Useful for props such as "source" in react-admin components
+export type ExtractRecordPaths<T extends RecordValues> =
+    // Trick that allows to check whether T was provided
+    [T] extends [never] ? string : RecordPath<T>;
```
fzaninotto, line 422: we used `AnyString` as the default value, now it's `string`. Isn't it a BC?

packages/ra-ui-materialui/src/field/FileField.tsx
```diff
@@ -114,16 +118,13 @@ export interface FileFieldProps<
     RecordType extends Record<string, any> = Record<string, any>,
 > extends FieldProps<RecordType> {
     src?: string;
-    title?: Call<Objects.AllPaths, RecordType> extends never
-        ? AnyString
-        : Call<Objects.AllPaths, RecordType> | AnyString;
+    title?: HintedString<ExtractRecordPaths<RecordType>>;
```
fzaninotto, line 121: This is a regression: we used to accept AnyString. Same for Imagefield.

fzaninotto (changes requested):

packages/ra-core/src/types.ts
djhi: No, `AnyString` is no longer necessary

packages/ra-ui-materialui/src/field/FileField.tsx
djhi: No `HintedString` does the same

packages/ra-ui-materialui/src/field/types.ts
djhi: No `HintedString` does the same

packages/ra-core/src/types.ts
djhi: More precisely, `AnyString` wasn't needed before. It was unnecessary already

fzaninotto (approved):

fzaninotto:
typecheck fails

fzaninotto (approved):

PR #10283, opened 2024-10-17
Update AutocompleteInput.tsx

Ferlorin:
## Problem

Breaks building.  No need since props is of type  React.HTMLAttributes<HTMLLIElement> & { key: any }

## Solution

Removal of the line

## How To Test

Build the project

## Additional Checks

- [bugfix] The PR targets `master` for a bugfix, or `next` for a feature

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.tsx
@@ -666,7 +666,6 @@ If you provided a React element for the optionText prop, you must also provide t
                 onInputChange={handleInputChange}
                 renderOption={(props, record: RaRecord) => {
                     // We have to extract the key because react 19 does not allow to spread the key prop
-                    // @ts-expect-error The key is indeed inside props but MUI does not provide the correct type
                     const { key: ignoredKey, ...rest } = props;
                     // We don't use MUI key which is generated from the option label because we may have options with the same label but with different values
                     const key = getChoiceValue(record);


slax57 (changes requested):
Thank you for the PR.

I'm sorry to say our CI seems to disagree with you.

There may be an issue in your build environment instead of the code itself.

Can you confirm you are using the `Makefile` commands for building? (like `make build`?)

Ferlorin:
@slax57 
When I say fail to build. i was getting errors when building my project with npm run build. However I re-installed everything from the beginning and a new .eslintrc file was created and it worked this time. However these issues are going red in the IDE when using typescript.
Its like there is a problem with the CI since IDEs are showing errors even if they just ignore them, however at some point those errors need to be minimized
image

slax57:
You need to use `yarn` to build the react-admin packages, otherwise you are completely bypassing the `yarn.lock` file, which is probably the root issue why your IDE is complaining.
Feel free to reopen if that does not fix the issue.

Ferlorin:
I use npm as a package manager to retrieve and build my project, I did not use directly the source. As you can see in the screenshot, the node_modules folder includes the packages that have been downloaded and contain thousands of errors.

slax57:
Thank you for these clarifications.
Still, it doesn't change to the fact that you are opening PRs to fix issues that we have no clear proof comes from the react-admin code. To me it's even the opposite: our CI checks, along with the fact that our packages are used by thousands of teams with no issue, tend to indicate that the issue very likely lies in your project's configuration instead.

At least, these issues deserve further analysis to better understand their root cause, before opening a PR.
If you are convinced there really is an issue in the react-admin code that needs to be fixed, then please open a new issue first, **including a reproduction**, so that we can discuss it.
Thanks.

PR #10288, opened 2024-10-21
Fix `<AutocompleteInput>` displays 'Create' option for choices that already exist when `createLabel` is provided

slax57:
## Problem

AutocompleteInput displays a 'Create' option for choices that already exist when the `createLabel` prop is provided.

image

Can be reproduced in the Storybook: https://react-admin-storybook.vercel.app/?path=/story/ra-ui-materialui-input-autocompleteinput--create-label

## Solution

Fix the condition under which the 'Create' option appears.

## How To Test

Run the story `/story/ra-ui-materialui-input-autocompleteinput--create-label` locally, or run the unit tests.

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
+++ b/packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
@@ -1168,7 +1168,7 @@ describe('<AutocompleteInput />', () => {
     });
 
     describe('onCreate', () => {
-        it("shouldn't include an option with the createLabel when the input is empty", async () => {
+        it("shouldn't include an option with the create label when the input is empty", async () => {
             const choices = [
                 { id: 'ang', name: 'Angular' },
                 { id: 'rea', name: 'React' },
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/AutocompleteInput.spec.tsx
```diff
@@ -1259,6 +1259,67 @@ describe('<AutocompleteInput />', () => {
...
+                        >
+                            <AutocompleteInput
+                                source="language"
+                                choices={choices}
+                                onCreate={handleCreate}
+                                createLabel="Start typing to create a new item"
+                            />
+                        </SimpleForm>
+                    </ResourceContextProvider>
+                </AdminContext>
```
djhi, line 1292: Why don't you use the story?

slax57: I copied the test just above which doesn't use it either, saving some time as it does almost exactly what I need in my own test.

slax57: Done anyway, for both tests

djhi (approved):

PR #10295, opened 2024-10-23
Fix smart count for no results message

thibault-barrat:
## Problem

#10291 introduces translation of the resource name in the no results message. Nevertheless, when there is no results, the smart count used for the translation should be 0.

## Solution

Update smart_count when calling `getResourceLabel` function.

### DIFF
--- a/packages/ra-core/src/core/useGetResourceLabel.ts
+++ b/packages/ra-core/src/core/useGetResourceLabel.ts
@@ -1,6 +1,6 @@
 import { useResourceDefinitions } from './useResourceDefinitions';
 import { useTranslate } from '../i18n';
-import { humanize, pluralize, singularize } from 'inflection';
+import { humanize, inflect } from 'inflection';
 
 /**
  * A hook which returns function to get a translated resource name. It will use the label option of the `Resource` component if it was provided.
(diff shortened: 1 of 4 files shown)


slax57 (approved):
Indeed, thanks!

slax57 (changes requested):
Tests appear to be failing though. Can you have a look?

slax57 (approved):
Nice catch, thanks!

packages/ra-core/src/core/useGetResourceLabel.ts
```diff
@@ -39,11 +39,7 @@ export const useGetResourceLabel = (): GetResourceLabel => {
                           smart_count: count,
                           _: resourceDefinition.options.label,
                       })
-                    : humanize(
-                          count > 1
-                              ? pluralize(resource)
-                              : singularize(resource)
-                      ),
+                    : humanize(inflect(resource, count)),
```
slax57, line 42: Actually `inflection` is a tool to manage strings for the English language, so this fix works well for English, but it may break other languages (like French).
I'll need to make sure it's OK to change this with the team. Let's put this PR on hold for now. Thank you anyways!

slax57: Consensus: applying English rules should be the default, using French gramary rules was a bug.
Users that use another language than English should provide translation keys for their resources.
Besides, this will only change the default value for resources that do not have a translation key, and when the smart count is 0, hence making this a pretty limited change.
TLDR: we're OK to merge!

slax57 (approved):

PR #10296, opened 2024-10-24
Fix undo logic not working when doing multiple deletions one by one

### LINKED ISSUE #10269: UNDO logic not working when doing multiple deletions one by one
**What you were expecting:**
The "UNDO" (cancel action) functionality is not working correctly. The issue caused by the so called undoableEventEmitter. This event emitter emits only the "end" event without any additional info about the action that will be performed. This causes an issue when I do multiple fast deletions. The issue is that the **second deletion triggers the emittion of "end" event which cases the first deletion action to get triggered even though the "UNDO" toast is still present on the UI and users can click the button but nothing will happen.** One possible solution for resolving the issue would be adding a unique identifier for the given action and passing this unique id to the undoable event emitter so undoable actions can be indipendently recognised.

**What happened instead:**
Each click of "UNDO" should be canceling the action that this "UNDO" is meant for.

**Steps to reproduce:**
Increase the "UNDO" default timeout so it doesn't hide the notification after 5 seconds so it's easier to reproduce. Open dev tools - network tab. Click two times the delete button inside of a list and click the first "UNDO" button. This will trigger the fist deletion instead of canceling it. From what I remember while testing I even think that it triggers both deletions and makes both "UNDO" toasts useless. 

**Related code:**
I'm posting links directly to the source code because the issue can be seen directly:

The event emitter that is shared:

https://github.com/marmelab/react-admin/blob/3a463d3a73455cc65a37c61852e35a26f7f3a2b2/packages/ra-core/src/dataProvider/undoableEventEmitter.ts#L3

The place where react admin is listening for the given event:

https://github.com/marmelab/react-admin/blob/3a463d3a73455cc65a37c61852e35a26f7f3a2b2/packages/ra-core/src/dataProvider/useDelete.ts#L415

One of the places (in my case the place) where the event is fired (upon clicking the second delete):

https://github.com/marmelab/react-admin/blob/3a463d3a73455cc65a37c61852e35a26f7f3a2b2/packages/ra-ui-materialui/src/layout/Notification.tsx#L87

**Other information:**
I think I provided the correct links and files but I was debugging the issue yesterday and by mistake I closed the files but I'm pretty sure that the links that I've shared are the correct ones.

**Environment**

* React-admin version: Latest
* React version: -
* Browser: -

fzaninotto:
Closes #10269

- [x] Add `UndoableMutationContextProvider` queue and associated hooks
- [x] Use it to take mutations in `Notification`
- [x] Use it to put mutations in data provider mutation hooks
- [x] Deprecate `undoableEventEmitter` (but keep it for BC)
- [x] Add unit tests for the "two mutations in a row" scenario

### DIFF
--- a/packages/ra-ui-materialui/src/layout/Notification.spec.tsx
+++ b/packages/ra-ui-materialui/src/layout/Notification.spec.tsx
@@ -1,0 +1,23 @@
+import * as React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+
+import { ConsecutiveUndoable } from './Notification.stories';
+
+describe('<Notification />', () => {
+    it('should confirm the first undoable notification when a second one starts', async () => {
+        const deleteOne = jest
+            .fn()
+            .mockImplementation((_resource, { id }) =>
+                Promise.resolve({ data: { id } })
+            );
+        const dataProvider = { delete: deleteOne } as any;
+        render(<ConsecutiveUndoable dataProvider={dataProvider} />);
+        await screen.findByText('Post deleted');
+        expect(deleteOne).toHaveBeenCalledTimes(0);
+        await waitFor(() => {
+            expect(deleteOne).toHaveBeenCalledTimes(1);
+        });
+        screen.getByText('ra.action.undo').click();
+        expect(deleteOne).toHaveBeenCalledTimes(1);
+    });
+});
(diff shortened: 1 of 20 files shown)


packages/ra-ui-materialui/src/layout/Notification.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
```
slax57, line 2: linter warning at this line

> 'waitFor' is defined but never used. Allowed unused vars must match /^_/u

slax57 (approved):

PR #10299, opened 2024-10-24
Fix DateInput messes up dates in some timezones

### LINKED ISSUE #10197: DateInput decrease the day by 1 when you change its value and trigger a form validation function
**What you were expecting:**
DateInput to always have the value sent to the validation function exactly as it displays on the component

**What happened instead:**
see below

**Steps to reproduce:**
The DateInput day value decrease by 1 when you log its value on a validation function, but only if you change the date initial value, for example if the initial value is today, when the issue happens, the day of the date is decreased by 1...
How to reproduce: 
- Add a validation prop to your form like so: `validate={validateCreate}`
- Set the forms validation modes to submit: `mode='onSubmit' reValidateMode='onSubmit'`
- Insert a default value on the DateInput, as a prop like this:
`defaultValue={new Date().toISOString().split('T')[0]}`
- Add a simple log to the validation form value for the DateInput (my DateInput source is called start_date): `console.log(values.start_date)`
- Open the page where you have your DateInput and press the save button
- You will see that the date is correct, since it has the initial value
- Change your Date to something else and press Save again, you will see the date has decreased by 1 day, and if you keep testing, you can see that it only happens if you change the date to something else other than the initial value

**Related code:**
Dont have a codesandbox, its very easy to reproduce this

**Environment**

* React-admin version:
`from package-lock: "node_modules/react-admin": { "version": "5.1.3", ... }`
* React version:
* Browser:

I would love to get some idea how to work around this issue

djhi:
## Problem

Fixes #10197

## Solution

We were trying to convert the input value into a `Date`. However, when passing a partial date (such as the html input value `yyyy-MM-dd`) to `Date`, it assumes the UTC timezone. When stringified again, this UTC date might be different than the one entered by users because of their timezone.

## How To Test

- Pass the original user value to react-hook-form.
- Don't try to convert dates strings to dates in parse

## Additional Checks


packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -131,20 +133,15 @@ export const DateInput = ({
...
-                  ? parse(localInputRef.current.value)
-                  : getStringFromDate(localInputRef.current.value);
-
-        if (newValue !== field.value) {
+        const isNewValueValid =
+            newValue === '' ||
+            (localInputRef.current.valueAsDate != null &&
+                !isNaN(new Date(localInputRef.current.valueAsDate).getTime()));
+
+        if (isNewValueValid) {
```
slax57, line 144: Shouldn't you also keep the check `newValue !== field.value`?
This would avoid calling `field.onChange()` if the value has not changed, and the input was only burred.

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -213,10 +210,11 @@ const convertDateToString = (value: Date) => {
     const yyyy = value.getFullYear().toString();
     const MM = (value.getMonth() + 1).toString();
     const dd = value.getDate().toString();
-    return `${yyyy}-${(pad + MM).slice(-2)}-${(pad + dd).slice(-2)}`;
+    const result = `${yyyy}-${(pad + MM).slice(-2)}-${(pad + dd).slice(-2)}`;
+    return result;
 };
 
-const dateRegex = /^\d{4}-\d{2}-\d{2}$/;
+const dateRegex = /^(\d{4}-\d{2}-\d{2}).*$/;
```
slax57, line 217: **nitpick (non-blocking):** You could probably optimize this regex by using `/^\d{4}-\d{2}-\d{2}/` instead of `/^(\d{4}-\d{2}-\d{2}).*$/` (notice the removal of `$`).
Just a nitpick though, feel free to ignore.

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -231,8 +229,10 @@ const getStringFromDate = (value: string | Date) => {
...
     // valid dates should not be converted
-    if (dateRegex.test(value)) {
-        return value;
+    // This regex only returns the date part of the provided string
+    const matches = dateRegex.exec(value);
+    if (matches) {
+        return matches[1];
     }
 
     return convertDateToString(new Date(value));
```
slax57, line 238: **question**: should we still attempt to call `convertDateToString` if the value does not match the regex? :thinking:

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -86,43 +84,47 @@ export const DateInput = ({
...
-                : parse
-                  ? parse(target.value)
-                  : getStringFromDate(target.value);
-
-        // Some browsers will return null for an invalid date so we only change react-hook-form value if it's not null
-        // The input reset is handled in the onBlur event handler
-        if (newValue !== '' && newValue != null) {
-            field.onChange(newValue);
-            valueChangedFromInput.current = true;
+    const handleChange = useEvent(
```
slax57, line 87: I'm surprised you didn't need to perform the same changes as you did to `handleBlur` to this function. Won't we still have the timezone issue in forms with `mode="onChange"`?

packages/ra-ui-materialui/src/input/DateInput.spec.tsx
```diff
@@ -99,25 +76,19 @@ describe('<DateInput />', () => {
...
-                        defaultValues={{ publishedAt: new Date('2021-09-11') }}
-                    >
-                        <DateInput {...defaultProps} />
-                    </SimpleForm>
-                </ResourceContextProvider>
-            </AdminContext>
+            <Basic
+                simpleFormProps={{
+                    onSubmit,
+                    defaultValues: { publishedAt: '2021-09-11' },
```
slax57, line 82: This value does not match the test description (date object). Is it intended?

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: Optimize?

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: RHF handles this itself

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: Yes, because `Date` also handles exotic formats that do not match ISO 8601

packages/ra-ui-materialui/src/input/DateInput.tsx
djhi: Updated

packages/ra-ui-materialui/src/input/DateInput.tsx
slax57: `.*` is a greedy operator, which can be avoided since we only actually care about the start of the string.
But anyways, on such small strings it would have almost no effect, so let's not bother with this.

packages/ra-ui-materialui/src/input/DateInput.stories.tsx
```diff
@@ -45,10 +68,16 @@ export const Validate = () => (
...
+}: {
+    children: React.ReactNode;
+    simpleFormProps?: Partial<SimpleFormProps>;
+}) => (
     <AdminContext i18nProvider={i18nProvider} defaultTheme="light">
         <Create resource="posts">
-            <SimpleForm>
+            <SimpleForm {...simpleFormProps}>
                 {children}
                 <FormInspector name="published" />
```
slax57, line 82: issue: some stories are using `published`, while others are using `publishedAt`, so the `FormInspector` doesn't always work. Can you fix it?

slax57 (changes requested):
Other than that it seems to work well! I emulated my position and locale to Honolulu, and saw no issues at all.
Also tested in Chrome and Firefox successfully. GJ!

fzaninotto (approved):

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -202,38 +205,65 @@ export type DateInputProps = CommonInputProps &
     Omit<TextFieldProps, 'helperText' | 'label'>;
 
 /**
- * Convert Date object to String
+ * Convert Date object to String, ignoring the timezone.
  *
  * @param {Date} value value to convert
  * @returns {String} A standardized date (yyyy-MM-dd), to be passed to an <input type="date" />
  */
 const convertDateToString = (value: Date) => {
```
fzaninotto, line 213: We still need this change for cases where the initial value is a non-standard date string (e.g. 02/02/2001)

PR #10300, opened 2024-10-25
Leverage canAccess in CRM demo

djhi:
## Problem

React Admin `5.3.0` introduced a better [access control mechanism](https://marmelab.com/react-admin/Permissions.html) but the CRM demo still uses the old permissions.

## Solution

Convert the usage of permissions to canAccess.

## How To Test

- Starts the demo

### As an admin (the default)
- Check that you can access the users admin menu
- Check that you can access the sales pages
- Note an email of a "simple" sale

### As the "simple" sale
- There is no *Users* menu item in the top right user menu dropdown
- Trying to access any sales URL by typing it directly in the browser address bar should redirect to the `/access-denied` pag

## Additional Checks


examples/crm/src/providers/fakerest/authProvider.ts
```diff
@@ -50,19 +51,21 @@ export const authProvider: AuthProvider = {
...
         const localUser = userItem ? (JSON.parse(userItem) as Sale) : null;
-        if (!localUser) {
-            return Promise.reject('user is not logged in');
-        }
+        if (!localUser) return false;
 
-        // We fetch permissions from server to avoid local storage tampering
-        const user = await dataProvider.getOne<Sale>('sales', {
+        // Get the matching sale
+        const { data } = await dataProvider.getOne<Sale>('sales', {
```
slax57, line 61: Correct me if I'm wrong, but I believe permissions were cached, whereas the `canAccess` method is called each time we need it. This would result in many calls to `dataProvider.getOne`.
Can't we instead call that at login, and store the `role` in memory or in localStorage?

slax57 (changes requested):

slax57 (approved):

PR #10304, opened 2024-10-26
fixes #10301 race condition in link generation

### LINKED ISSUE #10301: 5.2.2 -> 5.3.0 nested <ReferenceField link={false}><ReferenceField> generated link is flaky & invalid / used "parent.id" and not child.id
Given:
```
<Resource name="comments" representation="id" show={...} />
<Resource name="posts" representation="id" />
<Resource name="authors" representation="name" />

<List> <- comments
  <Datagrid>
    <ReferenceField source="post_id" reference="posts" link={false}> <-- false here is important
       <ReferenceField source="author_id" reference="author" link="show" />
    </ReferenceField>
  </Datagrid>
</List>
```
With such data / relation between `{ comment: {id: 'commentid', post: {id: "postid", author: {id: "authorid", name: "Shakespear"}} }`

On version 5.2.2
i would get `<Link to="#/author/authorid/show">Shakespear</Link>`

On version 5.3.0
During the first render (ie. refresh page) (when the resources are still being fetches async
i get `<Link to="#/author/postid/show">Shakespear</Link>` this link is broken

when i click the comment row
then go back (browser back) to comment list
the `<Link to="#/author/authorid/show">Shakespear</Link>` is valid (the same as in 5.2.2)
And this must be somehow related to all the resources now being (in cache)

I understand that a working demo would be best, but when i tried to replicate this in stackblitz & fake-rest-dataprovider
I can't get the same result

So pls let me know if i can't help to provide more context.

ps. adding a 3'rd "self" reference fixed the link issue (on first render), but that's ... just sad

```
<ReferenceField source="post_id" reference="posts" link={false}>
   <ReferenceField source="author_id" reference="author" link={false}> <-----
       <ReferenceField source="id" reference="author" link="show"> <----
    </ReferenceField>
</ReferenceField>
```

fkowal:
## Problem

fixes an issue (race condition) happening during reference link generation
bug introduced in 5.2.3 -> 5.3.0
https://github.com/marmelab/react-admin/commit/10a1c76f19bb5096f64f95cfb09d0a77a383455e

The issue is somewhat described in 
https://github.com/marmelab/react-admin/issues/10301

Closes #10301

## Solution

`path` is not `set` when async `record` comes in and `link` prop is a `string`

## How To Test

I hope it's clear that the `useEffect` should set this

### DIFF
--- a/packages/ra-core/src/routing/useGetPathForRecord.ts
+++ b/packages/ra-core/src/routing/useGetPathForRecord.ts
@@ -107,10 +107,7 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
                 );
                 return;
             }
-        }
-
-        // Handle the link function case
-        if (typeof link === 'function') {
+        } else if (typeof link === 'function') {
             const linkResult = link(record, resource);
             if (linkResult instanceof Promise) {
                 linkResult.then(resolvedPath => setPath(resolvedPath));
@@ -124,6 +121,14 @@ export const useGetPathForRecord = <RecordType extends RaRecord = RaRecord>(
                           type: linkResult,
                       })
                     : false
+            );
+        } else if (link) {
+            setPath(
+                createPath({
+                    resource,
+                    id: record.id,
+                    type: link,
+                })
             );
         }
     }, [


fzaninotto:
Thanks for your PR. I have created another one with a proper test to demonstrate the fix.

PR #10313, opened 2024-10-29
Fix `<SimpleList>` throws an error when no data in standalone mode

fzaninotto:
## Problem

Since #10184, published in 5.2.0, `<SimpleList>` and `<Datagrid>` throw an error when used in standalone mode (i.e., outside of a ListContext) with an empty `data`.

> useListContext must be used inside a ListContextProvider

## Solution

The problem comes from `ListNoResults`, which reads the `ListContext` to determine if there is an active filter.

https://github.com/marmelab/react-admin/blob/caca09ec556fa3052941071a4d19cd5eed2ea5ea/packages/ra-ui-materialui/src/list/ListNoResults.tsx#L15

## Solution

Use the alternative `useListContextWithProps` instead of `useListContext`.

### DIFF
--- a/packages/ra-ui-materialui/src/list/ListNoResults.tsx
+++ b/packages/ra-ui-materialui/src/list/ListNoResults.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { CardContent, Typography } from '@mui/material';
 import {
     useGetResourceLabel,
-    useListContext,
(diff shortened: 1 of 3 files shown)


slax57 (approved):

erwanMarmelab (commented):
Nice catch

PR #10314, opened 2024-10-29
Fix `<Edit>` ignores mutation meta when updating the `getOne` cache

### LINKED ISSUE #9542: When using `meta` in Edit, it breaks cache update in `useUpdate`
**What you were expecting:**
When I pass `queryOptions` and `mutationOptions` with meta in `Edit` component, then update the record with some new values, the cache in React-Query is updated with queryKey that _includes_ meta.

**What happened instead:**
React-Query cache item is updated with queryKey that misses meta.

**Steps to reproduce:**
1. Go to Posts screen.
2. Click on any post to edit it.
3. Change Title to new value.
4. Press Save - you are returned to List view - OK.
5. Title of the updated Post in List view is updated - OK.
6. Click on the same post to edit it.

Expected result:
Title text input has _new value_ in Edit screen (because the React-Query cache should have been updated with new record).

Actual result:
Title text input has _old value_ for 3 seconds in Edit screen, then it gets updated to new value (3s is a delay in a dataProvider).

**Related code:**
https://stackblitz.com/edit/github-aky8mx-pebq3a?file=src%2Fposts%2FPostEdit.tsx

```jsx
  const meta = { param: 123 }
  
  return (
    <Edit
      title={<PostTitle />}
      actions={<EditActions />}
      mutationOptions={{ meta }}
      queryOptions={{ meta }}
    >
```

I've set a timeout of 3 seconds in `delayedDataProvider`.

**Root cause**
`updateCache` in `useUpdate.ts` uses meta from options, that were destructured in `useEditController`. The meta value was dropped.
`useEditController.ts`:

```ts
    const {
        onSuccess,
        onError,
        meta: mutationMeta,
        ...otherMutationOptions
    } = mutationOptions;

    // ...

    const [update, { isLoading: saving }] = useUpdate<
        RecordType,
        MutationOptionsError
    >(resource, recordCached, {
        ...otherMutationOptions,
        mutationMode,
        returnPromise: mutationMode === 'pessimistic',
    });
```

`useUpdate.ts`
```ts
    const { id, data, meta } = params;
    // ...

    const updateCache = ({ resource, id, data }) => {
        // ...
        queryClient.setQueryData(
            [resource, 'getOne', { id: String(id), meta }],
            (record: RecordType) => ({ ...record, ...data }),
            { updatedAt }
        // ...
    }

    // ...
```

**Environment**

* React-admin version: 4.16.3
* React version: 17.0.2

fzaninotto:
Closes #9542

### DIFF
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -151,7 +151,7 @@ export const useEditController = <
                 : '',
     });
 
-    const recordCached = { id, previousData: record, meta: mutationMeta };
+    const recordCached = { id, previousData: record };
 
     const [update, { isPending: saving }] = useUpdate<RecordType, ErrorType>(
         resource,
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -151,7 +151,7 @@ export const useEditController = <
                 : '',
     });
 
-    const recordCached = { id, previousData: record };
+    const recordCached = { id, previousData: record, meta: mutationMeta };
 
     const [update, { isPending: saving }] = useUpdate<RecordType, ErrorType>(
         resource,
--- a/packages/ra-core/src/dataProvider/useUpdate.ts
+++ b/packages/ra-core/src/dataProvider/useUpdate.ts
@@ -115,7 +115,7 @@ export const useUpdate = <RecordType extends RaRecord = any, ErrorType = Error>(
     // otherwise the other side effects may not applied.
     const hasCallTimeOnSuccess = useRef(false);
 
-    const updateCache = ({ resource, id, data }) => {
+    const updateCache = ({ resource, id, data, meta }) => {
         // hack: only way to tell react-query not to fetch this query for the next 5 seconds
         // because setQueryData doesn't accept a stale time option
         const now = Date.now();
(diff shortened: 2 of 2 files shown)


erwanMarmelab (approved):

packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -151,7 +151,7 @@ export const useEditController = <
                 : '',
     });
 
-    const recordCached = { id, previousData: record };
+    const recordCached = { id, previousData: record, meta: mutationMeta };
```
slax57, line 154: This fixes the issue when the `meta` prop is passed during hook declaration time (i.e. when passed as `mutationOptions` in `<Edit>`), but won't work if the `meta` prop is passed at calltime (i.e. passed to the `<SaveButton>` for instance). I think we should instead pass the calltime `meta` to `updateCache()`, which would work in all cases.

slax57 (changes requested):

packages/ra-core/src/dataProvider/useUpdate.ts
```diff
@@ -273,6 +273,7 @@ export const useUpdate = <RecordType extends RaRecord = any, ErrorType = Error>(
                     resource: callTimeResource,
                     id: callTimeId,
                     data,
+                    meta: mutationOptions.meta ?? paramsRef.current.meta,
```
slax57, line 276: why not simply use `callTimeMeta` here? :thinking:

packages/ra-core/src/controller/edit/useEditController.ts
slax57, line 154: this should no longer be necessary with the new implementation of `useUpdate`, as `mutationMeta` is already passed to `update()` at line 244

slax57 (approved):

PR #10322, opened 2024-10-31
fix(ra-data-graphql-simple): Unify meta location in buildVariables

### LINKED ISSUE #10317: Sparse Field and meta handling different for create/update
**What you were expecting:**
I'm trying out the new `sparseField` feature of ra-data-graphql-simple added in #9392.
However when setting the property: `{ meta: { sparseFields: ['id'] } }` it's not taken into account for create/update mutations. The cause seems to be because `buildCreateUpdateVariables` in `buildVariables` only takes `id` and `data` destructured and ignores `meta`. The unit test for these two query types is written that the `meta` property is inside the `data` object, however the types suggest no such difference: `UpdateParams` still has `data: Partial<RecordType>` and `meta?: any` in its type declarations.

In short the below code should still use the sparse fields, but it's completely ignoring it:

```ts
updateComment(Resource.MatchComment, {
  id: commentId,
  data: {
    meta: { sparseFields: ['createdAt', 'id', 'matchId', 'employeeId', 'text', 'updatedAt', 'pinned'] },
    text: values.text,
  },
});
```

**What happened instead:**
Sparse fields are not used for create/update calls from `params.meta.sparseFields`, they do seem to work on `params.data.meta.sparseFields` however.

**Steps to reproduce:**
It should be easily reproducible/testable by updating the unit tests in `buildVariables.test.ts` for CREATE and UPDATE. `meta` property should be on the top level of the `params` variable, not inside `data`.

**Related code:**
Unit tests for create/update should use top-level `meta`.

```diff
-            const params = {
-                data: {
-                    meta: { sparseFields: [] },
-                },
-            };
+            const params = {
+                data: {},
+                meta: { sparseFields: [] },
+            };
```

**Other information:**

**Environment**

* React-admin version: 5.3.1
* Last version that did not exhibit the issue (if applicable): None, as it's a new feature.
* React version: 18.3.1
* Browser: Chrome

dricholm:
## Problem

Fixes #10317.
The sparse fields functionality works by reading the `meta.sparseFields` array. But for create and update mutations it looked for the `sparseFields` value in `data.meta`.

## Solution

Return the `meta` object in `buildCreateUpdateVariables` which is used for both create and update mutations.

## How To Test

Can be tested by setting the `meta` object for a create or update mutation and setting sparse fields on it:

```ts
{
  meta: { sparseFields: ['id'] }
}
```

Trigger the mutation and the generated selection should only include fields specified in `sparseFields`.

## Additional Checks

  - Not UI related, affects data provider. I'm not aware that it has any stories.

### DIFF
--- a/packages/ra-data-graphql-simple/src/buildVariables.test.ts
+++ b/packages/ra-data-graphql-simple/src/buildVariables.test.ts
@@ -19,6 +19,7 @@ describe('buildVariables', () => {
             },
         ],
     };
+
     describe('GET_LIST', () => {
         it('returns correct variables', () => {
             const params = {
@@ -55,8 +56,8 @@ describe('buildVariables', () => {
 
         it('should return correct meta', () => {
             const params = {
-                filter: {},
-                meta: { sparseFields: [] },
+                filter: { views: 100 },
+                meta: { sparseFields: ['field'] },
             };
 
             expect(
@@ -67,8 +68,8 @@ describe('buildVariables', () => {
                     {}
                 )
             ).toEqual({
-                filter: {},
-                meta: { sparseFields: [] },
+                filter: { views: 100 },
+                meta: { sparseFields: ['field'] },
             });
         });
     });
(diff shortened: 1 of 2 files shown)


djhi (approved):
Thanks!

PR #10324, opened 2024-10-31
fix(ra-ui-materialui):  Fix <TabbedFormView/> to support string ids with spaces

### LINKED ISSUE #9545: String Ids with spaces are not supported by `TabbedForm`
**What you were expecting:**
I have an entity with string Id that has a space in it. When I navigate to Edit view with `TabbedForm` I expect to see working tabbed form and first tab is opened by default.

**What happened instead:**
No tabs are rendered.

**Steps to reproduce:**

**Related code:**
I digged into the RA source code and found out, that TabbedFormView uses raw location.pathname to match with tab path. Obviously, spaces in URL are converted to %20 strings.

```js
const hidden = syncWithLocation
    ? !matchPath(
          `${resolvedPath.pathname}/${tabPath}`,
          location.pathname
      )
    : tabValue !== index;
```

Screenshot 2023-12-26 at 12 27 31

**Environment**

* React-admin version: 4.14.3
* React version: 18.2.0

geobde:
Bug Fix: TabbedForm Support for IDs with Spaces

### Problem:
`TabbedForm`  fails to render tabs when the entity ID contains spaces. 
This occurs because `TabbedFormView` compares the raw location.pathname (which contains 
encoded spaces as %20) with tab paths, causing the path matching to fail.

### Solution:
Fixed `getDecodedPathname` utility function that:
1. Splits the pathname into segments
2. Decodes URI components to handle %20 and other encoded characters
3. Removes any remaining spaces
4. Rejoins the path

This ensures consistent path matching regardless of spaces in entity IDs, allowing 
TabbedForm to properly render and navigate between tabs.

Fixes #9545


geobde:
Hello! No, we don't have the same issue with the `TabbedShowLayout` component because it doesn't use `location.pathname` to match the tab path.

geobde:
@djhi i just added a story & unit test

djhi:
I tried your PR locally and it seems it already work in v5 without the fix. Your issue mention react-admin v4 though. Can you confirm?

fzaninotto:
I can confirm the bug exists in v5

fzaninotto:
I can no longer reproduce the bug in v5. Your new story works fine without the patch. Using '1 1' as an id in the simple example works in postEdit and PostShow, too. 

Can you make sure you can still reproduce the bug on your side?

fzaninotto:

Feel free to comment with a reproduction.

PR #10326, opened 2024-11-04
[Doc] Improve `useUpdate` usage instructions

fzaninotto:
## Problem

`useUpdate` is a powerful hook, but it's not apparent when reading its documentation. 

The doc if its options is to be found in the "querying the API" chapter, but that's not necessarily where developers will look for it. 

## Solution

Develop the `useUpdate` doc, sometimes by duplicating part of the doc for "Querying the API".

### DIFF
--- a/docs/useUpdate.md
+++ b/docs/useUpdate.md
@@ -5,67 +5,326 @@ title: "useUpdate"
 
 # `useUpdate`
 
-This hook allows to call `dataProvider.update()` when the callback is executed, and update a single record based on its `id` and a `data` argument.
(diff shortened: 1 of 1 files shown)


docs/useUpdate.md
```diff
@@ -5,67 +5,326 @@ title: "useUpdate"
...
-    options
-);
+const [update, { isPending }] = useUpdate(resource, params, options);
+const handleClick = () => {
+    update();
+};
 ```
 
-This means the parameters can be passed either when calling the hook, or when calling the callback. It's up to you to pick the syntax that best suits your component. If you have the choice, we recommend passing the parameters when calling the `update` callback (second example below).
+It's up to you to pick the syntax that best suits your component. If you have the choice, we recommend passing the first syntax.
```
djhi, line 30:
```suggestion
It's up to you to pick the syntax that best suits your component. If you have the choice, we recommend using the first syntax.
```

djhi (approved):

PR #10337, opened 2024-11-07
Fix Datagrid header tooltip shows column names with a capital letter

fzaninotto:
## Problem

The tooltip that appears on hover of a sortable datagrid column shows the column name capitalized. 

> Sort by **Title** ascending

image

## Solution

Provide an uncapitalized version of the field to let i18n providers choose the right version to use. 

image

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/DatagridHeaderCell.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/DatagridHeaderCell.tsx
@@ -32,17 +32,22 @@ export const DatagridHeaderCell = (
             : // non active sort field, use default order
               field?.props.sortByOrder ?? 'ASC'
         : undefined;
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/list/datagrid/DatagridHeaderCell.tsx
```diff
@@ -32,17 +32,22 @@ export const DatagridHeaderCell = (
...
-                  label:
-                      typeof field.props.label === 'string'
-                          ? field.props.label
-                          : undefined,
-                  resource,
-                  source: field.props.source,
-              })
-            : undefined,
+        field: fieldLabel,
+        field_lower_first:
```
djhi, line 47: `field_lowercased`?

djhi (approved):

Issue #10323, opened 2024-10-31
Removal of useSafeSetState


dricholm:
**Is your feature request related to a problem? Please describe.**
I noticed there is a function called `useSafeSetState` in `ra-core/src/util/hooks.ts` which is intended to only call `setState` if the component is still mounted.
There was a warning for this in React before version 18 which warned us that the app could have a memory leak. This was widely misunderstood as the warning is only for a _possible_ memory leak, most cases were not memory leaks, just a no-op call.
I believe there's no advantage to having it now, it even adds some slight overhead so could lead to a performance boost. Although I assume not that much, unless lots of states are handled on a page.

**Describe the solution you'd like**
Replacing every `useSafeSetState` with `React.useState` and removing `useSafeSetState`.
Double-check that any possible subscriptions are unsubscribed so no memory leak could occur.

**Describe alternatives you've considered**
Current way works as well, just with more overhead.

**Additional context**
I would recommend this discussion which goes into more detail and has examples as well: https://github.com/reactwg/react-18/discussions/82

geobde:
@dricholm good point! i dont see the value of that hook

fzaninotto:
Good point. Would you like to open a PR to remove this hook?

djhi:
Closed by [#10341](https://github.com/marmelab/react-admin/pull/10341)

PR #10341, opened 2024-11-10
feat(ra-core): remove useSafeSetState and use React.useState instead

geobde:
Problem
The `useSafeSetState` hook was previously used to prevent warnings when attempting to call setState on unmounted components. However, with recent changes to React (removal of the "setState on unmounted component" warning), the use of this hook has become unnecessary and leads to unnecessary complexity.

Solution
Replaced the `useSafeSetState` hook with the `React.useState` hook. Since React no longer issues warnings when calling setState on unmounted components, the need for the safety check (i.e., using mountedRef and useEffect for cleanup) is no longer required.

Context
[Update to remove the "setState on unmounted component" warning
](https://github.com/reactwg/react-18/discussions/82)

Fixes https://github.com/marmelab/react-admin/issues/10323

### DIFF
--- a/examples/crm/src/login/LoginForm.tsx
+++ b/examples/crm/src/login/LoginForm.tsx
@@ -7,7 +7,7 @@ import {
     CircularProgress,
     Typography,
 } from '@mui/material';
-import { Form, useLogin, useNotify, useSafeSetState } from 'ra-core';
+import { Form, useLogin, useNotify } from 'ra-core';
 import { Login, TextInput } from 'react-admin';
 import { SubmitHandler } from 'react-hook-form';
 import { Link } from 'react-router-dom';
@@ -36,7 +36,7 @@ const StyledForm = styled(Form, {
 }));
 
 export const LoginForm = () => {
-    const [loading, setLoading] = useSafeSetState(false);
+    const [loading, setLoading] = React.useState(false);
     const login = useLogin();
     const notify = useNotify();
 
--- a/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
+++ b/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
@@ -9,7 +9,6 @@ import { AuthContext } from './AuthContext';
 import useLogout from './useLogout';
 import { useNotify } from '../notification/useNotify';
 import { AuthProvider } from '../types';
-import { useSafeSetState } from '../util';
 
 import { TestMemoryRouter } from '../routing';
 
@@ -40,7 +39,7 @@ const authProvider: AuthProvider = {
 };
 
 const TestComponent = ({ error }: { error?: any }) => {
-    const [loggedOut, setLoggedOut] = useSafeSetState(false);
+    const [loggedOut, setLoggedOut] = React.useState(false);
     const logoutIfAccessDenied = useLogoutIfAccessDenied();
     useEffect(() => {
         logoutIfAccessDenied(error).then(setLoggedOut);
--- a/packages/ra-core/src/util/hooks.ts
+++ b/packages/ra-core/src/util/hooks.ts
@@ -1,32 +1,7 @@
-import * as React from 'react';
-import { useState, useRef, useEffect, useCallback } from 'react';
+import { useRef, useEffect, useState } from 'react';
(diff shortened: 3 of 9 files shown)


examples/crm/src/login/LoginForm.tsx
```diff
@@ -7,7 +7,7 @@ import {
     CircularProgress,
     Typography,
 } from '@mui/material';
-import { Form, useLogin, useNotify, useSafeSetState } from 'ra-core';
+import { Form, useLogin, useNotify } from 'ra-core';
```
erwanMarmelab, line 10:
```suggestion
import { Form, useLogin, useNotify, useState } from 'ra-core';
```

examples/crm/src/login/LoginForm.tsx
```diff
@@ -36,7 +36,7 @@ const StyledForm = styled(Form, {
 }));
 
 export const LoginForm = () => {
-    const [loading, setLoading] = useSafeSetState(false);
+    const [loading, setLoading] = React.useState(false);
```
erwanMarmelab, line 39:
```suggestion
    const [loading, setLoading] = useState(false);
```

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -40,7 +39,7 @@ const authProvider: AuthProvider = {
 };
 
 const TestComponent = ({ error }: { error?: any }) => {
-    const [loggedOut, setLoggedOut] = useSafeSetState(false);
+    const [loggedOut, setLoggedOut] = React.useState(false);
```
erwanMarmelab, line 42:
```suggestion
    const [loggedOut, setLoggedOut] = useState(false);
```

Could you please import it

erwanMarmelab (changes requested):

examples/crm/src/login/LoginForm.tsx
```diff
@@ -7,7 +7,7 @@ import {
     CircularProgress,
     Typography,
 } from '@mui/material';
-import { Form, useLogin, useNotify, useSafeSetState } from 'ra-core';
+import { Form, useLogin, useNotify, useState } from 'ra-core';
```
fzaninotto, line 10: That's wrong, `ra-core` doesn't export a `useState` hook.

packages/ra-core/src/util/hooks.ts
```diff
@@ -1,33 +1,8 @@
-import * as React from 'react';
-import { useState, useRef, useEffect, useCallback } from 'react';
+import { useRef, useEffect, useState } from 'react';
 import isEqual from 'lodash/isEqual';
 
 // thanks Kent C Dodds for the following helpers
 
-export function useSafeSetState<T>(
```
fzaninotto, line 7: Please keep the export for backward compatibility, even if we don't use it. Mark it as deprecated instead.

fzaninotto (changes requested):

examples/crm/src/login/LoginForm.tsx
geobde: fixed!

examples/crm/src/login/LoginForm.tsx
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import { useState } from 'react';
```
fzaninotto, line 1: you need to keep the import * for JSX

fzaninotto (changes requested):

packages/ra-core/src/util/hooks.ts
fzaninotto: bump

geobde: useSafeSetState is back

fzaninotto (approved):

djhi (approved):

djhi:
We still support react 17. What about people still using it?

fzaninotto:
> We still support react 17. What about people still using it?

No, react-admin v5 requires React 18. 

https://marmelab.com/react-admin/Upgrade.html#react-18

djhi:
Thanks!

PR #10343, opened 2024-11-12
Fix `<SimpleList linkType={false}>` UI

erwanMarmelab:
## Problem

`linkType` should only remove the onClick action but it breaks the UI.

### Without `linkType`
image
### With `linkType={false}`
image

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList/SimpleList.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/SimpleList/SimpleList.stories.tsx
@@ -197,7 +197,7 @@ export const IconsAvatarsAndLinkType = () => {
                             }
                         />
                     }
-                    label="LinkType False"
+                    label="LinkType"
                 />
                 <Box>
                     <FormControlLabel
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/list/SimpleList/SimpleList.stories.tsx
```diff
@@ -137,6 +137,28 @@ export const FullApp = () => (
     </AdminContext>
 );
 
+export const AnyLink = () => (
```
fzaninotto, line 140:
```suggestion
export const LinkTypeFalse = () => (
```

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
```diff
@@ -286,6 +287,12 @@ const LinkOrNot = (
             <ListItemText
                 // @ts-ignore
                 component="div"
+                sx={{
+                    paddingX: 2,
+                    paddingY: 1,
```
fzaninotto, line 292: use `px`, `py` and `m` instead

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
```diff
@@ -297,6 +304,7 @@ const LinkOrNot = (
         <ListItemButton
             component={Link}
             to={createPath({ resource, id, type })}
+            sx={{ ...sx }}
```
fzaninotto, line 307: Why do you clone the object? Just pass it normally.

fzaninotto (changes requested):

djhi (approved):

PR #10344, opened 2024-11-12
Introduce `<FilterLiveForm>`

### LINKED ISSUE #5591: List Filter Value "jumps" if you enter a new value without waiting for a response from the old
**What you were expecting:**
The value entered in the filter field is expected to be independent of the response from the dataProvider.

**What happened instead:**
Instead, if you start typing a new filter value without waiting for a response from the old one, as there are asynchronous responses, the filter input "jumps", displaying the value that was in the previous request.
This is especially noticeable on slow connections when using the q filter or input fields that use a mask.

**Steps to reproduce:**
oDSTtXb8Xp

* React-admin version: 3.9.6

slax57:
## Problem

The `FilterList` is designed for buttons - not inputs. We provide the `FilterLiveSearch` component, which is a `TextInput` inside a form that submits on change. But some users want to be able to add custom form inputs (e.g. to select an author from a resource using an autocomplete), and re-writing the form wrapper is cumbersome.

## Solution

Introduce a new component to wrap form inputs in that case:

```tsx
export const EventFilterAside = () => (
    <Card>
        <CardContent>
            <FilterList label="Has ordered" icon={<MonetizationOnIcon />}>
                <FilterListItem
                    label="True"
                    value={{
                        nb_commands_gte: 1,
                        nb_commands_lte: undefined,
                    }}
                />
                <FilterListItem
                    label="False"
                    value={{
                        nb_commands_gte: undefined,
                        nb_commands_lte: 0,
                    }}
                />
            </FilterList>
            <FilterLiveForm>
                <TextInput source="q" />
            </FilterLiveForm>
            <FilterLiveForm>
                <ReferenceInput source="authorId" reference="authors" />
            </FilterLiveForm>
        </CardContent>
    </Card>
);
```

This PR also fixes #5591 :tada: 

## Todo

- [x] Introduce `<AutoSubmitFilterForm>`
- [x] Introduce `<FilterListWrapper>` to make it look nicer next to `<FilterList>`
- [x] Add stories
- [x] Add tests
- [x] Write doc
- [x] Try to fix [debounce issue](https://github.com/marmelab/react-admin/issues/5591)
  - [x] Story
  - [x] Fix
  - [x] Unit test
  - [x] Check if the behavior when there is validation at play is OK
- [x] Try to fix issue where `FilterLiveSearch` keeps applying old value
- [x] Use `AutoSubmitFilterForm` in `FilterForm` to keep the code more consistent
- [x] Use `AutoSubmitFilterForm` in `FilterLiveSearch` to keep the code more consistent

## How To Test

- `<FilterLiveForm>` logic: http://localhost:9010/?path=/story/ra-core-controller-list-filter-filterliveform--basic
- `<FilterLiveForm>` UI: http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterliveform--full-app
- `<FilterListSection>`: http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterlistsection--basic
- Filter debounce bug: http://localhost:9010/?path=/story/ra-ui-materialui-input-searchinput--race-condition
- `<FilterLiveSearch>` old value bug: http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterlivesearch--full-app
  1. Enter a filter value (e.g. 'foo')
  2. Navigate to Countries
  3. Navigate back to Books
  4. Clear the `<FilterLiveSearch>`
  5. Confirm the old value ('foo') does not come back

## Additional Checks


packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -30,35 +24,45 @@ import { FilterFormInput } from './FilterFormInput';
 import { FilterContext } from '../FilterContext';
 
 export const FilterForm = (props: FilterFormProps) => {
-    const { defaultValues, filters: filtersProps, ...rest } = props;
```
slax57, line 33: `defaultValues` is not used in all RA's code, and I tested it and it does not work, so I removed it to reduce tech debt. Shouldn't be a BC since it does not currently work.

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -30,35 +24,45 @@ import { FilterFormInput } from './FilterFormInput';
...
+    const { filters: filtersProps, ...rest } = props;
 
     const { setFilters, displayedFilters, filterValues } = useListContext();
     const filters = useContext(FilterContext) || filtersProps;
 
-    const mergedInitialValuesWithDefaultValues =
-        mergeInitialValuesWithDefaultValues(
-            defaultValues || filterValues,
-            filters
-        );
```
slax57, line 42: This whole call is actually useless, because:
1. `defaultValues` don't work
2. `mergeInitialValuesWithDefaultValues` actually does nothing, since there is a runtime check to prevent having both `alwaysOn` and `defaultValue`. It's probably a historic tech debt, that's why I removed it.

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -270,7 +252,7 @@ const getInputValue = (
     filterValues: Record<string, any>
 ) => {
     if (formValues[key] === undefined || formValues[key] === null) {
-        return '';
+        return get(filterValues, key, '');
```
slax57, line 255: same, historic bug (see updated unit test)

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
```diff
@@ -15,30 +15,37 @@ import {
...
+    const { filterValues, setFilters } = useListContext();
     const resource = useResourceContext(props);
 
-    const { debounce = 500, resolver, validate, children, ...rest } = props;
+    const {
+        debounce = 500,
+        resolver,
+        validate,
+        children,
+        component: Component,
```
djhi, line 32: Why don't you provide a default value here and avoid the ternary at L124?

packages/ra-ui-materialui/src/list/filter/FilterListWrapper.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { ReactNode } from 'react';
+import { Box, BoxProps, Typography } from '@mui/material';
+import { useTranslate } from 'ra-core';
+
+export const FilterListWrapper = (props: FilterListWrapperProps) => {
```
fzaninotto, line 6: What's the purpose of extracting this component?

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
slax57: I didn't know how to do it, but now I know! Thanks!

packages/ra-ui-materialui/src/list/filter/FilterListWrapper.tsx
slax57: I found it neat to separate the logic from the presentaiton. `<AutoSubmitFilterForm>` is a logic-only component, which makes it more versatile as it can be used in any UI.
But I still wanted an easy way to use `<AutoSubmitFilterForm>` alongside `<FilterList>` (which was the objective in the first place), so reusing part of the UI from `<FilterList>` felt like a good way to achieve that while keeping the DX not too bad.

djhi (commented):
We should also probably introduce:
- `FilterList.Item`
- `FilterList.Form`
-
**edit:** `FilterForm` is not used inside a `FilterList` so maybe not

packages/ra-ui-materialui/src/list/filter/FilterListWrapper.tsx
djhi, line 6: I would name this `FilterListSection`

docs/FilterLiveSearch.md
```diff
@@ -12,7 +12,7 @@ title: "The FilterLiveSearch Component"
 </video>
 
 
-The filter sidebar is not a form. Therefore, if your users need to enter complex filters, you'll have to recreate a filter form using react-hook-form (see the [Building a custom filter](./FilteringTutorial.md#building-a-custom-filter) for an example). However, if you only need one text input with a filter-as-you-type behavior, you'll find the `<FilterLiveSearch>` component convenient.
+The filter sidebar is not a form. Therefore, if your users need to enter complex filters, you'll have to recreate a filter form. This can easily be done thanks to the [`<AutoSubmitFilterForm>`](./AutoSubmitFilterForm.md) component. However, if you only need one text input with a filter-as-you-type behavior, you'll find the `<FilterLiveSearch>` component even more convenient.
```
djhi, line 15:
```suggestion
The filter sidebar is not a form. Therefore, if your users need to enter complex filters, you'll have to recreate a filter form. This can be done thanks to the [`<AutoSubmitFilterForm>`](./AutoSubmitFilterForm.md) component. However, if you only need one text input with a filter-as-you-type behavior, you'll find the `<FilterLiveSearch>` component even more convenient.
```

packages/ra-core/src/util/useDebouncedEvent.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    // Whenever callback or delay changes, we need to update the debounced callback
+    useEffect(() => {
+        debouncedCallbackRef.current = debounce(callback, delay);
+    }, [callback, delay]);
+
+    // The function returned by useCallback will invoke the debounced callback
+    // Its dependencies array is empty, so it never changes across re-renders
+    return useCallback((...args) => debouncedCallbackRef.current(...args), []);
+};
```
djhi, line 19: This doesn't have the checks we do in `useEvent` to prevent calling the callback inside a render. I'm not really sure of the consequences but the react team has made it a strong limitation of `useEvent` (see its code in our repo)

djhi (changes requested):

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+ * );
+ */
+export const AutoSubmitFilterForm = (props: AutoSubmitFilterFormProps) => {
+    const { filterValues, setFilters } = useListContext();
+    const resource = useResourceContext(props);
+
+    const {
+        debounce = 500,
+        resolver,
+        validate,
```
erwanMarmelab, line 97: When you have a non validated input you can't submit. But if you submit your `FilterList` in a same time, it will reset your `AutoSubmitForm` to the last validated state.
Do we need to consider it as an issue or as a normal way of work @fzaninotto @djhi ?

https://github.com/user-attachments/assets/63c241dd-30ae-4568-8724-8f6a44cabd6f

erwanMarmelab (approved):

erwanMarmelab:
Such a really nice job :mechanical_arm:

erwanMarmelab:
## Problem

With a `FilterLiveSearch` or an `AutoSubmitFilterForm`. If I write something and click on the `clear filters` button faster than my debounce, the new search query happens after the clearing.

It can easily happens when your app need to increase a bit the debounce.

Moreover, when it happens, your input seems to be empty… it’s confusing

## Solution

Make the `clear filters` button cancel next debounced search query

fzaninotto (changes requested):
Only reviewed documentation so far

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+ *     TextInput,
+ * } from 'react-admin';
+ *
+ * const BookListAside = () => (
+ *     <Card sx={{ order: -1, mr: 2, mt: 6, width: 250, height: 'fit-content' }}>
+ *         <CardContent>
+ *             <FilterList label="Century" icon={<CategoryIcon />}>
+ *                 <FilterListItem
+ *                     label="21st"
+ *                     value={{ year_gte: 2000, year_lte: null }}
```
fzaninotto, line 53:
```suggestion
 *                     value={{ year_gte: 2000, year_lte: undefined }}
```

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { useListContext } from '../useListContext';
+
+/**
+ * This component offers a convenient way to create a form that automatically
+ * updates the filters when the user changes its child input values.
+ *
+ * It fits nicely alongside a `<FilterList>` component, but you can also use it
+ * at other places to create your own filter UI.
+ *
+ * @example
```
fzaninotto, line 27: Perhaps a shorter example would be better for inline doc

packages/ra-core/src/controller/list/filter/AutoSubmitFilterForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+    <form {...props} />
+);
+
+export interface AutoSubmitFilterFormProps
+    extends Omit<UseFormProps, 'onSubmit' | 'defaultValues'> {
+    children: ReactNode;
+    validate?: ValidateForm;
+    debounce?: number | false;
+    resource?: string;
+    component?: React.ComponentType<any>;
```
fzaninotto, line 197: I'd name it `formComponent` instead, as it must be a form accepting an `onSubmit` prop. Also, we probably need to narrow down the type.

packages/ra-core/src/controller/list/filter/index.ts
fzaninotto, line 1: I'd put this file in the `form` directory instead. One level higher, plus that's where I'd look for it.

packages/ra-ui-materialui/src/list/filter/AutoSubmitFilterForm.stories.tsx
```diff
@@ -0,0 +1,513 @@
...
+    ListContextProvider,
+    required,
+    Resource,
+    useList,
+    useListContext,
+} from 'ra-core';
+import fakeRestDataProvider from 'ra-data-fakerest';
+import polyglotI18nProvider from 'ra-i18n-polyglot';
+import englishMessages from 'ra-language-english';
+import * as React from 'react';
```
fzaninotto, line 17: Move as first import

packages/ra-ui-materialui/src/list/filter/AutoSubmitFilterForm.stories.tsx
```diff
@@ -0,0 +1,513 @@
...
+    ],
+});
+
+const BookListAside = () => (
+    <Card sx={{ order: -1, mr: 2, mt: 6, width: 250, height: 'fit-content' }}>
+        <CardContent>
+            <FilterList label="Century" icon={<CategoryIcon />}>
+                <FilterListItem
+                    label="21st"
+                    value={{ year_gte: 2000, year_lte: null }}
```
fzaninotto, line 416:
```suggestion
                    value={{ year_gte: 2000, year_lte: undefined }}
```

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -1,105 +1,41 @@
...
-            // We can't rely on form state as it might not be synchronized yet
-            const isFormValid = await trigger();
-
-            // Check that the name is present to avoid setting filters when watch was
-            // triggered by a change on the ListContext values.
-            if (name && isFormValid) {
-                if (get(values, name) === '') {
-                    const newValues = cloneDeep(values);
-                    unset(newValues, name);
-                    setFilters(newValues, displayedFilters, true);
```
fzaninotto, line 75: This piece of logic wasn't ported. Are you sure it wasn't required?

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
```diff
@@ -1,105 +1,41 @@
...
+        </AutoSubmitFilterForm>
     );
 };
 
-export type FilterFormProps = FilterFormBaseProps & {
-    defaultValues?: FieldValues;
-};
+export type FilterFormProps = FilterFormBaseProps;
 
 export const FilterFormBase = (props: FilterFormBaseProps) => {
```
fzaninotto, line 37: I don't think it makes sense to separate the logic into two components anymore. I suggest you reintegrate FilterFormBase into FilterForm.

fzaninotto (changes requested):
Awesome!

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
slax57: I wasn't sure at first, but according to my tests, the new filterValues are properly cleaned up everytime. So apparently it's not required.
But maybe it would be safer to reintroduce it anyways. So I'll do it.

packages/ra-ui-materialui/src/list/filter/FilterForm.tsx
slax57: Actually we still need it because `FilterFormBase` needs to access the form context (`const form = useFormContext()`) created by `AutoSubmitFilterForm`.

docs/AutoSubmitFilterForm.md
```diff
@@ -0,0 +1,158 @@
+---
+layout: default
+title: "AutoSubmitFilterForm"
+---
+
+# `<AutoSubmitFilterForm>`
```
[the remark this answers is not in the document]
slax57: I went with FilterLiveForm

docs/img/AutoSubmitFilterForm.mp4
fzaninotto, line 1: You forgot to rename the images!

slax57: nice catch!

fzaninotto (approved):

djhi (changes requested):

PR #10357, opened 2024-11-18
[Doc] Improve Access Control for Custom Pages

fzaninotto:
## Problem

The current doc for securing custom pages with access control misses two things:

1. The AccessDenied page (the current syntax only shows an empty page when the user doesn't have access)
2. Instructions for tweaking the Menu

## Solution

Improve the doc

### DIFF
--- a/docs/CanAccess.md
+++ b/docs/CanAccess.md
@@ -42,4 +42,59 @@ const UserEdit = () => (
 | `accessDenied` |          | `ReactNode`    | -                     | The element displayed when users are denied access to the resource |
 | `error`        |          | `ReactNode`    | -                     | The element displayed when an error occurs while calling `authProvider.canAccess` |
 
+## Securing Custom Routes
+
+By default, there is no authentication or authorization control on custom routes. If you need to restrict access to a custom route, wrap the content with `<CanAccess>`. Remember to check the authentication status before with `<Authenticated>`:
+
+```tsx
+import { Authenticated, CanAccess, AccessDenied } from 'react-admin';
+
+export const LogsPage = () => (
+    <Authenticated>
+        <CanAccess resource="logs" action="read" accessDenied={<AccessDenied />}>
+            ...
+        </CanAccess>
+    </Authenticated>
+);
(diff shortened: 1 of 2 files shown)


docs/CanAccess.md
```diff
@@ -42,4 +42,59 @@ const UserEdit = () => (
...
+## Securing Custom Routes
+
+By default, there is no authentication or authorization control on custom routes. If you need to restrict access to a custom route, wrap the content with `<CanAccess>`. Remember to check the authentication status before with `<Authenticated>`:
+
+```tsx
+import { Authenticated, CanAccess, AccessDenied } from 'react-admin';
+
+export const LogsPage = () => (
+    <Authenticated>
+        <CanAccess resource="logs" action="read" accessDenied={<AccessDenied />}>
```
fzaninotto, line 54: The important part is the `accessDenied={<AccessDenied />}` part, which was missing.

docs/Permissions.md
```diff
@@ -253,30 +253,59 @@ const CommentsToolbar = ({ record }) => (
...
+            <Menu.Item primaryText="Logs" to="/logs" leftIcon={<SsidChartIcon />} />
+        </CanAccess>
+    </Menu>
+);
+```
+
 **Note**: You don't need to use `<CanAccess>` on the core react-admin page components (`<List>`, `<Create>`, `<Edit>`, `<Show>`) because they already have built-in access control.
 
-**Note**: You don't need to use `<CanAccess>` on custom pages if your admin uses [`requireAuth`](./Admin.md#requireauth).
+**Note**: You don't need to use `<Authenticated>` on custom pages if your admin uses [`requireAuth`](./Admin.md#requireauth).
```
djhi, line 308: I'd move that tip below the example that uses `<Authenticated>`

fzaninotto: I thought about that, too, but I feel that it interrupts the flow of the explanation for a particular case. So I'm -1 for this change.

djhi (approved):

PR #10362, opened 2024-11-20
Fix `useCanAccessResources` cannot have react-query options

djhi:
## Problem

`useCanAccessResources` currently uses `useQueries`. As such, it doesn't accept `useQuery` options like `enabled`.

## Solution

Refactor it to use `useQuery` instead.

## How To Test

- All tests should pass without modification

## Additional Checks

~~- [ ] The **documentation** is up to date~~

### DIFF
--- a/packages/ra-core/src/auth/useCanAccessResources.ts
+++ b/packages/ra-core/src/auth/useCanAccessResources.ts
@@ -74,12 +74,18 @@ export const useCanAccessResources = <
                 })
             );
 
+            const canAccessMap = queries.reduce(
+                (acc, { resource, canAccess }) => {
+                    acc[resource] = canAccess;
+                    return acc;
+                },
+                {} as Record<string, boolean>
+            );
+
             const result = resources.reduce(
                 (acc, resource) => ({
                     ...acc,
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/auth/useCanAccessResources.ts
```diff
@@ -75,10 +71,22 @@ export const useCanAccessResources = <
                     });
 
                     return { canAccess, resource };
-                },
-            };
-        }),
-        combine: combineSourceAccessResults<ErrorType>,
+                })
```
fzaninotto, line 74: suggestion: *reduce* that array of objects to make it an object with only the allowed resources. This will make the lookup below much simpler, and more effective (n finds : O(n2) => n object lookups : O(n))

fzaninotto (changes requested):

packages/ra-core/src/auth/useCanAccessResources.ts
djhi: Done

fzaninotto (approved):

PR #10368, opened 2024-11-20
Fix SortButton default translation

fzaninotto:
## Before

image

This is a regression caused by #10337

## After

image

## How to test

- Launch the ecommerce demo
- Go to the posters list
- See the sort button


djhi (approved):

djhi (commented):
Can't we add a test for this?

PR #10372, opened 2024-11-21
[Doc] Update `useUpdate` doc to explain `returnPromise` option

fzaninotto:


djhi (commented):
I know we have many such examples, but shouldn't we make them closer to real world usage (inside a custom button, etc)?

fzaninotto:
> I know we have many such examples, but shouldn't we make them closer to real world usage (inside a custom button, etc)?

This is the closest to real world usage I could find.

djhi:
> > I know we have many such examples, but shouldn't we make them closer to real world usage (inside a custom button, etc)?
> 
> This is the closest to real world usage I could find.

I meant have a real button with JSX and imports

fzaninotto:
> > > I know we have many such examples, but shouldn't we make them closer to real world usage (inside a custom button, etc)?
> > 
> > 
> > This is the closest to real world usage I could find.
> 
> I meant have a real button with JSX and imports

IMHO it dilutes the example. In addition, we don't do it for the other options.

Issue #10223, opened 2024-09-19
React Admin buttons don't trigger tooltips


michel-paiva:
**What you were expecting:**
When I wrap a React Admin custom button with a `Tooltip`, (for example `CreateButton`) I expect the `Tooltip` to be shown when I hover over it.

**What happened instead:**

The tooltip is not shown when hovering over a custom button wrapped by a `Tooltip`.

**Steps to reproduce:**

**Related code:**

Check the example where in the `postListActions` I have a `CreateButton` wrapped by a `Tooltip` and a material-ui [here](https://stackblitz.com/edit/github-htv2ru?file=src%2Fposts%2FPostList.tsx)

**Other information:**

This wasn't introduced now, it seems like an existing issue.
On `material-ui` doc they have [this section](https://stackblitz.com/edit/github-htv2ru?file=src%2Fposts%2FPostList.tsx) talking about custom child components

**Environment**

* React-admin version: 5.2

fzaninotto:
Thanks for the report. The problem comes from react-admin's `<Button>` and its derivatives (like `<CreateButton>`), which don't use `forwardRef`. 

A PR to fix this is welcome.

VikashChoudhary001:
Hi! The bug label was removed recently, so I wanted to confirm if this issue is still open for work or if it has already been resolved.

VikashChoudhary001:
Hi there,

It seems that the issue is caused by the fact that custom buttons like CreateButton do not natively trigger the mouseenter and mouseleave events required by the Tooltip component. To fix this, you can simply wrap the CreateButton inside a div to ensure the tooltip triggers on hover.

example ;-
`<Tooltip title="Create Post" disableInteractive>
  <div>
    <CreateButton />
  </div>
</Tooltip>
`
This approach ensures that the necessary mouse events are correctly fired, and the tooltip will work as expected.

Let me know if you need further clarification!

VikashChoudhary001:
Hi @fzaninotto ,

It seems that the issue is caused by the fact that custom buttons like `CreateButton` do not natively trigger the `mouseenter` and `mouseleave` events required by the `Tooltip` component. To fix this, you can simply **wrap the `CreateButton` inside a `div`** to ensure the tooltip triggers on hover.

Here’s an example of how to fix it:

```jsx
<Tooltip title="Create Post" disableInteractive>
  <div>
    <CreateButton />
  </div>
</Tooltip>
```
Let me know if you need further clarification!

[...]

PR #10376, opened 2024-11-23
feat: allow base button to be wrapped with tooltip

sbeaury:
## Problem

Solves #10223 

## Solution

Adds `React.forwardRef`so that MUI's tooltip can wrap RA custom button component 

## How To Test

This will now work

```tsx
<Tooltip title="This is a button">
  <Button>My Button</Button>
</Tooltip>
```

A `ButtonWithTooltip.tsx` component could be added

### DIFF
--- a/packages/ra-ui-materialui/src/button/Button.stories.tsx
+++ b/packages/ra-ui-materialui/src/button/Button.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { ReactNode } from 'react';
-import { createTheme, ThemeProvider, Stack } from '@mui/material';
+import { createTheme, ThemeProvider, Stack, Tooltip } from '@mui/material';
 import type { PaletteColor } from '@mui/material';
 import AddIcon from '@mui/icons-material/Add';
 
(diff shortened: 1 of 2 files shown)


fzaninotto:
I don't think this solves the linked issue, as other react-admin buttons don't forward the ref.

Also, can you add a story demonstrating this use case?

sbeaury:
> I don't think this solves the linked issue, as other react-admin buttons don't forward the ref.
> 
> Also, can you add a story demonstrating this use case?

It does, the issue was specifically for custom buttons (so any custom button deriving from the custom `Button` like `CreateButton`, `EditButton`, etc will inherit from this added `forwardRef`). Added the [example](https://github.com/marmelab/react-admin/pull/10376/commits/5f7698dca7b013824d389e2d80e1057c2e182c0a) in a story.

fzaninotto:
`CreateButton` and other high-level buttons don't use `forwardRef`, so even if the lower-level `Button` does use it, your change doesn't allow a `Tooltip` wrapping a `CreateButton`, which is the issue pointed in #10223.

PR #10349, opened 2024-11-13
[SimpleList] deprecate linkType, and implement rowClick as for the DataGrid

quentin-decre:
## Problem

The linkType for SimpleList provides less functionality than the rowClick for DataGrid. And also the inconsistency makes it harder to use.

## Solution

Implement rowClick prop as it works for Datagrid.
Keep the linkType prop working as previously, but show it as deprecated. If both are set, linkType takes precedence (because it will be simpler to remove from the code later that way).

## How To Test

Similar to data grid

## Additional Checks

- [ ] The PR targets `master` for a bugfix, or `next` for a feature
- [ ] The PR includes **unit tests** (if not possible, describe why)
- [X] The PR includes one or several **stories** (if not possible, describe why)
- [X] The **documentation** is up to date

PR #10385, opened 2024-11-27
Introduce SimpleList rowClick

djhi:
## Problem

Supersedes #10349

> The linkType for SimpleList provides less functionality than the rowClick for DataGrid. And also the inconsistency makes it harder to use.

## Solution

> Implement rowClick prop as it works for Datagrid.
Keep the linkType prop working as previously, but show it as deprecated. If both are set, linkType takes precedence (because it will be simpler to remove from the code later that way).

It also fixes an issue introduced by #10343 that produced nested `li` which is invalid HTML.

## How To Test

- `linkType`: https://react-admin-storybook-2x9vrs84o-marmelab.vercel.app/?path=%2Fstory%2Fra-ui-materialui-list-simplelist--link-type
- `rowClick`: https://react-admin-storybook-2x9vrs84o-marmelab.vercel.app/?path=%2Fstory%2Fra-ui-materialui-list-simplelist--row-click
- Fix for invalid html: https://react-admin-storybook-2x9vrs84o-marmelab.vercel.app/?path=%2Fstory%2Fra-ui-materialui-list-simplelist--icons-avatars-and-link-type. Open the devtool and highlight an item: it should not have nested `li`. Besides, [this error](https://github.com/marmelab/react-admin/actions/runs/11839487973/job/32991563043#step:5:96) should not appear anymore in the tests output.

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
+++ b/packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
@@ -1,11 +1,7 @@
-import * as React from 'react';
-import { styled } from '@mui/material/styles';
 import type { SxProps } from '@mui/material';
-import { isValidElement, ReactNode, ReactElement } from 'react';
 import {
     Avatar,
     List,
-    ListProps,
     ListItem,
     ListItemAvatar,
     ListItemButton,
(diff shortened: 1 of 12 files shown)


packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
```diff
@@ -272,47 +154,112 @@ export interface SimpleListProps<RecordType extends RaRecord = any>
     total?: number;
 }
 
-const LinkOrNot = (
-    props: LinkOrNotProps & Omit<ListItemProps, 'button' | 'component' | 'id'>
+const SimpleListItemContent = <RecordType extends RaRecord = any>(
```
djhi, line 157: I'd like to extract this to its own file but it uses `SimpleListClasses`

packages/ra-ui-materialui/src/list/SimpleList/SimpleListItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+                >
+                    {children}
+                </ListItemButton>
+            </ListItem>
+        );
+    }
+
+    return (
+        <ListItem
+            sx={{
```
djhi, line 108: No `disablePadding` here so that it looks the same as when using buttons

packages/ra-ui-materialui/src/list/SimpleList/SimpleList.tsx
djhi: To avoid recursive imports. Another option would be to extract the SimpleList styled root in its own file

slax57: Got it. Let's keep it that way for now.

packages/ra-ui-materialui/src/list/SimpleList/SimpleListItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const pathForRecord = useGetPathForRecord({
+        link: isFunctionLink ? false : linkType ?? rowClick,
+    });
+    const getPathForRecord = useGetPathForRecordCallback();
+    const handleClick = useEvent(
+        async (event: React.MouseEvent<HTMLAnchorElement>) => {
+            // No need to handle non function linkType or rowClick
+            if (!isFunctionLink) return;
+            if (!record) return;
+            event.persist();
```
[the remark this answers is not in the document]
djhi: TIL

slax57 (approved):

Issue #10382, opened 2024-11-25
Datagrids are conflicting when the resources use the same id


fxdave:
![Image](https://github.com/user-attachments/assets/4ab0c7dd-225b-40ef-9b5d-54640930551a)

**What you were expecting:**
first Datagrid is isolated from the second

**What happened instead:**
changing the first Datagrid's state changes the second Datagrid's state.

**Steps to reproduce:**
Have two Datagrid, both should have a row with the same id.

**Related code:**

```tsx
export const OrganizationShow = () => (
  <Show>
    <SimpleShowLayout>
      <TextField source="id" />
      <TextField source="name" />
      <NumberField source="customMaxAllowedProperties" />
      <ArrayField source="memberships">
        <Datagrid itemID="asd">
          <TextField source="id" />
          <ReferenceField source="userId" reference="users" />
          <TextField source="role" />
          <ReferenceField source="organizationId" reference="organizations" />
          <DateField source="createdAt" />
          <DateField source="updatedAt" />
        </Datagrid>
      </ArrayField>
      <ReferenceField reference="subscriptions" source="subscription.id" />
      <ArrayField source="portfolios">
        <Datagrid>
          <TextField source="id" />
          <TextField source="name" />
          <ReferenceField source="creatorId" reference="users" />
          <ReferenceField source="organizationId" reference="organizations" />
          <DateField source="createdAt" />
          <DateField source="updatedAt" />
        </Datagrid>
      </ArrayField>
    </SimpleShowLayout>
  </Show>
);
```

**Environment**

* React-admin version: 5.3.4
* React version: 18.3.1
* Browser: firefox 132.0.2-1

fzaninotto:
Thanks for your report. 

By default, the selection state is linked to the current resource. This is necessary so that an existing selection can be erased in a side effect, e.g. when deleting a record. 

Some list contexts can opt out of that shared global state (see for instance [`<ReferenceManyField storeKey>`](https://marmelab.com/react-admin/ReferenceManyField.html#storekey). We probably need to add support for the same `storeKey` to the `<ArrayField>` as well. A PR adding this support is welcome!

In the meantime, try passing a different `resource` for each `ArrayField`:

```diff
export const OrganizationShow = () => (
  <Show>
    <SimpleShowLayout>
      <TextField source="id" />
      <TextField source="name" />
      <NumberField source="customMaxAllowedProperties" />
-     <ArrayField source="memberships">
[...]

PR #10390, opened 2024-12-01
Chore(ArrayField): add support for `storeKey` to manage independent selection states

Aijeyomah:
## Problem
#10382

## Solution

- Introduced a `storeKey` property for `ArrayField` to enable distinct selection state storage for multiple instances of the same data source.
- Added documentation and examples to demonstrate usage.

https://www.loom.com/share/b6da1650760d4a83981f602341b667e7?sid=3fa2812e-1cdc-4266-9958-89c9ff731657

_Describe the solution this PR implements_

## How To Test

## Additional Checks

### DIFF
--- a/docs/ArrayField.md
+++ b/docs/ArrayField.md
@@ -71,12 +71,13 @@ const PostShow = () => (
 
 ## Props
 
-| Prop       | Required | Type              | Default | Description                              |
-|------------|----------|-------------------|---------|------------------------------------------|
-| `children` | Required | `ReactNode`       |         | The component to render the list.        |
-| `filter`   | Optional | `object`          |         | The filter to apply to the list.         |
-| `perPage`  | Optional | `number`          | 1000    | The number of items to display per page. |
-| `sort`     | Optional | `{ field, order}` |         | The sort to apply to the list.           |
+| Prop       | Required | Type              | Default | Description                                        |
+|------------|----------|-------------------|---------|----------------------------------------------------|
+| `children` | Required | `ReactNode`       |         | The component to render the list.                  |
+| `filter`   | Optional | `object`          |         | The filter to apply to the list.                   |
+| `perPage`  | Optional | `number`          | 1000    | The number of items to display per page.           |
+| `sort`     | Optional | `{ field, order}` |         | The sort to apply to the list.                     |
+| `storeKey` | Optional | `string`          |         | The key to use to store the records selection state|
 
 `<ArrayField>` accepts the [common field props](./Fields.md#common-field-props), except `emptyText` (use the child `empty` prop instead).
 
@@ -216,6 +217,35 @@ By default, `<ArrayField>` displays the items in the order they are stored in th
 </ArrayField>
 ```
 {% endraw %}
+
+## `storeKey`
+
+By default, `ArrayField` stores the selection state in localStorage so users can revisit the page and find the selection preserved. The key for storing this state is based on the resource name, formatted as `${resource}.selectedIds`.
+
+When displaying multiple lists with the same data source, you may need to distinguish their selection states. To achieve this, assign a unique `storeKey` to each `ArrayField`. This allows each list to maintain its own selection state independently.
+
+In the example below, two `ArrayField` components display the same data source (`books`), but each stores its selection state under a different key (`customOne.selectedIds` and `customTwo.selectedIds`). This ensures that both components can coexist on the same page without interfering with each other's state.
+
+```jsx
+<Stack direction="row" spacing={2}>
+    <ArrayField
+        source="books"
+        storeKey="customOne"
+    >
+        <Datagrid>
+            <TextField source="title" />
+        </Datagrid>
+    </ArrayField>
+    <ArrayField
+        source="books"
+        storeKey="customTwo"
+    >
+        <Datagrid>
+            <TextField source="title" />
+        </Datagrid>
+    </ArrayField>
+</Stack>
+```
 
 ## Using The List Context
 
--- a/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
@@ -14,6 +14,22 @@ import { Datagrid } from '../list';
 import { SimpleList } from '../list';
 import { ListContext, TwoArrayFieldsSelection } from './ArrayField.stories';
 
+beforeAll(() => {
+    jest.spyOn(console, 'error').mockImplementation((message, ...args) => {
+        if (
+            typeof message === 'string' &&
+            message.includes('React will try recreating this component tree')
+        ) {
+            return;
+        }
+        console.warn(message, ...args); // Still log other errors
+    });
+});
+
+afterAll(() => {
+    jest.restoreAllMocks();
+});
+
 describe('<ArrayField />', () => {
     const sort = { field: 'id', order: 'ASC' };
 
@@ -161,7 +177,6 @@ describe('<ArrayField />', () => {
 
     it('should not select the same id in both ArrayFields when selected in one', async () => {
         render(<TwoArrayFieldsSelection />);
-
         await waitFor(() => {
             expect(screen.queryAllByRole('checkbox').length).toBeGreaterThan(2);
         });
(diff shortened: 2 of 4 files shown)


Aijeyomah:
@fzaninotto Can you review? thnaks

djhi:
Thanks for the PR! As this is a new feature, can you target the `next` branch? Besides, in order to test it, we'll need stories and unit tests. Can you add some, taking inspiration from the `ListBase` component?

Aijeyomah:
> Thanks for the PR! As this is a new feature, can you target the `next` branch? Besides, in order to test it, we'll need stories and unit tests. Can you add some, taking inspiration from the `ListBase` component?

Yes, I can. Thanks

Aijeyomah:
> Thanks for the PR! As this is a new feature, can you target the `next` branch? Besides, in order to test it, we'll need stories and unit tests. Can you add some, taking inspiration from the `ListBase` component?

@djhi @fzaninotto I have updated the target branch as well as written the necessary stories and tests for the feature. Please re-review. Thanks

packages/ra-core/src/controller/list/useList.ts
```diff
@@ -91,12 +91,45 @@ export const useList = <RecordType extends RaRecord = any>(
...
+    }, [
+        resource,
+        setPage,
+        setPerPage,
+        initialPage,
+        initialPerPage,
+        page,
+        perPage,
+    ]);
+
```
slax57, line 132: I'm afraid you are going in the wrong direction here :grimacing:  . The fix for #10382 suggested by @fzaninotto was to allow to override the `storeKey` used to store the **selection state** (i.e. used in `useRecordSelection`), but not the **pagination** state.

You can have a look at what was done in `useReferenceManyFieldController` for an example:

https://github.com/marmelab/react-admin/blob/7b319c451d448fe9db00b76fa7ddcc10d4d6dedd/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts#L90-L92

packages/ra-core/src/controller/list/useList.storeKey.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+    <CoreAdminContext store={localStorageStore()} dataProvider={dataProvider}>
+        <CoreAdminUI layout={Layout}>
+            <CustomRoutes>
+                <Route path="/store" element={StorePosts} />
+                <Route path="/nostore" element={NoStorePosts} />
+            </CustomRoutes>
+            <Resource name="posts" />
+        </CoreAdminUI>
+    </CoreAdminContext>
+);
```
slax57, line 154: This Story is a bit misleading, `StorePosts` is still using a `storeKey`.

Besides, `storeKey={false}` is not the same as having no `storeKey` (for `NoStorePosts` ). I'd expect both lists in `ListsWithoutStoreKeys` to have no `storeKey`.

packages/ra-core/src/controller/list/useList.storeKey.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+            Go to Store List
+        </Link>
+        <Link aria-label="nostore" to={`/nostore`}>
+            Go to No-Store List
+        </Link>
+
+        <br />
+        {props.children}
+    </div>
+);
```
slax57, line 108: `Layout` should not be shared across stories since both stories do not implement the same routes

docs/ArrayField.md
```diff
@@ -217,6 +218,35 @@ By default, `<ArrayField>` displays the items in the order they are stored in th
 ```
 {% endraw %}
 
+## `storeKey`
+
+By default, `ArrayField` stores the selection state in localStorage so users can revisit the page and find the selection preserved. The key for storing this state is based on the resource name, formatted as `${resource}.selectedIds`.
+
+When displaying multiple lists with the same data source, you may need to distinguish their selection states. To achieve this, assign a unique `storeKey` to each `ArrayField`. This allows each list to maintain its own selection state independently.
+
+In the example below, two `ArrayField` components display the same data source (`books`), but each stores its selection state under a different key (`books.selectedIds` and `custom.selectedIds`). This ensures that both components can coexist on the same page without interfering with each other's state.
```
slax57, line 227:
```suggestion
In the example below, two `ArrayField` components display the same data source (`books`), but each stores its selection state under a different key (`customOne.selectedIds` and `customTwo.selectedIds`). This ensures that both components can coexist on the same page without interfering with each other's state.
```

slax57 (changes requested):
Too bad you took a wrong direction with the fix, but we appreciate the effort to provide good stories, tests, and documentation! Thanks!

Aijeyomah:
@slax57 Can you re-review? Thanks.

packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
```diff
@@ -12,7 +12,23 @@ import { NumberField } from './NumberField';
...
 import { Datagrid } from '../list';
 import { SimpleList } from '../list';
-import { ListContext } from './ArrayField.stories';
+import { ListContext, TwoArrayFieldsSelection } from './ArrayField.stories';
+
+beforeAll(() => {
+    jest.spyOn(console, 'error').mockImplementation((message, ...args) => {
+        if (
+            typeof message === 'string' &&
+            message.includes('React will try recreating this component tree')
```
fzaninotto, line 21: This isn't a normal error message - it probably comes from your story, your test, or your implementation.

packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
```diff
@@ -12,7 +12,23 @@ import { NumberField } from './NumberField';
 import { TextField } from './TextField';
 import { Datagrid } from '../list';
 import { SimpleList } from '../list';
-import { ListContext } from './ArrayField.stories';
+import { ListContext, TwoArrayFieldsSelection } from './ArrayField.stories';
+
+beforeAll(() => {
```
fzaninotto, line 17: IMHO, this is a footgun as it will hide potentially important errors. I'd prefer that you do the mock in individual tests and only when necessary

packages/ra-ui-materialui/src/field/ArrayField.stories.tsx
```diff
@@ -182,3 +182,70 @@ export const InShowLayout = () => (
...
+
+                            {/* Memberships ArrayField */}
+                            <ArrayField
+                                source="memberships"
+                                storeKey="organization_memberships"
+                            >
+                                <Datagrid
+                                    rowClick="toggleSelection"
+                                    bulkActionButtons={true}
+                                    isRowSelectable={() => true}
```
fzaninotto, line 224: Rows are selectable by default, you shouldn't need this line. Same for the second ArrayInput.

packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
```diff
@@ -158,4 +174,41 @@ describe('<ArrayField />', () => {
...
+
+    it('should not select the same id in both ArrayFields when selected in one', async () => {
+        render(<TwoArrayFieldsSelection />);
+        await waitFor(() => {
+            expect(screen.queryAllByRole('checkbox').length).toBeGreaterThan(2);
+        });
+
+        const checkboxes = screen.queryAllByRole('checkbox');
+
+        expect(checkboxes.length).toBeGreaterThan(3);
```
fzaninotto, line 186: By reading the test, I don't know why you wait for that. If you want to wait for the two lists to render, please use a `findByText` looking for the actual array values instead.

packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
```diff
@@ -158,4 +174,41 @@ describe('<ArrayField />', () => {
...
+
+        // Select an item in the memberships list
+        fireEvent.click(checkboxes[1]);
```
fzaninotto, line 189: Since you enabled rowClick, I suggest you trigger a click on a piece of text instead. This would make the test much more readable as it's not obvious what `checkboxes[1]` is.

packages/ra-ui-materialui/src/field/ArrayField.spec.tsx
```diff
@@ -158,4 +174,41 @@ describe('<ArrayField />', () => {
...
+        render(<TwoArrayFieldsSelection />);
```
fzaninotto, line 190: You shouldn't need to rerender. If you want to come back to the initial state, use more than one `it()` and group them in a `describe()`.

packages/ra-ui-materialui/src/field/ArrayField.tsx
```diff
@@ -79,9 +79,15 @@ const ArrayFieldImpl = <
 >(
     props: ArrayFieldProps<RecordType>
 ) => {
-    const { children, resource, perPage, sort, filter } = props;
+    const { children, resource, perPage, sort, filter, storeKey } = props;
     const data = useFieldValue(props) || emptyArray;
-    const listContext = useList({ data, resource, perPage, sort, filter });
+    const listContext = useList({
+        data,
+        resource: storeKey || resource, // Prioritize storeKey if provided
```
fzaninotto, line 86: I'd prefer that you add support for a custom `storeKey` in `useList`. I know that we currently use `resource` only for selection in `useList`, but this may change in the future.

fzaninotto (changes requested):

PR #10392, opened 2024-12-03
Fix `useLogin` should invalidate the `getPermissions` cache

slax57:
## Problem

Consider an Admin that has some pages that allow anonymous access, and also a `authProvider.getPermissions()` implementation, that, say, returns `'user'` when the user is logged in, and `undefined` when the user is not logged in.

If the user first accesses the anonymous parts of the Admin, the cached permission will be `undefined`. But if he then logs in, without leaving the Admin, then the cache is not updated, and the permissions still return `undefined` instead of `'user'`.

## Solution

Invalidate the `getPermissions` cache after a successful login, to force a refetch of the permissions.

## How To Test

https://react-admin-storybook-31pnpeyju-marmelab.vercel.app/?path=/story/ra-core-auth-uselogin--permissions-state

## Additional Checks

- [x] The **documentation** is up to date -> no need, bugfix

### DIFF
--- a/packages/ra-core/src/auth/useLogin.stories.tsx
+++ b/packages/ra-core/src/auth/useLogin.stories.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import usePermissions, { UsePermissionsResult } from './usePermissions';
+import usePermissions from './usePermissions';
 import { CoreAdminContext, TestMemoryRouter, useLogin, useLogout } from '..';
 
 export default {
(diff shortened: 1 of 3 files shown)


packages/ra-core/src/auth/useLogin.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+const UsePermissions = ({
+    children,
+}: {
+    children: (state: UsePermissionsResult<any, Error>) => React.ReactNode;
+}) => {
+    const permissionQueryParams = {
+        retry: false,
+    };
+    const res = usePermissions({}, permissionQueryParams);
+    return children(res);
```
djhi, line 48: Why use a child function here? You could just return the JSX you have in `inspectPermissionsState`, couldn't you?

slax57: Didn't give it too much though TBH, I just copied that code from the usePermissions stories

djhi (approved):

Issue #7763, opened 2022-05-27
Setting disableClearable prop on AutocompleteArrayInput gives build error


lvernaillen:
**What you were expecting:**
Setting the `disableClearable` prop on `AutocompleteArrayInput` should be possible without any build error. It removes the cross icon and removes the ability to clear the autocomplete input.

**What happened instead:**
Build error occurs stating Type 'true' is not assignable to type 'false'.
![image](https://user-images.githubusercontent.com/4180374/170704370-c4b24a72-06f3-4677-b9aa-26bd59dea821.png)

**Steps to reproduce:**
Use your `simple` application sandbox.
https://codesandbox.io/s/github/marmelab/react-admin/tree/master/examples/simple?file=/src/posts/TagReferenceInput.tsx
Update `src/posts/TagReferenceInput.tsx` by adding the `disableClearable` as shown below and see the compiler warning when hovering over the red squiggles.
```javascript 
<ReferenceArrayInput {...props} perPage={5} filter={{ published }}>
    <AutocompleteArrayInput
        create={<CreateTag />}
        optionText={`name.${locale}`}
        disableClearable  // ADD THIS PROP -> Type 'true' is not assignable to type 'false'
    />
</ReferenceArrayInput>
```
**Other information:**
Setting the prop `disableClearable` does work on a regular `AutocompleteInput` component.
This can be checked by adding the `disableClearable` prop to `AutocompleteInput` in `src/posts/PostEdit.tsx`.

A workaround is to force any type as done below. However that shouldn't be needed. Simply setting `disableClearable` should work.

```javascript 
<ReferenceArrayInput {...props} perPage={5} filter={{ published }}>
    <AutocompleteArrayInput
        create={<CreateTag />}
        optionText={`name.${locale}`}
        disableClearable={true as any}  // ADD THIS PROP -> this workaround works
    />
</ReferenceArrayInput>
```

**Environment**

* React-admin version: 4.1.0
* Last version that did not exhibit the issue (if applicable):
* React version: 17
* Browser: Chrome version 101.0.4951.67

djhi:
Reproduced. Thanks!

ethembynkr:
Same error exists in the AutocompleteInput component.

React-admin version: 4.16.19

fzaninotto:
@ethembynkr Can you please open a new issue?

PR #10393, opened 2024-12-03
Fix type of disableClearable prop for AutocompleteInput

smeng9:
Follow up of https://github.com/marmelab/react-admin/issues/7763#issuecomment-2232688294


djhi (approved):

djhi:
Thanks!

PR #10399, opened 2024-12-05
[Doc] Fix `<DateRangeInput>` import syntax

erwanMarmelab:
Fix the import of `DateRangeInput` in the OSS doc


djhi (changes requested):
Might be worth mentioning in the introduction that you might have to install the MUI package yourself

fzaninotto (approved):

PR #10418, opened 2024-12-16
`useUpdate`: allow to disable cache update from `meta`

slax57:
## Problem

After updating a record with `useUpdate`, regardless of the mutation mode or the mutation options, the cached data for the loaded queries (`getOne`, `getList`, ...) will be updated. This can cause unwanted refreshes of the UI, and can be a problem for instance in the context of a form, where we want to be able to save the current values without resetting the form (and losing newer changes) once the dataProvider resolves.

## Solution

Provide a way to disable the cache update.

This PR adds support for a `meta` prop named `disableCacheUpdate` allowing to do just that.

```tsx
const update = useUpdate();
update('foo', {
    id: 1,
    data: { bar: 'baz' },
    previousData: { id: 1, bar: 'bar' },
    meta: { disableCacheUpdate: true }, // cache won't be updated
});
```

## Additional fix

Also, this PR fixes a bug where the `meta` param was not used to construct the query keys when updating the cache, if the `meta` param was passed at call time, and the mutation mode was set to pessimistic.

## How To Test

Run the unit tests.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/controller/create/useCreateController.ts
+++ b/packages/ra-core/src/controller/create/useCreateController.ts
@@ -51,7 +51,11 @@ export const useCreateController = <
         MutationOptionsError,
         ResultRecordType
     > = {}
-): CreateControllerResult<RecordType> => {
+): CreateControllerResult<
+    RecordType,
+    MutationOptionsError,
+    ResultRecordType
+> => {
     const {
         disableAuthentication,
         record,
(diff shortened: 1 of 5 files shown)


djhi (commented):
Shouldn't we filter out `disableCacheUpdate` from the meta send to the dataProvider?

slax57:
This features turns out to introduce undesired side effects in the optimistic cache management (leading to situations where the UI is no longer updated at all), plus it does not allow to solve the original issue we had with the resetting form.
Hence I'll close it.

The only part we can salvage is the fix about the query key bug with `meta`, which is fixed in this PR: https://github.com/marmelab/react-admin/pull/10422

Issue #10427, opened 2024-12-20
AutoComplete suggestion list cover the popper dialog


a-pouramini:
I followed example code in this link:

https://marmelab.com/react-admin/AutocompleteInput.html#create

```
.....
  return (
    <ReferenceInput source={source} reference={reference} {...props}>
      <AutocompleteInput
        create={
          canCreate && (
            <CreateNewItemDialog />
          )
        }
      />
    </ReferenceInput>
  );
};

export default CustomReferenceArrayInput;

```
But when the dialog is poped up the suggestion list opens up and covers it.

slax57:
Can you please follow the issue template?

a-pouramini:
if you reporoduce the code suggested in your documentation, you may encounter this problem. However, I resolved it by applying z-index=1500 to the dialog. the z-index of suggestion list is 1300 and can cover the dialog if it's not set to higher.

PR #10442, opened 2025-01-10
[Doc] Explain `openOnFocus` in `<AutocompleteInput>` with StrictMode

djhi:
## Problem

With an `<Autocomplete>` which has a create component opening a Dialog, the list of choices reopens after clicking Create, which hides the Dialog.`

Fix https://github.com/marmelab/react-admin/issues/10427

## Solution

- Document the issue
- Fix the simple example

## How To Test

- Open the simple example in dev mode
- Edit a post and go to the *Miscellaneous* tab
- Create a new tag

=> No issue as we now have `openOnFocus` set to `false`

## Additional Checks

### DIFF
--- a/docs/AutocompleteArrayInput.md
+++ b/docs/AutocompleteArrayInput.md
@@ -212,6 +212,8 @@ const CreateRole = () => {
 ```
 {% endraw %}
 
+**Tip**: In development with `React.StrictMode`, you may run into an issue where the `<AutocompleteInput>` menu reopens after clicking the create item when the [`openOnFocus`](https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-openOnFocus) prop is set to `true` which is the default with React-admin. If that bothers you, set the `openOnFocus` prop to `false`.
+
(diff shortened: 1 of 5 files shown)


docs/AutocompleteArrayInput.md
```diff
@@ -212,6 +212,8 @@ const CreateRole = () => {
 ```
 {% endraw %}
 
+**Tip**: In development with `React.StrictMode`, you may run into an issue where the `<AutocompleteInput>` menu reopens after clicking the create item when the [`openOnFocus`](https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-openOnFocus) prop is set to `true` which is the default with React-admin. If that bothers you, set the `openOnFocus` prop to `false`.
```
slax57, line 215:
```suggestion
**Tip**: In development with `React.StrictMode`, you may run into an issue where the `<AutocompleteArrayInput>` menu reopens after clicking the create item when the [`openOnFocus`](https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-openOnFocus) prop is set to `true` which is the default with React-admin. If that bothers you, set the `openOnFocus` prop to `false`.
```

PR #10443, opened 2025-01-11
fix <Authenticated> - never render its children when checkAuth return…

adrien-may:
…s error

## Problem

Using <Authenticated>, we can witness a quick render with the children provided whereas it shouldn't. This happens because react-query returns an object with isPending=False but the callback for error has not been called yet.

## Solution

Checking for isError and isPending seems to be more reliable

## How To Test

I added a test to demonstrate the problem

## Additional Checks


fzaninotto (approved):

fzaninotto:
Thanks a lot for the well-written PR and the patch!

PR #10444, opened 2025-01-13
Allow graphql dataProviders to leverage the introspection results

djhi:
## Problem

See #9403
A GraphQL backend may support functionality that extends beyond the default Data Provider methods. One such example of this would be implementing GraphQL subscriptions and integrating [ra-realtime](https://marmelab.com/ra-enterprise/modules/ra-realtime) to power realtime updates in your React-Admin application. However, graphql dataFrovider implementations can't have access to the introspection made by `ra-data-graphql` and may have to reimplement the same logic.

## Solution

- Update `ra-data-graphql` so that the dataProvider has a `getIntrospection` method allowing custom implementations to leverage the introspection.
- Update `ra-data-graphql` so that the dataProvider has a `client` property allowing custom implementations to access the Apollo client easily whether it was provided or created by `ra-data-graphql`.

## Additional Checks

### DIFF
--- a/packages/ra-data-graphql/README.md
+++ b/packages/ra-data-graphql/README.md
@@ -199,6 +199,116 @@ The `./schema` file is a `schema.json` in `./src` retrieved with [get-graphql-sc
 
 > Note: Importing the `schema.json` file will significantly increase the bundle size.
 
+## Leveraging Introspection In Custom Methods
+
+If you need to build custom methods based on the introspection, you can leverage the `getIntrospection` method of the `dataProvider`. It returns an object with the following format:
+
+```js
+{
+    // The original schema as returned by the Apollo client
+    schema: {},
+    // An array of object describing the types that are compatible with react-admin resources
+    // and the methods they support. Note that not all methods may be supported.
+    resources: [
+        {
+            type: { name: 'name-of-the-type' }, // e.g. Post
+            GET_LIST: { name: 'name-of-the-query' }, // e.g. allPosts
+            GET_MANY: { name: 'name-of-the-query' }, // e.g. allPosts
+            GET_MANY_REFERENCE: { name: 'name-of-the-query' }, // e.g. allPosts
+            GET_ONE: { name: 'name-of-the-query' }, // e.g. Post
+            CREATE: { name: 'name-of-the-query' }, // e.g. createPost
+            UPDATE: { name: 'name-of-the-query' }, // e.g. updatePost
+            DELETE: { name: 'name-of-the-query' }, // e.g. deletePost
+        },
+    ],
+}
+```
+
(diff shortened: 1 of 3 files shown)


packages/ra-data-graphql/README.md
```diff
@@ -199,6 +199,116 @@ The `./schema` file is a `schema.json` in `./src` retrieved with [get-graphql-sc
...
+        //    2. resource/${resource}/${id}
+        const [, resourceName, id] = raRealTimeTopic;
+        const introspectionResults = await baseDataProvider.getIntrospection();
+        const resourceIntrospection = introspectionResults.resources.find(
+            resource => resource.type.name === resourceName
+        );
+        if (!resourceIntrospection) throw new Error(`Invalid resource ${resourceName}`);
+
+        const { query, queryName, variables } = buildQuery({ id, resource, resourceIntrospection });
+        const subscription = dataProvider.client
```
slax57, line 259:
```suggestion
        const subscription = baseDataProvider.client
```

?

packages/ra-data-graphql/README.md
```diff
@@ -199,6 +199,116 @@ The `./schema` file is a `schema.json` in `./src` retrieved with [get-graphql-sc
...
+        });
+
+        return Promise.resolve({ data: null });
+    },
+    unsubscribe: async (topic: string, subscriptionCallback: any) => {
+        subscriptions = subscriptions.filter(
+            subscription =>
+                subscription.topic !== topic ||
+                subscription.subscriptionCallback !== subscriptionCallback
+        );
```
slax57, line 278: **nitpick** (not really important for this example, but would improve completeness of the example): I believe you're supposed to call `subscription.unsubscribe()` on the matching subscription. Filtering out the subscription from the array is not enough.
Source: https://stackoverflow.com/questions/51477002/unsubscribe-subscription-in-apollo-client

packages/ra-data-graphql/src/index.ts
```diff
@@ -261,4 +274,9 @@ const getQueryOperation = query => {
     throw new Error('Unable to determine the query operation');
 };
 
+export type GetIntrospection = () => Promise<IntrospectionResult>;
+export type GraphqlDataProvider = DataProvider & {
+    getIntrospection: GetIntrospection;
+};
```
slax57, line 280: Shouldn't you also add `client` to the type?

packages/ra-data-graphql/README.md
```diff
@@ -199,6 +199,116 @@ The `./schema` file is a `schema.json` in `./src` retrieved with [get-graphql-sc
...
+import { RealTimeDataProvider } from '@react-admin/ra-realtime';
+import { IntrospectedResource } from 'ra-data-graphql';
+
+const subscriptions: {
+    topic: string;
+    subscription: any;
+    subscriptionCallback: any;
+}[];
+
+const baseDataProvider = buildDataProvider(/* */);
```
slax57, line 240: `buildDataProvider` should be imported

slax57 (changes requested):

slax57 (approved):

PR #10445, opened 2025-01-14
[TypeScript] Allow providing error type in dataProvider and controllers hooks

djhi:
## Problem

Many of our hooks do not allow to provide the `Error` type.

## Solution

Allow providing error type in dataProvider and controllers hooks

## Additional Checks

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -44,12 +44,14 @@ import { useResourceContext } from '../../core';
 export const useReferenceManyFieldController = <
     RecordType extends RaRecord = RaRecord,
     ReferenceRecordType extends RaRecord = RaRecord,
+    ErrorType = Error,
 >(
     props: UseReferenceManyFieldControllerParams<
         RecordType,
-        ReferenceRecordType
+        ReferenceRecordType,
+        ErrorType
     >
-): ListControllerResult<ReferenceRecordType> => {
+): ListControllerResult<ReferenceRecordType, ErrorType> => {
     const {
         debounce = 500,
         reference,
@@ -168,7 +170,7 @@ export const useReferenceManyFieldController = <
         isLoading,
         isPending,
         refetch,
-    } = useGetManyReference<ReferenceRecordType>(
+    } = useGetManyReference<ReferenceRecordType, ErrorType>(
         reference,
         {
             target,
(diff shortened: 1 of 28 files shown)


packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -180,20 +182,22 @@ export const useReferenceManyFieldController = <
         },
         {
             enabled: get(record, source) != null,
+            // @ts-expect-error FIXME The error type is not correctly inferred
```
slax57, line 185: That's because you forgot to use `ErrorType` at line 67

packages/ra-core/src/controller/list/ListController.tsx
```diff
@@ -17,12 +18,17 @@ import {
...
+>({
     children,
     ...props
 }: {
-    children: (params: ListControllerResult) => JSX.Element;
-} & ListControllerProps) => {
-    const controllerProps = useListController(props);
+    children: (
+        params: ListControllerResult<RecordType, ErrorType>
+    ) => JSX.Element;
```
slax57, line 30:
```suggestion
    ) => ReactNode;
```

?

packages/ra-core/src/controller/show/ShowController.tsx
```diff
@@ -17,12 +18,17 @@ import {
...
+>({
     children,
     ...props
 }: {
-    children: (params: ShowControllerResult) => JSX.Element;
-} & ShowControllerProps) => {
-    const controllerProps = useShowController(props);
+    children: (
+        params: ShowControllerResult<RecordType, ErrorType>
+    ) => JSX.Element;
```
slax57, line 30:
```suggestion
    ) => ReactNode;
```

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -111,14 +111,14 @@ export const ReferenceFieldView = <
 
     if (error) {
         return (
-            /* eslint-disable jsx-a11y/role-supports-aria-props */
```
slax57, line 114: eslint still complains about this line

slax57 (changes requested):

packages/ra-ui-materialui/src/field/ReferenceField.tsx
djhi: Fixed

slax57 (approved):

PR #10446, opened 2025-01-14
Fix useDelete does not invalidate cache in pessimistic mode

fzaninotto:
## Problem

Whe using pessimistic mode, a call to the delete callback of `useDelete` does not invalidate active queries. 

As a consequence, lists aren't refreshed. This creates buggy behaior in some cases.

For instance, in a list where each row has a DeleteButton in pessimistic mode, clicking on a delete button properly removes the deleted row, but does not update the list, which therefore has less rows than the `perPage`.

https://github.com/user-attachments/assets/33c748ae-7b8c-4843-a1b8-7176e1997005

## Solution

Related queries (getList, getMany, etc) are only invalidated on success on optimistic and undoable mode. They should be invalidated in all modes. 

https://github.com/user-attachments/assets/73766a3a-e18b-4177-b138-1a789245b99a

## How To Test

http://localhost:9010/?path=/story/ra-ui-materialui-button-deletebutton--in-list&args=mutationMode:pessimistic

## Additional Checks

### DIFF
--- a/packages/ra-core/src/dataProvider/useDelete.ts
+++ b/packages/ra-core/src/dataProvider/useDelete.ts
@@ -256,6 +256,16 @@ export const useDelete = <
             context: unknown
         ) => {
             if (mode.current === 'pessimistic') {
+                // update the getOne and getList query cache with the new result
+                const {
+                    resource: callTimeResource = resource,
+                    id: callTimeId = id,
+                } = variables;
+                updateCache({
+                    resource: callTimeResource,
+                    id: callTimeId,
+                });
+
                 if (
                     mutationOptions.onSuccess &&
                     !hasCallTimeOnSuccess.current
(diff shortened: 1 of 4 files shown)


packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -330,13 +318,6 @@ export const useDelete = <
             mode.current = mutationMode;
         }
 
-        if (mode.current === 'pessimistic') {
```
fzaninotto, line 333: moved down to allow snapshot (defined L350) to be accessible in pessimistic mode.

packages/ra-core/src/dataProvider/useDelete.ts
```diff
@@ -256,16 +256,6 @@ export const useDelete = <
...
-                // update the getOne and getList query cache with the new result
-                const {
-                    resource: callTimeResource = resource,
-                    id: callTimeId = id,
-                } = variables;
-                updateCache({
-                    resource: callTimeResource,
-                    id: callTimeId,
-                });
-
```
fzaninotto, line 268: In pessimistic mode, we don't populate the cache. We invalidate it.

djhi (approved):

PR #10453, opened 2025-01-17
Add vitest tutorial

smeng9:
Replaces https://github.com/marmelab/react-admin/pull/9332 since we no longer have issues in !9332

In Vitest 3, the browser mode is finally added. We don't need to rely on jsdom and module resolution for server. Everything will be tested and run exactly the same in a browser.

### DIFF
--- a/docs/Vite.md
+++ b/docs/Vite.md
@@ -147,6 +147,43 @@ const App = () => (
 export default App;
 ```
 
+## Unit Test Your App with Vitest
+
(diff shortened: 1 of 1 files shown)


djhi:
Vitest 3 is not released yet though, is it?

djhi (approved):

fzaninotto (approved):

fzaninotto:
Thanks!

PR #10464, opened 2025-01-23
Fix demo has multiple conflicting MUI packages

djhi:
## Problem

Following release 5.5, the demo shows larger padding than usual. This is because multiple conflicting versions of MUI packages are loaded. This only happens in the monorepo though.

## Solution

Use Vite aliases to force the resolution of MUI packages to the ones installed in the demo project.

## How To Test

- Run the demo
- Notice the padding is back to normal

## Additional Checks

### DIFF
--- a/examples/demo/vite.config.ts
+++ b/examples/demo/vite.config.ts
@@ -68,6 +68,21 @@ export default defineConfig(async () => {
                 //     find: 'scheduler/tracing',
                 //     replacement: 'scheduler/tracing-profiling',
                 // },
+                // The 2 next aliases are needed to avoid having multiple MUI instances
+                {
+                    find: '@mui/material',
+                    replacement: path.resolve(
+                        __dirname,
+                        'node_modules/@mui/material'
+                    ),
+                },
+                {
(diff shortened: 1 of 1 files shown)


examples/demo/vite.config.ts
```diff
@@ -68,6 +68,21 @@ export default defineConfig(async () => {
...
+                    ),
+                },
+                {
+                    find: '@mui/icons-material',
+                    replacement: path.resolve(
+                        __dirname,
+                        'node_modules/@mui/icons-material'
+                    ),
+                },
                 // we need to manually follow the symlinks for local packages to allow deep HMR
```
fzaninotto, line 86: How about the core MUI packages, which are indirect dependencies ? And how about other major dependencies overridden in the demo `package.json` (react and react-router)?

djhi: The core MUI packages are resolved from the MUI packages that require them. In our case, the packages from the demo node_modules. I haven't seen multiple react-router packages loaded (nor in vite, nor in the node_modules)

fzaninotto (approved):

PR #10469, opened 2025-01-24
Fix TabbedForm and TabbedShowLayout with react-router v7

### LINKED ISSUE #10458: TabbedForm does not work
I've got weird bug TabbedForm just stopped working shows only first tab, on first click to other tab system highlight it but do not redirect with any other click on tab highlighting disappears, but still only first tab rendered 

Image
 on each click, it concatenates URL 

Image

syncWithLocation={false} do not resolve the issue
reproducing on all browsers
```
Here is my dependencies 
"dependencies": {
    "@emotion/react": "^11.13.3",
    "@emotion/styled": "^11.13.0",
    "@mui/icons-material": "^5.16.7",
    "@mui/material": "^5.16.7",
    "@mui/x-date-pickers": "^7.18.0",
    "@tailwindcss/forms": "^0.5.7",
    "qs": "^6.13.0",
    "query-string": "^9.1.0",
    "ra-data-json-server": "^5.1.0",
    "ra-i18n-polyglot": "^5.4.2",
    "react": "^18.3.0",
    "react-admin": "^5.0.5",
    "react-dom": "^18.3.0",
    "react-multi-date-picker": "^4.5.2",
    "react-social-login-buttons": "^4.1.0",
    "reactjs-social-login": "^2.6.3"
  },
```

slax57:
## Problem

Fix #10458

React-router v7 reintroduced the change they introduced in 6.19.0, which was reverted in 6.20.1 due to it causing [issues](https://github.com/remix-run/react-router/issues/11052) in many apps, including react-admin.

## Solution

Introduce `useSplatPathBase` hook, which allows to compute the splat route's base path, allowing react-admin to create absolute paths for tabs (as opposed to relative paths previously), which will work both with react-router v6 and v7.

## How To Test

Wrap the `<Admin>` in the simple example with a `<HashRouter>`.
Then, you can set `<HashRouter future={{ v7_relativeSplatPath: true }}>` to test with the feature enabled (which is the default in v7), or set it to `false` to disable the feature.

Then, use `PostEdit` and `PostShow` to confirm the tabs are working OK.

## Additional Checks

- [ ] The PR includes **unit tests** (if not possible, describe why) -> can't unit test in v7
- [ ] The PR includes one or several **stories** (if not possible, describe why) -> no need for more stories

### DIFF
--- a/packages/ra-ui-materialui/src/detail/TabbedShowLayout.stories.tsx
+++ b/packages/ra-ui-materialui/src/detail/TabbedShowLayout.stories.tsx
@@ -1,19 +1,23 @@
 import * as React from 'react';
 import { Divider as MuiDivider } from '@mui/material';
 import {
-    RecordContextProvider,
-    ResourceContext,
     useRecordContext,
     WithRecord,
     TestMemoryRouter,
+    RaRecord,
+    testDataProvider,
+    ResourceContextProvider,
 } from 'ra-core';
 import { Labeled } from '../Labeled';
 import { TextField, NumberField } from '../field';
 import { TabbedShowLayout } from './TabbedShowLayout';
+import { AdminContext } from '../AdminContext';
+import { Route, Routes } from 'react-router';
+import { Show } from './Show';
 
 export default { title: 'ra-ui-materialui/detail/TabbedShowLayout' };
 
-const record = {
+const data = {
     id: 1,
     title: 'War and Peace',
     author: 'Leo Tolstoy',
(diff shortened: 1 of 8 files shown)


packages/ra-ui-materialui/src/form/TabbedFormView.tsx
```diff
@@ -35,9 +29,9 @@ export const TabbedFormView = (props: TabbedFormViewProps): ReactElement => {
         ...rest
     } = props;
     const location = useLocation();
-    const resolvedPath = useResolvedPath('');
```
slax57, line 38: `useResolvedPath` can no longer be used with both v6 and v7. See https://reactrouter.com/en/6.28.2/hooks/use-resolved-path

slax57:
There seems to be a bug with encoded paths... Investigating

slax57:
All issues fixed, back to RFR

packages/ra-ui-materialui/src/form/TabbedForm.spec.tsx
```diff
@@ -46,7 +46,8 @@ describe('<TabbedForm />', () => {
 
         const tabs = await screen.findAllByRole('tab');
         expect(tabs.length).toEqual(2);
-        await screen.findByLabelText('Title');
+        const titleInput = await screen.findByLabelText('Title');
+        expect(titleInput).toBeVisible();
```
slax57, line 50: Had to add this extra check because the current test did not detect the regression :grimacing:

packages/ra-ui-materialui/src/detail/TabbedShowLayout.stories.tsx
```diff
@@ -22,47 +26,73 @@ const record = {
...
+                        element={<Show sx={{ width: 600 }}>{children}</Show>}
+                    />
+                </Routes>
+            </ResourceContextProvider>
+        </AdminContext>
     </TestMemoryRouter>
 );
 
+export const Basic = () => (
+    <Wrapper>
```
slax57, line 64: I had to add a Wrapper to all stories to render the tabs in a nested route, otherwise the new implementation does not work.

djhi (approved):

PR #10472, opened 2025-01-24
Fix create-react-admin Vite alias and test

slax57:


packages/create-react-admin/src/generateAppTestFile.ts
```diff
@@ -33,7 +36,9 @@ test("should pass", async () => {
 
 	// Open the first post
 	fireEvent.click(await screen.findByText("Post 1"));
-	fireEvent.click(await screen.findByText("Edit"));
+  fireEvent.click(
```
djhi, line 39: same

djhi (approved):

PR #10473, opened 2025-01-24
[DNM] debug create-react-admin on CI

slax57:


slax57:
Tests are fixed, no more need for this PR

PR #10476, opened 2025-01-27
Update AuthRBAC.md

anthonycmain:
Updated to reflect that it is now part of core framework no longer just Enterprise

## Problem

RBAC is now part of the core framework

## Solution

Updated documentation to reflect this


fzaninotto:
Thanks fro your feedback. But the RBAC features are still part of the Enterprise Edition. What's open-source is  the Access Control part. 

Is anything misleading in the current doc?

PR #10479, opened 2025-01-27
Use Vitest browser to test react-admin

djhi:
## Problem

- Jest is slow
- Jest is unreliable (bundling issues, performances, etc.)
- JSDom has many differences from a real browser
- JSDom is incomplete
- We use vite already

## Solution

Use Vitest browser to test react-admin

- [x] Replace jest calls with vitest equivalents
- [ ] Replace testing-library calls with vitest equivalents

## How To Test

Run the usual test commands

## Additional Checks


fzaninotto:
We'll first migrate to vitest with js-dom, which should be minimal. then, and only then, we may use the browser mode. As a consequence, this PR is obsolete.

PR #10480, opened 2025-01-28
Make the CI fail on `console.error` in tests

erwanMarmelab:
## How To Test

Add a `console.error('foo', 'bar')` in a test and run it.

## Screenshot

image

### DIFF
--- a/test-setup.js
+++ b/test-setup.js
@@ -7,9 +7,14 @@ import '@testing-library/jest-dom/jest-globals';
 
 // Make the CI fail if console.error in tests
 let error = console.error;
-console.error = message => {
-    error.apply(message); // keep default behaviour
-    throw message;
+console.error = (...args) => {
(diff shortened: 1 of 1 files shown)


test-setup.js
```diff
@@ -5,6 +5,13 @@
 import '@testing-library/jest-dom';
 import '@testing-library/jest-dom/jest-globals';
 
+// Make the CI fail if console.error in tests
+let error = console.error;
+console.error = message => {
+    error.apply(message); // keep default behaviour
+    throw message;
```
slax57, line 12: **improvement**: you could log a line (or add this to the error you throw) indicating that the tests have failed because `console.error` was called. This will make debugging easier for future us.

slax57 (changes requested):

slax57 (approved):

djhi (approved):

Issue #10454, opened 2025-01-20
Add emptyText property to SelectArrayInput


MichaelTamm:
It would be nice to allow to specify what should be rendered inside a `SelectArrayInput` when no choice is selected. This feature is already present in the `SelectInput`.

For example:
```
const channelChoices = [
    {id: "email", name: "Email"},
    {id: "push", name: "Push Notification"},
];
<SelectArrayInput source="channels" choices={channelChoices} emptyText={<i>All Channels</i>} />
```

djhi:
Indeed! Can you open a PR?

Tomas025:
Hello, I'm a software engineering student I wanted to make my first contribution to an open source repository, can I act on this issue?

djhi:
Hi @Tomas025, of course! Be aware that you'll need to:
- target the `next` branch as this is a new feature
- add a story for this new prop in https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
- add tests in https://github.com/marmelab/react-admin/blob/master/packages/ra-ui-materialui/src/input/SelectArrayInput.spec.tsx using the story above
- add documentation for this new prop in https://github.com/marmelab/react-admin/blob/master/docs/SelectArrayInput.md

Tomas025:
Good afternoon, I uploaded a PR adding the empty Text, could you check please
#10504

nikghost17:
Hello is the issue still open ?
if Yes would like to work on it

kwitykwity:
Hello.... is this issue resolved or still open? If it's still open, I'd like to work on it.

fzaninotto:
this issue is still open but there is already a PR opened about it. If you want to contribute, please discuss on #11327

PR #10504, opened 2025-02-08
Added emptyText to SelectArrayInput component

Tomas025:
# Problem

Absence of the emptyText property in the SelectArrayInput component #10454 

## Solution

The empty Text property was added to the component, the emptyText option clears all fields that had been selected

## How To Test

It is possible to view the update in the storybook

## Additional Checks

- [ ✓ ] The PR targets `master` for a bugfix, or `next` for a feature
- [ ✓ ] The PR includes **unit tests** (if not possible, describe why)
- [ ✓ ] The PR includes one or several **stories** (if not possible, describe why)
- [ ✓ ] The **documentation** is up to date

### DIFF
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
@@ -261,7 +261,7 @@ export const EmptyText = () => (
                 { id: 'u002', name: 'Moderator' },
                 { id: 'u003', name: 'Reviewer' },
             ]}
-            emptyText={'Clear'}
+            emptyText='Clear'
             sx={{ width: 300 }}
         />
     </Wrapper>
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
@@ -251,6 +251,22 @@ export const DefaultValue = () => (
     </Wrapper>
 );
 
+export const EmptyText = () => (
+    <Wrapper>
+        <SelectArrayInput
+            source="roles"
+            choices={[
+                { id: 'admin', name: 'Admin' },
+                { id: 'u001', name: 'Editor' },
+                { id: 'u002', name: 'Moderator' },
+                { id: 'u003', name: 'Reviewer' },
+            ]}
+            emptyText={'Clear'}
+            sx={{ width: 300 }}
+        />
+    </Wrapper>
+);
+
 export const InsideArrayInput = () => (
     <Wrapper>
         <ArrayInput
--- a/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
+++ b/packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
@@ -230,14 +230,6 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
             ? [...(allChoices || []), createItem]
             : allChoices || [];
 
-    const renderEmptyItemOption = useCallback(() => {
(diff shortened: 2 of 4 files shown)


packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -351,6 +372,18 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
                     value={finalValue}
                     {...outlinedInputProps}
                 >
+                    {!isRequired && (
+                        <MenuItem
+                            value={''}
```
goharattiq, line 377:
```suggestion
                            value=''
```

packages/ra-ui-materialui/src/input/SelectArrayInput.stories.tsx
```diff
@@ -251,6 +251,22 @@ export const DefaultValue = () => (
...
+    <Wrapper>
+        <SelectArrayInput
+            source="roles"
+            choices={[
+                { id: 'admin', name: 'Admin' },
+                { id: 'u001', name: 'Editor' },
+                { id: 'u002', name: 'Moderator' },
+                { id: 'u003', name: 'Reviewer' },
+            ]}
+            emptyText={'Clear'}
```
goharattiq, line 264:
```suggestion
            emptyText='Clear'
```

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -217,6 +230,14 @@ export const SelectArrayInput = (props: SelectArrayInputProps) => {
             ? [...(allChoices || []), createItem]
             : allChoices || [];
 
+    const renderEmptyItemOption = useCallback(() => {
+        return typeof emptyText === 'string'
+            ? emptyText === ''
+                ? ' ' // em space, forces the display of an empty line of normal height
+                : translate(emptyText, { _: emptyText })
+            : emptyText;
+    }, [emptyText, translate]);
```
djhi, line 239: You don't need to wrap this in a `useCallback` here. I would even put the function body directly inside the `MenuItem` at L384

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
```diff
@@ -373,6 +406,7 @@ export type SelectArrayInputProps = ChoicesProps &
     Omit<FormControlProps, 'defaultValue' | 'onBlur' | 'onChange'> & {
         options?: SelectProps;
         disableValue?: string;
+        emptyText?: string | ReactElement;
```
djhi, line 409: I think you probably want a `ReactNode` here

djhi (changes requested):
Thanks!

packages/ra-ui-materialui/src/input/SelectArrayInput.tsx
Tomas025: OK, tonight I make the change, thanks for the suggestion

Tomas025:
I agree that it makes more sense a clear _button_, but I tried to replicate the way it was in `<SelectInput>` and I also realized that it `<NullableBooleanInput>` is also that way with an empty field in the choices

fzaninotto:
> I agree that it makes more sense a clear _button_, but I tried to replicate the way it was in `<SelectInput>` and I also realized that it `<NullableBooleanInput>` is also that way with an empty field in the choices

I understand your reasonning, and it makes sense. However, for new features, we look for the best balance between intuitive DX and good UX. In this case, I think the balance clearly points to a clear button as adornment.

fzaninotto:

PR #10505, opened 2025-02-10
Remove doc from published react-admin package

djhi:
## Problem

We currently include a copy of the `docs` directory in the published `react-admin` package. This makes it quite heavy and is probably useless.

## Solution

- remove the command that copy the `docs` directory when building the package.
- remove the `docs` directory from the included files of the `package.json` for `react-admin`

## Additional Checks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,10 +57,7 @@ build-ra-language-french:
 
 build-react-admin:
 	@echo "Transpiling react-admin files...";
-	@rm -rf ./packages/react-admin/docs
 	@cd ./packages/react-admin && yarn build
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -57,10 +57,7 @@ build-ra-language-french:
 
 build-react-admin:
 	@echo "Transpiling react-admin files...";
-	@rm -rf ./packages/react-admin/docs
```
fzaninotto, line 60: You should probably keep this one for those of us who have an old copy of the docs

erwanMarmelab (approved):

fzaninotto (approved):

Issue #10474, opened 2025-01-26
[DateInput] defaultValue does not override value from record


ogroppo:
**What you were expecting:**
if the value of the field from dataprovider is A and defaultValue is B - I expect B to override A

**What happened instead:**
Value from data provider always wins.

**Steps to reproduce:**
Follow the examples from jsDoc
```
 * @example
 * // If the initial value string contains more than a date (e.g. an hour, a timezone),
 * // these details are ignored.
 * <DateInput source="published_at" defaultValue="2021-09-11T20:46:20.000-04:00" />
 * // The input will display '2021-09-11' whatever the browser timezone.
 *
 * @example
 * // If the initial value is a Date object, DateInput converts it to a string
 * // and ignores the timezone.
 * <DateInput source="published_at" defaultValue={new Date("2021-09-11T20:46:20.000-04:00")} />
 * // The input will display '2021-09-11' whatever the browser timezone.
 *
```
source always wins over defaultValue

CODE
<img width="296" alt="Image" src="https://github.com/user-attachments/assets/a6eb8955-2d17-4d02-b598-ffa281cedb72" />
RESULT (data from source)
<img width="157" alt="Image" src="https://github.com/user-attachments/assets/4c5012a6-ea51-4c65-8308-6265a90e5c71" />

**Environment**

* React-admin version: 5.4.4
* React version: 18.3.1
* Browser: Chrome

ogroppo:
Also, slightly related. The documentation is misleading on page https://marmelab.com/react-admin/DateInput.html

Where it says 
```
Note: This conversion may change the date because of timezones. For example, the date string '2022-04-30T00:00:00.000Z' in Europe may be displayed as '2022-04-29' in Honolulu. If this is not what you want, pass your own [parse](https://marmelab.com/react-admin/Inputs.html#parse) function to <DateInput>.
```

That should replaced by
```
 * // If the initial value is a Date object, DateInput converts it to a string
 * // and ignores the timezone.
```
like the js docs, because the parse prop is for the output not the input

djhi:
Hi,

> if the value of the field from dataprovider is A and defaultValue is B - I expect B to override A

That's not how it works. Default values are only used when the value from the record is undefined.

Regarding the issue documentation, can you open a PR to fix it?

djhi:
It depends on your use case. Can you explain what you're trying to achieve?

ogroppo:
It's the answer to this question/problem of the docs:

```
[...]

PR #10506, opened 2025-02-10
[DateInput] Update docs to mention `format` prop for timezone adjustment

ogroppo:
## Problem

Problem is described [here](https://github.com/marmelab/react-admin/issues/10474), there is no way to manipulate the value coming from dataprovider because it will always win

## Solution

use `format` prop

## How To Test

Tested on storybook

## Additional Checks

- [x] The PR targets `master` for a bugfix

### DIFF
--- a/docs/DateInput.md
+++ b/docs/DateInput.md
@@ -37,7 +37,7 @@ The field value must be a string using the pattern `YYYY-MM-DD` (ISO 8601), e.g.
 
 In these cases, `<DateInput>` will automatically convert the value to the `YYYY-MM-DD` format, and will return a `string`, or `null` if the date is invalid.
 
-**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, use the [`format`](https://marmelab.com/react-admin/Inputs.html#format) prop to convert the date back to UTC and return the formatted string as `yyyy-MM-dd`.
+**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, use the [`format`](./Inputs.html#format) prop to convert the date to UTC and return the formatted string as `yyyy-MM-dd` to the input. In this scenario you likely want the outgoing value to stay consistent with UTC too, so provide your own [`parse`](./Inputs.md#parse) function to `<DateInput>` so that browser's timezone is stripped.
 
 ## Props
 
--- a/docs/DateInput.md
+++ b/docs/DateInput.md
@@ -37,7 +37,7 @@ The field value must be a string using the pattern `YYYY-MM-DD` (ISO 8601), e.g.
 
 In these cases, `<DateInput>` will automatically convert the value to the `YYYY-MM-DD` format, and will return a `string`, or `null` if the date is invalid.
 
-**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, pass your own `value` (perhaps a maipulation of the field from the context, to fix the offset) to `<DateInput>`.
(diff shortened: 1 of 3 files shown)


ogroppo:
updated to be a doc change only, thanks.

djhi (changes requested):

docs/DateInput.md
```diff
@@ -37,7 +37,7 @@ The field value must be a string using the pattern `YYYY-MM-DD` (ISO 8601), e.g.
 
 In these cases, `<DateInput>` will automatically convert the value to the `YYYY-MM-DD` format, and will return a `string`, or `null` if the date is invalid.
 
-**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, pass your own [`parse`](./Inputs.md#parse) function to `<DateInput>`.
+**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, use the [`format`](https://marmelab.com/react-admin/Inputs.html#format) prop to convert the date back to UTC and return the formatted string as `yyyy-MM-dd`.
```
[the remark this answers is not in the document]
ogroppo: sure!

packages/ra-ui-materialui/src/input/DateInput.tsx
```diff
@@ -36,6 +36,11 @@ import { InputHelperText } from './InputHelperText';
  * // The input will display '2021-09-11' whatever the browser timezone.
  *
  * @example
+ * // If you want to manipulate the value from the field to adjust its timezone, use format prop
+ * <DateInput source="published_at" format={value => new Date(value).toISOString().split("T")[0} />
+ * // The input will display the UTC day whatever the browser timezone.
```
slax57, line 41:
```suggestion
 * // If you want to manipulate the value from the field to adjust its timezone, use the format prop
 * <DateInput source="published_at" format={value => new Date(value).toISOString().split("T")[0] />
 * // The input will display the UTC day regardless of the browser timezone.
```

ogroppo: sure, amended for the other examples too

docs/DateInput.md
```diff
@@ -37,7 +37,7 @@ The field value must be a string using the pattern `YYYY-MM-DD` (ISO 8601), e.g.
 
 In these cases, `<DateInput>` will automatically convert the value to the `YYYY-MM-DD` format, and will return a `string`, or `null` if the date is invalid.
 
-**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, pass your own [`parse`](./Inputs.md#parse) function to `<DateInput>`.
+**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, use the [`format`](./Inputs.html#format) prop to convert the date to UTC and return the formatted string as `yyyy-MM-dd` to the input. In this scenario you likely want the outgoing value to stay consistent with UTC too, so provide your own [`parse`](./Inputs.md#parse) function to `<DateInput>` so that browser's timezone is stripped.
```
djhi, line 40:
```suggestion
**Note**: This conversion may change the date because of timezones. For example, the date string `'2022-04-30T00:00:00.000Z'` in Europe may be displayed as `'2022-04-29'` in Honolulu. If this is not what you want, use the [`format`](./Inputs.html#format) prop to convert the date to UTC and return the formatted string as `yyyy-MM-dd` to the input. In this scenario you likely want the outgoing value to stay consistent with UTC too, so provide your own [`parse`](./Inputs.md#parse) function to `<DateInput>` to transform it as needed.
```

I suggest this because the `DateInput` returns a stringin the `yyyy-mm-dd` format by default so you probably want to **add** some timezone (UTC or not).

djhi (changes requested):

docs/DateInput.md
ogroppo: ah yeah 🤦🏻

djhi (approved):

PR #10511, opened 2025-02-12
Refactor `create-react-admin` and support `ra-supabase`

djhi:
## Problem

The interactive mode forces us to document the choices to select, which makes tutorials a bit harder to follow. The recent introduction of the `--basic` flag only addresses one particular case.

We also want to promote `ra-supabase`.

## TODO

- [x] https://github.com/marmelab/ra-supabase/pull/83

## Solution

- Make the interactive mode optional. By default, create-react-admin will create an empty react-admin app (no data provider, no auth provider, no resource, npm)
    ```
    npx create-react-admin myAdmin
    ````

    Users who want to use interactive mode must specify it:
    ```
    npx create-react-admin --interactive myAdmin
    ```

    Users who want to override one option can do so in the command line
    ```
    npx create-react-admin --dataProvider ra-data-json-server --resources posts,users myAdmin
    ```

- Remove the `--basic` option, which is now useless
- Add support for provider shortcuts (`fakerest` -> `ra-data-fakerest`, etc)
- Add support for more package managers (`bun` and `pnpm`)
- Add support for `ra-supabase`

## How To Test

- `make build-create-react-admin install`
- Then test any combination of flags

It's difficult, even impossible for yarn, to test the detection of the package manager used to run the command:
- `bun run ./node_modules/.bin/create-react-admin` should use `bun` as the installer (a bun lock file will be generated)
- `npm run ./node_modules/.bin/create-react-admin` should use `npm` as the installer (a npm lock file will be generated)

## Additional Checks


packages/create-react-admin/templates/ra-supabase/README.md
```diff
@@ -0,0 +1,3 @@
+## Customize The Application
+
+Follow the instructions in your browser console once you start the application.
```
erwanMarmelab, line 3: Could you please specify that the user need to fill their `.env` file. It is better for beginners

packages/create-react-admin/src/cli.tsx
```diff
@@ -6,31 +6,88 @@ import App from './app.js';
...
+	  --data-provider  Set the data provider to use ("ra-data-fakerest", "ra-data-simple-rest", "ra-data-json-server", "ra-supabase" or "none")
 	  --auth-provider  Set the auth provider to use ("local-auth-provider" or "none")
 	  --resource       Add a resource that will be initialized with guessers (can be used multiple times). Set to "skip" to bypass the interactive resource step.
-	  --install        Set the package manager to use for installing dependencies ("yarn", "npm" or "skip" to bypass the interactive install step)
-	  --basic          Skip all the interactive steps and create a basic app with no data provider, no auth provider, no resources and install with npm
+	  --install        Set the package manager to use for installing dependencies ("yarn", "npm", "pnpm", "bun" or "skip" to bypass the interactive install step)
 
     Examples
 	  $ create-admin-app my-admin
 	  $ create-admin-app my-admin --data-provider ra-data-json-server --auth-provider local-auth-provider --resource posts --resource comments --install npm
```
slax57, line 80: Just realized those examples are wrong!

```diff
- create-admin-app
+ npx create-react-admin@latest
```

packages/create-react-admin/src/cli.tsx
```diff
@@ -55,17 +112,21 @@ const cli = meow(
...
-    const authProvider = cli.flags.basic ? 'none' : cli.flags.authProvider;
-    const install = cli.flags.basic ? 'npm' : cli.flags.install;
-    const resources =
-        cli.flags.basic || cli.flags.resource.includes('skip')
-            ? []
-            : cli.flags.resource;
+    const name = cli.input.length > 0 ? cli.input[0].trim() : undefined;
+    if (!name && !cli.flags.interactive) {
+        console.error(
+            'Please provide a name for your admin application: create-admin-app <name>'
```
slax57, line 118: same here

packages/create-react-admin/src/cli.tsx
```diff
@@ -6,31 +6,88 @@ import App from './app.js';
...
     {
         flags: {
             help: {
                 type: 'boolean',
                 alias: 'h',
             },
-            basic: {
+            interactive: {
                 type: 'boolean',
+                alias: 'i',
```
slax57, line 90: I see you did not update the choices for the `install` flag a few lines below (it's still `choices: ['yarn', 'npm', 'skip']`). Aren't the possible choices checked? :thinking: Because I'm still able to pick 'bun' or 'pnpm' according to my tests, so I'm surprised it works...

djhi: It looks like they're not checked, I fixed them anyway

packages/create-react-admin/src/cli.tsx
djhi: No, this one is actually required if not interactive

slax57: Yeah but what I meant is that the message still mentions `create-admin-app` which is incorrect :grimacing:

slax57 (approved):

PR #10515, opened 2025-02-14
Fix `useDataProvider` should only log errors in development mode

vytautassvirskas:
close https://github.com/marmelab/react-admin/issues/8738

### DIFF
--- a/packages/ra-core/src/dataProvider/useDataProvider.ts
+++ b/packages/ra-core/src/dataProvider/useDataProvider.ts
@@ -109,7 +109,7 @@ export const useDataProvider = <
                             .apply(dataProvider, args)
                             .then(response => {
                                 if (
-                                    process.env.NODE_ENV !== 'production' &&
+                                    process.env.NODE_ENV === 'development' &&
                                     reactAdminFetchActions.includes(type)
                                 ) {
                                     validateResponseFormat(response, type);


fzaninotto (approved):

fzaninotto:
A unit test fails...

vytautassvirskas:
> Any updates?

hey, I was quite busy last week. I will update tests today

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
```diff
@@ -229,7 +229,8 @@ describe('useDataProvider', () => {
         );
     });
 
-    it('should call getList and show error', async () => {
+    it('should call getList and show error in development environment', async () => {
+        process.env.NODE_ENV = 'development';
```
slax57, line 233: I'm afraid mutating `process.env.NODE_ENV` like that will leak to other tests and break test independence. Can you mock the result using `jest.mock` or `jest.spy` instead (and call `jest.restoreMocks` at the end)?

slax57 (changes requested):

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
vytautassvirskas: hey 👋 
I updated the tests in a slightly different manner than you suggested 🤞. If that's not suitable, I can implement the approach you recommended.

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
```diff
@@ -251,6 +259,63 @@ describe('useDataProvider', () => {
...
         );
+
+        process.env = originalEnv;
+    });
+
+    it('should call getList and not show error in test environment', async () => {
+        jest.resetModules();
+        process.env = {
+            ...originalEnv,
+        };
```
djhi, line 270: You can probably put all the new tests inside a `describe` with a `beforeEach` and `afterEach` to ensure nothing leaks.

packages/ra-core/src/dataProvider/useDataProvider.spec.tsx
slax57: I tried for a bit to come up with a cleaner solution, but was unsuccessful.
I could however confirm your solution prevents leakage of the `process.env.NODE_ENV` value, so it's good enough for me. I'll accept it.

slax57 (approved):

slax57:
Almost forgot to say: thank you for your contribution! :blush:

PR #10516, opened 2025-02-14
[chore] Improve release tooling

slax57:
## Problem

A large part of the react-admin release process is manual, thus cumbersome and prone to errors.

## Solution

Further automate the release process.

- [x] Generate and reorder the changelog entries automatically
- [x] Use Github API to close milestone and create new milestone
- [x] Use Github API to create Github Release and copy the changelog
- [x] ~~Script these steps and include them in the [npm lifecycle](https://docs.npmjs.com/cli/v11/using-npm/scripts#life-cycle-scripts)~~

## Out of scope

Some release steps are not automated yet (due to time constraints).
Examples:
- running the EE tests
- updating the docs
- updating the create-react-admin dependencies

## How To Test

These scripts support a `RELEASE_DRY_RUN` env variable allowing to skip all mutations to git, GitHub or npm, except 1 git commit (but not pushed!) from lerna.

First, create a `.env` file from the `.env.template` file and fill it in with your Github token.

Then run:

```
RELEASE_DRY_RUN=true make release
```

When this is done, don't forget to revert the commit created by lerna to update the versions:

```
git reset --hard HEAD~1
```

## Additional Checks


scripts/release.sh
```diff
@@ -0,0 +1,44 @@
...
+make install
+
+step "make build"
+make build
+
+step "manual tests: Run the EE tests"
+echo "Copy the the packages folder content inside the node_modules of ra-enterprise, then run a full build and run the tests"
+echo "Tip: You can use the 'copy-ra-oss-packages-to-ee.sh' script if you have it"
+echo "Press Enter when this is done"
+read
```
djhi, line 26: I think we should agree on a directory structure and do this automatically

djhi (commented):
This is really cool!

scripts/update-changelog.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const prOrder = [
+    'feature',
+    'fix',
+    '[doc]',
+    '[typescript]',
+    '[demo]',
+    '[website]',
+    '[storybook]',
+    '[demo]',
```
fzaninotto, line 17: demo is there twice

slax57:
Back to WIP as I need to check the github tag will happen before the `create-github-release` script is run

scripts/release.sh
slax57: yes I have plans for this, but will do that in a later PR (hopefully)

slax57:
All fixed! back to RFR!

djhi (changes requested):
Awesome!

PR #10518, opened 2025-02-15
Add LoginWithEmail to facilitate login with email

fzaninotto:
## Problem

To create a custom login form (e.g., to use email/password instead of userrname/password), developers have to copy an enorm amount of code from react-admin. 

## Solution

Provide a built-in `LoginWithEmail` component:

```tsx
import { Admin, LoginWithEmail } from 'react-admin';
import authProvider from './authProvider';

const App = () => (
    <Admin loginPage={LoginWithEmail} authProvider={authProvider}>
        ...
    </Admin>
);
```

LoginWithEmail

Improve composition of Login components,to allow to easily add custom content. 

```jsx
import { Box, Link } from '@mui/material';
import { Link as RouterLink } from 'react-router-dom';
import { Login, LoginForm } from 'react-admin';

const MyLogin = () => (
    <Login>
        <LoginForm />
        <Box textAlign="center" mb={1}>
            <Link component={RouterLink} to="/forgot-password">
                Forgot password?
            </Link>
        </Box>
    </Login>
);
```

LoginWithContent

I also updated the default login form to use a `<PasswordInput>` instead of a `<TextInput type="password">` so that we benefit from the "show/hide password" button. 

## How To Test

- The custom login forms of the  eCommerce and CRM demos still work, even though 90% of their code was removed.  
- The Login story was expanded to cover more cases
- Addes a LoginWithEmail story

## Additional Checks

- ~~[ ] The PR includes **unit tests** (current tests are enough)~~


packages/ra-ui-materialui/src/auth/LoginForm.tsx
```diff
@@ -46,22 +46,26 @@ export const LoginForm = (props: LoginFormProps) => {
...
+                {children || (
+                    <>
+                        <TextInput
+                            autoFocus
+                            source="username"
+                            label={translate('ra.auth.username')}
+                            autoComplete="username"
+                            validate={required()}
+                        />
+                        <PasswordInput
```
fzaninotto, line 61: **Notice**: I changed TextInput for PasswordInput here.

examples/crm/src/login/LoginForm.tsx
```diff
@@ -1,122 +1,42 @@
...
-                        Forgot your password?
-                    </Typography>
-                </CardActions>
-            </StyledForm>
-        </Login>
-    );
-};
-
-type FieldValues = Record<string, any>;
+export const LoginForm = () => (
```
fzaninotto, line 12: Couldn't use LoginWithEmail here as we also add the "forgot password" link.

packages/ra-ui-materialui/src/auth/LoginWithEmail.tsx
```diff
@@ -0,0 +1,30 @@
...
+                    type="email"
+                    validate={required()}
+                />
+                <PasswordInput
+                    source="password"
+                    label={translate('ra.auth.password', { _: 'Password' })}
+                    autoComplete="current-password"
+                    validate={required()}
+                />
+            </LoginForm>
```
djhi, line 27: Shouldn't we have a `LoginFormWithEmail` to make customization even easier (adding custom content)?

djhi (commented):
Nice improvement!

packages/ra-ui-materialui/src/auth/LoginWithEmail.tsx
fzaninotto: I think this would be too many components (not easy to document). Instead, I documented how to pass custom children.

djhi (approved):

PR #10522, opened 2025-02-17
Fix FilerLiveForm does not allow to clear filters

### LINKED ISSUE #10512: Issue with parse prop in filters after update (5.3.4 → 5.4.2)
**What you were expecting:**
When clicking the clear button, the input value should reset.
The state should be updated accordingly.
The filter should be removed from both the input and the URL.

**What happened instead:**
Clicking the clear button does not update the input state.
The filter remains in both the input field and the URL.
The list view does not refresh as expected.

**Steps to reproduce:**

1. Use a filter input with the parse prop.
2. Apply a filter.
3. Click the clear button.
4. Observe that the input state is not updated and the filter remains applied.

**Related code:**
https://codesandbox.io/p/github/BrenoWii/react-admin-sandbox/find-unexpected-behavior

**Other information:**

Image

**Environment**

* React-admin version: 5.4.2
* Last version that did not exhibit the issue (if applicable): 5.3.4
* React version: N/A
* Browser: Chrome

djhi:
## Problem

Fixes #10512. It seems that when calling `unset` which remove the property matching the filter input source from the filters, the value get re-added by react-hook-form.

## Solution

Set the filter value to an empty string which correctly removes it from the filter context.

## How To Test

In ra-core:
- https://react-admin-storybook-8h3fdlb89-marmelab.vercel.app/?path=/story/ra-core-form-filterliveform--parse-format
- Enter a number and removes it character by character until empty
- Enter a number then select the input value and remove it
- Enter a number then click the clear button

In ra-ui-materialui:
- https://react-admin-storybook-8h3fdlb89-marmelab.vercel.app/?path=/story/ra-ui-materialui-list-filter-filterliveform--parse-format
- Enter a number and removes it character by character until empty
- Enter a number then select the input value and remove it
- Enter a number then click the clear button

## Additional Checks

### DIFF
--- a/packages/ra-core/src/form/FilterLiveForm.spec.tsx
+++ b/packages/ra-core/src/form/FilterLiveForm.spec.tsx
@@ -5,6 +5,7 @@ import {
     GlobalValidation,
     MultipleFilterLiveForm,
     MultipleInput,
+    ParseFormat,
     PerInputValidation,
 } from './FilterLiveForm.stories';
 import React from 'react';
@@ -24,6 +25,28 @@ describe('<FilterLiveForm />', () => {
         fireEvent.change(input, { target: { value: 'foo' } });
         await screen.findByText('{"category":"deals","title":"foo"}');
         fireEvent.change(input, { target: { value: '' } });
+        await screen.findByText('{"category":"deals"}');
+    });
+
+    it.only('should allow to clear a filter value with parse/format', async () => {
+        render(<ParseFormat />);
+        const input = await screen.findByLabelText('document');
+        fireEvent.change(input, { target: { value: '123123123123' } });
+        await screen.findByText(
+            '{"category":"deals","document":"123123123123"}'
+        );
+        fireEvent.change(input, { target: { value: '' } });
+        await screen.findByText('{"category":"deals"}');
+    });
+
+    it.only('should allow to clear a filter value through a clear button with parse/format', async () => {
+        render(<ParseFormat />);
+        const input = await screen.findByLabelText('document');
+        fireEvent.change(input, { target: { value: '123123123123' } });
+        await screen.findByText(
+            '{"category":"deals","document":"123123123123"}'
+        );
+        fireEvent.click(screen.getByText('Clear'));
         await screen.findByText('{"category":"deals"}');
     });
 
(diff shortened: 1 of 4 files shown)


BrenoWii:
:+1:

packages/ra-core/src/form/FilterLiveForm.spec.tsx
```diff
@@ -27,6 +28,28 @@ describe('<FilterLiveForm />', () => {
         await screen.findByText('{"category":"deals"}');
     });
 
+    it.only('should allow to clear a filter value with parse/format', async () => {
```
WiXSL, line 31: `.only` spotted

packages/ra-core/src/form/FilterLiveForm.spec.tsx
```diff
@@ -27,6 +28,28 @@ describe('<FilterLiveForm />', () => {
...
+        const input = await screen.findByLabelText('document');
+        fireEvent.change(input, { target: { value: '123123123123' } });
+        await screen.findByText(
+            '{"category":"deals","document":"123123123123"}'
+        );
+        fireEvent.change(input, { target: { value: '' } });
+        await screen.findByText('{"category":"deals"}');
+    });
+
+    it.only('should allow to clear a filter value through a clear button with parse/format', async () => {
```
WiXSL, line 42: `.only` spotted

WiXSL (changes requested):

packages/ra-core/src/form/FilterLiveForm.stories.tsx
```diff
@@ -1,14 +1,20 @@
...
 import { ListContextProvider } from '../controller/list/ListContextProvider';
 import { useList } from '../controller/list/useList';
 import { useListContext } from '../controller/list/useListContext';
 
 export default { title: 'ra-core/form/FilterLiveForm' };
 
-const TextInput = props => {
-    const { field, fieldState } = useInput(props);
+const TextInput = ({ defaultValue = '', ...props }: InputProps) => {
+    const { field, fieldState } = useInput({ defaultValue, ...props });
```
slax57, line 17: why do you need to provide a `defaultValue` to `useInput`? I'm worried that `defaultValue = ''` is not the same as `defaultValue = undefined`

slax57 (commented):
Other than that, lgtm!

packages/ra-core/src/form/FilterLiveForm.stories.tsx
djhi: That avoid a warning in the story about uncontrolled/controlled inputs.

PR #10530, opened 2025-02-19
Add support for mutationMode in useCreate

djhi:
## Problem

We currently don't support `optimistic` and `undoable` `mutationMode` in `useCreate`, which prevent us from supporting offline first applications.

## Solution

Given that the new record identifier is generated client side, we can support  `optimistic` and `undoable` `mutationMode`.

- [ ] Add a full app story

## How To Test

- https://react-admin-storybook-git-create-mutation-mode-marmelab.vercel.app/?path=/story/ra-core-dataprovider-usecreate-optimistic--success-case
- https://react-admin-storybook-git-create-mutation-mode-marmelab.vercel.app/?path=/story/ra-core-dataprovider-usecreate-pessimistic--success-case
- https://react-admin-storybook-git-create-mutation-mode-marmelab.vercel.app/?path=/story/ra-core-dataprovider-usecreate-undoable--success-case

## Additional Checks

### DIFF
--- a/docs/useCreate.md
+++ b/docs/useCreate.md
@@ -97,10 +97,10 @@ const LikeButton = () => {
 ```tsx
 const notify = useNotify();
 const redirect = useRedirect();
-const like = { postId: record.id };
+
 const [create, { isPending, error }] = useCreate(
     'likes',
-    { data: like },
+    { data: { id: uuid.v4(), postId: record.id } },
     {
         mutationMode: 'optimistic',
         onSuccess: () => {
@@ -305,17 +305,17 @@ If this is not enough, you can use the `returnPromise` option so that the `creat
 This can be useful if the server changes the record, and you need the newly created data to create/update another record. 
 
 ```jsx
-const [create] = useCreate(
+const [createPost] = useCreate(
     'posts',
     { id: record.id, data: { isPublished: true } },
     { returnPromise: true }
 );
-const [create] = useCreate('auditLogs');
+const [createAuditLog] = useCreate('auditLogs');
 
 const createPost = async () => {
     try {
-        const post = await create();
-        create('auditLogs', { data: { action: 'create', recordId: post.id, date: post.createdAt } });
+        const post = await createPost();
+        createAuditLog('auditLogs', { data: { action: 'create', recordId: post.id, date: post.createdAt } });
     } catch (error) {
         // handle error
     }
--- a/docs/useCreate.md
+++ b/docs/useCreate.md
@@ -9,7 +9,7 @@ This hook allows to call `dataProvider.create()` when the callback is executed.
 
 ## Syntax
 
-```jsx
+```tsx
 const [create, { data, isPending, error }] = useCreate(
     resource,
     { data, meta },
@@ -19,7 +19,7 @@ const [create, { data, isPending, error }] = useCreate(
 
 The `create()` method can be called with the same parameters as the hook:
 
-```jsx
+```tsx
 create(
     resource,
     { data },
@@ -31,7 +31,7 @@ So, should you pass the parameters when calling the hook, or when executing the 
 
 ## Usage
 
-```jsx
+```tsx
 // set params when calling the hook
 import { useCreate, useRecordContext } from 'react-admin';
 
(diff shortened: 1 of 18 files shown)


erwanMarmelab:
Thank you for the full / nice doc and tests

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -106,91 +217,287 @@ export const useCreate = <
...
+        // as the previousData would be overwritten by the optimistic update
+        paramsRef.current = params;
 
-        if (returnPromise) {
-            return mutation.mutateAsync(
+        if (mutationMode) {
+            mode.current = mutationMode;
+        }
+
+        if (returnPromise && mode.current !== 'pessimistic') {
```
guilbill, line 368: Is it needed ? as the returnPromise is defined base on the mode https://github.com/marmelab/react-admin/blob/35bf17e551f3fcf1acb171870e7b711f6b427ed0/packages/ra-core/src/controller/create/useCreateController.ts#L147

packages/ra-ui-materialui/src/button/CreateButton.stories.tsx
```diff
@@ -50,6 +50,7 @@ const AccessControlAdmin = ({ queryClient }: { queryClient: QueryClient }) => {
     const [resourcesAccesses, setResourcesAccesses] = React.useState({
         'books.list': true,
         'books.create': false,
+        'books.delete': false,
```
guilbill, line 53: I don't get why we need this, is it a fix ?

docs/useCreate.md
```diff
@@ -61,6 +61,267 @@ const LikeButton = () => {
...
+- `optimistic`, and
+- `undoable`
+
+**Note**: For `optimistic` and `undoable` modes, the record `id` must be generated client side. Those two modes are useful when building local first applications.
+
+Here is an example of using the `optimistic` mode:
+
+```jsx
+// In optimistic mode, ids must be generated client side
+const id = uuid.v4();
```
guilbill, line 214: I was wondering if it was common to do it like this, generating the id client side. I understand the why, but I'm not sure on the how we solve this. Couldn't we have a mechanism witch a temporary id, replacing it as the data is fetched ?

guilbill (changes requested):
Tests and documentation are very clean! thanks for that, although, some errors in the doc, and I have some question about the `optimistic` mode and the fact that we generate the id client side.

packages/ra-ui-materialui/src/button/CreateButton.stories.tsx
djhi: It's a fix to avoid error messages in the tests output. Just a cleanup really

docs/useCreate.md
djhi: Yes it is common in offline/local first scenarios. A temporary mechanism would make handling relationships such as those with ReferenceOneInput or ReferenceManyToManyInput a nightmare to handle. This is a good starting point and if we do find a way to use temporary id, we can add it later.

packages/react-admin/src/Admin.tsx
```diff
@@ -115,6 +115,7 @@ export const Admin = (props: AdminProps) => {
         loginPage,
         notification,
         queryClient,
+        QueryClientProvider,
```
slax57, line 118: shouldn't we document this new prop?

packages/ra-core/src/dataProvider/useCreate.spec.tsx
```diff
@@ -186,9 +200,148 @@ describe('useCreate', () => {
...
+                expect(screen.queryByText('Hello World')).toBeNull();
+                expect(screen.queryByText('mutating')).toBeNull();
+            });
+            await screen.findByText('not found');
+        });
+        it('when undoable, displays result and success side effects right away and fetched on confirm', async () => {
+            render(<SuccessCaseUndoable timeout={10} />);
+            act(() => {
+                screen.getByText('Create post').click();
+            });
```
slax57, line 274: question: why do you need `act`? :thinking:

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -80,12 +92,111 @@ export const useCreate = <
...
+            ) =>
+                res && res.pages
+                    ? {
+                          ...res,
+                          pages: res.pages.map(page => ({
+                              ...page,
+                              data: updateColl(page.data),
+                          })),
+                      }
+                    : res,
```
slax57, line 181: Are you adding the new record to all pages of the infinite query? Won't that lead to the record appearing multiple times in the list? :thinking:

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -80,12 +92,111 @@ export const useCreate = <
...
+            [resource, 'getOne', { id: String(id), meta }],
+            (record: RecordType) => ({ ...record, ...clonedData }),
+            { updatedAt }
+        );
+        queryClient.setQueriesData(
+            { queryKey: [resource, 'getList'] },
+            (res: GetListResult) =>
+                res && res.data ? { ...res, data: updateColl(res.data) } : res,
+            { updatedAt }
+        );
```
slax57, line 165: shouldn't the total be updated to `total + 1`?

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -80,12 +92,111 @@ export const useCreate = <
...
+        queryClient.setQueriesData(
+            { queryKey: [resource, 'getManyReference'] },
+            (res: GetListResult) =>
+                res && res.data
+                    ? { data: updateColl(res.data), total: res.total }
+                    : res,
+            { updatedAt }
+        );
```
slax57, line 197: shouldn't the total be updated to `total + 1`?

slax57 (changes requested):

packages/react-admin/src/Admin.tsx
djhi: This was commited by mistake

packages/ra-core/src/dataProvider/useCreate.spec.tsx
djhi: copy/pasted from `useUpdate`

djhi: I confirm it's unnecessary

slax57 (approved):

PR #10536, opened 2025-02-20
[Doc] Fix dialogs `title` doc

erwanMarmelab:
## Problem

We have some typos in dialog `title`'s sections

## Solution

Fix those typos

### DIFF
--- a/docs/CreateDialog.md
+++ b/docs/CreateDialog.md
@@ -210,35 +210,32 @@ Unlike the `<Create>` components, with Dialog components the title will be displ
 Here is an example:
 
 ```tsx
-import React from 'react';
 import {
     List,
+    ListActions,
     Datagrid,
     SimpleForm,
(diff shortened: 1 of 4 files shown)


docs/CreateDialog.md
```diff
@@ -210,35 +210,32 @@ Unlike the `<Create>` components, with Dialog components the title will be displ
...
 ```tsx
-import React from 'react';
 import {
     List,
+    ListActions,
     Datagrid,
     SimpleForm,
     TextInput,
     DateInput,
+    DateField,
```
slax57, line 220: I don't think you need to import DateField in this example

slax57 (commented):
otherwise, lgtm

docs/CreateDialog.md
djhi, line 220:
```suggestion
```

PR #10541, opened 2025-02-21
Add a tips section on the home page

djhi:
## Problem

We have many tips that we regularly post on different medias. It's a shame we don't use them on the documentation itself

## Solution

- Add a tip section on the https://marmelab.com/react-admin/documentation.html page
- Fix the home made client side navigation to behave correctly including when using the browser back button
- Use modern scripts

## How To Test

- `make doc`
- click the *Documentation* link in the app bar
- press refresh and you should get a different tip each time
- Use links and the browser back button to navigate. Ensure the tips is only ever displayed on the documentation page and that the right side TOC is never displayed on the documentation page.

## Additional Checks

### DIFF
--- a/docs/js/ra-doc-exec.js
+++ b/docs/js/ra-doc-exec.js
@@ -5,7 +5,7 @@ import * as babel from 'https://esm.sh/prettier@3.5.1/plugins/babel';
 import * as estree from 'https://esm.sh/prettier@3.5.1/plugins/estree';
 import { marked } from 'https://esm.sh/marked@15.0.7';
 
-var tipElement, allMenus, navLinks, versionsLinks;
+var tipElement, tipContainer, allMenus, navLinks, versionsLinks;
 
 const showTip = async () => {
     const tips = await getContents('/assets/tips.md');
@@ -386,6 +386,9 @@ document.addEventListener('click', event => {
     if (tipElement) {
         tipElement.remove();
     }
+    if (tipContainer) {
+        tipContainer.remove();
+    }
     window.sessionStorage.setItem(
         'scrollIntoView',
         link.closest('.sidenav') ? 'false' : 'true'
@@ -432,6 +435,7 @@ window.addEventListener('popstate', () => {
 
 window.addEventListener('DOMContentLoaded', () => {
     tipElement = document.getElementById('tip');
+    tipContainer = document.getElementById('tip-container');
     allMenus = Array.from(document.querySelectorAll(`.sidenav a.nav-link`));
     navLinks = allMenus
         .filter(link => !link.classList.contains('external'))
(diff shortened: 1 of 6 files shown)


fzaninotto:
The feature doesn't work in my tests with `make docker-doc`.

guilbill (approved):

docs/js/ra-doc-exec.js
```diff
@@ -1,5 +1,43 @@
-/* global Prism, prettier, prettierPlugins */
-var allMenus, navLinks, versionsLinks;
+/* global Prism */
+import { transpileModule } from 'https://esm.sh/typescript@5.7.3';
+import * as prettier from 'https://esm.sh/prettier@3.5.1/standalone';
+import * as babel from 'https://esm.sh/prettier@3.5.1/plugins/babel';
+import * as estree from 'https://esm.sh/prettier@3.5.1/plugins/estree';
```
fzaninotto, line 5: Why did you add this plugin?

docs/js/ra-doc-exec.js
```diff
@@ -185,232 +214,9 @@ const buildJSCodeBlocksFromTS = async () => {
     applyPreferredLanguage();
 };
 
-// eslint-disable-next-line no-unused-vars
```
fzaninotto, line 188: Did you change anything in the code you've moved? The large diff makes this PR hard to review.

docs/js/ra-doc-exec.js
```diff
@@ -1,5 +1,43 @@
...
+import * as babel from 'https://esm.sh/prettier@3.5.1/plugins/babel';
+import * as estree from 'https://esm.sh/prettier@3.5.1/plugins/estree';
+import { marked } from 'https://esm.sh/marked@15.0.7';
+
+export const showTip = async () => {
+    const tipContainer = document.getElementById('tip-container');
+    const tipElement = document.getElementById('tip');
+    if (!tipElement) return;
+
+    const tips = await getContents('/assets/tips.md');
```
fzaninotto, line 13: Can you await both Promises in parallel instead of in series?

fzaninotto (changes requested):
Nice!

docs/js/ra-doc-exec.js
djhi: We had it before and we still need it so that prettier understand TS

docs/js/ra-doc-exec.js
djhi: No

docs/js/ra-doc-exec.js
djhi: Correction: it's actually unnecessary

PR #10544, opened 2025-02-24
[chore] Fix release script does not include the changelog in the tag

slax57:
## Problem

Release script introduced by https://github.com/marmelab/react-admin/pull/10516 let lerna create the git tag for the version, but the changelog was updated afterwards so it was not included in the tag.

## Solution

Delete the tag created by lerna, and recreate it ourselves after the changelog was updated.

## How To Test

Cannot really be tested without doing a real release.
However individual commands can be tested manually.

## Additional Checks

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -40,10 +40,21 @@ echo "Press Enter when this is done"
 read
 
 step "lerna version"
-./node_modules/.bin/lerna version --force-publish --no-push --no-git-tag-version
+# Running lerna version
+# This will create a commit and a tag
+./node_modules/.bin/lerna version --force-publish --no-push
 
 # Get the version from package.json
 npm_package_version=$(jq -r '.version' ./packages/react-admin/package.json)
+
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -40,11 +40,22 @@ echo "Press Enter when this is done"
...
 # Get the version from package.json
 npm_package_version=$(jq -r '.version' ./packages/react-admin/package.json)
 
+# Remove the tag created by lerna
+echo "Removing tag v${npm_package_version} created by lerna"
+git tag -d "v${npm_package_version}"
+if [ ! -z "$RELEASE_DRY_RUN" ]; then
+    # In dry-run mode, reset the last commit to avoid accidental push
+    echo "dry mode -- Resetting the workspace to the last commit"
+    git reset --soft HEAD~1
```
djhi, line 56: :+1:

djhi (approved):

PR #10545, opened 2025-02-24
Add support for offline/local first applications

djhi:
## Problem

React-admin currently don't handle well the loss of network connectivity: in any demo, if you go to a list page, then simulate being offline using the devtool and try to go to the second page of the same list, it will still display the first page data while the pagination will indeed indicate you are on page 2.

We handled it a bit better by at least showing users a notification when they try to load new data while offline but we can do better and allow for resumable mutations.

## Solution

- Needs #10530 
- [x] Make sure the `ListContext` include the `isPaused` and `isPlaceholderData` props from react-query and update components/hooks accordingly
- [x] Show the `empty` component when those two props are true (ideally we should have a dedicated component for that or handle the case in the default empty but this is achievable in userland)
- [x] Add a `mutationKey` to all mutations so that we can provide default functions for them at the `queryClient` level (required by react-query: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#persisting-offline-mutations)
- [x] Make the simple example work offline (no pwa yet though)
- [x] Add an `offline` prop to reference fields components
- [x] Add an `offline` prop to list components
- [x] Add an `offline` prop to show and edit views
- [x] Add an `offline` prop to reference inputs components
- [x] Change the notification message for pending mutations while offline
- [x] Update documentation

## Screenshots

_Edit view_
image

_Show view_
image

_DataTable_ (same for all list child components
image

## How To Test

- Open the simple example
- Use the browser devtool to go offline
- Click the posts create button and fill it
- Go back to the posts list after and notice the new item is there
- Open the devtool and note there is no create call to the dataProvider
- Use the browser devtool to go online
- Open the devtool and note there is a create call to the dataProvider

Then:
- Follow the same steps but select the `optimistic` option in the form toolbar first.
- Follow the same steps but click the `undoable` option in the form toolbar first.
- Follow the same steps but edit a record instead of creating it.
- Follow the same steps for both edit and create but enter `f00bar` as the post title to trigger an error once the network is online.

In stories for list components, reference components and detail views, simply switch to `offline` and hit the storybook refresh button

## Additional Checks


cypress/e2e/edit.cy.js
```diff
@@ -47,7 +47,11 @@ describe('Edit Page', () => {
         it('should redirect to list page after edit success', () => {
             // For some unknown reason, the click on submit didn't work in cypress
             // so we submit with enter
-            EditPostPage.setInputValue('input', 'title', 'Lorem Ipsum{enter}');
+            EditPostPage.setInputValue(
+                'input',
+                'title',
+                'Lorem Ipsum again{enter}'
```
djhi, line 53: Because react-query now persist queries and mutations for offline mode, the previous test now leaks into the second (e.g. this post has its title changed to *Lorem Ipsum*). I tried to configure *testIsolation* in Cypress but our version is probably too old

packages/ra-core/src/controller/list/useListController.ts
```diff
@@ -115,7 +115,9 @@ export const useListController = <
         error,
         isLoading,
         isFetching,
+        isPaused,
         isPending,
+        isPlaceholderData,
```
djhi, line 120: Those linter warnings make no sense

djhi:
I can't see a way to test this except e2e. Not sure if we can simulate network loss with cypress though

slax57:
@djhi since the simple demo has the react-query DevTools enabled, you can use them to simulate network loss I believe (clicking on the 'Wifi' icon does the job). A bit ugly but it should work!

docs/DataProviders.md
```diff
@@ -884,3 +884,131 @@ export default App;
...
+**Tip**: Note that unlike the [_Calling Custom Methods_ example](./Actions.md#calling-custom-methods), we passed `userId` to the `mutate` function. This is necessary so that React Query passes it too to the default function when resuming the mutation.
+
+Then, register a default function for it:
+
+```ts
+// in src/queryClient.ts
+import { addOfflineSupportToQueryClient } from 'react-admin';
+import { QueryClient } from '@tanstack/react-query';
+import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client';
+import { createSyncStoragePersister } from '@tanstack/query-sync-storage-persister';
```
erwanMarmelab, line 998: Why importing it ? We don't use it in this example

packages/ra-core/src/controller/input/useReferenceInputController.ts
```diff
@@ -118,6 +118,7 @@ export const useReferenceInputController = <RecordType extends RaRecord = any>(
         reference,
         // @ts-ignore the types of the queryOptions for the getMAny and getList are not compatible
         options: {
+            // @ts-ignore FIXME
```
erwanMarmelab, line 121: Why ? I don't see any typescript error

erwanMarmelab (commented):
Praise: very understandable doc !!

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -94,18 +135,44 @@ const backlinksDefaultValue = [
...
+    const mutationMode = useMutationMode();
     const dateDefaultValue = useMemo(() => new Date(), []);
     return (
-        <Create redirect="edit">
+        <Create
+            redirect="edit"
+            mutationMode={mutationMode}
+            transform={data => ({ ...data, id: getNewId(mutationMode) })}
+        >
+            <Alert severity="info" role="presentation">
```
fzaninotto, line 163: Please hide this alert when not in offline mode.

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -94,18 +135,44 @@ const backlinksDefaultValue = [
...
+                To test offline support, add either{' '}
+                <Link to="/posts/create?mutationMode=optimistic">
```
fzaninotto, line 165: Why don't you use a special location state instead?

packages/ra-core/src/dataProvider/addOfflineSupportToQueryClient.ts
```diff
@@ -0,0 +1,102 @@
...
+ *         }}
+ *     >
+ *         <Admin queryClient={queryClient} dataProvider={dataProvider}>
+ *             <Resource name="posts" {...posts} />
+ *             <Resource name="comments" {...comments} />
+ *         </Admin>
+ *     </PersistQueryClientProvider>
+ * );
+ */
+export const addOfflineSupportToQueryClient = ({
```
fzaninotto, line 83: As explained earlier, can't we create a new Query Client instead? or at least return the modified one.

packages/ra-core/src/dataProvider/dataFetchActions.ts
```diff
@@ -8,6 +8,14 @@ export const UPDATE_MANY = 'UPDATE_MANY';
 export const DELETE = 'DELETE';
 export const DELETE_MANY = 'DELETE_MANY';
 
+export const reactAdminMutations = [
```
fzaninotto, line 11: "data provider mutations" would be a better name IMHO

packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -176,7 +176,7 @@ export const useGetList = <
                   }
                 : result,
         [result]
-    ) as UseQueryResult<RecordType[], Error> & {
+    ) as unknown as UseQueryResult<RecordType[], Error> & {
```
fzaninotto, line 179: Why do you need this?

packages/ra-ui-materialui/src/list/ListView.tsx
```diff
@@ -76,20 +78,21 @@ export const ListView = <RecordType extends RaRecord = any>(
...
-            (total == null &&
-                hasPreviousPage === false &&
-                hasNextPage === false &&
-                // @ts-ignore FIXME total may be undefined when using partial pagination but the ListControllerResult type is wrong about it
-                data.length === 0)) &&
-        // the user didn't set any filters
-        !Object.keys(filterValues).length &&
-        // there is an empty page component
-        empty !== false;
+        (isPaused && isPlaceholderData) ||
```
fzaninotto, line 81: I don't understand the rationale of showing the empty page when offline. This gives a wrong information.

fzaninotto (changes requested):
I think the change of the ListContext shape affects more components; e.g. useListContextWithProps.

docs/DataProviders.md
```diff
@@ -884,3 +884,129 @@ export default App;
...
+
+```ts
+// in src/queryClient.ts
+import { addOfflineSupportToQueryClient } from 'react-admin';
+import { QueryClient } from '@tanstack/react-query';
+import { dataProvider } from './dataProvider';
+
+export const queryClient = new QueryClient();
+
+addOfflineSupportToQueryClient({
```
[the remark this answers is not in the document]
djhi: No, as it simply calls functions on the queryClient. There's no way around it

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -47,10 +71,16 @@ const PostCreateToolbar = () => {
                         notify('resources.posts.notifications.created', {
                             type: 'info',
                             messageArgs: { smart_count: 1 },
+                            undoable: mutationMode === 'undoable',
                         });
                         redirect('show', 'posts', data.id);
                     },
                 }}
+                transform={data => ({
```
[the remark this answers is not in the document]
djhi: Because the first one is the default and submit the `<Create>` form, hence, the `transform` set on `<Create>` will run

examples/simple/src/posts/PostCreate.tsx
djhi: Those are instructions to know how to test the mutation modes. Why do you want to hide them?

examples/simple/src/posts/PostCreate.tsx
djhi: I don't get this comment

packages/ra-ui-materialui/src/list/ListView.tsx
djhi: Should we introduce a new `offline` prop and `<Offline>` component?

examples/simple/src/Layout.tsx
```diff
@@ -1,21 +1,80 @@
...
+ * we add a default onSettled function to the mutation defaults here.
+ */
+const NotificationsFromQueryClient = () => {
+    const queryClient = useQueryClient();
+    const notify = useNotify();
+
+    queryClient.setMutationDefaults([], {
+        onSettled(data, error) {
+            if (error) {
+                notify(error.message, { type: 'error' });
```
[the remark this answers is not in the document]
djhi: Did you actually try it? This effect only runs when there are persisted mutations to resume. For instance, create a new post with title `f00bar` in optimistic mode while offline and click _Save and edit_ then close your browser tab. Reopen the tab, restore the network if needed. You'll see the server error as expected.
However, should you do the same while offline, the above function will not be called

examples/simple/src/Layout.tsx
```diff
@@ -1,21 +1,80 @@
...
+            </Layout>
+            <ReactQueryDevtools
+                initialIsOpen={false}
+                buttonPosition="bottom-left"
+            />
+        </>
+    );
+};
+
+const useIsOnline = () => {
```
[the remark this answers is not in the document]
djhi: Reproduced. However, I simply use react-query `onlineManager`. I don't know how to solve this.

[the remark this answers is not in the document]
djhi: After further tests, it appears toggling network in the firefox devtool does not trigger the `online` nor `offline` events. You have to be actually offline or online at the OS level. This is a bug in FF

djhi:
> I think the change of the ListContext shape affects more components; e.g. useListContextWithProps.

That's the only one left AFAIK

fzaninotto:
When going offline, I don't understand why I can't go to the detail view of a post when I've already seen it in the list view. The `getList` populates the cache, right?

fzaninotto:
The 'how to test' instructions must be updated, as the tester must select the optimistic mutation mode before submitting the form.

fzaninotto:
When doing an optimistic offline mutation, the loader in the app bar keeps spinning forever even tough the optimistic update is already applied. Is it a good UI? Instead, I think we should have a way to mention the pending updates, e.g. using a badge on the loader icon.

packages/ra-ui-materialui/src/list/index.ts
```diff
@@ -13,6 +13,7 @@ export * from './ListGuesser';
 export * from './ListNoResults';
 export * from './ListToolbar';
 export * from './ListView';
+export * from '../Offline';
```
fzaninotto, line 16: reexported from the wrong directory (the `..` is weird here)

fzaninotto (changes requested):
We're getting closer!

packages/ra-ui-materialui/src/Offline.tsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
djhi: Why not but it won't take much less space

packages/ra-core/src/dataTable/DataTableBase.tsx
```diff
@@ -206,6 +221,7 @@ export interface DataTableBaseProps<RecordType extends RaRecord = any> {
     hasBulkActions: boolean;
     hover?: boolean;
     empty: ReactNode;
+    offline: ReactNode;
```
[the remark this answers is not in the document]
djhi: But the `empty` is required?

packages/ra-core/src/controller/input/ReferenceInputBase.tsx
```diff
@@ -78,7 +79,9 @@ export const ReferenceInputBase = (props: ReferenceInputBaseProps) => {
         filter,
     });
 
-    return (
+    return controllerProps.isPaused ? (
```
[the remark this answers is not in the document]
djhi: That's the case

packages/ra-ui-materialui/src/field/ReferenceField.tsx
```diff
@@ -106,28 +105,41 @@ export interface ReferenceFieldProps<
...
+                <Typography
+                    component="span"
+                    variant="body2"
+                    sx={{ color: 'error.main' }}
+                >
+                    <Translate i18nKey="ra.notification.http_error">
+                        Server communication error
+                    </Translate>
+                </Typography>
+            </Stack>
```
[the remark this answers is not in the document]
djhi: We only had an icon for error before with no messages for users **not** using assistive tools

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -260,3 +327,95 @@ const CreateUser = () => {
...
+                    aria-controls={open ? 'split-button-menu' : undefined}
+                    aria-expanded={open ? 'true' : undefined}
+                    aria-label="select merge strategy"
+                    aria-haspopup="menu"
+                    onClick={handleToggle}
+                >
+                    <ArrowDropDownIcon />
+                </Button>
+            </ButtonGroup>
+            <Popper
```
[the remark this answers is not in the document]
djhi: I just copied their example: https://mui.com/material-ui/react-button-group/#split-button

docs/DataProviders.md
```diff
@@ -884,3 +884,129 @@ export default App;
...
+    >
+        <Admin queryClient={queryClient} dataProvider={dataProvider}>
+            <Resource name="posts" {...posts} />
+            <Resource name="comments" {...comments} />
+        </Admin>
+    </PersistQueryClientProvider>
+)
+```
+{% endraw %}
+
```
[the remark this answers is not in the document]
djhi: Added

djhi:
> When doing an optimistic offline mutation, the loader in the app bar keeps spinning forever even tough the optimistic update is already applied. Is it a good UI? Instead, I think we should have a way to mention the pending updates, e.g. using a badge on the loader icon.

The loading indicator now has a badge when users are offline and there are pending mutations. This badge also has a tooltip that explains it.

examples/simple/src/posts/PostCreate.tsx
djhi: Note for the next reviewer: we now have a menu button that allows to select the mutation mode instead

examples/simple/src/posts/PostCreate.tsx
djhi: We don't have this anymore

docs/DataProviders.md
```diff
@@ -888,3 +888,133 @@ export default App;
...
+export const queryClient = addOfflineSupportToQueryClient({
+    queryClient,
+    dataProvider,
+    resources: ['posts', 'comments'],
+});
+
+queryClient.setMutationDefaults('banUser', {
+    mutationFn: async (userId) => {
+        return dataProviderFn.banUser(userId);
```
slax57:
```suggestion
        return dataProvider.banUser(userId);
```

examples/simple/package.json
```diff
@@ -12,8 +12,10 @@
     "dependencies": {
         "@mui/icons-material": "^5.16.12",
         "@mui/material": "^5.16.12",
+        "@tanstack/query-sync-storage-persister": "5.47.0",
         "@tanstack/react-query": "^5.21.7",
         "@tanstack/react-query-devtools": "^5.21.7",
+        "@tanstack/react-query-persist-client": "5.47.0",
```
slax57:
```suggestion
        "@tanstack/query-sync-storage-persister": "^5.47.0",
        "@tanstack/react-query": "^5.21.7",
        "@tanstack/react-query-devtools": "^5.21.7",
        "@tanstack/react-query-persist-client": "^5.47.0",
```

examples/simple/src/Layout.tsx
```diff
@@ -1,21 +1,84 @@
...
+
+    return isOnline;
+};
+
+/**
+ * When react-query resumes persisted mutations through their default functions (provided in the getOfflineFirstQueryClient file) after the browser tab
+ * has been closed, it cannot handle their side effects unless we set up some defaults. In order to leverage the react-admin notification system
+ * we add a default onSettled function to the mutation defaults here.
+ */
+const NotificationsFromQueryClient = () => {
```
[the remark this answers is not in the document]
slax57: Bump

slax57 (commented):
Only reviewed the docs and the simple demo so far...

docs/DataProviders.md
```diff
@@ -888,3 +888,133 @@ export default App;
...
+
+```ts
+// in src/queryClient.ts
+import { addOfflineSupportToQueryClient } from 'react-admin';
+import { QueryClient } from '@tanstack/react-query';
+import { dataProvider } from './dataProvider';
+
+export const queryClient = new QueryClient();
+
+const queryClientWithOfflineSupport = addOfflineSupportToQueryClient({
```
Madeorsk: What's the point of `queryClientWithOfflineSupport` variable? Why is it not exported instead of `queryClient`? Is `queryClient` mutated when you use `addOfflineSupportToQueryClient`, so exporting it is not required? Current variable naming makes me think that `queryClient` doesn't have offline support.

EDIT: in following examples, it's much clearer.

```suggestion
const baseQueryClient = new QueryClient();

export const queryClient = addOfflineSupportToQueryClient({
    queryClient,
    dataProvider,
    resources: ['posts', 'comments'],
});
```

examples/simple/src/comments/CommentList.tsx
```diff
@@ -63,9 +64,10 @@ const exporter: Exporter = (records, fetchRelatedRecords) =>
     });
 
 const CommentGrid = () => {
-    const { data } = useListContext();
+    const { data, isPaused, isPlaceholderData } = useListContext();
     const translate = useTranslate();
 
+    if (isPaused && (data == null || isPlaceholderData)) return <Offline />;
```
Madeorsk, line 70: Maybe add a quick example about this in docs? At least something mentioning that you can use `isPaused` and `isPlaceholderData` to detect that you're in offline mode.

examples/simple/src/posts/PostCreate.tsx
```diff
@@ -26,63 +26,113 @@ import {
...
+    Stack,
+} from '@mui/material';
+import MoreButton from '@mui/icons-material/MoreVert';
+
+// Client side id generation. We start from 100 to avoid querying the post list to get the next id as we
+// may be offline and accessing this page directly (without going through the list page first) which would
+// be possible if the app was also a PWA.
+// We only do that for optimistic and undoable modes in order to not break any existing tests that expect
+// the id to be generated by the server (e.g. by FakeRest).
+let next_id = 100;
```
Madeorsk, line 48: Why 100? You could technically already have another post with ID 100, isn't it?
For client-generated ids, I use negative numbers, to quickly notice that they are not real IDs, and is usually enough to ensure that they are not conflicting with persisted data.

packages/ra-core/src/controller/input/useReferenceArrayInputController.ts
```diff
@@ -61,6 +61,7 @@ export const useReferenceArrayInputController = <
         error: errorGetMany,
         isLoading: isLoadingGetMany,
         isFetching: isFetchingGetMany,
+        isPaused: isPausedGetMany,
         isPending: isPendingGetMany,
         refetch: refetchGetMany,
     } = useGetManyAggregate<RecordType>(
```
Madeorsk, line 68: No `isPlaceholderData` here? What's the difference with `useReferenceArrayFieldController`?

packages/ra-core/src/dataProvider/addOfflineSupportToQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+ *         }}
+ *     >
+ *         <Admin queryClient={queryClient} dataProvider={dataProvider}>
+ *             <Resource name="posts" {...posts} />
+ *             <Resource name="comments" {...comments} />
+ *         </Admin>
+ *     </PersistQueryClientProvider>
+ * );
+ *
+ * @example <caption>Adding offline support with custom mutations</caption>
```
Madeorsk, line 46: Aren't these examples mixed up? The first one shows with a custom mutation and the second one without it.

packages/ra-core/src/dataProvider/addOfflineSupportToQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * const localStoragePersister = createSyncStoragePersister({
+ *     storage: window.localStorage,
+ * });
+ *
+ * addOfflineSupportToQueryClient({
+ *    queryClient,
+ *   dataProvider,
+ *   resources: ['posts', 'comments'],
+ * });
```
Madeorsk, line 65: It would be better to have consistent examples. Either we use the query client returned by `addOfflineSupportToQueryClient`, or we do not.

packages/ra-ui-materialui/src/detail/EditView.tsx
```diff
@@ -24,13 +26,25 @@ export const EditView = (props: EditViewProps) => {
...
     const { hasShow } = useResourceDefinition();
-    const { resource, defaultTitle, record, isPending } = useEditContext();
+    const { resource, defaultTitle, record, isPending, isPaused } =
+        useEditContext();
+
+    if (isPaused && record == null && offline) {
+        return (
+            <Root className={clsx('edit-page', className)} {...rest}>
+                <div className={clsx(EditClasses.main, EditClasses.noActions)}>
+                    <Content className={EditClasses.card}>{offline}</Content>
```
Madeorsk, line 43: Shouldn't you add `EditClasses.offline` somewhere?

packages/ra-ui-materialui/src/detail/ShowView.tsx
```diff
@@ -23,13 +25,24 @@ export const ShowView = (props: ShowViewProps) => {
...
 
-    const { resource, defaultTitle, record } = useShowContext();
+    const { resource, defaultTitle, isPaused, record } = useShowContext();
     const { hasEdit } = useResourceDefinition();
 
+    if (isPaused && record == null && offline) {
+        return (
+            <Root className={clsx('show-page', className)} {...rest}>
+                <div className={clsx(ShowClasses.main, ShowClasses.noActions)}>
+                    <Content className={ShowClasses.card}>{offline}</Content>
```
Madeorsk, line 40: Shouldn't you add `ShowClasses.offline` somewhere?

packages/ra-ui-materialui/src/field/ReferenceArrayField.tsx
```diff
@@ -149,11 +149,11 @@ export const ReferenceArrayFieldView = (
     props: ReferenceArrayFieldViewProps
 ) => {
     const { children, pagination, className, sx } = props;
-    const { isPending, total } = useListContext();
+    const { isPending, isPaused, total } = useListContext();
```
Madeorsk, line 152: `isPlaceholderData`?

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -94,8 +97,19 @@ export const ReferenceArrayInput = (props: ReferenceArrayInputProps) => {
         sort,
         filter,
     });
+    const { isPaused, allChoices } = controllerProps;
```
Madeorsk, line 100: `isPlaceholderData`?

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -94,8 +97,19 @@ export const ReferenceArrayInput = (props: ReferenceArrayInputProps) => {
...
 
-    return (
+    return isPaused && allChoices == null ? (
+        offline ?? (
+            <Labeled
```
Madeorsk: Shouldn't we use `<Labeled>` around `offline` variable too?

packages/ra-ui-materialui/src/input/ReferenceInput.tsx
```diff
@@ -64,15 +66,32 @@ import { AutocompleteInput } from './AutocompleteInput';
...
         );
     }
 
-    return <ReferenceInputBase {...rest}>{children}</ReferenceInputBase>;
+    return (
+        <ReferenceInputBase
+            {...rest}
+            offline={
+                offline ?? (
+                    <Labeled
```
Madeorsk: Shouldn't we use `<Labeled>` around `offline` variable too?

packages/ra-ui-materialui/src/Offline.tsx
```diff
@@ -0,0 +1,90 @@
...
+    },
+}));
+
+declare module '@mui/material/styles' {
+    interface ComponentNameToClassKey {
+        RaOffline: 'root';
+    }
+
+    interface ComponentsPropsList {
+        RaOffline: Partial<Offline>;
```
Madeorsk: Use `[PREFIX]` instead of `'RaOffline'`.

packages/ra-ui-materialui/src/Offline.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    interface Components {
+        RaOffline?: {
+            defaultProps?: ComponentsPropsList['RaOffline'];
```
Madeorsk: Use `typeof PREFIX` instead of `'RaOffline'`.

packages/ra-ui-materialui/src/input/ReferenceInput.stories.tsx
```diff
@@ -109,6 +109,54 @@ export const Basic = ({ dataProvider = dataProviderWithAuthors }) => (
...
+    >
+        <SimpleForm>
+            <LoadChildrenOnDemand>
+                <ReferenceInput reference="authors" source="author" />
+            </LoadChildrenOnDemand>
+        </SimpleForm>
+    </Edit>
+);
+
+export const Offline = ({ dataProvider = dataProviderWithAuthors }) => (
```
Madeorsk: Tested it and the default `AutocompleteInput` shows "Loading..." instead of the offline component that I expected. I think you need to change `useChoicesContext` to propagate `isPaused` and `isPlaceholderData` to show a better message in offline mode.

packages/ra-ui-materialui/src/layout/LoadingIndicator.tsx
```diff
@@ -30,18 +38,51 @@ export const LoadingIndicator = (inProps: LoadingIndicatorProps) => {
...
-                    {...rest}
-                />
-            )}
+            {loading ? (
+                isOffline ? (
+                    <Tooltip
+                        title={
+                            <Translate
+                                i18nKey="ra.notifications.pending_operations"
+                                options={{ smart_count: 1 }}
```
Madeorsk: Why `{ smart_count: 1 }`? Shouldn't it be `{ smart_count: pendingMutations.length }`?

packages/ra-ui-materialui/src/list/SingleFieldList.stories.tsx
```diff
@@ -106,21 +110,27 @@ export const LinkType = () => (
...
 
 export const Loading = () => (
-    <ListContextProvider value={{ isLoading: true } as any}>
+    <Wrapper listContext={{ isPending: true }}>
+        <SingleFieldList />
+    </Wrapper>
+);
+
+export const Offline = () => (
+    <Wrapper listContext={{ isPaused: true }}>
```
Madeorsk: This story shows nothing. Is it supposed to show the offline component?

examples/simple/src/posts/PostCreate.tsx
djhi: But they are real ids. They won't be replaced on the backend. I would use uuid but that's too much changes of the current data. Besides, we actually control the simple example data and know we don't have ids greater than 100. I believe this is enough to showcase and test the feature

examples/simple/src/Layout.tsx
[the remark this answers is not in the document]
djhi: Added a documentation section for this in `DataProviders.md`

yarn.lock
```diff
@@ -4319,13 +4355,39 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@tanstack/query-core@npm:5.81.2":
+  version: 5.81.2
+  resolution: "@tanstack/query-core@npm:5.81.2"
+  checksum: 36a6bddec2e7512015bcfbb0d7b0876fab418de9e0ef21ad403598276960e0b7d53efd62832ce462738ad22d9883e31cb5403eafc65dfd9b2f6744c22a9d8e42
+  languageName: node
+  linkType: hard
```
slax57: needs dedupe

docs/DataProviders.md
```diff
@@ -888,3 +888,170 @@ export default App;
...
+queryClient.setMutationDefaults('banUser', {
+    mutationFn: async (userId) => {
+        return dataProvider.banUser(userId);
+    },
+});
+```
+
+## Handling Errors For Resumed Mutations
+
+If you enabled offline support, users might trigger mutations while being actually offline. When they're back online, react-query will _resume_ those mutations and they might fail for other reasons (server side validation or errors). However, as users might have navigated away from the page that triggered the mutation, they won't see any notification.
```
slax57:
```suggestion
If you enabled offline support, users might trigger mutations while being actually offline. When they're back online, TanStack Query will _resume_ those mutations and they might fail for other reasons (server side validation or errors). However, as users might have navigated away from the page that triggered the mutation, they won't see any notification.
```

packages/ra-core/src/core/useIsOffine.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const useIsOffine = () => {
+    const [isOnline, setIsOnline] = React.useState(onlineManager.isOnline());
+
+    React.useEffect(() => {
+        const handleChange = () => {
+            setIsOnline(onlineManager.isOnline());
+        };
+        return onlineManager.subscribe(handleChange);
+    }, []);
```
slax57, line 18: This code looks odd... Shouldn't you subscribe on mount and unsubscribe on unmount? :thinking:

docs/DataProviders.md
```diff
@@ -888,3 +888,170 @@ export default App;
...
+To handle this scenario, you must register default `onError` side effects for all mutations (react-admin default ones or custom). If you want to leverage react-admin notifications, you can use a custom layout:
+
+```tsx
+// in src/Layout.tsx
+export const MyLayout = ({ children }: { children: React.ReactNode }) => {
+    const queryClient = useQueryClient();
+    const notify = useNotify();
+
+    React.useEffect(() => {
+        const mutationKeyFilter = []; // An empty array targets all mutations
```
slax57, line 1035: `mutationKeyFilter` is used nowhere I believe :thinking:

packages/ra-core/src/dataProvider/useCreate.ts
```diff
@@ -96,7 +96,11 @@ export const useCreate = <
         getMutateWithMiddlewares,
         ...mutationOptions
     } = options;
+
     const mode = useRef<MutationMode>(mutationMode);
+    useEffect(() => {
+        mode.current = mutationMode;
+    }, [mutationMode]);
```
slax57, line 103: Why do you need that? :thinking: 

Why only in useCreate? :thinking:

packages/ra-ui-materialui/src/Offline.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const Offline = (props: Offline) => {
+    const { icon, message: messageProp, variant = 'standard', ...rest } = props;
+    const translate = useTranslate();
+    const resource = useResourceContext(props);
+    const getResourceLabel = useGetResourceLabel();
+    if (!resource) {
+        throw new Error(
+            '<Offline> must be used inside a <Resource> component or provided a resource prop'
+        );
+    }
```
slax57, line 21: This limitation seems odd. If it is only required for the translation string, I think there are easy solutions to remove this limitation.

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -94,8 +97,17 @@ export const ReferenceArrayInput = (props: ReferenceArrayInputProps) => {
...
+    const { isPaused, allChoices } = controllerProps;
 
-    return (
+    return isPaused && allChoices == null ? (
+        <Labeled
+            source={props.source}
+            label={props.label}
+            resource={props.resource}
+        >
+            {offline ?? <Offline variant="inline" />}
```
slax57, line 108: why inline? :thinking:

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
```diff
@@ -91,11 +94,20 @@ export const ReferenceOneField = <
...
+            !controllerProps.isPaused &&
+            controllerProps.referenceRecord == null)
+    ) {
+        return empty;
+    }
+
+    if (controllerProps.isPaused && controllerProps.referenceRecord == null) {
+        return offline;
+    }
+
```
slax57, line 109: It would be nice to have a story to test this component too :innocent:

packages/ra-ui-materialui/src/detail/EditView.tsx
djhi: This class has been removed

packages/ra-ui-materialui/src/detail/ShowView.tsx
djhi: This class has been removed

PR #10549, opened 2025-02-25
Make `<TabbedForm tabs>` use the `onChange` prop

### LINKED ISSUE #10532: TabbedForm: I cannot get the tabindex somehow when I click on the tabs
**What you were expecting:**
`TabbedForm` onChange logic does not seem to work, so that I can get the current tabIndex (unless there is another way to achieve this)

**What happened instead:**
The `onChange` function is not triggered

**Steps to reproduce:**
This is how I have the logic setup:

```
    <TabbedForm
      sanitizeEmptyValues
      defaultValues={getDefaultValues({})}
      tabs={
        <TabbedFormTabs
          onChange={(tabIndex) => {
            console.log('log1-tabIndex2', tabIndex);
            setActiveTab(tabIndex);
          }}
        />
      }
    >
// ... more code here, with tabs, etc.
```

I believe the `onChange` will never be triggered, because you override this with a specific `handleChange` function which does not have a `onChange()` call at the end:

Image

**Environment**

* React-admin version: "^5.2.1"
* React version: "^18.3.1"
* Browser: N/A

erwanMarmelab:
Fixes #10532

### DIFF
--- a/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
+++ b/packages/ra-ui-materialui/src/form/TabbedFormView.tsx
@@ -37,6 +37,9 @@ export const TabbedFormView = (props: TabbedFormViewProps): ReactElement => {
         if (!syncWithLocation) {
             setTabValue(value);
         }
+        if (tabs.props.onChange) {
+            tabs.props.onChange(event, value);
+        }
     };
 
     const renderTabHeaders = () =>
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/form/TabbedFormView.tsx
```diff
@@ -37,6 +37,9 @@ export const TabbedFormView = (props: TabbedFormViewProps): ReactElement => {
         if (!syncWithLocation) {
             setTabValue(value);
         }
+        if (tabs.props.onChange) {
```
djhi, line 40: We really should rewrite this component with contexts.

djhi (approved):

PR #10552, opened 2025-02-27
[TypeScript] Fix buttons `icon` prop type

djhi:
## Problem

There's no reason to not accept a `ReactNode` for the `icon` prop. This would allow to pass `null` to remove the icon altogether.

## Solution

Fix the icon prop type.

## Additional Checks


fzaninotto:
IMHO, that's a new feature that should target `next` and be documented.

fzaninotto (approved):

PR #10553, opened 2025-02-27
[TypeScript] Remove usage of JSX.Element

djhi:
## Problem

`JSX.Element` is type JSX which is not tied to react (JSX can be used in other frameworks). We should use the correct types from React.

## Solution

- Replace usage of `JSX.Element` by their React equivalent.
- Use `ReactNode` instead of `ReactElement` where it makes sense.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/i18n/useTranslateLabel.ts
+++ b/packages/ra-core/src/i18n/useTranslateLabel.ts
@@ -1,4 +1,4 @@
-import { useCallback, ReactElement } from 'react';
+import { useCallback } from 'react';
 
 import { useTranslate } from './useTranslate';
 import { getFieldLabelTranslationArgs } from '../util';
@@ -16,7 +16,7 @@ export const useTranslateLabel = () => {
             resource,
         }: {
             source?: string;
-            label?: string | false | ReactElement;
+            label?: React.ReactNode;
             resource?: string;
         }) => {
             if (label === false || label === '') {
--- a/packages/ra-core/src/i18n/useTranslateLabel.ts
+++ b/packages/ra-core/src/i18n/useTranslateLabel.ts
@@ -1,4 +1,4 @@
-import { useCallback } from 'react';
+import { useCallback, ReactNode } from 'react';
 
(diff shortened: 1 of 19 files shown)


packages/ra-core/src/i18n/useTranslateLabel.ts
```diff
@@ -16,7 +16,7 @@ export const useTranslateLabel = () => {
             resource,
         }: {
             source?: string;
-            label?: string | false | ReactElement;
+            label?: React.ReactNode;
```
fzaninotto, line 19: `React` is not in scope. You should import `ReactNode` line 1.

packages/ra-ui-materialui/src/layout/AppBar.tsx
```diff
@@ -114,8 +114,8 @@ export interface AppBarProps extends MuiAppBarProps {
      */
     alwaysOn?: boolean;
     container?: React.ElementType<any>;
-    toolbar?: JSX.Element;
-    userMenu?: JSX.Element | boolean;
+    toolbar?: React.ReactNode;
+    userMenu?: React.ReactNode | boolean;
```
fzaninotto, line 118: Doesn't `ReactNode` already cover `boolean`?

packages/ra-ui-materialui/src/list/datagrid/types.ts
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { TableCellProps } from '@mui/material';
+
+/**
+ * This type is used to avoid relying on defaultProps which is deprecated.
+ */
+export type DatagridField = React.ReactElement & {
```
fzaninotto, line 7: this new type is never reexported

fzaninotto (changes requested):

packages/ra-ui-materialui/src/list/datagrid/types.ts
djhi: I'm not sure we should

PR #10555, opened 2025-02-27
Fix offline user experience

slax57:
## Problem

Our current behavior when users are offline is to show the available data even though it does not match the page parameters. For instance, you loaded the first page of a list then lose your connection but click on the _Go to page 2_ pagination button: you’ll see the first page anyway with no indication that we couldn’t load the requested data.

## Solution

In `useListController`, detect that we are in fetch mode and in pause (offline), and in that case show a notification. This will improve the user experience.

## Todo

- [x] Story
- [x] Dev
- [x] i18n
- [x] Tests

## How To Test

Story: http://localhost:9010/?path=/story/ra-core-controller-list-uselistcontroller--offline

Demo: `make run`
In the demo, you can use the Tanstack Query Devtools to simulate offline mode.

Screenshot_20250227_182525

## Additional Checks

## Screenshots

Screenshot_20250227_182124

### DIFF
--- a/packages/ra-language-english/src/index.ts
+++ b/packages/ra-language-english/src/index.ts
@@ -118,8 +118,7 @@ const englishMessages: TranslationMessages = {
             unsaved_changes:
                 "Some of your changes weren't saved. Are you sure you want to ignore them?",
             yes: 'Yes',
-            placeholder_data_warning:
-                'Could not fetch: lost connection to server',
+            placeholder_data_warning: 'Network issue: Data refresh failed.',
         },
         navigation: {
             clear_filters: 'Clear filters',
--- a/packages/ra-language-english/src/index.ts
+++ b/packages/ra-language-english/src/index.ts
@@ -118,7 +118,8 @@ const englishMessages: TranslationMessages = {
             unsaved_changes:
                 "Some of your changes weren't saved. Are you sure you want to ignore them?",
             yes: 'Yes',
-            placeholder_data_warning: 'Could not fetch the requested data',
+            placeholder_data_warning:
+                'Could not fetch: lost connection to server',
         },
         navigation: {
             clear_filters: 'Clear filters',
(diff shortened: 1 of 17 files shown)


packages/ra-language-english/src/index.ts
```diff
@@ -118,6 +118,7 @@ const englishMessages: TranslationMessages = {
             unsaved_changes:
                 "Some of your changes weren't saved. Are you sure you want to ignore them?",
             yes: 'Yes',
+            placeholder_data_warning: 'Could not fetch the requested data',
```
fzaninotto, line 121: I'd be more explicit with this message, mentioning lost connection to the server

slax57: I changed it to `'Could not fetch: lost connection to server'`. Is it better?

guilbill (approved):
Simple, easy to read, well tested, what else? JB quality 😉

packages/ra-language-english/src/index.ts
```diff
@@ -118,6 +118,8 @@ const englishMessages: TranslationMessages = {
             unsaved_changes:
                 "Some of your changes weren't saved. Are you sure you want to ignore them?",
             yes: 'Yes',
+            placeholder_data_warning:
+                'Could not fetch: lost connection to server',
```
erwanMarmelab, line 122: **suggestion:** (open to discussion) ‏
```suggestion
                'Could not fetch data: lost connection to server',
```

packages/ra-language-french/src/index.ts
```diff
@@ -122,6 +122,8 @@ const frenchMessages: TranslationMessages = {
             unsaved_changes:
                 "Certains changements n'ont pas été enregistrés. Êtes-vous sûr(e) de vouloir quitter cette page ?",
             yes: 'Oui',
+            placeholder_data_warning:
+                'Récupération impossible: connexion au serveur perdue',
```
erwanMarmelab, line 126: **suggestion:** (open to discussion) ‏
```suggestion
                'Récupération impossible des données: connexion au serveur perdue',
```

erwanMarmelab (changes requested):

fzaninotto:
Suggestion for the error message:

> Network issue: Data refresh failed.
>
> Problème réseau : Impossible de charger les données.

guilbill (approved):

PR #10556, opened 2025-02-28
Fix create-react-admin generated package.json when using yarn

djhi:
## Problem

For some reason, yarn v1 install multiple versions of the dependencies for which we support multiple major versions.
This leads to an error page as soon as you run the generated application.

## Solution

When the chosen installer is `yarn`, detect the installed version and if it is v1, add resolutions to the generated `package.json` file

## How To Test

- `make build-create-react-admin install`
- `cd ..` Required to avoid yarn errors when it runs in our monorepo
- `./react-admin/.bin/react-admin/node_modules/.bin/create-react-admin myadmin --install yarn`
- `cd myadmin`
- `cat package.json` check resolutions are present
- `yarn run dev` check the app runs

## Additional Checks

### DIFF
--- a/packages/create-react-admin/src/generateProject.ts
+++ b/packages/create-react-admin/src/generateProject.ts
@@ -2,6 +2,7 @@ import path from 'path';
 import fs from 'fs';
 import fsExtra from 'fs-extra';
 import url from 'url';
+import execa from 'execa';
 import merge from 'lodash/merge.js';
 import { ProjectConfiguration } from './ProjectState.js';
 import { generateAppFile } from './generateAppFile.js';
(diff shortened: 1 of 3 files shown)


slax57 (approved):
Fix works well.

But I realize we're still missing some pieces to properly support yarn version >= 3:
- it complains about the project not being in a workspace, and we need to add an empty `yarn.lock` file to fix the issue
- it installs dependencies in Plug n Play mode by default, which is not compatible with our current Vite config. The fix is to set `nodeLinker: node-modules` in `.yarnrc.yml` file

Do we want to fix these issues in this PR too?

slax57 (approved):

PR #10559, opened 2025-03-04
[typescript] Add SxProps to Error component

smeng9:
The styled component in error.tsx should accept sxprops to customize its styles


fzaninotto (approved):

fzaninotto:
Thanks!

PR #10591, opened 2025-03-13
feat: handle empty `emptyText` for `<ChipField>`

guilbill:
## Problem

When passing an empty value (undefined or null), a `<Chip>` is rendered with nothing in it.

## Solution

If the value is an empty string, render the emptyText instead (just like the DateField does).

If the emptyText is an empty string, the component should return null.

## How To Test

- Unit tests should be green.
- https://react-admin-storybook-7jlxn1ynw-marmelab.vercel.app/?path=/story/ra-ui-materialui-fields-chipfield--basic

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/field/ChipField.spec.tsx
+++ b/packages/ra-ui-materialui/src/field/ChipField.spec.tsx
@@ -27,12 +27,7 @@ const i18nProvider = polyglotI18nProvider(
 describe('<ChipField />', () => {
     it('should display the record value added as source', () => {
         const { getByText } = render(
-            <ChipField
-                className="className"
-                classes={{}}
-                source="name"
-                record={{ id: 123, name: 'foo' }}
-            />
+            <ChipField source="name" record={{ id: 123, name: 'foo' }} />
         );
         expect(getByText('foo')).not.toBeNull();
     });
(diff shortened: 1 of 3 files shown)


erwanMarmelab (approved):

packages/ra-ui-materialui/src/field/ChipField.tsx
```diff
@@ -18,8 +18,8 @@ const ChipFieldImpl = <
     const value = useFieldValue(props);
     const translate = useTranslate();
 
-    if (value == null && emptyText) {
-        return (
+    if (!value) {
+        return emptyText && emptyText.length > 0 ? (
```
erwanMarmelab, line 22: **suggestion:**
```suggestion
        if (!emptyText || emptyText.length === 0) return null;
        return (
```

packages/ra-ui-materialui/src/field/ChipField.tsx
```diff
@@ -28,7 +28,7 @@ const ChipFieldImpl = <
             >
                 {emptyText && translate(emptyText, { _: emptyText })}
             </Typography>
-        );
+        ) : null;
```
erwanMarmelab, line 31:
```suggestion
        );
```

erwanMarmelab (approved):

packages/ra-ui-materialui/src/field/ChipField.spec.tsx
```diff
@@ -89,4 +89,42 @@ describe('<ChipField />', () => {
 
         expect(getByText('Not found')).not.toBeNull();
     });
+
+    it('should return null when value and emptyText are an empty string', () => {
+        const { container } = render(
+            <ChipField
+                className="className"
+                classes={{}}
```
slax57, line 97: why do you need `className` and `classes`? :thinking:

slax57 (approved):

PR #10594, opened 2025-03-17
Add `<Translate>` component

erwanMarmelab:
## BC

The `Translate` type exported by `ra-core` was renamed `TranslateFunction`

## Problem

To output text, developers must use the `useTranslate` hook. It can be cumbersome.

## To Be Defined

A. args:

1.  `<Translate i18nKey="my-key" smart_count="18" />`
2.  `<Translate i18nKey="my-key" args={{smart_count:"18"}} />`

B. New name, we already have a type named `Translate`

## Solution

A new `ra-core` component to translate text.
```
<Translate i18nKey="ra.action.delete">Delete</Translate>
```
The default value is the child. The component should also accept translation options.
Test with interpolation, with `i18next` and `polyglot`.

## How To Test

https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-core-i18n-translate
https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-i18n-polyglot--translate-component
https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-i18n-i18next--translate-component

## Additional Checks

### DIFF
--- a/docs/Translate.md
+++ b/docs/Translate.md
@@ -1,0 +1,113 @@
+---
+layout: default
+title: "The Translate Component"
+---
+
+# `<Translate>`
+
+If you need to translate messages in your own components, React-admin provides the `<Translate>` component which displays your translated messages.
+
+# Usage
+
+```tsx
+const MyHelloButton = () => (
+    <Translate 
+        i18nKey="custom.hello_world"
+        component="button"
+    />
+);
+
+export default MyHelloButton;
+```
+
+**Tip:** You can directly use the translation function with [the `useTranslate` hook](./useTranslate.md).
+
+## Props
+
+| Prop        | Required | Type                    | Default            | Description                                             |
+| ----------- | -------- | ----------------------- | ------------------ | ------------------------------------------------------- |
+| `args`      | Optional | `Object`                | -                  | The arguments used for pluralization and interpolation. |
+| `children`  | Optional | `string`                | -                  | The default translation if the translation failed.      |
+| `component` | Optional | `ElementType`           | `div`              | The component to render as the root element.            |
+| `empty`     | Optional | `string` &#124; `false` | `"no translation"` | Message to be displayed if there is no translation.     |
+| `i18nKey`   | Required | `string`                | -                  | The translation key.                                    |
+
+Additional props are passed to the root element.
+
+## `args`: Pluralization and Interpolation
+
+Polyglot.js provides some nice features such as interpolation and pluralization, that you can use in react-admin.
+
+{%raw%}
+
+```tsx
+// in english.ts
+import englishMessages from 'ra-language-english';
+
+const messages = {
+    ...englishMessages,
+    custom: {
+        my_key: 'My Translated Key',
+        hello_world: 'Hello, %{my_world}!',
+        count_beer: 'Select one beer |||| Select %{smart_count} beers',
+    },
+};
+
+export default messages;
+```
+
+```tsx
+export const MyHelloButton = () => (
+    <Translate 
+        i18nKey="custom.hello_world"
+        component="button"
+        args={{ my_world: 'world' }}
+    />
+);
+
+export const SelectBeerButton = () => (
+    <Translate 
+        i18nKey="custom.count_beer"
+        component="button"
+        args={{ smart_count: 2 }}
+    />
+);
+```
+
+{%endraw%}
+
+## `children`
+
+You can provide a `children` to display if the translation function doesn't find a message with your `i18nKey`.
+
+```tsx
+const LoadingMessage = () => (
+    <Translate i18nKey="ra.page.loading" component="h3">
+        Loading
+    </Translate>
+);
+```
+
+## `component`
+
+The component to render as the root element.
+
+```tsx
+const LoadingMessage = () => <Translate i18nKey="ra.page.loading" component="h3" />;
+const MyHelloButton = () => <Translate i18nKey="custom.hello_world" component="button" />;
+const MarkedMessage = () => <Translate i18nKey="custom.myKey" component="mark" />;
+```
+
+## `empty`
+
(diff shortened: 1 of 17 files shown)


packages/ra-core/src/i18n/Translate.tsx
```diff
@@ -0,0 +1,32 @@
...
+    if (translatedMessage && translatedMessage !== i18nKey) {
+        return React.createElement(component, {}, translatedMessage);
+    }
+    if (children) {
+        return React.createElement(component, {}, children);
+    }
+    if (empty) {
+        return React.createElement(component, {}, empty);
+    }
+    return null;
```
djhi, line 23: I wouldn't even put it inside anything but a `Fragment` like we do for `ValidationError`:

```tsx
if (translatedMessage && translatedMessage !== i18nKey) {
        return <>{translatedMessage}</>;
}
```

djhi (changes requested):
We don't need the `component` prop

docs/Translate.md
```diff
@@ -0,0 +1,89 @@
...
+| `args`      | Optional | `Object`                | -                  | The arguments used for pluralization and interpolation. |
+| `children`  | Optional | `string`                | -                  | The default translation if the translation failed.      |
+| `empty`     | Optional | `string` &#124; `false` | `"no translation"` | Message to be displayed if there is no translation.     |
+| `i18nKey`   | Required | `string`                | -                  | The translation key.                                    |
+
+Additional props are passed to the root element.
+
+## `args`: Pluralization and Interpolation
+
+Polyglot.js provides some nice features such as interpolation and pluralization, that you can use in react-admin.
```
djhi, line 33: Don't mention polyglot, that's not the only i18n provider

packages/ra-core/src/i18n/Translate.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useTranslate } from './useTranslate';
+
+export const Translate = ({
+    i18nKey,
+    args,
+    children = 'no translation',
```
fzaninotto, line 7: I think it should default to empty striong if no translation and no default value.

packages/ra-core/src/i18n/Translate.tsx
```diff
@@ -0,0 +1,22 @@
...
+}: TranslateProps) => {
+    const translate = useTranslate();
+    const translatedMessage = translate(i18nKey, args);
+
+    if (translatedMessage && translatedMessage !== i18nKey) {
```
fzaninotto, line 12: Why the second condition? Please add a comment to explain it.

packages/ra-core/src/i18n/useTranslate.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback } from 'react';
 
-import { Translate } from '../types';
+import { TranslateFunction } from '../types';
```
fzaninotto, line 3: This is technically a BC break and it should appear front and foremost in the PR description

packages/ra-core/src/i18n/Translate.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <TestTranslationProvider messages={{}}>
+        <Translate i18nKey="custom.myKey">My Default Translation</Translate>
+    </TestTranslationProvider>
+);
+
+export const Args = () => (
+    <TestTranslationProvider
+        messages={{
+            custom: {
+                myKey: ({ price }) => `It cost ${price}.00 $`,
```
fzaninotto, line 37: Using functions is unusual. Can you make a test with interpolation instead?

https://marmelab.com/react-admin/TranslationTranslating.html#interpolation-pluralization-and-default-translation

packages/ra-core/src/i18n/Translate.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useTranslate } from './useTranslate';
+
+export const Translate = ({
+    i18nKey,
+    args,
+    children = 'no translation',
+}: TranslateProps) => {
+    const translate = useTranslate();
+    const translatedMessage = translate(i18nKey, args);
```
fzaninotto, line 10: I'd use the children as the `_` arg to make it the default.

docs/useTranslate.md
```diff
@@ -14,6 +14,8 @@ const translate = useTranslate();
 const translatedMessage = translate(translationKey, options);
 ```
 
+**Tip:** Instead of working with a hook, you can directly use [the `<Translate>` component](./Translate.md).
```
fzaninotto, line 17:
```suggestion
**Tip:** Instead of a hook, you can use [the `<Translate>` component](./Translate.md).
```

packages/ra-core/src/i18n/Translate.stories.tsx
erwanMarmelab: I can't. those functionalities are available with polyglot and i18next. We can't use those packages without in ra-core

packages/ra-i18n-i18next/README.md
```diff
@@ -29,7 +30,7 @@ const App = () => {
         options: {
             resources: {
                 en: {
-                    translations: convertRaTranslationsToI18next(englishMessages)
+                    translation: convertRaTranslationsToI18next(englishMessages)
```
djhi, line 33: why?

djhi (changes requested):

packages/ra-i18n-i18next/README.md
djhi: After investigation, you are correct and the doc is wrong. `translation` is the default namespace in i18next (see https://www.i18next.com/overview/configuration-options)

djhi (approved):

erwanMarmelab:
wrong based branch -> new PR: #10605

PR #10597, opened 2025-03-18
Add DataTable component

fzaninotto:
## Problem

`<Datagrid>` is fine, but since it uses child inspection, it leads to unexpected bugs when developers wrap children with custom components. For instance, it's not possible to use `<CanAccess>` to wrap a field:

```jsx
const PostList = () => (
  <List>
    <Datagrid>
      <TextField source="title" />
      <CanAccess action="read" resource="posts.views">
        <NumberField source="nb_views" />
      </CanAccess>
   </Datagrid>
 </List>
);
```

Besides, the signature of all field components is polluted by props that only serve in datagrid columns (e.g. sortBy, label, etc). In fact, field components mix 2 purposes: datagrid columns and record field formatting. 

## Solution

An alternative data table component named `<DataTable>` that no longer uses child inspection.

The following works as expected:

```jsx
const PostList = () => (
  <List>
    <DataTable>
      <DataTable.Col source="title" />
      <DataTable.Col source="published_at" field={DateField} />
      <CanAccess action="read" resource="posts.views">
        <DataTable.Col source="nb_views" field={NumberField} />
      </CanAccess>
   </DataTable>
 </List>
);
```

The new syntax covers 100% of the `<Datagrid>` features, and supports various new features. 

Pros:

- IDE-friendly, more discoverable syntax
- Allows wrapped columns (no WTF)
- Easier column styling (via `<DataTable.Col sx>`)
- Conditional formatting (via `<DataTable.Col cellSx>`)
- Built-in support for FunctionField (via `<DataTable.Col render>`)
- Support for column selector (no need for `<ConfigurableDatagrid>`)
- Optimized by default (no need for `optimized` prop)
- More type-safety than `Datagrid`
- The `sortBy` prop is no longer necessary (it's the `<DataTable.Col source>`)
- Allows to build custom reusable columns (e.g., `<DataTable.NumberCol>`)
- Fields no longer need to support column-only props (`sortable`, `sortByOrder`, `label`, etc). 
- Expandable rows now animate on expand
- In standalone mode, it's no longer necessary to set `bulkActionButtons={false}`

Cons:

- More characters to type
- In some cases, developers have to repeat the `source` both in the `<DataTable.Col>` and in the `Field`
- No backward compatibility, the two components will have to live side-by-side for some time

## How To Test

I've added a storybook for `DataTable` and `DataTable.Col`.

## Todos

- [x] Add `<DataTable>` component
- [x] Add `<DataTable.Col>` component
- [x] Add `<ColumnsButton>` component
- [x] Migrate demos to `<DataTable>`
- [x] Add documentation for `<DataTable>`

## Out of cope

- Make `<DataTable>` the recommended datagrid component yet (we'll wait for feedback first). 
- Replace every `<Datagrid>` in documentation with `<DataTable>`
- Replace every `<Datagrid>` in stories with `<DataTable>`

## Additional Checks

- [x] The PR targets `next` for a feature


docs/DataTable.md
```diff
@@ -0,0 +1,1804 @@
...
+| `cellSx`          | Optional | function                 | -                     | A function that returns the sx prop to apply to a cell.                             |
+| `className`          | Optional | string                  | -                     | The class name of the column.                                 |
+| `field`              | Optional | Component               | -                     | The field component to use for the column.                    |
+| `headerClassName`    | Optional | string                  | -                     | The class name of the header cell.                            |
+| `label`              | Optional | string                  | -                     | The column label, displayed in the header.                    |
+| `render`             | Optional | (record) => ReactNode | -               | A function to render the column content.                      |
+| `sortable`           | Optional | boolean                 | true                  | Whether the column is sortable.                               |
```
djhi, line 944: Same, I'd rather have a `disableSort` prop

docs/DataTable.md
```diff
@@ -0,0 +1,1804 @@
...
+![DataTable Column Type Safety](./img/DataTableColumnType.png)
+
+The most convenient way to benefit from this capability is to alias column components for your resource:
+
+```tsx
+import { List, DataTable, ReferenceField } from 'react-admin';
+import { type Review } from '../types';
+
+const Column = DataTable.Col<Review>;
+const ColumnNumber = DataTable.NumberCol<Review>;
```
djhi, line 1762: This one isn't used in the list below

examples/demo/src/orders/OrderList.tsx
```diff
@@ -152,169 +154,127 @@ const TabbedDatagrid = () => {
...
+        />
+        <ColumnNumber source="total_ex_taxes" options={currencyStyle} />
+        <ColumnNumber source="delivery_fees" options={currencyStyle} />
+        <ColumnNumber source="taxes" options={currencyStyle} />
+        <ColumnNumber
+            source="total"
+            options={currencyStyle}
+            sx={{ fontWeight: 'bold' }}
+        />
+    </DataTable>
```
djhi, line 206: If I'm not mistaken, all tables have the same columns except `DeliveredOrdersTable` that adds the `returned` column. Can't we have a `CommonColumn` const that has a `Fragment` containing the common cols. That's a good test and showcase of what the new `DataTable` allows too!

packages/ra-ui-materialui/src/list/datatable/DataTable.stories.tsx
```diff
@@ -0,0 +1,878 @@
...
+import fakeRestDataProvider from 'ra-data-fakerest';
+import {
+    CanAccess,
+    ListContextProvider,
+    Resource,
+    ResourceContextProvider,
+    useRecordContext,
+    useGetList,
+    useList,
+    type SortPayload,
```
djhi, line 11: linter warning

djhi (changes requested):
Awesome work! The DX improvement is great. I noticed that only the `DataTable` component can be targeted in the theme. Is it expected? Also, it would be great to support default props override through theme in all our new components (https://mui.com/material-ui/customization/creating-themed-components/#4-support-theme-default-props).

cypress/e2e/list.cy.js
```diff
@@ -254,7 +254,7 @@ describe('List Page', () => {
                 .should(el => expect(el).to.have.attr('aria-expanded', 'true'))
                 .should(el => expect(el).to.have.attr('aria-label', 'Close'));
 
-            cy.get('#13-expand').should(el =>
+            cy.get('#posts-13-expand').should(el =>
```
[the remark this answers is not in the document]
fzaninotto: ¯\_(ツ)_/¯

docs/DataTable.md
```diff
@@ -0,0 +1,1804 @@
...
+| `bulkActionToolbar`  | Optional | Element                 | `<BulkDelete Button>` | The component rendering the bulk action toolbar.         |
+| `empty`              | Optional | Element                 | `<Empty>`             | The component to render when the list is empty.                 |
+| `expand`             | Optional | Element                 |                       | The component rendering the expand panel for each row.   |
+| `expandSingle`       | Optional | Boolean                 | `false`               | Whether to allow only one expanded row at a time.             |
+| `header`             | Optional | Element                 | `<DataTable Header>`   | The component rendering the table header.                |
+| `hiddenColumns`      | Optional | Array                   | `[]`                  | The list of columns to hide by default.                          |
+| `footer`             | Optional | Element                 |                       | The component rendering the table footer.                |
+| `hover`              | Optional | Boolean                 | `true`                | Whether to highlight the row under the mouse.                 |
```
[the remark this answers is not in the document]
fzaninotto: This is actually a MUI `TableRow` prop, Id rather keep the same syntax. 

https://mui.com/material-ui/api/table-row/

docs/DataTable.md
```diff
@@ -0,0 +1,1804 @@
...
+            ...
+        </MyDataTable>
+    </List>
+);
+
+export default PostList;
+```
+{% endraw %}
+
+**Tip**: `<DataTable>` already sets `RaDataTable-rowEven` and `RaDataTable-rowOdd` classes on the rows, so you don't need to use a custom DataTable body to implement zebra stripes. You can just use [the `sx` prop](#sx-css-api) to set the background color of these classes.
```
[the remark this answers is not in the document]
fzaninotto: There is also a better way to do what you describe (cf the `rowSx` doc). I don't have any other example in mind.

examples/demo/src/orders/OrderList.tsx
fzaninotto: fixed

packages/ra-ui-materialui/src/list/datatable/context/DataTableDataContext.ts
```diff
@@ -0,0 +1,5 @@
+import { createContext, useContext } from 'react';
+
+export const DataTableDataContext = createContext<any[] | undefined>(undefined);
+
+export const useDataTableDataContext = () => useContext(DataTableDataContext);
```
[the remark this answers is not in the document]
fzaninotto: done

docs/DataTable.md
fzaninotto: fixed

examples/demo/src/orders/OrderList.tsx
```diff
@@ -152,169 +154,72 @@ const TabbedDatagrid = () => {
...
                 </>
             )}
         </Fragment>
     );
 };
 
+const Column = DataTable.Col<Order>;
+const ColumnNumber = DataTable.NumberCol<Order>;
+
+const OrdersTable = React.memo(
```
djhi, line 181: Nice component but it does not showcase Fragments to group columns that you can still use as the `DataTable` children which was impossible with `Datagrid` if I'm not mistaken. Your call though

djhi (changes requested):
Can you:
- add a i18nProvider to all stories, they would be nicer
- add more top margin or a title in all stories so that the bulk actions toolbar is not cropped
- improve the expand stories (the expand content is ugly
- improve the fullapp story expand content (same)

Also:
In the AccessControl story, when `none` is selected in the storybook control, you still have the hand mouse pointer which is confusing as there is no action.

Finally, this is awesome!

PR #10599, opened 2025-03-19
Improve compatibility with some package managers

djhi:
## Problem

Some package managers are stricter than others regarding dependencies (PNPM for instance). They require you declare any package you use in your own dependencies even if it was a dependency of another of your dependencies.

Some react-admin packages don't follow this rule.

## Solution

Add undeclared dependencies.

## How To Test

```sh
make build install
cd ..
./react-admin/node_modules/.bin/create-react-admin myadmin --install pnpm
cd myadmin
pnpm run dev
```

## Additional Checks

### DIFF
--- a/docs/CreateReactAdmin.md
+++ b/docs/CreateReactAdmin.md
@@ -97,6 +97,7 @@ npx create-react-admin@latest your-admin-name --resource posts --resource commen
 * `npm` (default)
 * `yarn`
 * `bun`
+* `pnpm`
 * `none` (if you want to install dependencies yourself)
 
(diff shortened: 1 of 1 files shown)


erwanMarmelab (approved):

djhi:
> * This fixes PNPM install alright, but how does this fix YARN + PnP?
It could fix it if we could detect whether PnP mode is active in the user environment (global config). For now, let's keep how we handle yarn as it is.

> * don't you need to update the Create React Admin doc?
Update incoming

PR #10605, opened 2025-03-20
Add `<Translate>` component

erwanMarmelab:
## BC

The `Translate` type exported by `ra-core` was renamed `TranslateFunction`

## Problem

To output text, developers must use the `useTranslate` hook. It can be cumbersome.

## To Be Defined

A. args:

1.  `<Translate i18nKey="my-key" smart_count="18" />`
2.  `<Translate i18nKey="my-key" args={{smart_count:"18"}} />`

B. New name, we already have a type named `Translate`

## Solution

A new `ra-core` component to translate text.
```
<Translate i18nKey="ra.action.delete">Delete</Translate>
```
The default value is the child. The component should also accept translation options.
Test with interpolation, with `i18next` and `polyglot`.

## How To Test

https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-core-i18n-translate
https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-i18n-polyglot--translate-component
https://react-admin-storybook-a85l2rp3y-marmelab.vercel.app/?path=/story/ra-i18n-i18next--translate-component

## Additional Checks


djhi (approved):

packages/ra-core/src/i18n/TestTranslationProvider.tsx
```diff
@@ -32,7 +32,7 @@ export const testI18nProvider = ({
                       ? typeof message === 'function'
                           ? message(options)
                           : message
-                      : options?._ || key;
+                      : options?._;
```
fzaninotto, line 35: Why this change?

packages/ra-core/src/i18n/Translate.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+describe('<Translate />', () => {
+    it('should render the translation', () => {
+        const { container } = render(<Basic />);
+        expect(container.innerHTML).toBe('My Translated Key');
+    });
+
+    it('should render the translation event if children is set', () => {
+        const { container } = render(
+            <TestTranslationProvider
```
fzaninotto, line 21: Instead of copying the code from the stories, you'd better render the stories themselves.

packages/ra-core/src/i18n/Translate.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+);
+
+export const NoTranslation = () => (
+    <TestTranslationProvider messages={{}}>
+        <Translate i18nKey="custom.myKey" />
+    </TestTranslationProvider>
+);
+
+export const NoTranslationWithChildren = ({
+    isChildrenANode = false,
```
fzaninotto, line 28: I'd make 2 stories instead of one - I think the argument is a bit artificial here.

packages/ra-core/src/i18n/Translate.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+NoTranslationWithChildren.args = { isChildrenANode: false };
+NoTranslationWithChildren.argTypes = {
+    isChildrenANode: { control: 'boolean' },
+};
+
+export const Args = () => (
+    <TestTranslationProvider
+        messages={{
+            custom: {
+                myKey: ({ price }) => `It cost ${price}.00 $`,
```
fzaninotto, line 54: It's weird to use the function syntax - we never use it elsewhere. Why don't you use the interpolation syntax instead?

fzaninotto: My bad, I saw in the previous PR that it's not possible.

docs/Translate.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Props
+
+| Prop       | Required | Type        | Default | Description                                                     |
+| ---------- | -------- | ----------- | ------- | --------------------------------------------------------------- |
+| `args`     | Optional | `Object`    | -       | The arguments used for pluralization and interpolation.         |
+| `children` | Optional | `ReactNode` | -       | The default content to display if the translation is not found. |
+| `i18nKey`  | Required | `string`    | -       | The translation key.                                            |
+
+## `args`: Pluralization and Interpolation
```
fzaninotto, line 28: This one is called `options` in `useTranslate`. I think we should use the same name.

packages/ra-core/src/i18n/TestTranslationProvider.tsx
erwanMarmelab: this implementation was wrong. If the translation function doesn't find a translation for the key, it should return the key. It should return anything. Like polyglot or i18next

erwanMarmelab:
Comments applied in #10615

PR #10607, opened 2025-03-20
Add generic type in datagrid

### LINKED ISSUE #10595: Making Datagrid Generic
**What you were expecting:**
I'm getting this error

> Conversion of type 'RecordType' to type 'MyGenericType' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

And I believe it is bad practice to continue with the unkown for this case

**What happened instead:**

I'm not able to pass a generic to the datagrid so that it changes the record in rowClick and other functions.

Image

1. If I try to pass the generic it says it doesn't accept it
2. If I don't pass the generic the error in the row function still remains and I can't overwrite it without passing unknow

**Steps to reproduce:**
1. Create any list screen
2. Use RowClick and assign a record with "as"

**Related code:**
Datagrid.tsx / DatagridProps

```
export interface DatagridProps<RecordType extends RaRecord = any>
    extends Omit<TableProps, 'size' | 'classes' | 'onSelect'> {
    /**
     * The component used to render the body of the table. Defaults to <DatagridBody>.
     *
     * @see https://marmelab.com/react-admin/Datagrid.html#body
     */
    body?: ReactElement | ComponentType;

    /**
     * A class name to apply to the root table element
     */
    className?: string;
```

**Other information:**

**Environment**

* React-admin version: 5.6.3
* Last version that did not exhibit the issue (if applicable): 5.3.4
* React version: 18.3.1
* Browser: Chrome

ogustavo-pereira:
## Problem
Closes #10595

## Solution
Since the datagrid is a React.forwardRef type component, I can't create a generic one easily, so I need help to solve and bring the best solution to this challenge.

I used the article below as a basis to solve the problem
https://www.totaltypescript.com/forwardref-with-generic-components

Variable names are for illustration purposes only, they can be better planned if the solution is to your liking.
_Describe the solution this PR implements_

## How To Test

In the datagrid component itself, you can create a test component passing a generic and seeing if this generic is going to the rowClick

At first I focused only on rowClick, which is my main problem today, if you need other places we can add it

image
```tsx
const Test = () => {
    return (
        <Datagrid<{ id: string; name: string }>
            rowClick={(id, resource, record) =>
                console.log({ id, resource, record })
            }   
        >
            <div></div>
        </Datagrid>
    );
};
```

## Additional Checks

- [ ] The PR targets `master` for a bugfix or a documentation fix, or `next` for a feature

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
@@ -117,9 +117,16 @@ const defaultBulkActionButtons = <BulkDeleteButton />;
  *     );
  * }
  */
-export const Datagrid: React.ForwardRefExoticComponent<
-    Omit<DatagridProps, 'ref'> & React.RefAttributes<HTMLTableElement>
-> = React.forwardRef<HTMLTableElement, DatagridProps>((props, ref) => {
+
+const fixedForwardRef = <T, P = {}>(
+    render: (props: P, ref: React.Ref<T>) => React.ReactNode
+): ((props: P & React.RefAttributes<T>) => React.ReactNode) =>
+    React.forwardRef(render) as any;
+
+const WDatagrid = <RecordType extends RaRecord = any>(
+    props: DatagridProps<RecordType>,
+    ref
+) => {
     const resourceFromContext = useResourceContext(props);
     const { canAccess: canDelete } = useCanAccess({
         resource: resourceFromContext,
@@ -299,7 +306,9 @@ export const Datagrid: React.ForwardRefExoticComponent<
             </OptionalResourceContextProvider>
         </DatagridContextProvider>
     );
-});
+};
+
+export const Datagrid = fixedForwardRef(WDatagrid);
 
 const createOrCloneElement = (element, props, children) =>
     isValidElement(element)
@@ -524,7 +533,7 @@ export interface DatagridProps<RecordType extends RaRecord = any>
      *     </List>
      * );
      */
-    rowClick?: string | RowClickFunction | false;
+    rowClick?: string | RowClickFunction<RecordType> | false;
 
     /**
      * A function that returns the sx prop to apply to a row.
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -117,9 +117,16 @@ const defaultBulkActionButtons = <BulkDeleteButton />;
...
  * }
  */
-export const Datagrid: React.ForwardRefExoticComponent<
-    Omit<DatagridProps, 'ref'> & React.RefAttributes<HTMLTableElement>
-> = React.forwardRef<HTMLTableElement, DatagridProps>((props, ref) => {
+
+const fixedForwardRef = <T, P = {}>(
+    render: (props: P, ref: React.Ref<T>) => React.ReactNode
+): ((props: P & React.RefAttributes<T>) => React.ReactNode) =>
+    React.forwardRef(render) as any;
```
djhi, line 124: Can you rename this to `genericForwardRef` (just like our `genericMemo`) and extract it in its own file directly under `src`? I have a feeling we're going to use it elsewhere

packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -117,9 +117,16 @@ const defaultBulkActionButtons = <BulkDeleteButton />;
...
+
+const WDatagrid = <RecordType extends RaRecord = any>(
```
djhi, line 126: `WDatagrid`? I suggest `DatagridComponent`

packages/ra-ui-materialui/src/list/types.ts
```diff
@@ -1,6 +1,6 @@
 import { Identifier, RaRecord } from 'ra-core';
 
-export type RowClickFunction = <RecordType extends RaRecord = RaRecord>(
+export type RowClickFunction<RecordType extends RaRecord = RaRecord> = (
```
djhi, line 3: We use this in `SimpleListItem` from `SimpleList` too. Can you make it accept a `RecordType` as well?

djhi (changes requested):
Thanks!

fzaninotto:
We're introducing an alternative Datagrid component called [DataTable](https://github.com/marmelab/react-admin/pull/10597) that is, by default, generic. It will be released in react-admin 5.8.

We won't be adding new updates to Datagrid, as we want to encourage the new component. So I'll close your PR and the related issue. 

Thanks anyway for your contribution.

PR #10608, opened 2025-03-21
Make `<RichTextInput>` fullWidth by default

erwanMarmelab:
## Problem

This is probably the only input that isn’t full width by default. As we can’t override its props with the theme, it’s cumbersome.

## Solution

- Make it full width by default (in a minor)
- Fix the stories in dark mode

## How To Test

## Additional Checks

### DIFF
--- a/packages/ra-input-rich-text/src/RichTextInput.stories.tsx
+++ b/packages/ra-input-rich-text/src/RichTextInput.stories.tsx
@@ -30,7 +30,9 @@ import {
 } from './RichTextInput';
 import { RichTextInputToolbar } from './RichTextInputToolbar';
 import {
+    Box,
     Button,
+    Card,
     List,
     ListItem,
     ListItemButton,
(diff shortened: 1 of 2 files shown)


packages/ra-input-rich-text/src/RichTextInput.stories.tsx
```diff
@@ -62,155 +64,173 @@ const i18nProvider: I18nProvider = {
...
+                        label="Body"
+                        source="body"
+                        fullWidth
+                    />
+                    <FormInspector />
+                </SimpleForm>
+            </Card>
         </ResourceContextProvider>
     </AdminContext>
 );
```
djhi, line 198: As we now enable fullWidth by default, can you change this story to show when it's **not** enabled?

djhi (changes requested):

packages/ra-input-rich-text/src/RichTextInput.stories.tsx
```diff
@@ -44,12 +46,12 @@ export default { title: 'ra-input-rich-text/RichTextInput' };
 const FormInspector = ({ name = 'body' }) => {
     const value = useWatch({ name });
     return (
-        <div style={{ backgroundColor: 'lightgrey' }}>
+        <Box sx={theme => ({ backgroundColor: theme.palette.divider })}>
```
djhi, line 49: Why not simply use a `Paper` like in other stories?

djhi: Nevermind, I misread this component as a wrapper for the stories

djhi (approved):

PR #10614, opened 2025-03-24
Introduce `useCloseNotification` hook

djhi:
## Problem

When passing a ReactNode with custom actions to `notify`, there's no way to close the notification programmatically (on the action click).

## Solution

Introduce a new context that returns a function to close the notification.

## How To Test

- Open the story: https://react-admin-storybook-2npud2c0z-marmelab.vercel.app/?path=/story/ra-ui-materialui-layout-notification--custom-notification-with-action
- Wait for the notification to disappear. There should be no console.log in the devtool
- Refresh the page
- Click the **Cancel** button. The notification should close and a log should be printed in the devtool.

## Additional Checks


docs/useNotify.md
```diff
@@ -216,4 +221,47 @@ export const ConnectionWatcher = () => {
 };
 ```
 
-Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
\ No newline at end of file
+Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
+
```
fzaninotto, line 225:
```suggestion

## Closing The Notification

```

docs/useNotify.md
```diff
@@ -47,6 +37,21 @@ notify('item.created', { type: 'info', messageArgs: { resource: 'post' } });
 notify('Element updated', { type: 'info', undoable: true });
 ```
 
+## Parameters
+
+The hook takes no argument and returns a callback. The callback takes 2 arguments:
```
fzaninotto, line 42: It's weird to mention 2 arguments and to illustrate it with a table with more than 2 rows. I'd convert the description of the options to a list.

docs/useNotify.md
```diff
@@ -216,4 +221,47 @@ export const ConnectionWatcher = () => {
 };
 ```
 
-Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
\ No newline at end of file
+Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
+
+If you have custom actions in your notification element, you can leverage the `useCloseNotification` hook to close the notification:
```
fzaninotto, line 226:
```suggestion
If you have custom actions in your notification element, you can leverage the `useCloseNotification` hook to close the notification programmatically:
```

docs/useNotify.md
```diff
@@ -216,4 +221,47 @@ export const ConnectionWatcher = () => {
...
+                setValue('myfield', 'a value');
+                notify(<SetFormValueNotification />);
+            }}
+            label="Set myfield value"
+        />
+    );
+};
+
+const SetFormValueNotification = () => {
+    const { setValue } = useFormContext();
```
fzaninotto, line 249: I think the `setValue` callback should be injected by the button. Otherwise, how can we be sure the notification is rendered inside the form context?

docs/useNotify.md
```diff
@@ -216,4 +221,47 @@ export const ConnectionWatcher = () => {
 };
 ```
 
-Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
\ No newline at end of file
+Note that if you use this ability to pass a React node, the message will not be translated - you'll have to translate it yourself using [`useTranslate`](./useTranslate.md).
+
+If you have custom actions in your notification element, you can leverage the `useCloseNotification` hook to close the notification:
+
```
fzaninotto, line 227: I think this example is bringing confusion. Earlier in the doc, we mention 'undoable' notifications (without explaining how to handle the undo). Here, you introduce a "cancellable" notificaiton, which seems like it's the same thing, but it uses a different syntax...

So either you should find another example for this feature, or we should document how to register an undoable action.

packages/ra-ui-materialui/src/layout/Notification.tsx
```diff
@@ -78,6 +79,11 @@ export const Notification = (props: NotificationProps) => {
         setOpen(false);
     }, [setOpen]);
 
+    const closeNotificationContext = React.useMemo(
```
fzaninotto, line 82: Why not use `useCallback` here?

packages/ra-ui-materialui/src/layout/useCloseNotification.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { CloseNotificationContext } from './CloseNotificationContext';
+
+export const useCloseNotification = () => {
```
fzaninotto, line 4: not reexported

packages/ra-ui-materialui/src/layout/CloseNotificationContext.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from 'react';
```
fzaninotto, line 1: I think this context (and the corresponding hook) should be in `ra-core`, just like `useNotify`.

packages/ra-ui-materialui/src/layout/Notification.tsx
djhi: I don't even need this line actually, `handleRequestClose` is already wrapped in a `useCallback`

docs/useNotify.md
djhi: Any suggestion for another example?

docs/useNotify.md
```diff
@@ -47,6 +37,26 @@ notify('item.created', { type: 'info', messageArgs: { resource: 'post' } });
...
+| --- | --- | --- | --- | --- |
+| `message` | Required | `string` | - | The message to display (a string, or a React node) |
+| `options` |  | `object` | - | The options |
+
+The `options` is an object that can have the following properties:
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `anchorOrigin` | `object` | - | The position of the notification. The default is `{ vertical: 'top', horizontal: 'right' }`. See [the Material UI documentation](https://mui.com/material-ui/react-snackbar/) for more details. |
+| `autoHideDuration` | `number | null` | - | Duration (in milliseconds) after which the notification hides. Set it to `null` if the notification should not be dismissible. |
```
slax57, line 54: I believe the default is 4000

docs/useNotify.md
```diff
@@ -47,6 +37,26 @@ notify('item.created', { type: 'info', messageArgs: { resource: 'post' } });
...
+| `messageArgs` | `object` | - | options to pass to the `translate` function (because notification messages are translated if your admin has an `i18nProvider`). It is useful for inserting variables into the translation. |
+| `multiLine` | `boolean` | - | Set it to `true` if the notification message should be shown in more than one line. |
+| `undoable` | `boolean` | - | Set it to `true` if the notification should contain an "undo" button |
+| `type` | `string` | - | The notification type (`info`, `success`, `error` or `warning` - the default is `info`) |
```
slax57, line 58: you could put `info` in the column for `Default`

slax57 (commented):
Other than that LGTM :white_check_mark:

slax57 (approved):
Good for me! I leave the final review to @fzaninotto

erwanMarmelab (approved):

PR #10617, opened 2025-03-25
[Doc] Add a FAQ entry for duplicate react-hook-form versions

djhi:
## Problem

Users are still regularily getting errors because of duplicate version of react-hook-form or react-router.

Example: https://stackoverflow.com/q/79464423/3975522

## Solution

Create an entry in the [FAQ](https://marmelab.com/react-admin/FAQ.html) for these issues.

## How To Test

- `make doc`
- http://0.0.0.0:4000/FAQ.html#i-get-an-error-about-control-being-null

## Additional Checks

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -6,10 +6,11 @@ title: "FAQ"
 # FAQ
 
 - [Can I have custom identifiers/primary keys for my resources?](#can-i-have-custom-identifiersprimary-keys-for-my-resources)
-- [I get warning about unique key for child in array](#i-get-warning-about-unique-key-for-child-in-array)
+- [I get warning about unique key for child in array](#i-get-a-warning-about-unique-key-for-child-in-array)
 - [How can I customize forms depending on its inputs values?](#how-can-i-customize-forms-depending-on-its-inputs-values)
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -6,10 +6,11 @@ title: "FAQ"
 # FAQ
 
 - [Can I have custom identifiers/primary keys for my resources?](#can-i-have-custom-identifiersprimary-keys-for-my-resources)
-- [I get warning about unique key for child in array](#i-get-warning-about-unique-key-for-child-in-array)
+- [I get warning about unique key for child in array](#i-get-a-warning-about-unique-key-for-child-in-array)
```
slax57, line 9:
```suggestion
- [I get a warning about unique key for child in array](#i-get-a-warning-about-unique-key-for-child-in-array)
```

slax57 (commented):
The doc is good :+1: 

But I think we should also add a FAQ entry for react-router, which often causes similar issues. 
It's already documented [here](https://marmelab.com/react-admin/Routing.html#troubleshooting), but IMO it would fit best in this FAQ section. So we should either duplicate the doc here or at least add a link to it.

PR #10622, opened 2025-03-26
Fix npm warnings when creating new app with create-react-admin

slax57:
## Problem

When installing dependencies for a new app created with `create-react-admin`, the following warnings appear in the console.

```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated @humanwhocodes/config-array@0.13.0: Use @eslint/config-array instead
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated @humanwhocodes/object-schema@2.0.3: Use @eslint/object-schema instead
npm warn deprecated eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options.
```

They are all due to the use of `eslint` v8, which is no longer maintained.

## Solution

Upgrade `create-react-admin`, and the whole monorepo, to `eslint` v9.

## How To Test

The CI checks should be enough to validate the monorepo config.

To test the `create-react-admin` fix:

1. `make build-create-react-admin`
2. In a different directory, run `./react-admin/packages/create-react-admin/node_modules/.bin/create-react-admin my-app -i`
3. Install the newly created `my-app` with NPM
4. Make sure there are no warnings

## Additional Checks


examples/simple/src/dataProvider.tsx
```diff
@@ -57,7 +57,7 @@ const addTagsSearchSupport = (dataProvider: DataProvider) => ({
                     ...params,
                     filter: item => {
                         const matchPublished =
-                            item.published == params.filter.published; // eslint-disable-line eqeqeq
```
djhi, line 60: I'm not sure about this: we got issues because of using `==` instead of `===`. Shouldn't we add this rule in the config?

djhi (changes requested):
Awesome! I can confirm I don't have any warnings when creating a new app

fzaninotto:
Please `yarn dedupe`

slax57:
Turns out it's not really linked to this PR, so back to RFR.

Just to be on the safe side, please confirm `make storybook` works on your machine before merging this PR. Thanks.

yarn.lock
```diff
@@ -1922,9 +1728,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@esbuild/aix-ppc64@npm:0.24.0":
-  version: 0.24.0
-  resolution: "@esbuild/aix-ppc64@npm:0.24.0"
+"@esbuild/aix-ppc64@npm:0.25.2":
```
djhi, line 1731: All esbuild deps have duplicates

djhi (changes requested):

yarn.lock
djhi: That's probably because we still use vite 5. Upgrading `vite` and `tsx` should target the same `esbuild` as storybook:
- https://github.com/vitejs/vite/blob/v6.2.4/packages/vite/package.json
- https://github.com/privatenumber/tsx/releases/tag/v4.19.3

image

djhi: We decided to tackle the other dependencies in another PR

djhi (approved):

PR #10636, opened 2025-04-02
Fix: Apply a default `defalutValue` on `ArrayInput`

erwanMarmelab:
## Problem

Instead of all other inputs, `ArrayInput` doesn't provide a default `defalutValue`.
This is an issue when calling the react-hook-form `reset` method as react-hook-form will not reset fields that don't have default values.

## Solution

Apply a default `defalutValue` on `ArrayInput`

## Additional Checks


djhi (changes requested):
A test that call reset on an `ArrayField` inside a `Create` would be great

djhi (approved):

PR #10643, opened 2025-04-04
Improve release script old version doc

djhi:
## Problem

This is an unnecessary manual step.

## Solution

Automate it

## How To Test

These scripts support a `RELEASE_DRY_RUN` env variable allowing to skip all mutations to git, GitHub or npm, except 1 git commit (but not pushed!) from lerna.

First, create a `.env` file from the `.env.template` file and fill it in with your Github token.

Then run:

```
RELEASE_DRY_RUN=true make release
```

Select the new minor version when Lerna asks for it.

When this is done, don't forget to revert the commit created by lerna to update the versions:

```
git reset --hard HEAD~1
```

## Additional Checks

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -48,11 +48,6 @@ echo "Test the 3 demos (simple, e-commerce, crm): check console & UI"
 echo "Press Enter when this is done"
 read
 
-step "manual task: Update the OldVersion.md file"
-echo "[Minor version only] Update the ./docs/OldVersions.md file to add the new minor version and update the previous one && commit"
-echo "Press Enter when this is done"
-read
-
 # Get the current version from package.json
 npm_previous_package_version=$(jq -r '.version' ./packages/react-admin/package.json)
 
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -74,6 +69,13 @@ if [ ! -z "$RELEASE_DRY_RUN" ]; then
 fi
 
 if [ "${npm_previous_package_version%.*}" != "${npm_current_package_version%.*}" ]; then
+    echo "New minor version - Updating the OldVersion.md file"
+    npm_previous_package_minor_version=${npm_previous_package_version%.*}
+    npm_current_package_minor_version=${npm_current_package_version%.*}
+    echo $npm_previous_package_minor_version
+    echo $npm_current_package_minor_version
+    sed -i "s/^- \[v$npm_previous_package_minor_version\].*/- [v$npm_current_package_minor_version](https:\/\/github.com\/marmelab\/react-admin\/blob\/master\/docs\/Admin.md)\n- [v$npm_previous_package_minor_version](https:\/\/github\.com\/marmelab\/react\-admin\/blob\/$npm_previous_package_version\/docs\/Admin.md\)/" docs/OldVersions.md
```
slax57, line 77:
```suggestion
    sed -i "s/^- \[v$npm_previous_package_minor_version\].*/- [v$npm_current_package_minor_version](https:\/\/github.com\/marmelab\/react-admin\/blob\/master\/docs\/Admin.md)\n- [v$npm_previous_package_minor_version](https:\/\/github\.com\/marmelab\/react\-admin\/blob\/v$npm_previous_package_version\/docs\/Admin.md\)/" docs/OldVersions.md
```

slax57, line 77: Also a nitpick: this will be committed with the next `git commit` instruction under the message `"Update create-react-admin templates dependencies for version..."`.

IMO we should either
- Create a separate commit for updating the `OldVersions.md` file
- Update the commit message to be more general, and make it clearer (in the code indentation) this commit includes both actions

slax57 (changes requested):

slax57 (approved):

PR #10657, opened 2025-04-11
Fix `<FilterLiveForm>` compatibility with react-hook-form 7.55.0

slax57:
## Problem

New version react-hook-form 7.55.0 causes bugs in filter forms using `<FilterLiveForm>`, where a previous value could be reapplied after a change.

### Reproduction 1

Sandbox: https://stackblitz.com/edit/github-ruowu9px?file=src%2Fposts%2FPostList.tsx
In the Post List (desktop mode), try to apply the newsletter filter more than three times.
You'll see that sometimes the previous value reapplies shortly after it was changed.

### Reproduction 2

In a local clone of the repository:
1. Update RHF to its latest version: `yarn up react-hook-form` (should be 7.55.0)
2. Run the Storybook: `make storybook`
3. Go to story http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterliveform--with-filter-list-section
4. Try to apply the newsletter filter more than three times. You'll see the same thing as Repro 1.

## Solution

`<FilterLiveForm>` has some custom code to reset the form whenever the filter values change externally. This is required to keep the form values in sync.
However this means the form is reset with values that do not necessarily match the fields registered into it.
Apparently with RHF 7.55.0 this causes the `watch` callback to be called more often than necessary, even on fields registered in the form but which have not changed. This, in turns, calls the `setFilters` callback too often.

The solution consists in only resetting fields which are actually registered in the form. Other filter values do not need to trigger a reset of the form. Hence we avoid calling `setFilters` when the form values did not change, avoiding the race condition with external changes.

## How To Test

To test the fix you need to use RHF v7.55.0: `yarn up react-hook-form`
Then, you can test:
- The unit test: `FilterLiveForm.spec.tsx`
- The story: http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterliveform--with-filter-list-section

## Additional Checks

### DIFF
--- a/packages/ra-core/src/form/FilterLiveForm.tsx
+++ b/packages/ra-core/src/form/FilterLiveForm.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import isEqual from 'lodash/isEqual';
 import cloneDeep from 'lodash/cloneDeep';
 import get from 'lodash/get';
+import merge from 'lodash/merge';
 import set from 'lodash/set';
 import { ReactNode, useEffect } from 'react';
 import { FormProvider, useForm, UseFormProps } from 'react-hook-form';
@@ -110,10 +111,7 @@ export const FilterLiveForm = (props: FilterLiveFormProps) => {
             return;
         }
         formChangesPending.current = true;
-        setFilters({
-            ...filterValues,
-            ...values,
-        });
+        setFilters(merge({}, filterValues, values));
     };
     const debouncedOnSubmit = useDebouncedEvent(onSubmit, debounce || 0);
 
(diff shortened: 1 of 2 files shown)


djhi (approved):

packages/ra-core/src/form/FilterLiveForm.tsx
```diff
@@ -111,7 +112,7 @@ export const FilterLiveForm = (props: FilterLiveFormProps) => {
             return;
         }
         formChangesPending.current = true;
-        setFilters(merge({}, filterValues, values));
+        setFilters(mergeObjNotArray(filterValues, values));
```
djhi, line 115: Can't you just use [`Array.isArray`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)?

packages/ra-core/src/form/FilterLiveForm.tsx
```diff
@@ -170,6 +171,16 @@ export interface FilterLiveFormProps
     >;
 }
 
+// Lodash merge customizer to merge objects but not arrays
+const mergeCustomizer = (objValue: any, srcValue: any) => {
+    if (isArray(srcValue)) {
```
djhi, line 176: Same

djhi (changes requested):

djhi (approved):

PR #10670, opened 2025-04-16
Fix `<RichTextField>` should render `emptyText` when value is an empty string

slax57:
## Problem

`<RichTextField>` does not render `emptyText` when the value is equal to `''`.

## Solution

Fix `<RichTextField>` to render `emptyText` when the value is an empty string.

## How To Test

Unit test + new Story

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/field/RichTextField.stories.tsx
+++ b/packages/ra-ui-materialui/src/field/RichTextField.stories.tsx
@@ -133,3 +133,23 @@ It is regarded as one of Tolstoy's finest literary achievements and remains a cl
         <RichTextField source="body" purifyOptions={{ ADD_ATTR: ['target'] }} />
     </RecordContextProvider>
 );
+
+export const Empty = ({ emptyText, body }) => (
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/field/RichTextField.stories.tsx
```diff
@@ -133,3 +133,23 @@ It is regarded as one of Tolstoy's finest literary achievements and remains a cl
...
+    body: '',
+};
+Empty.argTypes = {
+    emptyText: {
+        options: [undefined, 'empty'],
+        control: { type: 'inline-radio' },
+    },
+    body: {
+        options: [undefined, null, '', 'foo'],
+        control: { type: 'inline-radio' },
```
djhi, line 153:
```suggestion
        options: [undefined, null, 'empty string', 'foo'],
        mapping: {
            undefined: undefined,
            null: null,
            'empty string': '',
            'foo': 'foo',
        },
        control: { type: 'inline-radio' },
```

djhi (changes requested):
:+1:

PR #10680, opened 2025-04-17
Fix MUI7 compatibility by not using deprecated system props

smeng9:
## Problem

The system props is deprecated in v6 https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#system-props and v7 https://mui.com/material-ui/migration/upgrade-to-v7/#quality-of-life-improvements

Follow up of https://github.com/marmelab/react-admin/pull/10463

### DIFF
--- a/packages/ra-ui-materialui/src/form/FormTab.tsx
+++ b/packages/ra-ui-materialui/src/form/FormTab.tsx
@@ -44,7 +44,6 @@ export const FormTab = (props: FormTabProps) => {
     const renderContent = () => (
         <FormGroupContextProvider name={value.toString()}>
             <Stack
-                alignItems="flex-start"
                 style={hidden ? hiddenStyle : undefined}
                 className={contentClassName}
                 id={`tabpanel-${value}`}
@@ -53,6 +52,12 @@ export const FormTab = (props: FormTabProps) => {
                 // notes that aria-hidden="false" currently behaves inconsistently across browsers.
                 aria-hidden={hidden || undefined}
                 {...rest}
+                sx={[
+                    {
+                        alignItems: 'flex-start',
+                    },
+                    ...(Array.isArray(rest.sx) ? rest.sx : [rest.sx]),
+                ]}
             >
                 {children}
             </Stack>
--- a/packages/ra-ui-materialui/src/form/SimpleForm.tsx
+++ b/packages/ra-ui-materialui/src/form/SimpleForm.tsx
@@ -62,7 +62,12 @@ export const SimpleForm = (inProps: SimpleFormProps) => {
     return (
         <Form {...rest}>
             <Component className={className} sx={sx}>
-                <Stack alignItems="flex-start" {...sanitizeRestProps(props)}>
+                <Stack
+                    {...sanitizeRestProps(props)}
+                    sx={{
+                        alignItems: 'flex-start',
+                    }}
+                >
                     {children}
                 </Stack>
             </Component>
--- a/packages/ra-ui-materialui/src/layout/TitlePortal.tsx
+++ b/packages/ra-ui-materialui/src/layout/TitlePortal.tsx
@@ -3,13 +3,18 @@ import { Typography, TypographyProps } from '@mui/material';
 
 export const TitlePortal = (props: TypographyProps) => (
     <Typography
-        flex="1"
-        textOverflow="ellipsis"
-        whiteSpace="nowrap"
-        overflow="hidden"
         variant="h6"
         color="inherit"
         id="react-admin-title"
         {...props}
+        sx={[
+            {
+                flex: '1',
+                textOverflow: 'ellipsis',
+                whiteSpace: 'nowrap',
+                overflow: 'hidden',
+            },
+            ...(Array.isArray(props.sx) ? props.sx : [props.sx]),
+        ]}
     />
 );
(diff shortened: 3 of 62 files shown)


djhi (changes requested):
Thanks for the PR! However:
- The CRM and the simple examples still use MUI v5 and shouldn't be modified
- Our packages uses MUIv5 too in dev to ensure we don't use MUIv6 or v7 only features. That means our stories shouldn't be modified either

Please revert those changes

smeng9:
Hi @djhi,

Suggestions applied

packages/ra-ui-materialui/src/form/FormTab.tsx
```diff
@@ -53,6 +52,12 @@ export const FormTab = (props: FormTabProps) => {
                 // notes that aria-hidden="false" currently behaves inconsistently across browsers.
                 aria-hidden={hidden || undefined}
                 {...rest}
+                sx={[
+                    {
+                        alignItems: 'flex-start',
+                    },
+                    ...(Array.isArray(rest.sx) ? rest.sx : [rest.sx]),
+                ]}
```
djhi, line 60: I don't get this: why not just:

```suggestion
                sx={{
                        alignItems: 'flex-start',
                        ...rest.sx,
                }}
```

packages/ra-ui-materialui/src/form/SimpleForm.tsx
```diff
@@ -62,7 +62,12 @@ export const SimpleForm = (inProps: SimpleFormProps) => {
     return (
         <Form {...rest}>
             <Component className={className} sx={sx}>
-                <Stack alignItems="flex-start" {...sanitizeRestProps(props)}>
+                <Stack
+                    {...sanitizeRestProps(props)}
+                    sx={{
+                        alignItems: 'flex-start',
+                    }}
+                >
```
djhi, line 70: I'm not sure about this. How can I override the `sx` now?

packages/ra-ui-materialui/src/layout/TitlePortal.tsx
```diff
@@ -3,13 +3,18 @@ import { Typography, TypographyProps } from '@mui/material';
...
         {...props}
+        sx={[
+            {
+                flex: '1',
+                textOverflow: 'ellipsis',
+                whiteSpace: 'nowrap',
+                overflow: 'hidden',
+            },
+            ...(Array.isArray(props.sx) ? props.sx : [props.sx]),
+        ]}
```
djhi, line 18: Same here, do we really need the Array check?

packages/ra-ui-materialui/src/list/pagination/InfinitePagination.tsx
```diff
@@ -67,7 +67,16 @@ export const InfinitePagination = ({
...
-        <Box ref={observerElem} py={2} textAlign="center" sx={sx}>
+        <Box
+            ref={observerElem}
+            sx={[
+                {
+                    py: 2,
+                    textAlign: 'center',
+                },
+                ...(Array.isArray(sx) ? sx : [sx]),
+            ]}
```
djhi, line 78: here too

djhi (changes requested):

packages/ra-ui-materialui/src/form/FormTab.tsx
smeng9: Hi @djhi,

These are auto applied by the codemod tool in https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#system-props.

Main reason we did not directly unpack ...rest.sx, but we converted to an array is because the typing of SxProp can also be functions. MUI suggests using array values. https://mui.com/system/getting-started/the-sx-prop/#array-values

packages/ra-ui-materialui/src/form/SimpleForm.tsx
smeng9: Hi @djhi,

In the old sanitizeRestProps function, the sx property is already sanitized out https://github.com/marmelab/react-admin/blob/next/packages/ra-ui-materialui/src/form/SimpleForm.tsx#L99 

That means, in the original code it does not support overriding sx.

packages/ra-ui-materialui/src/layout/TitlePortal.tsx
smeng9: Yes, SxProps can be functions, which we cannot directly unpack

packages/ra-ui-materialui/src/list/pagination/InfinitePagination.tsx
smeng9: See previous comment

packages/ra-ui-materialui/src/form/SimpleForm.tsx
smeng9: We did not introduce or remove any new feature

djhi: My mistake. Thanks

smeng9:
Hi @djhi do you have additional questions regarding array check for sx prop?

The codemod for original mui upgrade is here https://github.com/mui/material-ui/pull/42282

djhi (approved):

djhi:
Thanks!

PR #10682, opened 2025-04-18
Fix `<ShowGuesser>` print incorrect code for reference arrays

djhi:
## Problem

The console.log for reference arrays is wrong:
```tsx
<ReferenceArrayField source="tags_ids" reference="tags"><TextField source="id" /></ReferenceArrayField>
```

It should be:
```tsx
<ReferenceArrayField source="tags_ids" reference="tags" />
```

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/detail/showFieldTypes.tsx
+++ b/packages/ra-ui-materialui/src/detail/showFieldTypes.tsx
@@ -14,9 +14,11 @@ import {
     RichTextField,
     TextField,
     UrlField,
+    ChipField,
 } from '../field';
 import { SimpleShowLayout, SimpleShowLayoutProps } from './SimpleShowLayout';
 import { InferredElement, InferredTypeMap, InputProps } from 'ra-core';
+import { SingleFieldList } from '../list';
 
 export const showFieldTypes: InferredTypeMap = {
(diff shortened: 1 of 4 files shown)


packages/ra-ui-materialui/src/detail/showFieldTypes.tsx
```diff
@@ -81,14 +83,16 @@ ${children.map(child => `            ${child.getRepresentation()}`).join('\n')}
...
-                Partial<Pick<InputProps, 'source'>>
-        ) => <TextField source="id" {...props} />,
-        representation: () => `<TextField source="id" />`,
+        component: () => (
+            <SingleFieldList>
+                <ChipField source="id" />
+            </SingleFieldList>
+        ),
+        representation: () =>
+            `<SingleFieldList><ChipField source="id" /></SingleFieldList>`,
```
slax57, line 95: Can't you remove the `referenceArrayChild` key altogether? I'm worried the rendered JSX does not match the JSX code printed in the console.

djhi: It might be used by guessers we don't maintain so I'd rather keep it

PR #10685, opened 2025-04-18
[chore] ESLint: (re)enable import rules

slax57:
## Problem

Rules brought by `eslint-plugin-import` were lost in the process of migrating to ESLint 9.x (#10622).

As a result, the `import/no-extraneous-dependencies` rule added in #10599 was lost too, hence we risk breaking compatibility with PNPM again.

## Solution

Reintroduce and properly configure `eslint-plugin-import`.

Doing so revealed new issues regarding dependencies, so this PR fixes them too.

## How To Test

`make lint` (but it's already done by the CI)

## Additional Checks


packages/ra-ui-materialui/package.json
```diff
@@ -37,6 +37,7 @@
         "@types/react-is": "^18.3.0",
         "@types/react-transition-group": "^4.4.10",
         "cross-env": "^5.2.0",
+        "csstype": "^3.1.3",
```
slax57, line 40: This dep is actually used in the code and was missing from the `package.json` file.

https://github.com/marmelab/react-admin/blob/a4703fd2e29c3ad849cde0d6d31bd5767027385f/packages/ra-ui-materialui/src/Labeled.tsx#L11

fzaninotto (approved):

PR #10686, opened 2025-04-21
Allow resource specific translations for pages and buttons

djhi:
## Problem

We have many translations that receive (or should receive) the current resource, for instance:

- `ra.page.create` (and other similar views)
- `ra.action.create` (and other similar actions)

However, in some languages such as French, some words are feminine and other masculine. This forces devs to have their own translations if they want:

- _New board_ : _Nouveau tableau_
- _New column_ : _Nouvelle colonne_

## Solution

Just like we did for notifications, accept resource specific translations in the form:

- `resources.posts.page.create`
- `resources.posts.action.create`

Those translations should fall back to the current translations.

I identified the following translations but there may be others:
- create default title
- edit default title
- show default title
- list and infinite list default title
- create button label

## How To Test

- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-core-controller-createbase--default-title
- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-core-controller-editbase--default-title
- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-core-controller-showbase--default-title
- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-core-controller-list-infinitelistbase--default-title
- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-core-controller-list-listbase--default-title
- https://react-admin-storybook-moe9v68sh-marmelab.vercel.app/?path=/story/ra-ui-materialui-button-createbutton--label

## Additional Checks


djhi:
> I'm a bit worried by the fact that the resource-specific translation keys for the CreateButton and the Create pages are the same. In my opinion, users should be able to have different translations for those (e.g. "Create" for the button and "New post" for the page).

They're not the same. One is `page.create`, the other `action.create`

erwanMarmelab (approved):

packages/ra-ui-materialui/src/button/CreateButton.tsx
```diff
@@ -63,6 +64,14 @@ const CreateButton = (inProps: CreateButtonProps) => {
     });
     const createPath = useCreatePath();
     const translate = useTranslate();
+    const getResourceLabel = useGetResourceLabel();
+    const label =
+        labelProp ??
```
fzaninotto, line 69: This is a regression: if a user provides a custom label, it is no longer translated. Same for the other buttons.

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -77,6 +84,15 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
     if (isValidElement(recordRepresentation)) {
         recordRepresentation = `#${record?.id}`;
     }
+    const label =
+        labelProp ??
+        translate(`resources.${resource}.action.delete`, {
+            recordRepresentation,
+            _: translate(`ra.action.delete`, {
+                name: getResourceLabel(resource, 1),
```
fzaninotto, line 92: we've computed a better `resourceName` just above, use it instead of this one.

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -77,6 +84,15 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
     if (isValidElement(recordRepresentation)) {
         recordRepresentation = `#${record?.id}`;
     }
+    const label =
```
fzaninotto, line 87: This is a regression: the label will be translated twice, because it's already translated by `Button`. Same for the other buttons.

fzaninotto (changes requested):

docs/Buttons.md
```diff
@@ -704,6 +704,38 @@ It also supports [all the other `<Button>` props](#button).
...
+};
+```
+
+You can also customize this label by specifying a custom `label` prop:
+
+```jsx
+export const PostCreateButton = () => (
+    <CreateButton label="New post" />
+);
+```
```
slax57, line 806: Should we also specify that this label accepts any custom translation key?

I see it's the case below for `DeleteButton` for example.

Same comment may apply to other components (e.g. I see it's the same for the `EditButton`)

docs/Buttons.md
```diff
@@ -780,7 +812,27 @@ You can also call it with a record and a resource:
...
+// in src/i18n/en.js
+import englishMessages from 'ra-language-english';
+
+export const en = {
+    ...englishMessages,
+    resources: {
+        posts: {
+            name: 'Post |||| Posts',
+            action: {
+                delete: 'Destroy %{recordRepresentation}'
```
slax57: Not very found of this example

```suggestion
                delete: 'Permanently remove %{recordRepresentation}'
```

docs/Buttons.md
```diff
@@ -1152,12 +1237,42 @@ export const PostShow = () => (
...
+        },
+    },
+    ...
+};
+```
+
+You can also customize this label by specifying a custom `label` prop:
+
+```jsx
+<ListButton label="Delete this comment" />
```
slax57:
```suggestion
<ListButton label="See all comments" />
```

packages/ra-core/src/i18n/useResourceTranslation.ts
```diff
@@ -0,0 +1,30 @@
+import type { ReactNode } from 'react';
+import { useTranslate } from './useTranslate';
+
+export const useResourceTranslation = (
+    params: UseResourceTranslationOptions
+) => {
+    const { resourceI18nKey, baseI18nKey, userText, options } = params;
+    const translate = useTranslate();
+
+    if (userText) {
```
slax57: IMO you should rather strictly compare to undefined, as currently it prevents the user from setting the label to anything falsy like `false` or an empty string.

```suggestion
    if (userText !== undefined) {
```

packages/ra-ui-materialui/src/button/Button.tsx
```diff
@@ -102,7 +117,7 @@ interface Props<RootComponent extends React.ElementType> {
     component?: RootComponent;
     to?: LocationDescriptor | To;
     disabled?: boolean;
-    label?: string;
+    label?: React.ReactNode;
```
slax57, line 120: You need to update `Buttons.md` to mention this new ability I believe.

Besides, it would be nice to also document the `label` prop on the component `<Button>` since it does a little more now.

slax57, line 120: Also, it's too bad we have neither stories nor tests to test this new feature :(

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -93,35 +131,9 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
             <Confirm
                 isOpen={open}
                 loading={isPending}
-                title={confirmTitleProp ?? confirmTitle}
-                content={confirmContentProp ?? confirmContent}
+                title={confirmTitle}
+                content={confirmContent}
```
slax57: Shouldn't you avoid double translation here too?

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -77,12 +81,46 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
...
+        resourceI18nKey: `resources.${resource}.message.delete_content`,
+        baseI18nKey: 'ra.message.delete_content',
+        options: {
+            recordRepresentation,
+            name: resourceName,
+            id: record?.id,
+            ...contentTranslateOptions,
+        },
+        userText: confirmContentProp,
+    });
```
slax57, line 124: Unless I'm mistaken this is neither covered by tests nor stories. It would be nice to have a way to actually test this.

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.tsx
```diff
@@ -39,12 +47,32 @@ export const DeleteWithUndoButton = <RecordType extends RaRecord = any>(
...
+            : recordRepresentationValue?.toString();
+    const label =
+        labelProp ??
+        translate(`resources.${resource}.action.delete`, {
+            recordRepresentation,
+            _: translate(`ra.action.delete`, {
+                name: getResourceLabel(resource, 1),
+                recordRepresentation,
+            }),
+        });
```
slax57: Why didn't you use `useResourceTranslation` here? :thinking:

packages/ra-ui-materialui/src/button/IconButtonWithTooltip.tsx
```diff
@@ -47,5 +54,5 @@ export const IconButtonWithTooltip = ({
 };
 
 export interface IconButtonWithTooltipProps extends IconButtonProps {
-    label: string;
+    label: React.ReactNode;
```
slax57, line 57: Does it really make sense to allow a ReactNode here? If I understand correctly, this label will only be used if it's a string anyways, right?

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
```diff
@@ -639,7 +639,7 @@ describe('<ReferenceField />', () => {
                     selector: 'a > span',
                 })
             );
-            await screen.findByText('ra.page.show');
+            await screen.findByText('resources.authors.page.show');
```
slax57, line 643: I realize this change may also affect EE tests. We need to be careful, and may have to adapt the EE tests to support both RA < 5.9 and RA >= 5.9.

packages/ra-ui-materialui/src/button/IconButtonWithTooltip.tsx
djhi: No if it's not a string it won't be passed to `aria-label` but will still be used in the `Tooltip` at L38

slax57: Right!

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
djhi: You're mistaken, there's a `Label` story in the `DeleteButton.stories.tsx` file that allows you to test this for both `DeleteWithConfirmButton` and `DeleteWithUndoButton`

slax57: for the `label` sure, but what about the `confirmTitle` and `confirmContent`?

djhi: Indeed, incoming

packages/ra-ui-materialui/src/button/Button.tsx
djhi: Added

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
djhi: Added

packages/ra-ui-materialui/src/field/ReferenceField.spec.tsx
djhi: yes

packages/ra-ui-materialui/src/button/Button.tsx
djhi: Added too

packages/ra-ui-materialui/src/button/UpdateWithConfirmButton.tsx
```diff
@@ -136,45 +144,56 @@ export const UpdateWithConfirmButton = (
...
+            ...contentTranslateOptions,
+        },
+        userText: confirmContentProp,
+    });
 
     return (
         <Fragment>
             <StyledButton
                 onClick={handleClick}
                 label={label}
```
slax57: you forgot to avoid double translation here I believe

slax57 (changes requested):
Other than that LGTM

slax57 (approved):

PR #10687, opened 2025-04-21
Fix Loading component timeout

WiXSL:
## Problem

`Loading` component is difficult to re-use because it has a hard-coded one second timeout inside of it.

## Solution

Add a `timeout` prop default to one second to make it more flexible

## Additional Checks


fzaninotto:
Thanks!

This is technically an evolution, though. Would you mind rebasing your PR on `next`?

PR #10688, opened 2025-04-22
create-react-admin: Upgrade to MUI v7

slax57:
## Problem

#10639 brought support for MUI v7.
To help minimizing technical debt, all new react-admin apps should start using that version right away.
But create-react-admin still uses MUI v6 by default.

## Solution

Upgrade create-react-admin to generate projects that use MUI v7 by default.

## How To Test

**Testing with yarn**

It's the easiest way to test, since we force the version of MUI packages using yarn resolutions.

1. `make build-create-react-admin`
2. `./packages/create-react-admin/node_modules/.bin/create-react-admin my-admin -i`
3. select yarn as package manager
4. `cd my-admin`
5. `yarn run dev`
6. Make sure you are using MUI v7 (`yarn list @mui/material`) and that the app runs fine

**Testing with npm/pnpm/bun**

It's more complex to test, because react-admin 5.7.3 still requires MUI <= 6, and react-admin 5.8.0 has not been released yet.

So you'd need to build the react-admin packages, and then force using this local version when resolving the dependencies.

To me this test adds very little to the test with yarn (basically it comes down to testing the resolution algorithm of the selected package manager).

## Additional Checks

~~- [ ] The **documentation** is up to date~~

### DIFF
--- a/packages/create-react-admin/templates/common/package.json
+++ b/packages/create-react-admin/templates/common/package.json
@@ -12,8 +12,8 @@
     "dependencies": {
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.0",
-        "@mui/icons-material": "^6.4.0",
-        "@mui/material": "^6.4.0",
+        "@mui/icons-material": "^7.0.1",
+        "@mui/material": "^7.0.1",
         "react": "^19.0.0",
         "react-admin": "^5.7.0",
         "react-dom": "^19.0.0",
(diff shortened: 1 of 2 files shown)


packages/create-react-admin/templates/common/package.json
```diff
@@ -12,8 +12,8 @@
     "dependencies": {
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.0",
-        "@mui/icons-material": "^6.4.0",
-        "@mui/material": "^6.4.0",
+        "@mui/icons-material": "^7.0.1",
+        "@mui/material": "^7.0.1",
```
slax57, line 16: One could argue we should also upgrade the minimum required version of `react-admin` to 5.8, but the latest version will be installed in any case, besides this would break the CI tests :grimacing:

djhi: And it will be done by the release script ;)

djhi (approved):

PR #10690, opened 2025-04-22
Add `<InPlaceEditor>` for edit-in-place

fzaninotto:
## Problem

To edit a value from a record, users have to leave the current screen to go to an Edit view (or use `EditInDialog`, but they will also lose context). 

## Solution

`<InPlaceEditor>` renders a field from the current record. On click, it switches to an editable state, allowing the user to change the value directly.

https://github.com/user-attachments/assets/ef71d699-3967-4e40-a897-0ae1ea9bb875

Use this component to let users edit parts of a record directly in the list or detail view. It is useful for quick edits without navigating to a separate edit page.

```jsx
import { Show, InPlaceEditor } from 'react-admin';
import { Stack, Box, Typography } from '@mui/material';

const CustomerShow = () => (
    <Show>
        <Stack direction="row" spacing={2}>
            <AvatarField />
            <CustomerActions />
            <Box sx={{ display: 'flex', justifyContent: 'center' }}>
                <Typography>Phone</Typography>
                <InPlaceEditor source="phone" />
            </Box>
            <Box sx={{ display: 'flex', justifyContent: 'center' }}>
                <Typography>Email</Typography>
                <InPlaceEditor source="email" />
            </Box>
            ...
        </Stack>
    </Show>
);
```

The field changes color on hover, to indicate that it is editable. The user can cancel the edit by pressing Escape. The field is saved automatically when the user clicks outside of it or presses Enter. While it is being saved, the field is disabled and a loading spinner is shown. If the save fails, an error message is displayed and the original value is restored.

This component can even help replace the `<EditableDatagrid>`:

https://github.com/user-attachments/assets/340b32c5-0583-451e-83f4-2336177ebb6e

## How To Test

http://localhost:9010/?path=/story/ra-ui-materialui-input-inplaceeditor--basic

## Additional Checks

- [x] Make the doc discoverable by adding links to it


packages/ra-ui-materialui/src/input/InPlaceEditor/InPlaceEditor.stories.tsx
```diff
@@ -0,0 +1,372 @@
...
+                                                            helperText={false}
+                                                            sx={{
+                                                                width: 50,
+                                                                '& .MuiInputBase-root':
+                                                                    {
+                                                                        marginTop: 0,
+                                                                        fontSize: 14,
+                                                                    },
+                                                            }}
+                                                        />
```
slax57, line 342: nitpick: this `SelectInput` should be `required`, because if you pick the empty option, then you can never edit it back (it renders an empty `div` so you can no longer click on it)

packages/ra-ui-materialui/src/input/InPlaceEditor/InPlaceEditor.Card.stories.tsx
```diff
@@ -0,0 +1,370 @@
...
+                                        SelectProps={{ defaultOpen: true }}
+                                        sx={{
+                                            '& .MuiInput-root': {
+                                                marginTop: 0,
+                                            },
+                                            '& .MuiSelect-select': {
+                                                textAlign: 'right',
+                                            },
+                                        }}
+                                    />
```
slax57, line 271: same issue here btw (nitpick)

packages/ra-ui-materialui/src/input/InPlaceEditor/InPlaceEditor.stories.tsx
```diff
@@ -0,0 +1,372 @@
...
+    mutationMode: 'pessimistic',
+    notifyOnSuccess: false,
+    showButtons: false,
+};
+Basic.argTypes = {
+    delay: { control: { type: 'number' } },
+    updateFails: { control: { type: 'boolean' } },
+    mutationMode: {
+        control: { type: 'select' },
+        options: ['optimistic', 'pessimistic', 'undoable'],
```
slax57, line 118: When testing `undoable` mode, the record does not seem to be refreshed and I can't see my changes being saved (I see the dataProvider call in the console though)

slax57 (changes requested):
Other than that this new component works very well and the UX is :ok_hand:

packages/ra-ui-materialui/src/input/InPlaceEditor/InPlaceEditor.stories.tsx
fzaninotto: Probably because you didn't enable `notifyOnSuccess` at the same time.

fzaninotto:
Thanks for the great review, I've taken most of your comments into account and added my comments where I disagree.

slax57 (approved):

dalen:
To use this in `<EditableDatagrid>`, doesn't the `handleEdit` event handler need to do `event.stopPropagation()`?

Or could some prop be added to enable that behavior?

PR #10696, opened 2025-04-24
[Doc] improve AutocompleteInput create example

Cimanel:
## Problem

Current examples of <AutocompleteInput create> in the docs leverage useCreate, and require calling create manually with data coming from self-managed controlled inputs.

## Solution

Update the doc examples to use <CreateBase> instead.

## How To Test

the story to test: http://localhost:9010/?path=/story/ra-ui-materialui-input-autocompleteinput--inside-reference-input-with-creation-support

## Additional Checks

### DIFF
--- a/docs/AutocompleteInput.md
+++ b/docs/AutocompleteInput.md
@@ -161,7 +161,6 @@ const BookCreate = () => (
             <ReferenceInput reference="authors" source="author">
                 <AutocompleteInput
                     create={<CreateAuthor />}
-                    optionText="name"
                 />
             </ReferenceInput>
         </SimpleForm>
@@ -214,7 +213,7 @@ const CreateAuthor = () => {
                 >
                     <SimpleForm defaultValues={{ name: filter }}>
                         <TextInput source="name" helperText={false} />
-                        <TextInput source="language" helperText={false} />
+                        <TextInput source="language" helperText={false} autoFocus />
                     </SimpleForm>
                 </CreateBase>
             </DialogContent>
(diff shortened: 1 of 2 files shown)


docs/AutocompleteInput.md
```diff
@@ -153,65 +153,71 @@ To allow users to add new options, pass a React element as the `create` prop. `<
...
+const BookCreate = () => (
     <Create>
         <SimpleForm>
-            <TextInput source="title" />
-            <ReferenceInput source="category_id" reference="categories">
-                <AutocompleteInput create={<CreateCategory />} />
+            <ReferenceInput reference="authors" source="author">
+                <AutocompleteInput
+                    create={<CreateAuthor />}
+                    optionText="name"
```
slax57, line 164: nitpick(non-blocking): not sure you need `optionText="name"`, I believe it's the default

slax57 (commented):
Other than that, LGTM! You're on the right track!

djhi (approved):

docs/AutocompleteInput.md
```diff
@@ -903,43 +907,42 @@ const PostCreate = () => {
...
+                onClick={onCancel}
+                sx={theme => ({
+                    position: 'absolute',
+                    right: 8,
+                    top: 8,
+                    color: theme.palette.grey[500],
+                })}
+            >
+                <CloseIcon />
+                </IconButton>>
```
slax57, line 930: something seems off at this line, looks like you're missing a closing tag? :thinking:

slax57 (changes requested):

slax57 (approved):

Issue #10695, opened 2025-04-24
Last filter can not be removed if multiple filters added


serkanc-kpler:
**What you were expecting:**

Last filter can be removed

**What happened instead:**

Last filter is displayed

**Steps to reproduce:**

Go to the list view, add two filters (do not leave them empty), fill both of them and then try to remove one by one.

**Related code:**

- https://stackblitz.com/edit/vitejs-vite-bnahyqyk?file=package.json

**Other information:**

**Environment**

* React-admin version: 5.7.3
* Last version that did not exhibit the issue (if applicable):
* React version: 19.1.0

serkanc-kpler:
Thank you for the quick feedback! If it takes too long to get the fix, I will consider downgrading react-hook-form. Thanks for the tip!

PR #10697, opened 2025-04-24
Fix `<FilterLiveForm>` compatibility with react-hook-form 7.55.0, part 2

slax57:
## Problem

Despite https://github.com/marmelab/react-admin/pull/10657, 2 issues are still present when using RHF version >= 7.55.0:
- https://github.com/marmelab/react-admin/issues/10695
- A bug occurs when using enterprise component `DatagridAG`: Applying a filter from ag-grid, then leaving the page and coming back to it, then clearing all filters with the 'Remove all filters' button => brings back the filter added previously via ag-grid

Closes https://github.com/marmelab/react-admin/issues/10695

## Solution

It turns out the fix brought by https://github.com/marmelab/react-admin/pull/10657 was not enough.
In a nutshell, this fix made it so that the filter form is not reset if changed values only affect fields that are not part of the form. But the form was still initialized with `defaultValues` equal to the `filterValues`, so external values could still leak into the form state.
Part of the solution is then to remove the form `defaultValues`, and rely only on the `useEffect` to (selectively) apply them.

Other part of the solution was to remove the `formChangesPending` ref, as it turns out this workaround no longer works with RHF v7.55.0 onwards, because they tend to call the `useWatch` callback multiple times, even for fields whose value has not changed.
Instead, properly managing the form state and only resetting when values known to the form are modified (i.e. part 1 of the fix above) seems to be enough to avoid re-applying old values.

## How To Test

2 new stories were added:
- http://localhost:9010/?path=/story/ra-core-form-filterliveform--multiple-filter-live-form-overlapping
- http://localhost:9010/?path=/story/ra-core-form-filterliveform--with-external-changes

Also, https://github.com/marmelab/react-admin/issues/10695 can be reproduced in story http://localhost:9010/?path=/story/ra-ui-materialui-list-filter-filterbutton--basic

Unit tests were added to cover all known issues.

=> Hence, to test the fix, I recommend:
1. Playing around with the added stories (the unit tests are already run by the CI)
2. Then, update RHF to its latest version by running `yarn up react-hook-form`, and run the tests (`make test-unit`), then play with the added stories

**Note:** You will notice that some tests **not related to filters** are failing when using RHF v7.46.1, which is something we'll need to address in a future PR, but all tests related to filters should pass with this PR.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/form/FilterLiveForm.spec.tsx
+++ b/packages/ra-core/src/form/FilterLiveForm.spec.tsx
@@ -1,12 +1,14 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { getFilterFormValues } from './FilterLiveForm';
 import {
     Basic,
     GlobalValidation,
     MultipleFilterLiveForm,
+    MultipleFilterLiveFormOverlapping,
     MultipleInput,
     ParseFormat,
     PerInputValidation,
+    WithExternalChanges,
 } from './FilterLiveForm.stories';
 import React from 'react';
 import { WithFilterListSection } from '../../../ra-ui-materialui/src/list/filter/FilterLiveForm.stories';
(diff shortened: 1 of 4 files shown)


slax57:
Some E2E tests are failing, back to WIP

slax57:
Fixed, back to RFR!

djhi (commented):
This looks good but maybe we should update react-hook-form to its latest version?

slax57:
I agree but we should do that on `next`. Besides, as I said, some other tests are failing which we'll need to investigate (soon).

PR #10699, opened 2025-04-25
docs: update field component example to use modern React patterns

dhfra21:
## Problem
     The documentation currently recommends using defaultProps for function components, which is being removed in React 19.

     ## Solution
     Update the example to use modern React patterns with parameter destructuring and default values.

     ## Additional Checks
     - [x] The PR targets `master` for a documentation fix
     - [x] The documentation is up to date

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -482,12 +482,10 @@ Here is a custom field displaying the full name:
 ```jsx
 import { useRecordContext } from 'react-admin';
 
-export const FullNameField = (props) => {
+export const FullNameField = ({ label = 'Name', ...props }) => {
     const record = useRecordContext(props);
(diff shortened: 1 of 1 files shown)


docs/Fields.md
```diff
@@ -482,12 +482,10 @@ Here is a custom field displaying the full name:
 ```jsx
 import { useRecordContext } from 'react-admin';
 
-export const FullNameField = (props) => {
+export const FullNameField = ({ label = 'Name', ...props }) => {
```
fzaninotto, line 485: This doesn't work. `Datagrid` and `SimpleShowLayout` inspect the label prop of their child, and a default value defined as you describe cannot be read by the parent.

fzaninotto:

PR #10700, opened 2025-04-25
[DOC] Rewrite autocompletearray input create doc example

Cimanel:
## Problem

Current examples of <AutocompleteArrayInput create> in the docs leverage useCreate, and require calling create manually with data coming from self-managed controlled inputs. 
https://marmelab.com/react-admin/AutocompleteArrayInput.html#creating-new-choices:~:text=OptionRenderer%20/%3E%3B-,Creating%20New%20Choices,-The%20%3CAutocompleteArrayInput%3E

## Solution

## How To Test

http://localhost:9010/?path=/story/ra-ui-materialui-input-autocompletearrayinput--inside-reference-array-input-with-creation-support

## Additional Checks

### DIFF
--- a/docs/AutocompleteArrayInput.md
+++ b/docs/AutocompleteArrayInput.md
@@ -714,6 +714,7 @@ Use the `create` prop when you want a more polished or complex UI. For example a
 import {
     AutocompleteArrayInput,
     Create,
+    CreateBase,
     ReferenceArrayInput,
     SimpleForm,
     TextInput,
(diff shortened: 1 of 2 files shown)


docs/AutocompleteArrayInput.md
```diff
@@ -746,43 +749,45 @@ const PostCreate = () => {
...
+                    resource="tags"
+                    mutationOptions={{
+                        onSuccess: tag => {
+                            onTagCreate(tag);
+                        },
+                    }}
+                >
+                    <SimpleForm defaultValues={{ name: filter }}>
+                        <TextInput
+                            source="tag"
```
slax57, line 784: Shouldn't the `source` be `"name"`? :thinking:

slax57 (changes requested):

slax57 (approved):

PR #10704, opened 2025-04-26
fix: translate category names in sidebar to French

### LINKED ISSUE #10703: Category names in sidebar are not translated to French
When switching the application language to French, the category names in the sidebar remain in English (e.g., Animals, Beard, Business, etc.), even though the section title is translated.

> Expected behavior:
> Category names should be displayed in French (e.g., Animaux, Barbe, Affaires, etc.)

> Steps to reproduce:
> 1. Switch the app language to French.
> 2. Observe the category names in the sidebar under “CATÉGORIES”.

Image

ghadabezine:
>Adds a categoryTranslations object in Aside.tsx to provide French translations for category names.
 >Uses this mapping to display the correct French name in the sidebar when the app is in French.
 > How to test:
> 1. Switch the app language to French.
> 2. Check the sidebar under “CATÉGORIES”.
> 3. Category names should now appear in French.
Closes #10703

### DIFF
--- a/examples/demo/src/products/Aside.tsx
+++ b/examples/demo/src/products/Aside.tsx
@@ -14,6 +14,22 @@ import {
 import { Category } from '../types';
 import { humanize } from 'inflection';
 
+const categoryTranslations: { [key: string]: string } = {
+    animals: 'Animaux',
+    beard: 'Barbe',
+    business: 'Affaires',
+    cars: 'Voitures',
+    city: 'Ville',
+    flowers: 'Fleurs',
+    food: 'Nourriture',
+    nature: 'Nature',
+    people: 'Personnes',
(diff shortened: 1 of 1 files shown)


fzaninotto:
This is not the right way to translate entities. Categories are records coming from the server. the client doesn't know if there will me more or less categories, or if their wording has changed. Therefore you can't use their name as a translation key on the frontend. 

The right "fix" for that would be to update the Categories model to store translations, and use [TranslatableFields](https://marmelab.com/react-admin/TranslatableFields.html) and [TranslatableInputs](https://marmelab.com/react-admin/TranslatableInputs.html) to render and edit them. 

But that's beyond the scope of this demo, which isn't a real app.

PR #10710, opened 2025-04-28
[DOC] selectInput create example + storie

Cimanel:
## Problem

Current examples of <SelectInput create> in the docs leverage useCreate, and require calling create manually with data coming from self-managed controlled inputs. 
https://marmelab.com/react-admin/SelectInput.html#create

## Solution

## How To Test

http://localhost:9010/?path=/story/ra-ui-materialui-input-selectinput--inside-reference-input-with-creation-support

## Additional Checks

### DIFF
--- a/docs/SelectInput.md
+++ b/docs/SelectInput.md
@@ -185,7 +185,7 @@ import {
 } from '@mui/material';
 
 const CreateAuthor = () => {
-    const { filter, onCancel, onCreate } = useCreateSuggestionContext();
+    const { onCancel, onCreate } = useCreateSuggestionContext();
 
     const onAuthorCreate = author => {
         onCreate(author);
--- a/docs/SelectInput.md
+++ b/docs/SelectInput.md
@@ -216,7 +216,7 @@ const CreateAuthor = () => {
                         },
                     }}
                 >
-                    <SimpleForm defaultValues={{ name: filter }}>
+                    <SimpleForm>
                         <TextInput source="name" helperText={false} />
                         <TextInput source="language" helperText={false} autoFocus />
                     </SimpleForm>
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/input/SelectInput.stories.tsx
```diff
@@ -1,24 +1,33 @@
...
+    CreateBase,
+    Resource,
+    TestMemoryRouter,
+    required,
+    useGetList,
+} from 'ra-core';
+import polyglotI18nProvider from 'ra-i18n-polyglot';
+import englishMessages from 'ra-language-english';
+import * as React from 'react';
+import { Admin, AdminContext } from 'react-admin';
```
slax57, line 21: Please do not import from `'react-admin'` as it will create a circular dependency

packages/ra-ui-materialui/src/input/SelectInput.stories.tsx
```diff
@@ -602,6 +611,99 @@ export const InsideReferenceInputWithError = () => (
     </TestMemoryRouter>
 );
 
+const CreateAuthor = () => {
+    const { filter, onCancel, onCreate } = useCreateSuggestionContext();
```
slax57, line 615: linter warning at this line

docs/SelectInput.md
```diff
@@ -158,66 +158,70 @@ To allow users to add new options, pass a React element as the `create` prop. `<
...
-    DialogActions,
     DialogContent,
-    TextField,
+    DialogTitle,
+    IconButton,
 } from '@mui/material';
 
-const CreateCategory = () => {
+const CreateAuthor = () => {
     const { filter, onCancel, onCreate } = useCreateSuggestionContext();
```
slax57, line 188:
```suggestion
    const { onCancel, onCreate } = useCreateSuggestionContext();
```

slax57:
The workaround in the stories can be removed once https://github.com/marmelab/react-admin/pull/10715 is merged

djhi (approved):

PR #10715, opened 2025-04-30
Fix `<AutocompleteInput>` and `<SelectInput>` renders undefined instead of the `createLabel` when `optionText` is a function or a `recordRepresentation` is set

slax57:
## Problem

Applicable to default `<SelectInput>` and also `<AutocompleteInput>` when a custom `createLabel` is set: when there is a `recordRepresentation` set on the referenced resource, or a function is passed to `optionText`, then the create label is rendered as "undefined undefined":

Screenshot_20250430_120935

This is due to `useSupportCreateSuggestion->getCreateItem`, which only supports the case when `optionText` is a `string` and otherwise falls back to setting the `name` prop.

However when rendering the choices, the `useChoices->getChoiceText` function always leverages `optionText` (or `recordRepresentation`, which is fed to `optionText`) if present, leading to an invalid choice text (often "undefined undefined").

## Solution

The solution this PR applies is to modify `useChoices->getChoiceText` to detect when the choice is the "create" choice, and in this case render the field targeted by `optionText` if it is a string, or its `name` field otherwise.

## How To Test

1. unit tests
2. stories (including Controls)
  i. http://localhost:9010/?path=/story/ra-ui-materialui-input-selectinput--create-label
  ii. http://localhost:9010/?path=/story/ra-ui-materialui-input-autocompleteinput--create-label

## Additional Checks

### DIFF
--- a/packages/ra-core/src/form/choices/useChoices.tsx
+++ b/packages/ra-core/src/form/choices/useChoices.tsx
@@ -55,8 +55,10 @@ export const useChoices = ({
     const getChoiceText = useCallback(
         choice => {
             if (choice?.id === createValue || choice?.id === createHintValue) {
-                // For the create choice, we always use the name prop as text
-                return get(choice, 'name');
+                return get(
+                    choice,
+                    typeof optionText === 'string' ? optionText : 'name'
+                );
             }
 
             if (isValidElement<{ record: any }>(optionText)) {
--- a/packages/ra-core/src/form/choices/useChoices.tsx
+++ b/packages/ra-core/src/form/choices/useChoices.tsx
@@ -27,6 +27,8 @@ export interface UseChoicesOptions {
     optionText?: OptionText;
     disableValue?: string;
     translateChoice?: boolean;
+    createValue?: string;
+    createHintValue?: string;
 }
 
 /*
@@ -45,11 +47,18 @@ export const useChoices = ({
     optionValue = 'id',
     disableValue = 'disabled',
     translateChoice = true,
+    createValue = '@@ra-create',
+    createHintValue = '@@ra-create-hint',
 }: UseChoicesOptions) => {
     const translate = useTranslate();
 
     const getChoiceText = useCallback(
         choice => {
+            if (choice?.id === createValue || choice?.id === createHintValue) {
+                // For the create choice, we always use the name prop as text
+                return get(choice, 'name');
+            }
+
             if (isValidElement<{ record: any }>(optionText)) {
                 return (
                     <RecordContextProvider value={choice}>
@@ -68,7 +77,7 @@ export const useChoices = ({
                   ? translate(String(choiceName), { _: choiceName })
                   : String(choiceName);
         },
-        [optionText, translate, translateChoice]
+        [createHintValue, createValue, optionText, translate, translateChoice]
     );
 
     const getChoiceValue = useCallback(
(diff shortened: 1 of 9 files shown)


packages/ra-ui-materialui/src/input/useSupportCreateSuggestion.tsx
```diff
@@ -62,7 +61,7 @@ export const useSupportCreateSuggestion = (
                             ? createHintValue
                             : createValue,
                 },
-                typeof optionText === 'string' ? optionText : 'name',
```
djhi, line 65: I believe this was made like this for people providing their own `renderOption` in `AutocompleteInput`. Now they'll have to check if the item is the createItem like us

slax57: Right! I tried to think about all cases that would prevent me from simplifying the code to using only the `name` field, but forgot this one!
I've reverted this change and made the `useChoices` hook compatible with all cases (hopefully).

packages/ra-core/src/form/choices/useChoices.tsx
```diff
@@ -27,6 +27,8 @@ export interface UseChoicesOptions {
     optionText?: OptionText;
     disableValue?: string;
     translateChoice?: boolean;
+    createValue?: string;
+    createHintValue?: string;
```
djhi, line 31: Aren't you introducing _new_ props indirectly to the `<AutocompleteInput>` and `<SelectInput>`? That should target the `next` branch

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -450,7 +486,7 @@ export const CreateDialog = () => (
     </Wrapper>
 );
 
-const CreateLabelInput = () => {
+const CreateLabelInput = ({ optionText }) => {
```
djhi, line 489: I know there are _just_ stories but we can do better than implicit any

packages/ra-ui-materialui/src/input/AutocompleteInput.stories.tsx
```diff
@@ -459,25 +495,53 @@ const CreateLabelInput = () => {
...
                 return newOption;
             }}
             createLabel="Start typing to create a new item"
+            optionText={optionText}
         />
     );
 };
 
-export const CreateLabel = () => (
+export const CreateLabel = ({ optionText }: any) => (
```
djhi, line 526: or explicit

packages/ra-ui-materialui/src/input/SelectInput.stories.tsx
```diff
@@ -321,31 +321,68 @@ export const OnCreate = () => {
     );
 };
 
-export const CreateLabel = () => {
-    const categories = [
-        { name: 'Tech', id: 'tech' },
-        { name: 'Lifestyle', id: 'lifestyle' },
+export const CreateLabel = ({ optionText }: any) => {
```
djhi, line 324: Same

djhi (changes requested):

packages/ra-core/src/form/choices/useChoices.tsx
slax57: Technically you're correct, but:
- For `<AutocompleteInput>` the props were already there, but non-documented
- For `<SelectInput>` the `createValue` prop was already there (non-documented), only the `createHintValue` prop is new
In both cases, these are undocumented props, so no need for a minor version IMO.

djhi (approved):

PR #10717, opened 2025-04-30
Fix prefetched resources are not used by ReferenceField

fzaninotto:
## Problem

`<ReferenceField>` relies on `getManyAggregate` to call the data provider. Each instance will call a `getMany(resource, [id])`, that react-admin will not  pass to the data provider, -but aggregate into a single `getMany(resource, [id1, id2, id3])`.

The prefetch feature pre-populates the `getMany(resource, [id1, id2, id3])` cache, but this isn't enough to prevent the actual data provider query.

## Solution

Prefetched records should be used to populate the cache of *individual*  getMany queries.

### DIFF
--- a/packages/ra-core/src/dataProvider/populateQueryCache.ts
+++ b/packages/ra-core/src/dataProvider/populateQueryCache.ts
@@ -37,6 +37,11 @@ export const populateQueryCache = ({
                 record,
                 { updatedAt }
             );
+            queryClient.setQueryData(
+                [resource, 'getMany', { ids: [String(record.id)] }],
+                [record],
+                { updatedAt }
+            );
         });
         const recordIds = data[resource].map(record => String(record.id));
         queryClient.setQueryData(


packages/ra-core/src/dataProvider/useGetOne.ts
```diff
@@ -64,24 +68,22 @@ export const useGetOne = <RecordType extends RaRecord = any, ErrorType = Error>(
     const onErrorEvent = useEvent(onError);
     const onSettledEvent = useEvent(onSettled);
 
-    const result = useQuery<RecordType, ErrorType>({
+    const result = useQuery<GetOneResult<RecordType>, ErrorType>({
```
fzaninotto, line 71: unfortunately, this is a breaking change: any other hook writing into the `getOne` cache (e.g., `useUpdate`) would need to be updated...

fzaninotto:
This can't be fixed in a backward compatible way. It will have to wait for the next major version.

PR #10719, opened 2025-04-30
fix: add default options to NumberField for better decimal precision

### LINKED ISSUE #10718: [NumberField] use a (higher) default for `maximumFractionDigits` option
NumberField is using, according to documentation, Intl.NumberFormat() which by mdn docs is "the default for plain number formatting is the larger of minimumFractionDigits and 3" i.e. 3
Example in Datagrid:
`0.001` => `0.001`
`0.0001` => `0` // very confusing for user, thinking the value is 0

Provide by default the maximum value for `maximumFractionDigits` (100) to Intl.NumberFormat so that user doesn't see `0` when the value is `0.0001`

transforming manually, tedious

saloua-ch:
## Description
Fixes the issue with small decimal numbers being displayed as 0 in the NumberField component.

## Changes
- Added default options to NumberField component to handle small decimal numbers correctly
- Updated documentation to reflect the new default options
- Added test case to verify the behavior with small decimal numbers

## Before
Small decimal numbers like 0.0001 would be displayed as 0 due to Intl.NumberFormat's default behavior.

## After
Small decimal numbers are now displayed correctly with their full precision by default.

## Related Issue
Fixes #10718

### DIFF
--- a/docs/NumberField.md
+++ b/docs/NumberField.md
@@ -63,7 +63,7 @@ If Intl is not available, `<NumberField>` outputs numbers as is (and ignores the
 | Prop        | Required | Type               | Default | Description                                                                      |
 | ----------- | -------- | ------------------ | ------- | -------------------------------------------------------------------------------- |
 | `locales`   | Optional | string             | ''      | Locale to use for formatting. Passed as first argument to `Intl.NumberFormat()`. |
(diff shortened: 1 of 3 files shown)


ogroppo (approved):
On the grounds NumberFormat is the underlying formatter, this is the solution that would match user's expectation 👍🏻

fzaninotto:
This is a breaking change, and we won't implement it for reasons explained in #10718.

PR #10720, opened 2025-05-01
fix: ensure NumberField displays small decimal values correctly

dhfra21:
## Problem

The `NumberField` component was not displaying small decimal values correctly. By default, it was truncating or rounding very small numbers (e.g., 0.0001) to 0, which could lead to data loss and incorrect representation of values. This was particularly problematic for scientific, financial, or precise measurements where small decimal values are important.

## Solution

I've modified the `NumberField` component to set a default `maximumFractionDigits` value of 100 in the options object. This ensures that small decimal values are displayed with their full precision by default. The change is implemented in a way that:

1. Preserves backward compatibility
2. Allows users to override the default behavior if needed
3. Maintains the existing formatting capabilities for other number types (currencies, percentages, etc.)

## How To Test

1. Create a record with a small decimal value:  `const record = { id: 1, value: 0.0001 };`
2. Render the NumberField component:  `<NumberField source="value" record={record} />`
3. Verify that the small decimal value is displayed correctly as "0.0001"
4. Test overriding the default behavior: ```
<NumberField 
    source="value" 
    record={record} 
    options={{ maximumFractionDigits: 2 }} 
/>
```
## Additional Checks

### DIFF
--- a/docs/NumberField.md
+++ b/docs/NumberField.md
@@ -90,7 +90,8 @@ When not provided, it uses the browser locale.
 
 Options passed to `Intl.NumberFormat()`. See [the Intl.NumberFormat documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) for the `options` prop syntax.
 
-{% raw %}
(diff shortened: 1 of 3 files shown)


fzaninotto:
As explained in #10718, we don't think there is a bug to fix, and you can override that default in a custom component, so we won't merge this PR.

Issue #10728, opened 2025-05-05
Theme Change Does Not Apply Instantly


ghadabezine:
Description
When switching themes using the theme switcher (e.g., in the demo app), the new theme is not applied immediately. The UI only updates to the new theme after a manual page refresh. This is unexpected—theme changes should be reflected instantly.

Steps to Reproduce:
1.Start the demo app.
2.Use the theme switcher in the AppBar to select a different theme.
3.Observe that the UI does not update to the new theme.
4.Refresh the page.
5.The new theme is now applied.

Expected Behavior:
The theme should update immediately after the user changes it, without requiring a page refresh.

Actual Behavior:
The theme only updates after a manual page refresh.

This is a video representation:
https://streamable.com/xcy7zs

fzaninotto:
That's a regression indeed, thanks for reporting. 

It is visible with a local checkout of the demo, not on https://marmelab.com/react-admin-demo/ (which is not up to date).

fzaninotto:
According to `git bisect`, this regression was introduced by 983d257466772180136e1bdde004bf52990be1f8 (!)

ghadabezine:
While the root cause may not be only in 983d257, the fact remains: users upgrading to 5.8.0 are affected, and the fix in this PR (key={themeName}) restores the expected behavior without introducing breaking changes.
It’s a minimal, non-invasive patch that ensures proper re-rendering when the theme changes, aligning with how react-admin used to behave.

fzaninotto:
The problem was introduced by updating react-router in #10712. We traced back the culprit to [react-router 7.1.4](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#v714), the first release to show the problem. By forcing the demo to react-router 7.1.3, the theme swapper works again.

PR #10729, opened 2025-05-05
fix: ensure theme changes apply instantly without refresh

ghadabezine:
this PR solves #10728

Theme Switching Now Applies Immediately

Added key={themeName} to the <Admin /> component to ensure it re-renders when the stored theme changes.

This fixes the problem where switching between light/dark/soft themes didn’t reflect without a manual page reload.

How to test:

Start the demo app.
>Use the theme switcher in the AppBar to select a different theme.
>The new theme is now applied.


fzaninotto:
Thanks, this works, but I believe that right fix is to look for the root cause of the original regression, which probably has other repercussions. We'll need to investigate further.

PR #10736, opened 2025-05-10
[Doc] Add a new third-party component `huyanhvn/react-admin-clipboard-field`

huyanhvn:
Add a new third-party component huyanhvn/react-admin-clipboard-field.

## Problem

None. This is just a suggestion.

## Solution

Hello team, I've created this wrapper for TextField that copies the text into clipboard. I've been finding it quite useful. If you think it can be useful for others too, please accept this suggestion into the docs. Thank you.

## How To Test

Just copy `src/clipboard-field.tsx` to your source directory and use ClipboardField anywhere you would use a TextField.

## Additional Checks

- [ ] The PR includes **unit tests** (if not possible, describe why): No, because this is a doc update.
- [ ] The PR includes one or several **stories** (if not possible, describe why): No, because this is a suggestion.

### DIFF
--- a/docs/Fields.md
+++ b/docs/Fields.md
@@ -615,10 +615,10 @@ const AuthorField = () => {
 
 You can find components for react-admin in third-party repositories.
 
-- [OoDeLally/react-admin-clipboard-list-field](https://github.com/OoDeLally/react-admin-clipboard-list-field): a quick and customizable copy-to-clipboard field.
 - [MrHertal/react-admin-json-view](https://github.com/MrHertal/react-admin-json-view): JSON field and input for react-admin.
 - [alexgschwend/react-admin-color-picker](https://github.com/alexgschwend/react-admin-color-picker): a color field.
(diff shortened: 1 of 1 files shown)


fzaninotto:
We already have one such field (three lines before the line you added), so my advice is to collaborate with the other contributor rather than compete against them - it doesn't bring value to have two small packahgs like that for react-admin users.

huyanhvn:
Actually the 2 fields behave differently and are used for different purposes.

fzaninotto:
Can you explain the differences in more detail? Because it's not obvious from the line you've added.

huyanhvn:
Sure. Here is my understanding:

`react-admin-clipboard-list-field`: a custom field that references a field in a record, then displays various string versions of that field and allows you to copies each into clipboard.
`react-admin-clipboard-field`: a wrapper around TextField, except with a copy-to-clipboard icon upon hovering that allows to copy field text into clipboard. Use it just like how you'd use TextField (on List page, Show page, forms, etc.).

slax57:
Indeed, to me they are different. Reopening.

slax57 (approved):

PR #10749, opened 2025-05-16
Add RecordField component

fzaninotto:
## Problem

`<SimpleShowLayout>` uses child inspection to decorate its children with labels. 

This isn't robust, as you just need to wrap a child component to break this decoration:

```jsx
consty PostShow = () => (
  <Show>
    <SimpleShowLayout>
      <TextField source="id" />
      <CanAccess resource="posts.title" action="read">
        <TextField source="title" />
      </CanAccess>
    </SimpleShowLayout>
  </Show>
);
```

## Solution

Provide a new clomponent to render a field and its label, `RecordField`:

```jsx
import { Show, RecordField } from 'react-admin';
import { Stack } from '@mui/material';

export const BookShow = () => (
    <Show>
        <Stack>
            <RecordField source="id" />
            <CanAccess resource="posts.title" action="read">
              <RecordField source="title" />
           </CanAccess>
        </Stack>
    </Show>
);
```

## How To Test

Use the fields/RecordField story

## Additional Checks


WiXSL (changes requested):

packages/ra-ui-materialui/src/field/RecordField.tsx
```diff
@@ -0,0 +1,170 @@
...
+    [`& .${RecordFieldClasses.label}`]: {
+        fontSize: '0.75em',
+        marginBottom: '0.2em',
+        color: (theme.vars || theme).palette.text.secondary,
+    },
+    [`&.${RecordFieldClasses.inline} .${RecordFieldClasses.label}`]: {
+        fontSize: '0.875rem',
+        display: 'block',
+        minWidth: 150,
+    },
```
slax57, line 147: FMI, why are you using `em` units with the default variant, but `rem` units for the inline variant?

packages/ra-ui-materialui/src/field/RecordField.tsx
```diff
@@ -0,0 +1,170 @@
...
+    [`& .${RecordFieldClasses.value}`]: {
+        flex: 1,
+    },
+}));
+
+declare module '@mui/material/styles' {
+    interface ComponentNameToClassKey {
+        RaRecordField: 'root' | 'content' | 'button' | 'icon';
+    }
```
slax57, line 156:
```suggestion
    interface ComponentNameToClassKey {
        RaRecordField: 'root' | 'label' | 'value' | 'inline';
    }
```

slax57 (changes requested):
other than that LGTM

docs/RecordField.md
```diff
@@ -0,0 +1,411 @@
...
+<RecordField source="title" empty="Missing title" />
+```
+
+`empty` also accepts a translation key, so you can have a localized string when the field is empty:
+
+```jsx
+<RecordField source="title" empty="resources.books.fields.title.missing" />
+```
+
+If you use the `render` prop, you can even use a React element as `empty` value.
```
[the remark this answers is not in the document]
slax57: I think it's no big deal for now, so this is non-blocking. But just for the record, my idea was that `RecordField` would be responsible for rendering the `empty` element, not `TextField`, nor any other component passed as `children` or `field`.

PR #10752, opened 2025-05-20
[Doc] Remove bad indication for "yarn create react-admin@latest"

erwanMarmelab:
## Problem

`yarn create react-admin@latest` doesn't work but we document it

## Solution

document `yarn create react-admin@latest` instead

## Additional Checks


fzaninotto:
You need to update the CreateReactAdmin.md and the Readme.md, too.

slax57 (changes requested):
I see more in `Readme.md` and `cli.tsx`

slax57 (approved):

PR #10762, opened 2025-05-28
Create a codemod to replace `Datagrid` by `DataTable `

erwanMarmelab:
## Problem

We have too much `Datagrid` that we want to transform to `DataTable` to do it manually

## Solution

Create a codemod to do it magically

## How To Test

1. Create a `test.tsx` and fill it with the example you want to transform. e.g.: the following one
```tsx
import { List, Datagrid, TextField } from 'react-admin';

const PostList = () => (
    <List>
        <Datagrid>
            <TextField source="id" label="Post number" />
            <TextField source="title" />
            <TextField source="body" />
        </Datagrid>
    </List>
);

export default PostList;
```
2. run the following command:
```txt
npx jscodeshift test.tsx \
    --extensions=ts,tsx \
    --parser=tsx \
    --transform=./packages/ra-core/codemods/replace-Datagrid-DataTable.ts
```

## To Do

- [ ] Write unit tests
- Create the codemod
    - [ ] replace imports
    - [ ] replace the `Datagrid` component
    - [ ] wrap the field with a `<DataTable.Col source="xxx">`
    - [ ] keep the `Datagrid` props
    - [ ] delete the `optimized` prop of the `Datagrid` component
    - [ ] replace the `rowStyle` prop of the `Datagrid` component by `rowSx`
    - [ ] replace the `& .RaDatagrid-xxxx` key of the `sx` prop of the `Datagrid` component by `& .RaDataTable-xxxx`
    - [ ] don't wrap a field if it is a `NumberField` **??? -> if yes, what do we want to do for the props of the initial field ?**

## Additional Checks

### DIFF
--- a/packages/ra-core/codemods/__testfixtures__/replace-Datagrid-DataTable-Basic.input.tsx
+++ b/packages/ra-core/codemods/__testfixtures__/replace-Datagrid-DataTable-Basic.input.tsx
@@ -1,0 +1,15 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import * as React from 'react';
+import { List, Datagrid, TextField } from 'react-admin';
+
+const PostList = () => (
+    <List>
+        <Datagrid>
+            <TextField source="id" label="Post number" />
+            <TextField source="title" />
+            <TextField source="body" />
+        </Datagrid>
+    </List>
+);
+
+export default PostList;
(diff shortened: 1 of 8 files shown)


dalen:
It would be great if this handled `DataGridConfigurable` with `omit` -> `hiddenColumns` as well.

Also, `sortable={false}` -> `disableSort` seems to be missing.

slax57:
Hi @dalen 
Thank you so much for this valuable feedback!
`sortable={false}` seems to be something that was forgotten indeed, so I'll create an issue for that right away!

`DatagridConfigurable` on the other hand is outside the scope of what we envisioned for the first version of the codemod, but it certainly would be a nice addition! -> Would you like to open a feature request to see if others would be interested?
And if you are willing to, you can also open a PR implementing that feature!
Thanks!

slax57:
@dalen on second look I realize that supporting `sortable` and `sortByOrder` was also out of scope. So both ideas are to be considered as new features to the codemod. (But they remain good ideas nevertheless!)
Thanks!

PR #10177, opened 2024-09-01
Reject promise in checkError with logoutUser: false and without redirectTo causes an error #10172

rktamil:
Add null check for redirect and When redirect  is not provided, I returned false and let the called handles the redirect.

This may not be ideal solution, happy to  discuss as this is my my first contribution.

Fixes #10172

PR #10763, opened 2025-05-30
fix(useLogoutAccessDenied): handle errors without redirect

carloshv93:
## Problem
Undefined redirectTo values throws error
Fix #10177 

## Solution
Create a handleRedirect function to properly handle either valid URL or undefined URLS
## Steps to reproduce
Follow steps from #10177 

## Additional Checks

## Evidence
image

### DIFF
--- a/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
+++ b/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
@@ -280,14 +280,17 @@ describe('useLogoutIfAccessDenied', () => {
                 </AuthContext.Provider>
             </TestMemoryRouter>
         );
-
         await waitFor(() => {
             expect(authProvider.logout).toHaveBeenCalledTimes(0);
             expect(notify).toHaveBeenCalledWith('Access denied', {
                 type: 'error',
             });
             expect(notify).toHaveBeenCalledTimes(1);
-            expect(screen.queryByText('Login page')).toBeNull();
+        });
+        await waitFor(() => {
+                expect(screen.queryByText('Login page')).toBeNull();
+            }, {
+            timeout: 2000
         });
     });
 });
--- a/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
+++ b/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
@@ -287,10 +287,5 @@ describe('useLogoutIfAccessDenied', () => {
             });
             expect(notify).toHaveBeenCalledTimes(1);
         });
-        await waitFor(() => {
-                expect(screen.queryByText('Login page')).toBeNull();
-            }, {
-            timeout: 2000
-        });
     });
 });
--- a/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
+++ b/packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
@@ -258,4 +258,36 @@ describe('useLogoutIfAccessDenied', () => {
             expect(screen.queryByText('unauthorized')).not.toBeNull();
         });
     });
+
+    it('should stay on same page if error have no redirectTo', async () => {
+        render(
+            <TestMemoryRouter>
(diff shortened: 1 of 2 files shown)


djhi (changes requested):
Thanks for the PR! Would you mind adding some too?

carloshv93:
> Thanks for the PR! Would you mind adding some too?
@djhi, I might not have understood what you're requesting. Could you clarify your change request a bit more? please

djhi:
> > Thanks for the PR! Would you mind adding some too?
> > @djhi, I might not have understood what you're requesting. Could you clarify your change request a bit more? please

Oups, sorry for the typo... Can you add some unit tests as well?

carloshv93:
> > > Thanks for the PR! Would you mind adding some too?
> > > @djhi, I might not have understood what you're requesting. Could you clarify your change request a bit more? please
> 
> Oups, sorry for the typo... Can you add some unit tests as well?

Unit test added

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
```diff
@@ -258,4 +258,36 @@ describe('useLogoutIfAccessDenied', () => {
...
+        );
+
+        await waitFor(() => {
+            expect(authProvider.logout).toHaveBeenCalledTimes(0);
+            expect(notify).toHaveBeenCalledWith('Access denied', {
+                type: 'error',
+            });
+            expect(notify).toHaveBeenCalledTimes(1);
+            expect(screen.queryByText('Login page')).toBeNull();
+        });
```
slax57: This test is probably not very robust, because you don't wait before checking that `'Login page'` is null. I suspect this tests would pass even without the fix. Can you wait for some time to let the redirection happen (if any) before checking that we are not redirected to the login page?

carloshv93: I wait for it for 2 secs now, so we can ensure there is no redirect.

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -102,14 +111,8 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
...
-                    } else {
-                        if (redirectTo.startsWith('http')) {
-                            // absolute link (e.g. https://my.oidc.server/login)
-                            window.location.href = redirectTo;
-                        } else {
-                            // internal location
-                            navigate(redirectTo);
-                        }
+                    } else if (redirectTo) {
+                        handleRedirect(redirectTo);
```
slax57, line 118: Also, I realize the dependency array of the `useCallback` is no longer up-to-date: you need to remove `navigate` and instead add `handleRedirect`

packages/ra-core/src/auth/useLogoutIfAccessDenied.ts
```diff
@@ -43,6 +43,15 @@ const useLogoutIfAccessDenied = (): LogoutIfAccessDenied => {
...
     const notify = useNotify();
     const navigate = useNavigate();
+
+    const handleRedirect = (url: string) => {
+        if (url.startsWith('http')) {
+            window.location.href = url;
+        } else {
+            navigate(url);
+        }
+    };
```
slax57: Which means you probably need to memoize this callback to avoid recreating a new one on each rerender

slax57 (changes requested):

packages/ra-core/src/auth/useLogoutIfAccessDenied.spec.tsx
slax57: Thanks, but this is not what I meant. `'Login page'` will be `null` immediately because you are not on the login page at first. Giving more time to react testing lib to test that is not gonna change the fact that the test will be valid immediately.

What you need to do instead is wait some time (e.g. 50 ms) **_before_** trying to check if  `'Login page'`  is present on the screen.

You can try this the other way around: try to write a test that fails without the fix, then re-apply the fix and confirm the test passes.

PR #10765, opened 2025-06-02
Fix `<SaveButton>` is not enabled when the form is prefilled (from `<CloneButton>` for instance)

### LINKED ISSUE #9217: Save button is disabled on form when using the clone button
**What you were expecting:**
In version 3 of react admin when you clicked on a clone button and the create form is shown the save button was enabled. 

**What happened instead:**
From version 4 the save button is by default disabled. So, the user must change a value on the form before being able to save

djhi:
## Problem

When users navigate to a page with a form prefilled from the router location, e.g. they clicked a `<CloneButton>`, the `<SaveButton>` is disabled by default because the form isn't dirty yet.

Fixes #9217.

## Solution

As react-hook-form does not provide a way to set the form values from an object while setting the form dirty, we made the  `<SaveButton>` check for the presence of a record in the router location too.

## How To Test

- Run the simple example
- Click any post
- Click the *Clone* button
- Check that all save buttons are enabled

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/button/SaveButton.tsx
+++ b/packages/ra-ui-materialui/src/button/SaveButton.tsx
@@ -18,6 +18,7 @@ import {
     useTranslate,
     warning,
     setSubmissionErrors,
+    useRecordFromLocation,
 } from 'ra-core';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/ra-ui-materialui/src/button/SaveButton.tsx
```diff
@@ -75,12 +76,14 @@ export const SaveButton = <RecordType extends RaRecord = any>(
...
-    // if alwaysEnable is undefined
+    // if alwaysEnable is undefined and the form wasn't prefilled
+    const recordFromLocation = useRecordFromLocation();
     const disabled = valueOrDefault(
         alwaysEnable === false || alwaysEnable === undefined
             ? undefined
             : !alwaysEnable,
-        disabledProp || !isDirty || isValidating || isSubmitting
+        (disabledProp || !isDirty || isValidating || isSubmitting) &&
+            recordFromLocation == null
```
slax57: IMO `recordFromLocation` shouldn't take precedence over everything, only `isDirty` IMO.

I suggest the following change:

```suggestion
        disabledProp ||
            (!isDirty && recordFromLocation == null) ||
            isValidating ||
            isSubmitting
```

slax57 (approved):

PR #10770, opened 2025-06-06
Add support for customization with MUI theme to fields and delete buttons

Madeorsk:
## Problem

Some fields, inputs and buttons are not customizable by themes. This adds a way to set custom styles or props on all fields and on delete button.

## Solution

Use styled components with a defined name in all fields and inputs.

Current limits:
- custom styles does not apply to empty fields

## How To Test

```shell
make storybook
```

Then view a field, input or button you want to test with and try to change the default theme. For example:

```diff
export const defaultLightTheme: ThemeOptions = deepmerge(
    defaultThemeInvariants,
    {
        //...
        components: {
            //...

+           RaEmailField: {
+               styleOverrides: {
+                   root: {
+                       color: 'red',
+                   },
+               },
+           },
        },
    }
);
```

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/button/DeleteWithUndoButton.stories.tsx
+++ b/packages/ra-ui-materialui/src/button/DeleteWithUndoButton.stories.tsx
@@ -58,32 +58,6 @@ const i18nProvider = polyglotI18nProvider(
     ]
 );
 
-const i18nProviderDefault = polyglotI18nProvider(
-    locale =>
-        locale === 'fr'
-            ? {
-                  ...frenchMessages,
-                  resources: {
-                      books: {
-                          name: 'Livre |||| Livres',
-                          fields: {
-                              id: 'Id',
-                              title: 'Titre',
-                              author: 'Auteur',
-                              year: 'Année',
-                          },
-                      },
-                  },
-              }
-            : englishMessages,
-    // Default locale
-    'en',
-    [
-        { locale: 'en', name: 'English' },
-        { locale: 'fr', name: 'Français' },
-    ]
-);
-
 const dataProvider = fakeRestDataProvider({
     books: [
         {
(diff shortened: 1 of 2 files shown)


packages/ra-ui-materialui/src/field/DateField.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+export default { title: 'ra-ui-materialui/fields/DateField' };
+
+export const Basic = ({
+    value,
+    theme,
+    ...props
+}: Partial<DateFieldProps> & { value?: Date | string; theme?: string }) => {
+    return (
+        <AdminContext
+            theme={theme === 'light' ? defaultLightTheme : defaultDarkTheme}
```
djhi, line 18: This does not show that we can actually target this component in the theme. Instead of allowing to change the theme mode, I suggest you pass a custom theme based on the `defaultLightTheme` with customization of the `DateField` styles and/or props. Same for all other components

djhi (changes requested):
Great job otherwise!

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.stories.tsx
```diff
@@ -0,0 +1,235 @@
...
+              },
+    // Default locale
+    'en',
+    [
+        { locale: 'en', name: 'English' },
+        { locale: 'fr', name: 'Français' },
+    ]
+);
+
+const i18nProviderDefault = polyglotI18nProvider(
```
djhi: Warning here, wrong copy/paste?

packages/ra-ui-materialui/src/field/ReferenceOneField.tsx
```diff
@@ -17,6 +17,12 @@ import {
 
 import { FieldProps } from './types';
 import { ReferenceFieldView } from './ReferenceField';
+import { sanitizeFieldRestProps } from './sanitizeFieldRestProps';
+import {
+    ComponentsOverrides,
```
djhi: lint warning here and below, wrong copy/paste?

PR #10776, opened 2025-06-10
Fix(show/edit): Avoid a redirect loop when useGetOne returns an error in shadcn-admin-kit

jonathanarnault:
## Problem

When a resource is not found in [`shadcn-admin-kit`](https://github.com/marmelab/shadcn-admin-kit), there is an infinite loop caused by a race condition between the router navigation and the query refresh

## Solution

Flush sync the navigation in `useEditController` and `useShowController` when the record fails to be retrieved.

## How To Test

Perform the test in the `shadcn-admin-kit` with the updated code

## Additional Checks

### DIFF
--- a/packages/ra-core/src/controller/edit/useEditController.ts
+++ b/packages/ra-core/src/controller/edit/useEditController.ts
@@ -3,7 +3,11 @@ import { useParams } from 'react-router-dom';
 
 import { useAuthenticated, useRequireAccess } from '../../auth';
 import { RaRecord, MutationMode, TransformData } from '../../types';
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/controller/edit/useEditController.ts
```diff
@@ -118,10 +123,19 @@ export const useEditController = <
                 (!isPendingAuthenticated && !isPendingCanAccess) ||
                 disableAuthentication,
             onError: () => {
+                console.warn('useEditController: record not found');
```
djhi: Shouldn't we remove this console?

PR #10782, opened 2025-06-12
Fix limitation preventing to declare create and show routes from the list view

slax57:
## Problem

If a resource declares both a list view and an edit view

```tsx
<Resource name="posts" list={PostList} edit={PostEdit} />
```

Then it's impossible to declare other routes (such as the create or show routes) from within the list view, because they overlap with the route declared for the edit view.

This prevents notably from being able to use a CreateDialog with a regular Edit view.
Hence this note in the [EE doc](https://react-admin-ee.marmelab.com/documentation/ra-form-layout#basic-usage-based-on-routing):

> **Note**: You can't use the `<CreateDialog>` and have a standard `<Edit>` specified on your `<Resource>`, because the `<Routes>` declarations would conflict. If you need this, use the `<CreateInDialogButton>` instead.

## Solution

Do not register the edit route if the location matches the create route or the show route.

## How To Test

- unit tests
- stories
  - http://localhost:9010/?path=/story/ra-core-core-resource--basic
  - http://localhost:9010/?path=/story/ra-core-core-resource--only-list
  - http://localhost:9010/?path=/story/ra-core-core-resource--with-all-dialogs
  - http://localhost:9010/?path=/story/ra-core-core-resource--with-create-dialog
  - http://localhost:9010/?path=/story/ra-core-core-resource--with-show-dialog
- Also http://localhost:9010/?path=/story/react-admin-resource--nested to make sure nested routes still work too
- And basepath:
  - http://localhost:9010/?path=/story/react-admin-admin--inside-router
  - http://localhost:9010/?path=/story/react-admin-admin--sub-path

## Additional Checks

### DIFF
--- a/docs/Resource.md
+++ b/docs/Resource.md
@@ -359,6 +359,7 @@ In order to display a list of songs for the selected artist, `<SongList>` should
 import { List, Datagrid, TextField, useRecordContext } from 'react-admin';
 import { useParams } from 'react-router-dom';
 import { Button } from '@mui/material';
+import EditIcon from '@mui/icons-material/Edit';
 
 export const SongList = () => {
     const { id } = useParams();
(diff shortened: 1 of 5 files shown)


djhi (approved):
Awesome stories, great job!

PR #10786, opened 2025-06-16
[Doc] Add a migration section in `<DataTable>` documentation that shows the codemod

djhi:
## Problem

We have a codemod to help people migrate from `<DataGrid>` to `<DataTable>`. However, it's not mentioned anywhere...
## Solution

Add a section  in `<DataTable>` documentation that shows the codemod


fblinks:
I guess that should be `replace-Datagrid-DataTable.ts ` instead of `replace-Admin-history.ts`

slax57 (approved):

PR #10791, opened 2025-06-19
Remove theme props override from view components

Madeorsk:
## Problem

useThemeProps was called 2 times in details / lists components

## Solution

Remove useThemeProps where required.

## How To Test

View List / InfiniteList / Show / Edit / Create "Themed" story.

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
+++ b/packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
@@ -452,6 +452,18 @@ export const Themed = () => (
                         perPage: 5,
                     },
                 },
+                RaList: {
+                    styleOverrides: {
+                        root: {
+                            background: 'pink',
+
+                            ['& .MuiListItemText-primary']: {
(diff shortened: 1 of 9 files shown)


packages/ra-ui-materialui/src/list/InfiniteList.stories.tsx
```diff
@@ -452,6 +452,18 @@ export const Themed = () => (
                         perPage: 5,
                     },
                 },
+                RaList: {
```
djhi, line 455: Does it work with `RaInfiniteList`?

Madeorsk: No, as RaInfiniteList uses ListView for rendering, which uses RaList as prefix for styling.

djhi: Can't you target it with `RaInfiniteList .RaList`?

djhi (approved):

PR #10796, opened 2025-06-20
[Doc] Update `<CreateDialog>` regarding limitation with old versions of react-admin

djhi:
## Problem

We can now have both `<CreateDialog>` and a standard `<Edit>` in version greater that `5.9.0`

## Solution

Update the documentation

### DIFF
--- a/docs/CreateDialog.md
+++ b/docs/CreateDialog.md
@@ -66,7 +66,7 @@ In the related `<Resource>`, you don't need to declare a `create` component as t
 <Resource name="customers" list={CustomerList} />
 ```
 
-**Note**: If you use a react-admin version older than `5.8.5`, you can't have both a `<CreateDialog>` and a standard `<Edit>` specified on your `<Resource>`, because the `<Routes>` declarations would conflict. If you need this, use the `<CreateInDialogButton>` instead.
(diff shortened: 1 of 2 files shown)


docs/CreateDialog.md
```diff
@@ -66,7 +66,7 @@ In the related `<Resource>`, you don't need to declare a `create` component as t
 <Resource name="customers" list={CustomerList} />
 ```
 
-**Note**: You can't use the `<CreateDialog>` and have a standard `<Edit>` specified on your `<Resource>`, because the `<Routes>` declarations would conflict. If you need this, use the [`<CreateInDialogButton>`](./CreateInDialogButton.md) instead.
+**Note**: If you use a react-admin version older than `5.8.5`, you can't have both a `<CreateDialog>` and a standard `<Edit>` specified on your `<Resource>`, because the `<Routes>` declarations would conflict. If you need this, use the `<CreateInDialogButton>` instead.
```
djhi:
```suggestion
**Note**: If you use a react-admin version older than `5.9.0`, you can't have both a `<CreateDialog>` and a standard `<Edit>` specified on your `<Resource>`, because the `<Routes>` declarations would conflict. If you need this, use the `<CreateInDialogButton>` instead.
```

Madeorsk (approved):

slax57:
DNM since #10782 had to be reverted (#10797)

PR #10808, opened 2025-06-27
Add ReferenceManyCountBase

fzaninotto:
The current ReferenceManyCount component is tied to MUI. To reuse the same (simple) logic in other UI kits, I moved that logic to `ra-core`.

```jsx
<ReferenceManyCountBase reference="comments" target="post_id" />
```

### DIFF
--- a/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
+++ b/packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
@@ -13,6 +13,7 @@ import usePaginationState from '../usePaginationState';
 import { useRecordSelection } from '../list/useRecordSelection';
 import useSortState from '../useSortState';
 import { useResourceContext } from '../../core';
+import { useRecordContext } from '../record';
 
 /**
  * Fetch reference records, and return them when available
@@ -55,7 +56,6 @@ export const useReferenceManyFieldController = <
     const {
         debounce = 500,
         reference,
-        record,
         target,
         filter = defaultFilter,
         source = 'id',
@@ -68,6 +68,7 @@ export const useReferenceManyFieldController = <
         >,
     } = props;
     const notify = useNotify();
+    const record = useRecordContext(props);
     const resource = useResourceContext(props);
     const dataProvider = useDataProvider();
     const queryClient = useQueryClient();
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/controller/field/useReferenceManyFieldController.ts
```diff
@@ -68,6 +68,7 @@ export const useReferenceManyFieldController = <
         >,
     } = props;
     const notify = useNotify();
+    const record = useRecordContext(props);
```
fzaninotto, line 71: This was done by callers of the hook. This logic should better be in ra-core.

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
```diff
@@ -77,7 +78,6 @@ export const ReferenceManyField = <
         target,
         queryOptions,
     } = props;
-    const record = useRecordContext(props);
```
fzaninotto, line 80: no longer necessary since the controller hook does it.

packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
```diff
@@ -51,39 +51,20 @@ export const ReferenceManyCount = <RecordType extends RaRecord = RaRecord>(
...
-        oneSecondHasPassed ? (
-            <CircularProgress size={14} />
-        ) : (
-            ''
-        )
-    ) : error ? (
-        <ErrorIcon color="error" fontSize="small" titleAccess="error" />
-    ) : (
-        total
+    const body = (
```
djhi, line 59: Just put it inside the main JSX directly? This `body` const is useless now IMO

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
```diff
@@ -69,6 +69,7 @@ export const ReferenceManyField = <
         page = 1,
         pagination = null,
         perPage = 25,
+        record,
```
djhi, line 71: lint warning now

packages/ra-ui-materialui/src/field/ReferenceManyCount.tsx
fzaninotto: it's used twice

packages/ra-ui-materialui/src/field/ReferenceManyField.tsx
fzaninotto: fixed

djhi (approved):

PR #10812, opened 2025-06-29
Fix Datagrid empty throws error when used in standalone mode

### LINKED ISSUE #10811: Enhance Datagrid to Autodetect Resource from useList
I am building custom list components that utilize `useList` and `ListContextProvider` to manage data and state outside of the standard `<Admin>` and `<Resource>` component hierarchy. In this setup, I explicitly define the resource name when calling useList (e.g., ```useList({ data: [], resource: 'Custom Posts' })```).

When the custom list data is empty, the `Datagrid` component, which is wrapped by `ListContextProvider`, attempts to render its empty state. This leads to the error: ```"<ListNoResults> must be used inside a <List> component"```. This error occurs because Datagrid's internal components (like the default Empty component which likely uses ListNoResults) rely on `useResourceContext` to infer the resource name. However, in this direct `ListContextProvider` setup, the `ResourceContext` is null, despite the resource being clearly defined within the list object provided by `useList`.

This forces me to explicitly Wrapping the `ListContextProvider` and Datagrid with a ResourceContextProvider (e.g., ```<ResourceContextProvider value="Custom Posts">``` to define resource.

I would like any component that is a child of `ListContextProvider` (such as Datagrid, SimpleList, FilterForm, etc.) to have direct access to and automatically infer the resource name defined in the list object returned by useList.

This enhancement would streamline custom list implementations, making `useList` and `ListContextProvider` more powerful and less error-prone for developers working outside of the strict <Resource> tree.

**Working Solution** explicitly Wrapping the `ListContextProvider` and `Datagrid` with a `ResourceContextProvider` (e.g., ``<ResourceContextProvider value="Custom Posts">``` to define resource.

**Additional context**
Here's the minimal example code demonstrating the issue:
```
import {
  Datagrid,
  ListContextProvider,
  NumberField,
  TextField,
  useList,
} from 'react-admin';

const CustomRoute = () => {
  const customResourceName = 'Custom Posts';
  const list = useList({
    data: [], // Empty data to trigger the 'no results' scenario
    resource: customResourceName, // Resource is defined here
  });

  return (
    <ListContextProvider value={list}>
      {/*
        This Datagrid currently throws "<ListNoResults> must be used inside a <List> component"
        because its internal useResourceContext cannot find the resource,
        even though it's defined in useList.
      */}
      <Datagrid>
        <NumberField source="id" />
        <TextField source="name" />
      </Datagrid>
    </ListContextProvider>
  );
};

export default CustomRoute;
```

fzaninotto:
Closes #10811

### DIFF
--- a/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
+++ b/packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
@@ -9,6 +9,7 @@ import {
     FC,
     ComponentType,
     ReactElement,
+    ReactNode,
     useMemo,
 } from 'react';
 import {
@@ -139,7 +140,7 @@ export const Datagrid: React.ForwardRefExoticComponent<
         header = DatagridHeader,
         children,
         className,
-        empty = DefaultEmpty,
+        empty,
         expand,
         bulkActionsToolbar,
         bulkActionButtons = canDelete ? defaultBulkActionButtons : false,
@@ -235,11 +236,7 @@ export const Datagrid: React.ForwardRefExoticComponent<
      * the Datagrid displays the empty component.
      */
     if (data == null || data.length === 0 || total === 0) {
-        if (empty) {
-            return empty;
-        }
-
-        return null;
+        return empty ?? <ListNoResults resource={resource} />;
     }
 
     /**
@@ -464,7 +461,7 @@ export interface DatagridProps<RecordType extends RaRecord = any>
      *     </List>
      * );
      */
-    empty?: ReactElement;
+    empty?: ReactNode;
 
     /**
      * A function that returns whether the row for a record is expandable.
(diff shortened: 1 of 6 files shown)


packages/ra-ui-materialui/src/list/datagrid/Datagrid.tsx
```diff
@@ -235,11 +236,7 @@ export const Datagrid: React.ForwardRefExoticComponent<
      * the Datagrid displays the empty component.
      */
     if (data == null || data.length === 0 || total === 0) {
-        if (empty) {
-            return empty;
-        }
-
-        return null;
+        return empty ?? <ListNoResults resource={resource} />;
```
djhi: If `empty` is `null` or `false`, it shouldn't return a default component. Only when it is `undefined`

packages/ra-ui-materialui/src/list/datatable/DataTable.tsx
```diff
@@ -183,7 +182,7 @@ export const DataTable = React.forwardRef(function DataTable<
             {...props}
             hasBulkActions={hasBulkActions}
             loading={loading}
-            empty={empty}
+            empty={empty ?? <ListNoResults resource={resourceFromContext} />}
```
djhi: Same

Issue #3677, opened 2019-09-12
[RadioButtonGroupInput] Hope Support disable some option


scq000:
When I use RadioButtonGroupInput component, I wish I can disable some choices by providing a `disableValue` field which name is `disabled`.

Example:
```
 const choices = [
     { id: 123, name: '12' },
     { id: 456, name: '18' },
     { id: 976, name: '30', disabled: true },
  ];
<RadioButtonGroupInput source="age" choices={choices} />
```

PR #10821, opened 2025-07-02
Add disable support for `RadioButtonGroupInput` and `CheckboxGroupInput` choices

WiXSL:
Implements #3677

## Additional Checks

### DIFF
--- a/docs/RadioButtonGroupInput.md
+++ b/docs/RadioButtonGroupInput.md
@@ -62,6 +62,7 @@ The form value for the source must be the selected value, e.g.
 | `optionValue`     | Optional | `string`                   | `id`    | Field name of record containing the value to use as input value                                                                        |
 | `row`             | Optional | `boolean`                  | `true`  | Display options in a compact row.                                                                                                      |
 | `translateChoice` | Optional | `boolean`                  | `true`  | Whether the choices should be translated                                                                                               |
+| `disableValue`    | Optional | `string`                   | `disabled` | The custom field name used in `choices` to disable some choices                                                                     |
 
 `<RadioButtonGroupInput>` also accepts the [common input props](./Inputs.md#common-input-props).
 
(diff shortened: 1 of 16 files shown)


djhi:
Thanks! Would you mind adding some tests?

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.spec.tsx
```diff
@@ -541,6 +541,33 @@ describe('<RadioButtonGroupInput />', () => {
...
+            </AdminContext>
+        );
+        fireEvent.mouseDown(screen.getByLabelText('Mastercard'));
+
+        expect(screen.getByText('VISA').getAttribute('class')).not.toContain(
+            'Mui-disabled'
+        );
+        expect(screen.getByText('Mastercard').getAttribute('class')).toContain(
+            'Mui-disabled'
+        );
```
djhi: Unfortunately, this is brittle. Can you check wheter the actual radio input has the `disabled` attribute set instead?

docs/RadioButtonGroupInput.md
```diff
@@ -94,6 +94,17 @@ const choices = [
...
+
+```jsx
+const choices = [
+    { id: 'tech', name: 'Tech' },
+    { id: 'lifestyle', name: 'Lifestyle' },
+    { id: 'people', name: 'People', disabled: true },
+];
+<RadioButtonGroupInput source="category" choices={choices} />
+```
+
```
djhi, line 108: Can you add a dedicated section for this prop?

WiXSL: Technically is not a prop of the component, but rather of the choices.
I'd change it up as so either way, see if it's ok

djhi: Sorry I meant the `disableValue` prop

djhi: It looks like this component does not accept it but `useChoices` actually does.

WiXSL: I didn't add the disableValue prop, just make it possible to pass a disable prop to the choices.

WiXSL: Ok, I'll try to add the `disableValue` prop as well

packages/ra-ui-materialui/src/input/RadioButtonGroupInput.tsx
```diff
@@ -277,6 +296,7 @@ export type RadioButtonGroupInputProps = Omit<CommonInputProps, 'source'> &
     RadioGroupProps & {
         options?: RadioGroupProps;
         source?: string;
+        disableValue?: string;
```
djhi: I believe we forgot to add this prop to the `ChoicesProps`. Can you add it there instead?

packages/ra-ui-materialui/src/input/RadioButtonGroupInputItem.tsx
```diff
@@ -53,6 +57,7 @@ export interface RadioButtonGroupInputItemProps
         Pick<ChoicesProps, 'optionValue' | 'optionText' | 'translateChoice'> {
     choice: any;
     source: any;
+    disableValue?: string;
```
djhi: Then you could pick it at L57

djhi (changes requested):

djhi (approved):

djhi:
Thanks!

PR #10828, opened 2025-07-04
Improve memory footprint of `useCanAccess` when it targets a record

djhi:
## Problem

When wrapping many different records in `<CanAccess>`, its memory footpring is huge. 

## Solution

This is because we pass the whole record in the `useQuery` parameters. We can only pass the record identifier.

## How It Was Tested

Before the fix:
image

After the fix:
image

Even with this small use case (records are quite simple), that's 10MB saved.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/auth/CanAccess.stories.tsx
+++ b/packages/ra-core/src/auth/CanAccess.stories.tsx
@@ -2,9 +2,9 @@ import * as React from 'react';
 import { Location } from 'react-router';
 import { QueryClient } from '@tanstack/react-query';
 import { AuthProvider } from '../types';
-import { CoreAdminContext } from '../core';
+import { CoreAdminContext } from '../core/CoreAdminContext';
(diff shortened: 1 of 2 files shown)


packages/ra-core/src/auth/CanAccess.stories.tsx
```diff
@@ -109,3 +109,45 @@ export const NoAuthProvider = () => (
...
+    };
+
+    Array.from({ length: 100 }, (_, i) => {
+        post[`property${i + 1}`] = `Another very long property ${i + 1}`;
+    });
+
+    return post;
+});
+
+export const ManyCalls = ({
```
fzaninotto, line 128: It would be interesting to have the react-query debugger on this story

djhi: Done

Madeorsk (approved):

PR #10829, opened 2025-07-07
Improve commerce demo bundle

djhi:
## Problem

The ecommerce demo bundle is quite large.

## Solution

- [x] Use echarts properly (we saved a bit more than 1Mb)
- [x] Lazy load the chart on dashboard

### DIFF
--- a/examples/demo/vite.config.ts
+++ b/examples/demo/vite.config.ts
@@ -47,7 +47,10 @@ export default defineConfig(async ({ mode }) => {
             }),
         ],
         define: {
-            'process.env': process.env,
+            'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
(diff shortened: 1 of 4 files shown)


examples/demo/vite.config.ts
```diff
@@ -47,7 +47,10 @@ export default defineConfig(async ({ mode }) => {
             }),
         ],
         define: {
-            'process.env': process.env,
+            'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
```
Madeorsk, line 50: I don't understand why you need `JSON.stringify` (same in `crm`).

djhi: It's necessary as Vite will put it as is otherwise and you'll end with an error like:
> Uncaught ReferenceError: development is not defined

PR #10830, opened 2025-07-07
Introduce `<ListIterator>`

djhi:
## Problem

Rendering a custom list is cumbersome. One has to create an iterator component that reads the list context.

The `<SingleFieldList>` is a good tradeoff, as one can just use a component that reads a `RecordContext` and it will be displayed several times.

But it’s tied to MUI, so we can’t use it in other UI kits.

## Solution

Add a `<ListIterator>` component, that simply iterates over data, creates a `RecordContext`, and renders its child n times. It shouldn’t render any wrapper, to let users choose the wrapper of their choice (e.g. `<ul>`)

It should render either `children` or a `render` prop

## TODO

- [x] Implementation
- [x] Tests & Stories
- [x] Update `<SingleFieldList>` and `<SimpleList>`
- [x] Documentation
 
## How To Test

- https://react-admin-storybook-git-list-iterator-marmelab.vercel.app/?path=/story/ra-core-controller-list-listiterator--using-render
- https://react-admin-storybook-git-list-iterator-marmelab.vercel.app/?path=/story/ra-core-controller-list-listiterator--using-children
- unit tests

## Additional Checks

### DIFF
--- a/docs/ListIterator.md
+++ b/docs/ListIterator.md
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: "ListIterator"
-storybook_path: react-admin-admin--basic
+storybook_path: ra-core-controller-list-listiterator--using-render
 ---
 
 # `<ListIterator>`
@@ -43,9 +43,8 @@ Here are all the props you can set on the `<AccordionForm>` component:
 | `data`      | Optional | `RaRecord[]`                       | -       | The records. Defaults to the `data` from the `ListContext`                                           |
 | `empty`     | Optional | `ReactNode`                        | `null`  | The content to display when there is no data                                                         |
 | `isPending` | Optional | `boolean`                          | -       | A boolean indicating whether the data is pending. Defaults to the `isPending` from the `ListContext` |
-| `pending`   | Optional | `ReactNode`                        | `null`  | The content to display while the data is loading                                                     |
+| `loading`   | Optional | `ReactNode`                        | `null`  | The content to display while the data is loading                                                     |
 | `render`    | Optional | `(record: RaRecord) => ReactNode`  | -       | A function that returns the content to render for each record                                        |
-| `resource`  | Optional | `string`                           | -       | The resource. Defaults to the `ResourceContext`                                                      |
 | `total`     | Optional | `number`                           | -       | The total number of records. Defaults to the `total` from the `ListContext`                          |
 
 Additional props are passed to `react-hook-form`'s [`useForm` hook](https://react-hook-form.com/docs/useform).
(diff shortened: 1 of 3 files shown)


docs/ListIterator.md
```diff
@@ -0,0 +1,250 @@
+---
+layout: default
+title: "ListIterator"
+storybook_path: react-admin-admin--basic
```
slax57: this doesn't look like the correct storybook path

docs/ListIterator.md
```diff
@@ -0,0 +1,250 @@
...
+| ----------- | -------- | ---------------------------------- | ------- | ---------------------------------------------------------------------------------------------------- |
+| `children`  | Optional | `ReactNode`                        | -       | The content to render for each record                                                                |
+| `data`      | Optional | `RaRecord[]`                       | -       | The records. Defaults to the `data` from the `ListContext`                                           |
+| `empty`     | Optional | `ReactNode`                        | `null`  | The content to display when there is no data                                                         |
+| `isPending` | Optional | `boolean`                          | -       | A boolean indicating whether the data is pending. Defaults to the `isPending` from the `ListContext` |
+| `pending`   | Optional | `ReactNode`                        | `null`  | The content to display while the data is loading                                                     |
```
slax57: Not sure about this naming. To my knowledge, this prop is usually named `loading` in other RA components. Shouldn't we also use `loading` here for consistency?

docs/ListIterator.md
```diff
@@ -0,0 +1,250 @@
...
+    });
+
+    return (
+        <OrderedList>
+            <ListIterator
+                isPending={isPending}
+                render={record => <ListItem>{record.title} - {record.views}</ListItem>}
+            />
+        </OrderedList>
+    );
```
slax57: You probably intended to also use `data` in this example

docs/ListIterator.md
```diff
@@ -0,0 +1,250 @@
...
+```jsx
+import { ListBase, ListIterator } from 'react-admin';
+import { OrderedList, ListItem } from 'my-favorite-ui-lib';
+
+const DashboardMostVisitedPosts = () => (
+    <ListBase resource="posts" sort={{ field: 'views', order: 'DESC' }} page={1} perPage={20}>
+        <OrderedList>
+            <ListIterator
+                render={record => <ListItem>{record.title} - {record.views}</ListItem>}
+            />
```
slax57: you forgot to use `resource` in this example.

Also, it's not obvious why `<ListIterator>` would require a `resource` to operate. Provided you have a `data` array, it should be all you need to iterate over it, correct?

EDIT: as I thought, it's actually never used. You can remove this prop altogether I believe.

PR #10833, opened 2025-07-08
Introduce `<ReferenceArrayInputBase>`

djhi:
## Problem

The `<ReferenceArrayInput>` from `ra-ui-materialui` actually don't use MUI except for its default children.

## Solution

Copy and adapt `<ReferenceArrayInput>` in `ra-core`.

## How To Test

- https://react-admin-storybook-3f2nu3m5l-marmelab.vercel.app/?path=/story/ra-core-controller-referencearrayinputbase--basic
- https://react-admin-storybook-3f2nu3m5l-marmelab.vercel.app/?path=/story/ra-core-controller-referencearrayinputbase--with-render
- https://react-admin-storybook-3f2nu3m5l-marmelab.vercel.app/?path=/story/ra-core-controller-referencearrayinputbase--with-error 
- tests

## Additional Checks

### DIFF
--- a/docs/ReferenceArrayInput.md
+++ b/docs/ReferenceArrayInput.md
@@ -114,7 +114,7 @@ See the [`children`](#children) section for more details.
 
 ## `children`
 
-By default, `<ReferenceInput>` renders an [`<AutocompleteArrayInput>`](./AutocompleteArrayInput.md) to let end users select the reference record.
+By default, `<ReferenceArrayInput>` renders an [`<AutocompleteArrayInput>`](./AutocompleteArrayInput.md) to let end users select the reference record.
 
 You can pass a child component to customize the way the reference selector is displayed.
 


packages/ra-core/src/controller/input/ReferenceArrayInputBase.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+        jest.spyOn(console, 'error').mockImplementation(() => {});
+        render(<WithError />);
+        await waitFor(() => {
+            expect(screen.queryByText('Error: fetch error')).not.toBeNull();
+        });
+    });
+    it('should pass the correct resource down to child component', async () => {
+        render(<Basic />);
+        await screen.findByText('Selected tags: 1, 3');
+    });
```
slax57: This test seems redundant with `should apply default values`, and doesn't seem to check the resource.
You should either fix it or remove it IMO.

packages/ra-core/src/controller/input/ReferenceArrayInputBase.tsx
```diff
@@ -0,0 +1,117 @@
...
+ * choices available through the ChoicesContext.
+ *
+ * Use it with a selector component as child, like `<SelectArrayInput>`
+ * or <CheckboxGroupInput>.
+ *
+ * @example
+ * export const PostEdit = () => (
+ *     <Edit>
+ *         <SimpleForm>
+ *             <ReferenceArrayInput source="tag_ids" reference="tags">
```
slax57: ReferenceArrayInput -> ReferenceArrayInputBase
(same applies to whole jsdoc)

packages/ra-core/src/controller/input/ReferenceArrayInputBase.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const { children, filter = defaultFilter, reference, render, sort } = props;
+    if (children && React.Children.count(children) !== 1) {
+        throw new Error(
+            '<ReferenceArrayInputBase> only accepts a single child (like <Datagrid>)'
+        );
+    }
+
+    if (!render && !children) {
+        throw new Error(
+            "<EditBase> requires either a 'render' prop or 'children' prop"
```
slax57:
```suggestion
            "<ReferenceArrayInputBase> requires either a 'render' prop or 'children' prop"
```

packages/ra-ui-materialui/src/input/ReferenceArrayInput.tsx
```diff
@@ -77,40 +70,22 @@ import { AutocompleteArrayInput } from './AutocompleteArrayInput';
...
 
 const defaultChildren = <AutocompleteArrayInput />;
-const defaultFilter = {};
 
 export interface ReferenceArrayInputProps
-    extends InputProps,
-        UseReferenceArrayInputParams {
-    children?: ReactElement;
+    extends Omit<ReferenceArrayInputBaseProps, 'children'> {
+    children?: React.ReactNode;
```
slax57: why not simply

```suggestion
    extends ReferenceArrayInputBaseProps {
```

?

slax57 (changes requested):
Why not also add a doc page, just like you dit for ListIterator?

packages/ra-core/src/controller/input/ReferenceArrayInputBase.spec.tsx
djhi: It is a bit redundant but it checks the resource as its in the text we look for: _Seleted **tags**_

packages/ra-core/src/controller/input/ReferenceArrayInputBase.tsx
djhi: Will do

packages/ra-core/src/controller/input/ReferenceArrayInputBase.spec.tsx
slax57: indeed, my bad!

packages/ra-core/src/controller/input/ReferenceArrayInputBase.tsx
```diff
@@ -0,0 +1,117 @@
...
+ * The enclosed component may filter results. ReferenceArrayInputBase create a ChoicesContext which provides
+ * a `setFilters` function. You can call this function to filter the results.
+ */
+export const ReferenceArrayInputBase = <RecordType extends RaRecord = any>(
+    props: ReferenceArrayInputBaseProps<RecordType>
+) => {
+    const { children, filter = defaultFilter, reference, render, sort } = props;
+    if (children && React.Children.count(children) !== 1) {
+        throw new Error(
+            '<ReferenceArrayInputBase> only accepts a single child (like <Datagrid>)'
```
slax57: Btw I'm just realizing this is wrong, `Datagrid` cant' be a child of `<ReferenceArrayInput>`. We should fix this error in both `ReferenceArrayInputBase` and `ReferenceArrayInput`.

PR #10836, opened 2025-07-11
Add timeout prop to loading components

### LINKED ISSUE #10835: Make Datagrid loading timeout configurable
**What you were expecting:**

I expect the Datagrid component skeleton to be displayed while the data is being retrieved, or for it to receive a loading property, so that external manipulation can be performed, since skeletons are not displayed.

**What happened instead:**

The Skeleton of the built-in Datagrid component isn't displaying; this is due to internal delays in the components added on purpose.

**Steps to reproduce:**

1. Use a list with the Datagrid.
2. Check the rendering of the default skeleton while loading the data.
3. Check the components that have a timeout.

**Related code:**

* Preferably, a sandbox forked from 
  - https://codesandbox.io/p/github/marmelab/react-admin-sandbox/main (v5)

  - https://stackblitz.com/github/marmelab/react-admin/tree/master/examples/simple (v5)
  - https://stackblitz.com/github/marmelab/react-admin/tree/4.x/examples/simple (v4)

```
```

**Other information:**

Additionally, you have the DatagridLoading component, which also uses a timer from its input.

**Environment**

* React-admin version: 6.9.1
* React version: 19.1 
* Browser: Chrome

WiXSL:
## Problem

Some loading components have a one second hard-coded wait before rendering

Fixes #10835

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/SimpleList/SimpleListLoading.tsx
+++ b/packages/ra-ui-materialui/src/list/SimpleList/SimpleListLoading.tsx
@@ -29,10 +29,11 @@ export const SimpleListLoading = (inProps: SimpleListLoadingProps) => {
         hasSecondaryText,
         hasTertiaryText,
         nbFakeLines = 5,
+        timeout = 1000,
         ...rest
     } = props;
 
-    const oneSecondHasPassed = useTimeout(1000);
+    const oneSecondHasPassed = useTimeout(timeout);
 
     return oneSecondHasPassed ? (
         <StyledList className={className} {...rest}>
@@ -109,6 +110,7 @@ export interface SimpleListLoadingProps extends ListProps {
     hasSecondaryText?: boolean;
     hasTertiaryText?: boolean;
     nbFakeLines?: number;
+    timeout?: number;
 }
 
 declare module '@mui/material/styles' {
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks! Would you mind updating the doc to allow users to discover this setting?

WiXSL:
> Thanks! Would you mind updating the doc to allow users to discover this setting?

Sure, but where exactly do I add the docs? I can't find a good place.

fzaninotto:
Good point! Which leads me to the question: How can anyone use these new props? DataTable and SimpleList don't expose their `loading` component.

WiXSL:
> Good point! Which leads me to the question: How can anyone use these new props? DataTable and SimpleList don't expose their `loading` component.

Mmm, you are right!
It is true that they don't expose these components, but they get exported.
I already did this with the `Loading` component to be able to re-use it.

Maybe the best solution is to add a `loading` prop to `SimpleList`, `DataTable`, and `Datagrid` components to be able to expose the loading component (in another PR) and merge this one as is.
What do you think?

fzaninotto:
I'd prefer that you add the loading prop in this PR. Otherwise, it adds no value to react-admin.

WiXSL:
> I'd prefer that you add the loading prop in this PR. Otherwise, it adds no value to react-admin.

Ok, I'll try to do that

fzaninotto:
Hi, any updates?

WiXSL:
Sorry, I've decided not to continue with this PR as not needed by me anymore

PR #10838, opened 2025-07-16
Update react-query depency to require at least v5.83

djhi:
## Problem

We received some requests and questions for enabling list components data fetching conditionally.
It's actually easier to do with react-query versions more recent than the one we install by default.

## Solution

- [x] Upgrade react-query
- [x] Document how to enabling data fetching conditionally
- [x] Add a story to show this use case

## How To Test

- https://react-admin-storybook-rarckfjb1-marmelab.vercel.app/?path=/story/ra-ui-materialui-list-list--conditional-data-fetching

## Additional Checks

### DIFF
--- a/docs/List.md
+++ b/docs/List.md
@@ -1192,8 +1192,7 @@ export const PostList = () => (
     >
         <WithListContext
             render={context =>
-                context.filterValues.q == null ||
-                context.filterValues.q === '' ? (
+                context.filterValues.q?.length > 2 ? (
                     <CardContentInner>
                         Type a search term to fetch data
                     </CardContentInner>
--- a/docs/List.md
+++ b/docs/List.md
@@ -1174,6 +1174,26 @@ const ProductList = () => (
 )
 ```
 
+## Enabling Data Fetching Conditionally
+
+You might want to allow data to be fetched only when at least some filters have been set. You can leverage TanStack react-query `enabled` option for that. It accepts a function that receives the query as its only parameter. As react-admin always format the `queryKey` as `[ResourceName, DataProviderMethod, DataProviderParams]`, you can check that there is at least a filter in this function:
+
+```tsx
+export const PostList = () => (
+    <List
+        filters={postFilter}
+        queryOptions={{
+            enabled: query => {
+                const listParams = query.queryKey[2] as GetListParams;
+                return listParams.filter.q?.length > 2;
+            }
+        }}
+    >
+        {/* DataGrid */}
+    </List>
+)
+```
+
 ## Accessing Extra Response Data
 
 If `dataProvider.getList()` returns additional metadata in the response under the `meta` key, you can access it in the list view using the `meta` property of the `ListContext`.
--- a/docs/List.md
+++ b/docs/List.md
@@ -1178,6 +1178,7 @@ const ProductList = () => (
 
 You might want to allow data to be fetched only when at least some filters have been set. You can leverage TanStack react-query `enabled` option for that. It accepts a function that receives the query as its only parameter. As react-admin always format the `queryKey` as `[ResourceName, DataProviderMethod, DataProviderParams]`, you can check that there is at least a filter in this function:
 
+{% raw %}
 ```tsx
 export const PostList = () => (
     <List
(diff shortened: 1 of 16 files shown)


docs/List.md
```diff
@@ -1174,6 +1174,26 @@ const ProductList = () => (
 )
 ```
 
+## Enabling Data Fetching Conditionally
```
djhi, line 1177: I'm not sure about where to put this documentation as it's actually `useGetList` that we configure.

fzaninotto: I think it's fine here 👍

packages/ra-core/package.json
```diff
@@ -57,8 +57,7 @@
         "react-router-dom": "^6.28.1 || ^7.1.1"
     },
     "dependencies": {
-        "@tanstack/react-query": "^5.21.7",
-        "clsx": "^2.1.1",
```
fzaninotto, line 61: 👍 good catch

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
```diff
@@ -248,7 +248,6 @@ export type UseInfiniteGetListOptions<
         GetInfiniteListResult<RecordType>,
         ErrorType,
         InfiniteData<GetInfiniteListResult<RecordType>>,
-        GetInfiniteListResult<RecordType>,
```
fzaninotto, line 251: does that mean that the type of `UseInfiniteQueryOptions` has changed in react-query? That would be a breaking change.

packages/ra-core/src/dataProvider/useGetList.ts
```diff
@@ -176,7 +176,7 @@ export const useGetList = <
                   }
                 : result,
         [result]
-    ) as UseQueryResult<RecordType[], Error> & {
+    ) as unknown as UseQueryResult<RecordType[], Error> & {
```
fzaninotto, line 179: Question: Why do we only need to do that when calling `dataProvider.getList()` and not for every other usage of `useQuery` (e.g. `useGetOne`) ?

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
djhi: No I believe we used it incorrectly before but it wasn't detected until this update

packages/ra-core/src/dataProvider/useGetList.ts
djhi: Because `useGetOne` returns the same format as react query and we made the decision to do things differently in all hooks that also return a total. Instead of `data.total` we wanted to have `total` directly on the result.

djhi: For the record, I'd love to change that in v6

packages/ra-core/src/dataProvider/useInfiniteGetList.ts
fzaninotto: I think it's due to https://github.com/TanStack/query/pull/8956. It doesn't seem to be a breaking change.

packages/ra-ui-materialui/src/list/List.stories.tsx
```diff
@@ -229,6 +230,48 @@ export const Filters = () => (
...
+    if (context.filterValues.q == null || context.filterValues.q === '') {
+        return (
+            <CardContentInner>
+                Type a search term to fetch data
+            </CardContentInner>
+        );
+    }
+
+    return <BookList />;
+};
```
slax57: This story is working, but since you are rendering a different child when the filter condition is not true, we can't really tell if the dataProvider is called or not. It would help to enable the fakerest debug logs.

slax57 (approved):

docs/List.md
```diff
@@ -1190,12 +1190,27 @@ export const PostList = () => (
             }
         }}
     >
-        {/* DataGrid */}
+        <WithListContext
+            render={context =>
+                context.filterValues.q == null ||
+                context.filterValues.q === '' ? (
```
slax57: Too bad you copied the wrong condition: in the docs the condition was `q.length > 2` and not `q !== ''` :sweat_smile:

fzaninotto (approved):

PR #10843, opened 2025-07-21
[Doc] Bootstrap headless documentation

djhi:
Bootstrap headless documentation with Astro Starlight

## How To Test

- `make doc-headless`
- Click the _Get started_ button
- Resize your window to be very small. In a new tab, enter `http://localhost:4321/ra-core/reference/example/`: note the sidenav scroll is sync

## Additional Checks


docs_headless/src/content/docs/reference/example.md
```diff
@@ -0,0 +1,12 @@
...
+description: A reference page in my new Starlight docs site.
+---
+
+Reference pages are ideal for outlining how things work in terse and clear terms.
+Less concerned with telling a story or addressing a specific use case, they should give a comprehensive outline of what you're documenting.
+
+## Tests
+
+- [General Concepts](/guides/architecture)
+- [Data Fetching](/guides/datafetchingguide)
```
slax57: It's too bad we can't use links to the .md files, because it means the md docs won't be browsable from Github directly. 
Can you try something like https://github.com/vernak2539/astro-rehype-relative-markdown-links?

docs_headless/src/content/docs/guides/Architecture.md
```diff
@@ -0,0 +1,304 @@
...
+
+![SPA lifecycle](../../../assets/SPA-lifecycle.png)
+
+The SPA architecture ensures that react-admin apps are [exceptionally fast](./Features.md#fast), easy to host, and compatible with existing APIs without requiring a dedicated backend. 
+
+To achieve this, react-admin utilizes an internal router, powered by `react-router`, to display the appropriate screen when the user clicks on a link. Developers can define routes using the [`<Resource>`](./Resource.md) component for CRUD routes and the [`<CustomRoutes>`](./CustomRoutes.md) component for other routes.
+
+For example, the following react-admin application:
+
+```jsx
```
slax57, line 19: I tried using syntax like `jsx{5}` but unfortunately it does not work. Do you think we could add support for it?

slax57 (commented):
Otherwise it indeed allows to get a really nice doc website with very few code, I love it!

docs_headless/src/content/docs/reference/example.md
slax57: Implemented! Works like a charm!

docs_headless/src/content/docs/guides/Architecture.md
slax57: Actually it's already possible! Just need to use `jsx {5}` instead of `jsx{5}`!

PR #10844, opened 2025-07-21
[Demo] Migrate from `fetch-mock` to `msw`

erwanMarmelab:
## Problem

`fetch-mock` brings us back some vulnerabilities.

## Solution

Change it for MSW

## To Do

- [x] demo
- [x] crm
- [x] simple

## Additional Checks


examples/demo/src/fakeServer/graphql.ts
```diff
@@ -1,29 +1,12 @@
...
+import { setupWorker } from 'msw/browser';
+import { getMswHandler } from 'fakerest';
 import generateData from 'data-generator-retail';
-import fetchMock from 'fetch-mock';
 
-export default () => {
-    const data = generateData();
-    const restServer = JsonGraphqlServer({ data });
-    const handler = restServer.getHandler();
-    const handlerWithLogs = (url: string, opts: any) =>
```
erwanMarmelab, line 9: We don't need to keep it. If we **don't** pass `quiet: true` to `worker.start`, it will automatically displays us the right logs :+1:

djhi (changes requested):
Good work, except you probably didn't test the graphql demo which is broken

examples/simple/package.json
```diff
@@ -35,5 +36,11 @@
         "react-simple-animate": "^3.5.3",
         "typescript": "^5.1.3",
         "vite": "^6.2.6"
+    },
+    "msw": {
+        "workerDirectory": [
+            "assets",
```
Madeorsk: Why `assets`? `mockServiceWorker.js` is in `public` directory, not `assets`.

fzaninotto:
needs rebase

examples/crm/src/providers/fakerest/fakeServer.ts
```diff
@@ -0,0 +1,13 @@
...
+import generateData from './dataGenerator';
+
+const handler = getMswHandler({
+    baseUrl: 'http://localhost:4000',
+    data: generateData() as CollectionItem,
+    middlewares: [withDelay(300)],
+});
+export const worker = setupWorker(http.all(/http:\/\/localhost:4000/, handler));
+
+export default () => worker;
```
fzaninotto, line 13: Why do you need this export? I don't see it imported anywhere.

yarn.lock
```diff
@@ -7361,6 +7496,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"cli-width@npm:^4.1.0":
```
fzaninotto, line 7499: This one is duplicated (albeit in a different major version). Can we avoid it?

yarn.lock
```diff
@@ -7734,6 +7876,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"cookie@npm:^0.7.2":
```
fzaninotto, line 7879: also duplicated

yarn.lock
```diff
@@ -9862,6 +10003,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"fakerest@npm:4.1.3, fakerest@npm:^4.1.3":
```
fzaninotto: also duplicated

yarn.lock
```diff
@@ -13308,9 +13442,9 @@ __metadata:
   linkType: hard
 
 "linkifyjs@npm:^4.1.0":
-  version: 4.3.2
-  resolution: "linkifyjs@npm:4.3.2"
-  checksum: 1a85e6b368304a4417567fe5e38651681e3e82465590836942d1b4f3c834cc35532898eb1e2479f6337d9144b297d418eb708b6be8ed0b3dc3954a3588e07971
+  version: 4.1.1
```
fzaninotto: please revert this one, linkify < 4.3.2 has a security vulnerability

yarn.lock
```diff
@@ -18802,6 +18992,18 @@ __metadata:
   languageName: node
   linkType: hard
 
+"tough-cookie@npm:^4.1.4":
```
fzaninotto: duplicate package

examples/simple/package.json
```diff
@@ -14,6 +14,7 @@
         "@mui/material": "^5.16.12",
         "@tanstack/react-query": "^5.83.0",
         "@tanstack/react-query-devtools": "^5.83.0",
+        "fakerest": "^4.1.3",
         "jsonexport": "^3.2.0",
         "lodash": "~4.17.5",
         "ra-data-fakerest": "^5.10.0",
```
fzaninotto: no longer needed. Also, missing ra-data-simple-rest.

examples/crm/src/providers/fakerest/fakeServer.ts
djhi: It's imported from the projet root index: https://github.com/marmelab/react-admin/pull/10844/files#diff-90bd1ee22f91d95ef929650cdc5d138faa08887243bdbc5167f2671ed876196dR5

yarn.lock
djhi: We shouldn't deduplicate different major versions

yarn.lock
djhi: Same with an additional note: before `1.0.0`, all versions can have breaking changes

djhi:
> The e-commerce demo doesn't work (either with REST or GraphQL). I get this error in the SAP when running `make run-demo`:

I can't reproduce this. Can you try resetting your `node_modules`?

fzaninotto (approved):
The error was caused by Firefox being Firefox. Rebooting it fixed it.

PR #10848, opened 2025-07-21
Add filter input to `<ColumnsButton>` when there are many columns

slax57:
## Problem

When a `<DataTable>` has many columns, the `ColumnsButton` dropdown is hard to use.

## Solution

When there are more than 5 columns, show an input in the top of the dropdown to filter the column list.

## How To Test

- http://localhost:9010/?path=/story/ra-ui-materialui-list-columnsbutton--few-columns
- http://localhost:9010/?path=/story/ra-ui-materialui-list-columnsbutton--label-types
- Can also be tested in the simple demo (`make run`)

## Additional Checks


djhi (commented):
Looks good. However, there's no way to interract with neither the filter input nor the columns switches with the keyboard (mutiple tabs just close the popover, and arrow keys don't help either.

PR #10857, opened 2025-07-28
Make all mutations react to their declaration time options changes

djhi:
## Problem

The mutation hooks currently put the `mutationMode` and `params` they receive at _declaration time_ in a ref but never update them from their props.

## Solution

Add effects that sync the refs with the props.

## How To Test

- stories
- tests

## Additional Checks

### DIFF
--- a/packages/ra-core/src/dataProvider/useDelete.spec.tsx
+++ b/packages/ra-core/src/dataProvider/useDelete.spec.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
-import { screen, render, waitFor } from '@testing-library/react';
+import { screen, render, waitFor, fireEvent } from '@testing-library/react';
 import expect from 'expect';
+import { QueryClient, useMutationState } from '@tanstack/react-query';
 
 import { CoreAdminContext } from '../core';
 import { RaRecord } from '../types';
@@ -19,7 +20,7 @@ import {
     ErrorCase as ErrorCaseUndoable,
     SuccessCase as SuccessCaseUndoable,
 } from './useDelete.undoable.stories';
-import { QueryClient, useMutationState } from '@tanstack/react-query';
+import { MutationMode, Params } from './useDelete.stories';
 
 describe('useDelete', () => {
     it('returns a callback that can be used with deleteOne arguments', async () => {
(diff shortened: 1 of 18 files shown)


packages/ra-core/src/dataProvider/useDelete.spec.tsx
```diff
@@ -75,6 +76,79 @@ describe('useDelete', () => {
...
+
+    it('uses the latest declaration time params', async () => {
+        jest.spyOn(console, 'log').mockImplementation(() => {});
+        const posts = [
+            { id: 1, title: 'Hello' },
+            { id: 2, title: 'World' },
+        ];
+        const dataProvider = {
+            getList: (resource, params) => {
+                console.log('getList', resource, params);
```
fzaninotto: you should not leave the console.log here (same in delete and in other hook tests)

packages/ra-core/src/dataProvider/useDelete.stories.tsx
```diff
@@ -0,0 +1,168 @@
...
+export default { title: 'ra-core/dataProvider/useDelete' };
+
+export const MutationMode = () => {
+    const posts = [
+        { id: 1, title: 'Hello' },
+        { id: 2, title: 'World' },
+    ];
+    const dataProvider = {
+        getList: (resource, params) => {
+            console.log('getList', resource, params);
```
fzaninotto: I think you should remove these logs and the console.log mocking on the test. These may have been useful during development, but now the story is explicit enough (and testable enough) not to rely on logs. 
Besides, the test aren't reverting the mock on console.log.

packages/ra-core/src/dataProvider/useDelete.stories.tsx
```diff
@@ -0,0 +1,168 @@
...
+};
+
+export const Params = ({ dataProvider }: { dataProvider?: DataProvider }) => {
+    const posts = [
+        { id: 1, title: 'Hello' },
+        { id: 2, title: 'World' },
+    ];
+    const defaultDataProvider = {
+        getList: (resource, params) => {
+            console.log('getList', resource, params);
```
fzaninotto: same, console.log is useless in tests

fzaninotto (changes requested):
1% code, 99% stories and tests, we're good ;)

PR #10867, opened 2025-07-29
Export `ValidationMessageFuncParams` and `ValidationMessageFunc` types

djhi:
## Problem

Some users need those types but they are not exported so they had to duplicate them in their code.

## Solution

Export them with better names.

## Additional Checks

### DIFF
--- a/packages/ra-core/src/form/validation/validate.ts
+++ b/packages/ra-core/src/form/validation/validate.ts
@@ -36,16 +36,18 @@ function isValidationErrorMessageWithArgs(
     return error ? error.hasOwnProperty('message') : false;
 }
 
-interface MessageFuncParams {
+export interface ValidationMessageFuncParams {
     args: any;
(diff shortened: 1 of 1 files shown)


packages/ra-core/src/form/validation/validate.ts
```diff
@@ -36,16 +36,18 @@ function isValidationErrorMessageWithArgs(
...
 
-interface MessageFuncParams {
+export interface ValidationMessageFuncParams {
     args: any;
     value: any;
     values: any;
 }
 
-type MessageFunc = (params: MessageFuncParams) => ValidationErrorMessage;
+export type ValidationMessageFunc = (
```
fzaninotto: No a big fan of this name. I suggest `ValidationMessageFactory`, `GetValidationMessage`, or `GenerateValidationMessage`. Append `Params` for the second type.

PR #10870, opened 2025-07-30
Implement reordering columns using the keyboard in `<ColumnsButton>`

djhi:
## Problem

Although it's possible to open the <ColumnsButton> menu with space and to interact with its items using the keyboard, it’s impossible to drag and drop them with the keyboard.

## Solution

Implement reordering columns using the keyboard.

## How To Test

- [Story](https://react-admin-storybook-njkolog9u-marmelab.vercel.app/?path=/story/ra-ui-materialui-list-datatable--full-app)
- Click in the background around the list t ofocus the story iframe
- Press _Tab_ until the `<ColumnsButton>` is active
- Press  _Space_ to open its menu
- Press _Tab_ to loop through its items
- Press  _Space_ to start dragging a column and use arrow keys to move it. Drop with _Space_
- You should still have the item focused
- You should be able to _Tab_ to the switch inside each item
- Check that the columns were indeed reordered in the menu and the datatable
- Press _Escape_ to close the menu

## Additional Checks

### DIFF
--- a/packages/ra-ui-materialui/src/list/datatable/ColumnsButton.spec.tsx
+++ b/packages/ra-ui-materialui/src/list/datatable/ColumnsButton.spec.tsx
@@ -29,7 +29,7 @@ describe('ColumnsButton', () => {
             screen
                 .getByRole('menu')
                 .querySelectorAll('li:not(.columns-selector-actions)')
-        ).toHaveLength(8); // 7 columns + the filter input li
+        ).toHaveLength(7);
         // Typing a filter
         fireEvent.change(
             screen.getByPlaceholderText('ra.action.search_columns'),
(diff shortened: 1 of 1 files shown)


Madeorsk (changes requested):
UX feels a bit weird.

- We can now focus 2 different elements with tab: the list item, and the checkbox itself. I don't like that, it's way harder to reach the end of the list (twice as many keys to press), and it may cause confusions. What do you think about allowing to use the keys to move the columns while the checkbox is focused?
- If I start moving an item with space, then just cancel without moving (pressing space again), the item is moved up or down. It shouldn't.
- Why does the toggle and the drag indicator icon disappear when we enter "dragging mode"?
- If I leave the selector while still in move mode (by pressing escape or clicking away), it keeps its "dragging" state, maybe just cancel the move in this case. (I can get back to a state where I can move again the previously selected column again by focusing something in MenuList, but it's not obvious).

Madeorsk:
Also, I think you should remove e4989e0e4dbb6142737cc2d64db22da86a15d6d1, we don't want to remove this as it's now merged in master.

PR #10876, opened 2025-08-01
Introduce `useDeleteController`

djhi:
## Problem

We currently have a lot of duplicated logic in `useDeleteWithConfirmController` and `useDeleteWithUndoController`. Besides they handle things that should only belong in the UI hooks/components (`onClick`, dialog handlers and state).

## Solution

- Introduce `useDeleteController` that can be used for all mutation modes
- Use it in `useDeleteWithConfirmController` and `useDeleteWithUndoController` as we want to keep them for backward compatibility
- Use it in the `DeleteWithUndoButton` and `DeleteWithConfirmButton`
- Modify the `DeleteWithConfirmButton` so that it always closes the confirmation dialog before running user provided side effects

## How To Test

- Buttons should work as before
- Tests should succeed as before

## Additional Checks

### DIFF
--- a/packages/ra-core/src/controller/button/useDeleteController.tsx
+++ b/packages/ra-core/src/controller/button/useDeleteController.tsx
@@ -1,0 +1,184 @@
+import { useCallback } from 'react';
+import { UseMutationOptions } from '@tanstack/react-query';
+
+import { useDelete } from '../../dataProvider';
+import { useUnselect } from '../';
+import { useRedirect, RedirectionSideEffect } from '../../routing';
+import { useNotify } from '../../notification';
+import { RaRecord, MutationMode, DeleteParams } from '../../types';
+import { useResourceContext } from '../../core';
+import { useTranslate } from '../../i18n';
+
+/**
+ * Prepare a set of callbacks for a delete button guarded by confirmation dialog
+ *
+ * @example
+ *
+ * const DeleteButton = ({
+ *     resource,
+ *     record,
+ *     redirect,
+ *     onClick,
+ *     ...rest
+ * }) => {
+ *     const {
+ *         open,
+ *         isPending,
+ *         handleDialogOpen,
+ *         handleDialogClose,
+ *         handleDelete,
+ *     } = useDeleteWithConfirmController({
+ *         resource,
+ *         record,
+ *         redirect,
+ *         onClick,
+ *     });
+ *
+ *     return (
+ *         <Fragment>
+ *             <Button
+ *                 onClick={handleDialogOpen}
+ *                 label="ra.action.delete"
+ *                 {...rest}
+ *             >
+ *                 {icon}
+ *             </Button>
+ *             <Confirm
+ *                 isOpen={open}
+ *                 loading={isPending}
+ *                 title="ra.message.delete_title"
+ *                 content="ra.message.delete_content"
+ *                 titleTranslateOptions={{
+ *                     name: resource,
+ *                     id: record.id,
+ *                 }}
+ *                 contentTranslateOptions={{
+ *                     name: resource,
+ *                     id: record.id,
+ *                 }}
+ *                 onConfirm={handleDelete}
+ *                 onClose={handleDialogClose}
+ *             />
+ *         </Fragment>
+ *     );
+ * };
+ */
+export const useDeleteController = <
+    RecordType extends RaRecord = any,
+    ErrorType = Error,
+>(
+    props: UseDeleteControllerParams<RecordType, ErrorType>
+): UseDeleteControllerReturn => {
+    const { mutationMode } = props;
+    const {
+        record,
+        redirect: redirectTo = 'list',
+        mutationOptions = {},
+        successMessage,
+    } = props as UseDeleteControllerParams<RecordType, ErrorType>;
+    const { meta: mutationMeta, ...otherMutationOptions } = mutationOptions;
+    const resource = useResourceContext(props);
+    const notify = useNotify();
+    const unselect = useUnselect(resource);
+    const redirect = useRedirect();
+    const translate = useTranslate();
+
+    const [deleteOne, { isPending }] = useDelete<RecordType, ErrorType>(
+        resource,
+        undefined,
+        {
+            onSuccess: () => {
+                notify(
+                    successMessage ??
+                        `resources.${resource}.notifications.deleted`,
+                    {
+                        type: 'info',
+                        messageArgs: {
+                            smart_count: 1,
+                            _: translate('ra.notification.deleted', {
+                                smart_count: 1,
+                            }),
+                        },
+                        undoable: mutationMode === 'undoable',
+                    }
+                );
+                record && unselect([record.id]);
+                redirect(redirectTo, resource);
+            },
+            onError: error => {
+                notify(
+                    typeof error === 'string'
+                        ? error
+                        : (error as Error)?.message ||
+                              'ra.notification.http_error',
+                    {
+                        type: 'error',
+                        messageArgs: {
+                            _:
+                                typeof error === 'string'
+                                    ? error
+                                    : (error as Error)?.message
+                                      ? (error as Error).message
+                                      : undefined,
+                        },
+                    }
+                );
+            },
+        }
+    );
+
+    const handleDelete = useCallback(() => {
+        if (!record) {
+            throw new Error(
+                'The record cannot be deleted because no record has been passed'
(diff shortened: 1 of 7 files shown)


packages/ra-core/src/controller/button/useDeleteController.tsx
```diff
@@ -0,0 +1,184 @@
...
+>(
+    props: UseDeleteControllerParams<RecordType, ErrorType>
+): UseDeleteControllerReturn => {
+    const { mutationMode } = props;
+    const {
+        record,
+        redirect: redirectTo = 'list',
+        mutationOptions = {},
+        successMessage,
+    } = props as UseDeleteControllerParams<RecordType, ErrorType>;
```
slax57: nitpick: can be merged

slax57: nitpick: the type cast seems unnecessary

packages/ra-core/src/controller/button/useDeleteController.tsx
```diff
@@ -0,0 +1,184 @@
...
+ *     onClick,
+ *     ...rest
+ * }) => {
+ *     const {
+ *         open,
+ *         isPending,
+ *         handleDialogOpen,
+ *         handleDialogClose,
+ *         handleDelete,
+ *     } = useDeleteWithConfirmController({
```
slax57: this jsdoc does not seem to match this hook

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -54,27 +54,84 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
...
+                                    smart_count: 1,
+                                }),
+                            },
+                            undoable: mutationMode === 'undoable',
+                        }
+                    );
+                    record && unselect([record.id]);
+                    redirect(redirectTo, resource);
+                }
+            },
```
slax57, line 98: Not sure about the choice not to use `useDeleteWithConfirmController` here. It looks like you had to move much code from ra-core over to ra-ui-materialui, which seems to go the opposite way of the ongoing efforts to enhance the headless core of RA.

To me things like handling notifications, side-effects, record selection state and redirection all belong to ra-core...

slax57 (changes requested):

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
djhi: This is because we need to close the dialog at L77. We decided to do it like it with @fzaninotto because we don't want UI concerns to be handled in the base controller. As we also don't want to have default side effects in the base controller that would call user provided side effects, we must duplicate them here.

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -75,112 +73,85 @@ const useDeleteWithConfirmController = <
...
+                                _: translate('ra.notification.deleted', {
+                                    smart_count: 1,
+                                }),
+                            },
+                            undoable: mutationMode === 'undoable',
+                        }
+                    );
+                    record && unselect([record.id]);
+                    redirect(redirectTo, resource);
+                },
```
Madeorsk, line 114: It's sad to have this duplicated again. Maybe introduce a `defaultOnSuccess` / `defaultOnError` in `useDeleteController` so you can use them here?

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -75,112 +73,85 @@ const useDeleteWithConfirmController = <
...
-            record,
-            resource,
-        ]
-    );
+        }
+        controllerHandleDelete();
+        if (typeof onClick === 'function') {
+            onClick(event);
+        }
+    });
```
Madeorsk, line 162: This `handleDelete` override is exactly the same in `useDeleteWithUndoController`. What about doing it in `useDeleteController` directly? Or you want to keep `useDeleteController` as generic as possible so it can be used for something else than a clickable component?

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
```diff
@@ -54,27 +54,84 @@ export const DeleteWithConfirmButton = <RecordType extends RaRecord = any>(
...
+                                        ? error
+                                        : (error as Error)?.message
+                                          ? (error as Error).message
+                                          : undefined,
+                            },
+                        }
+                    );
+                }
+            },
+        },
```
Madeorsk, line 123: Using something like `defaultOnSuccess` / `defaultOnError` like my other suggestion could help to deduplicate this code.

packages/ra-ui-materialui/src/button/DeleteWithUndoButton.tsx
```diff
@@ -50,14 +48,21 @@ export const DeleteWithUndoButton = <RecordType extends RaRecord = any>(
...
         record,
         resource,
         redirect,
-        onClick,
+        mutationMode: 'undoable',
         mutationOptions,
         successMessage,
     });
+    const handleClick: ReactEventHandler<any> = event => {
+        handleDelete();
```
Madeorsk, line 61: You're not stopping propagation here, so `rowClick` is called in `DataTable`. This can cause UI flickering with a show / edit view opened on row click.

Madeorsk (changes requested):
Are `useDeleteWithConfirmController` and `useDeleteWithUndoController` deprecated? I'm not sure that it is a good choice as we can see that it forces UI implementation to duplicate parts of the logic (that will most probably be duplicated again in other UI kits). A good compromise could be to provide a few helpers / primitive functions with consistent side effects (like the propagation stop on click).

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -75,112 +73,85 @@ const useDeleteWithConfirmController = <
     props: UseDeleteWithConfirmControllerParams<RecordType, ErrorType>
 ): UseDeleteWithConfirmControllerReturn => {
```
fzaninotto, line 75: mark the hook as deprecated

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
djhi: As discussed, we'll keep the duplication

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
djhi: As discussed:
1. both `useDeleteWithConfirmController` and `useDeleteWithUndoController` are deprecated
2. we don't want to export default side effects to reuse them

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
djhi: We want to keep useDeleteController as generic as possible so it can be used for something else than a clickable component

packages/ra-ui-materialui/src/button/DeleteWithConfirmButton.tsx
djhi: As discussed, we'll keep the duplication

packages/ra-core/src/controller/button/useDeleteWithConfirmController.tsx
```diff
@@ -75,112 +74,85 @@ const useDeleteWithConfirmController = <
...
     const {
-        record,
-        redirect: redirectTo = 'list',
         mutationMode,
         onClick,
-        mutationOptions = {},
```
slax57, line 82: I see one possible regression here: with the new implementation, if I'm not mistaken, you'll override the full `mutationOptions` object if one is passed. So if, say, the user overrides only the `onError` side-effect, you will loose both the default `onError` and `onSuccess` implementations. Whereas before you would only override the default `onError` but keep the default `onSuccess`. If I'm not mistaken.

djhi: Good catch!

slax57 (approved):

Madeorsk (approved):

